PR #695, opened 2020-11-17
Add ssh-key kwarg to CLI app connect

ehofesmann:
Be able to call use the CLI to connect to a remote app instead of having to manually open the ssh connection if you need to use an ssh key instead of password login.
```
fiftyone app connect --destination <destination> --ssh-key </path/to/key>
```
Also updated environment, CLI, and remote session docs with this and best practices of adding the `IdentityFile` to your `ssh_config` if it is used often.


benjaminpkane (approved):
Nice. LGTM!

brimoor (approved):
LGTM too

Issue #693, opened 2020-11-17
[BUG] Applying a scalar filter in the App doesn't fully work


brimoor:
With prod `v0.6.5.2`, using `fiftyone quickstart`, I get weird behavior when trying to apply a uniqueness filter. When I drag the slider endpoints, the slider widget disappears. Also, although the label stats in the filter bar update, as do the number of the images in the image grid, the label graphs do not update.

<img width="1308" alt="Screen Shot 2020-11-17 at 1 11 46 PM" src="https://user-images.githubusercontent.com/25985824/99429824-82f2e180-28d6-11eb-9351-fe177b048bbd.png">

PR #696, opened 2020-11-20
Dataset dropdown

benjaminpkane:

Code is still being cleaned, but ready for feedback. I found the previous logo and title layout to be prohibitively small when trying to add selection functionality, so the dataset name now lies in the row as its own element.

I also found the arrow dropdown to be clunky. I can add it, but I don't think our users need that to understand that is where to change datasets.

* Adds dataset selection via a FiftyOne style input/dropdown
* Label graphs now include Fields Sidebar filters (the extended view). See #693 
* Refresh the App by clicking the Logo or `FiftyOne`
Screenshot from 2020-11-20 15-24-58

Locally.

        notes for FiftyOne users.

* Added the ability to change datasets via the dataset name input in the header
* Added the ability to refresh the App by clicking the logo in the top left
* Added the inclusion of Fields Sidebar filters when calculating distribution plots

-   [ ] Other

### DIFF
--- a/fiftyone/server/main.py
+++ b/fiftyone/server/main.py
@@ -30,6 +30,7 @@ import eta.core.video as etav
 os.environ["FIFTYONE_SERVER"] = "1"
 import fiftyone.core.aggregations as foa
 import fiftyone.constants as foc
+import fiftyone.core.dataset as fod
 import fiftyone.core.fields as fof
 import fiftyone.core.labels as fol
 import fiftyone.core.media as fom
@@ -388,6 +389,17 @@ class StateHandler(tornado.websocket.WebSocketHandler):
 
         StateHandler.state["selected_objects"] = selected_objects
         await self.send_updates(ignore=self)
+
+    async def on_set_dataset(self, dataset_name):
+        """Event for setting the current dataset by name.
+
+        Args:
+            dataset_name: the dataset name
+        """
+        StateHandler.state["dataset"] = fod.load_dataset(
+            dataset_name
+        )._serialize()
+        await self.on_update(StateHandler.state)
 
     async def on_get_video_data(self, _id, filepath):
         """Gets the frame labels for video samples.
--- a/fiftyone/server/main.py
+++ b/fiftyone/server/main.py
@@ -300,6 +300,15 @@ class StateHandler(tornado.websocket.WebSocketHandler):
         awaitables = self.get_statistics_awaitables(only=self)
         asyncio.gather(*awaitables)
 
+    async def on_refresh(self):
+        """Event for refreshing an App client."""
+        StateHandler.state = fos.StateDescription.from_dict(
+            StateHandler.state
+        ).serialize()
+        awaitables = [self.send_updates(only=self)]
+        awaitables += self.get_statistics_awaitables(only=self)
+        asyncio.gather(*awaitables)
+
     async def on_fiftyone(self):
         """Event for FiftyOne package version and user id requests."""
         self.write_message(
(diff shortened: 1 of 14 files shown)


fiftyone/server/main.py
```diff
@@ -477,14 +498,19 @@ def get_statistics_awaitables(self, only=None):
...
 
     @classmethod
-    async def send_updates(cls, ignore=None):
+    async def send_updates(cls, ignore=None, only=None):
         """Sends an update event to the all clients, exluding the ignore
         client, if it is not None.
 
         Args:
             ignore (None): a client to not send the update to
+            only (None): the onle client to send the update to
```
brimoor, line 507: typo `onle`

fiftyone/server/main.py
```diff
@@ -597,6 +623,10 @@ async def on_distributions(self, group):
         else:
             results = []
 
+        for stage_dict in state.filters.values():
```
brimoor, line 626: maybe add a factory method to `DatasetView` for constructing a view from a list of stage dicts? this pattern appears multiple places in the code now, I believe

brimoor (changes requested):
Tested and works! One edge case I checked is that, if a new dataset is created while the App is open, refreshing the App should add that new dataset to the dataset listing dropdown, which is True.

I have ~3 separate comments for your consideration below:

1. An edge case I considered is what happens when the App is in a non-trivial state and then the view is changed. I considered these cases:
- A view is loaded
- Some samples are selected
- A label filter is applied

I see that the state does not reset when a new dataset is loaded. W.r.t. views, this could be considered a feature; e.g., if you want to toggle between two datasets and consider what a given view looks like under each dataset. Conversely, UUIDs should never be the same b/w datasets, so persisting the state of `X samples selected` between datasets is probably never what the user wants.

For example, I selected 15 samples with one dataset, then used the `Only show selected` option, then changed datasets, and got in the nonsensical state pictured below:

Screen Shot 2020-11-23 at 10 20 21 AM


I'd say the best best is to always reset the state when a new dataset is selected. I worry that strange things could happen if state is attempted to be persisted between datasets.

For example, a label filter may or may not be valid when a new dataset is loaded (depending on whether a field of the same name exists), and it seems like too much work to check all of these cases to determine if the state is valid to persist, or not.

2. What do you think about removing the `FiftyOne` text? It doesn't add value and we already have our logo up there.

3. A (not specific to this PR) comment about our dropdown menus: it can sometimes be hard to tell where I'm at in the list: EG can I scroll up or down from my current position, and what % in the list is my current scroll location. Do you have any ideas on options that can be selected to improve this?

One idea might be adjusting the height of the dropdown to show `X + 0.5` elements, if there are `N` total elements and `N > X`. This way the partial occlusion will convey that scrolling is possible.

fiftyone/server/main.py
benjaminpkane: Good point, improving this is in progress in order to best resolve #693

benjaminpkane:
State now resets when a dataset is changed and `FiftyOne` has been removed. I'd like to address scrolling in a separate PR that introduces a scrollbar library.

In terms of the server code, I would like to formalize things when filtering is generalized. #699 changes some things, as well. What I've learned in #699 will inform a general approach to filtering fields.

benjaminpkane:
I think I slightly prefer the header with the `FiftyOne`. Could go either way, though.

Screenshot from 2020-11-25 11-13-26

brimoor:
@benjaminpkane I agree, looks better with `FiftyOne` back in there

brimoor (approved):
LGTM

Issue #697, opened 2020-11-21
[BUG] COCO Sample Parser Inheritance


ehofesmann:
Loading COCO from the zoo using torch raises an error:
```
Uncaught exception
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/lambertw/anaconda3/envs/fiftyone/lib/python3.7/site-packages/fiftyone/zoo/__init__.py", line 180, in load_zoo_dataset
    name, splits=splits, dataset_dir=dataset_dir
  File "/Users/lambertw/anaconda3/envs/fiftyone/lib/python3.7/site-packages/fiftyone/zoo/__init__.py", line 121, in download_zoo_dataset
    cleanup=cleanup,
  File "/Users/lambertw/anaconda3/envs/fiftyone/lib/python3.7/site-packages/fiftyone/zoo/__init__.py", line 805, in download_and_prepare
    ) = self._download_and_prepare(split_dir, scratch_dir, split)
  File "/Users/lambertw/anaconda3/envs/fiftyone/lib/python3.7/site-packages/fiftyone/zoo/torch.py", line 375, in _download_and_prepare
    sample_parser,
  File "/Users/lambertw/anaconda3/envs/fiftyone/lib/python3.7/site-packages/fiftyone/zoo/torch.py", line 516, in _download_and_prepare
    raise ValueError("Unsupported SampleParser %s" % type(sample_parser))
ValueError: Unsupported SampleParser <class 'fiftyone.utils.coco.COCODetectionSampleParser'>
```

This is likely an inheritance issue due to the [`COCODetectionSampleParser`](https://github.com/voxel51/fiftyone/blob/0415d1fce56bf610c34204358ac79d3cfcda0d2a/fiftyone/utils/coco.py#L38)  and the [`ImageLabelsSampleParser`](https://github.com/voxel51/fiftyone/blob/0415d1fce56bf610c34204358ac79d3cfcda0d2a/fiftyone/zoo/torch.py#L496) both inheriting from `LabeledImageTupleSampleParser`.

PR #698, opened 2020-11-21
Resolves dataset zoo type bug

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/697.

The issue was that `COCODetectionSampleParser`'s base class was changed away from `ImageDetectionSampleParser` in some recent work, since it can also support loading segmentations as `Polylines` instead. This PR adopts a safer practice of choosing the dataset format based on the `label_cls` of the labels in the dataset.


benjaminpkane (approved):
LGTM

PR #699, opened 2020-11-24
Improved range sliders and loading states

### LINKED ISSUE #693: [BUG] Applying a scalar filter in the App doesn't fully work

benjaminpkane:

Resolves #693 

What is lacking right now is that all new extended statistics need to be computed when a confidence slider is changed, which shows loading indicators for all fields. This does not need to happen, as only the counts for the label field in question will change. Then again, when an `only_matches` option (blocked by only matches support for frame labels) is added, all extended statistics do need to be recomputed...

I think more granular loading will be best addressed when the Fields Sidebar is generalized.

Locally

        notes for FiftyOne users.

TBD

-   [ ] Other

### DIFF
--- a/electron/app/components/CheckboxGrid.tsx
+++ b/electron/app/components/CheckboxGrid.tsx
@@ -19,10 +19,9 @@ import NumericFieldFilter from "./NumericFieldFilter";
 
 const GLOBAL_ATOMS = {
   colorByLabel: atoms.colorByLabel,
-  includeLabels: atoms.filterIncludeLabels,
-  invertInclude: atoms.filterInvertIncludeLabels,
-  includeNoConfidence: atoms.filterLabelIncludeNoConfidence,
-  confidenceRange: atoms.filterLabelConfidenceRange,
+  includeLabels: selectors.filterIncludeLabels,
+  includeNoConfidence: selectors.filterLabelIncludeNoConfidence,
+  confidenceRange: selectors.filterLabelConfidenceRange,
   confidenceBounds: selectors.labelConfidenceBounds,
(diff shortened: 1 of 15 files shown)


benjaminpkane:
Ready for review

brimoor (changes requested):
I tested as per #693 and the label graphs still don't seem to be updating when either a label or scalar filter is applied.

Also, I loaded the `quickstart` dataset, then applied a uniqueness range filter, and then tried to create a `SortBy(uniqueness, True)` stage in the view bar, and I get the following error:

```
  File "main.py", line 132, in wrapper
  File "main.py", line 289, in on_message
    await event(**message)
  File "main.py", line 337, in on_update
    awaitables += self.get_statistics_awaitables()
  File "main.py", line 467, in get_statistics_awaitables
    view + _make_filter_stages(state.dataset, state.filters),
TypeError: unsupported operand type(s) for +: 'DatasetView' and 'list'
```

brimoor:
Also, this may be unrelated to this PR, but if I load the `quickstart` dataset and then directly try to create a `SortBy(uniqueness, True)` view stage, the label stats are stuck in a loading spinner state

benjaminpkane:
OK, things are looking better now. Apologies.

brimoor (approved):
Everything works now. Awesome!

PR #700, opened 2020-11-25
Detection mistakenness

ehofesmann:
Created a recipe to go with detection mistakenness. 

Some `evaluate_detections` updates were needed, namely being able to choose whether to match detections `classwise` and whether to save sample-level fields as outputs of the evaluation.


brimoor (approved):
LGTM. Nice recipe!

I made a quick editing pass to over the Brain docs and recipe to stay consistent with the changes I made to the docstring of the actual `compute_uniqueness()` method. I think things are looking consistent now

PR #704, opened 2020-12-01
Clarifying image format support

brimoor:
Upon closer inspection, Chromium does not seem to support viewing TIFF images. This PR updates the FAQ question about this and links to a better article on the topic.


benjaminpkane (approved):
LGTM

PR #708, opened 2020-12-11
In-App feedback

benjaminpkane:

Adds an in-app feedback form that submits to hub spot.

The form content itself is minimal at the moment. Right now an email is optional, and the questions are simple and brief. Happy to expand on these and change the tone.

The app context is included in the form as well (notebook, desktop, browser, or colab). I can also include the uuid of the installation, but it might be a little invasive (i.e. it removes anonymity).

Screenshot from 2020-12-11 12-25-59

Locally.

        notes for FiftyOne users.

Added an in-app feedback form

-   [ ] Other


brimoor:
Tested and works! FYI- I updated the email address on the form to an email group that we're all on.

A couple notes on my experience using the form:
- Definitely make email address mandatory. If they're willing to fill out a form, we'd like to put them on a mailing list to keep them updated on FiftyOne
- Can you tweak the CSS so that placeholder text is lighter (maybe light gray)? When I'm in progress of filling it out, my submission blends in with the placeholder text and it takes a second to register what I still need to fill out
- I was able to submit the form multiple times by clicking on `Submitted` after I already submitted the form. Can you disable clicking when the form is submitted?
- (take it or leave it) Can you add a timer that closes the dialog after a few seconds? Give the user long enough to process the submitted state, but not long enough that they have to get ahold of their mouse and locate the X button

Suggestions for copy
- Prefer `t-shirt` to `T-shirt`
- How about this for the main content:

```
We are super dedicated to making FiftyOne as valuable as possible for our users. If you're willing to jump on a quick Zoom call with us to chat about your use cases in more detail, let us know by checking the box below. We'll *mail you a free t-shirt* for your trouble :)
```

As the text implies, what about adding a checkbox that says something like `[ ] I'm open to a Zoom call (and a free t-shirt!)`

This gives users a way to say "I'm onboard with talking more", or "I'll give you some text now, but leave me alone after that".

brimoor (changes requested):
See https://github.com/voxel51/fiftyone/pull/708#issuecomment-747786016.

brimoor:
@benjaminpkane one more thought on this- can you make a way to hide the free t-shirt button? When taking screenshots and such, we'd rather not have that showing, for example. Maybe a config setting/environment variable to hide advertising? Something like that

benjaminpkane:
> @benjaminpkane one more thought on this- can you make a way to hide the free t-shirt button? When taking screenshots and such, we'd rather not have that showing, for example. Maybe a config setting/environment variable to hide advertising? Something like that

Sure, no problem.

benjaminpkane:
Comments addressed. The form will hide and the t-shirt button will disappear after submission. You can force the form button to reappear by clearing local storage and reloading the page.

https://developers.google.com/web/tools/chrome-devtools/storage/localstorage

Screenshot from 2020-12-18 08-48-49

brimoor:
Tested in all environments and the local storage solution works great!

One last comment on design- how about deactivating the submit button until the required fields are filled in? Currently the button looks clickable and I get a hand icon on hover, but can't submit.

Do you think we need to provide a small "dismiss this box without submitting" icon in the form so that users can get rid of it without providing an email if they really don't want to hear from us? Little bit worried that the "want a free t-shirt?" is bordering on intrusive advertising if there's no way to dismiss it without providing personal information.

brimoor (approved):
Works great! Left a couple comments for your consideration before merging

PR #710, opened 2020-12-16
Release v0.7.0

### LINKED ISSUE #719: [BUG] Sample stats don't load when creating `Take()` stages from the App
On `develop`, I observe that creating a `Take(10)` stage from the App loads the view (10 random samples), but the sample stats in the Fields Sidebar are stuck in spinning-icon state. However, if I then do `session.view = session.view`, then everything loads appropriately.

So it seems randomized views can be loaded via `session.view` but not created via the App.

Screen Shot 2020-12-21 at 9 50 07 AM

benjaminpkane:
Resolves #719


docs/source/release-notes.rst
```diff
@@ -3,6 +3,44 @@ FiftyOne Release Notes
...
+
+FiftyOne 0.7.0
+--------------
+*Released December 21, 2020*
+
+App
+^^^
+- Added web browser support, which is now the default settting
+- Added :ref:`IPython notebook support <notebooks>`, e.g. Jupyter and Google
+  Colab
```
brimoor, line 16: @benjaminpkane FYI I removed the VS Code mention here; lmk if that was wrong to do

brimoor (approved):
LGTM

PR #717, opened 2020-12-18
Session tweaks

benjaminpkane:
* Adds `auto` and `height` kwargs to `Session()` and `launch_app`.
* Setting `session.view` or `session.dataset` (and `session.clear_view()`) shows the App when `auto` is `True` in a notebook.
* Moved `pip`, `setuptools`, and `wheel` upgrade steps to notes in installation docs

### DIFF
--- a/docs/source/environments/index.rst
+++ b/docs/source/environments/index.rst
@@ -137,7 +137,9 @@ To use FiftyOne in a notebook, simply install `fiftyone` via `pip`:
 
     !pip install fiftyone
 
-and load datasets and create sessions as usual:
+and load datasets as usual. When you run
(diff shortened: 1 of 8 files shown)


benjaminpkane:
The nested scrolling is a common notebook feature. It can easily be changed by user action:

https://stackoverflow.com/questions/36757301/disable-ipython-notebook-autoscrolling

brimoor (approved):
LGTM. I made a few additional tweaks:
- allowing `desktop=True` in Jupyter (but not Colab) environments
- providing an `open_tab()` method that allows the App to be opened in a dedicated browser tab from Jupyter (but not Colab) notebooks

I also expanded the environment docs a bit to explain the various possible workflows from notebooks.

PR #718, opened 2020-12-20
Safari fixes

### LINKED ISSUE #716: [BUG] App not quite working in Safari
This may be difficult for you to test, but opening `fiftyone quickstart` in Safari doesn't quite work for me on `develop`. The App loads, but the header at the top seizes up and shrinks a bit (as well as other components like the view stage bars), and the App is not responsive to scrolling. Also looks like the proper fonts aren't loading for me.

tl;dr: it's not quite usable in its current state for me.

My Safari version is Version 14.0.1 (14610.2.11.51.10).

(Works great on my Chrome and Firefox browsers though!)

Screen Shot 2020-12-18 at 1 13 43 PM

benjaminpkane:
Resolves #716


brimoor (approved):
Works now; thanks!

PR #726, opened 2020-12-23
Adding support for overwriting an existing dataset

brimoor:
```py
import fiftyone as fo

dataset = fo.Dataset("test")

dataset = fo.Dataset("test")
# ValueError: Dataset 'test' already exists; use `fiftyone.load_dataset()` to load an existing dataset

dataset = fo.Dataset("test", overwrite=True)
# no error
```


benjaminpkane (approved):
LGTM

PR #727, opened 2020-12-23
Making default App port configurable via fiftyone config

brimoor:
Allows the default App port to be customized via your FiftyOne config:

```shell
fiftyone quickstart
# ... App launched. Point your web browser to http://localhost:5151

export FIFTYONE_DEFAULT_APP_PORT=2222
fiftyone quickstart
# ... App launched. Point your web browser to http://localhost:2222
```

Also allows the App to be launched via the CLI with no dataset loaded:

```shell
# Can use the UI to select the dataset to open
fiftyone app launch
```

### DIFF
--- a/docs/source/cli/index.rst
+++ b/docs/source/cli/index.rst
@@ -658,14 +658,14 @@ Launch the FiftyOne App.
 
 .. code-block:: text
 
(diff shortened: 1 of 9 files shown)


benjaminpkane (approved):
LGTM :rocket:

PR #728, opened 2020-12-28
Efficient implementations of field cloning/clearing/renaming, dataset view cloning/saving, and sample merging

brimoor:
Changelog
- Renames `Dataset.clone_field()` to `Dataset.clone_sample_field()` for consistency with other operations
- Re-implements `Dataset.clone_sample_field()` via efficient DB-only operations
- Adds a `Dataset.clone_frame_field()` method for cloning frame-level fields of video datasets
- Adds `DatasetView.clone_sample_field()` and `DatasetView.clone_frame_field()` methods that allow for cloning views into sample fields (e.g., after filtering labels in a list field)
- Adds a `DatasetView.save()` method that allows for saving the current view (overwriting the rest of the dataset)
- Adds a `DatasetView.clone()` method that allows for cloning the current view into a new dataset
- Adds a `Dataset._merge_samples()` method that performs a subset of the functionality exposed by `Dataset.merge_samples()` via efficient DB-only operations. This method is called internally by `merge_samples()` whenever possible

Notes
- There are some TODO items in here related to cloning views into video datasets, which can't be implemented yet because we haven't fully finished support for filtering frame-level fields yet
- The `DatasetView.clone_sample_field()` and `DatasetView.clear_sample_field()` methods require a MongoDB 4.4 feature that allows the `$merge` aggregation to write to the same collection

Example usage:

```py
import fiftyone as fo

sample1 = fo.Sample(
    filepath="image1.jpg",
    predictions=fo.Detections(
        detections=[
            fo.Detection(label="high", confidence=0.9, field=1),
            fo.Detection(label="high", confidence=0.6, field=2),
            fo.Detection(label="low", confidence=0.3, field=3),
            fo.Detection(label="low", confidence=0.1, field=4),
        ]
    )
)

sample2 = fo.Sample(
    filepath="image2.jpg",
    predictions=fo.Detections(
        detections=[
            fo.Detection(label="high", confidence=1.0, field=1),
            fo.Detection(label="high", confidence=0.8, field=2),
            fo.Detection(label="low", confidence=0.2, field=3),
        ]
    )
)

dataset = fo.Dataset()

# Clone low confidence predictions into new field
low_conf_view = dataset.filter_detections("predictions", F("confidence") < 0.5)
low_conf_view.clone_sample_field("predictions", "low_conf")

# Clone high confidence predictions into new field
high_conf_view = dataset.filter_detections("predictions", F("confidence") > 0.5)
high_conf_view.clone_sample_field("predictions", "high_conf")

print(dataset.head())

# Clear a view's fields
dataset[1:].clear_sample_field("low_conf")
dataset[1:].clear_sample_field("high_conf")

print(dataset.head())

# Clone a view into a dataset as a new dataset
dataset2 = (
    high_conf_view  # `predictions` will only contain high confidence
    .exclude_fields(["low_conf", "high_conf"])  # exclude new fields
    .limit(1)       # only include 1 sample
    .clone()
)
print(dataset2.head())

# Save the current view, overwriting the dataset
high_conf_view.exclude_fields(["low_conf", "high_conf"]).limit(1).save()
print(dataset.head())
```


fiftyone/core/collections.py
```diff
@@ -2031,16 +1947,85 @@ def _add_view_stage(self, stage):
...
+                applicable to video datasets
+
+        Returns:
+            the aggregation pipeline
+        """
+        raise NotImplementedError("Subclass must implement _pipeline()")
+
     def _aggregate(
-        self, pipeline=None, hide_frames=False, squash_frames=False
+        self,
```
brimoor, line 2011: Apologies for the diff here; I didn't change the implementation, I just added some content to the docstrings and reorganized the methods

brimoor:
Okay this is finally ready for review.

I was mucking around with some core functionality related to sample field operations like renaming, clearing, and cloning, which could be a bit scary, but I added reasonably thorough unit tests to check my work.

brimoor:
The unit tests will stop failing when the test environment has its mongodb install upgraded to 4.4.2

fiftyone/migrations/revisions/admin/v0_7_1.py
```diff
@@ -8,12 +8,8 @@
...
-    db.admin.command({"getParameter": 1, "featureCompatibilityVersion": 1})
     db.admin.command({"setFeatureCompatibilityVersion": "4.4"})
-    db.admin.command({"getParameter": 1, "featureCompatibilityVersion": 1})
 
 
 def down(db):
-    db.admin.command({"getParameter": 1, "featureCompatibilityVersion": 1})
-    db.admin.command({"setFeatureCompatibilityVersion": "3.6"})
-    db.admin.command({"getParameter": 1, "featureCompatibilityVersion": 1})
+    db.admin.command({"setFeatureCompatibilityVersion": "4.2"})
```
brimoor, line 15: I think this should be 4.2, not 3.6, because fiftyone<0.7 was always using MongoDB 4.2. I don't recall specifics, but I think we may be using some aggregation features that were introduced in 4.2...

brimoor:
@benjaminpkane I'm not sure what's causing the `INTERNALERROR` in the tests under Python 3.6 (but not 3.8). The `tests/unittests/dataset_tests` tests pass locally for me, running Python 3.6.8 on mac. Do they pass locally for you as well?

I think I may have seen an `INTERNALERROR` like this locally at one point during development, and I think that was because a `self.assertRaises()` in the tests wasn't be finding the right exception type (IE there was actually a bug in the code that I fixed, and there was also an error in pytest that raised an error during normal execution of `assertRaises()`). But I can't see why that would explain the issue here.

Any ideas on how to debug this if its specific to the test environment? IDK how to get in there and test things...

benjaminpkane:
Sometimes I get the same error locally. Not sure what to do about the tests yet. Circling back to this as soon as screenshots is finalized.

benjaminpkane (approved):
LGTM. Great stuff here. Largely deferring to the updated unit tests here.

If you agree, testing on 3.7 and not 3.6 for the time being is fine with me.

brimoor:
> LGTM. Great stuff here. Largely deferring to the updated unit tests here.
> 
> If you agree, testing on 3.7 and not 3.6 for the time being is fine with me.

Hmm okay strange. Thanks for finding a fix!

I'll make an issue to investigate why that change was necessary...

PR #729, opened 2020-12-28
Adding note for working with TIFF images

brimoor:
Adds a note to the FAQ explaining how to work with TIFF images (and other image types for which browser extensions are available) in the FiftyOne App


benjaminpkane (approved):
LGTM

PR #731, opened 2020-12-30
Improved feedback

### LINKED ISSUE #723: [FR] Local storage not working for colab notebooks
Unsure if anything can easily be done about this, but, the `Want a free t-shirt?` button comes back every time I do `session.show()` in Colab notebooks. 

Conversely, in Jupyter notebooks, I observe that FiftyOne remembers whether I have dismissed the `Want a free t-shirt?` button (due to browser local storage, as in non-notebook browser environments, I believe).

Screen Shot 2020-12-22 at 12 02 38 PM

benjaminpkane:
* Moves feedback state/storage from browser local storage to disk.
* Adds a third state (minimized) for when the user has hidden the `Want a free t-shirt` button but has not submitted the form. A t-shirt icon is displayed in place of the text button in this case.

Resolves #723.

Screenshot from 2020-12-30 08-10-41


benjaminpkane:
I am unsure why local storage was not behaving correctly in colab, and debugging it was proving to be cumbersome. So I opted to move the state out of the browser.

brimoor (changes requested):
Icon looks great!

Couple things I noticed when testing:
- I don't seem to be able to submit the form (with all fields filled, the submit button is inactive)
- In a browser that already had "form submitted" in local storage, I observed that I had to manually clear browser local storage before the state of the t-shirt icon would be tied to `~/.fiftyone/var/feedback.json` (IE I wanted to check that I could delete this file and refresh my browser to see the default "want a free t-shirt?" state again) 

On the second point, maybe this was the intended behavior to avoid people that have already interacted with the form from seeing the icon again when they upgrade; but, if only for devs, the need to clear local storage was a bit unexpected

benjaminpkane:
Pardon the delay. Comments addressed.

brimoor (approved):
LGTM. Thanks for the updates!

PR #733, opened 2020-12-30
Analytics fixes

benjaminpkane:
This PR ensures new COLAB import events are recorded. Currently, when first importing in COLAB, the DB service would register an new NONE import event, and then a returning COLAB import event would register. In general, almost all COLAB import events will be new.

It also removes on redundant import events from services, in general.

* Restricts import events to the main process by skipping service processes
* Changes release candidate installs to be considered development installs with respect to GA tracking


brimoor (approved):
LGTM

PR #735, opened 2020-12-30
Exposing ability to control install/error level when using zoo models via the CLI

brimoor:
```shell
# Load a few samples from COCO-2017
fiftyone zoo datasets load coco-2017 --split validation --dataset-name test --max-samples 3 --shuffle

# Apply a zoo model to the dataset
# If any requirements of the model are not met, a warning is issued rather than raising an error
fiftyone zoo models apply --error-level 1 faster-rcnn-resnet50-fpn-coco-torch test predictions

# Verify the `predictions` exist
fiftyone datasets info test

# View some samples
fiftyone datasets stream test

# Cleanup
fiftyone datasets delete test
```


benjaminpkane (approved):
LGTM!

PR #736, opened 2021-01-01
Adding support for customizing database directory

brimoor:
Allows the user to customize where their database is stored on disk. This may be desirable when working with very large datasets on machines where you'd prefer your data to be stored on another volume.

Also, as depicted below, this could be used to create separate "workspaces" containing different sets of datasets.

```shell
fiftyone config database_dir
# ~/.fiftyone/var/lib/mongo         # the default location

export FIFTYONE_DATABASE_DIR=~/fiftyone/__db__

fiftyone datasets list
# No datasets found

fiftyone quickstart
# Loads dataset into `FIFTYONE_DATABASE_DIR`
```


benjaminpkane (approved):
LGTM!

PR #738, opened 2021-01-05
Fix empty view stats

### LINKED ISSUE #737: [BUG] Field stats don't load on empty views
On `develop`, field label stats for empty views are stuck in a loading state rather than showing 0.

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "cifar10",
    split="test",
    max_samples=50,
    shuffle=True,
)

# Field label stats don't load properly
session.view = dataset.match(F("ground_truth.label") == "non_existent")
session.view = dataset.limit(0)
```

Screen Shot 2021-01-04 at 7 36 41 PM

benjaminpkane:
Resolves #737

A `None` value must be checked for because of `SampleCollection`s `__bool__` definition.

### DIFF
--- a/fiftyone/server/main.py
+++ b/fiftyone/server/main.py
@@ -221,14 +221,16 @@ class PollingHandler(tornado.web.RequestHandler):
         elif event == "deactivate":
             self.write_message({"type": "deactivate"})
 
-        elif event == "statistics":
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM

PR #739, opened 2021-01-06
Adding dataset and model listing pages to the docs

brimoor:
Adds `dataset_zoo/datasets.html` and `model_zoo/models.html` pages to the User Guide that list all available datasets and models in the zoo, respectively.

The dataset listings include description, size, manual download instructions (if necessary), links to appropriate sources, and example Python/CLI code for loading the datasets.

The model zoo listings are automatically generated via the new `docs/scripts/make_model_zoo_docs.py` script, and include descriptions, model size, required packages, links to appropriate sources and papers, and example Python code for loading a zoo dataset and running inference with the model.

Other additions
- Adds an `imagenet-sample` dataset to the dataset zoo that contains 1 example of each ImageNet class from the validation split, to enable convenient demonstration of the many models in the zoo trained on ImageNet
- Implements `kitti` in the base zoo package (previously required the TF backend, which "invented" splits that the original author did not provide, which we don't want to copy)
- Implements `caltech101` in the base zoo package (previously required the TF backend, and that was actually broken because TFDS has a bug that prevents the dataset from being downloaded properly)
- Adds `caltech256` to the zoo (also base package)

**Dataset zoo**

Screen Shot 2021-01-06 at 11 35 55 AM

Screen Shot 2021-01-06 at 11 36 54 AM

**Model zoo**

Screen Shot 2021-01-06 at 11 35 16 AM

Screen Shot 2021-01-06 at 11 35 44 AM


benjaminpkane:
Automatically generated files are in this diff?

docs/scripts/make_model_zoo_docs.py
```diff
@@ -0,0 +1,225 @@
...
+
+**Requirements**
+
+{% if base_packages %}
+-   Packages: ``{{ base_packages }}``
+
+{% endif %}
+-   CPU
+
+    -   Support? {{ supports_cpu }}
```
ehofesmann, line 67: I love the automatically generated docs. The only thing that stands out is that the formatting of this seems a bit awkward. I'd change it to something like 
```
- CPU Support
    - yes
    - tensorflow>2
```
for example

ehofesmann (commented):
There are a lot of models and datasets and it's a bit overwhelming to just see the list of them. It would be cool to separate them into cards like the `tutorials` so that you can search by individual tags and see an example picture of the outputs on the right side of the tags. Then when you click on them it brings you here to the details.

docs/scripts/make_model_zoo_docs.py
brimoor: Okay I went with

```
- CPU support
    - yes
    - Packages: `tensorflow>2`
```

I think `Packages` should be there for consistency with how it looks when there are compute-independent package requirements:

```
- Packages: `torch, torchvision`
- CPU support
    - yes
```

I don't really like the `yes/no` business at all, but it is also being used for `Exposes embeddings? yes/no`, so we should probably revisit in general in the future.

brimoor: >  It would be cool to separate them into cards like the tutorials so that you can search by individual tags and see an example picture of the outputs on the right side of the card. Then when you click on them it brings you here to the details.

Definitely agree on this point, but didn't have the time to make it happen at the moment. It will require some modifications to the code that we're inheriting from the PyTorch docs for the tutorials cards.

I'll make an issue out of this one.

brimoor: https://github.com/voxel51/fiftyone/issues/742.

brimoor:
> Automatically generated files are in this diff?

Yeah currently the contents of `docs/source/user_guide/model_zoo/models.rst` are partially auto-generated.

I'll make an issue to fully automate this. In the initial implementation of this, the script needed the models to actually be downloaded in order to generate, and I didn't want that to be a requirement to build the docs. I got rid of that, so it should be pretty straightforward to fully automate this in a follow-up PR.

I'll make an issue for this.

brimoor:
> > Automatically generated files are in this diff?
> 
> Yeah currently the contents of `docs/source/user_guide/model_zoo/models.rst` are partially auto-generated.
> 
> I'll make an issue to fully automate this. In the initial implementation of this, the script needed the models to actually be downloaded in order to generate, and I didn't want that to be a requirement to build the docs. I got rid of that, so it should be pretty straightforward to fully automate this in a follow-up PR.
> 
> I'll make an issue for this.

https://github.com/voxel51/fiftyone/issues/743.

ehofesmann (approved):

Issue #713, opened 2020-12-17
[FR] Add TF2 models from `tensorflow/models` to the model zoo


brimoor:
The FiftyOne Model Zoo currently only contains TF1.X models from https://github.com/tensorflow/models. However, in July, `tensorflow/models` upgraded its zoo to support TF2 and contain a number of shiny new TF2 models.

We should upgrade the https://github.com/voxel51/models fork of https://github.com/tensorflow/models to support the new TF2 models, and add them to the zoo. This should be doable in such a way that both the new TF2 models and the existing TF1 models are still usable (with appropriate `tensorflow` reinstalls in between).

brimoor:
@ehofesmann this is something that would be great for you to take a look at

PR #740, opened 2021-01-06
TF2 model zoo

ehofesmann:
Closes https://github.com/voxel51/fiftyone/issues/713.

Updates the model zoo:
- Support running models from [tensorflow/models/research/object_detection](https://github.com/tensorflow/models/tree/master/research/object_detection) that previously required `tensorflow<2` to run with `tensorflow>=2` installed
- Adds CenterNet from the [TF2 detection model zoo](https://github.com/tensorflow/models/blob/archive/research/object_detection/g3doc/tf2_detection_zoo.md) to the FiftyOne model zoo

Example usages:

```shell
# Create TF2 environment
conda create -n fo
conda activate fo
conda install tensorflow-gpu==2.2.0

cd fiftyone
bash install.bash

fiftyone zoo models apply --error-level 1 centernet-hg104-512-coco-tf2 quickstart predictions1
fiftyone zoo models apply --error-level 1 faster-rcnn-resnet50-coco-tf quickstart predictions2

# Create TF1 environment
conda create -n fo1
conda activate fo1
conda install tensorflow-gpu~=1.15

cd fiftyone
bash install.bash

fiftyone zoo models apply --error-level 1 faster-rcnn-resnet50-coco-tf quickstart predictions
```

Tested by running all TF zoo models and verifying that they work under both TF1 and TF2 whenever their listed requirements promises that they should.


brimoor (changes requested):
@ehofesmann in https://github.com/voxel51/fiftyone/pull/739, I split the TF and Torch models into separate `manifest.json` files. Can you please merge `develop` in here and then migrate your edits into the appropriate files?

Sorry for the inconvenience.

brimoor:
@ehofesmann regarding the `tf_slim` package install, I think that's something that needs to be moved to https://github.com/voxel51/eta/blob/develop/eta/tensorflow/install_models.bash, since my understanding is that SLIM models in ETA won't work unless that command is run (IE, the change here will make FiftyOne work, but ETA won't work).

Here's an ETA snippet that you can use to verify that a TFSlim model in ETA works:

```py
import eta.core.image as etai
import eta.core.learning as etal

# must run `eta install models` first
model = etal.load_default_deployment_model("mobilenet-v2-imagenet")

with model:
    img = etai.read("/path/to/image.png")
    print(model.predict(img))
```

ehofesmann:
> @ehofesmann regarding the `tf_slim` package install, I think that's something that needs to be moved to https://github.com/voxel51/eta/blob/develop/eta/tensorflow/install_models.bash, since my understanding is that SLIM models in ETA won't work unless that command is run (IE, the change here will make FiftyOne work, but ETA won't work).
> 
> Here's an ETA snippet that you can use to verify that a TFSlim model in ETA works:
> 
> ```python
> import eta.core.image as etai
> import eta.core.learning as etal
> 
> # must run `eta install models` first
> model = etal.load_default_deployment_model("mobilenet-v2-imagenet")
> 
> with model:
>     img = etai.read("/path/to/image.png")
>     print(model.predict(img))
> ```

Ah interesting, that would install `tf_slim` even if you aren't using a `Slim` model though. Should we add a flag to `eta install models --include-slim` or is it not that big of a deal to install it with eta models?

brimoor:
> > @ehofesmann regarding the `tf_slim` package install, I think that's something that needs to be moved to https://github.com/voxel51/eta/blob/develop/eta/tensorflow/install_models.bash, since my understanding is that SLIM models in ETA won't work unless that command is run (IE, the change here will make FiftyOne work, but ETA won't work).
> > Here's an ETA snippet that you can use to verify that a TFSlim model in ETA works:
> > ```python
> > import eta.core.image as etai
> > import eta.core.learning as etal
> > 
> > # must run `eta install models` first
> > model = etal.load_default_deployment_model("mobilenet-v2-imagenet")
> > 
> > with model:
> >     img = etai.read("/path/to/image.png")
> >     print(model.predict(img))
> > ```
> 
> Ah interesting, that would install `tf_slim` even if you aren't using a `Slim` model though. Should we add a flag to `eta install models --include-slim` or is it not that big of a deal to install it with eta models?

Hopefully installing a pip package doesn't do too much right? The actual source files are already downloaded.

As written, the `eta install models` command needs to install everything required to use any model. Otherwise things like https://github.com/voxel51/eta/blob/eb61aa3fd0d6618d14347b57762566ed31c8fae5/eta/classifiers/tfslim_classifiers.py#L39 and https://github.com/voxel51/eta/blob/eb61aa3fd0d6618d14347b57762566ed31c8fae5/install.bash#L283 would need to change, too.

fiftyone/zoo/models/manifest-tf.json
```diff
@@ -1237,6 +1237,41 @@
             },
             "tags": ["detection", "coco", "tf1"],
             "date_added": "2020-12-11 13:45:51"
+        },
+        {
```
brimoor, line 1241: @ehofesmann looks like you lost the other package updates that you originally had here in the merge

ehofesmann: What other package updates do you mean?

brimoor: The diffs you see in this commit https://github.com/voxel51/fiftyone/pull/740/commits/46f5eb3fc9503d1248e1cff73aed60c84dc1f0b5 no longer appear now

ehofesmann: Gotcha

brimoor (approved):
LGTM. I tested everything on CPU-only, and it sounds like you did similar for GPU-enabled setups.

brimoor:
@ehofesmann FYI now that the requirements allow `tensorflow>=2` or `tensorflow-gpu>=2`, `error-level 1` is no longer required in most cases because `conda install tensorflow-gpu` will cause `pip` to show a `tensorflow>=2` to be installed

PR #741, opened 2021-01-07
Firefox tweaks

benjaminpkane:
Proposing we hide scrollbars in firefox until a proper JS library can be used. Things are fairly consistent beyond scrollbars.

Before:
Screenshot from 2021-01-06 19-34-01
After:
Screenshot from 2021-01-06 19-34-46


brimoor (approved):
LGTM! I tested on my setup in Firefox and Chrome and things look good.

I had noticed the spacing issue in the before-screenshot that you provided on the right edge of the fields sidebar, but didn't connect the dots that it was a scrollbar issue. Glad its resolved now! (agree that lack of scrollbar is fine for now on Firefox)

PR #747, opened 2021-01-08
Release v0.7.1

### LINKED ISSUE #748: [BUG] Applying field filters in App with labels graphs open does not automatically refresh graphs
- Run `fiftyone quickstart`
- Click show graphs
- Apply a label filter, eg apply a `uniqueness` filter as shown below that results in only one image in the view

The uniqueness graph does not automatically update- it still shows the graph for all samples. If I click on `Labels` and then back to `Scalars` tab, then the graph has updated as expected.

The ideal would be the currently visible graph, if any, will automatically update when a view or extended view change occurs.

Screen Shot 2021-01-08 at 9 50 18 AM

benjaminpkane:
v0.7.1 release notes. Also resolves #748.


brimoor (approved):
LGTM

PR #757, opened 2021-01-08
Fix int field parsing in App

### LINKED ISSUE #754: [BUG] Can't create Limit(n) stage for n >= 1000 in App
Trying to create a `Limit(n)` stage from the App with `n >= 1000` fails on develop. The stack trace is below.

In the view bar `Limit(1234)` is converted to `Limit(1,234)` which suggests that there's a string vs numeric issue going on.

`session.view = dataset.limit(1234)` works, however, and the view bar shows `Limit(1234)`.

```
Server Error

An exception has been raised by the server. Your session has been reverted to its previous state.

  File "main.py", line 182, in wrapper
  File "main.py", line 445, in on_message
    await event(self, **message)
  File "main.py", line 533, in on_page
    pipeline = view._pipeline(hide_frames=True, squash_frames=True)
  File "/Users/Brian/dev/fiftyone/fiftyone/core/view.py", line 399, in _pipeline
    _pipeline.extend(s.to_mongo(self))
  File "/Users/Brian/dev/fiftyone/fiftyone/core/stages.py", line 1229, in to_mongo
    if self._limit <= 0:
TypeError: '<=' not supported between instances of 'NoneType' and 'int'
```

benjaminpkane:
Resolves #754.

* Convert int field submission to number when parsing
* Remove commas when casting from string to int

### DIFF
--- a/app/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
+++ b/app/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
@@ -70,10 +70,13 @@ export const PARSER = {
     validate: (value) => /[0-9A-Fa-f]{24}/g.test(value),
   },
   int: {
-    castFrom: (value) => String(value),
-    castTo: (value) => +value,
+    castFrom: (value) => String(value).replace(/\B(?=(\d{3})+(?!\d))/g, ","),
+    castTo: (value) => +value.replace(/[,\s]/g, ""),
     parse: (value) =>
-      value.replace(/[,\s]/g, "").replace(/\B(?=(\d{3})+(?!\d))/g, ","),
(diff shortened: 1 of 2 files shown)


brimoor (changes requested):
Hmm, I pulled this branch, ran `bash install.bash` and am still seeing the error from https://github.com/voxel51/fiftyone/issues/754 when trying to create a `Limit(1234)` stage in the App...

brimoor:
> Hmm, I pulled this branch, ran `bash install.bash` and am still seeing the error from #754 when trying to create a `Limit(1234)` stage in the App...

Weird, tried again with `yarn install && yarn build-web` and now it works. User error 🤓

brimoor (approved):
LGTM

PR #758, opened 2021-01-08
Adding images of all zoo datasets to the docs

brimoor:
Another 30MB of images into `docs/`... but its nice to see what the dataset looks like before downloading it!

Screen Shot 2021-01-08 at 4 18 23 PM


ehofesmann (approved):
LGTM! 

Thoughts:

1) It would be cool to have gifs for video datasets but that should definitely wait until we stop storing images in git.

2) It could be useful to show a screenshot of the label distribution for each dataset. It would only show the most frequent classes without having to scroll but it could still be nice to see when growing the docs.

brimoor:
Thanks for the comments; can you convert to an issue for future work?

PR #759, opened 2021-01-08
Create selectable cards for the Model Zoo docs

ehofesmann:
Converts the gigantic model tables to a tag-searchable grid of model cards (like the tutorials/recipes pages):

### Future work

This is currently only for the model zoo because the dataset zoo does not have as many datasets or tags to search by. I want to update the pytorch sphinx theme to:
1) Make the cards smaller
2) Be able to scroll through them instead of pages

### New listing page

Screen Shot 2021-01-11 at 11 04 25 AM

image

image

### You have to click on different pages to navigate through models
image

### Multiple selections are `or` rather than `and` 
image


brimoor (changes requested):
Awesome that you got this working!

What do you think about moving this to `model_zoo/models.rst` to replace the existing HTML tables, rather than including on the `model_zoo/index.rst` page? This interactive listing is strictly an improvement over the tables over there.

You could delete the currently separate TF and Torch tables and use your single interactive listing instead.

Another benefit of that is that the generation could be fully automatic (note that `docs/generate_docs.bash` automatically generates the `model_zoo/models.rst` file from source using the `docs/scripts/make_model_zoo_docs.py` script. That was a late-breaking addition I snuck into the last release rather than the manual copy+paste that was originally in https://github.com/voxel51/fiftyone/pull/739.

The reason I made `model_zoo/models.rst` separate from `model_zoo/index.rst` is that I was concerned about the landing page being too dense with content. With that being said, it would be nice to have a more obvious visual indication on the landing page that compels the reader to click through to the actual models. I'll think about that.

ehofesmann:
> Awesome that you got this working!
> 
> What do you think about moving this to `model_zoo/models.rst` to replace the existing HTML tables, rather than including on the `model_zoo/index.rst` page? This interactive listing is strictly an improvement over the tables over there.
> 
> You could delete the currently separate TF and Torch tables and use your single interactive listing instead.
> 
> Another benefit of that is that the generation could be fully automatic (note that `docs/generate_docs.bash` automatically generates the `model_zoo/models.rst` file from source using the `docs/scripts/make_model_zoo_docs.py` script. That was a late-breaking addition I snuck into the last release rather than the manual copy+paste that was originally in #739.
> 
> The reason I made `model_zoo/models.rst` separate from `model_zoo/index.rst` is that I was concerned about the landing page being too dense with content. With that being said, it would be nice to have a more obvious visual indication on the landing page that compels the reader to click through to the actual models. I'll think about that.

Done, for now I've moved the cards over to `model_zoo/models.rst`. I also like having an easily navigatable model listing in `index.rst` but that should probably wait until I've reformatted the cards to be smaller/easier to click through.

brimoor (dismissed):
LGTM; thanks for migrating to `models.rst` for now.

I agree with your future work idea to decrease the size of the cards. They're a bit big for such a large list of models (would also be nice to increase the pagination so more results can be seen)

docs/source/custom_directives.py
```diff
@@ -39,16 +39,26 @@ def run(self):
...
         tags = self.options.get("tags", "")
 
-        card_rst = _CUSTOM_CARD_TEMPLATE.format(
-            header=header,
-            description=description,
-            link=link,
-            image=image,
-            tags=tags,
-        )
+        if image:
```
brimoor, line 45: There was a bug here that prevented images from appearing on the tutorial/recipe cards that *do* have images. It's fixed now

brimoor (approved):
LGTM now (fixed a bug)

Issue #760, opened 2021-01-08
[BUG] SortBy does not create indexes on embedded fields


benjaminpkane:
It looks like some validation logic when creating indexes has regressed some sorting functionality. Sorting no longer works on embedded fields.

```py
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("cifar100")
view = dataset.sort_by("ground_truth.label")
```

Error
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-0ff0c1e85a06> in <module>
----> 1 d.sort_by("ground_truth.label")

~/code/fiftyone/fiftyone/core/collections.py in sort_by(self, field_or_expr, reverse)
   1516             a :class:`fiftyone.core.view.DatasetView`
   1517         """
-> 1518         return self._add_view_stage(fos.SortBy(field_or_expr, reverse=reverse))
   1519 
   1520     @view_stage

~/code/fiftyone/fiftyone/core/dataset.py in _add_view_stage(self, stage)
   2253 
   2254     def _add_view_stage(self, stage):
-> 2255         return self.view().add_stage(stage)
   2256 
   2257     def _pipeline(

~/code/fiftyone/fiftyone/core/collections.py in add_stage(self, stage)
    532                 a valid stage for this collection
    533         """
--> 534         return self._add_view_stage(stage)
    535 
    536     @view_stage

~/code/fiftyone/fiftyone/core/view.py in _add_view_stage(self, stage)
    467 
    468     def _add_view_stage(self, stage):
--> 469         stage.validate(self)
    470 
    471         view = copy(self)

~/code/fiftyone/fiftyone/core/stages.py in validate(self, sample_collection)
   2069             # Create an index on the field, if necessary, to make sorting
   2070             # more efficient
-> 2071             sample_collection.create_index(field_or_expr)
   2072 
   2073 

~/code/fiftyone/fiftyone/core/view.py in create_index(self, field, unique)
    265             unique (False): whether to add a uniqueness constraint to the index
    266         """
--> 267         self._dataset.create_index(field, unique=unique)
    268 
    269     def drop_index(self, field):

~/code/fiftyone/fiftyone/core/dataset.py in create_index(self, field, unique)
   2112         """
   2113         if field not in self.get_field_schema():
-> 2114             raise ValueError("Dataset has no field '%s'" % field)
   2115 
   2116         index_info = self._sample_collection.index_information()

[...]

PR #761, opened 2021-01-08
Adding back support for indexing by embedded fields

brimoor:
Fixes https://github.com/voxel51/fiftyone/issues/760, and adds a unit test to ensure it stays this way.

```py
import fiftyone as fo
import fiftyone.zoo as foz

session.view = dataset.sort_by("ground_truth.label")  # now works again
```


benjaminpkane (approved):
Thanks! LGTM

PR #762, opened 2021-01-10
Documentation notebook updates

benjaminpkane:
Updates all tutorial and recipe notebooks (except `open_images_evaluation`, which is moved to https://github.com/voxel51/fiftyone-examples) with screenshots using the new automatic screenshotting functionality. Minor adjustments to cells were made, as well.


brimoor:
@benjaminpkane Thanks for doing this! I'll give the notebooks a read-through this evening.

I agree that the Open Images notebook can be removed from tutorials; how about just moving it to https://github.com/voxel51/fiftyone-examples?

benjaminpkane:
Sounds good. Done.

brimoor (approved):
LGTM. I also took the liberty of removing the images in `tutorials/images` and `recipes/images` that are no longer used (replaced by embedded notebook images).

I also removed the `PyTorch_CIFAR10/` folder that has been around forever but never converted to an actual tutorial that appears in the docs.

One thing (can be dealt with later): I noticed when viewing the `recipes/detection_mistakenness.ipynb` notebook using Jupyter that the App outputs contain an `Activate` button above the App image that isn't present in the other notebooks. I'm guessing that's what the display change you made in this PR is handling? and that notebook was the first one you ran or something like that?

benjaminpkane:
Right, thanks for addressing the details. I've removed the `Activate` button from detection mistakenness. And you're correct that my template change addresses the issue.

PR #766, opened 2021-01-11
Removing unused server log directory

brimoor:
I've been using the v0.7.1 feature that enables `database_dir` to be customized via my FiftyOne config, and noticed that there was one more bit of code that was creating a directory in the old hard-coded location of `~/.fiftyone/var`.

@benjaminpkane This `server.log` seems to be unused; can you confirm that this change is safe?

### DIFF
--- a/fiftyone/server/main.py
+++ b/fiftyone/server/main.py
@@ -9,7 +9,6 @@ import asyncio
 import argparse
 from collections import defaultdict
 from copy import deepcopy
-import json
 import os
 import posixpath
 import traceback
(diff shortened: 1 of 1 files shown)


fiftyone/server/main.py
```diff
@@ -324,8 +321,10 @@ def _get_label_object_ids(label):
     list_field_name = type(label).__name__.lower()
     if hasattr(label, "id"):
         return [label.id]
-    elif list_field_name in label:
+
```
brimoor, line 324: these other diffs are just responding to what `pylint` told me

benjaminpkane (approved):
LGTM. This is safe. I'll make an issue about reintroducing adding server logs properly, as it is something that should be available to users long-term.

PR #770, opened 2021-01-12
Adding utils to transform images and videos 

brimoor:
Adds `fiftyone.utils.image.transform_images()` and `fiftyone.utils.video.transform_videos()` utilities along with corresponding CLI commands under the `fiftyone utils` namespace.

Also re-implements `SampleCollection.compute_metadata()` to use multiple processes, which greatly increases the efficiency of computing metadata for large datasets.

The pattern used to implement the multiprocess computation here for metadata and image transformation is somewhat generic; it would be nice to support such operations more formally so that other parallelizable operations can be as fast as possible, too.

#### Example usage

```py
import fiftyone.utils.metadata as foum

# Metadata
dataset.compute_metadata()  # uses n = # CPUs processes (fast)
dataset.compute_metadata(overwrite=True, num_workers=1)  # single thread (current behavior)

# Transform images
foui.reencode_images(dataset, ext=".jpg")  # re-encode as JPG
foui.transform_images(dataset, max_size=(512, 512))  # limit to 512 x 512
foui.transform_images(dataset, max_size=(512, 512), force_reencode=True, num_workers=1)  # single thread (slower)

# Transform videos
fouv.reencode_videos(dataset)  # re-encode as H.264 MP4s
fouv.transform_videos(dataset, max_size=(512, 512))  # limit to 512 x 512 frames
```


fiftyone/core/cli.py
```diff
@@ -2247,6 +2509,24 @@ def __call__(self, parser, namespace, values, option_string=None):
         setattr(namespace, self.dest, kwargs)
 
 
+class _StoreSizeTupleDictAction(argparse.Action):
```
benjaminpkane, line 2512: Neat!

benjaminpkane (approved):
LGTM!

PR #771, opened 2021-01-12
Fix label mistakes tutorial

benjaminpkane:
[PyTorch_CIFAR10](https://github.com/huyvnphan/PyTorch_CIFAR10) was recently updated and breaks the label mistakes tutorial when the `master` branch is used..

Cloning the repository at tag `v2.1` resolves the issue.


brimoor (approved):
Nice fix!

PR #773, opened 2021-01-13
A couple small tweaks

brimoor:
A couple small tweaks here:
- Allow FiftyOne config path to be customized by setting the `FIFTYONE_CONFIG_PATH` environment variable
- If `max_samples` is provided to `foz.load_zoo_dataset()`, include this in the dataset's default name to avoid clashing with other copies of the dataset that may have been loaded

The latter change is a small convenience for those that often experience the workflow below

```py
import fiftyone.zoo as foz

# I've already done this
dataset.persistent = True
dataset.name  # "coco-2017-validation"

# THIS PR: Load just a few samples temporarily for prototyping, separately
# from the existing `coco-2017-validation` dataset
    "coco-2017",
    split="validation",
    max_samples=100,
)

# PREVIOUSLY: I had to type extra characters to achive this
    "coco-2017",
    split="validation",
    dataset_name="extra-characters-to-type",
    max_samples=100,
)
```

### DIFF
--- a/docs/source/cli/index.rst
+++ b/docs/source/cli/index.rst
@@ -161,7 +161,7 @@ Tools for working with your FiftyOne config.
 
 .. code-block:: shell
 
-    # Print the location of your config
+    # Print the location of your config on disk (if one exists)
     fiftyone config --locate
 
 .. _cli-fiftyone-constants:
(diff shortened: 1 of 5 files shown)


benjaminpkane (approved):
LGTM :rocket:

PR #775, opened 2021-01-14
Declaring float type of bounding box coordinates

brimoor:
Per our documentation, bounding box coordinates are floats in `[0, 1] x [0, 1]`, so we declare as such in the data model.


benjaminpkane (approved):
LGTM

PR #776, opened 2021-01-14
Quickstart notebook

benjaminpkane:
See https://github.com/voxel51/fiftyone-examples/pull/9 for the notebook the homepage Colab button will link to.

In addtion the homepage Colab link, I have added a `max-width` style to notebook screenshots so the image is never wider (or bigger) than the native width of the image. As an example, this improves how jupyter notebook screenshots are displayed when the notebook is opened in Colab.


benjaminpkane:
Screenshot from 2021-01-14 08-58-11

brimoor (approved):
LGTM

PR #777, opened 2021-01-14
Adds HistogramValues aggregation stage and exposes aggregations directly on SampleCollection

brimoor:
## Histograms
Adds a `HistogramValues` aggregation stage that allows for computing histograms of numeric fields. See below for example usage.

**Example usage**

```py
import numpy as np
import matplotlib.pyplot as plt

import fiftyone as fo
import fiftyone.zoo as foz


def plot_hist(counts, edges, cumulative=False, density=False):
    counts = np.asarray(counts)
    edges = np.asarray(edges)

    if cumulative:
        heights = np.insert(np.cumsum(counts), 0, 0.0)
        if density:
            heights = heights / heights[-1]

        plt.step(edges, heights, where="pre")
    else:
        left_edges = edges[:-1]
        widths = edges[1:] - edges[:-1]

        if density:
            heights = (counts / np.sum(counts)) / widths
        else:
            heights = counts

        plt.bar(left_edges, heights, width=widths, align="edge")


# Manual bins
#BINS = [0.0, 0.25, 0.50, 0.75, 1.01]
#RANGE = None

# Equally-spaced bins
BINS = 28
RANGE = [0, 1]

# Dynamic equal-count bins
#BINS = 28
#RANGE = None

# Use aggregation
r = dataset.histogram_values("uniqueness", bins=BINS, range=RANGE)

# Use numpy
values = [s.uniqueness for s in dataset.select_fields("uniqueness")]
np_counts, np_edges = np.histogram(values, bins=BINS, range=RANGE)

# Plot results
plt.subplot(121)
plot_hist(r.counts, r.edges)
plt.title("histogram_values()")

plt.subplot(122)
plot_hist(np_counts, np_edges)
plt.title("np.histogram()")

plt.show(block=False)
```

histogram_values

### Adding list fields
When adding new list fields to dataset, use the type of the first element of the list to set the field type. If the user wants a generic list field, they can declare it first: `dataset.add_sample_field("list_field", fo.ListField)`.

Previously, the situation was opposite: the user had to manually declare fields if they wanted them to be treated as homogenous: `dataset.add_sample_field("list_field", fo.ListField, subfield=fo.FloatField())`.

I think that homogeneous fields are the more common case, so this default behavior is preferable. Also, this enables aggregations like `count()` and `distinct()` to work in the examples below without having to manually declare field types.

### SampleCollection methods

Exposes all aggregations directly as methods on `SampleCollection`, for consistency with how `ViewStage`s are also exposed. This allows for cleaner syntax:

```py
# previous
dataset.match().aggregate(fo.Count("countable_field"))

# now
dataset.match(...).count("countable_field")
```

Multiple aggregations can list be run via `sample_collection.aggregate([agg1, agg2, ...])` for efficiency, if desired.

**Example usage**

```py
import fiftyone as fo

dataset = fo.Dataset()

dataset.add_sample_field("list_field", fo.ListField)  # generic list field
dataset.add_samples(
    [
        fo.Sample(
            filepath="/path/to/image1.png",
            tags=["A"],
            list_field=[1, 2.0],
            countable_field="cat",
            countable_list_field=[1, 2, 3],
            numeric_field=1.0,
            numeric_list_field=[1.0, 2.0, 3.0],
        ),
        fo.Sample(
            filepath="/path/to/image2.png",
            tags=["B"],
            list_field=[3.0, 4],
            countable_field="dog",
            countable_list_field=[1, 2, 3],
            numeric_field=4.0,
            numeric_list_field=[1.5, 2.5],

        ),
    ]
)

# Get tags
print(dataset.distinct("tags").values)

# bounds()
print(dataset.bounds("numeric_field"))
print(dataset.bounds("numeric_list_field"))
dataset.bounds("list_field")  # error: generic list field

# count()
print(dataset.count())
print(dataset.count("countable_list_field"))
print(dataset.count("numeric_list_field"))

# count_values()
print(dataset.count_values("countable_field"))
print(dataset.count_values("countable_list_field"))

# distinct()
print(dataset.distinct("countable_field"))
print(dataset.distinct("countable_list_field"))
```

### DIFF
--- a/docs/source/user_guide/using_datasets.rst
+++ b/docs/source/user_guide/using_datasets.rst
@@ -623,8 +623,8 @@ dataset splits or mark low quality images:
         ]
     )
 
-    print(dataset.get_tags())
-    # {"test", "low_quality", "train"}
+    print(dataset.distinct("tags").values)
+    # ["test", "low_quality", "train"]
 
 The `tags` field can be treated like a standard Python `list`:
 
--- a/docs/source/user_guide/using_views.rst
+++ b/docs/source/user_guide/using_views.rst
@@ -53,8 +53,8 @@ You can access specific information about a view in the natural ways:
     len(view)
     # 10000
 
-    view.get_tags()
-    # ['test']
+    view.media_type
+    # "image"
 
 .. note::
 
--- a/fiftyone/__public__.py
+++ b/fiftyone/__public__.py
@@ -19,6 +19,7 @@ from .core.aggregations import (
     CountValues,
     Distinct,
     DistinctLabels,
+    HistogramValues,
 )
 from .core.dataset import (
     Dataset,
--- a/fiftyone/__public__.py
+++ b/fiftyone/__public__.py
@@ -88,6 +88,32 @@ from .core.models import (
     ModelManager,
 )
 from .core.sample import Sample
+from .core.stages import (
+    Exclude,
+    ExcludeFields,
+    ExcludeObjects,
+    Exists,
+    FilterField,
+    FilterLabels,
+    FilterClassifications,
+    FilterDetections,
+    FilterPolylines,
+    FilterKeypoints,
+    Limit,
+    LimitLabels,
+    MapLabels,
+    Match,
+    MatchTag,
+    MatchTags,
+    Mongo,
+    Shuffle,
+    Select,
+    SelectFields,
+    SelectObjects,
+    Skip,
+    SortBy,
+    Take,
+)
 from .core.session import (
     close_app,
     launch_app,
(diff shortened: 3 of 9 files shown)


benjaminpkane (approved):
LGTM. Direct methods for common aggregations is certainly useful. And `HistogramValues` is great!.

One comment is that It would be nice to have a way chain `Aggregation`s similar to how views are created.

brimoor:
> One comment is that It would be nice to have a way chain `Aggregation`s similar to how views are created.

Hmm can you think of a common use case for chaining aggregations? Running multiple in parallel with `aggregate()` definitely makes sense, but not quite sure if there's enough richness in our aggregations to have "primitives" that might be further operated on.

brimoor:
One related thought I did have, though, is it would be nice to expose a `utils` module that contains methods for things like my `plot_hist()` code in the example above that can be used to generate plots of our `Aggregation`s.

Similarly, we could provide utils for generating confusion matrix plots, PR curves, etc. All with native support for FiftyOne data types as input.

PR #778, opened 2021-01-14
Adopt generic aggregation classes

brimoor:
### Background

As of https://github.com/voxel51/fiftyone/pull/777, there are five core aggregation **operations** that we expose, which can be applied to fields of certain **types**:

```
# Fields

Countable types: (Bool, Int, String)
- field
- list field
- ``embedded.field``
- ``embedded.field`` within Label-list type

Numeric types: (Float, Int)
- numeric field
- numeric list field
- ``embedded.field`` (dict field or embedded doc field)
- ``embedded.field`` within Label-list type
```

```
# Operations

Generic fields
- count()

Countable fields
- distinct()
- count_values()

Numeric fields
- bounds()
- histogram_values()
```

We currently handle specific operations like counting labels (`CountLabels`) or computing the bounds of a `confidence` (`ConfidenceBounds`) via dedicated `Aggregation` classes. However, as we continue to add more types, the number of permutations grows. For example, when I added `HistogramValues` to https://github.com/voxel51/fiftyone/pull/777, our current pattern would indicate that I need to add a dedicated `HistogramConfidenceValues` class to handle building histograms for label confidences.

But, what if the user wants to compute an aggregation over a custom field they've added to their `Label` instances, for example?

This PR experiments with an alternative strategy where the core operation types are allowed to be applied via `embedded.field.name` syntax.

**Pros**
- Deprecates the need for `CountLabels()` and `ConfidenceBounds()`
- Allows for aggregations of arbitrary dynamic fields and things like the `confidence` of Label types, with no additional classes needing to be added

**Neutral**
- Allows for unexpected new functionality to work, like `bounds()` will now work on strings (lexicographic order) and documents (using some definition of document comparison that MongoDB has defined...)
- MongoDB tends to silently ignore unsupported data in aggregation pipelines, so if FiftyOne aggregations are run on non-sensical data, a result of some kind is returned without the user being alerted that they've done something strange

**Cons**
- Syntax is a bit less clean: more of the functionality is captured by magic ``"embedded.field.name"``  strings
- Allows the user to aggregate over things that don't necessarily make sense, like calling `count_values()` on a float field
- Inconsistent with `ViewStage`s, which do expose dedicated classes like `FilterLabels` vs `Match` for performing operations specifically on `Label` fields

### Example usage

```py
import fiftyone as fo

dataset = fo.Dataset()

dataset.add_sample_field("list_field", fo.ListField)  # generic list field
dataset.add_samples(
    [
        fo.Sample(
            filepath="/path/to/image1.png",
            tags=["A"],
            list_field=[1, 2.0],
            countable_field="cat",
            countable_list_field=[1, 2, 3],
            numeric_field=1.0,
            numeric_list_field=[1.0, 2.0, 3.0],
            classification=fo.Classification(label="cat", confidence=None),
            detections=fo.Detections(
                detections=[
                    fo.Detection(label="fox", bounding_box=[0, 0, 1, 1], confidence=0.5),
                    fo.Detection(label="squirrel", bounding_box=[0.0, 0.0, 1.0, 1.0], confidence=0.7),
                ]
            ),
        ),
        fo.Sample(
            filepath="/path/to/image2.png",
            tags=["A", "B"],
            list_field=[3.0, 4],
            countable_field="dog",
            countable_list_field=[1, 2, 3],
            numeric_field=4.0,
            numeric_list_field=[1.5, 2.5],
            classification=fo.Classification(label="dog", confidence=0.9),
            detections=fo.Detections(
                detections=[
                    fo.Detection(label="cow", bounding_box=[0, 0, 1, 1], confidence=0.3),
                    fo.Detection(label="emu", bounding_box=[0.0, 0.0, 1.0, 1.0], confidence=None),
                ]
            ),
        ),
    ]
)

# bounds()
print(dataset.bounds("numeric_field"))
print(dataset.bounds("numeric_list_field"))
print(dataset.bounds("list_field"))
print(dataset.bounds("classification.confidence"))
print(dataset.bounds("detections.detections.confidence"))

# count()
print(dataset.count())
print(dataset.count("countable_field"))
print(dataset.count("countable_list_field"))
print(dataset.count("numeric_list_field"))
print(dataset.count("detections.detections"))

# count_values()
print(dataset.count_values("tags"))
print(dataset.count_values("countable_field"))
print(dataset.count_values("countable_list_field"))
print(dataset.count_values("numeric_field"))  # works but isn't meaningful
print(dataset.count_values("numeric_list_field"))  # works but isn't meaningful
print(dataset.count_values("classification.label"))
print(dataset.count_values("detections.detections.label"))

# distinct()
print(dataset.distinct("tags").values)
print(dataset.distinct("countable_field"))
print(dataset.distinct("countable_list_field"))
print(dataset.distinct("classification.label"))
print(dataset.distinct("detections.detections.label"))

# histogram_values()
print(dataset.histogram_values("numeric_field", bins=2))
print(dataset.histogram_values("numeric_list_field", bins=2))
print(dataset.histogram_values("classification.confidence", bins=2, range=(0, 1)))
print(dataset.histogram_values("detections.detections.confidence", bins=2, range=(0, 1)))
```


benjaminpkane:
I am not opposed to this, even as things stand right now. I think replacing MongoEngine with [MongoDB Shemas](https://docs.mongodb.com/manual/core/schema-validation/) would really empower this way of thinking.

My one note is that embedded list field syntax, e.g. `detections.detections.confidence` is something we will have to own outside of MongoDB. Keeping an up to date MongoDB JSON schema will support this if/when it is added. Until then, `arbitrary.list_field.number` will be hard support.

brimoor:
> My one note is that embedded list field syntax, e.g. `detections.detections.confidence` is something we will have to own outside of MongoDB. Keeping an up to date MongoDB JSON schema will support this if/when it is added. Until then, `arbitrary.list_field.number` will be hard support.

Yeah I totally agree. Right now checks for `list_field` and `detections.detections` are hard-coded so that `unwind` can be applied as needed.

I did note that [unwind can be applied to non-array fields](https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/#non-array-field-path), which means we could go nuclear and just split on `.` and unwind every part of the field, just in case. Although 

[MongoDB claims](https://docs.mongodb.com/manual/core/aggregation-pipeline-optimization) that pipeline optimization should add the appropriate `$project` calls if an aggregation only uses certain fields, but I haven't investigated to confirm is this happens for our aggregation stages, and lots of `unwind`ing definitely is scary.

benjaminpkane:
Working through some necessary changes, and then I'll approve.

benjaminpkane:
Ok, I have my made updates. All App statistics are now computed using aggregations. Video and image datasets are both working, but this could use a secondary tester.

If testing, the app will need to rebuilt and browser cache cleared:
``
cd ./app && yarn install && yarn build-web
``

benjaminpkane (approved):
LGTM

benjaminpkane:
Still a couple bugs, actually. Will follow up.

benjaminpkane:
Ok, ready now.

```py
import fiftyone as fo
import fiftyone.zoo as foz

quickstart, session = fo.quickstart()
# test App filtering and plots

cifar = foz.load_zoo_dataset("cifar10")
session.dataset = cifar
# test App filtering and plots

quickstart_video = foz.load_zoo_dataset("quickstart-video")
for idx, sample in enumerate(quickstart_video):
    if idx % 2 == 0:
        label = "even"
        confidence = 1.0
    else:
        label = "odd"
        confidence = None
    sample["classification"] = fo.Classification(label="one", confidence=confidence)
    sample.save()

session.dataset = quickstart_video
# test App filtering and plots
```

brimoor:
@benjaminpkane thanks for picking this PR up and finishing it!

Only one comment from me after testing: the scalar distribution graphs are now pretty spread out and hard to read. Can you recover something closer to the previous visual layout before we merge this?

Here's what I see with the `quickstart` dataset now:
Screen Shot 2021-01-19 at 10 39 17 AM

And how it used to be:
Screen Shot 2021-01-19 at 10 39 23 AM

brimoor:
To clarify, there's two things about the current behavior that are undesirable for numeric graphs:
- There's probably too many bins. It's better to be able to see the whole distribution without scrolling horizontally
- It doesn't make sense to have tick labels on every bin for numeric distributions. Just a few is sufficient

benjaminpkane:
Comments addressed, including issue #796. 25 bins are now used for numeric histograms.

It is curios why there is 1 `other` value for `uniqueness` in the quickstart.
Screenshot from 2021-01-19 10-11-24

brimoor:
@benjaminpkane thanks for the updates! I know why there is one `other`. Annoyingly, mongodb uses `[min, max)` intervals for every bin, including the last bin. The largest uniqueness is always exactly 1, so you have to use bounds `[0, 1 + eps)` to cover everything.

More generally, since I believe you may be computing `(min, max) = bounds()` with a query, you'll have to use `(min, max + eps)` for all histograms to avoid this issue.

PR #782, opened 2021-01-14
Convert walkthrough.md to notebook in fiftyone-examples

brimoor:
Removes the `WALKTHROUGH.md` from this repository, opting instead to link to the walkthrough notebook introduced in https://github.com/voxel51/fiftyone-examples/pull/10.

**FOR DISCUSSION:** which notebook should `Try it in Colab` on the docs homepage link to?
(a) [quickstart.ipynb](https://github.com/voxel51/fiftyone-examples/blob/master/examples/quickstart.ipynb)
(b) [walkthrough.ipynb](https://github.com/voxel51/fiftyone-examples/blob/walkthrough-notebook/examples/walkthrough.ipynb)


ehofesmann:
It's a tough choice, we shouldn't have the linked colab be too complex (which the walkthrough one might be), but it needs to hit enough points to make the reader interested.

At the moment it feels like the quickstart really just introduces HOW to use FiftyOne datasets/notebooks. There is a little bit about WHY you would want to use FiftyOne in the complex view section but it really only starts to tease it.

I think the walkthrough it great, but probably too detailed to be someone's first exposure to FiftyOne. If we add two sections about evaluating a model and computing mistakenness to the quickstart, then it should provide enough of a reason for WHY someone would want to use it without making it too long.

brimoor:
> It's a tough choice, we shouldn't have the linked colab be too complex (which the walkthrough one might be), but it needs to hit enough points to make the reader interested.
> 
> At the moment it feels like the quickstart really just introduces HOW to use FiftyOne datasets/notebooks. There is a little bit about WHY you would want to use FiftyOne in the complex view section but it really only starts to tease it.
> 
> I think the walkthrough it great, but probably too detailed to be someone's first exposure to FiftyOne. If we add two sections about evaluating a model and computing mistakenness to the quickstart, then it should provide enough of a reason for WHY someone would want to use it without making it too long.

I agree with your analysis here. Any dissension @benjaminpkane?

@ehofesmann I know Ben is busy wrapping up a blog post, can you take a shot at minimally expanding the quickstart in `fiftyone-examples` per your comments?

I'll change the `Try it in Colab` link back to the quickstart here.

benjaminpkane:
No dissension here.

I agree that the walkthrough is a bit long for the purpose of a quick demo. And a true "why" is certainly missing in the quickstart notebook.

benjaminpkane (approved):
LGTM

PR #786, opened 2021-01-15
Prevent blocking eval_js in Colab

benjaminpkane:
If javascript run by Colab's `eval_js()`  returns a promise, it blocks until it resolves. Sometimes these promises never seem to resolve, even though the javascript is executed.

Resolves issues related to state communication between the App and Colab.


benjaminpkane:
One moment, let me fix this diff.

PR #787, opened 2021-01-15
Patch Release: Prevent blocking eval_js in colab

### LINKED ISSUE #779: [BUG] session.show() does not preserve view state
In a Jupyter/Colab notebook, running the following two cells will result in different samples in each view, whereas one would like to see the exact same samples.

The issue is that the `Shuffle()` stage is being recreated, which generates a new random seed internally.

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("coco-2017", split="validation", max_samples=100)
view = dataset.shuffle().limit(10)

```

```
session.show()
```

benjaminpkane:
If javascript run by Colab's `eval_js()` returns a promise, it blocks until it resolves. Sometimes these promises never seem to resolve, even though the javascript is executed.

Resolves issues related to state communication between the App and Colab.

I will resolve the merge conflict after tagging a patch release (`v0.7.1.4`). I am concerned there are some bugs on `develop` at the moment.

I have tested the entire quickstart with these updates in Colab. Release candidate used:
```
!pip install --extra-index-url https://test.pypi.org/simple/ fiftyone==0.7.1.4rc3
```


brimoor (approved):
LGTM

PR #800, opened 2021-01-20
Improved plotting

### LINKED ISSUE #798: [FR] Enhance usability of numeric distribution graphs
I have some reactions after using the current state of numeric distribution graphs in the App (see screenshots below for a typical use case).

Note that I'm marking this as `needs-discussion` to avoid wasting too much effort in the current graphing library if we will soon be moving away from it. But, I'm creating this issue nonetheless to document my thoughts on making the graphs as usable as possible.

Enhancements for numeric graphs
- Clarify in the graph that "other" bucket is exclusively `None`-valued fields. The actual aggregation uses `other` as a catch-all for any data outside of the specified bin edges, but we're in control of the bin edges and know that anything outside must be `None` 
- Hide the `other` bin when its count is zero

Integer-field-specific enhancements:
- Suppose the range of the field is `[0, n]`. For reasonably small values of `n`, say `n <= DEFAULT_NUM_BINS`, it would be desirable for the bins to be `[-0.5, 0.5, 1.5, ..., n + 0.5]` with tick labels `0, 1, ..., n` rather than the strange floating point values observed below
- When the range `[0, n]` is too large, I would revert to floating point-type style where where are a fixed number of equal-width bins and a small number of tick labels like ~5 or so

Example graph for an int field defined on 15 out of many samples in a dataset
("other" means `None`-valued, but this is not readily apparent to users from the current display):

Screen Shot 2021-01-19 at 2 48 15 PM

The same graph after applying an `Exists(field, True)` view stage. The tick labels are undesirable because this is an int-valued field with a small range (`[0, 8]` in this case):

Screen Shot 2021-01-19 at 2 48 49 PM

benjaminpkane:
Resolves #798.

Changes:
* Empty counts are filtered from plots
* Integers are binned up to 100, and then floating points take over
* ~5 tick marks are used when binning in floating point values

I can decrease the max number of integer bins, of course. Really, I could make nice clean bins for all integer ranges...maxing out at say 25 bins.

Screenshot from 2021-01-20 14-29-37
Screenshot from 2021-01-20 14-29-31
Screenshot from 2021-01-20 14-29-24

### DIFF
--- a/app/src/components/Distributions.tsx
+++ b/app/src/components/Distributions.tsx
@@ -21,6 +21,36 @@ const Container = styled.div`
   padding-left: 1rem;
 `;
 
+const TooltipDiv = styled.div`
+  box-sizing: border-box;
+  border-radius: 3px;
+  background-color: ${({ theme }) => theme.backgroundDarker};
+  box-shadow: 0 2px 25px 0 ${({ theme }) => theme.darkShadow};
+  color: ${({ theme }) => theme.fontDark};
+  border-radius: 2px;
+  padding: 0.5rem;
+  line-height: 1rem;
+  margin-top: 2.5rem;
+  font-weight: bold;
+  width: auto;
+  z-index: 802;
+`;
+
+const TooltipHeader = styled.div`
+  color: ${({ theme }) => theme.font};
+  display: flex;
+  padding-bottom: 0.5rem;
+`;
+
+const PlotTooltip = ({ title, count }) => {
+  return (
+    <TooltipDiv>
+      <TooltipHeader>{title}</TooltipHeader>
+      Count: {count}
+    </TooltipDiv>
+  );
+};
+
 class CustomizedAxisTick extends PureComponent {
   render() {
     const { x, y, payload, fill } = this.props;
@@ -53,7 +83,7 @@ const Title = styled.div`
 `;
 
 const Distribution = ({ distribution }) => {
-  const { name, data, ticks } = distribution;
+  const { name, data, ticks, type } = distribution;
   const [ref, { height }] = useMeasure();
   const barWidth = 24;
   const container = useRef(null);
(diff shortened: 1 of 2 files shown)


brimoor (changes requested):
That was fast, sweet!

Can you undo the change that removes empty bins from *numeric* histograms. When the x-axis is numeric, its important that gaps in the distribution be reflected in the curve, since the assumption is that a numeric axis is "to scale". Otherwise, a bimodal distribution might magically turn into a unimodal distribution, for example.

I suppose that categorical histograms will always omit empty bins, since the App isn't wired up to use any class lists or anything like that. I think that behavior is good for now.

Note: I realized when https://github.com/voxel51/fiftyone/pull/778 was merged that we had, in fact, previously been removing empty bins from the distributions. I never noticed b/c I assumed the x axis bins were all there!

brimoor:
Regarding the case of int fields with many distinct values, it's a matter of taste, but I think that ~25 bins max is about right. Not sure about you, but the ~50 or so bins in the third screenshot you shared are a bit much.

If you want to invest the time in the current plotting framework, here are some finer-grained recommendations:

Definition: "small int field" means an int field with less than MAX_NUM_BINS distinct values. "large int field" is the other case

Tooltips:
1. Hovering over either a (a) floating point bin, or (b) bin of a large int field would show something like `[min, max)` to clarify what values are present in the bin, and whether the endpoints are open or closed (particularly relevant for int bins)
2. For large int fields, it would be more user-friendly if the bin edges snapped to the nearest integer (so that the tooltips from 1 would be more readable)
3. For small int fields, the tooltip would just show `n`, not `[n - 0.5, n + 0.5)`

x-axis ticks:
1. For small int fields and float fields, the ticks are great as you have them
2. For large int fields, the ~5 ticks should snap to integers

benjaminpkane:
I believe I address all comments.

Screenshot from 2021-01-21 15-21-36

Screenshot from 2021-01-21 15-21-44

Screenshot from 2021-01-21 15-21-49

brimoor (approved):
Awesome, so usable! Before merging, how about changing the tooltip for int fields from `0` to `Value: 0` for consistency with `Rnage: [min, max)`?

PR #807, opened 2021-01-21
Prevent Colab from caching requests in notebooks

benjaminpkane:
Turns cache control off for all server requests in Colab. This prevents the responses from being saved in the notebook json.

Without Colab's cache control turned off, Colab notebook jsons will become huge while using the App, quickly becoming unusable.

See [here](https://colab.research.google.com/notebooks/snippets/advanced_outputs.ipynb#scrollTo=R8ZvCXC5A0wT)


brimoor (approved):
Good find!

Issue #805, opened 2021-01-21
[FR] Launch the App in a remote Jupyter notebook


ehofesmann:
I want to process a model in a jupyter notebook on a remote machine so I use:
```
jupyter notebook --no-browser --port=XXXX
```
and connect to it on my local machine through `ssh`.

Trying to launch the App inside of this notebook results in the following:
![image](https://user-images.githubusercontent.com/21222883/105418003-72c6ed80-5c0a-11eb-8238-15ccfcec6d4a.png)

benjaminpkane:
This requires some manual effort at the moment. You have to setup port forwarding for `fiftyone`, as well. And the local port must match the remote port.

This should be simplified, of course.

ehofesmann:
Oh that's awesome! I figured it might be difficult to get it to work at all, but it's good to know that it's already possible.

brimoor:
Sounds like a good FAQ solution. No need to engineer anything

PR #808, opened 2021-01-22
[DOCS] Update FAQ, use App in remote notebook

ehofesmann:
Added a section to the FAQ on how to use the FiftyOne App in a notebook that is connected to a remote machine. Resolves https://github.com/voxel51/fiftyone/issues/805

Screen Shot 2021-01-25 at 8 26 05 PM

### DIFF
--- a/docs/source/faq/index.rst
+++ b/docs/source/faq/index.rst
@@ -54,9 +54,9 @@ information about running FiftyOne in notebooks.
 Can I use FiftyOne in a remote notebook?
 ----------------------------------------
 
-Yes! A common notebook workflow is to `launch jupyter on a remote system and
-connect to it from a local browser 
+Yes! A common notebook workflow is to `launch jupyter remotely
 <https://ljvmiranda921.github.io/notebook/2018/01/31/running-a-jupyter-notebook>`_.
+and connect to it from a local browser 
 
 If you want to use the FiftyOne App in this notebook, you will also need to
 port forward FiftyOne from the remote machine to your local machine with the
--- a/docs/source/faq/index.rst
+++ b/docs/source/faq/index.rst
@@ -54,27 +54,40 @@ information about running FiftyOne in notebooks.
 Can I use FiftyOne in a remote notebook?
 ----------------------------------------
 
-Yes! A common notebook workflow is to `launch jupyter remotely
-<https://ljvmiranda921.github.io/notebook/2018/01/31/running-a-jupyter-notebook>`_.
-and connect to it from a local browser 
+Yes! A common notebook workflow is to `launch Jupyter remotely
+<https://ljvmiranda921.github.io/notebook/2018/01/31/running-a-jupyter-notebook>`_
+and connect to it from a local browser. 
 
 If you want to use the FiftyOne App in this notebook, you will also need to
 port forward FiftyOne from the remote machine to your local machine with the
 following.
 
-**You need to use the same port for FiftyOne on your remote and local machine.**
+To launch Jupyter remotely:
+
+.. code:: shell
+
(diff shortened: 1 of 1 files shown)


docs/source/faq/index.rst
```diff
@@ -49,6 +49,33 @@ App will launch/update in the output of your notebook cells!
 Check out the :ref:`notebook environment guide <notebooks>` for more
 information about running FiftyOne in notebooks.
 
+.. _faq-remote-notebook-support:
+
+Can I use FiftyOne in a remote notebook?
+----------------------------------------
+
+Yes! A common notebook workflow is to `launch jupyter on a remote system and
```
brimoor, line 57: comment on link style: can you hyperlink a smaller portion of the text so it doesn't dominate visually when scrolling through the FAQ? I believe in equal opportunity for all links :)

docs/source/faq/index.rst
```diff
@@ -49,6 +49,33 @@ App will launch/update in the output of your notebook cells!
...
+
+If you want to use the FiftyOne App in this notebook, you will also need to
+port forward FiftyOne from the remote machine to your local machine with the
+following.
+
+**You need to use the same port for FiftyOne on your remote and local machine.**
+
+On your remote machine:
+
+.. code:: shell
```
brimoor, line 69: how about adding the `jupyter notebook` command here too, for completeness. To show the full copy pasta for running this workflow

docs/source/faq/index.rst
```diff
@@ -49,6 +49,33 @@ App will launch/update in the output of your notebook cells!
...
+
+Yes! A common notebook workflow is to `launch jupyter on a remote system and
+connect to it from a local browser 
+<https://ljvmiranda921.github.io/notebook/2018/01/31/running-a-jupyter-notebook>`_.
+
+If you want to use the FiftyOne App in this notebook, you will also need to
+port forward FiftyOne from the remote machine to your local machine with the
+following.
+
+**You need to use the same port for FiftyOne on your remote and local machine.**
```
brimoor, line 65: Don't you have to use port 5151 exactly?

If the user runs `fo.launch_app(dataset)`, this will default to port 5151 on the remote machine, which is then forwarded to the local machine in this case.

brimoor (changes requested):
A helpful addition to the FAQ for sure! I left a couple questions/suggestions for you

docs/source/faq/index.rst
ehofesmann: Good call, pasta served 🍝

docs/source/faq/index.rst
ehofesmann: You're right, I just tested it with a different port and it didn't work. Fixed

brimoor (approved):
I'm a bit confused why you included `fiftyone app launch` in your update to this PR. Isn't the idea to work with the App directly in the Jupyter notebook?

After you do the port forwarding for the App, it will appear as desired in the cell outputs themselves.

Note that the App port doesn't *have* to be 5151, the user can customize that if they want. I added a note to the FAQ answer.

Port forwarding is a bit of a braintwister for sure.

ehofesmann:
> I'm a bit confused why you included `fiftyone app launch` in your update to this PR. Isn't the idea to work with the App directly in the Jupyter notebook?
> 
> After you do the port forwarding for the App, it will appear as desired in the cell outputs themselves.
> 
> Note that the App port doesn't _have_ to be 5151, the user can customize that if they want. I added a note to the FAQ answer.
> 
> Port forwarding is a bit of a braintwister for sure.

Good call, `fiftyone app launch` isn't necessary. It will just result in some `channel 2: open failed: connect failed: Connection refused` until the app is launched in the notebook.

PR #809, opened 2021-01-22
Aggregation and view improvements

brimoor:
### Change log
- Adds a `sum()` aggregation that computes the sum of (non-``None``) field values
- Updates the `count()` aggregation to omit ``None``-valued fields
- Adds a `map_values()` view stage that enables the user to apply `map_labels()`-like mapping operations to any sample fields (not just labels) <--- **does not support unwinding array fields yet**
- Adds an `if_else()` expression that allows for constructing conditional if-else expressions for use in view stages (see examples for usage of this)
- Adds support for manually unwinding list fields via the `embedded.list_field[].with.another_list_field[].field` syntax when computing aggregations. Such unwinding was (and still is) automatically performed on top-level list fields (i.e., `list_field -> list_field[]`) and Label list fields (i.e., `ground_truth.detections.label --> ground_truth.detections[].label`). This new syntax allows users to aggregate on arbitrary embedded lists that they may have added to their samples
- Removes the `match_tag()` operation, which is redundant with the existing `match_tags("tag")` syntax
- Allows `ListField`s of samples to be `None` if not set on a particular sample (previously they defaulted to `[]`, which was inconsistent with how other fields behave). This change does *NOT* apply to the `tags` field, which still defaults to `[]`

### Documentation updates
- Adds copy-paste-able code examples to all `ViewStage` class docs and their corresponding `SampleCollection.@view_stage` methods
- Adds copy-paste-able examples for all `Aggregation` class docs and their corresponding `SampleCollection.@aggregation` methods
- Updates the aggregations user guide page to provide a thorough overview and examples of the various aggregation functionality, including advanced usage tips
- Adds documentation for `map_labels()` and `map_values()` to the dataset view user guide page

### Example

Here's an example usage of `map_values()` + `sum()`.

> I do acknowledge that an `average()` aggregation might be prudent... but let's see.

```py
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

# Number of samples
dataset.count()

# Number of samples with `uniqueness` values
dataset.count("uniqueness")

# Compute the average of uniqueness values > 0.6
view = dataset.map_values("uniqueness", (F() > 0.6).if_else(F(), None))
print(view.sum("uniqueness").sum / view.count("uniqueness").count)

# Check with numpy
print(
    np.mean(
        [
            s.uniqueness for s in view.select_fields("uniqueness")
            if s.uniqueness is not None
        ]
    )
)

# Another way to do this
view = dataset.match(F("uniqueness") > 0.6)
print(view.sum("uniqueness").sum / view.count("uniqueness").count)
```


fiftyone/core/stages.py
```diff
@@ -1718,15 +1906,172 @@ def validate(self, sample_collection):
...
+        #
+
+        stage = fo.MapValues("numeric_field", ROOT.max(0))
+        view = dataset.add_stage(stage)
+
+        #
+        # Clip all negative values of `numeric_list_field` to zero
+        #
+
+        stage = fo.MapValues("numeric_list_field", ROOT.map(ROOT.max(0)))
```
benjaminpkane, line 1951: :fire:

benjaminpkane (approved):
LGTM! I like the `list[]` syntax. Lots of stuff in here that will inform future work.

Issue #804, opened 2021-01-21
[BUG] Clear selected samples when option is chosen from menu


brimoor:
If I select some samples and then choose an option like clear/only show selected samples, the selected samples should reset. Currently, they do not, which causes unintuitive issues such as samples that are not visible being selected when you chose the exclude samples option.

> Side comment about the selection menu: like other menus, it should close when you click away from the menu. Currently it does not, and there's no close option, which means you *must* perform a state-changing operation, even if you don't want to

There's also a related issue (there may be an existing issue about this) where the following workflow results in unexpected behavior:
- Select `n` samples --- now `n` samples are selected
- Choose only show selected --- only viewing `n` samples; those `n` samples are still selected
- Choose clear selection --- still viewing `n` samples; no samples are selected
- Delete the view stage that was created --- now viewing all samples; `n` samples are magically reselected

Definitely samples should not be magically reselected in the last bullet above, but I also think a safe behavior (which would actually resolve the first point in this issue) is to always clear the selection whenever the view changes. Changing the view probably invalidates whatever you may have selected and leaving the samples as selected will likely cause more cognitive overhead for you to undo than it would take to just reselect your samples of interest.

<img width="1275" alt="Screen Shot 2021-01-21 at 10 20 15 AM" src="https://user-images.githubusercontent.com/25985824/105371309-7b033680-5bd2-11eb-8261-d55c46a2d4ea.png">

brimoor:
This is now resolved!

Issue #781, opened 2021-01-14
[FR] Allow App label colors to be set via Session object


brimoor:
Continuing from https://github.com/voxel51/fiftyone/issues/779#issuecomment-760395980, it would be convenient as a user who is particular about colors (e.g., colorblind), to be able to manually set label colors used in the App via their `session` object.

A related comment: I've never particularly liked the `Refresh field colors` option because it causes some people (asking for a... friend) to endlessly click refresh and get frustrated as to why they are having to waste time doing that. This latter comment has two distinct aspects:
- Can better colors be chosen by default? (some unfinished work on this exists in https://github.com/voxel51/fiftyone/pull/494)
- Let the user customize via session object (this issue)

brimoor:
As of FiftyOne v0.7.2, the color pool from which colors are chosen can be configured via `session.config.color_pool`.

An open question is whether we should allow control of individual field colors as well rather than just the color pool. Leaving this issue open in a `needs-discussion` state until we make a decision on the latter point.

joihn:
Any workaround to achieve this feature ?

brimoor:
The `color_pool` defines the pool of colors from which field colors are chosen. So the closest workaround is to set that to the list of acceptable colors for you:

```py
import fiftyone as fo
import fiftyone.zoo as foz

fo.app_config.color_pool = ["#ff0000", "#00ff00"]

dataset = foz.load_zoo_dataset("quickstart")

session = fo.launch_app(dataset)

# You can also modify it dynamically like this
session.config.color_pool = ["#ff0000", "#00ff00"]
session.refresh()
```

As a matter of implementation, I believe field colors are chosen sequentially in the order that they appear in `print(dataset)`. So in this case, `ground_truth` gets `"ff0000"` and `predictions` gets `#00ff00`. This may change though.

As the screenshot below shows, if you have more fields, you'll get duplicate colors:

<img width="1275" alt="Screen Shot 2022-01-18 at 12 28 19 PM" src="https://user-images.githubusercontent.com/25985824/149987951-b78fda58-d8b7-4245-8957-79b69d5f927f.png">

brimoor:
One idea to allow these colors to be customized would be to add an optional `field_colors` setting that contains a `field -> color` mapping dict that takes precedence over `color_pool` when `field` is int it.

We could also add a `value_colors` or `label_colors` setting that provides a similar mapping dict when rendering colors in the App's "Color by value" mode.

joihn:
[...]

Issue #780, opened 2021-01-14
[FR] Allow user to configure App settings via their Session object


brimoor:
Continuing from https://github.com/voxel51/fiftyone/issues/779#issuecomment-760395980, it would be convenient as a user who is particular about colors (e.g., colorblind), to be able to manually set label colors used in the App via their `session` object.

More generally, it would be nice to be able to support configuring options such as the player51 view options (e.g., show object confidences) via the `Session` object so that users could programmatically configure certain default behaviors that they'd like to see in the App.

brimoor:
As of FiftyOne v0.7.2, this is now supported! See https://voxel51.com/docs/fiftyone/user_guide/config.html#configuring-the-app.

PR #810, opened 2021-01-22
App enhancements

### LINKED ISSUE #799: [BUG] App should handle selected/excluded fields of views
View stages like `select_fields()` and `exclude_fields()` allow the user to exclude certain fields from a dataset view. When a field is excluded:
- Accessing the excluded field on a `SampleView` from the view causes an `AttributeError`
- Excluded fields to not appear in `view.get_field_schema()`

The App currently seems to always use the dataset's schema rather than the view's schema to render the fields sidebar. But, for consistency with how the backend works, I'd say that excluded fields should be omitted from the fields sidebar and other places like the label graphs.

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.select_fields()  # select only default fields

print(dataset)  # full schema
print(view)  # non-selected fields are omitted from schema

dataset.first().ground_truth  # works
view.first().ground_truth   # AttributeError

```

### Current behavior

```py
# Works, but the `predictions` and `uniqueness` fields should not appear
session.view = dataset.select_fields("ground_truth")
```

Screen Shot 2021-01-20 at 12 36 37 PM

### A current bug

Raises an error currently:

```py
session.view = dataset.select_fields()
```

Screen Shot 2021-01-20 at 12 37 09 PM

benjaminpkane:
Ready for review. Jupyter needs a bug to be fixed, though. Will resolve tomorrow.

Resolves #799 , #804, #781, #780

* Fields are now properly filtered when selected or exclude stages exist in the view. Includes video dataset support
* Selecting and only showing or hiding samples in the App now works as one would expect
* Color pool and player51 options `Show confidence` and `Show attributes` are now config and session options (needs cleanup)


brimoor (changes requested):
I'm still getting an error when trying `session.view = dataset.select_fields() ` from https://github.com/voxel51/fiftyone/issues/799 on this branch

Screen Shot 2021-01-25 at 8 39 38 PM

brimoor:
I'm also still seeing some weird selection-clearing behavior from https://github.com/voxel51/fiftyone/issues/804. For example, I did the following:
- Selected 3 samples
- Chose `Clear selected samples` from the App
- Created a `Take(10)` view stage

Now the App reverts to showing `3 samples selected`

fiftyone/core/session.py
```diff
@@ -204,6 +209,9 @@ class Session(foc.HasClient):
         terminate the session.
 
     Args:
+        color_pool (None): a list of browser supported color strings from
```
brimoor, line 212: In the future I think the number of App-config-related parameters will get a bit unwieldy. How about encapsulating `color_pool, height, show_attributes, show_confidence` in a new `AppConfig` class that can be passed to `Session`?

By default, `AppConfig` would inherit default values from `fo.config` for any `None`-valued properties. Example usage:

```py
config = fo.AppConfig()  # default values from `fo.config`
config.show_attributes = False  # customize

session = fo.launch_app(dataset, config=config)

session.config.show_attributes = True  # let's change it back
```

fiftyone/core/session.py
```diff
@@ -363,6 +402,42 @@ def url(self):
...
+        """
+        return self.state.color_pool
+
+    @color_pool.setter
+    @_update_state()
+    def color_pool(self, color_pool):
+        self.state.color_pool = color_pool
+
+    @property
+    def show_attrs(self):
```
brimoor, line 418: mismatch between `show_attrs` and `show_attributes` here

fiftyone/core/session.py
```diff
@@ -363,6 +402,42 @@ def url(self):
...
+        return self.state.show_attrs
+
+    @show_attrs.setter
+    @_update_state()
```
brimoor, line 422: I agree that not auto-showing is the right approach. Nor would I want the App to automatically refresh in-place, in case I'm modifying multiple settings.

But, how hard would it be to make `session.refresh()` reflect these settings? I tried calling `session.refresh()` after changes `session.show_confidence` and didn't see any changes. But I'm unsure if that's because `session.refresh()` is not a full refresh like cmd+R, or if this is a current limitation of refreshing when expanded sample view is open.

Basically I'm wondering if now is a good time to make refreshing the App respect whether the expanded sample view is open :)

fiftyone/core/config.py
```diff
@@ -89,6 +106,24 @@ def __init__(self, d):
...
+        self.default_app_color_pool = self.parse_string_array(
+            d,
+            "default_app_color_pool",
+            env_var="FIFTYONE_DEFAULT_APP_COLOR_POOL",
+            default=_COLOR_POOL,
+        )
+        self.default_app_show_confidence = self.parse_bool(
+            d,
+            "default_app_show_confidence",
+            env_var="FIFTYONE_DEFAULT_APP_SHOW_CONFIDENCE",
```
brimoor, line 118: w.r.t. the `AppConfig` proposal above, perhaps the App-config-related variables should namespaced into a `fo.config.app_config` variable?

But the `EnvConfig` class doesn't handle parameters that are objects, not sure what the cleanest way to proceed there is... 🤔

brimoor (changes requested):

fiftyone/core/config.py
benjaminpkane: I think a `default_app_config_path` will do. I'll write an `AppConfig` tomorrow AM.

benjaminpkane:
Comments addressed. Still need to update documentation.

`AppConfig` is an `EnvConfig`. A session's config can only be updated with `update_config()`, which will keep a modal open, etc.

Triggering updates on nested changes, e.g.  `session.config.show_attributes` would require a bit of an overhaul, hence `update_config()`. Keeping properties like `session.show_attributes` that redirect do the config didn't seem appropriate.

brimoor:
> Triggering updates on nested changes, e.g. session.config.show_attributes would require a bit of an overhaul, hence update_config(). Keeping properties like session.show_attributes that redirect do the config didn't seem appropriate.

I don't think we need any automatic refreshing behavior of the `Session`s when their configs are updated. It would be fine to require `session.refresh()` or another state-updating action like `session.view = None` in order to see the changes. Are we on the same page there?

Sorry for any confusion; what I was saying last night is that I like that you chose to *not* auto-show when app config settings are changed, and I also don't think the active App should be auto-refreshed either (EG if multiple settings are going to be changed). It's fine to opt-in to refreshing the session after settings have changed.

benjaminpkane:
Ok, yes the `auto_show=True` was unintentional. I went ahead and added a CLI command and App config section to the configuring fiftyone user guide. Hopefully it is not overkill.

brimoor (changes requested):
@benjaminpkane just re-tested and almost everything is perfect now!

There's only two things I noticed:

1. Changing the dataset in the App resets the `AppConfig`, but I would expect the config to stay the same since it is the same `Session` object:

```py
import fiftyone as fo

app_config = fo.AppConfig({})
app_config.show_attributes = False
app_config.show_confidence = False
app_config.color_pool = ["#ee6600"]

# Open App with no dataset selected
session = fo.launch_app(config=app_config)
print(session.config)

# [select a dataset in the App]

print(session.config)  # reverts to the default config
```

2. Updating `session.config.color_pool` and then calling `session.refresh()` does not apply the color changes. I would expect any config changes to be applied after calling `session.refresh()`

brimoor:
Also, where do you think `port` should go? It is an App setting, so it could be more consistent to move it to `AppConfig`. On the other hand, this is probably not something that should be changed on-the-fly.

Perhaps the rule is:
- `AppConfig`: things that can be changed on-the-fly
- Other `launch_app(...)` kwargs: things that shouldn't be changed

Or maybe everything goes in `AppConfig` and certain things can't be changed once the config is applied to a `Session`? Hmm.

Not sure anything needs to be changed now, this just came to mind when reviewing here.

benjaminpkane:
> * `AppConfig`: things that can be changed on-the-fly
> * Other `launch_app(...)` kwargs: things that shouldn't be changed on an active session

Yes, those are the rules I went with. There are a few ways to let this evolve. Generally, right now `AppConfig` controls the UX of the App, not things more related to session configuration. This could evolve/change though.

> Or maybe everything goes in `AppConfig` and certain things can't be changed once the config is applied to a `Session`?

This is possible. I left `port`, etc. as kwargs mostly to minimize the scope of this PR (backward compatibility, etc.). Considering the above, it would turn `AppConfig` into more a `SessionConfig`. Which I have no problem with, just commenting.

brimoor:
> > * `AppConfig`: things that can be changed on-the-fly
> > * Other `launch_app(...)` kwargs: things that shouldn't be changed on an active session
> 
> Yes, those are the rules I went with. There are a few ways to let this evolve. Generally, right now `AppConfig` controls the UX of the App, not things more related to session configuration. This could evolve/change though.
> 
> > Or maybe everything goes in `AppConfig` and certain things can't be changed once the config is applied to a `Session`?
> 
> This is possible. I left `port`, etc. as kwargs mostly to minimize the scope of this PR (backward compatibility, etc.). Considering the above, it would turn `AppConfig` into more of a `SessionConfig`. Which I have no problem with, just commenting.

👍 noted, agree, and I like how it is now for this release

brimoor:
FYI I pushed some small changes that make the App config's location configurable in the same way as the main FiftyOne config itself. Very similar to how you have it.

Thanks for adding the App config info to the docs. You're a gentleman and a scholar!

I really love how configurable this PR makes the App. I think it fits nicely with the design of the rest of the system.

benjaminpkane:
> 2. Updating `session.config.color_pool` and then calling `session.refresh()` does not apply the color changes. I would expect any config changes to be applied after calling `session.refresh()`

I'm not sure I can reproduce this. I made a change that allows for pending config changes to not be overwritten by state changes from App interactions, though.

brimoor:
Hmm perhaps I should explain in more detail. I set my config to this:

```py
{
    "color_pool": [
        "#ee6600"
    ],
    "notebook_height": 800,
    "show_confidence": false,
    "show_attributes": true
}
```

expecting to see that all fields have that single color in the App. But they do not seem to obey `session.config`.

brimoor:
Also I'm seeing a weird behavior where `session.refresh()` reverts my config change rather than applying it:

```py
In [24]: session.config                                                                                                                                                              
Out[24]: 
{
    "color_pool": [
        "#ee6600"
    ],
    "notebook_height": 800,
    "show_confidence": false,
    "show_attributes": true
}

In [25]: session.config.show_confidence = True                                                                                                                                       

In [26]: session.config                                                                                                                                                              
Out[26]: 
{
    "color_pool": [
        "#ee6600"
    ],
    "notebook_height": 800,
    "show_confidence": true,
    "show_attributes": true
}

In [27]: session.refresh()                                                                                                                                                           

In [28]: session.config                                                                                                                                                              
Out[28]: 
{
    "color_pool": [
        "#ee6600"
    ],
    "notebook_height": 800,
    "show_confidence": false,
    "show_attributes": true
}
```

benjaminpkane:
> Hmm perhaps I should explain in more detail. I set my config to this:
> 
> ```python
> {
>     "color_pool": [
>         "#ee6600"
>     ],
>     "notebook_height": 800,
>     "show_confidence": false,
>     "show_attributes": true
> }
> ```
> 
> expecting to see that all fields have that single color in the App. But they do not seem to obey `session.config`.

Previously (and currently) the colors were "popped" from the pool until the pool was empty. And then random colors were generated by the App for any fields that were left.

Changing this to always "cycle" through the color pool now. Noting that random colors will never be used.

brimoor:
Oooh I see now. I thought I tried using `select_fields("ground_truth")` to get down to a single field (which would then match the single color I picked), but for some reason I thought that color didn't match up.

I like the cycling strategy that you're adopting now, because it gives the user deterministic control over which colors will actually be used.

benjaminpkane:
`session.refresh()` behavior is fixed. I didn't realize environment variables take precedence over what is provided to a `from_dict` call for `EnvConfigs`. Also fixed a bug in `fiftyone.core.config._set_settings()`.

brimoor:
Sorry for all the iterations on this PR... I feel like we're converging on stable behavior here, so I'll go ahead and approve to give you an escape route, but I noticed a couple things with the latest version:

1. It looks like `fo.app_config` is now tied to `session.config` in some unexpected ways. I think I can fix what's going on here, so I won't elaborate just yet.

2. The `Color by label` option in the App doesn't respect the `color_pool`. Also, unrelated to this PR, but still inconsistent with the UI imo, `Color by label` ignores the shuffle colors option.

3. You may want to throw in some error handling that will complain if `color_pool = []` or is otherwise bad data. Some weird things started happening to my App when I tried an empty list there

brimoor:
> 1. It looks like `fo.app_config` is now tied to `session.config` in some unexpected ways. I think I can fix what's going on here, so I won't elaborate just yet.

I added a couple missing `.copy()` calls to `fo.app_config` usages in `fiftyone/core/state.py` in order to avoid `session.config` changes unexpectedly having an effect on `fo.app_config`, which should never change unless the user directly changes it.

It's definitely a bit of a brain teaser handling the bidirectional communication b/w the app and the session. For example, I was a bit surprised to find that:

```py
import fiftyone as fo

app_config = fo.AppConfig()
session = fo.launch_app(config=app_config)

session.config is app_config  # False
```

when there is no explicit copy of user-provided configs in `Session.__init__()`, but then I realized that's because the session is getting deserialized from the App every now and then...

You may want to explicitly document to the user that passing a `config=` is a pass-by-value operation.

brimoor:
Okay one more comment: it would be nice if this were allowed:

```py
session = fo.launch_app()

# set the entire `config` in one go
session.config = fo.AppConfig()
```

brimoor (approved):

Issue #811, opened 2021-01-22
[FR] Consider removing `AggregationResult` classes


brimoor:
As I was working on https://github.com/voxel51/fiftyone/pull/809, I wondered if the `AggregationResult` class should be deprecated in favor of just returning the aggregation result itself. If there are multiple fields, as in the (only, currently) case of `HistogramValues`, they would be returned as a dict:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")

# current

avg = dataset.sum("uniqueness").sum / view.count("uniqueness").count

result = dataset.histogram_values("uniqueness", bins=50, range=(0, 1))
result.edges
result.counts

# proposed

avg = dataset.sum("uniqueness") / view.count("uniqueness")

result = dataset.histogram_values("uniqueness", bins=50, range=(0, 1))
result["edges"]
result["counts"]
```

The primary function of the `AggregationResult` is to give the results a `name` field, which I'm guessing came about because aggregations were originally *always* computed via `results = dataset.aggregate([agg1, agg2, agg3, ...])`. But, even then, the order of `results` is preserved between input and output, so it seems that the user should always know what aggregations they ran.

For Python users of aggregations, it seems that the more concise syntax may be appreciated.

benjaminpkane:
This is probably best. `AggregationResult`s were part of the initial Aggregations implementation to encapsulate the needs of the App server, i.e. identifying a result by `(type, field_path)`. This can be handled without `AggregationResult`s, though.

Issue #813, opened 2021-01-25
[FR] Add support for list field syntax in stages


benjaminpkane:
I was considering how one could answer questions like "What is the average bounding box area for bounding boxes with an area less than `0.2`?" purely with stages and aggregations. This would require `[]` syntax for stages like `MapValues` and perhaps a `MapValues` parameter that places the result in a new field.

Proposed API below:
```py
import fiftyone as fo
import fiftyone.zoo as foz

F = fo.ViewField
root = fo.root_field

area = F("bounding_box")[2] * F("bounding_box")[3]
dataset = foz.load_zoo_dataset("quickstart")

small_bboxes = dataset.filter_labels("predictions", area < 0.2)
# or dataset.filter("predictions.detections[]", area < 0.2)

areas = small_bboxes.map_values("predictions.detections[].bounding_box", root[2] * root[3], out="area")

avg = areas.sum("predictions.detections[].area") / areas.count("predictions.detections[]")
```

brimoor:
Yes indeed, I had the same thought over the weekend and have done some work on this. Hold this thought for a second :)

brimoor:
As of https://github.com/voxel51/fiftyone/pull/815, this is now possible via the following syntax:

```py
import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone import ViewField as F

dataset = foz.load_zoo_dataset("quickstart")

bbox_area = F("bounding_box")[2] * F("bounding_box")[3]

small_bbox_areas = (
    dataset
    .filter_labels("predictions", bbox_area < 0.2)
    .set_field("predictions.detections.area", bbox_area)
)

avg_small_bbox_area = (
    small_bbox_areas.sum("predictions.detections.area")
    / small_bbox_areas.count("predictions.detections.area")
)
```

brimoor:
Also, as of https://github.com/voxel51/fiftyone/pull/834, this computation is possible via the cleaner syntax:

```py
import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone import ViewField as F

dataset = foz.load_zoo_dataset("quickstart")

bbox_area = F("bounding_box")[2] * F("bounding_box")[3]

print(
    dataset
    .filter_labels("predictions", bbox_area < 0.2)
    .mean("predictions.detections[]", bbox_area)
)
```

PR #815, opened 2021-01-25
More view and aggregation updates

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/811 and https://github.com/voxel51/fiftyone/issues/813.

Changelog
- Replaces the `MapValues` view stage with a `SetField` stage that allows for setting new fields on a view using an arbitrary expression to define the value. Supports both sample-level and frame-level fields
- Removes `AggregationResult` class in favor of returning the raw aggregation results directly (cf. discussion in https://github.com/voxel51/fiftyone/issues/811)
- Adds support for the `[]` syntax (and automatic unwinding of common fields) to `SetField`, which supports processing of array fields and embedded array fields in the expected ways
- Adds a number of new `ViewExpression` methods exposing various string-related operations, some control flow things like case/switch statements, reductions, and other optimizations
- Fixes a subtle bug in `FilterLabels` that would overwrite any custom fields that the user may have added to their `Detections` instance (the issue here was that `$mergeObjects` does not merge subfields, it overwrites the entire top-level fields. True embedded merging required a bit more tedium)

Docs changes
- Updates `docs/source/user_guide/using_aggregations.rst` to reflect the new aggregation API
- Adds more examples to `docs/source/user_guide/using_views.rst` to whet the reader's appetite with the possibilities of `ViewExpression`s
- Adds thorough examples of using each `ViewExpression` method to the API reference docs

Example usages:

```py
import fiftyone as fo
import fiftyone.zoo as foz

#
# Sample-level fields
#

print(dataset.distinct("predictions.detections.label"))
view = dataset.map_labels("predictions", {"airplane": "OTHER"})
print(view.distinct("predictions.detections.label"))

# implement `map_labels()` via `set_field()`, for fun
print(dataset.distinct("predictions.detections.label"))
view = dataset.set_field("predictions.detections.label", F("label").map_values({"airplane": "OTHER"}))
print(view.distinct("predictions.detections.label"))

view = dataset.set_field("predictions.detections.is_cat", F("label") == "cat")
print(view.distinct("predictions.detections.is_cat"))

view = dataset.set_field("predictions.num_predictions", F("detections").length())
print(view.bounds("predictions.num_predictions"))

#
# Frame-level fields
#

print(dataset.distinct("frames.ground_truth_detections.detections.label"))
view = dataset.filter_labels("frames.ground_truth_detections", F("label") == "vehicle")
print(view.distinct("frames.ground_truth_detections.detections.label"))

# `map_labels()` now supports frame-level labels!
print(dataset.distinct("frames.ground_truth_detections.detections.label"))
view = dataset.map_labels("frames.ground_truth_detections", {"person": "PERSON"})
print(view.distinct("frames.ground_truth_detections.detections.label"))

# implement `map_labels()` via `set_field()`, for fun
print(dataset.distinct("frames.ground_truth_detections.detections.label"))
view = dataset.set_field(
    "frames.ground_truth_detections.detections.label",
    F("label").map_values({"person": "PERSON"})
)
print(view.distinct("frames.ground_truth_detections.detections.label"))
```


benjaminpkane:
My initial comment is that if we think stages like `FilterLabels`, and `LimitLabels` are here to stay (and not be generalized) for the foreseeable future, then `MapLabels` is not out of place.

> The difference between MapValues and SetField is whether the "root" document is the value or it's parent document.

Interesting, this is because `SetField` is not as useful if it does not have access to the parent?

brimoor:
> My initial comment is that if we think stages like `FilterLabels`, and `LimitLabels` are here to stay (and will not be generalized) for the foreseeable future, then `MapLabels` is not out of place.

Yeah I agree that `MapLabels` is consistent with `FilterLabels`, `LimitLabels`, etc. Labels are the most important data type, so I do think it makes sense to provide easy-to-use methods for those types.

brimoor:
> > The difference between MapValues and SetField is whether the "root" document is the value or it's parent document.
> 
> Interesting, this is because `SetField` is not as useful if it does not have access to the parent?

Actually its the other way around: the idea of `MapValues` is that it is exclusively useful for transforming values to other values. `SetField` gets access to the parent document of the field you are setting. In fact, `SetField` has a `root` flag that enables you to specify whether you would like the expression to be relative to the root sample document instead, in which case you have access to any information you want.

So from that standpoint, if something is on the chopping block, it should be `MapValues`.

I'd be interested to get your feedback on the current behavior of `SetField` w.r.t. the parent document vs. root sample document business. I thought about always defaulting to requiring the expression to always be defined relative to the root sample document, but, automatic array unwinding adds some complications there. I would have to do some logic on the `expr` argument to match any unwinding that may have taken place.

benjaminpkane:
Looking through the code, I don't see any `root=True` usage. It might also be a leaky abstraction to users, if I understand how `root=True` works (or would work). I'll respond again in the morning after looking at the implementation more, but a `root=True` option doesn't seem release critical.

benjaminpkane:
Very cool to see all the new additions to view expressions.

benjaminpkane:
I am happy to help update things server side when `AggregationResult`s have been removed. Might speed things along. I'll need to tweak some javascript code at a minimum unless serialized results do not change at all.

brimoor:
> I am happy to help update things server side when `AggregationResult`s have been removed. Might speed things along. I'll need to tweak some javascript code at a minimum unless serialized results do not change at all.

@benjaminpkane That would be great. I've removed `AggregationResult` and updated almost all Python code. There's a least one place that I left a todo that needs your attention in `fiftyone.server.main._gather_results()`.

The serialized format of the aggregation results has changed slightly, so you will need to make some small updates in various places.

One small thing I'm undecided about is which of these is preferable:

```py
#
# current behavior
#
# The motivation here is that np.histogram() returns a `(counts, edges)` tuple, but we also 
# return `other` and (a) three output arguments is starting to be a lot, and (b) `other` is 
# often probably not interesting
counts, edges, other = dataset.histogram_values(...)

# alternative
result = dataset.histogram_values(...)
# result = {"counts": counts, "edges": edges, "other": other}
```

Any thoughts here?

benjaminpkane:
`counts, edges, other` seems convenient to me. Once a user learns the format, I think they would prefer a tuple.

benjaminpkane:
Ok, server looks good now.

brimoor:
@benjaminpkane can you explain the purpose of the `hide_frames` and `hide_result` arguments of `ViewStage`s that process frame labels? The `SetField` and `MapLabels` stages now support frame labels, but I'm unsure what to do with those arguments and why they exist in general.

benjaminpkane:
`hide_result` applies Fields Sidebar filters to pagination (allowing for only matched results), but allows for the full sample info to still be sent so in the sample modal when a filter is changed, requiring more data to be presented, the sample has it.

For example, a user may select only `cat` detections in the sample grid for field, expand a sample (open the modal) and then remove the "only cats" filter. The modal has the data to show all detections because of `hide_result`. Perhaps requiring a new request to be sent when the modal is opened is preferred. `hide_result` seemed like a more appropriate MVP, though.

`hide_frames` can and should be removed. I can at least try to remove its usage in the server right now. It places frames in the `_frames` field. That's all.

benjaminpkane:
`hide_frames` was meant to be an optimization for pagination, but I think it is negligible at this. I just need to truncate pagination requests to only contain the first frame.

benjaminpkane (approved):
LGTM :rocket:

Issue #751, opened 2021-01-08
[FR] Add default confidence thresholds for some (or all?) detection models


brimoor:
Due to implementation details, some detection models in the zoo like `rfcn-resnet101-coco-tf` and `ssd-inception-v2-coco-tf` always emit a very large number (300 and 100, respectively, in these cases) of detections for every image. This is undesirable because it makes the raw output unreadable in the App and makes the `Sample` objects pretty huge, which causes things to be slow.

The `Dataset.apply_model()` method accepts an optional `confidence_thresh` argument that enables controlling the confidence threshold for *any model*, but many models also support a parameter like `confidence_thresh` directly in their associated `Config` files, so we could hard-code a low confidence threshold like 0.1 in the model zoo manifests, so that the default syntax `dataset.apply_model(model)` would result in reasonable behavior:

```
"default_deployment_config_dict": {
    "type": "fiftyone.core.eta_utils.ETAModel",
    "config": {
        "type": "eta.detectors.TFModelsDetector",
        "config": {
            "labels_path": "{{eta}}/tensorflow/models/research/object_detection/data/mscoco_complete_label_map.pbtxt",
            "confidence_thresh": 0.1
        }
    }
}
```

The only concern here is that some use cases like PR curves may want *every* detection to be available. The user could recover this behavior as follows:

```py
model = foz.load_zoo_model("rfcn-resnet101-coco-tf)
model.config.confidence_thresh = 0
```

although this is a bit undocumented. I tried to make `model.config.confidence_thresh` available on almost all models, but it's not formally defined in the `fiftyone.core.models.Model` interface.

Example raw output from offending models:

<img width="1278" alt="Screen Shot 2021-01-08 at 11 33 35 AM" src="https://user-images.githubusercontent.com/25985824/104041241-bc452080-51a6-11eb-86e9-fea5e8921511.png">

And result after filtering with confidence >= 0.1:

<img width="1276" alt="Screen Shot 2021-01-08 at 11 34 10 AM" src="https://user-images.githubusercontent.com/25985824/104041264-c49d5b80-51a6-11eb-83a8-e9ba6ac105c4.png">

ehofesmann:
I really don't want people to have to think about the model configs when they just want to run a model. 
Why not just set a default value of 0.1 to the `confidence_thresh` kwarg in `apply_model()`?

brimoor:
Yeah I suppose that might be the easiest solution. My two reservations on that approach were that:
[...]

PR #819, opened 2021-01-27
Adding small default confidence threshold to model zoo

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/751 by setting a default value of `confidence_thresh=0.3` in the `SampleCollection.apply_model()` method and related methods.

I considered a lower default threshold of 0.1, but some models became quite noisy at that level.

Prior to this change, some bad-behaving detection models would generate 100's of zero confidence detections that would clutter the output results and generally waste space and slow things down.

The raw output of a model can be recovered by passing `confidence_thresh=None` to `apply_model()`.

A best effort is made to implement this filtering efficiently: if the zoo model exposes a `model.config.confidence_thresh` parameter (which all current zoo models do), this parameter will be set to the `confidence_thresh` parameter of `apply_model()` so that the filtering is done as early in the pipeline as possible.

A possible concern with this change is that single label classifiers may not generate predictions for some images (if the confidence is too low), which, speaking from personal experience, can make you wonder if there is something wrong with the model zoo...

Example of applying every zoo model to a dataset:

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "coco-2017",
    split="validation",
    dataset_name=fo.get_default_dataset_name(),
    shuffle=True,
    max_samples=2,
)

for model_name in foz.list_zoo_models():
    try:
        zoo_model = foz.get_zoo_model(model_name)
        zoo_model.ensure_requirements()
    except:
        # can't run this model with your current setup
        continue

    model = foz.load_zoo_model(model_name, error_level=1)
    label_field = model_name.lower().replace("-", "_").replace(".", "_")

    # pass `confidence_thresh=None` here to see raw model outputs with no filtering
    dataset.apply_model(model, label_field=label_field)

```


ehofesmann (approved):
LGTM

Looking through predictions previously I found that thresholds range from around 0.1 to 0.5, so I think 0.3 is a good middle ground.

I ran your script to visualize predictions and 0.3 looks reasonable

PR #820, opened 2021-01-27
lstrip error when saving sample

ehofesmann:
There is a bug in the logic here: https://github.com/voxel51/fiftyone/blob/a07d3ddb4b4b20143b67f08b9bcf27a6e9311d5a/fiftyone/core/odm/mixins.py#L708

The goal is to match
```
        #   my_detections.detections          <- MATCH
        #   my_detections.detections.1        <- MATCH
        #   my_detections.detections.1.label  <- NO MATCH
```

But if the field name includes a number, then this will fail. For example,
```
        #   my_detections1.detections          <- MATCH
        #   my_detections1.detections.1        <- MATCH
        #   my_detections1.detections.1.label  <- MATCH
```

Code to replicate:
``` python
import fiftyone as fo

dataset.clone_sample_field("predictions", "predictions1")

view = dataset.filter_detections("predictions1", F("confidence") > 0.01)

sample = view.first()

for det in sample.predictions1.detections:
    det["test"] = "test"

sample.save()
```

```
ValueError: Modifying root of filtered list field 'predictions1.detections.1.test' is not allowed
```

The problem was that `k.lstrip(ff)` would strip the `1` in `detections.1.test` as well since `1` is in `predictions1`.

### DIFF
--- a/fiftyone/core/odm/mixins.py
+++ b/fiftyone/core/odm/mixins.py
@@ -715,7 +715,9 @@ class DatasetMixin(object):
             for d in update_doc.values():
                 for k in d.keys():
                     for ff in filtered_fields:
-                        if k.startswith(ff) and not k.lstrip(ff).count("."):
+                        if k.startswith(ff) and not k.replace(
+                            ff, "", 1
+                        ).lstrip(".").count("."):
                             raise ValueError(
                                 "Modifying root of filtered list field '%s' "
                                 "is not allowed" % k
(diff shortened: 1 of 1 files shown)


fiftyone/core/odm/mixins.py
```diff
@@ -715,7 +715,9 @@ def _extract_extra_updates(self, update_doc, filtered_fields):
             for d in update_doc.values():
                 for k in d.keys():
                     for ff in filtered_fields:
-                        if k.startswith(ff) and not k.lstrip(ff).count("."):
+                        if k.startswith(ff) and not k.replace(
```
brimoor, line 718: nit: prefer `k[len(ff):]` instead of `k.replace(ff, "", 1)`.

Same thing below.

brimoor (approved):
LGTM (left a small style comment for you before merging)

PR #824, opened 2021-01-28
Simplify frames data model and logic

benjaminpkane:
In response to #815, I propose two simplifications.

* Remove `hide_frames` and `squash_frames` pipeline kwargs
* Remove the "first frame" embedded document on samples (migration needs to be finished)

Neither of the above are needed anymore, thanks to #815 and my own improved understanding of where "frames" are headed.


brimoor (commented):
Oh man, I have no idea what most of the code you're deleting here does, so I'm 100% onboard with getting rid of it!

I'll wait to approve until until the migration is working so I can test it.

brimoor:
Question: with the deletion of the hard-coded first frame labels, will the first frame's labels still appear on initial app-load along with the thumbnail image for the video prior to loading the full data?

This is not directly related to this PR, but, in expanded sample mode, I suppose what one would like to see in JSON mode would be the labels for (only) the current frame in Player51. Showing labels for all frames would be way too much.

benjaminpkane:
> Question: with the deletion of the hard-coded first frame labels, will the first frame's labels still appear on initial app-load along with the thumbnail image for the video prior to loading the full data?

Yes, using `set_field`.
https://github.com/voxel51/fiftyone/blob/7daa1508771184ddf36b937b550890a43f02ea0d/fiftyone/server/main.py#L539

>This is not directly related to this PR, but, in the expanded sample modal for video datasets, I suppose what one would like to see in JSON mode would be the labels for (only) the current frame in Player51. Showing labels for all frames would be way too much.

JSON view still needs work.

benjaminpkane:
Ok, migration looks good

benjaminpkane:
The only problem is that the down migration will not run for users.

The only way to get down migrations to run for users is to introduce logic in the migrations runner that downloads "future" revisions from the web at runtime. A `v0_7_2.py` revision will never exist in a `0.7.1` install...

brimoor (approved):
LGTM. Love the net negative PR!

PR #830, opened 2021-01-29
Replacing links/images in README with absolute paths

brimoor:
The README is directly used to populate the PyPI homepage https://pypi.org/project/fiftyone, so it needs to have absolute links to everything or else they won't work on PyPI.

**Broken images currently on pypi**

Screen Shot 2021-01-29 at 10 17 27 AM


benjaminpkane (approved):
LGTM

PR #834, opened 2021-02-01
Adding support for aggregating on expressions

brimoor:
Adds support for aggregating on *expressions* applied to fields. 

Any expression is allowed, but here's a simple example:

```py
import fiftyone.zoo as foz

num_objects = F("detections").length()

# The `(min, max)` number of predictions per sample
print(dataset.bounds("predictions", expr=num_objects))

# The average number of predictions per sample
print(dataset.mean("predictions", expr=num_objects))
```

This new `Aggregation(field, expr)` syntax is available on all `Aggregation` classes and is effectively shorthand for operations that previously required a `set_field()` view stage to create the new field being aggregated over:

```py
# This is equivalent to the above
view = dataset.set_field("predictions.num_objects", F("detections").length())
print(view.bounds("predictions.num_objects"))
print(view.mean("predictions.num_objects"))
```

Other enhancements include:
- Added `Mean` and `Std` aggregations to compute the arithmetic mean and standard deviation of numeric fields, respectively
- `HistogramValues` will now automatically computes the bounds of the specified field by default (by running a `bounds()` aggregation). Previously, the default behavior was to use MongoDB's `$bucketAuto` to generate bin edges such that the bins have equal counts, but this is not typically what CV/ML users want; they want to see equal-width histograms. The previous behavior can be recovered by setting a new `auto=True` flag
- Fixes a bug in `ViewExpression.__getitem__` when the `expr[n:]` syntax is used
- Merged the `sort()` and `sort_by()` methods of `ViewExpression` into a single method with kwargs analogous to Python's `sorted()` builtin
- Added a number of new `ViewExpression` methods:
    - `prepend()`, `append()`, `insert()`, `extend()` - the familiar array methods
    - `std()` - compute the stdev of an array
    - `split(..., maxsplit)` - added support for `maxsplit` argument
    - `rsplit(..., maxsplit)` - right-hand version of `split()`
    - `literal()` - adds ability to write expressions that involve special MongoDB characters like `$` as literal characters
    - `rand()` - generates a random number in `[0, 1]` every time it is called
    - `range()`, `enumerate()`, `zip()` - more familiar array methods
- Updated `Using aggregations` docs page to include a discussion of aggregating on expressions
- Added examples of aggregating expressions to all `Aggregation` classes and `SampleCollection` methods

### DIFF
--- a/fiftyone/core/aggregations.py
+++ b/fiftyone/core/aggregations.py
@@ -9,6 +9,7 @@ import numpy as np
 
 import eta.core.utils as etau
 
+import fiftyone.core.expressions as foe
 import fiftyone.core.media as fom
 
 
@@ -19,16 +20,28 @@ class Aggregation(object):
     of a :class:`fiftyone.core.collections.SampleCollection` instance.
 
     Args:
-        field_name: the name of the field to compute on
+        field_name: the name of the field to operate on
+        expr (None): an optional
+            :class:`fiftyone.core.expressions.ViewExpression` or
+            `MongoDB expression <https://docs.mongodb.com/manual/meta/aggregation-quick-reference/#aggregation-expressions>`_
+            to apply to the field before aggregating
     """
 
-    def __init__(self, field_name):
+    def __init__(self, field_name, expr=None):
         self._field_name = field_name
+        self._expr = expr
 
     @property
     def field_name(self):
         """The field name being computed on."""
         return self._field_name
+
+    @property
+    def expr(self):
+        """The :class:`fiftyone.core.expressions.ViewExpression` or MongoDB
+        expression that will be applied to the field before aggregating, if any.
+        """
+        return self._expr
 
     def to_mongo(self, sample_collection):
         """Returns the MongoDB aggregation pipeline for this aggregation.
@@ -62,6 +75,11 @@ class Aggregation(object):
         """
         raise NotImplementedError("subclasses must implement default_result()")
 
+    def _parse_field_name(self, sample_collection):
+        return _parse_field_name(
+            sample_collection, self._field_name, expr=self._expr
+        )
+
 
 class AggregationError(Exception):
     """An error raised during the execution of an :class:`Aggregation`."""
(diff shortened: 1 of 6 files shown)


fiftyone/core/aggregations.py
```diff
@@ -631,28 +700,41 @@ def to_mongo(self, sample_collection):
...
+                np.linspace(self._range[0], self._range[1], bins + 1)
             )
-            return
+        else:
+            # Compute bin edges from bounds
+            self._num_bins = bins
 
-        # User-provided bin edges
-        self._edges = list(bins)
+    def _compute_bin_edges(self, sample_collection):
```
benjaminpkane, line 730: This is convenient, although it means if `n` `HistogramValues` aggregations are executed via `aggregate([...])`, then `n +1` queries will be executed. I was hoping to keep the invariant that one (or a constant) number of queries are executed when using the `aggregate()` method.

brimoor: yeah I agree, but my thinking was that `$bucketAuto` is the only fully automatic option, and it makes all histograms look like the uniform distribution, which is actively harmful for CV/ML applications (one wants to see the shape)... so I thought that shouldn't be the default. I think forcing the user to provide the range is the only other option, but that feels more painful to the user in most cases than just slowing down their query a touch

fiftyone/core/expressions.py
```diff
@@ -1706,12 +1788,21 @@ def reverse(self):
...
-        """Sorts this expression, which must resolve an array whose BSON
-        representation can be sorted by JavaScript's ``.sort()`` method.
+    def sort(self, key=None, reverse=False):
+        """Sorts this expression, which must resolve to an array.
+
+        If no ``field`` is provided, this array must contain elements whose
+        BSON representation can be sorted by JavaScript's ``.sort()`` method.
+
+        If a ``field`` is provided, the array must contain documents, which are
+        sorted by the specified field or embedded field.
```
benjaminpkane, line 1798: I'm not sure I understand this documentation? `field` refers to the `key` kwarg?

benjaminpkane, line 1798:
```suggestion
        If a ``field`` is provided, the array must contain documents, which are
        then sorted by the specified field or embedded field.
```

benjaminpkane (approved):
LGTM!

Issue #818, opened 2021-01-26
[FR] Ensure rendered object text is readable in Player51


benjaminpkane:
Right now there are various situations where a rendered Player51 object, e.g. a detection bounding box, does not have visible text showing it's label, confidence, etc.

Currently, all object text is placed on the top line of the bounding box (for detections). Instead, object text should be placed in the most visible way, even when the part of the object falls outside the frame.

brimoor:
I think tooltips will magically take care of this without any changes to the existing overlays :)

benjaminpkane:
Yes, I will close with that work.

Issue #709, opened 2020-12-16
[FR] Segmentation mask labels in the app


ehofesmann:
Provide a color key or add labels on the mask visualization itself for segmentation tasks. Not sure if this feature works in specific settings, but applying the deeplab model from the zoo did not provide labels making it pretty difficult to understand what was happening.

![image](https://user-images.githubusercontent.com/21222883/102387500-737ebb00-3f9e-11eb-8cf1-ec0d2a971387.png)
****

I'm sure this is in the backlog but I didn't see an issue about it so I figure I'd open on. Not top priority but this would be required to write a blog post about segmentations.

Looking in the docs, it was not clear what to expect segmentations to even look like. [We should add images here](https://voxel51.com/docs/fiftyone/user_guide/using_datasets.html?highlight=segmentation#objects-with-instance-segmentations) (probably for all label types)

brimoor:
I like the idea of providing a color-coded legend that can optionally be shown in this case.

ETA provides a `mask_index` field on classes like `ImageLabels` which is a dictionary mapping pixel values to labels for the pixels. This is stored on a per-image basis, which is a bit wasteful as typically all images in a dataset will follow the same legend. Nonetheless, FiftyOne could provide analogous functionality by adding a new `DictField` key `fiftyone.core.labels.Segementation.index` that allows the user to define the index for a mask.

However, the proper way to support this is probably to provide a mechanism to let a user store mask indexes in `dataset.info` that the App would refer to as necessary. In principle, the data model would need to allow for different mask indexes for each `Segmentation` field.

If we do this, we should also consider providing the analogous behavior for any `Label` classes that have string `label` fields (`Classification.label`, `Detection.label`, `Polyline.label`). In this case, the user could store int labels in the `label` fields and then the App would lookup the corresponding string labels in `dataset.info` at display time.

One thing I considered here was providing an optional `target` field that is an int. That way the user could choose how they want to store their labels:
- `target<int>` + `dataset.info` classes list
- directly provide `label<string>`

brimoor:
@benjaminpkane show label for current pixel on hover if a segmentation field is visible

benjaminpkane:
I am going to add a new field on `fiftyone.core.odm.DatasetDocument` (not worrying about naming right now), which will be a MongoEngine `MapField` mapping field names to a list of strings.

[...]

PR #838, opened 2021-02-02
[WIP] Targets, Player51 tooltips, new click behavior

### LINKED ISSUE #816: [FR] Change sample click interactions
As discussed offline, changing to a more standard click UI for sample expansion and selection in the grid would be an improvement.

 Required changes:
 * One click in the sample grid opens the expanded sample view of the clicked sample
 * On sample hover in the grid, a checkbox appears in a corner which can be clicked once for selection
 * Double-clicking has no behavior

benjaminpkane:
Resolves #816, #709

Changelog:
* Adds a `target` field to all relevant label types, `Detection`, `Polyline`, `Keypoints`, `Classification`
* Adds a detailed tooltip to display labels info in Player51's expanded sample view
* Adds target info API to datasets (WIP)
* Adds a chin to samples in the grid (WIP)
* Single-clicks now open the modal, and samples can be selected with a bottom right checkbox.

Screenshot from 2021-02-02 11-28-24


benjaminpkane:
Will undraft today.

benjaminpkane:
Undrafting. Still some issues, please see update description.

brimoor (changes requested):
Wow there was so much new and awesome stuff to chew on here! I've been using this feature for a few days now and letting my thoughts on how to best proceed gel. Here's where I landed as of now:

Discussion is welcome!

## Image chin

Now that I've seen the image chin in action, we're in a good position to
evaluate whether it is adding value. I'm really torn on this, because I see
pros and cons in both directions. I'll explain my thinking below:

### General comments

- I recommend not including the filepath in the chin; it's too much text for
  grid mode. This information is readily accessible by (now) single-clicking on
  the sample.
- I like hiding the chin when no fields are selected. This would also require
  some tweaks to the sample selection UI per my comments below. But, in any case,
  the reason I like hiding the chin when no samples are selected is that it
  provides a clean, media-only view, which can be important for taking in
  qualititive visual details of your dataset. Especially when its an unlabeled
  dataset that doesn't have labels yet.

### Pros

- The original impetus for the chin was to move the tags off the image itself
  so they don't occlude it
- The chin provides a logical place to put the sample selection checkbox.

### Cons

- I suspect that scrolling is not the most user-friendly behavior for the chin.
  The utility of the grid vs. expanded sample mode is the ability to take in
  information at a batch level. But if I have to scroll in  each chin
  individually to see certain things, this primary use case is violated.
  Therefore, if we do adopt an image chin, I'd be in favor of wrapping to
  multiple lines within the chin. Independent of that, there does likely need
  to be a maximum individual tag width, beyond which it is truncated with `...`
- Having the sample selection checkbox permanently visible clutters the sample
  grid, in my opinion. I'd rather have it appear only on hover. However, we
  also now have the possiblity of showing the tooltip on hover in grid mode,
  which may make selection on hover too cluttered. If we don't enable the
  tooltip in grid mode by default, then I think I'd prefer a google photos-style
  checkmark to appear in the upper left for the selection workflow. I also
  think there may be a possiblity of introducing a keyboard shortcut like
  shift + click to select samples rather than expanding them.
  shift + click + drag or shift + click + click (to select a rectangle of samples)
  would also be a nice feature here

## Targets

- For segmentations, we need a targets dict that maps mask values to label
  strings to be shown in the tooltips. This is now implemented and it is
  beautiful! For other label types, if we're going to include `target`
  attributes, I see these strictly as integer values that can be used to lookup
  `label` values in a provided dict (or rather a list; we can take this point
  separately). In other words, for a `Detection` or `Classification` field, the
  way I think we should implement this feature is to let the user either choose
  to specify a `label` or a `target`, but, if they provide a `target`, it would
  be used to lookup a `label` value when the App/library wants a string, not to
  populate an additional `target value` field that is distinct from `label`.

  With that said, allowing users to provide `target`s + lookup dict instead of
  `label`s will take some extra care on our part, because there are lots of
  other places in the code, including dataset import, export, and evaluation,
  that will need to aware that they may need to lookup `label`s based on
  `target` values. That scares me and makes me wonder if we want to make this
  change right now or not. One way to summarize this is to say that, yes, it is
  more space efficient to store target ints + a single lookup dict for large
  datasets with many predictions, but, our goal is not to provide the fastest
  implementation in the world; we're trying to provide the most informative
  human output, which means `label`s, not `targets`.

  All that said, I do like the idea of introducing an API for storing class
  lists for fields, which, if we do implement `target`s in the future, would be
  automagically be used if needed.

## Tooltips

This is a really powerful feature and I love it!

### Thoughts

- What if the tooltip were to always appear, even if you're not hovering over
  an object? The utility here would be standard image analysis where you want
  to get the coordinates of a particular pixel in the image. I'm not too sure
  about this though because I also like the notion that I'll discuss below of
  clicking-when-a-tooltip-is-visble adding that object to a new panel on the
  right. If this were implemented, then there would be nothing to do if the
  user clicks when no object is currently hovered. Perhaps this is okay though

### Scrolling/arrow up-down

- I like the idea of scrolling to view different labels. To me, the way this
  should work is that scrolling should only change the z-order of objects that
  are currently underneath your mouse pointer. And scrolling should change the
  tooltip that's currently being shown as well. Also, to make the feature
  easier to use, I don't think it should wrap-around; basically scrolling
  should iterate through the z-order of any objects under your current pointer,
  and if you reach the beginning or end of the list, then scrolling should stop.
  This will help the user get their bearings; they can scroll big-time up and
  then down and the number of z-order changes that occur will help them
  understand how many overlapping objects there are. With cyclic scrolling it
  may be hard to assess how many there are. With this said, the potential
  drawback here is that the user may get frustrated when scrolling in one
  direction does nothing (since they need to scroll in the other direction).
  There's definitely a lot of complexity here to consider. The main outcome
  I'm looking for is that there be a 1-1 correspondence between the tooltip
  and the top-most object and the object that will be selected if the user
  clicks.
- How about introducing a light transparent background fill on the currently
  hovered detection object, to make it more apparent what is currently
  selected. If we do this, it would need to be more transparent than the
  transparency used for instance segmentation masks, and so that the user could
  mostly see through it to see what the object is and any boxes that may be
  below it

### Single vs multiple objects per-tooltip

- I agree that, with the tooltip as verbose as it currently is, showing all
  objects in the tooltip might get pretty cluttered. Especially in cases where
  a detection model produces a number of duplicates. Also, if multiple labels
  were shown, then it would be less intuitive to make label section add the
  object whose tooltip is currenly being shown to a panel on the right (two
  options here would be to add either (a) all, or (b) only the top-most object
  in the tooltip to the righthand panel, but I worry that both of these are
  potentially confusing).
  On the other hand, showing all hovered objects in the tooltip would arguably
  provide for the best comparison experience when trying to evaluate a ground
  truth vs detected object that are close to overlapping. One consideration is
  whether this could be a configurable option (a) scrolling z-order with one
  object per tooltip, versus (b) all objects per tooltip. But, in the end, I
  think I like only allowing option (a)

### Feature requests

- Add support for configuring whether the tooltip appears in grid mode via
  `AppConfig`. I'm open to discussion here, but I think it may be desirable to
  *not* show the tooltip in grid mode by default. This could also (either
  instead or in addition to the config) be a toggle under the Options menu for
  the grid
- How about replacing `None` with `-` to minimize visual attention on missing
  fields?

### Segmentation tooltips

- We should add an `id` to `Segmentation` as you did with `Classification`.
  Basically all `Label` classes (except list of labels containers) should have
  an `id` field for consistency

- I'd recommend reordering the content so that the most useful content
  (the label string, followed by the numeric index) is on top:
  I suggested mask coords go after mask size so that it is left-aligned with
  the pixel coordinates, which also change then the mouse moves.

- I'm going back and forth on whether to show the mask size and mask dimensions.
  It is interesting information, and it is often distinct from the image
  dimensions (since models typically output masks of fixed sizes independent of
  image dimesions). However, speaking from experience, it was a bit confusing to
  see the image coords and mask coords changing independently. Perhaps it may
  help to clarify that these are "mask" coordinates, per the suggested tooltip
  I show below. Alternatively, perhaps we should just hide these?

```
x, y
------------------
<label>         <mask value>
label           mask value

<field>         <id>
field           id

<mask-size>     <mask-coords>
mask size       mask coords
```

### Detection tooltips

- Convert [x, y, w, h] to pixel coordinates, for brevity and also consistency
  with the image coordinates we show

- Here's a recommend reordering of the content to show the most important
  content first:

```
x, y
----------------------
<label>   <confidence>
label     confidence

<field>   <id>
field     id

<w, y>    <w, h>
top-left  size

(attributes)
```

### Polyline tooltips

- Converting to pixel coordinates (like detections)

- How about omitting the `closed` and `filled` values, since that's apparent
  from how the shape is drawn?

- Perhaps omit # keypoints? (not really relevant)

```
x, y
----------------------
<label>   <confidence>
label     confidence

<field>   <id>
field     id
```

### Keypoint tooltips

- Converting to pixel coordinates (like detections)

```
x, y
----------------------
<label>   <confidence>
label     confidence

<field>   <id>
field     id

<x, y>    <total #>
point     # points
```

### Classification tooltips

```
x, y
----------------------
<label>   <confidence>
label     confidence

<field>   <id>
field     id
```

### General tooltip comments

- I'm now wondering whether to put `field` and `id` or `label` and
  `confidence/value` on the first row of tooltips, since things like polylines
  and keypoints often may not have these values...

brimoor:
Also, as we discussed live last week and I alluded to in my comments above a few times, I'd like to propose the following new expanded sample feature:

## Selected objects panel

- When object(s) are selected in the App by clicking them, the tooltips (or versions of them) are added to a new `Selected objects` section in the righthand sidebar. This section gives the user the ability to review information about multiple objects of interest and do things like copy + paste their IDs.
- This section would also contain the current selected objects UI element, either in its current form as a dropdown, or expanded into an options-like menu, which would allow the user to take action on 
- Selected objects could be individually unselected by clicking an `x` on its tooltip in this panel (or by unselecting the object in the app).
- If selected objects are hidden, their tooltips would disappear from this section. Basically, there would be a 1-1 correspondence with all selected and currently visible objects in the media player
- This section would contain a UI element that would enable the user to type in a "tag" that could be submitted in order to add a tag to all currently selected objects

brimoor:
@voxel51/developers I think a good next step here may be to have another quick interactive session where anyone who has interest can take a look at the current state of these features (ie image chin and tooltips) and cast a vote on how to proceed on a couple key points.

benjaminpkane:
Still some work to do.

benjaminpkane:
Ok, I still need to clean up `Color by value` and `Refresh color` styles in the modal, and work through two small video updates. Other than that, things are ready. Will follow up when the above have been addressed.

Screenshot from 2021-02-26 15-36-25
Screenshot from 2021-02-26 15-35-55

benjaminpkane:
Ok, my final points have been resolved.

Regarding the chin, I ended up removing it. What I settled on is almost identical to before except values are squished with ellipses, and only one line of values can be present. Many values cause horizontal scrolling but I can remove that. I played with other implementations but they all seemed overly complex, not just in terms of implementation but also from a general UX perspective.

In short, my updates are what I propose after playing around with many variations of most the details we have discussed. A small point is I am not quite sure where to place the `Color by value` checkmark in the expanded sample view. See screenshot.

Screenshot from 2021-03-01 08-59-44
Screenshot from 2021-03-01 08-59-34

benjaminpkane:
Resolves #818

brimoor:
@benjaminpkane awesome looking forward to testing this afternoon and giving a review. One quick thing: what do you think about moving the options back to the sidebar? My reasons for this are:
- There may be more options in the future, and they're not important enough to show all the time
- I'd like to use the space above the grid for other "actions" like tagging samples/objects

benjaminpkane:
> @benjaminpkane awesome looking forward to testing this afternoon and giving a review!
> 
> One quick thing: what do you think about moving the options back to the sidebar? My reasoning is:
> 
> * There may be more options in the future, and they're not important enough to show all the time
> * I'd like to use the space above the grid for other "actions" like tagging samples/objects

Ok, no problem.

benjaminpkane:
I suppose I will say that we will certainly accrue more options, I just don't think they should reside in the sidebars as they stand. I see the sidebar as a place for filtering and customization with respect to the the shape of the user's dataset/view. And I think meta/global-options and actions should live somewhere or else or at least be properly distinguished as something else.

I'm not a designer of course, and I also think we agree on needing to distinguish these options. For consistency between the grid and modal I can move them back for now.

All of this is kind of a moot point at the moment if we wanted to make coloring options field specific, of course. Then they could go in filters.

Also noting the options we accrue will sooner than later be more than anything that can fit in the free space available without expandable components.

ehofesmann:
Nice work Ben! Pretty much all of my previous comments are resolved.

A few minor things I noticed that don't necessarily need to be changed:
1) It would be nice to be able to shift/ctrl click the check boxes in the grid view to select multiple samples. Since the checkbox is pretty small, selecting multiple objects back to back can be frustrating if you misclick.
2) I feel like clicking segmentations in the expanded view should do something. It's probably not feasible to select individual objects for semantic segmentations, but even just selecting the entire segmentation group could be useful for tagging purposes.
3) Scrolling/arrowing through z order of detections may not be very clear to new users. It could be useful to add some kind of note (or tooltip lol) mentioning this feature.

benjaminpkane:
Apologies, I the API has changed slightly as targets only apply to `Segmentation` fields now.

```py
import fiftyone as fo 
import fiftyone.zoo as foz 

dataset = dataset.limit(10) 
dataset = dataset.clone("tooltip-testing")
dataset.apply_model(model, "segmentations")
dataset.persistent = True

for s in dataset: 
    s["polyline"] = fo.Polyline(points=[[(0.5, 0.5), (0, 0.5), (0.5, 0)]], closed=True, filled=True) 
    s["keypoints"] = fo.Keypoints(keypoints=[fo.Keypoint(label="cat", points=[[0.5, 0.5]])]) 
    s["classification"] = fo.Classification(label="testing") 
    s.save()
 
dataset.default_mask_targets = {i: "idx-%d" % i for i in range(0, 256)}

session = fo.Session(dataset=dataset)
```

brimoor:
Agree with Eric, nice work Ben!!

I had a few reactions after using this for an evening:

Grid selection
- Definitely agree with Eric that providing a multi-select mode for images in the grid would be very useful. How about shamelessly copying Google Photos for this: once an image is selected, clicking *anywhere* in subsequent images will select/deselect them rather than expand them. Google Photos' shift click functionality would also be nice, but it could be made an issue for now if necessary. I believe how it works is that shift-clicking an image will select all images between that image and the last-selected image in the grid (or deselect all of them if every image in the list is already selected)

Label selection (expanded mode)
- I agree with Eric that it would be nice to support selecting segmentation fields, for consistency with how other fields are clickable. One reason is that, in the example code you show, selecting one bbox and then selecting "hide others" has no effect on segmentations, but I'd expect that it would be included like all other `Label` types. The most consistent action here would be to draw a dotted border around each colored region, but I suspect that would be a fair amount of work to support. An easier option would be to change the alpha value of the mask (make it 100% opaque I guess? since selection should mean *emphasis*)
- Related to the previous point, can you support selecting classification labels? That would complete the feature a bit more

z-order scrolling
- With my mac trackpad and mouse, I find that the z-scrolling is too fast for me to really control. I really think that removing the circular scrolling feature would help a lot here; reaching the bottom or top of the z-order really helps to limit wild scrolling and give a sense for how many objects are there. My other thought is: is there a way to limit scrolling sensitivity for users like me who don't have a "discrete" scroll wheel? I image that if my mouse's scroll wheel had "clicks" then the z-order scrolling would be much easier to use. This second point feels tricky to me... (and makes me want the first point more as a band-aid)

Naming
- We need to pick between `mask_targets` and `mask_index` for the segmentation mask labels. `mask_index` is used by `evaluate_segmentations()` and comes from ETA, and you're using `mask_targets` here. @ehofesmann any thoughts on the preferred terminology for the dict mapping mask pixel values to label strings?

Session refreshes
- I noticed that updating my app config and calling `session.refresh()` doesn't trigger the changes in the App (as the [docs promise](https://voxel51.com/docs/fiftyone/user_guide/config.html#launching-the-app-with-a-custom-config)). I have to ctrl+R in my browser in order to see the changes

Small stuff
- How do you feel about moving the checkboxes in the image grid from the upper-right to the upper-left? Feels more natural to me (and is what google photos uses)
- Can you make the animation on the `Refresh colors` button a bit faster? It feels sluggish. Also, the font color/style is different for that button currently

benjaminpkane:
> Related to the previous point, can you support selecting classification labels? That would complete the feature a bit more

Sure, it will be a small lift, but I can make it happen. Something I omitted out of prioritization.

>With my mac trackpad and mouse, I find that the z-scrolling is too fast for me to really control. I really think that removing the circular scrolling feature would help a lot here; reaching the bottom or top of the z-order really helps to limit wild scrolling and give a sense for how many objects are there. My other thought is: is there a way to limit scrolling sensitivity for users like me who don't have a "discrete" scroll wheel? I image that if my mouse's scroll wheel had "clicks" then the z-order scrolling would be much easier to use. This second point feels tricky to me... (and makes me want the first point more as a band-aid)

Can I propose that we turn off mousewheel z-ordering and only use the arrow up/down shortcuts? I think it would be nice to reserve wheeling for zooming in and out of the media.

Other points are noted and I can address. I'll note that clicking anywhere in the top shaded region selects the sample, but I can add true multi-selection. Thanks for the feedback Eric and Brian!

brimoor:
> > Related to the previous point, can you support selecting classification labels? That would complete the feature a bit more
> 
> Sure, it will be a small lift, but I can make it happen. Something I omitted out of prioritization.
> 
> > With my mac trackpad and mouse, I find that the z-scrolling is too fast for me to really control. I really think that removing the circular scrolling feature would help a lot here; reaching the bottom or top of the z-order really helps to limit wild scrolling and give a sense for how many objects are there. My other thought is: is there a way to limit scrolling sensitivity for users like me who don't have a "discrete" scroll wheel? I image that if my mouse's scroll wheel had "clicks" then the z-order scrolling would be much easier to use. This second point feels tricky to me... (and makes me want the first point more as a band-aid)
> 
> Can I propose that we turn off mousewheel z-ordering and only use the arrow up/down shortcuts? I think it would be nice to reserve wheeling for zooming in and out of the media.

Yeah I'm fine with deactivating mouse scrolling for now. I'd still vote for no circular z-ordering, even just for arrow keys though. I haven't investigated to see how long the z-ordering "lasts", but I think it would be fine for any one-off z order changes to be cleared whenever a field is toggled, the image changes, etc.

> Other points are noted and I can address. I'll note that clicking anywhere in the top shaded region selects the sample, but I can add true multi-selection. Thanks for the feedback Eric and Brian!

Yeah I did notice that clicking anywhere in the shaded region selects the samples. That helps with the initial click target, and full image clicking will take that to the next level in "selection mode".

ehofesmann:
> Naming
> 
> * We need to pick between `mask_targets` and `mask_index` for the segmentation mask labels. `mask_index` is used by `evaluate_segmentations()` and comes from ETA, and you're using `mask_targets` here. @ehofesmann any thoughts on the preferred terminology for the dict mapping mask pixel values to label strings?

I would prefer `mask_targets`, we may need to change `evaluate_segmentations()` but I feel like see `targets` more frequently than `index` in this context.

brimoor:
> > Naming
> > 
> > * We need to pick between `mask_targets` and `mask_index` for the segmentation mask labels. `mask_index` is used by `evaluate_segmentations()` and comes from ETA, and you're using `mask_targets` here. @ehofesmann any thoughts on the preferred terminology for the dict mapping mask pixel values to label strings?
> 
> I would prefer `mask_targets`, we may need to change `evaluate_segmentations()` but I feel like see `targets` more frequently than `index` in this context.

SGTM. Targets it is. I'll push a quick update that uses `mask_targets`, and now(!) can automatically lookup this dict when needed in places like `evaluate_segmentations()`

benjaminpkane:
> I noticed that updating my app config and calling session.refresh() doesn't trigger the changes in the App (as the docs promise). I have to ctrl+R in my browser in order to see the changes

`Show tooltip` was not getting updated (fixed), but generally speaking this is working for me.

```py
session.config.show_tooltip # True
session.config.show_tooltip = False
session.refresh() # tooltip hidden
```

Discussed comments have generally been addressed. Too rotate hover objects, down arrow is forward and up arrow is backward. https://github.com/voxel51/fiftyone/pull/838/commits/73cc800b93a8ae8946cd83ee6364f8892060ef3f can used for a merge.

brimoor:
This is working amanzing-ly! Love it!

brimoor:
Only thing I noticed when playing with the quickstart dataset is that a bug returned where clicking `reset` on a confidence filter resets the filter to the lowest observed value, but not necessarily back to "default" such that the field color will deactivate (ie., the issue is that field remain "active" after pressing `reset` right now)

brimoor:
I'm noticing an uncaught exception on this branch whenever exiting my Python session with an App instance open:

```
In [1]: import fiftyone as fo                                                                                                                                                        
In [2]: session = fo.launch_app()                                                                                                                                                    
App launched. Point your web browser to http://localhost:5151

In [3]: exit()                                                                                                                                                                       
Exception ignored in: <bound method Service.__del__ of <fiftyone.core.service.ServerService object at 0x12fd7cda0>>
  File "/Users/Brian/dev/fiftyone/fiftyone/core/service.py", line 86, in __del__
  File "/Users/Brian/dev/fiftyone/fiftyone/core/service.py", line 128, in stop
  File "/Users/Brian/dev/env/fifty3/lib/python3.6/site-packages/psutil/__init__.py", line 1362, in wait
TypeError: super() argument 1 must be type, not None
```

brimoor:
@benjaminpkane the two issues I mentioned are resolved, thanks for that 👍 

I wrapped up the mask targets work on this branch. The evaluation code now uses them if present, and its all documented, etc.

One note: I removed the `add_mask_targets()` and `delete_mask_targets()` methods since the same can be accomplished via `dataset.mask_targets["new"] = targets and `del dataset.mask_targets["new"]`. It's true that the latter does not do any field validation, but we're also not doing the work of renaming these things if, say, `rename_sample_field()` is called, so I think its for the best to keep it simple. Likewise, `del xxx` does require a manual `.save()` afterwards, but so do other in-place modifications, so I think its okay. (besides, our upcoming save context will obviate for `.save()` altogether).

I also added the necessary DB migration for the new dataset doc fields and added support for exporting/importing mask targets when using `Dataset.to_dict()` and `Dataset.export(..., fo.types.FiftyOneDataset)`. In general, I'd like to ensure that these two formats always contain the all information necessary to reconstruct a dataset fully.

benjaminpkane:
Ok, I added basic multi-selection. It is only useful for selecting a range of samples at the moment. It's not quite Google Photos, but its a start.

I'm realizing now that I should render label tags in tooltips now that they have been added. And perhaps any attribute on a `label`? Shortening to `{}` and `[]` for dicts and lists respectively? Note that right now, I limit it to 4 attributes rendered, and indicate there are more.

Right now `Detection.attributes` are rendered, but not attributes in general...

brimoor:
Yeah I'd definitely like to show a label's tags in the tooltip. And other direct attributes as well.

I'm fine with rendering list and dict attributes as `[...]` and `{...}` or something like that, without showing their contents. With the notable exception of `tags`, of course, which the user definitely should be able to see the elements, perhaps rendered as `list, of, tags`.

Hmm, how to handle many attributes? Noting that Eric's vote was for providing a mechanism (in the future) for selecting what attribute you want to see.

Three options come to mind for me:

(1) What you're doing: show up to `n` and indicate there are more if necessary
(2) Always show all attributes, and wait for Eric's feature to deal with users that provide many attributes
(3) Provide a expand/contract UI element in the bottom right of the tooltip that toggles between (1) and (2)

benjaminpkane:
Ok, I'm satisfied with where things are.

benjaminpkane:
Screenshot from 2021-03-05 13-39-57

brimoor:
Classification selection and tags are looking good! Things are looking stable and I'm happy to approve this PR now.

I have a couple final parting thoughts for you before merging:

- `Refresh colors` doesn't currently apply to segmentation masks; can you make it do so?

- I would recommend *not* sorting the tags alphabetically in the tooltip, since the underlying data structure is a list and will preserve the user's order if they have a reason for storing their tags in a certain order

- How about adopting the convention of hiding fields with no value from the tooltip? EG if a label has no tags, then don't include that entry in the tooltip at all. Similarly, if the label has no `label` or `confidence` (<-- ground truth), then hide those altogether

- I think that `label` and `confidence` should appear before other misc user-defined attributes in the tooltip, since they are in the formal data model. The remaining misc attributes can be in alphabetical order below

brimoor (approved):
LGTM

brimoor:
```py
import fiftyone as fo

detection = fo.Detection(
    label="test",
    bounding_box=[0.25, 0.25, 0.50, 0.50],
    confidence=0.95,
    tags=["there", "are", "really", "too", "many", "tags", "to", "show", "in", "the", "tooltip"],
    field1="hi",
    field2=15,
    field3=51.0,
    field4="hi",
    field5=15,
    field6=51.0,
    field7="hi",
    field8=15,
    field9=51.0,
    field10=["will", "these", "show", "up", "?"],
    field11={"how": "about", "these": "?"},
)

sample = fo.Sample(
    filepath="/Users/Brian/Desktop/test.png",
    ground_truth=fo.Detections(detections=[detection]),
)

dataset = fo.Dataset()
dataset.add_sample(sample)

```

Screen Shot 2021-03-07 at 12 07 38 PM

PR #842, opened 2021-02-04
Adds support for returning a single sample matching an expression

brimoor:
Adds a `SampleCollection.one()` method that returns a single `Sample` matching a given expression.

`collection.one(expr)` is essentially shorthand for `collection.match(expr).first()`, with an additional `exact=True` argument if you would like to strictly enforce that a single match must exist.

The impetus for this was trying to find a suitable replacement for accessing a sample by filepath over in https://github.com/voxel51/fiftyone/pull/841:

```py
# too much of a mouthful...
sample = dataset.match(F("filepath") == filepath).first()
```

which could now be written as:

```py
sample = dataset.one(F("filepath") == filepath)
```

Of course it has to be said that https://github.com/voxel51/fiftyone/pull/841 would enable the *even* shorter `dataset[filepath]` syntax, but I can imagine cases where extracting a sample based on a more general expression may be called for.

I marked this as needs discussion in case anyone has any thoughts about appropriate generalizations or restrictions of this method that make it easier to use. Ease of use is key here because this is purely shorthand for `match().first()`.

Example usage taken from docs:

```py
import fiftyone as fo
import fiftyone.zoo as foz

#
# Get a sample by filepath
#

# A random filepath in the dataset
filepath = dataset.take(1).first().filepath

# Get sample by filepath
sample = dataset.one(F("filepath") == filepath)

#
# Dealing with multiple matches
#

# Get a sample whose image is JPEG
sample = dataset.one(F("filepath").ends_with(".jpg"))

# Raises an error since there are multiple JPEGs
dataset.one(F("filepath").ends_with(".jpg"), exact=True)
```

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -180,6 +180,67 @@ class SampleCollection(object):
         """
         return [s for s in self[-num_samples:]]
 
+    def one(self, expr, exact=False):
+        """Returns a single sample in this collection matching the expression.
+
+        Examples::
+
+            import fiftyone as fo
+            import fiftyone.zoo as foz
+            from fiftyone import ViewField as F
+
+            dataset = foz.load_zoo_dataset("quickstart")
+
+            #
+            # Get a sample by filepath
+            #
+
+            # A random filepath in the dataset
+            filepath = dataset.take(1).first().filepath
+
+            # Get sample by filepath
+            sample = dataset.one(F("filepath") == filepath)
+
+            #
+            # Dealing with multiple matches
+            #
+
+            # Get a sample whose image is JPEG
+            sample = dataset.one(F("filepath").ends_with(".jpg"))
+
+            # Raises an error since there are multiple JPEGs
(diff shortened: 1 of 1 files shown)


benjaminpkane (approved):
I like this!

This is just an opinion, but I think naming it just `one()` would be nice. Up to you.

benjaminpkane:
Regarding the discussion requested here and in #841, the only alternative I can suggest is to allow view expressions in `collection.__getitem__()`, which is a shorthand for `match()`.

brimoor:
> I like this!
> 
> This is just an opinion, but I think naming it just `one()` would be nice. Up to you.

oooh I like `one()`. I'll make that change!

brimoor:
> Regarding the discussion requested here and in #841, the only alternative I can suggest is to allow view expressions in `collection.__getitem__()`, which is a shorthand for `match()`.

Yeah I agree this is the only further step in the syntax direction, but probably too much as `dataset[expression]` would be pretty unusual

Issue #827, opened 2021-01-28
[FR] Add string histograms to Scalars tab


brimoor:
It would be nice to see graphs for `string` and `list(string)`-type fields under the `Scalars` tab in the App:

```py
import string
import random

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart").clone()

for sample in dataset:
    sample["string"] = random.choice(string.ascii_uppercase)
    sample.save()

session = fo.launch_app(dataset)
```

Issue #833, opened 2021-01-29
[FR] Tweak text in App label filters


brimoor:
We've had some user feedback about the terminology used in the label filters in the App being unintuitive

(a) `+ add label` makes it sound like using this element will allow you to add labels to your dataset or something like that
(b) `Filter no confidence` was recently changed to its current text. Previously it was `Show no confidence` and that caused some confusion for users.

Discussion is welcome, but here's my proposal for text that is more indicative of the function:

(a) `+ filter by label`
(b) `Hide no confidence`

<img width="239" alt="Screen Shot 2021-01-29 at 6 25 07 PM" src="https://user-images.githubusercontent.com/25985824/106337574-83f1a900-625f-11eb-9c7c-add334a9555e.png">

brimoor:
https://github.com/voxel51/fiftyone/pull/848.

Issue #803, opened 2021-01-21
[FR] App filtering enhancements


brimoor:
The new filtering behavior introduced in https://github.com/voxel51/fiftyone/pull/793 is so awesome! The fact that the sample

After using it, I have a few suggestions for small enhancements to further increase the value:

1. **Hide label dropdowns after selection**. Currently the label selection dropdown in the fields sidebar stays open after you select an option, so that you can select multiple. However, I find myself getting flustered by the menu remaining open (since there's no dedicated close option), and would rather have it close by default and me have to click into the field again to reopen it rather than having to figure out how to **close** it after I performed a search. Note that using keystrokes is a very efficient way to use this menu option, and, since it autocompletes nicely for you, the fact that the menu closes each time is not a factor since it will automatically open again as you start typing

<img width="1274" alt="Screen Shot 2021-01-21 at 9 37 02 AM" src="https://user-images.githubusercontent.com/25985824/105367633-a71cb880-5bce-11eb-87b4-149b0c8b95c9.png">

2. **Excluding labels from a view**. Currently selecting a label value from the fields sidebar will show *only* that label. However, another common use case is to *exclude* a certain label. A good example of this is shown below where there is an `undefined` class ("I want to see everything that is not undefined")

Now how to implement this with a clean UI... What about adding a way to toggle a `[label x]` tag between include/exclude state by introducing a `+/-` symbol on the left side of the tag. By default, a new selection would appear as `[+ label x]`, where the `+` means you ARE viewing that label. If you click the tag, it would toggle to `[- label x]`, meaning that you are now *excluding* that label. You would then delete a tag by clicking *only* near the `x`.

brimoor:
So I did some thinking about the different types of filtering that one would like to perform for each field type:

All value types
1. (default) No filter, all samples
2. Hide samples with value == None
3. Show only samples with value == None

Numeric value
4. Show only samples with value in [a, b], excluding `None`

String value
5. Show only samples with value in {"list", "of", "strings"}
6. Hide samples with value in {"list", "of", "strings"}

Boolean value
7. Show samples with value in subset of {True, False}

Labels are composite types, you can filter by any other their fields, which have the above types.

The screenshot below shows the currently available filters, which are designed to support `1,2,4,5,7` and not yet `3,6`.

In order to support 3, one option is to generalize the `Filter value` checkbox to a three-state toggle like `Only None|Hide None|All`, and my proposal for supporting `6` is point 2 of the issue description above.

[...]

Issue #725, opened 2020-12-23
[FR] Refresh label colors doesn't apply to "color by label" setting


brimoor:
Currently clicking `Refresh label colors` doesn't affect the colors when `Color by label` is activated. It would be nice if it did (one may still want to refresh colors if they are conflicting)

<img width="1392" alt="Screen Shot 2020-12-23 at 11 13 53 AM" src="https://user-images.githubusercontent.com/25985824/103016618-4f782800-4510-11eb-84ea-555f21ce7623.png">

brimoor:
So as of v0.7.3, there are still some weird behaviors on this feature:

- In grid mode, you can click `Refresh field colors` only once when `Color by label` is selected, after which point it stops working
- In expanded sample mode, clicking `Refresh field colors` affects the grid mode (I can see in the background) but does not affect the expanded grid

<img width="1276" alt="Screen Shot 2021-02-18 at 2 28 52 PM" src="https://user-images.githubusercontent.com/25985824/108410827-e2be9880-71f5-11eb-9d42-6246664b8600.png">

<img width="1271" alt="Screen Shot 2021-02-18 at 2 28 43 PM" src="https://user-images.githubusercontent.com/25985824/108410833-e4885c00-71f5-11eb-825f-b830611b40db.png">

benjaminpkane:
Thanks, will address this week.

PR #848, opened 2021-02-06
Adds string and boolean field filters in the sidebar

benjaminpkane:
Resolves https://github.com/voxel51/fiftyone/issues/827, https://github.com/voxel51/fiftyone/issues/833, and https://github.com/voxel51/fiftyone/issues/833.

Also addresses points 1 and 2 from https://github.com/voxel51/fiftyone/issues/803.

Very happy with this refactor. Lots of technical debt has been banished.

* Adds string and boolean field filters in the sidebar
* Adds color by label support for classification and scalar field bubbles in the sample grid

Needs more testing, I have yet to test video and classifications. The implementation is complete. though.


brimoor (commented):
Amazing!

Can you also add support for this to boolean fields as well, treating `True` and `False` as strings? This would be a quick way to get filtering working on all primitive field types without worrying about new UI elements (the boolean toggle).

Also, can you make string/bool fields appear under the scalar graphs tab?

In fact, boolean fields currently *do* already appear in the scalar graphs (not strings), although the x-tick labels don't seem to show up.

My test code:

```py

import random

import fiftyone as fo
import fiftyone.zoo as foz

ANIMALS = [
    "bear", "bird", "cat", "cow", "dog", "elephant", "giraffe",
    "horse", "sheep", "zebra"
]

for sample in dataset.select_fields():
    sample["string"] = random.choice(ANIMALS)
    sample["bool"] = (random.random() < 0.5)
    sample.save()

```

brimoor:
One more thing on this branch: can you get per-label colors working for classifications, strings, and boolean fields? I think the per-label colors option doesn't apply to one or more of these.

These requests are motivated by viewing datasets that have been evaluated via https://github.com/voxel51/fiftyone/pull/849 in the App and wanting to filter by correct/incorrect (boolean field), TP/FP/FN/TN (string field), and wanting to see the different values for these fields in different colors (so that red = correct, green = incorrect, for example).

In searching for an existing issue about this, I found https://github.com/voxel51/fiftyone/issues/725, which is less-critical but related.

benjaminpkane:
Will be undrafted today.

benjaminpkane:
Undrafting

brimoor:
I think `Filter no label/value` is working in reverse- I'd expect it to remove objects/samples that have no label/value, but on the test code I put above, it removes all samples

brimoor:
Also, are you able to get string fields to appear under the scalars tab? I'm not seeing a graph for the `string` field when running my sample code above.

brimoor:
Here's some test code that has `None` values in all field types:

```py

import random

import fiftyone as fo
import fiftyone.zoo as foz

ANIMALS = [
    "bear", "bird", "cat", "cow", "dog", "elephant", "giraffe",
    "horse", "sheep", "zebra"
]

dataset.add_sample_field("string", fo.StringField)
dataset.add_sample_field("bool", fo.BooleanField)

with fo.ProgressBar() as pb:
    for sample in pb(dataset):
        for detection in sample["ground_truth"].detections:
            if random.random() < 0.1:
                detection.label = None

        for detection in sample["predictions"].detections:
            if random.random() < 0.1:
                detection.label = None

        if random.random() < 0.1:
            sample["uniqueness"] = None

        s = random.choice(ANIMALS)
        if random.random() < 0.1:
            s = None

        sample["string"] = s

        r = random.random()
        if r < 0.33:
            b = True
        elif r < 0.66:
            b = False
        else:
            b = None

        sample["bool"] = b
        sample.save()

print(dataset.filter_labels("ground_truth", F("label") == None).count("ground_truth.detections"))
print(dataset.filter_labels("predictions", F("label") == None).count("predictions.detections"))
print(dataset.match(F("uniqueness") == None).count())
print(dataset.match(F("string") == None).count())
print(dataset.match(F("bool") == None).count())

```

brimoor:
Also in running the test code with `None` labels above, I think I found a bug: the App crashes if objects have `label == None`.

brimoor:
I did a bit more thinking about filtering over in https://github.com/voxel51/fiftyone/issues/803#issuecomment-776817054.

Do you think you can sneak support into this PR for hiding the `Filter no values` toggles when there are no `None` values? This would reduce cognitive overload of the filtering options.

benjaminpkane:
> Do you think you can sneak support into this PR for hiding the `Filter no values` toggles when there are no `None` values? This would reduce cognitive overload of the filtering options.

Yes, will do.

benjaminpkane:
> Also, are you able to get string fields to appear under the scalars tab? I'm not seeing a graph for the `string` field when running my sample code above.

Working on it. I forgot about this request.

brimoor:
@benjaminpkane what do you think about adding (a) from https://github.com/voxel51/fiftyone/issues/833 in here? For labels it would be `+ filter by label` and for other values `+ filter by value`.

I'm not quite sure on part (b) from there anymore. May need more discussion.

benjaminpkane:
Will change to (a). Makes sense to me.

For (b) I would propose `Exclude no value` or `Exclude None` (where `None` is code highlighted.

brimoor:
ok yeah I like `exclude` for the None cases.

I think that terminology should be used for `Exclude true` and `Exclude false` in the boolean filters too, until we figure out something better as per https://github.com/voxel51/fiftyone/issues/803#issuecomment-776826126.

benjaminpkane:
Comments addressed, though I am still testing. Happy to tweak things still, like `Exclude ...` (See screenshots)
Screenshot from 2021-02-10 13-20-53
Screenshot from 2021-02-10 13-20-34

. Still digesting some comments made on issues outside this PR. May sneak some more changes in, though.

brimoor:
Looking good! I noticed a couple things in expanded sample mode:
- The `Exclude None` filter for labels doesn't seem to be working. Perhaps this is because the label has been cast to a string or something like that? On a related note: I think it would be preferable to render `label == None` as `label == ""` so that users can have boxes with no labels if that's what they'd like
- I noticed that applying label filter to a field that has been unselected does not cause the label fields background to change colors. The color change is important regardless of whether the label field is toggled on because it conveys whether their is an active filter on that field (in my opinion)

brimoor:
I hate to keep stringing along this PR, but do you think you can squeeze in hiding the `Exclude None` element when there are no `None`-valued items in the current view?

I bring this up because of a second point of clarification I just found: when using the `filter by label/value` option in the App to select certain labels you want to see, I think that `None` valued fields should also be hidden, since their label/value doesn't match the search any more than a non-None-non-matching label value does. If the `Exclude None` element were always shown, then this would necessitate auto-checking the `Exclude None` box when a filter is performed. But, it would be cleaner if the `Exclude None` element just disappeared since it is no longer relevant.

Screen Shot 2021-02-10 at 4 18 20 PM

benjaminpkane:
> I hate to keep stringing along this PR, but do you think you can squeeze in hiding the `Exclude None` element when there are no `None`-valued items in the current view?

In the extended view?

brimoor:
> > I hate to keep stringing along this PR, but do you think you can squeeze in hiding the `Exclude None` element when there are no `None`-valued items in the current view?
> 
> In the extended view?

yes in https://github.com/voxel51/fiftyone/pull/848#issuecomment-777044873 I'm back in extended view on the sample grid modal

benjaminpkane:
Ok, I'll address the new comments today. I am seeing some weird behavior from the state library (recoil), so I need to work through that.

benjaminpkane:
Ok, comments addressed, I hope. My `None` solution was to add it as a special case in the search results.

I could add an inversion/negation checkbox in the morning so the selected values can be excluded instead.

Screenshot from 2021-02-10 20-22-48
Screenshot from 2021-02-10 20-22-38

list.

brimoor:
Thanks for all the work on this!

I do think it would be really useful to have a field exclusion option. My thoughts on that are in point 2 of https://github.com/voxel51/fiftyone/issues/803, but it occurs to me that selecting one label and excluding another one independently doesn't make sense; it's either all selection or all exclusion. So I think your checkbox idea sounds promising.

Here's my last set of observations. And I'll go ahead and approve now to avoid holding you up :)

1. In expanded sample modal: the scalar fields are clickable but it doesn't do anything. Can you remove the click event (like is true for `tags`, for example)?

Screen Shot 2021-02-11 at 8 57 15 AM

2. Can you make list selection dropdowns auto-hide after something is selected? I think the typical case is to select one thing at a time, so defaulting to auto-hide saves the user time. It's easy to click into the modal again if multiple clicks are desired.

Screen Shot 2021-02-11 at 9 03 12 AM

3. I think if a range filter is applied to a numeric field, this should automatically remove any `None`-valued fields. This is how the equivalent view stage would work, btw, since `a < None < b == False` in MongoDB.
 
Screen Shot 2021-02-11 at 9 00 51 AM

brimoor (approved):
LGTM

PR #849, opened 2021-02-08
Adding support for evaluating classifications

brimoor:
Adds support for evaluating classifications via a new `fiftyone.utils.eval.classification` module. This module currently supports multiclass classification, binary classification, and top-k accuracy computations. In future work, we'll extend to support multilabel classification.

Most of the evaluation work is done via `sklearn.metrics`, and the following are exposed:
- Accuracy/precision/recall/F-beta, both at a dataset- and label-level
- Confusion matrix
- Average precision (binary)
- PR curve (binary)
- ROC curve (binary)
- Top-k classification accuracy

In order to implement this as efficiently as possible, I added a couple new features for batch getting/setting values on samples:
- A new `SampleCollection.values()` aggregation that extracts a "slice" of a dataset, e.g., a list of the values of a chosen field across all samples in a collection
- A new `SampleCollection.set_values()` method that allows for setting a field on each sample in a dataset using efficient bulk write operations

Making the above two changes makes classification evaluation 10x faster on a dataset of 30k samples compared to the same computations implemented by looping over samples and reading + saving them in the most efficient way I could (with `select_fields()`, etc)

**Example classification evaluation:**

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("imagenet-sample")
model = foz.load_zoo_model("resnet50-imagenet-torch")

view = dataset.take(100)
view.apply_model(model, "predictions")

results = foue.evaluate_classifications(view, "predictions", eval_field="eval", classes=classes)

# Prints a classification report
results.print_report()

# Plots a confusion matrix, restricted to 10 classes so it is legible
classes = view.distinct("predictions.label")[:10]

# An `eval` field was populated that indicates whether each sample was correctly classified
view.count_values("eval")

# View samples and filter by `eval` field (when boolean filtering is enabled)
```

**Example values() usage:**

```py
import fiftyone.zoo as foz

# List of uniqueness values on each sample
values = dataset.values("uniqueness")
print(len(values))
print(values[0])

# List of lists of detection labels on each sample
values = dataset.values("predictions.detections.label")
print(len(values))
print(values[0])

# Unwind the labels into a single list
values = dataset.values("predictions.detections[].label")
print(len(values))
print(values[0])
```


ehofesmann (commented):
Very nice! This implementation is really clean.

1) In my opinion, it would make more sense to specify `classes` in the `plot_confusion_matrix` or `print_report` methods rather than in `evaluate_classifications` since it's not even used in `evaluate_classifications` and I'd like to see a report with all results but a confusion matrix with only certain classes

2) This is not necessarily relevant to this PR, but a confidence threshold on classifications feels a bit weird. It makes sense for detections since there are so many we have to filter some out, but for a single classification on every sample, it seems like the default should be to not threshold it.

Thoughts?

ehofesmann:
Also I'd probably call `Slice values` just `Values` in the docs. It's just the raw list of values where every other aggregation like distinct or count are specific operations on these raw values.

brimoor:
> Very nice! This implementation is really clean.
> 
> 1. In my opinion, it would make more sense to specify `classes` in the `plot_confusion_matrix` or `print_report` methods rather than in `evaluate_classifications` since it's not even used in `evaluate_classifications` and I'd like to see a report with all results but a confusion matrix with only certain classes

Yeah good point. I like it

> 2. This is not necessarily relevant to this PR, but a confidence threshold on classifications feels a bit weird. It makes sense for detections since there are so many we have to filter some out, but for a single classification on every sample, it seems like the default should be to not threshold it. 

Yeah I agree; I noted this when introducing a default threshold in https://github.com/voxel51/fiftyone/pull/819. The problem is that one *does* want a default threshold for detection models (and presumably multilabel classifiers), but, as you say, probably *not* for single label classifiers. The trouble is that the models system does not currently declare what type of model you're using.

I'll probably deal with point 2 in a separate PR, because the classification framework needs to be able to handle missing predictions even if we update the model zoo so that classifiers won't produce missing values by default.

ehofesmann:
> > Very nice! This implementation is really clean.
> > 
> > 1. In my opinion, it would make more sense to specify `classes` in the `plot_confusion_matrix` or `print_report` methods rather than in `evaluate_classifications` since it's not even used in `evaluate_classifications` and I'd like to see a report with all results but a confusion matrix with only certain classes
> 
> Yeah good point. I like it
> 
> > 1. This is not necessarily relevant to this PR, but a confidence threshold on classifications feels a bit weird. It makes sense for detections since there are so many we have to filter some out, but for a single classification on every sample, it seems like the default should be to not threshold it.
> 
> Yeah I agree; I noted this when introducing a default threshold in #819. The problem is that one _does_ want a default threshold for detection models (and presumably multilabel classifiers), but, as you say, probably _not_ for single label classifiers. The trouble is that the models system does not currently declare what type of model you're using.
> 
> I'll probably deal with point 2 in a separate PR, because the classification framework needs to be able to handle missing predictions even if we update the model zoo so that classifiers won't produce missing values by default.

Yeah, it seemed like a confidence threshold would have been fine, but in practice, it felt pretty awkward. Maybe it would actually be better to hard code in a confidence threshold to the config of every model (0.3 for det, 0.0 for cls) and default `confidence_threshold` to `None`

brimoor:
> > Very nice! This implementation is really clean.
> > 
> > 1. In my opinion, it would make more sense to specify `classes` in the `plot_confusion_matrix` or `print_report` methods rather than in `evaluate_classifications` since it's not even used in `evaluate_classifications` and I'd like to see a report with all results but a confusion matrix with only certain classes
> 
> Yeah good point. I like it
> 

Point 1 has been addressed. All relevant methods accept an optional `classes` argument.

ehofesmann (approved):
LGTM!

fiftyone/core/dataset.py
```diff
@@ -966,6 +967,31 @@ def _add_samples_batch(self, samples, expand_schema):
 
         return [str(d["_id"]) for d in dicts]
 
+    def _bulk_update(self, sample_ids, updates, ordered=False):
```
brimoor, line 970: @benjaminpkane not sure if you have any reservations about this, or better ideas for how to set a field on each sample of a dataset in cases where the value can't be computed purely in an aggregation pipeline (in which case `set_field()` can be used for this).

I found `bulk_write()` to be significantly faster than the loop I commented out here.

fiftyone/utils/eval/classification.py
```diff
@@ -0,0 +1,499 @@
...
+            eval_field,
+            F().switch(
+                {
+                    (F(gt) == pos_label) & (F(pred) == pos_label): "TP",
+                    (F(gt) == pos_label) & (F(pred) != pos_label): "FN",
+                    (F(gt) != pos_label) & (F(pred) != pos_label): "TN",
+                    (F(gt) != pos_label) & (F(pred) == pos_label): "FP",
+                }
+            ),
+        ).save(eval_field)
```
brimoor, line 129: Here's a fairly complex update that *can* be performed purely as an aggregation pipeline

fiftyone/utils/eval/classification.py
```diff
@@ -0,0 +1,499 @@
...
+            top_k = np.argpartition(_logits, -k)[-k:]
+            _correct = target in top_k
+        else:
+            _correct = False
+
+        correct.append(_correct)
+
+    if eval_field:
+        samples._add_field_if_necessary(eval_field, fof.BooleanField)
+        samples.set_values(eval_field, correct)
```
brimoor, line 201: This one is too complex to be done within an aggregation, so we use `values()` to extract all the values from the DB and then `set_values()` to write the output in bulk.

This is *significantly* faster than the equivalent loop I commented out below.

fiftyone/core/collections.py
```diff
@@ -3345,9 +3496,10 @@ def _build_aggregation(self, aggregations):
     def _process_aggregations(self, aggregations, result, scalar_result):
         results = []
         for idx, agg in enumerate(aggregations):
-            try:
-                results.append(agg.parse_result(result[str(idx)][0]))
-            except:
+            _result = result[str(idx)]
+            if _result:
+                results.append(agg.parse_result(_result[0]))
```
brimoor, line 3501: @benjaminpkane I didn't want to have a naked `try-except` here because I'd like users to get errors if the run an invalid aggregation. I suspect you added this to deal with edge cases like empty fields/samples that may exist in the App. But, I believe I have explicitly handled those correctly here so that no App errors will arise.

But, your eyes on this would be appreciated.

fiftyone/core/collections.py
```diff
@@ -3243,23 +3390,20 @@ def aggregate(self, aggregations, _attach_frames=True):
...
         pipeline = self._pipeline(
             pipeline=facets, attach_frames=_attach_frames
         )
-        try:
-            # pylint: disable=no-member
-            result = next(self._dataset._sample_collection.aggregate(pipeline))
-        except StopIteration:
-            pass
+
+        result = self._dataset._sample_collection.aggregate(pipeline)
```
brimoor, line 3405: @benjaminpkane I couldn't quite figure out why the `try-except` was here previously. It seems like things wouldn't work if an exception was actually caught there anyway?  It seems that the `if not aggregations` check above is all that is required.

I wonder if this was something to do with handling empty datasets in the App.

The reason I was mucking around here is that I wanted errors to occur if someone runs an invalid aggregation on their data. Ideally the app's aggregations wouldn't rely on error catching for empty edge cases.

fiftyone/core/dataset.py
benjaminpkane: No issue here. At some point it would be nice to ensure that only generators are used for `bulk_write` and `insert_many`, but that is a scale issue.

fiftyone/utils/eval/classification.py
benjaminpkane: :rocket:

fiftyone/core/collections.py
benjaminpkane: Yes, I suppose `result` would not be declared. I think you're right to remove it.

fiftyone/core/collections.py
benjaminpkane: Yes, these changes explicitly handle everything. Thanks!

PR #851, opened 2021-02-08
Move default confidence thresholds to zoo model configs

brimoor:
https://github.com/voxel51/fiftyone/pull/819 introduced a default confidence threshold of 0.3 for all models. This had the positive impact of removing many spurious (or even hard-coded, due to model implementation) detections generated by detection models. However, a side effect is that single label classifiers sometimes generate no predictions now.

Fortunately, https://github.com/voxel51/fiftyone/pull/819 conveniently laid some ground work so that confidence thresholds can be applied on a per-model basis directly in their zoo configs.

So, as of this PR, all detection models have a default confidence threshold of 0.3 and other models have no thresholding.

A custom threshold can still be provided via the `apply_model(..., confidence_thresh=)` argument, which will override any behavior from the model's JSON config.

Example of applying every zoo model to a dataset:

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "coco-2017",
    split="validation",
    dataset_name=fo.get_default_dataset_name(),
    shuffle=True,
    max_samples=2,
)

for model_name in foz.list_zoo_models():
    try:
        zoo_model = foz.get_zoo_model(model_name)
        zoo_model.ensure_requirements()
    except:
        # can't run this model with your current setup
        continue

    model = foz.load_zoo_model(model_name, error_level=1)
    label_field = model_name.lower().replace("-", "_").replace(".", "_")

    # can pass `confidence_thresh=` here to set a custom confidence threshold
    dataset.apply_model(model, label_field=label_field)

```


ehofesmann (approved):
LGTM. It may be worth it to finetune the confidence thresholds for each detection model a bit now that we can do that, but at a glance they all look ok at 0.3.

PR #854, opened 2021-02-10
Evaluation API

brimoor:
This PR introduces an API for evaluating classifiers and object detectors.

The basic design is that you call one of the following convenient `evaluate_*()` methods on your dataset or view depending on the type of model you have:

```py
import fiftyone as fo

dataset = fo.load_dataset(...)

view = dataset.crazy_view()

# evaluate detections
results = view.evaluate_detections(gt_field, pred_field, ...)

# evaluate classifications (single-label, for now)
results = view.evaluate_classifications(gt_field, pred_field, ...)
```

Running any evaluation method returns a `results` object, which is an evaluation-method-specific instance of an `EvaluationResults` class. The `ClassificationResults` class was introduced in https://github.com/voxel51/fiftyone/pull/849; this PR introduces `DetectionResults`.

`DetectionResults` can currently report confusion matrices and such, but it cannot compute things like mAP, which I would like to see us add in the very near future.

The notion of an `EvaluationConfig` is also introduced, which enables the user to specify which evaluation method to use and optionally configure any parameters of the evaluation. For example, all of the following are now possible:

```py
# by default, COCO-style evaluation is used
results = view.evaluate_detections(...)

# specify COCO-style evaluation (using default parameter values) via a string `method` argument
# in the future, more methods will be supported
results = view.evaluate_detections(..., method="coco")

#
# customize the COCO-style evaluation
#

from fiftyone.utils.eval.coco import COCOEvaluationConfig

config = COCOEvaluationConfig()
config.iou = 0.5
config.classwise = False

results = view.evaluate_detections(..., config=config)

# customize the COCO-style evaluation using kwargs
# this maintains backwards-compatibility for existing users
results = view.evaluate_detections(..., iou=0.5, classwise=True)
```

Now, if you run any of these methods with no additional arguments, you will get your `results` object, but *no modifications, new fields, etc* will be added to your dataset. If you would like to populate evaluation data on your samples (e.g., TP/FP/FN for detection and binary classification), then you must pass an `eval_key` argument to the relevant evaluation method. This `eval_key` is used to generate fields/embedded fields/etc in which per-sample evaluation data is stored.

The details of the evaluation you ran are stored in your dataset's backing doc and can be accessed and acted on later:

```py
samples.list_evaluations()
# Returns a list of all evaluation keys for the given collection

eval_info = samples.get_evaluation_info(eval_key)
print(eval_info)
# retrieve information about an evaluation you previously ran

samples.delete_evaluation(eval_key)
# Deletes the evaluation results associated with the given evaluation key on the collection

samples.delete_evaluations()
# Deletes all evaluation results from the collection.
```

Running `delete_evaluation(eval_key)` will delete any evaluation results added to your dataset associated with a given `eval_key` (using the evaluation info stored in the dataset to figure out what kind of evaluation it was and how to undo it). The utility here is that the user doesn't need to worry about manually cleaning up whatever information may have been added to their dataset. They just say the word and it is gone!

With respect to detection evaluation, I refactored the existing `evaluate_detections()` method into a generic shell that will run any evaluation method specified by a `DetectionEvaluationConfig`, and moved the COCO-specific evaluation logic into a new `fiftyone.utils.eval.coco` submodule. The abstraction that I introduced here is that an evaluation method is defined by an "image evaluation function" with the following contract:

```py
    def evaluate_image(
        self, sample_or_frame, gt_field, pred_field, eval_key=None
    ):
        """Evaluates the ground truth and predicted objects in an image.

        Args:
            sample_or_frame: a :class:`fiftyone.core.Sample` or
                :class:`fiftyone.core.frames.Frame`
            pred_field: the name of the field containing the predicted
                :class:`fiftyone.core.labels.Detections` instances
            gt_field: the name of the field containing the ground truth
                :class:`fiftyone.core.labels.Detections` instances
            eval_key (None): an evaluation key for this evaluation

        Returns:
            a list of matched ``(gt_label, pred_label, pred_confidence)``
            tuples
        """
        raise NotImplementedError("subclass must implement evaluate_image()")
```

If an `eval_key` is passed, then the evaluation method has permission to add eval info to the sample's fields. Otherwise, it cannot. Regardless of that, the results of the evaluation are returned as a list of `(gt_label, pred_label, iou, pred_confs)` pairs.

COCO-style evaluation fits into this paradigm, and the idea is that most other evaluation techniques will too(?)

> Aside: w.r.t. evaluating videos, `evaluate_detections()` will apply `evaluate_image()` to all image frames by default. For true trajectory-level video evaluation, I think a separate `evaluate_video_detections()` method is called for

Open questions
- ~~Should certain parameters such as `classwise` and `iou` be refactored into the top-level `evaluate_detections()` method? Or will these be irrelevant to certain evaluation techniques and thus kept as `EvaluationConfig` parameters?~~

Other notable changes
- ~~I changed the default for COCO-style evaluation to `classwise=False`, because I think it makes sense to allow detections of different classes to be matched, so that the classification of the detector can be evaluated by default~~
- I made some improvements to `clone/rename/clear/deletion` of dataset/view sample/frame fields, specifically allowing multiple fields to be modified in a single transaction. I also improved the backend implementation to support modifying nested array fields in most cases

Example end-to-end usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

DATASET_NAME = "eval-detections-test"
EVAL_KEY = "eval-det"

if fo.dataset_exists(DATASET_NAME):
    dataset = fo.load_dataset(DATASET_NAME)
else:
    dataset = foz.load_zoo_dataset("quickstart").clone(DATASET_NAME)
    dataset.persistent = True

    "predictions",
    gt_field="ground_truth",
    eval_key=EVAL_KEY,
)

results.print_report()

# Restrict evaluation to 10 classes, for brevity
classes = dataset.distinct("predictions.detections.label")
classes = classes[:9] + [results.missing]

print(dataset.list_evaluations())
print(dataset.get_evaluation_info(EVAL_KEY))
dataset.delete_evaluation(EVAL_KEY)
print(dataset.list_evaluations())

#
# Customized detection evaluation
#

from fiftyone.utils.eval.coco import COCOEvaluationConfig

config = COCOEvaluationConfig()
config.iou = 0.5
config.classwise = False

    "predictions",
    gt_field="ground_truth",
    eval_key=EVAL_KEY,
    config=config,
)

results.metrics()

print(dataset.list_evaluations())
print(dataset.get_evaluation_info(EVAL_KEY))
fo.delete_evaluations(dataset)
print(dataset.list_evaluations())
```


fiftyone/utils/eval/coco.py
```diff
@@ -5,326 +5,180 @@
...
-        Ground truth:   detections.<pred_field>_eval
-        Predictions:    detections.<gt_field>_eval
-
-    Dictionaries are also added to each individual
-    :class:`fiftyone.core.labels.Detection` instance in the fields listed
-    below; these fields tabulate the IDs of the matching ground
-    truth/prediction for the detection at the specified IoU::
+    Args:
+        iou (0.75): the IoU threshold to use to determine matches
+        classwise (False): whether to only match objects with the same class
```
brimoor, line 22: @ehofesmann I'm curious to get your thoughts on whether `classwise` should be True or False by default. Does COCO-style evaluation use `classwise=True` by default?

It's useful to be able to analyze classification errors for detectors, but with `classwise=False` then matching objects aren't always "TP", which does mean that our TP/FP/FN indicators have an asterisk.

ehofesmann: I think that it should be `True` by default since that let's you evaluate the model as a whole and is what pycocotools does. Though I added `classwise` because setting it to `False` is really useful to evaluate just the detections. 

This definitely seems like one point that would need to be addressed in our "proper evaluation and metric" initiative we have with Jordi since you should incorporate the detection and classification abilities of a model into a metric separately.

ehofesmann (changes requested):
First off, this is beautiful! The idea of returning `results` is what I was struggling with when trying to contain everything in sample-level evaluation fields, this is a great solution to that. Also using `dataset.info` to store previous evaluation information is really smart.

One change request and two discussion points:

1) I would definitely request this be accessible from the top level `fo.evaluate_...` instead of having to import `fiftyone.utils.eval` unless you have a good reason not to.

2) I'm conflicted when it comes to using `configs` or just letting parameters like `classwise` and `iou` be kwargs at the top-level. I'm leaning towards making them kwargs since that's easiest for a new user to understand without having to learn anything new. Just need to look at the docs for all the existing kwargs and set them to what you want. But when we get to more complex evaluations that have very specific parameters only applicable to one eval type, having kwargs for those parameters could get messy since they're not always applicable. Maybe there's a middleground for things like `iou` that are used by any detection evaluation to be `kwargs` but if a parameter isn't useful to more than one model it's specified by the config. Can you give me some sage wisdom about your thoughts on this Brian?

3) I'm worried the current implementation of `evaluate_image` and then the result aggregation is a bit restrictive.

3a) I'm not totally sure of the specifics, but I think that Open Images evaluation also uses image-level labels, so right there, you would already need to be able to pass in a variable number of `gt` fields to `evaluate_image`. 

3b) Right now, the only aggregations performed seem to be counting tp/fp/fn. Adding mAP right now would be difficult since we also need confidences to compute PR curves. This makes me think that there should also be customizable `aggregate_image_results` functions similar to `evaluate_images` that will let us write different ways of doing things like computing mAP (since there are a few ways that could be done) etc.

fiftyone/utils/eval/coco.py
brimoor: ok thanks I reverted to `True` because that's what COCO uses

brimoor:
> First off, this is beautiful! The idea of returning `results` is what I was struggling with when trying to contain everything in sample-level evaluation fields, this is a great solution to that. Also using `dataset.info` to store previous evaluation information is really smart.
> 
> One change request and two discussion points:
> 
> 1. I would definitely request this be accessible from the top level `fo.evaluate_...` instead of having to import `fiftyone.utils.eval` unless you have a good reason not to.

yeah I was planning to this, and its now done! The PR description is updated to show that in use.

> 2. I'm conflicted when it comes to using `configs` or just letting parameters like `classwise` and `iou` be kwargs at the top-level. I'm leaning towards making them kwargs since that's easiest for a new user to understand without having to learn anything new. Just need to look at the docs for all the existing kwargs and set them to what you want. But when we get to more complex evaluations that have very specific parameters only applicable to one eval type, having kwargs for those parameters could get messy since they're not always applicable. Maybe there's a middleground for things like `iou` that are used by any detection evaluation to be `kwargs` but if a parameter isn't useful to more than one model it's specified by the config. Can you give me some sage wisdom about your thoughts on this Brian?

I figured you'd say this, I guess I was subconsciously asking so that you'd give me permission to expose those parameters at the top-level. That change is now in; `evaluate_detections()` always takes `iou` and `classwise` as parameters, which are passed to the actual evaluation method being used (which I suppose would ignore these parameters if they are not relevant to their implementation). There is still the option of passing a `config` for things like changing the name of the `iscrowd` COCO attribute, for example. Or they can be passed via `**kwargs` and that will work too (but the parameters don't need to be documented in the `evaluate_detections()` method docstring).

> 3. I'm worried the current implementation of `evaluate_image` and then the result aggregation is a bit restrictive.
> 
> 3a) I'm not totally sure of the specifics, but I think that Open Images evaluation also uses image-level labels, so right there, you would already need to be able to pass in a variable number of `gt` fields to `evaluate_image`.

Yeah good point. I updated the interface to take the entire `Sample` or `Frame` so that the method can access other stuff if it wants to. I also updated the interface to return the prediction confidences.

> 3b) Right now, the only aggregations performed seem to be counting tp/fp/fn. Adding mAP right now would be difficult since we also need confidences to compute PR curves. This makes me think that there should also be customizable `aggregate_image_results` functions similar to `evaluate_images` that will let us write different ways of doing things like computing mAP (since there are a few ways that could be done) etc.

So I think the cleanest way to handle this is to introduce subclasses of `DetectionResults` that expose new metrics. Like how `BinaryClassificationResults` subclasses from `ClassificationResults` and exposes PR curves and classification AP. I think we can safely introduce this in a separate PR once we get an implementation of mAP in.

brimoor:
@benjaminpkane I believe the windows tests are failing here because I un-skipped a few unit tests that involve cloning stuff (which was modified in this PR). I guess I have to re-skip them to get tests to pass?

I was crossing my fingers that the Windows issue would have resolved itself by now. (is it related to the fact that the cloning code is using new MongoDB version features?) 🤷‍♂️

benjaminpkane:
Yes, until I dig into the service tests/framework.

Ideally, I'd like to leverage a comms library for the service stuff. Any implementations we use will likely never be robust enough.

benjaminpkane:
```
is it related to the fact that the cloning code is using new MongoDB version features?
```
No, I'm quite sure this isn't the issue.

ehofesmann (approved):
LGTM!

PR #856, opened 2021-02-11
Defaulting only_matches to True

brimoor:
Changes the default value of the `only_matches` parameter of the `Filter*` view stages from `False` to `True`.

My reasoning here is that:
- The App has been using `True` by default for awhile, so this change is consistent with that
- It's more likely, in my opinion (and certainly in my experience as a user) that the user wants to hide empty samples when performing a search rather than seeing all the empty samples


benjaminpkane (approved):
LGTM, I agree `True` is the better default value.

PR #857, opened 2021-02-11
Release v0.7.3

benjaminpkane:


benjaminpkane:
There is an issue with numeric field sidebar filters. Will post a fix here today.

brimoor:
Hey @benjaminpkane what do you think about tweaking the look and feel of the `Exclude values` toggle slightly? Currently this option feels like its given more visual attention than it deserves and is a bit distracting.

I'm thinking:

- Only show it when labels are actually selected (otherwise it has no effect)
- Rather than making it a full size checkbox with large label, what about minifying this a bit, perhaps by making it a simple hyperlink option like the `reset` text that says something like `exclude instead?` (if we're currently showing only matches) or `filter instead?` (if we're currently excluding). This could be positioned below the tags, where it is now, or maybe could go in between `Labels` and `reset` in the header? (but only when at least one label is selected)

Screen Shot 2021-02-12 at 2 03 54 PM

benjaminpkane:
Right, I'll adjust.

benjaminpkane:
> * Only show it when labels are actually selected (otherwise it has no effect)
> * Rather than making it a full size checkbox with large label, what about minifying this a bit, perhaps by making it a simple hyperlink option like the `reset` text that says something like `exclude instead?` (if we're currently showing only matches) or `filter instead?` (if we're currently excluding). This could be positioned below the tags, where it is now, or maybe could go in between `Labels` and `reset` in the header? (but only when at least one label is selected)

Ok, updated.
Screenshot from 2021-02-12 17-11-56
Screenshot from 2021-02-12 17-10-58

brimoor:
FYI I committed a variation on the select|exclude UI that I like a bit better
Screen Shot 2021-02-13 at 1 14 21 AM

brimoor:
One bug I found while testing is that it seems that clicking `reset` on a confidence slider of a label does not get rid of the "active filter" highlight on the label field

Screen Shot 2021-02-13 at 1 12 19 AM

benjaminpkane:
> One bug I found while testing is that it seems that clicking `reset` on a confidence slider of a label does not get rid of the "active filter" highlight on the label field (i.e., clicking reset below does nothing)

Thanks, addressed.

brimoor (approved):
LGTM

PR #858, opened 2021-02-11
Adding support for cloning and saving views into video datasets

brimoor:
Resolves an outstanding `@todo` in the codebase: views that filter/modify frame-level labels can now be `.save()`d and `.clone()`d.

Other improvements
- Adds a `Frames.first()` method to get the first frame of a sample (`sample.frames.first()`)

Also resolves some outstanding bugs:
- Fixes a bug that had caused `frames` to disappear from `Sample.__repr__()`
- Fixes a bug that caused DatasetViews that modify frame labels from appropriately loading their changes in the `frames` field
- Fixes a bug that prevented `Sample.__repr__()` from working for video samples that aren't in datasets
- Fixes some bugs in managing in-memory `Frame` instances (`fiftyone.core.odm.mixin.Frame` needed `_reload_docs()` and `_reset_docs()` implementations)
- Before it disappeared, the `frames` entry in `Sample.__repr__()` was erroneously printed as a string; I fixed that and also am trying a new way to render the frames:

```
<Sample: {
    'id': '60215bc5fcd90c52400cc2d1',
    'media_type': 'video',
    'filepath': '/Users/Brian/fiftyone/quickstart-video/data/Ulcb3AjxM5g_053-1.mp4',
    'tags': BaseList([]),
    'metadata': None,
    'frames': <Frames: 120>,
}>
```

Example usage of `.save()` and `.clone()` with video dataset views:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# `dataset` contains vehicles, people, and road signs
print(dataset.count_values("frames.ground_truth_detections.detections.label"))

view = dataset.filter_labels(
    "frames.ground_truth_detections", F("label") == "vehicle"
)

# `view` only contains vehicles
print(view.count_values("frames.ground_truth_detections.detections.label"))

dataset2 = view.clone()

# `dataset2` only contains vehicles
print(dataset2)
print(dataset2.first())
print(next(dataset2.first().frames.values()))
print(dataset2.count_values("frames.ground_truth_detections.detections.label"))

view.save()

# now `dataset` only contains vehicles
print(dataset)
print(dataset.first())
print(next(dataset.first().frames.values()))
print(dataset.count_values("frames.ground_truth_detections.detections.label"))
```


brimoor:
@benjaminpkane I'm realizing there's something wrong with accessing the frames of `SampleView`s:

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.filter_labels(
    "frames.ground_truth_detections", F("label") == "vehicle"
)

print(dataset.count_values("frames.ground_truth_detections.detections.label"))
print(view.count_values("frames.ground_truth_detections.detections.label"))

print(dataset.first().frames.first())

# ERROR: shouldn't contain non-vehicles, but does
print(view.first().frames.first())
```

Any ideas?

brimoor:
More generally, I wonder if we can find a clean way to decouple loading the `Frames` of a video `Sample` from loading the sample itself. doing `dataset.first()` on the `quickstart-video` dataset takes a non-trivial amount of time, which is not ideal in situations where the user doesn't want to access the frames.

I'd rather make frame loading JIT if possible (noting that, when `DatasetView`s are involved, the aggregation pipeline may have internally attached and then detached frames in order to compute which samples are in the view):

```py
for sample in video_dataset:
    # `sample` loads fast because `frames` haven't been loaded yet
    print(sample.id)
    
    # now frames are loaded
    print(len(sample.frames))
```

benjaminpkane:
Yes, I have come across this. Peeking into things right now, not sure yet if there is a quick band-aid for this.

brimoor:
> Yes, I have come across this. Peeking into things right now, not sure yet if there is a quick band-aid for this.

Ok cool; I think there may be an easy fix... maybe there's something wrong with the "replacements" code in `Frames` that always overwrites existing data... The aggregations show the correct values, which suggests that `_pipeline(attach_frames=True)` should be doing the right thing.

I'd like to get to https://github.com/voxel51/fiftyone/pull/858#issuecomment-777752212 eventually, but that one is definitely not a quick bandaid :)

benjaminpkane:
Ok, I see the issue. https://github.com/voxel51/fiftyone/pull/858/commits/4b00351280c01de796809a1b1943d4e25b65c2ec does not test for it, will adjust it when I post a fix.

benjaminpkane:
Ok, I added a band-aid. `Frame` lacking a corresponding `FrameView` is the crux of the issue.

JIT is certainly attainable, or at least speed ups when frames aren't needed. Shouldn't be that much code, I'll look into it.

benjaminpkane (approved):
LGTM!

fiftyone/core/frame.py
```diff
@@ -361,7 +383,9 @@ class Frame(Document):
     """
 
     # Instance references keyed by [collection_name][sample_id][frame_number]
-    _instances = defaultdict(lambda: defaultdict(weakref.WeakValueDictionary))
+    _instances = defaultdict(
```
brimoor, line 386: @benjaminpkane doesn't this break the methods like `_rename_fields()` below? They don't index by `view`.

`SampleView`s aren't singletons, so can the behavior in `from_doc` below be to always create a new `Frame` instance if the `Frame` is being loaded in a view?

PR #864, opened 2021-02-14
Calculate COCO mAP

ehofesmann:
This PR expands on `evaluate_detections` to compute pycocotools style mAP and introduces a `COCODetectionResults` to print the mAP and plot PR curves from the evaluation.

```python
import fiftyone.zoo as foz
import fiftyone as fo

results = fo.evaluate_detections(dataset, "predictions", compute_mAP=True)

print(results.mAP())
# 0.3957...
# pycocotools also computes an mAP of 0.396

results.plot_pr_curves(classes=["person"])
```

image

I haven't forced a `classwise=False` calculation even though it would be nice so that we can get more interesting confusion matrices. The MSCOCO evaluation uses `True` by default, I'm thinking just let the user change it if they want.


fiftyone/utils/eval/coco.py
```diff
@@ -231,3 +473,113 @@ def _iscrowd(detection):
...
+                precision=precision, recall=recall
+            )
+            label = "AP = %.2f, class = %s" % (avg_precision, c)
+            display.plot(ax=ax, label=label, **kwargs)
+            plt.show(block=block)
+            displays.append(display.ax_)
+
+        return displays
+
+    def metrics(self, classes=None, average="micro", beta=1.0):
```
brimoor, line 558: I think its cleaner to leave `metrics()` alone and return mAP in a separate method, similar to `BinaryClassificationResults.average_precision()`

fiftyone/utils/eval/coco.py
```diff
@@ -231,3 +473,113 @@ def _iscrowd(detection):
...
+        for c in classes:
+            class_ind = self.classes.index(c)
+            precision = np.mean(self.precision[:, class_ind], axis=0)
+            recall = self.recall
+            avg_precision = np.mean(precision)
+            display = skm.PrecisionRecallDisplay(
+                precision=precision, recall=recall
+            )
+            label = "AP = %.2f, class = %s" % (avg_precision, c)
+            display.plot(ax=ax, label=label, **kwargs)
```
brimoor, line 552: Shouldn't there be a single `ax` return value that is the axes in which all PR curves are plotted?

fiftyone/utils/eval/coco.py
```diff
@@ -134,16 +330,36 @@ def _coco_evaluation(gts, preds, eval_key, config):
...
 
         # Highest confidence predictions first
-        preds = sorted(preds, key=lambda p: p.confidence or -1, reverse=True)
+        preds = sorted(preds, key=lambda p: p.confidence or -1, reverse=True)[
+            :100
+        ]
         objects["preds"] = preds
 
+        # Sort ground truth so crowds are last
+        gts_crowd = []
```
brimoor, line 339: you can write this as:

```py
gts = sorted(gts, key=lambda d: iscrowd(d)])
```

fiftyone/utils/eval/coco.py
```diff
@@ -134,16 +330,36 @@ def _coco_evaluation(gts, preds, eval_key, config):
         preds = objects["preds"]
 
         # Highest confidence predictions first
-        preds = sorted(preds, key=lambda p: p.confidence or -1, reverse=True)
+        preds = sorted(preds, key=lambda p: p.confidence or -1, reverse=True)[
```
brimoor, line 333: Although COCO may limit to 100 predictions, perhaps we don't want to do this? Or, at least, we should make this a parameter that the user can configure.

Certainly for the non-mAP part of this calculation, I think we should never cap the number of the predictions in this list, because the typical user of `evaluate_detections()` will expect that all of their predictions are evaluated, certainly for TP/FP/FN counts, for example.

fiftyone/utils/eval/coco.py
```diff
@@ -22,11 +36,13 @@ class COCOEvaluationConfig(DetectionEvaluationConfig):
         classwise (None): whether to only match objects with the same class
             label (True) or allow matches between classes (False)
         iscrowd ("iscrowd"): the name of the crowd attribute
+        iou_threshs([0.5::0.05::0.95]): 10 IoU thresholds used to compute mAP
     """
 
     def __init__(self, iscrowd="iscrowd", **kwargs):
         super().__init__(**kwargs)
         self.iscrowd = iscrowd
+        self.iou_threshs = [x / 100 for x in range(50, 100, 5)]
```
brimoor, line 45: you need to parse `iou_threshs` from `__init__()` in case the user provided a customized value

fiftyone/utils/eval/coco.py
```diff
@@ -22,11 +36,13 @@ class COCOEvaluationConfig(DetectionEvaluationConfig):
         classwise (None): whether to only match objects with the same class
             label (True) or allow matches between classes (False)
         iscrowd ("iscrowd"): the name of the crowd attribute
+        iou_threshs([0.5::0.05::0.95]): 10 IoU thresholds used to compute mAP
```
brimoor, line 39: nit: how about this docstring:

```
iou_threshs ([0.5::0.05::0.95]): a list of IoU thresholds to use when computing mAP and PR curves
```

brimoor (changes requested):
What do you think about *not* computing mAP values by default? I'm concerned that it will slow down evaluation quite a bit for large datasets, and may not be relevant to users in all cases (which is problematic while COCO-style evaluation is our only evaluation option).

Here's couple ideas that come to mind on how to handle this:

(a) Provide an optional `COCOEvaluationConfig` parameter that controls whether IoU sweeps are performed so that mAP and PR curves can be generated. The user has to opt-in to this in order to have the relevant information populated

(b) Never perform IoU sweeps inside `evaluate_detections()`; instead, pass the necessary parameters through to the `COCODetectionResults` object and let it perform the IoU sweeps only if the user calls a method like `mAP()` or `plot_pr_curves()` on the results object

I'm torn on how to best proceed here... since I see some cons to both approaches:

(a) If the user *doesn't* opt-in to IoU sweeps, then the `mAP()` and `plot_pr_curves()` methods of their results object will have to raise errors. Which indicates that the results object may need to be able to compute this information *anyway* as in (b)

(b) This will require the `SampleCollection` and other information be passed to the results object. This is not technically an issue, but it does break an existing convention that results objects contain all the necessary information to generate eval results without needing to touch the user's dataset again. Many of the methods of `EvaluationResults` do perform computations, though, so maybe this is okay...

What are your thoughts?

fiftyone/utils/eval/coco.py
brimoor: For example, the docstring mentions "classification results", but mAP is not a classification concept.

fiftyone/utils/eval/coco.py
brimoor: Perhaps the best approach is to convert this value `100` to a config parameter that can be set to `None` by the user if they do not want any limiting to be done.

Or, if you would rather think of IoU sweeps as *exactly* following COCO style, then leave it hard-coded to 100 but *only* for the IoU sweep portion of the evaluation.

Another COCO-specific parameter is `iscrowd`, which the evaluation routine provides support for customizing. And, importantly, evaluation will work if the user's data doesn't have a crowd parameter populated.

ehofesmann:
> What do you think about _not_ computing mAP values by default? I'm concerned that it will slow down evaluation quite a bit for large datasets, and may not be relevant to users in all cases (which is problematic while COCO-style evaluation is our only evaluation option).
> 
> Here's couple ideas that come to mind on how to handle this:
> 
> (a) Provide an optional `COCOEvaluationConfig` parameter that controls whether IoU sweeps are performed so that mAP and PR curves can be generated. The user has to opt-in to this in order to have the relevant information populated
> 
> (b) Never perform IoU sweeps inside `evaluate_detections()`; instead, pass the necessary parameters through to the `COCODetectionResults` object and let it perform the IoU sweeps only if the user calls a method like `mAP()` or `plot_pr_curves()` on the results object
> 
> I'm torn on how to best proceed here... since I see some cons to both approaches:
> 
> (a) If the user _doesn't_ opt-in to IoU sweeps, then the `mAP()` and `plot_pr_curves()` methods of their results object will have to raise errors. Which indicates that the results object may need to be able to compute this information _anyway_ as in (b)
> 
> (b) This will require the `SampleCollection` and other information be passed to the results object. This is not technically an issue, but it does break an existing convention that results objects contain all the necessary information to generate eval results without needing to touch the user's dataset again. Many of the methods of `EvaluationResults` do perform computations, though, so maybe this is okay...
> 
> What are your thoughts?

Man, that's a tough problem. I would prefer to have the user opt-in to running evaluation with IoU sweeps if they want `mAP()` and `plot_pr_curves()`. I don't like the idea of results being tied to a `SampleCollection`. If we were to go that way, it may make more sense to make results part of a `SampleCollection`.

Though I also don't like the idea of having `mAP()` raise errors based on what was passed into the `COCODetectionResults`. I feel like a solution may be to just return `DetectionResults` unless you opt-in to computing mAP, which would then return `COCODetectionResults` that lets you access `mAP()` and `plot_pr_curves()`.

brimoor:
> > What do you think about _not_ computing mAP values by default? I'm concerned that it will slow down evaluation quite a bit for large datasets, and may not be relevant to users in all cases (which is problematic while COCO-style evaluation is our only evaluation option).
> > Here's couple ideas that come to mind on how to handle this:
> > (a) Provide an optional `COCOEvaluationConfig` parameter that controls whether IoU sweeps are performed so that mAP and PR curves can be generated. The user has to opt-in to this in order to have the relevant information populated
> > (b) Never perform IoU sweeps inside `evaluate_detections()`; instead, pass the necessary parameters through to the `COCODetectionResults` object and let it perform the IoU sweeps only if the user calls a method like `mAP()` or `plot_pr_curves()` on the results object
> > I'm torn on how to best proceed here... since I see some cons to both approaches:
> > (a) If the user _doesn't_ opt-in to IoU sweeps, then the `mAP()` and `plot_pr_curves()` methods of their results object will have to raise errors. Which indicates that the results object may need to be able to compute this information _anyway_ as in (b)
> > (b) This will require the `SampleCollection` and other information be passed to the results object. This is not technically an issue, but it does break an existing convention that results objects contain all the necessary information to generate eval results without needing to touch the user's dataset again. Many of the methods of `EvaluationResults` do perform computations, though, so maybe this is okay...
> > What are your thoughts?
> 
> Man, that's a tough problem. I would prefer to have the user opt-in to running evaluation with IoU sweeps if they want `mAP()` and `plot_pr_curves()`. I don't like the idea of results being tied to a `SampleCollection`. If we were to go that way, it may make more sense to make results part of a `SampleCollection`.
> 
> Though I also don't like the idea of having `mAP()` raise errors based on what was passed into the `COCODetectionResults`. I feel like a solution may be to just return `DetectionResults` unless you opt-in to computing mAP, which would then return `COCODetectionResults` that lets you access `mAP()` and `plot_pr_curves()`.

Okay yeah I think this sounds good to me then:
- If a user does not request IoU sweeps, then they get a `DetectionResults` instance, even though "COCO-style" evaluation was performed (without the IoU sweeps, there's not much that's COCO-specific going on)
- If a user does request IoU sweeps, they get a `COCODetectionResults` object as you've implemented that contains the precision and recall numbers

ehofesmann:
Alrighty @brimoor , I addressed your comments and added a `compute_mAP` argument to the `COCOEvaluationConfig` that defaults to False so that `evaluate_samples` returns a `DetectionResults` object. It will return `COCODetectionResults` if `compute_mAP` is True.

fiftyone/utils/eval/coco.py
```diff
@@ -22,11 +32,23 @@ class COCOEvaluationConfig(DetectionEvaluationConfig):
...
+            computing mAP and PR curves
     """
 
-    def __init__(self, iscrowd="iscrowd", **kwargs):
+    def __init__(
+        self, iscrowd="iscrowd", compute_mAP=False, iou_threshs=None, **kwargs
+    ):
         super().__init__(**kwargs)
+
+        if compute_mAP and iou_threshs is None:
```
brimoor, line 46: I tweaked this so that `iou_threshs` is only populated if `compute_mAP == True`. Otherwise, the `get_evaluation_info()` would store a list of `iou_threshs` even when no sweep was performed, which is misleading

fiftyone/utils/eval/coco.py
```diff
@@ -89,39 +111,295 @@ def evaluate_image(
...
+        return COCODetectionResults(
+            matches, precision, recall, iou_threshs, classes, missing=missing
+        )
+
+
+class COCODetectionResults(DetectionResults):
+    """Class that stores the results of a COCO detection evaluation.
+
+    Args:
+        matches: a list of ``(gt_label, pred_label, iou, pred_confidence)``
```
brimoor, line 251: nit: the `matches` argument needed to be mandatory because the base class can't function without it

fiftyone/utils/eval/coco.py
```diff
@@ -89,39 +111,295 @@ def evaluate_image(
...
+            matches. Either label can be ``None`` to indicate an unmatched
+            object
+        precision: an array of precision values of shape
+            ``num_iou_threshs x num_classes x num_recall``
+        recall: an array of recall values
+        iou_threshs: the list of IoU thresholds
```
brimoor, line 257: I passed `iou_threshs` through here as well since the `precision` array's size depends on it (so this would be interesting information to the user to access via `results.iou_threshs`

fiftyone/utils/eval/coco.py
```diff
@@ -89,39 +111,295 @@ def evaluate_image(
...
+            the mAP float
+        """
+        if classes is not None:
+            class_inds = [self.classes.index(c) for c in classes]
+            classwise_AP = self._classwise_AP[class_inds]
+        else:
+            classwise_AP = self._classwise_AP
+
+        existing_classes = classwise_AP > -1
+        if not any(existing_classes):
```
brimoor, line 326: you wrote `len(existing_classes) == 0` here, but I think you meant `not any(existing_classes)` right?

fiftyone/utils/eval/detection.py
```diff
@@ -182,14 +192,50 @@ def evaluate_image(
...
-            eval_key (None): an evaluation key for this evaluation
+            eval_key (None): the evaluation key for this evaluation
 
         Returns:
             a list of matched ``(gt_label, pred_label, iou, pred_confidence)``
             tuples
         """
         raise NotImplementedError("subclass must implement evaluate_image()")
 
+    def generate_results(
```
brimoor, line 203: I renamed this to `generate_results()` to clarify what it does. Otherwise it can be confusing to see that `evaluate_image` is called only to call `evaluate_samples` later.

Also, since the main `evaluate_detections()` method calls it regardless of the evaluation method being run, you need to define it on the base `DetectionEvaluation` class (even though, yes, right now only COCO method exists)

brimoor (approved):
LGTM. It runs and matches pycocotools.

I have a couple comments about the behavior of `generate_results()` when the user specifies a `classes` and/or `missing` parameter, but we can take those offline

fiftyone/utils/eval/coco.py
ehofesmann: Ah yup!

fiftyone/utils/eval/detection.py
ehofesmann: Good call!

PR #866, opened 2021-02-15
Adding support for extracting values from nested array fields

brimoor:
The `values()` aggregation previously did not support cases where there were two or more array fields involved, for example extracting the labels for all detections of frame-level fields. This PR adds support for this.

This arose as I was working on evaluation for video datasets, where I wanted to write aggregations to extract all classifications or detections at the frame-level.

mea culpa: I haven't added aggregation tests for frame-level fields, but there are fairly comprehensive tests for sample-level fields, and this PR passes those. So at best this PR adds full support for video datasets, and at worst it replaces existing video bugs with new ones.

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

with fo.ProgressBar() as pb:
    for sample in pb(dataset):
        for frame in sample.frames.values():
            label = random.choice(["sunny", "cloudy", "rainy"])
            frame["weather"] = fo.Classification(label=label)

        sample.save()

def check(dataset, path, expr):
    print(dataset.distinct(path, expr=expr))
    print(dataset.count_values(path, expr=expr))

    values = dataset.values(path, expr=expr)
    print(len(values))  # number of samples
    print(len(values[0]))  # number of frames in first sample
    print(values[0][0])  # values for first frame of first sample

# Extract frame-level classifications
check(dataset, "frames.weather.label", None)
check(dataset, "frames.weather.label", F().upper())

# Extract frame-level object labels
check(dataset, "frames.ground_truth_detections.detections.label", None)
check(dataset, "frames.ground_truth_detections.detections.label", F().upper())
```

### DIFF
--- a/fiftyone/core/aggregations.py
+++ b/fiftyone/core/aggregations.py
@@ -77,8 +77,8 @@ class Aggregation(object):
         """
         raise NotImplementedError("subclasses must implement default_result()")
 
-    def _parse_field_name(self, sample_collection, auto_unwind=True):
-        return _parse_field_name(
+    def _parse_field_and_expr(self, sample_collection, auto_unwind=True):
+        return _parse_field_and_expr(
             sample_collection, self._field_name, auto_unwind, self._expr
         )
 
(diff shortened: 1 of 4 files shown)


benjaminpkane (approved):
LGTM

PR #868, opened 2021-02-17
Adding evaluation to user guide

brimoor:
Screen Shot 2021-02-17 at 11 36 08 AM


benjaminpkane (approved):
Reads great! I definitely have a better sense of things now.

PR #869, opened 2021-02-17
Adding support for gracefully skipping samples whose metadata cannot be computed

brimoor:
```py
import fiftyone as fo
import fiftyone.zoo as foz

# Test valid samples

dataset.compute_metadata()
dataset.compute_metadata()  # already computed, nothing happens
dataset.compute_metadata(overwrite=True)  # force overwrite

# Test invalid samples

dataset = fo.Dataset()
dataset.add_samples([fo.Sample(filepath=str(idx)) for idx in range(100)])

dataset.compute_metadata()  # warns that metadata couldn't be computed
dataset.compute_metadata(skip_failures=False)  # raises error
```


benjaminpkane (approved):
LGTM

Issue #817, opened 2021-01-26
[BUG] Inconsistent JSON in expanded sample view


benjaminpkane:
Right now there at least to situations in the expanded sample view where what is displayed by the JSON is not consistent with what is displayed by player51.

* Frame labels are not provided whatsoever for video samples, the `first_frame` data is displayed instead
* Selected or excluded objects are not filtered in the JSON

brimoor:
https://github.com/voxel51/fiftyone/issues/672 is a related issue here.

PR #877, opened 2021-02-19
Issues batch

### LINKED ISSUE #823: [FR] Always refresh App when session dataset/view is modified
Currently the App will not refresh if `session.dataset` or `session.view` are reset to the current dataset or view. But, the underlying dataset may have changed, so it would be desirable for the App to reload even if the session can't tell if something changed.

**Note:** you must copy + paste these commands in two separate chunks to see the bug. If you do it in one chunk, the session catches the updated data by the time the App loads.

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("cifar10", split="test", max_samples=500)

```

```py
sample = dataset.first()
sample.ground_truth.label = "CHANGED"
sample.save()

# None of these reload the App
session.dataset = dataset
session.view = None
session.refresh()

# pressing ctrl+R in the browser does reload the App, showing the changed label
```

### LINKED ISSUE #859: [FR] Decouple loading video samples from their frames?
_Originally posted by @brimoor in https://github.com/voxel51/fiftyone/issues/858#issuecomment-777752212_

I'd like to find a clean way to decouple loading the `Frames` of a video `Sample` from loading the sample itself. Currently we have this behavior:

```py
import fiftyone.zoo as foz

# takes a non-trivial amount of time since the frames are always 
# attached, even if I don't want them
print(dataset.first())
```

I'd rather make frame loading JIT if possible (noting that, when `DatasetView`s are involved, the aggregation pipeline may have internally attached and then detached frames in order to compute which samples are in the view):

```py
for sample in video_dataset:
    # `sample` loads fast because `frames` haven't been loaded yet
    print(sample.id)
    
    # now frames are loaded
    print(len(sample.frames))
```

### LINKED ISSUE #874: [FR] Add background to X? icons in view bar
Small UI request: it would be nice if the `X ?` icons in the view bar had a solid background so that they don't become cluttered with long view stages.

A fancy way to do this could be to introduce a transparency gradient on the left-hand side of this background so that the overhanging view stages gradually disappeared, indicating that there's something to see back there if one scrolls.

Screen Shot 2021-02-18 at 12 58 55 PM

### LINKED ISSUE #876: [BUG] Hiding selected objects does not work in App
In `fiftyone>=0.7.3`, options such as `Hide selected` and `Hide others` in expanded sample mode of the App do not cause the objects to actually disappear in the media player.

Screen Shot 2021-02-18 at 2 32 53 PM

benjaminpkane:
Issues
* Resolves #874
* Resolves #859
* Resolves #876
* Resolves #823

I began work on #817 but I will finish in another branch. It deserves its own PR.

Screenshot from 2021-02-19 12-16-34


brimoor:
> Issues
> 
> * Resolves #874

^I love this as a way of hinting to the user that they need to scroll. One related issue, though, is that its not possible to delete the last stage in a long chain because its `x` is behind the clear view `x`. Is there a way to keep the transparency but also make the stage component stop earlier so that the `x` is always clickable?

> * Resolves #859

^ awesome!

> * Resolves #876

^ works, thanks

> * Resolves #823

^ I'm seeing some weird behavior when testing this one with the code in the issue. If I run any *one* of the reload commands, nothing happens. But, if I run two or more of them, then the App refreshes. Can you reproduce this?

brimoor:
Related to the scrolling behavior in #874, any chance you can do something to address the user feedback that it can be unclear that our vertical scroll boxes have more contents because they don't look "scrollable"?

One idea was to make them `n + 0.5` units long if `n >= default height`. Or, perhaps make them `n + 1` units long and introduce the same transparency gradient from the view bar?

This gives me an idea: a consistent rule for both the view bar and the list boxes could be:
- if the length of the list exceeds the end of the current visible position, introduce a bit of transparency at the end of the component (and, I guess, the beginning of the component too). The transparency would disappear when one reaches the beginning or end of the list.

Under this strategy, the view stages would never actually appear below the `X?` icons.

If this is too much, please cry uncle :)

benjaminpkane:
> * if the length of the list exceeds the end of the current visible position, introduce a bit of transparency at the end of the component (and likewise for the beginning of the component too if the starting position is > 0%). The transparency would disappear when one reaches the beginning or end of the list.

I'll spend a few minutes. I like it. The logic shouldn't be hard. It's just getting the different css styles to play nicely with eachother.

benjaminpkane:
Ok, I've addressed  the comments and will tackle a general approach to scrollbars separately.

brimoor (approved):
LGTM, everything works beautifully now!

I noticed on this PR (and then also checked latest prod build, and saw it there too) that clicking the cube logo in the App no longer refreshes it.

Are you open to bringing back that option as an app-side equivalent to calling `session.refresh()`? I think it's natural for users to have such an option. Perhaps you removed it because refreshing the browser tab accomplishes the same thing. But it would be nice to have a similar refresh ability in notebooks and the desktop version of the App.

Also, clicking the cube and watching it rotate was just good fun 😆

benjaminpkane:
> I noticed on this PR (and then also checked latest prod build, and saw it there too) that clicking the cube logo in the App no longer refreshes it.
> 
> Are you open to bringing back that option as an app-side equivalent to calling `session.refresh()`? I think it's natural for users to have such an option. Perhaps you removed it because refreshing the browser tab accomplishes the same thing. But it would be nice to have a similar refresh ability in notebooks and the desktop version of the App.

Unintentional, fixed.

PR #878, opened 2021-02-19
Docs refresh

brimoor:
I did a pass over the docs and updated some stuff:

General
- Added some content about samples and fields to the "dataset basics" page, which has also been rebranded to a more generic `FiftyOne basics` page, with the hope that this will be more of a catch-all for general FiftyOne design ideas
- Added a few new capabilities to the `core capabilities` section of the main overview page
- Updated the descriptions of the Brain methods. They were slightly outdated

Recipes
- Refreshed the adding model predictions recipe and rebranded it to `adding classifier predictions`
- Added a new `adding object detections` recipe

Tutorials
- Renamed the `label mistakes` tutorial to `finding classification mistakes`
- Converted the existing detection mistakes to a `finding detection mistakes` tutorial

Best way to read the updates:

```shell
git checkout --track origin/docs-refresh
bash install.bash -d
bash docs/generate_docs.bash -c
# open `docs/build/html/index.html` in your browser
```


benjaminpkane (approved):
Looks good from my perspective!

PR #879, opened 2021-02-21
Patch python import logging

benjaminpkane:
Import logging is broken in v0.7.3 through v0.7.3.3 and fails silently due to a bare except.

This patch fixes the bug, and replaces the bare except with network exception errors only. I have added a test case as well so logging can be tested throughout development.

My mistake for not catching this earlier. The original bare except was of course inappropriate, as well.

Publishing patch release v0.7.3.4 now.

### DIFF
--- a/fiftyone/core/uid.py
+++ b/fiftyone/core/uid.py
@@ -6,6 +6,7 @@ Utilities for usage analytics.
 |
 """
 import os
+from socket import gaierror
 import threading
 import uuid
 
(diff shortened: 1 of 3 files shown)


brimoor (approved):
Ah my bad, looks like I introduced this when I moved this method to the `uid` module. :(

benjaminpkane:
> Ah my bad, looks like I introduced this when I moved this method to the `uid` module. :(

I also missed it in review and did not verify in pre-releases so I will definitely own this one. And the bare except.

Anyways, we should catch things earlier now. And App analytics are look fine.

Issue #882, opened 2021-02-22
[FR] Include "other" class when generating certain evaluation reports on a subset of classes


brimoor:
When running evaluation, most methods provide support for a `classes` argument, either directly to the evaluation method or on the relevant methods on the `EvaluationResults` object, that allow for evaluation reports to be generated on a subset of the classes in a field.

This comment may apply more generally, but, at least for confusion matrices, limiting the classes list will currently cause the confusion matrix to be square, which can be misleading because it doesn't capture incorrect predictions.

From the example below, I think classification reports already handle missing predictions correctly- the P/R/F1 values are appropriately computed. I believe it's just confusion matrices that I don't like.

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")

results = dataset.evaluate_detections("predictions", classwise=False)

# Get the 10 most common classes in the dataset
counts = dataset.count_values("ground_truth.detections.label")
classes = sorted(counts, key=counts.get, reverse=True)[:10]

# Print a classification report for the top-10 classes
results.print_report(classes=classes)  # looks fine(?)

# Plot confusion matrix for top-10 classes
results.plot_confusion_matrix(classes=classes)  # doesn't capture incorrect predictions; not helpful
```

![coco-confusion-matrix](https://user-images.githubusercontent.com/25985824/108753254-8075d880-7512-11eb-8511-bc580790e3ff.png)

brimoor:
As of https://github.com/voxel51/fiftyone/pull/884, the above code snippet will produce a confusion matrix as follows:

![Figure_1](https://user-images.githubusercontent.com/25985824/108882403-5ab00880-75d2-11eb-9917-2d8e5b71f4be.png)

PR #884, opened 2021-02-23
Adds support for showing "other" predictions in confusion matrix

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/882.

When restricting the classes included in confusion matrices via `plot_confusion_matrix(..., classes=classes)`, there will now be extra columns added if either of these cases are true:
- `missing`: there are predictions with GT labels in `classes` that are missing (eg false negatives)
- `other`: there are predictions with GT labels in `classes` but whose predicted labels don't appear in `classes`

The `missing` and `other` strings can be customized by the user as desired.

As per the discussion in https://github.com/voxel51/fiftyone/issues/882, I find this behavior more useful than omitting everything not in `classes` when studying confusion matrices, especially when the confusion matrices are for object detections.

This example shows all the permutations that are now possible:

```py
import fiftyone as fo
import fiftyone.zoo as foz

results = dataset.evaluate_detections("predictions", classwise=False)

# Get the 10 most common classes in the dataset
counts = dataset.count_values("ground_truth.detections.label")
classes = sorted(counts, key=counts.get, reverse=True)[:10]

# Print a classification report for the top-10 classes
# Shows only per-class metrics for the specified classes, but other predictions not in
# `classes` for GTs that are in `classes` are taken into account for P/R/F1 scores
```

```
               precision    recall  f1-score   support

       person       0.52      0.93      0.66       715
         kite       0.59      0.88      0.71       140
          car       0.16      0.74      0.27        61
         bird       0.63      0.76      0.69       111
       carrot       0.09      0.72      0.16        47
         boat       0.09      0.46      0.16        37
    surfboard       0.17      0.70      0.27        30
traffic light       0.32      0.79      0.45        24
     airplane       0.36      0.83      0.50        24
        bench       0.16      0.48      0.24        23

    micro avg       0.37      0.86      0.52      1212
    macro avg       0.31      0.73      0.41      1212
 weighted avg       0.47      0.86      0.59      1212
```

```py
# Default behavior: only show rows (GT) for the requested classes, but include
# "other" and "missing" columns that capture the rest of the predictions associated
# with the GT labels of interest
```

fig1

```py
# Do *not* include the extra prediction columns
results.plot_confusion_matrix(classes=classes, include_other=False)
```

fig2

```py
# `missing` is included in the classes list, so there is both a row (GT) and column (predictions)
# added for missing predictions. In addition, there is an extra column added for predictions
# with "other" labels
results.plot_confusion_matrix(classes=classes + [results.missing])
```

fig3

```py
# Include row and column for `missing`, but no "other" column for predictions
results.plot_confusion_matrix(classes=classes + [results.missing], include_other=False)
```

fig4


benjaminpkane:
I'm just thinking out loud, but would it be helpful in rare cases to allow the user to name the "other" class?

brimoor:
@ehofesmann it'd be nice to get your thoughts on this PR, as a user of detection confusion matrices.

ehofesmann (approved):
I think this is great! Showing `other` and `none` by default is definitely the way to go.

I could see some use cases related to what Ben said about naming the `other` class. Additionally, being able to specify what all of the `other` classes are. For example, if you are using an ImageNet model with 1000 classes, of which only 100 are relevant to you, but you only want to look at 10 in this confusion matrix. You would specify the 10 classes and the 90 other classes. 

That might be kind of an edge case that is better done by just evaluating a different view though.

I think it's good to go in the current form, you could allow renaming `other` if you want but it doesn't seem strictly necessary.

brimoor:
> I think this is great! Showing `other` and `none` by default is definitely the way to go.
> 
> I could see some use cases related to what Ben said about naming the `other` class. Additionally, being able to specify what all of the `other` classes are. For example, if you are using an ImageNet model with 1000 classes, of which only 100 are relevant to you, but you only want to look at 10 in this confusion matrix. You would specify the 10 classes and the 90 other classes.
> 
> That might be kind of an edge case that is better done by just evaluating a different view though.
> 

Hmm yeah that is a bit too specialized to add dedicated support, I think. The natural way to do that would be to `dataset.map_labels(...)` to rename the 90 "other" classes to a single label, which you could then include in your evaluation.

The main thing I was going for this this work is that including the `classes=` parameter will not magically omit predictions from your results. Anything that you didn't specify will be grouped into `(none)` or `(other)`, as necessary.

> I think it's good to go in the current form, you could allow renaming `other` if you want but it doesn't seem strictly necessary.

I actually did address @benjaminpkane's request: there's an optional `other_label` kwarg that you can use to rename the other class if you'd like. The missing class can also be renamed via the optional `missing` kwarg.

Issue #885, opened 2021-02-23
[BUG] Progress bars slowly expand in notebook contexts


brimoor:
When running cells with `fo.ProgressBar`s in them in notebook contexts, the bar slowly increases in width as the percentage increases. The intention (and what happens in other contexts) is that the bar maintains a fixed width, which is either the width of your terminal (by default) or a fixed value you specify via `fo.ProgressBar(..., max_width=)`.

![expanding-progress-bar](https://user-images.githubusercontent.com/25985824/108915936-91e6df80-75fb-11eb-8a48-c97d21c053e6.gif)

brimoor:
I did some digging and this is due to [a known Jupyter bug](https://github.com/jupyter/notebook/issues/4354) that causes certain unicode characters to be rendered with a non-monospaced width. This causes our progress bars to expand as the number of █ characters increases (I believe each one takes up two, rather than one, character widths).

I thought about fixing the underlying unicode issue by adding support for an ASCII-characters-only mode where █ would be replaced by `#`, but after trying it, I decided that I prefer to have my progress bars look solid ████ and expand slightly rather than look less solid `########` but fixed width.

PR #886, opened 2021-02-23
Limit progress bar width to 90 characters in notebook contexts

brimoor:
Limits the width of progress bars in notebook contexts to 90 characters. This is done to avoid the pesky problem of progress bars wrapping multiple lines in the HTML versions of notebooks that we publish to the web (the reader's viewport will never match the code runner's, so the fit-to-terminal option doesn't make sense for publishing content).

On a related note, you may have noticed that our progress bars expand gradually as their percentage increases. This is due to [a known Jupyter bug](https://github.com/jupyter/notebook/issues/4354) and I discuss this a bit in https://github.com/voxel51/fiftyone/issues/885. In any case, the 90 character limit comes in handy here too, since the buggy expansion caps out at about 100 characters, which doesn't cause wrapping for most practical browser window sizes.

### DIFF
--- a/fiftyone/core/context.py
+++ b/fiftyone/core/context.py
@@ -13,6 +13,16 @@ _NONE = "NONE"
 _context = None
 
 
+def is_notebook_context():
+    """Determines whether this process is running in a notebook context, either
+    Colab or Jupyter.
+
+    Returns:
+        True/False
+    """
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
Nice :fire:

PR #889, opened 2021-02-24
Adding low-level utilities for listing and deleting datasets

brimoor:
Adds low-level `list_datasets()` and `delete_dataset()` methods to `fiftyone.core.odm` that can list and delete datasets even if the dataset cannot be loaded via `fo.load_dataset()`, which may happen for example if the dataset's `DatasetDoc` has become corrupted (e.g., failed migration).

Also adds a `dry_run` flag to `drop_orphan_collections()` so that one can check for orphan collections without deleting them.


benjaminpkane (approved):
LGTM

Issue #825, opened 2021-01-28
[BUG] Labelbox import is creating incorrect filepath for samples when downloading the media


luke-iqt:
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS 10.15.7
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: fiftyone v0.7.1.4, Voxel51, Inc.
-   **Python version**: 3.8.3

### Commands to reproduce

Note: Here is a sample row from my labelbox export json file. If you save this to a .json file and then use it for the import, it should reproduce the error:
```json
[
[...]

PR #890, opened 2021-02-24
Fixing Labelbox bugs

ehofesmann:
Updating the Labelbox utilities to address some of the comments from https://github.com/voxel51/fiftyone/issues/825 as well as fixing a bug I ran into.


brimoor (approved):
LGTM

Issue #891, opened 2021-02-25
[FR] Add support for selecting objects using *only* their IDs


brimoor:
The [select_objects()](https://voxel51.com/docs/fiftyone/api/fiftyone.core.collections.html#fiftyone.core.collections.SampleCollection.select_objects) view stage supports creating a view that only contains certain objects, and, as shown in the example code in the docs, it works very naturally with `session.selected_objects`, which provides the necessary info automatically from a user's mouse clicks.

However, it also may be useful for a user to select objects using only their `IDs`, without having to worry about what field or sample the object is in. One way to accomplish this is via the pattern below:

```py
import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone import ViewField as F
from fiftyone.core.expressions import ObjectId

dataset = foz.load_zoo_dataset("quickstart")

# Grab some object IDs
object_ids = [
    dataset.first().ground_truth.detections[0].id,
    dataset.last().predictions.detections[0].id,
]

# Create a view that only contains these objects
_object_ids = [ObjectId(_id) for _id in object_ids]
view = (
    dataset
    .filter_labels(
        "ground_truth", F("_id").is_in(_object_ids), only_matches=False
    ).filter_labels(
        "predictions", F("_id").is_in(_object_ids), only_matches=False
    ).match(
        (F("ground_truth.detections").length() > 0)
        | (F("predictions.detections").length() > 0)
    )
)

session = fo.launch_app(view=view)
```

The following ideas come to mind:
(a) add optional support for passing a simple list of IDs to `select_objects()`, in which case the stage will internally use the pattern above to construct the view
(b) deprecate the more expressive but clunky current syntax and *only* support the above
(c) add a new stage that allows for selection via the above syntax

Perhaps (a) is the best best for now?

An important observation is that the pattern shown above will be quite slow for datasets with many `Label` fields.

Usage of the existing `select_objects()` stage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")

#
# Only include the objects currently selected in the App
#

session = fo.launch_app(dataset)

# Select some objects in the App...

view = dataset.select_objects(session.selected_objects)
```

brimoor:
[...]

PR #892, opened 2021-02-26
Utilities for tagging and visualizing samples and labels

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/891.

**BREAKING CHANGE**
- All notions of "object" have been replaced by "label" in reference to entities with `id` attributes that can be selected in the App. This includes the following changes:
    - `session.selected_objects` -> `session.selected_labels`
    - `SampleCollection.select_objects()` -> `SampleCollection.select_labels()`
    - `SampleCollection.exclude_objects()` -> `SampleCollection.exclude_labels()`
    - `SelectObjects` -> `SelectLabels`
    - `ExcludeObjects` -> `ExcludeLabels`

**DESCRIPTION**

This PR adds more full-featured support for tagging sample as well as individual labels, which I believe will be pretty heavily used by users as we roll out more upcoming features that make it easy to identify groups of samples/labels of interest :)

Specifically, the `select_labels()` and `exclude_labels()` stages support new syntaxes for selecting labels from a collection:

```py
    @view_stage
    def select_labels(self, labels=None, ids=None, tags=None, fields=None):
        """Selects only the specified labels from the collection.

        The returned view will omit samples, sample fields, and individual
        labels that do not match the specified selection criteria.

        You can perform a selection via one of the following methods:

        -   Provide one or both of the ``ids`` and ``tags`` arguments, and
            optionally the ``fields`` argument

        -   Provide the ``labels`` argument

        Args:
            labels (None): a list of dicts specifying the labels to select
            ids (None): a list of IDs of the labels to select
            tags (None): a list of tags of labels to select
            fields (None): a list of fields from which to select labels
        """
```

A few related tagging utilities were also added to `SampleCollection`:

- `tag_samples()` - add tag to samples, if necessary
- `untag_samples()` - remove tag from samples, if necessary
- `count_sample_tags()` - compute histogram of sample tags
- `tag_labels()` - add tag to labels in one or more fields, if necessary
- `untag_labels()` - remove tag from labels in one or more fields, if necessary
- `count_label_tags()` - histogram of label tags in one or more fields 

Also, since "tags" are now first-class citizens in FiftyOne, the `tags` attribute is now declared on all `Label` types.

Notes
- `count_sample_tags()` is a simple alias for `count_values("tags")`, but I added it nonetheless for consistency with the other tags operations
- In the case of a single label field, `count_label_tags()` is also a simple alias for `count_values("field.name.tags")`, but I thought it would be handy to have a version that will consider all label fields by default, similar to how `select_labels(tags=["tag"])` works when no `fields=` argument is provided.
- I chose not to implement `count_label_tags()` as an `Aggregation` because all other aggregations refer to a specific field while this one can refer to multiple fields. And, if a user really is interested in a single field, then they can already compute a tag histogram via the `count_values("field.name.tags")` syntax
- There might be a more efficient way to implement `count_label_tags()` as a single `$group` stage without having to use `$facet`, but I didn't invest the time to look into that

**DISCUSSION**

- ~~The term "object" is used in this PR for consistency with the existing `Session.selected_objects` and `SampleCollection.select_objects()` terminology. However, "object" really refers to any `Label` with an `id`, which includes `Classification`, `Detection`, `Polyline`, `Keypoint`, and `Segmentation`, and excludes label list types like `Detections`. I worry that "object" is a confusing term because it will likely be associated only with `Detection`, and not things like `Classification` that do not have a spatial component. I wonder if a more appropriate term would be "label"? Making this change would come at the price of a breaking API change~~

**VISUALIZATION**

This PR also provides some handy utilities (implemented via `matplotlib`, for now) for visualizing samples/objects that can optionally be connected to a `Session` instance and interactively explored via clicking/lassoing:

The relevant user-facing utilities are:
- `fiftyone.utils.plot.scatter.scatterplot()`- an interactive scatterplot of arbitrary 2D/3D points
- `fiftyone.utils.plot.location.location_scatterplot()` - an interactive scatterplot of lat-lon coordinates

**EXAMPLE VISUALIZATION**

location_scatterplot_opt

scatterplot

**EXAMPLE SELECTION USAGE**

The new examples in the `select_labels()` docstring demonstrate this functionality nicely:

```py
import fiftyone as fo
import fiftyone.zoo as foz

#
# Only include the labels currently selected in the App
#

# Select some labels in the App...

view = dataset.select_labels(labels=session.selected_labels)

#
# Create a view that only includes labels with the specified IDs
#

# Grab some label IDs
ids = [
    dataset.first().ground_truth.detections[0].id,
    dataset.last().predictions.detections[0].id,
]

view = dataset.select_labels(ids=ids)

print(view.count("predictions.detections"))

#
# Create a view that only includes labels with the specified tags
#

# Grab some label IDs
ids = [
    dataset.first().ground_truth.detections[0].id,
    dataset.last().predictions.detections[0].id,
]

# Give the labels a "test" tag
dataset = dataset.clone()  # create a copy since we're modifying data
dataset.select_labels(ids=ids).tag_labels("test")

print(dataset.count_label_tags())

# Retrieve the labels via their tag
view = dataset.select_labels(tags=["test"])

print(view.count("predictions.detections"))
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -136,7 +136,7 @@ jobs:
           git clone https://github.com/voxel51/eta eta --depth 1 --branch develop
           cd eta
           python setup.py bdist_wheel
-          pip install ./dist/*.whl
+          pip install ./dist/*.whl --force-reinstall
       # obtained from "Windows, pip" instructions at
       # https://pytorch.org/get-started/locally/ - unaware of a way to update
       # the version number automatically, but tests on other platforms should
--- a/app/src/components/Player51.tsx
+++ b/app/src/components/Player51.tsx
@@ -153,9 +153,6 @@ export default ({
     fps,
     overlayOptions,
   ]);
-  if (!thumbnail) {
-    console.log(selectedObjects);
-  }
 
   useEffect(() => {
     if (player && selectedObjects) {
--- a/app/src/components/Player51.tsx
+++ b/app/src/components/Player51.tsx
@@ -153,6 +153,9 @@ export default ({
     fps,
     overlayOptions,
   ]);
+  if (!thumbnail) {
+    console.log(selectedObjects);
+  }
 
   useEffect(() => {
     if (player && selectedObjects) {
--- a/app/src/components/SelectObjectsMenu.tsx
+++ b/app/src/components/SelectObjectsMenu.tsx
@@ -34,8 +34,8 @@ const SelectObjectsMenu = ({ sample, frameNumberRef }) => {
   const frameNumber = isVideo ? frameNumberRef.current : null;
 
   useSendMessage(
-    "set_selected_objects",
-    { selected_objects: convertSelectedObjectsMapToList(selectedObjects) },
+    "set_selected_labels",
+    { selected_labels: convertSelectedObjectsMapToList(selectedObjects) },
     null,
     [selectedObjects]
   );
@@ -63,7 +63,7 @@ const SelectObjectsMenu = ({ sample, frameNumberRef }) => {
   ).length;
 
   const _getObjectSelectionData = (object) => ({
-    object_id: object._id,
+    label_id: object._id,
     sample_id: sample._id,
     field: object.name,
     frame_number: object.frame_number,
@@ -99,7 +99,7 @@ const SelectObjectsMenu = ({ sample, frameNumberRef }) => {
     setHiddenObjects((hiddenObjects) =>
       addObjectsToSelection(
         hiddenObjects,
-        ids.map((object_id) => ({ object_id, ...selectedObjects[object_id] }))
+        ids.map((label_id) => ({ label_id, ...selectedObjects[label_id] }))
       )
     );
   };
--- a/app/src/containers/App.tsx
+++ b/app/src/containers/App.tsx
@@ -44,7 +44,7 @@ function App() {
   const handleStateUpdate = (state) => {
     setStateDescription(state);
     setSelectedSamples(new Set(state.selected));
-    setSelectedObjects(convertSelectedObjectsListToMap(state.selected_objects));
+    setSelectedObjects(convertSelectedObjectsListToMap(state.selected_labels));
   };
 
   useEventHandler(socket, "open", () => {
--- a/app/src/containers/Dataset.tsx
+++ b/app/src/containers/Dataset.tsx
@@ -139,7 +139,7 @@ function Dataset(props) {
     }
   }, [hideModal]);
 
-  useSendMessage("set_selected_objects", { selected_objects: [] }, !hideModal);
+  useSendMessage("set_selected_labels", { selected_labels: [] }, !hideModal);
 
   let src = null;
   let s = null;
--- a/app/src/utils/selection.ts
+++ b/app/src/utils/selection.ts
@@ -57,7 +57,7 @@ export const removeMatchingObjectsFromSelection = (
 ) => {
   const newSelection = { ...selection };
   if (Object.keys(filter).length) {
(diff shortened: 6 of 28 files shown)


benjaminpkane:
I think I'd like to remove the `objects` argument altogether, which I can add to this PR because there will need to be a couple App changes.

I also think renaming `object` to `label` now would be worthwhile, noting the breaking the change.

brimoor:
> I think I'd like to remove the `objects` argument altogether, which I can add to this PR because there will need to be a couple App changes.
> 
> I also think renaming `object` to `label` now would be worthwhile, noting the breaking the change.

SGTM. I like switching to "label" and removing the `objects` argument altogether.

The only concern I have (and this may be needless performance worry), is that, if we were to do the obvious thing and make `session.selected_labels` return a simple list of object IDs, then `samples.select_labels(ids=session.selected_labels)` might be a bit slow if there are many label fields, without the benefit of the `fields` argument. Any thoughts on this?

Regardless, if you could handle the updates, that'd be much appreciated 💯 💯 💯

benjaminpkane:
> > I think I'd like to remove the `objects` argument altogether, which I can add to this PR because there will need to be a couple App changes.
> > I also think renaming `object` to `label` now would be worthwhile, noting the breaking the change.
> 
> SGTM. I like switching to "label" and removing the `objects` argument altogether.
> 
> The only concern I have (and this may be needless performance worry), is that, if we were to do the obvious thing and make `session.selected_labels` return a simple list of object IDs, then `samples.select_labels(ids=session.selected_labels)` might be a bit slow if there are many label fields, without the benefit of the `fields` argument. Any thoughts on this?
> 
> Regardless, if you could handle the updates, that'd be much appreciated

Of course, no problem. I just looked at the `SelectObjects` implementation and now I think I've changed my mind on the `dict` argument. Regardless, I'll make a decision and address the above.

benjaminpkane:
The selection elements will be overhauled in the `tagging` branch, https://github.com/voxel51/fiftyone/pull/892/commits/66b61d7741ac837d39494c67c2463f12c0571231 is the bare minimum. In terms of presentation, `label` should now always be used in the App.

brimoor:
Hey @benjaminpkane when you get a chance, can you help get `SelectLabels` and `ExcludeLabels` working in the App? I think there may be issues with the fairly complex type specifications that these stages use:

```py
    @classmethod
    def _params(self):
        return [
            {
                "name": "labels",
                "type": "dict|NoneType",
                "placeholder": "[{...}]",
                "default": "None",
            },
            {
                "name": "ids",
                "type": "list<id>|id|NoneType",
                "placeholder": "...",
                "default": "None",
            },
            {
                "name": "tags",
                "type": "list<str>|str|NoneType",
                "placeholder": "...",
                "default": "None",
            },
            {
                "name": "fields",
                "type": "list<str>|str|NoneType",
                "placeholder": "...",
                "default": "None",
            },
        ]
```


```py
import fiftyone as fo
import fiftyone.zoo as foz

ids = [
    dataset.first().ground_truth.detections[0].id,
    dataset.last().predictions.detections[0].id,
]

# doesn't load correctly
session.view = dataset.select_labels(ids=ids)
```

benjaminpkane:
There is a small but tricky issue I will get to when it makes sense that causes parsing a `list<...>|NoneType` to fail but `NoneType|list<...>` works. In the meantime, https://github.com/voxel51/fiftyone/pull/892/commits/3da1ca7edf51ddf828a19078b97888ac07aea61d has fixed things for me.

brimoor:
Awesome, works!

benjaminpkane (approved):
LGTM. I'm not sure why the tests are failing. `develop` ETA has `etau.is_container`. Still investigating some.

PR #893, opened 2021-02-26
Adding support for saving run results on datasets

brimoor:
Adds support for saving the results of runs (currently, evaluations and brain methods) on a dataset, which can then be loaded later. Run results can be quite big, so they are stored in a separate collection via MongoDB's GridFS framework (which was quite natural to adopt).

Example usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", dataset_name="quickstart-save")
dataset.persistent = True

dataset.evaluate_detections(
    "predictions",
    gt_field="ground_truth",
    eval_key="eval_coco",
)
```

Later, in another process:

```py
import fiftyone as fo

dataset = fo.load_dataset("quickstart-save")

print(dataset.list_evaluations())

results = dataset.load_evaluation_results("eval_coco")

# Get the 10 most common classes in the dataset
counts = dataset.count_values("ground_truth.detections.label")
classes = sorted(counts, key=counts.get, reverse=True)[:10]

# Print a classification report for the top-10 classes
```

### DIFF
--- a/fiftyone/core/odm/dataset.py
+++ b/fiftyone/core/odm/dataset.py
@@ -106,11 +106,9 @@ class DatasetDocument(Document):
     evaluations = DictField(
         EmbeddedDocumentField(document_type=RunDocument), default=dict
     )
-    evaluation_results = DictField(default=dict)
     brain_methods = DictField(
         EmbeddedDocumentField(document_type=RunDocument), default=dict
     )
-    brain_results = DictField(default=dict)
     sample_fields = EmbeddedDocumentListField(
         document_type=SampleFieldDocument
     )
--- a/fiftyone/core/odm/dataset.py
+++ b/fiftyone/core/odm/dataset.py
@@ -106,6 +106,7 @@ class DatasetDocument(Document):
     evaluations = DictField(
         EmbeddedDocumentField(document_type=RunDocument), default=dict
     )
+    evaluation_results = DictField(default=dict)
     brain_methods = DictField(
         EmbeddedDocumentField(document_type=RunDocument), default=dict
     )
--- a/fiftyone/core/odm/dataset.py
+++ b/fiftyone/core/odm/dataset.py
@@ -110,6 +110,7 @@ class DatasetDocument(Document):
     brain_methods = DictField(
         EmbeddedDocumentField(document_type=RunDocument), default=dict
     )
+    brain_results = DictField(default=dict)
     sample_fields = EmbeddedDocumentListField(
         document_type=SampleFieldDocument
     )
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -136,7 +136,7 @@ jobs:
           git clone https://github.com/voxel51/eta eta --depth 1 --branch develop
           cd eta
           python setup.py bdist_wheel
-          pip install ./dist/*.whl
+          pip install ./dist/*.whl --force-reinstall
       # obtained from "Windows, pip" instructions at
       # https://pytorch.org/get-started/locally/ - unaware of a way to update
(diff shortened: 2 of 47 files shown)


benjaminpkane:
Will do. Looking forward to emerging from my App and Player51 deep-dive today. I'll at least start looking into this before Monday.

fiftyone/core/odm/dataset.py
```diff
@@ -106,9 +106,11 @@ class DatasetDocument(Document):
     evaluations = DictField(
         EmbeddedDocumentField(document_type=RunDocument), default=dict
     )
+    evaluation_results = DictField(default=dict)
     brain_methods = DictField(
         EmbeddedDocumentField(document_type=RunDocument), default=dict
     )
+    brain_results = DictField(default=dict)
```
benjaminpkane, line 113: `evaluations` and `evaluation_results` (and `brain_methods` and `brain_results`) are one-to-one, correct? I'm wondering if we can make `results` a field on `RunDocument`.

benjaminpkane: I would then use reference fields on a separate. I don't love the idea of leveraging mongoengine but the API is there, and access does not need to be particularly fast here.

brimoor: Yeah info and results are 1-1, so you're welcome to adopt the convention you mention in `odm/dataset.py` if that's cleaner.

The one note is that `Results` can be quite gigantic, which is why there are separate `RunInfo` and `RunResults` user-facing classes. My thinking was that the user should be able to call `print(samples.get_evaluation_info(eval_key))` to quickly get info about the run, and then call `results = samples. load_evaluation_results(eval_key)` separately to actually get the results if desired.

Only saying this b/c I wouldn't want `samples.get_evaluation_info()` to require loading the results part.

benjaminpkane:
I'm not an expert, but I don't  think there is a good way to get around pickling if array with a `object` `dtype` need to be supported.
I can easily change it back to pickling if that;s what we'd like.

brimoor:
> I'm not an expert, but I don't think there is a good way to get around pickling if arrays with an `object` `dtype` need to be supported.
> I can easily change it back to pickling if that;s what we'd like.

Now that runs are saved in a separate collection, I think its fine to just serialize numpy arrays with `numpy.tolist()`. Especially since string arrays need to be pickled... I wasn't aware of that. This was actually the default serialization for numpy arrays in ETA before I recently changed it... 🤦  

I now regret changing all of the evaluation results classes to store their data as numpy arrays... I think I'll revert that and we can store everything as lists.

benjaminpkane:
> > I'm not an expert, but I don't think there is a good way to get around pickling if arrays with an `object` `dtype` need to be supported.
> > I can easily change it back to pickling if that;s what we'd like.
> 
> Now that runs are saved in a separate collection, I think its fine to just serialize numpy arrays with `numpy.tolist()`. Especially since string arrays need to be pickled... I wasn't aware of that. This was actually the default serialization for numpy arrays in ETA before I recently changed it... 
> 
> I now regret changing all of the evaluation results classes to store their data as numpy arrays... I think I'll revert that and we can store everything as lists.

Ok, noting these documents still grow with respect to the size of the dataset/view. A bridge to be crossed later, when users need it changed!

brimoor:
Okay this is fully-functional now. The switch to GridFS was pretty painless, and I like it better this way anyway, since I think of results as a binary blob, which is how GridFS views it too :)

benjaminpkane (approved):
LGTM

benjaminpkane:
Well, actually looking at the tests now.

Issue #707, opened 2020-12-08
[FR] Flag to avoid copying data when exporting Dataset labels


ehofesmann:
I updated labels for a dataset in FiftyOne and want to use them in my own scripts to train a new model. The images are the same and I don't want to have to copy 118K images that I can just link back to them using the new labels. 

It's easy enough to take an `Exporter` and remove the step where files are copies but it would be nice to have a flag for copying data or just exporting labels. This would also be nice to have in the `fiftyone convert` command.

brimoor:
yep makes sense, I like it.

PR #895, opened 2021-02-26
Add Open Images v6 to Dataset Zoo

### LINKED ISSUE #853: [FR] Add Open Images to dataset zoo
Let's add the Open Images dataset to the [dataset zoo](https://voxel51.com/docs/fiftyone/user_guide/dataset_zoo/datasets.html)!

A specific feature request is the ability to download a sampling of the full dataset (e.g, certain classes, certain count from certain classes, etc). Individual image URLs are available, so this should be doable.

In addition, we should add Open Images-style evaluation as an option in the `fiftyone.utils.eval` module.

Dataset website
https://storage.googleapis.com/openimages/web/index.html

Download and data format instructions
https://storage.googleapis.com/openimages/web/download.html

Open Images-style evaluation (released as part of TF object detection API)
https://storage.googleapis.com/openimages/web/challenge_overview.html

ehofesmann:
Adds [Open Images V6](https://storage.googleapis.com/openimages/web/index.html) to the dataset zoo along with functions in `fiftyone.utils.openimages` to `get_classes()`, `get_segmentation_classes()`, `get_attributes()` that let you get information about the dataset.

The dataset also includes [Localized Narratives](https://google.github.io/localized-narratives/) that are an audio description of an image paired with a recording of the mouse cursor position. This is not yet supported.

Resolves #853

### Overview

This is the first dataset in the zoo to allow for partial downloads letting you specify the label types, classes, attributes, and max samples you want and it will only download relevant images and labels. Some changes were needed to make this happen, namely adding the properties `cleanup` and `supports_partial_download` to the base `ZooDataset` class that will avoid deleting existing data upon multiple calls of `load_zoo_dataset` for open images.

The implementation to load this dataset was just done manually without a sample parser or importer. It then exports the dataset as a `FiftyOneDataset` to disk. The `FiftyOneDatasetExporter` also needed to be updated to provide the option to not `export_media` and only save the serialized `json` of the dataset related to https://github.com/voxel51/fiftyone/issues/707.

```python
import fiftyone as fo
import fiftyone.utils.openimages as fouo
import fiftyone.zoo as foz

print(fouo.get_classes())
print(fouo.get_segmentation_classes())
print(fouo.get_attributes())

    "open-images-v6",
    label_types=["detections", "classifications", "relationships", "segmentations"],
    classes=["Man"],
    split="validation",
    max_samples=10,
)

print(dataset)

```
open-images

Adds object detection evaluation following how it is [done for Open Images](https://storage.googleapis.com/openimages/web/evaluation.html). 

There are a lot of similarities between this and COCO evaluation since they are both based off of Pascal VOC mAP calculation. A few notable differences are:
1) Open Images supports class hierarchies, if a hierarchy is provided, then ground truth objects are created for all parent classes of a detection.
2) Open Images has image-level positive and negative labels indicating the classes that are to be evaluated in each sample (others are ignored). If these fields are not specified then all classes are evaluated in the sample.
3) `iscrowd` is called `IsGroupOf`, and matching detections to a crowd ground truth object will result in 1 true positive when calculating the mAP, no matter the number of predictions that were matched with it. If no predictions are matched with it, it counts as 1 false positive.

New arguments are:
* `hierarchy` (None) - optional label hierarchy loaded following the form of Open Images
* `pos_lab_field` (None) - classifications field containing positive image-level labels
* `neg_lab_field` (None) - classifications field containing negative image-level labels
* `expand_hierarchy` (True) - whether to expand the hierarchy of gt detections and label fields
* `expand_pred_hierarchy` (False) - whether to expand hierarchy of predicted detections (standard open images protocol is for models to output all relevant objects in hierarchy)

```python
import fiftyone as fo
import fiftyone.zoo as foz
import random
random.seed(1)

    "open-images-v6", 
    "validation", 
    max_samples=10,
    label_types=["detections","classifications"]
)

# Add some fake predictions
dataset.clone_sample_field("detections", "predictions")
for s in dataset:
    negs = [l.label for l in s.negative_labels.classifications]
    for d in s.predictions.detections:
        d["confidence"] = random.random()
        if negs and random.random()<0.2:
            d["label"] = random.choice(negs)

    s.save()

hierarchy = dataset.info["hierarchy"]

results = fo.evaluate_detections(
    dataset, 
    "predictions", 
    "detections", 
    method="open-images",
    eval_key="eval",
    expand_pred_hierarchy = True, 
    hierarchy=hierarchy,
    pos_label_field="positive_labels", 
    neg_label_field="negative_labels"
) 
results.mAP()
```


brimoor:
> 7. (Maybe) Not every sample is annotated for every label type. Currently, specifying `max_samples` will download samples if they contain at least 1 specified label type. Could add an option to download samples only if they contain ALL specified label types.

^I'd say that, indeed, if the user specifies multiple label types and `max_samples`, then they'd only like samples that contain all label types

> 8. Download other subsets of [image-level labels like human-verified and machine-generated](https://storage.googleapis.com/openimages/web/download.html)

^doesn't sound important for initial release to me

> 9. Update docs
> 
> Need help with:
> a) How to store data on disk when only downloading specific classes at a time. The dataset will only be created with the relevant labels, so all other label information would not be saved. If we want to add new classes in the future, we would have to redownload the annotation information.

How big is the annotation data? If it's not too big then I'd say it is okay to force the user to re-download all the annotations if they expect to be able to run multiple zoo dataset loads without re-downloading data. With that said, in terms of how to cleanly handle this when Open Images is a zoo dataset, note that `download_zoo_dataset()` has a `cleanup` flag that defaults to `True`. This deletes the temporary directory that all zoo datasets use to download their raw files. You could write your code in such a way that files that already exist in the "temp" directory for Open Images are used w/o redownload (other datasets should already do this; the motivation for this feature was the ability to restart downloads that partially succeeded without redownloading files that already exist). Two options would then be to (a) always override `cleanup=False` for Open Images, or (b) just document for the user that, if they want multiple runs to be more efficient, then they should manually set `cleanup=False`.

> b) Determining the best approach to adding this to the zoo. There are other TODO items before this is relevant.
> c) Other multithreading examples in FO

Multithreading example: https://github.com/voxel51/fiftyone/blob/38d849a0eaecfc929906417a7f7aeb66cac8b6a3/fiftyone/core/metadata.py#L224-L231

> d) Would loading all samples into an array first cause memory issues?
> 
> https://github.com/voxel51/fiftyone/blob/ad5b1f6da0959ececeb6893a07cd899b4445d5b7/fiftyone/utils/openimages.py#L472

^unless there are gigantic segmentation masks in the samples, it would generally not be an issue to store them all in-memory; it's just JSON, and will be about as big as the uncompressed annotation files on disk.

> 
> 
> e) (Discussion) There are two reasons that a sample may not contain a type of label: 1-There are no annotations of that label type for that sample or 2-The specified classes do not exist of that label type but other classes do. Currently, both of these situations will make the field of that label type `None` for that sample. But in case `2`, that label type technically does exist, just not for your specified classes. So for example for detections, should it just be an empty list of detections rather than `None` in case 2?
> https://github.com/voxel51/fiftyone/blob/ad5b1f6da0959ececeb6893a07cd899b4445d5b7/fiftyone/utils/openimages.py#L532

Yes good observation; the difference is important. `None` should strictly be used for "unlabeled" whenever possible, while `[]` should be used when the sample was labeled but happened to not have any instances

fiftyone/utils/openimages.py
```diff
@@ -0,0 +1,1325 @@
...
+    dataset.export(
+        export_dir, dataset_type=fot.FiftyOneDataset, export_media=False
+    )
+    num_samples = len(dataset)
+    dataset.delete()
+
+    return num_samples, all_classes
+
+
+def get_attributes(dataset_dir=None, scratch_dir=None):
```
brimoor, line 249: No need to provide the `scratch_dir` argument here, right? The way I see it, the cleanest way for this method to function is:

- Check if dataset is downloaded in default location (or `dataset_dir`, if provided)
- If relevant files exist, try to load attributes from metadata or scratch dir
- If we fail, temporarily download the attributes file, load the attributes, and then delete the temporary file

Note in the last case that I'm recommending to cleanup the file, since this mode corresponds to a user who just wants to use our code to get the classes without consenting to storing anything on disk. (you can use `etau.TempDir()` to do this cleanly)

I'd also recommend mentioning in the docstring that this method can be called in isolation without having the dataset downloaded.

This comment applies to the other public methods below.

fiftyone/utils/openimages.py
```diff
@@ -0,0 +1,1325 @@
...
+                "metadata.json",
+            )
+
+        if os.path.exists(metadata_path):
+            metadata = etas.load_json(metadata_path)
+            if (
+                "info" in metadata.keys()
+                and "classes" in metadata["info"].keys()
+            ):
+                return metadata["info"]["classes"]
```
brimoor, line 326: There's a fair amount of code duplication in these methods when reading from `metadata["info"]`. You could refactor into a method like `_load_metadata_if_possible()` and call it from each public method here

brimoor (changes requested):
Lookin' good! I left a few comments for you to cleanup the interface just a touch

fiftyone/utils/openimages.py
brimoor: update: after running the method; since it does log where the file is downloaded, I'm okay with `get_attributes()` not cleaning up after itself automatically

brimoor:
@ehofesmann when I run your example code above, I get an empty dataset:

```py
Downloading https://storage.googleapis.com/openimages/v5/class-descriptions-boxable.csv to /Users/Brian/fiftyone/open-images-v6/tmp-download/general/class-descriptions-boxable.csv
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████|   93.8Kb/93.8Kb [2.6ms elapsed, 0s remaining, 40.6Mb/s] 
Downloading https://storage.googleapis.com/openimages/v5/classes-segmentation.txt to /Users/Brian/fiftyone/open-images-v6/tmp-download/general/classes-segmentation.txt
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████|   26.2Kb/26.2Kb [2.2ms elapsed, 0s remaining, 12.1Mb/s] 
Downloading https://storage.googleapis.com/openimages/v6/oidv6-attributes-description.csv to /Users/Brian/fiftyone/open-images-v6/tmp-download/general/oidv6-attributes-description.csv
 100% |███████████████████████████████████████████████████████████████████████████████████████████████████████████|    2.0Kb/2.0Kb [357.6us elapsed, 0s remaining, 8.3Mb/s] 
Downloading split 'validation' to '/Users/Brian/fiftyone/open-images-v6/validation'
Downloading https://storage.googleapis.com/openimages/2018_04/bbox_labels_600_hierarchy.json to /Users/Brian/fiftyone/open-images-v6/tmp-download/general/bbox_labels_600_hierarchy.json
 100% |█████████████████████████████████████████████████████████████████████████████████████████████████████████|  674.1Kb/674.1Kb [47.8ms elapsed, 0s remaining, 13.8Mb/s]     
Downloading https://storage.googleapis.com/openimages/v5/validation-annotations-bbox.csv to /Users/Brian/fiftyone/open-images-v6/tmp-download/validation/detections/validation-annotations-bbox.csv
 100% |█████████████████████████████████████████████████████████████████████████████████████████████████████████|  191.5Mb/191.5Mb [940.7ms elapsed, 0s remaining, 215.3Mb/s]      
Downloading https://storage.googleapis.com/openimages/v5/validation-annotations-human-imagelabels-boxable.csv to /Users/Brian/fiftyone/open-images-v6/tmp-download/validation/classifications/validation-annotations-human-imagelabels-boxable.csv
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████|   81.2Mb/81.2Mb [474.4ms elapsed, 0s remaining, 171.3Mb/s]      
Downloading https://storage.googleapis.com/openimages/v6/oidv6-validation-annotations-vrd.csv to /Users/Brian/fiftyone/open-images-v6/tmp-download/validation/relationships/oidv6-validation-annotations-vrd.csv
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████|   23.4Mb/23.4Mb [257.7ms elapsed, 0s remaining, 90.8Mb/s]      
No segmentations exist for classes: Bicycle

You can view the available segmentation classes via `get_segmentation_classes()`

Downloading https://storage.googleapis.com/openimages/v5/validation-annotations-object-segmentation.csv to /Users/Brian/fiftyone/open-images-v6/tmp-download/validation/segmentations/validation-annotations-object-segmentation.csv
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████|   22.3Mb/22.3Mb [251.3ms elapsed, 0s remaining, 88.7Mb/s]      
Writing annotations for validation split to disk
 100% |████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 0/0 [15.8ms elapsed, ? remaining, ? samples/s] 
Found 0 samples
Dataset info written to '/Users/Brian/fiftyone/open-images-v6/info.json'
Loading existing dataset 'open-images-v6-validation'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
Name:           open-images-v6-validation
Media type:     None
Num samples:    0
Persistent:     False
Tags:           []
Sample fields:
    filepath:        fiftyone.core.fields.StringField
    tags:            fiftyone.core.fields.ListField(fiftyone.core.fields.StringField)
    metadata:        fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.metadata.Metadata)
    detections:      fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.labels.Detections)
    positive_labels: fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.labels.Classifications)
    negative_labels: fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.labels.Classifications)
    relationships:   fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.labels.Detections)
    segmentations:   fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.labels.Detections)
```

brimoor:
@ehofesmann I haven't looked closely to see whether this would be easy or hard to implement, but it might be nice to provide `shuffle` and `seed` parameters that enable users to download random subsets of the dataset in concert with `max_samples`.

All three of these parameters are supported by all datasets when calling `load_zoo_dataset()`; you'd be supporting it for `download_zoo_dataset()` too.

Also, this reminds me: look at the documentation of `load_zoo_dataset()`, specifically how `**kwargs` are handled. Those `kwargs` are split between `download_zoo_dataset()` and the `DatasetImporter`. This may present problems since the `max_samples,shuffle,seed` parameters would clash.

The problem is... the user would expect `load_zoo_dataset("open-images-v6", split="validation", max_samples=10, shuffle=True)` to mean "load 10 random images from among the data I've already downloaded, not "load 10 more random images". For the latter, the user would call `download_zoo_dataset()` again.

ehofesmann:
> @ehofesmann when I run your example code above, I get an empty dataset:

Agh! Still some issues with the logic when providing specific `label_types` and `classes` I see, thought I fixed that when making examples for the docs. The following will work:
```python
    "open-images-v6",
    split="validation",
    max_samples=10,
)
```

> @ehofesmann I haven't looked closely to see whether this would be easy or hard to implement, but it might be nice to provide `shuffle` and `seed` parameters that enable users to download random subsets of the dataset in concert with `max_samples`.
> 
> All three of these parameters are supported by all datasets when calling `load_zoo_dataset()`; you'd be supporting it for `download_zoo_dataset()` too.
> 
> Also, this reminds me: look at the documentation of `load_zoo_dataset()`, specifically how `**kwargs` are handled. Those `kwargs` are split between `download_zoo_dataset()` and the `DatasetImporter`. This may present problems since the `max_samples,shuffle,seed` parameters would clash.
> 
> The problem is... the user would expect `load_zoo_dataset("open-images-v6", split="validation", max_samples=10, shuffle=True)` to mean "load 10 random images from among the data I've already downloaded, not "load 10 more random images". For the latter, the user would call `download_zoo_dataset()` again.

At the moment, `download_zoo_dataset()` gets priority, so `max_samples` will only ever be sent to the download function. That's a good point though, when loading 10 random images, you would expect them to be from disk if you already downloaded parts of the dataset before. I'll look into this.

fiftyone/utils/openimages.py
ehofesmann: I kind of like it cleaning up after itself. There is always the option for you to just download a subset of the dataset if you don't want to have to download the files every time.

ehofesmann:
OK! It's finally at a state where everything works. I've fixed the bug that you ran into and I've made it so that existing images are used before downloading new ones. 

This is in a pretty good state I think, though there are still more improvements that could be done. I began working on functionality to make loading existing subsets of a partially downloaded dataset more generic but that was taking a lot longer than I had time for right now. 

Also with the current implementation, a new FiftyOne dataset has to be constructed, saved, and then reloaded anytime open images is loaded from the zoo. As part of the generic loading, I was looking into how to only have to load the dataset once, save it, and then just use the loaded dataset. Loading it twice can become a serious issue if someone downloads the entire training split (500+ GB). Though once it's downloaded and saved, they could just load it directly with `dataset.add_dir`. 

Anyway, that might be too much for now. I'm curious what you're thoughts are on that @brimoor

brimoor:
> Also with the current implementation, a new FiftyOne dataset has to be constructed, saved, and then reloaded anytime open images is loaded from the zoo. As part of the generic loading, I was looking into how to only have to load the dataset once, save it, and then just use the loaded dataset. Loading it twice can become a serious issue if someone downloads the entire training split (500+ GB). Though once it's downloaded and saved, they could just load it directly with `dataset.add_dir`.
>
> Anyway, that might be too much for now. I'm curious what you're thoughts are on that @brimoor

I'm not quite sure what you're saying here. Each image is only *downloaded* once, right? Are you just pointing out that creating a FiftyOne dataset from data in a known format on disk can take some time (EG, 200 samples/sec on my laptop for COCO-formatted data)? I haven't looked at the implementation closely after your changes.

I think the way to do it is to organize all downloaded + parsed data in `/path/to/zoo/open-images-v6/<split>` folders. When users request a slice of data to be loaded into FiftyOne, this may or may not trigger downloading of more images into the `<split>` folders, and, if the user creates multiple FiftyOne datasets using `load_zoo_dataset("open-images-v6", ...)`, then they may have multiple datasets pointing to different subsets of the same files in the official `<split>` directories. This is all fine.

Are you talking about this or something else?

ehofesmann:
> > Also with the current implementation, a new FiftyOne dataset has to be constructed, saved, and then reloaded anytime open images is loaded from the zoo. As part of the generic loading, I was looking into how to only have to load the dataset once, save it, and then just use the loaded dataset. Loading it twice can become a serious issue if someone downloads the entire training split (500+ GB). Though once it's downloaded and saved, they could just load it directly with `dataset.add_dir`.
> > Anyway, that might be too much for now. I'm curious what you're thoughts are on that @brimoor
> 
> I'm not quite sure what you're saying here. Each image is only _downloaded_ once, right? Are you just pointing out that creating a FiftyOne dataset from data in a known format on disk can take some time (EG, 200 samples/sec on my laptop for COCO-formatted data)? I haven't looked at the implementation closely after your changes.
> 
> I think the way to do it is to organize all downloaded + parsed data in `/path/to/zoo/open-images-v6/<split>` folders. When users request a slice of data to be loaded into FiftyOne, this may or may not trigger downloading of more images into the `<split>` folders, and, if the user creates multiple FiftyOne datasets using `load_zoo_dataset("open-images-v6", ...)` with different parameter settings, then they may have multiple datasets pointing to different subsets of the same files in the official `<split>` directories. This is all fine.
> 
> Are you talking about this or something else?

Yes, each image is only downloaded once.

The way it is organized is similar to what you are saying. 
1) Images and raw annotations that are needed get downloaded into `/zoo/open-images/tmp-download/<split>`, then only the images get moved to `/zoo/open-images/<split>`. 
2) Label information is parsed from the raw annotations, but only for the specific samples that are requested. 
3) These labels are then added to a FiftyOne dataset which is then saved in `/zoo/open-images/<split>`. This `FiftyOneDataset` only stores the specific subset of images/labels that were requested and gets overwritten every time.
4) The code in `load_zoo_dataset()` then uses the `FiftyOneDatasetImporter` to import the saved `Dataset`

I just loaded 5000 samples from OI validation to compare the timing with COCO validation.
* 8.5 minutes - Downloading 5000 images (coco takes around 4 but is zipped)
* 5 minutes -  Parsing the raw annotations and adding `Samples` to the `Dataset` 
* 2 minutes - Saving the `FiftyOneDataset` to disk
* 3.5 minutes - Loading the `FiftyOneDataset` from disk

Note, the speed of loading the dataset from disk was around 68 samples/s vs coco 200 samples/second. I'm guessing it's because of the additional labels loaded for oi (segmentations and relationships fields).

Analyzing the speed of each step:
* Downloading images is already in parallel so there aren't too many easy gains there.
* Parsing annotations and creating samples also pretty difficult to speed up (it is using `dataset.add_samples()` so it's already batched)
* Saving the `FiftyOneDataset` to disk is required
* 3.5 minutes loading the dataset from disk could be avoided since we already loaded it

requirements/common.txt
```diff
@@ -1,4 +1,5 @@
 argcomplete==1.11.0
+boto3==1.17.36
```
brimoor, line 2: @ehofesmann FYI you need to include any new dependencies in `setup.py` as well in order for them to be automatically installed when someone installs the package from pypi.

The `requirements.txt` files just document the exact versions of packages that we use when developing, for reproducibility when working with source installs.

fiftyone/utils/eval/openimages.py
```diff
@@ -0,0 +1,793 @@
...
+
+    @classmethod
+    def _from_dict(cls, d, samples, **kwargs):
+        return super()._from_dict(
+            d, samples, precision=d["precision"], recall=d["recall"], **kwargs,
+        )
+
+
+_NO_MATCH_ID = ""
+_NO_MATCH_IOU = -1
```
brimoor, line 461: @ehofesmann can you use `None` instead of `-1` for non-matches? I made this change in #974 for COCO-style evaluation because it makes using the IoU slider filter in the App more intuitive (the bounds of the slider will correspond to the actually observed IoUs, not always extending to -1).

fiftyone/zoo/datasets/base.py
```diff
@@ -847,6 +848,162 @@ def _download_and_prepare(self, dataset_dir, scratch_dir, split):
...
+        self.seed = seed
+        self.shuffle = shuffle
+        self.image_ids = image_ids
+        self.image_ids_file = image_ids_file
+        self.num_workers = num_workers
+        self.version = version
+
+    @property
+    def name(self):
+        return "open-images"
```
brimoor, line 962: So for other datasets with versions like `coco-2014` and `coco-2017`, we put the "version" directly in the name of the zoo dataset.

In this case, it is currently undocumented, but users of Open Images can get a particular version (when more are supported) via the syntax `load_zoo_dataset("open-images", ..., version="v6")`, which is a 

Perhaps you should explicitly go with `OpenImagesV6Dataset` and `open-images-v6` in this class definition, for consistency with how we're handling other datasets? Another reason this makes sense is that the docstring of this class is specific to V6; those details will change for other versions of the dataset, suggesting that dedicated classes are indeed the right approach.

(Currently the content from the class docstring is manually copied over to generate the zoo docs, but someday I'd like to automate that.)

The underlying utils methods can still accept a `version` argument, I'm just referring specifying to the ZooDataset class here.

docs/source/user_guide/evaluation.rst
```diff
@@ -699,6 +694,505 @@ curves for your detections by passing the ``compute_mAP=True`` flag to
...
+- Ensure that precision is a non-increasing array.
+
+- Interpolate precision values so that they can be plotted with an array of
+  101 evenly spaced recall values.
+
+- Computing mAP: For every class that contains at least one ground truth
+  object, compute the AP by averaging the precision values over all 10 IoU
+  thresholds. Then take the mean of AP values of all classes to get mAP. 
+
+Open Images-style evaluation
```
benjaminpkane, line 822: Maybe add a `ref` here?

brimoor (approved):
LGTM

PR #896, opened 2021-03-01
Adding support for using DataLoaders when applying non-Torch models

brimoor:
Description coming soon


ehofesmann (commented):
It is strange that using a DataLoader doesn't improve performance even when using a GPU. From the looks of it, it seems like it should work for non-torch models just as well as for torch models. 

Did you actually see an improvement when using DataLoaders with torch models? If not this could be streamlined even more to never use a data loader.

Side questions: 
1) What is `TorchImageClassificationDataset` actually used for? I don't see any references to it, so is it just for outside purposes to quickly make a torch dataset? If so, I'll add a detection one in the future for a blog.
2) Why is it designed to provide a list of images and targets to the `TorchDataset`s instead of just providing a samples collection and field name and parsing the relevant information from the dataset in the class? I get that a dataset could change but a list of images/targets is fixed, however it just seems much easier to pass in your dataset than having to parse the images/targets yourself.

brimoor:
> It is strange that using a DataLoader doesn't improve performance even when using a GPU. From the looks of it, it seems like it should work for non-torch models just as well as for torch models.
> 
> Did you actually see an improvement when using DataLoaders with torch models? If not this could be streamlined even more to never use a data loader.
> 

I haven't tried using a Torch model w/o `DataLoader`, but I don't want to remove `DataLoader`s in that case because the Torch models expect torch tensors as input and the `DataLoader`s are a convenient way to load data in that format.

I did some searching online awhile back about writing `DataLoader`s that emit numpy arrays and was cautioned that it's not a good idea because the numpy arrays have to be pickled (slow) when transferring them between the worker processes and the main process. So, my assumption (not heavily tested) is that the cost of transferring the data was cancelling out the benefit of using multiple processes to load data.

It's also possible that data was being transferred between CPU and GPU unnecessarily; I didn't investigate more thoroughly because my main goal was just to simplify the code paths of `apply_model()` and `compute_embeddings()` in this PR.

I'd be happy to have your help investigating this in more detail if you have the bandwidth. I suspect that there is a way to make TF models more efficient...

> Side questions:
> 
> 1. What is `TorchImageClassificationDataset` actually used for? I don't see any references to it, so is it just for outside purposes to quickly make a torch dataset? If so, I'll add a detection one in the future for a blog.
>
> 2. Why is it designed to provide a list of images and targets to the `TorchDataset`s instead of just providing a samples collection and field name and parsing the relevant information from the dataset in the class? I get that a dataset could change but a list of images/targets is fixed, however it just seems much easier to pass in your dataset than having to parse the images/targets yourself.

Yeah looks like `TorchImageClassificationDataset` is not used at this point. The `TorchImagePatchesDataset` class is used, so I probably just added `TorchImageClassificationDataset` for completeness (this code is quite old).

I agree that it would in principle be a more useful abstraction to rewrite `TorchImageClassificationDataset` to take a FiftyOne dataset and the name of a `Classification` field as input and let the dataset object do the work of extracting the data.

But, to illustrate why it's not necessarily clear cut, in the case of `TorchImagePatchesDataset`, our current use cases require some custom logic to convert `Polylines` and other "region" labels into `Detection` instances that can then be used to extract patches. And there are edge cases to consider like what to do if there are no detections for a dataset (in the case of computing uniqueness on ROIs, we need something for every sample, so we use the entire image as a single ROI; but, one may choose differently when training a model, for example). I thought this logic was too use-case-specific to build into the `TorchImagePatchesDataset`, so instead I required the user to do that themselves beforehand and then pass in the `detections` that they want to use after that.

More generally, since writing torch `Dataset`s is pretty trivial, I question whether there's much value in us trying to provide builtin datasets that cover the user's needs. For example, in the case of detection, we could provide a torch dataset that will extract bounding boxes from a `label_field`, but what format should the coordinates be outputted in? `[x1, y1, x2, y2]` or `[x, y, w, h]`? relative coordinates or absolute coordinates? etc. The answers to these will depend on the user's model. Our dataset could provide a `parse_bbox()` method that the user could override to put the box in the correct format for their model, but I question whether a user would then find much value in having part of their training pipeline not "owned" by them. Why not just write the `Dataset` from scratch?

So basically, I'm onboard with adding Dataset classes when we have use cases for them (at least internally). And, when we do, making an effort to make them as generic as possible sounds like the way to go. But, I question if users would actually use them or just write their own data loaders.

ehofesmann:
> It's also possible that data was being transferred between CPU and GPU unnecessarily; I didn't investigate more thoroughly because my main goal was just to simplify the code paths of `apply_model()` and `compute_embeddings()` in this PR.
> 
> I'd be happy to have your help investigating this in more detail if you have the bandwidth. I suspect that there is a way to make TF models more efficient...

That's really interesting, I'd like to take a look at current approaches to loading TF data. I haven't done much with TF2 but I'm willing to bet that the new TF2 data format is similar to PyTorch `DataLoader`s.

>More generally, since writing torch Datasets is pretty trivial, I question whether there's much value in us trying to provide builtin datasets that cover the user's needs. For example, in the case of detection, we could provide a torch dataset that will extract bounding boxes from a label_field, but what format should the coordinates be outputted in? [x1, y1, x2, y2] or [x, y, w, h]? relative coordinates or absolute coordinates? etc. The answers to these will depend on the user's model. Our dataset could provide a parse_bbox() method that the user could override to put the box in the correct format for their model, but I question whether a user would then find much value in having part of their training pipeline not "owned" by them. Why not just write the Dataset from scratch?

I agree that the first step is to just make a blog showing the correct way to write your own `DataLoader` using FiftyOne datasets. 

I also agree that the default when writing your own `DataLoader` should be to take FiftyOne datasets as input, but that it might not be the best approach for more general code used within FiftyOne. 

Though, I do have a feeling that it would be useful to provide a basic `DataLoader` for something like detection. I'll wait to implement anything until I train some models with FiftyOne to see if it's actually worth it over just writing a custom `DataLoader`. But a couple of things make me believe that it might be:
1) There is a pseudo-standard way to load a detection model that is outlined in the primary PyTorch tutorial on the matter, also all TorchVision detectors require that `DataLoader` format. This would give us things like the default `[x,y,h,w]` format of boxes and output format of `__getitem__` that we need to provide.
2) If you use the TorchVision training examples, then it only takes around 50-100 lines to train a model on your dataset. If writing a custom `DataLoader` class takes up half of that, it would help streamline the process a good bit even if it's fairly easy to write.

ehofesmann (approved):
I think this PR is good for now! I've opened an issue related to TF data loading so we can get this PR merged https://github.com/voxel51/fiftyone/issues/905

PR #906, opened 2021-03-08
[WIP] App tagging

### LINKED ISSUE #904: [FR] Add support for assigning a tag to all selected samples or labels in the App
https://github.com/voxel51/fiftyone/pull/892 and https://github.com/voxel51/fiftyone/pull/901 added substantial functionality for selecting samples/labels of interest and programmatically assigning them `tags`.

This functionality will be further boosted by adding support for adding/remove tags from the currently selected samples/labels in the App.

**Tagging selected samples**

This could be as simple as adding a new option(s) under the existing dropdown that enables tagging/untagging the currently selected samples.

Fundamentally, I think it makes sense to force the user to *select* samples before tagging them, rather than, for example, allowing them to add a tag to all samples in the current view. However, for ease of use, perhaps we should add a `Select all samples` option to the dropdown that enables users to select all -> tag all with ease.

**Tagging selected labels**

Label selection via the App typically happens with expanded view open (for which there is an analogous FR https://github.com/voxel51/fiftyone/issues/835 for taking actions).

However, recent features such as `select_labels()` have made it possible to isolate a set of labels across multiple samples/fields in the App. Therefore, it makes sense to provide a label selection UI element (next to the sample selection UI element, say) on the sample grid that enables the user to tag/untag the currently selected labels.

One option here could be to force the user to *select* the labels in the App before tagging them. But, there is not currently the ability to select labels directly from grid mode (nor should there be, likely, since the cells are too small for that). One option could be to add a `Select all visible` option in the dropdown that would actually select all the visible objects, but I would only do this if we were to render the boxes/tags of selected labels with dotted border style. A separate option (although it is a bit less consistent with the sample selection convention mentioned above), would be to simply add a `Tag/untag visible labels` option(s) to the dropdown

Screen Shot 2021-03-08 at 10 45 43 AM

benjaminpkane:
Resolves #904 

Adds the ability to tag samples from the App in both the grid (and expanded sample view, forthcoming).

Screenshot from 2021-03-08 12-11-21
Screenshot from 2021-03-08 12-11-00


benjaminpkane:
I would say I only need a couple more hours to finalize this implementation...the modal still needs work.

I went with the principle that actions should be separated into separate categories like gmail. Actions in the modal have been moved to the bottom right (I can change that, but the previous location did not look good with the icons imo).

I could go on, but I hope the screenshots capture the new take on actions given Monday's discussions. It's not perfect, but I think something very close to this should be released on Thursday.

Screenshot from 2021-03-09 19-24-37
Screenshot from 2021-03-09 19-24-22
Screenshot from 2021-03-09 19-24-05
Screenshot from 2021-03-09 19-23-59
Screenshot from 2021-03-09 19-23-46
Screenshot from 2021-03-09 19-23-40

brimoor:
oooo I like it! 🔥 🚒 🧯💥

benjaminpkane:
State syncing in the modal is proving to be tricky. Or at least it's adding new requirements to the App's data flow. Hopefully today.

benjaminpkane:
Status screenshot.
Screenshot from 2021-03-15 22-57-11

benjaminpkane:
Undrafting. Video needs a little more work (incomplete data flow), but image dataset tagging is complete save for cosmetic details more monkey testing.

brimoor:
Nice work!! I find the tagging UI very intuitive, which is great because its a pretty powerful element.

Here's my observations after playing around with the feature for awhile:

**Small requests**
- How about removing `shown` from `+ tag 39 shown labels` in the tagging menu? So just `+ tag 39 labels`? Also, in the specific case when tagging labels among selected samples, I think the text needs to read `+ tag 39 selected labels` for consistency with `+ tag 10 selected samples`. For the first part, I thought about recommending `+ tag 39 visible labels` instead, but then `+ tag 39 visible labels in selected samples` seems like too much of a mouthful...
- I'd personally rather not see the image gradient at the bottom of the image when viewing sample-level tags. I think there's value to the upper-gradient on-hover because the checkbox can be hard to spot otherwise, but tags at the bottom already standout clearly so I'd rather err on the side of not cluttering the user's view of the image (especially since this is persistent, not only on-hover)
- When creating a new label, the "create" button says `Create "<tag>"`. How about making this more informative like `Add '<tag>' tag to <number> samples/labels`?
- For consistency with how tags are listed in the fields sidebar, I'd say that tags should only appear under the `Tags` section in expanded sample mode if they actually exist on that sample. EG, don't show the `X` if the tag exists on the dataset but not on that sample

**Larger(?) requests**
- In the tag dropdown, it would be very helpful if the list of pre-populated tags included tags that exist anywhere on the dataset, but not necessarily within my current selection. The use case here is when I have a tag called `fixme` that I apply to samples/labels whenever they have a problem. If I identify a new set of problem data, I'd like the `fixme` tag to appear as a checkbox so I don't have to type it from scratch. If these queries only ran when the tag UI element is opened, this would perhaps be more palatable to implement (no slowdown concerns)

**Bugs/observations**
- Sometimes clicking tag checkboxes seemed to cause an automatic save, while sometimes I had to manually press `apply`. This seemed to be intermittent, like the auto-saving was a bug. Now that I've used the interface, I really like having to opt-in to saving changes by clicking `apply` in all cases. Can we always require that?
- In expanded sample mode, a few times the tags/selected labels dropdowns were rendering behind the panel below (eg I couldn't see them). Closing the modal and refreshing the page seemed to fix this for me
- It feels like the App is a bit sluggish with these changes. Is this because there are some blocking requests for computing things like label tag counts? Are there any low-hanging options to do something about this? One thought is to only load label counts if the label tab of the tag-dropdown is actually pressed (if there is, in fact, a slowdown and this is the primary cause)

benjaminpkane:
>In the tag dropdown, it would be very helpful if the list of pre-populated tags included tags that exist anywhere on the dataset, but not necessarily within my current selection. The use case here is when I have a tag called fixme that I apply to samples/labels whenever they have a problem. If I identify a new set of problem data, I'd like the fixme tag to appear as a checkbox so I don't have to type it from scratch. If these queries only ran when the tag UI element is opened, this would perhaps be more palatable to implement (no slowdown concerns)

But do not union sample and labels tags, correct?

benjaminpkane:
All comments addressed. Things are faster now. Just need a little time tomorrow to wrap up video.

benjaminpkane:
Ok, I want to do a little more testing on a large video dataset, but things look to be pretty stable across video and image datasets.

I noticed one bug when tagging selected labels that I am still looking at too.

brimoor:
Awesome looking forward to trying this out! I'll circle back to do that this evening

brimoor (approved):
Wow, lots of code to make this happen! But a good bit was a frontend refactor I believe.

I stress tested pretty heavily on an image dataset and everything seemed to be working as expected. LGTM

brimoor:
A couple things I noticed on my last testing:

1. I always see `0` and `1` sample tags on my dataset in expanded sample mode (but I did not add these tags, nor do they disappear if I try to delete them via the UI)

Screen Shot 2021-03-18 at 6 04 50 PM

2. A small design thing: the `Add <tag> to samples/labels` message starts off as horizontally centered, but after I start typing, it actually seems that the text has a fixed left offset, which causes there to be a fair amount of empty space while my tag message is getting truncated on the RHS...

Perhaps you can take full use of available space by switching from centered to left-aligned when the user has typed a non-empty tag in the box?

Screen Shot 2021-03-18 at 6 11 04 PM

3. This might be(?) unrelated to the changes in this PR, but I noticed when testing tagging functionality with a `shuffle().limit(10)` stage in my view bar that submitting a tag change generates a new view. I guess this must be because a new shuffle stage with different internal state is being created during reload? I thought this issue was resolved way back when stages got UUIDs. Can you reproduce this?

benjaminpkane:
> Wow, lots of code to make this happen! But a good bit was a frontend refactor I believe.
> 
> I stress tested pretty heavily on an image dataset and everything seemed to be working as expected. LGTM

:+1: yes, chipping away at the big picture

PR #911, opened 2021-03-16
Adding respsonsive plotting interface

brimoor:
This PR adds an extensible plotting interface that enables connecting various types of responsive plots to a `Session` object.

A major subcomponent of this PR is adding a [plotly](https://plotly.com) backend for generating all types of plots in FiftyOne, including:
- Confusion matrices
- PR curve(s)
- ROC curve
- 2D or 3D scatterplots
- Geolocation plots
- Arbitrary plots defined by FiftyOne `Aggregation`(s) applied to a `DatasetView`

What do we mean by "responsive" plots? Well, two classes of plots are introduced:

### ViewPlot

These plots can be connected to a `Session` and will automatically recompute + update whenever the `Session`'s view changes.

### InteractivePlot

These plots can be connected to either the samples or labels in a `Session` object, via the `id`s of the relevant objects. When plots of this type are connected to a `Session` object, they will be bidirectionally interactive (if supported):
- Lasso/box-selecting points in a plot will automatically update the `Session`'s view to select the corresponding samples/labels in the App
- Updating a Session's view will automatically select the corresponding points in the plot

Methods such as `scatterplot()` and `location_scatterplot()` that generate interactive plots will use the `plotly` backend by default, but the existing [matplotlib](https://matplotlib.org) backend can be recovered by passing `backend="matplotlib"` to the relevant plotting methods.

However, I think we should deprecate the `matplotlib` backend altogether soon.

For the `plotly` backend, this interactivity is currently only supported in Jupyter notebooks. But in future work we can achieve this outside of notebooks by adopting [Dash](https://plotly.com/dash) to serve the plots.

### Connecting plots to sessions

All `Session` instances now have a `plots` attribute, which is an instance of a new `PlotManager` class that enables adding, connecting, disconnecting, freezing (in notebooks), and removing plots from a `Session` object.

See the examples below for usage.

### Organization

The plotting interface is exposed in a new `fiftyone.core.plots` subpackage, organized as follows:

```
fiftyone.core.plots
    plotly      # plotly implementations of InteractivePlots
    matplotlib  # matplotlib implementations of InteractivePlots
    views       # implementations of ViewPlots (plotly only)
```

### Other changes

- `ClassificationResults` and `DetectionResults` objects now store the IDs of the corresponding labels; for general reference and also so that they can be provided to the interactive plotting methods
- Adds support for testing `DatasetView` equality via `view1 == view2`
- Updated the `histogram_values()` aggregation to support all-None data
- The `values()` aggregation now supports extracting `id` fields from samples/frames/fields, with the following convention:
    - `values("id")` or `values("embedded.field.id")` will extract the IDs as strings
    - `values("_id")` or `values("embedded.field._id")` will extract the IDs as bson `ObjectID`s
- Fixed a bug in `SampleCollection._get_selected_labels()` that prevented the method from properly handling None-valued data

### Examples and documentation

Tutorials and documentation will be added in a separate PR. In the meantime, [examples-clean.ipynb.zip](https://github.com/voxel51/fiftyone/files/6198479/examples-clean.ipynb.zip) provides a variety of examples demonstrating usage of the plotting features.


fiftyone/utils/plot/plotly.py
```diff
@@ -0,0 +1,1913 @@
...
+    with Mapbox plots.
+
+    https://github.com/plotly/plotly.py/issues/2570
+    """
+    filepath = os.path.join(
+        os.path.dirname(os.__file__),
+        "site-packages",
+        "plotly",
+        "basedatatypes.py",
+    )
```
benjaminpkane, line 1883: This gives the incorrect path in my venv

brimoor: Hmm, can you patch it so it will either locate the file in other setups or just gracefully return?

benjaminpkane: Will do

benjaminpkane:
Ok. I've read through a lot of the code and really like the API. ViewGrid and ViewPlot are pretty cool. All of it, really.

As discussed, plots now display in Colab but are not interactive/responsive. If making a copy of the notebook and opening the copy in Colab, the plots are still rendered.

Plots will not be displayed in Jupyter if the notebook was created in and downloaded from Colab. The work to make the Colab-generated plots display in Jupyter will best coincide with a Dash backend.

Back to tagging.

benjaminpkane (approved):
LGTM! Will test more on develop.

Issue #910, opened 2021-03-15
[BUG] Error in filtered field schema


ehofesmann:
There is a schema-related bug that comes up when working with a view that had a field filtered.
```python
import fiftyone.zoo as foz
from fiftyone import ViewField as F

dataset = foz.load_zoo_dataset("quickstart")
view = dataset.filter_labels("ground_truth", F("label") == "person")
view.select_fields("filepath")
```

```
~/work/fiftyone/fiftyone/fiftyone/core/collections.py in __repr__(self)
     75
     76     def __repr__(self):
---> 77         return self.summary()
     78
     79     def __bool__(self):

~/work/fiftyone/fiftyone/fiftyone/core/view.py in summary(self)
    185         """
    186         aggs = self.aggregate(
--> 187             [foa.Count(), foa.Distinct("tags")], _attach_frames=False
    188         )
    189         elements = [

~/work/fiftyone/fiftyone/fiftyone/core/collections.py in aggregate(self, aggregations, _attach_frames)
   4253
   4254         pipeline = self._pipeline(
-> 4255             pipeline=facets, attach_frames=_attach_frames
   4256         )
   4257

~/work/fiftyone/fiftyone/fiftyone/core/view.py in _pipeline(self, pipeline, attach_frames, detach_frames, frames_only)
    650         _pipeline = []
    651         for s in self._stages:
--> 652             _pipeline.extend(s.to_mongo(self))
    653
    654         if pipeline is not None:

~/work/fiftyone/fiftyone/fiftyone/core/stages.py in to_mongo(self, sample_collection)
   1329
   1330     def to_mongo(self, sample_collection):
-> 1331         self._get_labels_field(sample_collection)
   1332
   1333         labels_field, is_frame_field = sample_collection._handle_frame_field(

~/work/fiftyone/fiftyone/fiftyone/core/stages.py in _get_labels_field(self, sample_collection)
   1371     def _get_labels_field(self, sample_collection):
   1372         field_name, is_list_field, is_frame_field = _get_labels_field(
-> 1373             sample_collection, self._field
   1374         )
   1375         self._is_frame_field = is_frame_field

~/work/fiftyone/fiftyone/fiftyone/core/stages.py in _get_labels_field(sample_collection, field_path)
   3343
   3344 def _get_labels_field(sample_collection, field_path):
-> 3345     field, is_frame_field = _get_field(sample_collection, field_path)
   3346
   3347     if isinstance(field, fof.EmbeddedDocumentField):

[...]

PR #912, opened 2021-03-17
Build view along with pipeline when aggregating views

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/910.

The issue was that `DatasetView._pipeline()` was passing the final view to `ViewStage.to_mongo(view)`, but technically it should have been passing a view that contains only stages `1-n` when serializing the `n`th stage. In most cases the two are equivalent, but not when certain view stages modify the schema, for example.

```py
import fiftyone.zoo as foz

view = dataset.filter_labels("ground_truth", F("label") == "person")

# now works
view.select_fields("filepath")
```

### DIFF
--- a/fiftyone/core/view.py
+++ b/fiftyone/core/view.py
@@ -49,9 +49,12 @@ class DatasetView(foc.SampleCollection):
         dataset: a :class:`fiftyone.core.dataset.Dataset`
     """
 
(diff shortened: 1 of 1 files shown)


benjaminpkane (approved):
Right, makes sense. LGTM

PR #918, opened 2021-03-19
[WIP] Open Images style detection evaluation

ehofesmann:
Depends on #895 

Adds object detection evaluation following how it is [done for Open Images](https://storage.googleapis.com/openimages/web/evaluation.html). 

There are a lot of similarities between this and COCO evaluation since they are both based off of Pascal VOC mAP calculation. A few notable differences are:
1) Open Images supports class hierarchies, if a hierarchy is provided, then ground truth objects are created for all parent classes of a detection.
2) Open Images has image-level positive and negative labels indicating the classes that are to be evaluated in each sample (others are ignored). If these fields are not specified then all classes are evaluated in the sample.
3) `iscrowd` is called `IsGroupOf`, and matching detections to a crowd ground truth object will result in 1 true positive when calculating the mAP, no matter the number of predictions that were matched with it. If no predictions are matched with it, it counts as 1 false positive.

New arguments are:
* `hierarchy` (None) - optional label hierarchy loaded following the form of Open Images
* `pos_lab_field` (None) - classifications field containing positive image-level labels
* `neg_lab_field` (None) - classifications field containing negative image-level labels
* `expand_hierarchy` (True) - whether to expand the hierarchy of gt detections and label fields
* `expand_pred_hierarchy` (False) - whether to expand hierarchy of predicted detections (standard open images protocol is for models to output all relevant objects in hierarchy)

```python
import fiftyone as fo
import fiftyone.zoo as foz
import random
random.seed(1)

    "open-images", 
    "validation", 
    max_samples=10,
    label_types=["detections","classifications"]
)

# Add some fake predictions
dataset.clone_sample_field("detections", "predictions")
for s in dataset:
    negs = [l.label for l in s.negative_labels.classifications]
    for d in s.predictions.detections:
        d["confidence"] = random.random()
        if negs and random.random()<0.2:
            d["label"] = random.choice(negs)

    s.save()

hierarchy = dataset.info["hierarchy"]

results = fo.evaluate_detections(
    dataset, 
    "predictions", 
    "detections", 
    method="open-images",
    eval_key="eval",
    expand_pred_hierarchy = True, 
    hierarchy=hierarchy,
    pos_lab_field="positive_labels", 
    neg_lab_field="negative_labels"
) 
results.mAP()
```

Since we need the label fields, we cannot just select gt and pred fields in `fiftyone.utils.eval.detection.evaluate_detections` so the property `additional_fields_required` was added to `DetectionEvaluationConfig` that defaults to False but is set to True for `OpenImagesEvaluationConfig`.


fiftyone/utils/eval/detection.py
```diff
@@ -183,6 +183,14 @@ def __init__(
         self.iou = iou
         self.classwise = classwise
 
+    @property
+    def additional_fields_required(self):
```
brimoor, line 187: style nit: we use active tense for method/property names. So prefer `requires_additional_fields` to `additional_fields_required`

fiftyone/utils/eval/openimages.py
```diff
@@ -0,0 +1,699 @@
...
+        iou (None): the IoU threshold to use to determine matches
+        classwise (None): whether to only match objects with the same class
+            label (True) or allow matches between classes (False)
+        iscrowd ("IsGroupOf"): the name of the crowd attribute
+        max_preds (None): the maximum number of predicted objects to evaluate
+            when computing mAP and PR curves. 
+        hierarchy (None): a dict containing a hierachy of classes for
+            evaluation following the structure 
+            ``{"LabelName": label, "Subcategory": [{...}, ...]}``
+        pos_lab_field (None): the name of the field containing
```
brimoor, line 45: nit: it's generally good to avoid abbreviations unless they really save space. So `pred` and `gt` are pretty ubiquitous in our codebase, and `pos` and `neg` are easy to understand. But `lab` only saves two characters compared to `label`, and the latter is more clear, so I'd recommend `pos_label_field` and `neg_label_field`.

Also, I believe these fields need to be `Classification` label fields, right? Whenever you reference a specific type like that, it's helpful to explicitly mention the class type `:class:`fiftyone.core.labels.Classification` so that readers of the API docs online can clickthrough to learn about the label type.

fiftyone/utils/eval/openimages.py
```diff
@@ -0,0 +1,699 @@
...
+        hierarchy (None): a dict containing a hierachy of classes for
+            evaluation following the structure 
+            ``{"LabelName": label, "Subcategory": [{...}, ...]}``
+        pos_lab_field (None): the name of the field containing 
+            image-level classifications that specify which classes should be 
+            evaluated in the image
+        neg_lab_field (None): the name of the field containing 
+            image-level classifications that specify which classes should not 
+            be evaluated in the image
+        expand_hierarchy (True): bool indicating whether to expand ground truth
```
brimoor, line 51: `expand_gt_hierarchy` for consistency with `expand_pred_hierarchy`?

fiftyone/utils/eval/openimages.py
```diff
@@ -0,0 +1,699 @@
...
+        self.max_preds = max_preds
+        self.hierarchy = hierarchy
+        self.pos_lab_field = pos_lab_field
+        self.neg_lab_field = neg_lab_field
+        self.expand_hierarchy = expand_hierarchy
+        self.expand_pred_hierarchy = expand_pred_hierarchy
+
+        if expand_pred_hierarchy:
+            if not hierarchy:
+                logger.info(
```
brimoor, line 90: suggestion: use `logger.warning()` when you are overriding the user's parameter value

fiftyone/utils/eval/openimages.py
```diff
@@ -0,0 +1,699 @@
...
+        if self.hierarchy:
+            (
+                self.hierarchy_keyed_parent,
+                self.hierarchy_keyed_child,
+                _,
+            ) = self._build_plain_hierarchy(self.hierarchy, skip_root=True)
+
+    @property
+    def method(self):
+        return "openimages"
```
brimoor, line 113: "open-images"

fiftyone/utils/eval/openimages.py
```diff
@@ -0,0 +1,699 @@
...
+
+    @property
+    def additional_fields_required(self):
+        return True
+
+    # Parse hierarchy, code from:
+    # https://github.com/tensorflow/models/blob/ec48284d0db7a67ab48a9bc13dc29c643ce0f197/research/object_detection/dataset_tools/oid_hierarchical_labels_expansion.py#L77
+    def _update_dict(self, initial_dict, update):
```
brimoor, line 121: nit: these private methods don't need to be class methods (they don't use `self`). In such cases I prefer to make them top-level private methods at the end of the module

fiftyone/utils/eval/openimages.py
```diff
@@ -0,0 +1,699 @@
...
+                initial_dict[key] = set(value_list)
+
+    def _build_plain_hierarchy(self, hierarchy, skip_root=False):
+        """Expands tree hierarchy representation to parent-child dictionary.
+
+        Args:
+            hierarchy: labels hierarchy .
+            skip_root: if true skips root from the processing (done for the case when all
+                classes under hierarchy are collected under virtual node).
+        Returns:
```
brimoor, line 142: nit: remove periods from args (for consistency with rest of codebase) and and newline before `Returns`

fiftyone/utils/eval/openimages.py
```diff
@@ -0,0 +1,699 @@
...
+        """Generates aggregate evaluation results for the samples.
+
+        This method performs Open Images-style evaluation as in 
+        :meth:`evaluate_image` to generate precision and recall curves for the
+        given IoU in ``self.config.iou``. In this case, a
+        :class:`OpenImagesDetectionResults` instance is returned that can
+        provide the mAP and PR curves.
+
+        Args:
+            samples: a :class:`fiftyone.core.SampleCollection`
```
brimoor, line 267: :class:`fiftyone.core.collections.SampleCollection`

brimoor (changes requested):
Glad to see this was quick to implement! I just saw some minor cosmetic things after a quick review

ehofesmann:
> Glad to see this was quick to implement! I just saw some minor cosmetic things after a quick review

I followed the instructions for their written instructions for performing evaluation. I still need to get an environment set up where I can pass some data through the official TF implementation to make sure the numbers are the same. I'm sure there will be some minor variations but god that code is hard to read...

Issue #922, opened 2021-03-23
[BUG] Evaluate detections without confidences


ehofesmann:
Currently, if the `pred_field` of `evaluate_detections` does not contain confidences, an error will occur:
```
inds = np.argsort(-np.array(confs), kind="mergesort")
TypeError: bad operand type for unary -: 'NoneType'
```
https://github.com/voxel51/fiftyone/blob/ca927c7aa354652f848e88d847fa4464a595ba4d/fiftyone/utils/eval/coco.py#L226

A check should be implemented to see if there are any `None` values in `confs` and raise a more informative error about the need for confidence values in the `pred_field` if you want to `compute_mAP`.

Alternatively, these `confs` could just be initialized to `1` by default if they are `None`, but I feel like it may be better to require the user to add confidences themselves.

PR #923, opened 2021-03-23
raise exception for None confidence in pred_field when computing mAP

dadounhind:

A check to see if there are any None values in confs and raise a more informative error about the need for confidence values in the pred_field if you want to compute_mAP. 

Resolves https://github.com/voxel51/fiftyone/issues/922.

All tests passed except :  FAILED unittests/view_tests.py::ViewExpressionTests::test_array - AttributeError: module 'eta.core.utils' has no attribute 'is_container'

        notes for FiftyOne users.

### DIFF
--- a/fiftyone/utils/eval/coco.py
+++ b/fiftyone/utils/eval/coco.py
@@ -226,8 +226,9 @@ class COCOEvaluation(DetectionEvaluation):
                 confs = [p[3] for p in tp] + [p[3] for p in fp]
                 if None in confs:
                     raise ValueError(
-                        "Average precision is a ranking based metric, detections in pred_field"
-                        " sample should have a confidence attribute "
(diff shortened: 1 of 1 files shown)


ehofesmann (changes requested):
Thanks a lot, @dadounhind, that error is much more helpful!

Just a minor reminder, with the developer installation of FiftyOne from source through `install bash -d`, the `-d` flag will add pre-commit hooks that lint and clean code when committing it.

This would format the code to something like:
```python
                if None in confs:
                    raise ValueError(
                        "Average precision is a ranking based metric,"
                        " detections in pred_field should have a confidence"
                        " attribute"
                    )
```

ehofesmann (approved):
Looks good, thanks!

Issue #924, opened 2021-03-23
[BUG] Copying views is broken for stages with `VALUE`


benjaminpkane:
I'm having difficulty understanding the logic of `ViewExpression.__deepcopy__` but it is where the issue is, I believe.

```py
from copy import copy

import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone.core.expressions import VALUE

F = fo.ViewField

dataset = foz.load_zoo_dataset("quickstart")
view = dataset.set_field("ground_truth.tags", F("$ground_truth.detections").reduce(VALUE.extend(F("tags")), []))
print(view._stages[-1]._expr)

copied_view = copy(view)
print(copied_view._stages[-1]._expr)
```

OUT:
```
{
    '$reduce': {
        'input': '$ground_truth.detections',
        'initialValue': [],
        'in': {'$concatArrays': ['$$value', '$$this.tags']},
    },
}
{
    '$reduce': {
        'input': '$ground_truth.detections',
        'initialValue': [],
        'in': {'$concatArrays': ['$value', '$tags']},
    },
}

```

The copied `expr` has one of its `$` removed.

brimoor:
Hmm looks like this is the problem:

```py
from copy import deepcopy
from fiftyone.core.expressions import VALUE

print(VALUE)
# '$$value'

print(deepcopy(VALUE))
# '$value'
```

PR #925, opened 2021-03-23
Fixing ViewField copying

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/924.


benjaminpkane (approved):
LGTM! Thanks for for fixing :raised_hands:. Still learning this side of views.

PR #926, opened 2021-03-23
Filter saving, and labels tags and tag matching in the Sidebar

### LINKED ISSUE #915: [FR] Add support for matching by sample/label tags in the fields sidebar
https://github.com/voxel51/fiftyone/pull/906 introduces *amazing* functionality for modifying the tags of samples/labels in datasets. This naturally leads to the following feature request: give me a simple way to filter my dataset to show only samples/labels with a given tag!

There are two conventions in the fields sidebar:

1. **Checkbox selection.** Each field has a checkbox, where:
- unchecked = field values hidden from grid
- checked = field values visible in grid

2. **Filtering widget.** Fields may have a `v` caret that opens a filtering modal that selectively chooses which values in that field to show

The tags field currently bends convention 1 because its individual values are given their own checkboxes. And there is no filtering widget.

I have two proposals for how to support tag filtering in the sidebar:

### Proposal 1

Rename the `TAGS` section to `SAMPLE TAGS`, and add a `LABEL TAGS` section. Within each section, continue to blow out each tag into separate entries as we currently do.

Add an eye icon on the RHS of each entry (in the same location as the `v` widget for other fields). Clicking the eye icon creates an extended view filter: only samples/labels matching the tag are shown. Multiple eye icons can be selected to match on multiple tags. Clicking an active eye icon removes the tag from the filter. If no eye icons are activated, no filtering is performed.

### Proposal 2

Replace the existing tags section with two entries:

```
[ ] sample tags
[ ] label tags
```

Toggling the checkbox for `sample tags` would show/hide all tags on all samples.

Now, what should toggling the `label tags` checkbox do? Here's a couple ideas:

(a) **Visual option.** All labels with tags get a visual emphasis. Reusing the dotted border may be confusing since it means selection in other contexts, but something similar to this (e.g,.  

(b) **Textual option.** All labels are decorated with their tags in text form. Such as:

- Classification tags: `<label>` could change to `<label> (<tags>)`
- Detections: `<tags>` appears where object `labels`  the label selection  

### Visual to spur inspiration

110369302-702c2600-8007-11eb-893f-f1af527ce9ac

### LINKED ISSUE #916: [FR] Add option to save extended view in App
We've talked for awhile about letting the user "save" their extended view, which would convert the fields sidebar state to additional stages in the view bar and reset the field filters.

Now that https://github.com/voxel51/fiftyone/pull/906 has introduced a row of options above the sample grid, how about adding a save icon to this tray whenever an extended view exists? This would provide a visual indication that the user can take action to save their view.

benjaminpkane:
Issues:
* Resolves #916 
* Resolves #915 

Changes:
* Adds saving the current filter state in the grid sidebar to the session view
* Adds label tags to the sidebar
* Adds matching on tags in the sidebar

Screenshot from 2021-03-23 13-44-35


benjaminpkane:
Label fields that are checked have an affect on the number of tags one would expect to see in counts. Which is in the very least confusing me...

So, technically, one might expect label tag counts to possible change when checking/unchecking label fields.

brimoor:
> What label fields are checked have an affect on the number of tags one would expect to see in counts.It seems like regardless of what decision is made, some users may find themselves confused.
> 
> One might expect label tag counts to possibly change when checking/unchecking label fields. I'm reluctant to kick off a new statistics request when a label field is checked/unchecked.

I think a good solution to this is that the `<label tag>: <count>` is computed at the sample-level regardless of how many fields the count is distributed over and whether those fields are visible via the label field checkboxes

brimoor:
In general, modulo a few bugs that we encountered while testing interactively, I think this feature is working well!

Here's my only thoughts:
- Make the tooltips when hovering over the checkboxes and target/down caret icon more informative, like `Only show samples with this tag` instead of `Only matches`. Then people will get the idea I think
- Adding a similarly informative tooltip to the save icon would be prudent I think. Like `Save current field filters as view stages`
- Consider alternatives to the target icon, which still doesn't feel like a "matching"-type operation

The only other option that comes to mind relative to the target icon and the difference between this selection UI and the label filters would be to consider a middle ground like the following:
- Leave the per-tag entries on the sidebar, but introduce the same down caret as for label/scalar filters. Clicking the down caret would reveal a filter tray with a single checkbox that says something like `Only show <tag> samples/labels`. The pro here is that all fields would have the same down caret; the con is that it takes two clicks to show a single tag, and the fact that this matching is OR-ed across all fields may be confusing (other filters are AND-ed). In fact, for this last reason, not sure I like this...

benjaminpkane:
Will undraft today. Working through some tricky queries that need to happen.

benjaminpkane:
Ok, things aren't quite done. Will demo tomorrow.

benjaminpkane:
Undrafting. After trying a few different icons, I have settled on the `eye` as it is likely to be the most familiar to users in this situation.

fiftyone/core/collections.py
```diff
@@ -4980,7 +4995,7 @@ def _record_list_field(field_name):
     if root_field_name in ("id", "_id"):
         root_field = None
     elif root_field_name not in schema:
-        if not allow_missing:
+        if not allow_missing and not root_field_name.startswith("_"):
```
brimoor, line 4987: per our online discussion: I think the best approach here is to add an `_allow_missing` parameter to `SetField()` so and use that instead of the `root_field` check here

fiftyone/core/stages.py
```diff
@@ -3206,9 +3217,10 @@ def _make_labels_pipeline(self, sample_collection):
         # our intention is not to remove other fields from the schema, only to
         # empty their sample fields in the returned view
         #
-        stage = SelectFields(list(self._labels_map.keys()))
-        stage.validate(sample_collection)
-        pipeline.extend(stage.to_mongo(sample_collection))
+        if self._select_fields:
```
brimoor: Per our online discussion- I think this parameter should be removed. The serialized `SelectFields()` stage here shouldn't affect the schema of the dataset, its just here to omit *all* labels from fields that don't appear in `_labels_map`, which is what the user wants (otherwise they would have included those fields in their selection).

It occurs to me that this might have been added in the spirit of helping the extended view get access to *all* data in case the user wants to open a sample in the expanded modal and view all data. But, I think the workaround/hack for that might be elsewhere, not here.

fiftyone/core/stages.py
```diff
@@ -3244,9 +3256,10 @@ def _make_pipeline(self, sample_collection):
         # our intention is not to remove other fields from the schema, only to
         # empty their sample fields in the returned view
         #
-        stage = SelectFields(fields)
-        stage.validate(sample_collection)
-        pipeline.extend(stage.to_mongo(sample_collection))
+        if self._select_fields:
```
brimoor: ditto

brimoor (approved):
LGTM! I left a few comments inline per our online discussion.

My only other comments are:
- Don't include a trivial `SelectLabels()` stage when saving an extended view that doesn't involve label selection
- Consider adding more information to the checkboxes in the fields sidebar, for consistency with the level of detail in the tooltip for the eye icon:
    - `Show <tag-field>` -> `Show <tag-field> tags` 
    - `Show <label-field>` -> `Show <label-field> labels`
    - `Show <scalar-field>` -> `Show <scalar-field> values`

PR #929, opened 2021-03-24
Colab installs from source

benjaminpkane:
Adds support for installing `fiftyone` from source in a Colab notebook.

The below can be used to test in a Colab notebook.
```
%%shell
git clone --depth 1 -b install https://github.com/voxel51/fiftyone.git
cd fiftyone
bash install.bash
```

Restarting the runtime after the install finishes is required. I'm unsure if that is avoidable.


benjaminpkane:
Ok, still need to fix at least one thing before this is functional

brimoor (approved):
LGTM. I tested by running [the quickstart](https://colab.research.google.com/github/voxel51/fiftyone-examples/blob/master/examples/quickstart.ipynb) in Colab with source install instructions in place of `pip install fiftyone`.

@benjaminpkane LMK if you don't agree with the other improvements I made to the install script. Now ETA source installs and manual MongoDB installs are only opt-in, and `fiftyone-brain` will be installed by default, unless you are a Voxel51 developer (`-v`).

In summary, the following are the syntaxes for the normal personas:

```
bash install.bash  # source installers, Google Colab, etc
bash install.bash -d  # non-Voxel51 developers
bash install.bash -v  # Voxel51 developers
```

brimoor:
> Ok, still need to fix at least one thing before this is functional

I guess there is one issue of note, which is that `bash install.bash -e` is **required** if the bleeding edge of FiftyOne happens to require the bleeding edge of ETA.

brimoor:
> > Ok, still need to fix at least one thing before this is functional
> 
> I guess there is one issue of note, which is that `bash install.bash -e` is **required** if the bleeding edge of FiftyOne happens to require the bleeding edge of ETA.

Okay, and related to this, I guess the source ETA install needs to come *after* the `pip install fiftyone`, which will install `voxel51-eta` from pip. (I wonder how this wasn't causing problems before??)

Anyway, I think it works now.

Issue #208, opened 2020-06-17
`fiftyone app connect` still tries to load the dataset locally


lethosor:
I'm testing on a Windows machine with no datasets available and connecting to a Linux machine with `cifar10` downloaded. Running `fiftyone app launch cifar10 --remote` on Linux, then `fiftyone app connect` on Windows brings up a working dashboard but produces these errors in the Windows console whenever I click on an image:

```
Exception in thread Thread-11:
Traceback (most recent call last):
  File "c:\users\test\env\lib\site-packages\mongoengine\queryset\base.py", line 260, in get
    result = next(queryset)
  File "c:\users\test\env\lib\site-packages\mongoengine\queryset\base.py", line 1590, in __next__
    raw_doc = next(self._cursor)
  File "c:\users\test\env\lib\site-packages\pymongo\cursor.py", line 1164, in next
    raise StopIteration
StopIteration

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "c:\users\test\env\lib\site-packages\fiftyone\core\dataset.py", line 1616, in _load_dataset
    _meta = foo.ODMDataset.objects.get(name=name)
  File "c:\users\test\env\lib\site-packages\mongoengine\queryset\base.py", line 263, in get
    raise queryset._document.DoesNotExist(msg)
fiftyone.core.odm.dataset.DoesNotExist: ODMDataset matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Program Files\Python38\lib\threading.py", line 932, in _bootstrap_inner
    self.run()
  File "C:\Program Files\Python38\lib\threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "c:\users\test\env\lib\site-packages\socketio\client.py", line 616, in _handle_eio_message
    self._handle_event(pkt.namespace, pkt.id, pkt.data)
  File "c:\users\test\env\lib\site-packages\socketio\client.py", line 500, in _handle_event
    r = self._trigger_event(data[0], namespace, *data[1:])
  File "c:\users\test\env\lib\site-packages\socketio\client.py", line 554, in _trigger_event
    return self.namespace_handlers[namespace].trigger_event(
  File "c:\users\test\env\lib\site-packages\socketio\namespace.py", line 18, in trigger_event
    return getattr(self, handler_name)(*args)
  File "c:\users\test\env\lib\site-packages\fiftyone\core\client.py", line 72, in on_update
[...]

PR #930, opened 2021-03-24
Updating remote session instructions

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/208.

This PR updates the remote session instructions to first recommend manual port forwarding on your local machine, which doesn't require a local FiftyOne install.

It also resolves https://github.com/voxel51/fiftyone/issues/208 by re-implementing the `fiftyone app connect` CLI method to simply execute the port forward command and then open a browser tab without calling `fo.launch_app()`.

I thought about removing `fiftyone app connect` completely, but, now that it is harmless, I think its okay to keep it as a convenience to users (as a second option). The one thing that *is* removed is the `--desktop` flag, which goes too far.

```py
# On remote machine
import fiftyone as fo
import fiftyone.zoo as foz

session = fo.launch_app(dataset, remote=True)
```

```
You have launched a remote App on port 5151. To connect to this App from another
machine, issue the following command to configure port forwarding:

ssh -N -L 5151:127.0.0.1:5151 [<username>@]<hostname>

where `[<username>@]<hostname>` refers to your current machine. The App can
then be viewed in your browser at http://localhost:5151.

Alternatively, if you have FiftyOne installed on your local machine, just run:

fiftyone app connect --destination [<username>@]<hostname> --port 5151

See https://voxel51.com/docs/fiftyone/user_guide/app.html#remote-sessions
for more information about remote sessions.
```

### DIFF
--- a/app/src/containers/Setup.tsx
+++ b/app/src/containers/Setup.tsx
@@ -32,8 +32,7 @@ const Code = styled.pre`
   border-radius: 3px;
 `;
 
-const remoteSnippet = `
-import fiftyone as fo
+const remoteSnippet = `import fiftyone as fo
 
 # Load your FiftyOne dataset
 dataset = fo.load_dataset(...)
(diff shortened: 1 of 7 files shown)


benjaminpkane (approved):
LGTM!

PR #937, opened 2021-03-29
Hide label attributes by default in Player51

benjaminpkane:
Per user feedback, it makes sense to not render attributes in detection bounding boxes now that the tooltip exists and is shown by default.


brimoor (approved):
LGTM

PR #938, opened 2021-03-29
Fixing detection evaluation for videos

brimoor:
The detection evaluation changes in ``fiftyone==0.7.4`` did not consider the case of evaluating videos. This PR fixes that.

@ehofesmann you will likely need to follow suit per this PR to support videos for Open Images-style evaluation in https://github.com/voxel51/fiftyone/pull/895.

Example usage taken from the docs, with `compute_mAP = True` added to test the full COCO evaluation workflow:

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone import ViewExpression as E

    "quickstart-video", dataset_name="video-eval-demo"
)

classes = dataset.distinct("frames.ground_truth.detections.label")

def jitter(val):
    if isinstance(val, list):
        return [jitter(v) for v in val]

    if random.random() < 0.10:
        return random.choice(classes)

    return val

dataset.clone_frame_field("ground_truth", "predictions")

gt_labels = dataset.values("frames.ground_truth.detections.label")
pred_labels = jitter(gt_labels)

dataset.set_values("frames.predictions.detections.label", pred_labels)
dataset.set_field(
    "frames.predictions.detections.confidence", E.rand()
).save("frames.predictions")

    "frames.predictions",
    gt_field="frames.ground_truth",
    eval_key="eval",
    compute_mAP=True,
)

print(dataset)
print(dataset.get_evaluation_info("eval"))
print("mAP: %f" % results.mAP())

results.print_report()
```


ehofesmann (approved):
LGTM!

PR #939, opened 2021-03-29
Adding trigonometric ViewExpressions and Gaussian random variable support

brimoor:
Adds `ViewExpression`s for computing trigonometric functions `cos()`, `sin()` and `tan()` and their hyperbolic and inverse variants.

Also adds a `ViewExpression.randn()` expression that can generate Gaussian random numbers.

Example usage:

```py
import numpy as np
import matplotlib.pyplot as plt
import scipy.stats as sps

import fiftyone as fo
from fiftyone import ViewExpression as E


def plot_hist(counts, edges, normalized=False):
    counts = np.asarray(counts)
    edges = np.asarray(edges)
    left_edges = edges[:-1]
    widths = edges[1:] - edges[:-1]
    if normalized:
        counts = counts / (counts * widths).sum()

    plt.bar(left_edges, counts, width=widths, align="edge")


dataset = fo.Dataset()
dataset.add_samples(
    [fo.Sample(filepath="image%d.jpg" % i) for i in range(10000)]
)

# `E.randn()` generates a new Gaussian random variable each time it is called
dataset.add_sample_field("randn", fo.FloatField)
dataset.set_field("randn", E.randn()).save("randn")

counts, edges, other = dataset.histogram_values("randn", bins=50)

plot_hist(counts, edges, normalized=True)

x = np.arange(min(edges), max(edges), 0.001)
plt.plot(x, sps.norm.pdf(x), "r")

plt.show(block=False)
```

gaussian

### DIFF
--- a/fiftyone/core/expressions.py
+++ b/fiftyone/core/expressions.py
@@ -11,6 +11,8 @@ import warnings
 
 import bson
 
+import numpy as np
+
 import eta.core.utils as etau
 
 from fiftyone.core.odm.document import MongoEngineBaseDocument
@@ -939,7 +941,7 @@ class ViewExpression(object):
         return ViewExpression({"$ln": self})
 
     def log(self, base):
-        """Computes logarithm base ``base`` of this expression, which must
+        """Computes the logarithm base ``base`` of this expression, which must
         resolve to a numeric value.
 
         Examples::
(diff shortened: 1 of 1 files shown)


benjaminpkane (approved):
LGTM!

One comment is that adding warnings or best practices examples  (saving the resulting values, for example)  to the `rand()` and `randn()` documentation might be helpful. `rand` and `randn()` cannot be reproduced and therefore have no use in the App, etc. when used on their own.

brimoor:
> LGTM!
> 
> One comment is that adding warnings or best practices examples (saving the resulting values, for example) to the `rand()` and `randn()` documentation might be helpful. `rand` and `randn()` cannot be reproduced and therefore have no use in the App, etc. when used on their own.

yeah good point. I added `warning` notes to the docstrings for each method

Issue #936, opened 2021-03-29
[BUG] Dataset statistics fail to load for larger datasets


brimoor:
There appears to be a wide variation in the time required for dataset statistics to load in the fields sidebar for larger datasets.

When working with the validation split of [bdd100k](https://voxel51.com/docs/fiftyone/user_guide/dataset_zoo/datasets.html#bdd100k), for example, I find that the loading icons in the screenshot below will persist for as little as 1-2 seconds to as many as 10-15 seconds or more. In the latter case, I typically click the cube to refresh the page, which will typically cause the stats to load faster if I try it a couple times.

This problem is not observed on datasets that are an order of magnitude smaller, such as [quickstart](https://voxel51.com/docs/fiftyone/user_guide/dataset_zoo/datasets.html#quickstart).

<img width="1108" alt="Screen Shot 2021-03-29 at 9 11 26 AM" src="https://user-images.githubusercontent.com/25985824/112842121-39c65180-906f-11eb-8a27-643d206512d7.png">

benjaminpkane:
These statistics are taking between 4 and 5 seconds for me. Never 10-15 seconds.

You can get a sense of speeds with the below, although it seems like `asyncio` adds some overhead. A regular `pymongo` cursor yields ~3 seconds for me.

```py
import eta.core.utils as etau

import fiftyone as fo
import fiftyone.core.state as fos
import fiftyone.types as fot

bdd = fo.Dataset.from("/path/to/bdd/validation", dataset_type=fot.BDDDataset)

stats = fos.DatasetStatistics(bdd)
aggs = stats.aggregations + stats.exists_aggregations

with etau.Timer():
    bdd.aggregate(aggs)
```

I can't see things taking longer than ~5 seconds. At that point the response will likely never resolve, which I am also looking into. It might be best to take this offline tomorrow morning.

benjaminpkane:
Resolved by #926

Issue #935, opened 2021-03-29
[BUG] App often crashes on develop when opening expanded sample modal


brimoor:
On `develop`, the App frequently crashes for me when trying to view a sample from [bdd100k](https://voxel51.com/docs/fiftyone/user_guide/dataset_zoo/datasets.html#bdd100k) in the expanded sample modal of the App.

The view loads but then errors out a few seconds after load.

```
index.js:7 TypeError: Cannot read property '1' of undefined
    at index.js:176
    at c (index.js:11)
    at f (index.js:11)
    at u (index.js:11)
    at Object.d [as get] (index.js:11)
    at vW (index.js:11)
    at IW (index.js:11)
    at pG (index.js:11)
    at TE (index.js:11)
    at FN (index.js:11)
```

benjaminpkane:
Likely resolved by #926, though I have not been able to reproduce.

If there is a specific view this occurs in that would be helpful. I cannot reproduce on the dataset view itself.

benjaminpkane:
Resolved by #942

Issue #921, opened 2021-03-22
[BUG] Settings menu erroneously appears in App screenshots


brimoor:
On `develop`, when freezing a session in a Jupyter notebook via `session.freeze()`, I often see this:

![download](https://user-images.githubusercontent.com/25985824/112070854-f33f9700-8b44-11eb-958c-5c934e0e79c7.png)

benjaminpkane:
I'm having trouble reproducing, but will add logic to p51 to force hide options in thumbnail mode.

brimoor:
Haven't seen this in ages.

PR #942, opened 2021-03-30
App issues

### LINKED ISSUE #934: [FR] Consider allowing multiple rows of image tags in the sample grid
The old behavior in the App was to let image tags stack up into multiple rows. More recently, the design was changed to always squash labels onto one row, which, as the example below illustrates, can make them rather unreadable for more densely annotated datasets.

There's at least two separate decision axes here:
- How to handle long individual tags?
- One row or multiple rows?

On balance, I find that the current design abbreviates tags too aggressively such that I can't fully read them. A recent example I had was an image with two tags: `[Parasitized] [Parasitized]`, but it was abbreviated as something like `[Pa...] [Pa...]` in a Jupyter notebook, which made it tough to parse.

On balance, I'm inclined to say that allowing multiple rows may be better. With that said, baking some level of abbreviation to handle `very long tags when they appear` would be appropriate. Perhaps the ideal would be some intelligent logic that would abbreviate more aggressively when there are more tags?

A related thought is that implementing a grid zoom option per https://github.com/voxel51/fiftyone/issues/749 would help to provide more real estate for individual samples in an environment with a fixed width App with dense annotations.

download

benjaminpkane:
Resolves #934 

With respect to #935, #936, I believe they have been implicitly resolved by #926. Will close unless there are objections.

Also resolves undocumented issues with sample-level label tags counts in the grid (#926) when filtered by a label field.

```py
import fiftyone as fo

tags = ["one", "two", "three", "four", "five", "six", "seven", "eight", "nine", "ten", "eleven", "twelve", "thirteen", "fourteen", "fifteen", "waytolongforatagbutwhoknowswhatuserswilldo"]
dataset.set_field("tags", tags).save()

session.refresh()
```
Screenshot from 2021-03-30 14-59-35


benjaminpkane:
Ok, working through a bug that is also on `develop`.

brimoor (approved):
LGTM.

I ran a notebook with ~10 App instances and still observed #921 once. I don't think this PR needs to be blocked to look into that more, but perhaps we should keep the issue open with `needs-evidence` or similar.

I'm comfortable with closing #935 and #936 per your comments in the description.

PR #943, opened 2021-03-31
Release v0.8.0

benjaminpkane:


docs/source/release-notes.rst
```diff
@@ -3,6 +3,160 @@ FiftyOne Release Notes
...
+  mask values (pixel values) using the new
+  :attr:`Dataset.mask_targets <fiftyone.core.dataset.Dataset.mask_targets>`
+  and
+  :attr:`Dataset.default_mask_targets <fiftyone.core.dataset.Dataset.default_mask_targets>`
+  fields
+- Fixed hiding, clearing, and only showing selected samples in the samples grid
+
+Brain
+^^^^^
+- Added a :meth:`compute_visualization() <fiftyone.brain.compute_visualization>` method that uses embeddings and dimensionality reduction methods to generate interactive visualizations of the samples and/or labels in a dataset. Check out :ref:`this user guide page <XXXXXXXX>` for details. Features include:
```
brimoor, line 33: FYI- for any OCD folks out there, these lines intentionally don't have line breaks at 80 characters. If they do, it somehow messes up the parsing of the nested lists.

In fact, RST format actually wants nested lists to have newlines before and after them (and this is how nested lists in docstrings are written), but something about the style sheet for the HTML pages causes the indentation to be gigantic for these nested lists.

Instead, it seems that *not* including the newlines will still produce nested lists, but with smaller indents, which looks better in HTML format.

benjaminpkane:
Will tag on approval. RC's look good.

brimoor:
> Will tag on approval. RCs look good.

FYI https://github.com/voxel51/fiftyone/pull/951 has a couple code changes that I'd like to include in the release

brimoor (approved):

Issue #944, opened 2021-03-31
[FR] More informative errors and better attribute handling on `Session`


benjaminpkane:
It would be beneficial to add more informative and standard error handling that informs users of incorrect assignments to `Session` attributes (incorrect `Session` use, in general)

As an example, the `Session.view` setter does not accept `Dataset`s only `DatasetView`s. Perhaps this can be made more flexible, as well.

Current server error from `session.view = dataset`:
```
Server Error

An exception has been raised by the server. Your session has been reverted to its previous state.

Traceback (most recent call last):
  File "main.py", line 179, in wrapper
    result = await func(self, *args, **kwargs)
  File "main.py", line 425, in on_message
    await event(self, **message)
  File "main.py", line 543, in on_update
    StateHandler.state = fos.StateDescription.from_dict(state).serialize()
  File "/home/ben/code/fiftyone/fiftyone/core/state.py", line 110, in from_dict
    view = fov.DatasetView._build(dataset, stages)
  File "/home/ben/code/fiftyone/fiftyone/core/view.py", line 716, in _build
    stage = fost.ViewStage._from_dict(stage_dict)
  File "/home/ben/code/fiftyone/fiftyone/core/stages.py", line 203, in _from_dict
    view_stage_cls = etau.get_class(d["_cls"])
TypeError: string indices must be integers
```

PR #945, opened 2021-03-31
Adding type checks for Session property setters

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/944.

#### Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

other_dataset = dataset.clone()

# these are all allowed
session.view = dataset.view()
session.view = other_dataset.view()
session.dataset = other_dataset
session.config = session.config.copy()
session.view = None
session.dataset = None
session.config = None

# these are not allowed and now raise informative error messages
session.dataset = dataset.view()
session.view = other_dataset
session.config = {}
session.plots = {}
```

#### Side note

I thought about performing the following silent "type casts", but I think its preferable to force the user to provide the correct type data to ensure they understand the difference between changing a view into a dataset versus changing the entire dataset
- `session.view = dataset` --> `session.view = dataset.view()`
- `session.dataset = view` --> `session.dataset = view._dataset`

### DIFF
--- a/fiftyone/core/session.py
+++ b/fiftyone/core/session.py
@@ -18,6 +18,7 @@ import fiftyone as fo
 import fiftyone.constants as focn
 import fiftyone.core.dataset as fod
 import fiftyone.core.client as foc
+from fiftyone.core.config import AppConfig
 import fiftyone.core.context as focx
 import fiftyone.core.frame as fof
 import fiftyone.core.media as fom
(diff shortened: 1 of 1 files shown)


benjaminpkane (approved):
Fast. :rocket: LGTM!

brimoor:
> I think other `Session` attributes could use errors as well. Will make a PR ~today.

What do you have in mind? The only other setters are `selected` and `selected_labels`, but I tend to hesitate to provide validation for potentially long lists of data. Perhaps the most I would do is a minimal check that the provided data is an iterable and the first element has the correct format

benjaminpkane:
> > I think other `Session` attributes could use errors as well. Will make a PR ~today.
> 
> What do you have in mind? The only other setters are `selected` and `selected_labels`, but I tend to hesitate to provide validation for potentially long lists of data. Perhaps the most I would do is a minimal check that the provided data is an iterable and the first element has the correct format

You're right. Punting.

PR #946, opened 2021-03-31
Validate keyword arguments in Session's init

benjaminpkane:
Adds validation to `Session.__init__` keyword arguments

```py
import fiftyone as fo

session = fo.launch_app(dataset=fo.Dataset().view())
```


brimoor (approved):
LGTM (so there were more places to address 😉 )

PR #950, opened 2021-04-01
Update docs homepage

ehofesmann:
A refresh of the homepage for [fiftyone.ai](http://fiftyone.ai), updating messaging and images.

image

I think there still needs to be an update to voxel51.com in order for the video to be loaded properly, not sure how Ben is loading that.


benjaminpkane (approved):
LGTM!

PR #955, opened 2021-04-02
Embeddings tutorial

brimoor:
Adds a tutorial for visualizing embeddings using the MNIST and BDD100K datasets as exemplars.

Screen Shot 2021-04-05 at 12 55 01 PM

Screen Shot 2021-04-05 at 12 55 10 PM


benjaminpkane (approved):
LGTM! Science + Visualization!!!

PR #959, opened 2021-04-06
Release hotfixes

brimoor:
Fixes some typos and adds some other minor enhancements to the new documentation introduced with v0.8.0 release.

Also fixes a bug that prevented `SegmentationResults` from being loaded via `dataset.load_evaluation_results()`.


brimoor:
FYI @benjaminpkane I already deployed these docs fixes to production

benjaminpkane (approved):
LGTM. I will tag right now unless there is a reason to wait

I'm tempted to start using semantic versioning where these changes would go out as v0.8.1.

I can release as v0.8.0.1 if there are objections.

brimoor:
> LGTM. I will tag right now unless there is a reason to wait
> 
> I'm tempted to start using semantic versioning where these changes would go out as v0.8.1.
> 
> I can release as v0.8.0.1 if there are objections.

I believe pretty much everything is tested so I don't anticipate any further changes.

No preference on versioning. Up to you. My only observation is that the code portion of these changes is very minor

PR #963, opened 2021-04-13
Grid zooming and slider improvements

### LINKED ISSUE #584: [FR] Limit integer label filter sliders to integers
Here's a dataset with some scalar values on its samples that have type `IntField`:

```
Name:           quickstart
Media type      image
Num samples:    200
Persistent:     False
Info:           {}
Tags:           ['validation']
Sample fields:
    media_type:   fiftyone.core.fields.StringField
    filepath:     fiftyone.core.fields.StringField
    tags:         fiftyone.core.fields.ListField(fiftyone.core.fields.StringField)
    metadata:     fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.metadata.Metadata)
    ground_truth: fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.labels.Detections)
    uniqueness:   fiftyone.core.fields.FloatField
    predictions:  fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.labels.Detections)
    tp_iou_0_75:  fiftyone.core.fields.IntField
    fp_iou_0_75:  fiftyone.core.fields.IntField
    fn_iou_0_75:  fiftyone.core.fields.IntField
```

It would be desirable if the label filter in the App respected the int data type and locked the slider into int values, rather than allowing/showing `[0.00    [34.92, 81.48]     97.00]` in the App

Screen Shot 2020-10-07 at 2 01 12 PM

### LINKED ISSUE #749: [FR] Add support for adjusting the number of samples shown in App sample grid
Currently the sample grid in the App always shows the same number of samples, regardless of the width of the App window. However, as a user, I often find myself wishing to see *more* samples in the grid.

Two ideas come to mind here:
(1) consider making the sample grid have fixed *absolute* width cells, such that increasing the width of the App window would result in more samples per row at some point
(2) provide a "zoom" slider in the upper right of the grid that allows the user to manually adjust the samples-per-grid setting

Currently the sample grid dynamically fits the images so that every row is flush on the left and right sides of the grid with equal height, which is awesome and should definitely remain.

Therefore, if we adopted (1), it would only work in "bins" so that as the App window width increases, more images are allowed per-row occasionally, while still satisfying the current flush-edges layout. 

It's probably simpler just to let the user manually control image density via option (2). IE, changing the App window width will not affect the grid layout; only adjusting the zoom will do that (always retaining flush edges).

**Current behavior**

Screen Shot 2021-01-08 at 10 04 02 AM

Screen Shot 2021-01-08 at 10 03 51 AM

benjaminpkane:
### Issues
* Resolves #749
* Resolves #584 

### Behavior

The min aspect ratio of the rows is determined by the zoom slider. Valid values fall between `[1, 9]`. The number of samples per page request is `ceil(min_aspect_ratio  * 4)`.

Currently, the default value remains 5, no matter the window size (width)

Integer scalar fields now have integer step sizes (max steps being 100) for their sliders.

### Possible discussion points
* Set the bounds of the slider (min and max aspect ratio) according to the window width
* Set the initial aspect ratio according to the initial window width
* Improve styles and/or change the location of the zoom slider

Screenshot from 2021-04-13 09-43-41
Screenshot from 2021-04-13 09-43-27
Screenshot from 2021-04-13 09-42-50


brimoor:
Works great! I tested the filters for int fields as well and everything looks good there.

With respect to the discussion points you mentioned, this is such a useful feature that it would be interesting to see how it looks when the zoom UI is permanently visible rather than hidden inside the settings menu. I'd try putting it to the right of `Viewing X samples`. Perhaps you could put a grid icon to the right of the slider to indicate that the UI controls something about the grid, then people will get the idea I think. Perhaps you could show a `Zoom` tooltip when hovering on the slider. Oooh and clicking the grid icon could reset to the default zoom. That would be handy :)

A couple other thoughts I had are:
- What if we only allowed integer values `[1, 9]` for the zoom parameter? Or to make it more typical, `[1, 10]`? Having fewer discrete values might actually be useful as it would help the user change the parameter in meaningful ways and also help them get back to exactly how they had it previously. I'm thinking that zoom = 1 would be fully zoomed out and zoom = 10 would be fully zoomed in. The numeric value would be inversely proportional to the number of images per row (which I think is good, since it avoids the temptation to think that zoom = 5 means exactly 5 images per row, which is not always true as the grid algorithm is more complicated than that)
- Would it make sense to show the numeric value of the current zoom when the slider is selected, similar to how numbers appear for sliders in the filter bar? Like the previous comment, this would help the user understand what their current setting is. This would be cleaner if the zoom levels were discrete per previous comment
- What about letting the user programmatically configure this setting via the `AppConfig`? I could see different users preferring different defaults here

The current default zoom level LGTM btw.

WRT whether the window width should affect the grid dimensions, I think I would like it either way. The current behavior (width does not affect grid dimensions) has never bothered me, and it has the pro of providing a convenient way to change the size of the images without messing with the grid. It also prevents needing to reload the grid if the window is made wider.

But, I could also get onboard with showing more images as the window width increases. I'd have to try it out to see. If we do go with this option, it would definitely not be desirable if my entire grid refreshed whenever the dimensions changed. I guess I would expect the currently loaded page to "slide up" and if the view got so wide that more images are required, then a new page request would be sent. Not sure on the details, sounds a bit more complicated.

benjaminpkane:
Thanks for the feedback. Will respond today or tomorrow. I've been heads down with thinking through label updates and optimizations.

benjaminpkane:
Comments addressed. There is a `default_grid_zoom` app config setting now. I chose `[0, 10]` for valid values for an odd number (with integer median).

Screenshot from 2021-04-16 14-19-24

brimoor (approved):
LGTM, works great! Here's some parting thoughts for your consideration:

- Can you make the grid icon clickable to reset to the default zoom? With tooltip indicating that this is what the click will do
- Wondering if you considered adding the label above the thermometer bulb of the zoom slider indicating the current zoom level (only when slider is being dragged)? My thought was that this would help put a number to the current zoom level. For example if you were trying to decide on a new default zoom to use in your app config
- Can you try arranging the elements per below? I think it's a bit cleaner to put the fixed size elements on the far end of the justified content (text will change depending on number of samples in view, and perhaps we'll show other info for other types of views there in the future)


```
Viewing X samples  |  [#####------] [:::]
```

PR #964, opened 2021-04-13
Proper handling of lists of expressions in view stages

### LINKED ISSUE #948: [FR] Add JSON type spec for view stage fields
The `expr` parameter of SetField can be a `ViewExpression` or an arbitrary MongoDB aggregation expression, which includes primitive values like ints, strings, lists, etc:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.add_sample_field("float_field", fo.FloatField)
dataset.set_field("float_field", 1).save("float_field")

dataset.add_sample_field("string_field", fo.StringField)
dataset.set_field("string_field", "hello").save("string_field")

dataset.add_sample_field("list_field", fo.ListField)
dataset.set_field("list_field", [1, 2]).save("list_field")

print(dataset.select_fields(["float_field", "string_field", "list_field"]).first())
```

However, our current view stage type specification is not expressive enough to cleanly define such possibilities. Currently, the `SetField.expr` spec is `NoneType|dict`, which wouldn't allow for ints, strings, and other valid field values.

An observation is that `json.loads()` handles all of these cases:

```py
import json

print(json.loads('1'))
print(json.loads('"hi"'))
print(json.loads('[1, 2]'))
print(json.loads('{"proper": "json"}'))
print(json.loads('null')
```

So, we could cleanly support this in the App by introducing a `json` field spec that allows any valid JSON value, and the App would simply use `JSON.parse()` to parse these values when constructing serialized view stages to send to the backend.

There may be other fields beyond `SetField.expr` for which this would be appropriate.

### LINKED ISSUE #962: [BUG] View stage expression parameters that are lists are not properly serialized
The following does not work, because the `SetField._kwargs()` method does not properly serialize the `expr` argument all the way to JSON (ViewExpressions are left unserialized):

https://github.com/voxel51/fiftyone/issues/948 may also need to be implemented in order to fully support the syntax below (where `expr` is a list, not a dict).

```py
import fiftyone as fo
import fiftyone.zoo as foz

upper_bbox = [F("bounding_box")[0], F("bounding_box")[1], F("bounding_box")[2], F("bounding_box")[3]/2.]

view = dataset.set_field("predictions.detections.bounding_box", upper_bbox)
```

brimoor:
* Resolves #962
* Resolves #948

```py
import fiftyone as fo
import fiftyone.zoo as foz

upper_bbox = [F("bounding_box")[0], F("bounding_box")[1], F("bounding_box")[2], F("bounding_box")[3] / 2]

view = dataset.set_field("predictions.detections.bounding_box", upper_bbox)

# previously did not work, now does
```

### DIFF
--- a/app/src/components/ViewBar/ViewBar.tsx
+++ b/app/src/components/ViewBar/ViewBar.tsx
@@ -83,6 +83,7 @@ const ViewBar = React.memo(() => {
   const [state, send] = useMachine(viewBarMachine);
   const [view, setView] = useRecoilState(selectors.view);
   const fieldPaths = useRecoilValue(selectors.fieldPaths);
+  console.log(view);
 
   useEffect(() => {
     send({
--- a/app/src/components/ViewBar/ViewBar.tsx
+++ b/app/src/components/ViewBar/ViewBar.tsx
@@ -83,7 +83,6 @@ const ViewBar = React.memo(() => {
   const [state, send] = useMachine(viewBarMachine);
   const [view, setView] = useRecoilState(selectors.view);
   const fieldPaths = useRecoilValue(selectors.fieldPaths);
-  console.log(view);
 
   useEffect(() => {
     send({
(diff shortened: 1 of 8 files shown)


benjaminpkane:
> @benjaminpkane would it be easy to add #948 to this PR for completeness?

Sure, I can address tomorrow.

benjaminpkane (approved):
Ok, this minimally works now.

Fixing this behavior also required allowing more than just top-level field names to be allowed (arbitrary strings). I would like to improve search results, autocomplete, etc. for `field` parameters as I know they are in a more or less "broken" state right now. For discussion offline, perhaps...

PR #965, opened 2021-04-14
Expanded view counts and selection

### LINKED ISSUE #651: [FR] Only include visible objects when selecting/unselecting all objects from App
Currently, choosing select all/unselect all from the object selection menu (in expanded sample mode in the App) will select all objects, regardless of whether they are currently visible in the media player.

I think a more useful behavior would be that select all/unselect all would only apply to the objects that are currently visible in the media player, which may exclude objects within fields that have been toggled off under display options, or may include objects that have been filtered via label filters (eg setting a confidence threshold). The reasoning here is that object selection is a visual workflow, so the "universe" of possible objects to include/omit should be limited to what is visible.

For example, the sample below has 19 detections and 5 polylines. If I uncheck `gt_detections` and then choose "select all" objects in the object selection menu, it selects all 19 + 5 objects, even though only the 5 polylines are currently visible.

Screen Shot 2020-10-29 at 4 32 39 PM

### LINKED ISSUE #672: [FR] Manually hiding objects should affect label counts and JSON
When objects are manually hidden by clicking on them in expanded sample view, the label counts in the display options menu (see below) and the contents of the filtered JSON (see below) should be updated accordingly.

What we're shooting for is that the stats (and JSON) **always** correspond to exactly what is shown in the media player.

Example:

All but 1 person is hidden, so `gt_person` count should be 1

Screen Shot 2020-11-04 at 10 26 50 PM

Likewise; the filtered JSON should only contain the one person detection that is visible:

Screen Shot 2020-11-04 at 10 26 57 PM

benjaminpkane:
### Issues

* Resolves #672  (although, I believe this was mostly resolved already)
* Resolves #651

### Behavior

With respect to #651, `Clear selection` and `Hide selected` are the only actions that do not account for the current visibility of labels. Actions that would result in a noop are not shown (e.g. no visible labels means that are not selected means `Select visible (current sample)` will not be shown.

Screenshot from 2021-04-14 08-50-29

### DIFF
--- a/app/src/components/Actions/ActionsRow.tsx
+++ b/app/src/components/Actions/ActionsRow.tsx
@@ -69,7 +69,13 @@ const Tag = ({ modal }) => {
   );
 };
 
-const Selected = ({ modal, frameNumberRef }) => {
+const Selected = ({
+  modal,
+  frameNumber,
+}: {
+  modal: boolean;
+  frameNumber?: number;
+}) => {
   const [open, setOpen] = useState(false);
   const selectedSamples = useRecoilValue(atoms.selectedSamples);
   const selectedObjects = useRecoilValue(selectors.selectedLabels);
@@ -101,7 +107,7 @@ const Selected = ({ modal, frameNumberRef }) => {
         <Selector
           modal={modal}
           close={() => setOpen(false)}
-          frameNumberRef={frameNumberRef}
+          frameNumber={frameNumber}
           bounds={bounds}
         />
       )}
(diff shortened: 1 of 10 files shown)


brimoor (approved):
LGTM. Works great!

Only comment (unrelated to changes in this PR) is that it's a bit strange that the `Hidden` entry in the filter widget is between `Labels` and `Confidence`. Perhaps if it were first then it wouldn't get lost in all the visual UI elements. Up to you

Screen Shot 2021-04-16 at 9 52 03 AM

PR #967, opened 2021-04-14
Adding ordered flag to select stage

brimoor:
Adds an optional `ordered=True` flag to `Select()` that will cause the samples in the returned view to appear in the same order as the list of IDs you provide:

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.limit(3)
reversed_view = dataset.select(view.values("id")[::-1], ordered=True)

print(view.values("id"))
print(reversed_view.values("id"))
```


benjaminpkane (approved):
LGTM!

PR #970, opened 2021-04-15
Adding support for controlling whether label selection/exclusion omits samples with no labels

brimoor:
Adds an optional `omit_empty` flag that controls whether samples with no labels are omitted from the view when using `select_labels()` or `exclude_labels()`.

The default behavior remains unchanged: only samples with at least one label are included in the view.

Depending on the workflow, `omit_empty=False` may be the desired behavior, since it ensures that no samples will ever be removed from the view. For example, `dataset.exclude_labels()` won't remove any labels, but it will remove any samples that don't have any labels to begin with, which might not be desirable if you are using `exclude_labels()` in an automated workflow to cleanup a dataset's annotations (but don't want samples to be deleted).

```py
import fiftyone as fo

# Select some labels in the App (or not, an empty list is also allowed)
labels = session.selected_labels

ids = dataset.first().ground_truth.detections[0].id

# Test `select_labels()`

session.view = dataset.select_labels()  # no filtering
session.view = dataset.select_fields("ground_truth").select_labels()  # 1 sample with no ground objects is removed
session.view = dataset.select_fields("ground_truth").select_labels(omit_empty=False)  # no filtering

session.view = dataset.select_labels(labels)
session.view = dataset.select_labels(labels, omit_empty=False)

session.view = dataset.select_labels(ids=ids)
session.view = dataset.select_labels(ids=ids, omit_empty=False)

# Test `exclude_labels()`

session.view = dataset.exclude_labels()  # no filtering
session.view = dataset.select_fields("ground_truth").exclude_labels()  # 1 sample with no ground objects is removed
session.view = dataset.select_fields("ground_truth").exclude_labels(omit_empty=False)  # no filtering

session.view = dataset.exclude_labels(labels)
session.view = dataset.exclude_labels(labels, omit_empty=False)

session.view = dataset.exclude_labels(ids=ids)
session.view = dataset.select_fields("ground_truth").exclude_labels(ids=ids)
session.view = dataset.select_fields("ground_truth").exclude_labels(ids=ids, omit_empty=False)
```

### DIFF
--- a/app/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
+++ b/app/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
@@ -29,12 +29,6 @@ export const toTypeAnnotation = (type: string): string => {
  * for details about numbers and javascript
  */
 export const PARSER = {
-  NoneType: {
-    castFrom: (value) => (value === null ? "None" : value),
-    castTo: () => null,
-    parse: () => "None",
-    validate: (value) => [null, "None", ""].some((v) => value === v),
-  },
   bool: {
     castFrom: (value) => (value ? "True" : "False"),
     castTo: (value) => ["true", "false"].indexOf(value.toLowerCase()) === 0,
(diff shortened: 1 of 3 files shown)


benjaminpkane (approved):
LGTM. A good detail to support.

PR #972, opened 2021-04-16
Stricter `str` stage parameter validation

### LINKED ISSUE #971: [BUG] Expression values for SortBy stage are not correctly rendered in the App
This issue may have arisen along with the new `json` type introduced in https://github.com/voxel51/fiftyone/pull/964.

The type of the `field_or_expr` parameter of `SortBy` is currently declared as `field|str|json`, which causes `sort_by("uniqueness")` to be properly rendered as `uniqueness` in the App, but `sort_by(F("ground_truth.detections").length())` is rendered as `Object [object]`, while it should be rendered as editable JSON. 

As a side note, if I reverse the order of the parameter type specification to `json|field|str`, then `sort_by(F("ground_truth.detections").length())` is correctly rendered as a JSON dict, but  `sort_by("uniqueness")` is treated as JSON and thus rendered as `"uniqueness"`, whereas it would be more user-friendly and consistent with other stage parameters if it were rendered as `uniqueness`.

I believe this is how the parsing engine currently works, but I'd say that composite types like `A|B|C` should be treated as ordered from left-to-right and the first valid type should be used to render the parameter. So the `field_or_expr` parameter should remain implemented as `field|str|json` so that the value will be treated as a `field` whenever possible.

So, ultimately, I believe the issue here is just that `Object [object]` should have been detected as `json` type and rendered appropriately.

```py
import fiftyone as fo
import fiftyone.zoo as foz

# looks good
session.view = dataset.sort_by("uniqueness")

# produces the screenshot below
session.view = dataset.sort_by(F("ground_truth.detections").length())
```

Screen Shot 2021-04-15 at 11 04 52 PM

benjaminpkane:
Resolves #971 

Obviously, there is so much to improve in the view bar, including parameter handling. This band-aid requires the `str` type to not be valid JSON...

* `iamastring` -> valid
*  `"notastring"` -> invalid
* `1` -> invalid
* `"[1]"` -> invalid
* `'{ "one": 1}'` -> invalid

Allows for proper handling of `str|json` types.

### DIFF
--- a/app/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
+++ b/app/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
@@ -165,7 +165,17 @@ export const PARSER = {
     castFrom: (value) => value,
     castTo: (value) => value,
     parse: (value) => value,
-    validate: () => true,
+    validate: (value) => {
+      if (typeof value !== "string") {
+        return false;
+      }
+      try {
+        JSON.parse(value);
+        return false;
+      } catch {
+        return true;
+      }
+    },
   },
 };
 
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -3603,7 +3603,11 @@ class SortBy(ViewStage):
     @classmethod
     def _params(cls):
         return [
-            {"name": "field_or_expr", "type": "field|str|json"},
+            {
+                "name": "field_or_expr",
+                "type": "field|str|json",
+                "placeholder": "field or expression",
+            },
             {
                 "name": "reverse",
                 "type": "bool",


brimoor (approved):
LGTM, confirmed that this resolves #972.

Question: is it relatively easy to re-introduce the down caret in the parameter component that toggles between field/string mode and JSON mode? The reasoning is that it's a bit intimidating to see the JSON editor pop up when creating the `SortBy` stage manually from the view bar. Suppose I just want to `SortBy("uniqueness")`, for example. If the component were in field/string mode by default, then I could type `uniqueness` then press `tab` to continue. Since it defaults to JSON mode, I have to use the mouse to press `Submit`.

More specifically, it would default to field mode because `field` appears first in `field|str|json`.

benjaminpkane:
>Question: is it relatively easy to re-introduce the down caret in the parameter component that toggles between field/string mode and JSON mode? The reasoning is that it's a bit intimidating to see the JSON editor pop up when creating the SortBy stage manually from the view bar. Suppose I just want to SortBy("uniqueness"), for example. If the component were in field/string mode by default, then I could type uniqueness then press tab to continue. Since it defaults to JSON mode, I have to use the mouse to press Submit.

> More specifically, it would default to field mode because field appears first in field|str|json.

Perhaps. The view bar has a lot of band-aids at the moment and needs a refresh. I'll respond on Monday.

PR #973, opened 2021-04-16
"How did you discover FiftyOne?" Form question

benjaminpkane:
Adds a "How did you discover FiftyOne?" question to the feedback/t-shirt form. An answer is required.


brimoor (approved):
LGTM! For your consideration: what about putting this question first, right below email? It seems to be the easiest to answer

PR #974, opened 2021-04-16
Patches and evaluation views

### LINKED ISSUE #947: [BUG] Case sensitive label matching in the App
When filtering labels for a field in the App, the search bar seems to only allow for lower case input. In Open Images, classes are capitalized leading to confusion when you search for the `Dog` label and nothing shows up unless you search for `dog`.

image

### LINKED ISSUE #981: [FR] Use checklist UI for categorical field filters with a small number of choices
It would be quite convenient if the string/boolean widgets in the filters sidebar of the App would switch to a multi-selectable list (see Scoop screenshot below) rather than the current UIs when there are a relatively small number of choices, say... <= 15?

The current autocomplete dropdown is good for string fields with too many choices to explicitly list, but when there are a relatively small number, it is quite helpful and more efficient to be presented with a full multi-selectable list of the possible options.

Another benefit is that the logic of the boolean filters feels backwards. It's much more natural if selecting something results in a positive action (show) not a negative action (hide).

Screen Shot 2021-04-24 at 12 01 01 PM

This feature request is inspired by fondly remembering how natural it was to use Scoop's filtering UI:

Screen Shot 2021-04-24 at 12 00 49 PM

### LINKED ISSUE #988: [FR] Add button to App to sort by visual similarity
Add a button to the App that is visible whenever one or more samples (or labels in expanded modal?) are selected that adds a `SortBySimilarity(session.selected, brain_key=brain_key)` stage (introduced in https://github.com/voxel51/fiftyone/pull/978) to the current view.

The `SampleCollection._get_similarity_keys()` method can be used to check for valid `brain_key`s to pass to the `SortBySimilarity` stage. If the user has multiple compliant keys, then the button should present an option to choose the appropriate key to use.

If the user has not indexed their dataset by similarity, it would be nice if the App UI would link to documentation explaining how to use this powerful feature.

115610576-0ca84000-a2b7-11eb-8e3f-4fe3db24b5df

### LINKED ISSUE #990: [FR] Add filepath to filters sidebar
Per some offline discussions and user feature requests, it would be nice if `filepath` were present in the filters sidebar in the App, with a search UI that enabled users to match arbitrary substrings of the filepath and then select the filepaths that they'd like to see in the grid.

This is related to the categorical dropdown enhancements from https://github.com/voxel51/fiftyone/issues/981, except that `filepath` is a case where there are too many rather too few options for the current design.

A proposed enhancement is to update the backend routes for these functions to check how many options there will be via, say, `distinct().limit(15)`, and then use the appropriate frontend UI depending on how many options there are.

brimoor:
Introduces new `to_patches()` and `to_evaluation_patches()` view stages that function as follows:
- `to_patches()`: creates a view that contains one sample per object patch in a field of the collection
- `to_evaluation_patches()`: creates a view based on the results of an evaluation that contains one sample for each TP/FP/FN example in the collection

Thanks to the work from https://github.com/voxel51/fiftyone/pull/985, this PR also includes support for activating these modes from the App via a new `Patches` icon in the App.

Also from https://github.com/voxel51/fiftyone/pull/985, the App also now includes a `Similarity` icon that enables sorting by visual similarity whenever samples are selected and you have indexed your dataset by similarity via `fiftyone.brain.compute_similarity()`.

These are popular feature requests that turned out to be much easier than expected to implement 🥇🥇🥇 

### Issues

Resolves #947
Resolves #990
Resolves #981
Resolves #988

### Notes

This feature required removing the unique `filepath` constraint on sample collections. I think I'm okay with this. Besides patch view, maybe the user really wants to include multiple versions of the same image in their dataset as separate samples. Of course, they could shoot themselves in the foot by accidentally adding duplicate images, but with great power comes great responsibility :)

Related, `dataset[filepath]` will now return the first sample (of multiple, potentially) with the specified filepath.

### Object patches example

```py
import fiftyone as fo
import fiftyone.zoo as foz

print(patches)

# This can also be achieved from the App... 
# Select patches icon -> Labels -> "ground_truth" in the App
session.view = patches
```

### Evaluation patches example

```py
import fiftyone as fo
import fiftyone.zoo as foz

eval_patches = dataset.to_evaluation_patches("eval")
print(eval_patches)

# This can also be achieved from the App...
# Select patches icon -> Evaluations -> "eval"
session.view = eval_patches
```

### Image similarity example

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

fob.compute_similarity(dataset, brain_key="image_similarity")

img_similarity = dataset.sort_by_similarity(dataset.first().id)
print(img_similarity)

# This can also be achieved from the App...
# Select some images -> similarity icon -> "image_similarity" -> apply
session.view = img_similarity
```

### Object similarity example

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.brain as fob

fob.compute_similarity(dataset, patches_field="ground_truth", brain_key="gt_similarity")

obj_similarity = patches.sort_by_similarity(patches.first().id)
print(obj_similarity)

# This can also be achieved from the App...
# Switch to ground truth patches mode, then select some patches -> similarity icon -> "gt_similarity" -> apply
session.view = obj_similarity
```


fiftyone/core/patches.py
```diff
@@ -0,0 +1,410 @@
...
+    @property
+    def name(self):
+        return self.dataset_name + "-patches"
+
+    def tag_samples(self, tags):
+        super().tag_samples(tags)
+
+        # Update source collection
+
+        def sync_fcn(view, fields=None):
```
benjaminpkane, line 154: So tagging `samples` in an evaluation patches view tags the root dataset labels, and the patches document, but not the labels in the patches documents?

brimoor: Yeah you're correct in pointing out the inconsistency here. Anything that happens to the root dataset needs to be reflected in the "linked" labels in the patches/evaluation views.

I took a bit of a pause on finishing (and testing, btw) the tagging behavior to let the options bounce around in my head a bit to see if a good approach would come to mind...

It might be best if tagging patches at the patch level did *NOT* apply to the labels. In fact, since patches are purely a view concept, this would create data that is not propagated to the root dataset at all, similar to what happens if you were to, for example, add a new field to a patches view.

Another option we discussed last week would be that tagging a patch would apply that tag to the root sample, but the "reflection" property mentioned about would dictate that all other patches from the same sample would need to automatically inherit that tag as well, but I don't like the idea of tagging one patch in a view causing updates to other documents in the view...

Where are your thoughts on all this at the moment?

benjaminpkane: This is confusing me

```py
import fiftyone.zoo as foz

patches = dataset.to_patches("predictions")
patches.tag_samples("test")

dataset.count_values("predictions.detections.tags")
patches.count_values("tags")
patches.count_values("predictions.tags")
dataset.to_patches("predictions").count_values("predictions.tags")
```

Out:
```
{'test': 5620}
{'test': 5620}
{}
{'test': 5620}
```

brimoor: I think the easiest and simplest conceptual path forward is just to change any relevant "SAMPLE" terminology to "PATCH" in the App when viewing patches views, and then remove the propagation of patch tag -> label tag stuff I started.

benjaminpkane: Ok, I'll move forward with that. No propagation is better.

This is a tricky feature, so I have concerns in general.  I will say, it is easier for me to conceptualize patches as just labels the themselves rather than a type of sample.

brimoor: Okay I removed the propagation of sample tags back to the source dataset.

Feel free to voice any concerns you have over the behavior of this feature. Definitely want to make sure that it's as clear as possible to the user!

brimoor: It's definitely undesirable from an implementation standpoint that data is duplicated and new collections are created, but I couldn't figure out how to implement the evaluation and frames transformations in a single pipeline. And, if we were able to figure that out, I'm a bit concerned that the aggregations would be too slow to run repeatedly every time (eg) a new page of samples from the view is requested in the App.

Hopefully, if I've done my job well, this implementation detail is completely hidden from the App, though.

This implementation is *almost* completely hidden from users, but one unfortunate deviation from "regular" views is that updating the root dataset will not automatically affect a patches view; it must be re-instantiated in order to pull any changes.

But, then again, regular datasets and views are not completely "self-syncing", since, as Jason discovered last week, if you're working with a dataset in multiple shells, you would need to call `dataset_reload(hard=True)` in one shell if you happen to add a new sample field in another shell.

Along both of these lines, I was thinking that it may be time to add a `reload()` method to the public interface of datasets/views that will handle refreshing datasets and views to deal with the aforementioned issues.

benjaminpkane:
Is it intentional that `fo.list_datasets()` contains patches datasets?

brimoor:
> Is it intentional that `fo.list_datasets()` contains patches datasets?

No I think they should be omitted but just hadn't implemented it yet in case the design were to change.

Two options that come to mind are to (a) give the patch datasets names that start with `_` and introduce the notion that such names are "private datasets" that aren't visible to users by default, or (b) add a property to the dataset doc to achieve the same effect (patch

One note on the latter point is that I think users should (and can, currently) be able to do `dataset.to_patches(...).clone()` as a syntax for creating a regular dataset out of a patches view.

Talking out loud: allowing cloning means, for example, that we'd have to continue to allow all datasets to have duplicate filepaths, not just the "private" patches datasets that are created for the patches feature.

benjaminpkane:
>Two options that come to mind are to (a) give the patch datasets names that start with _ and introduce the notion that such names are "private datasets" that aren't visible to users by default, or (b) add a property to the dataset doc marking the dataset as a patches dataset to achieve the same effect.

Either works, although (b) might be cleaner

benjaminpkane:
> Talking out loud: allowing cloning means, for example, that we'd have to continue to allow all datasets to have duplicate filepaths, not just the "private" patches datasets that are created for the patches feature.

Yes, fine with me

benjaminpkane:
> Two options that come to mind are to (a) give the patch datasets names that start with _ and introduce the notion that such names are "private datasets" that aren't visible to users by default, or (b) add a property to the dataset doc marking the dataset as a patches dataset to achieve the same effect.

Will handle

brimoor:
Hey @benjaminpkane I've been using this branch a bit more and ran into a few things:

- It looks like boolean `false` values are rendering as empty in the App
- Also, for both the `type` and `crowd` fields in the screenshot below, I've selected things but am not getting the `reset` button. Perhaps this was either a bug or intentional when the select|exclude toggle was removed. Still think its important to have the reset button so the user has an easy way to bail out and undo any filtering they may have performed

Screen Shot 2021-05-05 at 10 53 22 PM

- Are you open to letting the user scroll through all results in the dropdowns? I'd like to be able to scroll to see what else there is, otherwise it's hard to know how to refine my search to make the other options visible. If you're worried about bandwidth, then perhaps `195 more results` could be changed to a `Show all 195 results` button that would load the rest of the results and make it scrollable. Or just load all results from the start and change the text at the bottom to just `195 results`

Screen Shot 2021-05-05 at 10 55 22 PM

- Can you center the `Apply` button in the menu dropdowns, for consistency with how the `Reset` button is now centered in the filter widgets?

Screen Shot 2021-05-05 at 11 19 17 PM

benjaminpkane:
> Are you open to letting the user scroll through all results in the dropdowns? I'd like to be able to scroll to see what else there is, otherwise it's hard to know how to refine my search to make the other options visible. If you're worried about bandwidth, then perhaps 195 more results could be changed to a Show all 195 results button that would load the rest of the results and make it scrollable. Or just load all results from the start and change the text at the bottom to just 195 results

I'm not worried about 200 results, but 100,000. I've changed the search results limit to 200 from 5 and the container is scrollable.

Other comments addressed.

benjaminpkane (approved):
LGTM

PR #982, opened 2021-04-26
Adding support for big values() aggregations

brimoor:
The `values()` aggregation can return a potentially huge amount of data when applied to large datasets. When the size exceeds 16MB, this violates MongoDB's hard 16MB limit on a single document, which results in an error.

This PR introduces the concept of an aggregation with a "big result", which may be spread across multiple documents rather than a single document. Aggregations that are marked as having "big results" are run individually, not combined with other aggregations, since aggregating multiple `Aggregation`s at once uses the `$facet` stage, which collects all results into a single document that would violate the 16MB limit.

The `values()` aggregation is marked as having a big result by default, so this PR effectively allows running `values()` on arbitrarily large datasets.

@benjaminpkane I'm marking this as needs discussion to see if you have any thoughts on how to best proceed with this. This issue mainly arises because `values()` is not really an "aggregation", its a "slicing" tool. The benchmarking I did below suggests that *always* using `_big_result=True` for `values()` is probably fine. I'd rather not actually have the user need to manually tag their aggregations based on the size of their datasets.
 
```py
import eta.core.utils as etau

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("bdd100k", split="validation")

# yields a result that is slightly less than 16MB
view = dataset.limit(2000)

# new default behavior
with etau.Timer():
    values1 = view.values("ground_truth_detections", _raw=True)

# current behavior
with etau.Timer():
    values2 = view.values("ground_truth_detections", _big_result=False, _raw=True)

print(values1 == values2)

try:
    _ = dataset.values("ground_truth_detections", _big_result=False, _raw=True)  # error
except:
    print("Too big!")

_ = dataset.values("ground_truth_detections", _raw=True)  # works
```


benjaminpkane (approved):
So, `values()` is, of course, not really an aggregation. `big_result` looks relatively harmless, so this is fine with me for now, but I think `values()` should represent a different kind of entity and not inherit from `Aggregation`.

Specifically, I think it should return an iterator that yields from the cursor. If a user wants to create an in-memory list, they certainly can.

benjaminpkane:
Pulling values out of aggregations and allowing zipping across multiple fields (using `$facet`) is how I see the API long term...

brimoor:
> Pulling values out of aggregations and allowing zipping across multiple fields (using `$facet`) is how I see the API long term...

Yeah, but on the other hand, although `Values()` doesn't summarize like the other aggregations, it is technically the output of running a MongoDB "aggregation".

As you're alluding to above, the other defining property of our aggregations API is that you can run a set of them in a single pass over the collection, which I'd still like to preserve for `Values()` since we in fact use it ourselves, e.g., in https://github.com/voxel51/fiftyone/blob/a08582a9407cbaa7280989ef0d422527b309dbc3/fiftyone/utils/eval/classification.py#L193-L201.

I don't think `$facet` can be used for this in the context of cursor-based results since it always stuffs the *entire* aggregation results for each facet into a single doc. I guess we'd have to merge the actual pipeline definitions and use different field names for the `$set` for each "facet".

I think the immediate path forward I see is:
- `Values()` is still an aggregation, for now, and the implementation is updated to always run in "big result" mode
- Implementation is updated so that multiple `Values()` in a single call to `aggregate()` are computed in a single pass over the collection
- Any other non-`Values()` aggregations are computed in a second pass over the collection using `$facet`

benjaminpkane:
Ok, yes that makes sense. Yes, I was wrong about `$facet`. It only returns one document.

Yes, running simultaneous aggregations is a defining property. But I consider an aggregation something more along the lines of an [aggregation function in SQL](https://www.geeksforgeeks.org/aggregate-functions-in-sql/). So long-term, I think it should be moved.

Values does not need a grouping. There is at least one value for every document if null values are not omitted.

PR #983, opened 2021-04-26
Adding support for embedded documents in values() and set_values()

brimoor:
Adds full support for passing/retrieving embedded documents using `values()` and `set_values()`. Previously these methods only supported setting/getting primitive fields like `int`, `float`, `str`, etc. that required no conversion between Python and Mongo representations.

With this PR, `values()` is now a proper dual of `set_values()` in the sense that objects like `Detections`, `list(Detection)`, and so forth can be passed to `values()` and then directly retrieved from `set_values()`:

```py
dataset.set_values("embedded.field", values)
also_values = dataset.values("embedded.field")

values == also_values  # true, even if values is, say, a nested list of `Detections` objects
```

With this work, `set_values()` also now more fully delivers on the promise of equivalent loop-based syntaxes that it makes in its docstring; if you can write it as you see below, then you can do it more efficiently via `set_values(...)` instead.

```py
"""Sets the field or embedded field on each sample or frame in the
collection to the given values.

When setting a sample field ``embedded.field.name``, this function is
an efficient implementation of the following loop::

    for sample, value in zip(sample_collection, values):
        sample.embedded.field.name = value
        sample.save()

When modifying a sample field that contains an array, say
``embedded.array.field.name``, this function is an efficient
implementation of the following loop::

    for sample, array_values in zip(sample_collection, values):
        for doc, value in zip(sample.embedded.array):
            doc.field.name = value

        sample.save()

...
"""
```

Another notable change is that `set_values()` is now allowed to add new top-level fields to the dataset's schema, consistent with how one is allowed to set a new field on a `SampleView` object and then `.save()` it to commit the addition.

One must still explicitly declare new sample fields before using `set_field()`, however, since it is not straightforward to infer a proper type for any newly created fields.

New unit tests are added that provide relatively comprehensive testing of this functionality.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -770,7 +770,12 @@ class SampleCollection(object):
         return dict(counts)
 
     def set_values(
-        self, field_name, values, skip_none=False, _allow_missing=False
+        self,
+        field_name,
+        values,
+        skip_none=False,
+        expand_schema=True,
+        _allow_missing=False,
     ):
         """Sets the field or embedded field on each sample or frame in the
         collection to the given values.
@@ -821,7 +826,13 @@ class SampleCollection(object):
                 of ``values`` must be arrays of the same lengths
             skip_none (False): whether to treat None data in ``values`` as
                 missing data that should not be set
+            expand_schema (True): whether to dynamically add new sample/frame
+                fields encountered to the dataset schema. If False, an error is
+                raised if the root ``field_name`` does not exist
         """
+        if expand_schema:
+            self._expand_schema_from_values(field_name, values)
+
         field_name, is_frame_field, list_fields, _ = self._parse_field_name(
             field_name, allow_missing=_allow_missing
         )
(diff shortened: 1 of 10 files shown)


fiftyone/core/collections.py
```diff
@@ -821,21 +875,109 @@ def set_values(
...
+            and issubclass(field_type.document_type, fol._LABEL_LIST_FIELDS)
+            and isinstance(self, fov.DatasetView)
+        ):
+            if is_frame_field:
+                path = self._FRAMES_PREFIX + field_name
+            else:
+                path = field_name
+
+            # pylint: disable=no-member
+            if path in self._get_filtered_fields():
```
benjaminpkane, line 910: I'm not sure this will raise when intended. Or I don't understand when this is suppose to raise.

benjaminpkane (approved):
LGTM. I left one comment.

fiftyone/core/collections.py
brimoor: Oops there was a bug here, thanks for catching!

I updated this to gracefully switch to updating only the list elements rather than raising an error. Probably the user doesn't have any fields on the top-level document other than the nested list anyway. I log a warning just in case.

### Image example

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.filter_labels("predictions", F("confidence") > 0.9)

values = view.values("predictions")

view.set_values("predictions", values)
print(dataset.count("predictions.detections"))  # no data deletion
```

### Video example

```py
import fiftyone as fo
import fiftyone.zoo as foz

print(dataset.count("frames.ground_truth_detections.detections"))

view = dataset.filter_labels("frames.ground_truth_detections", F("label") == "vehicle")
print(view.count("frames.ground_truth_detections.detections"))

values = view.values("frames.ground_truth_detections")

print(dataset.count("frames.ground_truth_detections.detections"))
view.set_values("frames.ground_truth_detections", values)
print(dataset.count("frames.ground_truth_detections.detections"))  # no data deletion
```

PR #985, opened 2021-04-27
App patches

### LINKED ISSUE #947: [BUG] Case sensitive label matching in the App

### LINKED ISSUE #981: [FR] Use checklist UI for categorical field filters with a small number of choices

### LINKED ISSUE #990: [FR] Add filepath to filters sidebar

benjaminpkane:
### Issues
* Resolves #947
* Resolves #990
* Resolves #981 

### Description

Adds support for patches in the App.

* Renames `sample` to `patch` where relevant.
* Adds patches actions for both regular patches views and evaluation patches views

Discussion points:
* Will frame labels be supported? (requires my player51 updates, likely)
* Add helpful information about patches, e.g. creating evaluation keys for patches if none exist

Screenshot from 2021-04-27 14-55-53

### Object patches example

```py
import fiftyone as fo
import fiftyone.zoo as foz

# In App... select patches icon -> Labels -> "ground_truth" in the App

print(session.view)
```

### Evaluation patches example

```py
import fiftyone as fo
import fiftyone.zoo as foz

# In App... select patches icon -> Evaluations -> "eval"

print(session.view)
```

### Image similarity example

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

fob.compute_similarity(dataset, brain_key="image_similarity")

# In App... select some images -> similarity icon -> "image_similarity" -> apply

print(session.view)
```

### Objects similarity example

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.brain as fob

fob.compute_similarity(dataset, patches_field="ground_truth", brain_key="gt_similarity")

# In App... switch to ground truth patches mode, then select some patches -> similarity icon -> "gt_similarity" -> apply

print(session.view)
```


benjaminpkane:
Still finalizing some string filter functionality and working through edges cases, but this is ready for eyes/comments.

Screenshot from 2021-05-03 11-13-28
Screenshot from 2021-05-03 11-10-32

benjaminpkane:
Ready for review. Working through a couple expanded view style fixes, but functionality is done.

@brimoor Would you like me to keep the base as `patches-view`? Looking at #974 now.

benjaminpkane:
Supporting long strings like filepath was proving to be (overly) difficult. I don't think there is a reasonable solution in front of us, but happy to take it offline.

Currently, filepath, etc. remain left-aligned with ellipsis.

Screenshot from 2021-05-04 19-59-21

brimoor:
What if you summarized long list entries like `eta.core.utils.summarize_long_str(string, n, mode="middle")` does?

https://github.com/voxel51/eta/blob/eddd44981a0e8f2bdd213cb5e86f955d46f76456/eta/core/utils.py#L120-L150.

For example, `/home/ben/fiftyone/lots/of/crazy/folders/and/stuff/here/data/0001.jpg` might become `/home/ben/qui...data/001.jpg`.

The benefit of this approach is that it works regardless of whether its is the beginning or the end of the string that uniquely defines it.

Alternatively, if you only want to change behavior for filepaths in particular (since they are unique among most common types of strings that one is likely to load in that it is their *end* not beginning that uniquely defines them) you could try `summarize_long_str(string, n, mode="first")`.

In either case, the tooltip would show the full path. You'd have to adopt the same convention for the selected entries that appear below the filter in order for the user to make sense of what they've done.

benjaminpkane:
Yes I can add. I'll just note that our font is not monospaced.

brimoor:
Everything works amazingly well!

Here's a few reactions after using the UI for awhile:

**Filter widgets**

- How about centering the `Reset` button text for the filters? I see that things are left-aligned nicely, which I appreciate, but I kind of like `Reset` being centered because it visually separates from the other content, which makes sense because they have different semantic meanings ("reset" is not a label, its an action). If you feel strongly about how it is now, I defer to you
- I like how the select/hide toggle for boolean fields. What about also omitting it for string fields with < n unique values? In the screenshot below, it feels a bit weird that the toggle is visible in one case but not the other. I'm not sure what value of `n` you chose for the transition, but if n=10, say, then perhaps you're worried that it might be laborious to select 9/10 checkboxes to achieve the same thing as selecting one and then selecting exclude? If that's the case, I might recommend adding the select/exclude toggle to the boolean fields as well, for consistency

Screen Shot 2021-05-05 at 10 59 05 AM

**Patches icon**

- What if the patches icon were orange whenever > 0 label fields are visible (or even if the schema has > 0 label fields), similar to how the other icons light up when they are clickable? Yes, this would mean that the patches icon is almost always orange when working with detections, but it's an awesome feature right!?

**Tooltips**

- Proposed edit to the evaluation patches tooltip: `Switch to evaluation patches view for the "ground_truth" field`
- Proposed edit to the patches tooltip: `Switch to patches view for the "ground_truth" field`
- Proposed edit to tooltip of the apply button on sort by similarity: `Sort by similarity to the selected [images/objects/patches]` (bracketed term depends on brain key and view type

This is probably a taste thing, but just trying to make them a bit easier to parse

**Patches/similarity menus**

- What about always including the link to the documentation at the bottom of the lists, even if there are >= 1 entries? Then the user always has a link into the documentation to recall what the feature does and how to use it? Under the proposed change, there would not text that has `No label fields [arrow]`, there would just be a persistent entry at the top or bottom that says `About patch views`, `About evaluation views`, and `About similarity search`. In the case of sort by similarity with >= 1 entries, a whole modal populates, so perhaps instead of keeping an entry like this, you could add a `?` icon to the right of the `Sort by similarity` heading

Screen Shot 2021-05-05 at 11 08 00 AM

**Expanded modal**

- Render null values as `-` rather than `undefined` (latter is JS specific)

Screen Shot 2021-05-05 at 11 02 05 AM

**String tags**

- What about using the new middle-summarization technique when abbreviating long tags in the grid as well?
 
Screen Shot 2021-05-05 at 11 05 10 AM

brimoor:
Okay one small thing: what about auto-selecting the first radio button in the similarity modal by default? Then the user can directly click apply without clicking anything else.

Technically they can already click apply since the backend ViewStage will attempt to chose a reasonable default, but since the App already has the logic to render the possible values, might be a bit cleaner to show an option as selected so that the user knows what will be happening if they click apply.

brimoor:
> Okay one small thing: what about auto-selecting the first radio button in the similarity modal by default? Then the user can directly click apply without clicking anything else.
> 
> Technically they can already click apply since the backend ViewStage will attempt to chose a reasonable default, but since the App already has the logic to render the possible values, might be a bit cleaner to show an option as selected so that the user knows what will be happening if they click apply.

Actually I think there might be a small bug in the logic related to this. I ran `compute_similarity()` for both images and ground truth objects in my dataset, then switched to ground truth patches view, then ran sort by similarity without choosing a radio button, and I got an error explaining that my image similarity run couldn't be used for my current view. I thought this was because I messed up the default brain key logic on the backend, but I think its actually correct; it was the App that chose the wrong type of similarity result. The weird thing is that the radio button showed (only) my object similarity key, so it seems that some but not all parts of the App know what to pick by default.

benjaminpkane:
With respect sort by similarity issues, I'm seeing an error from this...
```py
import fiftyone.brain as fob
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart").limit(10).clone()
fob.compute_similarity(dataset, patches_field="ground_truth", brain_key="gt")

# AttributeError: 'PatchesView' object has no attribute '_get_ids_map'
patches.sort_by_similarity(patches.first().id, brain_key="gt")
```

brimoor:
> With respect sort by similarity issues, I'm seeing an error from this...
> 
> ```python
> import fiftyone.brain as fob
> import fiftyone.zoo as foz
> 
> dataset = foz.load_zoo_dataset("quickstart").limit(10).clone()
> fob.compute_similarity(dataset, patches_field="ground_truth", brain_key="gt")
> 
> patches = dataset.to_patches("ground_truth")
> 
> # AttributeError: 'PatchesView' object has no attribute '_get_ids_map'
> patches.sort_by_similarity(patches.first().id, brain_key="gt")
> ```

Ah my bad, I deleted it in https://github.com/voxel51/fiftyone/pull/974/commits/703c4563758691dfcefcaa4c9d96a4a2d0c3e6cc because it wasn't used anywhere, but I forgot that the brain might be using it. It's added back now https://github.com/voxel51/fiftyone/pull/974/commits/07945f8b7f3d2344bc8c06ee41c1585826e1f064.

FYI- I am currently writing unit tests for this functionality so this sort of stuff won't happen!

benjaminpkane:
Thanks! I've addressed everything I believe. Merging now, and can make any tweaks on `patches-view`.

I did not middle ellipsis `filepath`, etc. in the sample grid because hard coding a length would mean arbitrarily cutting of the value no matter the width of the image.

Issue #986, opened 2021-04-28
[FR] Add support for reloading dataset/sample backing documents when dataset has changed 


brimoor:
FiftyOne allows for a dataset to be manipulated simultaneously in multiple processes. However, if one process adds a new sample field, for example, then reading a sample in the other process will fail, since the sample document in the latter process doesn't contain the new field.

A minimal fix for this would be to introduce a `Dataset.reload()` method that can be used to manually refresh the backing documents for the dataset and any in-memory samples.

A more robust fix would be to gracefully call `Dataset.reload()` if loading a sample raises an error, which would automatically recover from such an error.

Example workflow between two shells that demonstrates the issue:

#### in shell 1

```py
import fiftyone as fo

dataset = fo.Dataset("simul")
sample = fo.Sample(filepath="image.png")
dataset.add_sample(sample)

print(sample)
```

#### shell 2

```py
import fiftyone as fo

dataset = fo.load_dataset("simul")
sample = dataset.first()

print(sample)
```

#### back in shell 1

```py
sample["new_field"] = 1
sample.save()

print(sample)
print(dataset.first())  # same
```

#### now back in shell 2

```py
print(sample)  # not updated
print(dataset.first())  # error
```

PR #992, opened 2021-04-29
Adding support for reloading dataset schemas + frames implementation overhaul

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/986.

Adds a `SampleCollection.reload()` method that allows for reloading the backing information (document, sample/frame classes, and any in-memory objects) for a dataset, which allows for editing a dataset's schema in one shell while working with it in another shell.

If an error is encountered while trying to load a sample from the DB, a `reload()` call is inserted, which allows for graceful recovery when the dataset's schema changes.

In implementing this, I also dove into the `Frames` class and made some significant improvements, including:
- Added `FrameView` and `FramesView` classes that are used to serve frames for dataset views
- Correct implementation of `Frames.__getitem__` when working with frame views. This method previously did not respect views, and it would not allow new fields to be added 
- Correct implementation of saving frames in `DatasetView`s that have filtered/excluded/selected fields. Previously `ReplaceOne` was always used, which would squash filtered data
- Added pretty thorough unit tests of expected behavior when working with video frames

Other enhancements:
- Adds an `Aggregation._needs_frames()` method that allows aggregations to declare whether they require frame labels. This enabled changing the default value of the `attach_frames` parameter to `False` for methods like `SampleCollection._pipeline()`, and instead **only** attaching frames if they are actually needed
- Adds `expand_schema` flags to a variety of `Sample`, `Frames`, and `Frame` methods to enable explicit control of whether it is allowed to expand a dataset's schema, similar to the same flag on `Dataset.add_sample()`
- Frames are now attached to samples in sorted order, which previously would have caused undefined behavior if users attempted something like `dataset.values("frames.field")` on a dataset whose frames were inserted in non-sorted order
- General cleanup of the `Document` interface, including abstracting the singleton behavior to metaclasses in the `fiftyone.core.singletons` module, and significant cleanup of the mess of code that was the `__getattr__`, `__getitem__`, etc. interfaces of `Document`, `Sample`, and `Frame`

### Image concurrency example

```py
# shell 1

import fiftyone as fo

dataset = fo.Dataset("simul")

sample = fo.Sample(filepath="image.png")
dataset.add_sample(sample)

print(sample)

# shell 2

import fiftyone as fo

dataset = fo.load_dataset("simul")

sample = dataset.first()

print(sample)

# shell 1

sample["new_field"] = 1
sample.save()

print(sample)
print(dataset.first())

# shell 2

print(sample)  # not yet updated
print(dataset.first())  # automatic reload!
```

### Video concurrency example

```py
# shell 1

import fiftyone as fo

dataset = fo.Dataset("simul-video")

sample.frames[1] = fo.Frame(hello="world")

dataset.add_sample(sample)

frame = sample.frames[1]

print(frame)

# shell 2

import fiftyone as fo

dataset = fo.load_dataset("simul-video")

sample = dataset.first()
frame = sample.frames[1]

print(frame)

# shell 1

frame["new_field"] = 1
sample.save()

print(frame)
print(dataset.first().frames[1])

# shell 2

print(frame)  # not yet updated
print(dataset.first().frames[1])  # automatic reload!
```


benjaminpkane (approved):
LGTM

Issue #1002, opened 2021-05-06
[BUG] MongoDB timeout when iterating through large datasets


ehofesmann:
Iterating through a detection dataset with ~40K images resulted in a pymongo error after roughly 1 hour.

``` shell
---------------------------------------------------------------------------
CursorNotFound                            Traceback (most recent call last)
<ipython-input-13-40dc943154af> in <module>
      5 header = "ImageID,Source,LabelName,Confidence,XMin,XMax,YMin,YMax,IsOccluded,IsTruncated,IsGroupOf,IsDepiction,IsInside"
      6 lines = [header]
----> 7 for sample in eval_view:
      8     img_id = sample.open_images_id
      9     for det in sample[gt_det_field].detections:
​
~/work/fiftyone/fiftyone/fiftyone/core/view.py in iter_samples(self)
    254         filtered_fields = self._get_filtered_fields()
    255 
--> 256         for d in self._aggregate(detach_frames=True):
    257             try:
    258                 doc = self._dataset._sample_dict_to_doc(d)
​
~/venvs/oi-test/lib/python3.6/site-packages/pymongo/command_cursor.py in next(self)
    257         # Block until a document is returnable.
    258         while self.alive:
--> 259             doc = self._try_next(True)
    260             if doc is not None:
    261                 return doc
​
~/venvs/oi-test/lib/python3.6/site-packages/pymongo/command_cursor.py in _try_next(self, get_more_allowed)
    268         """Advance the cursor blocking for at most one getMore command."""
    269         if not len(self.__data) and not self.__killed and get_more_allowed:
--> 270             self._refresh()
    271         if len(self.__data):
    272             coll = self.__collection
​
~/venvs/oi-test/lib/python3.6/site-packages/pymongo/command_cursor.py in _refresh(self)
    204                                     self.__collection.database.client,
    205                                     self.__max_await_time_ms,
--> 206                                     False))
    207         else:  # Cursor id is zero nothing else to return
    208             self.__killed = True
​
~/venvs/oi-test/lib/python3.6/site-packages/pymongo/command_cursor.py in __send_message(self, operation)
    138         try:
    139             response = client._run_operation_with_response(
--> 140                 operation, self._unpack_response, address=self.__address)
    141         except OperationFailure:
    142             kill()
​
~/venvs/oi-test/lib/python3.6/site-packages/pymongo/mongo_client.py in _run_operation_with_response(self, operation, unpack_res, exhaust, address)
   1370             address=address,
[...]

PR #1003, opened 2021-05-06
[WIP] Resolve cursor timeouts

benjaminpkane:
Work towards resolving #1002 ...

@ehofesmann I'd consider this more of an experiment than anything at this point.

### Background
MongoDB offers a "no timeout" option for typical `find()` cursors, but we use an "aggregation" cursor for sample iteration (and, in general, to support views) which do not have this option.

Our cursors do offer a batch size option. It may be that to many samples are being sent per batch which causes to long of wait between cursor requests.

The details of it are that the cursor may be sending too many samples at a time per cursor request, which leaves the cursor idle for too long.

### TLDR

On this branch, using `dataset_or_view.iter_samples(batch_size=N)` may alleviate the issue in #1002 . Maybe where `N=10` or `N=100`. It's hard to say, but `N=1` might be terribly slow.

### Example

```py
import fiftyone.zoo as foz

for sample in dataset.limit(100).iter_samples(batch_size=10):
    continue
```


benjaminpkane:
Thinking out loud, we could also just track the current index during iteration and create a new one with a `skip(idx)` stage if the cursor times out...

brimoor:
@benjaminpkane Thanks for taking a stab at some possible fixes for this!

[Looks like](https://docs.mongodb.com/manual/tutorial/iterate-a-cursor/#cursor-batches) MongoDB's default batching strategy is to use a batch size of 101 for the first batch, followed by an undetermined batch subsequently, which is probably as big as possible such that batch is <= 16MB.

Given the 16MB point, don't you think its more likely that the issue is the overall time and not the batching, especially since #1002 reports a failure after ~1 hour?

I think the `skip()` recovery might be a good option here.

If smaller batch sizes were a viable solution, I'd hate to decrease performance in the typical case to avoid problems only in extreme cases (and would rather users not have to manually configure their way out of errors in extreme cases either)

benjaminpkane:
```py
Given the 16MB point, don't you think its more likely that the issue is the overall time and not the batching, especially since #1002 reports a failure after ~1 hour?

I think the skip() recovery might be a good option here.

If smaller batch sizes were a viable solution, I'd hate to decrease performance in the typical case to avoid problems only in extreme cases (and would rather users not have to manually configure their way out of errors in extreme cases either)
```

My thinking was 16MB batches were too big, but I don't really believe that either. Anyways, closing, and will make a PR for `skip` recoveries.

PR #1004, opened 2021-05-07
Adding support for creating frames views into video datasets

brimoor:
### Frames views

Adds a `to_frames()` view stage that enables on-the-fly conversion of a video datasets into frames views, analogous to `to_patches()` and `to_evaluation_patches()`.

By default, the first time a sample is involved in a `to_frames()` stage, it is sampled into a directory of per-frame images. These image paths are then used as the `filepath` for the frame samples.

However, `sample_frames=False` can also be passed, in which case the frame samples will have their `filepath` set to the source video, the idea being that `(filepath, frame_number)` could be used by the App to render the appropriate frame. I have my concerns that this latter approach would not be fast enough to provide a good user experience in the App, as that would be a lot of video seeking + possible intra-key-frame decoding to do. Keeping an open mind about it though...

In addition, `to_frames()` can be further configured by providing any of the following additional parameters:
- `fps`: convert to frames at the specified frame rate rather than sampling every frame
- `max_fps`: a maximum frame rate at which to sample
- `sparse`: whether to only generate samples in the frames view for video frames that have frame documents. This is particularly powerful in conjunction with the new `select_frames()` and `match_frames()` stages, which enable filtering of frame documents

Here are the rules of frame views:
- Frame samples inherit all frame-level labels from the source video dataset, including their `frame_number`. They are also given a `sample_id` field that records the ID of the parent video sample
- The `id` of Frame samples is inherited from the `id` of the corresponding frame of the parent video sample
- Frame samples inherit any `tags` from their parent video samples (like patches views). Other sample-level fields are *not* inherited
- Any changes, deletions, or additions of fields on the frame samples are automatically synced to the frames of the source dataset
- Changes to default fields like `tags` and `metadata` are not synced to the source dataset (like patches views)

### Other additions

- Adds `last()`, `head()`, and `tail()` implementations to the `Frames` class
- Adds `exclude_fields()`, `select_frames()` and `match_frames()` stages that enable selecting specific frames of video collections via IDs or filter expressions, respectively

### Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

#
# Create a frames view for an entire video dataset
#

frames = dataset.to_frames()
print(frames)

session.view = frames

#
# Create a frames view that only contains frames with at least 10
# objects, sampled at a maximum frame rate of 1fps
#

num_objects = F("ground_truth_detections.detections").length()
view = dataset.match_frames(num_objects > 10)

frames = view.to_frames(max_fps=1, sparse=True)
print(frames)

session.view = frames
```


benjaminpkane:
I pretty much agree with all implementation decisions. I aim have a go/no-go with respect to `(filepath, frame_number)` usage in the App tomorrow.

benjaminpkane:
`(filepath, frame_number)` looks usable to me. It is slower, but I think it is a far better option for users who want look through a few frames without processing an entire video dataset. I think there is a lot of opportunity to optimize down the road, as well. I will share test code today or tomorrow in #1041.

The App easily supports both scenarios (image, or frame number and video) even right now. So supporting both is fine with me.

Depending on if we want to support patches views from frame views, `frame_number` would need to be added to `PatchesView` samples that come from a `FramesView`? Not sure if syncing could handle that situation at the moment.

It might be nice to see informative errors raise when patches views are applied to video datasets or frame views, if they are not supported.

brimoor:
> `(filepath, frame_number)` looks usable to me. It is slower, but I think it is a far better option for users who want look through a few frames without processing an entire video dataset. I think there is a lot of opportunity to optimize down the road, as well. I will share test code today or tomorrow in #1041.
> 
> The App easily supports both scenarios (image, or frame number and video) even right now. So supporting both is fine with me.
> 
> Depending on if we want to support patches views from frame views, `frame_number` would need to be added to `PatchesView` samples that come from a `FramesView`? Not sure if syncing could handle that situation at the moment.
> 
> It might be nice to see informative errors raise when patches views are applied to video datasets or frame views, if they are not supported.

Okay interesting, looking forward to trying it out!

I do want to support patches views into frame views, so I'll make sure that syncing and frame numbers are working (was on my list when I get back to this PR)

fiftyone/core/video.py
```diff
@@ -0,0 +1,582 @@
...
+            frame. A dimension can be -1 if no constraint should be applied.
+            The frames are resized (aspect-preserving) if necessary to meet
+            this constraint
+        max_size (None): an optional maximum ``(width, height)`` for each
+            frame. A dimension can be -1 if no constraint should be applied.
+            The frames are resized (aspect-preserving) if necessary to meet
+            this constraint
+        force_sample (False): whether to resample videos whose sampled frames
+            already exist
+        name (None): a name for the returned dataset
```
benjaminpkane, line 384: I think it would be nice to have a sparse option, where frames are only created for existing frame documents on the root dataset. And, perhaps, new frames can be propagated back to the root.

Noting it now, but perhaps this is a feature request for discussion.

benjaminpkane (approved):
LGTM!

benjaminpkane:
Things look good to me :rocket:

PR #1005, opened 2021-05-07
Graphs enhancements

### LINKED ISSUE #969: [FR] Add label tags graph to App graphs tab
For completeness, we should rename the `tags` graph to `sample tags` in the App and then add a `label tags` graph that plots the distribution of label tags, IE the same data that already exists in the `LABEL TAGS` section of the filters sidebar, just in graphical form.

Screen Shot 2021-04-15 at 10 37 19 AM

### LINKED ISSUE #980: [FR] Clicking on already opened graph tab should collapse it
It would be natural if clicking on an already-opened graphs tab would collapse the graphs section. This would present a more convenient alternative to mousing way over to the righthand side to click hide.

Screen Shot 2021-04-24 at 12 01 59 PM

benjaminpkane:
### Issues

* Resolves #969 
* Resolves #980
Screenshot from 2021-05-07 09-11-02

### DIFF
--- a/app/src/components/Distributions.tsx
+++ b/app/src/components/Distributions.tsx
@@ -170,7 +170,13 @@ const Distributions = ({ group }: { group: string }) => {
   useEffect(() => {
     setData([]);
     setLoading(true);
-  }, [JSON.stringify(view), JSON.stringify(filters), datasetName, refresh]);
+  }, [
+    JSON.stringify(view),
+    JSON.stringify(filters),
+    datasetName,
+    refresh,
+    group,
+  ]);
 
   if (loading) {
     return <Loading />;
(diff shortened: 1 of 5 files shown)


brimoor (approved):
Stated issues are resolved beautifully! One thing to fix: `Sample tags` doesn't change to `Patch tags` for patches view.

While you're in here, can you address another bug and one suggestion I noticed:

**Bug**: the tick labels for boolean fields don't appear (similar to tag bubble issue that was recently fixed for boolean fields)

Screen Shot 2021-05-07 at 2 38 50 PM

**Suggesion**: can you hide sections for empty graphs?

For example, when viewing `Label tags` graphs for the quickstart dataset (which has no label tags), I see the section headings with lots of empty space below each. What if you instead hid sections whose graphs would be empty, and, if all graphs in a tab are empty, just show "No <graph type>" like you do in the filters sidebar sections?

Screen Shot 2021-05-07 at 2 38 16 PM

PR #1007, opened 2021-05-07
Adding flag to extract eval patches from evaluation view

brimoor:
Previously, `to_evaluation_patches()` would ignore the collection to which the stage was being added and always extract patches from the view on which the evaluation was performed.

This PR changes the default behavior to instead extract patches from the collection to which `to_evaluation_patches()` is being applied. This may result in both omitting labels that were evaluated as well as including labels which were not evaluated, but that's okay. In the case of labels which were not evaluated, for example, each label will get its own patch with `type == None`.

This also resolves a bug in the current implementation where, effectively, `use_eval_view=True` was always being used when `to_evaluation_patches()` is called, but the view stages of the returned view were not correctly updated to reflect the fact that the evaluation view was being used.

### TODO

@benjaminpkane the `use_eval_true=True` parameter is a bit different from other view stages in that it can result in a view where the previously constructed view stages are deleted and replaced with different stages. A simple way to reproduce this effect is to run the snippet below. Can you look into whether the indicated error can be avoided?

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Evaluate part of the dataset

# Creates patches for contents of view, some of which were evaluated and some of which were not
session.view = dataset.skip(50).to_evaluation_patches("eval")

# Now in the App view bar, change `use_eval_view` parameter from `False` to `True`
# This raises an error in the App and in the terminal, although the view does seem to load afterwards

# Achieves the same result, but no errors as the view is created first
session.view = dataset.skip(50).to_evaluation_patches("eval", use_eval_view=True)
```

### Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Evaluate part of dataset

# Creates eval patches for entire dataset. Labels that weren't included in the
# evaluation get their own patches with ``type == None``
session.view = dataset.to_evaluation_patches("eval")

# Creates patches for contents of view, some of which were evaluated and some of which were not
session.view = dataset.skip(50).to_evaluation_patches("eval")

# Creates patches for evaluation view, regardless of prior view stages
session.view = dataset.skip(50).to_evaluation_patches("eval", use_eval_view=True)

# Test complete mismatch; all labels have ``type == None``
session.view = dataset.skip(100).to_evaluation_patches("eval")
```

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -3761,7 +3761,7 @@ class SampleCollection(object):
         return self._add_view_stage(fos.ToPatches(field))
 
     @view_stage
-    def to_evaluation_patches(self, eval_key):
+    def to_evaluation_patches(self, eval_key, use_eval_view=False):
         """Creates a view based on the results of the evaluation with the
         given key that contains one sample for each true positive, false
         positive, and false negative example in the collection, respectively.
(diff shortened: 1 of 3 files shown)


brimoor:
> Ok, this is tricky. Will try to resolve today

Another way to take this would to be to declare that views aren't allowed to change previous stages (that does break the existing notion of chaining view stages, after all), and I'll just remove the `use_eval_view=True` flag for now.

The essential element of this PR is really not `use_eval_view=True` but rather the fact that the new view should really only contain patches for the current view.

After all, the user can achieve the same effect as the `use_eval_view=True` flag as follows:

```py
view1 = view.to_evaluation_patches(eval_key, use_eval_view=True)

# equivalent
view2 = view.load_evaluation_view(eval_key).to_evaluation_patches()
```

PR #1011, opened 2021-05-07
Always show filter reset when applicable

### LINKED ISSUE #1008: [BUG] Filter reset button not appearing when using App filter widgets
On `develop`, the `Reset` filter button is not appearing in two cases where I would expect it to (since a non-trivial filtering operation has been performed):

- Unchecking `None` on a numeric scalar filter
- Checking any box(es) on a boolean filter

Screen Shot 2021-05-07 at 2 49 09 PM

benjaminpkane:
Resolves #1008 

Always show reset on filters when applicable

Screenshot from 2021-05-07 16-35-31


brimoor (approved):
LGTM

PR #1015, opened 2021-05-10
Mark filters to view before switching to patches

### LINKED ISSUE #1012: [FR] Switching to patches view should respect the current extended view
In the example below, switching to patches view (objects or evaluations) should first "bookmark" the extended view and then create the patches stage. This achieves the desired effect that the user likely has: "I'd like to see patches for the contents of the current sample grid"

Screen Shot 2021-05-07 at 8 35 39 PM

benjaminpkane:
Resolves #1012 

@brimoor should marking filters also happen when sorting by similarity?

### DIFF
--- a/app/src/components/Actions/ActionsRow.tsx
+++ b/app/src/components/Actions/ActionsRow.tsx
@@ -268,7 +268,7 @@ const SaveFilters = () => {
         setLoading(false);
         socket.send(packageMessage("save_filters", {}));
       }}
-      title={"Save current field filters as view stages"}
+      title={"Mark current field filters as view stages"}
     />
   ) : null;
 };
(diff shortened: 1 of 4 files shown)


app/src/components/Actions/ActionsRow.tsx
```diff
@@ -268,7 +268,7 @@ const SaveFilters = () => {
         setLoading(false);
         socket.send(packageMessage("save_filters", {}));
       }}
-      title={"Save current field filters as view stages"}
+      title={"Mark current field filters as view stages"}
```
brimoor, line 271: I'm onboard with the bookmark icon for this action, but I find "mark" in the tooltip a bit confusing. I'd recommend leaving `save` or perhaps `Convert current field filters to view stages` if you want something different

brimoor (approved):
LGTM. And yes good point, I think sorting by similarity should also save the current filters first.

What do you think about also including a `select()`/`select_labels()` stage if any samples/labels are selected before converting to patches view? This is consistent with how the tagging functionality behaves: if things are selected, the action only applies to them.

This latter point would only apply to patches, not to similarity, as selection has another meaning for similarity.

PR #1016, opened 2021-05-10
Clear group filters

### LINKED ISSUE #956: [FR] Add option to clear all filters on a field
The `SAMPLE TAGS` and `LABEL TAGS` sections in the filter sidebar have an icon that appears whenever one or more filters are applied, which can be clicked to clear all filters in that section.

It would be nice if the `LABEL FIELDS` and `SCALAR FIELDS` sections also supported this option, to allow for both a visual indicator that filters are applied and the ability to quickly clear them.

Screen Shot 2021-04-02 at 5 43 02 PM

benjaminpkane:
Resolves #956

Screenshot from 2021-05-10 09-02-05

### DIFF
--- a/app/src/components/Actions/Checker.tsx
+++ b/app/src/components/Actions/Checker.tsx
@@ -4,8 +4,9 @@ import { Checkbox } from "@material-ui/core";
 import numeral from "numeral";
 import styled from "styled-components";
 
-import { ItemAction, useHighlightHover } from "./utils";
+import { useHighlightHover } from "./utils";
 import { useKeydownHandler, useTheme } from "../../utils/hooks";
+import { ItemAction } from "./ItemAction";
 
 const CheckboxDiv = animated(styled(ItemAction)`
   display: flex;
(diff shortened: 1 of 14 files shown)


brimoor (changes requested):
ooh I love the filter icon with the ability to click to clear everything!

Was it your intention to remove the `Reset` button that appears when a slider-style filter is active? I still see the reset button when using checkbox-style filters, and I think the slider-style filters should still have their `Reset` buttons too.

brimoor (approved):
I love this feature! I'll go ahead and approve, but I noticed one thing to look into that I'll describe below. Not sure if this was introduced in this PR or has been around for awhile 🤷‍♂️ 

When using a numeric slider on a field that also includes `None` values, I observe the following:
- After making one adjustment to the slider endpoints, the `None` row disappears, but the filter includes both `None` values and values within the range.
- After making a second adjustment to the slider, the behavior changes and now `None` values are *not* allowed.

I believe the intended behavior (since the `None` row disappears after the first adjustment), is that if the slider is in a non-trivial position, then `None` values are always excluded.

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Evaluate part of dataset

# Try using the slider to filter the `eval_tp` field
```

PR #1018, opened 2021-05-10
Extract evaluation patches from current view

brimoor:
Previously, `to_evaluation_patches()` would ignore the collection to which the stage was being added and always extract patches from the view on which the evaluation was performed.

This PR changes the default behavior to instead extract patches from the collection to which `to_evaluation_patches()` is being applied. This may result in both omitting labels that were evaluated as well as including labels which were not evaluated, but that's okay. In the case of labels which were not evaluated, for example, each label will get its own patch with `type == None`.

This PR is a simplified implementation based on the discussion in #1007.

### Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Evaluate part of dataset

# Creates eval patches for entire dataset. Labels that weren't included in the
# evaluation get their own patches with ``type == None``
session.view = dataset.to_evaluation_patches("eval")

# Creates patches for contents of view, some of which were evaluated and some of which were not
session.view = dataset.skip(50).to_evaluation_patches("eval")

# Test complete mismatch; all labels have ``type == None``
session.view = dataset.skip(100).to_evaluation_patches("eval")

# Creates patches *exactly* the evaluation view
session.view = dataset.load_evaluation_view("eval").to_evaluation_patches("eval")
```


benjaminpkane (approved):
I like it. LGTM!

PR #1023, opened 2021-05-11
Restrict string field graphs to fields with small distinct counts

benjaminpkane:
I think there is a bright future for plots in FiftyOne.

In the meantime, I think silently restricting string field graphs to those with distinct counts less than 100 (currently). `tags` (redundant), `filepath` and `sample_id` are always omitted. Note that the graphs support lists with scalar subfields, unlike the sidebar.

A somewhat lazy solution to preventing excessive information in the App (and possible freezing) but I think there are bigger fish to fry. :fish:


brimoor (approved):
LGTM. Fish frying comment noted 😉

PR #1025, opened 2021-05-11
Adding documentation for patches views and sorting by visual similarity

brimoor:
Adds documentation on patches views and sorting by visual similarity to the Brain, App, Dataset Views, and Evaluation user guide pages.

Also includes a small new feature: brain results and evaluation results that are loaded via `SampleCollection.load_evaluation_results()` and `SampleCollection.load_brain_results()` are cached on the `Dataset` object in-memory so that subsequent retrievals of the results in the same session will be instant. This can be import when, for example, sorting by visual similarity on a large dataset, since creating the view stage requires access to the results, and it can take a few seconds to load a large binary blob from GridFS (much longer than the actually nearest neighbors search itself!)

Note- the similarity index on a dataset with 10K samples is only ~10MB, so I'm not worried about too much being cached in-memory.

Here are the links to use in the App:
- Object patches mode: `user_guide/app.html#viewing-object-patches`
- Evaluation patches mode: `user_guide/app.html#viewing-evaluation-patches`
- Sort by similarity: `user_guide/app.html#sorting-by-visual-similarity`


benjaminpkane (approved):
LGTM :+1:

PR #1027, opened 2021-05-11
Open Images docs banner

benjaminpkane:
Links to the open images tutorial

Screenshot from 2021-05-11 16-20-18


brimoor (approved):
Are you open to linking to the main website section like the banner on voxel51.com? I think that's preferable in the docs too.

PR #1034, opened 2021-05-16
Fixing multiprocess initialization bug when loading Open Images dataset (#1)

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1035.

Fixes a bug in the way that the `multiprocessing` module was being used that was causing Mac users running Python >= 3.8 to encounter an error when downloading the Open Images dataset.

Verified that this fixes the bug under Python 3.8:

```
conda create -n fifty3 python=3.8
conda activate fifty3

git clone https://github.com/voxel51/fiftyone
cd fiftyone
git checkout --track origin/oi-hotfix
bash install.bash
```

```py
import fiftyone.zoo as foz

# now works
    "open-images-v6", 
    split="validation", 
    label_types=["detections"], 
    max_samples=10,
)
```

### DIFF
--- a/fiftyone/utils/openimages.py
+++ b/fiftyone/utils/openimages.py
@@ -1575,8 +1575,6 @@ def _initialize_worker():
 
 def _do_s3_download(args):
     filepath, filepath_download = args
-
-    global s3_client
     s3_client.download_file(_BUCKET_NAME, filepath_download, filepath)
 
 
--- a/fiftyone/utils/openimages.py
+++ b/fiftyone/utils/openimages.py
@@ -1559,19 +1559,14 @@ def _download_specific_images(
             len(inputs),
         )
 
-    global s3_client
-    s3_client = None
-
     with fou.ProgressBar(total=len(inputs)) as pb:
-        with multiprocessing.Pool(num_workers) as pool:
+        with multiprocessing.Pool(num_workers, _initialize_worker) as pool:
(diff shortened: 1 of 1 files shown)


fiftyone/utils/openimages.py
```diff
@@ -1559,28 +1559,24 @@ def _download_specific_images(
...
         "s3",
         config=botocore.config.Config(signature_version=botocore.UNSIGNED),
     )
 
 
 def _do_s3_download(args):
-    _initialize_pool()
     filepath, filepath_download = args
+
+    global s3_client
```
ehofesmann, line 1579: This can be removed, you only need to use the global keyword when assigning the variable. https://stackoverflow.com/questions/9366212/correct-use-of-global-variables-in-python-3

brimoor:
@ehofesmann do you have any reason to expect that creating a boto client would fail for any reason? I did notice that if the initializer fails, then you do seem to get stuck in an infinite loop, but I don't see why it would ever fail. Creating the client probably does almost nothing

ehofesmann:
> @ehofesmann do you have any reason to expect that creating a boto client would fail for any reason? I did notice that if the initializer fails, then you do seem to get stuck in an infinite loop, but I don't see why it would ever fail. Creating the client probably does almost nothing

Really only if boto3 isn't installed properly. If you don't think that's an issue then this is good to go!

ehofesmann (approved):
LGTM

Issue #1035, opened 2021-05-16
[BUG] s3_client not defined when downloading Open Images dataset


brimoor:
**OS version**: macOS 10.14.6
**Python version**: 3.8
**FiftyOne version**: 0.9.1

When trying to download + load the Open Images dataset:

```py
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset(
    "open-images-v6",
    split="validation",
    label_types=["detections"],
    max_samples=10,
)
```

I encounter the following error:

```
-> 1275     _download_specific_images(valid_ids, split, dataset_dir, num_workers)
   1276 
   1277     if "segmentations" in label_types:
/opt/anaconda3/lib/python3.8/site-packages/fiftyone/utils/openimages.py in _download_specific_images(valid_ids, split, dataset_dir, num_workers, ext)
   1565     with fou.ProgressBar(total=len(inputs)) as pb:
   1566         with multiprocessing.Pool(num_workers) as pool:
-> 1567             for _ in pool.imap_unordered(_do_s3_download, inputs):
   1568                 pb.update()
   1569 
/opt/anaconda3/lib/python3.8/multiprocessing/pool.py in next(self, timeout)
    866         if success:
    867             return value
--> 868         raise value
    869 
    870     __next__ = next                    # XXX
NameError: name 's3_client' is not defined
```

PR #1036, opened 2021-05-16
Fixing multiprocess initialization bug when loading Open Images dataset (#2)

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1035.

Fixes a bug in the way that the `multiprocessing` module was being used that was causing Mac users running Python >= 3.8 to encounter an error when downloading the Open Images dataset.

Verified that this fixes the bug under Python 3.8:

```
conda create -n fifty3 python=3.8
conda activate fifty3

git clone https://github.com/voxel51/fiftyone
cd fiftyone
git checkout --track origin/oi-hotfix
bash install.bash
```

```py
import fiftyone.zoo as foz

# now works
    "open-images-v6", 
    split="validation", 
    label_types=["detections"], 
    max_samples=10,
)
```

### DIFF
--- a/fiftyone/utils/openimages.py
+++ b/fiftyone/utils/openimages.py
@@ -1559,27 +1559,24 @@ def _download_specific_images(
             len(inputs),
         )
 
-    global s3_client
-    s3_client = None
-
     with fou.ProgressBar(total=len(inputs)) as pb:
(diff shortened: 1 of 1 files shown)


ehofesmann:
Going with https://github.com/voxel51/fiftyone/pull/1034 instead

PR #1041, opened 2021-05-18
Looker

### LINKED ISSUE #390: Add support for zooming and panning in expanded sample view
It would be extremely useful to be able to zoom in/out and pan around the viewer in expanded sample mode. Typical use case is to inspect specific regions of large and/or complicated images

In terms of implementation, the obvious best practice is to follow the Google Maps-style interface:
- mouse scroll = zoom in/out
- click and drag = pan
- add `[+] [-]` controls in bottom right to support zooming in increments

Screen Shot 2020-08-13 at 3 02 03 PM

Screen Shot 2020-08-13 at 3 13 42 PM

### LINKED ISSUE #622: [FR] Color video play bar left of current timestamp
On the video player controls, it would be visually helpful if the timeline bar to the left of the current time indicator (the orange dot) were also rendered in orange, to help draw the eye to the current time dot (orange = look right, gray = look left).

As it currently stands, it can be hard to visually locate the orange dot on the timeline.

For reference, see YouTube player controls.

Screen Shot 2020-10-20 at 7 49 33 PM

### LINKED ISSUE #659: [FR] Enable audio for video samples in the App
In use cases like Eric's deep fakes blog, the audio associated with video samples may be a critical part of interpreting the dataset.

We should add support for playing audio in the App, when available. Some questions arise:

1) Play audio by default in grid view?
2) Play audio by default in expanded view?
3) If yes to 1), allow user to toggle audio in the App somewhere?
4) If yes to 2), allow user to toggle audio via an appropriate option in the Player51 menu?

Screen Shot 2020-11-02 at 12 46 24 PM

### LINKED ISSUE #844: [FR] Use image-size-independent font sizes for media overlays
The font size of the overlays on the media viewer in expanded sample mode seems to be proportional to image size or something like that. Often this results in tiny text like the example shown below.

I think it is more appropriate to use a constant font size, or at least one that depends on the App size and is invariant to sample statistics.

Example: bring your eyeglasses for this one!

Screen Shot 2021-02-03 at 9 30 36 PM

### LINKED ISSUE #907: [FR] Add support for switching between video and frame views in the App
When working with video datasets, we've had user requests for the ability to switch between "video view" and "frame view" in the App, where:
- video view: 1 sample per video (current behavior)
- frame view: 1 sample per frame (not currently supported)

This makes a lot of sense, as "frame view" would enable lots of filtering/searching operations that are not well-supported in video view.

The technical challenge here is that "frame view" will require random access to all frames, so the videos will need to be converted to sampled frames to support such a view.

Here's some example code in that direction that I worked up. It currently creates a separate dataset for the frame view:

```py
import os

import eta.core.video as etav

import fiftyone as fo
import fiftyone.zoo as foz


def videos_to_images(video_samples, image_dataset):
    """Adds a sample for each frame in the video dataset to the given image
    dataset.

    This method converts each video to a directory of sampled frames with the
    same base filepath in order to serve the image dataset.

    Args:
        video_samples: a :class:`fiftyone.core.collections.SampleCollection`
            of videos
        image_dataset: an image:class:`fiftyone.core.dataset.Dataset` to which
            to add the sampled frames
    """
    video_samples.compute_metadata()

    image_samples = []
    with fo.ProgressBar() as pb:
        for sample in pb(video_samples):
            video_path = sample.filepath
            frames_dir = os.path.splitext(video_path)[0]
            frames_patt = os.path.join(frames_dir, "%05d.jpg")

            frames = list(range(1, sample.metadata.total_frame_count + 1))
            etav.sample_select_frames(
                video_path, frames, output_patt=frames_patt
            )

            for frame_number, frame in sample.frames.items():
                frame_path = frames_patt % frame_number
                image_sample = fo.Sample.from_frame(frame, frame_path)
                image_samples.append(image_sample)

    image_dataset.add_samples(image_samples)


video_dataset = foz.load_zoo_dataset("quickstart-video")
video_view = video_dataset.take(2)
image_dataset = fo.Dataset("quickstart-video-frames")
videos_to_images(video_view, image_dataset)

session = fo.launch_app(image_dataset)
```

### LINKED ISSUE #940: [BUG] Add segmentation support for video samples in the App
Currently, the first frame's mask will render in the grid and expanded sample view, but later frames will not be updated.

Fully supporting this feature has been overlooked and is blocked ETA's `VideoLabels` supporting only one mask per frame, and the default serialization of numpy arrays by ETA.

More generally, this issue arises from the fact that Player51 does not natively support serialized FiftyOne `Sample`s.

### LINKED ISSUE #991: [FR] Add support for zooming to content in the App
Currently, viewing the patches and evaluation patches views introduced in https://github.com/voxel51/fiftyone/pull/974 in the App results in an App grids that usually contain multiple copies of the same (full) image with different subsets of labels in them.

Especially in the case of patches/evaluation patches, the user may want to zoom into the region of the image that only contains the labels, usually with a small amount of padding. However, the ability to "zoom to content" even in non-patches views would be quite useful and interesting.

Note that the extent of the content may involve computing the convex hull of the spatial labels (detections, polylines, keypoints, segmentations) across multiple label fields.

I can think of a few ways to approach this (not mutually exclusive):

Patch-specific implementations:

**Option A**: Implement a binary toggle, **only** for patch and evaluation patch views, which toggles between showing the full image and showing a zoomed-to-content version of the image with a fixed padding.

Note: the reason I say this is specific to patch views is that this would probably not do much for regular datasets, as the convex hull of all possible labels is probably quite close to `[0, 1] x [0, 1]`. 

Generic dataset implementations:

**Option B**: Implement a "zoom to content" button that can be activated on *any* dataset. Clicking this would zoom to the currently visible content in each sample. From a user understanding perspective, the complexity here is that activating this mode and then adding back labels that were hidden when the zoom occurred may result in labels that are not visible in the current viewport. So, perhaps when this button has been pressed, a "reset" button/link appears next to the button enabling the user to reset to full view. Or, they can click the "zoom to content" button again to re-zoom to the labels that are now visible.

**Option C**: Implement A or B as a slider that allows for adjusting the padding around the zoomed-to-content image. This could either smoothly deform between a `[0, 1] x [0, 1]` viewport and the zoomed viewport, or there could be a reset button that brings one back to `[0, 1] x [0, 1]` and the slider could adjust the padding around a centered-cropped viewport.

**Option D**: Support arbitrary zooming and panning of images in expanded sample mode (https://github.com/voxel51/fiftyone/issues/390). This likely makes sense at some point, independent of whether zoom-to-content is separately supported. Just covering the bases.

In grid mode, activating any of these options would presumably cause a full page reload, since the layout of the images based on their zoomed aspect ratios may change the layout (which is fine).

### LINKED ISSUE #1021: [FR] Show value counts for categorical choices in filters widgets
It would be very helpful to see the counts for each categorical value that appears as a checkbox-ed entry in a filter widget in the App.

This would apply to boolean fields, categorical fields in "small # options mode", as well as any values chosen using the `+ filter by label` dropdown for fields with many choices. We could also show counts for the `None` checkbox associated with numeric slider widgets, if visible (at least 1 `None`).

```
[x] fn (124)
[ ] fp (54)
[ ] tp (12)
```

With counts visible, there's a choice between sorting alphabetically and sorting in descending order of frequency. The graphs tab sorts in order of frequency, which feels and looks natural, so I'd recommend sorting entries by frequency in the filters widgets as well.

Also like the graphs tab, I'd recommend that `None` be treated separately and always be listed last, regardless of its count.

We could provide a `Sort values by` option in the settings menu to toggle between these modes?
- Frequency (and show counts)
- Alphabetically (no counts)    <-- current behavior

Screen Shot 2021-05-10 at 10 00 01 PM

### LINKED ISSUE #1073: Labels drawn with FiftyOne not fully inside the image
Labels drawn with FiftyOne are sometimes not visible since they go outside the image. Is there any fix for this?

### LINKED ISSUE #1140: [FR] Render segmentation pixel values in tooltip when no mask targets are available
When viewing `Segmentation` masks in the App with no `mask_targets` mapping, show

```
<pixel-value>
pixel
```

rather than

```
None
label
```

Screen Shot 2021-07-22 at 1 59 44 PM

benjaminpkane:
### Summary

Reorganizes the app codebase, splitting the project into three packages (and more later, likely)

* `app` - core react app
* `desktop` - desktop package wrapping the core `app` package
* `looker` - image and video visualizer supporting canvas overlay labels, video labels are streamed

Deprecates [player51](https://github.com/voxel51/player51/)

### Issues

* Resolves #991 
* Resolves #940
* Resolves #390 
* Resolves #844
* Resolves #907
* Resolves #1073
* Resolves #1021
* Resolves #622
* Resolves #659
* Resolves #1140


benjaminpkane:
I need to work through a couple build updates, and then I'll share a commit to checkout and install from.

benjaminpkane:
Things can be played with now, although video still needs work and optimizations and bugs fixes are ongoing.

Detections, patches, and the improved expanded view are the best things to look at at the moment.

`bash install.bash`

benjaminpkane:
Overlays are now in canvas in things are looking smooth. `Detection`s and `Segmentation`s are fairly stable. Should be on cruise control at this point. 

Screenshot from 2021-06-07 13-12-23

benjaminpkane:
Screenshot from 2021-06-09 15-28-26

Video still needs work, but patches and frames views are stable. Lots are small details I'm chipping away at, but getting close.

benjaminpkane:
Things are looking good. But video segmentations need some work.

At a minimum, I likely need to move some things off of the main thread with web workers. Maybe streaming, though I don't want that to delay things.

benjaminpkane:
Excluding build changes, things will be ready for eyes on Monday.

benjaminpkane:
I ended up prioritizing build changes because I needed to see how things were working in Firefox, etc. Finishing that up now.

benjaminpkane:
Ready for first impressions

```
bash install.bash
```

Screenshot from 2021-06-25 17-58-34

brimoor:
This is looking really good! 😍 🔥 💯 

Here's my reactions after initial testing this morning:

## grid view

- remove gradient animation from grid loading cells (prefer "flat" design)
- reduce spacing between grid cells (I liked the previous spacing)

## expanded view

#### HUD

- make HUD disappear immediately after mouse leaves window
- reset zoom icon looks too similar to fullscreen and zoom-to-content icons. just remove it?
- reduce scroll zoom sensitivity (does this need to be configurable?...)
- use M:SS / M:SS for video timestamp
- volume/playback sliders
    - only show playback rate slider when hovering on the playback icon or the slider itself (like youtube)
    - only show volume slider when hovering over audio icon or slider (like youtube)
    - with above two points in place, move audio controls to LHS of HUD (like youtube) so that RHS is consistent between image and video
- add thermometer bulb to video timebar when hovering over it (easier click target to drag)

#### keyboard shortcuts

- pressing esc should have the following precedence
    - when keyboard controls are open, close them
    - when settings menu is open, close it
    - when a non-trivial zoom is present, reset it
    - if full screen mode is active, revert to regular expanded modal
    - close the expanded modal
- "=" and "_" should also work for zoom, not just "+" and "-" (no need to list this in controls menu, though)
- dont show attributes or bounding box in tooltip (they dont show useful info)
- show "<" and ">" as video forward/backward controls (still allow "," and "." keystrokes though)
- new keyboard shortcut: pressing 0-9 zooms to 0%, 10%, 20%, ..., 90% timestamp in video

#### zoom

- since you let people zoom out beyond the full extend when scrolling, you should also let them pan outside the image to the same extent

#### RHS sidebar

- all dropdown field elements should autoselect the first result by default so that user can directly press enter without first arrowing down (both filter widgets and view stage parameter dropdowns)
- pressing escape when dropdown list is showing should close it

#### misc

- switching to JSON mode in expanded sample modal crashes the App for image datasets
- why does the blue border appear around the media player in certain circumstances? I think it should never appear
- navigating between videos or closing the modal when a video is playing causes the App to be stuck in a permanent loading state
- frame label counts in RHS of expanded modal for videos always show 0

ehofesmann:
Wow Ben! The video player feels so good!!

Brian covered a lot of the things I came across.

A couple of other things:
* Hovering over fullscreen says "Maximize (m)", but `m` is used for mute
* Use a different icon for zoom to content and fullscreen, maybe adding a slight space between the zoom controls and the fullscreen icon to group the zoom controls together
* When going through a video frame by frame, when you reach the end of the video and go forward a frame, you go to the beginning of the video. But when you're at the beginning of the video and go back a frame, it doesn't go to the last frame. It would be better if you can't loop around from the back so going forward a frame at the end of the video does nothing. 
* Clicking the hud controls bar should not make it disappear, at least not permanently. It happened a few times where I misclicked the seek bar, got rid of the controls, and they wouldn't come back unless I moved my mouse out of the window or pressed "c".

benjaminpkane:
Still working through some data flow changes

benjaminpkane:
Undrafting

benjaminpkane:
Ok, JSON is now viewed via Looker. Will add back copy to clipboard, and adjust styles in the morning with fresh eyes.
Screenshot from 2021-07-14 18-19-18

brimoor:
I had a chance to play around with videos, polylines, and segmentations in Looker this morning:

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

# Image dataset

dataset.name = "looker-image"
dataset.persistent = True

fob.compute_similarity(dataset, brain_key="image_sim")
fob.compute_similarity(dataset, patches_field="ground_truth", brain_key="obj_sim")

# Video dataset

dataset.name = "looker-video"
dataset.persistent = True

# Polyline dataset

    "coco-2017",
    split="validation",
    label_types=["detections", "segmentations"],
    use_polylines=True,
    max_samples=500,
    dataset_name="looker-polylines",
)
dataset.persistent = True

# Segmentation dataset

    "coco-2017",
    split="validation",
    label_types=["detections", "segmentations"],
    max_samples=500,
    dataset_name="looker-segmentations",
)
dataset.persistent = True
```

brimoor:
Here's my notes from my latest testing session:

### Help menu
- Suggest to use "Escape context" instead of "Escape window"
- Put "Next frame" directly below "Previous frame" in help menu for videos

### Graphs
- Trying to show graphs currently crashes the App with any dataset

### Video
- Add option to play video on loop in settings menu for video datasets
- Pressing 0 (to reset video playback) shows -1:XX.XX in video timestamp, should show 00.00:XX.XX
- Frame label fields should be selected by default when loading video datasets (currenly everything is deselected for me)
- Navigating between videos while video is playing does now work, but the loading icon displays for much longer than when navigating without a video curently playing. Is this a bug?

### Expanded modal
- Add "x" in upper left/right to close JSON mode, fullscreen, and expanded sample modal. Although pressing esc or pressing the setting again in the menu is possible, I feel a bit trapped when I get into these modes. I think users will too and will expect to see an "x" icon
- When objects are selected, it would be awesome if zoom to content would only zoom to the selected objects!
- Can I talk you into a few more `esc` functions for the hierarchy??
    - esc should close JSON before closing expanded sample modal
    - esc should unselect currently selected samples (both in grid and expanded sample)
    - esc should reset video to beginning if current frame number != 0
    - With these changes, the new total hierarchy would be something like this:
        - Escape help -> settings -> json -> selection -> playback -> fullscreen -> close

### JSON viewer
- JSON mode needs to not close on mouse click (a common use case here is to select some data, like a label ID). This is another reason for the "x" close icon feature request above
- JSON mode no longer shows only labels for currently visible content. Can we revert to only showing filtered content?

### Other
- # points in tooltip of polylines is always showing None for me. This isn't hugely important, I'd be fine with removing this instead of fixing it
- Data with instance segmentations not loading correctly (see `looker-segmentations` dataset)

ehofesmann:
Great stuff! I played around with the current version of Looker for a bit and have a few comments.

I also noted the issues with plots and with the JSON closing on click that Brian mentioned.

Additional notes:

**Player:**
* When clicking zoom to content I found myself wanting to click it again to zoom back out to the full image. It seems natural to me for it to always zoom to content unless you are already at the exact zoom level and location that it would go to, in that case it would zoom all the way out instead.
* When viewing videos, the settings bar feels like it disappears slightly too fast (2 seconds). I would up it to 3 seconds matching YouTube.

**Sidebar:**
* The colors for fields change between the grid view and the expanded view
* "Sort filter results by" is a bit unclear at the moment, it took like 10 seconds to figure out what it meant. I would suggest changing it to something like "Sort labels sidebar by" since ATM this really only applies to label fields/filter results, right? If it's meant to only sort the filter results, I'd change it to something like "Sort label filters by" or "Sort label classes by"
* As I mentioned, "Sort filter results by" currently sorts both the label fields and the filter results. I'm not sure if that is intended, but it makes it so that the label fields switch around when clicking through samples if you sort by count instead of value. If it is intended I'd recommend defaulting to value instead of count. 
image 
image


* Filtering out all labels by sorting by confidence of one field also shows "0 of X" in the other field even though those aren't filtered out (see the ground truth of the image below)
image

**Other**:
* Allow page up/page down keys in the grid view and when viewing the JSON of a sample. (Though maybe hold off on allowing it in the grid view until the grid is virtualized to avoid people running into slowdown issues even faster)

brimoor:
> When clicking zoom to content I found myself wanting to click it again to zoom back out to the full image. It seems natural to me for it to always zoom to content unless you are already at the exact zoom level and location that it would go to, in that case it would zoom all the way out instead.

I thought about requesting this too (it does seem natural), but held off in the case the definition of "exact zoom level" was dubious given that we're talking about floating point numbers. I previously requested `esc` to reset zoom for this reason. And the user can already double-click to achieve this as well. If we do introduce this feature, the look-and-feel of the zoom-to-content icon in the tray would need to change whenever its behavior would be "reset".

brimoor:
> "Sort filter results by" is a bit unclear at the moment, it took like 10 seconds to figure out what it meant. I would suggest changing it to something like "Sort labels sidebar by" since ATM this really only applies to label fields/filter results, right? If it's meant to only sort the filter results, I'd change it to something like "Sort label filters by" or "Sort label classes by"

If sorting continues to apply to both top-level fields and field values (see below), then perhaps "Sort sidebar contents by" would be more clear? I wouldn't use "label" or "class", because this applies to tags, bools, and strings. If only field values are sorted (see below), then perhaps "Sort field values by"?

> As I mentioned, "Sort filter results by" currently sorts both the label fields and the filter results. I'm not sure if that is intended, but it makes it so that the label fields switch around when clicking through samples if you sort by count instead of value. If it is intended I'd recommend defaulting to value instead of count.

I see what Eric is getting at here. It's a bit tricky. I definitely like having field values inside the expanded sections sorted by count by default, because the frequency of things inherently determines their importance (or you can opt into alphabetical sorting if you prefer).

Does the same reasoning apply to top-level fields? Well, the difference there is that users typically wouldn't have as many top-level fields, so they might not ever get out-of-hand. Also, with top-level fields, we have a third option not applicable to field values- in the status quo, fields are sorted by the order they were added to the dataset. The good thing about this is that, in most cases, `ground_truth` would appear on top, and the newest field would always appear last. And it doesn't cause fields to change order when flipping through different samples (I can see how that could get annoying).

Putting it all together, my thinking is:
(a) field values should definitely be sorted by count by default, with option to change to alphabetical
(b) the ordering of tag values in the sidebar should match setting (a), since tag values are like field values
(c) I think it would be too complicated to add a separate, independent control over how non-tag top-level fields are sorted in the sidebar. Secondly, natural ordering of label fields never bothered me, so perhaps don't change it for now? My second vote would be to continue to link these fields with ordering from (a) and just go with the fact that this could change field orders on a per-sample basis when navigating the expanded sample modal

brimoor:
> Allow page up/page down keys in the grid view and when viewing the JSON of a sample. (Though maybe hold off on allowing it in the grid view until the grid is virtualized to avoid people running into slowdown issues even faster)

Not sure about this one myself. Continuous scrolling is pretty intuitive. In JSON mode, I would want arrow keys to be reserved to help me select contents with shift key held down

ehofesmann:
> I thought about requesting this too (it does seem natural), but held off in the case the definition of "exact zoom level" was dubious given that we're talking about floating point numbers. I previously requested `esc` to reset zoom for this reason. And the user can already double-click to achieve this as well. If we do introduce this feature, the look-and-feel of the zoom-to-content icon in the tray would need to change whenever its behavior would be "reset".

Sure, the "exact zoom level" may not be necessary since it will pretty much always be off slightly if the user just zoomed in themselves. 
I see where you're coming from. Personally, I still think it's reasonable that the zoom goes out when clicking the crop button a second time without having adjusted the zoom or shifting the image. Even without changing the icon, just so that clicking it multiple times actually does something. Though I think I'd be happy with at least adding a second line saying "Reset zoom (double-click/esc)" or so when hovering over the crop icon.

> ... 
Putting it all together, my thinking is:
(a) field values should definitely be sorted by count by default, with option to change to alphabetical
(b) the ordering of tag values in the sidebar should match setting (a), since tag values are like field values
(c) I think it would be too complicated to add a separate, independent control over how non-tag top-level fields are sorted in the sidebar. Secondly, natural ordering of label fields never bothered me, so perhaps don't change it for now? My second vote would be to continue to link these fields with ordering from (a) and just go with the fact that this could change field orders on a per-sample basis when navigating the expanded sample modal

I agree with pretty much everything here. "Sort sidebar contents by" sounds good. Leaving top-level fields as-is (with no option to change it) and sorting field values/tags by count by default is best.

> Not sure about this one myself. Continuous scrolling is pretty intuitive. In JSON mode, I would want arrow keys to be reserved to help me select contents with shift key held down

Just to clarify, I'm suggesting PageUp/PageDown, not arrow keys, which at the moment don't seem to be used for anything. I agree with what you said, scrolling is intuitive but the JSON can get pretty long so faster PgUp/PgDn would be nice.

brimoor:
> > I thought about requesting this too (it does seem natural), but held off in the case the definition of "exact zoom level" was dubious given that we're talking about floating point numbers. I previously requested `esc` to reset zoom for this reason. And the user can already double-click to achieve this as well. If we do introduce this feature, the look-and-feel of the zoom-to-content icon in the tray would need to change whenever its behavior would be "reset".
> 
> Sure, the "exact zoom level" is not necessary since it will pretty much always be off slightly if the user just zoomed in themselves.
> I see where you're coming from. Personally, I still think it's reasonable that the zoom goes out when clicking the crop button a second time without having adjusted the zoom or shifting the image. Even without changing the icon, just so that clicking it multiple times actually does something. Though I think I'd be happy with at least adding a second line saying "Reset zoom (double-click/esc)" or so when hovering over the crop icon.
> 

I'm onboard with the idea that a second click/keypress of zoom to content without any intermediate zooming/panning would reset the view. As an implementation detail, I'd implement it via a simple flag that is set when the action is taken and then unset by any other action, as a cheap way of determining whether the zoom to content action is currently in "zoom" or "reset"  mode.

brimoor:
> > Not sure about this one myself. Continuous scrolling is pretty intuitive. In JSON mode, I would want arrow keys to be reserved to help me select contents with shift key held down
> 
> Just to clarify, I'm suggesting PageUp/PageDown, not arrow keys, which at the moment don't seem to be used for anything. I agree with what you said, scrolling is intuitive but the JSON can get pretty long so faster PgUp/PgDn would be nice.

oh I see. I forget about page up/down because mac keyboards don't have dedicated keys for those. I think it is `fn + up/down arrow` but I never use it. I'm not opposed to those hotkeys but agree that waiting until the grid is faster seems prudent

benjaminpkane:
> Data with instance segmentations not loading correctly (see looker-segmentations dataset)

```py
import fiftyone.zoo as foz

    "coco-2017",
    split="validation",
    label_types=["detections", "segmentations"],
    max_samples=500,
    dataset_name="looker-segmentations",
)
dataset.persistent = True
```

I don't see masks on this dataset.
Screenshot from 2021-07-15 16-45-17

I can see these...
```py
import numpy as np

for sample in dataset:
    for det in sample.ground_truth.detections:
        det.mask = np.asarray([[1, 0], [0, 1]])
        sample.save()
```

Screenshot from 2021-07-15 16-49-20

benjaminpkane:
Ok, most comments have been addressed. Any active field and coloring bugs will be addressed in #1134. I did not add "remove selection" to the escape sequence, although I will try to add that in #1134 as well.

brimoor (approved):
LGTM. Amazing work! 💪 🥇

PR #1047, opened 2021-05-20
Fix the Labelbox ID used when importing samples

### LINKED ISSUE #1046: [BUG] Labelbox Import Uses Incorrect Labelbox ID?
### System information

-   **OS Platform and Distribution**: Linux Ubuntu 20.04
-   **FiftyOne version (run `fiftyone --version`)**:  v0.9.1
-   **Python version**: 3.6.9

### Commands to reproduce

We used upload_media_to_labelbox() to upload images from Voxel51 to Labelbox and then performed bounding box labeling. We then exported a JSON file from Labelbox. We ran the import_from_labelbox() function and encountered this error message for each labeled sample:

```
Skipping labels for unknown Labelbox ID 'ckou3botb2y0d0yw47ftmggdr'; provide a `download_dir` if you wish to download media and create samples for new media
```

We were able to overcome this problem by using the "DataRow ID" field instead of the "ID" field from the exported JSON file.

We believe this problem results from the upload_media function, which uses the Labelbox SDK create_data_row() function, which creates a data row object and then the ID of the data row is saved into Voxel51. The ID in the exported JSON file refers to the ID of the label, not the data row.
https://github.com/voxel51/fiftyone/blob/496e402db4eb2974acb12d26f4ffe926ced2d7b0/fiftyone/utils/labelbox.py#L393-L396

### Recommended Fix

We used the following fix and it worked for us.

We changed the "ID" field below:

https://github.com/voxel51/fiftyone/blob/496e402db4eb2974acb12d26f4ffe926ced2d7b0/fiftyone/utils/labelbox.py#L131

to "DataRow ID"

ehofesmann:
Resolves #1046 

The json format of exported Labelbox annotations follows this format:
```
        [
            {
                "DataRow ID": <labelbox-id>,
                "Labeled Data": <url-or-None>,
                "Label": {...},
                ....
            }
        ]
```

Where the `labelbox_id_field` of the FiftyOne Samples stores the "DataRow ID" field of the Labelbox JSON, not the "ID" field as was previously implemented.

Labels can now be properly imported into an existing FiftyOne dataset from Labelbox annotations.

### DIFF
--- a/fiftyone/utils/labelbox.py
+++ b/fiftyone/utils/labelbox.py
@@ -61,8 +61,7 @@ def import_from_labelbox(
 
         [
             {
-                "ID": <labelbox-id>,
-                "DataRow ID": <datarow-id>,
+                "DataRow ID": <labelbox-id>,
                 "Labeled Data": <url-or-None>,
                 "Label": {...}
             }
@@ -483,7 +482,7 @@ def convert_labelbox_export_to_import(inpath, outpath=None, video_outdir=None):
             )
 
             dout_map[uuid] = {
-                "ID": uuid,
+                "DataRow ID": uuid,
                 "Labeled Data": None,
                 "Label": {"frames": frames_outpath},
             }
@@ -491,7 +490,7 @@ def convert_labelbox_export_to_import(inpath, outpath=None, video_outdir=None):
 
         if uuid not in dout_map:
             dout_map[uuid] = {
-                "ID": uuid,
+                "DataRow ID": uuid,
                 "Labeled Data": None,
                 "Label": {"objects": [], "classifications": []},
             }
--- a/fiftyone/utils/labelbox.py
+++ b/fiftyone/utils/labelbox.py
@@ -62,6 +62,7 @@ def import_from_labelbox(
         [
             {
                 "ID": <labelbox-id>,
+                "DataRow ID": <datarow-id>,
                 "Labeled Data": <url-or-None>,
                 "Label": {...}
             }
@@ -128,7 +129,7 @@ def import_from_labelbox(
     # ref: https://github.com/Labelbox/labelbox/blob/7c79b76310fa867dd38077e83a0852a259564da1/exporters/coco-exporter/coco_exporter.py#L33
     with fou.ProgressBar() as pb:
         for d in pb(d_list):
-            labelbox_id = d["ID"]
+            labelbox_id = d["DataRow ID"]
 
             if labelbox_id in id_map:
                 # Get existing sample


benjaminpkane (approved):
LGTM!

fiftyone/utils/labelbox.py
```diff
@@ -62,6 +62,7 @@ def import_from_labelbox(
         [
             {
                 "ID": <labelbox-id>,
+                "DataRow ID": <datarow-id>,
```
brimoor, line 65: I would recommend updating the documentation like so:

```

        [
            {
                "DataRow ID": <labelbox-id>,
                "Labeled Data": <url-or-None>,
                "Label": {...},
                ...
            }
        ]
```

The point of the documentation is to help the user understand what value they should be storing in the `labelbox_id_field` of their samples

brimoor (changes requested):
Please also replace "ID" with "DataRow ID" in the `convert_labelbox_export_to_import()` function for completeness.

The snippet below is how I tested import/export when originally writing this integration.

The trouble was that I guess I messed up ID vs data row ID in both places, so I thought things were working. Oh and the fact that I wrote the integration based purely on reading documentation and didn't actually try uploading any labels to Labelbox 🤦 

```py
from uuid import uuid4

import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.labelbox as foul


LABELBOX_EXPORT_PATH = "/tmp/labelbox-image-export.json"
LABELBOX_IMPORT_PATH = "/tmp/labelbox-image-import.json"
LABELBOX_ID_FIELD = "labelbox_id"

    "bdd100k", split="validation", shuffle=True, max_samples=10
).clone()

# Generate fake Labelbox IDs, since we haven't actually uploaded there
for sample in dataset:
    sample[LABELBOX_ID_FIELD] = str(uuid4())
    sample.save()

# Export labels in Labelbox format
foul.export_to_labelbox(
    dataset,
    LABELBOX_EXPORT_PATH,
    labelbox_id_field=LABELBOX_ID_FIELD,
    label_prefix="",  # all fields
)

# Convert to Labelbox import format
foul.convert_labelbox_export_to_import(
    LABELBOX_EXPORT_PATH, LABELBOX_IMPORT_PATH
)

# Import labels from Labelbox
foul.import_from_labelbox(
    dataset,
    LABELBOX_IMPORT_PATH,
    label_prefix="lb",
    labelbox_id_field=LABELBOX_ID_FIELD,
)

# Verify that we have two copies of the same labels
```

ehofesmann:
@brimoor Good call, I didn't try `convert_labelbox_export_to_import()` previously. 

Also FYI, I did `import_from_labelbox()` with an actual dataset from Labelbox and it is working properly now.

brimoor (approved):

PR #1048, opened 2021-05-21
Adding support for merging list fields, plus other goodies

brimoor:
### Merging improvements

Adds a number of useful options to `Dataset.merge_samples()` and `Document.merge()`, including:

- `fields (None)`: an optional field or iterable of fields to which to restrict the merge. If provided, fields other than these are omitted from `samples` when merging or adding samples. Can also be dict mapping source field names to destination field names
- `omit_fields (None)`: an optional field or iterable of fields to exclude from the merge. If provided, these fields are omitted from `samples`, if present, when merging or adding samples
- `merge_lists (True)`: whether to merge the elements of list fields (e.g., `tags`) and label list fields (e.g., `Detections` fields) rather than  merging the entire top-level field like other field types. For label lists fields, existing `Label` elements are either replaced (when `overwrite` is True) or kept (when `overwrite` is False) when their `id` matches a label from the provided samples

The default value of `merge_lists` is True, which means that the default behavior of `merge_samples()` has changed slightly, as list fields will now have their elements merged, rather than either overwriting or skipping the entire field.

A few other improvements were added to `Dataset.merge_samples()`, including:
- The `Dataset.merge_samples(overwrite=False)` syntax will now be implemented via efficient aggregation pipelines rather than loading the samples in-memory. The only case where in-memory samples is used is when a `key_fcn` is provided
- Fixes a bug where non-unique indexes would be permanently upgraded to unique indexes when merging (the user may later want to add duplicate `filepath` samples, for example, which should always be allowed, except when merging on `filepath`)

Extensive unit tests are added to ensure that all the various options work as expected.

### Other enhancements

- Added optional `Document.copy(fields=None, omit_fields=None)` parameters to only include a subset of a document's fields when copying it
- Fixed a bug that prevented `ViewExpression.contains(value)` from working when `value` is a `ViewExpression` rather than a primitive value

### Performance improvements

Also, in **PERFORMANCE NEWS**, importing datasets may now be up to 2x faster or more!

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Now 2x faster on my machine
```

The issue was that `Dataset.add_samples()` used to choose a batch size of 128 for image datasets and 1 for video datasets, **BUT** for empty datasets with no `media_type`, it conservatively had to choose a batch size of 1 in case the samples happened to be videos, which would happen when importing *any* dataset using `load_zoo_dataset()` as well as `Dataset.from_*()` factory methods. 🤦 🤦 🤦 

So, I finally got around to implementing a dynamic batching strategy in `Dataset.add_samples()` whereby samples will be inserted into the database in dynamically-sized batches such that the time between insertions is ~0.2 seconds. This value was chosen so that the progress bar always updates at a reasonable frame rate.

Note that `Dataset.add_samples(samples)` accepts not just an in-memory list of samples, but also things like generators, which could, for example, involve parsing some binary data, reading from disk, downloading a file from the web, etc. In any case, the point is that the progress bar tracks more than just DB insertion time, it may also magically track sample preparation times if the user provides a generator. I think that's pretty nifty. 

### Example usage

A primary motivating use case for this feature is exporting annotations to an annotation vendor, where annotations could be edited and then loaded back into a dataset. The following snippet demonstrates how such a workflow is now possible:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Export low confidence predictions for correction
low_conf_preds = dataset.filter_labels("predictions", F("confidence") < 0.5)
low_conf_preds.export("/tmp/cvat/export", fo.types.CVATImageDataset, label_field="predictions")

# Add, delete, or modify annotations in CVAT

# Load edits
edits = fo.Dataset.from_dir("/tmp/cvat/export", fo.types.CVATImageDataset, label_field="")

# Merge edits into dataset
dataset.filter_labels("predictions", F("confidence") >= 0.5).save()  # deletes edited predictions
dataset.merge_samples(edits, fields={"detections": "predictions"})  # add replacements
```

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -463,33 +463,18 @@ class SampleCollection(object):
 
         return field_name in self.get_frame_field_schema()
 
-    def validate_fields_exist(self, field_or_fields):
-        """Validates that the collection has fields with the given names.
+    def validate_fields_exist(self, fields):
+        """Validates that the collection has field(s) with the given name(s).
 
-        If ``field_or_fields`` contains an embedded field name such as
-        ``field_name.document.field``, only the root ``field_name`` is checked
-        for existence.
+        If embedded field names are provided, only the root field is checked.
 
         Args:
-            field_or_fields: a field name or iterable of field names
+            fields: a field name or iterable of field names
 
         Raises:
             ValueError: if one or more of the fields do not exist
         """
-        if etau.is_str(field_or_fields):
-            fields = [field_or_fields]
-        else:
-            fields = field_or_fields
-
-        if self.media_type == fom.VIDEO:
-            frame_fields = list(
-                filter(lambda n: n.startswith(self._FRAMES_PREFIX), fields)
-            )
-            fields = list(
-                filter(lambda n: not n.startswith(self._FRAMES_PREFIX), fields)
-            )
-        else:
-            frame_fields = []
+        fields, frame_fields = self._split_frame_fields(fields)
 
         if fields:
             schema = self.get_field_schema(include_private=True)
@@ -522,7 +507,7 @@ class SampleCollection(object):
 
             for field in frame_fields:
                 # We only validate that the root field exists
-                field_name = field.split(".", 2)[1]  # removes "frames."
+                field_name = field.split(".", 1)[0]
                 if (
                     field_name not in frame_schema
                     and field_name not in default_frame_fields
@@ -5360,6 +5345,15 @@ class SampleCollection(object):
             "    %s %s" % ((field_name + ":").ljust(max_len), str(field))
             for field_name, field in field_schema.items()
         )
+
+    def _split_frame_fields(self, fields):
+        if etau.is_str(fields):
+            fields = [fields]
+
+        if self.media_type != fom.VIDEO:
+            return fields, []
+
+        return fou.split_frame_fields(fields)
 
     def _parse_field_name(
         self,
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -1439,6 +1439,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         insert_new=True,
         expand_schema=True,
         omit_default_fields=False,
+        merge_lists=False,
         overwrite=True,
         include_info=True,
         overwrite_info=False,
@@ -1472,8 +1473,11 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
                 if a sample's schema is not a subset of the dataset schema
             omit_default_fields (False): whether to omit default sample fields
                 when merging. If ``True``, ``insert_new`` must be ``False``
+            merge_lists (False): whether to merge top-level list fields and the
+                elements of label list fields. If ``True``, this parameter
+                supercedes the ``overwrite`` parameter for list fields
             overwrite (True): whether to overwrite (True) or skip (False)
-                existing sample fields
+                existing fields
             include_info (True): whether to merge dataset-level information
                 such as ``info`` and ``classes``. Only applicable when
                 ``samples`` is a
@@ -1488,6 +1492,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
             isinstance(samples, foc.SampleCollection)
             and key_fcn is None
             and overwrite
+            and not merge_lists
         ):
             _merge_samples(
                 samples,
@@ -1498,6 +1503,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
                 insert_new=insert_new,
                 expand_schema=expand_schema,
                 omit_default_fields=omit_default_fields,
+                merge_lists=merge_lists,
                 include_info=include_info,
                 overwrite_info=overwrite_info,
             )
@@ -1544,6 +1550,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
                             sample,
                             omit_fields=omit_fields,
                             omit_none_fields=omit_none_fields,
+                            merge_lists=merge_lists,
                             overwrite=overwrite,
                             expand_schema=expand_schema,
                         )
@@ -3817,6 +3824,7 @@ def _merge_samples(
     insert_new=True,
     expand_schema=True,
     omit_default_fields=False,
+    merge_lists=False,
     include_info=True,
     overwrite_info=False,
 ):
@@ -3940,6 +3948,8 @@ def _merge_samples(
                 }
             }
         )
+
+    # @todo support `overwrite=False` and `merge_lists=True`
 
     sample_pipeline.append(
         {
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -1416,7 +1416,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         if fields is not None:
             if etau.is_str(fields):
                 fields = [fields]
-            else:
+            elif not isinstance(fields, dict):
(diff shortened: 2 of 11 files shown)


ehofesmann (changes requested):
This is awesome! Everything works well from what I've seen, I just have a couple of points of discussion:

1) 
The field names should not have to match. In the case where the `edit` dataset isn't just a throw-away, we don't want users to be required to rename their sample fields. It would be much more natural to provide arguments to `merge_samples` that cover field names from both datasets. Something like `existing_field` and `incoming_field` (there are probably better words for this)

2)
I'm trying to figure out the proper workflow for this entire process. When I was testing this, I took an image and uploaded it to CVAT and added new annotations. From there, I downloaded the new annotations file (and optionally the images). The annotation file has no reference to filepaths, only filenames. The issue comes in when trying to figure out the best way to import those new annotations. We don't have the option to import a dataset `from_dir` without an associated image directory. You could just move the downloaded annotations file to the same directory as your images, but then you would manually need to move the original `labels.xml` so you can add your new downloaded `labels.xml` to that dir. You could symlink/copy them to a new directory and easily import your dataset `from_dir` there, but then you need to use a `key_fcn` to match on fileNAME instead of path which sounds like it is inefficient.
In my mind, there are two ways to tackle this:
* Allow importing datasets with only labels and then make it easier to match samples based on filename instead of filepath without requiring a `key_fcn`.
* Update relevant importers to let you specify the `data` directory and labels files separately

I'm sure there are other approaches that may be better.



Also nice work on the dataset loading optimization! On my machine, it was actually more like a 3x speedup to load cifar10. For 1000 samples of Open Images, it was more like a 25% speedup, but there more processing happening in that so it's still a nice improvement.

brimoor:
> This is awesome! Everything works well from what I've seen, I just have a couple of points of discussion:
> 
> 
> The field names should not have to match. In the case where the `edit` dataset isn't just a throw-away, we don't want users to be required to rename their sample fields. It would be much more natural to provide arguments to `merge_samples` that cover field names from both datasets. Something like `existing_field` and `incoming_field` (there are probably better words for this)
> 

Hmm I do see your point (eg, the code in the PR description exhibits the point you make; it would be a bit cleaner if `rename_sample_field()` weren't explicitly needed). Perhaps I could allow `fields` to be a dict mapping source to destination fields. My hesitation in pulling the trigger on that would be that "merging" documents has a particular meaning---producing a document that contains the fields of both documents together, with some choice of overwrite---and if we add too many more bells and whistles, it starts to go beyond the method name. But, I'll think about it.

> 
> I'm trying to figure out the proper workflow for this entire process. When I was testing this, I took an image and uploaded it to CVAT and added new annotations. From there, I downloaded the new annotations file (and optionally the images). The annotation file has no reference to filepaths, only filenames. The issue comes in when trying to figure out the best way to import those new annotations. We don't have the option to import a dataset `from_dir` without an associated image directory. You could just move the downloaded annotations file to the same directory as your images, but then you would manually need to move the original `labels.xml` so you can add your new downloaded `labels.xml` to that dir. You could symlink/copy them to a new directory and easily import your dataset `from_dir` there, but then you need to use a `key_fcn` to match on fileNAME instead of path which sounds like it is inefficient.
> In my mind, there are two ways to tackle this:
> 
> * Allow importing datasets with only labels and then make it easier to match samples based on filename instead of filepath without requiring a `key_fcn`.
> * Update relevant importers to let you specify the `data` directory and labels files separately
> 
> I'm sure there are other approaches that may be better.
> 
> Also nice work on the dataset loading optimization! On my machine, it was actually more like a 3x speedup to load cifar10. For 1000 samples of Open Images, it was more like a 25% speedup, but there more processing happening in that so it's still a nice improvement.

Thanks for sharing your workflow, it's definitely helpful to consider concrete use cases. I have a few thoughts here:

1. Regarding the desire to import labels without duplicating the media, I think https://github.com/voxel51/fiftyone/pull/1050 provides a helpful ingredient for that by supporting `data.json` files in place of actual media files, since you could now construct a `data.json` file for your to-be-imported dataset without interfering with the existing dataset's files that you're merging into. A next step could be to make the `data.json` file optional for dataset types where the labels file itself contains filenames or filepaths. The trouble here is that if the labels truly only contain `filename` with no directory, then there's not enough information there to create a *new* Sample, it would only be sufficient when merging using `key_fcn = lambda sample: os.path.basename(sample.filepath)`. But, when combined with point 2 below, maybe that's okay.

Perhaps we could add a `Dataset.from_labels(labels_path, dataset_type)` factory method, which would *only* support labeled dataset types where a single labels file contains all necessary information. Maybe it could have an optional `data_dir` argument which could specify a directory name in which all `filename`s in the labels files are presumed to reside.

2. We could add `merge_dir()` and `merge_importer()` functions to `Dataset` that would be similar to `add_dir()` and `add_importer()` but would add support for specifying a `key_field` or `key_fcn` that would be used to decide at-load-time whether to create a new Sample or merge the data into existing samples. This method could support arguments like `insert_new` that would provide the ability to specify that all data you're merging in should correspond to existing samples, and thus, for example, its not a problem if the labels file only contains `filename`s, because you would presumably have used a `key_fcn` that matches only based on filenames. In any case, the utility of `merge_dir()` would be to specifically serve use cases like importing edits without having to explicitly create a temporary dataset that is then thrown away. But, maybe it is more work than its worth, since addressing point 1 on its own would already support the workflow via a simple create + merge workflow in two steps.

ehofesmann:
> 1. Regarding the desire to import labels without duplicating the media, I think #1050 provides a helpful ingredient for that by supporting `data.json` files in place of actual media files, since you could now construct a `data.json` file for your to-be-imported dataset without interfering with the existing dataset's files that you're merging into. A next step could be to make the `data.json` file optional for dataset types where the labels file itself contains filenames or filepaths. The trouble here is that if the labels truly only contain `filename` with no directory, then there's not enough information there to create a _new_ Sample, it would only be sufficient when merging using `key_fcn = lambda sample: os.path.basename(sample.filepath)`. But, when combined with point 3 below, maybe that's okay.
> 2. Following on 1, perhaps we could add a `Dataset.from_labels(labels_path, dataset_type)` factory method, which would _only_ support labeled dataset types where a single labels file contains all necessary information. Maybe it could have an optional `data_dir` argument which could specify a directory name in which all `filename`s in the labels files are presumed to reside.
> 3. We could add `merge_dir()` and `merge_importer()` functions to `Dataset` that would be similar to `add_dir()` and `add_importer()` but would add support for specifying a `key_field` or `key_fcn` that would be used to decide at-load-time whether to create a new Sample or merge the data into existing samples. This method could support arguments like `insert_new` that would provide the ability to specify that all data you're merging in should correspond to existing samples, and thus, for example, its not a problem if the labels file only contains `filename`s, because you would presumably have used a `key_fcn` that matches only based on filenames. In any case, the utility of `merge_dir()` would be to specifically serve use cases like importing edits without having to explicitly create a temporary dataset that is then thrown away. But, maybe it is more work than its worth, since addressing point 1 on its own would already support the workflow via a simple create + merge workflow in two steps.

I really like the idea of a `Dataset.from_labels(labels_path, data_path, dataset_type)`, I do think that being able to specify a `data_dir` is the way to go (`from_labels` could also allow for a `data.json` I would think)

I agree that `merge_dir()` and `merge_importer()` are unnecessary at the moment.

brimoor:
Okay I've converted the second part of the discussion to an issue https://github.com/voxel51/fiftyone/issues/1061, since it goes beyond the scope of this PR.

brimoor:
@ehofesmann FYI I addressed point 1 in this PR. The updated CVAT example in the PR description shows that you can now rename fields during a merge w/o an explicit need for `rename_sample_field()`!

Still planning to address point 2 in a separate PR once #1050 is merged.

PR #1051, opened 2021-05-22
Adding a match_labels() stage

brimoor:
### Label matching

Adds a new `match_labels()` view stage that enables matching samples that have specific labels, but without actually filtering the non-matching labels.

Matches can be specified via either `ids` or `tags`, like `select_labels()`, or via a `filter` expression that defines the matching labels, like `filter_labels()`.

This stage is useful in cases such as inspecting detection evaluation results where one may want to inspect an *entire* sample that contains a false positive prediction without actually filtering the true positives, since one needs to see the whole situation to diagnose why a particular label was deemed a false positive.

### Other updates

- Adds an optional `bool` flag to `MatchTags()` that allows for optionally matching samples **without** the specified tags
- Adds missing implementations of `get_filtered_fields()` to `SelectLabels` and `ExcludeLabels`
- Fixes a bug in `LimitLabels()` that would cause views to contain empty label lists if the source dataset contains `None`-valued fields

### Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

#
# Only show samples whose labels are currently selected in the App
#

# Select some labels in the App...

view = dataset.match_labels(labels=session.selected_labels)

#
# Only include samples that contain labels with the specified IDs
#

# Grab some label IDs
ids = [
    dataset.first().ground_truth.detections[0].id,
    dataset.last().predictions.detections[0].id,
]

view = dataset.match_labels(ids=ids)

print(len(view))
print(view.count("predictions.detections"))

#
# Only include samples that contain labels with the specified tags
#

# Grab some label IDs
ids = [
    dataset.first().ground_truth.detections[0].id,
    dataset.last().predictions.detections[0].id,
]

# Give the labels a "test" tag
dataset = dataset.clone()  # create copy since we're modifying data
dataset.select_labels(ids=ids).tag_labels("test")

print(dataset.count_values("ground_truth.detections.tags"))
print(dataset.count_values("predictions.detections.tags"))

# Retrieve the labels via their tag
view = dataset.match_labels(tags="test")

print(len(view))
print(view.count("predictions.detections"))

#
# Only include samples that contain labels matching a filter
#

filter = F("confidence") > 0.99
view = dataset.match_labels(filter=filter, fields="predictions")

print(len(view))
print(view.count("predictions.detections"))
```

### DIFF
--- a/fiftyone/__public__.py
+++ b/fiftyone/__public__.py
@@ -138,6 +138,7 @@ from .core.stages import (
     GeoWithin,
     MapLabels,
     Match,
+    MatchLabels,
     MatchTags,
     Mongo,
     Shuffle,
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1984,6 +1984,9 @@ class SampleCollection(object):
 
         -   Provide the ``tags`` argument to exclude labels with specific tags
 
+        If multiple criteria are specified, labels must match all of them in
+        order to be excluded.
+
         By default, the exclusion is applied to all
         :class:`fiftyone.core.labels.Label` fields, but you can provide the
         ``fields`` argument to explicitly define the field(s) in which to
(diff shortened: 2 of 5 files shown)


benjaminpkane (approved):
LGTM :+1:

PR #1052, opened 2021-05-24
Adding support for exporting image patches

### LINKED ISSUE #1049: [FR] Export patches view as classification dataset
It would be useful to construct a cropped patches view of a `Detections` field and be able to export it to disk as a classification dataset where every sample is a cropped object classified as the detection label.

```python
import fiftyone as fo
import fiftyone.zoo as foz

patches_view = dataset.to_patches("ground_truth")

patches_view.export(
    export_dir="/tmp/classification_dataset",
    dataset_type=fo.types.ImageClassificationDirectoryTree,
    label_field="ground_truth",
)
```

brimoor:
Resolves #1049.

When a dataset export involving a spatial field (eg `Detections`) and a classification dataset format (eg `ImageClassificationDirectoryTree`) is detected, the export routine will automatically assume that the user wants to export the *image patches*.

Also adds support for automatically coercing single label fields like `Detection` into `Detections` if the exporter expects list-type fields.

One more thing: previously label type coercions were only applied when a single `label_field` is being exported. Now, label coercions are applied if possible even when multiple fields are being exported.

### Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart").limit(5).clone()

    sample["attribute"] = fo.Classification(label=str(idx))
    sample.save()

view = dataset.to_patches("ground_truth")

#
# Automatic upgrade from `Detection` to `Detections`
#
# This also demonstrates that exporters gracefully handle path collisions when
# the same image is exported multiple times
#
view.export("/tmp/quickstart/coco", fo.types.COCODetectionDataset)

#
# Exporting unlabeled image patches
#
# If `label_field` is omitted, the raw images are exported. However, since a `label_field` is
# provided that is a valid patches type, image patches are exported instead
#
dataset.export("/tmp/quickstart/images1", fo.types.ImageDirectory, label_field="ground_truth")
view.export("/tmp/quickstart/images2", fo.types.ImageDirectory, label_field="ground_truth")

#
# Exporting labeled image patches
#
# `label_field` is a `Detections` field, so the image patches are exported as a classification dataset
#
dataset.export("/tmp/quickstart/tree1", fo.types.ImageClassificationDirectoryTree, label_field="ground_truth")

#
# Automatic export of labeled image patches
#
# In this case, `label_field` is not explicitly specified, so the routine first checks for a `Classification`
# field to export, and then checks (and finds) a spatial field to export patches
#
view.export("/tmp/quickstart/tree2", fo.types.ImageClassificationDirectoryTree)

#
# Multi-field exports
#
# BDD format supports both classifications and detections. So, the first applicable detections field
# (`ground_truth`) is exported, and the `attribute` field is coerced into lists and exported as well
#
# In the second variation, only `attribute` is exported, still with coercion into lists for compatibility 
#
dataset.export("/tmp/quickstart/bdd1", fo.types.BDDDataset)
dataset.export("/tmp/quickstart/bdd2", fo.types.BDDDataset, label_field="attribute")
```

### DIFF
--- a/docs/source/user_guide/export_datasets.rst
+++ b/docs/source/user_guide/export_datasets.rst
@@ -81,6 +81,108 @@ a |DatasetView| into any format of your choice via the basic recipe below.
     particular label field that you wish to export. This is necessary your
     FiftyOne dataset contains multiple label fields.
 
+.. _export-label-coercion:
+
+Label type coercion
+-------------------
+
+For your convenience, the
+:meth:`export() <fiftyone.core.collections.SampleCollection.export>` method
+will automatically coerce the data to match the requested export types in a
+variety of common cases listed below.
+
+Object patches
+~~~~~~~~~~~~~~
+
+When exporting in either an unlabeled image or image classification format, if
+a spatial label field (|Detection|, |Detections|, |Polyline|, or |Polylines|)
+is provided to
+:meth:`export() <fiftyone.core.collections.SampleCollection.export>`, the
+object patches of the provided samples will be exported.
+
+.. code-block:: python
+    :linenos:
+
+    import fiftyone as fo
+    import fiftyone.zoo as foz
+
+    dataset = foz.load_zoo_dataset("quickstart")
+
+    # No label field is provided; only images are exported
+    dataset.export("/tmp/quickstart/images", fo.types.ImageDirectory)
+
+    # A detections field is provided, so the object patches are exported as a
+    # directory of images
+    dataset.export(
+        "/tmp/quickstart/patches",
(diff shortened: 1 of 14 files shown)


ehofesmann (approved):
Very nice! I like that it will automatically export patches when specifying a classification dataset type for a detections field.

I have added a section about this in the `Object patches views` docs:
image

brimoor:
@ehofesmann thanks for the docs! I think they make more sense on the export user guide page, so I migrated them over there and left a `:note` on the views page.

PR #1053, opened 2021-05-27
Syntax sugar for running multiple aggregations of the same type

brimoor:
Adds support for executing an aggregation on multiple fields via an abbreviated syntax:

```
tuple, of, results = SampleCollection.<aggregation_method>(("tuple", "of", "fields"), **kwargs)
```

Example usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# old way
ids, filepaths = dataset.aggregate([fo.Values("id"), fo.Values("filepath")])

# new option
ids, filepaths = dataset.values(["id", "filepath"])
```


benjaminpkane (approved):
LGTM! Convenient

PR #1055, opened 2021-05-27
Adding support for rendering edges on plotly scatterplots

brimoor:
Adds an optional `edges` argument to the `scatterplot()` and `location_scatterplot()` methods of `fiftyone.core.plots.plotly.scatterplot()` that enables drawing undirected edges between scatterpoints.

Edges are always rendered in a separate trace, which has no selection callbacks (the corresponding nodes should be selected).

Also gets rid of the `ManualInteractiveScatter` class, which wasn't actually being used, because it was going to be a slight pain to support edges traces for that class.

This will be useful for upcoming brain features 😄 

Screen Shot 2021-05-27 at 11 22 01 AM

Screen Shot 2021-05-27 at 11 22 32 AM

### DIFF
--- a/fiftyone/core/plots/plotly.py
+++ b/fiftyone/core/plots/plotly.py
@@ -44,7 +44,7 @@ def plot_confusion_matrix(
     gt_field=None,
     pred_field=None,
     colorscale="oranges",
-    layout=None,
+    **kwargs,
 ):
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
LGTM :+1:

PR #1059, opened 2021-05-29
Adding support for applying Lightning Flash models to datasets

ehofesmann:
Adds support for applying [Lightning Flash](https://github.com/PyTorchLightning/lightning-flash) models to datasets via `SampleCollection.apply_model()` and `SampleCollection.compute_embeddings()`.

This integration requires `flash==0.4.0` or later.

This branch doubles as the release branch for `fiftyone==0.10.0`.

Predictions example:

```py
from flash.image import ObjectDetector
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart").exclude_fields("predictions").clone()

model = ObjectDetector.load_from_checkpoint("https://flash-weights.s3.amazonaws.com/object_detection_model.pt")

view = dataset.take(5)

```

Embeddings example:

```py
from flash.image import ImageEmbedder
import fiftyone as fo
import fiftyone.zoo as foz

model = ImageEmbedder(backbone="resnet101", embedding_dim=128)

view = dataset.take(5)
embeddings = view.compute_embeddings(model)

print(embeddings.shape)
```

See https://github.com/PyTorchLightning/lightning-flash/pull/360 for many more examples!


fiftyone/core/models.py
```diff
@@ -182,6 +197,55 @@ def apply_model(
         )
 
 
+def apply_flash_model(
```
brimoor, line 200: Exposing this here is not necessary, since `apply_model()` itself accepts Flash models. As a matter of design taste, I'd recommend keeping Flash-specific code in `fiftyone.utils.flash`.

fiftyone/core/models.py
```diff
@@ -78,6 +81,18 @@ def apply_model(
         skip_failures (True): whether to gracefully continue without raising an
             error if predictions cannot be generated for a sample
     """
+    if isinstance(model, fcm.Task):
```
brimoor, line 84: I'd prefer to add an `fiftyone.utils.flash.is_flash_model()` method that is used here, to avoid the need to explicitly import `flash` in this file (even though it is only a `lazy_import`).

It's nice to keep all usages of utils in a single place, similar to how we treat `tensorflow`, `torch`, etc.

fiftyone/utils/flash.py
```diff
@@ -0,0 +1,93 @@
...
+            logits
+        batch_size (None): an optional batch size to use. Only applicable for
+            image samples
+        num_workers (None): the number of workers to use when loading images.
+            Only applicable for Torch models
+        skip_failures (True): whether to gracefully continue without raising an
+            error if predictions cannot be generated for a sample
+    """
+    serializer = _get_fo_serializer(model, confidence_thresh)
+    model.serializer = serializer
```
brimoor, line 68: There is a `fiftyone.core.utils.SetAttributes` context manager that you can use to replace `model.serializer` only while the code is running.

It's generally bad manners to permanently modify a user's object.

fiftyone/utils/flash.py
```diff
@@ -0,0 +1,93 @@
...
+    with fou.ProgressBar() as pb:
```
brimoor, line 69: I haven't investigated in great detail, but shouldn't Flash provide support for running prediction on an entire dataset via a single call to `predict()` that would then use their fancy data loading + distributed setup to perform the inference?

In other words, shouldn't you instead use the syntax developed in https://github.com/voxel51/lightning-flash/pull/2 here?

This may be a question for the Flash team if you can't find anything on your own.

brimoor (changes requested):

fiftyone/core/models.py
ehofesmann: If we check every model that gets passed into `apply_model()` for if it's a flash model, then anytime that `apply_model` gets called you would need to have flash installed even if we lazily import it. Is there any way around it? Maybe checking if it's a FiftyOne model instead?

brimoor: yeah you definitely need to implement such that Flash doesn't have to be installed.

One way to implement that is to use `inspect.getmro()` + `etau.get_class_name()` so that you can use string comparison to determine if `model` is an instance of `fcm.Task`

brimoor: https://stackoverflow.com/questions/1401661

fiftyone/utils/flash.py
ehofesmann: So I've updated this to just use `predict()`:
https://github.com/voxel51/fiftyone/blob/f8a65320b2ba23ec070f69bd3b2a5f5b294d0bde/fiftyone/utils/flash.py#L72-L76

The only issue now is that there is no progress bar, I guess we just have to live with that?

fiftyone/utils/flash.py
```diff
@@ -0,0 +1,140 @@
...
+        skip_failures (True): whether to gracefully continue without raising an
+            error if predictions cannot be generated for a sample
+    """
+    if batch_size is not None:
+        print("Flash models only support the default batch size")
+
+    serializer = _get_fo_serializer(model, confidence_thresh, store_logits,)
+    with fou.SetAttributes(model, serializer=serializer):
+        predictions = model.predict(samples.values("filepath"))
+        for sample, prediction in zip(samples, predictions):
```
brimoor, line 71: As long as `predictions` contains a list of `Label` instances and not dicts of labels, `samples.set_values(label_field, predictions)` will be way faster here.

The explicit for loop requires loading all samples into memory, adding labels, and saving them. One by one

fiftyone/utils/flash.py
```diff
@@ -0,0 +1,140 @@
...
+    mro = inspect.getmro(model_class)
+    for c in mro:
+        class_name = etau.get_class_name(c)
+        if class_name == "flash.core.model.Task":
+            return True
+
+    return False
+
+
+def normalize_labels(filepaths, predictions):
```
brimoor, line 93: rename to `normalize_detections()` for clarity

fiftyone/utils/flash.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def _get_fo_serializer(model, confidence_thresh, store_logits):
+    """Initializes the FiftyOne serializer to be used for the given task"""
+    previous_serializer = model.serializer
+    if isinstance(model, fcc.ClassificationTask):
+        prev_args = dict(inspect.getmembers(model.serializer))
+        multi_label = False
+        if "multi_label" in prev_args:
+            multi_label = prev_args["multi_label"]
```
brimoor, line 121: `multi_label = prev_args.get("multi_label", False)` is a clean way to write these 3 lines

fiftyone/utils/flash.py
```diff
@@ -0,0 +1,140 @@
...
+            logits
+        batch_size (None): an optional batch size to use. Only applicable for
+            image samples
+        num_workers (None): the number of workers to use when loading images.
+            Only applicable for Torch models
+        skip_failures (True): whether to gracefully continue without raising an
+            error if predictions cannot be generated for a sample
+    """
+    if batch_size is not None:
+        print("Flash models only support the default batch size")
```
brimoor, line 66: Prefer `logger.info()` over `print()`, since this allows users to configure all logging to, for example, be written to disk, if they'd like

fiftyone/utils/flash.py
```diff
@@ -0,0 +1,140 @@
...
+    confidence_thresh=None,
+    store_logits=False,
+    batch_size=None,
+    num_workers=None,
+    skip_failures=True,
+):
+    """Applies the given ``flash.core.model.Task`` to the samples in the 
+    collection.
+
+    This method supports all the following cases:
```
brimoor, line 41: These things aren't true anymore

docs/source/integrations/lightning_flash.rst
```diff
@@ -0,0 +1,446 @@
...
+the predictions of Flash models to a FiftyOne |Dataset| for visualiziation and
+analysis.
+
+The following Flash tasks are supported by FiftyOne:
+
+- :ref:`Image Classification <flash:image_classification>`
+- :ref:`Image Object Detection <flash:object_detection>`
+- :ref:`Image Semantic Segmentation <flash:semantic_segmentation>`
+- :ref:`Video Classification <flash:video_classification>`
+
```
brimoor, line 21: @ehofesmann image embeddings task are also supported :)

docs/source/integrations/lightning_flash.rst
```diff
@@ -0,0 +1,446 @@
...
+            
+            import fiftyone as fo
+            import fiftyone.zoo as foz
+        
+            from flash import Trainer
+            from flash.image import ObjectDetectionData, ObjectDetector
+            from flash.image.detection.serialization import FiftyOneDetectionLabels
+        
+            # 1. Load your FiftyOne dataset
+            dataset = foz.load_zoo_dataset("quickstart", max_samples=40)
```
brimoor, line 128: Might be nice to add a comment here that you're using views into one dataset, but using different datasets for each split is also allowed

docs/source/integrations/lightning_flash.rst
```diff
@@ -0,0 +1,446 @@
...
+you can use the FiftyOne integrations to
+add generate and add model predictions to your |Dataset| or |DatasetView|.
+
+
+Apply model
+-----------
+
+The easiest way to generate predictions on an existing |Dataset| or |DatasetView| is
+to use the :meth:`apply_model() <fiftyone.core.collections.SampleCollection.apply_model>`
+function, passing in your Flash model.
```
brimoor, line 365: I would mention here that, behind the scenes, FiftyOne will construct the appropriate Flash `Trainer()` and use it to perform possibly-distributed inference to generate the predictions.

Speaking of, we should allow `kwargs` to be passed to `apply_model()` that, in the case of Flash models, will be passed to `Trainer(**kwargs)`, so that the user can provide arguments like `num_gpus=32` to configure the distributed inference.

docs/source/integrations/lightning_flash.rst
```diff
@@ -0,0 +1,446 @@
...
+.. default-role:: code
+
+We have collaborated with the team behind 
+`PyTorch Lightning Flash <https://github.com/PyTorchLightning/lightning-flash>`_ to make it
+as easy as possible to train Flash :class:`models <flash:flash.core.model.Task>`
+on your FiftyOne datasets and to add
+the predictions of Flash models to a FiftyOne |Dataset| for visualiziation and
+analysis.
+
+The following Flash tasks are supported by FiftyOne:
```
brimoor, line 15: supported -> supported natively

docs/source/integrations/lightning_flash.rst
```diff
@@ -0,0 +1,446 @@
+.. _flash:
+
+PyTorch Lightning Flash Integration
+===================================
+
+.. default-role:: code
+
+We have collaborated with the team behind
```
brimoor, line 8: recommended change:

```
We've collaborated with the
`PyTorch Lightning Flash <https://github.com/PyTorchLightning/lightning-flash>`_
team to make it easy to train :class:`Flash models <flash:flash.core.model.Task>`
on your FiftyOne datasets and add predictions from Flash models to your
:ref:`FiftyOne datasets <using-datasets>` for visualization and analysis, all in just
a few lines of code.
```

docs/source/integrations/lightning_flash.rst
```diff
@@ -0,0 +1,446 @@
...
+        
+            # 1. Load your FiftyOne dataset
+            dataset = foz.load_zoo_dataset("cifar10", split="test", max_samples=40)
+            train_dataset = dataset.shuffle(seed=51)[:20]
+            test_dataset = dataset.shuffle(seed=51)[20:25]
+            val_dataset = dataset.shuffle(seed=51)[25:30]
+            predict_dataset = dataset.shuffle(seed=51)[30:40]
+        
+            # 2. Load the Datamodule
+            datamodule = ImageClassificationData.from_fiftyone_datasets(
```
brimoor, line 60: We're renaming to a single `from_fiftyone()` method, right?

docs/source/integrations/lightning_flash.rst
```diff
@@ -0,0 +1,446 @@
...
+- :ref:`Video Classification <flash:video_classification>`
+
+Support for future Flash tasks is on the horizon.
+
+.. _flash-model-training:
+
+Model training
+______________
+
+One of the primary uses of Flash is to load an existing model and finetune it
```
brimoor, line 29: Suggested change (insert appropriate class name):

```
You can easily train or finetune Flash models on your
:ref:`FiftyOne datasets <using-datasets>` with just a few lines of code using
Flash's builtin :meth:`DataSource.from_fiftyone() <flash~flash.core.XXX.DataSource.from_fiftyone>`
method, which is implemented for each of the Flash tasks shown below:
```

docs/source/integrations/lightning_flash.rst
```diff
@@ -0,0 +1,446 @@
...
+        
+            # 7. Generate predictions
+            model = ObjectDetector.load_from_checkpoint(
+                "https://flash-weights.s3.amazonaws.com/object_detection_model.pt"
+            )
+            model.serializer = FiftyOneDetectionLabels()
+
+            predictions = trainer.predict(model, datamodule=datamodule)
+
+            # 7b. Flatten batched predictions
```
brimoor, line 173: nit: I'd remove `7b` and just add a comment like `# flatten batches` to the end of the itertools line, for brevity

brimoor (changes requested):

docs/source/integrations/lightning_flash.rst
ehofesmann: Duh! I knew I missed something...

brimoor (approved):

benjaminpkane:
Released

PR #1077, opened 2021-06-22
Fixes display issues for the `sample_id` patches view field

benjaminpkane:
Previously, the `sample_id` tag would not show up on samples in the grid. `None` would also be displayed in the expanded view  for the `sample_id` field.

Screenshot from 2021-06-22 07-45-26
Screenshot from 2021-06-22 07-45-17


brimoor (approved):
LGTM

PR #1050, opened 2021-05-21
Add flags to only write labels to disk when exporting and to import media from a filepath mapping

ehofesmann:
Resolves #707 

Adds an `export_media` flag to the relevant dataset exporters that allows for control over how images and videos are handled when exporting datasets via `SampleCollection.export()`. The options for `export_media` are:
* `True` (default) - copy media files to `data` directory as normal
* `False` - do not copy media files, instead write a `data.json` file containing mappings of uuids to filepaths
* `"move"` - move media fields instead of copying as normal
* `"symlink"` - create a symbolic link to media files instead of copying as normal

Also adds a `data_json` (default `False`) flag to the relevant dataset importers allowing them to read the `data.json` file written by an exporter.

Compatible datasets:
* BDDDataset
* COCODetectionDataset
* CVATImageDataset
* CVATVideoDataset
* FiftyOneDataset (doesn't support `data_json`, not necessary since it stores filepaths)
* FiftyOneImageClassificationDataset
* FiftyOneImageDetectionDataset
* GeoJSONImageDataset (doesn't support `data_json`, that info is already stored in `filepath` in labels)
* KITTIDetectionDataset
* LegacyFiftyOneDataset
* VOCDetectionDataset
* YOLODataset

This is available to most `LabeledDatasets`, except for the directory trees and TFRecords datasets which prevents me from making a blanket statement that all `LabeledDatasets` support this.

```python
import fiftyone as fo
import fiftyone.zoo as foz
import inspect

def run_test(dtype, dataset, label_field="ground_truth", label_field_class="ground_truth_class"):
    try:
        imp = dtype().get_dataset_importer_cls()
        exp = dtype().get_dataset_exporter_cls()
        imp_args = inspect.getfullargspec(imp).args 
        exp_args = inspect.getfullargspec(exp).args 
    except TypeError:
        print("Skipping %s" % dtype.__name__)
        return


    print("Testing %s" % dtype.__name__)
    data_dir = "/tmp/test_%s/" % dtype.__name__
    data_dir_2 = "/tmp/test_%s_default/" % dtype.__name__

    _label_field = label_field
    if "Classification" in dtype.__name__ or "Video" in dtype.__name__:
        _label_field = label_field_class

    dataset.export(
        export_dir=data_dir_2,
        dataset_type=dtype,
        label_field=_label_field,
    )
    
    test_imp_1 = fo.Dataset.from_dir(
        dataset_dir = data_dir_2,
        dataset_type=dtype,
    )

[...]

PR #1078, opened 2021-06-23
Adding support for customizing dataset imports and exports

### LINKED ISSUE #707: [FR] Flag to avoid copying data when exporting Dataset labels

brimoor:
Resolves #707, and deprecates #1050.

Adds a variety of new syntaxes for importing and exporting datasets that allow, among other things, independent control of the data and labels locations for many standard formats via new `data_path`, `labels_path`, and `export_media` parameters (plus additional parameters on a per-dataset basis).

The docs from `SampleCollection.export()` describe how these parameters function:

```
            export_dir (None): the directory to which to export the samples in
                format ``dataset_type``. This parameter may be omitted if you
                have provided appropriate values for the ``data_path`` and/or
                ``labels_path`` parameters.

            data_path (None): an optional parameter that enables explicit
                control over the location of the exported media for certain
                export formats. Can be any of the following:

                -   a folder name like "data" or "data/" specifying a subfolder
                    of ``export_dir`` in which to export the media
                -   an absolute directory path in which to export the media. In
                    this case, the ``export_dir`` has no effect on the location
                    of the data
                -   a filename like "data.json" specifying the filename of a
                    JSON manifest file in ``export_dir`` generated when
                    ``export_media`` is ``"manifest"``
                -   an absolute filepath specifying the location to write the
                    JSON manifest file when ``export_media`` is ``"manifest"``.
                    In this case, ``export_dir`` has no effect on the location
                    of the data

                If None, a default value of this parameter will be chosen based
                on the value of the ``export_media`` parameter. Note that this
                parameter is not applicable to certain export formats such as
                binary types like TF records

            labels_path (None): an optional parameter that enables explicit
                control over the location of the exported labels. Only
                applicable when exporting in certain labeled dataset formats.
                Can be any of the following:

                -   a type-specific folder name like "labels" or "labels/" or a
                    filename like "labels.json" or "labels.xml" specifying the
                    location in ``export_dir`` in which to export the labels
                -   an absolute directory or filepath in which to export the
                    labels. In this case, the ``export_dir`` has no effect on
                    the location of the labels

                For labeled datasets, the default value of this parameter will
                be chosen based on the export format so that the labels will be
                exported into ``export_dir``

            export_media (None): controls how to export the raw media. The
                supported values are:

                -   ``True``: copy all media files into the output directory
                -   ``False``: don't export media. This option is only useful
                    when exporting labeled datasets whose label format stores
                    sufficient information to locate the associated media
                -   ``"move"``: move all media files into the output directory
                -   ``"symlink"``: create symlinks to the media files in the
                    output directory
                -   ``"manifest"``: create a ``data.json`` in the output
                    directory that maps UUIDs used in the labels files to the
                    filepaths of the source media, rather than exporting the
                    actual media

                If None, an appropriate default value of this parameter will be
                chosen based on the value of the ``data_path`` parameter. Note
                that some dataset formats may not support certain values for
                this parameter (e.g., when exporting in binary formats such as
                TF records, "symlink" is not an option)
```

This is a lot of information, but I have endeavored to make the usage as intuitive as possible. For example:
- If you provide `export_dir`, you'll get a data + labels export with default filenames/labels names (current behavior)
- If you provide `labels_path` + `export_media=True/"move"/"symlink"`, then you'll get a data + labels export 
- If you provide `labels_path` + `data_path`, you'll get a data + labels export where the media are copied (IE `export_media=True` is chosen by default)
- If you only provide `labels_path`, then you'll get a labels-only export (IE `export_media` defaults to `False`)
- If you specify `export_media="manifest"`, then `data_path` will automatically default to `data.json` rather than `data/`
- If you provide a JSON filepath for `data_path`, then `export_media` will default to `"manifest"` 

### Other changes

- Adds new `merge_dir()` and `merge_importer()` functions that combine `from_dir() + merge_samples()` into a single method call.
- Renamed `YOLODataset` to `YOLOv4Dataset`, and added a new `YOLOv5Dataset` type for importing/exporting in [YOLOv5 format](https://github.com/ultralytics/yolov5)
- Updated `GeoJSONImageDataset` to `GeoJSONDataset` to reflect, both in name and function, that this type can handle both image and video datasets

### TODO
- Update loading datasets user guide page
- Add import/export unit tests

### Examples

The example below demonstrates labels-only imports and exports using the new syntax:

```py
import os

import fiftyone as fo
import fiftyone.zoo as foz

#
# Part 1: Labels-only exports
#

# In this workflow, no images are ever copied/moved. Everything points here
IMAGES_DIR = os.path.dirname(dataset.first().filepath)

GT_PATH = "/tmp/coco/ground_truth.json"
PREDICTIONS_PATH = "/tmp/coco/predictions.json"

# Export ground truth labels
dataset.export(
    labels_path=GT_PATH,
    label_field="ground_truth",
)

# Export predictions
dataset.export(
    labels_path=PREDICTIONS_PATH,
    label_field="predictions",
)

#
# Part 2: Labels-only imports
#

# Load ground truth
    data_path=IMAGES_DIR,
    labels_path=GT_PATH,
    label_field="ground_truth",
)

# Add predictions to dataset
dataset2.merge_dir(
    data_path=IMAGES_DIR,
    labels_path=PREDICTIONS_PATH,
    label_field="predictions",
)

print(dataset2)
print(dataset2.count("ground_truth.detections"))
print(dataset2.count("predictions.detections"))
```


benjaminpkane:
Goes without saying that this is awesome. Manifests are my favorite :rocket: :rocket: :rocket:

docs/source/user_guide/export_datasets.rst
```diff
@@ -47,13 +47,50 @@ a |DatasetView| into any format of your choice via the basic recipe below.
...
+        #
+        # Export the labels in the `ground_truth` field in COCO format, and
+        # write a `data.json` file describing the locations of the source media
+        # rather than exporting the actual media
+        #
+        dataset_or_view.export(
+            export_dir="/path/for/export",
+            dataset_type=fo.types.COCODetectionDataset,
+            label_field="ground_truth",
+            export_media=False,
```
ehofesmann, line 79: Shouldn't this be `export_media="manifest"` if you say it is writing a `data.json` file?

fiftyone/utils/yolo.py
```diff
@@ -218,24 +514,218 @@ def label_cls(self):
...
+        self._writer = YOLOAnnotationWriter()
+
+        self._parse_classes()
 
-        self._filename_maker = fou.UniqueFilenameMaker(
-            output_dir=self._data_dir,
+        self._media_exporter = foud.ImageExporter(
+            self.export_media,
+            export_path=self.data_path,
+            supported_modes=(True, False, "move", "symlink"),
```
ehofesmann, line 534: What is the reason that YOLO exporters don't provide a `manifest` option?

docs/source/user_guide/export_datasets.rst
brimoor: yep you're correct; forgot to update this documentation 😅

fiftyone/utils/yolo.py
brimoor: YOLOv4 stores the labels in the same folder as the images, and YOLOv5 does a find-replace on `/images/` -> `/labels/` to locate the labels coresponding to images, so there's really not a valid use case I could find for trying to support the "manifest" option.

There are a fair number of annoying nuances/peculiarities to consider for each dataset format...

benjaminpkane (approved):
Approving for release

fiftyone/utils/yolo.py
ehofesmann: In my mind, the "manifest" option is always useful so that you can easily export only labels and then reimport them again directly from that folder. "symlink" provides the same workflow but there may be situations where you would rather have a single manifest file than a directory of symlinks. That being said, it's definitely not quite as useful for YOLO as for other formats so if you want to leave it the way it is now, that's alright by me.

ehofesmann (approved):
Looks good to me! I'll leave the YOLO comments and unit tests up to you, but I'm happy with it. This import/export functionality is going to be pretty valuable in a lot of use cases.

fiftyone/utils/yolo.py
brimoor: Okay I see your point. Let's take this one offline

PR #1082, opened 2021-06-23
Filter classifications in the sample grid

### LINKED ISSUE #1081: [BUG] Small inconsistency in App filtering behavior for Classifications fields
There's a small inconsistency in filtering behavior for `Classifications` (plural) fields in the App.

In the example below, I search for `Accordian` in the sidebar, and I correctly see only the two samples that contain accordians, but all other labels for those two samples are still appearing, whereas for `Detections` fields, for example, applying such a filter would remove all non-Accordian boxes from the grid.

If I then click the bookmark to save the view stage, the correct behavior is then observed (only Accordian tags appear).

The desired behavior is that matching labels on any type of list field will hide non-matching `Label` instances.

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "open-images-v6",
    split="validation",
    label_types=["detections", "classifications"],
    classes=["Fedora", "Piano"],
    max_samples=25,
)

```

Screen Shot 2021-06-23 at 4 14 46 PM

Screen Shot 2021-06-23 at 4 14 57 PM

benjaminpkane:
Resolves #1081

### DIFF
--- a/app/src/components/Sample.tsx
+++ b/app/src/components/Sample.tsx
@@ -128,6 +128,7 @@ const SampleInfo = React.memo(({ id }) => {
   const colorByLabel = useRecoilValue(atoms.colorByLabel(false));
   const labelTypes = useRecoilValue(selectors.labelTypesMap);
   const sample = useRecoilValue(atoms.sample(id));
+  const filters = useRecoilValue(labelFilters(false));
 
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM

PR #1083, opened 2021-06-24
COCO Evaluation Updates

ehofesmann:
After testing models on the quickstart and entire validation set, COCO evaluation seems to have been correct the entire time and matches the COCO API implementation. The only issue that I came across when comparing evaluations between FiftyOne and COCO was the way that values were being exported.

When exporting COCO labels, the image and annotation IDs need to start from one in order to be compatible with COCO API functionality since it expects 1-indexed IDs. For example, when evaluating with the COCO API and computing matches, the matched box IDs are stored in a Numpy array initialized to zeros.

Additionally, rounding bounding box coordinates was able to noticeably affect results when the IoU of a box was very close to a threshold. This showed up even just in the 200 sample quickstart dataset meaning it is likely fairly common.

**Question**: Are there any reasons that we can't just 1-index COCO ids and remove rounding? If we don't want to remove rounding entirely, we should at least round further than 1 decimal place.


fiftyone/utils/coco.py
```diff
@@ -511,8 +511,10 @@ def label_cls(self):
         return fol.Detections
 
     def setup(self):
-        self._image_id = -1
-        self._anno_id = -1
+        self._data_dir = os.path.join(self.export_dir, "data")
```
brimoor, line 514: I believe `_data_dir` and `_labels_path` are leftover from a merge conflict and need to be removed

fiftyone/utils/coco.py
```diff
@@ -511,8 +511,10 @@ def label_cls(self):
...
+        self._labels_path = os.path.join(self.export_dir, "labels.json")
+        self._image_id = 0
+        self._anno_id = 0
```
brimoor, line 517: 1-based is cool with me

brimoor (approved):

PR #1084, opened 2021-06-25
Use extended view for string field search results

### LINKED ISSUE #1065: [FR] Choose filter widget style based on view, not dataset
It would be nice to update the logic in the filter sidebar to choose whether to use search-style vs explicit-checkbox-list style widgets based on the number of distinct values in the **view** rather than in the **dataset**.

In the example below, I created a view that reduced the number of distinct sample IDs from 1000+ to 10, but the `id` filter is still the search-style option, while it would be more usable if it switched to checkbox-style since the filterable set is now smaller.

Screen Shot 2021-06-09 at 4 47 35 PM

benjaminpkane:
Resolves #1065


benjaminpkane:
Closing, see #1065

PR #1085, opened 2021-06-25
Adding select_by() and exclude_by() stages

brimoor:
Adding variants of `Select()` and `Exclude()` that allow for selecting/excluding by categorial fields other than just `id`.

These are useful, for example, when your dataset contains an "external" ID field like `coco_id` (int) or `open_images_id` (str) and you'd like to add predictions to your dataset based on these external IDs.

```py
import fiftyone as fo

dataset = fo.Dataset()
dataset.add_samples(
    [
        fo.Sample(filepath="image%d.jpg" % i, int=i, str=str(i))
        for i in range(100)
    ]
)

#
# Create a view containing samples whose `int` field have the given
# values
#

view = dataset.select_by("int", [1, 51, 11, 41, 21, 31])
print(view.head(6))

#
# Create a view containing samples whose `str` field have the given
# values, in order
#

view = dataset.select_by(
    "str", ["1", "51", "11", "41", "21", "31"], ordered=True
)
print(view.head(6))
```

### DIFF
--- a/fiftyone/__public__.py
+++ b/fiftyone/__public__.py
@@ -124,6 +124,7 @@ from .core.plots import (
 from .core.sample import Sample
 from .core.stages import (
     Exclude,
+    ExcludeBy,
     ExcludeFields,
     ExcludeFrames,
     ExcludeLabels,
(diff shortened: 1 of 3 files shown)


benjaminpkane:
LGTM :+1:

benjaminpkane (approved):
LGTM :+1:

PR #1086, opened 2021-06-26
Optimizing implementation of batch aggregations

brimoor:
A common pattern in various places in the codebase are batch aggregations like this:

```py
list, of, results = dataset.values(["list", "of", "fields"])

list, of, results = dataset.aggregate([fo.Values("list"), fo.Values("of"), fo.Values("fields")])
```

For aggregations **other than** `Values()`, such operations are already quite efficient, as they are executed via `$facet` with a single pass over the documents in the DB.

For `Values()`, `$facet` is not an option because it returns results in a single document, which could easily violate the 16MB limit. Thus, currently, if an aggregation is executed that contains multiple `Values()` instances, those aggregations are run in series, each requiring a separate pass over the DB.

For `Values()` operations that involve unwinding of fields, I don't see any obvious ways to avoid these multiple passes. However, when multiple `Values()` calls with no unwinding are executed as a batch, they could be executed in a single pass over the database by merging their `$project` stages into a single stage. This PR implements this merging + subsequent unpacking of the aggregation results into separate lists.

The benchmarks below show that this change can 2x+ the performance of aggregations that involve multiple `Values()` invocations.

```py
import eta.core.utils as etau
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("bdd100k", split="validation")

#
# before and after: uses facets (but would not scale >16MB)
# Time elapsed: 1.1 seconds
#

with etau.Timer():
    dataset.values(
        [
            "id",
            "filepath",
            "ground_truth_detections.detections.label",
            "ground_truth_detections.detections.confidence",
            "ground_truth_polylines.polylines.label",
            "ground_truth_polylines.polylines.confidence",
            "ground_truth_polylines.polylines.filled",
            "ground_truth_polylines.polylines.closed",
        ],
        _big_result=False,  # forces $facet to be used
    )

#
# before: runs aggregations in series
# Time elapsed: 1.6 seconds
#
# after: single aggregation pass
# Time elapsed: 709.9ms
#

with etau.Timer():
    dataset.values(
        [
            "id",
            "filepath",
            "ground_truth_detections.detections.label",
            "ground_truth_detections.detections.confidence",
            "ground_truth_polylines.polylines.label",
            "ground_truth_polylines.polylines.confidence",
            "ground_truth_polylines.polylines.filled",
            "ground_truth_polylines.polylines.closed",
        ]
    )
```

### DIFF
--- a/docs/source/user_guide/using_aggregations.rst
+++ b/docs/source/user_guide/using_aggregations.rst
@@ -382,12 +382,22 @@ The example below demonstrates this capability:
         ]
     )
 
+    #
     # Count the number of keypoints in the dataset
-    count = dataset.count("keypoints.points[]")
+    #
+    # The `points` list attribute is declared on the `Keypoint` class, so it is
+    # automatically unwound
+    #
+    count = dataset.count("keypoints.points")
     print(count)
     # 5
 
-    # Compute the values in the custom `friends` field of the predictions
+    #
+    # Compute the values in the `friends` field of the predictions
+    #
+    # The `friends` list attribute is a dynamic custom attribute, so we must
+    # explicitly request that it be unwound
+    #
     counts = dataset.count_values("classes.friends[]")
     print(counts)
     # {'dog': 1, 'squirrel': 2, 'rabbit': 1}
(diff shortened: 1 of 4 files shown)


benjaminpkane (approved):
LGTM!

PR #1091, opened 2021-06-30
Disconnect timeout before closing waited sessions

### LINKED ISSUE #1090: [BUG] Refreshing browser tab exits session.wait()
Refreshing the App window in the browser causes `session.wait()` to exit, but the intention is that `session.wait()` should only exit when the App window is actually closed.

This is worth addressing because refreshing the App window via ctrl + R is a common practice for App users.

```py
import fiftyone as fo 

session = fo.launch_app()

# should block until App is closed, but ctrl + R also breaks
session.wait()  
```

benjaminpkane:
Adds a one second timeout before closing waited sessions (`session.wait()`). Resolves premature closes due to page refreshes.

Resolves #1090

### DIFF
--- a/fiftyone/server/main.py
+++ b/fiftyone/server/main.py
@@ -206,7 +206,7 @@ def _catch_errors(func):
 
 _notebook_clients = {}
 _deactivated_clients = set()
-_DISCONNECT_TIMEOUT = 2  # seconds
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM

Issue #1095, opened 2021-07-01
[BUG] Do not run downgrades automatically


benjaminpkane:
Upgrades and downgrades currently happen automatically. For downgrades, the issue with this is that the current version of `fiftyone` does not have the full (future) revision history from which it is downgrading from.

This leads to datasets (or the database) incorrectly being labeled as successfully downgraded, and ultimately more confusion when working through database and dataset migration/version issues.

Related #1094

Issue #1094, opened 2021-07-01
[BUG] The fields "{'db_field'}" do not exist on the document "SampleFieldDocument"


luke-iqt:
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 20.04
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: fiftyone v0.9.3, Voxel51, Inc.
-   **Python version**: Python 3.6.9

### Describe the problem

I have just began encountering the following error when I try to load a Voxel51 dataset. The dataset had been migrated to a newer version of Voxel a while ago, but had worked fine after that migration. This problem started after the instance had been restarted. Is it possible the underlying MongoDB has been corrupted? 

### Code to reproduce issue

```
import fiftyone as fo
dataset_name = "jsm-test-dataset" # Name of the Voxel51 Dataset to use
dataset = fo.load_dataset(dataset_name) 
```

Log:

```
---------------------------------------------------------------------------
FieldDoesNotExist                         Traceback (most recent call last)
<ipython-input-2-bce0580c7d7e> in <module>
      3 import os
      4 
----> 5 dataset = fo.load_dataset(dataset_name)

/usr/local/lib/python3.6/dist-packages/fiftyone/core/dataset.py in load_dataset(name)
     97         ValueError: if no dataset exists with the given name
     98     """
---> 99     return Dataset(name, _create=False)
    100 
    101 

/usr/local/lib/python3.6/dist-packages/fiftyone/core/singletons.py in __call__(cls, name, _create, *args, **kwargs)
     31         ):
     32             instance = cls.__new__(cls)
---> 33             instance.__init__(name=name, _create=_create, *args, **kwargs)
     34             name = instance.name  # `__init__` may have changed `name`
     35             cls._instances[name] = instance

/usr/local/lib/python3.6/dist-packages/fiftyone/core/dataset.py in __init__(self, name, persistent, overwrite, _create, _migrate, _patches)
    239                 self._sample_doc_cls,
    240                 self._frame_doc_cls,
--> 241             ) = _load_dataset(name, migrate=_migrate)
    242 
    243         self._evaluation_cache = {}

/usr/local/lib/python3.6/dist-packages/fiftyone/core/dataset.py in _load_dataset(name, migrate)
   3465     try:
   3466         # pylint: disable=no-member
[...]

PR #1098, opened 2021-07-02
Don't allow migrations involving future FiftyOne versions

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1094
Resolves https://github.com/voxel51/fiftyone/issues/1095

Running a migration that involves a future version of the FiftyOne package should not be allowed, since the currently installed version of `fiftyone` is not aware of any upward/downward migrations that may be introduced in future versions of the package.

**For discussion:**
Although this is technically correct, I wonder if there is some middle ground we can achieve here, since this will now *always* raise an error whenever anyone tries to downgrade their installation without following the docs instructions, while many times no migration would be required...

This can be tested as follows:

```shell
# manually increase version in `setup.py` to a future version, say 0.11.2
pip install -e .

# undo `setup.py` change
pip install -e .
fiftyone datasets info quickstart  # raises error shown below
```

```
OSError: You must have fiftyone>=0.11.2 installed in order to migrate from v0.11.2 to v0.11.1, but you are currently running fiftyone==0.11.1.

See https://voxel51.com/docs/fiftyone/getting_started/install.html#downgrading-fiftyone for information about downgrading FiftyOne.
```

Updated documentation:

Screen Shot 2021-07-02 at 2 59 45 PM

### DIFF
--- a/docs/source/getting_started/install.rst
+++ b/docs/source/getting_started/install.rst
@@ -232,9 +232,9 @@ Here's the workflow for downgrading to an older version of FiftyOne:
     pip install fiftyone==$VERSION
 
 If you are reading this after encountering an error resulting from downgrading
-your ``fiftyone`` package without first running ``fiftyone migrate``, don't
-worry, you simply need to reinstall the newer version of FiftyOne and then
-follow these instructions.
+your ``fiftyone`` package without first running
+:ref:`fiftyone migrate <cli-fiftyone-migrate>`, don't worry, you simply need to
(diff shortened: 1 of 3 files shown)


benjaminpkane (approved):
LGTM. I would be fine with only logging a warning that their dataset has a future version and may need proper downgrade. As long as the dataset retains its future version.

benjaminpkane:
I defer to you

brimoor:
> I defer to you

Okay let's go with this conservative approach for now 🚀

PR #1104, opened 2021-07-07
Fixed aws error 404 bug on windows

Pascal0902:
At the moment, an error 404 is raised when calling "_download_images_if_necessary" on windows, since windows style paths are used when downloading from the s3 storage. This changes should prevent this bug. For details, see stackoverflow: https://stackoverflow.com/questions/68229288/fail-to-load-a-subpart-of-open-images-v6-with-fiftyone/

Added a regex to convert windows stype paths to linux for s3 downloads.

Tested on windows 10 with python 3.9. Changes should not have any effect on other os.

        notes for FiftyOne users.

-   [X] Other

### DIFF
--- a/fiftyone/utils/openimages.py
+++ b/fiftyone/utils/openimages.py
@@ -14,6 +14,7 @@ import multiprocessing
 import os
 import random
 import warnings
+import re
 
 import boto3
 import botocore
@@ -1628,7 +1629,7 @@ def _download_images_if_necessary(
     num_existing = 0
     for image_id in image_ids:
         fp = os.path.join(data_dir, image_id + ".jpg")
-        fp_download = os.path.join(split, image_id + ".jpg")
+        fp_download = re.sub(r"\\", "/", os.path.join(split, image_id + ".jpg"))
         if not os.path.isfile(fp):
             inputs.append((fp, fp_download))
         else:


fiftyone/utils/openimages.py
```diff
@@ -1628,7 +1629,7 @@ def _download_images_if_necessary(
     num_existing = 0
     for image_id in image_ids:
         fp = os.path.join(data_dir, image_id + ".jpg")
-        fp_download = os.path.join(split, image_id + ".jpg")
+        fp_download = re.sub(r"\\", "/", os.path.join(split, image_id + ".jpg"))
```
benjaminpkane, line 1632:
```suggestion
        fp_download = os.path.normpath(os.path.join(split, image_id + ".jpg"))
```

fiftyone/utils/openimages.py
```diff
@@ -14,6 +14,7 @@
 import os
 import random
 import warnings
+import re
```
benjaminpkane, line 17:
```suggestion
```

benjaminpkane (changes requested):
Hi @Pascal0902. Good catch. Thanks for the patch!

I think `os.path.normpath()` will accomplish the same thing? If that works for you (I'm not using Windows), we can get this merged.

benjaminpkane (approved):
I'm sorry, you're converting **from** a windows path. LGTM. Thank!

PR #1105, opened 2021-07-07
Handling unlabeled datasets in COCODetectionDatasetImporter

brimoor:
Previously, providing a `classes` requirement to `COCODetectionDatasetImporter` when loading an unlabeled dataset would raise an error. Now, the `classes` will be ignored in this case and a warning to this effect printed.

This change allows the code below to successfully run (with a warning), which is a better behavior than an unhelpful error message:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# warns that classes requirement must be ignored for test split, which is unlabeled
    "coco-2017",
    splits=["train", "test"],
    classes=["person", "dog", "car", "clock", "horse"],
    only_matching=True,
    include_id=True,
    max_samples=100,
)

print(dataset)
```

### DIFF
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -382,7 +382,15 @@ class COCODetectionDatasetImporter(
         image_metadata = fom.ImageMetadata(width=width, height=height)
 
         if self._annotations is None:
-            return image_path, image_metadata, None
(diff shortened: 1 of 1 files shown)


ehofesmann (approved):
LGTM, thanks for the quick turnaround!

PR #1109, opened 2021-07-08
Add progress flag to track iter_samples() progress

brimoor:
Adds an optional `progress` flag to `SampleCollection.iter_samples()` that provides a convenient way to render a progress bar tracking the progress of an operation:

```py
import fiftyone.zoo as foz

    pass

for sample in dataset.take(100).iter_samples(progress=True):
    pass
```


benjaminpkane (approved):
LGTM!

PR #1111, opened 2021-07-08
`Session.open_tab()` in notebooks and remote warning

### LINKED ISSUE #1102: [BUG] session.open_tab() hangs in remote Jupyter notebooks
When working in a [remote Jupyter notebook](https://voxel51.com/docs/fiftyone/faq/index.html#can-i-use-fiftyone-in-a-remote-notebook), `session.open_tab()` hangs indefinitely.

I don't think it's possible to open a browser tab in the remote notebook workflow, so the desired behavior would be to either raise an error or log a message explaining how to view the App instance.

More generally, there may be other methods that need remote-notebook detection (`_remote` only handles App instances created with the `remote=True` flag).

benjaminpkane:
Resolves #1102

Allows notebooks sessions to open a tab, and changes pure remote sessions to warn (can revert, but I'm thinking this is more friendly). Comments/changes in the language are welcome.

Tested in Colab.

### DIFF
--- a/fiftyone/core/session.py
+++ b/fiftyone/core/session.py
@@ -20,10 +20,7 @@ import fiftyone.core.dataset as fod
 import fiftyone.core.client as foc
 from fiftyone.core.config import AppConfig
 import fiftyone.core.context as focx
-import fiftyone.core.frame as fof
-import fiftyone.core.media as fom
 import fiftyone.core.plots as fop
-import fiftyone.core.sample as fosa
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM. Tested in a remote notebook and it works beautifully!

`open_tab()` is a really nice way to work in notebooks 💪

brimoor:
I followed your lead and converted a couple other errors to warnings.

PR #1112, opened 2021-07-08
Improvements to Open Images partial downloads

brimoor:
This PR makes some improvements to how `classes` and `attrs` requirements are handled for partial downloads of the Open Images dataset.

### Change log

The changes are as follows:
- "classifications" are excluded from `classes` requirements, since they only describe the label schema and do not actually indicate the presence of that label type. The exception to this is that, if **only** "classifications" labels are requested, then the `classes` requirement **will** be applied
- `classes` requirements were previously being AND-ed across all label types that support classes ("detections" and "segmentations"), but the more natural behavior is to OR across label types. In other words, if you want samples containing particular classes, they can appear across any label type you requested, not necessarily ALL label types. (When `max_samples` is specified, priority will still be given to samples that really do contain the classes in each label type)

In addition, pretty thorough unit tests are now included in `tests/intensive/dataset_zoo_tests.py` to ensure that the expected behaviors are occurring when partially downloading COCO and Open Images.

### Editorial comment

These partial download implementations have turned out to be quite annoying, as its actually not so straightforward to provide an intuitive behavior across all of the partial download parameters that we support... With the data loaded into FiftyOne, querying for exactly what you want becomes much easier 😃 

### Example usage

This PR was motivated by the example below. It so happens that `Violin` is only used for detections, not segmentations, but, nonetheless, I would expect the code below to load 25 samples with violins. Currently, it returns an empty dataset because it is requiring "Violin" to be present as both a detection and segmentation label.

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "open-images-v6",
    split="validation",
    classes="Violin",
    max_samples=25,
)

# currently fail but now succeed
assert len(dataset) == 25
assert len(dataset.match_labels(fields="detections", filter=F("label") == "Violin")) == 25
```

### DIFF
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -324,10 +324,9 @@ class COCODetectionDatasetImporter(
             default="labels.json",
         )
 
-        label_types = _parse_label_types(label_types)
-
+        _label_types = _parse_label_types(label_types)
         if include_id:
-            label_types.append("coco_id")
+            _label_types.append("coco_id")
 
         super().__init__(
             dataset_dir=dataset_dir,
@@ -346,6 +345,7 @@ class COCODetectionDatasetImporter(
         self.use_polylines = use_polylines
         self.tolerance = tolerance
 
+        self._label_types = _label_types
         self._info = None
         self._classes = None
         self._supercategory_map = None
@@ -394,7 +394,7 @@ class COCODetectionDatasetImporter(
 
         label = {}
 
-        if "detections" in self.label_types:
+        if "detections" in self._label_types:
             detections = _coco_objects_to_detections(
                 coco_objects,
                 frame_size,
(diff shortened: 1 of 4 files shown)


ehofesmann (approved):
LGTM

I was a little hesitant about not including `classifications` when specifying classes since it could be useful if you provide a class that doesn't exist explicitly in other label types but is a parent in the hierarchy of labels that do exist. However, I don't like that it used to pick up samples where the `classes` you provide appear in the `negative_labels`.

brimoor:
> I was a little hesitant about not including `classifications` when specifying classes since it could be useful if you provide a class that doesn't exist explicitly in other label types but is a parent in the hierarchy of labels that do exist. However, I don't like that it used to pick up samples where the `classes` you provide appear in the `negative_labels`.

Yeah I see what you mean, but it'll be much more natural to work with hierarchies with the data fully in FiftyOne.

Issue #1101, opened 2021-07-06
[FR] polyline(poly2d) evaluation in 'evaluate_detections'


lukelin-linkernetworks:
### Proposal Summary

the 'evaluate_detections' function does not offer polyline (poly2d) evaluation at this moment.

### Motivation

-   What is the use case for this feature?
     When I have two datasets with 'ground_truth' and 'predictions', but both contained poly2d annotations.

-   Why is this use case valuable to support for FiftyOne users in general?
     Because there are the situations that objects are not annotated in bounding box but more accurate in polylines.

-   Why is it currently difficult to achieve this use case?  
    The current 'evaluate_detections' only supports classification, detection, and segmentation, and only bbox can be apply to detection.

### What areas of FiftyOne does this feature affect?

-   [ ] App: FiftyOne application
-   [x] Core: Core `fiftyone` Python library
-   [ ] Server: FiftyOne server

### Details

There is a library called "shapely" which has "poly_iou".

### Willingness to contribute

The FiftyOne Community encourages new feature contributions. Would you or
another member of your organization be willing to contribute an implementation
of this feature?

-   [ ] Yes. I can contribute this feature independently.
-   [ ] Yes. I would be willing to contribute this feature with guidance from
        the FiftyOne community.
-   [ ] No. I cannot contribute this feature at this time.

brimoor:
Hey @lukelin-linkernetworks 👋 check https://github.com/voxel51/fiftyone/pull/1114 out!

You can do a source install of that branch as follows:

```shell
git clone https://github.com/voxel51/fiftyone
cd fiftyone
git checkout --track origin/poly-eval
bash install.bash
```

I'd be interested to hear what you think of the implementation there if you have a chance to try it out 😄

PR #1114, opened 2021-07-09
Adds support for evaluating polygons and instance segmentations

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1101.

### Description

The `evaluate_detections()` method now supports evaluating `Polylines` and `Detections` with instance masks. Both COCO and Open Images flavored evaluations are supported, as the only difference for these new cases is how IoU is computed.

The user will be prompted to install [Shapely](https://github.com/Toblerity/Shapely) the first time they run an evaluation that requires polgyon computations. Thanks to the awesome Shapely library, IoU computation on polygons should be quite robust, as things like self-overlapping and self-intersecting shapes will be appropriately handled.

**Note:** I decided against including `shapely` as a builtin requirement because, while it is cross-platform and stable, is does rely on compiled C++ code and thus some users may want/need to tweak their install slightly beyond what `pip install shapely` provides. Also, polygon evaluation is a fairly niche thing.

**Note:** Currently, instance segmentations are first converted to polygons with a default tolerance of 2 pixels and then IoU calculations are performed on those. If users are interested in computing IoUs directly from the array masks, then we could look into it, but the implementation would take a bit more work that I didn't feel like doing just yet.

### Misc

One small additional change added here is the removal of the `config` parameter from `compute_*()`, which was intended to offer an alternative to specifying `method` + `**kwargs` to configure the evaluation method.

This change was motivated by realizing that the configs hold the names of the GT and predicted fields, but the predicted field is also a mandatory positional argument for the `compute_*()` methods themselves, which is confusing. Also, removing the `config` option avoids the potential confusion resulting from multiple syntaxes to achieve the same thing...

### Example usage

```py
import random
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

# Create dataset

def jitter(labels, delta=0.03):
    if isinstance(labels, fo.Detections):
        for detection in labels.detections:
            alpha = random.random()
            bounding_box = np.array(detection.bounding_box)
            bounding_box += alpha * delta * np.array([1, 1, 0, 0])
            detection.bounding_box = bounding_box.tolist()
            detection.confidence = alpha

    if isinstance(labels, fo.Polylines):
        for polyline in labels.polylines:
            alpha = random.random()
            points = []
            for _points in polyline.points:
                _points = np.array(_points)
                _points += alpha * delta * np.ones(_points.shape)
                points.append(_points.tolist())

            polyline.points = points
            polyline.confidence = alpha

    return labels

    "coco-2017",
    split="validation",
    label_types=["detections", "segmentations"],
    max_samples=50,
)

# Add polylines
with fo.ProgressBar() as pb:
    for sample in pb(dataset):
        sample["polylines"] = sample["segmentations"].to_polylines()
        sample.save()

# Add mock predictions
with fo.ProgressBar() as pb:
    for sample in pb(dataset):
        sample["pred_det"] = jitter(sample["detections"].copy())
        sample["pred_seg"] = jitter(sample["segmentations"].copy())
        sample["pred_pol"] = jitter(sample["polylines"].copy())
        sample.save()

print(dataset)

# Detection evaluation
results1 = dataset.evaluate_detections(
    "pred_det",
    gt_field="detections",
    eval_key="det_coco",
)
print(results1.metrics())

# Instance segmentation evaluation
results2 = dataset.evaluate_detections(
    "pred_seg",
    gt_field="segmentations",
    eval_key="seg_coco",
    use_masks=True,
)
print(results2.metrics())

# Polygon evaluation
results3 = dataset.evaluate_detections(
    "pred_pol",
    gt_field="polylines",
    eval_key="pol_coco",
)
print(results3.metrics())

# Polygon evaluation but use bounding boxes for IoU calculations
results4 = dataset.evaluate_detections(
    "pred_pol",
    gt_field="polylines",
    eval_key="pol_coco",
    use_boxes=True,
)
print(results4.metrics())
```

### Documentation

Screen Shot 2021-07-09 at 1 55 48 AM

### DIFF
--- a/docs/source/integrations/coco.rst
+++ b/docs/source/integrations/coco.rst
@@ -335,7 +335,7 @@ You can also explicitly request that COCO-style evaluation be used by setting
 the ``method`` parameter to ``"coco"``.
 
 See :ref:`this page <evaluating-detections>` for more information about using
-FiftyOne to analyze detection models.
+FiftyOne to analyze object detection models.
 
 .. note::
 
@@ -373,13 +373,13 @@ populated on each sample and its predicted/ground truth objects:
         FP: sample.<eval_key>_fp
         FN: sample.<eval_key>_fn
 
--   The fields listed below are populated on each individual |Detection|
-    instance; these fields tabulate the TP/FP/FN status of the object, the ID
-    of the matching object (if any), and the matching IoU::
+-   The fields listed below are populated on each individual object instance;
+    these fields tabulate the TP/FP/FN status of the object, the ID of the
+    matching object (if any), and the matching IoU::
 
-        TP/FP/FN: detection.<eval_key>
-              ID: detection.<eval_key>_id
-             IoU: detection.<eval_key>_iou
+        TP/FP/FN: object.<eval_key>
+              ID: object.<eval_key>_id
+             IoU: object.<eval_key>_iou
 
 .. note::
 
@@ -404,7 +404,7 @@ The example below demonstrates COCO-style detection evaluation on the
     dataset = foz.load_zoo_dataset("quickstart")
     print(dataset)
 
-    # Evaluate the detections in the `predictions` field with respect to the
+    # Evaluate the objects in the `predictions` field with respect to the
     # objects in the `ground_truth` field
     results = dataset.evaluate_detections(
         "predictions",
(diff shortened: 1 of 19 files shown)


ehofesmann (approved):
LGTM! Good call updating the docs references of `detection` to `object`. I'm curious how close this gets to the COCO API evaluation, but matching those numbers exactly isn't crucial for this.

PR #1115, opened 2021-07-09
Adding support for gracefully handling evaluations/brain methods from future FiftyOne versions

brimoor:
Previously, if a user downgraded from a future version of FiftyOne, a dataset could have evaluations and brain methods with new parameters or completely new methods altogether. In such cases, calling methods like `dataset.get_evaluation_info()` and `dataset.delete_evaluation()` after downgrade would result in uncaught exceptions as the old version of FiftyOne does not know how to handle these future concepts.

This PR does not offer any magic for backporting future methods, but it does make the handling of future evaluations/brain methods more graceful in the following ways:
- `RunConfig` will now gracefully ignore new parameters rather than raise an error. This enables configs from future versions where additional parameters were added to an evaluation/brain method to be loaded without error in older versions of FiftyOne
- Added a more helpful error message when `load_evaluation_info()` and `load_brain_info()` fail due to version-related issues suggesting that the user re-run the method on their current FiftyOne version
- Perhaps most helpfully, ensures that `delete_evaluation()` and `delete_brain_run()` will successfully remove record of the run from the DB even if a version mismatch prevents us from running the `cleanup()` method. This paves the way for the user to regenerate their results if desired
- Added a warning to the `down` migration for v0.11.2 that warns that it might not be possible to work with results from future runs in older versions of FiftyOne (this has always been true, but now we're actually warning the user about this)

Some utility methods for deleting inaccessible evaluations/brain methods have also been added to `fiftyone.core.odm`.

Example usage:

```shell
# I bumped the version in `setup.py` to v0.11.2 in this branch
pip install -e .
```

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.persistent = True

dataset.evaluate_detections(
    "predictions",
    gt_field="ground_truth",
    eval_key="eval",
)
```

```shell
fiftyone migrate --all -v 0.11.1
# note warnings
```


fiftyone/core/collections.py
```diff
@@ -1887,7 +1888,10 @@ def _get_visualization_keys(self, **kwargs):
     def _get_brain_runs_with_type(self, run_type, **kwargs):
         brain_keys = []
         for brain_key in self.list_brain_runs():
-            brain_info = self.get_brain_info(brain_key)
+            try:
+                brain_info = self.get_brain_info(brain_key)
+            except:
+                continue
```
benjaminpkane, line 1894: Why hide these exceptions? A comment might be helpful.

benjaminpkane (approved):
LGTM! Well thought out.

fiftyone/core/collections.py
brimoor: Yeah good point. I didn't want to raise the error because I don't want one unusable brain result to affect the ability to affect other ones. But, it wouldn't hurt to warn the user that something was not usable, even if they continually see it until they finally delete that brain result from their dataset (or upgrade, etc)

Issue #1108, opened 2021-07-08
[FR] Add 'license' field to COCO2014/2017 dataset view


Yosshi999:
### Proposal Summary
COCO dataset has a list of licenses in the 'licenses' field and each entry in the 'images' field has an index of the license list.
It will be useful when we can filter this dataset by licenses like:
```
dataset.match(ViewField("metadata.license").is_in(
    ("Attribution License", "Attribution-ShareAlike License", "United States Government Work")
))
```

### Motivation

-   Filtering out the images whose license is not suitable for the user's purpose.

### What areas of FiftyOne does this feature affect?

-   [ ] App: FiftyOne application
-   [x] Core: Core `fiftyone` Python library
-   [ ] Server: FiftyOne server

### Details
I think just adding 'license' field to [ImageMetadata](https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/metadata.py#L55) is simplest.
It seems there is a license field in [export_sample()](https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/coco.py#L711) but I'm not sure when this code is executed.

### Willingness to contribute

The FiftyOne Community encourages new feature contributions. Would you or
another member of your organization be willing to contribute an implementation
of this feature?

-   [ ] Yes. I can contribute this feature independently.
-   [x] Yes. I would be willing to contribute this feature with guidance from
        the FiftyOne community.
-   [ ] No. I cannot contribute this feature at this time.

brimoor:
Hi @Yosshi999 👋 this is a great idea! I added support fir this in https://github.com/voxel51/fiftyone/pull/1116 😄 

You can do a source install and try it out like so:

```shell
git clone https://github.com/voxel51/fiftyone
cd fiftyone
git checkout --track origin/coco-license
bash install.bash
```

PR #1116, opened 2021-07-10
Adding support for loading COCO license data

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1108.

`COCODetectionDatasetImporter` now provides an optional `include_license` flag that will load the license data for the images, if available.

```
include_license (False): whether to include the license ID of each
    sample in the loaded labels, if available. Supported values are:

    -   ``"False"``: don't load the license
    -   ``True``/``"name"``: store the string license name
    -   ``"id"``: store the integer license ID
    -   ``"url"``: store the license URL

    Note that the license descriptions (if available) are always loaded
    into ``dataset.info["licenses"]`` and can be used to convert
    between ID, name, and URL later
```

Example usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "coco-2017",
    split="validation",
    classes="person",
    max_samples=50,
    include_license=True,
)

print(dataset.info["licenses"])
print(dataset.count_values("license"))

session.view = dataset.match(
    F("license").is_in(
        [
            "Attribution License",
            "Attribution-ShareAlike License",
        ]
    )
)

session.view = dataset.match(F("license").contains_str("NoDerivs"))
```

It's also quite convenient to use the filter sidebar in the App to search for licenses of interest:

Screen Shot 2021-07-10 at 10 34 16 AM

### DIFF
--- a/docs/source/integrations/coco.rst
+++ b/docs/source/integrations/coco.rst
@@ -133,6 +133,14 @@ COCO-2014 and COCO-2017 by passing them to
 -   ``include_id``: whether to include the COCO ID of each sample in the loaded
     labels. By default, this is False
 
+-   ``include_license``: whether to include the COCO license of each sample in
+    the loaded labels, if available. The supported values are:
+
+    -   ``"False"`` (default): don't load the license
+    -   ``True``/``"id"``: store the integer license ID
+    -   ``"name"``: store the string license name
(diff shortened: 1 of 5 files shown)


ehofesmann (approved):
LGTM!

PR #1121, opened 2021-07-12
[WIP] Optimizing frame access for video collections

brimoor:
> **WIP**: This would be nice, but it doesn't quite work yet because the `DatasetView` may contain stages like `Skip()`. To properly implement this, I think we'd need to add a `ViewStage.to_frames_mongo()` method that would allow views to define their frame-specific operations, decoupled from any sample-level filtering that may be happening.

Optimizes methods of the `fiftyone.core.frame.FramesView` class such as `keys()` and `__getitem__()` that retrieve information about the contents of frame views for a particular sample in a *view into* a video collection.

Previously, these methods would naively run the entire view's aggregation pipeline and then `select(sample.id)` to access the information for the sample of interest. Now, the sample is first selected (since, by construction, we know that the sample must belong in the view, since the user has their hands on a `FramesView` instance for that sample) and then the relevant pipeline transformations are performed.

This is much more efficient as, now, all aggregation pipelines invoked by `FramesView` methods now only select and process a single sample's worth of frames.

Example usage:

```py
import eta.core.utils as etau

import fiftyone as fo
import fiftyone.zoo as foz

# Load a video dataset
dataset.compute_metadata()

# Create a reasonably complex view
bbox_area = (
    F("bounding_box")[2] * F("$metadata.frame_height")
    * F("bounding_box")[3] * F("$metadata.frame_width")
)
view = dataset.filter_labels(
    "frames.ground_truth",
    (F("label") == "vehicle") & (bbox_area > 200 ** 2)
)

# `fiftyone==0.11.1`: 20.9 seconds
# this PR: 3.2 seconds
with etau.Timer():
    for sample in view:
        for frame_number in sample.frames:
            frame = sample.frames[frame_number]
```

### DIFF
--- a/fiftyone/core/frame.py
+++ b/fiftyone/core/frame.py
@@ -701,8 +701,10 @@ class FramesView(Frames):
         return list(self._view.get_frame_field_schema().keys())
 
     @property
-    def _frames_view(self):
-        return self._view.select(self._sample.id)
+    def _frames_pipeline(self):
+        return [{"$match": {"_id": self._sample._id}}] + self._view._pipeline(
+            frames_only=True
+        )
 
     def add_frame(self, frame_number, frame, expand_schema=True):
         """Adds the frame to this instance.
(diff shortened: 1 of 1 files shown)


brimoor:
https://github.com/voxel51/fiftyone/pull/1244 brought this to the finish line.

PR #1122, opened 2021-07-13
Catch exception when trying to patch plotly on read-only FS

SNeugber:

Patching plotly fails when running in a read-only filesystem (e.g. inside [Singularity](https://sylabs.io/guides/3.8/user-guide/)).
This just catches the exception and logs it. 

Tested by running fiftyone in Singularity and verifying that it doesn't crash the process anymore.

        notes for FiftyOne users.

-   [ ] Other


brimoor (approved):
LGTM. Thanks for the patch!

PR #1125, opened 2021-07-13
Dataset import/export enhancements

brimoor:
This PR includes a variety of enhancements related to importing and exporting datasets, including:
- Adding support for recursively listing data directories when importing datasets (see below for details)
- Removed all usage of the `attributes` dict on label types such as `Detection`. The feature still exists in the data model, but all dataset formats that previously used `attributes` now simply store this information as dynamic properties on the `Label` instances
- Added a new `Label.iter_attributes()` method that allows for iterating over the dynamic attributes of a `Label` instance (also including any attributes that happen to be stored within the now-mostly-deprecated `attributes` property, for backwards compatibility reasons)
- Adding general support for controlling whether/which object attributes are imported/exported in formats that support arbitrary object attributes (COCO, VOC, BDD, FiftyOne types, etc)
- Adds unit tests for importing and exporting in all standard formats, including `export_dir` syntax, `data_path`/`labels_path` syntax, and dynamic attributes
- Improves the import/export documentation, including:
    - Emphasizing that a simple Python for loop is often the best approach to loading data into FiftyOne
    - De-emphasizing the use of `SampleParser`s when importing custom data, in favor of either Python loops or `DatasetImporter`s

### Example of recursive data listing

A common user question is whether their `data/` directory can contain nested subfolders, which was not previously supported. Now, all relevant dataset types will support this naturally: when `data_path` contains nested subfolders, the corresponding label files should use `path/to/subfolder/filename.ext` rather than just `filename.ext` to refer to the matching media.

For example, importing COCO data in the following format is supported:

```
data_path/
    im1.jpg
    ...
    nested/
        im2.jpg
        ...
        subfolders/
            im3.jpg
            ...
```

```
{
    "info": {...},
    "licenses": [...],
    "categories": [...],
    "images": [
        {
            "file_name": "im1.jpg",
            ...
        },
        {
            "file_name": "nested/im2.jpg",
            ...
        },
        {
            "file_name": "nested/subfolders/im3.jpg",
            ...
        },
        ...
    ],
    "annotations": [...]
}
```

And here's an example usage:

```py
import eta.core.utils as etau

import fiftyone as fo
import fiftyone.zoo as foz

#
# Create a COCO formatted dataset with images (one, in this case) stored in
# nested subfolders
#

dataset.limit(1).export(
    label_field="ground_truth",
    export_dir="/tmp/coco",
    overwrite=True,
)

etau.move_file(
    "/tmp/coco/data/000880.jpg",
    "/tmp/coco/data/path/to/subfolder/000880.jpg",
)

l = etas.read_json("/tmp/coco/labels.json")
l["images"][0]["file_name"] = "path/to/subfolder/000880.jpg"
etas.write_json(l, "/tmp/coco/labels.json")

# `<dataset_dir>/data` is recursively listed
d1 = fo.Dataset.from_dir(
    dataset_dir="/tmp/coco",
)
print(d1.first().filepath)

# `data_path` is recursively listed
d2 = fo.Dataset.from_dir(
    data_path="/tmp/coco/data",
    labels_path="/tmp/coco/labels.json",
)
print(d2.first().filepath)
```


tests/unittests/import_export_tests.py
```diff
@@ -0,0 +1,1675 @@
...
+        self.assertEqual(len(view), len(dataset2))
+        self.assertEqual(
+            view.count("predictions.detections"),
+            dataset2.count("predictions.detections"),
+        )
+        self.assertEqual(
+            view.distinct("predictions.detections.age"),
+            dataset2.distinct("predictions.detections.age"),
+        )
+        self.assertEqual(
```
benjaminpkane, line 681: This is failing, as booleans are converted strings when outputting as XML.

```
======================================================================
FAIL: test_voc_detection_dataset (__main__.ImageDetectionDatasetTests)
----------------------------------------------------------------------
  File "/home/ben/code/fiftyone/tests/unittests/decorators.py", line 21, in wrapper
    return func(*args, **kwargs)
  File "import_export_tests.py", line 686, in test_voc_detection_dataset
    self.assertEqual(
AssertionError: Lists differ: [False, True] != ['False', 'True']

First differing element 0:
False
'False'

- [False, True]
+ ['False', 'True']
?  +     +  +    +
```

ehofesmann: Since there's no way to tell if the booleans in an XML are meant to be booleans or just strings, do you think it would be better to convert them to `bool` automatically or just leave them as strings?

ehofesmann: I'd say that importing strings is fine for formats that don't have an option for boolean data types

benjaminpkane: Yes, I'm fine with the current behavior. I don't know VOC, but if they don't support booleans I think this is a good best effort. XML itself is certainly capable of holding booleans, if we wanted to look into that at some point.

So a user will have to convert booleans after import. FWM

brimoor: Oops I tried to implement this such that things would be converted to more specific types than string, if possible, during import. But I guess I messed that up for VOC? I think it is working for other XML types like CVAT. Take a look there for copy-paste code?

benjaminpkane: > Oops I tried to implement this such that things would be converted to more specific types than string, if possible, during import. But I guess I messed that up for VOC? I think it is working for other XML types like CVAT. Take a look there for copy-paste code?

Ok, cool. I'll try to adjust.

ehofesmann (approved):
LGTM, ready to merge once tests and VOC are updated

fiftyone/utils/cvat.py
```diff
@@ -1231,32 +1108,22 @@ def __init__(self, occluded=None, attributes=None):
         self.attributes = attributes or []
 
     def _to_attributes(self):
-        attributes = {a.name: a.to_attribute() for a in self.attributes}
+        attributes = {a.name: a.value for a in self.attributes}
 
         if self.occluded is not None:
-            attributes["occluded"] = fol.BooleanAttribute(value=self.occluded)
+            attributes["occluded"] = self.occluded
```
benjaminpkane, line 1114: Hmmm, this errors on occluded boolean attributes. @ehofesmann is this expected?

```py
import fiftyone as fo

dataset = fo.Dataset()
sample = fo.Sample("path/to/image.png")
sample["detection"] = fo.Detection(occluded=True, bounding_box=[0, 0, 1, 1]) 
dataset.add_sample(sample)

dataset.export(
    export_dir="/tmp/cvat-test-export", dataset_type=fo.types.CVATImageDataset
)
```

benjaminpkane:
```
<ipython-input-12-5be5aac31b16> in <module>
      6 dataset.add_sample(sample)
      7 
----> 8 dataset.export(
      9     export_dir="~/tmp/cvat-test-export", dataset_type=fo.types.CVATImageDataset
     10 )

~/code/fiftyone/fiftyone/core/collections.py in export(self, export_dir, dataset_type, data_path, labels_path, export_media, dataset_exporter, label_field, label_prefix, labels_dict, frame_labels_field, frame_labels_prefix, frame_labels_dict, overwrite, **kwargs)
   5557 
   5558         # Perform the export
-> 5559         foud.export_samples(
   5560             self,
   5561             dataset_exporter=dataset_exporter,

~/code/fiftyone/fiftyone/utils/data/exporters.py in export_samples(samples, export_dir, dataset_type, data_path, labels_path, export_media, dataset_exporter, label_field_or_dict, frame_labels_field_or_dict, num_samples, **kwargs)
    289         )
    290 
--> 291     write_dataset(
    292         samples,
    293         sample_parser,

~/code/fiftyone/fiftyone/utils/data/exporters.py in write_dataset(samples, sample_parser, dataset_exporter, num_samples, sample_collection)
    342         (UnlabeledImageDatasetExporter, LabeledImageDatasetExporter),
    343     ):
--> 344         _write_image_dataset(
    345             dataset_exporter,
    346             samples,

~/code/fiftyone/fiftyone/utils/data/exporters.py in _write_image_dataset(dataset_exporter, samples, sample_parser, num_samples, sample_collection)
    670                 else:
    671                     # Export sample
--> 672                     dataset_exporter.export_sample(
    673                         image_or_path, metadata=metadata
    674                     )

~/code/fiftyone/fiftyone/utils/data/exporters.py in __exit__(self, *args)
   1006 
   1007     def __exit__(self, *args):
-> 1008         self.close(*args)
   1009 
   1010     def setup(self):

~/code/fiftyone/fiftyone/utils/cvat.py in close(self, *args)
    489         if self._task_labels is None:
    490             # Compute task labels from active label schema
--> 491             cvat_task_labels = CVATTaskLabels.from_cvat_images(
    492                 self._cvat_images
    493             )

~/code/fiftyone/fiftyone/utils/cvat.py in from_cvat_images(cls, cvat_images)
    741 
    742                 if anno.occluded is not None:
--> 743                     schema.add_object_attribute("occluded", anno.occluded)
    744 
    745                 for attr in anno.attributes:

~/code/fiftyone/eta/eta/core/frames.py in add_object_attribute(self, label, attr)
   1395             attr: an Attribute
   1396         """
-> 1397         self.objects.add_object_attribute(label, attr)
   1398 
   1399     def add_object_attributes(self, label, attrs):

~/code/fiftyone/eta/eta/core/objects.py in add_object_attribute(self, label, attr)
   1726         """
   1727         self._ensure_has_object_label(label)
-> 1728         self.schema[label].add_object_attribute(attr)
   1729 
   1730     def add_object_attributes(self, label, attrs):

~/code/fiftyone/eta/eta/core/objects.py in add_object_attribute(self, attr)
   1234             attr: an `eta.core.data.Attribute`
   1235         """
-> 1236         self.attrs.add_attribute(attr)
   1237 
   1238     def add_object_attributes(self, attrs):

~/code/fiftyone/eta/eta/core/data.py in add_attribute(self, attr)
   1244             attr: an Attribute
   1245         """
-> 1246         name = attr.name
   1247         if name not in self.schema:
   1248             schema_cls = attr.get_schema_cls()

AttributeError: 'bool' object has no attribute 'name'
```

tests/unittests/import_export_tests.py
benjaminpkane: I see the same issue for CVAT.

```py
import fiftyone as fo

dataset = fo.Dataset()
sample = fo.Sample("/home/ben/data/num-1-100/001.png")
sample["detection"] = fo.Detection(boolean=True, bounding_box=[0, 0, 1, 1]) 
dataset.add_sample(sample)

dataset.export(
    export_dir="~/tmp/cvat-test-export", dataset_type=fo.types.CVATImageDataset
)

    dataset_dir="/tmp/cvat-test-export",
    dataset_type=fo.types.CVATImageDataset,
    label_field="predictions",
)

print(dataset2.first())
```

Out:
```
<Sample: {
    'id': '60ff2a4fb11c60c518138e17',
    'media_type': 'image',
    'filepath': '/tmp/cvat-test-export/data/001.png',
    'tags': BaseList([]),
    'metadata': <ImageMetadata: {
        'size_bytes': None,
        'mime_type': None,
        'width': 128,
        'height': 128,
        'num_channels': None,
    }>,
    'predictions_detections': <Detections: {
        'detections': BaseList([
            <Detection: {
                'id': '60ff2a4fb11c60c518138e16',
                'attributes': BaseDict({}),
                'tags': BaseList([]),
                'label': 'None',
                'bounding_box': BaseList([0.0, 0.0, 1.0, 1.0]),
                'mask': None,
                'confidence': None,
                'index': None,
                'boolean': 'True',
            }>,
        ]),
    }>,
}>
```

benjaminpkane:
I'm going to finish the workflow updates, but I think I'll hold off on merging this today.

fiftyone/utils/cvat.py
ehofesmann: I'll push a fix for this. Double checking that there are no other issues

benjaminpkane:
So currently we output `True` and `False` as `"True"` and `"False"` in XML. The current behavior is we do not attempt to convert these back. We use `xmltodict` which makes no mention support booleans.

I am not opposed to leaving these values as strings, though. The other option would be explicitly convert all `"True"` and `"False"` values when loading the XML. See [here](https://github.com/voxel51/fiftyone/blob/6d99d2cf146285d64ab48406aa0457e138dd3775/fiftyone/core/utils.py#L598). Or outputting boolean values with an attribute that does not interfere with the formats involved (CVAT, VOC, etc.).

But again, I am fine with current behavior with respect to this PR. But I want to make sure there is agreement.

brimoor:
I definitely neglected to appropriately handle parsing attribute types for CVAT like I wanted, but that's now fixed in https://github.com/voxel51/fiftyone/pull/1125/commits/218286f00c964cab7271edaec26cf7a272130540.

For VOC, I fixed a sneaky bug in https://github.com/voxel51/fiftyone/pull/1125/commits/417462fedeb4a39adf8ae9ae872032369530359d, but after that, tests were passing locally for me but not in CI. I am guessing the reason is that `distutils` was being used and isn't always available? I got rid of `distutils` in https://github.com/voxel51/fiftyone/pull/1125/commits/2dab1a11fa5dedb11a8f1c24a8e7cdc2ecd88cfc and now the import/export tests pass.

There was then one more failing test due to trying expand the schema of a dataset when `add_samples()` is presented with a Sample with a new field with value `None`. Since that isn't specific to this PR, I copped out and just avoided the broken case for now: https://github.com/voxel51/fiftyone/pull/1125/commits/b632b13694eaa427bb6887a94ba2c2677affd705.

benjaminpkane:
Alright, very weird. One import/export test hangs on windows. Something to with paths I presume. That took way too long, punting.

PR #1127, opened 2021-07-14
Adding CVAT integration for creating, modifying, and deleting labels in a FiftyOne Dataset

ehofesmann:
This PR adds the ability to upload, modify, and reload annotations from FiftyOne into CVAT using only `annotate()` and `load_annotations()` methods on a `SampleCollection`. 

CVAT uses 3 levels of annotation configurations, projects, tasks, and jobs. With this API, you can create tasks and jobs. Specifically, for every label field you want to modify or create, a new task will be created containing one or more jobs with the images in the `SampleCollection` that is being annotated.

CVAT only allows for one video per task, so for video datasets, a new task with a single job is created for every sample and additionally for every label field that is being modified or created.

Label fields each require a separate task since each task only allows for a single label schema that may not be consistent between fields. If annotations are added in CVAT of a type that differs from the specified label field, then the user is prompted to enter a new field name when `load_annotations()` is called.

**Example usage**
See `docs/source/integrations/cvat.rst` for full documentation and plenty of examples :)


fiftyone/utils/cvat.py
```diff
@@ -2377,6 +2425,1904 @@ def write(
...
+        password = fo.annotation_config.cvat_password
+
+        if username is None or password is None:
+            logger.info(
+                "No config or environment variables found for "
+                "authentication. Please enter CVAT login information. Set the "
+                "environment variables `FIFTYONE_CVAT_USERNAME` and "
+                "`FIFTYONE_CVAT_PASSWORD` to avoid this in the future."
+            )
+            self.prompt_username_password(host="CVAT")
```
benjaminpkane, line 2613: I think the you want to return here with the prompt `dict`?

benjaminpkane, line 2613:
```suggestion
            return self.prompt_username_password(host="CVAT")
```

ehofesmann: Ah, yup!

fiftyone/core/collections.py
```diff
@@ -5540,6 +5540,77 @@ def export(
...
+                specified in the `label_schema`. This is required for a new
+                `label_field` if `label_schema` is not provided. For existing label fields, if
+                neither `classes` nor `label_schema` is given, default classes are used
+                if available, otherwise classes are parsed from existing labels in the
+                label field
+            attributes (True): a list of string attributes or dictionary of attribute
+                name to type, values, and default values that will be the default for
+                every label field without attributes specified through the
+                `label_schema`. `True` indicates loading all values for existing
+                label fields. `False` indicates loading no attributes
```
benjaminpkane, line 5582: I'm a little confused here. What does "loading" mean? Attributes will be sent to the labeling backend? What does the `dict` options look like?

```
{ "attr_name": {"type": ..., values: [...], "default": ...}}
```

fiftyone/core/collections.py
```diff
@@ -5540,6 +5540,77 @@ def export(
...
+        label_type=None,
+        classes=None,
+        attributes=True,
+        media_field="filepath",
+        launch_editor=False,
+        **kwargs,
+    ):
+        """Exports the samples and a label field to the given annotation
+        backend.
+
```
benjaminpkane, line 5557: Perhaps add an `Examples::` section

fiftyone/core/config.py
```diff
@@ -198,6 +198,30 @@ def _validate(self):
             self.default_ml_backend = self.default_ml_backend.lower()
 
 
+class AnnotationConfig(EnvConfig):
```
benjaminpkane, line 201: Hmmm, this config will grow as more providers as added? Not terribly important now, but maybe each provider should have a separate config.

fiftyone/utils/cvat.py
```diff
@@ -2377,6 +2425,2054 @@ def write(
...
+                    d["status_code"],
+                    d["request"],
+                    d["url"],
+                    d["reason"],
+                    d["_content"],
+                )
+            )
+
+    def delete(self, url):
+        """Send delete request to the given CVAT API URL
```
benjaminpkane, line 2554:
```suggestion
        """Sends a delete request to the configured CVAT backend.
```

docs/source/integrations/cvat.rst
```diff
@@ -0,0 +1,986 @@
...
+<https://cvat.org>`_ or through a 
+`custom installation and self-hosted server. <https://openvinotoolkit.github.io/cvat/docs/administration/basics/installation/>`_
+No matter the server being used, an account will be required and the username
+and password must be provided to FiftyOne.
+
+CVAT uses three levels of abstraction for annotation workflows: projects,
+tasks, and jobs. A job contains one or more images and can be assigned to a
+specfic annotator or reviewer. A task defined the label schema to use for
+annotation and contains multiple jobs. A project also allows for a label schema
+and contains multiple tasks.
```
benjaminpkane, line 102: Typos, I think. Could use edits.

docs/source/integrations/cvat.rst
```diff
@@ -0,0 +1,986 @@
...
+Setup
+_____
+
+
+Server URL
+----------
+
+Both `cvat.org <https://cvat.org>`_ and custom CVAT servers are supported. The
+following attributes allow specification of the server URL that you have an
+account on and to which you want to upload data:
```
benjaminpkane, line 123:
```suggestion
account with and to which you want to upload data:
```

docs/source/integrations/cvat.rst
```diff
@@ -0,0 +1,986 @@
...
+* `url`: base url of the CVAT server (e.g. `cvat.org` or `localhost`)
+* `port`: four digit port of the custom CVAT server if applicable
+* `https`: boolean indicating whether the URL is `https` (`True`) or `http`
+  (`False`)
+
+
+The environment variables `FIFTYONE_CVAT_URL`, `FIFTYONE_CVAT_PORT`, and
+`FIFTYONE_CVAT_HTTPS` can be set to avoid providing them in every 
+:meth:`annotate() <fiftyone.core.collections.SampleCollection.annotate>`
+call or they can be set in the `~/.fiftyone/annotation_config.json`.
```
benjaminpkane, line 134: Link to how to construct an `annotation_config.json`?

fiftyone/core/collections.py
brimoor: FYI I updated the integration docs page. Hopefully this is explained better now

fiftyone/core/collections.py
brimoor: I added a link to the integration docs page to cover the examples request here

fiftyone/core/config.py
brimoor: yeah, I agree. Still pondering how to best implement this, considering that I'd like for users to be able to write their own annotation plugin (supporting exact same syntax as natively supported backends) without editing code within the FiftyOne package.

docs/source/integrations/cvat.rst
brimoor: yeah, documentation should be added to the FiftyOne config user guide page on this, too

fiftyone/utils/cvat.py
```diff
@@ -2278,6 +2312,20 @@ def to_eta_attribute(self):
 
         return etad.CategoricalAttribute(self.name, self.value)
 
+    def to_attribute(self):
+        """Returns a :class:`fiftyone.core.labels.Attribute` representation of
+        the attribute.
+        Returns:
+            a :class:`fiftyone.core.labels.Attribute`
+        """
+        if isinstance(self.value, bool):
```
brimoor, line 2321: We're trying to get away from using the `fiftyone.core.labels.Attribute` classes. I wanted to delete them in recent work I did, but stopped short since `label.attributes` may still be used by some datasets out there.

Can you avoid doing this type conversion?

fiftyone/utils/cvat.py
```diff
@@ -2377,6 +2425,2066 @@ def write(
...
+    server url and credentials.
+
+    This API provides methods to easily get, put, post, patch, and delete tasks
+    and jobs through the formatted urls specified by the CVAT REST API.
+
+    Additionally, samples and label schemas can be uploaded and annotations
+    downloaded through this class.
+
+    Args:
+        url ("cvat.org"): URL of the CVAT server to which to upload samples
```
brimoor, line 2441: How about deleting the `https` parameter and just asking the user to put `https://cvat.org`, `http://my-cvat.com` or `localhost` or whatever into the `url` parameter? Fewer parameters = better.

You can deduce/strip out the appropriate `http/s` value if you really need it.

fiftyone/utils/cvat.py
```diff
@@ -2377,6 +2425,2066 @@ def write(
...
+            labels (None): the label schema to use for the created task
+            segment_size (None): maximum number of images to load into a job.
+                Not applicable to videos
+            image_quality (75): an int in `[0, 100]` determining the image
+                quality to upload to CVAT
+            task_assignee (None): the username to assign the created task(s)
+            task_name ("FiftyOne_annotation"): a name for the CVAT task
+
+        Returns:
+            a tuple of
```
brimoor, line 2694: FYI I updated some of the `Returns:` sections here. Sphinx does not allow `Returns:` to have multiple "coloned" arguments like `Args:` does. You have to fiddle with it like you see here to make it look similar.

Generating the docs and inspecting how it looks is the easiest way I've found to learn what works. Inspecting existing code and copying its patterns is always good too.

fiftyone/utils/cvat.py
```diff
@@ -2377,6 +2425,2066 @@ def write(
...
+                upload to CVAT
+            label_schema: a dictionary containing the description of label
+                fields, classes and attribute to annotate
+            media_field ("filepath"): string field name containing the paths to
+                media files on disk to upload
+            segment_size (None): maximum number of images to load into a job.
+                Not applicable to videos
+            image_quality (75): an int in `[0, 100]` determining the image
+                quality to upload to CVAT
+            task_assignee (None): the username to assign the created task(s)
```
brimoor, line 2837: FYI I removed the word "sequentially" from these docstrings so that there could be one source of truth for documenting details like how assignees are chosen- the integration page docs

docs/source/integrations/cvat.rst
```diff
@@ -0,0 +1,1034 @@
...
+  applicable to videos
+- `image_quality`: an int in `[0, 100]` determining the image quality to upload
+  to CVAT
+- `task_assignee`: a username to assign the generated tasks
+- `job_assignees`: a list of usernames to assign jobs
+- `job_reviewers`: a list of usernames to assign job reviews
+
+.. _cvat-label-schema:
+
+Label schema
```
brimoor, line 318: I combined the `Label schema` and `Attributes` sections and spent some time trying to make it a bit easier to understand all the different options at one's fingertips. Hopefully you agree it reads a bit easier now!

A latent goal while doing my editing was to "anonymize" as much of the copy as possible by removing unnecessary mentions of "CVAT" so that much of this can be upgraded to an `Annotations` user guide page soon.

brimoor (changes requested):

brimoor:
As per our live discussion, I will work on storing the annotation info to disk rather than returning it in-memory in a separate PR over the weekend. I will also work on some changes to the `FiftyOneAnnotationConfig` per the comments I left in the PR.

I won't be touching this PR anymore, unless any of the edits I pushed earlier today introduced any bugs... 😟

fiftyone/utils/cvat.py
ehofesmann: I do still need this to be able to use the quickstart dataset in examples since `iscrowd` is stored as a FiftyOne attribute. Though we should probably update that dataset so that `iscrowd` and `area` are stored directly on labels instead of in attributes

brimoor (approved):
LGTM

PR #1130, opened 2021-07-19
width height mismatch from coco formatted data

shortcipher3:

Found a bug where for non-square images the bounding boxes were being imported wrong into Voxel51.  Found that width and height had been swapped in one line of the coco utils.  All the documentation gives a `(width, height)` format, which matches coco format, but one line has the `frame_size` swapped to a `(height, width)` format.

Swapped width and height normalization parameters in test data and observed that although the numbers in the prediction file don't match pixel locations the boxes are drawn correctly. 

        notes for FiftyOne users.
Bugfix in MSCOCO Detection Import Utility to correctly draw bounding boxes.

-   [ ] Other


brimoor (approved):
LGTM. Thanks for catching this 👍

PR #1133, opened 2021-07-20
Fixing loading of zoo datasets with partial downloads (#1129)

### LINKED ISSUE #1129: [BUG] Open-images-v6 download check does not include metadata
### Instructions

I downloaded the OpenImages-V6 dataset the first time like this, by accident (using `download_zoo_dataset` instead of `load_zoo_dataset`):

```python
fiftyone.zoo.datasets.download_zoo_dataset(
        name='open-images-v6',
        dataset_dir='~/data/open-images-v6',
        label_types='classifications',
        num_workers=8
    )
```

I ended up downloading the whole dataset, which is fine. But I noticed I only had `classifications.csv`. So I tried to download the rest, figuring I could use `download_zoo_dataset` again.

```python
fiftyone.zoo.datasets.download_zoo_dataset(
        name='open-images-v6',
        dataset_dir='~/data/open-images-v6',
        label_types=['detections','classifications','relationships','segmentations']
        num_workers=8
    )
```

gave the output:

```console
Split 'test' already downloaded
Split 'validation' already downloaded
```

Okay great, so then I tried to load it:

```python
fiftyone.zoo.datasets.load_zoo_dataset(
        name='open-images-v6',
        dataset_dir='~/data/open-images-v6',
        label_types=['detections','classifications','relationships','segmentations'],
        num_workers=8
    )
```

But I got an output such as this:

```console
Split 'train' already downloaded                                                                                                          Split 'test' already downloaded
Split 'validation' already downloaded                                                                                                     Loading 'open-images-v6' split 'train'                                                                                                    Uncaught exception
Uncaught exception
  File "/home/collin_m_mccarthy/torchvx/utils/fiftyone_download.py", line 60, in <module>
    download_dataset(config)
  File "/home/collin_m_mccarthy/torchvx/utils/fiftyone_download.py", line 42, in download_dataset
    fiftyone.zoo.datasets.load_zoo_dataset(
  File "/home/collin_m_mccarthy/.conda/envs/torchvx/lib/python3.9/site-packages/fiftyone/zoo/datasets/__init__.py", line 273, in load_zoo_dataset
    dataset.add_importer(
  File "/home/collin_m_mccarthy/.conda/envs/torchvx/lib/python3.9/site-packages/fiftyone/core/dataset.py", line 2755, in add_importer
    return foud.import_samples(
  File "/home/collin_m_mccarthy/.conda/envs/torchvx/lib/python3.9/site-packages/fiftyone/utils/data/importers.py", line 107, in import_samples
    with dataset_importer:
  File "/home/collin_m_mccarthy/.conda/envs/torchvx/lib/python3.9/site-packages/fiftyone/utils/data/importers.py", line 694, in __enter__
    self.setup()
  File "/home/collin_m_mccarthy/.conda/envs/torchvx/lib/python3.9/site-packages/fiftyone/utils/openimages.py", line 260, in setup
    ) = _setup(
  File "/home/collin_m_mccarthy/.conda/envs/torchvx/lib/python3.9/site-packages/fiftyone/utils/openimages.py", line 722, in _setup
    attrs_map = _get_attrs_map(dataset_dir, download=download)
  File "/home/collin_m_mccarthy/.conda/envs/torchvx/lib/python3.9/site-packages/fiftyone/utils/openimages.py", line 792, in _get_attrs_map    attrs_csv = _get_general_metadata_file(
  File "/home/collin_m_mccarthy/.conda/envs/torchvx/lib/python3.9/site-packages/fiftyone/utils/openimages.py", line 785, in _get_general_metadata_file
    _download_file_if_necessary(filepath, url, quiet=0, download=download)
  File "/home/collin_m_mccarthy/.conda/envs/torchvx/lib/python3.9/site-packages/fiftyone/utils/openimages.py", line 1590, in _download_file_if_necessary
    raise ValueError("File '%s' is not downloaded" % filepath)
ValueError: File '/home/collin_m_mccarthy/data/open-images-v6/train/metadata/attributes.csv' is not downloaded
```

It looks like when it checks to see if everything is downloaded it's not checking to make sure the metadata is downloaded as well. I believe it should, so at a minimum `load_zoo_dataset` works correctly and downloads it automatically if it is not.  Based on the documentation it sounds like that's how it's supposed to work?

Thank you,
Collin

-   [X] `Core`: Core `fiftyone` Python library issue

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community.
-   [X] No. I cannot contribute a bug fix at this time.

brimoor:
Resolves #1129.

The logic in `load_zoo_dataset()` had gotten a bit too fancy and was attempting to skip downloading a dataset if the correct number of images were already downloaded. However, as #1129 demonstrated, this was not sufficient, as it could be the case that all images are downloaded but some additional annotations may need to be downloaded.

Now, for zoo datasets that support partial downloads, the `_download_and_prepare()` method will *always* be called, which gives the dataset full control to decide if additional content needs to be downloaded. Furthermore, the `_download_and_prepare()` methods for COCO and Open Images were optimized as much as possible to avoid unnecessary file downloads and unnecessary loading of annotations into memory (when possible).

The reason we got into this situation in the first place is that I wanted the logging output to be as specific as possible. IE, when a dataset split is already downloaded, I wanted to see this:

```
Split 'validation' already downloaded
Loading 'open-images-v6' split 'validation'
```

which is what one sees for zoo datasets that don't support partial downloads. However, with this change, the user will now always see something like this this for zoo datasets that *do* support partial downloads but are already downloaded:

```
Downloading split 'validation' to '/Users/Brian/fiftyone/open-images-v6/validation' if necessary
Necessary images already downloaded
Existing download of split 'validation' is sufficient
Loading 'open-images-v6' split 'validation'
```

Example usage:

```py
import fiftyone.zoo as foz

# Downloads entire split w/ subset of label types
    "open-images-v6",
    split="validation",
    label_types=["classifications"],
)
print(dataset)

dataset.delete()

# Downloads remaining label types
# This would previously fail; now succeeds
    "open-images-v6",
    split="validation",
    label_types=["detections", "classifications", "relationships", "segmentations"],
)
print(dataset)
```


ehofesmann (dismissed):
I'm all for streamlining the check for now. 

I should point out that `fiftyone.utils.openimages.is_download_required` is pretty fast for `validation`, but for the `train` set it takes upwards of 5 minutes on my laptop to load in all of the dataframes into memory. This has to be performed at least one more time when loading the dataset (and another time if additional downloads are required). The issue is that we really do need to read and parse the labels into memory to see if we need to download additional data for a dataset that supports partial downloads in most situations. Possible solutions are to:
1) Add additional logic that can make this check faster in situations where we don't need to parse labels (ex: if only max samples is specified, we only really need to check for images and the existence of label files). However, this brings us back to the issue this PR addresses of adding unnecessary complexity.
2) Avoid checking if downloads are necessary and just go ahead and download whatever is necessary at the same time that you check it.
3) Let it take a few minutes to load in datasets with >1M samples

I'll go ahead and approve this and let you decide if we should go with 1 or 2 above

brimoor:
Good point, 5 minutes is a long time, especially redundant loads just to make the logging output more "concise". 1 or 2. Hmmm...

brimoor:
@ehofesmann I updated this in response to your helpful feedback. The updated PR description describes what I did.

If you could test this before merging, that would be much appreciated! There are zoo dataset tests in `tests/intensive/dataset_zoo_tests.py`. I don't have sufficient internet speeds at the moment to download large datasets.

ehofesmann (approved):
LGTM, tests are passing and partial download datasets always calling `_download_and_prepare` is a good solution!

PR #1134, opened 2021-07-20
Flashlight

benjaminpkane:
Replaces the react infinite scroll grid component with a virtualized infinite scroll grid. Only visible and nearly visible samples are rendered to the DOM tree.


benjaminpkane:
My intention is to undraft this tomorrow. I need to finish selections in the grid and expanded view  `->` (next) and `->` previous actions,  add sample updates after tagging, and I want to make a few optimizations for rapid scrolling.

I'm also thinking about setting a minimum grid size value based on the window width. A thin window could mean 100+ samples in the grid which is a lot, particularly for video.

benjaminpkane:
Still needs integration, but v1 virtualization is finalized.

benjaminpkane:
Trying to make it great, even for video. Almost there.

benjaminpkane:
Undrafting. Adding deselection to the escape sequence now.

benjaminpkane:
Ok, the pixel animation is only shown when loading the initial page (first rows). When scrolling fast, and updates need to occur, the layout is shown, but not the media. In other words, a grey a layout.

benjaminpkane:
Extended view filtering has a regression. Working through it.

brimoor (approved):
LGTM! Flashlight is working *really* well. The grid for a 100k image dataset is super responsive and loads really fast for me, and really loving how toggling things on and off is always instantaneous now.

Awesome work 🔥 🥇 

The only things I noticed are:

- Clicking the grid icon on a video dataset allows you to achieve a zoom outside of the more limited range that you're wanting to enforce for video datasets. This also makes me wonder where the users app setting will be appropriately clipped if it is too small for video datasets? Better check that. Side point, I think zoom = 5 looks reasonable for video. Perhaps extend the allowable range to include it? (min is 6 currently?)

- For video datasets, it does take a bit longer to load a page than for image datasets, and you can't scroll past the current page until it fully loads. This can lead to some confusion between whether you've reached the end of the view or whether you need to wait a bit until you can continue scrolling. This is really more of an observation than anything concrete, since the real way to address this is full scrubbing support, which I know is in your icebox. But, I believe a limiting factor is the number of ffmpeg "elements" (currently 5?) that are working on prepping the videos; perhaps you could bump that up a bit to speed things up? I certainly am not noticing any performance complaints from my machine when scrubbing through video datasets

PR #1151, opened 2021-07-31
Open Images download hotfix

brimoor:
Fixes a bug where label files were not being downloaded when running a `load_zoo_dataset()` call that does not include `classes` or `attrs` options in an environment where Open Images has never been downloaded:

```py
import fiftyone.zoo as foz

foz.delete_zoo_dataset("open-images-v6", split="validation")

# previously would fail, now succeeds
dataset = foz.load_zoo_dataset("open-images-v6", split="validation", max_samples=5)
```


benjaminpkane (approved):
LGTM

PR #1152, opened 2021-08-02
Only load matching samples when classes/attrs requirements are provided

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1153.

Fixes a bug where `classes/attrs` requirements were being ignored when loading the Open Images dataset with no `max_samples` requirement.

The *download* portion of the code was respecting the requirements, but the *import* portion was not (when no `max_samples` requirement was provided), so, as the example below demonstrates, running multiple loads could currently cause issues:

```py
import fiftyone.zoo as foz

# Download some samples
foz.load_zoo_dataset(
    "open-images-v6",
    split="validation",
    max_samples=50,
)

# This previously loaded all downloaded samples, but now only loads `House` samples, as expected
    "open-images-v6",
    split="validation",
    classes=["House"],
)
```


benjaminpkane (approved):
LGTM

PR #1154, opened 2021-08-02
Adding session.wait() instructions in key places

brimoor:
A common user issue is using `launch_app()` in a script without including `session.wait()`.

This PR attempts to document the correct usage of the FiftyOne App from scripts in the primary places where (I think?) new users might be going wrong:

Screen Shot 2021-08-02 at 12 36 17 PM

Screen Shot 2021-08-02 at 12 36 26 PM

Screen Shot 2021-08-02 at 12 36 47 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,9 +86,14 @@ dataset = foz.load_zoo_dataset("quickstart")
 session = fo.launch_app(dataset)
 ```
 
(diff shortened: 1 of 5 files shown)


benjaminpkane (approved):
LGTM!

PR #1156, opened 2021-08-04
App user guide updates

benjaminpkane:
* Adds notes about cropping to content in the patches sections
* Adds `Using the image visualizer` and `Using the video visualizer` sections


brimoor:
Lookin' good! 🥇 

Here's a few things I found on my latest test:

Help menu
- Replace `Escape help > JSON > settings > zoom > fullscreen > clear > close` help text with just `Escape the current context`. I think more is less here, and people will figure it out. It'll also help with line wrapping on 50% width screens :)
- Organize zoom in/out and next/previous frame in settings menu so that each pair is vertically adjacent (easier to "group" them visually that way). Currently it takes time to visually locate the sister setting within each pair

Looker HUD
- Persist audio/playback settings while arrowing through samples (they currently reset for each sample, unlike other settings)

Settings menus
- Horizontally center `Refresh colors [icon]` text within button, for consistency with other menu buttons

Label tags
- Tooltip for label tag eye icon should read: `Only show labels with the "<tag>" tag or other selected tags`. Both in grid and expanded sample views. It currently says "samples", not "labels"

Grid
- Loading an empty view in the App (eg `limit(0)`) causes an infinite loading state, but should instead show an empty grid so the user knows that nothing is loading

Video
- Maximizing a video using the HUD controls while it is playing freezes the App

Extended view
- Combining a label tag selection with a label field filter isn't working correctly. For example, in the screenshot below, I have some labels with the `duplicate` tag. I clicked the eyeball icon to only show those labels. Then I tried to further filter by using the label filter widget for a field (I wanted to only see `person` labels with the `duplicate` tag). But this resulted in many samples being shown in the grid even though in the upper right it says 1 of 500 samples

Expanded modal
- When performing actions such as tagging labels in the expanded sample modal, applying a change doesn't refresh the modal's data (eg tags don't appear in the tooltip until I close and reopen the modal)
    - Corollary: add support for reopening expanded modal to same sample when the App is refreshed via ctrl + R?
    - Corollary: add refresh button to expanded modal so user doesn't have to close/reopen or ctrl + R?

Screen Shot 2021-08-05 at 7 26 38 PM

benjaminpkane:
> For video datasets, it does take a bit longer to load a page than for image datasets, and you can't scroll past the current page until it fully loads. This can lead to some confusion between whether you've reached the end of the view or whether you need to wait a bit until you can continue scrolling. This is really more of an observation than anything concrete, since the real way to address this is full scrubbing support, which I know is in your icebox. But, I believe a limiting factor is the number of ffmpeg "elements" (currently 5?) that are working on prepping the videos; perhaps you could bump that up a bit to speed things up? I certainly am not noticing any performance complaints from my machine when scrubbing through video datasets

This only has to do with `ffprobe` in the server pagination requests. It is `<video/>` elements in the browser which are pooled for creating the first frame thumbnail. That process occurs after the layout is already on screen. It doesn't block list growth.

I'll add a check for existing metadata to skip the media inspection (image and video)

benjaminpkane:
Comments addressed

brimoor:
> Comments addressed

Confirmed that everything LGTM

brimoor (approved):
LGTM

~~I still need to add the GIFs for the app docs before releasing though~~ GIFs are in!

brimoor:
Couple small things I observed after doing some work today:

- I'd recommend not ever making the settings menu orange. Currently it turns orange, I believe, when switching into patches mode because of the new setting to toggle b/w full image and cropped mode, but I think its a bit confusing because the settings menu contains multiple things and its not clear what is being indicated. Also in the future, the novelty of this setting would wear off, so best to "future proof" the App by not introducing a transient highlight now I think

- In patches views, the similarity action appears (and turns orange) whenever I open the expanded sample modal, which makes sense because it is valid to perform a similarity search based on whatever labels are visible in that sample (if the user has generated an appropriate similarity index). However, the similarity action does *not* appear when opening the expanded sample modal for a non-patches view (only when the user selects some labels), but it does make sense to do a similarity search based on the current sample, same as if the user selected that sample from grid view

Since there are some subtle cases, I'll list the desired behavior of similarity search:
- **Non-patches grid view**: when sample(s) are selected, show similarity action. Menu should show all similarity indices with `patches_field==None`, and performing a search should use all selected samples as query IDs
- **Patches gird view**: when sample(s) are selected, show similarity action. Menu should show all similarity indices with a `patches_field` that matches the label field (either field, in the case of evaluation views), and performing a search should use all labels within the selected patches that match the `patches_field` of the chosen index as query IDs
- **Non-patches expanded view**: when no labels are selected, show similarity action. Menu shows all similarity indices with `patches_field==None`, and performing a search should use the current sample as query ID. When labels are selected, still show similarity action. Menu now shows all similarity indices with a `patches_field` that matches the label field(s) from which any selected labels come from. Performing a search uses all labels from the `patches_field` of the chosen similarity index as query IDs
- **Patches expanded view**: always show similarity action. Manu shows all similarity indices with `patches_field` matching a label field in the view (there are two in the case of evaluation patches). Performing a search uses all labels from the `patches_field` of the chosen similarity index as query IDs if no labels are selected, or only the selected labels if labels are selected

benjaminpkane:
Comments addressed.

PR #1158, opened 2021-08-05
Adding support for importing DICOM datasets

brimoor:
Adds support for importing medical image datasets stored in [DICOM format](https://en.wikipedia.org/wiki/DICOM) (`*.dcm`).

The canonical [pydicom](https://github.com/pydicom/pydicom) library is used to read the files.

### Notes

Exporting in DICOM format is not supported as the format requires many specific metadata fields that would not make sense to populate via FiftyOne. IE read-only is the primary workflow with DICOM.

This PR supports basic reading of static images. 3D or 4D image types (e.g., CT scans) will take a bit more work to properly support (but is doable). Leaving that for future work.

### Example usage

```py
import pydicom
from pydicom.data import get_testdata_file

import fiftyone as fo

dicom_path = get_testdata_file("MR_small.dcm")

    dicom_path=dicom_path,
    images_dir="/tmp/dicom",
    dataset_type=fo.types.DICOMDataset,
    label_field="",
)

```

Or, only loading specific attributes:

```py
    dicom_path=dicom_path,
    images_dir="/tmp/dicom",
    dataset_type=fo.types.DICOMDataset,
    keywords=["PatientName", "StudyID"],
    label_field="",
)

print(dataset)
```

Screen Shot 2021-08-05 at 11 24 21 AM

### Documentation

Screen Shot 2021-08-05 at 11 39 14 AM


docs/source/user_guide/dataset_creation/datasets.rst
```diff
@@ -2666,6 +2669,164 @@ as follows:
...
+        DICOM_PATH='/path/to/*.dcm'  # glob pattern of DICOM files
+        # DICOM_PATH='/path/to/DICOMDIR'  # DICOMDIR file
+
+        # Create the dataset
+        fiftyone datasets create \
+            --name $NAME \
+            --type fiftyone.types.DICOMDataset \
+            --kwargs \
+                dicom_path=$DICOM_PATH \
+                keywords=PatientName,StudyID  # load specific attributes
```
benjaminpkane, line 2822:
```suggestion
                label_field="" \
                keywords=PatientName,StudyID  # load specific attributes
```

benjaminpkane (approved):
LGTM

docs/source/user_guide/dataset_creation/datasets.rst
brimoor: Yeah, I'm actually planning a more general update to how `label_field` is used so that `ground_truth` won't be prepended when there are >1 fields being imported.

So I'll punt on this change for now because after the aforementioned change the current syntax will do the right thing.

PR #1160, opened 2021-08-05
Gracefully ignore None-valued new fields when expanding dataset schemas

brimoor:
The impetus for this PR is that I wanted the following to work (which it now does):

```py
import fiftyone as fo

dataset = fo.Dataset()

samples = [
    fo.Sample(filepath="image1.jpg", ground_truth=None),
    fo.Sample(filepath="image2.jpg", ground_truth=fo.Classification()),
]
dataset.add_samples(samples)

print(dataset)  # has `ground_truth` field
```

Previously, an error would be raised because we weren't able to infer an appropriate field type for `ground_truth` after only seeing the first sample.

A consequence of this change is that, in the following example, no `ground_truth` field will be added to the dataset, since we don't know what type to use for the field:

```py
import fiftyone as fo

sample = fo.Sample(filepath="image.jpg", ground_truth=None)

dataset = fo.Dataset()
dataset.add_sample(sample)

print(dataset)  # no `ground_truth` field
```

I think this consequence is fine, since our data model ensures complete equivalence between `None`-valued and missing fields on samples.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -1394,7 +1394,16 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         if validate:
             self._validate_samples(samples)
 
-        dicts = [sample.to_mongo_dict() for sample in samples]
+        dicts = []
+        for sample in samples:
+            d = sample.to_mongo_dict()
+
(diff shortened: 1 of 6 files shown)


benjaminpkane (approved):
~~LGTM~~ I didn't notice evaluation tests were failing. My mistake.

PR #1168, opened 2021-08-09
Improved handling of default label field when importing/exporting datasets

brimoor:
This PR scratches an itch that has been bugging me for awhile with regards to the default behavior of label fields when importing/exporting datasets.

### Dataset imports

Previously, `label_field="ground_truth"` was predominantly used in the codebase as a default value when importing labeled datasets. This was fine and good for dataset types with a single label field, but, for multitask label formats like BDD, Open Images, COCO, etc., the default behavior was to prepend `"ground_truth_"` to all keys of the imported labels dictionary, which was a bit verbose and annoying.

As of this PR, the default value of `label_field=None` is used everywhere, with this default behavior:

```
            label_field (None): controls the field(s) in which imported labels
                are stored. Only applicable if the dataset importer used is a
                :class:`fiftyone.utils.data.importers.LabeledImageDatasetImporter`.
                If the importer produces a single
                :class:`fiftyone.core.labels.Label` instance per sample, this
                argument specifies the name of the field to use; the default is
                ``"ground_truth"``. If the importer produces a dictionary of
                labels per sample, this argument specifies a string prefix to
                prepend to each label key; the default in this case is to
                directly use the keys of the imported label dictionaries as
                field names
```

I think this is the best of all worlds:
- For unlabeled datasets where `label_field` doesn't even make sense, seeing `label_field=None` as a default value in the code is more sensible than seeing `"ground_truth"`
- For labeled datasets with a single label field, the current default behavior is maintained (`"ground_truth"` is used)
- For multitask labeled datasets, the keys of the `DatasetImporter.label_cls` dict are directly used as field names by default, which means that importing `quickstart-video`, for example, yields a frame-level field called `detections` rather than the verbose `ground_truth_detections`:

```py
import fiftyone.zoo as foz
print(foz.load_zoo_dataset("quickstart-video"))
```

```
Name:        quickstart-video
Media type:  video
Num samples: 10
Persistent:  False
Tags:        []
Sample fields:
    id:       fiftyone.core.fields.ObjectIdField
    filepath: fiftyone.core.fields.StringField
    tags:     fiftyone.core.fields.ListField(fiftyone.core.fields.StringField)
    metadata: fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.metadata.Metadata)
Frame fields:
    id:           fiftyone.core.fields.ObjectIdField
    frame_number: fiftyone.core.fields.FrameNumberField
    detections:   fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.labels.Detections)
```

### Exporting datasets

The `export()` method previously provided three mutually exclusive parameters `label_field`, `labels_prefix`, and `labels_dict` specifying which label fields to export. 

As of this PR, these have been condensed into a single `label_field` parameter with the following semantics:

```
            label_field (None): controls the label field(s) to export. Only
                applicable to labeled image datasets or labeled video datasets
                with sample-level labels. Can be any of the following:

                -   the name of a label field to export
                -   a glob pattern of label field(s) to export
                -   a list or tuple of label field(s) to export
                -   a dictionary mapping label field names to keys to use when
                    constructing the label dictionaries to pass to the exporter

                Note that multiple fields can only be specified when the
                exporter used can handle dictionaries of labels. By default,
                the first field of compatible type for the exporter is used
```

IMO this is cleaner and  still covers all cases, with the additional ability to now directly specify a list of label fields to export without having to construct a dictionary that maps label fields to keys for the labels dictionary passed to the exporter (in this case the keys default to the label field names themselves).

In addition, `DatasetExporter.label_cls` can now be a list/tuple of label field types, which declares that the exporter can export a single label field that can be one of multiple types. `ImageSegmentationDirectoryExporter` is an example of a dataset format that follows this pattern, as it can export segmentations, instances, or polylines, but it should only receive one mask per sample (previously it was implemented to overlay multiple masks into a single mask for export, but that was more of a hack to fit the data model than an intended feature).


brimoor:
A couple discussion points:
- Does it bother anyone that the parameter name is `label_field` even though it *can* refer to multiple label fields, a prefix, etc? There is somewhat of a convention in the rest of the codebase that if a field can refer to one or more things, it has a plural name (`label_fields`). But, `label_field` is quite central to the codebase at this point and changing that would definitely cause people annoyance, so removing `label_field` is out of the question. I also though about providing a `label_fields` alias for the parameter so that the user could provide either... There is somewhat of a precedent for this with the `split` and `splits` parameters of `load_zoo_dataset()`, but I hesitate to intentionally introduce more parameters that differ by one character like that...
- As the documentation changes show, this PR will change the field names when datasets like `quickstart-video` and `bdd100k` are imported. I think this is not too bad since these are primarily demonstrative datasets and, if the user has imported these and made them persistent (the only way a problem would arise if a user tries to run the updated docs examples), then the user could always "fix" field not found errors that arise by renaming the field on their existing import or just deleting and re-importing. But, is anyone worried about field potential "your examples are broken" issues from users?

benjaminpkane:
> As the documentation changes show, this PR will change the field names when datasets like quickstart-video and bdd100k are imported. I think this is not too bad since these are primarily demonstrative datasets and, if the user has imported these and made them persistent (the only way a problem would arise if a user tries to run the updated docs examples), then the user could always "fix" field not found errors that arise by renaming the field on their existing import or just deleting and re-importing. But, is anyone worried about field potential "your examples are broken" issues from users?

If you're saying that it should only affect persisted datasets from examples, that is small potatoes to me. I don't have an issue with that.

benjaminpkane:
> Does it bother anyone that the parameter name is label_field even though it can refer to multiple label fields, a prefix, etc? There is somewhat of a convention in the rest of the codebase that if a field can refer to one or more things, it has a plural name (label_fields). But, label_field is quite central to the codebase at this point and changing that would definitely cause people annoyance, so removing label_field is out of the question. I also though about providing a label_fields alias for the parameter so that the user could provide either... There is somewhat of a precedent for this with the split and splits parameters of load_zoo_dataset(), but I hesitate to intentionally introduce more parameters that differ by one character like that...

I agree with your approach from what I've read. Ideally, maybe we ask users to provide functions (we'll give you some information, give us back a string, etc.). Functions can't be serialized, though.

In general, handing off nuance and complexity about how to name things is best left with users imo.

ehofesmann:
1) I'm fine with overloading `label_field` to support multiple fields and dictionaries. After thinking on our discussion last week, I get how providing more arguments can make it difficult to manage what to use for new users. Just FYI, this never ended up coming up for the CVAT PR since we decided to only allow `label_field` to be a string and to use `label_schema` for multiple fields. 

2) Agreed with Ben, this doesn't seem like it would be a huge issue. I'm not sure how many users would have made the example datasets persistent

brimoor:
okay this is ready to go now

benjaminpkane (approved):
LGTM

PR #1180, opened 2021-08-11
Zoom fix for empty samples

### LINKED ISSUE #1175: [BUG] ESC key has no effect in expanded sample modal when sample has no label fields
On `fiftyone==0.12.0`, pressing ESC in the expanded sample modal to reset zoom to full image has no effect in the example below, where a view with all non-default fields excluded is loaded in the App.

I'm guessing the issue is that there is a context check related to zoom-to-content or similar that is higher in the priority order which is "capturing" the ESC event and doing nothing. Such check needs to be skipped when it is inapplicable.

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Open expanded sample modal, zoom in, then press ESC
# Nothing happens, but view should reset to full image
```

benjaminpkane:
Resolves #1175 

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Open expanded sample modal, zoom in, then press ESC
# Nothing happens, but view should reset to full image
```

### DIFF
--- a/app/packages/looker/src/index.ts
+++ b/app/packages/looker/src/index.ts
@@ -555,7 +555,7 @@ export class FrameLooker extends Looker<HTMLVideoElement, FrameState> {
 
     if (this.state.zoomToContent) {
       toggleZoom(this.state, this.currentOverlays);
-    } else if (this.state.setZoom && this.pluckedOverlays.length) {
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM. Tested and works

PR #1181, opened 2021-08-11
Fix range slider blurs

### LINKED ISSUE #1176: [BUG] Grid zoom tooltip does not always disappear after click release
The tooltip of the grid zoom slider is supposed to always disappear after click-release, which it normally does, but it reliably is stuck in the visible state whenever I drag to the extremes of the slider.

Screen Shot 2021-08-10 at 9 55 19 PM

benjaminpkane:
Resolves #1176


brimoor (approved):
LGTM. Tested and works

PR #1182, opened 2021-08-11
Fix next/previous click controls in notebooks

### LINKED ISSUE #1174: [BUG] Cannot navigate between samples in expanded sample modal in a Jupyter notebook
With `fiftyone==0.12.0` in a Jupyter notebook with the expanded sample modal open, clicking the left/right arrows and also left/right arrow keys to navigate between samples has no effect.

benjaminpkane:
Resolves #1174

Fixes the click arrows for next/previous actions in notebook Lookers.

Regarding the arrow keys, notebooks currently require the shift key for wheeling and arrow shortcuts, as the App is within an iframe and these events interfere with scrolling (horizontal or vertical) unless `Shift` key is active.

I am unaware of any horizontal scrolling that could be possible, so I could remove the shift key requirement for left/right arrows. But it would interfere, if horizontal scrolling is available.

Screenshot from 2021-08-11 08-52-51

### DIFF
--- a/app/packages/looker/src/elements/base.ts
+++ b/app/packages/looker/src/elements/base.ts
@@ -44,9 +44,7 @@ export abstract class BaseElement<
       this.events[eventType] = (event) =>
         handler({ event, update, dispatchEvent });
       this.element &&
-        this.element.addEventListener(eventType, this.events[eventType], {
-          passive: eventType === "wheel",
-        });
+        this.element.addEventListener(eventType, this.events[eventType]);
     }
   }
   applyChildren(children: BaseElement<State>[]) {
--- a/app/packages/looker/src/elements/common/actions.ts
+++ b/app/packages/looker/src/elements/common/actions.ts
@@ -31,10 +31,6 @@ import {
 import { dispatchTooltipEvent } from "./util";
 import closeIcon from "../../icons/close.svg";
 
-export const IS_NOTEBOOK = new URLSearchParams(window.location.search).get(
-  "notebook"
-);
-
 const readActions = <State extends BaseState>(
   actions: ControlMap<State>
 ): ControlMap<State> => {
@@ -119,7 +115,7 @@ export const controls: Control = {
 
 export const wheel: Control = {
   title: "Zoom",
-  shortcut: IS_NOTEBOOK ? "Shift+Wheel" : "Wheel",
+  shortcut: "Wheel",
   eventKeys: null,
   detail: "Zoom in and out",
   action: () => null,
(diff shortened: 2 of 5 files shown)


brimoor:
Ah, shift key within notebooks.

One observation is that the grid view "captures" the user's scroll events as they scroll within their browser, so perhaps we should follow the same pattern in expanded view and allow `wheel` rather than `Shift + wheel`?

Truth be told, it may have been better to require `Shift + wheel` to scroll the grid in Jupyter notebooks from the start. But, it might cause confusion if we made that change a this point, unless we adopted the Google Maps approach of throwing up a `Use CMD + Wheel to zoom` overlay when the user is scrolling on top of an embedded map (importantly, I believe that is done without "capturing" the scroll event; the user still gets to scroll down the page).

What are your thoughts on how best to proceed?

nit: for any shift shortcuts that remain, can you please put spaces before/after the `+` in the help menu, e.g., `Shift + wheel`.

benjaminpkane:
Hmm, I would be fine with leaving the grid as is for now. But I see your point about consistency, etc.

Scroll handling would ideally be accounted for in all scrollable containers (sidebars, plots, etc.), which could be cumbersome to maintain/stand up quickly. I would be happy with making an for issue with `needs discussion`.

brimoor:
Since `esc` works, it would feel weird if `shift` were required for the arrow keys. So I like removing shift requirement for everything except possibly scrolling.

@benjaminpkane Any thoughts on `wheel` vs `shift + sheel` for zoom in/out and the grid?

I can see how it would be annoying if scrolling down a notebook would suddenly start zooming into an image, but that's also true for the grid view and we've been allowing that...

brimoor:
Can confirm that pressing the left/right arrow buttons in Jupyter does now work.

benjaminpkane:
So I removed all of the shift requirements, but note that I can't prevent the notebook from scrolling when a user zooms. The App is an iframe, and has no control over what the parent page does.

So we should add something in the App or documentation about using shift to prevent the notebook itself from scrolling when they use arrow keys or zoom with the mouse wheel.

brimoor:
> So I removed all of the shift requirements, but note that I can't prevent the notebook from scrolling when a user zooms. The App is an iframe, and has no control over what the parent page does.
> 
> So we should add something in the App or documentation about using shift to prevent the notebook itself from scrolling when they use arrow keys or zoom with the mouse wheel.

It's nice that `shift + up/down arrow` is still an option to prevent the browser window from scrolling. For me on macOS in Chrome, `shift + wheel` does not (and did not previously) prevent my browser window from scrolling. So zooming in/out always causes a bit of jitter in my browser. Are you able to completely prevent your browser from scrolling? Is this a shift vs ctrl vs cmd thing?

benjaminpkane:
> Are you able to completely prevent your browser from scrolling in your setup with shift? If so, is this a shift vs ctrl vs cmd thing?

Yes, shift disables all scroll events in chrome on my ubuntu machine.

brimoor (approved):
LGTM. I added a note about shift key to the docs:

Screen Shot 2021-08-16 at 2 20 18 PM

brimoor:
> > Are you able to completely prevent your browser from scrolling in your setup with shift? If so, is this a shift vs ctrl vs cmd thing?
> 
> Yes, shift disables all scroll events in chrome on my ubuntu machine.

Hmm okay, mileage may vary across browser + OS I guess.

PR #1186, opened 2021-08-12
fixed loading yolov5 yaml with multiply paths

imyhxy:

This commit fixed loading yolov5 yaml with multiply paths for one dataset split. 

The change is minimum, and I have tested it with my own yolov5 yaml.

The yolov5 yaml supports using a list of paths in each data split. e.g.

```yaml
train: [/path/to/split/one,
        /path/to/split/two]
test: /path/to/test/set
```
Current yolov5 yaml load function will  raise `AttributeError` as follow:

```python
AttributeError: 'list' object has no attribute 'endswith'
```

        notes for FiftyOne users.

-   [ ] Other


brimoor (approved):
LGTM. Thanks for the fixes!

PR #1188, opened 2021-08-12
Adding a group_by() view stage

brimoor:
Adds a `group_by()` stage that allows for reorganizing the samples in a collection so that they are grouped by a specified field or expression. The groups can also be sorted.

A use case for this came up last night as I was working on brain stuff; happy to discuss offline if anyone is interested.

A note of warning here is that, since the `$group` aggregation used here stores a list of sample documents in each group, this stage could easily produce documents internally in the MongoDB pipeline that exceed 100MB and must therefore leverage the `useDiskSpace = true` option (which we have enabled). So, for very large collections with large groups, this stage could be quite intensive. But, I didn't see any easy way to avoid this when implementing the stage. Anybody know if I missed any easy solution?

Example usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Take a random sample of 1000 samples and organize them by ground
# truth label with groups arranged in decreasing order of size
view = dataset.take(1000).group_by(
    "ground_truth.label",
    sort_expr=F().length(),
    reverse=True,
)

print(view.values("ground_truth.label"))
print(
    sorted(
        view.count_values("ground_truth.label").items(),
        key=lambda kv: kv[1],
        reverse=True,
    )
)
```

### DIFF
--- a/fiftyone/__public__.py
+++ b/fiftyone/__public__.py
@@ -139,6 +139,7 @@ from .core.stages import (
     LimitLabels,
     GeoNear,
     GeoWithin,
+    GroupBy,
     MapLabels,
     Match,
     MatchFrames,
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -2937,6 +2937,56 @@ class SampleCollection(object):
         )
 
     @view_stage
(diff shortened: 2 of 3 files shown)


benjaminpkane (approved):
LGTM

PR #1189, opened 2021-08-12
Interactive plot selection improvements

brimoor:
This PR enhances the behavior of selection callbacks when using `InteractivePlot` instances connected to sessions, such as confusion matrices and embeddings visualizations.

Specifically, `InteractivePlot` has a new `selection_mode` property that can be set by the user to change the behavior of App updates when selections are made in a connected plot. The settings are:

```py
    @property
    def selection_mode(self):
        """The current selection mode of the plot.

        Only applicable when ``link_type == "labels"``.

        This property controls how the current view is updated in response to
        updates from :class:`InteractivePlot` instances that are linked to
        labels:

        -   ``"select"``: show only the selected labels
        -   ``"match"``: show unfiltered samples containing the selected labels
        -   ``"patches"``: show the selected labels in a patches view

        .. note::

            ``"patches"`` mode is only supported if an ``init_patches_fcn`` was
            provided when constructing this plot.
        """
```

Depending on the use case, any of `select` vs `match` vs `patches` may be more desirable behavior. The best way to understand this is probably just to run the attached Jupyter notebook and see for yourself.

This PR also changes the default behavior of interactive confusion matrices and object embeddings plots to use patches mode by default. This means that:
- By default, clicking on cells in a confusion matrix will show the corresponding *evaluation patches* in the App (`patches` mode), rather than entire samples with only the selected objects visible (`select` mode)
- By default, selecting scatter points in an object embeddings visualization will show the corresponding *object patches* in the App (`patches` mode), rather than entire samples with only the selected objects visible (`select` mode)

**NOTE**
- This PR includes a change to the implementation of the `SortBySimilarity` stage that requires the corresponding PR from the brain in order to properly function

**TODO** (will be done in a follow-up PR)
- Add documentation for `selection_mode` to user guide
- Add updated documentation for similarity brain page to user guide
- Update similarity tutorial

Other changes
- Refactors the functionality of `ClassificationResults` into a `BaseEvaluationResutls` class so that it is no longer the case that `DetectionsResults` and `SegmentationResults` are subclasses of `ClassificationResults`, which doesn't make sense

Example notebook demonstrating usage:
[plot_selection_tests.ipynb.zip](https://github.com/voxel51/fiftyone/files/6976764/plot_selection_tests.ipynb.zip)


brimoor:
This PR's functionality is used by the CVAT tutorial from https://github.com/voxel51/fiftyone/pull/1127, which is now merged, so I'm going to go ahead and merge this one, too.

PR #1193, opened 2021-08-13
App list fields filtering

### LINKED ISSUE #1067: [FR] Add support for visualizing + filtering list fields in the App
We should support list fields in the App. Currently, `ListField` is unsupported, and a workaround (for string lists) is to load them as a `Classifications` label field.

Now, the question is: how should we handle typed (eg numeric, string, boolean) vs untyped list fields?

If the `ListField` has a `subfield` declared, then I'd say we should use the appropriate widget type (eg slider for numerics) in the sidebar. The filtering operation would match samples whose list contains one or more matches.

If the `ListField` does not have a `subfield` declared, then I'd say we should treat all the entries as generic categorical data and use the categorical-style checkbox filtering widget to handle the possible mix of strings, booleans, and numerics that may live in the untyped list. A detail here is that non-primitive typed values like dicts would likely need to be omitted.

A corollary to this App-centric feature request is that we should probably consider automatically assigning `subfield` when dynamically adding homogeneous list fields to datasets (ie remove need for `add_sample_field()` in the example below).

```py
import fiftyone as fo

sample = fo.Sample(
    filepath="image.png",
    list=["list", "of", "strings"],
    cls=fo.Classifications(
        classifications=[
            fo.Classification(label="list"),
            fo.Classification(label="of"),
            fo.Classification(label="strings"),
        ]
    )
)

dataset = fo.Dataset()

# currently necessary to declare that `list` can only contain strings
dataset.add_sample_field("list", fo.ListField, subfield=fo.StringField)

dataset.add_sample(sample)

```

Screen Shot 2021-06-11 at 12 57 55 PM

benjaminpkane:
Resolves #1067

Changes:
* Adds list field filtering in the grid for `ListField`s with a defined `subfield` of type `BooleanField`, `FloatField`, `IntField` or `StringField`
* Adds automatic subfield typing when expanding a dataset's schema and the first list instance seen meets the one of the above primitive types (i.e. every item is of the same primitive type)
* Adds support for listing the items of a `ListField` in the expanded sample view
* Improves the performance of the `DatasetStatistics` aggregations by a factor of ~2x

```py
import fiftyone as fo
from random import random

for sample in dataset:
    sample["list_ints"] = [int(random() * 10) for i in range(5)]
    sample["list_strs"] = sample.filepath.split("/")
    sample["list_bools"] = [True, False]
    sample["list_floats"] = [random() for i in range(5)]
    sample.save()
```

Screenshot from 2021-08-13 13-47-31
Screenshot from 2021-08-13 13-46-56


tests/unittests/dataset_tests.py
```diff
@@ -759,11 +763,29 @@ def test_expand_schema(self):
         dataset.add_sample(sample)
         schema = dataset.get_field_schema()
 
+        # Scalars
```
brimoor, line 766: Thanks for the tests!

brimoor (changes requested):
This is looking good! List field filtering is working great and the ~2x speedup on stats is a great step (would like to push it a bit further soon 😄 but this a good starter)

One bug I noticed is that the App crashes when trying to load the `Scalars` graph tab on any dataset (likely related to list field handling).

Couple small points:
- How about adding a `+` sign as  in `+4 more` to the list field tags that summarize the extra list field values?
- I'm not really liking `PRIMITIVES` as the section name for non-label/tag fields. It just feels like strange terminology. What about just using the more verbose `OTHER FIELDS` and `UNSUPPORTED FIELDS` for those two sections?
- The name of the `Scalars` graph tab needs to be updated to match whatever is chosen in previous bullet
- Remove `Currently` from `Currently unsupported fields` tooltip in App sidebar. No need to imply any future work in the UI 😅

brimoor:
One observation is that, with the current implementation of list field type inference, if the first sample added to a dataset happens to have `list_field=[]`, then an untyped list field will be added, even if later samples have non-empty lists that would imply an appropriate `subfield` for the list.

This type of issue is that the reason that I added [this check](https://github.com/voxel51/fiftyone/blob/629059e410d246e3da4fd722467bc8e9e9bed6f4/fiftyone/core/dataset.py#L4079-L4081) to skip None-valued fields when expanding the schema of a dataset. A consequence of that check is that, if every sample in a batch has a `None` value, then no field will be added, which is okay because `None` values are stripped from sample dicts before inserting them into the DB and thus no schema violation will occur when data is added.

If we were to add similar "lazy subfield inference" for list fields, then I guess we would have to also strip `list_field=[]` from samples when `list_field` has not yet been defined, which probably goes too far because `None != []`...

Any ideas?

I would be okay with leaving it how it is currently implemented, since users have tools to declare/correct field type issues after dataset creation. But, if there is a way to improve the type inference, I'm onboard to make it "do the right thing" as much as possible.

benjaminpkane:
Well `[]` is falsey in python and I'll note the following behavior:
```
import fiftyone.zoo as foz

view = dataset.limit(1)
sample = view.first()
sample["test"] = "value"
sample.save()

view = view.select_fields("test")
print(next(view._aggregate())) # test has "value"

sample.test = None
sample.save()

print(next(view._aggregate())) # test field no longer in document
```

So noting the above I don't think `[] == None` is very drastic.

benjaminpkane:
I've fixed an unlisted issue in this work. Previously, larger datasets did not have a count value for some selected values in string field filters. Specifically, the statistics do not carry every count value for say, 10,000 filepaths, only the first 200.

I've added some logic to carry over the count value from search results so when a value is selected, it will have a count.

Before:
Screenshot from 2021-08-16 11-46-45

After:
Screenshot from 2021-08-16 11-46-05

brimoor:
> Well `[]` is falsey in python and I'll note the following behavior:
> 
> ```python
> import fiftyone.zoo as foz
> 
> dataset = foz.load_zoo_dataset("quickstart")
> 
> view = dataset.limit(1)
> sample = view.first()
> sample["test"] = "value"
> sample.save()
> 
> view = view.select_fields("test")
> print(next(view._aggregate())) # test has "value"
> 
> sample.test = None
> sample.save()
> 
> print(next(view._aggregate())) # test field no longer in document
> ```
> 
> So noting the above I don't think `[] == None` is very drastic.

Hmm I'm not following what the conclusion from the above code is. Yes, `None` and missing being completely interchangeable is an intentional thing in the data model, but `None` and `[]` aren't interchangeable- eg `sample.list_field.append()` only works for `[]`, which users might count on if they intentionally defaulted a sample's list field value to `[]`. It's worth noting that our default list field `tags` is configured to be non-nullable (`sample.tags = None` will compile but you'll still have `sample.tags == []` after), but all non-default fields are nullable, so I don't think we should change the default behavior of all `ListField` to be non-nullable.

tl dr; I think I'm just okay with how it currently is. Worst case, user might get a `ListField` when they wanted a `ListField(StringField)`, and, if that happens, they'll need to fix that manually.

brimoor (approved):
LGTM. Can you look into these two things before merging?
- I like the `OTHER FIELDS` change. Can you also update the graphs tab to show `Other fields` instead of just `Other`?
- A bug I'm noticing on this branch ~~(have not cross-referenced to see if this is true on `develop` as well)~~ is that my `SAMPLE TAGS` sidebar section is always showing `No sample tags` even when my dataset has tags

Issue #1192, opened 2021-08-13
[BUG] Deleting dataset fields does not update DatasetDoc


brimoor:
In the example below, I `dataset.delete_frame_field("off_limits")` on my dataset, and then the field does not appear to be in my dataset's schema anymore:

```
Name:        short-clips-v4
Media type:  video
Num samples: 150
Persistent:  True
Tags:        []
Sample fields:
    id:       fiftyone.core.fields.ObjectIdField
    filepath: fiftyone.core.fields.StringField
    tags:     fiftyone.core.fields.ListField(fiftyone.core.fields.StringField)
    metadata: fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.metadata.Metadata)
Frame fields:
    id:           fiftyone.core.fields.ObjectIdField
    frame_number: fiftyone.core.fields.FrameNumberField
    detections:   fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.labels.Detections)
```

However, it still exists in `dataset._doc.frame_fields`, which causes confusion for the App, exporting, etc.

brimoor:
update: not the App's fault

PR #1194, opened 2021-08-14
Fixing Issue #1192

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1192.

The issue was that the `name` of renamed fields was being updated in `DatasetDocument`, but not its corresponding `db_field`.

Also adds an easier to read error when trying to rename a dataset to a name that is already in use, and fixes a bug with `Dataset.add_archive()` where archives were being extracted to `/path/to/dir/dir` rather than `/path/to/dir`.

Mea culpa, there was (and still is no) unit test that checked for the issue that gave rise to https://github.com/voxel51/fiftyone/issues/1192, but I manually verified that things are now correct.


benjaminpkane (approved):
LGTM!

PR #1198, opened 2021-08-16
Fix initial patch zoom for grid thumbnails

benjaminpkane:
On `develop`, patches are not set in the thumbnails. Not an issue in v0.12.0.

Screenshot from 2021-08-16 13-09-58


brimoor (approved):
LGTM

PR #1208, opened 2021-08-19
App JSON improvements

### LINKED ISSUE #1199: [FR] Don't condense sample JSON format in App
When viewing JSON for a sample of a dataset with a label list field like `Detections`, I observe that the JSON in the App doesn't quite match the actual structure of the document:

The structure of the document is `dataset.first().to_dict()`:

```
{'filepath': '/Users/Brian/fiftyone/quickstart/data/000880.jpg',
 'tags': ['validation'],
 'metadata': None,
 'ground_truth': {'_cls': 'Detections',
  'detections': [{'_id': {'$oid': '5f452471ef00e6374aac53c8'},
    '_cls': 'Detection',
    'attributes': {'area': {'_cls': 'NumericAttribute',
      'value': 73790.37944999996},
     'iscrowd': {'_cls': 'NumericAttribute', 'value': 0.0}},
    'tags': [],
    'label': 'bird',
    'bounding_box': [0.21084375, 0.0034375, 0.46190625, 0.9442083333333334],
    'eval': 'tp',
    'eval_id': '5f452c60ef00e6374aad9394',
    'eval_iou': 0.8575063187115628},
    ...
```

but I see this, where `detections` subfield has been deleted and `ground_truth` is directly a list field:

Screen Shot 2021-08-16 at 3 44 12 PM

I'm not sure if this was intentionally done for JSON mode to abbreviate things, or if the App converts to this format for its own reasons and this internal format is what is being shown. But, in either case, I think it is preferable to print the actual document structure, because, for example, `ground_truth.detections.label` will be required when writing aggregations, not `ground_truth.label`.

A related note here is that `ObjectId`s are always printed as strings, which is good, because the ability to work with these IDs as if they are strings is something that we intentionally support in the data model. However, for consistency, it would be desirable if `Label._id` were shown as `Label.id`, to match `Sample.id`. A few related observations here:
- `Frame._sample_id` is hidden from `print(frame)`, so it should be hidden from JSON view in App, too
- Patches views introduce `_sample_id (ObjectId)` which is exposed as `sample_id (str)` in the data model
- Frame patches views introduce both `_sample_id (ObjectId)` and `_frame_id (ObjectId)` that are exposed as `sample_id (str)` and `frame_id (str)`.

The way we're handling `_field (ObjectId) -> field (str)` is definitely a bit annoying to manually manage everywhere... fully admit that

benjaminpkane:
Resolves #1199 

Changes:
* always translate `_id` to `id` for relevant for labels and samples
* Do not show `_sample_id` within frames in video JSON
* Only show `shape` for masks
* Translate other underscore fields if they have a defined field mapping (e.g. `_media_type` to `media_type`)


app/packages/looker/src/worker.ts
```diff
@@ -169,7 +186,8 @@ const getSendChunk = (uuid: string) => ({
     let buffers: ArrayBuffer[] = [];
 
     value.frames.forEach((frame) => {
-      buffers = [...buffers, ...processMasks(frame)];
+      delete frame._sample_id;
```
brimoor, line 189: Can this be easily generalized? The idea is to exclude all private fields except ones that we're manually opting into showing (for both sample- and frame-level fields).

For samples, `_rand` is currently the only private field we're not showing, and `_sample_id` is the only frame-level field, but presumably more will come in the future.

brimoor (changes requested):
The JSON changes LGTM here.

However, I noticed that the dashed line for selected labels is broken (look-and-feel doesn't change upon selection), so requesting changes rather than approving as there may be something subtly broken here that may be causing other issues I didn't happen to observe.

app/packages/looker/src/worker.ts
benjaminpkane: I'll put it on the backend, but really this goes back to a data model overhaul. The App is a client, so it shouldn't have to modify representation ever.

benjaminpkane: Ok, this line was unnecessary. Pardon the delay.

I think `filterSample` in [`index.ts`](https://github.com/voxel51/fiftyone/pull/1208/files#diff-d526c437fbb7dd06767fb2fb657232d0354eccffee28c5ef4536871d23d0e0b6),  which handles `_sample_id`,  is fairly robust and general. It leverages the fields maps (frame and sample). So I'm thinking it's fine to leave this in `Looker` until some more server work is done.

brimoor: 👍

brimoor (changes requested):
The label selection in the expanded sample modal is working again (L&F changes), but the selection checkboxes in the grid view aren't updating properly

app/packages/looker/src/index.ts
```diff
@@ -141,6 +141,7 @@ export abstract class Looker<
       }
 
       if (eventType === "selectthumbnail") {
+        console.log(this.sample);
```
brimoor, line 144: remove log?

app/packages/looker/src/worker.ts
```diff
@@ -97,6 +97,7 @@ const processSample = ({ sample, uuid }: ProcessSample) => {
     ];
   }
 
+  mapId(sample);
```
brimoor, line 100: Hmm, I see. Are there any other places that need to be checked? Little bit scary.

Sorry about starting a waterfall of data representation changes via this JSON feature request (yes, data model improvement conversations noted)

benjaminpkane: Yes, this is very much "for now" code. But, basically, the structure is the App itself still handles samples with `_id`, and within Looker everything is handled as `id`.

My bugs, didn't help. Giving it another pass now, but things look accounted for.

brimoor (approved):
LGTM

PR #1211, opened 2021-08-20
Add Apple Silicon support and a configurable database connection

benjaminpkane:
Changes:
* Adds Apple Silicon wheels for `fiftyone-desktop` and `fiftyone-db`. `fiftyone-db` is hollow for arm64 Macs, as MongoDb is not yet available for the platform, so it is just placeholder because `fiftyone-db` is required
* Adds `database_uri` to `FiftyOneConfig` for defining your own MongoDB connection
* Corresponding documentation updates


docs/source/user_guide/config.rst
```diff
@@ -231,6 +238,23 @@ For example, you can customize your FiftyOne config at runtime as follows:
         show_progress_bars=True,
     )
 
+.. _configuring-mongodb-connection:
+
+Configuring a MongoDB connection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, FiftyOne is installed with it's own `mongod` database
```
brimoor, line 246: Question: how does the manual DB configuration now work for environments other than Apple silicon where we don't provide a prebuilt `fiftyone-db`?

The hollow `fiftyone-db` package was a nice idea for Apple silicon! Can/should that be extended for *all* unsupported environments so that everyone can always `pip install fiftyone`? The hollow package could be setup to print an error pointing to this section in the docs if it is ever asked to spin up the DB with no `database_uri`.

docs/source/user_guide/config.rst
```diff
@@ -231,6 +238,23 @@ For example, you can customize your FiftyOne config at runtime as follows:
...
+distribution. This database is managed by FiftyOne automatically as a service
+that runs whenever at least one FiftyOne Python client is alive.
+
+As an alternative to this managed database service, a valid
+`MongoDB connection string URI <https://docs.mongodb.com/manual/reference/connection-string/>`_
+can be configured via the `database_uri` setting, and FiftyOne will connect to
+your own MongoDB instance instead. FiftyOne requires MongoDB version `4.4`.
```
brimoor, line 253: Are there any concerns with admin DB migrations when the user is using their own database? IE might they fail due to permissions, etc?

If so, should this be documented here?

docs/source/user_guide/config.rst
benjaminpkane: The only wheels we are missing at this point are linux i686 and win32. It doesn't seem important to me to add them. They are sunset platforms.

Filling out more support for linux distributions could be done, but I'd like to leave it out here. Detecting if the default db should not be used (SUSE system), and whether or not `mongod` is available via an alternative `fiftyone-db-*` package is a separate concern (implementation wise).

I'll make an issue. And add to the docs here that we do not support 32 bit systems.

docs/source/user_guide/config.rst
benjaminpkane: Sure, I'll add a note.

docs/source/user_guide/config.rst
brimoor: How does the FiftyOne install process look for someone on a linux distribution like SUSE today?

After seeing how the Apple silicon experience was, I'm just a bit confused on how people on unsupported distributions were able to follow the instructions we used to provide to get running in the first place.

benjaminpkane: They would have no issue installing `fiftyone`. `pip` does not distinguish linux distributions. I only mention SUSE because it is listed [here](https://www.mongodb.com/try/download/community). But for example, I am on Ubuntu 20.04 and the default 18.04 `fiftyone-db` binary works fine for me.

So, `mongod` might fail for a SUSE user. I need to check. We need to add a `fiftyone-db-suse` in that case.

brimoor: Ah I see. So basically everyone with a 64 bit machine can `pip install fiftyone` after this PR, but users in unsupported linux environments (for example) might encounter an error when trying to fire up the database.

But this PR reduces the priority for a `fiftyone-db-suse` package (for example) **IF** the logging message that occurs when the DB service fails to start points the user to your docs on manual DB setup

fiftyone/core/odm/database.py
```diff
@@ -7,42 +7,130 @@
...
+            raise ConnectionError("Could not connect to `mongod`") from e
+
+        raise RuntimeError("Failed to validate `mongod` version") from e
+
+    min_ver, max_ver = foc.MONGODB_VERSION_RANGE
+
+    if min_ver <= version < max_ver:
+        return
+
+    if version >= max_ver and config.database_uri is not None:
```
brimoor, line 122: Why is the "try upgrading FiftyOne" recommendation added specifically in the case when `database_uri is not None` here?

In general, is the idea that the user might have previously installed a newer version of FiftyOne that did an admin migration but they since reverted?

benjaminpkane: Yes, if they connect to a mongo 5.0 instance (after we add support/transition) and they downgrade to 0.12.0 for some reason.

It's a small case, I suppose. In the near term, I think it will be more likely that their mongo version is too old.

brimoor: I see. Wouldn't this apply equally to FiftyOne-managed dbs and `database_uri` dbs though? Both should have up/down admin migrations run whenever the relevant fiftyone versions are installed.

I'm latching onto this primarily to make sure I understand the change, and this special case just confused me. If the "try upgrading" message were printed in all cases, I wouldn't have commented in the first place.

benjaminpkane: This validation is using the version of the mongo binary. `fiftyone-db` should always be carrying the correct one.

This check has nothing to do with admin migrations. Upgrading and downgrading `fiftyone` should always upgrade/downgrade `fiftyone-db` accordingly, no?

brimoor: I see, I was thinking about things like `db.admin.command({"setFeatureCompatibilityVersion": "4.4"})` in our (only) admin migration, but that doesn't actually change the binary version, of course.

brimoor (approved):
LGTM. This is working for me, so if you're satisfied that the HOLD issues you mentioned are resolved, then merge away!

PR #1212, opened 2021-08-20
Resolves #1178

### LINKED ISSUE #1178: [FR] Validate desktop package version when launching desktop App
Suppose I am a civilian desktop App user of `fiftyone==0.11.2.1`, which I installed as follows:

```shell
pip install fiftyone==0.11.2.1
pip install 'fiftyone[desktop]'
```

Then I upgrade my FiftyOne package [as instructed](https://voxel51.com/docs/fiftyone/getting_started/install.html#upgrading-fiftyone):

```shell
pip install --upgrade fiftyone
```

If I then try to launch the App:

```shell
fiftyone quickstart --desktop
```

I see an unhandled error like so:

```
An exception has been raised by the server. Your session has been reverted to its previous state.

  File "main.py", line 272, in wrapper
  File "main.py", line 516, in on_message
    event = getattr(self, "on_%s" % message.pop("type"))
AttributeError: 'StateHandler' object has no attribute 'on_page'
```

The solution is to upgrade my `fiftyone-desktop` package too, but, as a civilian, I didn't know that I needed to do that and the error message above doesn't help.

Proposals:
- Should we manually verify that the desktop package has a compliant version before launching the App, and, if not, print a helpful error message telling the user what to do? And/or, is there something more automatic we can do to always keep the desktop package in sync?
- And/or, add `fiftyone-desktop` upgrade instructions to [upgrade section](https://voxel51.com/docs/fiftyone/getting_started/install.html#upgrading-fiftyone)?

Related note: 

The [install guide](https://voxel51.com/docs/fiftyone/getting_started/install.html#fiftyone-desktop-app) shows `pip install fiftyone-desktop`, but `pip install 'fiftyone[desktop]'` would probably be better because the latter will install the proper desktop version for your `fiftyone` install, while the former seems to always install the latest version of the desktop package.

ErfanTagh:
Resolves https://github.com/voxel51/fiftyone/issues/1178.


fiftyone/core/session.py
```diff
@@ -27,6 +27,7 @@
 import fiftyone.utils.templates as fout
 from fiftyone.core.state import StateDescription
 
+import pkg_resources
```
benjaminpkane, line 30: Standard library imports go in the first block of imports. After `import logging` in this case.

fiftyone/core/session.py
```diff
@@ -356,6 +350,40 @@ def __init__(
             self.open()
             return
 
+    def import_desktop():
+
+        """Imports the `fiftyone-desktop` package and verifies that the correct
+        version is installed.
+        """
```
benjaminpkane, line 357: You can make this a private method in the module (outside of `Session`) at the end of `session.py` after
`def _display_colab():`. No docstring is necessary.

package/desktop/setup.py
```diff
@@ -187,5 +187,6 @@ def write_wheelfile(self, *args, **kwargs):
         "Programming Language :: Python :: 3.9",
     ],
     python_requires=">=3.6",
+    install_requires = ['fiftyone'],
```
benjaminpkane, line 190: You can remove this.
```suggestion
```

fiftyone/core/session.py
benjaminpkane:
```py
def _import_desktop():
    try:
        import fiftyone.desktop
    except ImportError as e:
        raise ValueError(
            "You must `pip install fiftyone[desktop]` in order to launch the "
            "desktop App"
        ) from e

    # Get `fiftyone-desktop` requirement for current `fiftyone` install
    fiftyone_dist = pkg_resources.get_distribution("fiftyone")
    requirements = fiftyone_dist.requires(extras=["desktop"])
    desktop_req = [r for r in requirements if r.name == "fiftyone-desktop"][0]

    desktop_dist = pkg_resources.get_distribution("fiftyone-desktop")

    if not desktop_req.specifier.contains(desktop_dist.version):
        raise ValueError(
            "fiftyone==%s requires fiftyone-desktop%s, but you have "
            "fiftyone-desktop==%s installed.\n"
            "Run `pip install fiftyone[desktop]` to install the proper "
            "desktop package version" % (
                fiftyone_dist.version,
                desktop_req.specifier,
                desktop_dist.version
            )
        )
```

benjaminpkane (changes requested):
Thanks @ErfanTagh. A couple comments.

fiftyone/core/session.py
ErfanTagh: Thanks for the comments, I applied them

fiftyone/core/session.py
```diff
@@ -356,6 +350,9 @@ def __init__(
             self.open()
             return
 
+    
+        
+
```
benjaminpkane, line 355:
```suggestion
    
        
        
```

fiftyone/core/session.py
```diff
@@ -340,14 +341,7 @@ def __init__(
...
 
-            try:
-                import fiftyone.desktop  # pylint: disable=unused-import
-            except ImportError as e:
-                if not focn.DEV_INSTALL:
-                    raise ValueError(
-                        "You must install the 'fiftyone-desktop' package "
-                        "in order to launch a desktop App instance"
-                    ) from e
+            import_desktop()
```
benjaminpkane, line 344:
```suggestion
            _import_desktop()
```

fiftyone/core/session.py
```diff
@@ -1074,3 +1071,32 @@ def capture(img, width):
             )
 
     output.register_callback("fiftyone.%s" % uuid.replace("-", "_"), capture)
+
+
+def import_desktop():
```
benjaminpkane, line 1076:
```suggestion
def _import_desktop():
```

benjaminpkane (approved):

benjaminpkane (changes requested):
@ErfanTagh 

Can install the dev environment and make sure `pre-commit` is installed. There are still some formatting issues.

```
  pip install -r requirements/dev.txt
  pre-commit install
  pip install -e .
```

Or

```
bash install.bash -d
```

And then remove those 3 lines you added. Some more formatting will occur.

fiftyone/core/session.py
```diff
@@ -340,14 +341,7 @@ def __init__(
                     "Cannot open a Desktop App instance from a Colab notebook"
                 )
 
-            try:
-                import fiftyone.desktop  # pylint: disable=unused-import
```
brimoor, line 344: @benjaminpkane in the current implementation, I think this is allowing a release desktop package to be used when running a source install. I don't see much use for that because we can't guarantee that any app build will work with the bleeding edge of any branch.

Should we just update so `_import_desktop()` is never called if `focn.DEV_INSTALL == True`?

benjaminpkane: Yes, ok

brimoor (approved):
Hi @ErfanTagh thanks for taking this on 💪 

We would like to include this in a release today, so I'm going to go ahead and approve + merge it and we'll handle the small cleanup items in a separate PR.

PR #1213, opened 2021-08-22
Adding annotation page to user guide

brimoor:
Adds an `Annotating Datasets` page to the user guide that explains the annotation API in a backend-agnostic fashion. This includes a section at the end about implementing a custom annotation backend.

This also contributes some other bits
- Fixed a bug where attributes like `confidence` would be overwritten by None when they weren't explicitly included as an attribute to be annotated
- Allows `label_type` to be a `Label` or scalar `Field` rather than a string, if desired
- Updates `annotate()` to support patches views, and raises an error is empty or frames views are passed, which are not supported
- Added support for importing/exporting/cloning datasets with annotation runs

I also did an editing pass over the annotation tutorial:
- Found some old `"default_value"` syntax which needed to be updated to just `"default"`. Also found a few other typos
- Updated the thumbnail image for the tutorial. The card template used by the docs always displays thumbnail images as squares, so the source image needs to be square to avoid deformation
- Cleaned up the verb tense a bit. Active verbiage like "Let's do XX" is preferable for tutorials, as supposed to "XX can be done". And, either way, consistency is key

Finally, there is some cosmetic cleanup of the code:
- Removing the `set_*config_settings()` methods from `fiftyone.core.config`, which were pretty silly. Cleaner to just directly edit the relevant config objects.
- Uses `iter_samples(progress=True)` rather than an explicit `ProgressBar` usage, for brevity
- Addressing pylint warnings in affected modules


benjaminpkane (approved):
LGTM

PR #1219, opened 2021-08-24
Patch release v0.13.1

benjaminpkane:
Resolves:
* `id` bubbles are not displayed when they are checked

### DIFF
--- a/app/packages/looker/src/elements/common/tags.ts
+++ b/app/packages/looker/src/elements/common/tags.ts
@@ -95,11 +95,17 @@ export class TagsElement<State extends BaseState> extends BaseElement<State> {
               value: label,
             })),
         ];
-      } else if (
-        fieldsMap &&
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM. Thanks for the release notes 🥇

fiftyone/migrations/revisions/v0_13_0.py
```diff
@@ -36,7 +36,7 @@ def down(db, dataset_name):
         config = run_doc.get("config", {})
         config_cls = config.get("cls", "")
         if config_cls == "fiftyone.brain.similarity.SimilarityConfig":
-            config.pop("model")
-            config.pop("metric")
+            config.pop("model", None)
```
brimoor, line 39: This shouldn't ever be important, but I happened to be working with a dataset in a broken state where this was needed, so I thought I'd throw it in anyway.

I prefer to write the migrations as fail-safe as possible

PR #1222, opened 2021-08-25
Adding support for datetime fields

brimoor:
Adds support for storing date information in FiftyOne datasets.

Timezones and date-based queries are naturally supported 🎵👂 

### TODO

- [ ] Add App-side support for filtering based on datetime fields

### Timezone example

Timezone handling is pretty clean:

```py
from datetime import datetime, timedelta
import pytz

import fiftyone as fo


dataset = fo.Dataset()

# These are all the epoch
date1 = datetime(1970, 1, 1, 0, 0, 0)
utcdate1 = date1.replace(tzinfo=pytz.utc)
date2 = utcdate1.astimezone(pytz.timezone("US/Eastern"))
date3 = utcdate1.astimezone(pytz.timezone("US/Pacific"))

# FiftyOne treats naive datetimes as UTC implicitly
sample1 = fo.Sample(filepath="image1.png", date=date1)

# FiftyOne accepts timezone-aware datetimes too
sample2 = fo.Sample(filepath="image2.png", date=date2)
sample3 = fo.Sample(filepath="image3.png", date=date3)

dataset.add_samples([sample1, sample2, sample3])

fo.config.timezone = "US/Eastern"
dataset.reload()

print(sample1.date.tzinfo.zone == "US/Eastern")
print((sample1.date - utcdate1).total_seconds())
print((sample1.date - sample2.date).total_seconds())
print((sample1.date - sample3.date).total_seconds())

fo.config.timezone = None
dataset.reload()

print(sample1.date.tzinfo is None)
print((sample1.date - date1).total_seconds())
print((sample1.date - sample2.date).total_seconds())
print((sample1.date - sample3.date).total_seconds())
```

### Dataset view example

And of course, support for date-based dataset views is there, too!

```py
from datetime import datetime, timedelta

import fiftyone as fo

dataset = fo.Dataset()
dataset.add_samples(
    [
        fo.Sample(
            filepath="image1.png",
            creation_date=datetime(2021, 8, 24, 1, 0, 0),
        ),
        fo.Sample(
            filepath="image2.png",
            creation_date=datetime(2021, 8, 24, 2, 0, 0),
        ),
        fo.Sample(
            filepath="image3.png",
            creation_date=datetime(2021, 8, 24, 3, 0, 0),
        ),
    ]
)

query_date = datetime(2021, 8, 24, 2, 1, 0)
query_delta = timedelta(minutes=30)

# Samples with creation date after 2021-08-24 02:01:00
view = dataset.match(F("creation_date") > query_date)
print(view)

# Samples with creation date within 30 minutes of 2021-08-24 02:01:00
view = dataset.match(abs(F("creation_date") - query_date) < query_delta)
print(view)
```

### DIFF
--- a/docs/source/user_guide/config.rst
+++ b/docs/source/user_guide/config.rst
@@ -292,7 +292,7 @@ Configuring a timezone
 ----------------------
 
 By default, FiftyOne loads all datetimes in FiftyOne datasets as naive
-`datetime` objects in UTC time.
+`datetime` objects expressed in UTC time.
 
 However, you can configure FiftyOne to express datetimes in a specific timezone
 by setting the `timezone` property of your FiftyOne config.
@@ -334,8 +334,8 @@ Or, you can even dynamically change the timezone while you work in Python:
 
 .. note::
 
-    This setting does not affect the internal database representation of
-    datetimes, which are always stored in UTC timestamps.
+    The `timezone` setting does not affect the internal database representation
+    of datetimes, which are always stored as UTC timestamps.
 
 .. _configuring-fiftyone-app:
 
--- a/docs/source/user_guide/config.rst
+++ b/docs/source/user_guide/config.rst
@@ -24,6 +24,10 @@ FiftyOne supports the configuration options described below:
 |                               |                                     |                               | specifying a custom MongoDB database to which to connect. See                          |
 |                               |                                     |                               | :ref:`this section <configuring-mongodb-connection>` for more information.             |
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
+| `timezone`                    | `FIFTYONE_TIMEZONE`                 | `None`                        | An optional timzone string. If provided, all datetimes read from FiftyOne datasets     |
+|                               |                                     |                               | will be expressed in this timezone. See :ref:`this section <configuring-timezone>` for |
+|                               |                                     |                               | more information.                                                                      |
++-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
 | `dataset_zoo_dir`             | `FIFTYONE_DATASET_ZOO_DIR`          | `~/fiftyone`                  | The default directory in which to store datasets that are downloaded from the          |
 |                               |                                     |                               | :ref:`FiftyOne Dataset Zoo <dataset-zoo>`.                                             |
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
@@ -100,6 +104,7 @@ and the CLI:
         {
             "database_dir": "~/.fiftyone/var/lib/mongo",
             "database_uri": null,
+            "timezone": null,
             "dataset_zoo_dir": "~/fiftyone",
             "dataset_zoo_manifest_paths": null,
             "default_app_config_path": "~/.fiftyone/app_config.json",
@@ -135,6 +140,7 @@ and the CLI:
         {
             "database_dir": "~/.fiftyone/var/lib/mongo",
             "database_uri": null,
+            "timezone": null,
             "dataset_zoo_dir": "~/fiftyone",
             "dataset_zoo_manifest_paths": null,
             "default_app_config_path": "~/.fiftyone/app_config.json",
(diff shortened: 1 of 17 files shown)


docs/source/user_guide/config.rst
```diff
@@ -24,6 +24,10 @@ FiftyOne supports the configuration options described below:
 |                               |                                     |                               | specifying a custom MongoDB database to which to connect. See                          |
 |                               |                                     |                               | :ref:`this section <configuring-mongodb-connection>` for more information.             |
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
+| `timezone`                    | `FIFTYONE_TIMEZONE`                 | `None`                        | An optional timzone string. If provided, all datetimes read from FiftyOne datasets     |
```
benjaminpkane, line 27: Should the default be `local`?

brimoor: My thinking was that timezones are pretty error prone, especially for teams working across different timezones, and defaulting to UTC was my way of suggesting to the world that UTC is a good convention to use

benjaminpkane:
Ready for :eyes: 

```py
from datetime import datetime, timedelta as d
import fiftyone as fo


now = datetime.now()


    sample["ms"] = now - d(milliseconds=idx)
    sample["seconds"] = now - d(seconds=idx)
    sample["minutes"] = now - d(minutes=idx)
    sample["hours"] = now - d(hours=idx)
    sample["days"] = now - d(days=idx)
    sample["weeks"] = now - d(weeks=idx)
    sample.save()

session.refresh()
```

brimoor:
@benjaminpkane thanks for the handy App example code.

A couple observations:
- Are you open to `YYYY/MM/DD` format for dates? `MM/DD/YYYY` is definitely no good (silly americans), but I personally prefer large -> small to help 
- Related to above, perhaps we could provide an App setting to let the user configure their date rendering string if they prefer something different?
- I get an aggregation error when trying to view the stats graphs for your example dataset:
```
pymongo.errors.OperationFailure: The 'boundaries' option to $bucket must be sorted, but elements 0 and 1 are not in ascending order (2021-10-03T14:23:22.206Z is not less than 2021-10-03T14:23:22.206Z)., full error: {'ok': 0.0, 'errmsg': "The 'boundaries' option to $bucket must be sorted, but elements 0 and 1 are not in ascending order (2021-10-03T14:23:22.206Z is not less than 2021-10-03T14:23:22.206Z).", 'code': 40194, 'codeName': 'Location40194'}
```

I like how you're doing some conditional rendering to provide optimized filtering experiences depending on the available range of a particular datetime field. Specifically, if all dates lie in the same day, you show that separately. I noticed that for some of the intermediate cases like `minutes` and `hours` fields of your example dataset, the endpoints were a bit hard to make sense of. I'll think about a concrete recommendation a bit more, but I think we can tweak the logic a bit more to make the tooltips have a bit more granularity when it is necessary

brimoor:
okay so I'm thinking something like this gist: https://gist.github.com/brimoor/bcf8afa3b17225ed9ed528fcb6ef90fc.

My idea is to split off `YYYY/MM/DD` when all datetimes are in the same day (like you did), but then additionally show only the necessary number of "significant digits" so that the 100 increments on the slider (or whatever the number might be) will always show different values at every point.

As an implementation detail, since the above strategy effectively truncates times, the filtering should probably be implemented as `left <= value < right + delta`, where `delta` = 1 unit of the smallest time component shown in the slider. This *must* be done at the top-right endpoint, at least, or else the largest value may not be reachable by the filters.

One thing I don't like about the above approach is that, for the `hours` field, where the "significant digit" is hours and the slider spans multiple days, the tooltip would show:

```
2021/09/26 08
2021/09/26 10
2021/09/26 12
2021/09/26 14
...
```

Whenever any sub-day times at all need to be shown, I thought about just padding with zeros, eg:

```
2021/09/26 08:00:00
2021/09/26 10:00:00
2021/09/26 12:00:00
2021/09/26 14:00:00
```

but these tooltips would be quite wide and likely have too much overlap?

I also considered tweaking the "common datetime" logic to pull out as much commonality as possible, eg:

```
common: 2021/09/26 08:XX

slider:
00
02
04
...
```

This would achieve the shortest tooltips, but is probably not worth the non-standard confusion.

While typing this, I'm now thinking that, if hours/minutes/seconds of any kind are needed to provide the desired "resolution" on the tooltip, then it's probably best to just show all of them, rather than padding with zeros...

benjaminpkane:
Can we just use `/` instead of `-`. I spent an hour trying to fix it, no luck.

brimoor:
@benjaminpkane When `fo.config.timezone = None`, the backend shows everything in UTC, but the App shows local timezone. Can you update so the App also shows UTC?

Also, can you update so that the App pulls the current `fo.config.timezone` value every time it refreshes? The docs state that this is allowed:

```py
# Change timezone on-the-fly
fo.config.timezone = "US/Eastern"
dataset.reload()
```

so for consistency it would be nice if `session.refresh()` pulled that change too.

Also, one final gotcha: since `DateField` data is stored in MongoDB as timestamps and loaded by PyMongo initially as timezone-aware `datetime`s, we have to manually convert back to UTC before casting from `datetime` to `date`, otherwise the date will be incorrect. Here's how this is done in the backend: https://github.com/voxel51/fiftyone/blob/168afc8238182196c05665b5f50f2c85a08ca5d6/fiftyone/core/fields.py#L94-L101.

and I had to wire up things like aggregations to use that method as well in order for `date` fields to work correctly: https://github.com/voxel51/fiftyone/blob/168afc8238182196c05665b5f50f2c85a08ca5d6/fiftyone/core/aggregations.py#L250-L252

When viewing the example dataset in this PR in the App, dates are one day off when the App is rendering in timezones like `US/Eastern` that are behind UTC.

benjaminpkane:
> so for consistency it would be nice if session.refresh() pulled that change too.

It does

PR #1227, opened 2021-08-26
Adding a temporal detection label type

brimoor:
Adds a `TemporalDetection` label type that allows a label to be assigned a `[first, last]` frame support in a video.

Also adds support for importing/exporting datasets with `TemporalDetection(s)` label fields in a new `FiftyOneTemporalDetectionDataset` type with format documented at the bottom of this PR.

Finally, for consistency with other `FiftyOne*Dataset` formats, adds support for importing/exporting predictions confidences in `FiftyOneImageClassificationDataset` format. On the export side, this is on an opt-in basis by passing an optional `include_confidence=True` flag during export.

### Basic usage

```py
import fiftyone as fo

sample = fo.Sample(
    filepath="video1.mp4",
    events=fo.TemporalDetections(
        detections=[
            fo.TemporalDetection(label="party", support=[30, 60]),
            fo.TemporalDetection(label="meeting", support=[45, 75]),
        ]
    )
)

dataset = fo.Dataset()
dataset.add_sample(sample)

print(dataset)
print(dataset.first())
```

### Creation from timestamps

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "quickstart-video",
    max_samples=1,
    shuffle=True,
    dataset_name=fo.get_default_dataset_name()
)
dataset.compute_metadata()

# Create two events from timestamps (in seconds)
sample = dataset.first()
event1 = fo.TemporalDetection.from_timestamps(
    [1, 2], label="party", sample=sample
)
event2 = fo.TemporalDetection.from_timestamps(
    [1.5, 2.5], label="meeting", sample=sample
)
sample["events"] = fo.TemporalDetections(detections=[event1, event2])
sample.save()

print(dataset)
print(dataset.first())
```

### FiftyOneTemporalDetectionDataset format

Screen Shot 2021-10-02 at 11 25 25 AM

### DIFF
--- a/app/packages/app/src/recoil/selectors.ts
+++ b/app/packages/app/src/recoil/selectors.ts
@@ -9,6 +9,7 @@ import {
   makeLabelNameGroups,
   VALID_LIST_TYPES,
   LIST_FIELD,
+  UNSUPPORTED_IMAGE,
 } from "../utils/labels";
 import { packageMessage } from "../utils/socket";
 import { viewsAreEqual } from "../utils/view";
@@ -335,10 +336,15 @@ export const fieldSchema = selectorFamily({
   },
 });
 
-const labelFilter = (f) => {
+const getLabelFilter = (video: boolean, dimension: string) => (f) => {
+  if (!f.embedded_doc_type) {
+    return false;
+  }
+
+  const type = f.embedded_doc_type.split(".").slice(-1)[0];
   return (
-    f.embedded_doc_type &&
-    VALID_LABEL_TYPES.includes(f.embedded_doc_type.split(".").slice(-1)[0])
+    ((dimension !== "frame" && video) || !UNSUPPORTED_IMAGE.includes(type)) &&
+    VALID_LABEL_TYPES.includes(type)
   );
 };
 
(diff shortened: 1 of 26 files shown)


benjaminpkane:
Should an exception be raised when adding a `VideoClassification` to an image dataset?

brimoor:
> Should an exception be raised when adding a `VideoClassification` to an image dataset?

yes I suppose I should add some validation. This is (I believe) the first `Label` type that doesn't make sense in certain situations, like on image samples and directly on frames.

brimoor:
> > Should an exception be raised when adding a `VideoClassification` to an image dataset?
> 
> yes I suppose I should add some validation. This is (I believe) the first `Label` type that doesn't make sense in certain situations, like on image samples and directly on frames.

So we don't currently do any "contextual" validation of `Document` fields at all; validation is assumed to be completely definable by the `Field.validate()` method. Rather than building in some logic to validate fields against their parent `Document`, perhaps its cleaner just to make `VideoClassification` appear as an unsupported image-sample and frame-level field in the App for now?

I believe the App already has to keep track of such a list of things it knows it can render, and I'd like to avoid adding too much validation logic in the backend since that stuff may change as we shed MongoEngine.

benjaminpkane:
OK, the TODOs should be addressed now.

brimoor:
Okay I updated `VideoClassification` -> `TemporalDetection` and everything's working!

PR #1228, opened 2021-08-26
Clips views

brimoor:
Adds a `to_clips()` view stage that can extract clips defined by `(filepath, support)` from a video dataset.

Clip views are proper first-class citizens in the FiftyOne ecosystem:
- Clip views are `DatasetView`s, so they can be filtered, searched, etc by chaining view stages as usual
- Clip views can be visualized in the App
- Clip views can be exported via `export()` in unlabeled or labeled video dataset formats (and the corresponding clips will be extracted from the source videos if needed)
- Other methods like `apply_model()`, `compute_embeddings()`, and `draw_labels()` can be applied to clip views

See the documentation in the PR for more details on the semantics of clips views, and plenty of examples.

Also adds an optional `other_fields` parameter to the `to_patches()` and `to_evaluation_patches()` stages that allows for optionally including additional fields from the parent samples on the patches views, which can provide desirable metadata for certain operations.

### Example usages

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.compute_metadata()

sample = dataset.first()
sample["events"] = fo.TemporalDetections(
    detections=[
        fo.TemporalDetection.from_timestamps(
            [1, 2], label="party", sample=sample
        ),
        fo.TemporalDetection.from_timestamps(
            [3, 4], label="meeting", sample=sample
        ),
    ]
)
sample.save()

sample = dataset[:2].last()
sample["events"] = fo.TemporalDetections(
    detections=[
        fo.TemporalDetection.from_timestamps(
            [0.5, 1.5], label="meeting", sample=sample
        ),
        fo.TemporalDetection.from_timestamps(
            [2.5, 3.5], label="party", sample=sample
        ),
    ]
)
sample.save()

#
# Create a clips view that contains one clip for each video classification
#

clips1 = dataset.to_clips("events")
print(clips1)
print(clips1.values("events.label"))
print(clips1.count("frames"))
print(clips1.values("support"))

#
# Create a clips view that contains one clip for each contiguous segment that
# contains at least one person
#

people = F("detections.detections").filter(F("label") == "person")
clips2 = dataset.to_clips(people.length() > 0)
print(clips2)
print(clips2.values("support"))

#
# Equivalent way to achieve previous result
#

clips3 = (
    dataset
    .filter_labels("frames.detections", F("label") == "person")
    .to_clips("frames.detections")
)
print(clips3)
print(clips3.values("support"))

session = fo.launch_app(view=clips3)
```

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -4570,7 +4570,7 @@ class SampleCollection(object):
             field_or_expr: can be any of the following:
 
                 -   a :class:`fiftyone.core.labels.TemporalDetection`,
-                    :class:`fiftyone.core.labels.TemporalDetections`, or
+                    :class:`fiftyone.core.labels.TemporalDetections`,
                     :class:`fiftyone.core.fields.FrameSupportField`, or list of
                     :class:`fiftyone.core.fields.FrameSupportField` field
                 -   a frame-level label list field of any of the following
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -5663,7 +5663,7 @@ class ToClips(ViewStage):
         field_or_expr: can be any of the following:
 
             -   a :class:`fiftyone.core.labels.TemporalDetection`,
-                :class:`fiftyone.core.labels.TemporalDetections`, or
+                :class:`fiftyone.core.labels.TemporalDetections`,
                 :class:`fiftyone.core.fields.FrameSupportField`, or list of
                 :class:`fiftyone.core.fields.FrameSupportField` field
             -   a frame-level label list field of any of the following types:


benjaminpkane:
What are the rules about multiple clips stages? For example, given the above snippet:
```py
view = dataset.to_clips("events").to_clips(people.length() > 0)
```

This does not seem to behave as expected.

brimoor:
> What are the rules about multiple clips stages? For example, given the above snippet:
> 
> ```python
> view = dataset.to_clips("events").to_clips(people.length() > 0)
> ```
> 
> This does not seem to behave as expected.

Oh I didn't think about that. I'll add an exception that prevents it for now, but I might be able to get it to work. Looks like it's just the frame lookup logic that's broken. The clipping logic works.

brimoor:
> > What are the rules about multiple clips stages? For example, given the above snippet:
> > ```python
> > view = dataset.to_clips("events").to_clips(people.length() > 0)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This does not seem to behave as expected.
> 
> Oh I didn't think about that. I'll add an exception that prevents it for now, but I might be able to get it to work. Looks like it's just the frame lookup logic that's broken. The clipping logic works.

Update: supporting `to_clips().to_clips()` was actually pretty easy. That's in now (and should be allowed from the App too; why not?)

benjaminpkane:
This is almost ready. But, using the video quickstart, I'm confused why the following still shows people and signs:

```
F = fo.ViewField
(
    dataset
    .filter_labels("frames.detections", F("label") == "vehicle")
    .to_clips("frames.detections")
    .count_values("frames.detections.detections.label")
)
```

Out
```
{'person': 1108, 'vehicle': 7511, 'road sign': 2726}
```

brimoor:
> This is almost ready. But, using the video quickstart, I'm confused why the following still shows people and signs:
> 
> ```python
> (
>     dataset
>     .filter_labels("frames.detections", F("label") == "vehicle")
>     .to_clips("frames.detections")
>     .count_values("frames.detections.detections.label")
> )
> ```
> 
> Out
> 
> ```
> {'person': 1108, 'vehicle': 7511, 'road sign': 2726}
> ```

This is by design. Unlike other views, `to_clips()` only uses previous filters to select clips; but then all frame labels for the clips will be present when working with the clips views. There are some notes about this in the method docstrings and also the new docs content on clips views, but who reads docs, right?

I should clarify that when I say "by design", I don't really mean in the sense that I *wanted* it to be like this (since I'm sure your confusion will be shared by others, as this behavior is inconsistent with other view stages), but that this approach avoids the need to create a duplicate frames collection, which is better for efficiency reasons (both view creation time and subsequent view edit times).

I considered trying to copy all previous view stage logic "into" the clips view, but it's not always possible for complex views whose contents might depend on, say, the number of frames in the video. Also, the silver lining is that, if the default behavior were to be that previous filtering is respected, then users who want the current behavior (which is valid, if not always preferred) wouldn't really have a good way to achieve it...

benjaminpkane:
Ok, ready for eyes.

benjaminpkane:
Some notes:

`ListField(FrameSupportField)` are a bit of a spanner in the works (my own doing). `dataset.count("support_list")` should return the number of `FrameSupportField`s technically, while `dataset.bounds("support_list")` should unwind the tuples to compute numeric bounds...

I have not (yet) made any adjustments for `support` fields in the `Other fields` graphs tab.

In particular, comments are welcome regarding grid bubbles for support fields and video classification labels. The below shows the current behavior.

```py
import fiftyone as fo

dataset, session = fo.quickstart(video=True)
dataset.add_sample_field("support", fo.FrameSupportField)
dataset.add_sample_field("support_list", fo.ListField, subfield=fo.FrameSupportField)
session.view = dataset.limit(1)

sample = dataset.first()
sample["support"] = [1, 1]
sample["support_list"] = [[i, i] for i in range(1, 5)]
sample["event"] = fo.VideoClassification(label="event", support=[1, 1])
sample["events"] = fo.VideoClassifications(
    classifications=[
        fo.VideoClassification(label="events", support=[i, i]) for i in range(1, 5)
    ]
)
sample.save()
session.refresh()
```

brimoor:
@benjaminpkane clips in the App are working great! Here's the only things I noticed:
- Some samples vs patches vs clips localization text needs updating in the graphs tabs. In clips view, the tab reads "Patch tags" but should read "Clip tags", and the "No sample tags" text should be "No clip tags". This latter bug is likely also an issue in regular patches views
- I get an error when trying to switch between any image dataset and any video dataset in the App
- After using the support filtering capability, I think we should switch to instead only showing clips that fall completely within the specified range, eg `(F("support")[0] >= 10) & (F("support")[1] <= 20)`

fiftyone/core/collections.py
```diff
@@ -7140,13 +7267,15 @@ def _parse_field_name(
     omit_terminal_lists,
     allow_missing,
 ):
-    unwind_list_fields = set()
+    unwind_list_fields = []
+    explicit_unwinds = set()
```
brimoor, line 7271: @benjaminpkane the change you made here is breaking [this unit test](https://github.com/voxel51/fiftyone/blob/866b17a436a3399abc379b1bea4ad4fd56bf67ea/tests/unittests/view_tests.py#L1303-L1310).

What were you trying to achieve? This unwinding code is tricky; I can take a shot at supporting what you need

benjaminpkane: I want to support `ListField(FrameSupportField)` fields in the App sidebar filters. This requires computing the bounds on a nested list.

```py
import fiftyone.zoo as foz


dataset.add_sample_field("support_list", fo.ListField, subfield=fo.FrameSupportField)
sample = dataset.first()
sample["support_list"] = [[1, 1], [2, 2]]

dataset.bounds("support_list")
```

brimoor: @benjaminpkane ah I see, I added support for this via the syntax: `dataset.bounds("support_list[][]")`

brimoor: Okay even better just `dataset.bounds("support_list")` now works again

PR #1229, opened 2021-08-27
Adding a Heatmap label type

brimoor:
Adds a `Heatmap` label type that can be used to store array data whose values are interpreted as continuous values along an App-configurable colormap.

`Heatmap` is similar but distinct from the existing `Segmentation` label type, which stores categorical array data whose 
pixels correspond to (eg) string labels rather than numerical intensity-like data. 

The optional `Heatmap.range` parameter allows the user to specify the intended dynamic range of a heatmap. If not specified, a range of `[0, 1]` will be assumed if the array contains floating points values and `[0, 255]` if the heatmap contains integer values.

In "color by field" mode in the App, heatmaps are rendered in a single color with opacity proportional to the absolute value of the pixel value. So, for example, when `range=[0, 1]`, 0 is 0% transparent and 1 is 100% opaque, while when `range=[-50, 100]`, -50 is 50% opaque, 0 is 0% opaque, and 100 is 100% opaque.

In "color by value" mode, heatmaps are instead rendered with a fixed transparency using a colorscale that can be configured via the new `colorscale` property of the App config.

Since `plotly` is already a strict dependency of FiftyOne, we use it to handle colorscale -> colormap generation, as they already provide [quite sophisticated support](https://plotly.com/python/colorscales/) for that. The `colorscale` App config parameter can be set to any named plotly colorscale like `"viridis"` (there are many), but the user could even set this to a completely custom colorscale during execution and `AppConfig.get_colormap()` would handle it. See the added user guide docs for details and examples.

### Simple usage

```py
import numpy as np
import fiftyone as fo

sample = fo.Sample(
    heatmap=fo.Heatmap(map=np.random.random(size=(64, 64))),
)

print(sample)
```

### Image example

```py
import numpy as np
import fiftyone as fo
import fiftyone.zoo as foz

def make_heatmap(metadata):
    height = metadata.height // 2
    width = metadata.width // 2
    x, y = np.meshgrid(np.linspace(-1, 1, width), np.linspace(-1, 1, height))

    sign = np.sign(np.random.randn())
    x0, y0 = np.random.random(2) - 0.5
    heatmap = sign * np.exp(-np.sqrt((x - x0) ** 2 + (y - y0) ** 2))

    return fo.Heatmap(map=heatmap, range=[-1, 1])

dataset.compute_metadata()

for sample in dataset:
    sample["heatmap"] = make_heatmap(sample.metadata)
    sample.save()

# Select `Settings -> Color by value` in the App
# Heatmaps will now be rendered using your default colorscale (printed below)
print(session.config.colorscale)

# Switch to a different named colorscale
session.config.colorscale = "RdBu"
session.refresh()

# Use a custom colorscale
session.config.colorscale = [
    [0.00, "rgb(166,206,227)"],
    [0.25, "rgb(31,120,180)"],
    [0.45, "rgb(178,223,138)"],
    [0.65, "rgb(51,160,44)"],
    [0.85, "rgb(251,154,153)"],
    [1.00, "rgb(227,26,28)"],
]
session.refresh()
```

### Video example

```py
import numpy as np
import fiftyone as fo
import fiftyone.zoo as foz

def drift(z):
    znew = ((z + 0.3 * np.random.randn() + 1) % 2) - 1
    return 0.9 * z + 0.1 * znew

def dancing_normal(h, w, n):
    x, y = np.meshgrid(np.linspace(-1, 1, w), np.linspace(-1, 1, h))
    x0, y0 = 0, 0

    maps = []
    for _ in range(n):
        x0, y0 = drift(x0), drift(y0)
        heatmap = np.exp(-np.sqrt((x - x0) ** 2 + (y - y0) ** 2))  # float heatmap
        # heatmap = np.round(255 * heatmap).astype(np.uint8)  # integer heatmap
        maps.append(heatmap)

    return maps

dataset = foz.load_zoo_dataset("quickstart-video", max_samples=1).select_fields().clone()
dataset.compute_metadata()

sample = dataset.first()

height = sample.metadata.frame_height // 16
width = sample.metadata.frame_width // 16
num_frames = len(sample.frames)
heatmaps = dancing_normal(height, width, num_frames)

for frame, heatmap in zip(sample.frames.values(), heatmaps):
    frame["heatmap"] = fo.Heatmap(map=heatmap)

sample.save()

```

### DIFF
--- a/app/packages/looker/src/worker.ts
+++ b/app/packages/looker/src/worker.ts
@@ -113,8 +113,10 @@ const DESERIALIZE = {
 };
 
 const mapId = (obj) => {
-  obj.id = obj._id;
-  delete obj._id;
+  if (obj._id) {
+    obj.id = obj._id;
+    delete obj._id;
+  }
   return obj;
 };
 


benjaminpkane:
I'm going to put colorscale handling on the frontend as using the supplied python methods is proving to be messy.

benjaminpkane:
Ready for eyes.

brimoor:
@benjaminpkane the implementation is looking good on this! I do notice some latency in the grid view when scrolling through datasets with dense heatmaps, but it's certainly not prohibitive.

A couple things I noticed after playing with this:
- "Color by field" mode doesn't seem to work for me when I load an integer-valued mask (everything is rendered as fully transparent)
- I don't think the `color_by_transparency` setting is working in "Color by value" mode? Toggling it doesn't seem to have any effect for me

Also, interested to get @benjaminpkane and @ehofesmann's thoughts on this:

After playing with both segmentations and heatmaps in the App, I'm thinking this may be a good time to introduce a slider to allow users to adjust the transparency of all overlays in the App. I sometimes find that heatmaps are not visible enough, while sometimes I find the opposite. The "optimal" setting can vary based on image characteristics and map characteristics. At the same time, we've also had users requesting the ability to adjust the transparency of overlays for other spatial types like detections. Especially in crowded scenes when a touch of transparency would help them make sense of both the detections and the content behind them.

As for how to implement this, I guess this would either be in the settings menu, in which case it would apply to both the grid view and expanded view, or as a slider directly in the visualizer HUD, in which case it would only apply to the grid. For overlay elements that already support transparency, this setting would basically determine the maximum transparency: `[0, alpha]` range rather than `[0, 1]` propagated to the mask for rendering

Other design factors (my gut reactions):
- Do we allow per-field transparency configuration? (no, seems too complicated)
- Does transparency apply to the tooltip? (no)
- Does this setting affect tag bubbles in the grid? (no)

brimoor:
@benjaminpkane update: the video example in the PR description doesn't load in the App for me (stuck in loading state)

benjaminpkane:
Noting the current performance, and the addition of a transparency slider, I need to move color updates (array updates) for the masks off of the main thread.

brimoor:
> Noting the current performance, and the addition of a transparency slider, I need to move color updates (array updates) for the masks off of the main thread.

If it's going to be a big job then I recommend we have a design chat before diving into transparency slider, to make sure it's what we want

benjaminpkane:
> > Noting the current performance, and the addition of a transparency slider, I need to move color updates (array updates) for the masks off of the main thread.
> 
> If it's going to be a big job then I recommend we have a design chat before diving into transparency slider, to make sure it's what we want

The implication is that mask colors will update asynchronously. I have to touch every mask pixel in a for loop each time the coloring is updated. And for video, future frame masks would probably need to block playback if they are not ready...

brimoor:
yeah I get it. Even without a slider to further complicate things, it would be nice if the main thread weren't be burdened by the complexities of array overlays. A brainstorming chat will help I think

brimoor:
Summarizing the live discussion: let's try introducing a single slider under the settings menu that allows the user to adjust the transparency of all overlays, both in the grid and in the expanded modal. The slider would range from `[0, 1]` and default to `alpha = 0.7`, and have the following effects:
- Segmentations, instance masks, and polygon interiors are rendered at transparency `alpha`
- Bboxes, polylines, etc are rendered at transparency `min(1, alpha / 0.7)`
- Heatmaps would be rendered with transparency range `[0, alpha]`

Personally I don't think it's critical for this setting to apply to classification bubbles in the grid (especially if they will soon be migrated to the canvas), but that's just my two cents.

benjaminpkane:
> @benjaminpkane update: the video example in the PR description doesn't load in the App for me (stuck in loading state)

```py
import numpy as np
import fiftyone as fo
import fiftyone.zoo as foz

def drift(z):
    znew = ((z + 0.3 * np.random.randn() + 1) % 2) - 1
    return 0.9 * z + 0.1 * znew

def dancing_normal(h, w, n):
    x, y = np.meshgrid(np.linspace(-1, 1, w), np.linspace(-1, 1, h))
    x0, y0 = 0, 0

    maps = []
    for _ in range(n):
        x0, y0 = drift(x0), drift(y0)
        heatmap = np.exp(-np.sqrt((x - x0) ** 2 + (y - y0) ** 2))  # float heatmap
        # heatmap = np.round(255 * heatmap).astype(np.uint8)  # integer heatmap
        maps.append(heatmap)

    return maps

dataset = foz.load_zoo_dataset("quickstart-video", max_samples=1).select_fields().clone()
dataset.compute_metadata()

sample = dataset.first()

height = sample.metadata.frame_height // 16
width = sample.metadata.frame_width // 16
num_frames = len(sample.frames)
heatmaps = dancing_normal(height, width, num_frames)

for frame, heatmap in zip(sample.frames.values(), heatmaps):
    frame["heatmap"] = fo.Heatmap(map=heatmap)

sample.save()

```

`// 16` (not `// 4`) will work. 4 exceeds the 16MB document size limit. I'll work on error handling.

benjaminpkane:
Need one more day on this.

Unfortunately I cannot work around the current lookup limitation in the App (produced by the video snipped) because the extended view needs all frames attached for filtering frame labels.

benjaminpkane:
Ok, this is now functional. Masks are now fully processed in a worker (off of the main thread). I want to make a small improvement to color refreshes in the grid in the morning with fresh eyes.

benjaminpkane:
> For simplicity, we could remove the `colorscale_transparency` setting and always use the latter behavior in "Color by value" mode. Then we have the following behavior:
> 
> * Color by field: transparency captures pixel value
> * Color by value: color captures pixel value

I think this is best for the time being

brimoor:
SGTM

brimoor:
Okay I was playing with `Segmentation` masks that are either boolean or have only 1 non-zero mask target (which work beautifully now!), and I noticed a couple small bugs:
- In color by value mode, `id=None` shows in the tooltip, but it should still show
- After toggling between color by field/value, I sometimes get in a state where selecting the mask does not visibly toggle the selected label UI. This is similar I believe to how the sample selection UI sometimes stops working as I reported in my previous comment

PR #1233, opened 2021-08-30
Adding support for compound sort criteria

brimoor:
Adds support for specifying compound sort criteria when using the `sort_by()` stage.

Example usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

#
# Performs a compound sort where samples are first sorted in
# descending or by number of detections and then by ascending order
# of uniqueness for samples with the same number of predictions
#
view = dataset.sort_by(
    [
        (F("predictions.detections").length(), -1),
        ("uniqueness", 1),
    ]
)

num_objects, uniqueness = view[:5].values(
    [F("predictions.detections").length(), "uniqueness"]
)
print(list(zip(num_objects, uniqueness)))
```

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -4122,8 +4122,8 @@ class SampleCollection(object):
 
     @view_stage
     def sort_by(self, field_or_expr, reverse=False):
-        """Sorts the samples in the collection by the given field or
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
LGTM

PR #1240, opened 2021-08-30
Docs formatting tweaks

brimoor:
We document the parameters that can be passed to `load_zoo_dataset()` for the COCO and Open Images datasets in a couple places in the docs. We do this because:
- The possible parameters are defined in a few different classes and thus linking directly to any one class doesn't fully cover the options
- These two datasets are quite popular, so we want to minimize friction in learning how to download and start using them

This PR switches from ``monospace``-style formatting to **bold** (*default*) style formatting, for consistency with how function arguments are rendered in the actual API docs. This visually helps make the parameters more readable, I think.


benjaminpkane (approved):
LGTM

PR #1244, opened 2021-09-02
Frames improvements

brimoor:
This PR makes two enhancements to frame-level labels for video datasets.

## Optimizing frame access for frame-filtering views

Previously, accessing the frames for a video sample in a dataset view that involves frame-level filtering would use `view.select(sample_id)._aggregate(frames_only=True)` to retrieve the frames. This is inefficient because selecting the sample of interest after running `view` makes iterating over all the frames of all samples `O(n^2)`, while it could be `O(n)` if we first selected only the sample of interest.

This PR updates the implementation in the `FramesView` class to use the select-then-aggregate approach, which unfortunately is a bit tricky as per the comments in the code.

Note the 18x performance improvement in the example below (the gains are particularly large in this case because the runtime of the loop is dominated by the cost of the DB aggregations involved):

```py
import eta.core.utils as etau
import fiftyone as fo
import fiftyone.zoo as foz

# A video dataset with 40 densely labeled videos
dataset.add_samples(dataset)
dataset.add_samples(dataset)

# A view that involves frame-level filtering
# We also include a pernicious skip stage to ensure the implementation is working 
view = dataset.filter_labels("frames.detections", F("label") == "vehicle").skip(1)

# this PR: Time elapsed: 271.3ms
# develop: Time elapsed: 4.9 seconds
with etau.Timer():
    for sample in view:
        for frame_number in sample.frames:
            pass
```

## Overwriting existing frames

Previously it was not allowed to overwrite an existing frame, but now this is allowed:

```py
import fiftyone as fo

sample.frames[1] = fo.Frame(hello="world")

dataset = fo.Dataset()
dataset.add_sample(sample)

# Previously would raise an error, now is allowed
sample.frames[1] = fo.Frame(goodbye="world")
sample.save()
```

I think this change makes sense because `Frames.__getitem__()` behaves like a defaultdict- it will create a new frame if the frame does not exist. But this could lead to confusion if a user accesses a frame in a read-only context and then later wants to populate the frame and then becomes confused when they see "cannot overwrite existing frame".


benjaminpkane (approved):
LGTM. I moved and renamed the optimization method in order to make use of it in the server.

benjaminpkane:
Fixing the tests

brimoor:
> LGTM. I moved and renamed the optimization method in order to make use of it in the server.

Sweet, I moved the method to `fiftyone.core.view` for clarity and also moved my hacky new stage registry to `fiftyone.core.stages` so that it is harder to forget to update it.

PR #1245, opened 2021-09-02
Configurable session waiting

benjaminpkane:
Adds an integer keyword argument `wait` to `Session.wait()` and a corresponding `--wait` argument to relevant CLI commands.

`session.wait(-1)` permanently blocks execution until a keyboard interrupt, which is useful in cases where you are running an App session that you want to serve "forever".

`session.wait(10)` might be more appropriate than the default wait time (3 seconds) for remote sessions over slow internet connections, where refreshing the App via ctrl + R in the browser could take >3 seconds to re-establish a connection.

```py
import fiftyone as fo

# ex: serve forever
session.wait(-1)

# ex: wait 10 seconds before continuing execution
session.wait(10)  
```

### DIFF
--- a/fiftyone/core/session.py
+++ b/fiftyone/core/session.py
@@ -860,28 +860,28 @@ class Session(foc.HasClient):
     def wait(self, wait=3):
         """Blocks execution until the App is closed by the user.
 
-        Closing a desktop App will always exit the wait. For browser Apps, all
-        connected windows (tabs) must be closed.
+        For browser Apps, all connected windows (tabs) must be closed before
+        this method will unblock.
+
+        For desktop Apps, all positive ``wait`` values are equivalent;
+        execution will immediately unblock when the App is closed.
 
         Args:
             wait (3): the number of seconds to wait for a new App connection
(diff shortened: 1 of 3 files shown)


fiftyone/core/session.py
```diff
@@ -857,21 +857,31 @@ def no_show(self):
         """
         return fou.SetAttributes(self, _auto=False)
 
-    def wait(self):
+    def wait(self, wait=1):
```
brimoor, line 860: Can you make zero or any negative number for `wait` serve forever?

brimoor (approved):
LGTM. Hope you're okay with a couple tweaks I made:
- Added support for wait < 0 for desktop sessions (it still makes sense to wait forever, its just that all positive values are equivalent)
- Updated so that a strictly negative value is required to wait forever. To me, `sesison.wait(0)` seems to imply that we'll immediately exit, not wait forever. `session.wait(-1)` seems more distinctive

FYI very annoying, but currently any changes to CLI args must be manually added to the CLI docs.

benjaminpkane:
Yes, good adjustments. Thanks.

PR #1247, opened 2021-09-03
Release v0.13.2

### LINKED ISSUE #1225: [BUG] Incorrect statistics for classifcations in the App
When working with the `lfw` zoo dataset, classification counts are incorrect for the base view.

The filter should show `236`, not `236 of 530`.

Screenshot from 2021-08-25 15-03-00

benjaminpkane:
* Ensures forks, in particular windows-style forks, have a database port set and therefore do not attempt to create their own DatabaseService
* Only log import events from the main process
* Resolves #1225


brimoor (approved):
LGTM!

PR #1248, opened 2021-09-03
[BUG] Fix error when loading CVATImageDataset with None values

ehofesmann:
Fixes error that is thrown when loading a `CVATImageDataset` if any of the attributes of the label schema do not contain a list of specified values.

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -814,7 +814,7 @@ class CVATTaskLabels(object):
                 _attributes.append(
                     {
                         "name": attribute["name"],
-                        "categories": attribute["values"].split("\n"),
+                        "categories": (attribute["values"] or "").split("\n"),
                     }
                 )
 
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -796,7 +796,7 @@ class CVATTaskLabels(object):
     @classmethod
     def from_labels_dict(cls, d):
         """Creates a :class:`CVATTaskLabels` instance from the ``<labels>``
-        tag of a CVAT image annotation XML file.
+        tag of a CVAT annotation XML file.
 
         Args:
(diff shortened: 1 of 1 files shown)


fiftyone/utils/cvat.py
```diff
@@ -811,11 +811,10 @@ def from_labels_dict(cls, d):
             attributes = _ensure_list(_tmp.get("attribute", []))
             _attributes = []
             for attribute in attributes:
+                _values = attribute.get("values", None)
+                _categories = _values.split("\n") if _values else []
```
brimoor, line 815: `"".split("\n")` is `[""]`, so I updated this so that missing values will result in `[]` instead, which is more reasonable I think.

This `CVATTaskLabels` is just the annotation schema, it doesn't affect things much. But, nonetheless, cleaner I think.

ehofesmann: Good call!

fiftyone/utils/cvat.py
```diff
@@ -1146,9 +1145,10 @@ def _parse_anno_dict(d):
 
         attributes = []
         for attr in _ensure_list(d.get("attribute", [])):
-            name = attr["@name"].lstrip("@")
-            value = _parse_attribute(attr["#text"])
-            attributes.append(CVATAttribute(name, value))
+            if "#text" in attr:
```
brimoor, line 1148: Related patch, '#text' is apparently not present when a label attribute with empty value is parsed from the XML. 

This change will just omit the attribute in that case, which is fine, since our codebase has a strict convention that None and missing are interchangeable. I considering inserting `None`, but there is other code in this module that tries to guess attribute types based on their values in some cases. Might not actually have been an issue (didn't fully investigate), but type inference would have failed on `None`, so it felt dangerous.

brimoor (approved):
LGTM

PR #1250, opened 2021-09-03
[BUG] Fix importing COCO, VOC formats when exported from CVAT

ehofesmann:
Exporting annotations from CVAT creates artifacts that slightly differ from existing formats. This PR updates existing importers to avoid errors and allow importing of attributes when exported from CVAT.

For example, the COCO format when exported from CVAT contains:
```json
        {
            "id": 34,
            "image_id": 9,
            "category_id": 2,
            "segmentation": [],
            "area": 6499.0,
            "bbox": [
                28.0,
                254.0,
                97.0,
                67.0
            ],
            "iscrowd": 0,
            "attributes": {
                "label_id": "5f452471ef00e6374aac53ca",
                "attribute:iscrowd": "0.0",
                "attribute:area": "4827.32605",
                "occluded": false
            }
        },
```

The Pascal VOC format contains:
```
<annotation>
  <folder></folder>
  <filename>000000.jpg</filename>
  <source>
    <database>Unknown</database>
    <annotation>Unknown</annotation>
    <image>Unknown</image>
  </source>
  <size>
    <width>640</width>
    <height>480</height>
    <depth></depth>
  </size>
  <segmented>0</segmented>
  <object>
    <name>bird</name>
    <occluded>0</occluded>
    <bndbox>
      <xmin>135.0</xmin>
      <ymin>2.0</ymin>
      <xmax>431.0</xmax>
      <ymax>455.0</ymax>
    </bndbox>
    <attributes>
      <attribute>
        <name>label_id</name>
        <value>5f452471ef00e6374aac53c8</value>
      </attribute>
      <attribute>
        <name>attribute:iscrowd</name>
        <value>0.0</value>
      </attribute>
      <attribute>
        <name>attribute:area</name>
        <value>73790.37944999996</value>
      </attribute>
    </attributes>
```

Loaded into Python, this becomes:
```python
attributes = {
    ....,
    "bndbox": ...,
    "attributes": {
        "attribute": [
            {
                "name": "label_id",
                "value": "5f452471ef00e6374aac53c8",
            }, ...
        ]
    }
}
```

* `YOLOv4Dataset` does not support attributes so it loads without issue
* CVAT does not export TF Records following the TF Object Detection API so `TFObjectDetectionDataset` is not relevant

### DIFF
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -1121,10 +1121,10 @@ class COCOObject(object):
         Returns:
             a :class:`COCOObject`
         """
+        # Handles CVAT exported attributes
+        d.update(d.pop("attributes", {}))
+
         if extra_attrs is True:
-            if "attributes" in d:
-                d.update(d["attributes"])
-                del d["attributes"]
             return cls(**d)
 
         if etau.is_str(extra_attrs):
@@ -1135,13 +1135,7 @@ class COCOObject(object):
         else:
             attributes = {}
 
-        # Handles CVAT exported attributes
-        if "attributes" in d and extra_attrs:
-            for f in extra_attrs:
-                if attributes[f] is None:
-                    attributes[f] = d["attributes"].get(f, None)
-
-        args = dict(
+        return cls(
             id=d.get("id", None),
             image_id=d.get("image_id", None),
             category_id=d.get("category_id", None),
@@ -1151,9 +1145,8 @@ class COCOObject(object):
             score=d.get("score", None),
             area=d.get("area", None),
             iscrowd=d.get("iscrowd", None),
+            **attributes,
         )
-        args.update(attributes)
-        return cls(**args)
 
     def _get_label(self, classes):
         if classes:
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -1122,6 +1122,9 @@ class COCOObject(object):
             a :class:`COCOObject`
         """
         if extra_attrs is True:
+            if "attributes" in d:
+                d.update(d["attributes"])
+                del d["attributes"]
             return cls(**d)
 
         if etau.is_str(extra_attrs):
@@ -1132,7 +1135,13 @@ class COCOObject(object):
         else:
             attributes = {}
 
-        return cls(
+        # Handles CVAT exported attributes
+        if "attributes" in d and extra_attrs:
+            for f in extra_attrs:
+                if attributes[f] is None:
+                    attributes[f] = d["attributes"].get(f, None)
+
+        args = dict(
             id=d.get("id", None),
             image_id=d.get("image_id", None),
             category_id=d.get("category_id", None),
@@ -1142,8 +1151,9 @@ class COCOObject(object):
             score=d.get("score", None),
             area=d.get("area", None),
             iscrowd=d.get("iscrowd", None),
-            **attributes,
         )
+        args.update(attributes)
+        return cls(**args)
 
     def _get_label(self, classes):
         if classes:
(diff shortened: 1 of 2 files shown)


fiftyone/utils/voc.py
```diff
@@ -783,13 +803,16 @@ def _parse_attribute(value):
     except:
         pass
 
-    if value in {"True", "true"}:
-        return True
+    try:
```
brimoor, line 806: why would these checks every fail?

fiftyone/utils/voc.py
```diff
@@ -580,8 +582,18 @@ def to_detection(self, frame_size, extra_attrs=True):
         label = self.name
         bounding_box = self.bndbox.to_detection_format(frame_size)
 
+        # Handles CVAT exported attributes
+        cvat_attrs = self.attributes.get("attributes", None)
```
brimoor, line 586: I think it makes more sense to do this attribute parsing in `__init__()` so that if we add other `to_*()` methods they don't have to re-implement this logic, which they would presumably also require

fiftyone/utils/voc.py
```diff
@@ -183,6 +183,7 @@ def setup(self):
                 for p in etau.list_files(self.labels_path, recursive=True)
             }
         else:
+            logger.info("No labels found at %s" % self.labels_path)
```
brimoor, line 186: I'm planning a pass over all importers to handle these logging type messages. So perhaps just remove for now for consistency with the others?

The complexity was that I was trying to support unlabeled splits as well, where you wouldn't necessarily want to see this logging message because you in fact did not specify a labels path (but this is printed due to class internal defaults)

fiftyone/utils/coco.py
```diff
@@ -1122,6 +1122,9 @@ def from_anno_dict(cls, d, extra_attrs=True):
             a :class:`COCOObject`
         """
         if extra_attrs is True:
+            if "attributes" in d:
```
brimoor, line 1125: I think the changes here could be reduced to just this one line at the beginning of the method:

```py
# Handles CVAT exported attributes
d.update(d.pop("attributes", {}))
```

As currently written, the attribute handling logic is implemented twice here, which is what initially caught my eye.

brimoor (changes requested):

fiftyone/utils/voc.py
ehofesmann: The attributes are loaded in as a dictionary like the example in the PR description, so this call sometimes passes in a `dict`:
https://github.com/voxel51/fiftyone/blob/f12d3dc463e79b3b518f369a6e7892bbc9a7acb0/fiftyone/utils/voc.py#L529-L532

fiftyone/utils/voc.py
ehofesmann: Sounds good

fiftyone/utils/coco.py
ehofesmann: Good call!

fiftyone/utils/voc.py
ehofesmann: I moved the cvat attribute parsing so that this isn't an issue.

brimoor (approved):
LGTM

fiftyone/utils/voc.py
```diff
@@ -525,6 +525,13 @@ def from_annotation_dict(cls, d):
         """
         name = d["name"]
         bndbox = VOCBoundingBox.from_bndbox_dict(d["bndbox"])
+
+        # Handles CVAT exported attributes
+        if "attributes" in d:
```
brimoor, line 530: I threw in these `if "attributes" in d` checks just to avoid a bunch of operations on empty dictionaries that would happen in the 99.999% of use cases where CVAT-style attributes are not involved

PR #1205, opened 2021-08-19
Include array type in mask overlay construction

benjaminpkane:
Resolves #1204

Issue #1204, opened 2021-08-19
[BUG] display of segmentation masks with dype uint8 is broken


falckt:
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux, Ubuntu 21.04 and 18.04.5
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: fiftyone v0.12.0, Voxel51, Inc.
-   **Python version**: Python 3.8.10 (fresh conda environment)

### Commands to reproduce

[Example data](https://github.com/voxel51/fiftyone/files/7012549/test.tar.gz)

```bash
fiftyone datasets create --name uint_mask --dataset-dir $TEST_DIR --type fiftyone.types.ImageSegmentationDirectory
fiftyone app launch uint_mask
```

### Describe the problem

Segmentation masks that are added with dtype `unit8` appear broken in the app. This is the case when importing a dataset using `ImageSegmentationDirectory` as above, as well as when manually adding such a sample to a dataset (c.f. example code below). When changing the dtype to `int` everything works as expected as seen in the code example below.

The behavior is easiest to observe with the code example below. If you look at the segmentation mask in the app, `int_mask` shows as expected, while `uint8_mask` is completely broken. There are a few pixels visible in a the upper left part of the image, but the label values are completely off. Probably something in the serialization/deserealization logic goes wrong and the data is unpacked incorrectly.

A quick fix is changing https://github.com/voxel51/fiftyone/blob/7c061ea22030ea355357df78a8c61598d784ea02/fiftyone/utils/data/importers.py#L2717 to
```python
    mask = etai.read(mask_path, cv2.IMREAD_UNCHANGED).astype(int)
```

### Code to reproduce issue

```python
# generate test data
import cv2
import numpy as np

img = np.zeros((400, 400, 3), np.uint8)
img = cv2.circle(img, (200, 200), 100, (255, 255, 255), -1)

mask = img[..., 0]

EXAMPLE_PATH = "/tmp/test_img.png"
cv2.imwrite(EXAMPLE_PATH, img)

# minimal example
import fiftyone as fo

ds = fo.Dataset()

ds.add_sample(
    fo.Sample(
        EXAMPLE_PATH,
        uint8_mask=fo.core.labels.Segmentation(mask=mask),
        int_mask=fo.core.labels.Segmentation(mask=mask.astype(int)),
    )
)

session = fo.launch_app(ds)

[...]

PR #1253, opened 2021-09-07
Adding support for instance/semantic segmentation annotation

### LINKED ISSUE #1243: [BUG] Instance segmentation masks that consist of separated regions are not loaded into CVAT properly
FiftyOne stores instance segmentations under the `mask` attribute of a `Detection`. This is a binary mask so it can consist of multiple unconnected regions. When uploaded to CVAT, the mask is first converted to polylines and uploaded as a CVAT polygon. At the moment, only the first polyline is uploaded to CVAT. This section needs to be updated (as well as the corresponding section in `_create_polyline_shapes()`):

https://github.com/voxel51/fiftyone/blob/e43c00ab96282d1f016e4f57f806c8f48feff6bc/fiftyone/utils/cvat.py#L4170

### Example
```python
import fiftyone as fo
import fiftyone.zoo as foz

    "open-images-v6",
    split="validation",
    image_ids=["0022bffa9abfb554"],
    dataset_name="seg_bug",
)

```
image
```python
results = dataset.annotate(
    "seg_bug",
    label_field="segmentations",
    launch_editor=True,
)
```
image

**Possible solution**:

Create a new `instance_id` attribute for segmentations uploaded to CVAT to match together multiple polygon annotations.

ehofesmann:
Requires https://github.com/voxel51/eta/pull/523
Resolves #1243 

Adds support for semantic and instance segmentation when using FiftyOne's annotation API:
- `label_type="segmentation"`: semantic segmentation
- `label_type="instances"`: instance segmentation

Additional changes
- Added support for supplying mask targets for semantic segmentation tasks
- Added support for polyline vs polygon annotations
- When annotating video tracks in CVAT, labels are now interpolated between keyframes as necessary upon import
- Added a `mask_types="stuff"/"thing"` argument to `Segementation.to_detections()` that allows for control over whether to generate one instance per class (stuff) or one instance per connected region (thing) when converting segmentation masks to detections
- Added a `Segmentation.to_polylines()` method
- Added support for importing/exporting segmentation masks with > 256 classes as uint16 masks when working with `ImageSegmentationDirectory` format
- Fixed a bug that prevented multiple imports of the same annotation run from working as expected when a label is deleted but then later re-added
- Fixed a bug that prevented annotations for new label fields of video datasets from being imported properly
- Fixed a bug that would cause things like polygons with < 3 points to be deleted when editing existing annotations in CVAT
- API cleanup

The video annotation import problem was solved by adding an `ensure_frames()` method to `Dataset`/`DatasetView` that enables methods that rely on `Frame` instances to exist to ensure that this is indeed the case. 

The polygon with < 3 points problem was solved by requiring the annotation backend to keep track of which IDs it was able to upload for annotation, rather than relying on a generic `AnnotationBackend.build_id_map()` method that assumes that all labels in the view were able to be uploaded.

### Detection/instance annotation

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "open-images-v6",
    split="validation",
    max_samples=25,
    label_types=["segmentations", "detections"],
)

# Populate a field with detections and instances
dataset.clone_sample_field("detections", "boxes_and_instances")
dataset.clone_sample_field("segmentations", "tmp")
dataset.merge_labels("tmp", "boxes_and_instances")
print(dataset)

# Edit boxes
dataset.annotate(
    "detections",
    label_field="boxes_and_instances",
    label_type="detections",
    launch_editor=True,
)

print(dataset.get_annotation_info("detections"))
print(dataset.load_annotation_view("detections"))

dataset.load_annotations("detections")
session.refresh()

# Edit instances
dataset.annotate(
    "instances",
    label_field="boxes_and_instances",
    label_type="instances",
    launch_editor=True,
)

dataset.load_annotations("instances")
session.refresh()

# Boxes are chosen by default
dataset.annotate(
    "default",
    label_field="boxes_and_instances",
    launch_editor=True,
)

dataset.load_annotations("default")
session.refresh()
```

### Polyline/polygon annotation

```py
import random
import fiftyone as fo
import fiftyone.zoo as foz

    "open-images-v6",
    split="validation",
    max_samples=20,
    label_types="segmentations",
)

# Populate some polylines and polygons
for sample in dataset:
    filled = bool(random.random() < 0.5)
    sample["polylines"] = sample.segmentations.to_polylines(filled=filled)
    sample.save()

print(dataset)

# Edit polygons
dataset.annotate(
    "polygons",
    label_field="polylines",
    label_type="polygons",
    launch_editor=True,
)

print(dataset.get_annotation_info("polygons"))
print(dataset.load_annotation_view("polygons"))

dataset.load_annotations("polygons")
session.refresh()

# Edit polylines
dataset.annotate(
    "polylines",
    label_field="polylines",
    label_type="polylines",
    launch_editor=True,
)

dataset.load_annotations("polylines")
session.refresh()

# Polygons are chosen by default
dataset.annotate(
    "default",
    label_field="polylines",
    launch_editor=True,
)

dataset.load_annotations("default")
session.refresh()
```

### Semantic segmentation annotation

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "quickstart",
    dataset_name="segmentation-eval-demo",
    max_samples=10,
    shuffle=True,
)

CLASSES = (
    "tvmonitor"
)
dataset.default_mask_targets = {
    idx: label for idx, label in enumerate(CLASSES.split(","))
}

dataset.apply_model(model, "semantic_segmentations")

dataset.annotate(
    "segmentations",
    label_field="semantic_segmentations",
    launch_editor=True,
)

print(dataset.get_annotation_info("segmentations"))

dataset.load_annotations("segmentations")
session.refresh()

results = dataset.load_annotation_results("segmentations")
dataset.load_annotations("segmentations")
```

### Existing video field annotation

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.limit(1)

view.annotate(
    "video",
    launch_editor=True,
)

print(dataset.get_annotation_info("video"))

view = dataset.load_annotation_view("video")

dataset.load_annotations("video")
session.refresh()
```

### New video field annotation

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.clear_frames()

view = dataset.limit(1)

view.annotate(
    "video",
    classes=["vehicle"],
    attributes={
        "occluded": {
            "type": "radio",
            "values": [0, 0.2, 0.5, 1],
            "default": 0,
            "mutable": False,
        }
    },
    launch_editor=True,
)

print(dataset.get_annotation_info("video"))

view = dataset.load_annotation_view("video")

dataset.load_annotations("video")
session.refresh()
```


fiftyone/core/labels.py
```diff
@@ -561,7 +561,7 @@ def to_segmentation(self, mask=None, frame_size=None, target=255):
 
         mask, _ = _parse_to_segmentation_inputs(mask, frame_size, None)
         _render_instance(mask, self, target)
-        return Segmentation(mask=mask)
+        return Segmentation(mask=mask.astype(np.int))
```
brimoor, line 564: `_parse_to_segmentation_inputs()` creates a `uint8` mask by default, so if we need to use `int`, then it would be best to specify it there rather than do a cast in all of these return statements.

Is there a particular reason you need `int` as opposed to `uint8`? 

In fact, now that I think of it, this change probably breaks things like [segmentation exports](https://github.com/voxel51/fiftyone/blob/5063f248a247ffc4a8f56d2d19201d3a190527e4/fiftyone/utils/data/exporters.py#L2398-L2431), which need to write masks as images, which, unless we do some work, will require 8 bit grayscale.

I'm struggling a bit to decide what dtypes to use/allow for semantic segmentations. Whenever FiftyOne internal code creates masks, I've tended to use `uint8` to save space and for export reasons mentioned above, while [the docs](https://voxel51.com/docs/fiftyone/user_guide/using_datasets.html#semantic-segmentation) say "any integer type", and the code itself doesn't do any validation at all...

On the one hand, the smallest possible dtype for arrays stored in the database is a pretty big deal, as `uint8` (8 bits) is 4x-8x smaller and thus faster to work with than `int` (defaults to 32 bits, unless a large enough value is inserted, in which case it silently upgrades to 64 bits, I believe).

On the other hand, this does implicitly limit to 255 classes + background, which might be too few for some users (but this would be rare, I think, given that segmentations are often stored on disk as images).

But, of course, 8 bits only allows for 256 classes

fiftyone/core/labels.py
```diff
@@ -1278,6 +1278,62 @@ def to_image_labels(self, name=None):
         """
         return etai.ImageLabels(mask=self.mask, tags=self.tags)
 
+    def to_detections(self, mask_targets=None):
```
brimoor, line 1281: Is there a use case for this?

If I wanted to do a transformation like this, I would likely want the default (or at least an option) to create one detection per connected region rather than always one per class. In fact, I could see the user wanting to be able to specify this on a per-class basis ("stuff-like" vs "thing-like" distinction used in some papers).

I might also consider the name `to_instances()` for this method, since, although  the data type returned is `Detections`, as it is a bit more specific.

But, in any case, first question: how this is used right now?

fiftyone/core/labels.py
ehofesmann: Right, I really should have pointed this out specifically. The reason I did that was because the App wasn't showing masks stored in `uint8`, only when stored in `int` so I assumed that when the docs said "any integer type", it meant `int`. I agree that `uint8` should be the default and it may have just been an App issue. Have you encountered that?

fiftyone/core/labels.py
ehofesmann: **Current use case**: CVAT processes all segmentations as polygons so I needed a way to go from `Segmentation` to `Polyline`. Since detections already had a `to_polyline()` method, I just converted from `Segmentation` to `Detections` and then go to polylines from there.

The fact that it creates one instance per class is fine since that is how it is annotated in CVAT, so splitting it into instances would just be unnecessary computation for this use case.

I agree that this isn't the most useful method otherwise. Also `Segmentations.to_polylines()` may be preferable, but looking through the code it didn't seem like it would be too much more efficient than `Segmentations` -> `Detections` -> `Polylines` so I just left it as is.

fiftyone/core/labels.py
brimoor: Oh I see. That was reported in https://github.com/voxel51/fiftyone/issues/1204 but Ben fixed it in https://github.com/voxel51/fiftyone/pull/1205 a few weeks ago

brimoor: Since the App should now any integer type, I'll take a whack at improving the export logic so that non-uint8 segmentations can be appropriately exported (I guess as 16 bit images

If the cast isn't need here anymore, then I think we should remove it and revisit the issue of segmentation mask typing separately.

fiftyone/core/labels.py
brimoor: I see. Just because I like useful utilities, I think I'll take a whack at upgrading this method to allow for customizing stuff vs thing-like behavior, and the CVAT logic can use stuff-only.

I'll probably throw in `to_polylines()` as well, for completeness.

fiftyone/core/labels.py
ehofesmann: There it is! I didn't rebuild my development App since that update apparently. Yeah, this change should be removed then.

brimoor (approved):
LGTM

brimoor (approved):
LGTM

Issue #1252, opened 2021-09-06
Cityscapes Label Data not importing correctly [BUG]


ShaneGilroy:
### Commands to reproduce

```py
# Load Citypersons Dataset

import fiftyone as fo
import fiftyone.zoo as foz

# The path to the source files that you manually downloaded
source_dir = "/Users/gilro/Desktop/Datasets/Cityscapes"

dataset = foz.load_zoo_dataset(
    "cityscapes",
    split="validation",
    source_dir=source_dir,
    person_annos=True,
    fine_annos=True,
)

session = fo.launch_app(dataset)
```

### Describe the problem

**When loading the Citypersons dataset, labels are not loaded with the images**. The directory is set up as instructed in the documentation and the "person_annos" parameters are forced to true. 

### Code to reproduce issue

```py
import fiftyone as fo
import fiftyone.zoo as foz

source_dir = "/Users/gilro/Desktop/Datasets/Cityscapes"

dataset = foz.load_zoo_dataset(
    "cityscapes",
    split="validation",
    source_dir=source_dir,
    person_annos=True,
    fine_annos=True,
)

session = fo.launch_app(dataset)
```

Provide a reproducible test case that is the bare minimum necessary to generate
the problem.

### Other info / logs

Source directory set up as instructed below:

![image](https://user-images.githubusercontent.com/38457008/132221021-4e1a0612-f72e-4322-9d8b-00c189b42360.png)

Output - "No Label Fields" as below:

![image](https://user-images.githubusercontent.com/38457008/132221120-be5994b3-363f-46e5-91a5-c6515e565f65.png)

Include any logs or source code that would be helpful to diagnose the problem.
If including tracebacks, please include the full traceback. Large logs and
files should be attached.

### What areas of FiftyOne does this bug affect?

-   [x] `App`: FiftyOne application issue
-   [ ] `Core`: Core `fiftyone` Python library issue
-   [ ] `Server`: Fiftyone server issue

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another
member of your organization be willing to contribute a fix for this bug to the
FiftyOne codebase?

-   [ ] Yes. I can contribute a fix for this bug independently.
-   [x] Yes. I would be willing to contribute a fix for this bug with guidance
        from the FiftyOne community.
-   [ ] No. I cannot contribute a bug fix at this time.

brimoor:
@ShaneGilroy When I run the code you shared the dataset loads properly.

[...]

PR #1256, opened 2021-09-08
Adding missing zip to cityscapes docs

brimoor:
Fixes a documentation bug discovered when resolving https://github.com/voxel51/fiftyone/issues/1252.


benjaminpkane (approved):
LGTM

PR #1258, opened 2021-09-09
Setup tweaks

brimoor:
Small tweaks to align with teams.

@benjaminpkane is the `fiftyone_desktop` -> `fiftyone-desktop` change safe? I believe so since PyPI translates underscores to hyphens anyway, right?


benjaminpkane (approved):
Yes, it's safe. LGTM

PR #1260, opened 2021-09-10
Adding Labelbox annotation backend

ehofesmann:
This PR adds a [Labelbox](https://labelbox.com) backend to FiftyOne's [annotation API](https://voxel51.com/docs/fiftyone/user_guide/annotation.html).

Complete documentation is available (after next release) at https://voxel51.com/docs/fiftyone/integrations/labelbox.html.

### Example usage

First create an account at https://app.labelbox.com.

```shell
pip install labelbox

export FIFTYONE_LABELBOX_API_KEY=...
```

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Step 1: Load your data into FiftyOne

# Step 2: Locate a subset of your data requiring annotation

# Create a view that contains only high confidence false positive model
# predictions, with samples containing the most false positives first
most_fp_view = (
    dataset
    .filter_labels("predictions", (F("confidence") > 0.8) & (F("eval") == "fp"))
    .sort_by(F("predictions.detections").length(), reverse=True)
)

# Retrieve the sample with the most high confidence false positives
sample_id = most_fp_view.first().id
view = dataset.select(sample_id)

# Step 3: Send samples to Labelbox

# A unique identifier for this run
anno_key = "labelbox_basic_recipe"

label_schema = {
    "new_ground_truth": {
        "type": "detections",
        "classes": dataset.distinct("ground_truth.detections.label"),
        "attributes": {
            "iscrowd": {
                "type": "radio",
                "values": [True, False],
            },
        },
    },
}

view.annotate(
    anno_key,
    backend="labelbox",
    label_schema=label_schema,
    launch_editor=True,
)
print(dataset.get_annotation_info(anno_key))

# Step 4: Perform annotation in Labelbox and save the tasks

# Step 5: Merge annotations back into FiftyOne dataset

# Load the view that was annotated in the App
view = dataset.load_annotation_view(anno_key)

# Step 6: Cleanup

# Delete tasks from Labelbox
results.cleanup()

# Delete run record (not the labels) from FiftyOne
```


brimoor (approved):
LGTM.

Per offline discussions, I updated the Labelbox backend to explicitly raise an error if an existing field is passed to `annotate()`, since uploading existing labels is not yet suported.

To go along with that, I updated all the examples in the integration docs to be sure that no pre-existing fields are ever used. I also updated the "Editing labels with a free Labelbox account" section to show my recommended workflow for editing labels (using a new, temporary label field to store edits).

I read through the tutorial and see that there may need to be some work done there to show valid workflows for "editing" existing labels. I was planning to just copy my recommended best practice over from the integration docs, but I noticed some strange things like how some images contain `skis` but that doesn't appear in any of the label schema definitions that I saw. So, I got cold feet and didn't update the code in the tutorials.

brimoor:
@ehofesmann FYI- After making the docs updates to this PR, I decided that it wasn't overly pressing to add a `SampleCollection.build_label_schema()` method that would generate an annotation-style label schema for one or more existing label fields; the examples in the docs here are fairly clean as is.

Two reasons for this:
-  Such a method would have to take a `backend` parameter, because EG the `type`s chosen for attributes depend on what backend you choose (what widgets they support)
- Ben's work on embedded document schemas will provide a similar field + attribute schema capability, so I want to make sure an annotation-specific method isn't confusing

My current thinking is that we might just add a utility method in `fiftyone.utils.annotation` like so:

```py
build_label_schema(sample_collection, fields, backend="cvat"):
    pass
```

ehofesmann:
> 



> LGTM.
> 
> Per offline discussions, I updated the Labelbox backend to explicitly raise an error if an existing field is passed to `annotate()`, since uploading existing labels is not yet suported.
> 
> To go along with that, I updated all the examples in the integration docs to be sure that no pre-existing fields are ever used. I also updated the "Editing labels with a free Labelbox account" section to show my recommended workflow for editing labels (using a new, temporary label field to store edits).
> 
> I read through the tutorial and see that there may need to be some work done there to show valid workflows for "editing" existing labels. I was planning to just copy my recommended best practice over from the integration docs, but I noticed some strange things like how some images contain `skis` but that doesn't appear in any of the label schema definitions that I saw. So, I got cold feet and didn't update the code in the tutorials.

Thanks for the updates! I've gone ahead and added your method to the tutorial.

PR #1261, opened 2021-09-10
Limit stats results

benjaminpkane:
Limits string field histogram results to the first 200, sorted by count in descending order.

```py
import fiftyone as fo
import fiftyone.zoo as foz

F = fo.ViewField

dataset = foz.load_zoo_dataset("cifar100")
dataset.add_sample_field("list_str", fo.ListField, subfield=fo.StringField)

dataset.set_field("list_str", [F("filepath")]).save() # currently crashes the App when `Other fields` is opened

session = fo.Session(dataset)
```

### DIFF
--- a/app/packages/app/src/components/Distributions.tsx
+++ b/app/packages/app/src/components/Distributions.tsx
@@ -12,6 +12,7 @@ import { useMessageHandler, useSendMessage } from "../utils/hooks";
 import * as selectors from "../recoil/selectors";
 import { AGGS } from "../utils/labels";
 import { filterStages } from "./Filters/atoms";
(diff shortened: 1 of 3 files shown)


brimoor (changes requested):
Confirmed that the limiting behavior works as expected, but now the sample tag graphs fail to render when testing with the example snippet you shared.

benjaminpkane:
Right, fixed

> Confirmed that the limiting behavior works as expected, but now the sample tag graphs fail to render when testing with the example snippet you shared.

Right, thank you. Fixed.

brimoor (approved):
LGTM

PR #1262, opened 2021-09-11
Fixing video dataset reload bug

brimoor:
Fixes a subtle bug that prevented `dataset.reload()` from working for video datasets where the user has a `Frame` instance from the dataset in-memory.

A unit test is added to ensure correct behavior.


benjaminpkane (approved):
LGTM

Issue #1265, opened 2021-09-13
[BUG] Interactive confusion matrix callbacks don't work for large datasets 


brimoor:
A race condition seems to occur when selecting cells in an interactive confusion matrix for a sufficiently large dataset that prevents the App from updating as expected.

The example below (when run as cells in a Jupyter notebook) reliably causes this error:

```py
import random
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone import ViewField as F

tmp_dataset = foz.load_zoo_dataset("coco-2017", split="validation", label_types=["detections"])
```

```py
def jitter(detections, delta, gamma, classes):
    detections = detections.copy()

    for detection in detections.detections:
        alpha = random.random()

        if alpha < gamma:
            detection.label = random.choice(classes)

        bounding_box = np.array(detection.bounding_box)
        bounding_box += alpha * delta * np.array([1, 1, 0, 0])
        detection.bounding_box = bounding_box.tolist()
        detection.confidence = alpha

    return detections
```

```py
counts = tmp_dataset.count_values("ground_truth.detections.label")
classes = sorted(counts, key=counts.get, reverse=True)[:20]

# Create a detection dataset containing 20 classes
dataset = (
    tmp_dataset
    .filter_labels("ground_truth", F("label").is_in(classes))
    .clone()
)

dataset.default_classes = None

# Add predictions
jit = lambda labels: jitter(labels, 0.03, 0.1, classes)
for sample in dataset.iter_samples(progress=True):
    sample["predictions"] = jit(sample["ground_truth"])
    sample.save()

# Perform evaluation
dataset.evaluate_detections("predictions", gt_field="ground_truth", eval_key="eval")

dataset.name = "confusion-matrix-test"
dataset.persistent = True
```

```py
import fiftyone as fo

dataset = fo.load_dataset("confusion-matrix-test")
results = dataset.load_evaluation_results("eval")
```

```py
session = fo.launch_app(dataset)
```

```py
plot = results.plot_confusion_matrix()
plot.show(height=800)
session.plots.attach(plot, name="eval")
```

PR #1266, opened 2021-09-13
Fixing #1265

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1265.

The previous implementation of interactive plotting callbacks relied on all session/plot updates finishing within 0.5 seconds. However, as https://github.com/voxel51/fiftyone/issues/1265 demonstrates, loading views created, for example, by selecting cells in a confusion matrix containing many thousands of labels in a large dataset can easily exceed 0.5 seconds. In such cases, a race condition currently ensues that causes an infinite chain of session/plot updates.

This PR improves the callback logic to prevent such self-update loops. The time-based threshold is still there (now 1 second rather than 0.5) but it's primary function now is just to prevent unnecessary updates due to incessant clicking in the App/plots.

This PR also makes a small change to the behavior of interactive confusion matrices: when updating a confusion matrix in response to an App update, the confusion matrix will "remember" the loaded view, such that any subsequent selections in the confusion matrix will only contain labels from the loaded view, not the entire confusion matrix (until either the plot or the App is reset, at which point this view will be forgotten). This change allows for certain workflows like seeing how a confusion matrix changes in response to selecting a subset of the available samples, where, after selecting a subset of samples, the user would then like to select cells in the confusion matrix to further drill down into specific classes of interest from within the chosen subset of samples (previously selecting cells in a confusion matrix would always load *all* labels in the cell from the evaluation run).

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -704,17 +704,41 @@ class SampleCollection(object):
     def _get_selected_labels(self, ids=None, tags=None, fields=None):
         view = self.select_labels(ids=ids, tags=tags, fields=fields)
 
-        sample_ids = view.values("id")
+        label_fields = view._get_label_fields()
+        if not label_fields:
+            return []
 
-        labels = []
-        for label_field in view._get_label_fields():
+        paths = ["id"]
+        is_list_fields = []
+        is_frame_fields = []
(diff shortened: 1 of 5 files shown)


benjaminpkane (approved):
LGTM!

Issue #1268, opened 2021-09-14
[FR] Please consider upgrading mongoDB to 5.0


robmarkcole:
### Proposal Summary
I would assume this is in the pipeline, but perhaps there are backwards compatibility issues. I am on an M1 mac and despite following the docs am unable to get mongodb-community@4.4 running. However mongodb-community@5.0 runs without issue.

brimoor:
Hi @robmarkcole 👋 

Just to be clear, you're saying that you found [these instructions](https://voxel51.com/docs/fiftyone/user_guide/config.html#configuring-mongodb-connection) but unfortunately you're having issues getting MongoDB v4.4 installed and running yourself?

[This user](https://github.com/voxel51/fiftyone/issues/1165) seemed to indicate that `brew install mongodb-community@4.4` was sufficient, but I was a little surprised about that because I was under the impression that compiling with Rosetta may also be required. Unfortunately I don't have access to an M1 mac to test that.

The user in the above issue also reported issues with other packages like `scikit-image` and `sciki-learn` that seemingly also might be blockers for getting FiftyOne running on M1 macs.

Upgrading to MongoDB v5 is definitely something that we can consider. It was only released July 13th and we haven't started investigating if there are any breaking changes in the new version, but I'm guessing it shouldn't be too much of an issue. Although we warn that we need MongoDB v4.4, that's mostly because are using v4.4 features and so earlier versions will not work. However, I wouldn't be surprised if you can use a MongoDB v5 database now, especially if you [set the feature compatibility version](https://docs.mongodb.com/manual/reference/command/setFeatureCompatibilityVersion) to v4.4 on your database:

```
db.admin.command({"setFeatureCompatibilityVersion": "4.4"})
```

You would probably need to be running a [source install](https://github.com/voxel51/fiftyone#installing-from-source) to do that though, as you would need to edit [this line](https://github.com/voxel51/fiftyone/blob/5063f248a247ffc4a8f56d2d19201d3a190527e4/fiftyone/constants.py#L80) to make FiftyOne allow you to use MongoBD v5.

robmarkcole:
I was able to `brew install mongodb-community@4.4` as well as 5.0, but there was an issue running 4.4 whilst 5.0 ran without issue. RE skimage and sklearn I am regularly using them on M1, so this appears resolved nowadays. 

I will now try the edit you suggest

brimoor:
I was able to install MongoDB v5 and successfully load the quickstart dataset in the App.

[...]

PR #1269, opened 2021-09-14
Allowing MongoDB v5 custom database usages

brimoor:
Motivated by https://github.com/voxel51/fiftyone/issues/1268, this PR tweaks the database version verification logic to allow MongoDB versions greater than v4.4 to be used when connecting to a self-installed database, since the preliminary investigations in https://github.com/voxel51/fiftyone/issues/1268 seem to indicate that this will work, and users in environments like Apple Silicon may find it easier to install MongoDB v5 than older versions.

The documentation for connecting to a self-installed database is also improved, including a full example.


benjaminpkane (approved):
LGTM

PR #1273, opened 2021-09-20
ActivityNet Integration

ehofesmann:
# ActivityNet Integration

This PR adds ActivityNet 100 and 200 to the [FiftyOne Dataset Zoo](https://voxel51.com/docs/fiftyone/user_guide/dataset_zoo/index.html) as well as adding ActivityNet-style temporal detection evaluation.

activitynet-100-validation

## ActivityNet Zoo Datasets

The [FiftyOne Dataset Zoo](https://voxel51.com/docs/fiftyone/user_guide/dataset_zoo/index.html) provides support for loading both the ActivityNet 100 and ActivityNet 200 datasets.

Like all other zoo datasets, you can use `load_zoo_dataset()` to download and load an ActivityNet split into FiftyOne:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Download and load 10 samples from the validation split of ActivityNet 200
    "activitynet-200",
    split="validation",
    max_samples=10,
)

```

**Note**: ActivityNet 200 is a superset of ActivityNet 100 so we have made sure to only store one copy of every video on disk.

### Partial Downloads

In addition, FiftyOne provides parameters that can be used to efficiently download specific subsets of the ActivityNet dataset, allowing you to quickly explore different slices of the dataset without downloading the entire split.

When performing partial downloads, FiftyOne will use existing downloaded data first if possible before resorting to downloading additional data from YouTube.

```py
import fiftyone as fo
import fiftyone.zoo as foz

#
# Load 4 samples from the validation split that
# contain the actions "Bathing dog" and "Walking the dog"
# with a maximum duration of 20 seconds
#
# Videos that contain all `classes` will be prioritized first, followed
# by videos that contain at least one of the required `classes`. If
# there are not enough videos matching `classes` in the split to meet
# `max_samples`, only the available videos will be loaded.
#
# Videos will only be downloaded if necessary
#

    "activitynet-200",
    split="validation",
    classes=["Bathing dog", "Walking the dog"],
    max_duration=20,
    max_samples=4,
)

```

### Full Split Downloads

Many videos have been removed from YouTube since the creation of ActivityNet. Due to this, if you do not specify any partial download parameters classes, max_duration, or max_samples (shown above) and thus the entire split is requested, then you are required to manually download the entire dataset.

In order to manually download the entire source dataset, you must fill out t[his form](https://docs.google.com/forms/d/e/1FAIpQLSeKaFq9ZfcmZ7W0B0PbEhfbTHY41GeEgwsa7WobJgGUhn4DTQ/viewform) which will give you access to the dataset through Google Drive for 7 days.

After downloading the full dataset, it can be loaded into FiftyOne:

```py
import fiftyone as fo
import fiftyone.zoo as foz

source_dir = "/path/to/dir-with-activitynet-zips"

# Load the entire ActivityNet-200 dataset into FiftyOne
dataset = foz.load_zoo_dataset("activitynet-200", source_dir=source_dir)

```

## ActivityNet Evaluation

This PR also adds support for ActivityNet-style evaluation for `TemporalDetection` labels though the [`evaluate_detections()`](https://voxel51.com/docs/fiftyone/user_guide/evaluation.html#detections) method. The resulting mAP has been verified to match that computed by the [ActivityNet evaluation API](https://github.com/activitynet/ActivityNet/tree/master/Evaluation).

While both the ActivityNet evaluation API and FiftyOne can be used to compute mAP, the benefit of using FiftyOne is that it also stores sample- and label-level results in the form of IoUs and TP/FP/FN results. This combined with the powerful querying language of FiftyOne allows you to analyze the exact success and failure modes of your model more easily than ever. Additionally, FiftyOne lets you compute PR curves and confusion matrices from these evaluation results as well.

The example below demonstrates ActivityNet-style temporal detection evaluation on the ActivityNet 200 dataset from the Dataset Zoo:

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

# Load subset of ActivityNet 200
classes = ["Bathing dog", "Walking the dog"]
    "activitynet-200",
    split="validation",
    classes=classes,
    max_samples=10,
)
print(dataset)

# Generate some fake predictions for this example
random.seed(51)
for sample in dataset:
    for det in sample.predictions.detections:
        det.support[0] -= random.randint(-10,10)
        det.support[1] -= random.randint(-10,10)
        det.support[0] = max(det.support[0], 1)
        det.support[1] = max(det.support[1], det.support[0] + 1)
        det.confidence = random.random()
        det.label = random.choice(classes)

    sample.save()

# Evaluate the segments in the `predictions` field with respect to the
# segments in the `ground_truth` field
    "predictions",
    gt_field="ground_truth",
    eval_key="eval",
    compute_mAP=True,
)

# Get the model mAP
print(results.mAP())

# Print a classification report for the classes

# Print some statistics about the total TP/FP/FN counts
print("TP: %d" % dataset.sum("eval_tp"))
print("FP: %d" % dataset.sum("eval_fp"))
print("FN: %d" % dataset.sum("eval_fn"))

# Create a view that has samples with the most false positives first, and
# only includes false positive segments in the `predictions` field
view = (
    dataset
    .sort_by("eval_fp", reverse=True)
    .filter_labels("predictions", F("eval") == "fp")
)

# Visualize results in the App
```


fiftyone/utils/activitynet.py
```diff
@@ -0,0 +1,1038 @@
...
+            -   a filename like ``"labels.json"`` specifying the location of
+                the labels in ``dataset_dir``
+            -   an absolute filepath to the labels. In this case,
+                ``dataset_dir`` has no effect on the location of the labels
+
+            If None, the parameter will default to ``labels.json``
+        compute_metadata (False): whether to produce
+            :class:`fiftyone.core.metadata.VideoMetadata` instances for each
+            video when importing
+        shuffle (False): whether to randomly shuffle the order in which the
```
brimoor, line 171: I think `ActivityNetDatasetImporter` needs to implement the `classes` and `max_duration` parameters from `download_activitynet_split()`, otherwise if the user downloads a full split but then runs:

```py
    "activitynet-200",
    split="validation",
    classes=["Bathing dog", "Walking the dog"],
    max_duration=20,
)
```

all samples will be loaded, not just the ones matching `classes` and `max_duration` that would be loaded if the same method were being run for the very first time.

This was also true for COCO and OpenImages, where we had to implement the partial loading logic in both places.

brimoor: Perhaps related, I successfully ran these:

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "activitynet-100",
    split="validation",
    max_samples=10,
)

    "activitynet-200",
    split="validation",
    classes=["Bathing dog", "Walking the dog"],
    max_duration=20,
    max_samples=4,
)
```

But then I tried (after exiting and reopening ipython):

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "activitynet-200",
    split="validation",
    classes=["Bathing dog", "Walking the dog"],
)
```

and got the following error:

```
~/dev/fiftyone/fiftyone/utils/activitynet.py in download_activitynet_split(dataset_dir, split, source_dir, classes, max_duration, copy_files, num_workers, shuffle, seed, max_samples, version)
    122             sample_ids,
    123             num_downloaded_samples,
--> 124         ) = manager.download_necessary_samples(download_config)
    125         num_samples = num_existing + num_downloaded_samples
    126

~/dev/fiftyone/fiftyone/utils/activitynet.py in download_necessary_samples(self, config)
    549             requested_num,
    550             num_workers,
--> 551             split,
    552         )
    553         requested_sample_ids.extend(downloaded_samples)

~/dev/fiftyone/fiftyone/utils/activitynet.py in _download_requested_samples(self, ids, class_samples, shuffle, requested_num, num_workers, split)
    585
    586             downloaded_ids = self._separate_versions_and_attempt_to_download(
--> 587                 ids, class_samples, requested_num, num_workers, split,
    588             )
    589

~/dev/fiftyone/fiftyone/utils/activitynet.py in _separate_versions_and_attempt_to_download(self, ids, samples_info, num_samples, num_workers, split)
    608                 a100_ids,
    609                 samples_info,
--> 610                 min(num_a100_ids, num_samples),
    611                 num_workers,
    612             )

TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

fiftyone/utils/activitynet.py
```diff
@@ -0,0 +1,1038 @@
...
+        if self._classes is not None:
+            info["classes"] = self._classes
+
+        if self._taxonomy is not None:
+            info["taxonomy"] = self._taxonomy
+
+        return info
+
+
+class ActivityNetDownloadConfig(object):
```
brimoor, line 242: Following up on some offline conversation- this data sharing between the 100 and 200 dataset versions has been working well for me, so I have no problem with keeping the implementation as-is

fiftyone/utils/activitynet.py
```diff
@@ -0,0 +1,1038 @@
...
+    def data_json_path(self, split):
+        return os.path.join(self.split_dir(split), "data.json")
+
+    def labels_path(self, split):
+        return os.path.join(self.split_dir(split), "labels.json")
+
+    def error_path(self, split):
+        return os.path.join(self.split_dir(split), "download_errors.json")
+
+    def ensure_data_dirs(self):
```
brimoor, line 801: nit: I noticed that the split directories are created regardless of whether I've actually downloaded any content for that split. All other zoo datasets only create the directories if the split has actually been downloaded

I sometimes check the directory structure as a quick way of seeing what splits I've downloaded. Would be nice if this were possible here, too

fiftyone/utils/eval/activitynet.py
```diff
@@ -0,0 +1,592 @@
...
+            classes=classes,
+            missing=missing,
+            samples=samples,
+        )
+        self.precision = np.asarray(precision)
+        self.recall = np.asarray(recall)
+        self.iou_threshs = np.asarray(iou_threshs)
+        self._classwise_AP = classwise_AP.mean(0)
+
+    def plot_pr_curves(self, classes=None, backend="plotly", **kwargs):
```
brimoor, line 336: https://github.com/voxel51/fiftyone/pull/1490 added two new things to this method for COCO evaluation
- Added an `iou_thresh` parameter to plot a specific threshold's curves
- Passed confidence thresholds through to the plotting routine (for tooltips)

Can you add the same features here please?

brimoor (changes requested):
Good stuff! Just left a couple comments for you that shouldn't be too complicated to add before we release this 📈

fiftyone/utils/data/importers.py
```diff
@@ -676,7 +676,11 @@ def _load_data_map(data_path, ignore_exts=False, recursive=False):
                 )
 
             data_map = etas.read_json(data_path)
-            return {to_uuid(k): v for k, v in data_map.items()}
+            data_root = os.path.dirname(data_path)
+            return {
+                to_uuid(k): os.path.join(data_root, v)
```
brimoor, line 681: This didn't need to be implemented as an optional `abspath=True` behavior: it's good to always interpret relative paths w.r.t. the location of the manifest

Note that `os.path.join("/this/is/not/used", "/a/b/c.txt") == "/a/b/c.txt"`, so extra conditional logic to handle whether `v` is absolute or not is not necessary.

brimoor:
When I do my first download, I see this:

```shell
Downloading split 'validation' to '/Users/Brian/fiftyone/activitynet-200/validation' if necessary
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████|   35.9Mb/35.9Mb [597.9ms elapsed, 0s remaining, 60.0Mb/s]      
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████|   37.7Mb/37.7Mb [609.7ms elapsed, 0s remaining, 61.8Mb/s]      
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████|   37.7Mb/37.7Mb [601.1ms elapsed, 0s remaining, 62.7Mb/s]      
Downloading videos...
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████|   35.9Mb/35.9Mb [628.0ms elapsed, 0s remaining, 57.1Mb/s]      
Failed to download video 'https://www.youtube.com/watch?v=-F7QWQA8Eh8': Video unavailable
Failed to download video 'https://www.youtube.com/watch?v=-79MZQX4CEA': Video unavailable
Failed to download video 'https://www.youtube.com/watch?v=-U4lNtzVQ8s': Sign in to confirm your age                                                                         
Failed to download video 'https://www.youtube.com/watch?v=-0r0HEwAYiQ': 'NoneType' object has no attribute 'span'
Failed to download video 'https://www.youtube.com/watch?v=-VcxQ6i6Ejk': Sign in to confirm your age                                                                          
Failed to download video 'https://www.youtube.com/watch?v=-Z98HU6T7J8': This video has been removed for violating YouTube's policy on violent or graphic content
Failed to download video 'https://www.youtube.com/watch?v=-ZDCHvzbnoU': Video unavailable                                                                                    
 100% |███████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [17.0s elapsed, 0s remaining, 0.6 videos/s]    
Dataset info written to '/Users/Brian/fiftyone/activitynet-200/info.json'
Loading existing dataset 'activitynet-200-validation-10'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
```

Three questions:
- Why are there four total downloads before the videos are downloaded? Two each of 35.9Mb and 37.7Mb files? I haven't looked at the code, but this seems strange. I could see one for the 100 labels and another for the 200 labels, but what else?
- It's misleading that the progress bar immediately after `Downloading videos...` is not actually tracking video downloads
- In general, progress bars are only value-add for steps that take a long time. What about just omitting the non-video progress bars?

fiftyone/utils/activitynet.py
```diff
@@ -0,0 +1,1175 @@
...
+    ):
+        download_urls = []
+        download_paths = []
+        for sample_id in ids:
+            sample_info = samples_info[sample_id]
+            download_path = os.path.join(videos_dir, "v_%s.mp4" % sample_id)
+            download_urls.append(sample_info["url"])
+            download_paths.append(download_path)
+
+        downloaded, errors = fouy.download_youtube_videos(
```
brimoor, line 698: I didn't investigate the code in detail, but in your implementation, do you rely on the downloaded video exactly matching `download_path` above? Because it might not; eg it could be a different format if no MP4 streams are available.

ehofesmann:
> Three questions:
> 
> * Why are there four total downloads before the videos are downloaded? Two each of 35.9Mb and 37.7Mb files? I haven't looked at the code, but this seems strange. I could see one for the 100 labels and another for the 200 labels, but what else?
> * It's misleading that the progress bar immediately after `Downloading videos...` is not actually tracking video downloads
> * In general, progress bars are only value-add for steps that take a long time. What about just omitting the non-video progress bars?
> 
> BTW I do like seeing the "Failed to download" warnings because it serves as a good reminder that partial downloads from YouTube are strictly for protoyping; if you need specific files, you should download the full dataset from the maintainers.

1) Yes, it needs 2 downloads to get the 100 and 200 labels, but it needs it in 2 different places. I've updated the code to just copy it to the second place instead of downloading again.

2) I haven't seen that, in the code, it prints that it is downloading videos just before downloading them so it may have been a printing issue? Either way, resolved by 3

3) Agreed, the label downloads are quiet now.

fiftyone/utils/activitynet.py
ehofesmann: Thanks for pointing that out. No, it doesn't exactly need to match. MP4 streams are just preferred.

brimoor:
I checked other zoo datasets and they do often show progress bars for label downloads, so I went ahead and removed `quiet=True` for consistency. The duplicate downloads was the main point point of confusion.

brimoor (approved):
LGTM

PR #1277, opened 2021-09-20
Expose CVAT API parameters to upload higher quality video

ehofesmann:
When uploading data to CVAT manually, the options for `use_cache` and `use_zip_chunks` are checked by default. https://openvinotoolkit.github.io/cvat/docs/manual/basics/creating_an_annotation_task/#advanced-configuration
image

Through the CVAT REST API, these options are set to `False` by default. This PR exposes these parameters to the user when calling `SampleCollection.annotate()` with the CVAT backend similar to the existing `image_quality`.

The primary issue that this resolves is that when `use_zip_chunks` is False, there is a noticeable smoothing and reduction in quality when uploading videos at an `image_quality` of 100.

When `use_zip_chunks=True`:
image

When `use_zip_chunks=False`:
image

This can be tested with the following code:

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset.annotate(
    "high_quality",
    label_field="high_q_field",
    label_type="scalar",
    classes=["test"],
    image_quality=100,
    launch_editor=True,
    use_zip_chunks=True,
)

dataset.annotate(
    "low_quality",
    label_field="low_q_field",
    label_type="scalar",
    classes=["test"],
    image_quality=100,
    launch_editor=True,
    use_zip_chunks=False,
)
```

### DIFF
--- a/docs/source/integrations/cvat.rst
+++ b/docs/source/integrations/cvat.rst
@@ -404,6 +404,13 @@ provided:
     Not applicable to videos
 -   **image_quality** (*75*): an int in `[0, 100]` determining the image
     quality to upload to CVAT
+-   **use_cache** (*True*): option to use a cache when preparing data chunks.
+    This will reduce the task creation time as data will be processed
+    on-the-fly and stored in a cache when requested
+-   **use_zip_chunks** (*True*): for videos only, this option will force to use
+    zip chunks as compressed data. Note, setting this option to `False`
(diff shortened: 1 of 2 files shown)


brimoor (approved):
LGTM. I added the `chunk_size` parameter as well, which allows for customizing the number of image/video frames in each chunk that is uploaded (seems to affect both cases).

PR #1278, opened 2021-09-20
App video issues

### LINKED ISSUE #1274: [BUG] Sound not muted in the App by default
When playing a video with sound in the App, the volume shows that it is muted by default but the sound still plays. If you unmute and mute again, then it will actually be muted.

This can be replicated by loading in a couple of videos with sound (not `quickstart-video`) . For example, you can use the following code with #1273 

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("activitynet-200", split="validation", max_samples=2)
```
Screenshot from 2021-09-20 13-49-522

### LINKED ISSUE #1275: [BUG] Video sound does not stop playing when exiting expanded view
When viewing videos with sound in the expanded view in the App, if the video is not paused when going to the next sample or exiting the expanded view then the audio will continue to play with no way to stop it outside of refreshing the App.

This can be replicated by loading in a couple of videos with sound (not `quickstart-video`) . For example, you can use the following code with #1273 

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("activitynet-200", split="validation", max_samples=2)
```
1) Click on a video to bring up the expanded view
2) Start playing it and turn on the sound
3)  Exit the expanded view or arrow over to a new sample

Going back to the original sample does not let you stop the audio from playing, if you play the video again, it will start to play the audio on top of what is already playing.

### LINKED ISSUE #1276: [BUG] Video playback time prepends a 0 to seconds
When viewing a video in the App, the displayed seconds prepend a 0 when the number of digits increases (when it goes from 9 to 10 seconds)

For example, this is 12 seconds into a 1 minute 15 second video:
Screenshot from 2021-09-20 13-49-522

This can be replicated by loading in a couple of videos longer than 10 seconds (not `quickstart-video`) . For example, you can use the following code with #1273 

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("activitynet-200", split="validation", max_samples=2)
```

benjaminpkane:
### Issues

* Resolves #1276
* Resolves #1275
* Resolves #1274

### DIFF
--- a/app/packages/looker/src/elements/util.ts
+++ b/app/packages/looker/src/elements/util.ts
@@ -78,7 +78,7 @@ const stringifyNumber = function (
   pad: boolean = false
 ): string {
   let str = "";
-  if (pad) {
+  if (pad && number <= 10) {
     str += "0" + number;
   } else if (number === 0) {
     str = "0";
--- a/app/packages/looker/src/elements/video.ts
+++ b/app/packages/looker/src/elements/video.ts
@@ -344,7 +344,7 @@ export class VideoElement extends BaseElement<VideoState, HTMLVideoElement> {
   private release: () => void;
   private src: string;
   private update: StateUpdate<VideoState>;
-  private volume: number = 0;
+  private volume: number;
   private waitingToPause: boolean = false;
   private waitingToPlay: boolean = false;
   private waitingToRelease: boolean = false;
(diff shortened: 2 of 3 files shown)


ehofesmann (approved):
LGTM, this fixed all of my issues. Thanks!

PR #1279, opened 2021-09-21
Expose the mutable property for CVAT attributes

ehofesmann:
When annotating object tracks on videos in CVAT, the `mutable` property of the attributes determines if the attribute can change in value from one frame to another (`mutable=True`) or if the value of the attribute must remain fixed throughout the track (`mutable=False`). At the moment (and by default in the future unless specified), all attributes are set to `mutable=True`.

This PR allows you to configure the mutable property of attributes in the label schema. It is expected that any future backend that allows for object attribute annotation in video will also support some notion of `mutable`.

Annotating new immutable attributes:
```python
import fiftyone as fo
import fiftyone.zoo as foz

attrs = {
    "occluded": {
        "type": "radio",
        "values": [0, 0.25, 0.5, 1],
        "default": 0,
        "mutable": False,
    }
}

dataset.annotate(
    "annot",
    attributes=attrs,
    launch_editor=True,
)

# Annotate some of the attributes in CVAT
input()

dataset.load_annotations("annot", cleanup=True)
```

Loading existing immutable attributes from FiftyOne into CVAT:
```python
attrs = {
    "occluded": {
        "type": "radio",
        "values": [0, 0.25, 0.5, 1],
        "default": 0,
        "mutable": False,
    }
}

dataset.annotate(
    "annot2",
    attributes=attrs,
    launch_editor=True,
)

# Change some of the attributes in CVAT
input()

dataset.load_annotations("annot2", cleanup=True)
```

### Updated documnetation
Screenshot from 2021-09-21 15-16-34

### DIFF
--- a/docs/source/integrations/cvat.rst
+++ b/docs/source/integrations/cvat.rst
@@ -532,10 +532,12 @@ attribute that you wish to label:
             "type": "radio",
             "values": [True, False],
             "default": True,
+            "mutable": True,
         },
         "weather": {
             "type": "select",
             "values": ["cloudy", "sunny", "overcast"],
+            "mutable": False,
         },
         "caption": {
             "type": "text",
@@ -545,7 +547,7 @@ attribute that you wish to label:
     view.annotate(
         anno_key,
         label_field="new_field",
-        label_type="detections",
+        label_type="frames.detections",
         classes=["dog", "cat", "person"],
         attributes=attributes,
     )
@@ -564,6 +566,7 @@ For CVAT, the following `type` values are supported:
 -   `checkbox`: a boolean checkbox UI. In this case, `default` is optional and
     `values` is unused
 
+
 When you are annotating existing label fields, the `attributes` parameter can
 take additional values:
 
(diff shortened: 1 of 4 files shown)


brimoor:
@ehofesmann how about instead adding `mutable` to the top-level `attributes` dict as an optional attribute that defaults to `True` for frame-level fields? Seems simpler to me.

There aren't any backwards compatibility issues as long as this attribute is implemented as optional (and anyways, same would apply to `other`, right?)

ehofesmann:
> @ehofesmann how about instead adding `mutable` to the top-level `attributes` dict as an optional attribute that defaults to `True` for frame-level fields? Seems simpler to me.
> 
> There aren't any backwards compatibility issues as long as this attribute is implemented as optional (and anyways, same would apply to `other`, right?)

Do you mean something like this:
1) 
```python
attrs = {
    "occluded": {
        "type": "radio",
        "values": [0, 0.25, 0.5, 1],
        "default": 0,
    },
    "mutable": False,
}
```

or like this:
2) 
```python
attrs = {
    "occluded": {
        "type": "radio",
        "values": [0, 0.25, 0.5, 1],
        "default": 0,
        "mutable": False,
    }
}
```

The first option wouldn't work because users might want to have some mutable and some immutable attributes on the same object.

The second option can work, I just thought it would be cleaner to separate out backend-specific options with the `other`. It also makes it clean to let the backend handle formatting any of the `other` properties it needs: https://github.com/voxel51/fiftyone/blob/7d7a3e0c2bcc5076949b82471926f771c4c15e3a/fiftyone/utils/annotations.py#L535-L536

 I'm open to changing this if you feel strongly about it though.

There shouldn't be any backwards compatibility issues either way. Previous label schemas will continue to default to `mutable=True` and these new label schemas run on an old version would just ignore `other`.

brimoor:
I meant like this:

```
attrs = {
    "occluded": {
        "type": "radio",
        "values": [0, 0.25, 0.5, 1],
        "default": 0,
        "mutable": False,
    }
}
```

And to clarify, I don't see this as a backend-specific thing. I would document `mutable` in the backend-agnostic API, since it always makes sense to specify whether a temporal attribute is mutable or not.

If a certain backend doesn't support one option, it would then warn that it can't comply.

ehofesmann:
@brimoor That's a good point, `mutable` should be a standard property across backends. I'll remove `other` until something comes up that would require it. 

Just FYI, it seems that Labelbox does not have this option at the moment, all attributes are always mutable. However, they also don't allow uploading object labels for videos, so this won't be an issue.

ehofesmann:
@brimoor Your suggestions are in

brimoor (approved):
LGTM

PR #1283, opened 2021-09-22
Relax `opencv-python-headless` version requirement

### LINKED ISSUE #1282: [FR] update opencv-python-headless
### Proposal Summary
Currently this repo requires opencv-python-headless<=4.4.0.46. To cut a long story short there are no wheels available for python3.9 and I am unable to install fiftyone (I am using docker `image: jupyter/scipy-notebook:latest`). However version `4.5.3.56` is available for install without issue, and I propose updating the requirement for this dependency.

benjaminpkane:
Removes the pin on `opencv-python-headless`. The original reason for this was to prevent building wheels from source for new installs during the few hours the source dist was available, but binary dists were not.

`opencv-python-headless` is popular package, only updated a few times a year, and therefore a loose requirement is likely best.

### Issues

* Resolves #1282

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,6 +78,7 @@ setup(
         "mongoengine==0.20.0",
         "motor>=2.3,<3",
         "numpy",
+        "opencv-python-headless>=4.4,<5",
         "packaging",
         "pandas",
         "Pillow>=6.2",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -78,6 +78,7 @@ def get_version():
         "mongoengine==0.20.0",
         "motor>=2.3,<3",
         "numpy",
+        "opencv-python-headless>=4.4,<5",
```
brimoor, line 81: Is there a reason to include a version requirement at all? I'm not aware of anything we're using that has or will change anytime soon.

brimoor (approved):
LGTM (left a comment for consideration)

setup.py
benjaminpkane: I'll remove

PR #1284, opened 2021-09-22
Only warn during `mongod` compatibility validation

benjaminpkane:
Relaxes `mongod` validation to only warn if an incompatible version is found. One reason for this change is it enables connections to Amazon DocumentDB (no promises).

Tested with this [guide](https://docs.aws.amazon.com/en_us/documentdb/latest/developerguide/connect-ec2.html)

### DIFF
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -123,7 +123,7 @@ def _validate_db_version(config, client):
     if config.database_validation and not (min_ver <= version < max_ver):
         raise RuntimeError(
             "Found `mongod` version %s, but only [%s, %s) are compatible. "
-            "You can suppress this exception if desired by setting your "
+            "You can suppress this exception by setting your "
             "`database_validation` config parameter to `False`. See "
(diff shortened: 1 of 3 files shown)


fiftyone/core/odm/database.py
```diff
@@ -126,13 +126,13 @@ def _validate_db_version(config, client):
         # that the user has set their feature compatibility version
         # https://docs.mongodb.com/manual/reference/command/setFeatureCompatibilityVersion
         if version < min_ver:
-            raise RuntimeError(
-                "Found `mongod` version %s, but %s or later is required"
+            logger.warning(
```
brimoor, line 129: I hate to go too architecture astronaut-y here, but if we're going to allow this, then perhaps we should also remove the `database_uri` conditional and also provide a config setting that users who are intentionally violating this can set to avoid seeing this warning

benjaminpkane: Ok, I've added the `database_validation` setting and simplified things to one exception. I've changed it back to a `RuntimeError` so a user really has to understand what they are doing.

brimoor (approved):
LGTM

PR #1296, opened 2021-09-23
Including missing GT/predictions by default in confusion matrices

brimoor:
Makes a small change to the default behavior of confusion matrices plotted via the following syntax:

```py
plot = results.plot_confusion_matrix(classes=classes)
plot.show()
```

Previously, if `classes` are provided, no row/column for missing GT/predictions would be included in the confusion matrix unless you explicitly included `results.missing` in the `classes` list.

Now, the row/column for missing GT/predictions is *always* included if there are any, unless you pass the optional `include_missing=False` flag to explicitly remove them.

I think this is a better default behavior, especially for detection evaluation. The issue arose because "missing" is not generally an issue for classification evaluation, and both cases share the same code.

As the examples below demonstrate, the new default behavior now "shows the right thing" for both detection and classification.

### Detection example

```py
import fiftyone as fo
import fiftyone.zoo as foz

results = dataset.evaluate_detections("predictions", eval_key="eval")

results.print_report()
plot = results.plot_confusion_matrix()
plot.show()

results.print_report(classes=["person", "cat", "dog"])
plot = results.plot_confusion_matrix(classes=["person", "cat", "dog"])
plot.show()

plot = results.plot_confusion_matrix(classes=["person", "cat", "dog"], include_other=False)
plot.show()

plot = results.plot_confusion_matrix(classes=["person", "cat", "dog"], include_missing=False, include_other=False)
plot.show()
```

### Classification example

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("cifar10", split="test", max_samples=1000, shuffle=True)

def jitter(val):
    if random.random() < 0.10:
        return random.choice(classes)

    return val

predictions = [
]

results = dataset.evaluate_classifications("predictions", gt_field="ground_truth", eval_key="eval_simple")

results.print_report()
plot = results.plot_confusion_matrix()
plot.show()

results.print_report(classes=["airplane", "cat", "dog"])
plot = results.plot_confusion_matrix(classes=["airplane", "cat", "dog"])
plot.show()

plot = results.plot_confusion_matrix(classes=["airplane", "cat", "dog"], include_other=False)
plot.show()
```

### DIFF
--- a/fiftyone/utils/eval/base.py
+++ b/fiftyone/utils/eval/base.py
@@ -73,6 +73,9 @@ class BaseEvaluationResults(foe.EvaluationResults):
 
     def _get_labels(self, classes, include_missing=False):
         if classes is not None:
+            if include_missing and self.missing not in classes:
+                classes = list(classes) + [self.missing]
+
             return np.asarray(classes)
 
         if include_missing:
@@ -91,7 +94,7 @@ class BaseEvaluationResults(foe.EvaluationResults):
         Returns:
             a dict
         """
-        labels = self._get_labels(classes, include_missing=False)
+        labels = self._get_labels(classes)
 
         if self.ytrue.size == 0 or labels.size == 0:
             d = {}
@@ -136,7 +139,7 @@ class BaseEvaluationResults(foe.EvaluationResults):
         Returns:
             a dict
         """
-        labels = self._get_labels(classes, include_missing=False)
+        labels = self._get_labels(classes)
 
         accuracy = _compute_accuracy(
             self.ytrue, self.ypred, labels=labels, weights=self.weights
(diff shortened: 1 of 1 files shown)


ehofesmann (approved):
LGTM, I agree that this behavior should be the default.

The only gripe I have is that when specifying classes and including missing and other, then the missing+other cell of the confusion matrix will likely be a much higher value than the remaining cells which blows out the color gradient. 

image

That being said, I'm not sure we can do much about that if you want to see both missing and other at the same time. The only solution would be to set that cell to `0` or some value like `None` or `-1`, but I don't really like that. If the cell appears then it should be populated with the correct value.

brimoor:
@ehofesmann yep see what you mean, but I think this new behavior is the best trade-off of all considerations as a default case.

yoeldk:
@brimoor  I think a confusion matrix should always be a squared NXN matrix, so "other" should also be added to the rows. I don't see a reason to prefer "classes predicted as others" over "others predicted as classes" by having "other" only in the columns as it appears in the image above.

PR #1297, opened 2021-09-24
Adding support for filtering object trajectories

brimoor:
Adds an optional `filter_labels(..., trajectories=True)` flag that allows for matching entire object trajectories (ie, a sequence of labels with the same `index` in a video) for which a given filter matches the object in at least one frame of the video.

This is useful, for example, for matching object trajectories based on time-varying attributes, eg., viewing object trajectories that are occluded at some point during the video.

### Example usage

This example is a bit contrived since one would expect `type` to be an immutable attribute:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Only include object trajectories whose type is `sedan` in at least one frame
view = dataset.filter_labels("frames.detections", F("type") == "sedan", trajectories=True)

session.view = view
```

A more realistic example would be a case where the objects have an `occluded = {'light', 'heavy', None}` attribute that can vary between frames:

```py
# Only include object trajectories that are heavily occluded in at least one frame
view = dataset.filter_labels("frames.detections", F("occluded") == "heavy", trajectories=True)
```

### DIFF
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -1155,8 +1155,8 @@ class FilterField(ViewStage):
     def __init__(self, field, filter, only_matches=True, _new_field=None):
         self._field = field
         self._filter = filter
-        self._new_field = _new_field or field
         self._only_matches = only_matches
+        self._new_field = _new_field or field
         self._is_frame_field = None
         self._validate_params()
 
@@ -1634,9 +1634,9 @@ class FilterLabels(ViewStage):
     ):
         self._field = field
         self._filter = filter
-        self._new_field = _new_field or field
         self._only_matches = only_matches
         self._trajectories = trajectories
+        self._new_field = _new_field or field
         self._prefix = _prefix
         self._labels_field = None
         self._is_frame_field = None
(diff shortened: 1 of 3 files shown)


fiftyone/core/stages.py
```diff
@@ -1795,9 +1889,66 @@ def _get_frames_list_field_only_matches_expr(field):
...
+        )
+    else:
+        raise ValueError(
+            "Cannot filter trajectories for field '%s' of type %s"
+            % (field, label_type)
+        )
+
+    indexes_expr = F("frames").reduce(reduce_expr, [])
+
+    set_pipeline = [{"$set": {"_indexes": indexes_expr.to_mongo()}}]
```
benjaminpkane, line 1926: Converting the index lists to a set with `{ $setIntersection: [ "$_indexes", "$_indexes" ] }` or similar might be a small optimization. Or only creating index sets with `$addToSet`.

benjaminpkane (approved):
LGTM! Very useful

fiftyone/core/stages.py
brimoor: Oh good point.

I tried using set arithmetic both during the `$reduce` and at the end like you mentioned, and they were actually pretty similar in runtime, but, since `$addToSet` isn't directly supported in `$reduce`, the `$setUnion` workflow likely wouldn't scale as well for very large datasets, so I went with the "at the end" option.

This change makes the pipeline ~35% faster on my test dataset. Thanks for the idea! 💪

PR #1300, opened 2021-09-25
Adding a to_trajectories() stage

brimoor:
Adds a `to_trajectories(field)` view stage that generates a clips view that contains one clip per object trajectory (ie., a collection of objects with the same `index` in a video) in a frame-level field of a video dataset.

Unlike `to_clips(field, trajectories=True)`, which also generates one clip per object trajectory (although this syntax is not prominently documented) but includes all frame-level labels, `to_trajectories()` selects *only* the single object defining the trajectory in each clip.

Note that `to_trajectories()` is the video analogue of `to_patches()` for image datasets.

### TODO

- Resolve the `@todo` in `fiftyone.core.state`

### Discussion

I'd like to let this PR bake a bit, because, with the current implementation, I've empirically observed that the aggregations for trajectories views can be quite expensive to compute, and it might be good to optimize before officially releasing this feature.

Under the hood, all frame-level labels are attached to each clip and then filtered via a `filter_labels()` stage to retrieve the defining object trajectory for each clip. The good thing about this is that (like other clips views) it avoids creating a temporary frame collection to serve the view. The bad thing is that the computations quickly explode because the number of dense frames that must be filtered in this setup is `O(# bboxes)`, which can be much greater than computing stats for the source video collection, which is `O(# video frames)`.

For example, in the snippet below, we have:

```
len(dataset)  # 10
dataset.count("frames")  # 1279

len(trajectories)  # 109
trajectories.count("frames")  # 7565
```

### Example usage

Here we create a trajectories view that contains one clip per vehicle in the dataset:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Create a trajectories view for the vehicles in the dataset
trajectories = (
    dataset
    .filter_labels("frames.detections", F("label") == "vehicle")
    .to_trajectories("frames.detections")
)
print(trajectories)

session = fo.launch_app(view=trajectories)
```

### DIFF
--- a/fiftyone/core/clips.py
+++ b/fiftyone/core/clips.py
@@ -17,6 +17,7 @@ import fiftyone.core.fields as fof
 import fiftyone.core.labels as fol
 import fiftyone.core.media as fom
 import fiftyone.core.sample as fos
+import fiftyone.core.stages as fost
 import fiftyone.core.validation as fova
 import fiftyone.core.view as fov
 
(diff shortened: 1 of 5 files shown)


fiftyone/core/clips.py
```diff
@@ -406,6 +407,131 @@ def _sync_source_keep_fields(self):
...
+            trajectory_stages.append(exclude_stage)
+
+        #
+        # Select correct trajectory
+        #
+
+        filter_stage = fost.FilterLabels(
+            clips_stage.field,
+            (
+                (F("label") == F("$" + field + ".label"))
```
swheaton, line 524: Don't fully understand what this is. I'm reading "$" means look at the sample itself not the embedded document.
So does this basically group clips into equal label/index combos?

Might just be my relative ignorance of ViewExpressions

fiftyone/core/collections.py
```diff
@@ -6183,6 +6183,57 @@ def to_clips(self, field_or_expr, **kwargs):
...
+
+                -   :class:`fiftyone.core.labels.Detections`
+                -   :class:`fiftyone.core.labels.Polylines`
+                -   :class:`fiftyone.core.labels.Keypoints`
+            **kwargs: optional keyword arguments for
+                :meth:`fiftyone.core.clips.make_clips_dataset` specifying how
+                to perform the conversion
+
+        Returns:
+            a :class:`fiftyone.core.clips.ClipsView`
```
swheaton, line 6233: technically this is true, but shouldn't we say `TrajectoriesView` ?

swheaton (commented):
tests?

code looks fine, though I'm still not 100% sure I understand what's happening

fiftyone/core/clips.py
brimoor: yeah what this is doing is filtering the labels in `clips_stage.field` so that only the ones whose `label` attribute match the `field.label` attribute at the sample-level are retained.

In trajectory views, every clip has a sample-level field that contains the unique `(label, index)` for the unique object associated with the clip.

fiftyone/core/collections.py
brimoor: good point, will update 👍

brimoor:
> tests?

test added 👍

swheaton (approved):
thanks for adding tests! lgtm

PR #1301, opened 2021-09-26
Adding set operation expressions

brimoor:
Adds `ViewExpression` methods for some previously missing set operations, including `is_subset()`, `set_equals()`, `unique()`, `union()`, `intersection()`, `difference()`, and `contains(..., all=True)`.

See the added documentation for example usages.


benjaminpkane (approved):
LGTM. Set operations for the win!

PR #1304, opened 2021-09-28
Add server address option to App sessions

benjaminpkane:
Adds a server address in the following places:

* `FiftyOneConfig.default_app_address`
* `address` in commands that launch a session `Session`

The default is `None`, which maintains current behavior. One use for this parameter is to restrict session's `localhost`.


brimoor (approved):
LGTM. I added some docs explaining why one might use this

PR #1306, opened 2021-09-28
[WIP] Add ActivityNet-style VideoClassification evaluation

ehofesmann:
Add support for evaluation of the temporal activity detection challenge from ActivityNet. Labels like temporal activity detections which are video-level annotations containing a class name and frame range exist in FiftyOne as `VideoClassification` labels. 

This PR adds:
* `SampleCollection.evaluate_video_classifications()` with the default method being `"activitynet"`-style

**Example**

```python
import fiftyone as fo
import fiftyone.zoo as foz

import random
random.seed(51)

def create_fake_preds(dataset):
    for sample in dataset:
        for vc in sample.predictions.classifications:
            vc.support[0] -= random.randint(-10,10)
            vc.support[1] -= random.randint(-10,10)
            vc.support[0] = max(1, vc.support[0])
            vc.support[1] = max(1, vc.support[1])
            vc.confidence = random.random()
            if random.random() > 0.9:
                vc.label = "Other"

        sample.save()

dataset = foz.load_zoo_dataset("activitynet-200", split="validation", max_samples=10)
create_fake_preds(dataset)

results = dataset.evaluate_video_classifications(
    "predictions",
    compute_mAP=True,
    eval_key="eval",
)

print(results.mAP())

```

### TODO

* Compare resulting mAP with the official implementation

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1840,6 +1840,59 @@ class SampleCollection(object):
             **kwargs,
         )
 
+    def evaluate_video_classifications(
+        self,
+        pred_field,
+        gt_field="ground_truth",
+        eval_key=None,
+        classes=None,
+        missing=None,
+        method="activitynet",
+        iou=0.50,
+        classwise=True,
+        **kwargs,
+    ):
+        """Evaluates the video classification predictions in the given collection with
+        respect to the specified ground truth labels. These labels are often used
+        for tasks like temporal action detection.
+    
+        Args:
+            pred_field: the name of the field containing the predicted
+                :class:`fiftyone.core.labels.VideoClassification` instances
+            gt_field ("ground_truth"): the name of the field containing the ground
+                truth :class:`fiftyone.core.labels.VideoClassification` instances
+            eval_key (None): an evaluation key to use to refer to this evaluation
+            classes (None): the list of possible classes. If not provided, classes
+                are loaded from :meth:`fiftyone.core.dataset.Dataset.classes` or
+                :meth:`fiftyone.core.dataset.Dataset.default_classes` if
+                possible, or else the observed ground truth/predicted labels are
+                used
+            missing (None): a missing label string. Any None-valued labels are
+                given this label for results purposes
+            method ("activitynet"): a string specifying the evaluation method to use.
+                Supported values are ``("activitynet")``
+            iou (0.50): the IoU threshold to use to determine segment matches
+            classwise (True): whether to only match segments with the same class
+                label (True) or allow matches between classes (False)
+            **kwargs: optional keyword arguments for the constructor of the
+                :class:`VideoClassificationEvaluationConfig` being used
+    
+        Returns:
+            a :class:`VideoClassificationResults`
+        """
+        return foue.evaluate_video_classifications(
+            self,
+            pred_field,
+            gt_field=gt_field,
+            eval_key=eval_key,
+            classes=classes,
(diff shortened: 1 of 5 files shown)


brimoor:
@ehofesmann following up on our online discussion earlier today, I'm leaning towards exposing this feature under `evaluate_classifications()`, unless we rename `VideoClassification` to something else.

Here's my line of reasoning:
- "Actual" video classification evaluation, where you have a video dataset and `Classification` labels, is done via `evaluate_classifications()`. So it would likely be confusing to have another method called `evaluate_video_classifications()` that *doesn't* do "actual" video classification
- Re-using `evaluate_classifications()` does not cause any conflicts, because the evaluation protocol to use is strictly defined by the fact that one is working with fields of type `VideoClassification`
- Although this evaluation protocol is technically video activity "detection" and does therefore need to deal with many of the same concepts as object detection evaluation, using `evaluate_detections()` seems more confusing than `evaluate_classifications()`, since the `Label` type is `VideoClassification`

If it would help, it's not too late to rename `VideoClassification` to something more "detection-y".

Like we discussed online, another angle to consider is whether the `ClassificationResults` class exposes the methods one would want. Currently, `ClassificationResults` and `DetectionResults` expose the same methods, but `COCODetectionResults` and `OpenImagesDetectionResults` additionally expose `plot_pr_curves()` and `mAP()`. I we wanted those methods for this PR's work, it might suggest that we should go the "detection" route for all of this.

ehofesmann:
> @ehofesmann following up on our online discussion earlier today, I'm leaning towards exposing this feature under `evaluate_classifications()`, unless we rename `VideoClassification` to something else.
> 
> ...
> 
> ActivityNet uses "Temporal Action Localization" to describe this task. Doesn't immediately bring any suitable replacements for `VideoClassification` to mind for me...

All good points. Point number 1 about "actual" video classification is enough for me to agree that we cannot keep `evaluate_video_classifications()`. I think that we should just bite the bullet and rename `VideoClassification` now before any of the related PRs are out. 

My vote for renaming `VideoClassification` would be to replace it with `TemporalDetection`. 
* Temporal detection/localization is what I have come across most often when referring to these types of labels
* It is clear that it is related to a `Detection` and that it will require a label plus some other kind of positional attribute (`support` vs `bounding_box` in this case). 
* Adding `Temporal` should make it clear that it only applies to video data
* It leaves the option open for a `SpatioTemporalDetection` in the future if indexed `Detections` aren't enough for some reason (the name is a bit long though....)

The main concern that I have with this is that users will expect it to accept a `bounding_box` since it has the word `Detection`.... unless we just let it have an optional `bounding_box`? :exploding_head:

brimoor:
> My vote for renaming `VideoClassification` would be to replace it with `TemporalDetection`.
> 
> * Temporal detection/localization is what I have come across most often when referring to these types of labels
> * It is clear that it is related to a `Detection` and that it will require a label plus some other kind of positional attribute (`support` vs `bounding_box` in this case).
> * Adding `Temporal` should make it clear that it only applies to video data
> * It leaves the option open for a `SpatioTemporalDetection` in the future if indexed `Detections` aren't enough for some reason (the name is a bit long though....)
> 
> The main concern that I have with this is that users will expect it to accept a `bounding_box` since it has the word `Detection`.... unless we just let it have an optional `bounding_box`? 🤯

Hmm, also all good points. I think now we ponder a bit on naming while you finish the implementation. Swapping out names at the end should be easy enough.

PR #1308, opened 2021-09-29
Respecting keyframes when annotating video tracks

brimoor:
This PR adds support for respecting keyframes when adding/editing video track annotations.

In particular, the following general convention is established and implemented for the CVAT backend:
- When downloading new video tracks, keyframes annotations will be marked with a `keyframe=True` attribute
- When uploading existing video track annotations, if a given label field has at least one label with `keyframe=True`, then *only* keyframes are uploaded. Otherwise, all instances of the object are uploaded

**TODO**
- When downloading edits to existing video tracks with keyframes, we need to update non-keyframe labels in-place (currently only the keyframes are updated)

### Example usage

Step 1: add new video detections

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video").select_fields().clone()
dataset.clear_frames()

view = dataset.limit(1)

view.annotate(
    "new_labels",
    label_type="detections",
    classes=["vehicle"],
    attributes={
        "type": {
            "type": "select",
            "values": ["sedan", "suv", "truck"],
            "mutable": False,
        },
        "occluded": {
            "type": "radio",
            "values": [0, 0.2, 0.5, 1],
            "default": 0,
        }
    },
    launch_editor=True,
)
print(dataset.get_annotation_info("new_labels"))

# Add tracks in CVAT...

dataset.load_annotations("new_labels", cleanup=True)
session.refresh()

print(dataset.count_values("frames.detections.detections.keyframe"))
```

Step 2: edit existing video detections

```py
view.annotate(
    "edit_labels",
    launch_editor=True,
)
print(dataset.get_annotation_info("edit_labels"))

# Edit tracks in CVAT...

dataset.load_annotations("edit_labels", cleanup=True)
session.refresh()

print(dataset.count_values("frames.detections.detections.keyframe"))
```

### DIFF
--- a/docs/source/integrations/cvat.rst
+++ b/docs/source/integrations/cvat.rst
@@ -180,6 +180,20 @@ The easiest way to get started is to use the default server
 `cvat.org <https://cvat.org>`_, which simply requires creating an account and
 then providing your authentication credentials as shown below.
 
+.. note::
+
+    CVAT is the default annotation backend used by FiftyOne. However, if you
+    have changed your default backend, you can opt-in to using CVAT on a
+    one-off basis by passing the optional `backend` parameter to
+    :meth:`annotate() <fiftyone.core.collections.SampleCollection.annotate>`:
+
+    .. code:: python
+
+        view.annotate(anno_key, backend="cvat", ...)
+
+    Refer to :ref:`these instructions <annotation-setup>` to see how to
+    permanently change your default backend.
+
 Authentication
 --------------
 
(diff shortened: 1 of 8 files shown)


ehofesmann (changes requested):
This is looking good except for one change. At the moment, the annotation on the last frame gets repeated to the end of the video if it is not `outside`. This means that all frames from the last keyframe to the end are also keyframes.

The required change is to just remove the keyframe attribute of the duplicated annotations here:
https://github.com/voxel51/fiftyone/blob/995a418b0e6c614d7d86256e63b4c464efd9ec1f/fiftyone/utils/cvat.py#L3609-L3610

brimoor:
> This is looking good except for one change. At the moment, the annotation on the last frame gets repeated to the end of the video if it is not `outside`. This means that all frames from the last keyframe to the end are also keyframes.
> 
> The required change is to just remove the keyframe attribute of the duplicated annotations here:
> 
> https://github.com/voxel51/fiftyone/blob/995a418b0e6c614d7d86256e63b4c464efd9ec1f/fiftyone/utils/cvat.py#L3609-L3610

Good catch, fixed!

brimoor:
@ehofesmann is this approvable now? 😄

PR #1311, opened 2021-10-01
Increase service listener timeout

benjaminpkane:
Increase the default timeout to allow for the worst case, e.g. windows workflows.

Obviously, even 10 seconds is far slower than what we want for user experience. In due time.


brimoor (approved):
LGTM

PR #1313, opened 2021-10-01
Adding support for read-only labels/attributes during annotation

brimoor:
Adds optional flags that allow for configuring whether labels/attributes can be edited during an annotation run:
- Passing `allow_label_edits=False` to `annotate()` declares that existing object's `label`s cannot be changed
- Including the  `read_only=True` key when uploading existing label attributes declares that the attribute's value should be uploaded to the annotation backend for informational purposes, but the attribute's value cannot be changed

As noted in the docs, annotation backends may not be able to obey these parameters in their editing interface, but these settings will always be enforced when the annotations are imported back into FiftyOne.

For example, the snippet below uploads the vehicle tracks in a video dataset along with their existing `type` attributes and requests that a new `make` attribute be populated without allowing edits of any other kind:

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.take(1)

anno_key = "cvat_read_only_attrs"

# Upload existing `type` attribute as read-only and add new `make` attribute
attributes = {
    "type": {
        "type": "text",
        "read_only": True,
    },
    "make": {
        "type": "text",
        "mutable": False,
    },
}

view.annotate(
    anno_key,
    classes=["vehicle"],
    attributes=attributes,
    allow_additions=False,
    allow_deletions=False,
    allow_label_edits=False,
    allow_spatial_edits=False,
    launch_editor=True,
)
print(dataset.get_annotation_info(anno_key))

# Populate `make` attribute in CVAT

```

### DIFF
--- a/docs/source/integrations/cvat.rst
+++ b/docs/source/integrations/cvat.rst
@@ -718,10 +718,42 @@ configure an annotation run for this as follows:
         allow_spatial_edits=False,
     )
 
+You can also include a `read_only=True` parameter when uploading existing
+label attributes to specify that the attribute's value should be uploaded to
+the annotation backend for informational purposes, but any edits to the
+attribute's value should not be imported back into FiftyOne.
+
+For example, if you have vehicles with their `make` attribute populated and you
+want to populate a new `model` attribute based on this information without
+allowing changes to the vehicle's `make`, you can configure an annotation run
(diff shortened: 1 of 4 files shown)


ehofesmann (approved):
LGTM!

PR #1281, opened 2021-09-22
Annotation improvements

brimoor:
Change log
- Renames `semantic_segmentation` label type to `segmentation`
- Renames `segmentation(s)` label types to `instance(s)`
- Added support for supplying mask targets for semantic segmentation tasks
- Added support for polyline vs polygon annotations
- Added a `mask_types="stuff"/"thing"` argument to `Segementation.to_detections()` that allows for control over whether to generate one instance per class (stuff) or one instance per connected region (thing) when converting segmentation masks to detections
- Added a `Segmentation.to_polylines()` method
- Added support for importing/exporting segmentation masks with > 256 classes as uint16 masks when working with `ImageSegmentationDirectory` format
- Fixed a bug that prevented multiple imports of the same annotation run from working as expected when a label is deleted but then later re-added
- Fixed a bug that prevented annotations for new label fields of video datasets from being imported properly
- Fixed a bug that would cause things like polygons with < 3 points to be deleted when editing existing annotations in CVAT
- API cleanup

The video annotation import problem was solved by adding an `ensure_frames()` method to `Dataset`/`DatasetView` that enables methods that rely on `Frame` instances to exist to ensure that this is indeed the case. 

The polygon with < 3 points problem was solved by requiring the annotation backend to keep track of which IDs it was able to upload for annotation, rather than relying on a generic `AnnotationBackend.build_id_map()` method that assumes that all labels in the view were able to be uploaded.

### Detection/instance annotation

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset(
    "open-images-v6",
    split="validation",
    max_samples=25,
    label_types=["segmentations", "detections"],
)

# Populate a field with detections and instances
dataset.clone_sample_field("detections", "boxes_and_instances")
dataset.clone_sample_field("segmentations", "tmp")
dataset.merge_labels("tmp", "boxes_and_instances")
print(dataset)

session = fo.launch_app(dataset)

# Edit boxes
dataset.annotate(
    "detections",
    label_field="boxes_and_instances",
    label_type="detections",
    launch_editor=True,
)

print(dataset.get_annotation_info("detections"))
print(dataset.load_annotation_view("detections"))

dataset.load_annotations("detections")
session.refresh()

# Edit instances
dataset.annotate(
    "instances",
    label_field="boxes_and_instances",
    label_type="instances",
    launch_editor=True,
)

dataset.load_annotations("instances")
session.refresh()

# Boxes are chosen by default
dataset.annotate(
[...]

PR #1315, opened 2021-10-04
Labelbox integration updates

brimoor:
Updates the Labelbox integration to work with the annotation API changes introduced in https://github.com/voxel51/fiftyone/pull/1281.

Note: I updated the `invite_users` parameter to `members` because
(a) the "invite" part is only applicable if the user is not already a member of your Labelbox project, and, in practice, organizations would probably already have members added, so no "invite" is simpler
(b) the [Labelbox docs](https://docs.labelbox.com/docs/add-manage-members) seem to use "members" rather than "users" (although I do see that their API appears to still use "user" language)


fiftyone/utils/labelbox.py
```diff
@@ -1245,7 +1129,7 @@ class LabelboxAnnotationResults(foua.AnnotationResults):
     """
 
     def __init__(
-        self, samples, config, id_map, project_id, is_video, backend=None,
+        self, samples, config, id_map, project_id, is_video, backend=None
     ):
         super().__init__(samples, config, backend=backend)
```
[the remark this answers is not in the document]
brimoor: So now is a good time to mention that I haven't actually tested any of this code. Hoping to get to that this afternoon

fiftyone/utils/labelbox.py
```diff
@@ -110,30 +105,26 @@ class LabelboxBackend(foua.AnnotationBackend):
...
+            "scalar",
+        ]
 
     @property
     def supported_attr_types(self):
-        return {"text", "select", "radio", "checkbox"}
+        return ["text", "select", "radio", "checkbox"]
 
     @property
     def default_attr_type(self):
```
ehofesmann, line 130: `default_attr_type()` -> `recommended_attr_tool()`

fiftyone/utils/labelbox.py
```diff
@@ -531,87 +462,160 @@ def upload_data(
...
+        classes_as_attrs = config._classes_as_attrs
+
         if not classes_as_attrs:
-            # @todo
+            # @todo implement this
             raise NotImplementedError(
-                "Annotating classes at the top level is not yet supported."
+                "Annotating classes at the top level is not yet supported"
             )
+
```
ehofesmann, line 489: The current version raises an error when attempting to upload existing labels. This means that even the updated existing label workflow where you create and upload a new label field with the tagged labels will raise an error. 

I'm not sure if it should be in this PR or a separate one, but we need to add a check/parameter to determine if the user has a paid labelbox account or not. This is what the `upload_annotations` arg was doing previously.

brimoor: It raises an error because uploading existing labels isn't implemented yet :)

I updated the "edit existing labels using free Labelbox" workflow to not use an existing field. Instead it recommends populating a new field which is then merged into the existing field within FiftyOne

brimoor: My issue with `upload_annotations=False` is that I see it purely as a workaround to make the "edit existing labels using free Labelbox" workflow in this work possible with a handful less keystrokes. It's not something I think we should expose in general because it says "hey, you have existing labels but we'll make it easy for you add more (likely duplicate) labels to the field".

I like forcing the user to do a manual workaround if they really want to do this workflow. We put in the work to support uploading existing labels for a "proper" editing experience, so I think the user should see the error to this effect and then either (a) upgrade to paid Labelbox, or (b) switch to CVAT backend to have a proper editing experience

ehofesmann: Ah I see the issue, I missed the `@todo retrieve label schema for existing 'ground_truth' field` in the existing labels workflow. Without that I assumed you wanted to allow automatic uploading of existing label schemas.

I agree that requiring users to manually provide the label schema and create a new field is the correct way for us to do this. 

Will the `SampleCollection.get_label_schema("ground_truth")` method be added in a separate PR?

brimoor: I'll add it, just need to find the time

PR #1316, opened 2021-10-04
Remove duplicate page stage

### LINKED ISSUE #1314: [BUG] App grid view does not contain all samples
On `develop`, when I load the quickstart dataset in the App and start scrolling, the grid view exhausts before it gets to the 200th sample:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Scroll to the end of the grid

# Now show the actual last sample, which differs from above
session.view = dataset.skip(len(dataset) - 1)
```

benjaminpkane:
Resolves #1314


brimoor (approved):
LGTM

PR #1317, opened 2021-10-04
Fixing a subtle run results cloning bug

brimoor:
Fixes a subtle bug in the implementation of cloning runs (annotations, evaluations, brain runs) when things like `dataset.clone()` or `dataset.merge_samples(..., include_info=True)` are invoked.

On `develop`, the code below fails, but now works:

```py
import fiftyone as fo

files = ['one', 'two', 'three']

classif_1 = ['a', 'a', 'c']
classif_2 = ['a', 'b', 'c']

data = [('first_classif', classif_1), ('second_classif', classif_2)]

full_ds = fo.Dataset()
for dataset_name, labels in data:
    samples = []
    for sample_name, label in zip(files, labels):
        sample = fo.Sample(filepath = sample_name)
        annotation = fo.Classification(label=label)
        sample[dataset_name] = annotation
        samples.append(sample)

    ds =  fo.Dataset()
    ds.add_samples(samples)

    full_ds.merge_samples(ds)

# previously fails because results weren't being saved on the correct dataset
results = full_ds.evaluate_classifications(
    'second_classif', gt_field='first_classif', eval_key='eval'
)
results.print_report()
```

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -4764,10 +4764,11 @@ def _clone_runs(dst_dataset, src_doc):
     # memory and then writing a new copy for the destination dataset
     #
 
-    dst_doc.annotation_runs = deepcopy(src_doc.annotation_runs)
-    for anno_key, run_doc in dst_doc.annotation_runs.items():
+    for anno_key, run_doc in src_doc.annotation_runs.items():
+        _run_doc = deepcopy(run_doc)
+        dst_doc.annotation_runs[anno_key] = _run_doc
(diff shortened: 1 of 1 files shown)


ehofesmann (approved):
LGTM, it works! 

If I'm understanding this right, cloning/merging basically resulted in linking the runs of one dataset to another, so when you try to access/add new runs to the cloned dataset, they didn't technically exist on that dataset.

brimoor:
Right, I assumed that `deepcopy()` would do what I wanted, but I was wrong. Setting a class attribute of a MongoEngine class (eg `DatasetDoc.annotations = blah`) is a big no-no.

PR #1318, opened 2021-10-04
Allow a list of task assignees for CVAT

### LINKED ISSUE #1290: [FR] Allow CVAT task assignees to be a list for video tasks
When uploading video samples to CVAT, every video is loaded into a new task. The current `task_assignee` argument of the CVAT `annotate()` method only accepts a single username that is then applied to every video task. It should be possible to provide a list of usernames that will be applied like the `job_assignees` argument to sequential video tasks. 

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video", max_samples=3)

results = dataset.annotate(
    "video",
    task_assignees=["user1", "user2", "user3"],
    launch_editor=True,
)
```

Images are all loaded into a single task so this functionality would be irrelevant.

ehofesmann:
Resolve #1290 

When annotating videos in CVAT, each video is uploaded to a separate task (unlike images that are loaded into one task). The annotation parameter `task_assignee` allows you to specify the username to assign to the generated task. 

This PR allows `task_assignee` to accept a list of usernames that will be applied, round robin-style, to the created video tasks.

If a single string username is provided to `task_assignee`, it will be applied to all created tasks like before.

### Example

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video", max_samples=3).clone()

anno_key = "task_assignees"
results = dataset.annotate(
    anno_key,
    task_assignee=["user1", "user2"],
    launch_editor=True,
)

```

### DIFF
--- a/docs/source/integrations/cvat.rst
+++ b/docs/source/integrations/cvat.rst
@@ -443,7 +443,9 @@ provided:
     in reduced video quality in CVAT due to size limitations on ZIP files that
     can be uploaded to CVAT
 -   **chunk_size** (*None*): the number of frames to upload per ZIP chunk
--   **task_assignee** (*None*): a username to assign the generated tasks
+-   **task_assignee** (*None*): the username to assign the generated tasks.
+    This argument can be a list of usernames when annotating videos as each
+    video is uploaded to a separate task 
 -   **job_assignees** (*None*): a list of usernames to assign jobs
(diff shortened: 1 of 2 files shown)


brimoor (approved):
LGTM

Issue #1320, opened 2021-10-05
[BUG] Not loading class-specific attributes after annotation


ehofesmann:
The following example from PR #1305 is not working when attempting to load modifications to class-specific attributes. For example, `colorful`, `type`, and `age` are not being updated when downloading annotations, but `iscrowd` and `area` are updated.

```python
import random
import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone import ViewField as F

dataset = foz.load_zoo_dataset("quickstart")
dataset = dataset[-1:].select_fields("ground_truth").clone()

for sample in dataset:
    for detection in sample.ground_truth.detections:
        if detection.label == "car":
            detection.type = random.choice(["sedan", "suv", "truck"])
        elif detection.label == "kite":
            detection.colorful = random.choice([False, True])
        elif detection.label == "person":
            detection.age = random.randint(1, 51)

    sample.save()

dataset.annotate(
    "test",
    label_field="ground_truth",
    classes=[
        {"classes": ["car"], "attributes": True},
        {"classes": ["kite"], "attributes": True},
        {"classes": ["person"], "attributes": True},
    ],
    attributes=["area", "iscrowd"],
    launch_editor=True,
)
print(dataset.get_annotation_info("test"))

# Create annotations in CVAT...

dataset.load_annotations("test", cleanup=True)
sesion = fo.launch_app(dataset)
```

PR #1323, opened 2021-10-05
Fixing serialization of CVAT attribute values

brimoor:
Resolves another issue in the example code from https://github.com/voxel51/fiftyone/issues/1320.

It turns out that CVAT requires booleans to be serialized as exactly `"true"` and `"false"` in order for them to be properly linked with the checkbox tool type.


ehofesmann (approved):
LGTM, good catch on the checkbox issue.

PR #1331, opened 2021-10-12
Catch page and frame stream errors

benjaminpkane:
### Page errors

When an error is encountered during pagination, e.g. from the server, the App will now catch the error and show the error page.

```py
import fiftyone as fo

sample = dataset.first()
sample["inf"] = float("inf")
sample.save()
session.refresh()
```

### Frame stream errors

Similar to paging through samples, when an error is encountered when streaming frames, Looker will now show the error view.

```py
import fiftyone as fo

dataset, session = fo.quickstart(video=True)
sample = dataset.first()
sample[60]["inf"] = float("inf")
sample.save()
session.refresh() # the video will now error
```

### DIFF
--- a/app/packages/app/src/components/Actions/Options.tsx
+++ b/app/packages/app/src/components/Actions/Options.tsx
@@ -1,4 +1,4 @@
-import { Autorenew } from "@material-ui/icons";
+import { Autorenew, OpacityRounded } from "@material-ui/icons";
 import React from "react";
 import { constSelector, useRecoilState, useRecoilValue } from "recoil";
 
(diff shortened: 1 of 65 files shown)


benjaminpkane:
> One observation is that on develop right now, the video plays until the 60th frame and then hangs, while now it immediately errors. Was that intentional? If it is an easy change, I think it would be a helpful context clue for debugging reasons if the video plays until the actual bad data is about to be rendered (in the case of bad frame data).

The error is shown as soon as it is observed, i.e. when a frame chunk response 500s. Maybe it's overkill to go to a full error state in this state. But an error is an error.

PR #1332, opened 2021-10-12
Release v0.14.0

### LINKED ISSUE #1346: [BUG] App grid does not populate when showing only tagged labels using extended view
I loaded a video dataset in the App and then tagged one label on the 25th frame of a video:

When I then use the filter sidebar to show only the selected labels, I see "1 of 10 samples" as expected but the actual grid is empty.

If I then click the bookmark icon to save the filter as a view stage, the sample *does* appear as expected. So, this seems to be an extended view issue.

```py
import fiftyone as fo
import fiftyone.zoo as foz

```

Screen Shot 2021-10-14 at 10 47 21 PM

benjaminpkane:
### Issues

* Resolves #1346

### DIFF
--- a/fiftyone/core/clips.py
+++ b/fiftyone/core/clips.py
@@ -327,6 +327,7 @@ def make_clips_dataset(
                 :class:`fiftyone.core.fields.FrameSupportField`, or list of
                 :class:`fiftyone.core.fields.FrameSupportField` field
             -   a frame-level label list field of any of the following types:
+
                 -   :class:`fiftyone.core.labels.Classifications`
                 -   :class:`fiftyone.core.labels.Detections`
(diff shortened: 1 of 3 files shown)


brimoor:
@benjaminpkane I was testing the dataset from https://github.com/voxel51/fiftyone/pull/1228 in the App on this PR and it looks like the special rendering for `support` fields that you did was lost in the merge madness?

Screen Shot 2021-10-14 at 8 47 58 PM

brimoor:
@benjaminpkane not sure if you finished your tweaks, but the `support` label tags currently still look like they do in my screenshot above. Also, grid tags for (at least) ID and frame number fields are not currently rendering:

Screen Shot 2021-10-14 at 11 02 30 PM

Issue #1309, opened 2021-09-30
[BUG] Resolve `nan` and `inf` issues in the App


benjaminpkane:
`nan` and `inf` values in a dataset currently break the App. Notably, `nan` and `inf` are not valid JSON, and the values break other App components like numeric bounds values.

Breaking code:
```py
import fiftyone as fo


dataset, session = fo.quickstart()
sample = dataset.first()

sample["nan"] = float("nan")
sample["inf"] = float("inf")
sample["keypoint"] = fo.Keypoint(label="nan", points=[(float("nan"), float("nan"))])
sample.save()
session.refresh()
```

PR #1350, opened 2021-10-15
Graceful handling of nan and inf field values

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1309.

Updates the App and the `histogram_values()` aggregation to gracefully handle nan and +/- inf values in floating point fields.

Also adds an optional `safe=True` flag to aggregations that cause nan/inf to be treated as None, which can be desirable in some cases. 

```py
import fiftyone as fo

dataset = fo.Dataset()
dataset.add_samples(
    [
        fo.Sample(filepath="image1.png", float=1.0),
        fo.Sample(filepath="image2.png", float=-float("inf")),
        fo.Sample(filepath="image3.png", float=float("inf")),
        fo.Sample(filepath="image4.png", float=float("nan")),
        fo.Sample(filepath="image5.png", float=None),
        fo.Sample(filepath="image6.png"),
    ]
)

# nan/inf are always grouped into `other`
print(dataset.histogram_values("float"))  # other = 5

# By default, nan/inf will be present in floating point aggregations
print(dataset.bounds("float"))  # (nan, inf)
print(dataset.count("float"))  # 4 
print(dataset.distinct("float"))  # [nan, -inf, 1.0, inf]
print(dataset.count_values("float"))  # {None: 2, -inf: 1, 1.0: 1, nan: 1, inf: 1}
print(dataset.mean("float"))  # nan
print(dataset.sum("float"))  # nan
print(dataset.std("float"))  # nan

# But the `safe=True` option will instead treat them as missing
print(dataset.bounds("float", safe=True))  # (1.0, 1.0)
print(dataset.count("float", safe=True))  # 1
print(dataset.distinct("float", safe=True))  # [1.0]
print(dataset.count_values("float", safe=True))  # {None: 5, 1.0: 1}
print(dataset.mean("float", safe=True))  # 1.0
print(dataset.sum("float", safe=True))  # 1.0
print(dataset.std("float", safe=True))  # 0.0

# The App gracefully handles nan/inf too
```


brimoor:
@benjaminpkane can you contribute the necessary App fixes here when you get some bandwidth?

benjaminpkane:
The current behavior of bounds would still break range sliders. Can we add a setting to `Bounds()` that filters out invalid numbers.

brimoor:
yeah I can add an optional `safe=True` flag that treats nan/inf as missing. I think the default should be `safe=False` though because API users should get the bad data if it exists

brimoor:
@benjaminpkane when testing this branch, I'm seeing a couple bugs:
- When I switch a numeric field filter to `exclude` mode, the label tags stop appearing in the grid (the filtering works, they just stop rendering)
- I am able to open the example code from the PR description in the App. However, if I then try to load any other dataset, the App crashes (even after restarting a fresh Python session) with the following error:

```
vendor.0d490864.js:27 TypeError: (void 0) is not a function
    at index.9eb2adc0.js:403
    at a (vendor.0d490864.js:27)
    at y (vendor.0d490864.js:27)
    at b (vendor.0d490864.js:27)
    at w (vendor.0d490864.js:27)
    at vendor.0d490864.js:27
    at vendor.0d490864.js:27
    at Object.U [as get] (vendor.0d490864.js:27)
    at getNodeLoadable (vendor.0d490864.js:27)
    at h (vendor.0d490864.js:27)
```

benjaminpkane:
>I am able to open the example code from the PR description in the App. However, if I then try to load any other dataset, the App crashes (even after restarting a fresh Python session) with the following error:

Hmm I am unable to reproduce this

brimoor:
> > I am able to open the example code from the PR description in the App. However, if I then try to load any other dataset, the App crashes (even after restarting a fresh Python session) with the following error:
> 
> Hmm I am unable to reproduce this

What additional information can I provide? Here's what I'm seeing and having a hard time making sense of:
- Running the code in the PR description works
- Running the code in [this comment](https://github.com/voxel51/fiftyone/pull/1350#issuecomment-950011018) raises the error. But I can successfully view `session.view = dataset.select_fields()`
- Trying to load all other existing datasets I have raises the error

benjaminpkane:
Ok, things should work now

benjaminpkane (approved):

brimoor:
> Ok, things should work now

Works, thanks!

PR #1353, opened 2021-10-17
Adding a Regression label type

brimoor:
Adds a `Regression` label type that can store regression values and their associated confidences and other custom attributes.

An `evaluate_regressions()` method is also added that provides support for computing common regression evaluation metrics as well as generating scatterplots that can be used to analyze predicted regressions w.r.t. ground values, optionally stratified by an additional categorical or numeric variable.

### Example usage

```py
import random
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

# Populate some fake regression + weather data
for idx, sample in enumerate(dataset, 1):
    ytrue = random.random() * idx
    ypred = ytrue + np.random.randn() * np.sqrt(ytrue)
    confidence = random.random()
    sample["ground_truth"] = fo.Regression(value=ytrue)
    sample["predictions"] = fo.Regression(value=ypred, confidence=confidence)
    sample["weather"] = random.choice(["sunny", "cloudy", "rainy"])
    sample.save()

print(dataset)

results = dataset.evaluate_regressions("predictions", gt_field="ground_truth", eval_key="eval")

# Print some standard regression evaluation metrics
results.print_metrics()

# Plot a scatterplot of the results colored by `weather` and scaled by `confidence`
plot = results.plot_results(labels="weather", sizes="predictions.confidence")
plot.show()

# Show the samples with the smallest regression error
session.view = dataset.sort_by("eval")

# Show the samples with the largest regression error
session.view = dataset.sort_by("eval", reverse=True)
```

```
mean squared error        59.69
root mean squared error   7.73
mean absolute error       5.48
median absolute error     3.57
r2 score                  0.97
explained variance score  0.97
max error                 31.77
support                   200
```

regression_evaluation_plot

Regression scatterplots can also be attached to the App to enable interactive exploration of regression results:

regression-evaluation


brimoor:
@benjaminpkane thanks for the regression App support!

PR #1357, opened 2021-10-18
Include row and column for missing/other in confusion matrices

brimoor:
This PR extends the work from https://github.com/voxel51/fiftyone/pull/1296 by including both a row (GT) and column (predictions) for "missing" and "other" when generating confusion matrices for a subset of the available classes of an evaluation run.

Terminology:
- `classes`: a subset of the available classes to study
- "other": examples that appear in `classes` but are matched to examples whose label does not appear in `classes`
- "missing": unmatched examples (eg false positives or false negatives)

As of this PR, the default behavior is to include rows/columns for "missing" and "other" in the plotted confusion matrix only if those rows/columns contain non-zero data. Note that, when studying a `classes` subset, the `(other/missing, other/missing)` entries of the confusion matrix are artificially set to zero, since they describe data that lies completely outside of the `classes` subset of interest.

An optional `log_colorscale=True` flag is also added that applies the colorscale on a log scale, which is useful to better visualize variations in smaller values when large values are also present.

Thanks to @yoeldk for the comment in https://github.com/voxel51/fiftyone/pull/1296#issuecomment-934134687 to inspire this reasonable change 😄 

### Detections example

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "predictions", eval_key="eval", classwise=False
)

results.print_report()
results.print_report(classes=["person", "cat", "dog"])

plot = results.plot_confusion_matrix()
plot.show()

# Show specific classes, including `missing` and `other` only if relevant
# Also use a log colorscale to better visualize variations near zero
plot = results.plot_confusion_matrix(
    classes=["person", "cat", "dog"], log_colorscale=True
)
plot.show()

# Don't show missing and other
plot = results.plot_confusion_matrix(
    classes=["person", "cat", "dog"], include_missing=False, include_other=False
)
plot.show()
```

### Classifications example

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

    "cifar10", split="test", max_samples=1000, shuffle=True
)

def jitter(val):
    if random.random() < 0.10:
        return random.choice(classes)

    return val

predictions = [
]

    "predictions", gt_field="ground_truth", eval_key="eval_simple"
)

results.print_report()
results.print_report(classes=["airplane", "cat", "dog"])

plot = results.plot_confusion_matrix()
plot.show()

# Show specific classes, including `other` only if relevant
# Also use a log colorscale to better visualize variations near zero
plot = results.plot_confusion_matrix(
    classes=["airplane", "cat", "dog"], log_colorscale=True
)
plot.show()

# Don't show other
plot = results.plot_confusion_matrix(
    classes=["airplane", "cat", "dog"], include_other=False
)
plot.show()
```


ehofesmann (approved):
LGTM

PR #1369, opened 2021-10-22
Interactive sidebar

### LINKED ISSUE #1232: [FR] Opening the expanded view during grid selection
Per user feedback, it would be useful if one could still open the expanded sample view even during grid selection. `Ctrl + click` could be used to override the default click selection when in selection mode in  the grid.

### LINKED ISSUE #1337: [FR] App gets stuck when unplayable video is encountered
When working with the HMDB51 dataset as follows:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("hmdb51", split="test")
fouv.reencode_videos(dataset)

view = dataset.match(F("ground_truth.label") == "catch")

```

There are some `catch` videos for which the App seems unable to play them for whatever reason, even after re-encoding.

The grid view functions fine until I hover over an unsupported video, at which point all subsequent hover operations don't work (infinite loading state):

Screen Shot 2021-10-13 at 10 02 36 PM

Similarity, in the expanded modal, when I am arrowing between videos, when I reach an unplayable video, the arrow keys stop working:

Screen Shot 2021-10-13 at 10 02 48 PM

Both of these indicate some kind of uncaught exception that is clogging things up and needs to be handled more gracefully.

### LINKED ISSUE #1347: [BUG] Pressing ESC when viewing a single frame clip in the App enters a broken state
I created a clips view with a clip containing exactly 1 frame by tagging a single bounding box in a video dataset, using the filter sidebar to show only tagged labels, and then using the clips menu to convert to clips on the field containing the tagged label.

```py
import fiftyone as fo
import fiftyone.zoo as foz

```

I am able to view the clip as expected in the expanded sample modal (shown below). However, if I press ESC immediately after opening the expanded modal, the App escapes into a broken state where the labels disappear. The expected behavior is that the expanded modal closes.

Screen Shot 2021-10-14 at 10 47 37 PM

### LINKED ISSUE #1419: [FR] Show error stack traces in the App
Currently if an error occurs while loading an App page, one of two things happens:
1. The generic error page pictured below is loaded
2. A non-fatal server error dialog appears in the bottom-right of the App, and the backend stack trace is printed to the user's shell

Errors can occur for a number of reasons
- **User error**: they tried to load an invalid view in the App
- **User error**: they are trying to view a video dataset in the App w/o ffmpeg installed
- **Internal error**: something that's our fault

In user error cases, the user needs to see the stack trace from the backend server to understand what went wrong. Option 2 achieves that, but the stack trace is not visible when working from notebooks. Option 1 doesn't show the trace at all, which is bad.

One solution here is to simplify to *always* showing the fatal error App page but also including the relevant JS or backend stack trace. The backend stack trace is generally necessary because the user needs to fix things. The JS stack trace might be helpful for github issue creation purposes, but in general these types of errors are not the user's fault, so we could instead not show the full error and instead either provide instructions to access the error from the browser's console or, perhaps, collapse the error by default and let the user opt-in to seeing it, if desired.

Screen Shot 2021-11-12 at 2 35 48 PM

### LINKED ISSUE #1503: [BUG] App cannot render empty heatmaps
On `fiftyone==0.14.2`, the App fails to render samples with a `Heatmap` or `Segmentation` field whose array is empty:

```py
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

sample = dataset.first()
sample["heatmap"] = fo.Heatmap(map=np.zeros((0, 0)))
sample["segmentation"] = fo.Segmentation(mask=np.zeros((0, 0)))
sample.save()

```

### LINKED ISSUE #1509: [BUG] The epoch incorrectly renders as 0 in the App
The epoch incorrectly renders as `0` rather than `1970-1-1` in the App grid:

```py
from datetime import date, datetime
import fiftyone as fo

sample = fo.Sample(
    date=date(1970, 1, 1),
    datetime=datetime(1970, 1, 1, 0, 0),
)

dataset = fo.Dataset()
dataset.add_sample(sample)

```

Screen Shot 2022-01-04 at 3 10 23 PM

Screen Shot 2022-01-04 at 3 10 31 PM

### LINKED ISSUE #1521: [BUG] Gracefully handle attempts to load a deleted dataset in the App
Currently, if the dataset on a `Session` is deleted, an error is raised by the App when loading the page and the user cannot escape the error page without modifying the session via python.

benjaminpkane:
## Issues

* Resolves #1521
* Resolves #1509
* Resolves #1503
* Resolves #1419
* Resolves #1347
* Resolves #1337
* Resolves #1232


benjaminpkane:
I would say there is another day of details to get through, but interacting with the new sidebar is functional (dragging entries). The goal here is that directions are not needed, so no directions are provided :slightly_smiling_face: 

```
bash install.bash

fiftyone quickstart
```

Still connecting the updates to Looker, loading states, and filter queries.
sidebar

benjaminpkane:
Ran into a dependency issue. Connecting looker and the modal now.

benjaminpkane:
A couple more details came up, notably handling schema changes due to patches and clips views.

benjaminpkane:
Undrafting

fiftyone/core/collections.py
```diff
@@ -440,6 +442,76 @@ def _get_default_frame_fields(
             include_private=include_private, use_db_fields=use_db_fields
         )
 
+    def get_field(self, path):
```
benjaminpkane: `SampleCollection.get_field()` is the new method for accessing any `Field` instance on a schema via its path.

```py
import fiftyone.zoo as foz

quickstart = foz.load_zoo_dataset("quickstart")
print(quickstart.get_field("ground_truth.detections.confidence"))

video_quickstart = foz.load_zoo_dataset("quickstart-video")
print("frames.detections.detections", video_quickstart.get_field("frames.detections.detections")
```

One point I did not address is dynamic `document_type` updates. Perhaps this should be addressed at some point. The obvious example relates to metadata:
```py
quickstart.compute_metadata(num_workers=1, skip_failures=False) # multiprocessing isn't working...

metadata_field = quickstart.get_field("metadata")
print(metadata_field) # fiftyone.core.Metadata, not fiftyone.core.metadata.ImageMetadata
```

fiftyone/core/collections.py
```diff
@@ -440,6 +442,76 @@ def _get_default_frame_fields(
...
+            if idx != last and isinstance(field, (fof.ListField)):
+                field = field.field
+
+            if isinstance(field, fof.EmbeddedDocumentField):
+                schema = field.get_field_schema()
+                _add_mapped_fields_as_private_fields(schema)
+
+        return field
+
+    def _resolve_field(self, path):
```
benjaminpkane: `SampleCollection._resolve_field()` will translate regular paths to database paths, e.g. `detection.id` to `detection._id`.

fiftyone/core/collections.py
```diff
@@ -7779,117 +7825,15 @@ def _parse_field_name(
...
 
     return (
-        field_name,
+        ".".join(resolved_keys),
         is_frame_field,
         unwind_list_fields,
         other_list_fields,
-        id_to_str,
+        isinstance(field, fof.ObjectIdField)
+        and not field_name.split(".")[-1].startswith("_"),
```
benjaminpkane: `id_to_str` logic

fiftyone/core/dataset.py
```diff
@@ -4511,6 +4511,19 @@ def _create_indexes(sample_collection_name, frame_collection_name):
...
+    for field_name, field in doc_cls._fields.items():
+        if isinstance(field, fof.EmbeddedDocumentField):
+            doc = foo.SampleFieldDocument.from_field(field)
+            field = doc.to_field()
+            field._set_parent(doc_cls)
+            doc_cls._fields[field_name] = field
+            setattr(doc_cls, field_name, field)
+
+    for field_doc in field_docs or []:
+        doc_cls._declare_field(field_doc.to_field())
```
benjaminpkane: When creating sample and frame backing document classes, default emedded document field instances, e.g. `Metadata`, must be explicitly recreated and populated with dynamic fields. If they are not recreated, then the default instance is polluted and future datasets would load with dynamic fields on other datasets.

fiftyone/core/fields.py
```diff
@@ -177,14 +178,27 @@ def __str__(self):
 
         return etau.get_class_name(self)
 
+    def validate(self, value):
+        if isinstance(self.field, EmbeddedDocumentField):
+            for v in value:
+                self.field.validate(v, clean=False, expand=True)
```
benjaminpkane: Validate every document in a `ListField` with an `EmbeddedDocumentField` field. A pain point for performance, likely.

fiftyone/core/fields.py
```diff
@@ -600,12 +614,110 @@ class EmbeddedDocumentField(mongoengine.fields.EmbeddedDocumentField, Field):
...
+                self.error("field does not have field '%s' declared" % name)
+            elif isinstance(field, EmbeddedDocumentField):
+                field.validate(field_value, clean=False, expand=expand)
+            else:
+                field.validate(field_value)
+
+        if isinstance(value, foo.DynamicEmbeddedDocument):
+            value._set_parent(self)
+
+        super().validate(value, clean)
```
benjaminpkane: Core validation traversal on `EmbeddedDocumentField`

fiftyone/core/fields.py
```diff
@@ -600,12 +614,110 @@ class EmbeddedDocumentField(mongoengine.fields.EmbeddedDocumentField, Field):
...
+
+    def get_field_schema(
```
benjaminpkane: `EmbeddedDocumentField.get_field_schema()` allows for schema traversal.

```py
import fiftyone.zoo as foz

quickstart = foz.load_zoo_dataset("quickstart")

gt_detection_schema = quickstart.get_field_schema()["ground_truth"].get_field_schema()["detections"].field.get_field_schema()
```

fiftyone/core/fields.py
```diff
@@ -600,12 +614,110 @@ class EmbeddedDocumentField(mongoengine.fields.EmbeddedDocumentField, Field):
...
+        return filtered_fields
+
+    def to_python(self, value):
+        doc = super().to_python(value)
+
+        if isinstance(doc, foo.DynamicEmbeddedDocument):
+            doc._set_parent(self)
+        return doc
+
+    def _save_field(self, field, keys):
```
benjaminpkane: The `_parent` of an `EmbeddedDocumentField` is either a sample or frame backing document **class** or an `EmbeddedDocumentField` **instance**.

frame and sample document backing classes have an equivalent `_save_field` method, but it is a `@classmethod`.

These `_save_field()` methods propagate schema changes back up to the dataset document for saving.

fiftyone/core/odm/dataset.py
```diff
@@ -22,12 +22,19 @@
     TargetsField,
 )
 
-from .document import Document, EmbeddedDocument, BaseEmbeddedDocument
+from .document import Document
+from .embedded_document import EmbeddedDocument, BaseEmbeddedDocument
 from .runs import RunDocument
 
 
 def create_field(
```
benjaminpkane: `create_field()` is now a recursive method

benjaminpkane: A notable limitation of current schemas is a `ListField(ListField(OtherField))` cannot be described by the data model, i.e. a `SampleFieldDocument`. This was already the case, of course.

fiftyone/core/odm/embedded_document.py
```diff
@@ -0,0 +1,202 @@
...
+
+food = fou.lazy_import("fiftyone.core.odm.dataset")
+
+
+class BaseEmbeddedDocument(MongoEngineBaseDocument):
+    """Base class for documents that are embedded within other documents and
+    therefore are not stored in their own collection in the database.
+    """
+
+    _parent = None
```
benjaminpkane: A `BaseEmbeddedDocument` instance has an `EmbeddedDocumentField` as its `_parent`. For propagating updates.

fiftyone/core/odm/embedded_document.py
```diff
@@ -0,0 +1,202 @@
...
+            subfield=subfield,
+            **kwargs,
+        )
+        self._declare_field(field)
+
+    def _declare_field(self, field):
+        if self._parent is None:
+            self._custom_fields[field.name] = field
+        else:
+            self._parent._save_field(field, [field.name])
```
benjaminpkane: `BaseEmbeddedDocument._declare_field` propagates updates.

fiftyone/core/odm/utils.py
```diff
@@ -0,0 +1,159 @@
...
+    if isinstance(value, datetime):
+        return fof.DateTimeField
+
+    if isinstance(value, date):
+        return fof.DateField
+
+    return None
+
+
+def _merge_implied_fields(implied_fields):
```
benjaminpkane: `_merge_implied_fields()` merges a list of implied field kwargs recursively. It has not been thoroughly tested...

ehofesmann:
This is really well done, the new sidebar and the App, in general, feel great! Some initial bugs that I came across:

* In the grid view, scrolling down and then back up removes the selection overlay when hovering over samples
* The dropdown for labels in the grid and expanded views is not working for me, primitives work fine 

image

benjaminpkane:
> In the grid view, scrolling down and then back up removes the selection overlay when hovering over samples

Thanks for the notes. Can you clarify what you mean by selection overlay? I'm not sure I understand.

ehofesmann:
> Thanks for the notes. Can you clarify what you mean by selection overlay? I'm not sure I understand.

By selection overlay, I just meant the checkbox and fade to black at the top of the image:
selection_overlay

Basically, there is no way to select images after scrolling back up, I can only click on them and go into the expanded view. After exiting the expanded view, the selection boxes appear again.

brimoor:
@benjaminpkane here's a couple issues I found just now while testing the dev deployment:

1) Actions menu is not properly displaying (see below)

Screen Shot 2022-03-21 at 5 28 18 PM

2) When editing a label group name, it would be helpful to have two things:
- In editing mode, consider adding something like a save icon (same location as the pencil icon) that gives the user a visual indicator that they should click it to save. I like how currently ENTER or doing anything that gives focus to something else seems to commit the change (but see next point)
- After committing an edit, render something to indicate that the edit was accepted. Currently nothing happens and its not obvious that the change was successful. One idea is to render a checkmark icon that fades away after a few ms. Or maybe you change the look-and-feel of the div in editing mode and revert to normal when the edit is saved 

I'm currently editing the group name in the screenshot below, but its not obvious from a UX standpoint:
Screen Shot 2022-03-21 at 5 28 47 PM

3) Consider adding some more validation logic to the field group names:
- Don't allow `""`
- When merging groups (either by adding a new group name or editing an existing group name to match another), the sidebar elements temporarily renders in the wrong order until you refresh or drag the group

benjaminpkane:
Superseded by #1713. Remaining comments will be addressed there

PR #1370, opened 2021-10-23
Apache beam I/O

brimoor:
This PR adds some utilities that leverage [Apache Beam](https://beam.apache.org) to parallelize import, merge, and export operations:
- `beam_import()`: a parallelized alternative to `Dataset.add_samples()`
- `beam_merge()`: a parallelized alternative to `Dataset.merge_samples()`
- `beam_export()`: a parallelized alternative to `SampleCollection.export()`

For now these are just standalone utilities, but in the near future I think it could make sense to add an optional `num_workers` parameter to the main API methods that will optionally leverage Beam to accelerate the operations.

### Notes

The current implementation of `beam_export()` is only well-suited for export formats that are inherently sharded. For example, it is ideal for exporting a sharded TF record dataset, but not ideal for exporting in COCO format, since the current implementation lacks a way to "reduce" shards back into a single JSON file (if that's what the user wants). I'm pondering what, if anything, to add to the implementation to address this.

### Import example

```py

import fiftyone as fo
import fiftyone.utils.beam as foub

samples = range(10000)

def make_sample(idx):
    return fo.Sample(filepath="image%d.png" % idx, uuid=idx)

#
# Option 1: build the samples on the workers
#

dataset = fo.Dataset()

foub.beam_import(dataset, samples, parse_fcn=make_sample)
print(dataset)

#
# Option 2: build the samples in the main thread
# This is generally not preferred but may be necessary if your ``parse_fcn`` is not serializable
#

dataset = fo.Dataset()

samples = map(make_sample, samples)

foub.beam_import(dataset, samples)
print(dataset)
```

### Merge example

```py
import fiftyone as fo
import fiftyone.utils.beam as foub
import fiftyone.zoo as foz

samples = iter(dataset.select_fields("predictions"))

foub.beam_merge(dataset, samples, fields={"predictions": "predictions2"})

print(dataset.count("predictions2.detections"))
```

### Export example

```py
from apache_beam.options.pipeline_options import PipelineOptions

import fiftyone as fo
import fiftyone.utils.beam as foub
import fiftyone.zoo as foz

# Use multithreading instead of the default multiprocessing
options = PipelineOptions(
    runner="direct",
    direct_num_workers=10,
    direct_running_mode="multi_threading",
)

foub.beam_export(
    dataset,
    num_shards=20,
    options=options,
    dataset_type=fo.types.TFObjectDetectionDataset,
    label_field="ground_truth",
    tf_records_path="/tmp/beam/tf.records-%05d-of-00020",
)
```

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -818,11 +818,6 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         self._sample_doc_cls.add_implied_field(field_name, value)
         self._reload()
 
-    def _add_implied_sample_field_if_necessary(self, field_name, value):
-        self._add_sample_field_if_necessary(
-            field_name, **foo.get_implied_field_kwargs(value)
-        )
-
     def add_frame_field(
         self,
         field_name,
@@ -889,11 +884,6 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
 
         self._frame_doc_cls.add_implied_field(field_name, value)
         self._reload()
-
-    def _add_implied_frame_field_if_necessary(self, field_name, value):
-        self._add_frame_field_if_necessary(
-            field_name, **foo.get_implied_field_kwargs(value)
-        )
 
     def rename_sample_field(self, field_name, new_field_name):
         """Renames the sample field to the given new name.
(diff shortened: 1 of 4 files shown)


benjaminpkane (approved):
LGTM. I don't think much work needs to be done to accommodate COCO, etc. Users have other options, if the format itself is the issue.

PR #1388, opened 2021-11-01
Add model predictions section to docs

ehofesmann:
Adds a section on how to add model predictions in general. This section also includes loading YOLO and COCO formatted predictions specifically.

image


brimoor (approved):
LGTM

PR #1391, opened 2021-11-03
Track label IDs using CVAT server IDs

ehofesmann:
Labels IDs must be tracked when annotating existing labels in CVAT so that the annotations can be merged back into FiftyOne properly. Previously, the `label_id` of each label was only tracked through a text attribute. This raises issues where the `label_id` attribute can easily be erased when:
1) The class of a label is changed
2) An annotator accidentally modifies the `label_id` attribute

Internally, CVAT stores it's own unique ID for each label. However, this ID was not used in the original implementation of this integration since the CVAT server ID is [subject to change](https://github.com/openvinotoolkit/cvat/issues/893#issuecomment-578020576). 

This PR adds support for tracking the CVAT server ID in addition to the `label_id` attribute in order to add redundancy to the tracking of label IDs. 

It should be noted that it is still entirely possible that the `label_id` attribute is changed and the server ID of the label gets changed resulting in a label getting overwritten when loading annotations.

### Example

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=5).clone()

anno_key = "server_id_task"
label_schema = {
    "ground_truth": {
        "attributes": {
            "iscrowd": {
                "type": "text",
            }
        }
    }
}
results = dataset.annotate(
    anno_key,
    label_schema=label_schema,
    launch_editor=True,
)

# In CVAT, change the class of one object and save
# (this erases the `label_id`)

session.refresh()
```

Previously, this example would have loaded the modified detection as a new label, removing additional attributes like `area` that were not loaded into CVAT. This PR now uses the backup server ID mapping to access the label ID and merge the modified label properly.

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -16,7 +16,6 @@ import warnings
 import webbrowser
 
 from bson import ObjectId
-from bson.errors import InvalidId
 import jinja2
 import numpy as np
 import requests
@@ -2848,7 +2847,7 @@ class CVATAnnotationResults(foua.AnnotationResults):
             samples,
             config,
             d["id_map"],
-            d.get("server_id_map", {"tags": {}, "shapes": {}, "tracks": {}}),
+            d.get("server_id_map", {}),
             d.get("project_ids", []),
             d["task_ids"],
             job_ids,
@@ -3948,29 +3947,29 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
             num_uploaded_tags = len(anno_resp["tags"])
             num_uploaded_tracks = len(anno_resp["tracks"])
 
-        server_id_map = self._create_server_id_map(anno_resp, attr_id_map)
-
-        return server_id_map
+        return self._create_server_id_map(anno_resp, attr_id_map)
 
     def _create_server_id_map(self, anno_resp, attr_id_map):
-        # Create mapping from CVAT server ID to FiftyOne ObjectId for each
-        # label
-        server_id_map = {"tags": {}, "shapes": {}, "tracks": {}}
         label_id_map = {}
         for class_id, class_attr_map in attr_id_map.items():
             for attr_name, attr_id in class_attr_map.items():
                 if attr_name == "label_id":
                     label_id_map[class_id] = attr_id
 
+        server_id_map = {}
         for anno_type, anno_list in anno_resp.items():
-            if anno_type not in server_id_map:
+            if anno_type not in ("tags", "shapes", "tracks"):
                 continue
+
+            id_map = {}
             for anno in anno_list:
                 server_id = anno["id"]
-                class_id = anno["label_id"]
+                label_attr_id = label_id_map[anno["label_id"]]
                 for attr in anno["attributes"]:
-                    if attr["spec_id"] == label_id_map[class_id]:
-                        server_id_map[anno_type][server_id] = attr["value"]
+                    if attr["spec_id"] == label_attr_id:
+                        id_map[server_id] = attr["value"]
+
+            server_id_map[anno_type] = id_map
 
         return server_id_map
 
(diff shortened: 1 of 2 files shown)


fiftyone/utils/cvat.py
```diff
@@ -3939,6 +3947,48 @@ def _upload_annotations(
...
+        return server_id_map
+
+    def _map_cvat_server_id(self, anno_resp, attr_id_map):
+        # Create mapping from CVAT server ID to FiftyOne ObjectId for each
+        # label
+        server_id_map = {}
+        dup_ids = []
+        types = ["shapes", "tags", "tracks"]
+        attr_id_map_rev = {}
+        for class_id, class_attr_map in attr_id_map.items():
```
brimoor, line 3961: You only care about the `label_id` attribute in this method, so it would be cleaner to just extract its ID here and then use it below (no need to map back to attribute strings later)

fiftyone/utils/cvat.py
```diff
@@ -3939,6 +3947,48 @@ def _upload_annotations(
...
+
+            for anno in anno_list:
+                server_id = anno["id"]
+                class_id = anno["label_id"]
+                for attr in anno["attributes"]:
+                    spec_id = attr["spec_id"]
+                    value = attr["value"]
+                    attr_name = attr_id_map_rev[class_id][spec_id]
+                    if attr_name == "label_id":
+                        if server_id in server_id_map:
```
brimoor, line 3977: You don't need to worry about duplicate server IDs (if they change we can assume they will be new, not reused)

fiftyone/utils/cvat.py
```diff
@@ -5396,13 +5466,18 @@ def __init__(
         label_dict,
         class_map,
         attr_id_map,
+        server_id_map,
```
brimoor, line 5469: You'll need to add `server_id_map = d.get("server_id_map", {})` to `_from_dict()` to ensure backwards compatibility of this new feature

fiftyone/utils/cvat.py
```diff
@@ -3939,6 +3947,48 @@ def _upload_annotations(
             num_uploaded_tags = len(anno_resp["tags"])
             num_uploaded_tracks = len(anno_resp["tracks"])
 
+        server_id_map = self._map_cvat_server_id(anno_resp, attr_id_map)
+
+        return server_id_map
+
+    def _map_cvat_server_id(self, anno_resp, attr_id_map):
```
brimoor, line 3954: How about `_create_server_id_map()`? More consistent with the other method names

brimoor (changes requested):
I like this as an additional layer of safety against label ID changes 💯

fiftyone/utils/cvat.py
ehofesmann: I need to check if server IDs are shared between `tracks`, `shapes`, and `tags`. If not, then there could be duplicates. If they are shared, I'll remove this.

brimoor: If they aren't globally unique then you'd have to store nested server ID maps for each shape type, right?

ehofesmann: Right. Fixed. They aren't globally unique, every newly uploaded/annotated label is assigned an integer ID server-side that is the previously assigned integer ID incremented by 1 . Each group (`tags`, `tracks`, and `shapes`) store their own ID, so assuming every server starts with ID 0 or 1 for each group, there can be overlapping server IDs for a tag, track, and shape.

fiftyone/utils/cvat.py
ehofesmann: I think I resolved this. I'm not 100% sure what you meant by this, though.

brimoor (approved):
LGTM

PR #1392, opened 2021-11-03
Fix count values for colab

### LINKED ISSUE #1390: [BUG] App label filter not working in Colab
I ran through the [quickstart](https://colab.research.google.com/github/voxel51/fiftyone-examples/blob/master/examples/quickstart.ipynb) in Colab (using `fiftyone==0.14.0`) and found an App error when trying to use a `label` filter in the App.

The error also occurred when using the same label filter from the expanded modal. However, other filters such as confidence and the alternate low-choices mode of the label filter were working as expected, so I believe this is strictly related to the autocomplete-style filter UI.

I cannot reproduce this error using `fiftyone==0.14.0` outside of colab.

Screen Shot 2021-11-03 at 10 54 54 AM

benjaminpkane:
Resolves #1390

### DIFF
--- a/fiftyone/server/main.py
+++ b/fiftyone/server/main.py
@@ -377,6 +377,7 @@ class PollingHandler(tornado.web.RequestHandler):
                     message = {"state": StateHandler.state}
 
             if event in {
+                "count_values",
                 "distinct",
                 "distributions",
                 "get_video_data",


brimoor (approved):
LGTM

PR #1393, opened 2021-11-03
Add support for non-sequential updates in set_values()

brimoor:
In certain use cases, one may have a batch of updates that need to be applied to a dataset, where the "keys" for these updates are non-ID values such as the `filepath` of the images for which new information is available, or an external ID of the sample (eg the `coco_id` of samples loaded from a COCO-style dataset).

This PR adds a new `set_values(field, values, key_field=key_field, ...)` syntax where `values` can be a dict mapping the `key_field` of the samples that should be updated to the values that you want to set on the samples. 

In other words, this new syntax is an efficient implementation of the following loop:

```py
for key, value in values.items():
    sample = sample_collection.one(F(key_field) == key)
    sample.field = value
    sample.save()
```
 
Example usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "coco-2017",
    split="validation",
    classes=["cat", "dog"],
    max_samples=10,
    shuffle=True,
    include_id=True,
)

# Some COCO IDs whose samples we want to update
coco_ids = dataset.take(2).values("coco_id")

# A dict mapping COCO IDs to the values we want to set
values = {i: str(i) for i in coco_ids}

# Apply the changes, using `coco_id` as the key
dataset.set_values("coco_id_str", values, key_field="coco_id")

# Verify that the correct samples were updated
updated_view = dataset.exists("coco_id_str")
assert set(values.keys()) == set(updated_view.values("coco_id"))
```


ehofesmann (approved):
LGTM.

Something to consider is that it might be nice for `key_field` to default to either sample id/filepath (parsed to determine which one) since those are probably the most common keys to use when setting values.

brimoor:
> LGTM.
> 
> Something to consider is that it might be nice for `key_field` to default to either sample id/filepath (parsed to determine which one) since those are probably the most common keys to use when setting values.

I'd rather make the user explicitly provide `key_field` to declare their intention of what keys they are using. The "easy mode" version is just to provide a list of values at correspond to the order of the samples in the collection.

PR #1394, opened 2021-11-03
Adding a compute_max_ious() utility

brimoor:
Adds a handy `compute_max_ious()` utility that computes the maximum IoU between spatial objects in the same sample/frame in a given label field.

This utility enables the slick workflow below, which is a great way to detect and remove duplicate annotations. A recipe version of this workflow is added to the docs, as well.

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("coco-2017", split="validation", max_samples=1000)

foue.compute_max_ious(dataset, "ground_truth", classwise=True)

# Retrieve detections that overlap above a chosen threshold
dups_view = dataset.filter_labels("ground_truth", F("max_iou") > 0.75)

#
# Option 1: tag duplicates in the App
#

session = fo.launch_app(view=dups_view)

# Add 'duplicate' tags to labels in the App...

dataset.delete_labels(tags="duplicate")

#
# Option 2: send duplicates to CVAT for editing
#

anno_key = "remove_dups"
dups_view.annotate(anno_key, label_field="ground_truth", launch_editor=True)

# Remove duplicates in CVAT...

# Verify that duplicates no longer exist in FiftyOne
session = fo.launch_app(view=dups_view)
```


ehofesmann (approved):
LGTM. Love the recipe, nice and concise but hits a lot of important points.

Noting that `compute_max_ious()` [loads all labels into memory](https://github.com/voxel51/fiftyone/blob/e1a5a18632627fe7969973fe687ca310fa0a18fb/fiftyone/utils/eval/detection.py#L237) which we may want to change in the future.

PR #1402, opened 2021-11-08
Use threading for COCO and Open Images S3 downloads

### LINKED ISSUE #1259: [BUG] Large RAM usage when downloading Open Images dataset
### System information

-   **OS Platform and Distribution (windows 10):
-   **FiftyOne installed from pip :
-   **FiftyOne version : v0.13.2;
-   **Python version: 3.9:

### Commands to reproduce

# shell commands  
_________________________________________
set HTTP_PROXY=http://127.0.0.1:1081
set HTTPS_PROXY=http://127.0.0.1:1081
python .\PycharmProjects\pythonProject\main.py

#code in main.py
_______________________________________________
import fiftyone.zoo as foz
import os

#dataset = foz.load_zoo_dataset("quickstart")
dataset1 = foz.load_zoo_dataset(
              "open-images-v6",
              split="train",
              label_types=["detections"],
              classes=["Bathtub"],
              max_samples=1000,
              seed=51,
              shuffle=True,
              dataset_name="open-images-Bathtub"
          )

_________________________________________
### Describe the problem
###This is happend after the CSV was parsed and it started to download the images. Crazy increase of memory,as  shown in image below:

image

ehofesmann:
In this PR:
* Move S3 downloading functionality do a separate AWS utility
* Convert the multiprocessing when downloading COCO and Open Images data to threading to resolve #1259 

### Example

**Open Images**
```python
import fiftyone as fo
import fiftyone.zoo as foz

d1 = foz.load_zoo_dataset(
    "open-images-v6",
    max_samples=100,
    dataset_name="oitest1",
    split="validation",
    num_workers=1,
)

d2 = foz.load_zoo_dataset(
    "open-images-v6",
    max_samples=200,
    dataset_name="oitest2",
    split="validation",
    num_workers=20,
)
```

**COCO**
```python
import fiftyone as fo
import fiftyone.zoo as foz

d1 = foz.load_zoo_dataset(
    "coco-2017",
    max_samples=100,
    dataset_name="cocotest1",
    split="validation",
    num_workers=1,
)

d2 = foz.load_zoo_dataset(
    "coco-2017",
    max_samples=200,
    dataset_name="cocotest2",
    split="validation",
    num_workers=20,
)
```


fiftyone/utils/aws.py
```diff
@@ -0,0 +1,127 @@
...
+
+import fiftyone.core.utils as fou
+
+import eta.core.utils as etau
+
+
+logger = logging.getLogger(__name__)
+
+
+def download_from_s3(
```
brimoor, line 27: How about `download_public_s3_files()` instead to clarify the scope of this method?

fiftyone/utils/aws.py
```diff
@@ -0,0 +1,127 @@
...
+def _build_inputs(urls, s3_client, download_dir=None):
+    inputs = []
+    for url, filepath in urls.items():
+        bucket_name, object_path = _parse_url(url)
+        if filepath is None:
+            filepath = os.path.join(download_dir, object_path)
+        if not os.path.isfile(filepath):
+            inputs.append((bucket_name, object_path, filepath, s3_client))
+        else:
+            logger.warning("File `%s` already exists, skipping..." % filepath)
```
brimoor, line 102: Suggestion: provide an optional `overwrite=True` setting that will re-download any existing files (in case they changed, say).

Not needed by Open Images, but nice to have as a general purpose utility I think

fiftyone/utils/aws.py
```diff
@@ -0,0 +1,127 @@
...
+        config=botocore.config.Config(
+            signature_version=botocore.UNSIGNED,
+            max_pool_connections=max(10, num_workers),
+        ),
+    )
+
+    inputs = _build_inputs(urls, s3_client, download_dir=download_dir)
+    num_samples = len(inputs)
+
+    if num_samples == 0:
```
brimoor, line 82: nit: you could do `if not urls: return` earlier in the function to avoid constructing other things like `s3_client` if they won't be used

fiftyone/utils/aws.py
```diff
@@ -0,0 +1,127 @@
...
+        return 0
+
+    if num_workers == 1:
+        _single_thread_download(inputs)
+    else:
+        _multi_thread_download(inputs, num_workers)
+
+    return num_samples
```
brimoor, line 90: The way this function is written, the user will always know exactly how many files will be downloaded (also this return value isn't documented), so I'd just omit. No return value needed

brimoor (approved):
LGTM. Definitely a good change- threads will do just fine for downloads, and this avoids Windoze issues.

I left a couple small comments for your consideration before merging

fiftyone/utils/aws.py
ehofesmann: This has to be after the `inputs` are constructed because this is only relevant if all of the filepaths already exist and overwrite is `False`. The filepaths are checked in `_build_inputs()` though.

PR #1406, opened 2021-11-09
[Server] Handle `None` when guessing mime type

### LINKED ISSUE #1399: [BUG] Image paths not ending with an extension make app fail to open
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux
-   **FiftyOne version (run `fiftyone --version`)**: 0.14.0
-   **Python version**: 3.8

### Describe the problem

If images in a dataset contain path which do not end with an extension e.g. `/path/to/image` instead of `/path/to/image.jpg`, when launching the fiftyone app, it does not load and shows a 500 error without any clear error message.

I can provide a piece of code to reproduce if really needed but I think my description is fairly straightforward.

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community.

benjaminpkane:
Resolves #1399

```py
import fiftyone as fo

sample = fo.Sample(filepath="/path/to/img")
dataset = fo.Dataset()
dataset.add_sample(sample)

session = fo.Session(sample)
```

In the case of images, the App will work as images are the default, but there is an expectation of properly named media files (with extensions) to distinguish between videos and images in both the Python API and App...

### DIFF
--- a/fiftyone/server/utils.py
+++ b/fiftyone/server/utils.py
@@ -127,7 +127,7 @@ def read_metadata(filepath, metadata=None):
         dict
     """
     mimetype, _ = mimetypes.guess_type(filepath)
-    if mimetype.startswith("video/"):
+    if mimetype and mimetype.startswith("video/"):
         if metadata:
             width = metadata.get("frame_width", None)
             height = metadata.get("frame_height", None)


brimoor (approved):
LGTM (also agree though that `/path/to/img` is heavily discouraged)

PR #1407, opened 2021-11-10
Fixed model oom

dangraf:

Added a function "MaxResize" in utils.torch.py and added support for the pipeline to allow an optional parameter "image_max_dim" and "image_max_size" (very similar to image_min_dim and image_min_size". 
This parameter was also added for the model zoo for models with variable image size and was set to 2048 for largest dim of the image.

The MaxResize function is tested in "tests/misc/torch_test.py" but no test is added to verify the full pipeline.
Maybe this test should be located in the "unit-test".

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -502,6 +502,7 @@ class MinResize(object):
         size = (int(round(alpha * h)), int(round(alpha * w)))
         return F.resize(pil_image_or_tensor, size, **self._kwargs)
 
+    
 class MaxResize(object):
     """Transform that resizes the PIL image or torch Tensor, if necessary, so
     that its maximum dimensions do not cause memory overflow.
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -114,6 +114,12 @@ class TorchImageModelConfig(foc.Config):
         image_min_dim (None): resize input images during preprocessing, if
             necessary, so that the smaller image dimension is at least this
             value
+        image_max_size (None): resize the input images during preprocessing, if
+            necessary to prevent OOM, so that the image dimensions are at 
+            most this ``(width, height)``
+        image_max_dim (None): resize input images during preprocessing, if
+            necessary, so that the largest image dimension is at most this
+            value to prevent out of memory
         image_size (None): a ``(width, height)`` to which to resize the input
             images during preprocessing
         image_dim (None): resize the smaller input dimension to this value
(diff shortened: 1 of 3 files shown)


fiftyone/utils/torch.py
```diff
@@ -114,6 +114,12 @@ class TorchImageModelConfig(foc.Config):
         image_min_dim (None): resize input images during preprocessing, if
             necessary, so that the smaller image dimension is at least this
             value
+        image_max_size (None): resize the input images during preprocessing, if
+            necessary to prevent OOM, so that the image dimensions are at
```
brimoor, line 118: nit: I'd recommend removing the "to prevent OOM" commentary from the docstrings in this PR. It's one reason why this parameter might be used, but there could be other reasons

fiftyone/utils/torch.py
```diff
@@ -386,6 +399,11 @@ def _build_transforms(self, config):
         elif config.image_dim:
             transforms.append(torchvision.transforms.Resize(config.image_dim))
 
+        if config.image_max_size:
+            transforms.append(MaxResize(config.image_max_size))
+        elif config.image_max_dim:
+            transforms.append(MinResize(config.image_max_dim))
```
brimoor, line 405: `MinResize` -> `MaxResize`

tests/misc/torch_tests.py
```diff
@@ -12,9 +12,47 @@
 import torch
 import torchvision
 
+import numpy as np
+from PIL import Image
+
 import fiftyone as fo
 import fiftyone.utils.torch as fout
 
+def _get_fake_img(h,w):
```
brimoor, line 21: Thanks for the tests!

fiftyone/utils/torch.py
```diff
@@ -485,6 +503,43 @@ def __call__(self, pil_image_or_tensor):
         size = (int(round(alpha * h)), int(round(alpha * w)))
         return F.resize(pil_image_or_tensor, size, **self._kwargs)
 
+class MaxResize(object):
```
brimoor, line 506: nit: our style uses two newlines between top-level functions/classes. Can you add one more here please?

brimoor (approved):
LGTM. Thanks for doing this! I left a couple small comments for you; if you could address before merging I'd appreciate it 💯

fiftyone/utils/torch.py
dangraf: I've just fixed the issues you addressed. Feels good to give back something since I really like this library. Good work!

PR #1414, opened 2021-11-12
Fixes #1413

### LINKED ISSUE #1413: [BUG] Unable to clone view whose parent dataset has a brain run
On `fiftyone==0.14.0`, Trying to clone a view into a dataset with a brain run:

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

fob.compute_similarity(dataset, brain_key="img_sim")

dataset2 = dataset.limit(1).clone()
```

raises the following error:

```
ValueError: Number of sample IDs (1) does not match number of embeddings (200). You may have missing data/labels that you need to omit from your view
```

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1413.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -4715,8 +4715,8 @@ def _clone_dataset_or_view(dataset_or_view, name):
 
     clone_dataset = load_dataset(name)
 
-    # Clone run results
-    if (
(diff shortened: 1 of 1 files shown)


benjaminpkane (approved):
LGTM

PR #1418, opened 2021-11-12
[WIP] Raise error when ffmpeg is not installed when viewing video datasets in the App

### LINKED ISSUE #1255: [FR] Raise error when loading video datasets in the App without ffmpeg installed
[Per the docs](https://voxel51.com/docs/fiftyone/getting_started/troubleshooting.html#videos-do-not-load-in-the-app), installing `ffmpeg` is required in order to view video datasets in the App.

However, currently, if the App does not find `ffprobe`/`ffmpeg` it will silently assume a square video and happily render potentially misaligned labels on the videos in the App.

We should update things so that this cannot happen. Automatically installing `ffmpeg` in all environments is likely out of the question (would be nice, as then things would "just work"), so probably the best option is to raise an informative error when `ffmpeg` is missing.

brimoor:
Resolves #1255, but blocked until https://github.com/voxel51/fiftyone/issues/1419 is resolved.

```
brew uninstall ffmpeg
```

```py
import fiftyone as fo
import fiftyone.zoo as foz

```

### DIFF
--- a/fiftyone/server/utils.py
+++ b/fiftyone/server/utils.py
@@ -13,6 +13,7 @@ import io
 import struct
 import warnings
 
+import eta.core.utils as etau
 import eta.core.video as etav
 import PIL.Image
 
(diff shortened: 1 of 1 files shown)


benjaminpkane:
Stale

PR #1424, opened 2021-11-15
Server metadata optmizations

benjaminpkane:
Updates metadata computation in the server to use coroutines so each sample can be gathered in parallel.


brimoor (approved):
LGTM

PR #1429, opened 2021-11-17
Raising error when annotating video label field not starting with "frames."

ehofesmann:
This PR raises an error when requesting spatial annotations and the `label_field` doesn't start with `frames.`, as expected.

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video", max_samples=1).clone()

results = dataset.annotate(
    "frames_test",
    label_field="new_field",
    label_type="instances",
    classes=["test"],
)
```

```
ValueError: Invalid label field 'new_field'. Spatial labels of type 'instances' being annotated on a video must be stored in a frame-level field, i.e., one that starts with 'frames.'
```

### DIFF
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -400,6 +400,7 @@ def _build_label_schema(
                 )
             )
 
+        # Converting to return type normalizes for single vs multiple labels
         _return_type = _RETURN_TYPES_MAP[_label_type]
         _is_trackable = _is_frame_field and _return_type in _TRACKABLE_TYPES
 
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -376,6 +376,8 @@ def _build_label_schema(
 
     _label_schema = {}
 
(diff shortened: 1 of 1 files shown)


brimoor (changes requested):
Isn't https://github.com/voxel51/fiftyone/issues/1430 still an issue?

I'm onboard with logging a warning rather than raising an error at task creation time, but, in that case, shouldn't we still update the download logic so that the frame-level labels are treated as new, unexpected labels that can be downloaded if desired? The fact that an unhandled exception occurs during download is not ideal.

ehofesmann:
> 

You're right, that is still an issue. Updating the download logic is a bit tricky given how unexpected fields are currently defined. Currently, unexpected annotations are just any annotations not matching the given `label_type`, this is not the issue here so some notion of unexpected frame annotation would need to be added. For now, I will just let this raise an error until there is a clear reason that we would want to allow both sample- and frame-level spatial annotations in one run.

brimoor (approved):
LGTM

PR #1433, opened 2021-11-18
Retain original filenames when annotating in CVAT

### LINKED ISSUE #1432: Is it possible to pass the original filenames to CVAT?
When uploading the images from fiftyone to CVAT, the filenames are lost. When syncing with a dataset repository, however, they are crucial as there is no other way of getting the original filenames to the annotations. I think #1235 could possibly fix this, if it was implemented. Are there currently any other workarounds? The labeling order in a given task in CVAT is not important to me.

ehofesmann:
Resolves #1432 

CVAT data is sorted alphabetically by filename after they are uploaded. This can cause issues when uploading annotations that are not in the same order. To resolve this, we modify the filenames of the media to ensure the correct ordering in CVAT. 

Previously, the filenames were simply overwritten with an increasing integer value. This PR prepends this integer value to the filenames in order to retain the original filename while ordering the media correctly.

The alternative is to re-order the annotations to match the original filenames sorted alphabetically, however, the view that is being annotated may be ordered in a specific way for a reason so we should not re-order it ourselves upon upload.

image

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -3370,7 +3370,7 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
             # IMPORTANT: CVAT organizes media within a task alphabetically by
             # filename, so we must give CVAT filenames whose alphabetical order
             # matches the order of `paths`
-            filename = "%06d%s" % (idx, os.path.splitext(path)[1])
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM, clever solution

ehofesmann:
@brimoor Can you merge this? The tests are failing for unrelated reasons.

PR #1436, opened 2021-11-18
Better handling of relative and user paths

brimoor:
Requires https://github.com/voxel51/eta/pull/532.

Allows for `~` to be included in config settings such as `database_dir`.

```shell
export FIFTYONE_DATABASE_DIR=~/fiftyone/database

# ~ is now expanded, so downstream users of this setting don't need to worry about it
fiftyone config
```


benjaminpkane (approved):
LGTM

PR #1437, opened 2021-11-19
Optimizing image metadata calculation

brimoor:
Previously `ImageMetadata.build_for()`---which was used by `SampleCollection.compute_metadata()` to populate metadata for image samples---required loading the entire image into memory.

This PR optimizes the implementation to use `PIL.Image.open()`, which only reads the bare minimum content.

Also, all `Metadata.build_for()` methods can now accept publicly GET-able URLs.

```py
import fiftyone as fo
import fiftyone.zoo as foz

# This was 50% faster when applied to BDD100k validation split (10k high resolution images) on my machine
dataset.compute_metadata(overwrite=True)
```

### DIFF
--- a/fiftyone/core/metadata.py
+++ b/fiftyone/core/metadata.py
@@ -9,8 +9,10 @@ import itertools
 import logging
 import multiprocessing
 import os
+import requests
 
-import eta.core.image as etai
+from PIL import Image
+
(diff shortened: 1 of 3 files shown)


benjaminpkane (approved):
LGTM :racehorse:

PR #1438, opened 2021-11-19
Optimize worker label processing

### LINKED ISSUE #1400: [FR] Optimize segmentation mask rendering
`fiftyone==0.14.0` changed the way that array data such as `Segmentation` and `Heatmap` are rendered. While this enabled the very helpful ability to adjust the transparency/colorscale of these data types, it significantly increased render times.

For example, below we visualize masks whose width/height are > 1024 pixels, which is currently too slow to make using the App desirable:  

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "open-images-v6",
    split="validation",
    max_samples=100,
)

# Defines pixels values for each class in the segmentation masks
mask_targets = {i: label for i, label in enumerate(dataset.default_classes, 1)}
dataset.default_mask_targets = mask_targets

    dataset, "segmentations", "masks", mask_targets=mask_targets
)

```

benjaminpkane:
Resolves #1400 

The change in for loops brought speeds back to normal. 8-10x faster than `develop`. My mistake.

### DIFF
--- a/app/packages/looker/src/index.ts
+++ b/app/packages/looker/src/index.ts
@@ -105,7 +105,23 @@ const workerCallbacks = {
   ],
 };
 
-const labelsWorker = createWorker(workerCallbacks);
+const getLabelsWorker = (() => {
+  // one labels worker seems to be best
+  // const numWorkers = navigator.hardwareConcurrency || 4;
+  const numWorkers = 1;
+
+  const workers = [];
+  for (let i = 0; i < numWorkers; i++) {
+    workers.push(createWorker(workerCallbacks));
+  }
+
+  let next = -1;
+  return () => {
+    next++;
+    next %= numWorkers;
+    return workers[next];
+  };
+})();
 
 export abstract class Looker<
   State extends BaseState = BaseState,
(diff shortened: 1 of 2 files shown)


brimoor (approved):
LGTM! Confirmed that this gets us back to fiftyone==0.13.3 levels 🥳 

One of those increasingly rare times when for, not for-each is required.

Out of curiosity, what is going on when when the falling pixels loading state is being shown? Once the grid appears, the images tile in as the worker draws them quite quicktly, but the most expensive part of https://github.com/voxel51/fiftyone/issues/1400 is pre-grid-appearance. If that could be optimized then we'd **really** be in business!

benjaminpkane:
> Out of curiosity, what is going on when when the falling pixels loading state is being shown? Once the grid appears, the images tile in as the worker draws them quite quicktly, but the most expensive part of #1400 is pre-grid-appearance. If that could be optimized then we'd really be in business!

It means the server is waiting for the first page of data. Perhaps there is an opportunity to optimize, but nothing obvious. I'll have to think about it.

PR #1439, opened 2021-11-19
[HOLD] Connect CVAT tasks to git repositories

### LINKED ISSUE #1427: [FR] Support CVAT Dataset Repositories
### Proposal Summary

Support for dataset repositories in CVAT during task creation and import.

### Motivation

Ideally, the repository information can be sent to CVAT when creating a task in fiftyone. When importing the labeled data back to fiftyone, the labels are automatically pushed to the specified repository using the endpoint `<cvat_host>/git/repository/push/<task_id>`

It allows for easy backups of the labeled data, along with information about the labeler, job reviewer etc.

    organization?
We want to use the CVAT dataset repository field for synchronization and backups when a labeling task is finished. It is also important to us, to save the metadata about labeler and reviewer to assure high dataset quality.

    insufficient)
To the best of my knowledge, this feature is not supported at the moment. I saw it's possible to set the values for `job_assignees`, `job_reviewers` etc. in `CVATBackendConfig` but there is no option for a dataset repository or did I overlook it?

-   [ ] App: FiftyOne application

### Willingness to contribute

of this feature?

        the FiftyOne community.

ehofesmann:
Resolves #1427 

This PR adds these arguments to the CVAT annotation backend:
* `git_repository` (None): the url of the git repository to link with the created tasks and to which to upload annotations

* `push_to_git` (True): whether to automatically push annotations to git whenever samples are uploaded or downloaded and a `git_repository` is provided
* `git_lfs` (False): whether to use the git LFS (Large File Storage) to store annotations

CVAT provides the ability to link an annotation task to a git dataset repository, allowing you to upload the labels of the task to a zip file in the repository.

The `git_repository` argument can be used to define the URL of the Git repository when annotating a FiftyOne `SampleCollection`. Alternatively, you can provide the URL and a path to the annotations within the repository formatted as URL [PATH] defined here: https://openvinotoolkit.github.io/cvat/docs/manual/basics/creating_an_annotation_task/#dataset-repository

When samples are uploaded and when annotations are downloaded, the labels in the task are automatically pushed to the `git_repository` unless the `push_to_git` parameter is set to `False`.

### Example: 

Note: The first time that you connect CVAT to a git repository, you will need to add the SSH key that gets printed to your git account.

```python
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.take(1)

anno_key = "cvat_github"
label_field = "ground_truth"
git_repository = "https://github.com/username/repo-name"
git_repository_alt = "https://github.com/username/repository-name [path/to/annotations.xml]"
push_to_git = True # Default
git_lfs = False # Default

# Create task and connect to GitHub
view.annotate(
    anno_key,
    label_field=label_field,
    git_repository=git_repository,
    push_to_git=push_to_git,
    git_lfs=git_lfs,
    launch_editor=True,
)

# Annotate in CVAT

```

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -2524,9 +2524,9 @@ class CVATBackendConfig(foua.AnnotationBackendConfig):
         occluded_attr (None): an optional attribute name containing existing
             occluded values and/or in which to store downloaded occluded values
             for all objects in the annotation run
-        git_repo (None): the url of the GitHub repository to link with the
+        git_repo (None): the url of the git repository to link with the
             created tasks and to which to upload annotations
-        push_to_git (True): whether to automatically push annotations to Git
+        push_to_git (True): whether to automatically push annotations to git
             whenever samples are uploaded or downloaded and a `git_repo` is
             provided
     """
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -2525,7 +2525,8 @@ class CVATBackendConfig(foua.AnnotationBackendConfig):
             occluded values and/or in which to store downloaded occluded values
             for all objects in the annotation run
         git_repository (None): the url of the git repository to link with the
-            created tasks and to which to upload annotations
+            created tasks and to which to upload annotations. The string can
+            also contain the path to the file within the repository `URL [PATH]`
         push_to_git (True): whether to automatically push annotations to git
             whenever samples are uploaded or downloaded and a
             `git_repository` is provided
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -2524,11 +2524,13 @@ class CVATBackendConfig(foua.AnnotationBackendConfig):
         occluded_attr (None): an optional attribute name containing existing
             occluded values and/or in which to store downloaded occluded values
             for all objects in the annotation run
-        git_repo (None): the url of the git repository to link with the
+        git_repository (None): the url of the git repository to link with the
             created tasks and to which to upload annotations
         push_to_git (True): whether to automatically push annotations to git
-            whenever samples are uploaded or downloaded and a `git_repo` is
-            provided
+            whenever samples are uploaded or downloaded and a
+            `git_repository` is provided
+        git_lfs (False): whether to use the git LFS (Large File Storage) to
+            store annotations
     """
 
     def __init__(
@@ -2550,8 +2552,9 @@ class CVATBackendConfig(foua.AnnotationBackendConfig):
         project_name=None,
         project_id=None,
         occluded_attr=None,
-        git_repo=None,
+        git_repository=None,
         push_to_git=True,
+        git_lfs=False,
         **kwargs,
     ):
         super().__init__(name, label_schema, media_field=media_field, **kwargs)
@@ -2567,8 +2570,9 @@ class CVATBackendConfig(foua.AnnotationBackendConfig):
         self.project_name = project_name
         self.project_id = project_id
         self.occluded_attr = occluded_attr
-        self.git_repo = git_repo
+        self.git_repository = git_repository
         self.push_to_git = push_to_git
+        self.git_lfs = git_lfs
 
         # store privately so these aren't serialized
         self._username = username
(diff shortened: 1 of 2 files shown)


fiftyone/utils/cvat.py
```diff
@@ -2523,6 +2524,11 @@ class CVATBackendConfig(foua.AnnotationBackendConfig):
         occluded_attr (None): an optional attribute name containing existing
             occluded values and/or in which to store downloaded occluded values
             for all objects in the annotation run
+        git_repo (None): the url of the git repository to link with the
+            created tasks and to which to upload annotations
+        push_to_git (True): whether to automatically push annotations to git
```
brimoor, line 2529: Is there a purpose for `push_to_git`? Isn't it safe to assume that if a `git_repo` is provided then they always want their labels posted?

fiftyone/utils/cvat.py
```diff
@@ -2523,6 +2524,11 @@ class CVATBackendConfig(foua.AnnotationBackendConfig):
         occluded_attr (None): an optional attribute name containing existing
             occluded values and/or in which to store downloaded occluded values
             for all objects in the annotation run
+        git_repo (None): the url of the git repository to link with the
```
brimoor, line 2527: Let's use `git_repository` here. In general I'd like to avoid using abbreviations unless the parameter name would be realllly long. `attr` is a bit of a grandfathered in exception.

fiftyone/utils/cvat.py
```diff
@@ -3312,6 +3338,52 @@ def delete_tasks(self, task_ids):
             for task_id in pb(list(task_ids)):
                 self.delete_task(task_id)
 
+    def create_git(self, task_id, git_repo):
+        json_data = {
+            "path": git_repo,
+            "lfs": False,
```
brimoor, line 3344: Since storing zips in git can quickly get out-of-hand spacewise, I imagine users of this feature will want a `git_lfs` flag exposed so they can opt-in to using LFS if desired.

brimoor (changes requested):
So how does the CVAT git feature work? I see that it stores zips of labels in the git repo by task ID, but what happens to older zips (eg the one from upload) when a new version is posted (eg after one ore more downloads)? Are the old ones just overwritten?

fiftyone/utils/cvat.py
niclaswue: Yes, the older zip files just seem to be overwritten by newer ones. It is also possible to export the raw .xml file by just specifying an xml path instead of a zip file path.

niclaswue:
Thank you for making a PR, this looks pretty good. Is it already tested? I also assumed `path` just takes the repo url, however after getting an "IndexError" all the time, I found out, you have to additionally pass along the file path of the annotation file inside the repository. See this issue: https://github.com/openvinotoolkit/cvat/issues/3190

fiftyone/utils/cvat.py
ehofesmann: It's not necessarily a given. When adding a `git_repo`, it links it to the created tasks. In the CVAT UI, this repo can be manually synced which users may want to do themselves rather than during upload/download time.

image

ehofesmann:
> Thank you for making a PR, this looks pretty good. Is it already tested? I also assumed `path` just takes the repo url, however after getting an "IndexError" all the time, I found out, you have to additionally pass along the file path of the annotation file inside the repository. See this issue: [openvinotoolkit/cvat#3190](https://github.com/openvinotoolkit/cvat/issues/3190)
> 
> So I would say, we could add an additional argument like `git_repository_filepath` that can be used to specify the exact location of the annotation data. This can then also be used to specify the output format by either specifying a zip or xml path. The file path is then just appended to the `git_repo` url in angled brackets. There should be a sensible default value here so that the api can be used like in your example.

I have tested this using a GitHub dataset repository and it works as expected. Thanks for pointing out the `URL [PATH]` syntax, it works out of the box by providing that to the `git_repository` argument so there should not be a need for `git_repository_filepath`. The issue that you linked seems to be a problem that needs to be resolved within CVAT. Until it is resolved, you can just continue to explicitly provide the path to the `git_repository` argument in our API. 

Let me know if you come across any other issues!

ehofesmann:
> So how does the CVAT git feature work? I see that it stores zips of labels in the git repo by task ID, but what happens to older zips (eg the one from upload) when a new version is posted (eg after one ore more downloads)? Are the old ones just overwritten?

Whenever the repository is first synced (either on upload on our end or manually by the user in CVAT) a new branch is created in the repo based on the unique task id and name. On future syncs, this branch gets overwritten and updated with the annotations as they appear at the time of syncing.

niclaswue:
> I have tested this using a GitHub dataset repository and it works as expected. Thanks for pointing out the `URL [PATH]` syntax, it works out of the box by providing that to the `git_repository` argument so there should not be a need for `git_repository_filepath.` The issue that you linked seems to be a problem that needs to be resolved within CVAT. Until it is resolved, you can just continue to explicitly provide the path to the `git_repository` argument in our API.

Very interesting, I tested this with an on premise GitLab repository. You're right then, this is totally something CVAT need to fix on their end.

ehofesmann:
> Very interesting, I tested this with an on premise GitLab repository. You're right then, this is totally something CVAT need to fix on their end.

@niclaswue Does adding the `[PATH]` to the `git_repository` parameter allow this to work for your GitLab repo?

brimoor:
What about tweaking this to expose this functionality as a method that users call manually on their `Results` object, similar to how things like task statuses can be checked now?

My reasoning is that:
1. This feature is, after all, tangential to FiftyOne, as our recommended practice is to use FiftyOne as the one source of truth for datasets, not some other data sink
2. Given 1, I'd rather not cloud our API with parameters that are tangential to the recommended use of our API
3. Maybe the user wants to save a version of their CVAT labels at a time other than upload/download time. They can't do that currently, but they could if this feature were exposed directly on the `Results` object

ehofesmann:
Agree with Brian that this is not a best practice, also this has gone stale

PR #1443, opened 2021-11-23
Fix error when downloading VOC dataset

kamilc:

Without the fix contained in this PR, running `fiftyone zoo datasets download voc-2012` fails with an error message saying that a list is not hashable. While the fix itself seems naive (I unfortunately didn't have enough availability to explore the codebase for a longer time) - it seems to be working fine, fixing the original issue.

I've tested it manually, by running the download for `voc-2012` and then loading the dataset and looking at the samples.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/voc.py
+++ b/fiftyone/utils/voc.py
@@ -800,6 +800,9 @@ def _parse_attribute(value):
     except:
         pass
 
+    if isinstance(value, list):
+        return value
+
     if value in {"True", "true"}:
         return True
 


ehofesmann (changes requested):
Thanks for fixing this error!

Upon further testing, this was actually a more general issue when importing any `VOCDetectionDataset` that contains attributes other than integers, floats, or strings.

For example, if you [manually download VOC2012](http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar) and attempt to import it using `fo.Dataset.from_dir(...)` then this error would arise for dict attributes that it contains as well.

Rather than checking for lists specifically, it would be more robust to check specifically if the attribute is a string and return anything else:
```python
def _parse_attribute(value):
    try:
        return int(value)
    except:
        pass

    try:
        return float(value)
    except:
        pass

    if isinstance(value, str):
        if value in {"True", "true"}:
            return True

        if value in {"False", "false"}:
            return False

        if value == "None":
            return None

    return value
```

brimoor:
@kamilc thanks for finding and fixing this!

I'm going to merge your work in now and we'll handle the generalization per Eric's request separately.

PR #1445, opened 2021-11-23
More robust attribute parsing

brimoor:
Implements https://github.com/voxel51/fiftyone/pull/1443#pullrequestreview-813998106.


ehofesmann (approved):
LGTM

PR #1449, opened 2021-11-25
Adding support for passing headers to CVAT API

brimoor:
This PR adds support for configuring custom headers to be included in all requests to a CVAT server.

This can be useful for self-hosted CVAT servers in environments like AWS that may require such headers in order to make properly authenticated requests.

## Example usage

Store custom headers in `~/.fiftyone/annotation_config.json`:

```
{
    "default_backend": "cvat",
    "backends": {
        "cvat": {
            "config_cls": "fiftyone.utils.cvat.CVATBackendConfig",
            "url": "http://localhost:8080",
            "username": "...",
            "password": "...",
            "headers": {"hello": "world"}
        }
    }
}
```

Then use the annotation API like normal:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = dataset.select_fields("ground_truth").limit(1).clone()

results = dataset.annotate("test", label_field="ground_truth")

# Verify that headers were loaded
print(api._session.headers)

dataset.load_annotations("test", cleanup=True)
```

### DIFF
--- a/docs/source/integrations/cvat.rst
+++ b/docs/source/integrations/cvat.rst
@@ -223,6 +223,7 @@ You can also store your credentials in your
     {
         "backends": {
             "cvat": {
+                ...
                 "username": ...,
                 "password": ...
             }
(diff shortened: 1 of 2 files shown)


ehofesmann (approved):
LGTM

Issue #1451, opened 2021-11-26
[BUG] Torch datasets not loading


benjaminpkane:
Torch datasets are not loading in `0.14.2` and `0.14.1` it seems. `0.14.0` seems to not have the issue.

```py
import fiftyone as fo

dataset = foz.load_zoo_dataset("cifar10", split="test")
```

Output:
```
Downloading split 'test' to '/home/ben/fiftyone/cifar10/test'
Files already downloaded and verified
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-a56cbfe97d58> in <module>
----> 1 dataset = foz.load_zoo_dataset("cifar10", split="test")

~/venvs/fnew/lib/python3.6/site-packages/fiftyone/zoo/datasets/__init__.py in load_zoo_dataset(name, split, splits, label_field, dataset_name, dataset_dir, download_if_necessary, drop_existing_dataset, overwrite, cleanup, **kwargs)
    201             overwrite=overwrite,
    202             cleanup=cleanup,
--> 203             **download_kwargs,
    204         )
    205         zoo_dataset = info.get_zoo_dataset()

~/venvs/fnew/lib/python3.6/site-packages/fiftyone/zoo/datasets/__init__.py in download_zoo_dataset(name, split, splits, dataset_dir, overwrite, cleanup, **kwargs)
    119         splits=splits,
    120         overwrite=overwrite,
--> 121         cleanup=cleanup,
    122     )
    123 

~/venvs/fnew/lib/python3.6/site-packages/fiftyone/zoo/datasets/__init__.py in download_and_prepare(self, dataset_dir, split, splits, overwrite, cleanup)
   1002                     num_samples,
   1003                     classes,
-> 1004                 ) = self._download_and_prepare(split_dir, scratch_dir, split)
   1005 
   1006                 # Add split to ZooDatasetInfo

~/venvs/fnew/lib/python3.6/site-packages/fiftyone/zoo/datasets/torch.py in _download_and_prepare(self, dataset_dir, scratch_dir, split)
    196             download_fcn,
    197             get_class_labels_fcn,
--> 198             sample_parser,
    199         )
    200 

~/venvs/fnew/lib/python3.6/site-packages/fiftyone/zoo/datasets/torch.py in _download_and_prepare(dataset_dir, scratch_dir, download_fcn, get_class_labels_fcn, sample_parser)
    534         )
    535     else:
[...]

PR #1452, opened 2021-11-26
Fixing Torch/TF zoo dataset loading bug

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1451.

```py
import fiftyone as fo
import fiftyone.zoo as foz

try:
    foz.delete_zoo_dataset("cifar10")
except:
    pass

```


benjaminpkane (approved):
LGTM :+1:

Issue #1453, opened 2021-11-28
[BUG] evaluate_detections mAP with classes specify got error


vuvalini:
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Both Ubuntu 18.04.5 /  MacOS 10.14 tested 
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: 0.14.2
-   **Python version**: Python 3.7

### Describe the problem

Error raised while  evaluate_detections with compute_mAP and classes specify.

### Code to reproduce issue

```python
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")

results = dataset.evaluate_detections(
    "predictions",
    gt_field="ground_truth",
    eval_key="eval_predictions",
    classes= ["bottle"],
    compute_mAP=True,
)
```

### Other info / logs

ValueError raised

```bash
Traceback (most recent call last):
  File "/Applications/PyCharm.app/Contents/plugins/python-ce/helpers/pydev/pydevd.py", line 1483, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/Applications/PyCharm.app/Contents/plugins/python-ce/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "********************************/scratch.py", line 12, in <module>
    compute_mAP=True,
  File "********************************/fiftyone/fiftyone/core/collections.py", line 1958, in evaluate_detections
    **kwargs,
  File "********************************/fiftyone/fiftyone/utils/eval/detection.py", line 205, in evaluate_detections
    samples, matches, eval_key=eval_key, classes=classes, missing=missing
  File "********************************/fiftyone/fiftyone/utils/eval/coco.py", line 312, in generate_results
    precision[iou_threshs.index(t)][classes.index(c)] = q
ValueError: 'bird' is not in list
python-BaseException
```

### What areas of FiftyOne does this bug affect?

-   [ ] `App`: FiftyOne application issue
-   [x] `Core`: Core `fiftyone` Python library issue
-   [ ] `Server`: Fiftyone server issue

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another
member of your organization be willing to contribute a fix for this bug to the
FiftyOne codebase?

-   [ ] Yes. I can contribute a fix for this bug independently.
[...]

PR #1454, opened 2021-11-29
Fixing #1453

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1453.

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "predictions",
    gt_field="ground_truth",
    eval_key="eval",
    method="coco",
    compute_mAP=True,
)
print(results.mAP())  # same as `develop`

dataset.delete_evaluations()

# This now works
    "predictions",
    gt_field="ground_truth",
    eval_key="eval",
    classes=["bottle"],
    method="coco",
    compute_mAP=True,
)
print(results.mAP())
```

### DIFF
--- a/fiftyone/utils/eval/coco.py
+++ b/fiftyone/utils/eval/coco.py
@@ -192,10 +192,11 @@ class COCOEvaluation(DetectionEvaluation):
         Returns:
             a :class:`DetectionResults`
         """
-        gt_field = self.config.gt_field
-        pred_field = self.config.pred_field
+        config = self.config
+        gt_field = config.gt_field
+        pred_field = config.pred_field
 
-        if not self.config.compute_mAP:
+        if not config.compute_mAP:
             return DetectionResults(
                 matches,
                 eval_key=eval_key,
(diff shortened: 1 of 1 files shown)


fiftyone/utils/eval/coco.py
```diff
@@ -675,6 +573,120 @@ def _compute_matches(
...
+    # Compute precision-recall
+    # https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocotools/cocoeval.py
+    precision = -np.ones((num_threshs, num_classes, 101))
+    recall = np.linspace(0, 1, 101)
+    for idx, _thresh_matches in enumerate(thresh_matches):
+        for c, matches in _thresh_matches.items():
+            c_idx = class_idx_map.get(c, None)
+            num_gt = matches["num_gt"]
+
+            if c_idx is None or num_gt == 0:
```
brimoor, line 650: Accounting for `c_idx == 0` was the actual error that is fixed here

fiftyone/utils/eval/coco.py
```diff
@@ -489,20 +389,18 @@ def _coco_evaluation_iou_sweep(gts, preds, config):
         gts, preds, id_keys, iou_key, config, max_preds=config.max_preds
     )
 
-    matches_dict = {
-        i: _compute_matches(
+    return [
```
brimoor, line 392: Using floating point numbers as dict keys is dangerous as subtle numeric precision errors can cause lookups to fail. This hadn't arisen in the previous implementation *yet*, but better to avoid it anyway

fiftyone/utils/eval/coco.py
```diff
@@ -206,110 +207,9 @@ def generate_results(
...
-
-                inds = np.searchsorted(rec, recall, side="left")
-                try:
-                    for ri, pi in enumerate(inds):
-                        q[ri] = pre[pi]
-                except:
-                    pass
-
-                precision[iou_threshs.index(t)][classes.index(c)] = q
+        precision, recall, iou_threshs, classes = _compute_pr_curves(
```
brimoor, line 210: Refactoring tasks into encapsulated methods helps with readability

ehofesmann (approved):
LGTM, thanks for the notes!

PR #1455, opened 2021-11-30
Handle deleted tasks during annotation with CVAT

### LINKED ISSUE #1431: [FR] Make CVAT annotation run cleanup graceful
Currently, `CVATAnnotationResults.cleanup()` will raise an exception if not all of the tasks exist.

However, I think a better behavior would be to instead log a warning but continue on anyway, since the end result is the same: the run is now deleted.

This came up for me when I killed an expensive in-progress upload but then wanted to use the API to clean it up.

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video").limit(3).select_fields().clone()

dataset.annotate(
    "test",
    label_field="ground_truth",
    label_type="detections",
    classes=["vehicle"],
    launch_editor=True,
)

# Manually delete one of the tasks in CVAT...

# This will raise an error when trying to delete the task that was manually deleted
dataset.load_annotations("test", cleanup=True)
```

ehofesmann:
Resolves #1431

Adds support for gracefully continuing an import of a CVAT annotation run that involves deleted tasks.

Deleted tasks are treated as if they never existed; IE any existing labels that were uploaded to those tasks are **not** deleted from the dataset; those sample(s) are left untouched.

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video").limit(2).select_fields().clone()

dataset.annotate(
    "test",
    label_type="detections",
    classes=["vehicle"],
    launch_editor=True,
)

# Manually delete one of the tasks in CVAT...

dataset.load_annotations("test", cleanup=True)
```

### DIFF
--- a/docs/source/integrations/lightning_flash.rst
+++ b/docs/source/integrations/lightning_flash.rst
@@ -36,7 +36,9 @@ In order to use the Lightning Flash integration, you'll need to
 
 .. code-block:: shell
 
-    pip install lightning-flash
+    # This integration currently requires these versions explicitly
+    pip install lightning-flash==0.4.0
+    pip install pytorch-lightning==1.3.6
 
 Depending on the type of Flash tasks that you intend to use, you will also need
 to install some package extras:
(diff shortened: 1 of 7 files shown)


brimoor (commented):
I updated the implementation of this feature a bit.

Setting `allow_deletions=False` when deleted tasks are encountered is a hack, because, after all, the user may want `allow_deletions=True` for the remaining tasks. I believe the best behavior is to treat deleted tasks as if they were never included in the run in the first place.

brimoor (approved):
LGTM

Issue #1462, opened 2021-12-07
[BUG] Reserved attribute name discrepancy


benjaminpkane:
Note the discrepancy below:

```py
import fiftyone as fo

# works
sample = fo.Sample(filepath="image.png", objects="object")
dataset = fo.Dataset()
dataset.add_sample(sample)

# works
dataset = fo.Dataset()
dataset.add_sample_field("objects", fo.StringField)

# error
sample = fo.Sample(filepath="image.png")
dataset = fo.Dataset()
dataset.add_sample(sample)
sample["objects"] = "object"
```

brimoor:
There's at least one place in the documentation that should be updated once this is addressed:

<img width="771" alt="Screen Shot 2021-12-07 at 7 11 31 PM" src="https://user-images.githubusercontent.com/25985824/145125418-74648bd2-d399-43cf-a77d-ff434d2c406a.png">

brimoor:
Update: as per https://github.com/voxel51/fiftyone/pull/1469, I think the proper way for us to address this will be to update our `Document` classes so that fields can have arbitrary names and can always be accessed via `document[field]`, and `document.field` will default to a class method rather than user-defined fields (eg `field="to_dict"`) in the case of a clash.

Unfortunately, this is not easily achievable until we get rid of MongoEngine.

PR #1469, opened 2021-12-10
Partial work on #1462

brimoor:
I spent more time than I'd like to admit trying to properly fix https://github.com/voxel51/fiftyone/issues/1462, but this is where I got.

Ideally we would update the implementation of our `Document` classes so that fields can have arbitrary names and can always be accessed via `document[field]`, which is important in the case where a field clashes with a method name on the class, like `field="to_dict"`, which currently breaks things, since `document.to_dict` will currently return the user's data rather than invoke the method. This needs to be updated so that `document.to_dict` will always reference the method rather than the user's data in the case of clashes. Unfortunately, that's not easily achievable until we get rid of MongoEngine.

For now, I just removed the error that prevents `sample["objects"] = "object"` from being allowed, since as the code below shows, the other pathways for assigning an `objects` field are *not* currently blocked.

tl dr; until we get rid of MongoEngine, behavior is undefined if the user happens to choose a field name that clashes with a method name.

```py
import fiftyone as fo

# works
sample = fo.Sample(filepath="image.png", objects="object")
dataset = fo.Dataset()
dataset.add_sample(sample)

# works
dataset = fo.Dataset()
dataset.add_sample_field("objects", fo.StringField)

# previously would raise an error; now is allowed
sample = fo.Sample(filepath="image.png")
dataset = fo.Dataset()
dataset.add_sample(sample)
sample["objects"] = "object"
```

### DIFF
--- a/fiftyone/core/odm/document.py
+++ b/fiftyone/core/odm/document.py
@@ -33,21 +33,6 @@ class SerializableDocument(object):
 
         return self.to_dict() == other.to_dict()
 
-    def __getattr__(self, name):
-        try:
-            return object.__getattr__(self, name)
-        except AttributeError:
-            pass
-
-        return self.get_field(name)
-
-    def __setattr__(self, name, value):
-        try:
-            object.__getattr__(self, name)
-            object.__setattr__(self, name, value)
-        except AttributeError:
-            self.set_field(name, value)
-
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
LGTM

PR #1478, opened 2021-12-16
Return sample IDs with unexpected annotations

### LINKED ISSUE #1477: [FR] Return info for unexpected labels
At the moment I am using
```
dataset.load_annotations(run, skip_unexpected=True)
```
to load annotations from CVAT into fiftyone. This logs a warning in the console when unexpected labels occur, otherwise it silently ignores the labels. For me, the code is running on a server where no interactive prompt is desired as this would halt the server. Therefore, my only option for `skip_unexpected` is `True`. I think it would be really nice, if there was some return value which indicates whether unexpected labels occured. Or even better, the ability to tag samples with unexpected labels. This way I can exclude them from my dataset later or relabel them afterward.
What do you think?

ehofesmann:
Resolves #1477

When unexpected annotations have been added during an annotation run, the only way to load them currently is with access to an interactive prompt which may not always be available in all workflows. This PR replaces the old `skip_unexpected=True/False` parameter of `load_annotations()` with a more powerful `unexpected` parameter with the following options:

```
        unexpected ("prompt"): how to deal with any unexpected labels that
            don't match the run's label schema when importing. The supported
            values are:

            -   ``"prompt"``: present an interactive prompt to direct/discard
                unexpected labels
            -   ``"ignore"``: automatically ignore any unexpected labels
            -   ``"return"``: return a dict containing all unexpected labels,
                or ``None`` if there aren't any
```

Example usage:

```python
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=3).clone()

anno_key = "unexpected"
dataset.annotate(
    anno_key,
    label_field="ground_truth",
    launch_editor=True,
)

# Add types other than detections and polygons

unexpected_annos = dataset.load_annotations(anno_key, unexpected="return", cleanup=True)
print(unexpected_annos)
```

### DIFF
--- a/docs/source/integrations/cvat.rst
+++ b/docs/source/integrations/cvat.rst
@@ -1003,6 +1003,20 @@ to see the available keys on a dataset.
     However, you can pass `cleanup=True` to delete all information associated
     with the run from the backend after the annotations are downloaded.
 
+Note that CVAT cannot explicitly prevent annotators from creating labels that
+don't obey the run's label schema. However, you can pass the optional
+`unexpected` parameter to
+:meth:`load_annotations() <fiftyone.core.collections.SampleCollection.load_annotations>`
+to configure how to deal with any such unexpected labels that are found. The
+supported values are:
+
+-   `"prompt"` (**default**): present an interactive prompt to direct/discard
+    unexpected labels
+-   `"ignore"`: automatically ignore any unexpected labels
+-   `"return"`: return a dict containing all unexpected labels, if any
+
+See :ref:`this section <cvat-unexpected-annotations>` for more details.
+
 .. _cvat-managing-annotation-runs:
 
 Managing annotation runs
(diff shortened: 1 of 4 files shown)


brimoor (changes requested):
Shouldn't you return the unexpected labels themselves (not just the sample IDs)? It seems to me that if the user doesn't want to use `skip_unexpected=True` then they definitely want a way to get ahold of the actual labels.

The point of avoiding an interactive prompt definitely makes sense. A full solution to me could be to upgrade the `skip_unexpected=True/False` to a more generic `unexpected` argument with this behavior:
- `"prompt"` (default): interactively prompt users to deal with unexpected labels if they exist 
- `"skip"`: skip unexpected labels
- `"return"`: return a dict containing all unexpected labels, or ``None`` if there aren't any
- a dict mapping label types to field names in which to store unexpected labels of these types, if encountered. In this case, any unexpected labels with types not in this dict will be ignored

After typing this out, the last setting seems pretty silly. If the user knows enough to configure this setting, they wouldn't have unexpected labels in the first place.

brimoor (approved):
LGTM. I went ahead and implemented the `unexpected` parameter per my proposal.

Issue #1223, opened 2021-08-25
[FR] Update similarity view stage in-place when multiple consecutive searches are performed


brimoor:
When working with visual similarity from the App, I find myself doing the click image + sort by similarity workflow many times. Currently, each time results in a new view stage, which starts to get out-of-hand in the view bar.

Instead, it would be nice if consecutive searches with the same `k` value did not result in a new stage, since the first sort is completely unnecessary in that case.

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")
fob.compute_similarity(dataset, brain_key="image_sim")

# Do multiple image similarity sorts
session = fo.launch_app(dataset)
```

<img width="1273" alt="Screen Shot 2021-08-24 at 11 57 39 PM" src="https://user-images.githubusercontent.com/25985824/130724915-4430e0c3-0597-4bbc-b2ff-fb9d4246ca8d.png">

cavesdev:
Hello! Is someone already working on this?

benjaminpkane:
@cavesdev no, this is open. Feel free to work on this.

brimoor:
This has been addressed as of `fiftyone==0.20`!

PR #1481, opened 2021-12-23
Update visual similarity stage in-place when possible in App workflows

cavesdev:
Solves #1223 

Visual similarity now updates the stage if the k value is unchanged from the previous one. 

Quick test:
```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

fob.compute_similarity(dataset, brain_key="image_sim")

query_id = dataset.first().id
stage = fo.SortBySimilarity(query_id)
view = dataset.add_stage(stage)

query_id = dataset.last().id
stage = fo.SortBySimilarity(query_id)
view = view.add_stage(stage)

session.view = view
session.wait()
```

Before:
Screen Shot 2021-12-22 at 20 13 49

After:
Screen Shot 2021-12-22 at 20 09 51

Additionally, you can try the click image + sort by similarity UI workflow. 

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/view.py
+++ b/fiftyone/core/view.py
@@ -779,9 +779,30 @@ class DatasetView(foc.SampleCollection):
             view = stage.load_view(self)
         else:
             view = copy(self)
-            view._stages.append(stage)
+
+            if not self._add_sort_stage(stage, view._stages):
+                view._stages.append(stage)
 
         return view
+
+    @staticmethod
+    def _add_sort_stage(new_stage, stages):
+        if not type(new_stage) == fost.SortBySimilarity:
+            return False
+
+        sort_views = [
+            i for i, x in enumerate(stages) if type(x) == fost.SortBySimilarity
+        ]
+
+        if len(sort_views) > 0:
+            sort_views.reverse()
+            last_sort_view = sort_views[0]
+
+            if stages[last_sort_view].k == new_stage.k:
+                stages[last_sort_view] = new_stage
+                return True
+
+        return False
 
     def _get_filtered_schema(self, schema, frames=False):
         if schema is None:


benjaminpkane:
Hi @cavesdev, I like what you did here, but we are looking for an App-side solution (TypeScript). I can point you in the right direction if that is in your wheelhouse.

cavesdev:
I see that stages are mainly handled in the backend, that is why I implemented the function there. 
Is there a function that *updates* the stage instead of adding a new one? 

As I understand, this function is the one that *adds* a new stage.

https://github.com/voxel51/fiftyone/blob/06975961f5ee649dd36429feb21c959dfc0744ed/app/packages/app/src/components/Actions/Similar.tsx#L75-L89

brimoor:
@benjaminpkane can you help @cavesdev figure out the App-side implementation per his question?

@cavesdev thanks for taking this issue on! Like Ben mentioned, we think an App-side implementation makes the most sense here because, for API-based usage, we think it's best to avoid any "magic" optimizations and instead allow users to chain arbitrary view stages together, even if, strictly speaking, certain stages are nullified by future stages.

The duplicate `SortBySimilarity` stages from #1223 is strictly a consequence of the point-and-click workflows in the UI, so an App-side resolution seems appropriate.

benjaminpkane:
Sorry, I missed this. No, there is not a function that updates a stage.

In the hook you referenced (within `useRecoilCallback`) you will have to interact with the `view` selector. 

```ts 
const params = await snapshot.getPromise(sortBySimilarityParameters);
const queryIds = await getQueryIds(snapshot, params.brainKey);
const view = await snapshot.getPromise(selectors.view);
set(similaritySorting, true);
set(atoms.modal, null);

// `view` is an array of serialized stages (readonly).
// Copy and modify for the new sorting (adding or editing a sort by stage)

set(selectors.view
```

benjaminpkane:
I'm not totally on board with the idea of inspecting and modifying the view indirectly (outside the view bar). I'd rather see sorting by similarity moved to the extended view in some way. Anyways, this is fine with me for now.

brimoor:
> I'm not totally on board with the idea of inspecting and modifying the view indirectly (outside the view bar). I'd rather see sorting by similarity moved to the extended view in some way. Anyways, this is fine with me for now.

Yeah I agree that similarity searches in the App shouldn't immediately be saved to the view bar. The extended view is the right place for it. Moving the similarity UI to the sidebar will take a bit of work though.

Is there a clean way to move the *logic* into the extended view but just keep the "state" of any active similarity search saved in the similarity menu itself? If the user executes a new similarity search, it would strictly overwrite the previous one (unless they first bookmark the extended view). The only requirement I see is that there be a clear visual indicator so that the user knows when they have a similarity search active that they might want to clear. I guess simply keeping the similarity icon visible even when no samples are selected could be enough.

benjaminpkane:
@cavesdev Noting the above, to separate the similarity sorting from the current view, you'll need some knowledge of [recoil](https://recoiljs.org/).

Specifically, you will want to create a new `atom` that contains the sorting parameters. It can be defined `Similar.tsx`. Then you will include that atom as a part of `page` requests (defined in [Flashlight.tsx](https://github.com/voxel51/fiftyone/blob/develop/app/packages/app/src/components/Flashlight.tsx) and on the backend receive the new paramater(s) in [main.py](https://github.com/voxel51/fiftyone/blob/b9b5b060affa34eb3ccfaf40adafbd5fe219603a/fiftyone/server/main.py#L224).

Because we are only sorting, only page queries need to be aware of the setting.

There are obviously more details, UX, etc. But if you are able to get sorting to work outside the view bar, we can then add pinning the sorting with filters and other details.

cavesdev:
Sorry for the late reply, new year was around the corner. Thanks for the pointers (and the opportunity to work on this issue!), working on it!

benjaminpkane (approved):
@cavesdev We're going to try to get this into the next release. I can take it from here. It requires subtle changes in a handful of places.

I can point you in the direction a bug or two which may be more manageable, if you wish to make another contribution. Cheers

Issue #1482, opened 2021-12-23
[BUG] Lightning Flash integration doesn't work with current Flash version


brimoor:
**tl;dr** our Flash integration needs to be updated to support the latest Flash version

Our [Lightning Flash integration](https://voxel51.com/docs/fiftyone/integrations/lightning_flash.html) was originally written for these versions of Flash/Pytorch Lightning:

```
pip install lightning-flash==0.4.0
pip install pytorch-lightning==1.3.6
```

With the above versions of those packages installed, I am still able to run the detection and embeddings examples from our integration docs. However, I am no longer able to load the pre-trained image checkpoint:

```
rm -rf ~/.cache/torch/hub/checkpoints
```

```py
from flash.image import ImageClassifier

model = ImageClassifier.load_from_checkpoint(
    "https://flash-weights.s3.amazonaws.com/image_classification_model.pt"
)
```

```
AttributeError: Can't get attribute 'image_loader' on <module 'flash.image.data' from '/Users/Brian/dev/env/fifty3/lib/python3.6/site-packages/flash/image/data.py'>
```

If I upgrade to the latest version of Lightning Flash (0.6.0 as of this writing) then our examples no longer work.

Example code I tried running:

```py
from flash.image import ImageClassifier, ObjectDetector

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=5)

model = ObjectDetector.load_from_checkpoint(
    "https://flash-weights.s3.amazonaws.com/object_detection_model.pt"
)
dataset.apply_model(model, label_field="flash_detector")

model = ImageClassifier.load_from_checkpoint(
    "https://flash-weights.s3.amazonaws.com/image_classification_model.pt"
)
dataset.apply_model(model, label_field="flash_classifier")
```

MrBlenny:
Would love to see a fix for this.
I've tried updating the imports but there seem to be some other issues. With a little bit of direction I'd be happy to make a PR :)

andife:
I also quite interested in a new working version of integrationof the "semantic segmentation" example / use case. In the meanwhile up to the time the fix is available maybe adding a comment that newer flash version are not supported yet, would be nice. Thank you!

PR #1483, opened 2021-12-23
Note explicit Flash version dependency

brimoor:
Until https://github.com/voxel51/fiftyone/issues/1482 is addressed, this PR simply documents the version of Flash/PyTorch-Lightning for which our Flash integration was implemented.


benjaminpkane (approved):
LGTM

PR #1411, opened 2021-11-11
Updating dependencies for build workflows

brimoor:
Adds `requirements/extras.txt` for extra packages that can be cross-platform installed via pip for certain optional features.

These requirements are now installed in docs/package build workflows (these had gotten out-of-date and were breaking the docs builds of late).

Unfortunately things like `torch`, `tensorflow`, and `pycocotools` can't be included in this because they require Windows-specific installs.

PR #1485, opened 2021-12-24
Reverting nbsphinx version

brimoor:
The current section highlight in the RHS table of contents in the docs is currently broken in production. I believe the `nbsphinx` version change in https://github.com/voxel51/fiftyone/pull/1411 was the culprit.


benjaminpkane (approved):
LGTM

PR #1491, opened 2021-12-27
Documenting default COCO label_types

brimoor:
Users have (rightly) been confused by the fact that `COCODetectionDatasetImporter` defaults to `label_types="detections"`.

This PR updates the documentation to explicitly mention this and to clearly document how to load additional label types like segmentations if they are present.

Screen Shot 2021-12-27 at 3 10 30 PM


benjaminpkane (approved):
LGTM

PR #1492, opened 2021-12-28
[HOLD] Adding support for saving changes made to AnnotationResults

brimoor:
Adds an `AnnotationResults.save()` method that allows for saving changes to an existing annotation run.

In order to implement this feature, the `AnnotationResults` object needs knowledge of the `anno_key`, and it seemed to me that the best way to make that happen was to add support for passing run keys when loading all `RunResults` objects. But, this does increase the footprint of this PR, so I've marked it as a HOLD for now to ponder...

```py
# Load a CVAT annotation run

# Use API to modify tasks

# Or, delete tasks this way
results.delete_tasks(...)

# Or, delete tasks this way
results.cleanup()

# This PR adds this syntax...
results.save()

# ...because the way to achieve this before is ugly
results._backend.save_run_results(dataset, anno_key, results)
```

### DIFF
--- a/fiftyone/core/runs.py
+++ b/fiftyone/core/runs.py
@@ -473,7 +473,7 @@ class Run(Configurable):
         results_str = run_doc.results.read().decode()
 
         try:
-            run_results = RunResults.from_str(results_str, view, config)
+            run_results = RunResults.from_str(results_str, view, config, key)
         except Exception as e:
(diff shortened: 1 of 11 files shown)


brimoor:
Implemented in https://github.com/voxel51/fiftyone/pull/2696 instead.

PR #1493, opened 2021-12-28
Adding support for passing data map directory to importers

brimoor:
Adds support for directly passing a `filename -> filepath` mapping dict to the `data_path` parameter of importers.

Previously one could only pass the path to a JSON path on disk containing the mapping, but in many workflows it is more natural to construct + provide the mapping in-memory.


benjaminpkane (approved):
LGTM

PR #1495, opened 2021-12-29
Sanitize CVAT URLs

ehofesmann:
This PR allows CVAT server URLs to contain a trailing `/` without raising errors when working with our CVAT API.

```python

# Works
url = "https://cvat.org"
api = fouc.CVATAnnotationAPI("cvat", url)

# Previously raised 405 error, now works
url = "https://cvat.org/"
api = fouc.CVATAnnotationAPI("cvat", url)
```


brimoor (approved):
LGTM

PR #1505, opened 2022-01-03
Proper handling of ID fields in set_values(..., key_field="id")

brimoor:
Fixes a bug that prevented `set_values(field_name, values, key_field="id")` from working as expected:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Previously this would raise an error
values = {k: v for k, v in zip(*dataset.values(["id", "filepath"]))}
dataset.set_values("filepath2", values, key_field="id")

# Now the above code runs, and this assertion passes
assert dataset.values("filepath") == dataset.values("filepath2")
```


ehofesmann (approved):
LGTM

PR #1506, opened 2022-01-04
Adding support for setting frame-level fields by frame number in set_values()

brimoor:
When setting frame-level fields via `set_values(..., values, ...)`, this PR adds support for providing `values` whose elements are dicts mapping frame numbers to values.

Previously, the elements of `values` in this case could only be lists corresponding 1-1 with frames, which was inconvenient for at least these reasons:
- In many cases, one may have values for a sparse set of frames << all frames. With the list syntax, this would have previously required inserting `[]` elements in `values` for the skipped frames
- `Frame` documents are only created on-demand, so the user may not readily know which frames have been declared to construct the 1-1 list
- Point 2 also meant that all frames must have previously been declared; ie `set_values()` could not populate fields of previously uninstantiated frames

The dict syntax in this PR resolves all three of these issues, and, in doing so, becomes a far more convenient way to set frame-level fields of video datasets.

Example usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

#
# Method 1
# Here `num_detections` is a list of lists corresponding to existing frames
#

num_detections = dataset.values(F("frames").map(F("detections.detections").length()))

dataset.set_values("frames.num_detections1", num_detections)

#
# Method 2
# Both the samples and frame numbers are specified via dicts
#

values = {}
for filepath, frame_counts in zip(dataset.values("filepath"), num_detections):
    values[filepath] = {fn: count for fn, count in enumerate(frame_counts, 1)}

dataset.set_values("frames.num_detections2", values, key_field="filepath")

print(dataset.bounds("frames.num_detections1"))
print(dataset.bounds("frames.num_detections2"))
```


ehofesmann (approved):
LGTM, frame number dicts are a great update, much easier to use.

Issue #1502, opened 2022-01-03
[BUG] Filtering by a datetime field makes the server raise an exception


FGeri:
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 20.04.2 LTS
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: v0.14.2
-   **Python version**: 3.8

### Commands to reproduce

-

### Describe the problem

When trying to filter by the sliders on a datetime field in the app, the server raises an uncaught exception, hence the filtering is not done.

### Code to reproduce issue

-

### Other info / logs

```
An exception has been raised by the server. Your session has been reverted to its previous state.

Traceback (most recent call last):
  File "main.py", line 313, in wrapper
    result = await func(self, *args, **kwargs)
  File "main.py", line 556, in on_message
    await event(self, **message)
  File "main.py", line 624, in on_filters_update
    await self.send_statistics(view, filters=filters, extended=True)
  File "main.py", line 1031, in send_statistics
    results = await view._async_aggregate(aggregations)
  File "/home/fhe2bp/venvs/datalake/lib/python3.8/site-packages/fiftyone/core/collections.py", line 6784, in _async_aggregate
    _results = await foo.aggregate(collection, pipelines)
  File "/home/fhe2bp/venvs/datalake/lib/python3.8/site-packages/fiftyone/core/odm/database.py", line 191, in _do_async_pooled_aggregate
    return await asyncio.gather(
  File "/home/fhe2bp/venvs/datalake/lib/python3.8/site-packages/fiftyone/core/odm/database.py", line 201, in _do_async_aggregate
    return await cursor.to_list(1)
  File "/home/fhe2bp/venvs/datalake/lib/python3.8/site-packages/motor/core.py", line 1417, in _to_list
    result = get_more_result.result()
  File "/home/fhe2bp/venvs/datalake/lib/python3.8/site-packages/motor/core.py", line 1626, in _on_started
    pymongo_cursor = future.result()
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/fhe2bp/venvs/datalake/lib/python3.8/site-packages/pymongo/collection.py", line 2502, in aggregate
    return self._aggregate(_CollectionAggregationCommand,
[...]

PR #1508, opened 2022-01-04
Fixing expressions containing near-epoch dates

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1502.

The issue was that apparently MongoDB's `{$toDate: arg}` for some reason seems to require `arg` to be a `float` or a `long` but **NOT** an `int`. The lack of `int` support meant that expressions involving datetimes too close to the epoch would result in MongoDB expressions involving ints that would raise an error!

```py
from datetime import datetime

import fiftyone as fo

dataset = fo.Dataset()
dataset.add_samples(
    [
        fo.Sample(
            filepath="image1.png",
            capture_date=datetime(1970, 1, 1, 1, 0, 0),
        ),
        fo.Sample(
            filepath="image2.png",
            capture_date=datetime(1970, 1, 1, 2, 0, 0),
        ),
        fo.Sample(
            filepath="image3.png",
            capture_date=datetime(1970, 1, 1, 3, 0, 0),
        ),
    ]
)

# Used to fail; now works
print(dataset.match(F("capture_date") > datetime(1970, 1, 1, 2, 0, 0)))
```

### DIFF
--- a/fiftyone/core/expressions.py
+++ b/fiftyone/core/expressions.py
@@ -4616,6 +4616,7 @@ def _do_to_mongo(val, prefix):
         return [_do_to_mongo(v, prefix) for v in val]
 
     if isinstance(val, (date, datetime)):
+        # The arg needs must be float (not int) to avoid errors near the epoch
         return {"$toDate": fou.datetime_to_timestamp(val)}
 
     if isinstance(val, timedelta):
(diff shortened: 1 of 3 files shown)


ehofesmann (approved):
LGTM

PR #1510, opened 2022-01-05
Import existing CVAT datasets into FiftyOne (#2)

brimoor:
An alternative implementation of https://github.com/voxel51/fiftyone/pull/1456 that also implements the TODO items mentioned there.

Other changes
- Added new ``unexpected="keep"`` option to `load_annotations()`
- `%06d_` is no longer prepended to video filenames in CVAT, since it is not necessary (always one video per task)
- Optimized database I/O in `_create_shapes_tags_tracks()` and `_build_frame_id_map()`

## Image dataset example

```py
import os

import fiftyone as fo
import fiftyone.zoo as foz

dataset = dataset.select_fields("ground_truth").clone()

# Create a CVAT project with labels
task_size = 10
results = dataset.annotate(
    "test",
    label_field="ground_truth",
    task_size=task_size,
    segment_size=3,
    project_name="test",
)
# print(results.task_ids)  # if using `task_ids` instead of `project_name`

#
# In order to load annotations, we need a mapping from CVAT image filenames to
# filepaths on local disk
#
# Since we're using a CVAT task uploaded via FiftyOne, the mapping is a bit
# weird
#
data_map = {}
for batch in fou.iter_batches(dataset.values("filepath"), task_size):
    for idx, filepath in enumerate(batch):
        filename = "%06d_%s" % (idx, os.path.basename(filepath))
        data_map[filename] = filepath

# Default: import into fields matching the label type (eg "detections")
fouc.import_annotations(
    dataset,
    project_name="test",
    data_path=data_map,
)

# But, we can instead prompt for field names
fouc.import_annotations(
    dataset,
    project_name="test",
    data_path=data_map,
    label_types="prompt",
)

# Or we can explicitly specify where to import
fouc.import_annotations(
    dataset,
    project_name="test",
    data_path=data_map,
    label_types={"detections": "ground_truth2"},
)

```

## Video dataset example

```py
import os

import fiftyone as fo
import fiftyone.zoo as foz

# Create a CVAT project with labels
results = dataset.annotate(
    "test",
    project_name="test-video",
)

dataset.rename_frame_field("detections", "detections_orig")

fouc.import_annotations(dataset, project_name="test-video")

print(dataset.count_values("frames.detections_orig.detections.label"))
print(dataset.count_values("frames.detections.detections.label"))
```

### DIFF
--- a/docs/source/integrations/cvat.rst
+++ b/docs/source/integrations/cvat.rst
@@ -2113,6 +2113,86 @@ every 10th frame as a keyframe to provide a better editing experience in CVAT:
     with the shape's contents. See :ref:`this section <cvat-limitations>` for
     details.
 
+.. _cvat-importing:
+
+Importing datasets from CVAT
+____________________________
+
+The previous examples showed ways to link existing an FiftyOne |Dataset| to
+annotation tasks and projects in CVAT. However, you may have a CVAT project or
+task that you want to use to create a |Dataset|.
+
+One way to create a |Dataset| using existing data and annotations from
+CVAT is to simply export the CVAT dataset to disk and import it using the
+:ref:`CVATImageDataset <CVATImageDataset-import>`
+or 
+:ref:`CVATVideoDataset <CVATVideoDataset-import>`
+types.
+
+Alternatively, you can use the
+:func:`import_dataset() <fiftyone.utils.cvat.import_dataset>` CVAT utility to provide
+paths to local data and the name of a CVAT project or list of CVAT task IDs
+to automatically create a |Dataset|.
+
+The following parameters are supported by :func:`import_dataset() <fiftyone.utils.cvat.import_dataset>`:
+
+-   **data_path**: a parameter that enables explicit control
+    over the location of the media. Can be any of the following:
+
+    -   an absolute directory path where the media files reside
+    -   an absolute filepath specifying the location of the JSON data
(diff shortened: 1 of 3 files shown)


ehofesmann (approved):
LGTM! There are two issues that I resolved in #1513 (`data_path` points to a json file and that `build_frame_id_map` needed to ensure frames). We may want to merge that in before merging this into develop.

One thing to note is that different tasks in CVAT can use the same filename for different images. This caused some weird behavior, but after thinking about it, I decided that we should just continue to treat them as if they were the same image and leave it to the user to ensure unique filenames in CVAT. We may want to think about logging a warning about it though.

brimoor:
> LGTM! There are two issues that I resolved in #1513 (`data_path` points to a json file and that `build_frame_id_map` needed to ensure frames). We may want to merge that in before merging this into develop.
> 
> One thing to note is that different tasks in CVAT can use the same filename for different images. This caused some weird behavior, but after thinking about it, I decided that we should just continue to treat them as if they were the same image and leave it to the user to ensure unique filenames in CVAT. We may want to think about logging a warning about it though.

Thanks for the review 💪 Your comments made me realize there was previously a bug in situations where a user provides a `data_path` that only includes a subset of the files in the specified tasks. I updated things so that it correctly handles that case now, which required adding the same fixes you mentioned in #1513.

Issue #1494, opened 2021-12-29
[BUG] Cannot import name '_registerMatType' from 'cv2.cv2'


Saransh-cpp:
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Google Colab
-   **FiftyOne installed from (pip or source)**: `pip`
-   **FiftyOne version (run `fiftyone --version`)**: `0.14.2`
-   **Python version**: `3.7.12`

### Commands to reproduce

Run the following in a `Colab` notebook -
```
%pip install fiftyone
import fiftyone as fo
```

### Describe the problem

Running the commands above in a Colab notebook (`quickstart.ipynb`) gives an import error. My initial guess is that this is a version conflict issue but I couldn't explore this a lot as I am not familiar with `fiftyone` (a new user). 

### Code to reproduce issue

```py
import fiftyone as fo
```

### Other info / logs

```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-9-8c082ccbb59f> in <module>()
----> 1 import fiftyone as fo

9 frames
/usr/local/lib/python3.7/dist-packages/fiftyone/__init__.py in <module>()
     23 __version__ = _foc.VERSION
     24 
---> 25 from fiftyone.__public__ import *
     26 from fiftyone.core.uid import log_import_if_allowed as _log_import
     27 from fiftyone.migrations import migrate_database_if_necessary as _migrate

/usr/local/lib/python3.7/dist-packages/fiftyone/__public__.py in <module>()
      7 """
      8 import fiftyone.core.config as foc
----> 9 import fiftyone.core.odm as foo
     10 
     11 config = foc.load_config()

/usr/local/lib/python3.7/dist-packages/fiftyone/core/odm/__init__.py in <module>()
     55     validate_fields_match,
     56 )
---> 57 from .sample import (
     58     DatasetSampleDocument,
     59     NoDatasetSampleDocument,

/usr/local/lib/python3.7/dist-packages/fiftyone/core/odm/sample.py in <module>()
     63 import fiftyone.core.fields as fof
     64 import fiftyone.core.frame_utils as fofu
---> 65 import fiftyone.core.metadata as fom
     66 import fiftyone.core.media as fomm
     67 import fiftyone.core.utils as fou

/usr/local/lib/python3.7/dist-packages/fiftyone/core/metadata.py in <module>()
     15 
     16 import eta.core.utils as etau
[...]

PR #1511, opened 2022-01-05
Add comment about OpenCV installation to notebooks

ehofesmann:
Manages #1494 for the time being by adding a note to tutorials and recipes to install OpenCV if the import error is occurring.

image


brimoor (approved):
LGTM. I like the note version 👍

PR #1518, opened 2022-01-07
Docs updates

brimoor:
- Removes COCO banner
- Updates install count (but @benjaminpkane that code is not being used anymore right? can it just be removed?)


benjaminpkane (approved):
LGTM

PR #1524, opened 2022-01-11
Add utility to download videos from YouTube

ehofesmann:
Adds a `fiftyone.utils.youtube.download_youtube_videos()` method that provides a highly customizable and efficient mechnism for downloading videos from YouTube.

Features include:
- Ability to specify a desired resolution and codec to download (if possible)
- Multiple threads (full videos) or processes (video clips) are used to make the download efficient
- Support for gracefully skipping non-existent (eg deleted) videos
- Ability to specify a long list of videos and a maximum number of videos that should be downloaded, so that missing videos can be gracefully skipped until the requested number of videos is downloaded
- Ability to download specific clip segments of videos, efficiently without downloading the entire video first

Example usages:

```py
import fiftyone as fo
import fiftyone.utils.youtube as fouy

# By default, the highest resolution progressive MP4 stream is downloaded
downloaded, _ = fouy.download_youtube_videos(
    urls=["https://www.youtube.com/watch?v=rC_beNmWPCM"],
    download_dir="/tmp/youtube",
)
print(fo.VideoMetadata.build_for(downloaded[0]))

# Download the highest resolution stream of any kind
downloaded, _ = fouy.download_youtube_videos(
    urls=["https://www.youtube.com/watch?v=rC_beNmWPCM"],
    download_dir="/tmp/youtube",
    ext=None,
    only_progressive=False,
    resolution="highest",
)
print(fo.VideoMetadata.build_for(downloaded[0]))

# Download an .mov stream, if possible
downloaded, _ = fouy.download_youtube_videos(
    urls=["https://www.youtube.com/watch?v=rC_beNmWPCM"],
    download_dir="/tmp/youtube",
    ext=".mov",
)
print(fo.VideoMetadata.build_for(downloaded[0]))

# Download the (progressive MP4) stream with resolution closest to 400p
downloaded, _ = fouy.download_youtube_videos(
    urls=["https://www.youtube.com/watch?v=rC_beNmWPCM"],
    download_dir="/tmp/youtube",
    resolution="400p",
)
print(fo.VideoMetadata.build_for(downloaded[0]))

# Don't log warnings
downloaded, _ = fouy.download_youtube_videos(
    urls=["https://www.youtube.com/watch?v=rC_beNmWPCM"],
    download_dir="/tmp/youtube",
    resolution="400p",
    ext=".mov",
    quiet=True,
)
print(fo.VideoMetadata.build_for(downloaded[0]))

# Explicitly specify where to download the video
# Note that the file extension might change if a compatible stream isn't found
# Use the `downloaded` argument to see what videos are actually named
downloaded, _ = fouy.download_youtube_videos(
    urls=["https://www.youtube.com/watch?v=rC_beNmWPCM"],
    video_paths=["/tmp/youtube/Voxel51 Overview.mov"],
)
print(downloaded[0])

# By default, non-existent videos are gracefully skipped
# You can use the returned `errors` to act on missing videos
_, errors = fouy.download_youtube_videos(
    urls=["https://www.youtube.com/watch?v=XXXXXXXXXXX"],
    download_dir="/tmp/youtube",
)
print(errors)

# Or you can make missing videos be fatal
fouy.download_youtube_videos(
    urls=["https://www.youtube.com/watch?v=XXXXXXXXXXX"],
    download_dir="/tmp/youtube",
    skip_failures=False,
)

# YouTube videos are often taken down; you can specify a long list of `urls`
# together with a `max_videos` if you only need a certain number of videos and
# are willing to skip missing videos
fouy.download_youtube_videos(
    urls=[
        "https://www.youtube.com/watch?v=XXXXXXXXXXX",
        "https://www.youtube.com/watch?v=YYYYYYYYYYY",
        "https://www.youtube.com/watch?v=ZZZZZZZZZZZ",
        "https://www.youtube.com/watch?v=rC_beNmWPCM",
    ],
    download_dir="/tmp/youtube",
    max_videos=1,
)

# Use `clip_segments` to download specific clips into videos
downloaded, _ = fouy.download_youtube_videos(
    urls=["https://www.youtube.com/watch?v=rC_beNmWPCM"],
    video_paths=["/tmp/youtube/brian.mov"],
    clip_segments=[(91, 95)],
)
print(fo.VideoMetadata.build_for(downloaded[0]))
```


fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+def _do_download(args):
+    url, video_path, clip_segment, download_dir, ext, buffer_seconds = args
+    ydl_opts, output_path = _build_ydl_opts(
+        ext, video_path, clip_segment, download_dir
+    )
+
+    try:
+        if clip_segment in [None, [None, None], [0, None]]:
```
brimoor, line 265: Need to be careful that `[None, None] != (None, None)`. I'd recommend doing that parsing in `_build_tasks_list()` so that, by the time we get here, `clip_segment` is either `None` or `(start, stop)`.

fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+):
+    if isinstance(urls, str):
+        urls = [urls]
+    if isinstance(clip_segments, tuple):
+        clip_segments = [clip_segments]
+
+    if isinstance(urls, list):
+        if download_dir is None:
+            raise ValueError(
+                "When `urls` is a list, `download_dir` is required but was found to be `None`."
```
brimoor, line 140: nit: line limit violation

fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+            used
+        ext (None): the extension to use to store the downloaded videos or a
+            list of extensions corresponding to the given ``urls``. By default,
+            the default extension of each video is used 
+        buffer_seconds (5): the number of additional seconds before and after
+            each clip to download to ensure nearest keyframes are captured if
+            `clip_segments` is provided
+
+    Returns:
+        urls of successfully downloaded videos
```
brimoor, line 92: how about a dict mapping URLs to local paths, for convenience?

fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+                    buffer_seconds=buffer_seconds,
+                )
+
+        return True, url, None
+
+    except Exception as e:
+        if isinstance(e, youtube_dl.utils.DownloadError):
+            # pylint: disable=no-member
+            return False, url, str(e.exc_info[1])
+        return False, url, "other"
```
brimoor, line 297: `str(e)` rather than `"other"`?

fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+            else:
+                errors[error_type].append(url)
+    return downloaded, errors
+
+
+def _multi_thread_download(tasks, max_videos, num_workers):
+    downloaded = []
+    errors = defaultdict(list)
+    with fou.ProgressBar(total=max_videos, iters_str="videos") as pb:
+        with multiprocessing.dummy.Pool(num_workers) as pool:
```
brimoor, line 210: `multiprocessing.dummy.Pool` uses threads, not processes.

Ideally, full video downloads would be done using a thread pool, since the bottleneck is network bandwidth, while clip downloads would be done using a process pool, since here the bottleneck is some combination of network and compute (`ffmpeg`) bandwidth.

fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+            (None, 8.3),
+        ]
+
+    Args:
+        urls: either a list of video urls, or a dict
+            mapping these urls to locations on disk. If `urls` is a list, then
+            the `download_dir` argument is required
+        download_dir (None): the output directory used to store the downloaded
+            videos. This is only used if `urls` is a list
+        clip_segments (None): a list of int or float tuples indicating start
```
brimoor, line 77: Things will be undefined when `urls` is a dict but `clip_segments` is a list, right?

fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+                _url = info_dict.get("url", None)
+                if _url is None:
+                    _url = info_dict["requested_formats"][0]["url"]
+
+                _id = info_dict.get("id", None)
+                _title = info_dict.get("title", _id)
+                _ext = info_dict.get("ext", ext)
+
+                output_path = output_path.replace(
+                    "%(title)s", _title.replace("/", "")
```
brimoor, line 280: A clean way to do this is to construct strings like `/path/for/${title}${ext}` if placeholders are needed when building the tasks. Then here you can just do:

```
output_path = output_path.format(title=title, ext=ext)
```

This will fill the patterns if they exist but not do anything if they don't exist (manually provided user path).

fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+
+    if len(clip_segments) != num_videos:
+        raise ValueError(
+            "Found %d `clip_segments` and %d `urls`, but the lengths of these "
+            "iterables must match" % (len(clip_segments), num_videos)
+        )
+
+    ext_list = _parse_list_arg(ext, num_videos)
+    urls_list = _parse_list_arg(list(urls.keys()), num_videos)
+    paths_list = _parse_list_arg(list(urls.values()), num_videos)
```
brimoor, line 159: It would be cleaner to render your `$(title)$(ext)` patterns here so that the tasks are just `(url, path, [segment])`.

fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+        clip_segments = [clip_segments]
+
+    if isinstance(urls, list):
+        if download_dir is None:
+            raise ValueError(
+                "When `urls` is a list, `download_dir` is required but was found to be `None`."
+            )
+        urls = {url: None for url in urls}
+
+        etau.ensure_dir(download_dir)
```
brimoor, line 144: Rather than `ensure_dir()` here, better to ensure the base directory of every output path (they could be arbitrary if the user provided them) at actual download time.

fiftyone/utils/youtubedl.py
brimoor: Processes could *always* be used on Linux/Mac, but the trouble is that they are sketchy on Windows.

So, the ideal compromise might be preferable to offer this functionality as separate methods:
- `download_youtube_videos()` --- thread pool
- `download_youtube_clips()` --- process pool

brimoor (changes requested):

fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+        return True, url, None
+
+    except Exception as e:
+        if isinstance(e, youtube_dl.utils.DownloadError):
+            # pylint: disable=no-member
+            return False, url, str(e.exc_info[1])
+        return False, url, "other"
+
+
+def _download_ffmpeg(url, output_path, clip_segment, buffer_seconds=5):
```
brimoor, line 300: This is nearly a re-implementation of what you can achieve with:

```py

fouv.extract_clip(url, output_path, timestamps=clip_segment)
```

It'd be nice to avoid code duplication. If that other method needs some additional parameters to deal with accuracy or other issues, then please add them there :)

fiftyone/utils/youtubedl.py
brimoor: I guess the best solution here is to have `urls`, `video_paths`, and `clip_segments` arguments that are all lists. 

This is specifically for clip downloads in a `download_youtube_clips()` method. For full video downloads in a `download_youtube_videos()` method, you could still allow `urls` to be a dict if you want.

fiftyone/utils/youtubedl.py
brimoor: Currently the task creation logic is partly here but partly in the download method, requiring the reader to fully understand both methods before they can make sense of what either method is doing

fiftyone/utils/youtubedl.py
```diff
@@ -0,0 +1,343 @@
...
+    errors = defaultdict(list)
+    with fou.ProgressBar(total=max_videos, iters_str="videos") as pb:
+        with multiprocessing.dummy.Pool(num_workers) as pool:
+            for is_success, url, error_type in pool.imap_unordered(
+                _do_download, tasks
+            ):
+                if is_success:
+                    if len(downloaded) < max_videos:
+                        downloaded.append(url)
+                        pb.update()
```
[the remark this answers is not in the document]
brimoor: ^you do it this way in the single-threaded implementation

fiftyone/utils/youtubedl.py
brimoor: ^example of task creation logic bleeding into the download implementation. Standardize as much as possible in `_build_tasks_list()` for simplicitly.

fiftyone/utils/youtubedl.py
brimoor: It would also be more useful if the errors dict also mapped `urls -> error messages`. The input is basically a list of URLs, so the most likely actions that will be taken based on the output is to check each `url` to see what happened to it.

fiftyone/utils/youtubedl.py
ehofesmann: It turned out to be pretty clean to switch between threads and processes without using two functions.

fiftyone/utils/youtubedl.py
ehofesmann: A dict for URLs doesn't really work since if you are downloading clips, you may have multiple local paths for one URL. I like the idea of returning url/local path tuples though.

brimoor: A list of tuples wouldn't quite work either in the sense that if you provided a list of clip segments to download that includes multiple clips from the same video and only one of the clips was actually downloaded because of a `max_videos` parameter, you wouldn't be able to deduce from the return values which of the clips was actually downloaded.

How about instead returning dicts that map integer index into the provided lists of `urls` and `clip_segments`. Both for successful downloads and errors. This would allow the caller to figure out what was attempted and succeeded, what was attempted and failed, and what was not attempted at all.

brimoor (approved):
LGTM.

@ehofesmann Note that I added a few more features. YouTube videos are a bit complex with the different stream types and resolutions available, and I felt this method would be a bit more powerful with some more features.

My basic design principle in choosing what/how to expose certain options is that this method is intended to efficiently download a potentially large number of videos. Therefore, whenever possible, the provided options are treated as recommendations, not requirements.

If the user wants very fine-grained, mandatory control of what they're downloading. They should just directly use `pytube`.

Also, please note that I changed the return format slightly. You may need to tweak how zoo datasets use this utility.

fiftyone/utils/youtube.py
```diff
@@ -0,0 +1,461 @@
...
+
+logger = multiprocessing.get_logger()
+
+
+def download_youtube_videos(
+    urls,
+    download_dir=None,
+    video_paths=None,
+    clip_segments=None,
+    ext=".mp4",
```
brimoor, line 33: I think `ext` makes more sense as a single *hint* rather than a list of strict requirements, because in general the user would have no way to know what codecs might be available for a particular watch URL

fiftyone/utils/youtube.py
```diff
@@ -0,0 +1,461 @@
...
+    only_progressive=True,
+    resolution="highest",
```
brimoor, line 35: I read the pytube docs and found that `ext=".mp4", only_progressive=True, resolution="highest"` should generally result in a 720p video being downloaded. This makes sense as a default behavior I think.

Certain videos from popular creators might have crazy 4k streams available (stored as non-progressive, non-MP4 streams, per my research), which are generally not useful for ML use cases. However, the user can still pass `ext=None, only_progressive=False` to get their hands on the absolute highest quality stream.

fiftyone/utils/youtube.py
```diff
@@ -0,0 +1,461 @@
...
+            downloading videos. By default, ``multiprocessing.cpu_count()`` is
+            used
+        skip_failures (True): whether to gracefully continue without raising
+            an error if a video cannot be downloaded
+        quiet (False): whether to suppress logging, except for a progress bar
+
+    Returns:
+        a tuple of
+
+        -   **downloaded**: a dict mapping integer indexes into ``urls`` to
```
brimoor, line 115: I think a return dict structure like this is, in general, the only way to succinctly describe what was downloaded and what failed in a situation where `len(urls) > max_videos` and certain videos are not available in the requested `ext` and thus the user needs a way to deduce the names of the returned videos.

fiftyone/utils/youtube.py
brimoor: FYI- an `ext` argument is used quite a bit throughout the codebase, and in those places it always includes the `.`. The original reasoning was for consistency with what you get  `os.path.splitext()`, but at this point we should continue with the convention regardless just for API consistency.

PR #1525, opened 2022-01-12
Adding support for defining custom EmbeddedDocument classes

brimoor:
Adds support for defining custom `EmbeddedDocument` and `DynamicEmbeddedDocument` classes, which can then be used to populate arbitrary top-level sample fields or nested fields.

A new `module_path` config setting is added where you can register the modules in which your custom embedded document classes are defined. Any modules in this list will automatically be imported whenever FiftyOne is imported so that datasets with your custom fields can be loaded.

## Example usage

First, define some custom embedded document classes in a `foo.bar` module:

```py

class Test(foo.EmbeddedDocument):
    field = fof.IntField()

class DynamicTest(foo.DynamicEmbeddedDocument):
    field = fof.IntField()
```

Next, add `foo.bar` to FiftyOne's `module_path` config setting so that the module will always be imported whenever you import FiftyOne: 

```shell
export FIFTYONE_MODULE_PATH=foo.bar

# Verify module path
fiftyone config
```

You're now free to use the `Test` and `DynamicTest` classes as you please, whether this be top-level sample fields or nested fields:

```py
# foo/bar.py
import fiftyone as fo
import foo.bar

sample = fo.Sample(
    test=Test(field=1),
    dynamic_test=DynamicTest(field=2, foo="bar"),
    dict={
        "test": Test(field=3),
        "dynamic_test": DynamicTest(field=4, spam="eggs"),
    },
)

dataset = fo.Dataset()
dataset.add_sample(sample)

dataset.name = "test"
dataset.persistent = True
```

The dataset can be loaded in future sessions with no additional work:

```py
import fiftyone as fo

dataset = fo.load_dataset("test")
print(dataset.first())
```

```
<Sample: {
    'id': '61de37ae47f305672d5040d6',
    'media_type': 'image',
    'filepath': '/path/to/image.png',
    'tags': BaseList([]),
    'metadata': None,
    'test': <Test: {'field': 1}>,
    'dynamic_test': <DynamicTest: {'field': 2, 'foo': 'bar'}>,
    'dict': BaseDict({
        'test': <Test: {'field': 3}>,
        'dynamic_test': <DynamicTest: {'field': 4, 'spam': 'eggs'}>,
    }),
}>
```

### DIFF
--- a/docs/source/user_guide/config.rst
+++ b/docs/source/user_guide/config.rst
@@ -76,6 +76,8 @@ FiftyOne supports the configuration options described below:
 | `model_zoo_manifest_paths`    | `FIFTYONE_MODEL_ZOO_MANIFEST_PATHS` | `None`                        | A list of manifest JSON files specifying additional zoo models. See                    |
 |                               |                                     |                               | :ref:`adding models to the zoo <model-zoo-add>` for more information.                  |
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
+| `module_path`                 | `FIFTYONE_MODULE_PATH`              | `None`                        | A list of modules that should be automatically imported whenever FiftyOne is imported. |
++-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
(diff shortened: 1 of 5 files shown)


benjaminpkane (approved):
LGTM

PR #1528, opened 2022-01-12
Add hollow support for 32 bit systems

benjaminpkane:
Adds `fiftyone-db` wheels for i686 linux systems and win 32 systems.


brimoor (approved):
LGTM

PR #1530, opened 2022-01-12
Publish source distributions

benjaminpkane:
Adds a source distribution for the published package, which can installed with:

```
pip install fiftyone --no-binary fiftyone
```

in which case, the opencv dependency is dynamically configured based on the environment, .e.g. Google's Colab.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,7 +50,7 @@ jobs:
           fi
       - name: Build dist
         run: |
-          python setup.py bdist_wheel
+          python setup.py bdist_wheel sdist
       - name: Upload dist
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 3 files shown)


brimoor (approved):
LGTM.

I think there's a typo in the PR description. Just

```
pip install --no-binary fiftyone
```

should do, right?

benjaminpkane:
--no-binary` takes an argument.

```
  --no-binary <format_control>
                              Do not use binary packages. Can be supplied multiple times, and each time adds to the existing value. Accepts either ":all:" to disable all binary packages, ":none:" to
                              empty the set (notice the colons), or one or more package names with commas between them (no colons). Note that some packages are tricky to compile and may fail to install
                              when this option is used on them.
```



`pip install --no-binary fiftyone` is the same as `pip install`, i.e. no requirement is requested

brimoor:
Ah I see, carry on 📈

PR #1547, opened 2022-01-19
Resolve bug when annotating existing fields with multiple label types

ehofesmann:
When annotating an existing label field, there is logic in place to automatically detect the type of the label field being annotated. Some label fields can be associated with multiple annotation label types (detections can be bounding boxes or segmentation instances, polylines can be polylines or polygons). 

There is a bug when annotating an existing label field but all samples in the view have `None` values for the field. This PR resolves this bug by simplifying the logic to detect existing label types.

## Example

```python
import fiftyone as fo
import fiftyone.zoo as foz

d1 = foz.load_zoo_dataset("quickstart", max_samples=4).clone()
dataset = d1[:3].clone()

anno_key="anno_key"
dataset.annotate(
    anno_key, 
    label_field="segs", 
    label_type="polylines",
    classes=["a", "b"], 
    launch_editor=True,
)

# Annotate a polygon in CVAT

dataset.add_sample(fo.Sample(filepath=d1.last().filepath))
view = dataset.exists("segs", False)

anno_key2 = "anno_key2"
view.annotate(
    anno_key2, 
    label_field="segs", 
    label_type="polylines",
    classes=["a", "b"], 
    launch_editor=True,
)

# ERROR
```

### DIFF
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -693,55 +693,17 @@ def _get_existing_label_type(samples, backend, label_field, field_type):
 
     fo_label_type = field_type.document_type
 
-    if issubclass(fo_label_type, (fol.Detection, fol.Detections)):
-        # Must check for detections vs instances
-        _, label_path = samples._get_label_field_path(label_field)
-        _, mask_path = samples._get_label_field_path(label_field, "mask")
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM

Issue #1545, opened 2022-01-19
[FR] Display similarity score in the App


Rusteam:
### Proposal Summary

Show similarity score the App when sorted by similarity.
Possible, add a view for similarity search.

### Motivation

-   What is the use case for this feature?

> When sorting images by similarity we are only able to see the order of images but not their scores.
That's quite an important step in exploring a feature extractor/embeddings.

-   Why is this use case valuable to support for FiftyOne users in general?

> Similarity search is a usual case when we're dealing with large number of classes 
and few images per class, or even without class labels. Now, users have to switch another
tool like python plotting libs or custom frontend to visualize similarities/matches and scores.
If fiftyone could have sim.score displayed in the least, or a dedicated view 
for similarity search that would make it very useful for this task.

-   Why is this use case valuable to support for your project(s) or
    organization?

> We deal with cases where there's large number of classes (few thousand) and 
few images per class (1-5).

-   Why is it currently difficult to achieve this use case? (please be as
    specific as possible about why related FiftyOne features and components are
    insufficient)

> `sortBySimiliarty` does not display a score and it's hard to tell if there's any difference
between rank 1 or rank 2 or rank 3.

### What areas of FiftyOne does this feature affect?

-   [x] App: FiftyOne application
-   [ ] Core: Core `fiftyone` Python library
-   [x] Server: FiftyOne server

### Details

Need guidance.

### Willingness to contribute

The FiftyOne Community encourages new feature contributions. Would you or
another member of your organization be willing to contribute an implementation
of this feature?

-   [ ] Yes. I can contribute this feature independently.
-   [x] Yes. I would be willing to contribute this feature with guidance from
        the FiftyOne community.
-   [ ] No. I cannot contribute this feature at this time.

brimoor:
Hey @Rusteam, good idea 💪 I implemented this in https://github.com/voxel51/fiftyone/pull/1548 and it will be included in the next release.

In the meantime, check out [this example](https://voxel51.com/docs/fiftyone/user_guide/brain.html#cifar-10-example) for some ways that you can access embedding distances in the context of unique/duplicate detection.

Rusteam:
> In the meantime, check out [this example](https://voxel51.com/docs/fiftyone/user_guide/brain.html#cifar-10-example) for some ways that you can access embedding distances in the context of unique/duplicate detection.

Indeed that's a clever workaround. Thanks

PR #1548, opened 2022-01-20
Adding support for populating query distances when sorting by similarity 

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1545.

Adds an optional `dist_field` parameter that can specify the name of a (new or existing) float field in which to store query distances when [sorting by similarity](https://voxel51.com/docs/fiftyone/user_guide/using_views.html#visual-similarity):

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

fob.compute_similarity(dataset, brain_key="img_sim")

view = dataset.sort_by_similarity(dataset.shuffle().first().id, dist_field="dist")

session = fo.launch_app(view)
```

This feature can be used when [sorting by similarity in the App](https://voxel51.com/docs/fiftyone/user_guide/app.html#sorting-by-visual-similarity), but currently the `dist_field` must be added in the view bar after initially creating the view:

sim-opt

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -119,8 +119,8 @@ jobs:
       - name: Install fiftyone
         run: |
           pip install -r requirements/extras.txt
-          pip install --pre --extra-index-url https://test.pypi.org/simple/ fiftyone-brain voxel51-eta
           pip install -e .
+          pip install --pre --extra-index-url https://test.pypi.org/simple/ fiftyone-brain voxel51-eta
       - name: Install ETA from source
         if: ${{ !startsWith(github.ref, 'refs/heads/rel') && !startsWith(github.ref, 'refs/tags/') }}
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,7 +120,7 @@ jobs:
         run: |
           pip install -r requirements/extras.txt
           pip install -e .
-          pip install --pre --extra-index-url https://test.pypi.org/simple/ fiftyone-brain voxel51-eta
+          pip install --update --pre --extra-index-url https://test.pypi.org/simple/ fiftyone-brain voxel51-eta
       - name: Install ETA from source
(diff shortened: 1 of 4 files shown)


brimoor:
@benjaminpkane can you add `dist_field` as an optional field in the similarity menu?

benjaminpkane:
LGTM. I will make App updates in #1369

benjaminpkane:
It's a bit messy that you are changing the schema here. Working through it.

brimoor:
> It's a bit messy that you are changing the schema here. Working through it.

Yeah I believe a `dataset.reload()` is called for to cause the schema to update

brimoor:
If we've built things correctly, when the user then tries to access a sample from the main process, a `dataset.reload()` should automatically be triggered when they encounter data in an unexpected `dist_field`.

Maybe now is the time to get something similar working in the App process? That way if the user adds a new field from another process, the App will automatically reflect this upon the next refresh

benjaminpkane:
That is beyond the point. The issue is you are asking for more hidden (extended view) state, but this stage changes schema. That breaks the rules of the extended view. I'm working through it.

brimoor:
> That is beyond the point. The issue is you are asking for more hidden (extended view) state, but this stage changes schema. That breaks the rules of the extended view. I'm working through it.

I see. FWIW, we do not allow `DatasetView`s to declare new fields (they must be declared on the dataset itself), so the same can be true for the extended view: if `dist_field` is used, it always causes an irreversible field to be added to the dataset, and also the values are permanently written to the dataset.

I fully admit that this violates the notion of what a view stage should do, but I haven't found the motivation to allow `DatasetView` schemas to be "temporarily" expanded, primarily because I also haven't found a good way to implement attaching an arbitrary set of ephemeral field data to a view.

benjaminpkane:
Note that this also opaquely makes synchronous db calls which will block when used by the server

brimoor:
> Note that this also opaquely makes synchronous db calls which will block when used by the server

Yep but just a drop in the bucket compared to the linear algebra that it already does to compute nearest neighbors

PR #1564, opened 2022-01-25
Update Dockerfile example

ehofesmann:
[Installing Python through apt results in packages](https://stackoverflow.com/questions/9387928/whats-the-difference-between-dist-packages-and-site-packages) being stored in `dist-packages` instead of `site-packages`. This PR updates the Dockerfile example of automatically launching the App.


brimoor (approved):
LGTM

PR #1566, opened 2022-01-27
Handle empty Keypoint in App

### LINKED ISSUE #1562: [BUG] App can't render samples with empty Keypoint data
The App grid fails to render when a `Keypoint` with `points=[]` is encountered.

In this case, it should still render the image and other fields, just with no keypoint data

```py
import fiftyone as fo

sample = fo.Sample(
    filepath="/path/to/image.jpg",
    keypoints=fo.Keypoints(keypoints=[fo.Keypoint(label="test")]),
)

dataset = fo.Dataset()
dataset.add_sample(sample)

```

Screen Shot 2022-01-24 at 10 40 24 AM

benjaminpkane:
Resolves #1562

### DIFF
--- a/app/packages/looker/src/overlays/keypoint.ts
+++ b/app/packages/looker/src/overlays/keypoint.ts
@@ -20,6 +20,10 @@ export default class KeypointOverlay<
   }
 
   containsPoint(state: Readonly<State>): CONTAINS {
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM

PR #1572, opened 2022-01-28
Encode filepath as URI component

### LINKED ISSUE #1571: [BUG] App fails to load media filepaths that need encoding
Example filename that will cause a 404 in the App: `PreDiff_Normal_A-00.14.07.343-00.14.34.155.mp4#t=0.2.jpg`
error

benjaminpkane:
Resolves #1571


brimoor (approved):
LGTM

PR #1573, opened 2022-01-28
View save

brimoor:
Previously, `DatasetView.save()` would both save changes to the samples in the view *and* delete any samples that were not in the view.

This PR splits that behavior into two separate methods:
- `save()`: save changes to (only) the samples in the view to the underlying dataset
- `keep()`: delete any samples that are not in the view to the underlying dataset

This PR also adds a `keep_frames()` method that provides a slight variation on the above functionality that is relevant for video datasets: deleting omitted frames from a video view **only** for the samples in the view.

### Updated Documentation

Ordinarily, when you define a `DatasetView` that extracts a specific subset of a dataset and its fields, the underlying `Dataset` is not modified. However, you can use `save()` to save the contents of a view you've created to the underlying dataset:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Capitalize some labels
rename_view = dataset.map_labels("predictions", {"cat": "CAT", "dog": "DOG"})
rename_view.save()

print(dataset.count_values("predictions.detections.label"))
# {'CAT': 35, 'DOG': 49, ...}

# Discard all predictions with confidence below 0.3
high_conf_view = dataset.filter_labels(
    "predictions", F("confidence") > 0.3, only_matches=False
)
high_conf_view.save()

print(dataset.bounds("predictions.detections.confidence"))
# (0.3001, 0.9999)
```

Note that calling `save()` on a `DatasetView` will only save modifications to samples that are in the view; all other samples are left unchanged.

You can use `keep()` to delete samples from the underlying dataset that do not appear in a view you created:

```py
print(len(dataset))
# 200

# Discard all samples with no people
people_view = dataset.filter_labels("ground_truth", F("label") == "person")
people_view.keep()

print(len(dataset))
# 94
```

### DIFF
--- a/docs/source/user_guide/using_views.rst
+++ b/docs/source/user_guide/using_views.rst
@@ -713,10 +713,11 @@ in the sense that:
     by iterating over the contents of the view or calling
     :meth:`set_values() <fiftyone.core.collections.SampleCollection.set_values>`
     will be reflected on the source dataset
--   Calling :meth:`save() <fiftyone.core.patches.PatchesView.save>` on an
-    object patches view (typically one that contains additional view stages
-    that filter or modify its contents) will sync any |Label| edits or
-    deletions with the source dataset
+-   Calling :meth:`save() <fiftyone.core.patches.PatchesView.save>` or
+    :meth:`keep() <fiftyone.core.patches.PatchesView.keep>` on an object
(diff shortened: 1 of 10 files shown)


benjaminpkane (approved):
LGTM

PR #1575, opened 2022-02-01
Fix builds

benjaminpkane:
A test was added that Windows fails on. Punting. Also fixes builds so RC candidates are installed.


brimoor (approved):
LGTM. Before merging, can you check if `@skipwindows` can be used in place of these:

```
@unittest.skip("TODO: Fix workflow errors. Must be run manually")
```

that exist in some of the unit tests?

benjaminpkane:
For another day

PR #1580, opened 2022-02-03
Apply patch to YouTube download utilities

ehofesmann:
There was a recent update from YouTube that is causing issues with video downloading utilities like Pytube used in FiftyOne. This PR adds a patch to work around this issue for the current version of Pytube.

```python
import fiftyone.utils.youtube as fouy

urls = ["https://www.youtube.com/watch?v=---9CpRcKoU"]

fouy.download_youtube_videos(urls, download_dir="/tmp/youtube_test")
```

Previously would raise error:
```
Failed to download video 'https://www.youtube.com/watch?v=---9CpRcKoU': 'NoneType' object has no attribute 'span'
```


brimoor (approved):
LGTM

Issue #1563, opened 2022-01-24
[FR] Add support for storing and rendering skeletons for keypoint fields


brimoor:
When working with keypoint models, one often has an associated skeleton defining connectivity between the keypoints that should be rendered as edges (see screenshot below).

Currently, to achieve this, one must store two copies of the data: one in a `Keypoints` field and another in a `Polylines` field. The issues with this approach are:
- Data is duplicated
- Performing filtering operations (eg to see only specific objects) is cumbersome, because both fields must be individually filtered.

Instead, it would be ideal to have `default_skeleton` and `skeleton` fields analogous to `default_classes/classes` and `default_mask_targets/mask_targets` that allow users to attach skeletons to their dataset and/or specific fields, which would then be used by the App to dynamically render connections between points. The simplest data structure for this would be:

```
skeleton = {
    "labels": ["nose", "left eye", "right eye", ...],
    "edges": [[0, 1], [1, 2], ...],
}
```

which assigns semantic labels to each point and then draws edges `points[0] -> points[1]`, `points[1] -> points[2]`, etc.

Although, perhaps we may want to also allow for the possibility of per-`Keypoint.label` skeleton definitions?

In the example code below, [KeypointDetectorOutputProcessor](https://github.com/voxel51/fiftyone/blob/50c8fc08a57b049cdbb34406249c86f7249ceac1/fiftyone/utils/torch.py#L802) manually performs the `Polylines` population so that the desired visualization is achieved (albeit clunky for the reasons described above):

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("coco-2017", split="validation")

view = dataset.take(20)

model = foz.load_zoo_model("keypoint-rcnn-resnet50-fpn-coco-torch")
view.apply_model(model, label_field="predictions")

session = fo.launch_app(view)
```

<img width="1151" alt="Screen Shot 2022-01-24 at 11 31 49 AM" src="https://user-images.githubusercontent.com/25985824/150824031-8f301790-e156-4ec0-a356-2f8c691e406a.png">

benjaminpkane:
Stale. Resolved

PR #1583, opened 2022-02-04
Add support for per-point confidences in Keypoint class

### LINKED ISSUE #1581: [FR] Add support for per-point confidence/visibility
### Proposal Summary

Add support for per-point confidence/visibility - currently Keypoint label class (which represents a set of points associated with an instance) is a flat list of (x, y) which doesn't provide additional per-point information. In addition, Coco import skips over the point which visibility = 0 (not visible/annotated) which can break the implicit mapping between joint -> point since num_joints != num_points and there's no information on which joint was skipped.

### Motivation

This is needed to correctly import COCO keypoint dataset as well as filter sample based on per-point confidence...etc.

Anyone using keypoint detection would probably need this since most of models outputs per-point confidence and there needs to be a way to distinguish in ground truth whether the point is visible or not + mapping between point to joint.

    organization?
My project needs keypoint dataset - was hoping to leverage existing COCO dataset however the import breaks the mapping so any partially annotated keypoints (e.g. missing things) are useless.

    insufficient)
One way to workaround would be to only use fully annotated samples (no v=0), which would probably throw out majority of keypoint annotations.

### Details

(Use this section to include any additional information about the feature. If
you have a proposal for how to implement this feature, please include it here.)

We probably need to fix COCOObject::to_keypoints to not skip over v == 0 case and encode this visibility information to Keypoint by

Option 1. Expand KeypointsField to include confidence/visibility (not sure how it would support "optional-ness")
Option 2. Add a separate field "confidences"/"visibilities" in Keypoint class which maps to "points"

### Willingness to contribute

of this feature?

        the FiftyOne community.

I would love to contribute, however I'm not a Python expert + new to FiftyOne so would probably need some guidance.

stbdang:
Fixes #1581 

Replacing Keypoint.confidence to Keypoint.confidences
Fixing COCO import to not skip over v=0, instead set visible to false
Populating optional visible field in coco.py/torch.py

Tested locally on COCO-2017 with keypoint import. This is the sample output from COCO import

```
<Keypoint: {
    'id': '61fc66d8c50045c0978ff8ed',
    'attributes': BaseDict({}),
    'tags': BaseList([]),
    'label': 'person',
    'points': BaseList([
        BaseList([0.585480093676815, 0.3328125]),
        BaseList([0.629976580796253, 0.2625]),
        BaseList([0.4894613583138173, 0.271875]),
        BaseList([nan, nan]),
        BaseList([0.26697892271662765, 0.2796875]),
        BaseList([0.9391100702576113, 0.5359375]),
        BaseList([0.09836065573770492, 0.6109375]),
        BaseList([nan, nan]),
        BaseList([nan, nan]),
        BaseList([0.8758782201405152, 0.596875]),
        BaseList([0.48243559718969553, 0.6734375]),
        BaseList([nan, nan]),
        BaseList([nan, nan]),
        BaseList([nan, nan]),
        BaseList([nan, nan]),
        BaseList([nan, nan]),
        BaseList([nan, nan]),
    ]),
    'confidences': None,
    'index': None,
    'visible':
    BaseList([
        True, True, True, False, True, True, True, False, False, True, True, False, False, False, False, False, False,
    ]),
    'num_keypoints': 8,
}>
```

confidence->confidences might be something to mention in the release notes.

        notes for FiftyOne users.

Keypoint.confidence was replaced with Keypoint.confidences which represent confidence per each point.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/eta.py
+++ b/fiftyone/utils/eta.py
@@ -753,7 +753,6 @@ def from_keypoint(keypoints):
     return fol.Keypoint(
         label=keypoints.label,
         points=keypoints.points,
-        confidence=keypoints.confidence,
         index=keypoints.index,
         tags=keypoints.tags,
         **attributes,
--- a/fiftyone/utils/eta.py
+++ b/fiftyone/utils/eta.py
@@ -730,7 +730,6 @@ def to_keypoints(keypoint, name=None, extra_attrs=True):
     return etak.Keypoints(
         name=name,
         label=keypoint.label,
-        confidence=keypoint.confidence,
         index=keypoint.index,
         points=keypoint.points,
         attrs=attrs,
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -862,6 +862,7 @@ class KeypointDetectorOutputProcessor(OutputProcessor):
             ]
 
             points = [(p[0] / width, p[1] / height) for p in kpts]
+            visible = [(p[3] > 0) for p in kpts]
 
             _detections.append(
                 fol.Detection(
@@ -875,7 +876,7 @@ class KeypointDetectorOutputProcessor(OutputProcessor):
                 fol.Keypoint(
                     label=self.class_labels[label],
                     points=points,
-                    confidence=score,
+                    visible=visible,
                 )
             )
 
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -845,11 +845,12 @@ class KeypointDetectorOutputProcessor(OutputProcessor):
         labels = output["labels"].detach().cpu().numpy()
         scores = output["scores"].detach().cpu().numpy()
         keypoints = output["keypoints"].detach().cpu().numpy()
+        keypoints_scores = torch.sigmoid(output["keypoints_scores"].detach().cpu()).numpy()
 
         _detections = []
         _keypoints = []
         _polylines = []
-        for box, label, score, kpts in zip(boxes, labels, scores, keypoints):
+        for box, label, score, kpts, kpt_scores in zip(boxes, labels, scores, keypoints, keypoints_scores):
             if confidence_thresh is not None and score < confidence_thresh:
                 continue
 
@@ -862,7 +863,7 @@ class KeypointDetectorOutputProcessor(OutputProcessor):
             ]
 
             points = [(p[0] / width, p[1] / height) for p in kpts]
-            visible = [(p[3] > 0) for p in kpts]
+            visible = [bool(p[2] > 0) for p in kpts]
 
             _detections.append(
                 fol.Detection(
@@ -877,6 +878,7 @@ class KeypointDetectorOutputProcessor(OutputProcessor):
                     label=self.class_labels[label],
                     points=points,
                     visible=visible,
+                    confidences=kpt_scores.tolist(),
                 )
             )
 
--- a/fiftyone/core/labels.py
+++ b/fiftyone/core/labels.py
@@ -851,7 +851,7 @@ class Keypoint(_HasID, _HasAttributesDict, Label):
     Args:
         label (None): a label for the points
         points (None): a list of ``(x, y)`` keypoints in ``[0, 1] x [0, 1]``
-        confidence (None): a confidence in ``[0, 1]`` for the points
+        confidences (None): a list of confidence in ``[0, 1]`` for each point
         index (None): an index for the keypoints
         attributes ({}): a dict mapping attribute names to :class:`Attribute`
             instances
@@ -859,7 +859,7 @@ class Keypoint(_HasID, _HasAttributesDict, Label):
 
     label = fof.StringField()
     points = fof.KeypointsField()
-    confidence = fof.FloatField()
+    confidences = fof.ListField(field=fof.FloatField(), null=True)
     index = fof.IntField()
 
     def to_shapely(self, frame_size=None):
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -973,17 +973,17 @@ class COCOObject(object):
         label = self._get_label(classes)
 
         points = []
+        visible = []
         for x, y, v in fou.iter_batches(self.keypoints, 3):
             if v == 0:
-                continue
-
-            points.append((x / width, y / height))
+                points.append((float("nan"), float("nan")))
+                visible.append(False)
+            else:
+                points.append((x / width, y / height))
+                visible.append(True)
 
         return fol.Keypoint(
-            label=label,
-            points=points,
-            confidence=self.score,
-            **self.attributes,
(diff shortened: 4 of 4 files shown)


fiftyone/utils/eta.py
```diff
@@ -753,7 +753,6 @@ def from_keypoint(keypoints):
     return fol.Keypoint(
         label=keypoints.label,
         points=keypoints.points,
-        confidence=keypoints.confidence,
```
brimoor, line 756: There's also a `to_keypoints()` method above where you'll need to remove usage of `keypoint.confidence`

fiftyone/utils/torch.py
```diff
@@ -875,7 +876,7 @@ def _parse_output(self, output, frame_size, confidence_thresh):
                 fol.Keypoint(
                     label=self.class_labels[label],
                     points=points,
-                    confidence=score,
+                    visible=visible,
```
brimoor, line 879: I think you mentioned that models like `keypoint-rcnn-resnet50-fpn-coco-torch` expose keypoint confidences too? Can you parse and store those here?

Here's some example code you can use to test the model:

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "coco-2017",
    split="validation",
    classes="person",
    max_samples=10,
    shuffle=True,
)

```

brimoor (commented):
Thanks for this! Left a couple small comments for your consideration but overall this looks good!

@benjaminpkane can you update the App to render per-point `confidences` for `Keypoint` instead of the now-deprecated `confidence` for every point?

fiftyone/utils/eta.py
stbdang: Oops yeah, let me do that

fiftyone/utils/torch.py
stbdang: I actually implemented it earlier and decided to not include it in this PR, because it seemed per-point confidences need further processing (sigmoid?). The output values looked like this.

` 'keypoints_scores': tensor([[16.0222, 18.2254, 17.0186, 11.8371, 13.9270,  6.0463,  7.8868,  7.1934,
          6.9537,  7.8964, 11.9600, -0.3767, -2.9283, -2.6453, -1.6882, -2.4117,
         -3.3753],
        [ 3.4408,  7.4437,  3.9761,  6.6165,  4.9816, -0.2632, -1.3853, -1.6286,
         -1.7510,  1.1719, -1.5498, -1.6644, -1.3459,  0.8909, -2.4248,  0.2460,
         -2.5256],
        [ 1.0068,  1.3324,  1.4261,  3.2719,  2.5410, -0.2752, -0.0260,  1.1362,
          1.7262,  1.0567, -0.1256, -1.9884, -2.3250, -2.7186, -1.9209, -2.9283,
         -2.7841]`
         
         I could either relax the description of confidences to be arbitrary float value (which changes the meaning of the field) and pass this value as is or apply sigmoid-like function to the output. However, since this output isn't really documented on torchvision side, I decided to not populate it for now. (https://pytorch.org/vision/stable/_modules/torchvision/models/detection/keypoint_rcnn.html#keypointrcnn_resnet50_fpn)
    
If you have insight into the expected value of this "keypoint_scores", let me know!

fiftyone/utils/eta.py
stbdang: Addressed in the following commit.

fiftyone/utils/torch.py
brimoor: I think your intuition apply sigmoid is likely to be correct. How about wiring that up and then checking if `sigmoid(keypoints_scores)` make sense with some of the model outputs?

That'll be easy once the App supports renderings `confidences`, but in the meantime you can probably figure it out by printing some keypoints and eyeballing in the App

stbdang: Added per-point confidences, was debating whether to apply confidence_thresh to each point or leave it to the user since this probably means visible flag needs to be toggled off if it's lower than threshold (the model outputs all 1's for its keypoint prediction). I left it intact for now.

stbdang: I think it might be useful for the app to handle filtering (so user can adjust threshold and see the visualization change dynamically) rather than having a static threshold value.

benjaminpkane: I'm not on board with this approach. If we want App filtering per point, then points should be treated as real entity with `id`s. Parallel lists is not a safe way to store data.

brimoor: I agree that we need to consider filtering support carefully.

One important point here is that a `Keypoint` object definitively represents a semantic grouping of points. In that sense, if we *do* support filtering individual keypoints within an object, it is a different sort of operation than filtering other things like bounding boxes. An example of this is the upcoming [keypoint skeleton](https://github.com/voxel51/fiftyone/issues/1563) that we'll be adding. If a user is allowed to filter points in a skeleton by confidence, the `Keypoint` representation will have to be implemented such that we can always associate individual points with their skeleton vertex. This effectively means each point needs an "index", which suggests a list is indeed a reasonable data structure.

I'm not in favor of assigning `ObjectId`s to individual points within a `Keypoint`, for a number of reasons, including:
- Introduces an additional level of ID nesting not present for other label types
- Sets a precedent that we should add IDs to bounding box coordinates and polyline vertexes
- Would necessitate non-trivial breaking changes to the data model

My current proposal would be that filtering `Keypoint.points` by confidence can be implemented by either:
- Replacing low confidence points by `NaN`, since there is already a precedent that `NaN`-valued points are allowed but cause the point to not be rendered
- Replacing the corresponding `visible[i]` with `False`
 
With respect to breaking data model changes: this project is sufficiently popular at this point(!) that refactoring `Keypoint.points` into something other than a list of lists gives me pause, since it may break existing user code. @stbdang made the shrewd observation that models like `keypoint-rcnn-resnet50-fpn-coco-torch` do not produce any notion of `Keypoint.confidence`, only individual point confidences. So, I'm willing to move from `Keypoint.confidence` to `Keypoint.confidences` under the assumption that there are not likely to be important workflows that rely on filtering by `Keypoint.confidence`

brimoor: Hey @stbdang really appreciate your work on this. There's two remaining blockers here:

1. We need to make an App tweak one way or the other to reflect the fact that `confidence` is no longer a field on `Keypoint` classes. Specifically, we either need to remove the `confidence` filter altogether for `Keypoint`-type fields, or we need to keep the filter but update the filtering logic per my comments above so that it will insert `NaN` values to achieve the desired effect of "hiding" those keypoints
2. I'm on the fence on whether to introduce the convention of populating a `visible` field for keypoints right now, because this isn't implemented by other label types like `Detection` and `Polyline`

brimoor (approved):
LGTM. I changed the base branch to a new `keypoints` feature branch where we can handle the remaining App updates to get this ready for merging into `develop`.

Thanks again for your contribution!

PR #1586, opened 2022-02-04
Fixing frame aggregations

brimoor:
Resolves an aggregation bug that would previously cause an error when writing aggregations on video datasets that involve applying expressions directly to `F("frames")`.

Examples:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.compute_metadata()

# Previously failed, now works
num_objects = F("frames").map(F("detections.detections").length()).max()
print(dataset.count_values(num_objects))

# Previously failed, now works
clips = dataset.to_clips(F("detections.detections").length() > 5)
duration = F("frames").length() / F("$metadata.frame_rate")
print(clips.bounds(duration))
```

### DIFF
--- a/fiftyone/core/aggregations.py
+++ b/fiftyone/core/aggregations.py
@@ -1704,6 +1704,7 @@ def _parse_field_and_expr(
     if field_name is None:
         field_name, expr = _extract_prefix_from_expr(expr)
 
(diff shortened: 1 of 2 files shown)


ehofesmann (approved):
LGTM

PR #1601, opened 2022-02-16
Keypoint updates

### LINKED ISSUE #1581: [FR] Add support for per-point confidence/visibility

brimoor:
Resolves #1581

Requires https://github.com/voxel51/eta/pull/556

## Change log

- [x] Implement keypoint skeletons as per #1563
- [x] Implement skeleton rendering in `draw_labels()`
- [x] Add a `filter_keypoints()` view stage that applies per-`point` filters to keypoint objects
- [x] Remove `visible` usage in favor of `NaN`-based filtering
- [x] Update `draw_labels()` code to gracefully handle NaN-valued keypoints
- [x] Documentation
- [x] Add keypoint skeleton (`label` and `edges`) rendering in the App
    - [x] Add an option to the App's settings menu to toggle whether skeletons are rendered
    - [x] Add skeleton labels and any `ListField` field values (whose length is `len(points)`) to the tooltip of individual points
- [x] Add support for custom per-point attributes on keypoints (occluded, etc)
- [x] Update App keypoint filtering logic:
    - [x] Add a `skeleton.label` filter that allows for filtering keypoints by their skeleton label
    - [x] Add a `confidence` filter that allows for filtering keypoints by `Keypoint.confidence`
- [x] ~~Convert `KeypointSkeleton` to just a dict?~~
- [x] ~~Add support for per-`label` skeletons in a field?~~

App implementation details
- The `Keypoint.points` filter logic should insert `NaN` values rather than removing the elements from the list (since the skeleton is defined based on list index position)
- Any `Keypoint.points` entry that contains a NaN coordinates should not be rendered in the App
- Any skeleton edges between one or more points with NaN coordinates should not be rendered
- The new `filter_keypoints()` view stage supports filtering by per-point confidences and per-point skeleton labels

## Example usage

```py
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("coco-2017", split="validation").clone()

dataset.default_skeleton = model.skeleton

view = (
    dataset
    .match_labels(fields="ground_truth", filter=F("label") == "person")
    .limit(2)
)

# Add some custom attributes
for sample in view:
    for keypoint in sample["predictions_keypoints"].keypoints:
        keypoint["occluded"] = [np.random.randn() > 0 for _ in range(len(keypoint.points))]
        keypoint["age"] = np.random.randint(51)

    sample.save()

print(view.bounds("predictions_keypoints.keypoints.confidence"))
print(dataset.default_skeleton)

session = fo.launch_app(view)

session.view = view.filter_keypoints(
    "predictions_keypoints",
    filter=F("confidence") > 0.99999,
)

session.view = view.filter_keypoints(
    "predictions_keypoints",
    labels=["left eye", "right eye"],
)

view.draw_labels(
    "/tmp/keypoints",
    label_fields="predictions_keypoints",
    overwrite=True,
)
```

### Keypoint skeleton construction

```py
skeleton = fo.KeypointSkeleton(
    labels=[
        'nose',
        'left eye',
        'right eye',
        'left ear',
        'right ear',
        'left shoulder',
        'right shoulder',
        'left elbow',
        'right elbow',
        'left wrist',
        'right wrist',
        'left hip',
        'right hip',
        'left knee',
        'right knee',
        'left ankle',
        'right ankle',
    ],
    edges=[
        [11, 5, 3, 1, 0, 2, 4, 6, 12],
        [9, 7, 5, 6, 8, 10],
        [15, 13, 11, 12, 14, 16],
    ]
)
```


brimoor:
@ehofesmann @benjaminpkane can you take a look at this PR and share any thoughts about the current keypoint interface?

I'm leaning towards *not* implementing the two outstanding backend TODO items in the PR description, but want to get your thoughts.

brimoor:
@ehofesmann thanks for your detailed thoughts here.

Per your comments, I went ahead and nixed the per-label skeleton feature. I also didn't want to add it and was hoping for some confirmation bias :)

Regarding `Keypoint` and `Keypoints`, I agree that `Keypoint` being singular is confusing on its face, since, after all, it represents a collection of one or more points. However, since we do need a way to represent "points representing one person" versus "groups of points representing multiple people", our hand is basically forced. I'm okay with it though because this naming is semantically consistent with `Detection` vs `Detections`, for example.

Allowing `Keypoint.points` to be either a dict or a list is not something that we can do in FiftyOne right now. We could theoretically allow the user to choose between `Keypoint.points_list` or `keypoint.points_dict`, but I don't want to introduce that complexity right now. That would be similar to allowing users to choose between `Classification.label_str` and `Classification.target_int`, for example. Also thought about it, but not acting on it just yet.

We could add a factory method like `KeypointSkeleton.from_connections(labels, connections)` that would allow you construct a skeleton from a more human-readable data format like `connections=[("left eye", "right eye"), ...]`. But on the implementation side, this is always going to need to be converted to integer indices for the actual rendering. The choice basically comes down to whether the data model should be optimized for human-readability or computational-readiness/efficiency.

brimoor:
Here's a direction we may want to go:

```py
class Keypoint(Label):
    label = fof.StringField()
    points = fof.ListField(fof.EmbeddedDocumentField(Point))

class Point(Label):
    x = fof.FloatField()
    y = fof.FloatField()
    confidence = fof.FloatField()
```

fiftyone/utils/labels.py
```diff
@@ -341,3 +343,115 @@ def classification_to_detections(sample_collection, in_field, out_field):
             image[out_field] = fol.Detections(detections=[detection])
 
         sample.save()
+
+
+def filter_keypoints(sample_collection, field, expr=None, labels=None):
```
benjaminpkane, line 348: Shouldn't this proivde an `only_matches` option? If the confidence filter does not filter empty keypoints than it is better to do this all client side in the App.

brimoor: Hmm yeah that does make sense. If a filter omits all points in a `Keypoint`, remove the entire object.

I think I can add that

benjaminpkane:
This is ready excluding some testing and confidence filtering. Once `only_matches` is added keypoint filtering I can finish.

benjaminpkane:
keypoints

fiftyone/utils/labels.py
brimoor: @benjaminpkane ok added the `only_matches` flag.

I'm thinking about converting this utility to a proper `FilterKeypoints(ViewStage)` and `SampleCollection.filter_keypoints()` view stage. Any objections?

benjaminpkane: I might prefer `FilterLabels` to support keypoints automatically. Perhaps in addition to `FilterKeypoints`. `FilterKeypoints` is somewhat specific but I suppose thats fine for now.

brimoor: I thought about merging this functionality into `FilterLabels`, but the problem is that will require introducing new optional kwargs, say `points_filter` and `points_labels`, that are only applicable when the field being filtered happens to be of type `Keypoint(s)`.

benjaminpkane: Ok. Cause to refactor `Keypoint` in the future

benjaminpkane:
Ok, ready

brimoor:
> Ok, ready

@benjaminpkane Almost there! I only saw two issues:
- In the grid view, when I use the `skeleton.label` filter widget, I get a `FilterLabels` stage (which isn't valid) but it should be a `FilterKeypoints` stage
- In the expanded modal, the tooltip does not appear when hovering over certain keypoints when `NaN` values are involved. Try loading the `filter_keypoints()` views in the code snippet in the PR description to see this

benjaminpkane:
Fixed. 

> In the grid view, when I use the skeleton.label filter widget, I get a FilterLabels stage (which isn't valid) but it should be a FilterKeypoints stage

I was unable to reproduce the above. Let me know if it is still an issue

brimoor:
> Fixed.
> 
> > In the grid view, when I use the skeleton.label filter widget, I get a FilterLabels stage (which isn't valid) but it should be a FilterKeypoints stage
> 
> I was unable to reproduce the above. Let me know if it is still an issue

LGTM, thanks! (probably I forgot to rebuild the App or something)

PR #1526, opened 2022-01-12
Adding support for boolean view indexing

brimoor:
Adds a boolean view indexing syntax:

```py
view = sample_collection[bool_array]
```

as a shorthand for:

```py
ids = itertools.compress(sample_collection.values("id"), bool_array)
view = sample_collection.select(ids)
```

Also adds a boolean expression indexing syntax:

```py
view = sample_collection[bool_expr]
```

as shorthand for:

```py
view = sample_collection.match(bool_expr)
```

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone import ViewField as F

dataset = foz.load_zoo_dataset("quickstart")

expr = F("uniqueness") > 0.7

# Existing syntax
view1 = dataset.match(expr)
print(view1)

# New boolean array syntax
bools = dataset.values(expr)  # list/array of booleans
view2 = dataset[bools]
print(view2)

# New boolean expression syntax
view3 = dataset[expr]
print(view3)
```

PR #1617, opened 2022-02-24
Docs updates

brimoor:
Changelog
- Documents https://github.com/voxel51/fiftyone/pull/1525
- Documents https://github.com/voxel51/fiftyone/pull/1526
- Documents `Dataset.load_evaluation_results()`

### DIFF
--- a/docs/source/user_guide/using_datasets.rst
+++ b/docs/source/user_guide/using_datasets.rst
@@ -454,6 +454,19 @@ whenever accessing the sample from the |Dataset|:
 You can use :ref:`dataset views <using-views>` to perform more sophisticated
 operations on samples like searching, filtering, sorting, and slicing.
 
+.. note::
+
+    Accessing a sample by its integer index in a |Dataset| is not allowed. The
+    best practice is to lookup individual samples by ID or filepath, or use
+    array slicing to extract a range of samples, and iterate over samples in a
+    view.
+
+    .. code-block:: python
+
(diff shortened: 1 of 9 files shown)


docs/source/user_guide/using_datasets.rst
```diff
@@ -2335,11 +2348,12 @@ associated metadata.
...
 .. note::
 
-    In most cases, it is recommended to :ref:`add custom attributes as fields
-    directly on<using-labels>` the |Label| object. However, a typical use case
-    for this feature, as opposed to simply storing custom attributes directly
-    on the |Label| object, is to store predictions and associated
-    confidences of a classifier applied to the object patches.
+    In most cases, it is recommended to
+    :ref:`add custom attributes as fields directly on <using-labels>` the
+    |Label| object. However, a typical use case for this feature, as opposed to
```
ehofesmann, line 2353:
```suggestion
    |Label| object. 
    
    .. code::
    
        sample["label_field"]["custom"attribute"] = 51
    
    However, a typical use case for this feature, as opposed to
```

I'm not sure if we should really keep this section at all, but if so, having a code example of how we recommend they store attributes first could be useful.

brimoor: Good point. I converted this to an actual deprecation warning and included some example code showing what you *should do*

ehofesmann (approved):
LGTM!

PR #1618, opened 2022-02-24
Storing database version in the db

brimoor:
Stores the database version in a new `config` collection.

Previously, the DB version was stored on disk in a `${fiftyone.config.database_dir}/config.json` file, which was problematic when connecting to centralized databases via `database_uri`, because each new client would think it was connecting to a new database and run all upward admin migrations.

Upgrading to and downgrading from this branch is seamless.


benjaminpkane (approved):
LGTM

PR #1619, opened 2022-02-24
Use full video support in draw_labels()

brimoor:
Resolves https://github.com/voxel51/eta/issues/555.

Previously, the example below would not render the video-level attribute on the video because video-level attributes were only added to frames that have (possibly empty) `Frame` instances.

Now, video-level attributes will always be rendered on frames `1, 2, ..., sample.metadata.total_frame_count`.

```py
import fiftyone as fo
import fiftyone.utils.annotations as foua

sample = fo.Sample(filepath="/Users/Brian/Desktop/water.mp4")
sample["weather"] = fo.Classification(label="sunny")

foua.draw_labeled_video(sample, "/tmp/labeled.mp4")
```

### DIFF
--- a/fiftyone/core/metadata.py
+++ b/fiftyone/core/metadata.py
@@ -209,14 +209,18 @@ class VideoMetadata(Metadata):
         )
 
 
-def compute_sample_metadata(sample, skip_failures=False):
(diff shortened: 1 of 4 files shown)


ehofesmann (approved):
LGTM

Issue #1629, opened 2022-03-04
[BUG] Coco Object Detection Data Loader fails to read annotation json files if it contains "info":  NULL


Bukkster:
### System information

-  **OS Platform and Distribution (e.g., Linux Ubuntu 16.04):** Windows 11 PRO 64-bit, Version: 10.0.22000 , Build 22000,
-   **FiftyOne installed from:** pip in conda environment
-   **FiftyOne version (run `fiftyone --version`)**: FiftyOne v0.14.3, Voxel51, Inc.
-   **Python version**: 3.7.12

### Commands to reproduce

```
import fiftyone as fo

# Define dataset name & paths 
val_name = "Objects365Val"
val_data_path = "C:/Users/UserName/Coding/Datasets/Objects365/Images/val"
val_labels_path = "C:/Users/UserName/Coding/Datasets/Objects365/Annotations/val.json"

# Load the dataset
obj365ValDataset = fo.Dataset.from_dir(
    name = val_name,
    dataset_type=fo.types.COCODetectionDataset,
    data_path=val_data_path,
    labels_path=val_labels_path,
)
```

### Describe the problem

Categories fail to load when when I try to use the coco object detection data loader with the Objects 365 dataset which has a coco formatting (except without supercatagorey field).  I get the following Type Error:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~\AppData\Local\Temp\ipykernel_26932\4192680555.py in <module>
      8     dataset_type=fo.types.COCODetectionDataset,
      9     data_path=val_data_path,
---> 10     labels_path=val_labels_path,
     11 ) 
     12 #label_map_path=label_map_path, class_path=class_path,

C:\Programs\Anaconda3\envs\fiftyone\lib\site-packages\fiftyone\core\dataset.py in from_dir(cls, dataset_dir, dataset_type, data_path, labels_path, name, label_field, tags, **kwargs)
   3553             label_field=label_field,
   3554             tags=tags,
-> 3555             **kwargs,
   3556         )
   3557         return dataset

C:\Programs\Anaconda3\envs\fiftyone\lib\site-packages\fiftyone\core\dataset.py in add_dir(self, dataset_dir, dataset_type, data_path, labels_path, label_field, tags, expand_schema, add_info, **kwargs)
   2341             tags=tags,
   2342             expand_schema=expand_schema,
-> 2343             add_info=add_info,
   2344         )
   2345 

[...]

PR #1639, opened 2022-03-14
Update fiftyone/utils/coco.py to fix bug #1629

Bukkster:

Redefined "info" variable initialization in line 1273 _parse_coco_detection_annotations definition so that it will always be a dictionary type even if info in the annotation json file is defined as null, NoneType, or anything other than DictType.

This avoids value assignment issues with line 1280 or 1283 in coco.py

Tested by trying to load Objects365 object detection database, which in its annotation json files, defines:
```
"info":  null,
```
Code used to test fix:

```
import fiftyone as fo

val_name = "obj365Val"
val_data_path = "C:/Users/UserName/Datasets/Objects365/Images/val"
val_labels_path = "C:/Users/UserName/Datasets/Objects365/Annotations/val.json"

obj365ValDataset = fo.Dataset.from_dir(
    name = val_name,
    data_path=val_data_path,
    labels_path=val_labels_path,
)
```

## Release Notes

Bug Fix [#1629]

        notes for FiftyOne users.

voxel51/fiftyone/fiftyone/utils/coco.py

-   [ ] Other

### DIFF
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -1270,12 +1270,13 @@ def load_coco_detection_annotations(json_path, extra_attrs=True):
 
 def _parse_coco_detection_annotations(d, extra_attrs=True):
     # Load info
-     info = {}
-        if d.get("info", {}) is not None
-            info.update(d.get("info", {}))
-            
+    info = d.get("info", None)
     licenses = d.get("licenses", None)
     categories = d.get("categories", None)
+    
+    if info is None:
+        info = {}
+        
     if licenses is not None:
         info["licenses"] = licenses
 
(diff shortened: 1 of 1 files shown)


fiftyone/utils/coco.py
```diff
@@ -1270,7 +1270,10 @@ def load_coco_detection_annotations(json_path, extra_attrs=True):
 
 def _parse_coco_detection_annotations(d, extra_attrs=True):
     # Load info
-    info = d.get("info", {})
+     info = {}
```
brimoor, line 1273: I'd recommend writing it like this:

```py
info = d.get("info", None)
if info is None:
    info = {}
```

It's a bit easier to read 😄

brimoor (changes requested):

fiftyone/utils/coco.py
Bukkster: Thanks Just made the change

brimoor (approved):
LGTM

brimoor:
Thanks for the patch @Bukkster!

PR #1640, opened 2022-03-14
Removing deprecated skip_unlabeled parameter from docs

brimoor:
At some point `skip_unlabeled` was removed from the `LabeledDatasetImporter` interface, but there were some stray instances remaining in the docs that needed to be removed.


ehofesmann (approved):
LGTM

PR #1641, opened 2022-03-15
Add Creating Views recipe

ehofesmann:
This PR adds a new recipe notebook covering how to create views and how to use view fields and expressions. Specifically noting how to work with nested lists and examples of different types of view expressions.


brimoor (approved):
LGTM

PR #1642, opened 2022-03-15
Add destination field argument when loading annotations

ehofesmann:
This PR adds the `destination_field` argument to `fiftyone.utils.annotations.load_annotations` allowing you to specify the name of the field to which to load annotations. This is only available for annotation runs operating off of a single label field.

While it is possible to clone sample fields to perform the same function, this simplifies that workflow, removing the necessity of managing additional fields. 

One example is reannotating subsets of a large dataset one at a time. With this parameter, you can simply pass in the existing field to `annotate()`, then specify the `destination_field` when loading annotations to be able to immediately call `dataset.exists("destination_field")` and get the samples that have been reannotated with no additional overhead.

## Example:

```python
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=2).clone()

anno_key = "test_destination_field"
results = dataset.annotate(anno_key, label_field="ground_truth")

dataset.load_annotations(anno_key, cleanup=True, destination_field="test_field")
```


docs/source/integrations/cvat.rst
```diff
@@ -1957,6 +1964,58 @@ attributes between annotation runs.
    :alt: cvat-occ-widget
    :align: center
 
+
+.. _cvat-destination-field:
```
brimoor, line 1968: nit: stick to just one blank line between blocks in all cases (see rest of file).

I wish that our pre-commit hook would automatically enforce this like it does for Python, but unfortunately it doesn't. Goal is just that everyone use the same style to minimize diffs.

docs/source/integrations/labelbox.rst
```diff
@@ -687,7 +687,7 @@ take additional values:
 Note that only scalar-valued label attributes are supported. Other attribute
 types like lists, dictionaries, and arrays will be omitted.
 
-.. note:: 
+.. note::
```
brimoor, line 690: nit: please tweak your editor so that it only removes trailing whitespace on lines that you edit, not the entire file.

Looks like yours is configured to process the whole file, which can lead to large cosmetic diffs whenever you touch a file.

fiftyone/utils/annotations.py
```diff
@@ -945,7 +945,12 @@ def _format_attributes(backend, attributes):
 
 
 def load_annotations(
-    samples, anno_key, unexpected="prompt", cleanup=False, **kwargs
+    samples,
+    anno_key,
+    destination_field=None,
```
brimoor, line 950: suggestion: consider `dest_field` for brevity. Really up to you though

fiftyone/utils/annotations.py
```diff
@@ -957,6 +962,9 @@ def load_annotations(
     Args:
         samples: a :class:`fiftyone.core.collections.SampleCollection`
         anno_key: an annotation key
+        destination_field (None): an optional name of a new destination field
```
brimoor, line 965: Please add this parameter to `SampleColleciton.load_annotations()` too, for completeness

brimoor (approved):
LGTM. Just left a couple small comments for your consideration

PR #1647, opened 2022-03-17
Resolve bug when annotating views with mapped labels

ehofesmann:
When creating an annotation task, there is currently a bug where views using the `MapLabels` stage will result in labels that were mapped not being uploaded to the annotation backend. This PR ensures that labels that were mapped are uploaded and that the mapping is reversed upon loading the annotated labels.

This is only applicable for labels where the class was not modified, any labels that were created or the class was modified will use the mapped class name. This is because a label mapping can result in multiple class names mapped to one, so there is no way to reverse the mapping for labels that were changed or created.

## Example

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = (
    foz.load_zoo_dataset("quickstart", max_samples=1)
    .select_fields("ground_truth")
    .clone()
)

labels = dataset.distinct("ground_truth.detections.label")
label_map = {l: l.upper() for l in labels}

view = dataset.map_labels("ground_truth", label_map)

anno_key = "anno_key"
results = view.annotate(
    anno_key, backend="cvat", label_field="ground_truth",
)

# Create/delete/modify a detection

print(dataset.values("ground_truth.detections.label"))
# ["BIRD", "bird"]
```

The newly created label has the class name "BIRD" and the unmodified labels are unmapped back to "bird".

### DIFF
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -769,10 +769,6 @@ def _get_classes(
     if classes:
         return classes
 
-    return _get_existing_classes(samples, label_field)
-
-
-def _get_existing_classes(samples, label_field):
     _, label_path = samples._get_label_field_path(label_field, "label")
     return sorted(
         set(samples._dataset.distinct(label_path))
@@ -1054,7 +1050,7 @@ def load_annotations(
                         _merge_scalars(samples, annos, results, new_field)
                     else:
                         _merge_labels(
-                            samples, annos, results, new_field, anno_type,
+                            samples, annos, results, new_field, anno_type
                         )
                 else:
                     if label_field:
@@ -1543,7 +1539,7 @@ def _merge_label(
     only_keyframes=False,
 ):
     for field in _DEFAULT_LABEL_FIELDS_MAP.get(type(label), []):
-        if field == "label" and not allow_label_edits:
+        if not allow_label_edits and field == "label":
             continue
 
         if not allow_index_edits and field == "index":
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -26,7 +26,6 @@ import fiftyone.core.fields as fof
 import fiftyone.core.labels as fol
 import fiftyone.core.media as fomm
 import fiftyone.core.metadata as fom
-import fiftyone.core.stages as fos
 import fiftyone.core.utils as fou
 import fiftyone.core.validation as fov
 import fiftyone.utils.eta as foue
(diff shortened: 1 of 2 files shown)


fiftyone/utils/annotations.py
```diff
@@ -768,8 +769,26 @@ def _get_classes(
...
+    return _get_existing_classes(samples, label_field)
+
+
+def _get_existing_classes(samples, label_field):
+    mapped_labels = _get_labels_map(samples, label_field)
+
     _, label_path = samples._get_label_field_path(label_field, "label")
-    return samples._dataset.distinct(label_path)
+    classes = samples._dataset.distinct(label_path)
+    mapped_classes = [mapped_labels.get(c, c) for c in classes]
```
brimoor, line 780: I would recommend a different, simpler approach here.

Just to level set, the original thinking behind using `samples._dataset` to populate a default classes list here rather than `samples` was that, if the user has constructed a view they want to edit that only contains a subset of labels, they would probably still like for the full list of classes on that label field to be made available in the annotation backend.

For example, say they have tagged some objects with incorrect labels that they want to edit, and it so happens that only `dog` labels are in the view and the desired action is to replace these with `german shepherd` or `terrier`, which are also in the full dataset schema.

The impetus for this PR is that this strategy doesn't account for the fact `samples` may be a view that dynamically modifies the labels, which I agree needs to be handled.

However, the issue with the chosen approach here is that `map_labels()` is only one of many possible ways to edit labels. For example, the view could use something like `set_field()` and `F("label").upper().concat("-OOPS")` to do the transformation. Or maybe the view just goes off the rails and uses `mongo()`...

Therefore, my recommendation would be to simply replace:

```py
classes = samples._dataset.distinct(label_path)
```

with 

```py
classes = sorted(set(samples._dataset.distinct(label_path)) | set(samples.distinct(label_path)))
```

Then we get the best of both worlds: any label in the dataset's classes OR the view's classes can be used, and there's no fancy transformations done at import time.

brimoor (changes requested):

fiftyone/utils/annotations.py
ehofesmann: Great call! This does end up making any label mapping permanent on the dataset, however, this does seem to be the best way to handle most cases. I've made these updates.

brimoor (approved):
LGTM

fiftyone/utils/annotations.py
```diff
@@ -26,6 +26,7 @@
 import fiftyone.core.labels as fol
 import fiftyone.core.media as fomm
 import fiftyone.core.metadata as fom
+import fiftyone.core.stages as fos
```
brimoor, line 29: This import is no longer needed

PR #1649, opened 2022-03-18
Add note about outdated metadata in transform_images

willyd:

Add documentation about outdated metadata after resizing images with `transform_images`.

Not tested. It's documentation only. Please LMK if the format is off.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/image.py
+++ b/fiftyone/utils/image.py
@@ -74,6 +74,15 @@ def transform_images(
     The ``filepath`` of the samples are updated to point to the transformed
     images.
 
+    .. note::
+
+        This method will not update any outdated metadata fields on the collection
+        after resizing. This can cause annotations to show in the wrong place. To avoid
+        this, the recommended approach is to call:
+
+            sample_collection.compute_metadata(overwrite=True)
+
+
     Args:
         sample_collection: a
             :class:`fiftyone.core.collections.SampleCollection`
--- a/fiftyone/utils/image.py
+++ b/fiftyone/utils/image.py
@@ -76,9 +76,9 @@ def transform_images(
 
     .. note::
 
-        This method will not update any outdated metadata fields on the collection
-        after resizing. This can cause annotations to show in the wrong place. To avoid
-        this, the recommended approach is to call:
+        This method will not update any outdated metadata fields on the
+        collection after resizing. This can cause annotations to show in the
+        wrong place. To avoid this, the recommended approach is to call::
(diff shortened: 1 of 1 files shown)


fiftyone/utils/image.py
```diff
@@ -74,6 +74,15 @@ def transform_images(
     The ``filepath`` of the samples are updated to point to the transformed
     images.
 
+    .. note::
+
+        This method will not update any outdated metadata fields on the collection
+        after resizing. This can cause annotations to show in the wrong place. To avoid
+        this, the recommended approach is to call:
```
brimoor, line 81: Using `call::` instead of `call:` will cause the content below to be rendered as code in the docs, which is nice

fiftyone/utils/image.py
```diff
@@ -74,6 +74,15 @@ def transform_images(
     The ``filepath`` of the samples are updated to point to the transformed
     images.
 
+    .. note::
+
+        This method will not update any outdated metadata fields on the collection
+        after resizing. This can cause annotations to show in the wrong place. To avoid
```
brimoor, line 80: Our style guide requests < 80 character line limit. Could you update the line breaks to satisfy that please? :)

brimoor (approved):
Thanks for this!

Before we merge, could you add this note to the other methods in this module that take a `sample_collection` as input and could resize the images/videos?

fiftyone/utils/image.py
willyd: Ok I will change that.

fiftyone/utils/image.py
willyd: Sure thing.

willyd:
> Before we merge, could you add this note to the other methods in this module that take a `sample_collection` as input and could resize the images/videos?

Yes of course.

willyd:
@brimoor 

I don't see any **public** methods that take in a collection and would potentially resize the images. Is your request about another module or about private methods?

brimoor:
> @brimoor
> 
> I don't see any other **public** methods that take in a collection and would potentially resize the images. Is your request about another module or about private methods?

Ah my bad, I was thinking of the `fiftyone.utils.video` module, which has an analogous method for transforming videos.

willyd:
All right then. LMK if you require any changes.

brimoor:
@willyd all set, thanks for the contrib!

Issue #1654, opened 2022-03-22
[BUG] Error when annotating sample-level video label with Labelbox


ehofesmann:
Labelbox supports annotating sample-level classifications on videos. However, there is currently a bug when attempting to load these annotations.

## Example:

```python
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video", max_samples=1).clone()

results = dataset.annotate(
    "anno_key",
    label_field="class",
    classes=["test"],
    launch_editor=True,
    label_type="classification",
    backend="labelbox",
)

dataset.load_annotations("anno_key", cleanup=True)
```

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/work/fiftyone/fiftyone/fiftyone/core/document.py in get_field(self, field_name)
    155         try:
--> 156             value = self._doc.get_field(field_name)
    157         except AttributeError:

~/work/fiftyone/fiftyone/fiftyone/core/odm/mixins.py in get_field(self, field_name)
    261             raise AttributeError(
--> 262                 "%s has no field '%s'" % (self._doc_name(), field_name)
    263             )

AttributeError: Frame has no field 'class'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
~/work/fiftyone/fiftyone/fiftyone/core/document.py in __getitem__(self, field_name)
     67         try:
---> 68             return self.get_field(field_name)
     69         except AttributeError as e:

~/work/fiftyone/fiftyone/fiftyone/core/document.py in get_field(self, field_name)
    690
--> 691         value = super().get_field(field_name)
    692

~/work/fiftyone/fiftyone/fiftyone/core/document.py in get_field(self, field_name)
    158             raise AttributeError(
--> 159                 "%s has no field '%s'" % (self.__class__.__name__, field_name)
    160             )

AttributeError: FrameView has no field 'class'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
<ipython-input-5-3e9c1f4b7f94> in <module>
----> 1 dataset.load_annotations("anno_key", cleanup=True)

~/work/fiftyone/fiftyone/fiftyone/core/collections.py in load_annotations(self, anno_key, unexpected, cleanup, **kwargs)
   6396         """
   6397         return foua.load_annotations(
-> 6398             self, anno_key, unexpected=unexpected, cleanup=cleanup, **kwargs,
   6399         )
   6400

[...]

PR #1655, opened 2022-03-22
Disallow video-level annotation

### LINKED ISSUE #1430: [FR] Prevent users from specifying sample-level spatial fields when annotating videos
Currently the annotation API lets users define sample-level spatial fields like `detections` for a video annotation task. However, importing such data will raise an error.

The issue is that the correct syntax is to prefix frame-level fields with `frames.`, but this is currently hard to self-debug.

We should do one or both of the following things to make this self-diagnosable:
- Raise an error at schema creation time alerting the user that they cannot annotate sample-level spatial fields for video
- Update the import logic so that frame-level labels are treated as new, unexpected annotations rather than treating them as sample-level labels (they are not)

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video").limit(1).select_fields().clone()

dataset.annotate(
    "test",
    label_field="ground_truth",
    label_type="detections",  # this should instead be "frames.detections"
    classes=["vehicle"],
    launch_editor=True,
)

# Create a video track in CVAT...

```

```
Downloading labels from CVAT...
Download complete
TypeError                                 Traceback (most recent call last)
<ipython-input-4-07cea798f021> in <module>
----> 1 dataset.load_annotations("test")

~/dev/fiftyone/fiftyone/core/collections.py in load_annotations(self, anno_key, skip_unexpected, cleanup, **kwargs)
   6291             skip_unexpected=skip_unexpected,
   6292             cleanup=cleanup,
-> 6293             **kwargs,
   6294         )
   6295 

~/dev/fiftyone/fiftyone/utils/annotations.py in load_annotations(samples, anno_key, skip_unexpected, cleanup, **kwargs)
   1027                         label_info=label_info,
   1028                         global_attrs=global_attrs,
-> 1029                         class_attrs=class_attrs,
   1030                     )
   1031             else:

~/dev/fiftyone/fiftyone/utils/annotations.py in _merge_labels(samples, anno_dict, results, label_field, label_type, label_info, global_attrs, class_attrs)
   1282             only_keyframes = True
   1283 
-> 1284         _update_tracks(samples, label_field, anno_dict, only_keyframes)
   1285 
   1286     id_map = results.id_map.get(label_field, {})

~/dev/fiftyone/fiftyone/utils/annotations.py in _update_tracks(samples, label_field, anno_dict, only_keyframes)
   1558 
   1559         for _frame_id, _label_ids, _indexes in zip(
-> 1560             _frame_ids, _frame_lids, _frame_inds
   1561         ):
   1562             _label_ids = _to_list(_label_ids)

TypeError: zip argument #2 must support iteration
```

ehofesmann:
Resolves #1430 

Adds a property to annotation backends to decide whether to allow annotation of video-level labels. CVAT does not provide a way to annotate labels at a video-level, and while Labelbox does, there is currently a bug that needs to be resolved before the annotations can be loaded: #1654 

Note: This primarily refers to classifications, spatial label can never be annotated at a video-level as that concept is not well defined.

## Example:

```python
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video", max_samples=1).clone()

results = dataset.annotate(
    "anno_key",
    label_field="class",
    classes=["test"],
    launch_editor=True,
    label_type="classification",
)

dataset.load_annotations("anno_key", cleanup=True)
```

Previously would raise an error when loading annotations which would have resulted in wasted effort by annotators. Now raises an error when calling `annotate()`:

```
ValueError: Invalid label field 'class'. Backend 'cvat' does not support annotating video fields at a sample-level. Labels must be stored in a frame-level field, i.e., one that starts with 'frames.'
```

### DIFF
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -407,13 +407,21 @@ def _build_label_schema(
         _return_type = _RETURN_TYPES_MAP[_label_type]
         _is_trackable = _is_frame_field and _return_type in _TRACKABLE_TYPES
 
-        if is_video and _return_type in _SPATIAL_TYPES and not _is_frame_field:
-            raise ValueError(
-                "Invalid label field '%s'. Spatial labels of type '%s' being "
-                "annotated on a video must be stored in a frame-level field, "
-                "i.e., one that starts with 'frames.'"
-                % (_label_field, _label_type)
-            )
+        if is_video and not _is_frame_field:
+            if not backend.supports_video_sample_fields:
+                raise ValueError(
+                    "Invalid label field '%s'. Backend '%s' does not support "
+                    "annotating video fields at a sample-level. Labels must be "
+                    "stored in a frame-level field, i.e., one that starts with "
+                    "'frames.'" % (_label_field, backend.config.name)
+                )
+            elif _return_type in _SPATIAL_TYPES:
+                raise ValueError(
+                    "Invalid label field '%s'. Spatial labels of type '%s' "
+                    "being annotated on a video must be stored in a "
+                    "frame-level field, i.e., one that starts with 'frames.'"
+                    % (_label_field, _label_type)
+                )
 
         # We found an existing field with multiple label types, so we must
         # select only the relevant labels
@@ -1782,6 +1790,15 @@ class AnnotationBackend(foa.AnnotationMethod):
         existing video track annotations.
         """
         raise NotImplementedError("subclass must implement supports_keyframes")
+
+    @property
+    def supports_video_sample_fields(self):
+        """Whether this backend supports annotating video labels at a
+        sample-level or only at a frame-level.
+        """
+        raise NotImplementedError(
+            "subclass must implement supports_video_sample_fields"
+        )
 
     @property
     def requires_label_schema(self):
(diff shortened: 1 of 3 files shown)


fiftyone/utils/annotations.py
```diff
@@ -1783,6 +1791,15 @@ def supports_keyframes(self):
         """
         raise NotImplementedError("subclass must implement supports_keyframes")
 
+    @property
+    def supports_video_sample_fields(self):
+        """Whether this backend supports annotating video labels at a
+        sample-level or only at a frame-level.
```
brimoor, line 1797: nit: I'd remove "or only at a frame-level" here.

Maybe in the future we'll have backends that don't support videos at all, in which case this property would be False but frame-level labels aren't supported either (presumably a new `supports_videos=False` parameter would be added)

fiftyone/utils/annotations.py
```diff
@@ -407,13 +407,21 @@ def _build_label_schema(
...
 
-        if is_video and _return_type in _SPATIAL_TYPES and not _is_frame_field:
-            raise ValueError(
-                "Invalid label field '%s'. Spatial labels of type '%s' being "
-                "annotated on a video must be stored in a frame-level field, "
-                "i.e., one that starts with 'frames.'"
-                % (_label_field, _label_type)
-            )
+        if is_video and not _is_frame_field:
+            if not backend.supports_video_sample_fields:
```
brimoor, line 411: nit: I'd check for the second error first. If a user has provided a spatial label field, they most likely just forgot `frames.`, and the second error is more apropos for that

brimoor (approved):
LGTM. Couple small comments for ya

PR #1659, opened 2022-03-25
Remove `quickstart` from selection test

benjaminpkane:
Removes the `quickstart` dataset from the `selection_tests.py` to avoid download failures and improve test speeds.


brimoor (approved):
LGTM

PR #1662, opened 2022-03-29
Adding a concat() view stage

brimoor:
Adds a `concat()` view stage that allows for concatenating one collection onto another:

```py
view = view1.concat(view2)

len(view) == len(view1) + len(view2)  # True

# equivalent syntax sugar
view = view1 + view2
```

The implementation strictly prohibits concatenating views into different datasets, because allowing that would cause undefined behavior when iterating over the contents of the concatenated view, since a reference to the correct parent dataset is required in order to perform operations like schema expansion, sample deletion, etc.

One important subtlety that this implementation accounts for is concatenating "generated" views like patches/clips/frames that involve temporary collections, since it is important for efficiency reasons that temporary collections *not* be re-generated when serializing/deserializing concatenated views (which happens when working with `session.view`, for example).

## TODO

- [x] Add documentation to the user guide
- [x] Is it a problem that users can define concatenated views that contain duplicate samples (duplicate `id`s)? Both in the API and also in the App (eg selecting a sample)

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

#
# Concatenate two regular views
#

view1 = dataset.match(F("uniqueness") < 0.2)
view2 = dataset.match(F("uniqueness") > 0.7)

view = view1.concat(view2)

print(view1)
print(view2)
print(view)

session = fo.launch_app(view)

#
# Concatenate two patches views
#

gt_patches = dataset.to_patches("ground_truth")

patches1 = gt_patches[:10]
patches2 = gt_patches[-10:]
patches = patches1 + patches2

print(patches1)
print(patches2)
print(patches)

session.view = patches
```


benjaminpkane:
I will leave concrete after testing on Monday. But just reading through this brings to mind some benefits of introducing a formal way to define "semantic" id's when working with multiple views on a single document.

In MongoDB it would look like:

```
{ _id: { _id: "...", name: "name"} }
```

Or in this case maybe:
```
{ _id: { _id: "...", concatIndex: 0} }
```

This is something that the App could take advantage of, and provide some more meaning to users. I'm not saying it needs to a part of this work, but it would certainly complement it.

brimoor:
@benjaminpkane I'm fine with how the App handles duplicate sample IDs right now: it only renders the last instance in the grid, but no errors occur.

So no action required on this PR if you don't want to think about duplicate IDs right now.

benjaminpkane (approved):
LGTM

PR #1410, opened 2021-11-11
Added the FIW dataset

visionjo:
## What changes are proposed in this pull request?

Adding FIW to fiftyone's datazoo.

## How is this patch tested? If it is not, please explain why.

Locally, in terminal and via ipython.

## Release Notes
* Added fiw to datazoo
* Created documentation
* Show demo with test splits
### Is this a user-facing change that should be mentioned in the release notes?

-   [ ] No. You can skip the rest of this section.
-   [x] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

(Details in 1-2 sentences. You can just refer to another PR with a description
if this PR is part of a larger change.)
Families in the Wild (FIW), a public benchmark for recognizing families via facial images, is now
available as a dataset in fiftyone's datazoo. Easy to download, explore, and run experiments. This is 
a continued WIP.

#### TO DO
- [ ] Create FIW dataset class
- [ ] Parse the family relationship matrices and add lists of relatives (i.e., one list per relationship type)
- [ ] Add task one, and demo end-to-end evaluation
- [ ] Add face features as optional download
- [ ] Incorporate metrics and other means of analysis (perhaps integrate with MLFlow for experiment tracking)
- [ ] Add task 2
- [ ] Add task 3
- [ ] Add multimedia

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [x] Core: Core `fiftyone` Python library changes
-   [x] Documentation: FiftyOne documentation changes
-   [ ] Other

PR #1663, opened 2022-03-29
Add FIW to the dataset zoo

ehofesmann:
This PR extends #1410 to add the Families in the Wild dataset to the FiftyOne dataset zoo. This is an image dataset of faces as well as labels about the relationships and kinships between them.

Families in the Wild is a public benchmark for recognizing families via facial images. The dataset contains over 26,642 images of 5,037 faces collected from 978 families. A unique Family ID (FID) is assigned per family, ranging from F0001-F1018 (i.e., some families were merged or removed since its first release in 2016). The dataset is a continued work in progress. Any contributions are both welcome and appreciated!

Faces were cropped from imagery using the five-point face detector MTCNN from various phototypes (i.e., mostly family photos, along with several profile pics of individuals (facial shots). The number of members per family varies from 3-to-26, with the number of faces per subject ranging from 1 to >10.

Various levels and types of labels are associated with samples in this dataset. Family-level labels contain a list of members, each assigned a member ID (MID) unique to that respective family (e.g., F0011.MID2 refers to member 2 of family 11). Each member has annotations specifying gender and relationship to all other members in that respective family.

Within FiftyOne, each sample corresponds to a single face image and contains primitive labels of the Family ID, Member ID, etc. The relationship labels are stored as multi-label Classifications, where each classification represents one relationship that the member has with another member in the family. The number of relationships will differ from one person to the next, but all faces of one person will have the same relationship labels.

Additionally, the labels for the [Kinship Verification task](https://competitions.codalab.org/competitions/21843) are also loaded into this dataset through FiftyOne. These labels are stored as classifications just like relationships, but the labels of kinship differ from those defined above. For example, rather than Parent, the label might be fd representing a Father-Daughter kinship or md for Mother-Daughter.

In order to make it easier to browse the dataset in the FiftyOne App, each sample also contains a face_id field containing a unique integer for each face of a member, always starting at 0. This allows you to filter the face_id field to 0 in the App to show only a single image of each person.

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("fiw")

```

image


fiftyone/zoo/datasets/base.py
```diff
@@ -1157,6 +1158,98 @@ def _get_raw_dir(self, dataset_dir):
         return os.path.join(root_dir, "raw")
 
 
+class FIWDataset(FiftyOneDataset):
+    """Families in the Wild is a public benchmark for recognizing families via
```
brimoor, line 1162: Small point: this description is shorter than the dataset's entry in the User Guide.

In the future, I'd like to remove the redundancy here by automatically generating the RST from these class docstrings (as is done for the model zoo). That will take a bit of work to achieve, but the point is that, if/when we do that, you might lose the longer description in the `.rst`.

tldr: consider updating this description to match the `.rst`.

brimoor (approved):
LGTM

PR #1664, opened 2022-03-31
Always populate iscrowd field when exporting COCO labels

brimoor:
Previously exporting in [COCO format](https://voxel51.com/docs/fiftyone/user_guide/export_datasets.html#cocodetectiondataset) would only include the `iscrowd` attribute in the JSON if the attribute is actually populated on the collection being exported. This was problematic for certain downstream stages of COCO workflows, which will assume that `iscrowd: 0` is always populated if no crowd data is available.

This PR will now export `iscrowd: 0` if now crowd data is available.

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.delete_sample_field("ground_truth.detections.attributes")

dataset.export(
    labels_path="/tmp/labels.json",
    label_field="ground_truth",
)
```

```shell
# Previously would not include `iscrowd` attribute, but now does
python -m json.tool /tmp/labels.json
```

### DIFF
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -656,7 +656,8 @@ class COCODetectionDatasetExporter(
             -   ``False``: do not export extra attributes
             -   a name or list of names of specific attributes to export
         iscrowd ("iscrowd"): the name of a detection attribute that indicates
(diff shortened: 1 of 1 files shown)


ehofesmann (approved):
LGTM

PR #1666, opened 2022-04-04
Embedded document schema updates (take 2)

brimoor:
Change log
- Adds a database migration that infers embedded document schemas for existing datasets
- Updates the schema of the `metadata` field when a dataset's `media_type` is set


benjaminpkane (approved):
LGTM. Thanks!

PR #1667, opened 2022-04-05
Adding database admin config setting

brimoor:
Adds a `database_admin` config setting that can optionally be used to disallow database/dataset migrations.

Also fixes some migration edge cases that were missed in https://github.com/voxel51/fiftyone/pull/1666.


benjaminpkane (approved):
LGTM

PR #1671, opened 2022-04-06
Update PyTorch Lightning Flash integration

### LINKED ISSUE #1482: [BUG] Lightning Flash integration doesn't work with current Flash version

ehofesmann:
Resolves #1482 

This PR updates the integration with pytorch lightning flash to work with the updates from the past two versions that resulted in breaking changes for the previous implementation. The integration now works with flash versions >= 0.7.0.
Now, the tasks of classification, detection, segmentation, and embedding generation for images and classification for video work again. Additionally, the `dataset.apply_model(flash_model)` syntax can be used again. 

There were various changes made in lightning flash, the most relevant being:
* `model.predict()` no longer being supported, only `trainer.predict(model, datamodule)`
* data transformations/preprocessing updates moving it from a `data_pipeline` stored in the model
* Serializers being renamed to `output` and their parameters changed
* Changes in now model parameters and transformation arguments are supplied to datamodule, trainers, and models
* Removing pretrained models from examples

This PR also adds intensive tests copying all examples in our integration documentation to allow them to be tested more easily and regularly to track future chagnes.


docs/source/integrations/lightning_flash.rst
```diff
@@ -509,48 +557,67 @@ Flash models with FiftyOne serializers will directly return predictions as
...
 
-    from flash.image import ObjectDetector
-    from flash.image.detection.serialization import FiftyOneDetectionLabels
+    from flash import Trainer
+    from flash.image import ImageClassificationData, ImageClassifier
+    from flash.core.classification import FiftyOneLabelsOutput
 
     # Load your dataset
-    dataset = foz.load_zoo_dataset("quickstart", max_samples=5)
+    dataset = foz.load_zoo_dataset("quickstart", max_samples=5).clone()
```
brimoor, line 586: suggestion: remove `clone()` for simplicity?

I sometimes include it when using the entire `quickstart` dataset in an example that adds/edits fields, since many examples use the full dataset. But, when only a subset is being used, clashes are less likely

fiftyone/utils/flash.py
```diff
@@ -11,15 +11,17 @@
 
 import numpy as np
 
-import fiftyone.core.metadata as fom
+import fiftyone.core.labels as fol
+import fiftyone.core.media as fom
 import fiftyone.core.utils as fou
 
-fou.ensure_import("flash>=0.4")
+fou.ensure_import("flash>=0.6dev")
```
brimoor, line 18: In the docs and PR description you said `0.7`, not `0.6dev`. Use `0.7` for consistency?

fiftyone/utils/flash.py
```diff
@@ -96,6 +98,7 @@ def compute_flash_embeddings(
     embeddings_field=None,
     batch_size=None,
     num_workers=None,
+    transform_kwargs=None,
```
brimoor, line 101: Although it is a small breaking change, I think updating `trainer_kwargs` to be a dict (defaulting to `trainer_kwargs=None`) rather than `**trainer_kwargs` would make sense, for consistency with `transform_kwargs` here.

Make this change in analogous methods like `apply_flash_model()` too.

With the above changes in-place, you could then rename the `**trainer_kwargs` arguments of `SampleCollection.apply_model()` and `SampleCollection.compute_embeddings()` to just `**kwargs`, and update the docs to just explain that these are additional model-specific keyword arguments passed through to the underlying inference implementation (after all, users may want to pass `transform_kwargs` now). This would be an improvement because its bad to have Flash-specific things in the docstring of `SampleCollection` methods; the latter should be generic.

tests/intensive/lightning_flash_tests.py
```diff
@@ -0,0 +1,440 @@
...
+
+    python tests/intensive/lightning_flash_tests.py
+
+| Copyright 2017-2022, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
+
+import numpy as np
```
brimoor, line 14: nit: remove empty lines between imports and module docstring

tests/intensive/lightning_flash_tests.py
```diff
@@ -0,0 +1,440 @@
+"""
+Tests for the :mod:`fiftyone.utils.flash` module.
```
brimoor, line 2: Thanks for adding these! 💯 🚀

brimoor (changes requested):
Lookin' good! Just left a few small suggestions for you

fiftyone/utils/flash.py
ehofesmann: I'm gonna go with 0.7dev here and in the docs so that it won't fail when using source builds of flash, sound good?

brimoor: 👍

brimoor (approved):
LGTM

PR #1674, opened 2022-04-08
Fix CVAT CSRF bug

### LINKED ISSUE #1668: [BUG] CSRF Failed when connecting to CVAT servers
When connecting to some CVAT servers like [cvat.org](https://cvat.org), the login fails with the following error:
```
Exception: 403 error for request <PreparedRequest [POST]> to url https://cvat.org/api/tasks with the reason Forbidden. 
Error content: b'{"detail":"CSRF Failed: Referer checking failed - no Referer."}'
```

This can be circumvented by adding the `Referer` header pointing to the server API login url:
```python
results = dataset.annotate(
    anno_key,
    label_field=label_field,
    headers={"Referer": "https://cvat.org/api/auth/login"},
)
```

The reason for this issue is still unclear to me, but it can be resolved by checking the login attempt for a 403 error and automatically adding the referer header.

ehofesmann:
Resolves #1668 

Automatically add the `Referer` header to CVAT requests session if it does not exist to avoid the CSRF error.

This now works without needing to manually provide the `Referer` header.
```python
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=1).clone()

anno_key = "csrf_bug"
results = dataset.annotate(
    anno_key,
    label_field="ground_truth",
    url="https://cvat.org",
    username=USERNAME,
    password=PASSWORD,
)

```

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -3592,6 +3592,12 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
             self._server_version = 1
             self._login(username, password)
 
+        self._add_referer()
+
+    def _add_referer(self):
+        if "Referer" not in self._session.headers:
+            self._session.headers["Referer"] = self.login_url
+
     def close(self):
         """Closes the API session."""
         self._session.close()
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM

PR #1679, opened 2022-04-12
Consistent node version

benjaminpkane:
Use a consistent node version in the install script and App README (16.4.2 LTS).


brimoor (approved):
LGTM

Issue #1675, opened 2022-04-08
[BUG] 'KeyError' when loading dataset with relative paths on Windows


ace42588:
When attempting to load a COCO format dataset from disk in a Windows environment, an uncaught exception occurs if the labels json uses relative paths for the image. Output:
`Uncaught exception
Traceback (most recent call last):
  File "C:\Users\Trace\Documents\Work\Ping\load_coco_formatted.py", line 3, in <module>
    dataset = fo.Dataset.from_dir(
  File "C:\Python39\lib\site-packages\fiftyone\core\dataset.py", line 3664, in from_dir
    dataset.add_dir(
  File "C:\Python39\lib\site-packages\fiftyone\core\dataset.py", line 2454, in add_dir
    return self.add_importer(
  File "C:\Python39\lib\site-packages\fiftyone\core\dataset.py", line 2997, in add_importer
    return foud.import_samples(
  File "C:\Python39\lib\site-packages\fiftyone\utils\data\importers.py", line 129, in import_samples
    sample_ids = dataset.add_samples(
  File "C:\Python39\lib\site-packages\fiftyone\core\dataset.py", line 1378, in add_samples
    for batch in batcher:
  File "C:\Python39\lib\site-packages\fiftyone\core\utils.py", line 900, in __next__
    batch.append(next(self._iter))
  File "C:\Python39\lib\site-packages\fiftyone\utils\coco.py", line 415, in __next__
    image_path = self._image_paths_map[filename]
KeyError: 'midv2019/images/DG/DG01_01.tif'`

This appears to be caused by the image_paths_map having normalized key values. Normalizing the filename in the `__next__` def of the COCODetectionDatasetImporter seems to resolve the issue.
https://github.com/voxel51/fiftyone/blob/1986f1162fb91c4f57d8413dca44fee9644378b6/fiftyone/utils/coco.py#L410
becomes `filename = os.path.normpath(next(self._iter_filenames))`

brimoor:
Thanks for catching this! I think this same bug (`/` vs `\`) may exist in other places too, but we'll get it patched ASAP

PR #1684, opened 2022-04-14
Fixing #1675

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1675.

Also, all import formats will support Unix (`"path/to/image.jpg"`) or Windows-style (`"path\\to\\image.jpg"`) paths in labels files and data manifests, which will be properly normalized on all OSes.


ehofesmann (approved):
LGTM

PR #1687, opened 2022-04-19
Automatically delete non-persistent datasets when using custom database URIs

### LINKED ISSUE #1345: [FR] Add support for automatic cleanup of non-persistent datasets when using a database URI
When using a [custom DB hookup](https://voxel51.com/docs/fiftyone/user_guide/config.html#configuring-a-mongodb-connection), non-persistent datasets are not automatically cleaned up when there are no active database connections, as they are when using the default database.

For many workflows, this leads to an accumulation of throwaway datasets, so it would be nice if non-persistent datasets were automatically deleted when using a custom DB, too.

One important point here is that multiple users on different machines may be using a custom DB simultaneously. So, only the database, not any one Python session, knows when it is appropriate to execute the cleanup.

j053y:

Resolves #1345

Automatically delete non-persistent datasets when last of some number of concurrent database connections exits.

1. Start a client and create a new non-persistent dataset
2. Start one or more additional clients and list datasets (the non-persistent dataset should appear in this list)
3. Exit all but one client and list datasets (the non-persistent dataset should still appear in this list)
4. Exit the final client 
5. Start a new client and list datasets (the non-persistent dataset should no longer appear in this list)

        notes for FiftyOne users.

Non-persistent datasets are automatically cleaned up when using the built-in database or a custom database URI.

-   [ ] Other

### DIFF
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -95,6 +95,7 @@ def establish_db_conn(config):
         _connection_kwargs["port"] = int(established_port)
     if config.database_uri is not None:
         _connection_kwargs["host"] = config.database_uri
+        _db_service = fos.ExternalDatabaseService()
     elif _db_service is None:
         if os.environ.get("FIFTYONE_DISABLE_SERVICES", False):
             return
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -155,7 +155,7 @@ def _async_connect():
 
 
 def get_master_connection_count():
-    """Counts the number of master connections to database from fiftyone app.
+    """Counts the number of master connections to database from fiftyone.
 
     Returns:
         -   Integer count of master connections
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -155,7 +155,7 @@ def _async_connect():
 
 
 def get_master_connection_count():
-    """Counts the number of master connections to database from fiftyone.
+    """Counts the number of master connections to database from FiftyOne.
 
     Returns:
         -   Integer count of master connections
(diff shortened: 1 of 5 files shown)


fiftyone/core/service.py
```diff
@@ -241,6 +241,32 @@ def stop(self):
...
 
+class ExternalDatabaseService(Service):
+    """Service that manages the external MongoDB database."""
+
+    service_name = "ext-db"
+    allow_headless = True
+
+    @property
+    def command(self):
+        return ["sleep", "14d"]
```
benjaminpkane, line 252: This command fails on Mac. Windows will have to be supported too. It might be worthwhile to tweak https://github.com/voxel51/fiftyone/blob/develop/fiftyone/service/main.py to support command-less services that just has a cleanup call.

It seems a bit rude to start unnecessary processes on a users computer...

j053y: Good to know. I can definitely look into making a command-less service.

j053y: I modified service to run without a command, so there is no unnecessary sub-process and behavior should be consistent across each OS.

fiftyone/core/odm/database.py
```diff
@@ -124,7 +125,9 @@ def establish_db_conn(config):
 
             raise error
 
-    _client = pymongo.MongoClient(**_connection_kwargs)
+    _client = pymongo.MongoClient(
+        **_connection_kwargs, appname=foc.DATABASE_APPNAME
+    )
     _validate_db_version(config, _client)
```
benjaminpkane, line 131: Perhaps start the external service after the DB version (and connection) are validated? Right now the service starts even for invalid URIs

benjaminpkane:
Things look good to me other than my last comment! 🚀  I'll defer to @brimoor for approval

j053y:
Per Ben's recommendation, I moved the service initialization call to run after the external database is validated.

j053y:
Using a different approach  #1697

PR #1692, opened 2022-04-21
Database name, admin, and type

brimoor:
This PR adds some useful utilities for configuring database access.

### Database name

A new `database_name` config setting is added that allows for users to customize their MongoDB database name:

```py
import fiftyone as fo

print(fo.list_datasets())  # [...]
print(fo.config.database_name)  # fiftyone
```

```shell
export FIFTYONE_DATABASE_NAME=fiftytwo
```

```py
import fiftyone as fo

print(fo.list_datasets())  # []
print(fo.config.database_name)  # fiftytwo

foo.drop_database()  # cleanup
```

```shell
unset FIFTYONE_DATABASE_NAME
```

### Database/client types

Notions of database type and client type are added, and now database connections are refused if a Python client's type does not match the database's type.

The client type is hard-coded in `fiftyone.constants.CLIENT_TYPE`, and the database's type is stored in a new `type` field of the `config` collection.

This is an internal mechanism that exists to prevent OSS Python clients from connecting to non-OSS databases.

### Database admin

A new `database_admin` config setting is added. This parameter defaults to `True`, but clients can set it to `False`, in which case database migrations are not allowed (other than lazily migrating a dataset to the current database version, which is normal usage):

```shell
fiftyone migrate -v 0.15.0.1
export FIFTYONE_DATABASE_ADMIN=false
```

```py
import fiftyone as fo  # ERROR: only admins can migrate
```

```shell
unset FIFTYONE_DATABASE_ADMIN
```

```py
import fiftyone as fo
# Migrating database to v0.15.1
```

The documentation explains how this feature may be useful for OSS users leveraging `database_uri` to work on a shared database:

Screen Shot 2022-04-21 at 1 26 33 PM


j053y:
LGTM

j053y (approved):

PR #1697, opened 2022-04-22
Automatically delete non-persistent datasets when using custom database URIs

j053y:
Resolves https://github.com/voxel51/fiftyone/issues/1345
This is an alternate solution to #1687. 

Automatically delete non-persistent datasets upon sole connection to the database. This happens on connection time instead of exit.

Quickly test by running the following steps:
1. Open a new terminal session (Session # 1) and run `python`:
    ```python
    import fiftyone as fo
    dataset = fo.Dataset()
    fo.list_datasets()  # non-persistent dataset is listed
    ```

2. Keep Session # 1 running.

3. Open another terminal session (Session # 2):
    ```bash
    fiftyone datasets list  # non-persistent dataset is listed
    ```

4. Exit `python` in  Session # 1.
5. Back to Session # 2:
    ```bash
    fiftyone datasets list  # non-persistent dataset should no longer be listed
    ```

Should work for both a built-in and external Mongo database.

### DIFF
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -161,9 +161,7 @@ def establish_db_conn(config):
     )
     _validate_db_version(config, _client)
 
-    conn_count = _get_master_connection_count()
-    if conn_count <= 1:
-        _delete_non_persistent_datasets(config, _client)
+    _autodelete_non_persistent_datasets(config, _client)
 
     connect(config.database_name, **_connection_kwargs)
 
@@ -194,9 +192,31 @@ def _async_connect():
         )
 
 
-def _delete_non_persistent_datasets(config, client):
-    db = client[config.database_name]
+def _autodelete_non_persistent_datasets(config, client):
+    try:
+        cursor = client.admin.aggregate(
+            [
+                {"$currentOp": {"allUsers": True}},
+                {"$project": {"appName": 1, "command": 1}},
+                {
+                    "$match": {
+                        "appName": foc.DATABASE_APPNAME,
+                        "command.ismaster": 1,
+                    }
+                },
+            ]        )
 
+    except Exception as e:
+        logger.warning(
+            'Skipping automatic non-persistent dataset cleanup. This action requires read access of the "admin" database.'
+        )
+        return
+
+    conn_count = len(list(cursor))
+    if conn_count is None or conn_count > 1:
+        return
+
+    db = client[config.database_name]
     has_datasets = False
     for doc in db.datasets.find({"persistent": False}):
         has_datasets = True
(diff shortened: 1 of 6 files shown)


j053y:
This works on Linux, like my previous PR. I'm curious of the behavior on another OS, specifically OS X.

fiftyone/core/odm/database.py
```diff
@@ -173,15 +179,58 @@ def _connect():
...
+        )
+
+
+def _delete_non_persistent_datasets(config, client):
+    db = client[config.database_name]
+
+    has_datasets = False
+    for doc in db.datasets.find(
+        {"persistent": False},
+        {
```
brimoor, line 203: Side comment: I appreciate the optimization here, but FYI the dataset doc is intended to always be quite small, so I wouldn't bat an eye if you just used `db.datasets.find({"persistent": False})`

brimoor (approved):
LGTM. This one works for me on mac!

brimoor:
Okay two comments:
- Shall we just delete the `delete_non_persistent_datasets()` call from `DatabaseService.cleanup()`?
- @benjaminpkane this PR requires access to `db.admin` at import time for all users. Any concerns with this?

fiftyone/core/odm/database.py
brimoor: oh yeah that's fine, I was just saying you could use `for doc in db.datasets.find({"persistent": False})` in this method 😉

j053y: Oh I see. My mistake, that makes total sense :+1:

brimoor:
> * @benjaminpkane this PR requires access to `db.admin` at import time for all users. Any concerns with this?

A try-except with a warning that says "automatic non-persistent dataset cleanup requires admin db access" would be fine.

fiftyone/core/odm/database.py
benjaminpkane: @brimoor Aren't there grid fs files associated with a dataset that need to be cleaned up? Should [this function](https://github.com/voxel51/fiftyone/blob/cf0a75d9ec5352db8027a105bcd300c54fab0d53/fiftyone/core/dataset.py#L176) be used?

benjaminpkane:
The default DB service runs without auth enabled so admin calls are fine. Auth is also noted [here](https://voxel51.com/docs/fiftyone/user_guide/config.html#configuring-a-mongodb-connection) for URI users.

fiftyone/core/odm/database.py
j053y: That's the one I thought @brimoor was talking about. See my first comment in this thread about circular dependencies.

benjaminpkane: Ok, I'll take a look

brimoor: Good catch @benjaminpkane: yes, `Dataset.delete()` really does need to be called.

@j053y `fiftyone.core.odm.database.delete_dataset()` isn't intended to be a "public" method, more of a last resort utility to (mostly) get a corrupted dataset out of the user's deployment

j053y: @brimoor I don't see anyway to use `Dataset.delete()` with major refactoring due to the circular import problem. Also I'm not currently using `fiftyone.core.odm.database.delete_dataset()`

brimoor: I brought up `fiftyone.core.odm.database.delete_dataset()` just to clarify that its logic is equivalent to your logic in `_autodelete_non_persistent_datasets()`, but that's okay for that method because it is designed to be an internal-only thing.

However, `_autodelete_non_persistent_datasets()` needs to do a full deletion like `Dataset.delete()` does or else users will accumulate orphan content in the DB.

brimoor (approved):
LGTM

PR #1698, opened 2022-04-22
Adding progress bar to CVAT uploads

### LINKED ISSUE #1695: [FR] Add progress bar in CVAT upload
To upload big datasets to CVAT, or datasets with where each sample size is large (4k images), I set the `task_size` and `segment_size` low enough, for example around `100`.  However, during the uploading process, we have no indication of the process, rather than a simple log `Uploading samples to CVAT...` and `Upload complete` in the end.

Simple enough, we can add `ProgressBar` here https://github.com/voxel51/fiftyone/blob/cf0a75d9ec5352db8027a105bcd300c54fab0d53/fiftyone/utils/cvat.py#L4134

as follows:
```
with fo.ProgressBar(total= num_samples, iters_str="batches") as pb:
    for idx, offset in enumerate(range(0, num_samples, batch_size)):
        ...
        pb.update(batch_size)
```
and probably remove [this](https://github.com/voxel51/fiftyone/blob/cf0a75d9ec5352db8027a105bcd300c54fab0d53/fiftyone/utils/cvat.py#L3171) and [this](https://github.com/voxel51/fiftyone/blob/cf0a75d9ec5352db8027a105bcd300c54fab0d53/fiftyone/utils/cvat.py#L3173) log.

ioangatop:

Adds:
- Progress bar in CVAT upload.

Resolve #1695 

A simple feature that affects only the terminal visualization.

        notes for FiftyOne users.

-   [  ] App: FiftyOne application changes
-   [  ] Build: Build and test infrastructure changes
-   [  ] Core: Core `fiftyone` Python library changes
-   [  ] Documentation: FiftyOne documentation changes
-   [ x ] Other

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -3167,11 +3167,7 @@ class CVATBackend(foua.AnnotationBackend):
 
     def upload_annotations(self, samples, launch_editor=False):
         api = self.connect_to_api()
-
-        logger.info("Uploading samples to CVAT...")
         results = api.upload_samples(samples, self)
-        logger.info("Upload complete")
-
         api.close()
 
         if launch_editor:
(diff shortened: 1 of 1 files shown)


ehofesmann (changes requested):
Thanks for the contribution. This looks great, I just have a few minor formatting recommendations, and then it should be good to merge!

ioangatop:
@ehofesmann great thank you!

ioangatop:
@ehofesmann did some formatting changes, way simpler than before! Let me know if prefer it this way

ehofesmann:
> @ehofesmann did some formatting changes, way simpler than before! Let me know if prefer it this way

Nice find on the `pb.start()` method! Unfortunately, the `ProgressBar` is designed as a context manager meaning it needs to be called using the syntax `with fou.ProgressBar(...) as pb:`, otherwise the `pb.close()` method may not be called correctly resulting in errors.

Basically, the way you had it originally was good. 

Something that would be nice is to have the progress bar only print if there is more than one batch being uploaded. In many cases, only a single batch of images is being used and the progress bar will not actually update until the end. For single batches, it should probably just print out `"Uploading samples to CVAT...."` without printing the bar.

brimoor:
> Something that would be nice is to have the progress bar only print if there is more than one batch being uploaded. In many cases, only a single batch of images is being used and the progress bar will not actually update until the end. For single batches, it should probably just print out `"Uploading samples to CVAT...."` without printing the bar.

^the cleanest way to implement this would be to add it to the `ProgressBar` class, which is [implemented in a different package](https://github.com/voxel51/eta/blob/022d37d7d04dbf24917f095af247cb43e284e018/eta/core/utils.py#L1433). Instead, perhaps a good compromise is to set `iters_str="tasks"` so that the bar at least conveys that the progress was trivial because it was tracking task creation and there was only one task (the default for image datasets, which is the most common use case).

I imagine you'd like a progress bar at download time too, right? That would be [added here](https://github.com/voxel51/fiftyone/blob/78658ef1896f4339cfefb588c5bbe4eef22e87ad/fiftyone/utils/cvat.py#L4266) 🤗

ioangatop:
@ehofesmann makes sense! I went back to the first style, and now the bar is enabled only when there are multiple batch sizes! I thought that the log `Uploading samples to CVAT....` should be printed also when the pbad is printed, as the first batch can take some time, and the user would have an indication that the program didn't lag. However, we do have the option to print the log only when the pbar is quiet. Let me know!

@brimoor I would prefer the `iters_str="samples"`, it would calculate the upload per image, which I believe its more insightful, but also the user sees how many images he will upload in total, regardless of the batch size, which also serves as a hint to the user that he is uploading the correct data

Now when there is a single batch, it will print:
```
Uploading samples to CVAT...
```

and when there are multiple:
```
Uploading samples to CVAT...
 100% |█████████████| 10/10 [5.7s elapsed, 0s remaining, 1.8 samples/s] 
```

ehofesmann:
Thanks for the updates! I agree, I also prefer always printing `Uploading samples to CVAT...` but only providing a progress bar for `samples` when there are multiple batches. 

There are a couple more formatting suggestions, but you can resolve those automatically by using our pre-commit hooks described [here](https://github.com/voxel51/fiftyone/blob/develop/CONTRIBUTING.md#developing-changes-to-fiftyone). You just need to install them from inside your cloned repository with:
```
pre-commit install
```

Note that there were recently some updates to the versions of these pre-commit hooks, so you may need to merge `voxel51/develop` into your branch first before installing them.

ioangatop:
@ehofesmann  great, thank you! And thank you very much for the comments and for contributing to this great tool!
I hope I can contribute more as we move forward!

PR #1699, opened 2022-04-22
Non-persistent dataset cleanup tweak

brimoor:
See https://github.com/voxel51/fiftyone/pull/1697.

### DIFF
--- a/fiftyone/__public__.py
+++ b/fiftyone/__public__.py
@@ -5,13 +5,11 @@ FiftyOne's public interface.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-import fiftyone.core.config as foc
-import fiftyone.core.odm as foo
+import fiftyone.core.config as _foc
 
-config = foc.load_config()
-annotation_config = foc.load_annotation_config()
-app_config = foc.load_app_config()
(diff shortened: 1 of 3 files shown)


fiftyone/__public__.py
```diff
@@ -5,13 +5,11 @@
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-import fiftyone.core.config as foc
-import fiftyone.core.odm as foo
+import fiftyone.core.config as _foc
```
brimoor, line 8: `_foc` is just to avoid polluting the `fiftyone` package namespace (should have done it earlier)

fiftyone/core/odm/database.py
```diff
@@ -172,6 +172,8 @@ def establish_db_conn(config):
             % (config.type, foc.CLIENT_TYPE)
         )
 
+    _delete_non_persistent_datasets_if_necessary()
```
brimoor, line 175: Important to do this last because https://github.com/voxel51/fiftyone/pull/1692 added a check to this method that may reject certain client connections

j053y (approved):
LGTM

PR #1700, opened 2022-04-22
[WIP] Add Notebook Zoo

ehofesmann:
Goal: Add a zoo for notebooks that can be accessed directly through FiftyOne, easily utilize and pass variables from Python scripts/sessions to a zoo notebook, and make it easily extensible for community contributions.

This PR currently introduces a testing notebook that can be launched through a Python session that accepts a view and label field to be used in the notebook. The idea is that a notebook is able to specify any arguments that it needs and they can be passed directly to `foz.launch_notebook()`. Variables are passed using the built-in IPython [store magic](https://ipython.readthedocs.io/en/stable/config/extensions/storemagic.html?highlight=store), serializing variables into a database accessible across IPython sessions. Any thoughts/concerns are greatly appreciated.

## Example

```python
import fiftyone.zoo as foz

view = dataset[1:15]
field = "ground_truth"

foz.launch_notebook(
    "test",
    view=view,
    field="ground_truth",
)
```

This automatically launches this notebook which can then load the `view` and `field` variables that were provided:
image

Main TODOs:

* Finalize how the interaction between FiftyOne and a notebook should look
* Automatically serialize/deserialize views in a single function with other variables
* Add descriptions for variables required by a notebook to know what to pass into `launch_notebook()`
* Figure out how to best accept notebooks from a URL (or github repo?) and parse it into the manifest
* See how this works for remote workflows

### DIFF
--- a/fiftyone/zoo/__init__.py
+++ b/fiftyone/zoo/__init__.py
@@ -6,3 +6,4 @@ voxel51.com
 """
 from .datasets import *
 from .models import *
+from .notebooks import *
--- a/fiftyone/zoo/notebooks/__init__.py
+++ b/fiftyone/zoo/notebooks/__init__.py
@@ -1,0 +1,31 @@
+"""
+The FiftyOne Notebook Zoo.
+
+| Copyright 2017-2022, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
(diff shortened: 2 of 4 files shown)


ehofesmann:
This is no longer necessary since plugins are available

PR #1711, opened 2022-04-27
[WIP] Fields migration

brimoor:
This PR contains the cherry-picked commits from https://github.com/voxel51/fiftyone/pull/1666 with a migration update that only records default embedded document fields (not dynamic ones).

I included the `Metadata` -> `ImageMetadata`/`VideoMetadata` upgrade because I think it is a good thing to do.


benjaminpkane:
Merged into #1713. Thanks!

PR #1713, opened 2022-04-28
Routing and Sidebar

### LINKED ISSUE #1232: [FR] Opening the expanded view during grid selection

### LINKED ISSUE #1337: [FR] App gets stuck when unplayable video is encountered

### LINKED ISSUE #1347: [BUG] Pressing ESC when viewing a single frame clip in the App enters a broken state

### LINKED ISSUE #1419: [FR] Show error stack traces in the App

### LINKED ISSUE #1503: [BUG] App cannot render empty heatmaps

### LINKED ISSUE #1509: [BUG] The epoch incorrectly renders as 0 in the App

### LINKED ISSUE #1521: [BUG] Gracefully handle attempts to load a deleted dataset in the App

### LINKED ISSUE #1731: [BUG] Keypoint skeleton not showing in the App for videos
When viewing a video in the App that has keypoints, the skeleton is not drawn when specified through `dataset.skeletons`, only if it is provided through `dataset.default_skeleton`.

## Example
```python
import fiftyone as fo
import fiftyone.zoo as foz

img_dataset = foz.load_zoo_dataset("quickstart", max_samples=1).select_fields().clone()
vid_dataset = foz.load_zoo_dataset("quickstart-video", max_samples=1).select_fields().clone()

skeleton = fo.KeypointSkeleton(
    labels=["point1", "point2"],
    edges=[[0,1]],
)

img_dataset.skeletons["kps"] = skeleton
vid_dataset.skeletons["kps"] = skeleton

kps = fo.Keypoints(
    keypoints=[
        fo.Keypoint(label="test", points=[[0.5,0.5], [0.7,0.7]])
    ],
)

sample = img_dataset.first()
sample["kps"] = kps
sample.save()

sample = vid_dataset.first()
sample.frames[1]["kps"] = kps
sample.save()

session = fo.launch_app(img_dataset)
# Skeleton is drawn for image dataset

session.dataset = vid_dataset
# No skeleton for video dataset

vid_dataset.default_skeleton = skeleton
session.refresh()
# Skeleton is drawn for video dataset
```

benjaminpkane:
### Features

* Interactive Sidebar (dragging, grouping)
* Path filtering via `EmbeddedDocument` schemas, e.g. `metadata.size_bytes`
* Routed datasets: `/datasets/:dataset-name`
* Improved App error handling with tracebacks

### Changes

* Server sent events (SSE) in lieu of websockets (`/events` route)
* New lightweight `Session` events framework (to be replaced with `GraphQL` subscriptions)
* New [`starlette`](https://www.starlette.io/) ASGI server
* Initial GraphQL API release (transitional work) with [`strawberry-graphql`](https://strawberry.rocks/) server and [`react-relay`](https://relay.dev/) client
* React 18
* Custom router designed for the "Render-as-You-Fetch" and `react-relay`
* Drops Python 3.6 support

### Issues

* Resolves #1521
* Resolves #1509
* Resolves #1503
* Resolves #1419
* Resolves #1347
* Resolves #1337
* Resolves #1232
* Resolves #1731

### TODO
* [x] Tests
* [x] Finish new session events (e.g. sample selection)
* [x] Test and update notebook support (e.g. screenshots)
* [x] Test and update desktop App support
* [x] Update relevant documentation
* [x] Address sidebar comments
* [x] Keypoints
* [x]  App bugs
* [x] Colab
* [x] Windows


benjaminpkane:
I've undrafted this work. Though there are still some outstanding details. See TODOs

benjaminpkane:
I encourage anyone with time to monkey test these updates. Specifically sessions, sessions in notebooks, desktop, and any and all field/label types in the app.

brimoor:
Make sure you update the README badges, docs, etc to reflect the min Python 3.7 requirement (just came to mind when reading the description here, didn’t have a chance to grep the code)

benjaminpkane:
Ok, skeletons for frames work now, but they use the field name and not the path, e.g. `frames.kps` uses the `kps` skeleton.

This is related to `.` not being supported in the dictionary because of MongoDB/mongoengine. I'm making an issue

ehofesmann (approved):
Gotcha, that's fine, the video skeletons are working for me without `frames.`. Everything else LGTM

nirgmultinarity:
> Ok, skeletons for frames work now, but they use the field name and not the path, e.g. `frames.kps` uses the `kps` skeleton.
> 
> This is related to `.` not being supported in the dictionary because of MongoDB/mongoengine. I'm making an issue

Video skeletons only partially work for me without `frames.` now, the labels are shown but the edges are not being drawn.

PR #1714, opened 2022-04-28
IoU utilities

brimoor:
Adds a `fiftyone.utils.iou` module with the following utilities:

- `compute_ious()`: computes the pairwise IoUs between two sets of objects
- `compute_segment_ious()`: computes the pairwise IoUs between two sets of temporal detections
- `compute_max_ious()`: populates an attribute on each label in the given spatial field(s) that records the max IoU between the object and another object in the same sample/frame
- `find_duplicates()`: returns IDs of duplicate labels in the given field of the collection, as defined by labels with an IoU greater than a chosen threshold with another label in the field

The `compute_ious()`, `compute_max_ious()`, and `compute_segment_ious()` utilities are being upgraded from the `fiftyone.utils.eval` module to a more findable place. Also a few enhancements were added:

- `compute_ious()` now avoids double-computations when the IoU matrix is symmetric
- `compute_max_ious()` now has an `other_field` argument that supports computing IoUs between different fields

### Documentation

Updated the "Removing duplicate objects" recipe to include the `find_duplicates()` method.

### Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.iou as foui

# Populate a `max_iou` attribute on each object
foui.compute_max_ious(dataset, "ground_truth", iou_attr="max_iou", classwise=True)

# Retrieve detections that overlap above a chosen threshold
dups_view = dataset.filter_labels("ground_truth", F("max_iou") > 0.75)
```

Option 1: manually tag duplicates in the App:

```py
session = fo.launch_app(view=dups_view)

# Add 'duplicate' tags to labels in the App...

dataset.delete_labels(tags="duplicate")
```

Option 2: use `find_duplicates()` to automatically tag and remove duplicates:

```py
dup_ids = foui.find_duplicates(
    dataset, "ground_truth", iou_thresh=0.75, method="greedy", classwise=True
)

dataset.select_labels(ids=dup_ids).tag_labels("duplicate")
session.view = dataset.match_labels(ids=dup_ids)

# Remove any spurious `duplicate` tags in the App...

dataset.delete_labels(tags="duplicate")
```

Option 3: send duplicates to CVAT for action:

```py
anno_key = "remove_dups"
dups_view.annotate(anno_key, label_field="ground_truth", launch_editor=True)

# Remove duplicates in CVAT...

# Verify that duplicates no longer exist in FiftyOne
session = fo.launch_app(view=dups_view)
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,9 @@
 repos:
   - repo: https://github.com/asottile/blacken-docs
-    rev: v1.12.1
+    rev: v1.12.0
     hooks:
       - id: blacken-docs
-        additional_dependencies: [black==22.3.0]
+        additional_dependencies: [black==21.12b0]
         args: ["-l 79"]
         exclude: ^docs/theme/|\.rst$
   - repo: https://github.com/ambv/black
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,13 @@
 repos:
   - repo: https://github.com/asottile/blacken-docs
-    rev: v1.6.0
+    rev: v1.12.1
     hooks:
       - id: blacken-docs
-        additional_dependencies: [black==19.10b0]
+        additional_dependencies: [black==22.3.0]
         args: ["-l 79"]
         exclude: ^docs/theme/|\.rst$
   - repo: https://github.com/ambv/black
-    rev: 19.10b0
+    rev: 22.3.0
     hooks:
       - id: black
         language_version: python3
--- a/app/packages/app/src/components/Actions/Options.tsx
+++ b/app/packages/app/src/components/Actions/Options.tsx
@@ -45,39 +45,43 @@ export const RefreshButton = ({ modal }) => {
 };
 
 const ColorBy = ({ modal }) => {
-  const [colorByValue, setColorByValue] = useRecoilState(
-    selectors.appConfigOption({ modal, key: "color_by_value" })
+  const [colorBy, setColorBy] = useRecoilState<string>(
+    selectors.appConfigOption({ modal, key: "color_by" })
   );
 
   return (
     <>
       <PopoutSectionTitle>Color by</PopoutSectionTitle>
+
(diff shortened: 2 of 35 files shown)


ehofesmann (commented):
This is a great addition! I love the new recipe. Everything looks pretty good, I just have a few thoughts:

When trying out the simple vs greedy approaches, the greedy method did not take much longer on the COCO validation split, even after doubling the number of labels and using `classwise=False`, it may be worth setting it to the default method.

Recipe:
* Instead of saying "duplicate labels" use "duplicate object labels" or even "duplicate objects", since this only applies to detections/polygons
* Maybe add another sentence to the intro about what IoU is and how you will be using it to detect overlapping objects, potentially link to  [our blog post](https://medium.com/p/45a511185be1)
* Move the `dataset.untag_labels("duplicate")` cell to the previous section so it doesn't get in the way if someone only follows the automatic duplicate detection section

brimoor:
> This is a great addition! I love the new recipe. Everything looks pretty good, I just have a few thoughts:
> 
> When trying out the simple vs greedy approaches, the greedy method did not take much longer on the COCO validation split, even after doubling the number of labels and using `classwise=False`, it may be worth setting it to the default method.

I'm expecting the simple method to become relatively faster than greedy once we optimize the data model I/O. Let's wait and see if that holds true before switching the default here.

> 
> Recipe:
> 
> * Instead of saying "duplicate labels" use "duplicate object labels" or even "duplicate objects", since this only applies to detections/polygons

^good call, done!

> * Maybe add another sentence to the intro about what IoU is and how you will be using it to detect overlapping objects, potentially link to  [our blog post](https://medium.com/p/45a511185be1)

^good call, done!

> * Move the `dataset.untag_labels("duplicate")` cell to the previous section so it doesn't get in the way if someone only follows the automatic duplicate detection section

^this method won't raise an error if there are no labels with that tag, so I think it's okay to leave here

ehofesmann (approved):
LGTM!

Issue #1724, opened 2022-05-03
[FR] Add a database save context


brimoor:
## Background

A common FiftyOne workflow is to iterate over a collection, modify the sample, and save the result.

This is currently achieved via code like this:

```py
# collection may be a Dataset or a DatasetView

for sample in collection:
    sample["somefield"] = f(sample)
    sample.save()
```

The `for` loop reads samples from the database using a single cursor (resuming from the last read sample if the iteration takes so long that the cursor times out), which means that the read operation is relatively efficient.

However, each `sample.save()` call results in a new database connection, which presents a performance bottleneck.

For large datasets, minimizing the runtime of the above pattern is a high priority.

Currently, the **read operation** can be optimized by using `select_fields()` to limit which fields of the dataset are read from the database:

```py
# Suppose f() only needs `required_field`  
for sample in collection.select_fields("required_field"):
    sample["somefield"] = f(sample)
    sample.save()
```

Currently, the only way to optimize the **write operations** is to leverage `values()` and `set_values()` to "vectorize" the code:

```py
# One db connection
inputs = collection.values("required_field")

outputs = [f(value) for value in inputs]

# One db connection
collection.set_values("somefield", outputs)
```

with the limitation that `inputs` and `outputs` must fit in memory.

## Feature request

An alternative to the above for optimizing the **write operations** would be to introduce what we might call a database "save context" that would defer save operations until a sufficiently large batch is accumulated and then execute the write using bulk operations:

```py
with collection.save_context(batch_size=100):
    for sample in collection:
        sample["somefield"] = f(sample)
        sample.save()
```

The above would decrease the number of database connections by 100x, for example. Some benchmarking would need to be done, but I suspect this will significantly decrease the runtime of the loop.

Syntactic sugar could be added to merge the above with the for loop:

```py
for sample in collection.save_context(batch_size=100):
    sample["somefield"] = f(sample)
    sample.save()
```

Or the save context could be added to the existing `iter_samples()` method:

```py
# Render a progress bar and batch save operations!
for sample in collection.iter_samples(progress=True, batch_size=100):
    sample["somefield"] = f(sample)
    sample.save()
```

## Discussion points

Should we allow `sample.save()` to be omitted, with the understanding that the save context implicitly implies that all changes will be saved?

```py
for sample in collection.save_context(batch_size=100):
    sample["somefield"] = f(sample)
```

One concern I have with the above is that users could encounter problems if they decide to remove the save context (thinking of it only as an optimization):

[...]

PR #1727, opened 2022-05-05
Adding a dataset save context

j053y:
This PR adds a dataset "save context" that allows for aggregating sample updates into bulk save operations:

```py
# Existing syntax: no batching
for sample in dataset:
    # Edit sample here
    sample.save()

# Syntax 1
for sample in dataset.iter_samples(autosave=True):
    # Edit sample here

# Syntax 2
with dataset.save_context() as context:
    for sample in dataset:
        # Edit sample here
        context.save(sample)
```

By default, updates are dynamically batched such that database connections occur every 0.2 seconds (this is the same strategy used by `add_samples()`), but this can be customized via the optional `batch_size` kwarg.

### Example usage

Here's the `iter_samples(autosave=True)` syntax:

```py
import random as r
import string as s

import fiftyone as fo
import fiftyone.zoo as foz

def make_label():
    return "".join(r.choice(s.ascii_letters) for i in range(10))

# No save context
    sample.ground_truth.label = make_label()
    sample.save()

# Save in batches of 10
for sample in dataset.iter_samples(progress=True, autosave=True, batch_size=10):
    sample.ground_truth.label = make_label()

# Save every 0.5 seconds
for sample in dataset.iter_samples(progress=True, autosave=True, batch_size=0.5):
    sample.ground_truth.label = make_label()
```

And here's the `save_context()` syntax:

```py
# No save context
    sample.ground_truth.label = label()
    sample.save()

# Save in batches of 10
with dataset.save_context(batch_size=10) as context:
        sample.ground_truth.label = make_label()
        context.save(sample)

# Save every 0.5 seconds
with dataset.save_context(batch_size=0.5) as context:
        sample.ground_truth.label = make_label()
        context.save(sample)
```
   
### Benchmarking

I created a script to gather some time metrics around different batch sizes.

The script is very long running because it works with some bigger datasets and runs tests multiple times. I've have already run it and added the results into the [this Google Sheet](https://docs.google.com/spreadsheets/d/17i-8fxgbuZnb8nms-aXarMLrj1aS-SqD9xxwnXEMFes/edit#gid=474581980).

If we need even more speed increases, I suggest we create/optimize a way of getting Document diffs, but in the spirit of incremental progress and keeping tasks small I think that work should be done somewhere else and not included in this PR.

```py
import statistics
import time
import uuid

import fiftyone as fo

SAMPLE_COUNTS = [80, 400, 2000, 10000, 50000]
BATCH_SIZES = [1, 10, 25, 50, 100, 1000]
NUMBER_OF_TESTS_PER_BATCH = 10

results = {}
for sample_count in SAMPLE_COUNTS:
    dataset_name = f"samples:n={sample_count}"

    try:
        dataset = fo.load_dataset(dataset_name)
        print(f"Loaded DATASET[{dataset_name}]")
    except ValueError:
        dataset = fo.Dataset(dataset_name, persistent=True)
        print(f"Created DATASET[{dataset_name}]")
        print(f"Adding samples to DATASET[{dataset_name}]...")
        dataset.add_samples(
            [
                fo.Sample(filepath=f"/some/path/dummy{i}.png")
                for i in range(sample_count)
            ]
        )

    results[dataset_name] = {}
    for size in BATCH_SIZES:
        durations = []
        with fou.ProgressBar(
            total=NUMBER_OF_TESTS_PER_BATCH, iters_str="tests"
        ) as pb:
            print(f"Running tests where BATCH SIZE={size}...")

            for i in pb(range(NUMBER_OF_TESTS_PER_BATCH)):
                start = time.time()

                for sample in dataset.iter_samples(autosave=True, batch_size=size):
                    sample[
                        "somefield"
                    ] = f"{dataset_name};size={size};{uuid.uuid4()}"

                durations.append(round(time.time() - start, 4))

        results[dataset_name][size] = durations


for dataset_name, resultsByBatchSize in results.items():
    print()
    print(f"DATASET[{dataset_name}] ".ljust(79, "="))
    for batch_size, durations in resultsByBatchSize.items():
        print(f"batch_size={batch_size}:")
        print(f" - Durations : {', '.join([str(d) for d in durations])}")
        print(f" - Mean      : {round(statistics.mean(durations), 4)}")
        print(f" - Median    : {round(statistics.median(durations), 4)}")
        print(f" - Min       : {round(min(durations), 4)}")
        print(f" - Max       : {round(max(durations), 4)}")
```


fiftyone/core/dataset.py
```diff
@@ -4572,6 +4620,61 @@ def _reload_docs(self, hard=False):
...
+            batch_size: number of `sample.save()` to aggregate into
+                one bulk call
+
+        Returns:
+            :class:`fiftyone.core.dataset.Dataset`
+        """
+        self._sample_save_batch_size = batch_size
+        return self
+
+    def batch_save_sample(self, sample):
```
brimoor, line 4662: Suggestion: make this a private method because we don't intend *users* to directly use it.

Feel free to object to this if you want, but we currently use private methods for everything that a user won't call, including methods that are called by other classes, like this one here.

fiftyone/core/sample.py
```diff
@@ -464,6 +464,9 @@ def save(self):
                 "Cannot save a sample that has not been added to a dataset"
             )
 
+        if self._dataset.allow_sample_save_batching:
+            return self._dataset.batch_save_sample(self)
```
brimoor, line 468: If you track down how the document is saved in non-batching mode, you get to this code:

https://github.com/voxel51/fiftyone/blob/1e276f32d2b9b9242f3a6fde3f6a8bfa37aefa4c/fiftyone/core/odm/document.py#L359-L478

Does `doc._get_update_doc()` do everything that's done there in terms of clearing changed fields, etc?

Side note: I believe this code was copied from MongoEngine into our repo by a previous engineer. I don't know why, but presumably they wanted to tweak something about the behavior? Not sure...

fiftyone/core/sample.py
```diff
@@ -464,6 +464,9 @@ def save(self):
...
+
```
brimoor, line 469: `self.frames.save()` still needs to be called here, or else any edits to the frames of video samples won't be saved here!

brimoor (changes requested):
Good start here!

It occurs to me that #1724 should have been more (or less, really) specific: we'd like to support save contexts for `DatasetView`s too, not just `Dataset`.

As the examples below demonstrate, users are probably even more likely to be iterating over views than entire datasets:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Example 1
# Since I don't need to read any fields for this operation, I include `select_fields()` to optimize reads
# But, this means I'm iterating over a view, but FiftyOne is designed to let me do that without me having to care about the distinction
for sample in dataset.select_fields().iter_samples(progress=True, batch_size=10):
    sample["hello"] = "world"
    sample.save()

# Example 2
# Here I'm slicing, so again I'm working with a view
for sample in dataset[:100].iter_samples(progress=True, batch_size=10):
    sample["spam"] = "eggs"
    sample.save()
```

Also, note that, when working with video samples, there are also `Frame` objects in play, which can be saved individually via `frame.save()` or at the sample-level via `sample.frames.save()`, or by just relying on `sample.save()` to catch all sample and frame-level changes.

```py
import fiftyone as fo

sample.frames[1] = fo.Frame()
sample.frames[2] = fo.Frame()

dataset = fo.Dataset()
dataset.add_sample(sample)

# Option 1: rely on `sample.save()` to capture all sample and frame-level edits
dataset1 = dataset.clone()
for sample in dataset1
    sample["foo"] = "bar"
    for frame in sample.frames.values():
        frame["foo"] = "bar"

    sample.save()

# Option 2: call `frame.save()` on each frame individually
dataset2 = dataset.clone()
for sample in dataset2:
    for frame in sample.frames.values():
        frame["foo"] = "bar"
        frame.save()

# Option 3: call `sample.frames.save()` to save all frame edits
dataset3 = dataset.clone()
for sample in dataset3:
    for frame in sample.frames.values():
        frame["foo"] = "bar"
    
    sample.frames.save()
```

I think the database save context should capture `Frame.save()` and `Frames.save()` events as well.

brimoor:
I think some unit tests involving the new save context are called for here in order to be sure that everything is behaving as expected (eg image and video datasets).

brimoor:
Okay one more comment: when adding a feature to the public API, please add documentation for it to the [User Guide section of the docs](https://voxel51.com/docs/fiftyone/user_guide/index.html). In this case, the relevant pages are probably:
- Dataset save contexts: https://github.com/voxel51/fiftyone/blob/develop/docs/source/user_guide/using_datasets.rst
- View save contexts: https://github.com/voxel51/fiftyone/blob/develop/docs/source/user_guide/using_views.rst

fiftyone/core/dataset.py
j053y: There is a bit of code smell to me, other classes should respect public/private methods (even if we wrote them). While maybe not too bad in this case, following this pattern creates a tight coupling between classes that should be solved by abstraction layers.

fiftyone/core/sample.py
j053y: It was just a method I found that seemed to fit the need. If there's a preferred way to get the update doc I'm open to it.

brimoor: I don't have a "preference" per se, just pointing out that different code is being run to generate the updates now depending on code path

fiftyone/core/dataset.py
brimoor: Yeah some refactoring would definitely be required to satisfy both of these design constraints simultaneously:
(a) All public methods are intended to be user-callable
(b) Classes don't call other classes' private methods

Currently, (a) is achieved at the expense of (b) in some cases.

j053y: Fair enough. I can make it private for consistency.

fiftyone/core/sample.py
j053y: I see now that to save for frames and views I'm going to have to change my approach. I might do a little abstracting on this class to be able to pull out the guts of a save call that way I can aggregate them.

brimoor:
@j053y ready for another review pass here?

convention: we use the "re-request review" option to prompt this

j053y:
I've taken a new approach to accomplish this. the `iter_samples` method now accepts an `autosave` variable. This allows anything sample or frame under the iteration of a dataset or view to be automatically saved (in a batch).

```py
import fiftyone as fo


DEFAULT_DATASET_NAME = "autosave-dataset"
VIDEO_DATASET_NAME = "autosave-dataset-video"


default_dataset = fo.Dataset(DEFAULT_DATASET_NAME)
default_dataset.add_samples(
    [fo.Sample("some/path/to/file") for _ in range(10)]
)

video_dataset = fo.Dataset(VIDEO_DATASET_NAME)
for _ in range(10):
    for i in range(1, 11):
        sample.frames[i] = fo.Frame()
    video_dataset.add_sample(sample)

key, value = "hello", "dataset"
for sample in default_dataset.iter_samples(autosave=True):
    sample[key] = value


key, value = "helloselect", "datasetviewselect"
for sample in default_dataset.select_fields().iter_samples(autosave=True):
    sample[key] = value


key, value = "helloindex", "datasetviewindex"
for sample in default_dataset[:100].iter_samples(autosave=True):
    sample[key] = value

key, sample_value, frame_value = "hellovideo", "sample", "frame"
for sample in video_dataset.iter_samples(autosave=True):
    sample[key] = sample_value
    for frame in sample.frames.values():
        frame[key] = frame_value
```

brimoor (approved):
LGTM

PR #1728, opened 2022-05-05
Fixing GeoTIFF list bug

brimoor:
Previously an error would occur when passing a list of GeoTIFF images to the `image_path` argument of `GeoTIFFDatasetImporter`. Now this works as expected:

```py
import fiftyone as fo

    image_path=["/path/to/image.tif"],
    dataset_type=fo.types.GeoTIFFDataset,
)
```


brimoor:
@voxel51/developers I'd like to include this in this week's release.

benjaminpkane (approved):
LGTM

PR #1736, opened 2022-05-10
Avoid non-critical errors in startup services

brimoor:
Every so often, a user will report that they've managed to get a dataset with `name==None` into their database, despite the code's best efforts to prevent this from occurring. This currently results in an error:

```
In fiftyone/core/dataset.py", line 4591, in _list_datasets:
    return sorted(conn.datasets.distinct("name"))
TypeError: '<' not supported between instances of 'str' and 'NoneType'
```

because `sorted()` does not like None. This PR patches that.

More generally, this PR throws a try-except around everything in the non-persistent dataset cleanup routine that runs at import time, because errors that prevent `import fiftyone as fo` from succeeding are catastrophic! The user has no access to the rest of the library to potentially fix any non-critical errors that exist.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -4583,19 +4583,17 @@ def _get_random_characters(n):
 def _list_datasets(include_private=False):
     conn = foo.get_db_conn()
 
+    if include_private:
+        query = {}
+    else:
+        # Datasets whose sample collections don't start with `samples.` are
+        # private e.g., patches or frames datasets
+        query = {"sample_collection_name": {"$regex": "^samples\\."}}
+
     # We don't want an error here if `name == None`
     _sort = lambda l: sorted(l, key=lambda x: (x is None, x))
 
-    if include_private:
-        return _sort(conn.datasets.distinct("name"))
-
-    # Datasets whose sample collections don't start with `samples.` are private
-    # e.g., patches or frames datasets
-    return _sort(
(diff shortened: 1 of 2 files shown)


fiftyone/core/dataset.py
```diff
@@ -4583,12 +4583,15 @@ def _get_random_characters(n):
...
-        return sorted(conn.datasets.distinct("name"))
+        return _sort(conn.datasets.distinct("name"))
 
     # Datasets whose sample collections don't start with `samples.` are private
     # e.g., patches or frames datasets
-    return sorted(
+    return _sort(
         conn.datasets.find(
             {"sample_collection_name": {"$regex": "^samples\\."}}
         ).distinct("name")
```
j053y, line 4597: Nitpick. I prefer to avoid repeated code in situations like this.
```py
    query = {} # {"name": {"$ne": None}} if you want to let Mongo do your filtering
    if not include_private:
        # Datasets whose sample collections don't start with `samples.` are private
        # e.g., patches or frames datasets
        query["sample_collection_name"] = {"$regex": "^samples\\."}}    

    # We don't want an error here if `name == None`    
    return sorted(
        conn.datasets.find(query).distinct("name"),
        key=lambda x: (x is None, x) # remove this line if you let Mongo filter
    )
```

j053y (approved):
LGTM

PR #1739, opened 2022-05-10
update pandas for arm64 compatibility

findtopher:

The current version of pandas is incompatible with arm64 builds - this change makes `install.bash` work without further changes.

Bump `pandas` from 1.1.5 (December, 2020) to 1.4.2 (April, 2022)
Bump `pytz` from 2019.3 to 2022.1 (`pandas` dependency)

Built on Ubuntu 20.4 locally, and Apple M1 Pro running 12.3.1 (Monterey) - basic local functionality tested

        notes for FiftyOne users.

Upgrades `pandas` from 1.1.5 to 1.4.2

-   [ ] Other

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -12,12 +12,12 @@ mongoengine==0.20.0
 motor==2.3.0
 ndjson==0.3.1
 packaging==20.3
-pandas==1.1.5
+pandas==1.4.2
 plotly==4.14.3
 pprintpp==0.4.0
 psutil==5.7.0
 pymongo==3.11.0
-pytz==2019.3
+pytz==2022.1
 PyYAML==5.4
 scikit-learn>=0.23.2
 scikit-image>=0.16.2


findtopher:
Would like a couple of opinions on this change before I roll it in.

benjaminpkane:
We currently support Python 3.6+ (dropping 3.6 next release)

Looks like install might fail for Python 3.6/3.7. Note that workflows don't use requirements, so only install script users would have issues

brimoor (commented):
Ah okay, the requirements here are in `requirements/common.txt`, not in `setup.py`.

The CI tests are passing, but that's not a surprise because the workflow installs via the `setup.py` requirements, which don't have any pinned versions:

https://github.com/voxel51/fiftyone/blob/4035cf729e01ab76d9cf84df616055a4fc971a34/.github/workflows/build.yml#L121-L130

https://github.com/voxel51/fiftyone/blob/4035cf729e01ab76d9cf84df616055a4fc971a34/setup.py#L31-L67

brimoor:
@voxel51/voxel51-developers does anyone have any recommendations on how to best handle dependency versions?

Here's how we currently do things:
- `setup.py`: any explicitly known version requirements for non-developer installs, wheel builds, etc.
- `requirements/*.txt`: we tag exact dependency versions to ensure reproducibility between developer environments

Developer installs use `install.bash`, which installs requirements from the pinned versions:

https://github.com/voxel51/fiftyone/blob/4035cf729e01ab76d9cf84df616055a4fc971a34/install.bash#L124-L132

Don't want to reinvent the wheel here, just want to know if anyone has a best practice in mind. The current model would just be to upgrade pinned versions every time a new developer comes along that needs a different min package version (eg Apple Silicon).

brimoor (approved):
LGTM

PR #1745, opened 2022-05-12
Black update

brimoor:
#1703 updated the `black` version used by our pre-commit hooks. The newer version prefers a different line break strategy that has been causing pesky cosmetic diffs in PRs since #1703.

This PR prophylactically applies the updated pre-commit hook to the entire project to prevent extra diffs going forward.


j053y (approved):
LGTM

PR #1746, opened 2022-05-12
Optimizing non-persistent dataset deletion

brimoor:
The previous implementation of `delete_non_persistent_datasets()` required reading every `DatasetDocument` into memory, which was wasteful. Now only non-persistent datasets are loaded.


j053y (approved):
LGTM

Issue #1741, opened 2022-05-11
[BUG] Cannot open CocoDataset in after upgrade from 0.14.4 to 0.15.1 


hovnatan:
### System information

-   **OS Platform and Distribution MacOS M1**:
-   **FiftyOne installed from pip**:
-   **FiftyOne version FiftyOne v0.15.1, Voxel51, Inc.**:
-   **Python version 3.9.12**:

### Describe the problem

I get `fiftyone.core.odm.dataset.DoesNotExist: Document does not exist` on the exact same code after upgrading from 0.14.4 to 0.15.1

```
    coco_dataset = fo.Dataset.from_dir(
        dataset_type=fo.types.COCODetectionDataset,
        dataset_dir="./images_path",
        data_path="images/",
        labels_path="labels.json",
        label_types="segmentations",
   )
   session = fo.launch_app(coco_dataset)
```

### Other info / logs
```
 100% |█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [9.7ms elapsed, 0s remaining, 307.8 samples/s]
Uncaught exception
Traceback (most recent call last):
  File "/Users/hovnatan/run_fifty_one.py", line 10, in <module>
    session = fo.launch_app(coco_dataset)
  File "/Users/hovnatan/fiftyone/fiftyone/core/session.py", line 142, in launch_app
    _session = Session(
  File "/Users/hovnatan/fiftyone/fiftyone/core/session.py", line 339, in __init__
    state.dataset._reload()
  File "/Users/hovnatan/fiftyone/fiftyone/core/dataset.py", line 4558, in _reload
    self._doc.reload()
  File "/Users/hovnatan/venv_mmseg/lib/python3.9/site-packages/mongoengine/document.py", line 746, in reload
    raise self.DoesNotExist("Document does not exist")
fiftyone.core.odm.dataset.DoesNotExist: Document does not exist
```

### What areas of FiftyOne does this bug affect?

-   [ ] `App`: FiftyOne application issue
-   [x] `Core`: Core `fiftyone` Python library issue
-   [ ] `Server`: Fiftyone server issue

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another
member of your organization be willing to contribute a fix for this bug to the
FiftyOne codebase?

-   [ ] Yes. I can contribute a fix for this bug independently.
-   [ ] Yes. I would be willing to contribute a fix for this bug with guidance
        from the FiftyOne community.
[...]

PR #1742, opened 2022-05-11
Use previous "cleanup" method

j053y:
Reverting to old cleanup method for non-persistent datasets. Enabling automatic cleanup for external Mongo connections will require a bigger refactor than originally estimated. 

Fixes #1741

## Testing
Make sure your config has no reference to an external Mongo database.

Run the following code in the terminal twice to ensure non-persistent datasets are cleaned up.
```py
import fiftyone as fo

fo.list_datasets() # []

fo.Dataset('test)
fo.list_datasets() # ["test"]

exit()
```

PR #1747, opened 2022-05-13
Delete non-persistent datasets at shutdown

j053y:
This is an alternate to #1742. 

This allows us to retain the feature of automatically deleting non-persistent datasets. It is done when the last session disconnects from the database as opposed to start-up.

### Test

- Open a new terminal session (Session # 1) and run `python`:

    ```py 
    import fiftyone as fo
    dataset = fo.Dataset('test')
    ```

- Keep Session # 1 running.

- Open another terminal session (Session # 2):

    `fiftyone datasets list  # non-persistent dataset ('test') is listed`

- Exit python in Session # 1.

- Back in Session # 2:

    `fiftyone datasets list  # non-persistent dataset ('test') should no longer be listed`

Should work for both a built-in and external Mongo database.

### Config

If users wish to disable this feature they can do so by adding `disable_automatic_dataset_cleanup` to their config or setting the ENV variable `FIFTYONE_DISABLE_AUTOMATIC_DATASET_CLEANUP`.


j053y:
Fixes https://github.com/voxel51/fiftyone/issues/1741 by not running any cleanup code on start-up. There is the potential for the counter to get out of sync and need to be reset manually.

fiftyone/core/odm/database.py
```diff
@@ -209,39 +210,47 @@ def _async_connect():
...
-            )
-        )
-    except:
-        logger.warning(
-            "Skipping automatic non-persistent dataset cleanup. This action "
-            "requires read access of the 'admin' database"
-        )
+def _track_db_connection(config, client):
+    # Using an env variable so that this only happens once per "main" process
+    if config.disable_automatic_dataset_cleanup or os.environ.get(
```
brimoor, line 215: suggestion: use `os.environ.get("__FIFTYONE_CONN_TRACKER__") != "1")` here to avoid unexpected behavior if someone thinks that they can, for example, set `__FIFTYONE_CONN_TRACKER__ = 0` to have the same effect as unsetting the env var

brimoor (commented):
Hmm, per your comments, this approach does seem risky. `atexit` promises that registered functions will run at "normal termination", but there could be all kinds of abnormal failures that would cause the counter to become permanently positive. My gut tells me that over any sufficient period of time, non-persistent datasets will never be cleaned up with this approach.

The only truly reliable test I can think of as to whether it is safe to cleanup non-persistent datasets at any given time is if MongoDB tells us we are the last active connection. Of course that was your first inclination too, so the question arises: do you have any thoughts about what exactly was going wrong in some environments with #1697?

Idea: what if you put the logic from #1697 into the cleanup function in this PR? This would definitely avoid counter-sync errors, and it's okay if it doesn't run in critical shutdowns because it will run next time 🤞 Also, since this code doesn't run until shutdown (and only in the main process thanks to your env var trick), perhaps any failure cases are more limited?

fiftyone/core/config.py
```diff
@@ -197,6 +197,12 @@ def __init__(self, d=None):
         self.timezone = self.parse_string(
             d, "timezone", env_var="FIFTYONE_TIMEZONE", default=None
         )
+        self.disable_automatic_dataset_cleanup = self.parse_bool(
```
brimoor, line 200: The trouble here is that, in a multiple connection environment, everybody would have to set this to `True` to actually prevent dataset cleanup.

brimoor:
Other options:
1. Revert to #1742. Teams/external DB users must run `delete_non_persistent_datasets()` themselves, but vanilla OSS users will continue to enjoy automatic cleanup
2. Revert to #1742 but include (Teams)/recommend (external DB users) a cron job that handles non-persistent dataset cleanup in a presumably 
3. Keep `persistent`, but completely disable automatic cleanup in all cases. Provide user guidance that "the non-persistent property is simply a promise to yourself that this dataset is not important and can be automatically cleaned up via `delete_non_persistent_datasets()` at a time of your choosing"
4. Get rid of non-persistent datasets altogether. Deletion is fully manually, on a per-dataset basis

fiftyone/core/odm/database.py
j053y: Good catch. I agree specific is better in this case.

fiftyone/core/config.py
j053y: Yeah I was thinking about a single user that might happen to have their own instance of Mongo already running, not a team. But I can remove that no problem.

j053y:
I think it's a good idea and worth trying to combine the approaches of this and https://github.com/voxel51/fiftyone/pull/1697. I'll work to make those changes and see if I can get some consistency.

j053y:
Alright, I've merged the approaches. It seems to be working but I would encourage someone who got the error outlined in https://github.com/voxel51/fiftyone/issues/1741 to try as I could not originally reproduce that.  

@brimoor If this approach does not work, I think we should go with other option #4. If we can't reliably delete non-persistent datasets for all types of users, I don't think they make sense as concept.

brimoor (approved):
LGTM after cleanup

PR #1749, opened 2022-05-14
Cloning/merging bug fixes

brimoor:
Found a couple bugs in `merge_samples()`, `rename_frame_field()`, and `clone_frame_field()`. This PR fixes them.

### Merging samples

```py
import fiftyone as fo
import fiftyone.zoo as foz

small_dataset = dataset[:100].select_fields("ground_truth").clone()
model_predictions = dataset[50:150]

# Previously would raise an error; now works
small_dataset.merge_samples(
    model_predictions,
    fields="predictions",
    insert_new=False,
    include_info=False,
)

assert len(small_dataset.exists("predictions")) == 50
```

### Renaming/cloning frame fields

```py
import fiftyone as fo
import fiftyone.zoo as foz

frame = dataset.first().frames.first()

# Previously wouldn't populate `vehicle_type`, now does
dataset.clone_frame_field(
    "detections.detections.type",
    "detections.detections.vehicle_type",
)
assert (
    frame.detections.detections[0].vehicle_type
    == frame.detections.detections[0].type
)

# Previously wouldn't populate `still_type`, now does
dataset.rename_frame_field(
    "detections.detections.vehicle_type",
    "detections.detections.still_type",
)
assert (
    frame.detections.detections[0].still_type
    == frame.detections.detections[0].type
)

dataset.clear_frame_field("detections.detections.still_type")
assert frame.detections.detections[0].still_type is None

dataset.delete_frame_field("detections.detections.still_type")
assert "still_type" not in frame.detections.detections[0]
```

### DIFF
--- a/tests/unittests/dataset_tests.py
+++ b/tests/unittests/dataset_tests.py
@@ -284,6 +284,17 @@ class DatasetTests(unittest.TestCase):
         common12 = common12_view.first()
         self.assertEqual(common12.field, common2.field)
 
+        # Merge specific fields, no new samples
+
+        dataset1c = dataset1.clone()
+        dataset1c.merge_samples(dataset2, fields=["field"], insert_new=False)
+        self.assertEqual(len(dataset1c), 2)
+        common12_view = dataset1c.match(F("filepath") == common_filepath)
+        self.assertEqual(len(common12_view), 1)
+
+        common12 = common12_view.first()
+        self.assertEqual(common12.field, common2.field)
+
         # Merge a view with excluded fields
 
         dataset21 = dataset1.clone()
--- a/tests/unittests/dataset_tests.py
+++ b/tests/unittests/dataset_tests.py
@@ -1004,61 +1004,61 @@ class DatasetTests(unittest.TestCase):
         dataset = fo.Dataset()
         sample = fo.Sample(
             filepath="image.jpg",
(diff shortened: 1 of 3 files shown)


tests/unittests/dataset_tests.py
```diff
@@ -1004,91 +1015,158 @@ def test_rename_embedded_fields(self):
...
-            "predictions.new_field", "predictions.field"
+            "predictions.detections.field_copy",
+            "predictions.detections.field",
+        )
+        self.assertIsNotNone(sample.predictions.detections[0].field)
+        with self.assertRaises(AttributeError):
+            sample.predictions.detections[0].field_copy
+
+    @drop_datasets
+    @unittest.skip("TODO: Fix workflow errors. Must be run manually")
```
brimoor, line 1107: Not sure why the other tests I edited were being omitted due to Windows errors, but I continued omitting these too since they are effectively the same.

I did, however, manually run these tests locally before making this PR... 😢

benjaminpkane: Regarding skipped tests, any test that is currently skipped is because they randomly fail or completely stall in the workflow. Hopefully we can resolve these soon.

As a best practice going forward, I do think tests should be owned by the contributor for all platforms. "It works on my machine" is an unnecessary distraction for other contributors.

benjaminpkane (approved):
LGTM

PR #1767, opened 2022-05-22
[WIP] Adding examples to public API docstrings

brimoor:
This PR adds examples to the docstrings of all non-trivial public API methods.

Checklist
- [x] `fiftyone.core.dataset.Dataset`
- [ ] `fiftyone.core.view.DatasetView`
- [ ] `fiftyone.core.collections.SampleCollection`
- [ ] `fiftyone.core.sample.Sample`
- [ ] others?

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -434,7 +434,27 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
 
     @property
     def media_type(self):
-        """The media type of the dataset."""
+        """The media type of the dataset.
+
+        The media type of empty datasets is ``None``, and the appropriate type
+        will be automatically inferred based on the ``filepath`` of the first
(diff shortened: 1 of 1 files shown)


mwoodson1 (dismissed):
I left some specific notes, mostly regarding documentation but here is some high level feedback

- The documentation for the `FiftyOneTorchDataset` class is much too long imo and should be relocated to a user guide or tutorial documentation
- A lot of the documentation seems distributed training focused. Some users may use this feature without training on multiple machines, especially our open source users. Documentation for both is important but should be separated.
- There are a lot of gotchas with this feature. The biggest one I see is only mentioned in the `basic_example.ipynb` example stating you need to call `to_torch` again after sampling, otherwise it's not suitable for multiprocessing. Put as many of these into warnings in the code itself or make sure to call all of them out in one place to avoid confusion.

PR #1771, opened 2022-05-24
Add installation of mongodb v6 for experimental apple silicon support

ritch:

Add experimental support for mongodb 6 for apple silicon.

The install has been tested. Changes will still be required to enable the experimental support within fiftyone itself.

        notes for FiftyOne users.

Need to discuss before releasing. Essentially this adds the installation bits for enabling `bash install.bash -m` for a different version of mongodb that supports apple silicon. This would need to be released as experimental until we actually support v6. 

This is mostly a useful utility for development of features unrelated to mongo, which is why I'd like to add it.

For this set of changes, the usage is as follows:

```
bash install.bash -m -x
```

 - the `-m` option to install mongo in the `~/.fiftyone` dir stays mostly the same
 - `-x` enables installing the macos + arm64 v6 rc build of mongodb

Note: this also addresses a bug in the installer, which will throw when attempting to check the version on apple silicon when `-m` is set.

-   [x] Other - install

### DIFF
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -165,13 +165,13 @@ def establish_db_conn(config):
                     "connection to your own MongoDB instance or cluster "
                 )
 
-            if fou.is_arm_mac():
-                raise FiftyOneConfigError(
-                    "MongoDB is not yet supported on Apple Silicon Macs. "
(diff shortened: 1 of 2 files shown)


ritch:
Note "bump matplotlib" is out of scope fore this PR - will move to another.

Issue #1761, opened 2022-05-20
[FR] Add keyboard shortcut to temporarily hide all labels in the App


brimoor:
Often one needs a quick way to hide all labels in the expanded modal so that the image beneath them can be inspected. Currently the only way to achieve this is to uncheck all label fields, but there could be many, and maybe I forget which ones I had selected.

Having the option to hold `SPACE` (for example) to temporarily hide all labels would be super convenient.

**Update**: `SPACE` controls play/pause for videos, so something else may be needed here.

Note that all keyboard shortcuts should be documented under the `?` menu in the HUD.

<img width="1141" alt="Screen Shot 2022-05-20 at 6 29 10 PM" src="https://user-images.githubusercontent.com/25985824/169620531-008afc06-cd9d-4e78-90b7-63a124ed9c15.png">

benjaminpkane:
Resolved by #1779

PR #1779, opened 2022-05-25
Add keyboard shortcut to temporarily hide all labels in the App

ritch:
See background in https://github.com/voxel51/fiftyone/issues/1761.

This feature adds the ability to temporarily hide all overlays/tooltips by pressing the "Shift" key. This is supported while viewing samples in the `Looker` modal image and video modes.

Specifically this adds:

 - a new control in the `Looker`
 - a corresponding button in the `Looker` controls
 - support for [ControlEventKeyTypes](https://github.com/voxel51/fiftyone/pull/1779/files#diff-15d94c8a67ce8f2705adf8b91d31fb9ccd9c57157c787f75de8ad58363a26bbfR61) which allow for specifying which style event your hotkey supports (HOLD = press and hold, KEY_DOWN/DEFALUT = current behavior).
 - a new `Atom` for persisting the show/hide overlay value in recoil
 - glue between the `Looker` and recoil (based on fullscreen)

To implement a control with "hold" supprt, you must specify it's `ControlEventKeyTypes` as `HOLD` as well as add a new method: `control.afterAction()`. This `afterAction` method will be called on `keyup` allowing for the control to make updates after the user depresses the key.

NOTE: in a typical hold event model I would usually debounce the keydown event. I've left that optimization out, if we notice performance issues we can look into adding that or something similar.


app/packages/looker/src/state.ts
```diff
@@ -86,6 +86,7 @@ interface BaseOptions {
   showIndex: boolean;
   showJSON: boolean;
   showLabel: boolean;
+  showLabels: boolean;
```
ritch, line 89: Should probably rename this to `showOverlays`

requirements/common.txt
```diff
@@ -9,7 +9,7 @@ httpx==0.7.7
 hypercorn==0.13.2
 Jinja2==3.0.3
 kaleido==0.2.1
-matplotlib==3.2.1
+matplotlib==3.5.2
```
ritch, line 12: ignore...need to remove

app/packages/app/src/recoil/atoms.ts
```diff
@@ -68,6 +68,11 @@ export const fullscreen = atom<boolean>({
   default: false,
 });
 
+export const showLabels = atom<boolean>({
```
ritch, line 71: I'm not sure if this is needed. Is there another atom / place that allows mutating the options? I'm just following the fullscreen toggle implementation.

app/packages/looker/src/elements/util.ts
```diff
@@ -24,6 +24,8 @@ export const ICONS = Object.freeze({
...
   options:
     "data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24'%3E%3Cpath fill='rgb(238, 238, 238)' d='M12,15.5A3.5,3.5 0 0,1 8.5,12A3.5,3.5 0 0,1 12,8.5A3.5,3.5 0 0,1 15.5,12A3.5,3.5 0 0,1 12,15.5M19.43,12.97C19.47,12.65 19.5,12.33 19.5,12C19.5,11.67 19.47,11.34 19.43,11L21.54,9.37C21.73,9.22 21.78,8.95 21.66,8.73L19.66,5.27C19.54,5.05 19.27,4.96 19.05,5.05L16.56,6.05C16.04,5.66 15.5,5.32 14.87,5.07L14.5,2.42C14.46,2.18 14.25,2 14,2H10C9.75,2 9.54,2.18 9.5,2.42L9.13,5.07C8.5,5.32 7.96,5.66 7.44,6.05L4.95,5.05C4.73,4.96 4.46,5.05 4.34,5.27L2.34,8.73C2.21,8.95 2.27,9.22 2.46,9.37L4.57,11C4.53,11.34 4.5,11.67 4.5,12C4.5,12.33 4.53,12.65 4.57,12.97L2.46,14.63C2.27,14.78 2.21,15.05 2.34,15.27L4.34,18.73C4.46,18.95 4.73,19.03 4.95,18.95L7.44,17.94C7.96,18.34 8.5,18.68 9.13,18.93L9.5,21.58C9.54,21.82 9.75,22 10,22H14C14.25,22 14.46,21.82 14.5,21.58L14.87,18.93C15.5,18.67 16.04,18.34 16.56,17.94L19.05,18.95C19.27,19.03 19.54,18.95 19.66,18.73L21.66,15.27C21.78,15.05 21.73,14.78 21.54,14.63L19.43,12.97Z' /%3E%3C/svg%3E",
+  hideLabels:
```
benjaminpkane, line 27: These SVGs come from this website. 
* https://materialdesignicons.com/

It's a bit messy that icons can from here, or the React app. Open to changing things in the future.

app/packages/looker/src/elements/common/actions.ts
```diff
@@ -142,6 +142,29 @@ export const previous: Control = {
...
+
+export const hideLabels: Control = {
+  title: "Hide labels",
+  shortcut: "h",
+  eventKeys: "h",
+  detail: "Toggles visibility of all labels",
+  action: (update, dispatchEvent) => {
+    update(({ options: { showLabels } }) => {
+      if (showLabels) {
+        dispatchEvent("options", { showLabels: false });
```
benjaminpkane, line 154: So as you may have noticed, there a quite a few options that define a Looker's state.

The word "options` is also overused. Specifically, the "options" event is for AppConfig options which  have configurable defaults that come from the server

* https://github.com/voxel51/fiftyone/blob/499a53e34c58585b9ac0bf0d7d7e6e5b1030aec4/fiftyone/server/query.py#L207
* https://github.com/voxel51/fiftyone/blob/499a53e34c58585b9ac0bf0d7d7e6e5b1030aec4/fiftyone/core/config.py#L270

The default for this should always be true, though.

app/packages/looker/src/processOverlays.ts
```diff
@@ -19,6 +19,7 @@ const processOverlays = <State extends BaseState>(
   let classifications = null;
 
   for (const overlay of overlays) {
+    if (!state.options.showLabels) continue;
```
benjaminpkane, line 22: Feel free to return immediately in this function

benjaminpkane (approved):
LGTM! Looks like HOLD was straightforward to add. Clean solution.

I wouldn't worry about debouncing. Updates are supposed to run in a single animation frame regardless.

brimoor:
Couple things I saw while testing:
- Tooltip still shows `h` as hotkey, should be `shift`
- Agree with Ben that, to me, the icons should be swapped: strikethrough should be shown when the labels are hidden, not when they're visible
- If I click on the new icon, then drag somewhere else (eg up onto canvas), then release, the show labels event is not triggered. I'm now in a state where I'm not holding `shift` or clicking the icon, but the labels are still hidden. Ideally this couldn't happen

ritch:
Fixed the issues mentioned above. This should be ready to go now.

PR #1782, opened 2022-05-25
updating pylint to 2.13.9

neuralsyn:
The existing pylint version mishandles various typing scenarios. For example when using Optional[T] you will receive Value 'Optional' is unsubscriptable. The latest version resolves this issue.


benjaminpkane (approved):
LGTM. Thanks!

PR #1784, opened 2022-05-25
adding missing ipython_genutils dependency

neuralsyn:
Jupyter doesn't pull in ipython_genutils automatically thus producing a module not found error during the docs build.  This explicitly adds it as a dependency to resolve the issue.


benjaminpkane (approved):
LGTM

PR #1785, opened 2022-05-25
v0.16.1 Revision

benjaminpkane:
The intention of the v0.16.0 revision was to add definitions for embedded document field schemas within the dataset document. The revision contains a bug that omits list fields like `Detections`, though.

This v0.16.1 revision accounts for list fields.

### DIFF
--- a/fiftyone/migrations/revisions/v0_16_1.py
+++ b/fiftyone/migrations/revisions/v0_16_1.py
@@ -1,6 +1,10 @@
 """
 FiftyOne v0.16.1 revision.
 
+This revision is identical to v0.16.0 except it does not affect
+app sidebar groups and it fixes label list fields, e.g. Detections,
+which were skipped in v0.16.0
+
 | Copyright 2017-2022, Voxel51, Inc.
 | `voxel51.com <https://voxel51.com/>`_
 |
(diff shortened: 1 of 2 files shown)


fiftyone/migrations/revisions/v0_16_1.py
```diff
@@ -0,0 +1,296 @@
...
+            except Exception as e:
+                print(
+                    "Failed to infer schema of frame field '%s' of type "
+                    "'%s': %s" % (name, embedded_doc_type, e)
+                )
+                field["fields"] = []
+        else:
+            field["fields"] = []
+
+    dataset_dict["app_sidebar_groups"] = None
```
brimoor, line 65: Remove this to avoid squashing any sidebar groups the user created in v0.16.0 :)

fiftyone/migrations/revisions/v0_16_1.py
```diff
@@ -0,0 +1,296 @@
+"""
+FiftyOne v0.16.1 revision.
```
brimoor, line 2: Probably worth adding a comment there that explains what this migration is fixing. Otherwise it will look weird because, the way it is implemented, it seems to duplicate much of the v0.16.0's functionality.

brimoor (approved):
My bad on the failed migration 😢 thanks for patching it!

I can't remember, is it possible in v0.16.0 for the user to get (eg by declaring) non-default subfields into their dataset schema? If so then this migration would silently delete them as written.

PR #1787, opened 2022-05-25
[WIP} proposal for e2e tests

ritch:
So far this is for discussion only. This may drastically change before landing. Using this draft PR for commenting.


benjaminpkane:
Initial impressions are Cypress looks like a great fit. I don't have experience with a test framework like this, but I hope we can keep installation time to a minimum. As an aside, the current workflows are already a bit slow and could use some caching improvements.

Your action items LGTM. I can help next week with adding initial tests or updating workflows/installation if you'd like. Or any other ways I can support.

One thought that came to mind looking at this is whether we should move the current `./app` yarn workspace to the repository root? And just have one yarn workspace. The Python package is defined at the root, I don't see why yarn couldn't coexist there.

ritch:
> One thought that came to mind looking at this is whether we should move the current ./app yarn workspace to the repository root? And just have one yarn workspace. The Python package is defined at the root, I don't see why yarn couldn't coexist there.

Probably. I like the idea of one flat list of node/browser packages. So anything I can do to get to that model, LMK.

That being said this PR is actually going in the wrong direction in some ways:

I thought the style here would all for simpler writing of tests, but I need to go back to the `cy.myCommand()` style. Since cypress has this weird thing about being like jquery for no real benefit AFAICT.

The other thing I'm thinking is pulling in this plugin that will let us target elements via react component names. Lets say you want to click on the `<Looker />`... well just do this:

```js
cy.react('Looker').click()
// or even better - target some more generic component with props:
cy.react('Button', {title: 'Labels'}).click()
```

I'll rework this PR to demonstrate this as well as the cli (and hopefully python) execution. Then we should meet to discuss (anyone else reading along is welcome, just ping me on slack).

e2e/README.md
```diff
@@ -0,0 +1,25 @@
+This directory contains a suite of black box and ui integration tests.
```
benjaminpkane, line 1: I would add steps on how to run locally to the README. Feel free to leave the relevant TODOs

I'm currently running tests with `yarn start` and `yarn cypress run` (in another shell).  The quickstart tests are failing for me right now. Need to look into it more.

e2e/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
benjaminpkane, line 1: Feel free to remove

e2e/cypress/integration/links.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+import TestContext from "../support/TestContext"
+const path = require('path')
+
+let ctx;
+let proc;
+
+let LINK = 'https://voxel51.com/'
+describe('links', () => {
+    before(async () => {
+      const FIXTURES = path.join(__dirname, '..', 'fixtures')
```
benjaminpkane, line 12: I think one image is fine, but the repository is already quite big. Just noting that future tests should dynamically create image data, or use the zoo, etc.

e2e/cypress/integration/links.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        sample = fo.Sample(
+            filepath="${IMG_PATH}",
+            link="${LINK}",
+        )
+        
+        dataset = fo.Dataset()
+        dataset.add_sample(sample)
+        
+        session = fo.launch_app(dataset)
```
benjaminpkane, line 26: The below will prevent browser opens.

```suggestion
        session = fo.launch_app(dataset, remote=True)
```

benjaminpkane (approved):
Thought I'd leave some comments. Fully aware this is still WIP. 

As discussed offline, what has already been added is great! We can finally test the App.

PR #1790, opened 2022-05-26
Fixing App regressions

benjaminpkane:
Fixing a few regressions introduced in v0.16.0

* Fixes expanded samples in clips views which appeared to be empty
* Fixes sort by similarity with a `dist_field`
* Fixes "color by" for simple values (classifications, tags, etc.)
* Fixes changing datasets when sort by similarity is set
* Fixes mask and map coloring
* Fixes fortran array handling for masks and maps


app/packages/app/src/recoil/aggregations.ts
```diff
@@ -132,6 +132,7 @@ export const aggregations = selectorFamily<
     }
 
     const dataset = get(selectors.datasetName);
+    get(aggregationsTick);
```
ritch, line 135: I see this elsewhere in this file, but it's not clear why. Does this force an update or linking of some sort?

benjaminpkane: Yes, it forces an update. It could/should be removed. Just need to check behavior doesn't change.

PR #1793, opened 2022-05-26
adding filename exception logging for failed xml parsing

neuralsyn:
Currently when an invalid or malformed xml file is parsed an `ExpatError` will be raised.  This change logs the filename that produced this error before re-raising it to the calling code.

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -6,6 +6,11 @@ Core utilities.
 |
 """
 import atexit
+from base64 import b64encode, b64decode
+from collections import defaultdict
+from contextlib import contextmanager
+from copy import deepcopy
+from datetime import date, datetime
 import hashlib
 import importlib
 import inspect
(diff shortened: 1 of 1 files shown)


fiftyone/core/utils.py
```diff
@@ -662,8 +666,12 @@ def load_xml_as_json_dict(xml_path):
     Returns:
         a JSON dict
     """
-    with open(xml_path, "rb") as f:
-        return xmltodict.parse(f.read())
+    try:
+        with open(xml_path, "rb") as f:
+            return xmltodict.parse(f.read())
+    except ExpatError as ex:
+        logger.error(f"Failed to read {xml_path}: {ex}")
```
brimoor, line 673: nit: per [style guide](https://github.com/voxel51/fiftyone/blob/develop/STYLE_GUIDE.md#logging) we currently prefer directly raising exceptions to `logger.error()`:

```py
raise ExpatError(f"Failed to read {xml_path}: {e}")
```

fiftyone/core/utils.py
```diff
@@ -19,15 +14,24 @@
...
 import platform
+import posixpath
 import signal
 import struct
 import subprocess
 import timeit
 import types
 import zlib
 
+from base64 import b64decode
```
brimoor, line 26: nit: per [style guide](https://github.com/voxel51/fiftyone/blob/develop/STYLE_GUIDE.md#imports), we currently ignore `import`/`from` for grouping purposes.

IE, import blocks are defined solely based on whether it is builtin/third-party/project, not by import strategy used.

brimoor (approved):
LGTM after addressing nits

PR #1794, opened 2022-05-26
Fix bad format string

robert-perrotta:

The ValueError raised when asking for an unsupported plotting backend was specified in a way that caused the following error: `TypeError: not enough arguments for format string`. I put the pair of format string arguments in a tuple following the example of a similar ValueError on [line 873](https://github.com/voxel51/fiftyone/blob/1a9c9bf0fe3f1a5a332629e2f711ab8d4358a7c4/fiftyone/core/plots/base.py#L873).

I ran the same code that revealed the error (edited from [here](https://voxel51.com/docs/fiftyone/user_guide/evaluation.html#map-and-pr-curves)), and this time saw the correct error message.

```python
import fiftyone as fo
import fiftyone.zoo as foz

print(dataset)

# Performs an IoU sweep so that mAP and PR curves can be computed
    "predictions",
    gt_field="ground_truth",
    compute_mAP=True,
)

print(results.mAP())
# 0.3957

plot = results.plot_pr_curves(backend="bokeh", classes=["person", "kite", "car"])
```

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/plots/base.py
+++ b/fiftyone/core/plots/base.py
@@ -615,8 +615,7 @@ def _parse_backend(backend):
     if backend not in available_backends:
         raise ValueError(
             "Unsupported plotting backend '%s'; supported values are %s"
-            % backend,
-            available_backends,
+            % (backend, available_backends)
         )
 
     return backend


benjaminpkane (approved):
Thanks @robert-perrotta. LGTM!

PR #1801, opened 2022-05-27
FFmpeg App Error

### LINKED ISSUE #1799: [BUG] Video Overlay Aspect Ratio/Zoom/Alignment issue
### Instructions

Thank you for submitting an issue. Please refer to our
[issue policy](https://www.github.com/voxel51/fiftyone/blob/develop/ISSUE_POLICY.md)
for information on what types of issues we address.

**Please fill in this template to ensure a timely and thorough response.**

-   Place an "x" between the brackets next to an option if it applies. Example:
    -   [x] Selected option
-   Please delete this section (all content above this line) before submitting
    the issue

### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
Mac OS 12.4

-   **FiftyOne installed from (pip or source)**:
Seeing this in both.

-   **FiftyOne version (run `fiftyone --version`)**:
0.16.0, also seeing in develop

-   **Python version**:
3.9.13

### Commands to reproduce

```
fiftyone app view --zoo-dataset quickstart-video
```

### Describe the problem

Video labels seem misaligned. Looks like the video is zoomed in.

### Code to reproduce issue

None - just open the app to a video dataset.

### Other info / logs

Screen Shot 2022-05-27 at 8 05 12 AM
Screen Shot 2022-05-27 at 8 17 22 AM

benjaminpkane:
Resolves #1799 

Previously the App server provided fallback metadata. v0.16.0 was supposed add missing FFmpeg errors to the App, but it got lost in a merge conflict, I believe.

Perhaps there is a middle ground where some errors still return fallback metadata?

TODO: add unit test


brimoor (approved):
I verified that the "ffmpeg not found" error is appropriately raised in the App.

Regarding whether placeholder metadata should continue to be used in other situations, I'll approve and defer to you on a final decision. Here's my two cents:

I haven't thought deeply about how many cases there are when fallback metadata being used would result in a usable App experience. Certainly not when the dataset contains spatial labels, but if there aren't spatial labels then "something" may well be better than "nothing" (eg retryable network error). If there are people out there experiencing retryable failures, then this change would start raising fatal errors for them. Perhaps the safest option (only raise errors when we know what to recommend) should be preferred.

PR #1805, opened 2022-05-31
Fix App page views

benjaminpkane:
v0.16.0 and v0.16.1 configure GA but do not send the `pageview` event.

Will publish a patch today or tomorrow.


brimoor (approved):
LGTM

Issue #1807, opened 2022-05-31
[BUG] Problems with patches views and session.view property


brimoor:
### System Information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS 11.6.5
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: 0.16.1
-   **Python version**: 3.9

### Bugs

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")
patches = dataset.to_patches("ground_truth")

session = fo.launch_app(dataset)

session.view = patches  # fails with error below
```

```
Error: invalid path uniqueness
at http://localhost:5151/assets/index.4033b8d5.js:191:11861
at a (http://localhost:5151/assets/vendor.f3a8bf7d.js:51:287465)
at w (http://localhost:5151/assets/vendor.f3a8bf7d.js:51:276411)
at http://localhost:5151/assets/vendor.f3a8bf7d.js:51:274973
```

```py
# Recover from failed state
session.dataset = None
session.refresh()

session = fo.launch_app(patches)  # this actually works

session.view = dataset.limit(10)  # fails with error below
```

```
Error: invalid path sample_id
at http://localhost:5151/assets/index.4033b8d5.js:191:11861
at a (http://localhost:5151/assets/vendor.f3a8bf7d.js:51:287465)
at w (http://localhost:5151/assets/vendor.f3a8bf7d.js:51:276411)
at http://localhost:5151/assets/vendor.f3a8bf7d.js:51:274973
```

brimoor:
Resolved by https://github.com/voxel51/fiftyone/pull/1808.

PR #1808, opened 2022-05-31
[WIP] Release v0.16.2

benjaminpkane:
### Includes

* #1786 
* #1801 
* #1805

### DIFF
--- a/app/packages/app/src/Root/Root.tsx
+++ b/app/packages/app/src/Root/Root.tsx
@@ -64,7 +64,7 @@ const getUseSearch = (prepared: PreloadedQuery<RootQuery>) => {
     const { data, refetch } = usePaginationFragment(
       graphql`
         fragment RootDatasets_query on Query
-          @refetchable(queryName: "DatasetsPaginationQuery") {
+        @refetchable(queryName: "DatasetsPaginationQuery") {
           datasets(search: $search, first: $count, after: $cursor)
             @connection(key: "DatasetsList_query_datasets") {
             total
@@ -109,7 +109,6 @@ const DatasetLink: React.FC<{ value: string; className: string }> = ({
 };
 
 export const useGA = (prepared: PreloadedQuery<RootQuery>) => {
-  const [gaInitialized, setGAInitialized] = useState(false);
   const query = usePreloadedQuery<RootQuery>(rootQuery, prepared);
 
   const info = useFragment(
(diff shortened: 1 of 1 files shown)


brimoor:
As of https://github.com/voxel51/fiftyone/pull/1808/commits/a624f434889abb737662e455fe5ea63ab4319dc9, when I run this code:

```py
import fiftyone as fo
import fiftyone.zoo as foz

clips = dataset.to_clips("frames.detections")

session.view = clips
```

and then refresh the App by pressing ctrl + R in my browser, I get this error:

```
Error: invalid path support
at http://localhost:5151/assets/index.60546a6e.js:191:11861
at a (http://localhost:5151/assets/vendor.fa360434.js:51:287465)
at w (http://localhost:5151/assets/vendor.fa360434.js:51:276411)
at B (http://localhost:5151/assets/vendor.fa360434.js:51:277622)
at http://localhost:5151/assets/vendor.fa360434.js:51:278916
at http://localhost:5151/assets/vendor.fa360434.js:51:278888
at Object.F [as get] (http://localhost:5151/assets/vendor.fa360434.js:51:278910)
at getNodeLoadable (http://localhost:5151/assets/vendor.fa360434.js:51:235406)
at http://localhost:5151/assets/vendor.fa360434.js:51:277128
at u2.getLeafNode (http://localhost:5151/assets/vendor.fa360434.js:51:268153)
```

Also, if I then try to escape this error by pressing `x` in the error form, the error still persists. It also persists if I first do this:

```py
session.view = None
```

and then press the `x`. However, I can now escape the error via ctrl + R in the browser.

I'm stating this second part because it would be ideal if pressing `x` would more reliably bring the App back to a functional state. One off-the-cuff idea there is:

- If the error occurred while loading a view, reload with the view cleared (`session.view = None`)
- If the error occurred while loading a dataset, reload with the dataset cleared (`session.dataset = None`)
- If any other error occurred, just reload the current route

benjaminpkane:
Comments addressed.

Resetting from the error page with the `X` is equivalent to `session.dataset = None` now. Adding in more scenarios would likely get complicated.

brimoor:
Just verified that #1807 and #1809 are resolved 👍

brimoor:
> Resetting from the error page with the `X` is equivalent to `session.dataset = None` now. Adding in more scenarios would likely get complicated.

Hmm not sure about always resetting to `session.dataset = None`... feels like this might be more annoying than helpful because it'll take clicks to get back to where you where. Likely you did a sequence of things, all of which were valid except the very last thing. I guess in that sense what would be cool is to keep a small history of pages and just pop the bad one from the stack...

But anyway, don't want to overdesign right now. The whole point of having the `x` is to escape to a valid state, so, I guess if it must always do one thing, then `session.dataset = None` is the only thing that's guaranteed to work.

Issue #1760, opened 2022-05-20
[FR] Add support for toggling between multiple media fields in the App


brimoor:
In many use cases, it makes sense for a `Sample` to have multiple versions of its media. For example:
- Providing different resolution versions of a video
- Storing anonymized versions of media (eg with faces/license plates blurred)
- Storing a low resolution version of a video for use in the grid vs in the expanded modal ("manual thumbnailing")

Note that we're not talking about multiple views of the scene (eg left vs right camera) here, we're talking about different versions of the exact same scene. Any `Detection` labels, for example, would be renderable on any version of the media.

This could be achieved by adding a `MediaField` subtype of `StringField` that allows the user to declare that the field contains a media path that should be included in a dropdown menu in the App that controls which media field is being used to source media files.

The required `filepath` field would have this type, but so could new fields:

```py
import fiftyone as fo

dataset = fo.Dataset()
dataset.add_sample_field("anonymized_image_path", fo.MediaField)

sample = fo.Sample(
    filepath="/path/to/image.jpg",
    anonymized_image_path="/path/to/anonymized-image.jpg",
)
dataset.add_sample(sample)

print(dataset)
```

```
Name:        2022.05.20.18.08.13
Media type:  image
Num samples: 1
Persistent:  False
Tags:        []
Sample fields:
    id:                    fiftyone.core.fields.ObjectIdField
    filepath:              fiftyone.core.fields.MediaField
    tags:                  fiftyone.core.fields.ListField(fiftyone.core.fields.StringField)
    metadata:              fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.metadata.Metadata)
    anonymized_image_path: fiftyone.core.fields.MediaField
```

Additional elements
- Add support for programmatically changing media fields via the [App config](https://voxel51.com/docs/fiftyone/user_guide/config.html#configuring-the-app)?
- Add support for configuring which media field is used by default in both the grid and (separately?) the expanded modal? For example, use `filepath` for the expanded modal but `thumbnail_path` in the grid. Such configuration should be stored on the `DatasetDocument`

Should the dropdown go under settings? Or somewhere else?

<img width="1148" alt="Screen Shot 2022-05-20 at 5 59 59 PM" src="https://user-images.githubusercontent.com/25985824/169618205-01b47c61-3013-4b5b-8c74-46489947ef56.png">

ritch:
@brimoor this has been in for a while, closing. LMK if there is something worth keeping this open for.

PR #1737, opened 2022-05-10
Adding saved views

brimoor:
This PR adds support for saving views on datasets under a `name` that can then be loaded at a later time.

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone import ViewField as F

dataset = foz.load_zoo_dataset("quickstart")
dataset.persistent = True

results = dataset.evaluate_detections(
    "predictions",
    gt_field="ground_truth",
    eval_key="eval",
)

view = (
    dataset
    .select_fields("ground_truth")
    .filter_labels("ground_truth", F("label") == "cat")
)

dataset.save_view("test", view)
```

Later, in another session:

```py
import fiftyone as fo

dataset = fo.load_dataset("quickstart")

results = dataset.load_evaluation_results("eval")
view = dataset.load_view("test")
```

PR #1811, opened 2022-06-01
[WIP] Media fields

ritch:
wip

https://github.com/voxel51/fiftyone/issues/1760


ritch:
> @ritch FYI #1737 may be a helpful reference for some other housekeeping items that come with features like this. Including:
> 
> * DB up/down migrations (for any changes to data model)
> * Docs updates (for any noteworthy additions to API or App)
> * Unit tests (okay, they're really integration tests)

I've looked through the PR and I see what you are referring to. Lets discuss how you tested the migration code. Otherwise seems straight forward. Docs might need a couple reviews to get right (not sure where all its worth mentioning the field type and filepath type change).

ritch:
Going with a different approach in: https://github.com/voxel51/fiftyone/pull/1841

Closing this, but keeping the branch to pull some code from.

PR #1812, opened 2022-06-01
OpenLABEL importer updates

ehofesmann:
This PR revamps the OpenLABEL importers to be much more flexible with regards to loading attributes and adds the ability to load keypoint pose data and correspond them to a given skeleton.

Specifically, it adds the `skeleton` and `skeleton_key` arguments to the OpenLABEL importers that let you provide a `KeypointSkeleton` used to load keypoints of the skeleton in the correct order and a `skeleton_key` that provides the key into the OpenLABEL annotation dictionary to the label of each point that matches the labels in the skeleton.

```python
    data_path=self.images_dir,
    labels_path=labels_path,
    dataset_type=fo.types.OpenLABELImageDataset,
    label_types="keypoints",
    skeleton_key=skeleton_key,
    skeleton=skeleton,
)
```

A new test was added for these keypoint poses that can be run to test this PR.


brimoor (approved):
LGTM

PR #1830, opened 2022-06-07
[WIP] Dynamic schemas, type annotations, and JSON validation

benjaminpkane:
### Changes

* Removes mongoengine
* Adds support for full dynamic schemas
* Adds support for modeling data with type annotations
* Replaces `fiftyone.core.odm` with `fiftyone.core.data`
* FASTER

A full description will be added soon


benjaminpkane:
I will share quickstart numbers when I can confirm it is loading correctly. Time permitting

PR #1838, opened 2022-06-07
Fix bug when using zoo models with newest version of Torchvision

### LINKED ISSUE #1789: [BUG] loading segmentation torch models errror, has no attribute 'load_state_dict_from_url'
### System information

-   **OS Platform and Distribution**: `Ubuntu 20.04`
-   **FiftyOne installed from (pip or source)**: `pip`, from within `conda` environment
-   **FiftyOne version (run `fiftyone --version`)**: `FiftyOne v0.16.0, Voxel51, Inc.`
-   **Python version**: `Python 3.9.7`
-   **Torch version**: `1.11.0`
-   **Torchvision version**: `0.12.0`

### Code to reproduce

```
import fiftyone.zoo as foz
```

### Describe the problem
The model fails to load (was taken from the [semantic segmentation evaluation doc](https://voxel51.com/docs/fiftyone/user_guide/evaluation.html#semantic-segmentations). Ultimately the error is  `AttributeError: module 'torchvision.models.segmentation.deeplabv3' has no attribute 'load_state_dict_from_url'`. This applies to all the PyTorch segmentation models listed [here](https://voxel51.com/docs/fiftyone/user_guide/model_zoo/models.html).

### Other info / logs
```
Input In [2], in <cell line: 3>()
      1 import fiftyone.zoo as foz
----> 3 model = foz.load_zoo_model("deeplabv3-resnet101-coco-torch")

File ~/miniconda3/lib/python3.9/site-packages/fiftyone/zoo/models/__init__.py:207, in load_zoo_model(name, download_if_necessary, install_requirements, error_level, **kwargs)
    204 config_dict = deepcopy(model.default_deployment_config_dict)
    205 model_path = model.get_path_in_dir(models_dir)
--> 207 return fom.load_model(config_dict, model_path=model_path, **kwargs)

File ~/miniconda3/lib/python3.9/site-packages/fiftyone/core/models.py:1505, in load_model(model_config_dict, model_path, **kwargs)
   1499         raise ValueError(
   1500             "Model config must implement the %s interface"
   1501             % etal.HasPublishedModel
   1502         )
   1504 # Build model
-> 1505 return config.build()

File ~/miniconda3/lib/python3.9/site-packages/eta/core/learning.py:296, in ModelConfig.build(self)
    289 def build(self):
    290     """Factory method that builds the Model instance from the config
    291     specified by this class.
    292 
    293     Returns:
    294         a Model instance
    295     """
--> 296     return self._model_cls(self.config)

File ~/miniconda3/lib/python3.9/site-packages/fiftyone/utils/torch.py:230, in TorchImageModel.__init__(self, config)
    226 self._device = torch.device("cuda:0" if self._using_gpu else "cpu")
    227 self._using_half_precision = (
    228     self.config.use_half_precision is True
    229 ) and self._using_gpu
--> 230 self._model = self._load_model(config)
    231 self._no_grad = None
    232 self._benchmark_orig = None

File ~/miniconda3/lib/python3.9/site-packages/fiftyone/utils/torch.py:471, in TorchImageModel._load_model(self, config)
    468 def _load_model(self, config):
    469     self._download_model(config)
--> 471     model = self._load_network(config)
    473     model = model.to(self._device)
    474     if self._using_half_precision:

File ~/miniconda3/lib/python3.9/site-packages/fiftyone/zoo/models/torch.py:97, in TorchvisionImageModel._load_network(self, config)
     94 kwargs = config.entrypoint_args or {}
     95 model_dir = fo.config.model_zoo_dir
---> 97 monkey_patcher = _make_load_state_dict_from_url_monkey_patcher(
     98     entrypoint, model_dir
     99 )
    100 with monkey_patcher:
    101     # Builds net and loads state dict from `model_dir`
    102     model = entrypoint(**kwargs)

File ~/miniconda3/lib/python3.9/site-packages/fiftyone/zoo/models/torch.py:114, in _make_load_state_dict_from_url_monkey_patcher(entrypoint, model_dir)
    108 """Monkey patches all instances of ``load_state_dict_from_url()`` that are
    109 reachable from the given ``entrypoint`` function in the
    110 :mod:`torchvision:torchvision.models` namespace so that models will be
    111 loaded from ``model_dir`` and not from the Torch Hub cache directory.
    112 """
    113 entrypoint_module = inspect.getmodule(entrypoint)
--> 114 load_state_dict_from_url = entrypoint_module.load_state_dict_from_url
    116 def custom_load_state_dict_from_url(url, **kwargs):
    117     return load_state_dict_from_url(url, model_dir=model_dir, **kwargs)

AttributeError: module 'torchvision.models.segmentation.deeplabv3' has no attribute 'load_state_dict_from_url'
```

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community.

ehofesmann:
Resolves #1789

We monkey patch the `load_state_dict_from_url` function used by torch models in the FiftyOne Model Zoo so that we can choose to load them from the model zoo directory on disk. As of torchvision v0.12.0, the structure of some models has been changed so the existing monkey patch no longer worked. This PR updates the location of the `load_state_dict_from_url` in the torchvision module and changes the namespace where it is applied.

To test, you can run this code from the evaluate segmentations example:
```python
import fiftyone.zoo as foz

# Load a few samples from COCO-2017
    "quickstart",
    max_samples=3,
).clone()

# Add DeepLabv3-ResNet50 predictions to dataset
```

### DIFF
--- a/fiftyone/zoo/models/manifest-torch.json
+++ b/fiftyone/zoo/models/manifest-torch.json
@@ -61,7 +61,7 @@
             "default_deployment_config_dict": {
                 "type": "fiftyone.zoo.models.torch.TorchvisionImageModel",
                 "config": {
-                    "entrypoint_fcn": "torchvision.models.segmentation.segmentation.deeplabv3_resnet101",
+                    "entrypoint_fcn": "torchvision.models.segmentation.deeplabv3_resnet101",
                     "entrypoint_args": { "pretrained": true },
                     "output_processor_cls": "fiftyone.utils.torch.SemanticSegmenterOutputProcessor",
                     "mask_targets_path": "{{eta-resources}}/voc-labels.txt",
@@ -98,7 +98,7 @@
             "default_deployment_config_dict": {
                 "type": "fiftyone.zoo.models.torch.TorchvisionImageModel",
                 "config": {
-                    "entrypoint_fcn": "torchvision.models.segmentation.segmentation.deeplabv3_resnet50",
+                    "entrypoint_fcn": "torchvision.models.segmentation.deeplabv3_resnet50",
                     "entrypoint_args": { "pretrained": true },
                     "output_processor_cls": "fiftyone.utils.torch.SemanticSegmenterOutputProcessor",
                     "mask_targets_path": "{{eta-resources}}/voc-labels.txt",
@@ -350,7 +350,7 @@
             "default_deployment_config_dict": {
                 "type": "fiftyone.zoo.models.torch.TorchvisionImageModel",
                 "config": {
-                    "entrypoint_fcn": "torchvision.models.segmentation.segmentation.fcn_resnet101",
+                    "entrypoint_fcn": "torchvision.models.segmentation.fcn_resnet101",
                     "entrypoint_args": { "pretrained": true },
                     "output_processor_cls": "fiftyone.utils.torch.SemanticSegmenterOutputProcessor",
                     "mask_targets_path": "{{eta-resources}}/voc-labels.txt",
@@ -387,7 +387,7 @@
             "default_deployment_config_dict": {
                 "type": "fiftyone.zoo.models.torch.TorchvisionImageModel",
                 "config": {
-                    "entrypoint_fcn": "torchvision.models.segmentation.segmentation.fcn_resnet50",
+                    "entrypoint_fcn": "torchvision.models.segmentation.fcn_resnet50",
                     "entrypoint_args": { "pretrained": true },
                     "output_processor_cls": "fiftyone.utils.torch.SemanticSegmenterOutputProcessor",
                     "mask_targets_path": "{{eta-resources}}/voc-labels.txt",
--- a/fiftyone/zoo/models/torch.py
+++ b/fiftyone/zoo/models/torch.py
@@ -6,6 +6,7 @@ FiftyOne Zoo models provided by :mod:`torchvision:torchvision.models`.
 |
 """
 import inspect
+from packaging import version
 
 import eta.core.utils as etau
 
@@ -111,6 +112,8 @@ def _make_load_state_dict_from_url_monkey_patcher(entrypoint, model_dir):
     loaded from ``model_dir`` and not from the Torch Hub cache directory.
     """
     entrypoint_module = inspect.getmodule(entrypoint)
+    if version.parse(torchvision.__version__) >= version.parse("0.12.0"):
+        entrypoint_module = torchvision._internally_replaced_utils
     load_state_dict_from_url = entrypoint_module.load_state_dict_from_url
 
     def custom_load_state_dict_from_url(url, **kwargs):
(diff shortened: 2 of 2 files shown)


fiftyone/zoo/models/manifest-torch.json
```diff
@@ -61,7 +61,7 @@
             "default_deployment_config_dict": {
                 "type": "fiftyone.zoo.models.torch.TorchvisionImageModel",
                 "config": {
-                    "entrypoint_fcn": "torchvision.models.segmentation.segmentation.deeplabv3_resnet101",
+                    "entrypoint_fcn": "torchvision.models.segmentation.deeplabv3_resnet101",
```
brimoor, line 64: If the old path no longer works, then looks like you'll have to add a `torchvision>=` requirement to these, right?

fiftyone/zoo/models/torch.py
```diff
@@ -120,5 +123,5 @@ def custom_load_state_dict_from_url(url, **kwargs):
         entrypoint_module,
         custom_load_state_dict_from_url,
         fcn_name=load_state_dict_from_url.__name__,
-        namespace=torchvision.models,
+        namespace=torchvision,
```
brimoor, line 126: Does this change depend on the specific torchvision version? IE will this code still work on `torchvision < 0.12.0`?

fiftyone/zoo/models/manifest-torch.json
brimoor: If we don't have to force the very latest `torchvision` version, then that'd be nice. But, if there's no easy way to support older and newer versions, then adding a `torchvision>=XXX` is fwm

brimoor (changes requested):

fiftyone/zoo/models/manifest-torch.json
ehofesmann: `torchvision.models.segmentation` imports everything from inside of `segmentation`, so the additional `.segmentation` was unnecessary and this still worked when I tested on `torchvision < 0.12.0`

fiftyone/zoo/models/torch.py
ehofesmann: I'm gonna be honest, I'm not 100% sure how changing the namespace affects this function, I assumed that it just makes it so that all `load_state_dict_from_url` will be replaced with the custom function in the entirety of `torchvision` rather than just in `torchvision.models`. But I did test this with `torchvision < 0.12.0` and it worked.

brimoor: Was this change necessary to get this PR to work?

brimoor (approved):
LGTM

fiftyone/zoo/models/torch.py
brimoor: I guess it probably is required or else you wouldn't have done it 😅

ehofesmann: Ah, you're right. This wasn't necessary. I was under the assumption that since the `load_state_dict_from_url` function definition was moved outside of `torchvision.models` that the namespace would also have to change, but it seems that the namespace only defines where to look for uses of that function and replace those. I'll revert it.

PR #1839, opened 2022-06-07
Update how CVAT task and project pages are requested

ehofesmann:
This PR changes the functions that iterate through CVAT tasks and projects by directly building the request url for a specific page rather than using the url provided in the `next` response from the previous page. There can be issues where this url is malformed and does not contain `https` or a port.

This approach is what is used by the [CVAT CLI.](https://github.com/openvinotoolkit/cvat/blob/d7560bbd39fec68f944515c2591dda74f3764b90/utils/cli/core/core.py#L50-L73) 

To test:

```python
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=10).clone()
results = dataset.annotate(
    "anno_key",
    label_field="ground_truth",
    task_size=2,
    projects="test_project",
    url="https://cvat.org",
    username=CVAT_USERNAME,
    password=CVAT_PASSWORD,
)

print(api.list_tasks())
print(api.list_projects())
```

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -3491,12 +3491,18 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
     def projects_url(self):
         return "%s/projects" % self.base_api_url
 
+    def projects_page_url(self, page_number):
+        return "%s/projects?page=%d" % (self.base_api_url, page_number)
+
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM. Nice detective work here 💪 📈

PR #1841, opened 2022-06-08
[HOLD] Dataset Media Fields

### LINKED ISSUE #1760: [FR] Add support for toggling between multiple media fields in the App

ritch:
Screen Shot 2022-06-14 at 6 19 28 PM

```py
import fiftyone as fo
import fiftyone.zoo as foz


for i, sample in enumerate(dataset): 
    sample["thumbnail"] = f"/Users/ben/data/thumbnails/quickstart-{i}.png" 
    foui.transform_image(sample.filepath, sample.thumbnail, size=(64, -1)) 
    sample.save()

dataset.app_config.media_fields = ["filepath", "thumbnail"]
dataset.app_config.grid_media_field = "thumbnail"

session = fo.Session(dataset)
```

### Issues

* Resolves #1760 

### Questions?
* How much validation should be done? E.g. when a media field is deleted. Should we validate `grid_media_field` is a `media_field`?
* Are we extending these to the global app config as well? What if a dataset does not have the right shape?

### Todo

* Documentation


app/packages/app/src/components/Actions/Options.tsx
```diff
@@ -229,6 +229,8 @@ type OptionsProps = {
 };
 
 const Options = ({ modal, bounds }: OptionsProps) => {
+  const dataset = useRecoilValue(atoms.dataset);
+  console.log(dataset);
```
ritch, line 233: This logs:

```
{id: '62a0acd249dac2bfcfa32ccc', name: '2022.06.08.07.06.10', mediaType: 'image', sampleFields: Array(8), frameFields: Array(0), …}
appSidebarGroups: null
brainMethods: []
createdAt: "2022-06-08T14:06:10.175000"
defaultMaskTargets: {}
defaultSkeleton: null
evaluations: []
frameFields: []
id: "62a0acd249dac2bfcfa32ccc"
lastLoadedAt: "2022-06-08T14:06:12.639000"
maskTargets: {}
mediaType: "image"
name: "2022.06.08.07.06.10"
sampleFields: (8) [{…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}]
skeletons: []
version: "0.16.1"
viewCls: null
```

fiftyone/core/odm/dataset.py
```diff
@@ -391,3 +398,4 @@ class DatasetDocument(Document):
     app_sidebar_groups = ListField(
         EmbeddedDocumentField(document_type=SidebarGroupDocument), default=None
     )
+    app_config = EmbeddedDocumentField(document_type=DatasetAppConfigDocument)
```
benjaminpkane, line 401:
```suggestion
    app_config = EmbeddedDocumentField(
        document_type=DatasetAppConfigDocument,
        default=lambda: DatasetAppConfigDocument(
            grid_media_field="filepath", media_fields=["filepath"]
        ),
    )
```

fiftyone/core/dataset.py
```diff
@@ -245,6 +246,11 @@ def __init__(
             )
 
         self._doc = doc
+        if self._doc.app_config == None:
```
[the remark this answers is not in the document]
benjaminpkane: I think the `default` kwarg can be used when defining the field. Providing a lambda. See suggestion

fiftyone/core/dataset.py
```diff
@@ -360,6 +362,14 @@ def media_type(self, media_type):
         self._doc.media_type = media_type
         self._set_metadata(media_type)
 
+    @property
+    def app_config(self):
+        return self._doc.app_config
+
+    @app_config.setter
+    def app_config(self, config):
+        self._doc.app_config = config
```
brimoor, line 371: `self._doc.save()` is needed here

brimoor (changes requested):
I propose that we merge this into a `feature/dataset-app-config` feature branch and hold off on releasing until we also expose/migrate all of `fityone.core.config.AppConfig` into `fiftyone.core.odm.DatasetAppConfig` (or decide we never want to do that). My goal is to more fully round out the new `dataset.app_config` concept before releasing.

Other review comments
- Need to add logic in `delete_sample_field()` and `rename_sample_field()` that will appropriately update `DatasetAppConfig.media_fields` as necessary
- Prefer to migrate `app_sidebar_groups` into `app_config.sidebar_groups`, for cleanliness
- These features will need documentation before releasing

brimoor:
Following up on some questions from the PR description:

Q: Are we extending these to the global app config as well? What if a dataset does not have the right shape?
A: I don't think we should bring the new `Dataset.app_config` into the `fiftyone.AppConfig` (as pointed out, these new settings depend on the shape of the dataset and thus can't be global). Instead I think we should bring `fiftyone.AppConfig` into `Dataset.app_config` 🤗  There are some order of precedence details to get right

Q: How much validation should be done? E.g. when a media field is deleted. Should we validate grid_media_field is a media_field?
A: I think we should automatically handle deletion/renaming of a field that's included in `media_fields`, but, since we don't have a `MediaField` concept *only a `StringField` that may contain filepaths, but may also just contain "foo" and "bar"), we shouldn't attempt to validate that a `media_fields` choice is valid. Just wait for "invalid image" to appear in the App grid

benjaminpkane:
Stale. Merged in to #1954

PR #1848, opened 2022-06-09
Label studio integration

Rusteam:

A new annotation backend with [Label Studio](labelstud.io).
At the moment, images only.

Integration testing is done with connecting to a local instance of Label Studio.

1. Get Label Studio up-and-running with `pip` or `docker`
```
pip install label-studio
label-studio start
# or docker
 docker run -d -p 8080:8080 \
            --name labelstudio \
            -v /tmp/labelstudio:/data \
            labelstudio/labelstudio
```

3. Get access token from account settings, set url and token in the environment:
```
export FIFTYONE_LABELSTUDIO_URL=<url>
export FIFTYONE_LABELSTUDIO_TOKEN=<token>
```
4. 

        notes for FiftyOne users. 
-  [x] Integration page write-up.

Annotate image classes, detections, polygons, etc. with Label Studio.

-   [ ] Other


fiftyone/utils/labelstudio.py
```diff
@@ -0,0 +1,835 @@
...
+        project_name: the name of the project to use on the Label Studio server
+
+    """
+
+    def __init__(
+        self,
+        name: str,
+        label_schema: str,
+        media_field="filepath",
+        url="http://localhost:8080",
```
ehofesmann, line 80: The default `url` can be set to `None` since it will already have a default value provided through `config.py` and this default value will not be used.

tests/intensive/labelstudio_tests.py
```diff
@@ -0,0 +1,563 @@
+"""Test Label Studio integration.
```
ehofesmann, line 1: Amazing! Thanks for adding these tests!

fiftyone/utils/labelstudio.py
```diff
@@ -0,0 +1,835 @@
...
+        child_tag="Label",
+        label=fol.Segmentation,
+    ),
+    "scalar": dict(
+        parent_tag="Number",
+        child_tag=None,
+        label=fol.Regression,
+    ),
+}
+_LABEL2TYPE = {v["label"].__name__: k for k, v in _LABEL_TYPES.items()}
```
ehofesmann, line 835: Nit: Generally we prefer to explicitly type out variable names.
```suggestion
_LABEL_TO_TYPE = {v["label"].__name__: k for k, v in _LABEL_TYPES.items()}
```

fiftyone/utils/labelstudio.py
```diff
@@ -0,0 +1,835 @@
...
+        Args:
+            project_id: project id
+        """
+        self.make_request(
+            "DELETE",
+            f"/api/projects/{project_id}",
+        )
+
+
+class LabelStudioAnnotationResults(foua.AnnotationResults):
```
ehofesmann, line 437: Note, that this class requires the [`_from_dict()`](https://github.com/voxel51/fiftyone/blob/8cab046e7fe98ad04e37bab70183e03e128904bf/fiftyone/utils/annotations.py#L2094) method to allow the run results to be loaded properly from the database. You can see an example of it [here](https://github.com/voxel51/fiftyone/blob/8cab046e7fe98ad04e37bab70183e03e128904bf/fiftyone/utils/labelbox.py#L1378)

fiftyone/utils/labelstudio.py
```diff
@@ -0,0 +1,835 @@
...
+    @property
+    def api_key(self):
+        return self._api_key
+
+    @api_key.setter
+    def api_key(self, value):
+        self._api_key = value
+
+
+class LabelStudioBackend(foua.AnnotationBackend):
```
ehofesmann, line 105: This class is required to implement the [`recommend_attr_tool()`](https://github.com/voxel51/fiftyone/blob/8cab046e7fe98ad04e37bab70183e03e128904bf/fiftyone/utils/annotations.py#L1834) method which basically defines the type of annotation widget that will be used for a given attribute, generally based on the type of attribute that is being parsed.

ehofesmann (changes requested):
Wow! Thanks for all the work you're putting into this, this is awesome!

It looks like you still have TODOs you are working through, but I went ahead and left some recommendations that you can take a look at. Let me know if you have any questions as you're working on this!

fiftyone/utils/labelstudio.py
Rusteam: I haven't added attribute handling yet. Is it strictly necessary at the start?

ehofesmann: No problem! It makes sense to get everything working first before tackling attributes.

fiftyone/utils/labelstudio.py
```diff
@@ -0,0 +1,835 @@
...
+        project_name=None,
+        **kwargs,
+    ):
+        super().__init__(name, label_schema, media_field=media_field, **kwargs)
+
+        self.url = url
+        self.project_name = project_name
+
+        self._api_key = (
+            api_key if api_key else os.getenv("FIFTYONE_LABELSTUDIO_TOKEN")
```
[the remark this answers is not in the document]
Rusteam: fixed now, make sure to set the api_key variable before running `pytest labelstudio_tests.py` as it will require the token for each test

fiftyone/utils/labelstudio.py
Rusteam: I left the home page for LS as the default address however it will not lead to anywhere. if running locally users have to set localhost explicitely

fiftyone/utils/labelstudio.py
Rusteam: are we closing this one for now?

ehofesmann: So at the moment, if attributes are provided, then `recommend_attr_tool()` will be called. This can happen by default if passing in an existing label field where attributes are parsed automatically. The way to avoid it, you can set `attributes=False`, like this:
```python
dataset.annotate(anno_key, label_field="existing_field", attributes=False, backend="labelstudio")
```

From what I can tell, there's two ways to go about this:
1) Raise a `NotImplementedError` in `LabelStudioBackend.recommend_attr_tool()` that instructs the user to set `attributes=False` until attributes are supported by this backend.
2) Add a flag that automatically ignores all attributes for a given backend [like this](https://github.com/voxel51/fiftyone/commit/fa55781dda7b25dff0432581f3740db43ce18fd1).

While it would be smoother to automatically ignore attributes, I'm a bit worried about that behavior being unexpected for users. What are your thoughts @Rusteam and @brimoor ?

Another alternative is to just add attribute support, but this issue can come up with future backends too so it would be useful to discuss.

brimoor: Raising a `NotImplementedError` is a reasonable thing to do if we plan to implement it soon, but I guess the crux of the decision is that it would be annoying that the simplest syntax:

```py
dataset.annotate(anno_key, label_field="existing_field", backend="labelstudio")
```

would fail since the public `annotate()` API has already decided that `attributes=True` is always the default. Seems like good motivation to get attributes support done too 🤗 

I would prefer to only introduce `supports_attributes` if we think there will be annotation backends that don't support label attributes at all, not just because we haven't implemented it yet.

If we do introduce `supports_attributes`, then we could update the default to `attributes=None` in the public `annotate(...)` API and then [document](https://voxel51.com/docs/fiftyone/user_guide/annotation.html#requesting-annotations) that the default value for the `attributes` parameter will be set to the value of `supports_attributes` property of the annotation backend being used. That would help address Eric's concern of certain backends silently ignoring attributes.

Rusteam: Seems like implementing attributes is the only viable solution :)

Rusteam: Do I understand correctly that attributes correspond to label tags? Is it usually applied to a label with multiple instances (detections, polygons, etc.) and can be something like `[occluded, partial, etc.]`?
If so, I don't think Label Studio supports per-label annotations. Their [control tags](https://labelstud.io/tags/relation.html) require a label to be attached to input data `toName`. I will double-check with their community

ehofesmann: Hmm, it seems you're right. From looking into Labelstudio more, it doesn't seem like they have a way to add attributes directly.

In FiftyOne, the idea of an attribute is any field added to a `Label` object other than the `Label.label` representing the class/category. The closest I found are the `Relations` that you linked or a [`TextArea`](https://labelstud.io/tags/textarea.html) (which is really meant for captioning and doesn't support numerical/boolean types that a lot of attributes use) neither of which are a clean way to support attributes. 

I'll dig a bit deeper, but it seems that Labelstudio may actually require `supports_attributes=False` until [this is resolved](https://github.com/heartexlabs/label-studio/issues/794).

Rusteam: iWould it beeter to not throw an error by default but rather show a warning that attributes are not supported?

Rusteam:
> Ah, yup, I needed to upgrade `label-studio`, I was on 1.4 previously. We'll need to make sure to set the requirements to `label-studio>=1.5.0`. Things are working great now!

That's on server side. We do not expect fiftyone to handle that, right? Maybe we can query the server version and give a warning if it's below 1.5.0? As for sdk version, can we specify a specific version with `fou.lazy_import()`?

Rusteam:
Hey there, any updates on this?

ehofesmann:
@Rusteam Sorry for the delay! Yeah, you can require a minimum version of the sdk through the lazy import. It looks like this:
```python
ls = fou.lazy_import(
    "label_studio_sdk", callback=lambda: fou.ensure_package("label_studio_sdk>=0.0.13")
)
```

But you're right, getting the version of the server would need to be done through a request and manual verification of the version. It seems that this would do it:
```python
from packaging import version
server_version = self._client.make_request("GET", "/api/version").json()["release"]
version.parse(server_version) >= version.parse("1.5.0")
```

ehofesmann:
After this and the `supports_attributes` updates, I want to play with it a bit more, but then it will likely be ready to merge! Thanks again for making this PR, this is really good work!

fiftyone/utils/labelstudio.py
ehofesmann: @Rusteam Thanks for all the updates you've been making to this! I've opened a PR (https://github.com/Rusteam/fiftyone/pull/1) on your fork into this branch adding the `supports_attributes` updates (as well as a few other things like Labelstudio docs). Take a look at that when you get a chance and let me know if you want to discuss anything there, if not, you can go ahead and merge it in.

Rusteam:
thanks @ehofesmann  for a clear guide on how to use this integration and other updates.

ehofesmann (approved):
I think we're ready to get this out in the next release. I've opened https://github.com/voxel51/fiftyone/pull/2006 for some tweaks after this is merged.

Thanks for all of your work @Rusteam! :heart:

AndreaPi:
Hi! I would be very interested in this integration. Is there an ETA for when it will be merged into FiftyOne? Thanks!

brimoor:
It's already merged and available as of `fiftyone==0.18`!

https://voxel51.com/docs/fiftyone/integrations/labelstudio.html

AndreaPi:
@brimoor ouch, you’re right, I didn’t notice the `merged` label! Thanks

Issue #1755, opened 2022-05-20
[FR] Upgrade quickstart dataset format


brimoor:
The `quickstart` zoo dataset uses the deprecated `attributes` dict to store object attributes:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")

print(dataset.first()["ground_truth"].detections[0])
```

```
<Detection: {
    'id': '5f452471ef00e6374aac53c8',
    'attributes': BaseDict({
        'area': <NumericAttribute: {'value': 73790.37944999996}>,
        'iscrowd': <NumericAttribute: {'value': 0.0}>,
    }),
    'tags': BaseList([]),
    'label': 'bird',
    'bounding_box': BaseList([
        0.21084375,
        0.0034375,
        0.46190625,
        0.9442083333333334,
    ]),
    'mask': None,
    'confidence': None,
    'index': None,
}>
```

[Per the docs](https://voxel51.com/docs/fiftyone/user_guide/using_datasets.html#label-attributes), the `attributes` field will be removed, and dynamic attributes should be used instead:

```
<Detection: {
    'id': '5f452471ef00e6374aac53c8',
    'tags': BaseList([]),
    'label': 'bird',
    'bounding_box': BaseList([
        0.21084375,
        0.0034375,
        0.46190625,
        0.9442083333333334,
    ]),
    'mask': None,
    'confidence': None,
    'index': None,
    'area': 73790.37944999996,
    'iscrowd': 0.0,
}>
```

We should upgrade the [quickstart dataset](https://github.com/voxel51/fiftyone/blob/22e34e91deb1d2e2fe6316ec81714e0c55015523/fiftyone/zoo/datasets/base.py#L2433) to use dynamic attributes, per our recommendation.

brimoor:
Closing this issue won't involve a PR. The goal is just to update the format of the [quickstart files](https://github.com/voxel51/fiftyone/blob/d81007d630106f1819d95943df937fc112da3b3a/fiftyone/zoo/datasets/base.py#L2433) so that, when it the dataset downloaded via `download_zoo_dataset()` in the future, the dataset will use the latter format for its `Detection` objects.

It should be noted that the `quickstart` dataset is used very frequency in example snippets, so changing its format would, in general, be something that should be done with care to avoid breaking example code. However, we've been planning to delete the `attributes` dict for awhile:
[...]

PR #1857, opened 2022-06-09
Update docs to match new quickstart attribute format

ehofesmann:
Updates tutorials/recipes that use the quickstart dataset to match the new format of attributes from #1755


brimoor (approved):
LGTM. Thanks! 💪

Issue #1722, opened 2022-05-03
[FR] Stop passing classes and default_classes to export and evaluation methods


brimoor:
**Background**

Datasets can optionally [store class lists](https://voxel51.com/docs/fiftyone/user_guide/using_datasets.html#storing-class-lists) in their `Dataset.classes` and `Dataset.default_classes` properties.

The purpose of these properties is to store a complete and ordered class list for individual label fields (`classes`) or an entire dataset (`default_classes`), respectively. This information can be necessary because it may not be inferable from the observed labels in a field; eg, although a model/label schema included 10 total classes, perhaps the dataset/view does not actually contain instances of all 10 classes.

**The problem**

Currently, if `default_classes`/`classes` is available, then methods like `export()` and `evaluate_XXX()` will silently use the classes list, which may cause unexpected behavior if these lists are out of sync with the contents of that field.

For example:

```py
import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone import ViewField as F

dataset = foz.load_zoo_dataset("coco-2017", split="validation")

# Zoo datasets usually have classes lists populated
print(dataset.default_classes)

# I created a view that only contains cats and dogs, but since the dataset has
# `default_classes` populated, the exported labels will contain all 80 classes
# in its `categories` field, which is not what I wanted
view = dataset.filter_labels("ground_truth", F("label").is_in(["cat", "dog"]))
view.export(
    labels_path="/tmp/coco.json",
    dataset_type=fo.types.COCODetectionDataset,
)
```

```
	"categories": [{
		"id": 0,
		"name": "0",
		"supercategory": null
	}, {
		"id": 1,
		"name": "person",
		"supercategory": "person"
	}, {
		"id": 2,
		"name": "bicycle",
		"supercategory": "vehicle"
	}, {
		"id": 3,
		"name": "car",
		"supercategory": "vehicle"
	}, {
```

Users have been tripped up by this:
- https://github.com/PyTorchLightning/lightning-flash/issues/763#issuecomment-924540154
- https://github.com/voxel51/fiftyone/issues/1715

**Proposed solution**

A better default behavior is to just export whatever is actually in the label field, implicitly computing the available classes list whenever necessary, and instead require the user to manually opt-in to providing their classes list via the optional `classes` kwarg if they want a specific classes list to be used (by convention, such a kwarg always exists whenever the relevant method can leverage a manually defined classes list).

Continuing from the example code above:

```py
[...]

PR #1858, opened 2022-06-09
Stop passing classes and default_classes to export and evaluation methods

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1722.

This is scratching a pretty deep itch, and it feels oh so good 🥇 

The fact that certain API methods would silently search for class lists on `Dataset.classes` and `Dataset.default_classes` often caused unexpected behavior for users that had first used something like `filter_labels()` to select a subset of classes to work with (see https://github.com/voxel51/fiftyone/issues/1722 for details).

**Lesson learned**: don't be fancy, let the user store class lists, but prefer relying on the user to explicitly request certain behavior:

```py
# bad: silently access dataset.default_classes to locate classes
dataset.some_method(...)

# good: let user explicitly tell us what to do
dataset.some_method(..., classes=dataset.default_classes)
```

~~**NOTE**: there's still one important place that *will* silently infer class lists: the annotation API. My reasoning there is, even if the user first did something like `filter_labels()` to extract a view of interest to annotate, they are very likely to want *all possible classes from their schema* available for use when annotating.~~

^removed default classes/mask targets from the annotation API too

### DIFF
--- a/docs/source/integrations/coco.rst
+++ b/docs/source/integrations/coco.rst
@@ -325,7 +325,8 @@ to add them to your dataset as follows:
     ]
 
     # Add COCO predictions to `predictions` field of dataset
-    fouc.add_coco_labels(coco_dataset, "predictions", predictions)
+    classes = coco_dataset.default_classes
+    fouc.add_coco_labels(coco_dataset, "predictions", predictions, classes)
 
     # Verify that predictions were added to two images
     print(coco_dataset.count("predictions"))  # 2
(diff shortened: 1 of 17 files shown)


ehofesmann (approved):
This LGTM, but I would argue that this could be extended even to the annotation API. If you filter out some classes from the view, it still uses the underlying dataset (and the view) to get the entire list of classes so you don't need to be concerned about missing labels: https://github.com/voxel51/fiftyone/blob/aa076d0208dccf95626f568790185f7893ecd852/fiftyone/utils/annotations.py#L780-L783

Also, I have serious doubts that many users actually use or even know about the existence of `dataset.classes` and `dataset.default_classes`, so in the case that they do want to specify classes that don't exist yet, they would probably be providing them manually anyway. Though for the same reason, there's also not much of a reason to change the way the annotation API works currently. It would only trip people up if they have default classes defined for one field and go to annotate another field with different classes, which I suppose could happen but I've never seen it yet.

PR #1863, opened 2022-06-09
Add retries for CVAT requests

ehofesmann:
When making calls to the CVAT REST API, some of the requests can raise errors due to server issues requiring methods like `dataset.annotate()` and `dataset.load_annotations()` to be run multiple times before succeeding. This PR introduces retries around any `requests` made by our the CVAT annotation integration.

A test was added to `fiftyone/tests/intensive/cvat.py` that is able to verify that this works correctly.

### DIFF
--- a/fiftyone/core/config.py
+++ b/fiftyone/core/config.py
@@ -654,17 +654,20 @@ def _get_installed_packages():
 
 
 class HTTPRetryConfig(object):
-    """Values used to configure the behavior of the retry logic
-    of HTTP calls made throughout the app
+    """Values used to configure the behavior of the retry logic of HTTP calls
+    made throughout the library.
+
+    NOTE: calls made directly through storage clients (GCS, S3) use their own
+    internal retry logic implementation and may not perfectly match this
+    configuration. This configuration is for direct HTTP requests.
     """
 
     # HTTP codes that should trigger a retry
     RETRY_CODES = {408, 429, 500, 502, 503, 504, 509}
 
-    # Exponential backoff factor.
+    # Exponential backoff factor
     # See https://github.com/litl/backoff/blob/master/backoff/_wait_gen.py#L17
     FACTOR = 0.1
 
-    # Maximum number of times to execute a retry
-    # before throwing an exception
+    # Maximum number of times to execute a retry before throwing an exception
     MAX_TRIES = 10
--- a/fiftyone/core/config.py
+++ b/fiftyone/core/config.py
@@ -651,3 +651,20 @@ def _get_installed_packages():
     except:
         logger.debug("Failed to get installed packages")
         return set()
+
+
+class HTTPRetryConfig(object):
+    """Values used to configure the behavior of the retry logic
+    of HTTP calls made throughout the app
+    """
+
+    # HTTP codes that should trigger a retry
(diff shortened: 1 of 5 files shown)


fiftyone/core/config.py
```diff
@@ -651,3 +651,20 @@ def _get_installed_packages():
...
         return set()
+
+
+class HTTPRetryConfig(object):
+    """Values used to configure the behavior of the retry logic
+    of HTTP calls made throughout the app
+    """
+
+    # HTTP codes that should trigger a retry
+    RETRY_CODES = {408, 429, 500, 502, 503, 504, 509}
```
findtopher, line 662: wouldn't the 500 codes be errors that won't fix themselves with retries?

findtopher (commented):
I like your test, I just worry about retries on `5/d/d` error codes.  I'm happy to have someone else override my opinion, but I'm not comfortable approving this without understanding the plan.

fiftyone/core/config.py
Anddraca: Yeah, this is confusing -- the logic in the decorator/library that uses these codes implements a negation:
```
giveup=lambda e: hasattr(e, "error_code")
        and e.error_code not in HTTPRetryConfig.RETRY_CODES,```

Comment (findtopher) [reply fiftyone/core/config.py:662]: Then could we change the comment to say `HTTP codes that will not trigger a retry`?  That would help clarify things the next time I say dumb stuff 😝 

This makes sense to me.

**Review by findtopher — APPROVED** [review]
Thanks for the clarification - this makes sense in that context.

File: fiftyone/core/config.py (hunk 1 above)
Comment (Anddraca) [reply fiftyone/core/config.py:662]: That is absolutely true. I'm the one that wrote that original comment and I had my understanding backwards when I wrote it :D

Comment (Anddraca) [reply fiftyone/core/config.py:662]: And I copied them from what Brian had originally made :grin:  -- There was no small amount of discussion around these codes; they were mostly lifted from the various cloud service documentations

File: fiftyone/utils/cvat.py [hunk 2]
```diff
@@ -3655,6 +3657,15 @@ def _login(self, username, password):
                 "csrftoken"
             ]
 
+    @backoff.on_exception(
+        backoff.expo,
+        requests.exceptions.RequestException,
+        factor=HTTPRetryConfig.FACTOR,
+        max_tries=HTTPRetryConfig.MAX_TRIES,
+        giveup=lambda e: hasattr(e, "error_code")
```
brimoor, line 3665: This logic says "never giveup if we get an exception that doesn't have an error code".

First question: why is `hasattr(e, "error_code")` needed?

Comment: the goal here is to only retry in specific cases where the server gave us a code that we know is retryable, so if we get something unexpected, that would be a giveup condition, not a retry condition

brimoor: Probably should specifically test for that exception then rather than blindly assuming all errors without codes are retryable

brimoor:
@ehofesmann can we get this merged this week?

ehofesmann:
The retries in this PR wrap request calls that are used in a lot of places and can raise a variety of different errors that may need to be handled differently. I plan to discuss this more with @Anddraca offline and will circle back around to this. Closing this PR until then and opening an issue instead.

Issue #1522, opened 2022-01-10
[BUG] Inconsistent naming in loading COCO Dataset with include_id=True


devrimcavusoglu:
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 20.04
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: FiftyOne v0.14.2, Voxel51, Inc.
-   **Python version**: Python 3.9.7

### Describe the problem

There's a inconsistency for loading COCO dataset when using `include_id=True`. By default `include_id` is `False`, and loading a COCO dataset inherently loads the dataset by creating `"ground_truth"` field for samples and `"detections"` as a sub-field of `"ground_truth"` which is perfectly fine. However, when we try to preserve COCO ids with `include_id=True`, the name for ground_truth detections are now created as `"detections"` field and `"detections"` as a subfield, and it saves the COCO ids in "coco_id" field. To better illustrate the nested form, it can be shown as follows:

```json
{
    ...,
    "detections": { "detections": [...]},
    "coco_id": 174,
}
``` 

The first problem is the naming inconsistency between `"ground_truth"` and `"detections"` when you switch from `include_id=False` to `include_id=True` (subfields' name remain as `"detections"` for both). The expected behaviour is to get the same field name whether or not using `include_id` as `False/True`. Imho, "ground_truth" is better naming convention here.

The second problem is that when we try to fix this problem with `label_field` parameter, e.g `label_field="ground_truth"`, it makes things worse as the new field name is concatenated string `"<label_field> + detections"`, so if our `label_field` is `"ground_truth"` the detections field becomes `"ground_truth_detections"`. Moreover, it also does the same to the field `"coco_id"` as it becomes `"ground_truth_coco_id"`. 

### Code to reproduce issue

The behaviour described above can be observed with the following code.

```python
coco_dataset = fo.Dataset.from_dir(
    name="coco-dataset",
    dataset_type=fo.types.COCODetectionDataset,
    dataset_dir=IMAGES_DIR,
    labels_path="/path/to/coco.json",
    label_field="ground_truth",
    include_id=True
)

>>> print(sample)
<...
"ground_truth_detections": "detections": <[...]>,
"ground_truth_coco_id": 174
...>
```

### What areas of FiftyOne does this bug affect?

-   [ ] `App`: FiftyOne application issue
[...]

PR #1864, opened 2022-06-10
Adding support for label field dicts when importing labeled datasets

brimoor:
Adds support for passing `label_field` as a dict mapping label keys to field names when importing datasets using `Dataset.from_dir()` and all related factory methods.

As the example below illustrates, this provides more fine-grained support for directly specifying the field names to use when importing multitask dataset formats such as BDD:

```py
import random
import os

import fiftyone as fo
import fiftyone.zoo as foz

#
# Export a dataset with detections and classifications
#

data_path = os.path.dirname(dataset.first().filepath)

for sample in dataset.select_fields().iter_samples(progress=True):
    sample["weather"] = fo.Classification(label=random.choice(["sunny", "cloudy"]))
    sample.save()

dataset.export(
    labels_path="/tmp/bdd.json",
    dataset_type=fo.types.BDDDataset,
    label_field=["ground_truth", "weather"],
)
print(dataset)

# Now import with default label field names
    data_path=data_path,
    labels_path="/tmp/bdd.json",
    dataset_type=fo.types.BDDDataset,
)
print(dataset2)

# New syntax: import with one field name customized
dataset3 = fo.Dataset.from_dir(
    data_path=data_path,
    labels_path="/tmp/bdd.json",
    dataset_type=fo.types.BDDDataset,
    label_field={"detections": "ground_truth"},
)
print(dataset3)
```

### DIFF
--- a/docs/source/user_guide/dataset_creation/datasets.rst
+++ b/docs/source/user_guide/dataset_creation/datasets.rst
@@ -5164,14 +5164,20 @@ should implement is determined by the type of dataset that you are importing.
         importer = CustomLabeledImageDatasetImporter(...)
         label_field = ...
 
+        if isinstance(label_field, dict):
+            label_key = lambda k: label_field.get(k, k)
+        elif label_field is not None:
+            label_key = lambda k: label_field + "_" + k
+        else:
+            label_field = "ground_truth"
+            label_key = lambda k: k
+
         with importer:
             for image_path, image_metadata, label in importer:
                 sample = fo.Sample(filepath=image_path, metadata=image_metadata)
 
                 if isinstance(label, dict):
-                    sample.update_fields(
-                        {label_field + "_" + k: v for k, v in label.items()}
-                    )
+                    sample.update_fields({label_key(k): v for k, v in label.items()})
                 elif label is not None:
                     sample[label_field] = label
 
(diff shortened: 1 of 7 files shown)


ehofesmann (approved):
LGTM

Though this could get confusing due to the label fields of importers changing depending on what is being imported. Like [this issue](https://github.com/voxel51/fiftyone/issues/1522) where `ground_truth` turns into `detections` when `coco_id` is provided when importing a coco dataset (I see you are working on updating this for coco, we might as well update this for all formats). This might be a good time to consider moving away from defaulting to `"ground_truth"` and always requiring an importer to return a dict of labels.

This could let us get rid of `label_cls` when defining importers which is a bit confusing for new contributors. Without it, the importer could just return a dict of whatever labels it wants for each sample which is much easier to implement. Something like OpenLABEL where all fields are parsed directly from the annotations already has `label_cls` returning `None`. There are a couple of changes that would need to be made, primarily getting rid of [this optimization](https://github.com/voxel51/fiftyone/blob/d0d1784430810312ef6819c68b600797bebdce29/fiftyone/utils/data/parsers.py#L178-L183) which I don't have any idea of how much this helps.

brimoor:
Just documenting that @ehofesmann makes some good points above and were discussed IRL

PR #1877, opened 2022-06-13
Remove thumbnail frame classifications 

### LINKED ISSUE #1837: [BUG] Frame-level labels incorrectly appear as None in App grid view
On `fiftyone>=0.16`, `None` is incorrectly rendered for frame-level `Classification` fields in the App grid view.

I do not believe we are rendering frame-level fields in the grid view as the video plays, so the expected behavior here is that nothing would be rendered for the `frames.predictions` field in the grid, in both "poster" and "hover" mode.

```py
import fiftyone as fo

sample1 = fo.Sample(filepath="/path/to/video.mp4")

sample2 = fo.Sample(filepath="/path/to/video.mp4")
sample2.frames[1] = fo.Frame(predictions=fo.Classification(label="cat"))

dataset = fo.Dataset()

```

Screen Shot 2022-06-07 at 12 14 57 PM

benjaminpkane:
Resolves #1837

### DIFF
--- a/app/packages/looker/src/elements/common/tags.ts
+++ b/app/packages/looker/src/elements/common/tags.ts
@@ -263,6 +263,10 @@ export class TagsElement<State extends BaseState> extends BaseElement<State> {
           path
         );
 
+        if (field === null) {
+          continue;
+        }
+
         const pushList = (renderer, value) => {
           let count = 0;
           let rest = 0;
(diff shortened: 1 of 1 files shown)


ritch:
Fix verified:

```python
import fiftyone as fo
import os

vids = os.environ['TEST_VIDEOS']

sample1 = fo.Sample(filepath=f'{vids}/A.mp4')

sample2 = fo.Sample(filepath=f'{vids}/B.mp4')
sample2.frames[1] = fo.Frame(predictions=fo.Classification(label="cat"))

print(sample1.filepath)

dataset = fo.Dataset()
dataset.name = 'testing'

```
Screen Shot 2022-06-16 at 7 38 53 AM

ritch (approved):
LGTM

PR #1878, opened 2022-06-13
Maintain active dataset fields

### LINKED ISSUE #1852: [BUG] Changing session.view resets field visibility choices
On `fiftyone==0.16.2`, updating `session.view` resets any field visibility toggles I may have set (eg, unselected all label fields), and forces the defaults (all label fields visible). I don't think this used to be the case though?

This came up when I was trying to work with an interactive plot. I just wanted to see images with no labels, but every time I made a selection in the linked plot, the labels kept re-appearing, which was annoying. I don't recall facing this issue before.

Of course persisting sidebar settings is a bit tricky because views can change the label schema.

benjaminpkane:
Resolves #1852

### DIFF
--- a/app/packages/app/src/Root/Datasets/Dataset.tsx
+++ b/app/packages/app/src/Root/Datasets/Dataset.tsx
@@ -9,9 +9,7 @@ import { useStateUpdate } from "../../utils/hooks";
 import { DatasetQuery } from "./__generated__/DatasetQuery.graphql";
 import { datasetName } from "../../recoil/selectors";
 import transformDataset from "./transformDataset";
-import * as atoms from "../../recoil/atoms";
 import { filters } from "../../recoil/filters";
-import { _activeFields } from "../../recoil/schema";
 import { State } from "../../recoil/types";
 import { similarityParameters } from "../../components/Actions/Similar";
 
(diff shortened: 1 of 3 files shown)


app/packages/app/src/Root/Datasets/Dataset.tsx
```diff
@@ -97,17 +98,24 @@ export const Dataset: Route<DatasetQuery> = ({ prepared }) => {
...
-    update(({ reset }) => {
+    update(({ reset, get }) => {
       reset(filters);
-      reset(_activeFields({ modal: false }));
       reset(similarityParameters);
 
+      const newDataset = transformDataset(dataset);
+
+      const oldDataset = get(atoms.dataset);
+      oldDataset && console.log(oldDataset.id, newDataset.id);
```
ritch, line 108: Stray console.log?

benjaminpkane: Yes, thanks

app/packages/app/src/Root/Datasets/Dataset.tsx
```diff
@@ -97,17 +98,24 @@ export const Dataset: Route<DatasetQuery> = ({ prepared }) => {
...
+      if (!oldDataset || oldDataset.id !== newDataset.id) {
+        reset(_activeFields({ modal: false }));
```
ritch, line 108: Is there a reason this isn't in `useStateUpdate()`? Looks like there are similar things there.

ritch (approved):
FIx LGTM. Left a few comments, but shouldn't hold up the fix.

app/packages/app/src/Root/Datasets/Dataset.tsx
benjaminpkane: Good point, fixed

PR #1882, opened 2022-06-13
Fix selection and expansion for bad media files

### LINKED ISSUE #1865: [BUG] Cannot select sample with error in grid
When a sample's media fails to load, it enters an error state. This seems t prevent some actions such a selection and consequently tagging. Sample tagging and selection should be possible regardless of the error state
Screen Shot 2022-06-10 at 9 54 26 AM
.

benjaminpkane:
Resolves #1865

```py
import fiftyone as fo

dataset = fo.Dataset("bad")
sample = fo.Sample(filepath="bad.png")  
dataset.add_sample(sample)

# image
session = fo.Session(dataset)

dataset = fo.Dataset("bad-video")
sample = fo.Sample(filepath="bad.mp4")  
dataset.add_sample(sample)

# video
session.dataset = dataset
```

### DIFF
--- a/app/packages/looker/src/elements/common/thumbnail.module.css
+++ b/app/packages/looker/src/elements/common/thumbnail.module.css
@@ -11,6 +11,7 @@
   width: 100%;
   height: auto;
   position: absolute;
+  z-index: 1;
   top: 0;
   left: 0;
   opacity: 0;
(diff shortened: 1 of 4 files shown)


brimoor (approved):
LGTM

PR #1883, opened 2022-06-13
[WIP] Plugins

ritch:
See [here](https://github.com/voxel51/fiftyone/blob/a746c787d414cf004aac474ddc3bb971c3af25a4/app/packages/plugins/README.md) for the latest info on plugins.


ritch:
Plugins are now in the groups branch...this is no longer needed.

PR #1909, opened 2022-06-27
Fix custom_parser implementation in recipe

idow09:


brimoor (approved):
LGTM, thanks for the patch!

PR #1925, opened 2022-07-05
Adding weighted_sample() and balanced_sample() utils

brimoor:
Adds a `fiftyone.utils.random` module with two new helpful randomized sampling utils:
- `weighted_sample()`: Generates a random sample of size ``k`` from the given collection such that the probability of selecting each sample is proportional to the given per-sample weights
- `balanced_sample()`: Generates a random sample of size ``k`` from the given collection such that the expected histogram of ``path`` values in the sample is uniform

Also migrates the existing `fiftyone.utils.splits.random_split()` util to the `fiftyone.utils.random` module, because it seems like a better organization (put all randomized utils in one-place).

Example usage:

```py
import fiftyone as fo
import fiftyone.utils.random as four
import fiftyone.zoo as foz

# Sample proportional to label length
weights = dataset.values(F("ground_truth.label").strlen())
view1 = four.weighted_sample(dataset, 10000, weights)

# Now take a balanced sample from this unbalanced sample
view2 = four.balanced_sample(view1, 2000, "ground_truth.label")

# Plot results
plot1 = fo.CategoricalHistogram("ground_truth.label", init_view=dataset)
plot2 = fo.CategoricalHistogram(
    "ground_truth.label",
    order=lambda kv: -len(kv[0]),  # order by label length
    init_view=view1,
)
plot3 = fo.CategoricalHistogram("ground_truth.label", init_view=view2)
plot = fo.ViewGrid([plot1, plot2, plot3])
plot.show()
```

Screen Shot 2022-07-05 at 1 34 44 PM


ehofesmann (changes requested):
Love these utils!

I would prefer to name them something other than `XXX_sample()` since the term `sample` is pretty overloaded. Even in the example, your views were named `sample1` and `sample2` which gets confusing. Maybe `weighted_take()` or `balanced_take()`? It's not a view stage though so that might also be a bit confusing...
On that note, it might make more sense to just return the view (more similar to `take()`) directly rather than tagging the samples. You can always let the user `tag_samples()` afterwards if they want.

Also, I believe some of our blogs/examples also use `random_split()`, I'll double check and get a list of those that I'll update when this is released.

brimoor:
@ehofesmann thanks for the ideas!

I agree that always using tags to encode the splits was unnecessary complexity. My goal was to match the API of the `random_split()` method, which does use tags. But I think I found a cleaner approach 🤗 :

I made `tags` an optional argument for `weighted_sample()` and `balanced_sample()`, so no tags are used by default, and I added a new syntax to `random_split()`:

```
list, of, view = four.random_split(dataset, [list, of, fracs])
```

Regarding the "split" terminology, I agree that it's a bit overloaded with our `Sample`, but on the other hand taking a random "sample" is standard terminology in statistics, so I think it might be best to stick with the sample terminology in the `fiftyone.utils.random` module. Cool with that?

ehofesmann (approved):
I like the tag as an argument, good idea!
I suppose you're right, it being in the `random` module should add enough context for it to not be too confusing.

Send it!

PR #1927, opened 2022-07-06
Adding new_ids=True option to Dataset.add_collection()

brimoor:
Adds an optional `new_ids=True` option to `Dataset.add_collection()` that generates new sample/frame IDs when adding the samples to the collection.

This is useful, for example, to quickly exponentiate the size of a dataset for testing:

```py
# Make dataset 2 ** 4 times larger
for _ in range(4):
    dataset.add_collection(dataset, new_ids=True)
```

### Image example

```py
import fiftyone as fo
import fiftyone.zoo as foz

new_ids = dataset.add_collection(dataset, new_ids=True)

assert len(dataset) == 400
assert len(set(dataset.values("id"))) == 400
```

### Video example

```py
import fiftyone as fo
import fiftyone.zoo as foz

new_ids = dataset.add_collection(dataset, new_ids=True)

assert len(dataset) == 20
assert len(set(dataset.values("frames.id", unwind=True))) == 2558
```

### DIFF
--- a/tests/unittests/dataset_tests.py
+++ b/tests/unittests/dataset_tests.py
@@ -843,6 +843,57 @@ class DatasetTests(unittest.TestCase):
             )
 
     @drop_datasets
+    def test_add_collection(self):
+        sample1 = fo.Sample(filepath="image.jpg", foo="bar")
+        dataset1 = fo.Dataset()
+        dataset1.add_sample(sample1)
+
+        sample2 = fo.Sample(filepath="image.jpg", spam="eggs")
+        dataset2 = fo.Dataset()
+        dataset2.add_sample(sample2)
+
+        # Merge dataset
+        dataset = dataset1.clone()
+        dataset.add_collection(dataset2)
+
+        self.assertEqual(len(dataset), 2)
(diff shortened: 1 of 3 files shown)


tests/unittests/dataset_tests.py
```diff
@@ -842,6 +842,57 @@ def test_merge_samples_and_labels(self):
...
+        # Merge view
+        dataset = dataset1.clone()
+        dataset.add_collection(dataset2.exclude_fields("spam"))
+
+        self.assertEqual(len(dataset), 2)
+        self.assertTrue("spam" not in dataset.get_field_schema())
+        self.assertIsNone(dataset.last()["foo"])
+
+    @drop_datasets
+    def test_add_collection_new_ids(self):
```
manivoxel51, line 873: nice tests 💥

manivoxel51 (approved):
nice work Brian! 💯 

it mostly makes sense. specially after reading the tests

ehofesmann (approved):
LGTM, this would have definitely come in handy in the past when benchmarking datasets. Also love the tests!

PR #1931, opened 2022-07-07
Fix refresh after in App view change

benjaminpkane:
0n `0.16.5` after changing the view via the view bar, the App will break if clicking the logo to refresh the page.

Specifically, the browser history was previously not kept up to date with the new view when it was changed in the App.


brimoor (approved):
LGTM

Issue #1935, opened 2022-07-11
[FR] Add percentile (quantile) aggregation


tadejsv:
### Proposal Summary

I would like an aggregation that could compute a percentile/quantile. These values can sometimes be more informative than just mean / standard deviation (which are available now).

### What areas of FiftyOne does this feature affect?

-   [ ] App: FiftyOne application
-   [x] Core: Core `fiftyone` Python library
-   [ ] Server: FiftyOne server

### Willingness to contribute

The FiftyOne Community encourages new feature contributions. Would you or
another member of your organization be willing to contribute an implementation
of this feature?

-   [ ] Yes. I can contribute this feature independently.
-   [x] Yes. I would be willing to contribute this feature with guidance from
        the FiftyOne community.
-   [ ] No. I cannot contribute this feature at this time.

tadejsv:
Great, thank you 🙏

PR #1937, opened 2022-07-12
Adding quantiles aggregation

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1935.

Adds a `Quantiles` aggregation that can be used to compute the quantiles of numeric data.

### Notes

- Unfortunately, MongoDB has no builtin method for computing quantiles, so the implementation performs a full sort of the values. Thus `quantiles()` requires `O(n)` memory (DB-side, not client-side), not `O(1)` like other aggregations
- I benchmarked `$group` + `JS sort` versus `$sort` + `$group` on a dataset of 100K samples, and the JS sort version was faster
- I also tried https://stackoverflow.com/a/60694525/16823653, which was extremely slow

### Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

print(dataset.quantiles("uniqueness", [0.25, 0.50, 0.75]))
# [0.2202, 0.3377, 0.6255]

print(dataset.quantiles("uniqueness", 0.9))
# 0.6949

print(dataset.quantiles("predictions.detections.confidence", [0.25, 0.50, 0.75]))
# [0.0923, 0.2025, 0.5627]

print(dataset.quantiles("predictions.detections.confidence", 0.9))
# 0.9435
```

### DIFF
--- a/docs/source/user_guide/using_aggregations.rst
+++ b/docs/source/user_guide/using_aggregations.rst
@@ -282,6 +282,35 @@ collection:
     print(dataset.mean("predictions.detections.confidence"))
     # 0.34994137249820706
 
+.. _aggregations-quantiles:
+
+Quantiles
+_________
+
+You can use the
+:meth:`quantiles() <fiftyone.core.collections.SampleCollection.quantiles>`
+aggregation to compute the quantile(s) of the (non-``None``) values of a field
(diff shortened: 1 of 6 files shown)


benjaminpkane (approved):
LGTM

PR #1940, opened 2022-07-13
Extending stats() method to sample collections

brimoor:
This PR extends the existing `Dataset.stats()` method to all sample collections.

Some private methods for computing per-sample/frame sizes are also added. I could make them public but wasn't quite sold that this is something that belongs in the public namespace yet. Open to thoughts on this point though 🤗 

### Image dataset

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Full dataset stats
fo.pprint(dataset.stats(include_media=True))

# Full view stats (matches above)
fo.pprint(dataset.view().stats(include_media=True))

# Non-trivial view stats
fo.pprint(dataset[:100].select_fields().stats(include_media=True))

# Per sample sizes
fo.pprint(dataset[:10]._get_per_sample_bytes())
```

### Video dataset

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

# Full dataset stats
fo.pprint(dataset.stats(include_media=True))

# Full view stats (matches above)
fo.pprint(dataset.view().stats(include_media=True))

# Non-trivial view stats
fo.pprint(dataset[:5].select_fields().stats(include_media=True))

# Per-sample sizes
fo.pprint(dataset[:5]._get_per_sample_bytes())

# Per-sample total frame sizes
fo.pprint(dataset[:5]._get_per_sample_frames_bytes())

# Per-frame sizes
frame_ids = random.sample(dataset.values("frames.id", unwind=True), 5)
fo.pprint(dataset.select_frames(frame_ids)._get_per_frame_bytes())
```


benjaminpkane (approved):
LGTM. I'll note that these stats do not include dataset document size, or run info. Maybe that should be mentioned somewhere?

brimoor:
> LGTM. I'll note that these stats do not include dataset document size, or run info. Maybe that should be mentioned somewhere?

Good point 👍 I'll document this before merging

PR #1944, opened 2022-07-14
CVAT optimizations

ehofesmann:
There are a few instances in which the CVAT integration scaled poorly in regard to the number of tasks that exist on the CVAT server and the number of annotation runs on a dataset. This PR implements optimizations to improve these issues
Specifically:

1) When loading annotations, the existing tasks were explicitly listed only to determine if a given task id exists. This can take a few minutes if there are thousands of tasks on the CVAT server, which is much longer than any other operation for a small annotation run on only a few samples.
2) When performing operations on the `CVATAnnotationResults` like `get_status()`, it was previously connecting to the API and closing the connection every single time. Now, you can pass in an API object which will be used directly, allowing you to avoid needing to authenticate CVAT credentials every single time you want to `get_status()`, which can be a lot of there are hundreds of annotation runs on a dataset.


docs/source/integrations/cvat.rst
```diff
@@ -2413,6 +2413,26 @@ for that annotation run:
...
+    :class:`CVATAnnotationAPI <fiftyone.utils.cvat.CVATAnnotationAPI>` object as
+    defined in the next section. The benefit is that this allows you to avoid
+    authentication in each
+    method and significantly speeds up these calls if there are
+    multiple being performed at a time.
+
+    .. code-block:: python
+        :linenos:
+
+        api = results.connect_to_api()
```
brimoor, line 2430: My two cents: I think it'd be cleaner to use a pattern like this:

```py
class CVATBackend(object):

    def __init__(self):
        self._api = None

    def get_api(self):
        if self._api is None:
            self._api = self.connect_to_api()

        return self._api

    def use_api(self, api):
        self._api = api


class CVATAnnotationResults(object):

    def use_api(self, api):
        self._backend.use_api(api)

    def get_status(self):
        api = self._backend.get_api()
```

The example of sharing an `api` across results would be written like this:

```py
for anno_key in dataset.list_annotation_runs():

    # Optional optimization
    results.use_api(api)

    results.print_status()
    results.other_method()
    results.another_method()
```

In my view, the pattern of sharing an API instance across multiple results instances is too subtle of a use case to warrant polluting the entire public namespace of the `AnnotationResults` class with optional `api=None` args.

The real win here is that, if I understand correctly, the current implementation will recreate API instances within a single `CVATAnnotationResults` object; which is definitely wasteful and should be removed.

brimoor (changes requested):

docs/source/integrations/cvat.rst
ehofesmann: I like that, I was trying to figure out the best way to do something like this. Where I still get stuck is how to handle closing the API connection. This is currently occurring automatically at the end of methods like `print_status()` and if it always closes the connection, then you can reuse the same API object. What do you think is the best way to handle that?

brimoor: Is there a reason you added `api.close()` in the first place? I don't think it's actually necessary.

`Requests.Session` is an abstraction that maintains a pool of TCP connections under the hood for any requests to use. Closing the session clears the pool, but I'm pretty sure it does a good job of cleaning up after itself as you make requests.

The primary place you need to manually ensure things are closed is when doing a streaming request:

```py
# Use context manager to ensure streaming request is always closed
with session.get(url, stream=True) as response:
    ....
```

The context manager ensures the `session` will be told when to release the TCP connection that the request was using back to the pool.

brimoor: If you really wanted to design this out, you could use a pattern like this:

```py
class CVATAnnotationAPI(object):

    def __init__(self):
        self._api = None

    def __enter__(self):
        return self

    def __exit__(self, *args):
        self.close()


class CVATBackend(object):

    def __init__(self):
        self._api = None

    def __enter__(self):
        self.get_api().__enter__()
        return self

    def __exit__(self, *args):
        if self._api is not None:
            self._api.__exit__(*args)

    def get_api(self):
        if self._api is None:
            self._api = self.connect_to_api()

        return self._api

    def use_api(self, api):
        self._api = api


class CVATAnnotationResults(object):

    def __enter__(self):
        self._backend.__enter__()
        return self

    def __exit__(self, *args):
        self._backend.__exit__(*args)

    def use_api(self, api):
        self._backend.use_api(api)

    def get_status(self):
        api = self._backend.get_api()
```

With the end result that, if a user really wanted to be sure that no TCP connections exist after a block of code, they could use this syntax:

```py
with results:
    # do stuff
```

brimoor:
@ehofesmann can we get this merged this week?

brimoor:
@ehofesmann LMK what you think about adding https://github.com/voxel51/fiftyone/pull/1997 to this PR before merging it 📈

brimoor (approved):
LGTM

@ehofesmann I recall you mentioning that tests might need to be updated? Just a reminder in case you're not ready to merge yet

ehofesmann:
> LGTM
> 
> @ehofesmann I recall you mentioning that tests might need to be updated? Just a reminder in case you're not ready to merge yet

I looked into it more and the existing tests were sufficient, thank for checking on that though!

PR #1949, opened 2022-07-15
Fix App masks

benjaminpkane:
Resolves a bug introduced in #1943.

Mask values are dicts, not lists. Fixes the default value.


brimoor (approved):
LGTM

PR #1955, opened 2022-07-18
Groups (adding App support)

ritch:


brimoor:
@ritch Just noticed this PR adds 1.1M lines! Looks like this is mostly due to some example `.pcd` data included in the plugin example docs. Can we purge the data from the repo and instead make it available via a Google Drive link or similar?

I'd like to keep the repo stats realistic and also keep the cloned repository size under control.

Issue #1966, opened 2022-07-26
[BUG] Cannot connect classification confusion matrix to App


brimoor:
On `fiftyone==0.16.5`, running this code to connect a confusion matrix for a `Classification` task to the App should be valid:

```py
import fiftyone as fo

filepath = "/path/to/image.jpg"

sample1 = fo.Sample(
    filepath=filepath,
    gt=fo.Classification(label="cat"),
    pred=fo.Classification(label="cat", confidence=0.99),
)
sample2 = fo.Sample(
    filepath=filepath,
    gt=fo.Classification(label="dog"),
    pred=fo.Classification(label="dog", confidence=0.8),
)
sample3 = fo.Sample(
    filepath=filepath,
    gt=fo.Classification(label="dog"),
    pred=fo.Classification(label="cat", confidence=0.1),
)

dataset = fo.Dataset()
dataset.add_samples([sample1, sample2, sample3])

results = dataset.evaluate_classifications("pred", gt_field="gt", eval_key="eval")

plot = results.plot_confusion_matrix()
plot.show()

session = fo.launch_app(dataset)

# Raises the error shown below
session.plots.attach(plot)
```

but currently raises the following error:

```
ValueError: Invalid label field type <class 'fiftyone.core.labels.Classification'>. Extracting patches is only supported for the following types: (<class 'fiftyone.core.labels.Detections'>, <class 'fiftyone.core.labels.Polylines'>)
```

The problem is that the plot's `selection_mode` defaults to `"patches"`, but that shouldn't happen for classification tasks.

Therefore, one can workaround this issue by manually setting the plot's `selection_mode` to `"select"` or `"match"` prior to attaching the plot to the App:

```py
# Avoids the error
plot.selection_mode = "match"
session.plots.attach(plot)
```

PR #1967, opened 2022-07-26
Fixing #1966

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1966

How I tested (in addition to unit test coverage):
- Checking that error in https://github.com/voxel51/fiftyone/issues/1966 no longer occurs
- Checking that [this code](https://voxel51.com/docs/fiftyone/user_guide/evaluation.html#confusion-matrices) still runs


ehofesmann (approved):
LGTM

PR #1975, opened 2022-07-29
[WIP] Catch UnicodeDecodeError when reading invalid file

democat3457:

Catch a UnicodeDecodeError if you pass in an invalid file to `_read_file_lines()`. This is (from my experience) caused by passing in an image rather than a txt file, which can happen in a particular scenario where yolo labels are loaded from a folder containing both images and label text files (where the corresponding label file may not exist).

Pass an image to `_read_file_lines`. (is there something else I'm supposed to put here?)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/yolo.py
+++ b/fiftyone/utils/yolo.py
@@ -1095,7 +1095,11 @@ def _write_yaml_file(d, path, **kwargs):
 
 def _read_file_lines(path):
     with open(path, "r") as f:
-        lines = [l.strip() for l in f.read().splitlines()]
+        try:
+            lines = [l.strip() for l in f.read().splitlines()]
+        except UnicodeDecodeError:
+            logger.warning("Error while decoding file %s, ignoring", path)
+            return []
         return [l for l in lines if l]
 
 


brimoor:
Hi @democat3457 👋 

Can you explain more about how your YOLO data is formatted on disk to cause this issue?

It sounds like the problem is that FiftyOne thinks that an image file in your setup should contain labels, so I'd rather find and fix that issue rather than saying "oh well, we'll just say image file == no labels"

democat3457:
@brimoor Gotcha! The way we are storing our dataset on disk is that we put images and YOLO label files in the same folder (obviously differing from the normal format of train/val split). I use something similar to the following procedure to load our dataset:
1. Load images only using `fo.Dataset.from_images_dir` on the dataset folder
2. Load ground-truth labels using `fouy.add_yolo_labels` on the dataset folder <- errors under specific circumstance, below
3. Load model detection labels using `fouy.add_yolo_labels` on the detection output folder

Because we pass in the dataset folder as a string to no. 2, the match type in `add_yolo_labels` is selected as `'uuid'`.

The situation where the UnicodeDecodeError occurs is when an image exists, but no corresponding label file with the same basename exists. I'm assuming that somehow, the image file is getting passed into `load_yolo_annotations` rather than adding it to `bad_paths`. Additionally, it's really hard to troubleshoot because there is no print for which file the UnicodeDecodeError occurred on, so debug prints have to be added manually into the python source files to figure out which image has no corresponding label file.

brimoor:
@democat3457 Ah thanks for clarifying! I think the best solution here is for you to instead update this logic: 

https://github.com/voxel51/fiftyone/blob/44fcf0d98594af21bc446c116ea692c4934d7322/fiftyone/utils/yolo.py#L72-L75

to only match on `.txt` files:

```py
        txt_map = {
            os.path.splitext(p)[0]: os.path.join(labels_path, p)
            for p in etau.list_files(labels_path, recursive=True)
            if p.endswith(".txt")
        }
```

Then the `add_yolo_labels()` method should have the behavior you want 🙌

PR #1976, opened 2022-08-02
GeoJSONField Map Plot

benjaminpkane:
Functionally complete and ready to demo with `quickstart-geo`, etc. Wrapping up build/plugin details.

```
bash install.bash
```

One question is if selection should directly apply a new `Select` stage to the view bar, or maintain an extended view stage like sort by similarity in the App that can overwrite itself. And similarly be saved view pinning/saving the extended view.

Screen Shot 2022-08-02 at 5 32 02 PM


brimoor (changes requested):
Slick stuff! Love how the map updates to a tight crop of the current view whenever the extended view is updated!

I definitely think that the map plugin should be wired up to the extended view, not directly to the view bar. Every selection strictly overrides any previous selection; no need for stages to accumulate in the view bar.

A couple feature requests:
- Can you add a reset button that resets to a tight crop of the current extended view?
- Can you add `ESC` as a keyboard shortcut for the reset action? 😄 
- Can you add a `?` button that links to a to-be-written (by me) documentation section on using this feature? See below for more context

Usability issues:

I'm seeing some inconsistent behavior with "submitting" a lasso selection. I think I can either press `enter` or double-click the last point to submit the selection, but sometimes the view doesn't update. I can demo this live if necessary. Could selections involving "binned bubbles" be an issue?

brimoor:
Question: does this plugin expose any settings? EG the ability to configure how aggressively things are binned, or the marker sizes?

benjaminpkane:
Wrapping up addressing comments now.

> Question: does this plugin expose any settings? EG the ability to configure how aggressively things are binned, or the marker sizes/colors?

Yes, it can. Everything regarding the data and styling is customizable through the package we are using.

brimoor:
> Wrapping up addressing comments now.
> 
> > Question: does this plugin expose any settings? EG the ability to configure how aggressively things are binned, or the marker sizes/colors?
> 
> Yes, it can. Everything regarding the data and styling is customizable through the package we are using.

Sweet! If you can provide an example `app_config.json` entry with all the possible values then I'd be happy to document it

benjaminpkane:
> > Wrapping up addressing comments now.
> > > Question: does this plugin expose any settings? EG the ability to configure how aggressively things are binned, or the marker sizes/colors?
> > 
> > 
> > Yes, it can. Everything regarding the data and styling is customizable through the package we are using.
> 
> Sweet! If you can provide an example `app_config.json` entry with all the possible values then I'd be happy to document it

Ok, the main configuration here is the circle sizing.

```json
{
  "plugins": {
    "map": {
      "circle-color": "rgb(244, 113, 6)",
      "circle-opacity": 0.7,
      // Use step expressions (https://docs.mapbox.com/mapbox-gl-js/style-spec/#expressions-step)
      "circle-radius": ["step", ["get", "point_count"], 20, 10, 30, 25, 40],
      "point-radius": 4
    }
  }
}
```

benjaminpkane:
Comments addressed

benjaminpkane:
Working through a couple bugs

brimoor:
New controls look good!

Bug:
- Currently the view bar is broken (doesn't accept new stages, or the bookmark icon)

Feature request:
- I don't believe it is currently possible to disable binning. Can you support a config setting like `circle-radius: null` for that? Or perhaps a `circle-threshold: <min-num-points>` setting that enables the user to say that they don't want binning if less than a certain number of points are visible in the current viewport. The reason here is that, for some applications like reviewing drive sequences, it is important to see individual points to isolate when the vehicle was at a particular location

Question:
- What's the status on the need for a mapbox API token? Is one required to exceed a certain volume of API calls? Can one be provided via the App config?

app/packages/map/src/state.ts
```diff
@@ -0,0 +1,79 @@
...
+  clustering?: boolean;
+  clusters?: {
+    textPaint: mapboxgl.SymbolPaint;
+    paint: mapboxgl.CirclePaint;
+  };
+  pointPaint?: mapboxgl.CirclePaint;
+  mapboxAccessToken: string;
+}
+
+export const defaultSettings = Object.freeze({
```
benjaminpkane, line 14: default settings are here

benjaminpkane:
Ok, comments addressed! The token is now a setting, bug fixed, and clustering is optional.

brimoor:
Sweet, I'll give it a test tonight.

For documentation purposes, what are the requirements around Mapbox API tokens? Is this working out of the box with a default token or no token at all but will stop working at some point?

PR #1997, opened 2022-08-11
Upgrading AnnotationAPI reuse to interface level

brimoor:
@ehofesmann I upgraded the `connect_to_api()` notion to an interface-level thing. So, for example, now the Labelbox backend can use the same API-sharing features "for free". The updated docs show the idea.

LMK what you think about this proposed upgrade to add to #1944 before merging it 😄

### DIFF
--- a/docs/source/integrations/cvat.rst
+++ b/docs/source/integrations/cvat.rst
@@ -2366,18 +2366,19 @@ Using the CVAT API
 ------------------
 
 You can use the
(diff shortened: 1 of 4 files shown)


ehofesmann (approved):
I like it! This is much better, there wasn't any reason that `connect_to_api()` needed to be tied to the annotation results object. I believe will need to update the CVAT tests after this is merged into #1944

PR #1999, opened 2022-08-12
Switching to list-based storage of saved views

brimoor:
Makes some changes to the saved view dataset model that will be necessary to implement the frontend design we'd like to have for this feature:
- Switches to list-based storage of saved views on `DatasetDocument` so that views can be ordered and reordered in the frontend
- Adds `created_at`, `last_modified_at`, and `last_loaded_at` fields
- Adds `color` (with validation) and `description` fields
- Adds a `url_name` field that contains a `url-friendly-name` for view names `URL Friendly Name!`
- Adds a `get_view_info()` method to retrieve the editable info about a saved view
- Adds an `update_view_info()` method to edit a saved view's info

Relevant private methods that the App backend can use:
- `_reorder_views()`: change the order of the saved views on a dataset
- `_validate_view_name()`: validate that a new saved view name is available (both name and URL name)

## URL-friendly names discussion

Here's how `to_url_name()` is currently defined:

```py
def to_url_name(name):
    """Returns a URL-friendly version of the given string.

    The following strategy is used to generate URL-friendly strings:

        -   The characters ``A-Za-z0-9-_.`` are left unchanged
        -   Whitespace and ``+`` are converted to ``-``
        -   All other characters are omitted
        -   All consecutive ``-`` characters are reduced to a single ``-``
        -   All leading and trailing ``-`` are stripped
        -   The resulting string must be ``[1, 100]`` characters in length

    Examples::

        Input name                       | URL-friendly name
        ---------------------------------+-----------------------
        coco_2017                        | coco_2017
        c+o+c+o 2-0-1-7                  | c-o-c-o-2-0-1-7
        cat.DOG                          | cat.DOG
        ---name----                      | name
        Brian's #$&@ (Awesome?) Dataset! | Brians-Awesome-Dataset
        sPaM     aNd  EgGs               | sPaM-aNd-EgGs
    """
```

Below are my notes that led me to propose the above behavior as our default:

#### Github repository names
Github allows `[A-Za-z0-9_.-]` and transforms all other characters to `"-"`
https://stackoverflow.com/a/59082561/16823653

#### Github branches and tags
Github allows `A-Za-z0-9_.-/`
https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names

#### Our use case
We'll use URL-friendly dataset and saved view names like so:

```
https://fiftyone.ai/my-dataset?view=my-view
```

Considerations
- If we wanted to allow `.` in names, we can't use them as document keys in MongoDB collections (currently we're not using names as keys)
- We can't allow `/` in dataset names (without escaping) because it will interfere with routing `fiftyone.ai/bad/dataset/name`

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -2304,7 +2304,9 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
             _view_doc = self._get_view_doc(name, pop=True)
             _view_doc.delete()
 
+        index = max([v.index or 0 for v in self._doc.views], default=-1) + 1
         now = datetime.utcnow()
+
         view_doc = foo.ViewDocument(
             dataset_id=self._doc.id,
             name=name,
@@ -2315,7 +2317,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
                 json_util.dumps(s)
                 for s in view._serialize(include_uuids=False)
             ],
-            index=len(self._doc.views),
+            index=index,
             created_at=now,
             last_modified_at=now,
         )
@@ -2444,10 +2446,18 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         return self._doc.views[idx]
 
     def _ordered_views(self):
-        return sorted(self._doc.views, key=lambda v: v.index)
+        return sorted(self._doc.views, key=lambda v: v.index or 0)
 
     def _reorder_views(self, new_order):
-        index_map = {o: i for i, o in enumerate(new_order)}
+        old_indexes = set(v.index for v in self._doc.views)
+        new_indexes = set(new_order)
+        if old_indexes != new_indexes:
+            raise ValueError(
+                "New indexes %s don't match old indexes %s"
+                % (new_indexes, old_indexes)
+            )
+
+        index_map = {old: new for new, old in enumerate(new_order)}
 
         for view_doc in self._doc.views:
             view_doc.index = index_map[view_doc.index]
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -6,7 +6,6 @@ FiftyOne datasets.
 |
 """
 from collections import defaultdict
-from copy import deepcopy
 from datetime import datetime
 import fnmatch
 import itertools
@@ -26,13 +25,9 @@ import eta.core.serial as etas
 import eta.core.utils as etau
 
 import fiftyone as fo
-import fiftyone.core.aggregations as foa
-import fiftyone.core.annotation as foan
-import fiftyone.core.brain as fob
 import fiftyone.constants as focn
 import fiftyone.core.collections as foc
-import fiftyone.core.expressions as foex
-import fiftyone.core.evaluation as foe
+import fiftyone.core.expressions as foe
 import fiftyone.core.fields as fof
 import fiftyone.core.frame as fofr
 import fiftyone.core.labels as fol
@@ -268,7 +263,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         if isinstance(id_filepath_slice, slice):
             return self.view()[id_filepath_slice]
 
-        if isinstance(id_filepath_slice, foex.ViewExpression):
+        if isinstance(id_filepath_slice, foe.ViewExpression):
             return self.view()[id_filepath_slice]
 
         if etau.is_container(id_filepath_slice):
@@ -2264,9 +2259,16 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         Returns:
             a list of saved view names
         """
-        return sorted(self._doc.views.keys())
+        return [view_doc.name for view_doc in self._doc.views]
 
-    def save_view(self, name, view, overwrite=False):
+    def save_view(
+        self,
+        name,
+        view,
+        description=None,
+        color=None,
+        overwrite=False,
+    ):
         """Saves the given view into this dataset under the given name so it
         can be loaded later via :meth:`load_view`.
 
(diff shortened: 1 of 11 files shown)


fiftyone/core/dataset.py
```diff
@@ -2293,27 +2295,91 @@ def save_view(self, name, view, overwrite=False):
...
             dataset_id=self._doc.id,
+            name=name,
+            url_name=fou.to_url_name(name),
+            description=description,
+            color=color,
             view_stages=[
                 json_util.dumps(s)
                 for s in view._serialize(include_uuids=False)
             ],
+            index=len(self._doc.views),
```
manivoxel51, line 2318: what does the index do?

fiftyone/core/dataset.py
```diff
@@ -2293,27 +2295,91 @@ def save_view(self, name, view, overwrite=False):
...
+            info["name"] = "a new name"
+
+            dataset.update_view_info("test", info)
+
+        Args:
+            name: the name of a saved view
+            info: a dict whose keys are a subset of the keys returned by
+                :meth:`get_view_info`
+        """
+        view_doc = self._get_view_doc(name)
```
manivoxel51, line 2364: is it worth checking if the view exists?

fiftyone/core/dataset.py
```diff
@@ -2337,23 +2403,23 @@ def load_view(self, name):
         Returns:
             a :class:`fiftyone.core.view.DatasetView`
         """
-        if name not in self._doc.views:
-            raise ValueError("Dataset has no saved view '%s'" % name)
+        view_doc = self._get_view_doc(name)
```
manivoxel51, line 2406: same here. Is it worth checking that the view exists

fiftyone/core/fields.py
```diff
@@ -159,6 +160,16 @@ class StringField(mongoengine.fields.StringField, Field):
     pass
 
 
+class ColorField(StringField):
+    """A string field that holds a hex color string like '#FF6D04'."""
```
manivoxel51, line 164: 👍

fiftyone/core/utils.py
```diff
@@ -1468,3 +1470,78 @@ def _load_until(self, goal_position):
...
+def to_url_name(name):
+    """Returns a URL-friendly version of the given string.
+
+    The following strategy is used to generate URL-friendly strings:
+
+        -   The characters ``A-Za-z0-9-_.`` are left unchanged
+        -   Whitespace and ``+`` are converted to ``-``
+        -   All other characters are omitted
+        -   All consecutive ``-`` characters are reduced to a single ``-``
+        -   All leading a trailing ``-`` are stripped
```
manivoxel51, line 1499: did you mean `leading and trailing`

fiftyone/core/utils.py
```diff
@@ -1468,3 +1470,78 @@ def _load_until(self, goal_position):
...
+    if c in _SAFE_CHARS:
+        return c
+
+    if c in _WHITESAPCE_CHARS:
+        return "-"
+
+    return ""
+
+
+def to_url_name(name):
```
manivoxel51, line 1490: I wish we could use a lib for this logic. it is pretty critical little piece of code and rather not maintain it :) https://www.npmjs.com/package/url-safe-string seems to possibly be able to cover most scenarios. totally up to you :)

manivoxel51 (approved):
nice work @brimoor! 🍨 

left couple of comments

fiftyone/core/dataset.py
brimoor: That logic is now in `_get_view_doc()`

fiftyone/core/dataset.py
brimoor: `index` is so the views can be ordered. The figma design allows them to be reordered so that you can have your favorite view listed first, for example.

I added `_ordered_view()` and `_reorder_views()` as private methods to demonstrate how `index` is used. Also added some unit tests for this functionality.

LMK if you have any thoughts on a better way to implement "views can be reordered"

fiftyone/core/utils.py
brimoor: ty!

fiftyone/core/utils.py
brimoor: Hmm I assumed that there wouldn't be anything out there flexible enough to do what I had in mind, but it does look like that package does everything except stripping consecutive, leading, and trailing `-`.

Another point of discussion: I guess we'll want the URL validation logic always (only) in the backend, since it needs to run against data in the database to verify availability.

I added some more of my notes from last night when trying to decide what to propose for generating URL slugs for dataset names (I added it for saved views here, but we need to add it for dataset names too).

Any thoughts on whether you like the current logic (implementation aside)?

fiftyone/core/dataset.py
manivoxel51: makes sense 👍

fiftyone/core/utils.py
manivoxel51: I like it! `-` makes sense as a separator for url friendliness. curious why you chose to allow `.` and `_`. To me, it simpler and more readable to always enforce a single separator character, `-`.

brimoor: I definitely agree that a smaller character set for URLs is more readable.

The only consideration I had was that, if we were to convert `_.` to `-` and also convert everything to lowercase, for example (also more readable URLs) then a user would have a smaller set of available view names.

For example, if they name a view `my_view`, then they wouldn't be allowed to name another view `My View!`, because they'd both have URL name `my-view`.

Now, should the user really be using both `my_view` and `My View!`? We could definitely decide for them that they cannot do that, if folks feel strongly about it.

fiftyone/core/dataset.py
brimoor: FYI I removed `index` for now and am instead relying on the ordering of the `DatasetDocument.views` list to convey this information

PR #2003, opened 2022-08-17
v0.17.0rc0

benjaminpkane:
Feature release including plugins, geo location maps, group datasets, and point clouds.


brimoor (approved):
LGTM

PR #2006, opened 2022-08-18
Label studio updates

ehofesmann:
Updates the label studio integration with the following:
* Raises error when annotating video, Classifications, and scalar labels which are not currently supported
* Adds support for non-submitted tasks
* Tracks label ids to properly merge preexisting annotations
* Use default annotation config in test
* Update file docstring headers


brimoor (approved):
LGTM

@ehofesmann can you test once more before merging to make sure my editing pass didn't break anything? 😄

PR #2009, opened 2022-08-20
Package-wide logging config

brimoor:
https://github.com/voxel51/eta/pull/567 removed ETA's configuration of the root logger, which was a good change because the root logger shouldn't have been touched.

However, FiftyOne was implicitly built assuming all of its `logging.INFO` messages would be printed, which is no longer true:

```py
import fiftyone.zoo as foz

# This used to print some persistent output, but with latest `eta`, it will not
```

This PR re-introduces package-wide logging at `INFO` level by default, and exposes a `fo.config.logging_level` setting that can be used to customize this behavior if desired:

```shell
export FIFTYONE_LOGGING_LEVEL=WARNING
```


benjaminpkane (approved):
LGTM 👍

PR #2011, opened 2022-08-21
Adding ipywidgets<8 requirement

brimoor:
`ipywidgets==8.0.0` was [released August 18th](https://pypi.org/project/ipywidgets/#history), but unfortunately [it looks like](https://pythontechworld.com/issue/plotly/plotly.py/3686) it is not compatible with `plotly<=5.9.0`, as the following error is raised (confirmed locally):

```
AttributeError: type object 'DOMWidget' has no attribute '_ipython_display_'
```

Therefore, this PR adds an `ipywidgets<8` requirement to FiftyOne.

### DIFF
--- a/docs/source/user_guide/plots.rst
+++ b/docs/source/user_guide/plots.rst
@@ -145,7 +145,7 @@ below for your environment:
 
     .. code-block:: shell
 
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
LGTM

PR #2012, opened 2022-08-22
Save context tweaks

brimoor:
Completes the implementation of the dataset save context started in https://github.com/voxel51/fiftyone/pull/1727.

### Change log

- Added support for dynamically batching samples to achieve a target number of database connections per second
- Updated the default batching strategy to save batches every 0.2 seconds (this is the same strategy used by `add_samples()`)
- Added a `save_context()` syntax
- Added proper reloading of `Sample`/`Frame` singletons when applying save contexts to dataset *views*
- Added documentation
- Added unit tests

### Example usage

The `iter_samples(autosave=True)` syntax is the same:

```py
import random as r
import string as s

import fiftyone as fo
import fiftyone.zoo as foz

def make_label():
    return "".join(r.choice(s.ascii_letters) for i in range(10))

# No save context
    sample.ground_truth.label = make_label()
    sample.save()

# Save in batches of 10
for sample in dataset.iter_samples(progress=True, autosave=True, batch_size=10):
    sample.ground_truth.label = make_label()

# Save every 0.5 seconds
for sample in dataset.iter_samples(progress=True, autosave=True, batch_size=0.5):
    sample.ground_truth.label = make_label()
```

And here's how the new `save_context()` syntax looks:

```py
# No save context
    sample.ground_truth.label = label()
    sample.save()

# Save in batches of 10
with dataset.save_context(batch_size=10) as context:
        sample.ground_truth.label = make_label()
        context.save(sample)

# Save every 0.5 seconds
with dataset.save_context(batch_size=0.5) as context:
        sample.ground_truth.label = make_label()
        context.save(sample)
```

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -9,9 +9,11 @@ from collections import defaultdict
 import fnmatch
 import itertools
 import logging
+import numbers
 import os
 import random
 import string
+import timeit
 import warnings
 
 from bson import ObjectId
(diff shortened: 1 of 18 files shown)


fiftyone/core/collections.py
```diff
@@ -69,49 +71,97 @@ def registrar(func):
...
+            raise ValueError(
+                "Dataset context '%s' cannot save sample from dataset '%s'"
+                % (self._dataset.name, sample._dataset.name)
+            )
 
-    def save_sample(self, sample):
-        self._samples.append(sample)
-        if len(self._samples) >= self._batch_size:
-            self._save_samples()
+        sample_op, frame_ops = sample._save(deferred=True)
```
brimoor, line 128: I'm calling `_save()` here rather than waiting until `_save_batch()` is called to spread the computation across calls to `save()`. If the user is printing a progress bar for the *input* samples, this makes it smoother.

fiftyone/core/collections.py
```diff
@@ -69,49 +71,97 @@ def registrar(func):
...
+        sample_collection: a
+            :class:`fiftyone.core.collections.SampleCollection`
+        batch_size (None): the batching strategy to use. Can either be an
+            integer specifying the number of samples to save in a batch, or a
+            float number of seconds between batched saves
+    """
+
+    def __init__(self, sample_collection, batch_size=None):
+        if batch_size is None:
+            batch_size = 0.2
```
brimoor, line 88: Time-based seems like a reasonable default to me because it controls the DB connection rate (soon, the API request rate), which is something one often wants to limit. 

A fixed batch size would produce wildly different connection rates since the size of `Sample` objects may vary from tiny to 16MB

fiftyone/core/odm/document.py
```diff
@@ -523,164 +524,108 @@ class Document(BaseDocument, mongoengine.Document):
...
+            clean (True): whether to call the document's ``clean()`` method.
+                Only applicable when ``validate`` is True
 
         Returns:
             self
         """
+        self._save(deferred=False, validate=validate, clean=clean, **kwargs)
+        return self
+
+    def _save(self, deferred=False, validate=True, clean=True, **kwargs):
```
brimoor, line 543: Went ahead and merged the deferred vs non-deffered logic into a single function to avoid duplication

j053y (approved):

PR #2015, opened 2022-08-22
Error handling for Looker3D

ritch:
Fixes crashing due to either missing files or misconfiguration. Instead displays error message.

### DIFF
--- a/app/packages/components/src/components/Loading/Loading.module.css
+++ b/app/packages/components/src/components/Loading/Loading.module.css
@@ -5,8 +5,10 @@
   align-items: center;
   width: 100%;
   height: 100%;
+  padding: 0 2rem;
   color: var(--font-dark);
   background: transparent;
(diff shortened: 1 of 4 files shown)


benjaminpkane:
Maybe we can use the `@fiftyone/components`[error boundary component](https://github.com/voxel51/fiftyone/tree/develop/app/packages/components/src/components/ErrorBoundary) here?

Another thing that came up here is that the activator may use improper settings, which is not contained by this error boundary. In that case, errors would be caught by the global error boundary. That's ok with me, just noting it.

ritch:
> Maybe we can use the @fiftyone/components[error boundary component](https://github.com/voxel51/fiftyone/tree/develop/app/packages/components/src/components/ErrorBoundary) here?

I tried this initially and it wouldn't handle all errors. I do want to de-duplicate the two.

Issue #1903, opened 2022-06-21
[BUG] YOLOv5 yaml path ignored


victor1cea:
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 22.04
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: 0.16.3
-   **Python version**: 3.9.12

### Describe the problem
The 'path' from dataset.yaml in YOLOv5 datasets is ignored/not handled.
Example of dataset.yaml stored
```
names:
[cars]
nc: 1
train: ./images/train
val: ./images/val
test: ./images/test
path: some/path # ignored
```
`path` should be used like a base path for the splits e.g. train path would actually be `some/path/images/train`. The importer ignores this and the splits are not found. The exporter does not write `path: ...` in the exported `dataset.yaml`.
### Code to reproduce issue

Provide a reproducible test case that is the bare minimum necessary to generate
the problem.
```
# read any yolov5 and have values for "path" key in dataset.yaml
yolo_dataset: fo.Dataset = fo.Dataset(name)
for split in splits:
    yolo_dataset.add_dir(
        yaml_path=yaml_path,
        dataset_type=fo.types.YOLOv5Dataset,
        split=split,
        tags=split,
        include_all_data=True
    )

# export the dataset
split_view = self.yolo_dataset.match_tags(split)
split_view.export(export_dir,
                  yaml_path=export_dir + '/dataset.yaml',
                  dataset_type=fo.types.YOLOv5Dataset,
                  label_field='ground_truth',
                  split=split,
                  include_confidence=True
                  )

```

### What areas of FiftyOne does this bug affect?

-   [ ] `App`: FiftyOne application issue
-   [x] `Core`: Core `fiftyone` Python library issue
-   [ ] `Server`: Fiftyone server issue

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another
member of your organization be willing to contribute a fix for this bug to the
FiftyOne codebase?

-   [x] Yes. I can contribute a fix for this bug independently.
-   [ ] Yes. I would be willing to contribute a fix for this bug with guidance
        from the FiftyOne community.
-   [ ] No. I cannot contribute a bug fix at this time.

brimoor:
Hi @victor1cea 👋 

[...]

PR #2016, opened 2022-08-23
Fix Issue #1903 path variable

victor1cea:

Adds support for `path`  variable in `dataset.yaml` for YOLOv5 datasets, as mentioned in Issue #1903.
The docs have been updated too. 

Created a dataset and tested the new possible file structures e.g. `dataset.yaml` located outside of the actual dataset directory.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/docs/source/user_guide/dataset_creation/datasets.rst
+++ b/docs/source/user_guide/dataset_creation/datasets.rst
@@ -2225,6 +2225,7 @@ where `dataset.yaml` contains the following information:
 
 .. code-block:: text
 
+    path: <dataset_dir> # optional
     train: ./images/train/
     val: ./images/val/
 
@@ -2238,7 +2239,9 @@ See `this page <https://docs.ultralytics.com/tutorials/train-custom-datasets>`_
 for a full description of the possible format of `dataset.yaml`. In particular,
 the dataset may contain one or more splits with arbitrary names, as the
 specific split being imported or exported is specified by the `split` argument
-to :class:`fiftyone.utils.yolo.YOLOv5DatasetImporter`.
+to :class:`fiftyone.utils.yolo.YOLOv5DatasetImporter`. `dataset.yaml` can also
+be located outside of `<dataset_dir>` as long as the optional `path` is set to
+point to `<dataset_dir>`.
 
 .. note::
 
@@ -2259,7 +2262,9 @@ where `<target>` is the zero-based integer index of the object class label from
 `[0, 1] x [0, 1]`, and `<confidence>` is an optional detection confidence in
 `[0, 1]`.
 
-Unlabeled images have no corresponding TXT file in `labels/`.
+Unlabeled images have no corresponding TXT file in `labels/`. The label file
+path for each image is obtained by replacing `images/` with `labels/` in the
+respective image path.
 
 The image and labels directories for a given split may contain nested
 subfolders of parallelly organized images and labels.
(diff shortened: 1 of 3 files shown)


brimoor (approved):
LGTM. Thanks! 🚀

alexk-ede:
I'm not sure if I should open a separate issue or not.

I updated to  fiftyone==0.16.6
and 
I'm basically exporting it via
        split_view.export(
            export_dir=export_dir,
            dataset_type=dataset_type,
            label_field=label_field,
            split=split_out,
            classes=final_labels_list)

(dataset_type is fo.types.COCODetectionDataset ofc)

But it's still not writing the path to the dataset.yaml file.

victor1cea:
@alexk-ede the changes I made are only for YOLOv5 datasets. Also, how are you using `dataset.yaml`  with COCO datasets? because COCO uses `.json` files afaik.

alexk-ede:
@victor1cea Oh, sry for the confusion, I copy pasted the wrong dataset_type. 
I'm just getting some labels out of coco and oi6 and exporting them to yolov5.

So in my custom converter function it's currently set as a default argument dataset_type=fo.types.YOLOv5Dataset. 
And yolov5 accepts the dataset only when I manually add the path variable. Even after the 0.16.6 update.

alexk-ede:
So here is a minimal example (based on the yolov5 export example), 
which highlights actually 2 problems.

```
import fiftyone as fo
import fiftyone.zoo as foz

coco_classes = ["car"]
coco2017_dataset1 = foz.load_zoo_dataset(
    "coco-2017",
    classes=coco_classes,
)

classes_filter = coco_classes # using all

## adapted from the example https://voxel51.com/docs/fiftyone/user_guide/export_datasets.html#yolov5dataset
export_dir = "yolov5-dataset"
label_field = "ground_truth"  # for example

# The splits to export
splits = ["train", "val"]

# All splits must use the same classes list
classes = classes_filter

matching_view = coco2017_dataset1.filter_labels(label_field, F("label").is_in(classes_filter))

# The dataset or view to export
# We assume the dataset uses sample tags to encode the splits to export  
dataset_or_view = matching_view


# Export the splits
for split in splits:
    split_view = dataset_or_view.match_tags(split)
    split_view.export(
        export_dir=export_dir,
        dataset_type=fo.types.YOLOv5Dataset,
        label_field=label_field,
        split=split,
        classes=classes,
        export_media="symlink", # speedup
    )
```

This yields

```
 100% |█████████████| 12251/12251 [27.0s elapsed, 0s remaining, 455.3 samples/s]      
Directory 'yolov5-dataset' already exists; export will be merged with existing files
 100% |█████████████████████| 0/0 [86.6ms elapsed, ? remaining, ? samples/s] 
```

and a `dataset.yaml` with only

```
names:
- car
nc: 1
train: .\images\train\
val: .\images\val\
```


So 1. this still doesn't export the path, so this dataset just doesn't work with yolov5 unless it's placed in the correct folder where yolov5 expects it to be.
(and the second problem is that the example code doesn't properly export the val part of the dataset, bc it's usually called validation in the datasets, but it's unrelated to this issue)

victor1cea:
@alexk-ede  Oh, I see now. The `path` can be read, but will never be exported. I could fix it this week.
Regarding the second problem,  you could change the tags of the samples from the validation split like this:

```
for sample in dataset:
    if 'validation' in sample.tags:
        sample.tags = ['val']
        sample.save()
```

PR #2017, opened 2022-08-23
Adding client compatibility versions

brimoor:
This PR introduces a `fiftyone.constants.COMPATIBLE_VERSIONS` constant that allows a given FiftyOne version to declare that it is compatible with a range of other versions.

When clients running this patch encounter a dataset with a compatible version, they will not migrate the dataset (unless a migration was specifically requested via `fiftyone migrate`); instead the dataset will be loaded and used as-is.

For example, since this patch declares its compatibility as `>=0.16.3,<0.17`, you can connect to a database at version `0.16.3` without causing anything to be migrated.

This PR also converts all of the classes used in `DatasetDocument` to be non-strict, meaning that they will no longer raise a `ValidationError` if unknown fields are encountered when loading dataset docs from the database. This is a future-proofing change that will allow for wider compatibility between versions. For example, with this change, this branch could be compatible with a future client version that adds new fields to its dataset's `DatasetDocument` instances.

### DIFF
--- a/docs/source/cli/index.rst
+++ b/docs/source/cli/index.rst
@@ -749,7 +749,7 @@ FiftyOne deployments.
 
 .. code-block:: shell
 
-    # Migrate the database and all datasets to the current package version
+    # Migrate the database and all datasets to the current client version
     fiftyone migrate --all
 
 .. code-block:: shell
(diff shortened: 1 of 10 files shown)


benjaminpkane (approved):
LGTM 👍

Issue #1996, opened 2022-08-11
[BUG] Add validation to dataset names


ehofesmann:
Dataset names should be validated for unsupported characters before being saved. Many non-alphanumeric characters result in issues when attempting to load the dataset in the App. 

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart").clone()
dataset.name = "test/1"

session = fo.launch_app(dataset)
```

```
404: /datasets/test%2F1 not found
```

brimoor:
More generally, we'll need to introduce notions of "human-readable" vs "URL-friendly" dataset names.

IE a dataset with name `ErIc's CrAzY DaTaSeT` will be reachable at `/datasets/erics-crazy-dataset`.

We could still add validation to the human-readable name, but I do think spaces should be allowed in the dataset name, but they can't be allowed in the URL.

swheaton:
URL safe slugs are used now.

PR #2019, opened 2022-08-24
Adding dataset URL names and descriptions

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1996.

Adds a `url_name` and `description` to datasets:

```py
import fiftyone as fo

dataset = fo.Dataset("my dataset!")
dataset.description = "Hello, world!"

print(dataset.name)  # my dataset!
print(dataset._doc.url_name) # my-dataset

dataset = fo.Dataset("my-dataset")
# ValueError: Dataset name 'my-dataset' is not available
```


fiftyone/core/dataset.py
```diff
@@ -78,6 +78,32 @@ def dataset_exists(name):
...
+    Returns:
+        the URL name
+
+    Raises:
+        ValueError: if the name is not available
+    """
+    url_name = fou.to_url_name(name)
+
+    query = {"$or": [{"name": name}, {"url_name": url_name}]}
+    if skip is not None:
```
manivoxel51, line 97: curious what a use case is  for skip 🤔

fiftyone/migrations/revisions/v0_18_1.py
```diff
@@ -0,0 +1,147 @@
...
+    for d in db.datasets.find(
+        {"_id": {"$ne": _id}}, {"name": 1, "url_name": 1}
+    ):
+        if "name" in d:
+            existing_names.add(d["name"])
+
+        if "url_name" in d:
+            existing_url_names.add(d["url_name"])
+
+    if name is None:
```
manivoxel51, line 56: could name ever be none? i thought name is required for a dataset

fiftyone/migrations/revisions/v0_18_1.py
```diff
@@ -0,0 +1,147 @@
...
+        return c
+
+    if c in _WHITESAPCE_CHARS:
+        return "-"
+
+    return ""
+
+
+def _to_url_name(name):
+    if not isinstance(name, str):
```
manivoxel51, line 107: 👍

fiftyone/migrations/revisions/v0_18_1.py
```diff
@@ -0,0 +1,147 @@
...
+        raise ValueError("Expected string; found %s: %s" % (type(name), name))
+
```
manivoxel51, line 109: should we make sure the len(name) > 0?

fiftyone/migrations/revisions/v0_18_1.py
```diff
@@ -0,0 +1,147 @@
...
+    last = ""
+    for c in name:
+        s = _sanitize_char(c)
+        if s and (s != "-" or last != "-"):
+            safe.append(s)
+            last = s
+
+    url_name = "".join(safe).strip("-")
+
+    if len(url_name) < _URL_NAME_LENGTH_RANGE[0]:
```
manivoxel51, line 126: could url_name have length zero? that should raise an error i think

fiftyone/migrations/revisions/v0_18_1.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+    return url_name
+
+
+def _get_default_dataset_name(existing_names):
+    now = datetime.now()
+    name = now.strftime("%Y.%m.%d.%H.%M.%S")
+    if name in existing_names:
+        name = now.strftime("%Y.%m.%d.%H.%M.%S.%f")
```
manivoxel51, line 145: unique indeed

manivoxel51 (approved):
Nice work @brimoor ! 🍨

fiftyone/migrations/revisions/v0_18_1.py
brimoor: It shouldn't be possible, but I like to be extra extra careful in migrations

fiftyone/migrations/revisions/v0_18_1.py
brimoor: The `url_name` is strictly shorter than `name`, so it implicitly verifies the name's length 👍

fiftyone/migrations/revisions/v0_18_1.py
brimoor: Yeah that's what this is verifying: it ensures that `len(name) >= len(url_name) >= 1`

fiftyone/migrations/revisions/v0_18_1.py
brimoor: FYI- I hate copying code, but we do have to do a one time copy of code whenever certain logic is required in migrations, since they have to run independent of library version.

In this case, I've copy-pasted the logic from `fiftyone.core.dataset.get_default_dataset_name()` and `fiftyone.core.utils.to_url_name()`.

fiftyone/core/dataset.py
brimoor: So that `my dataset` can be renamed to `my-dataset` 🤗

manivoxel51 (approved):
nice work Brian! 🍨

fiftyone/core/dataset.py
```diff
@@ -6137,6 +6176,7 @@ def _create_dataset(
     dataset_doc = foo.DatasetDocument(
         id=_id,
         name=name,
+        url_name=url_name,
```
kaixi-wang, line 6179: since this is not in prod yet, for consistency with teams, should we use `slug`?

cc @j053y

brimoor: I will try raising this again on `saved-views` branch

PR #2021, opened 2022-08-24
common json panel for 2d/3d lookers

ritch:
Refactor the JSON panel for use in both 2d/3d lookers.


benjaminpkane (approved):
Smooth. LGTM!

Small comments:
* opacity differs from Looker2d's shortcuts panel
* Looker3d HUD actions don't have titles/tooltips
* I noticed that the active sample can change with JSON open, but that JSON data does not change with it. Maybe it can change with it?

I will open a PR cleaning up the old implemntation.

PR #2022, opened 2022-08-24
add mediafield support to looker3d

ritch:
- adds proper media field support for looker3d
 - removes multifield support from looker3d settings


benjaminpkane (approved):
LGTM 👍

PR #2023, opened 2022-08-24
Fix documentation builds

benjaminpkane:
Documentation builds are failing, like to changes with `pip` and editable installs.


brimoor (approved):
LGTM

PR #2024, opened 2022-08-24
refactor tooltips for 3d

ritch:
This moves most of the logic into a single hook used in the 2d/3d lookers.


app/packages/state/src/hooks/useTooltip.ts
```diff
@@ -0,0 +1,91 @@
+import { useRecoilState, atom } from "recoil";
+import { useEffect } from "react";
+const tooltipState = atom({
```
benjaminpkane, line 3: Maybe add explicit typing here?

benjaminpkane (approved):
Implementation LGTM! 👏

benjaminpkane:
One comment I forgot to mention is that looker3d does not have a pointer cursor on label hover

PR #2025, opened 2022-08-25
Persistent group slice in expanded view

benjaminpkane:
Updates the modal to maintain the active slice between next/previous navigation actions. For example, with the groups quickstart, upon clicking on the right slice and navigating to the next group, the right slice will still be shown.
Screen Shot 2022-08-25 at 3 24 43 PM


brimoor (changes requested):
Persistent slice is working great!

The only issue I'm seeing on this branch is something I've seen before (but was resolved) that, when using prev/next controls in the modal, sometimes I can see the grid updating in the background.

The most reliable way to see this is to launch `quickstart-groups`, open the modal, select the point cloud slice, hover mouse over Looker2D, then press right arrow key.

Doing this always results in the grid updating in the background for me. Also the bug occurs where the grid zoom resets, but the actual slider doesn't reset, so the App is in an inconsistent state.

benjaminpkane:
Ok, addressed

brimoor (approved):
LGTM. Perfect! 🏅

PR #2026, opened 2022-08-25
Fix group session updates

benjaminpkane:
Fixes session updates for empty view and selected sample/label updates from the Python client.


brimoor (approved):
LGTM

Issue #1902, opened 2022-06-21
[BUG]


sadjadasghari:
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 16.04
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: v0.15.1
-   **Python version**: 3.8.10

### Commands to reproduce

```
foul.import_from_labelbox(dataset_test, json_path, label_prefix=None, download_dir='/home/sadjad/Workspace/yolov5/datasets/temp/', labelbox_id_field='labelbox_id')
```

### Describe the problem

When trying to import from a labelbox project with video entities, the method fails to get the video labels. 

### Code to reproduce issue

```
import fiftyone.core.media as fomm
dataset_test = fo.Dataset()
dataset_test.media_type = fomm.VIDEO
json_path = '/home/sadjad/Downloads/temp/export-2022-05-19T15 18 44.922Z.json'
foul.import_from_labelbox(dataset_test, json_path, label_prefix=None, download_dir='/home/sadjad/Workspace/yolov5/datasets/temp/', labelbox_id_field='labelbox_id')
```
### Other info / logs

```
---------------------------------------------------------------------------
WebSessionError                           Traceback (most recent call last)
/tmp/ipykernel_15238/2211279030.py in <module>
      8 client.api_key = API_KEY
      9 json_path = '/home/sadjad/Downloads/temp/export-2022-05-19T15 18 44.922Z.json'
---> 10 foul.import_from_labelbox(dataset_import, json_path, label_prefix=None, download_dir='/home/sadjad/Workspace/yolov5/datasets/temp/', labelbox_id_field='labelbox_id')

~/.virtualenvs/yolov5/lib/python3.8/site-packages/fiftyone/utils/labelbox.py in import_from_labelbox(dataset, json_path, label_prefix, download_dir, labelbox_id_field)
   1491                     sample.metadata.frame_height,
   1492                 )
-> 1493                 frames = _parse_video_labels(d["Label"], frame_size)
   1494                 sample.frames.merge(
   1495                     {

~/.virtualenvs/yolov5/lib/python3.8/site-packages/fiftyone/utils/labelbox.py in _parse_video_labels(video_label_d, frame_size)
   2102 def _parse_video_labels(video_label_d, frame_size):
   2103     url_or_filepath = video_label_d["frames"]
[...]

PR #2027, opened 2022-08-25
[HOLD] updated _parse_video_labels function to support url_or_filepath with …

sadjadasghari:
…API_KEY

changes to _parse_video_labels function in utils/labelbox.py to support video label urls that need API_KEY to be downloaded. 

tested locally with installing via bash install.bash -d

        notes for FiftyOne users.

Supports importing video datasets from Labelbox and fixes a bug where the importer fails to fetch video labels files from Labelbox. 

-   [ ] Other

### DIFF
--- a/fiftyone/utils/labelbox.py
+++ b/fiftyone/utils/labelbox.py
@@ -2121,7 +2121,9 @@ def _make_mask(instance_uri, color):
 # https://labelbox.com/docs/exporting-data/export-format-detail#video
 def _parse_video_labels(video_label_d, frame_size):
     url_or_filepath = video_label_d["frames"]
-    label_d_list = _download_or_load_ndjson(url_or_filepath)
+    headers = {"Authorization": "Bearer %s" % API_KEY}
+    response = requests.get(url_or_filepath, headers=headers)
+    label_d_list = ndjson.loads(response.text)  # _download_or_load_ndjson(url_or_filepath)
 
     frames = {}
     for label_d in label_d_list:


sadjadasghari:
this fix is related to the bug https://github.com/voxel51/fiftyone/issues/1902

fiftyone/utils/labelbox.py
```diff
@@ -2121,7 +2121,9 @@ def _make_mask(instance_uri, color):
 # https://labelbox.com/docs/exporting-data/export-format-detail#video
 def _parse_video_labels(video_label_d, frame_size):
     url_or_filepath = video_label_d["frames"]
-    label_d_list = _download_or_load_ndjson(url_or_filepath)
+    headers = {"Authorization": "Bearer %s" % API_KEY}
```
brimoor, line 2124: `API_KEY` isn't defined here.

I guess one approach would be to allow it to be manually provided as a kwarg. However, our [Labelbox integration](https://voxel51.com/docs/fiftyone/integrations/labelbox.html#authentication) provides a way to access one's API key as follows:

```shell
export FIFTYONE_LABELBOX_API_KEY=XXXXXXXX
```

```py
import fiftyone as fo

print(fo.annotation_config.backends["labelbox"]["api_key"])
# XXXXXXXX
```

sadjadasghari: Yes, I wasn't sure about if it is ok to provide api_key as a kwarg, but to me the second solution (exporting api key) makes more sense (then we will need to add that as an instruction for calling the 'import_from_labelbox' method).

PR #2029, opened 2022-08-26
Fixing match_frames() bug

brimoor:
Fixes a bug in `match_frames()` that was causing `sample.frames` to lookup all frames, even if the view filters the frames:

```py
import fiftyone as fo
import fiftyone.zoo as foz

num_objects = F("detections.detections").length()
view = dataset.match_frames(num_objects == 28)
sample = view.first()

assert len(view) == 1

# Used to fail because all frames were being attached; now succeeds
assert len(sample.frames) == view.count("frames")
```

### DIFF
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -6609,8 +6609,9 @@ _STAGES = [
 # Registry of stages that promise to only reorder/select documents
 _STAGES_THAT_SELECT_OR_REORDER = {
     # View stages that only reorder documents
-    SortBy,
     GroupBy,
+    SortBy,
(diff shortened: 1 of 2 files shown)


benjaminpkane:
@brimoor This PR has all of its changes pushed? I don't see any implementation changes

brimoor:
> @brimoor This PR has all of its changes pushed? I don't see any implementation changes

Yeah the fix was just to remove `MatchFrames` from `_STAGES_THAT_SELECT_OR_REORDER`.

The unit test check I added used to fail and now passes.

benjaminpkane (approved):
LGTM

PR #2031, opened 2022-08-26
Fix README slack links

benjaminpkane:
Fixes the second of two README slack links. Currently checking other links, as well.


brimoor (approved):
LGTM after all links are checked

PR #2032, opened 2022-08-26
Looker3d hotkeys

ritch:
**New HelpPanel**

Adds a new generic react HelpPanel which displays shortcuts/help items. These are provided by the component using the `useHelpPanel` hook, which allows the content / rendering/component to exist in totally different context.

However this does not replace the support for the existing Looker Help Panel. That should be possible in the future. This is only used in the 3d looker at the moment.

**Hot Key Support for Looker3D**

This tries to mimic the exact logic/style of Escape and other shortcuts in the 2d looker.

**Known Issue**

If you hover over the 2d looker while a panel (eg. JSONPanel) is open, pressing escape will both close the JSONPanel and the entire Modal. This is due to the common "reading state in an event bug" where the state is actually the initially rendered state. However fixing this is non trivial and should be taken on as a follow up.


brimoor:
Just tested this branch and documented a few issues I found in our punch list doc

benjaminpkane (approved):
LGTM!

PR #2035, opened 2022-08-29
Adding grouped dataset importer/exporter interfaces

brimoor:
Adds `GroupDatasetImporter` and `GroupDatasetExporter` interfaces for importing/exporting grouped collections in/to custom formats on disk.


docs/source/user_guide/dataset_creation/datasets.rst
```diff
@@ -5865,3 +6047,44 @@ corresponding to the type of dataset that you are working with.
...
+                class for exporting datasets of this type to disk.
+
+                Returns:
+                    a :class:`fiftyone.utils.data.exporters.GroupDatasetExporter`
+                    class
+                """
+                # Return your custom GroupDatasetExporter class here
+                pass
+
+    Note that, as this type represents an unlabeled image dataset, its importer
```
benjaminpkane, line 6088: Is this correct? Seems like an odd statement

benjaminpkane (approved):
LGTM

docs/source/user_guide/dataset_creation/datasets.rst
brimoor: Good catch! Copy paste error 😿

PR #2038, opened 2022-08-30
Group statistics and tagging mode

benjaminpkane:
Adds a group context to grid and modal options that causes the entire group to be used for computing aggregations and tagging.


brimoor (changes requested):
The `group|slice` toggle works great!

Here's a few things I'm seeing:

1. Since `slice` is the default, can you switch order in menu to be `slice | group`?
2. Can you update the text above the grid as follows?

Slice mode
- Full dataset: show `200 groups`
- Filter by selecting `Person_sitting`: show `5 of 200 groups`

Group mode
- Full dataset: show `200 groups (600 samples)`
- Filter by selecting `Person_sitting`: show `5 of 200 groups (15 of 600 samples)`

```py
import fiftyone as fo
import fiftyone.zoo as foz

```

3. The tag menu in the modal is currently pulling/pushing changes to the entire view, but should only be applying to the current group
4. When adding a tag in `group` mode, if I select a sample in the grid and type a new tag, the menu suggests that I'll be adding `1` instance of the tag, but actually I'll be tagging all samples in the group, so the outcome will be `3` tags. It would be preferable if the menu showed `3` instead of `1`

benjaminpkane:
> Full dataset: show 200 groups (600 samples)
Filter by selecting Person_sitting: show 5 of 200 groups (15 of 600 samples)

This will add a new aggregation, this is ok? Can group be missing a slice?

brimoor:
Yep, would require both `view.select_group_slices(_allow_mixed=True).count()` and `view.count()`, but that's okay.

benjaminpkane:
Ok I believe comments have been addressed

brimoor:
I'm still seeing some issues:

Grid
- Create a `limit(10)` stage in the view bar. In "group" mode, we should see `10 groups (30 samples)`, but we see `10 groups (10 samples)`
- Select two samples in the grid. in "group" mode, we should see `+ tag 6 selected samples` (or even better, `+ tag 6 samples in selected groups`), but we see `+ tag 2 selected samples`. Once I start typing a new tag, however, I do correctly see `Add "XXX" tag to 6 samples`

Sample modal
- Switching to "group" mode doesn't update the stats in the sidebar
- In "group" mode, the sample stats in the tag modal are showing the full view, not the current group

Tweaks
- In "slice" mode, please change the text to `XXX groups` rather than `XXX samples`. I think showing the "groups" text is important here because it's the only place

Unrelated to this PR (?):
- Looker3D dies after pressing the bookmark icon in the grid (no longer displays when opening the modal)
- Filtering by label ID doesn't work when the point cloud sample is active (no labels ever match the filter)

benjaminpkane:
Ok, everything is addressed but:
>Filtering by label ID doesn't work when the point cloud sample is active (no labels ever match the filter)

I can look into that in another PR if not already addressed

brimoor:
@ritch resolved the 3D label selection in https://github.com/voxel51/fiftyone/pull/2056 🎉

brimoor:
@benjaminpkane I'm seeing a build error:

```
sidebarSampleId' is not exported by ../state/src/index.ts, imported by src/components/Actions/utils.tsx
```

benjaminpkane:
> @benjaminpkane I'm seeing a build error:
> 
> ```
> sidebarSampleId' is not exported by ../state/src/index.ts, imported by src/components/Actions/utils.tsx
> ```

My mistake. Fixed

benjaminpkane:
Offline comments have been addressed here, as well

brimoor:
Almost home! One last pass:

General tagging bugs
- In both slice/group stats modes, with no samples selected, the tagging menu should say `+ tag XXX samples`, not `+ tag XXX selected samples`, since the tagging operation will be applied to the entire view
- In group stats mode, with `X` samples selected, the tagging menu says `+ tag X selected samples/labels`, but should say `+ tag Y selected samples/labels`, where `Y` is the number of samples/labels across the groups, not just the current slice. This is true for both the grid view and the modal

Modal
- Applying a tag edit from the modal causes the grid view to reload, but nothing one does from the modal should cause the grid to reload (too expensive)
- nit: scrolling on the sidebar causes the grid view behind to scroll, but shouldn't

brimoor (approved):
LGTM after final round of comments are addressed

brimoor:
@benjaminpkane confirmed that everything is working now **EXCEPT** that the sidebar now doesn't fully refresh after tags are applied (one must manually refresh page)

And one nit: while you're in there, can you please reset the selection state after any action is submitted in the tag menu?

PR #2039, opened 2022-08-30
Groups zoo fixes

brimoor:
- Fixes a bug that prevented `load_zoo_dataset("kitti-multiview", split="train")` from working properly
- Fixes the `itemRotation` setting for both `kitti-multiview` and `quickstart-groups` (the latter was achieved by uploading a new zip to Google Drive


benjaminpkane (approved):
LGTM

PR #2041, opened 2022-08-30
Remove old session module

benjaminpkane:
Removes the session module that was replaced by the session subpackage in #1713


brimoor (approved):
LGTM

PR #2044, opened 2022-08-31
Session group slice

benjaminpkane:
Adds bi-directional state updates between the App and Python for changes to a group dataset's current `group_slice`


fiftyone/core/session/session.py
```diff
@@ -986,6 +986,9 @@ def _attach_listeners(session: "Session"):
         session, "_state", event.state
     )
     session._client.add_event_listener("state_update", on_state_update)
+    session._client.add_event_listener(
+        "state_update", lambda e: print(e.state.dataset)
```
brimoor, line 990: Remove?

fiftyone/server/query.py
```diff
@@ -217,6 +218,9 @@ async def resolver(
             dataset.id = ObjectId()
             dataset.media_type = view.media_type
 
+        if dataset.media_type == fom.GROUP:
+            dataset.group_slice = view.group_slice
```
brimoor, line 222: Hmm maybe I should support for views having a custom `group_slice` that doesn't necessarily match their parent dataset's

brimoor: @benjaminpkane FYI just added support for setting `view.group_slice` independently of `dataset.group_slice` to this branch.

I guess the correct "synced" behavior is that changing the group slice in the App would immediately update both `session.dataset.group_slice` and `session.view.group_slice`.

Can you update the session logic as necessary so that:
- `session.dataset.group_slice` and `session.view.group_slice` matches the App's group slice setting? Since they should always match, it shouldn't be necessary to directly set `view.group_slice` here, so I guess maybe no code change is necessary here.
- Setting `session.view = view` will pull the view's group slice for the App setting. This will involve a slight state tweak

benjaminpkane: Ok,, I believe these are addressed

brimoor:
@benjaminpkane couple things I'm seeing on this branch:

- After calling `session.refresh()`, the view bar stops working in the App (clicking add stage doesn't work)
- After adding a view stage to the view bar in the App, eg `Limit(10)`, I get the following error when clicking into the slice selector in the App:

```
TypeError: Cannot read properties of undefined (reading 'filter')
at get (http://localhost:5151/assets/index.ca78d4cd.js:303:7171)
at J (http://localhost:5151/assets/index.ca78d4cd.js:38:57755)
at K (http://localhost:5151/assets/index.ca78d4cd.js:38:58795)
at http://localhost:5151/assets/index.ca78d4cd.js:38:60327
at ft (http://localhost:5151/assets/index.ca78d4cd.js:38:60090)
at Object.We [as get] (http://localhost:5151/assets/index.ca78d4cd.js:38:60320)
at bSe (http://localhost:5151/assets/index.ca78d4cd.js:38:15271)
at OSe (http://localhost:5151/assets/index.ca78d4cd.js:38:17436)
at http://localhost:5151/assets/index.ca78d4cd.js:38:38067
at http://localhost:5151/assets/index.ca78d4cd.js:38:38138
```

- Not sure if this is a side effect of the first point, but calling `session.refresh()` doesn't pull the latest `group_slice` from Python:

```py
import fiftyone as fo
import fiftyone.zoo as foz

print(dataset.group_slice)  # 'left'

# group slice doesn't update in the App
dataset.group_slice = "right"
session.refresh()
```

benjaminpkane:
> After calling session.refresh(), the view bar stops working in the App (clicking add stage doesn't work)

Ok comments addressed. I couldn't reproduce the above. Let me know if it's still an issue.

brimoor (approved):

PR #2048, opened 2022-08-31
Adding support for linking directly to imported subpackage contents

brimoor:
This PR adds the following block of code to all `__init__.py` files that import symbols that are intended to be exposed in FiftyOne's public namespace using `path.to.package.symbol` rather than `path.to.package.module.symbol`:

```py
# This enables Sphinx refs to directly use paths imported here
__all__ = [
    k
    for k, v in globals().items()
    if not k.startswith("_") and not isinstance(v, types.ModuleType)
]
```

The reason I'm adding this code is to allow Sphinx to understand references to imported symbols. For example, we allow users to use `fiftyone.types.Dataset` rather than the full `fiftyone.types.dataset_types.Dataset`. But, prior to this PR, docstrings had to use the full `:class:fiftyone.types.dataset_types.Dataset`. Now docstrings may choose to use the abbreviated `:class:fiftyone.types.Dataset` (and I updated all such references in the docs to use the abbreviated form as a proof of concept).

I believe @benjaminpkane was thinking along similar lines when he manually added `__all__` to `fiftyone/core/session.__init__.py` when he split `fiftyone.core.session` into submodules, because that was required to avoid breaking (or having to manually change) all `:class:fiftyone.core.session.Session` references in the code/docs. This PR effectively future-proofs all other such cases.

### Implementation detail

The funny `globals()` syntax is because it would be tedious (although, tbh, more proper) to update certain imports like this:

```py
from .base import *
from .converters import *
from .exporters import *
from .importers import *
from .ingestors import *
from .parsers import *
```

into explicit imports and corresponding `__all__` statements.

### DIFF
--- a/docs/source/_includes/substitutions.rst
+++ b/docs/source/_includes/substitutions.rst
@@ -81,11 +81,11 @@
 .. |UnlabeledVideoSampleParser| replace:: :class:`UnlabeledVideoSampleParser <fiftyone.utils.data.parsers.UnlabeledVideoSampleParser>`
 .. |LabeledVideoSampleParser| replace:: :class:`LabeledVideoSampleParser <fiftyone.utils.data.parsers.LabeledVideoSampleParser>`
 
-.. |DatasetType| replace:: :class:`Dataset <fiftyone.types.dataset_types.Dataset>`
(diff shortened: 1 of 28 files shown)


benjaminpkane (approved):
LGTM! I didn't know this helps with Sphinx, nice

PR #2052, opened 2022-09-01
[WIP] Embed support

ritch:


app/packages/state/src/hooks/useStateUpdate.ts
```diff
@@ -86,6 +86,7 @@ const useStateUpdate = () => {
       }
 
       if (dataset) {
+        console.log("dataset?", dataset);
```
benjaminpkane, line 89: leftover logging?

app/packages/dataset/src/Dataset.tsx
```diff
@@ -0,0 +1,136 @@
...
+              console.log({ colorscale, config, state });
+
+              setState((s) => ({ colorscale, config, state }));
+
+              break;
+            }
+          }
+        },
+        onclose: () => {
+          // clearModal();
```
benjaminpkane, line 123: This should happen I think

benjaminpkane (approved):
LGTM! I think the `@fiftyone/app` package.json could be cleaned up if you see anything obvious, i.e. removing dependencies.

Also saw some generated graphql code still in `./Root`

PR #2053, opened 2022-09-01
Improving groups migration

brimoor:
The original impetus for this PR was to update the v0.17.0 migration so that a valid `DatasetAppConfig` is inserted by the migration itself. Previously `app_config == {}` was being inserted, which led to default values being populated by `load_dataset()` but *not* getting persisted to the database because `doc.save()` wasn't detecting any changes. This is bad because anything (eg GraphQL) that tries to pull directly from the DB wouldn't have the correct data available to it.

Along the way, I removed some unnecessary code from `_load_dataset()` that was redeclaring `sample_fields` and `frame_fields` and then immediately re-saving the dataset doc, even though this data was already properly defined. The implementation of `_load_dataset()` now has the property that the only field that may be changed is `last_loaded_at`, which makes sense. I verified that this change is valid like so:

```shell
fiftyone migrate -n quickstart -v 0.8
```

```py
import fiftyone as fo
```

### DIFF
--- a/fiftyone/core/cli.py
+++ b/fiftyone/core/cli.py
@@ -2592,11 +2592,6 @@ class MigrateCommand(Command):
             )
             return
 
-        fom.migrate_database_if_necessary(
-            destination=args.version,
-            verbose=args.verbose,
-        )
-
         if args.dataset_name:
             for name in args.dataset_name:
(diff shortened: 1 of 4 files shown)


benjaminpkane (approved):
LGTM

PR #2054, opened 2022-09-01
Add Detectron2 tutorial

ehofesmann:
Adds a tutorial showing how to load FiftyOne datasets into detectron2 model training pipelines, trains a segmentation model, then loads the predictions back into FiftyOne and performs evaluation.


brimoor (approved):
LGTM. Love that we have an example that includes training a model 💯 🍨

PR #2055, opened 2022-09-02
Optimizing group selection in server get_view()

brimoor:
Optimizes group selection in `fiftyone.server.view.get_view()`, similar to what is done for sample selection.


benjaminpkane (approved):
LGTM

PR #2058, opened 2022-09-02
Removing cv2 colab fix

brimoor:
Removes the warning below, since I'm no longer seeing this issue in Colab 🥳 🎈 

> If the above import fails due to a `cv2` error, it is an issue with OpenCV in Colab environments. [Follow these instructions to resolve it.](https://github.com/voxel51/fiftyone/issues/1494#issuecomment-1003148448)

~~Adds a suggestion to install `fiftyone==0.15.1` when working in Google Colab, since the App is currently buggy in Colab with `fiftyone>=0.16,<0.17`~~ <-- no longer needed, thank you https://github.com/voxel51/fiftyone/pull/2069!

### DIFF
--- a/docs/source/recipes/convert_datasets.ipynb
+++ b/docs/source/recipes/convert_datasets.ipynb
@@ -20,9 +20,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "If you haven't already, install FiftyOne:\n",
-    "\n",
-    "*If you're working in Google Colab, update the cell below to install `fiftyone==0.15.1`*"
+    "If you haven't already, install FiftyOne:"
    ]
   },
   {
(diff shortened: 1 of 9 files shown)


ehofesmann (commented):
Do we really need to add the v0.15.1 suggestion if we can get this fixed in v0.17, which would be the next release anyway?

brimoor:
> Do we really need to add the v0.15.1 suggestion if we can get this fixed in v0.17, which would be the next release anyway?

Yeah good point, how about we leave this PR in an open state so that we can remove it if the colab fixes make it into v0.17.

Note that it would be valuable to merge https://github.com/voxel51/fiftyone-examples/pull/41 ASAP because fiftyone.ai pulls directly from the `master` branch over there, so users will immediately benefit pre v0.17 release.

ehofesmann (approved):
LGTM! Ready to merge after #2069

Issue #2000, opened 2022-08-12
[BUG] Unique filenames are not generated when drawing labels


ehofesmann:
When drawing labels on samples and saving them to disk, it is expected that unique filenames will be created if multiple samples in the view have the same filename/path. However, currently these unique filenames aren't generated and only one copy of each file will be written.

```python
from fiftyone import ViewField as F
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart").clone()
sample = dataset.first()
dataset.add_sample(sample)

dup_view = dataset.match(F("filepath") == sample.filepath)
dup_view.draw_labels("/tmp/draw-labels", label_fields=["ground_truth"])
```

Resulting directory:
```
/tmp/draw-labels
    00880.jpg
```

Expected result:
```
/tmp/draw-labels
    00880.jpg
    00880-2.jpg
```

PR #2060, opened 2022-09-02
Adding rel_dir to export() and draw_labels()

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/2000

Other goodies:
- Added a `rel_dir` parameter on `draw_labels()` and `export()`---as well as all builtin exporters---which allows for the possibility of preserving nested subdirectory structure of the input dataset's media
- Adds an optional `abs_paths` parameter to all relevant exporters that allows for storing absolute paths to media in the exported labels


ehofesmann (approved):
LGTM!

PR #2061, opened 2022-09-02
Adding plugins docs/env var

brimoor:
Change log
- Adds `plugins_dir` to the FiftyOne config (env var `FIFTYONE_PLUGINS_DIR`)
- Adds a pointer to the plugin README to the App User Guide page
- Editorial pass over plugin README 🤓


ritch (approved):
LGTM!

Issue #2067, opened 2022-09-07
Quotation marks needed for fiftyone[desktop]


GeorgePearse:
### URL(s) with the issue:

https://voxel51.com/docs/fiftyone/getting_started/install.html#installing-fiftyone-desktop

### Description of proposal (what needs changing):
```
pip install fiftyone[desktop]
```
does not work, it needs quotation marks

e.g.
```
pip install "fiftyone[desktop]"
```

### Willingness to contribute

-   [X] Yes. I can contribute a documentation fix independently.
-   [ ] Yes. I would be willing to contribute a document fix with guidance from
        the FiftyOne community.
-   [ ] No. I cannot contribute a documentation fix at this time.

brimoor:
Ah good catch, feel free to make a PR to patch it 🤗

brimoor:
Closed by https://github.com/voxel51/fiftyone/pull/2068.

PR #2068, opened 2022-09-07
Update install.rst

GeorgePearse:

Resolution of https://github.com/voxel51/fiftyone/issues/2067, include quotation marks for 

```
pip install "fiftyone[desktop]"
```

Tested by user

### DIFF
--- a/docs/source/getting_started/install.rst
+++ b/docs/source/getting_started/install.rst
@@ -117,7 +117,7 @@ install as follows:
 
 .. code-block:: shell
 
-  pip install fiftyone[desktop]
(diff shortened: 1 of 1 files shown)


findtopher (approved):
This is primarily a zsh issue, but testing shows that it works fine in bash so the quotes are probably a good idea.

May need to dig deeper into the docs and make sure that all the commands work with zsh (the OS X standard now) - but the unescaped `[]`s are definitely a problem for zsh.

GeorgePearse:
> 

Cheers, realised it may be operating system thing as I wrote it.

PR #2069, opened 2022-09-07
Fix Colab sessions and screenshotting

benjaminpkane:
[Here](https://colab.research.google.com/drive/1Vy2idD04D4ylLBxMopLrgf8zWSkObXCp?usp=sharing) is a notebook to test with


brimoor (approved):
Sessions are working again! 💯 🙇 

nit: is it possible to suppress the ID that's printed when the App is launched? 
Screen Shot 2022-09-07 at 10 28 03 PM

benjaminpkane:
> nit: is it possible to suppress the ID that's printed when the App is launched?

This is a result of `session.show()` returning an id. If we don't need that behavior, we can remove it.

PR #2070, opened 2022-09-08
Fix CVAT bug when checking if a task exists

ehofesmann:
On CVAT > v2.0 servers, when checking if a task exists, the existing method would fail if there are too many tasks. 
Previously, this was accomplished by searching for a task id by `cvat_url/api/task?id=XXX` which would return the query task for CVATv1 servers, but may not for CVATv2 servers. This error would result in annotation runs unable to be loaded

After this PR, the `task_exists()` method now attempts to get the status of a task and returns `False` if it encounters a 404 error. 

A test was added to capture this issue on CVATv2 servers.

Note: This also fixes a bug where annotations were being duplicated to other images in a task due to the fact that images also contain an `outside` property now in CVATv2 servers.


brimoor (approved):
LGTM

PR #2071, opened 2022-09-08
Update evaluate detections notebook to use `only_matches=False` when filtering labels

### LINKED ISSUE #2050: [BUG] evaluate_detections calculates wrong precision and recall after filter_labels
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 20.04.01 LTS
-   **FiftyOne version (run `fiftyone --version`)**: FiftyOne v0.16.6, Voxel51, Inc.
-   **Python version**: 3.8.5

### Commands to reproduce

Jupyter notebook included in "Code to reproduce issue" section.

### Describe the problem

The tutorial at https://voxel51.com/docs/fiftyone/tutorials/evaluate_detections.html  shows that we can use filter_labels() to pick out only high confidence predictions, and then use evaluate_detections() to calculate precision and recall. Normally, decreasing the confidence threshold should always decrease the precision and increase the recall. However, sometimes the opposite occurs, even when closely following the tutorial code.

The attached notebook demonstrates this using entirely code and data from the tutorial. With confidence threshold 0.997, the "bowl" class has precision 0.80. Reducing the confidence threshold to 0.9968, however, *increases* the precision to 0.83.

It seems that when evaluate_detections() is called on a view of the dataset, it does not populate the eval_tp, eval_fp, and eval_fn fields for points not in that view. Note that on line 4, these variables are all None. I believe this is causing the precision and recall to be calculated incorrectly: the system isn't being penalized for making no detections on this image. eval_fn should instead be equal to the number of ground truth detections (26).

I believe that when evaluate_detections() is calculated on a filtered view of the dataset, everything outside of that view should have eval_tp and eval_fp set to 0, and eval_fn set to len(ground_truth.detections).

### Code to reproduce issue

[bugtrack3.zip](https://github.com/voxel51/fiftyone/files/9471850/bugtrack3.zip)

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community.

ehofesmann:
Resolves #2050 

When filtering labels, any samples that no longer contain labels would normally be removed from the view. However, this is not desired when performing evaluations since it can skew your results between views. We set only_matches=False so that all samples will be retained, even if some no longer contain labels.


brimoor (approved):
LGTM

PR #2079, opened 2022-09-09
App groups fixes

benjaminpkane:
Resolves App bugs on the `groups` release branch


brimoor (approved):
Things are getting stable! Just seeing these two issues:
- When in group stats mode, tagging samples only applies to the current slice, but should include all slices
- When tagging labels in the modal, the pinned sample is not correctly updated when the point cloud sample is active (the tags are always applied to an image sample; probably the active slice from the initial modal state)

PR #2081, opened 2022-09-09
fix pinned sample not clearing

ritch:
In a group modal, selecting the point cloud sample would not be cleared. This would cause a graphql query to execute in an unexpected state (non-groups). This patch clears out the selected sample when the modal is closed.


brimoor (approved):
LGTM

brimoor (approved):
LGTM

PR #2083, opened 2022-09-12
fix L3D hotkey targeting

ritch:
Fixes case where "keyboard shortcuts shouldn’t have “focus” when using the tagging menu (eg as I type the tag “test”, the E and T shortcuts take effect behind)"


brimoor (approved):
LGTM. Thanks 🚀

PR #2086, opened 2022-09-14
Don't regenerate IDs when copying Group instances

brimoor:
Alternative to https://github.com/voxel51/fiftyone/pull/2085 that makes it so that *only* `Group` instances don't have their IDs changed when they are copied.

Why is this change needed? Because it is normal to have multiple `Group` instances with a shared `id`, and regenerating this ID in a `sample.copy()` operation would break the group association.


brimoor:
Talked with @benjaminpkane offline and am going to go ahead and merge this one

PR #2087, opened 2022-09-14
Fix CVAT task id bug and add task download progress bar

ehofesmann:
This PR resolves an issue with CVAT >v2.0 servers where `import_annotations()` can fail to list tasks for existing projects due to a change in the CVAT API. Previously, the tasks for a project were listed as a dictionary of tasks that we needed to access the `id` of, now they are stored as a list of integer ids directly.

This PR fixes that bug as well as adds a progress bar when downloading annotations from CVAT for more than one task.

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -3903,7 +3903,13 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
             the list of task IDs
         """
         resp = self.get(self.project_url(project_id)).json()
-        return [task["id"] for task in resp.get("tasks", [])]
+        tasks = []
+        for task in resp.get("tasks", []):
+            if isinstance(task, int):
+                tasks.append(task)
+            else:
+                tasks.append(task["id"])
+        return tasks
 
     def create_task(
         self,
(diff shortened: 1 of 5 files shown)


fiftyone/utils/cvat.py
```diff
@@ -4372,152 +4378,163 @@ def download_annotations(self, results):
...
-            frames = data_resp["frames"]
-
-            label_fields = labels_task_map_rev[task_id]
-            label_types = self._get_return_label_types(
-                label_schema, label_fields
-            )
+        with fou.ProgressBar(
+            total=len(task_ids),
+            iters_str="tasks",
+            quiet=len(task_ids) == 1,
```
brimoor, line 4384: nit: if `quiet` is not explicitly included as a kwarg to `ProgressBar`, then the default will be set based on `fo.config.show_progress_bars`. So, I would recommend passing `quiet=True` only when there's one task and, otherwise, not including `quiet` so that the `fo.config.show_progress_bars` will determine whether progress is shown

fiftyone/utils/cvat.py
```diff
@@ -3903,7 +3903,13 @@ def get_project_tasks(self, project_id):
             the list of task IDs
         """
         resp = self.get(self.project_url(project_id)).json()
-        return [task["id"] for task in resp.get("tasks", [])]
+        tasks = []
+        for task in resp.get("tasks", []):
+            if isinstance(task, int):
```
brimoor, line 3908: An alternative here would be to condition the logic explicitly on the CVAT server version. At min, I'd suggest documenting why `tasks` could have different contents here, for our future sanity.

brimoor (approved):
LGTM

PR #2088, opened 2022-09-14
Add the option to load a view or dataset for run results

ehofesmann:
Currently, a bug can arise when load annotation results through the following:
* Create a view that adds a stage referencing a field being annotated
* Create an annotation run with this view
* Delete the field
* Load the annotation run
* ERROR: The annotation results cannot be loaded since the run view needs to be constructed [here](https://github.com/voxel51/fiftyone/blob/e94ccf26703bde87be16cd5143bd54014001ef2b/fiftyone/core/runs.py#L472) in order to load the run results, but one of the fields doesn't exist.

The workaround is for the user to manually add the sample/frame field back on the dataset and rerun `load_annotations()`, however, this seemingly should not be necessary since we allow brand new fields to be annotated in the same manner.

Since the annotation run view is never actually used when loading annotations for an annotation run, loading the dataset directly should suffice. This PR adds the option to `load_run_results()` to load the dataset directly rather than loading a view, which `load_annotation_results()` then uses.

My only concern with this change is that a future annotation backend may require the results to contain the annotation view rather than dataset for some unforseen reason.

### DIFF
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -185,7 +185,7 @@ def annotate(
             subclass of the backend being used
 
     Returns:
-        an :class:`AnnnotationResults`
+        an :class:`AnnotationResults`
     """
     # @todo support this?
     if samples._dataset._is_frames:
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -1029,7 +1029,8 @@ def load_annotations(
         ``None``, unless ``unexpected=="return"`` and unexpected labels are
         found, in which case a dict containing the extra labels is returned
     """
-    results = samples.load_annotation_results(anno_key, **kwargs)
+    dataset = samples._root_dataset
+    results = dataset.load_annotation_results(anno_key, **kwargs)
     annotations = results.backend.download_annotations(results)
     label_schema = results.config.label_schema
 
@@ -1064,7 +1065,7 @@ def load_annotations(
                 # Expected labels
                 if label_type == "scalar":
                     _merge_scalars(
-                        samples,
+                        dataset,
                         annos,
                         results,
                         label_field,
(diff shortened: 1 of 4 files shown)


fiftyone/utils/annotations.py
```diff
@@ -1571,22 +1572,20 @@ def _merge_labels(
...
+def _ensure_label_field(dataset, label_field, fo_label_type):
+    field, is_frame_field = dataset._handle_frame_field(label_field)
     if is_frame_field:
-        if not samples.has_frame_field(field):
-            samples._dataset.add_frame_field(
-                field,
-                fof.EmbeddedDocumentField,
-                embedded_doc_type=fo_label_type,
-            )
+        dataset.add_frame_field(
```
brimoor, line 1578: FYI this method was recently updated to only add the field _if necessary_

brimoor (approved):
LGTM!

In fact, I added a commit that updates the logic of `load_annotations()` everywhere to explicitly operate on the full dataset rather than (potentially) a view into it, because, fundamentally, annotation runs are about making permanent changes to the dataset.

As you mentioned, all the logic was already operating on the full dataset anyway, whenever that distinction could have mattered.

It's actually better this way because, previously, everything was being done to the `samples` argument passed to `load_annotations()`, which could have been anything, neither the full dataset nor the actual view on which the annotation run was executed, which definitely doesn't make sense (not a bug, but just confusing logic).

brimoor:
LMK if you have any concerns with the addition I made. Probably better test it to make sure I didn't miss anything 😄

brimoor:
> My only concern with this change is that a future annotation backend may require the results to contain the annotation view rather than dataset for some unforseen reason.

Definitely a good thing to consider, but I don't see any reason to allow a backend to make edits that depend on the actual run view.

PR #2089, opened 2022-09-15
Update Qdrant tutorial

ehofesmann:
Updates Qdrant tutorial to use `fiftyone.utils.random` instead of `fiftyone.utils.splits`.


brimoor (approved):
LGTM

Issue #2096, opened 2022-09-19
[BUG] Dataset creation is not multiprocessing-safe


elsampsa:
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 18.04.6
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: 0.16.6
-   **Python version**: 3.8

### Commands to reproduce

- Use separate mongodb server (not the default local one)
- Launch two python scripts _simultaneously_
- Both scripts clone the same dataset

### Describe the problem

When cloning a dataset, fiftyone seems to create mongodb collections with names based on timestamps.  If the same dataset is cloned _almost_ simultaneously, there will be a conflict in the collection name

Come on, when using timestamps, you must always include time information up to microseconds..! :)  Or otherwise use uuids.

I am giving an explicit name to the dataset when cloning (based in uuid):
```
dataset=dataset.clone(some_name)
dataset.persistent=True
```
but even so, fiftyone uses at some internal stage (which crashes) the name ``samples.2022.09.19.10.46.48`` for mongodb (see the logs below).

### Code to reproduce issue

n/a

### Other info / logs

Logs from two processes launched near-simultaneously (by a batch job system):

Process 1:
```
Traceback (most recent call last):
  File "/mnt/wekamount/RI-Users/sampsa.riikonen/venv3/lib/python3.8/site-packages/fiftyone/core/odm/document.py", line 590, in _save
    _id = collection.insert_one(doc).inserted_id
  File "/mnt/wekamount/RI-Users/sampsa.riikonen/venv3/lib/python3.8/site-packages/pymongo/collection.py", line 705, in insert_one
    self._insert(document,
  File "/mnt/wekamount/RI-Users/sampsa.riikonen/venv3/lib/python3.8/site-packages/pymongo/collection.py", line 620, in _insert
    return self._insert_one(
  File "/mnt/wekamount/RI-Users/sampsa.riikonen/venv3/lib/python3.8/site-packages/pymongo/collection.py", line 609, in _insert_one
    self.__database.client._retryable_write(
  File "/mnt/wekamount/RI-Users/sampsa.riikonen/venv3/lib/python3.8/site-packages/pymongo/mongo_client.py", line 1552, in _retryable_write
    return self._retry_with_session(retryable, func, s, None)
[...]

PR #2097, opened 2022-09-19
Use UUIDs in dataset collection names

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/2096.

Dataset creation was not previously multiprocess-safe due to potential name clashes in the sample collection names. Now UUIDs are used.


benjaminpkane (approved):
LGTM

PR #2103, opened 2022-09-20
[HOTFIX] Backward Compatible App

benjaminpkane:
Fixes a bug that would prevent 0.17 clients from successfully connecting to databases containing older dataset versions.

This bug only affects users who are setting `FIFTYONE_DATABASE_ADMIN=false` (this is not the default) and trying to connect to a v0.16.6 database using a v0.17 client.


brimoor (approved):
LGTM

PR #2109, opened 2022-09-21
Move plugin imports

benjaminpkane:
Moves App plugin imports to the `Root` component to guarantee plugins are active.


brimoor (approved):
LGTM

PR #2119, opened 2022-09-23
Theme refresh

benjaminpkane:
Adds the new `ThemeProvider` component and removes the old `Theme` component from `@fiftyone/components`.

Introduces basic usage of an `@mui/joy` theme.

There should be no actual changes the what is seen on screen. Working through a couple more details.


app/packages/dataset/src/Dataset.tsx
```diff
@@ -55,20 +55,24 @@ export function Dataset({ datasetName, environment }) {
...
-          datasetQueryRef={datasetQueryRef}
-          initialState={initialState}
-        >
-          <ViewBar />
-          <CoreDataset />
-        </DatasetLoader>
-      </Suspense>
-      <div id="modal" />
-    </Container>
+    <ThemeWrapper value={themePalette}>
```
ritch, line 62: When this is `<Fragment  value={false} />` won't it throw?

Maybe...

```ts
  const themePalette = theme?.palette;
  const ThemeWrapper = themePalette ? ThemeContext.Provider : Fragment;
  const themeProps = themePalette ? {value: themePalette} : {};
  return (
    <ThemeWrapper {...themeProps}>
```

imanjra: Good idea! It didn't throw when I omitted theme prop but it's better to do this.

ritch (approved):
Not sure why the loaders.ts refactor is in this PR? Those changes LGTM anyway.

As far as the theme I'm mostly curious about how the buttons and some other components compare. This seems to clean up the icons and colors, but I'm still noticing hardcoded padding and different components used.

That being said I think this is worth merging as is 👍

app/packages/components/src/components/ThemeProvider/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+    light: {
+      palette: {
+        background: {
+          body: "#F6F8FA",
+          level1: "#FFF",
+          level2: "#FFF",
+          level3: "#FFF",
+        },
+        neutral: {
+          softBg: "rgba(27, 31, 36, 0.15)",
```
ritch, line 92: For consistency, can we include these in the `colors`?

imanjra: Yah, I agree. I will move hard-coded colors to colors variable

PR #2125, opened 2022-09-29
improve plugin install docs

ritch:
Added more info around manually installing plugins.


brimoor (approved):
LGTM.

nit: I don't think you've done a [dev install](bash install.bash -d
) in your environment. There should be a pre-commit hook in place that will [auto-format your .md files](https://github.com/voxel51/fiftyone/blob/develop/.prettierrc) (eg 79 character line length)

PR #2129, opened 2022-09-30
Dataset Reloading on Click

ritch:
When embeded, the `Dataset` component will refresh on all clicks, since this window handler causes a perceived state change.


benjaminpkane (approved):
LGTM

Issue #1791, opened 2022-05-26
[FR] Add support for MongoDB transactions and sessions


benjaminpkane:
In order to take advantage of more powerful features in MongoDB like session transactions and snapshots, the MongoDB database must be configured as a replica set, even if it is still just a single node

There is likely better documentation on the issue out there, but here is a starting point:
* https://hevodata.com/learn/mongodb-transactions-on-single-node/

PR #2135, opened 2022-10-03
Refactor merge_samples() to safely cleanup in case of errors

brimoor:
Refactors the implementation of `merge_samples()` so that, if something goes wrong in the `$merge` aggregations, the temporary indexes and frame fields that are created in order to facilitate the merge will be automatically cleaned up.

Prior to this change, if a `$merge` failed on a video dataset, the datasets involved would be in broken states until the following operations were manually performed on them:

```py
# Cleanup after failed merge
dataset._frame_collection.drop_index("_merge_key_1_frame_number_1")
dataset._frame_collection.update_many({}, {"$unset": {"_merge_key": ""}})
```

The fatal error was the unexpected `_merge_key` values, but leaving extra indexes laying around is also a party foul.

**Update:** adds some additional `try-finally` blocks in various places throughout the codebase


benjaminpkane (approved):
LGTM. This is where transactions are valuable. So I'd recommend we look into https://github.com/voxel51/fiftyone/issues/1791 before we start adding more code like this.

PR #2136, opened 2022-10-03
Fix JSON/Help outside click handling

ritch:
Fixes a regression where json / help panels do not close on outside clicks.

Also moves hooks into fos pkg. I wanted these util hooks to go into the utilities package, but that caused a weird circular dep issue.


benjaminpkane (approved):
LGTM

ritch:
This causes a weird issue with clicking on the associated action button. Since its an outside click both the outside click and action button click cause unexpected state updates. I have a fix that I'm including in this PR.

PR #2141, opened 2022-10-06
Automatic Docker address + moving instructions to the docs

brimoor:
## Change log

- Automatically sets `address="0.0.0.0"` when running in Docker. Thanks @ritch for the idea!
- Moves the Docker instructions from the GitHub README to the docs
- Explicitly documents that `session = fo.launch_app(..., address="0.0.0.0")` is required when working inside Docker (in case `is_docker()` doesn't work for any reason)
- Adds a note + pointer to the Docker documentation under the App docs where `launch_app()` is introduced

## New Docker section of the environments page

Screen Shot 2022-10-06 at 3 30 44 PM

## New pointer in the launch App docs

Screen Shot 2022-10-06 at 4 35 30 PM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,6 +97,9 @@ RUN pip --no-cache-dir install --upgrade pip setuptools wheel ipython
 COPY dist dist
 RUN pip --no-cache-dir install dist/*.whl && rm -rf dist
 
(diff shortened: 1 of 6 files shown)


benjaminpkane (approved):
LGTM

PR #2145, opened 2022-10-10
fix:  create annotation run crashes if some but not all samples are labeled

oddeirikigland:

Creating an annotation run works fine, but when you add new images to the dataset and want to create a new annotation run with all the images, both the new and old, labelstudio crashes. This PR fixes this since labels are only imported from samples that have labels. 

Versions:

fiftyone version: 0.17.2
Tested with both label-studio-sdk 0.0.14 and 0.0.15 

        notes for FiftyOne users.

-   [ ] Other

## To reproduce:
```
d.annotate(
    anno_key=annotation_key,
    label_schema={
        gt_field: {
            "type": "keypoints",
            "classes": classes,
        }
    },
)
```
Where d is a fiftyone dataset and some of the samples in d have values in `gt_field` and some do not.

### DIFF
--- a/fiftyone/utils/labelstudio.py
+++ b/fiftyone/utils/labelstudio.py
@@ -257,12 +257,10 @@ class LabelStudioAnnotationAPI(foua.AnnotationAPI):
                         },
                     )
                     for smp in samples.select_fields(label_field)
-                    if smp[label_field]
                 }
                 id_map[label_field] = {
                     smp.id: _get_label_ids(smp[label_field])
                     for smp in samples.select_fields(label_field)
-                    if smp[label_field]
                 }
 
         return tasks, predictions, id_map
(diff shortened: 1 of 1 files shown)


ehofesmann (changes requested):
Thanks for catching this bug and making this PR!

This solution does work, but it may be better to still add the sample id to the `predictions` and `id_map`, just with empty labels.

This would allow `predictions` to still include the full result information about the image even if there are no labels to upload. Also, it can be useful to store information in the `id_map` about the exact samples that were uploaded even if they had no labels.

So basically instead of changing the lines here, I recommend updating the `export_label_to_label_studio()` and `_get_label_ids()` methods to handle `None` values. For eaxmple:
```python
def _get_label_ids(label):
    if label is None:
        return []
    elif isinstance(
...
```

fiftyone/utils/labelstudio.py
```diff
@@ -636,7 +636,11 @@ def export_label_to_label_studio(label, full_result=None):
         a dictionary or a list in Label Studio format
     """
     # TODO model version and model score
-    if _check_type(label, fol.Classification, fol.Classifications):
+    if label is None:
+        result_value = {}
+        ls_type = ""
+        ids = [0]
```
oddeirikigland, line 642: which values should ls_type and ids have?

oddeirikigland:
Changed as requested @ehofesmann

fiftyone/utils/labelstudio.py
ehofesmann:
```suggestion
        ls_type = None
        ids = []
```

`ids` will need to be an empty list in this case. Since `ids` and `result_value` are empty, `ls_type` will not be used. I would just set it to `ls_type = None`.

ehofesmann (approved):
Amazing! Thanks for this fix @oddeirikigland !

PR #2148, opened 2022-10-10
add note to plugin docs

ritch:
Adds a clarification to plugin docs about relative paths.


findtopher (approved):
Looks like a winner to me! 🚀

PR #2150, opened 2022-10-11
Fix view expression sort documentation

ehofesmann:
View expressions have a `.sort()` method, `.sort_by()` is to sort samples at the dataset-level.


brimoor (approved):
LGTM

PR #2159, opened 2022-10-13
only warn when plugin re-registered

ritch:
Originally a plugin registered with the same name would cause an assertion to throw. This relaxes that to a warning, since in some environments HMR causes plugins to re-register without issue.

The original intent was to prevent multiple plugins from sharing the same name. The warning will replace the assertion/error.


imanjra (approved):
lgtm!

PR #2179, opened 2022-10-15
DB config cleanup

brimoor:
Change log
- Updates the docs to clarify that FiftyOne only requires MongoDB v4.4 or later
- Removes the v0.7.1 "admin" migration. In general editing past migrations is 👎, but this migration wasn't actually required and we shouldn't be automating things that require admin database permissions


kaixi-wang (approved):
lgtm

PR #2181, opened 2022-10-17
add support for readOnly mode

ritch:
In order to support cases where the app should not display actions that write to mongodb, this PR adds the ability to load the app in a "readOnly" mode. For now this only applies to the Tagging functionality, but any new or other features which behave similarly should be hidden in this mode.

### DIFF
--- a/app/packages/dataset/src/Dataset.tsx
+++ b/app/packages/dataset/src/Dataset.tsx
@@ -8,7 +8,7 @@ import {
   usePreLoadedDataset,
   ViewBar,
 } from "@fiftyone/core";
-import { useRecoilValue } from "recoil";
+import { useRecoilValue, useSetRecoilState } from "recoil";
 import * as fos from "@fiftyone/state";
 import { darkTheme, getEventSource, toCamelCase } from "@fiftyone/utilities";
 import { useEffect, useState, Suspense } from "react";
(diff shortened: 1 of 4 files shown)


app/packages/dataset/src/index.tsx
```diff
@@ -28,28 +28,42 @@ const DatasetWrapper = () => {
 };
 
 function LoadableDataset() {
-  const [datasetName, setDatasetName] = React.useState("quickstart");
+  const [settings, setSettings] = React.useState({
```
ritch, line 31: This was used to test the readOnly flag in the isolated test app.

app/packages/dataset/src/index.tsx
```diff
@@ -28,28 +28,42 @@ const DatasetWrapper = () => {
...
+      : null;
+    const readOnly = readOnlyInputRef.current
+      ? readOnlyInputRef.current.checked
+      : false;
+    onChange((s) => ({ ...s, dataset, readOnly }));
   }
   return (
     <form onSubmit={load}>
-      <input ref={ref} type="text" defaultValue={current} />
+      <input
```
imanjra, line 60: Nice!

app/packages/dataset/src/Dataset.tsx
```diff
@@ -26,15 +26,17 @@ enum Events {
...
-export function Dataset({ datasetName, environment, theme }) {
+export function Dataset({ datasetName, environment, theme, readOnly }) {
   theme = theme || darkTheme;
 
   const [initialState, setInitialState] = useState();
   const [datasetQueryRef, loadDataset] = useDatasetLoader(environment);
+  const setReadOnly = useSetRecoilState(fos.readOnly);
 
   useEffect(() => {
+    setReadOnly(readOnly);
```
imanjra, line 37: Maybe we can have a state with multiple external configs?

imanjra (approved):

benjaminpkane (approved):
LGTM as well 👍

PR #2182, opened 2022-10-17
Fix relay compilation and ensure future compilations in builds

benjaminpkane:
Fixes relay compilation in @fiftyone/core and ensures future breaking changes will be caught in workflows via the `yarn build` script in @fiftyone/app.

Also removes unused prettier code in the app and adds __generated__ files created by relay to the .prettierignore which I find more convenient than compiling and reformatting every time via pre-commit.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,3 @@
 *.html
 voxel51-website.js
+**/__generated__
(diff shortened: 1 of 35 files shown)


ritch (approved):
LGTM

Issue #2163, opened 2022-10-14
[FR] Store and visualize custom field-level information


ehofesmann:
Similar to `dataset.info` it would be useful to be able to store some form of `dataset.field_name.info` that can contain a dictionary (or other structure) of any arbitrary information that the user may be interested in keeping in FiftyOne that corresponds to the same field of all samples in a dataset. 

It is important for this to be visualizable in the App as well.

Some example use cases are if you are training models, you will end up with multiple label fields containing model predictions. It would be useful to have a text description that users can see in the App to know what the goal/process of each training run was.

Similarly, when integrating with an experiment tracker like weights and biases, it is useful to store a link to the corresponding WandB run in FiftyOne, but the only way to be able to visualize a clickable link in the App is to duplicate it into a new field in every sample.

brimoor:
This can be achieved by adding (say) `description` and `info` fields to [SampleFieldDocument](https://github.com/voxel51/fiftyone/blob/c5c3269c0bc000c02848f071554089f1d79569ae/fiftyone/core/odm/dataset.py#L139).

Here's how this could work through the SDK:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")

schema = dataset.get_field_schema()
field = schema["ground_truth"]
print(field.description)

field.description = "Updated description"
field.save()

new_dataset = fo.Dataset()

new_dataset.add_sample_field(
    "ground_truth",
    fo.EmbeddedDocumentField,
    embedded_doc_type=fo.Detections,
    description="Annotations from project XXXX",
)
```

PR #1825, opened 2022-06-06
Adding support for dynamic embedded document field schemas

brimoor:
## Change log

- Adds support for declaring dynamic embedded document fields on the dataset's schema via `add_sample_field()` and `add_frame_field()`
- Added support for selecting/excluding embedded document fields via `select_fields()` and `exclude_fields()`
- Added a `dynamic=True` flag that can be passed to dataset factory methods that will cause all dynamic embedded document attributes that are encountered to be automatically added to the dataset's schema
- Added a `schema()` aggregation that can be used to compute the observed type(s) of arbitrarily nested embedded documents
- Added `get_dynamic_field_schema()` and `add_dynamic_sample_fields()` methods for automatically detecting and declaring dynamic sample fields
- Added `get_dynamic_frame_field_schema()` and `add_dynamic_frame_fields()` methods for detecting and declaring dynamic frame fields
- Added `flat=True` option to `get_field_schema()` and `get_frame_field_schema()` methods that returns all embedded document fields as top-level keys

## Notes

- The only default behavior that this PR changes is that `evaluate_detections()` will automatically add the dynamic attributes that it populates to the dataset's schema
- Dynamic attributes are not declared by default by `add_samples()`, `from_dir()`, etc
- As a result, there is no decrease in performance in the default case

## Example usage

Previously undeclared dynamic attributes can now be declared:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")
fo.pprint(dataset.get_dynamic_field_schema())

# Declare dynamic attributes
dataset.add_dynamic_sample_fields()
fo.pprint(dataset.get_dynamic_field_schema())

# Verify that they exist in the dataset's schema
fo.pprint(dataset.get_field_schema(flat=True))

# Dynamic attributes are available in the App for filtering
session = fo.launch_app(dataset)

# Dynamic attributes are carried over to patches views too
session.view = dataset.to_patches("ground_truth")
```

Screen Shot 2022-10-20 at 3 43 11 PM

Detection evaluation automatically declares the dynamic attributes that it populates:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")

dataset.evaluate_detections("predictions", gt_field="ground_truth", eval_key="eval")
fo.pprint(dataset.get_field_schema(flat=True))

session = fo.launch_app(dataset)

# Dynamic attributes can be excluded
# This syntax selects only the default fields on the detections
[...]

PR #2187, opened 2022-10-20
Adding support for field descriptions (option 1)

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/2163

Adds support for storing field-level descriptions on datasets.

## Notes

I vote for https://github.com/voxel51/fiftyone/pull/2216 instead of this implementation 🤗 

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.set_field_metadata(
    "ground_truth",
    description="Ground truth annotations",
)

dataset.set_field_metadata(
    "ground_truth.detections.area",
    description="Area of the box, in pixels^2",
)

print(field.description)  # Ground truth annotations

field = dataset.get_field("ground_truth.detections.area")
print(field.description)  # Area of the box, in pixels^2
```

Appropriate validation is in-place:

```py
# None of these are allowed
dataset.set_field_metadata("foo", description="bar")  # bad field
dataset.set_field_metadata("ground_truth.detections.foo", description="bar")  # bad attribute
dataset.set_field_metadata("ground_truth", foo="bar")  # bad metadata key
dataset.set_field_metadata("ground_truth", description=False)  # bad metadata value
```

Field descriptions may also be provided when the field is declared, if desired:

```py
import fiftyone as fo

dataset = fo.Dataset()
dataset.add_sample_field(
    "int_field", fo.IntField, description="An integer field"
)

field = dataset.get_field("int_field")
print(field.description)  # An integer field
```

### DIFF
--- a/docs/source/user_guide/using_datasets.rst
+++ b/docs/source/user_guide/using_datasets.rst
@@ -334,11 +334,6 @@ to update a field's metadata at any time:
     field = dataset.get_field("ground_truth.detections.area")
     print(field.description)  # Area of the box, in pixels^2
 
-.. note::
-
-    Did you know? You can view/edit field metadata directly
-    :ref:`in the App <fiftyone-app>`!
-
 .. _custom-app-config:
 
 Custom App config
--- a/docs/source/user_guide/using_datasets.rst
+++ b/docs/source/user_guide/using_datasets.rst
@@ -280,6 +280,65 @@ Datasets can also store more specific types of ancillary information such as
     the dataset's :meth:`info <fiftyone.core.dataset.Dataset.info>` property
     in-place to save the changes to the database.
 
+.. _storing-field-metadata:
+
+Storing field metadata
+----------------------
+
+You can store metadata such as descriptions on the :ref:`fields <using-fields>`
+of your dataset.
+
+One approach is to manually declare the field with
+:ref:`add_sample_field() <fiftyone.core.dataset.Dataset.add_sample_field>` with
+the appropriate metadata provided:
+
+.. code-block:: python
+    :linenos:
+
+    import fiftyone as fo
+
+    dataset = fo.Dataset()
+    dataset.add_sample_field(
+        "int_field", fo.IntField, description="An integer field"
+    )
+
+    field = dataset.get_field("int_field")
+    print(field.description)  # An integer field
+
+You can also use
+:ref:`set_field_metadata() <fiftyone.core.dataset.Dataset.set_field_metadata>`
+to update a field's metadata at any time:
+
+.. code-block:: python
+    :linenos:
+
+    import fiftyone as fo
+    import fiftyone.zoo as foz
+
+    dataset = foz.load_zoo_dataset("quickstart")
+    dataset.add_dynamic_sample_fields()
+
+    dataset.set_field_metadata(
+        "ground_truth",
+        description="Ground truth annotations",
+    )
+
+    dataset.set_field_metadata(
+        "ground_truth.detections.area",
+        description="Area of the box, in pixels^2",
+    )
+
+    field = dataset.get_field("ground_truth")
+    print(field.description)  # Ground truth annotations
+
+    field = dataset.get_field("ground_truth.detections.area")
+    print(field.description)  # Area of the box, in pixels^2
+
+.. note::
+
+    Did you know? You can view/edit field metadata directly
+    :ref:`in the App <fiftyone-app>`!
+
 .. _custom-app-config:
 
 Custom App config
--- a/docs/source/user_guide/using_datasets.rst
+++ b/docs/source/user_guide/using_datasets.rst
@@ -289,8 +289,8 @@ You can store metadata such as descriptions on the :ref:`fields <using-fields>`
 of your dataset.
 
 One approach is to manually declare the field with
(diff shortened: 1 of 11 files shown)


brimoor:
@ehofesmann any thoughts on how to support editing field descriptions from the App? 

A simple tooltip would be a reasonable way to view the description, but editing the description from the App may be nice too.

brimoor:
@ehofesmann Also curious if you have any thoughts on the `dataset.set_field_metadata()` syntax. Love it? Hate it?

ehofesmann:
This is awesome already! Some thoughts:

* Descriptions are a good start, but we really need to be able to store an arbitrary amount of data on fields. Something like a dictionary is something that might be better than a description string to start out. In terms of visualization, it could then use a the same structure as the json in the sample modal. Mainly because in the W&B use case, you will end up with many different URLs for each model run, so a single string field wouldn't be enough.

* I'm not sure needing to support editing through the App is necessary right out of the gate. The description of fields isn't particularly something that needs to be modified constantly. (Not saying it wouldn't be nice). Especially if we go with the dict route, it would be great to be able to edit arbitrary JSON in the sample modal as well so that would kill 2 birds with one stone (probably in a follow-up PR imo)

* Syntax: What about something like this:
```python
dataset["ground_truth"].description = "my field is cool"
dataset["ground_truth"].info = {"other_stuff_1": 51}

dataset["ground_truth.detections.area"].description = "Area of the box, in pixels^2",

field = dataset["ground_truth"]
print(field.description)
```
This could allow for extensions something like
```python
dataset["ground_truth"].values  # Equivalent to: dataset.values("ground_truth")
```

kaixi-wang:
I'm a little confused by what is valid/not valid. How are the first two bad examples different from the valid examples? Is it that they are 3 letters?

[GOOD] `dataset.set_field_metadata( "ground_truth", description="Ground truth annotations",)`
[BAD] `dataset.set_field_metadata("foo", description="bar")`  # bad field


[GOOD] `dataset.set_field_metadata("ground_truth.detections.area", description="Area of the box, in pixels^2",)`
[BAD] `dataset.set_field_metadata("ground_truth.detections.foo", description="bar")`  # bad attribute

brimoor:
@kaixi-wang the `set_field_metadata()` method is designed to let you set metadata (currently only a `description` property) on any fields or embedded fields that have *already been defined* on your dataset. So here's why each of these is not allowed:

```py
# This dataset doesn't have a field named `foo` defined
dataset.set_field_metadata("foo", description="bar")

# This dataset's `ground_truth.detections` field doesn't have an attribute named `foo` defined
dataset.set_field_metadata("ground_truth.detections.foo", description="bar")

# `foo` is not a valid metadata field. Currently only `description` is allowed
dataset.set_field_metadata("ground_truth", foo="bar")  # bad metadata key

# `description` must be a string
dataset.set_field_metadata("ground_truth", description=False)
```

And these are allowed:

```py
# This dataset has a `ground_truth` field
dataset.set_field_metadata("ground_truth", description="...")

# This dataset's `ground_truth.detections` field has an `area` attribute defined
dataset.set_field_metadata("ground_truth.detections.area", description="...")
```

You can see what fields/embedded fields exist on a dataset using the following syntax (which was just added in https://github.com/voxel51/fiftyone/pull/1825 🤗 ):

```py
print(dataset.get_field_schema(flat=True))
```

brimoor:
@ehofesmann Hmm interesting: if we allowed attaching an arbitrary dict of info to a field, how would you want this info displayed in the App? I can see why one might want a `description` and a `url`, for example. But the real value of storing a URL would be that you have a way to click on it from the Ap.

One side comment is that, although I added `dataset.info`, I don't love it, because it will be challenging to build a UI for visualizing arbitrary key-value pairs in a way that makes the information look good, be clickable (URLs, for example), and also editable.

The `dataset["ground_truth"]` syntax is definitely concise, but I'm worried that we might be overloading `dataset[]` too much as these things are already supported:

```
dataset[sample_id]
dataset[filepath]
dataset[1:10]
```

One other thing I should mention is note that `field.save()` is not currently something you can do:

```py

# Not allowed
field.description = "foo"
field.save()
```

because the `Field` class doesn't currently store a reference to its parent dataset, which would be required in order for `field.save()` or other things like potentially `field.values()` to work. We could add it of course, but just pointing it out.

ehofesmann:
@brimoor How I'm imagining we could display this info in the App:
* The description field could be special in that if you hover over a field on the sidebar, it could show the description in a tooltip. This wouldn't work for an arbitrary number of custom field metadata, nor for links that you need to click since the tooltip would just move away. But a description is exactly what I would want to see on hover.
* We could add some clickable icon/area on the fields in the sidebar that have field metadata populated which would show a modal similar to the JSON overlay, but over the grid view of the App. This would allow users to see a full page of all of the custom field metadata they added (like model hyperparameters and urls they could click)

Yeah, overloading the `dataset[]` syntax might be an issue. I was just thinking in terms of how pandas operates when you call `dataframe[]` but the catch is that if you want to do `dataset[1:10]`, the equivalent would be `dataframe.iloc[1:10]` so we swapped what needs a method call vs what is built into the brackets. (I still prefer our syntax)

brimoor:
Closing in favor of https://github.com/voxel51/fiftyone/pull/2216.

PR #2189, opened 2022-10-21
Migrating runs/views to separate collections

brimoor:
Migrates saved views and annotation/brain/evaluation runs into separate `views` and `runs` collections, respectively. 

This is preferable to directly storing these documents on the `DatasetDocument` because both types of documents store serialized views, whose descriptions can be quite large (eg, `dataset.select(["very", "long", "list", "of", "ids"])`).


fiftyone/core/odm/views.py
```diff
@@ -13,14 +13,14 @@
...
+from .document import Document
 
 
-class ViewDocument(EmbeddedDocument):
+class ViewDocument(Document):
     """Backing document for dataset views."""
 
     # strict=False lets this class ignore unknown fields from other versions
-    meta = {"strict": False}
+    meta = {"collection": "views", "strict": False}
```
kaixi-wang, line 23: ohh so this is how you add a collection

PR #2190, opened 2022-10-21
Adding support for populating default App sidebar groups via SDK

brimoor:
Adds a `DatasetAppConfig.default_sidebar_groups()` method that can be used to populate the default sidebar groups programmatically.

Also documents the App's sidebar mode and groups features.

```py
import fiftyone as fo
import fiftyone.zoo as foz

# dataset = foz.load_zoo_dataset("quickstart-video")

sidebar_groups = fo.DatasetAppConfig.default_sidebar_groups(dataset)
dataset.app_config.sidebar_groups = sidebar_groups
print(dataset.app_config)

```


fiftyone/core/odm/dataset.py
```diff
@@ -357,6 +380,128 @@ def _rename_paths(self, paths, new_paths):
...
+            primitives,
+            other,
+            unsupported,
+            frames=True,
+        )
+
+    sidebar_groups = [
+        SidebarGroupDocument(name="tags", expanded=True),
+        SidebarGroupDocument(name="label tags", expanded=True),
+        SidebarGroupDocument(name="metadata", paths=metadata, expanded=False),
```
kaixi-wang, line 424: Not something that needs to be addressed by this PR, but I noticed all the metadata fields are prefixed with `metadata`, which causes some of the longer field names to be hidden. Since the metadata section already has a heading, is there a reason we need to display the prefix?

kaixi-wang:
From the example usage, it appears that the config is saved on the dataset? Does this means anyone who loads the dataset will now have those settings?

fiftyone/core/odm/dataset.py
brimoor: Yeah that's a good observation. A couple comments:
1. This PR is duplicating server-side the clientside logic that already exists and is performed to construct default sidebar groups if none have been explicitly configured
2. The reason that the `metadata` section uses `metadata.XXX` for the individual field names is that the group name is arbitrary, so if the user renamed it to something other than `METADATA`, then they might be confused where the `XXX` field is coming from

brimoor:
> From the example usage, it appears that the config is saved on the dataset? Does this means anyone who loads the dataset will now have those settings?

Yep good observation 👍  OSS doesn't have any notion of users; everything that is persistent is saved on the dataset itself. In Teams, I could see both global and per-user being useful. Some organization would be generally useful, while other users might have personal preferences for things being organized differently. We'd need to build support for the latter case into the Teams project if we want to support it.

kaixi-wang (approved):
lgtm. thanks for answering my questions!

Issue #1270, opened 2021-09-16
[BUG] Video out of sync with label overlay


doronpor:
### System information

-   Linux Ubuntu 18.04.2 LTS
-   installed from source
-   version 0.13.2
-   python version 3.6.9

### Commands to reproduce

```
1. Create a video with the details bellow
2. Load a database with high amount of labels
3. hover over the Thumbnail or play a video in full screen
```

### Describe the problem

Video occasionally gets out of sync with label overlay, i.e. when hovering on the thumbnail
or when playing video of the sample, the image would start playing in a high rate and 
the labels would catch-up after the video well after the video would stop playing.

### Code to reproduce issue

**Video created with the following parameters**:

_ffmpeg output parameters_:
["-f", "mp4", "-vcodec", "libx264", "-vb",  "4096K",
"-minrate", "4096K", "-maxrate", "4096K", "-bufsize", "250k",                                                                                                                                     "-pix_fmt", "yuv420p", "-vsync", "0"]
_video resolution_: 2688 x 1520
_frame rate_: 4 fps
_Labels_: There are high amount of labels on the video (7 layers including detections and polylines)

**connection information**:

Server is running in linux system. 
Application running on windows remote system (edge or chrome)

### What areas of FiftyOne does this bug affect?

-   [X] `App`: FiftyOne application issue
-   [ ] `Core`: Core `fiftyone` Python library issue
-   [ ] `Server`: Fiftyone server issue

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another
member of your organization be willing to contribute a fix for this bug to the
FiftyOne codebase?

-   [ ] Yes. I can contribute a fix for this bug independently.
-   [ ] Yes. I would be willing to contribute a fix for this bug with guidance
        from the FiftyOne community.
-   [X] No. I cannot contribute a bug fix at this time.

brimoor:
Hi @doronpor 👋 

Thanks for reporting this. I noticed some occasional frame-label misalignments when doing some in-progress work on video features recently as well. We'll take a look at this and report back ASAP here

doronpor:
Hi Brimoor,
Any success with identifying the issue ?

benjaminpkane:
HI @doronpor. I've made a video per your description and have tested it with a ~200 detections, but am not seeing any issues.

https://github.com/voxel51/fiftyone/pull/1228 has a fix in the video player related to sparse frame labels, but that does not sound like your issue.

[...]

PR #2191, opened 2022-10-24
App sidebar mode and optimizations

benjaminpkane:
### Change log

- Significantly optimizes the performance of the sidebar by lazily computing statistics only for currently visible fields
- Added new sidebar modes `all|best|fast` with updated default behavior that further optimizes the performance of the App for large datasets
- Added support for configuring the sidebar mode dynamically in the App and programmatically on a per-dataset basis
- Added support for programmatically configuring sidebar groups and default expansion states on a
  per-dataset basis

### Issues resolved

* Resolves https://github.com/voxel51/fiftyone/issues/1270
* Resolves https://github.com/voxel51/fiftyone/issues/2255
* Resolves https://github.com/voxel51/fiftyone/issues/2244


benjaminpkane:
Stale

brimoor:
@benjaminpkane can you add a loading state to the `Loading` indicators in the filter widgets? Something as simple as animating the dots:

```
Loading.
Loading..
Loading...
```

This would help convey that something is happening, like the spinners for the counts.

benjaminpkane:
> @benjaminpkane can you add a loading state to the `Loading` indicators in the filter widgets?
> 
> Something as simple as animating the dots:
> 
> ```
> Loading.
> Loading..
> Loading...
> ```
> 
> This would help convey that something is happening, like the spinners for the counts.

Ok, I added the above

brimoor (approved):
LGTM

PR #2193, opened 2022-10-25
Allow frame views to edit source frame filepaths

brimoor:
Previously, editing the `filepath` of a frame view would not sync to the source dataset's frame `filepath`, but this should be allowed, since frame views were intended to sync all relevant frame fields of the source dataset. This PR resolves the bug.

```py
import fiftyone as fo
import fiftyone.zoo as foz

frames = dataset.to_frames(sample_frames=True)

sample = frames.first()
sample.filepath = sample.filepath.upper()
sample.save()

# Previously wasn't changed; now is
print(dataset.first().frames.first().filepath)

upper_view = frames.set_field("filepath", F("filepath").upper())
frames.set_values("filepath", upper_view.values("filepath"))

# Previously wasn't changed; now is
print(dataset.values("frames.filepath", unwind=True)[:5])

frames.set_field("filepath", F("filepath").upper().concat(".bak")).save()

# Previously wasn't changed; now is
print(dataset.values("frames.filepath", unwind=True)[:5])
```

### DIFF
--- a/fiftyone/core/video.py
+++ b/fiftyone/core/video.py
@@ -142,6 +142,21 @@ class FramesView(fov.DatasetView):
     def media_type(self):
         return fom.IMAGE
 
+    def _get_sample_only_fields(
+        self, include_private=False, use_db_fields=False
+    ):
+        sample_only_fields = set(
+            self._get_default_sample_fields(
+                include_private=include_private, use_db_fields=use_db_fields
+            )
+        )
+
+        # If sample_frames != dynamic, `filepath` can be synced
+        if self._frames_stage.config.get("sample_frames", None) != "dynamic":
+            sample_only_fields.discard("filepath")
+
+        return sample_only_fields
+
     def _get_default_sample_fields(
         self, include_private=False, use_db_fields=False
     ):
--- a/fiftyone/core/video.py
+++ b/fiftyone/core/video.py
@@ -152,7 +152,8 @@ class FramesView(fov.DatasetView):
         )
 
         # If sample_frames != dynamic, `filepath` can be synced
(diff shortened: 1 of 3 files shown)


fiftyone/core/video.py
```diff
@@ -142,6 +142,22 @@ def name(self):
...
+    ):
+        sample_only_fields = set(
+            self._get_default_sample_fields(
+                include_private=include_private, use_db_fields=use_db_fields
+            )
+        )
+
+        # If sample_frames != dynamic, `filepath` can be synced
+        config = self._frames_stage.config or {}
+        if config.get("sample_frames", None) != "dynamic":
```
manivoxel51, line 156: nit: alternatively you could return early if and remove this if
```
...
if config.get("sample_frames", None) == "dynamic":
  return
... rest ...
```

fiftyone/core/video.py
```diff
@@ -273,16 +289,14 @@ def _sync_source_sample(self, sample):
         self._sync_source_schema()
 
         dst_dataset = self._source_collection._root_dataset
-        default_fields = set(
```
manivoxel51, line 276: nit: could removing the `set` have duplicate consequences 🤔

manivoxel51 (approved):
nice work Brian! 🍦 

nit: your description of the bug fix is great! It could be useful to show the output of print() statements in the description

fiftyone/core/video.py
brimoor: In the new implementation, `_get_sample_only_fields()` directly returns set, so we're all set ✔️

fiftyone/core/video.py
brimoor: Hmm yeah could go either way

PR #2201, opened 2022-10-27
Support cuda devices for CLIP zoo model

ehofesmann:
When running the CLIP zoo model on a cuda device, it would currently fail due to the `text_features` not being properly placed on the cuda device as well causing a device mismatch when running a forward pass.

This PR places the `text_features` tensor on the same device as the model itself to resolve this issue. Tested both on a cuda device and cpu device.


brimoor (approved):
LGTM 🚀

Issue #2095, opened 2022-09-19
[FR] Optimize Apps' label tagging implementation


brimoor:
The App's label tagging implementation needs some optimization. For example, if I select a single label and add a tag to it on a dataset with 10k samples and 200k labels, the operation takes ~30 seconds or so.

There are a few cases that need to be considered:
- Whether tags are being added, removed, or both
- Whether the user is operating on selected labels or the entire view

When operating on selected labels, the operation can be O(1) by first selecting the relevant samples -> fields -> IDs:

```py
fields = list(set(d["field"] for d in selected_labels))
sample_ids = list(set(d["sample_id"] for d in selected_labels))
ids = list(set(d["label_id"] for d in session.selected_labels))

view = dataset.select(sample_ids).select_labels(ids=ids, fields=fields)
view.tag_labels(tags)
```

When removing a label tag from an entire view, it would be faster to first use `select_labels(tags=)` to isolate the tag being removed:

```py
dataset.select_labels(tags=tags).untag_labels(tags)
```

Obviously if a label tag is being added to an entire dataset, then there's not much room for optimization (many mutations must take place). But, in most cases, the user will only be doing a very small number of mutations, so the implementation should be optimized so that those common cases will be fast.

PR #2203, opened 2022-10-27
Optimizing tagging implementations

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/2095.

Optimizes the sample/label tagging implementations in the SDK and App.

The previous implementation would trigger update operations on every sample/label in the dataset in situations such as calling `dataset.tag_samples(tag)` on a dataset where all but a few samples already have the tag of interest.

The new implementation selects only the samples/labels that need edits, which is a big optimization in the typical case of applying a small number of edits. 

### Notes

1. The `changes` made by the App's tagging UI had to be split into additions and deletions, but that is fine because the real win is that the number of update operations is now proportional to the number of actual edits

2. All tagging implementations still involve reading the tags into memory and then using `set_values()` to commit the updated tags. I believe it should be possible to instead read only the IDs into memory and express the corresponding edits using update operators, but I haven't invested the time in that yet...

### Experiments

I also experimented with the following implementations of sample tagging, which are 100% aggregation pipelines whose last stage is `$merge`. However, I found that these implementations are roughly 50% slower when a small number of changes are made and only 50% faster when a large number of changes are made, so it didn't seem worth it.

```py
def tag_samples(self, tags):
    tags = [tags] if etau.is_str(tags) else list(tags)

    # We only need to process samples that are missing a tag of interest
    view = self.match_tags(tags, bool=False, all=True)

    add_tags = F("tags").extend(E(tags).difference(F("tags")))
    tag_expr = (F("tags") != None).if_else(add_tags, tags)
    view.set_field("tags", tag_expr).save("tags")


def untag_samples(self, tags):
    tags = [tags] if etau.is_str(tags) else list(tags)

    # We only need to process samples that have a tag of interest
    view = self.match_tags(tags)

    del_tags = F("tags").filter(~E(tags).contains([F()]))
    view.set_field("tags", del_tags).save("tags")
```


benjaminpkane (approved):
LGTM 🐎

PR #2205, opened 2022-10-28
[DRAFT] Saved views server support

kaixi-wang:

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/server/routes/view.py
+++ b/fiftyone/server/routes/view.py
@@ -1,0 +1,34 @@
+"""
+FiftyOne Server ``/view`` route.
+| Copyright 2017-2022, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+import glob
+import os
+import numpy as np
+
+from starlette.endpoints import HTTPEndpoint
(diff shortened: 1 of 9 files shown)


fiftyone/server/routes/view.py
```diff
@@ -0,0 +1,34 @@
...
+from starlette.requests import Request
+
+import eta.core.serial as etas
+import fiftyone.server.view as fosv
+
+import fiftyone as fo
+from fiftyone.server.decorators import route
+
+
+class View(HTTPEndpoint):
```
manivoxel51, line 21: curious when we would be using the http endpoint 🤔

kaixi-wang: I was initially thinking of using endpoints to trigger the SDK functions, but there is probably a more acceptable way of triggering `fo.load_dataset` from the app...

@benjaminpkane Do you have any recommendations? Is another way of calling the SDK functions Brian already wrote using graphQL? Or is the only other option to not use 'load_view' and 'save_view' and just implement the mutations/queries for saved views directly?

PR #2207, opened 2022-10-28
Transparent looker background

benjaminpkane:
Changes the looker background to be transparent so background color is configured by the parent element. Fixes tile coloring in the grid.


imanjra (approved):
lgtm! 🎨

PR #2208, opened 2022-10-31
More tagging optimization

brimoor:
Implements my idea from https://github.com/voxel51/fiftyone/pull/2203: all tagging operations are now done via update operations!

```py
{"$addToSet": {"tags": {"$each": tags}}}  # add tags
{"$pullAll": {"tags": tags}}  # remove tags
```

This is a much better implementation than the current one because:
- The tag edits are no longer applied in-memory; `values()` and `set_values()` are no longer used
- When tagging labels in generated views (patches/frames/clips), the relevant changes are applied to the source collection only on the affected labels by ID, whereas previously the entire field was synced to the source collection via a `$merge` aggregation

### DIFF
--- a/fiftyone/core/clips.py
+++ b/fiftyone/core/clips.py
@@ -5,8 +5,8 @@ Clips views.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-from copy import deepcopy
 from collections import defaultdict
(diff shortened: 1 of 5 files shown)


benjaminpkane (approved):
LGTM

PR #2214, opened 2022-11-01
Export state from dataset module

ritch:
Allows for managing state externally.

See [here](https://github.com/voxel51/fiftyone/pull/2214/files#diff-df67c1016008538fab6072e014adc3a196aeb28234134be32382edf99f41c043R41) for example usage.

### DIFF
--- a/app/packages/dataset/src/index.ts
+++ b/app/packages/dataset/src/index.ts
@@ -1,6 +1,8 @@
-export {Dataset} from './Dataset'
-import {getEnvironment, RelayEnvironmentKey} from '@fiftyone/state'
+export { Dataset } from "./Dataset";
+import { getEnvironment, RelayEnvironmentKey } from "@fiftyone/state";
 
 export function getEnvProps() {
-  return {environment: getEnvironment(), environmentKey: RelayEnvironmentKey}
+  return { environment: getEnvironment(), environmentKey: RelayEnvironmentKey };
 }
+
+export * as fos from "@fiftyone/state";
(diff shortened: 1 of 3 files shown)


app/packages/dataset/src/index.tsx
```diff
@@ -25,14 +25,39 @@ const DatasetWrapper = () => {
   );
 };
 
+const EXAMPLE_VIEW = [
+  {
+    _cls: "fiftyone.core.stages.Limit",
+    kwargs: [["limit", 10]],
+    _uuid: "020b33dd-775a-4b4a-a865-4901e2e6ee43",
```
kaixi-wang, line 32: is this is just a hardcoded view to demo changeView()? Can _cls be a list that positionally aligns with kwargs? Or how would you apply multiple stages?

kaixi-wang: Oh yeah that makes sense, I missed the `[]`

app/packages/dataset/src/index.ts
```diff
@@ -1,6 +1,8 @@
...
+import { getEnvironment, RelayEnvironmentKey } from "@fiftyone/state";
 
 export function getEnvProps() {
-  return {environment: getEnvironment(), environmentKey: RelayEnvironmentKey}
-}
\ No newline at end of file
+  return { environment: getEnvironment(), environmentKey: RelayEnvironmentKey };
+}
+
+export * as fos from "@fiftyone/state";
```
imanjra, line 8: Nice!

imanjra (approved):
lgtm

Issue #2164, opened 2022-10-14
[BUG] Error when loading a group dataset with alternate media fields in the App


ehofesmann:
The App raises an error when you have a group dataset with an alternate media field populated, but not for all samples:

```python
import fiftyone as fo
import fiftyone.utils.image as foui
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-groups").clone()
sample = dataset.first()
sample["thumbnail"] = "/tmp/quickstart-1.png"
foui.transform_image(sample.filepath, sample.thumbnail, size=(64, -1)) 
sample.save() 
dataset.app_config.media_fields = ["filepath", "thumbnail"]
dataset.app_config.grid_media_field = "thumbnail"
dataset.save()

session = fo.launch_app(dataset)
```

```
Cannot return null for non-nullable field MediaURL.url.
File "/home/eric/.pyenv/versions/3.9.13/envs/fo/lib/python3.9/site-packages/graphql/execution/execute.py", line 628, in await_result
return await completed
File "/home/eric/.pyenv/versions/3.9.13/envs/fo/lib/python3.9/site-packages/graphql/execution/execute.py", line 456, in get_results
await gather(*(results[field] for field in awaitable_fields)),
File "/home/eric/.pyenv/versions/3.9.13/envs/fo/lib/python3.9/site-packages/graphql/execution/execute.py", line 632, in await_result
self.handle_field_error(error, return_type)
File "/home/eric/.pyenv/versions/3.9.13/envs/fo/lib/python3.9/site-packages/graphql/execution/execute.py", line 666, in handle_field_error
raise error
File "/home/eric/.pyenv/versions/3.9.13/envs/fo/lib/python3.9/site-packages/graphql/execution/execute.py", line 628, in await_result
return await completed
File "/home/eric/.pyenv/versions/3.9.13/envs/fo/lib/python3.9/site-packages/graphql/execution/execute.py", line 841, in get_completed_results
await gather(
File "/home/eric/.pyenv/versions/3.9.13/envs/fo/lib/python3.9/site-packages/graphql/execution/execute.py", line 821, in await_completed
self.handle_field_error(error, item_type)
File "/home/eric/.pyenv/versions/3.9.13/envs/fo/lib/python3.9/site-packages/graphql/execution/execute.py", line 666, in handle_field_error
raise error
[...]

PR #2217, opened 2022-11-01
Fix unset media field

ritch:
Fixes [this](https://github.com/voxel51/fiftyone/issues/2164) issue by allowing mediaFields to be null.

Also adds improved error messages when a mediaField is null in Looker3D.


brimoor (approved):
Works, thanks!

PR #2218, opened 2022-11-01
Set view name on saved views

kaixi-wang:

- Set the `name` property of the `DatasetView` to the `ViewDocument.name` from which the view was loaded from. 
- This will make routing directly to a view simpler as the saved view name (if it exists) can be accessed via `view.name`

- Updated save view unittest to check for name property

        notes for FiftyOne users.

-   [ ] Other


fiftyone/core/dataset.py
```diff
@@ -2841,7 +2841,7 @@ def save_view(
         """
         if view._root_dataset is not self:
             raise ValueError("Cannot save view into a different dataset")
-
+        view._set_name(name)
```
kaixi-wang, line 2844: Note: We are only allowing the view.name to be set if the view is loaded from a saved view. Thus, the only way to get the name of a DatasetView to be a specific string is to call `dataset.save_view("<view_name>")` or `dataset.update_view_info("<current_view_name>", {"name": "<new_view_name>"}).

brimoor: Or `load_view()`! Which would be the typical way to get your hands on a saved view.

I don't think this is relevant to this PR, but an important consideration will be that accessing a saved view loaded from within the App via `session.view` should also have the saved view name in-place in Python

fiftyone/core/view.py
```diff
@@ -238,7 +244,10 @@ def default_group_slice(self):
 
     @property
     def name(self):
-        """The name of the view."""
+        """The name of the view with which it was saved. Defaults to `<dataset_name>-view` for unsaved views."""
```
brimoor, line 247: nit: unfortunately the linter doesn't automatically enforce line length limits for comments, but our style is to always keep everything to 79 characters or fewer.

My two cents would be to document like this:

```py
    @property
    def name(self):
        """The name of the view.
    
        If this view is a saved view, its saved name will be returned.
        Otherwise, a default name will be returned.
        """
```

fiftyone/core/view.py
```diff
@@ -317,6 +326,7 @@ def summary(self):
             a string summary
         """
         elements = [
+            ("View name:", self.name),
```
brimoor, line 329: How about only including `View name` if the view is actually a saved view?

When it is not a saved view, the default name is not meaningful and I think it's helpful to reduce clutter whenever possible on these heavily used outputs.

fiftyone/core/view.py
```diff
@@ -1331,6 +1341,7 @@ def _add_view_stage(self, stage):
             media_type = stage.get_media_type(self)
             if media_type is not None:
                 view._set_media_type(media_type)
+            view._set_name(None)
```
brimoor, line 1344: This needs to be outside the `if` block. The `stage.has_view` case also returns a different view which is not the saved view.

fiftyone/core/view.py
brimoor: Related suggestion: how about adding a property like `DatasetView.is_saved_view` to allow users to determine if a view is a saved view or not?

brimoor (changes requested):
Lookin' good, left a couple comments for your consideration 😎

fiftyone/core/dataset.py
kaixi-wang: I set the name on `load_view` [here](https://github.com/voxel51/fiftyone/pull/2218/files#diff-d4170c49dd9265f224bc766f1675ff3a6690cb4103051867110901c46f787abeR2950) as well, but the method name got collapsed in the diff. Is this what you were referring to?

Agree. I'm handling the App side of it in another PR

fiftyone/core/view.py
kaixi-wang: ah didn't notice the line length didn't auto-format correctly. I'll see if there's a setting in my IDE to do it automatially.

fiftyone/core/dataset.py
brimoor: Yeah no action, I was just commenting that the primary way a user will get their hands on a saved view is by calling `load_view()` or accessing `session.view` after loading a saved view in the App

brimoor: I see that you handled `load_view()` in your implementation. Nice 💪

fiftyone/core/view.py
kaixi-wang: On second thought, thinking about the saved view selection dropdown impl/the usage in the App, does it make sense to return the `<dataset_name>-view` as the default? Maybe it would make life easier to make the default something along the lines of `Unsaved`. What do you think?

brimoor: It sounds like you're thinking of `DatasetView.name` as a way to communicate to the App whether the current view is saved. For that purpose, I would recommend using the `is_saved` pattern I mentioned:

```py
dataset_name = view._dataset.name
stages = view._serialize()
saved_name = view.name if view.is_saved else None  # use None for unsaved here
```

For the SDK, `SampleCollection.name` is declared mostly for cosmetic reasons. Here's one place where its used currently:

```py
import fiftyone as fo
import fiftyone.zoo as foz

d = view.to_dict()

fo.pprint(d)

"""
{
    'name': 'quickstart-1-view',
    'version': '0.18.0',
    'media_type': 'image',
    ...
}
"""

dataset2 = fo.Dataset.from_dict(d)
print(dataset2.name)  # quickstart-1-view
```

It's not really essential that the exported JSON contains the source dataset name as root, but my thinking before saved views existed was just to leave a breadcrumb in the JSON back to the original dataset.

brimoor: However, now that saved views exist, perhaps we should just return `name == None` for unsaved views? That would be the most clear way to indicate to the user that views *can* have well-defined names, but this one doesn't have one.

I don't think any existing usages of `DatasetView.name` would mind if `name == None`.

brimoor: If we do go with this change, then I would be fine with getting rid of `DatasetView.is_saved` and just having the convention that unsaved views have `name=None`.

brimoor: FYI I pushed a commit that fixes the only potential problem I saw if we allow `DatasetView.name=None`

fiftyone/core/clips.py
```diff
@@ -86,13 +91,15 @@ def __init__(
...
         self.__stages = _stages
+        self.__name = _name
 
     def __copy__(self):
         return self.__class__(
             self._source_collection,
             deepcopy(self._clips_stage),
             self._clips_dataset,
             _stages=deepcopy(self.__stages),
+            _name=self.__name,
```
brimoor, line 102: Included `name` here because it makes sense to me that copying a saved view should still be the same saved view

fiftyone/core/clips.py
```diff
@@ -74,7 +74,12 @@ class ClipsView(fov.DatasetView):
     """
 
     def __init__(
-        self, source_collection, clips_stage, clips_dataset, _stages=None
+        self,
```
brimoor, line 77: `ClipsView` is a `DatasetView` subclass so it needed `name` implementation too

fiftyone/core/patches.py
```diff
@@ -90,7 +90,12 @@ class EvaluationPatchView(_PatchView):
 
 class _PatchesView(fov.DatasetView):
     def __init__(
-        self, source_collection, patches_stage, patches_dataset, _stages=None
+        self,
```
brimoor, line 93: Patches views needed the `name` implementation too

fiftyone/core/view.py
```diff
@@ -123,6 +129,7 @@ def __copy__(self):
             _stages=deepcopy(self.__stages),
             _media_type=self.__media_type,
             _group_slice=self.__group_slice,
+            _name=self.__name,
```
brimoor, line 132: I included name in `__copy__` implementation with the idea that copying a saved view is still the same saved view

tests/unittests/dataset_tests.py
```diff
@@ -2118,38 +2118,56 @@ def test_saved_views(self):
...
-        self.assertTrue(dataset.has_view("test"))
-        self.assertListEqual(dataset.list_views(), ["test"])
+        self.assertTrue(dataset.has_view(view_name))
+        self.assertListEqual(dataset.list_views(), [view_name])
 
         self.assertIsNone(last_loaded_at1)
         self.assertIsNotNone(last_modified_at1)
 
-        also_view = dataset.load_view("test")
+        still_saved_view = deepcopy(view)
```
brimoor, line 2141: Added some more tests for copying, view chaining, and `is_saved` property

tests/unittests/patches_tests.py
```diff
@@ -265,6 +267,25 @@ def test_to_patches(self):
         with self.assertRaises(KeyError):
             sample["ground_truth"]
 
+        # Test saving a patches view
```
brimoor, line 270: Added some tests for patches/frames/clips views below here

brimoor (approved):
LGTM. Note that there are some subclasses of `DatasetView` that needed this implementation too

PR #2225, opened 2022-11-03
Remove old session client

benjaminpkane:
Removes the old client module that has been replaced by [fiftyone.core.session.client](https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/session/client.py).


kaixi-wang (approved):
lgtm!

PR #2226, opened 2022-11-03
fix typo in colorscale route

kaixi-wang:

- fix typo: `coloscales` -> `colorscales` so that the get request will work (https://github.com/voxel51/fiftyone/blob/develop/fiftyone/server/routes/colorscales.py#L19)

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/server/routes/__init__.py
+++ b/fiftyone/server/routes/__init__.py
@@ -29,7 +29,7 @@ from .values import Values
 routes = [
     ("/aggregate", Aggregate),
     ("/aggregations", Aggregations),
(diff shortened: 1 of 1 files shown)


benjaminpkane:
Good catch. I think this can be removed, actually. It's not used anywhere. GraphQL can be used instead.

kaixi-wang:
> Good catch. I think this can be removed, actually. It's not used anywhere. GraphQL can be used instead.

Yeah I didn't find any usages, so I just removed the route

benjaminpkane (approved):
LGTM 🧹

PR #2227, opened 2022-11-03
Optimizing ordered select implementation

brimoor:
In the typical case where the number of selected IDs is small compared to the number of total documents, this implementation of ordered `select()`, `select_by()` and `select_groups()` should be more efficient. `$match` first, then `$set`.


benjaminpkane (approved):
LGTM

PR #2228, opened 2022-11-03
Update CVAT integration and docs to point to app.cvat.ai

ehofesmann:
CVAT has recently moved from hosting their server from cvat.org to app.cvat.ai. This PR updates the integration to point to app.cvat.ai by default and updates corresponding the references in the docs and tutorials.


brimoor (approved):
LGTM

PR #2230, opened 2022-11-03
Sync router state

### LINKED ISSUE #2229: [BUG] View bars swallows first new stage I try to create
On `develop`, the view bar will reliably fail to load a `Limit(10)` stage when I am viewing the `quickstart-geo`, `quickstart-groups`, and `bdd100k` datasets. This does not happen on the `quickstart` dataset. Likely only a problem for larger datasets?

This only happens the first time I try to use the view bar after loading any of these datasets. The next time I try to use the view bar, it works (until I change datasets again).

benjaminpkane:
Resolves #2229. This is a syncing issue with recoil/relay and the url state. @kaixi-wang per you offline comment, Recoil Sync could be a great fit for keeping state in sync. The current implementation isn't ideal.

### DIFF
--- a/app/packages/state/src/hooks/useSetView.ts
+++ b/app/packages/state/src/hooks/useSetView.ts
@@ -5,7 +5,7 @@ import { useMutation } from "react-relay";
 import { useRecoilTransaction_UNSTABLE, useRecoilValue } from "recoil";
 import { State, stateSubscription, view } from "../recoil";
 import { RouterContext } from "../routing";
-import { getDatasetName, transformDataset } from "../utils";
+import { transformDataset } from "../utils";
 import useSendEvent from "./useSendEvent";
 import useStateUpdate from "./useStateUpdate";
 import * as fos from "../";
@@ -14,6 +14,7 @@ const useSetView = () => {
   const send = useSendEvent(true);
   const updateState = useStateUpdate();
   const subscription = useRecoilValue(stateSubscription);
+  const router = useContext(RouterContext);
   const [commit] = useMutation<setViewMutation>(setView);
 
   const onError = useErrorHandler();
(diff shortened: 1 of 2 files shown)


brimoor (changes requested):
@benjaminpkane now every time I update the view the entire App reloads. Can you avoid that?

brimoor:
@benjaminpkane I am still seeing the issue pictured above. Also setting `session.view = None` gets me to an invalid state pictured below:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Create a Limit(10) stage in the App (whole app refreshes)

session.view = None

# App is now in inconsistent state pictured below
```

Screen Shot 2022-11-04 at 7 46 57 PM

kaixi-wang:
> Resolves #2229. This is a syncing issue with recoil/relay and the url state. @kaixi-wang per you offline comment, Recoil Sync could be a great fit for keeping state in sync. The current implementation isn't ideal.

I'm currently trying to get it working for routing to saved views, but if it works well, we can probably use it for datasets as well.

benjaminpkane:
Things should work now

app/packages/state/src/hooks/useSetView.ts
```diff
@@ -40,6 +41,14 @@ const useSetView = () => {
             },
             onError,
             onCompleted: ({ setView: { dataset, view: value } }) => {
+              console.log(router.history.location.state);
```
brimoor, line 44: remove `console.log`?

brimoor (approved):
LGTM

brimoor (changes requested):
Hmm okay nevermind, still seeing some state sync issues.

Here's my diagnosis of the GIF:
1. Grid loads wrong number of samples when `limit(5)` view is loaded from Python. I believe the first page of results from the previous state is being shown (not shown in GIF but I cannot scroll past first page). Refreshing in browser resolves the issue
2. Loading `limit(1)` stage now shows first page of previous view's  results in grid
3. Loading `limit(1)` a second time syncs everything correctly, but now the view bar is non-responsive

benjaminpkane:
I can't reproduce any of the above. Can you do a clean install and rebuild. And clear your cache, of course.

`cd ./app && rm -rf node_modules && yarn install && yarn build`

brimoor:
My normal build is to use `install.bash`. I ran `cd ./app && rm -rf node_modules && yarn install && yarn build` this time and am still seeing the issues I reported.

```
sse-starlette==0.10.3
starlette==0.20.4
strawberry-graphql==0.138.1
```

benjaminpkane:
OK, we will have to take this offline. Still unable to reproduce

PR #2231, opened 2022-11-03
Downgrade starlette

j053y:
I didn't notice this got bumped in the previous PR. The version was too high for FastAPI and VoxelHub to be able to use it. Dropping it back down.


brimoor (approved):
LGTM

PR #2234, opened 2022-11-04
fix toggling field not capturing click some times

imanjra:

Fixes an issue where toggle field does not register click sometime if you move cursor right after click a field or checkbox.

Visually inspected to ensure clicking and moving the cursor away is registered as toggle but drag or selection is not.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Sidebar/Entries/Draggable.tsx
+++ b/app/packages/core/src/components/Sidebar/Entries/Draggable.tsx
@@ -29,6 +29,9 @@ const Draggable: React.FC<
   return (
     <>
       <animated.div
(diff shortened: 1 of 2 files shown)


benjaminpkane (changes requested):
I like what you did here. In truth, I think you can just use the builtin `onClick`. This old approach is there for historical reasons.

You'd just need to stop propagation in the Draggable component's onClick event.

imanjra:
Thanks @benjaminpkane for the suggestion! I have updated to just onClick.

benjaminpkane (approved):
LGTM. Thanks for tweak!

PR #2240, opened 2022-11-04
Migration using removed pymongo method 

j053y:
migration 0.6.2. collection_names -> list_collection_names


brimoor (changes requested):
@j053y I believe the `404.html` file that is deleted by this PR needs to exist. 

I assume this wasn't intentional on your part (empty files and git don't get along well). @benjaminpkane can elaborate if needed

j053y:
@brimoor Just noticed that as well, added back in.

brimoor (approved):
LGTM

PR #2241, opened 2022-11-04
Add Field Description and Info to App

ritch:


brimoor (changes requested):
@ritch okay I tried to break the tooltip with this:

```py
import fiftyone as fo
import fiftyone.zoo as foz

field = dataset.get_field("ground_truth.detections.area")
field.description = "Contrary to popular belief, Lorem Ipsum is not simply random text. It has roots in a piece of classical Latin literature from 45 BC, making it over 2000 years old. Richard McClintock, a Latin professor at Hampden-Sydney College in Virginia, looked up one of the more obscure Latin words, consectetur, from a Lorem Ipsum passage, and going through the cites of the word in classical literature, discovered the undoubtable source. Lorem Ipsum comes from sections 1.10.32 and 1.10.33 of 'de Finibus Bonorum et Malorum' (The Extremes of Good and Evil) by Cicero, written in 45 BC. This book is a treatise on the theory of ethics, very popular during the Renaissance. The first line of Lorem Ipsum, 'Lorem ipsum dolor sit amet..', comes from a line in section 1.10.32.\n\nThe standard chunk of Lorem Ipsum used since the 1500s is reproduced below for those interested. Sections 1.10.32 and 1.10.33 from 'de Finibus Bonorum et Malorum' by Cicero are also reproduced in their exact original form, accompanied by English versions from the 1914 translation by H. Rackham."
field.info = {
    "a-really-really-really-long-key-here": {'a': {'b': {'c': {'d': 'e'}}}},
    "int-value": 1,
    "bool-value": True,
    "float-value": 1.2345,
    "url2": "https://github.com/voxel51/fiftyone/commit/250f29cdc3ba64418e811031fabfa5606b9017a0",
    "url3": "https://github.com/voxel51/fiftyone/commit/250f29cdc3ba64418e811031fabfa5606b9017a0",
    "url4": "https://github.com/voxel51/fiftyone/commit/250f29cdc3ba64418e811031fabfa5606b9017a0",
    "url5": "https://github.com/voxel51/fiftyone/commit/250f29cdc3ba64418e811031fabfa5606b9017a0",
    "url6": "https://github.com/voxel51/fiftyone/commit/250f29cdc3ba64418e811031fabfa5606b9017a0",
    "url7": "https://github.com/voxel51/fiftyone/commit/250f29cdc3ba64418e811031fabfa5606b9017a0",
    "url8": "https://github.com/voxel51/fiftyone/commit/250f29cdc3ba64418e811031fabfa5606b9017a0",
}
field.save() 

```

and here's what I saw:

1. Long values need `...` handling to avoid overflow (it's fine if the whole URL doesn't show because you can just click on it to see where it goes)

Screen Shot 2022-11-04 at 10 47 49 PM

2. Two things here: (a) boolean values aren't rendering (should see `True`). (b) looks like the column width logic is being applied based on all keys, but should only be applied to keys that are actually rendered (in this case `a-really-really-long-key` has a non-renderable value, so it's long length shouldn't affect the column width):

Screen Shot 2022-11-04 at 10 50 12 PM

3. I'm still seeing the legacy tooltip too. Can you disable that?

Screen Shot 2022-11-04 at 10 43 46 PM
Screen Shot 2022-11-04 at 10 44 14 PM

ritch:
Fixed the issues mentioned above:

image

I did not see the legacy tooltips in some of the screenshots above but I did see one. I removed two cases in the code, so I think we should be good but I did not confirm those other cases.

brimoor (approved):
LGTM, thanks @ritch!

PR #2243, opened 2022-11-04
Build 404 for SPA intex.html handling

benjaminpkane:
A change in starlette default 404 handling now requires a 404.html to be present in the static directory for the Static file route to resolve on app routes, e.g. `dataset/my-dataset`. Adding a dummy 404.html to the public folder in `@fiftyone/app` accommodates the peculiar behavior.


brimoor (approved):
LGTM

PR #2247, opened 2022-11-05
toggle sidebar entry only on clicking header

imanjra:

Improve sidebar toggle entry to only toggle when header of entry is clicked and prevent toggle on click and drag to avoid unexpected toggle on selection of entry text.  

### Demo
Kapture 2022-11-05 at 18 57 13

Tested visually by interacting with field to ensure entry is toggled only when expected. 

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Sidebar/Entries/RegularEntry.tsx
+++ b/app/packages/core/src/components/Sidebar/Entries/RegularEntry.tsx
@@ -28,6 +28,7 @@ type RegularEntryProps = React.PropsWithChildren<{
   heading: ReactNode;
   left?: boolean;
   onClick?: MouseEventHandler;
+  onHeaderClick?: MouseEventHandler;
   title: string;
   trigger?: (
     event: React.MouseEvent<HTMLDivElement>,
@@ -47,11 +48,15 @@ const RegularEntry = React.forwardRef(
       heading,
       left = false,
       onClick,
+      onHeaderClick,
       title,
       trigger,
     }: RegularEntryProps,
     ref
   ) => {
+    const tolerance = 5;
+    const headerClickStart = useRef<MouseEvent>();
+
     return (
       <Container
         ref={ref}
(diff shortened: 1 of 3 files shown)


app/packages/core/src/components/Sidebar/Entries/RegularEntry.tsx
```diff
@@ -63,7 +68,24 @@ const RegularEntry = React.forwardRef(
...
       >
         <Draggable color={color} entryKey={entryKey} trigger={trigger}>
-          <Header style={{ justifyContent: left ? "left" : "space-between" }}>
+          <Header
+            onMouseDown={(event: MouseEvent) => {
+              headerClickStart.current = event;
+            }}
+            onMouseUp={(event: MouseEvent) => {
+              if (!onHeaderClick) return;
+              const startX = headerClickStart.current.pageX;
```
imanjra, line 77: @brimoor re-added this bit of code to detect drag event.

brimoor (approved):
LGTM, thanks! I defer to @benjaminpkane if he would like to discuss the implementation.

One tiny detail---and I am not sure it is possible to have this and still maintain the nicely responsive click-to-toggle behavior, so take this with a grain of salt---it would be nice if double-clicking to select the field text did not cause the visibility to toggle:

imanjra:
@brimoor I was thinking of preventing the toggle on double click, but from what I've gathered so far, it will add a slight delay to check for and skip a single click in case of double click. I'll investigate further to see if there is good middle ground.

app/packages/core/src/components/Sidebar/Entries/RegularEntry.tsx
benjaminpkane: This is functional, of course, but it's an ambiguous click zone imo.

benjaminpkane:
From experience, there is not a solution where double-click and click both work cleanly in the browser.

PR #2252, opened 2022-11-07
Field desc fixups

ritch:
- Increase the delay before the tooltips appear in the App
- Invert for field tooltip in light mode (lighter background and darker table row background, for consistency with how the tooltip matches the fields sidebar entries in dark mode)
- In light mode, use a lighter color for the placeholder squares that appear in the grid when scrolling down


brimoor (approved):
LGTM, thanks!

imanjra (approved):
lgtm!

PR #2268, opened 2022-11-09
rename saved view methods

kaixi-wang:
Change log
- Rename `views` to `saved_views` for clarity
- Added `save` to related methods for consistency


fiftyone/core/clips.py
```diff
@@ -275,12 +275,12 @@ def reload(self):
...
         # Regenerate the clips dataset
         #
-        # This assumes that calling `load_view()` when the current clips
+        # This assumes that calling `load_saved_view()` when the current clips
         # dataset has been deleted will cause a new one to be generated
         #
 
         self._clips_dataset.delete()
-        _view = self._clips_stage.load_view(self._source_collection)
+        _view = self._clips_stage.load_saved_view(self._source_collection)
```
brimoor, line 283: This should not have been renamed; `ViewStage.load_view()` is something different

fiftyone/core/odm/database.py
```diff
@@ -147,7 +147,7 @@ def establish_db_conn(config):
         _connection_kwargs["port"] = int(established_port)
     if config.database_uri is not None:
         _connection_kwargs["host"] = config.database_uri
-    elif _db_service is None:
+    if _db_service is None:
```
brimoor, line 150: Revert this! It's unrelated to this PR and is also not what we want (would cause a mongo service to start even if user has `database_uri` configured)

fiftyone/core/stages.py
```diff
@@ -76,7 +76,7 @@ def __eq__(self, other):
     @property
     def has_view(self):
         """Whether this stage's output view should be loaded via
-        :meth:`load_view` rather than appending stages to an aggregation
+        :meth:`load_saved_view` rather than appending stages to an aggregation
```
brimoor, line 79: revert all (not saved views)

fiftyone/core/patches.py
```diff
@@ -289,12 +289,12 @@ def reload(self):
         #
         # Regenerate the patches dataset
         #
-        # This assumes that calling `load_view()` when the current patches
+        # This assumes that calling `load_saved_view()` when the current patches
```
brimoor, line 292: revert all (not saved views)

fiftyone/core/video.py
```diff
@@ -259,12 +259,12 @@ def reload(self):
         #
         # Regenerate the frames dataset
         #
-        # This assumes that calling `load_view()` when the current patches
+        # This assumes that calling `load_saved_view()` when the current patches
```
brimoor, line 262: revert all (not saved views)

fiftyone/core/view.py
```diff
@@ -1379,8 +1379,8 @@ def _slice(self, s):
     def _add_view_stage(self, stage):
         stage.validate(self)
 
-        if stage.has_view:
-            view = stage.load_view(self)
+        if stage.has_saved_view:
```
brimoor, line 1382: revert (not saved views)

docs/source/user_guide/using_views.rst
```diff
@@ -129,7 +129,7 @@ your dataset under a name of your choice:
     dataset.save_view("cats-view", cats_view)
 
 Then you can conveniently use
-:meth:`load_view() <fiftyone.core.dataset.Dataset.load_view>` to load the view
+:meth:`load_saved_view() <fiftyone.core.dataset.Dataset.load_saved_view>` to load the view
```
brimoor, line 132: nit: style guide says 79 character line limit

fiftyone/core/dataset.py
```diff
@@ -3193,8 +3193,9 @@ def delete_view(self, name):
         view_doc = self._get_view_doc(name, pop=True)
         self._doc.save()
 
-    def delete_views(self):
+    def delete_saved_views(self):
         """Deletes all saved views from this dataset."""
+        print("Deleting saved views: ", self.list_saved_views())
```
brimoor, line 3198: remove print

fiftyone/core/dataset.py
```diff
@@ -3193,8 +3193,9 @@ def delete_view(self, name):
...
         self._doc.saved_views = []
         self._doc.save()
```
brimoor, line 3201: You should rename private saved view-related methods here too, including:
- `_get_view_doc`
- `_views`
- `_reorder_views`
- `_validate_view_name`

fiftyone/utils/data/exporters.py
```diff
@@ -1708,7 +1708,7 @@ def log_collection(self, sample_collection):
         # Exporting the information below only makes sense when exporting an
         # entire dataset
 
-        if dataset.has_views:
+        if dataset.has_saved_views:
```
brimoor, line 1711: Should rename `"views"` to `"saved_views"` in these importer/exporter JSON too, for consistency

brimoor (changes requested):

fiftyone/core/dataset.py
kaixi-wang: oh I meant to ask, should we put some kind of warning? It seems too easy to accidentally delete them all.

brimoor: That could be a design choice we could make, but it's not used anywhere else right now so I wouldn't add it here:

```py
dataset.delete_samples(...)
dataset.delete_evaluations()
dataset.delete()
```

fiftyone/utils/data/exporters.py
kaixi-wang: should `export_views` then be `export_saved_views` and likewise for imports or do you think it's fine without prefixing with `saved`

brimoor: Hmm, well we're in it this far, so I guess so right!?!

brimoor: fwiw I did go through this thought process when I initially added this stuff, and at that time I chose my personal preference of going for as few keystrokes as possible (so `view` not `saved_view`). But, if our stance is that users may get confused seeing just "view" since this is not referring to any old view they may have created, then I would go with principal #2 instead: be super consistent on public APIs

kaixi-wang: yeah I think consistency is best here, especially since most people will have autocomplete to minimize the keystrokes anyways

fiftyone/core/dataset.py
kaixi-wang: removed `_reorder_views` per the discussion earlier

fiftyone/core/dataset.py
```diff
@@ -3076,11 +3076,13 @@ def save_view(
...
             raise ValueError("Cannot save view into a different dataset")
 
-        url_name = self._validate_view_name(name, overwrite=overwrite)
+        view._set_name(name)
+        url_name = self._validate_saved_view_name(name, overwrite=overwrite)
 
         now = datetime.utcnow()
 
         view_doc = foo.ViewDocument(
+            id=ObjectId(),
```
kaixi-wang, line 3085: @brimoor I added an id field to the view document so that the front end has a unique identifier for saved_view components

brimoor (approved):
LGTM

PR #2276, opened 2022-11-09
App patches

benjaminpkane:
Patches server aggragations so all field types can be shown in the sidebar, regardless of filtering. Also excludes gs:// and s3:// paths from URL detections and fixes link font coloring (removes blue highlighting, a theme regression)


app/packages/core/src/components/Sidebar/Entries/FilterablePathEntry.tsx
```diff
@@ -305,11 +305,11 @@ const FilterableEntry = React.memo(
               field={field}
               color={color}
               expandedPath={expandedPath}
-              template={({ label, hoverHanlders, hoverTarget, container }) => (
+              template={({ hoverHanlders, hoverTarget, container }) => (
```
brimoor, line 308: nit: typo

app/packages/utilities/src/index.ts
```diff
@@ -354,6 +354,15 @@ const isURL = (() => {
   const nonLocalhostDomainRE = /^[^\s\.]+\.\S{2,}$/;
 
   return (string) => {
+    return true;
```
brimoor, line 357: I think it would be best to only treat `^http[s]://` and perhaps `^localhost:`and `^0.0.0.0:` as URLs, since users could provide all of the following:

```
s3://bucket/path
gs://bucket/path
[custom-minio-alias]://bucket/path
```

brimoor: I defer to you on how to finalize, but avoiding linking `[custom-minio-alias]://bucket/path` while still linking other things seems ill-defined

brimoor: If you don't want to think about it right now, also fine punting. The currently implemented behavior covers the 90% case

brimoor (approved):
LGTM

app/packages/utilities/src/index.ts
benjaminpkane: Apologies, I overlooked. Will address tomorrow

PR #2283, opened 2022-11-10
[HOLD] saved views UI and API connection 

manivoxel51:

- This is the UI for dataset views selection.
- works in dark/light mode
- Next step is mutation and fixing the color selection

https://user-images.githubusercontent.com/109545780/200677282-a9f4b632-ae3b-4608-9656-3c2920a6bcb0.mov

Create/Edit modal
Screen Shot 2022-11-10 at 5 43 40 PM

When view param doesn't match any existing views

https://user-images.githubusercontent.com/109545780/203185028-871d66dc-ed8f-4a21-a183-9f48e5e588f5.mov

When edit/create saved view name already exists

Screen Shot 2022-11-21 at 3 51 24 PM

(Details)

        notes for FiftyOne users.

-   [ ] Other


kaixi-wang (approved):
lgtm

brimoor:
@manivoxel51 for OSS I think we should remove the `Everyone with ...` text from the view creation UI

manivoxel51:
@benjaminpkane let me know if you see anything that can be improved

app/packages/state/src/recoil/selectors.ts
```diff
@@ -40,6 +40,19 @@ export const mediaType = selector({
   },
 });
 
+export const savedViewsSelector = selector({
+  key: "datasetViews",
+  get: ({ get }) => {
+    console.log("dataset", get(atoms.dataset));
+    const savedViews = get(atoms.dataset)?.savedViews;
```
benjaminpkane, line 47: Can we add `savedViews` to the dataset type in "../types.ts"? The selector should have implicit typing then.

app/packages/state/src/hooks/useHover.ts
```diff
@@ -0,0 +1,26 @@
+import { useRef, useState, useEffect } from "react";
+
+function useHover() {
```
benjaminpkane, line 3: I would explicitly type what is returned. Right now the type is a an array and not a tuple.

app/packages/state/src/hooks/useHover.ts
```diff
@@ -0,0 +1,26 @@
...
+      if (node) {
+        node.addEventListener("mouseover", handleMouseOver);
+        node.addEventListener("mouseout", handleMouseOut);
+        return () => {
+          node.removeEventListener("mouseover", handleMouseOver);
+          node.removeEventListener("mouseout", handleMouseOut);
+        };
+      }
+    },
+    [ref.current] // Recall only if ref changes
```
benjaminpkane, line 21: This dep is probabaly unnecessary. Refs don't trigger updates

app/packages/components/src/components/Selection/Option.tsx
```diff
@@ -0,0 +1,126 @@
+import React from "react";
+
+import styled from "styled-components";
+import { useHover } from "@fiftyone/state";
+import { Edit, Check } from "@mui/icons-material";
+import { useTheme } from "@fiftyone/components";
+
+import Tooltip from "@mui/joy/Tooltip";
+
+const Box = styled.div`
```
benjaminpkane, line 10: Just thinking out loud based on the discussions we have been having, styled components is used heavily in the App, but I actually prefer CSS modules because they have code hints etc. Styled components are harder to read.

My point is, once we decide how want to style, we should stick with that and slowly remove everything else. Modules, styled components, or something else.

app/packages/state/src/hooks/useHover.ts
benjaminpkane, line 21: It might be safer to have this hook return `onMouseOver` and`onMouseOut` event handlers rather the ref. Just because the ref itself won't trigger updates

fiftyone/server/query.py
```diff
@@ -190,7 +190,7 @@ def modifier(doc: dict) -> dict:
         doc["frame_fields"] = _flatten_fields([], doc.get("frame_fields", []))
         doc["brain_methods"] = list(doc.get("brain_methods", {}).values())
         doc["evaluations"] = list(doc.get("evaluations", {}).values())
-        doc["saved_views"] = doc.get("saved_views", [])
+        doc["saved_views"] = doc.get("saved_views", doc.get("views", []))
```
benjaminpkane, line 193: I think you will need to update the `schema.graphql` and add the new field to the dataset queries. There are two, right now. Probably should only be one.
* https://github.com/voxel51/fiftyone/blob/develop/app/packages/core/src/Dataset.ts
* https://github.com/voxel51/fiftyone/blob/develop/app/packages/relay/src/mutations/setView.ts

benjaminpkane:
Looks really good! Left a few comments

fiftyone/server/query.py
```diff
@@ -356,11 +356,19 @@ def saved_view(
...
         if ds.has_views & ds.has_view(view_name):
-            for view in ds._doc.saved_views:
-                if view.name == view_name:
+            for the_view_name in ds.list_views:
+                if the_view_name == view_name:
                     return view
         return
 
+    @gql.field
+    def saved_views(self, dataset_name: str) -> t.List[SavedView]:
```
kaixi-wang, line 365: Is this necessary? The list of saved views is already added to the Dataset graphql object, so you should be able to just directly read it from the query result already

manivoxel51: I separated it so I can create a refetchable fragments and refresh the views separate from reloading the entire dataset.

app/packages/state/src/recoil/selectors.ts
manivoxel51: cool, if you mean to add a SavedView type in ./types and have the selector be typed for its return type - done!

app/packages/components/src/components/Selection/Option.tsx
manivoxel51: agreed on a consensus on styling one way. we had a short discussion on this on slack. since we are moving to shared component lib with mui, both styled components and css modules will get changed / removed. I personally like styled components since you have access to themes, props passing is possible, and I like the inheritance approach to build reusable ones.

fiftyone/server/query.py
manivoxel51: good catch. line 193 should be reverted back to `doc["saved_views"] = doc.get("saved_views", [])`. 

we have views in db initially - then changed to saved_views. I was testing around

manivoxel51:
@kaixi-wang I'll merge this to your branch and work on the following separately since this PR is too large and hard to maintain. also amazing work on fixing the build and simplifying couple of things in the frontend 🍨

PR #2288, opened 2022-11-10
Patch v0.16.6 compatibility

brimoor:
v0.18 purports to be backwards-compatible with v0.16.6 datasets. However, this is not currently the case because it does not appropriately handle missing `app_config`s.

```shell
fiftyone migrate -v 0.16.6 -n quickstart
export FIFTYONE_DATABASE_ADMIN=false
```

```py
import fiftyone as fo

db = foo.get_db_conn()

# Scenario 1: entire `app_config` is missing

assert "app_config" not in db.datasets.find_one({"name": "quickstart"})
assert dataset.version == "0.16.6"
assert dataset.app_config is not None  # previously failed; now doesn't

session = fo.launch_app(dataset)  # TODO: fix this

# Scenario 2: any part(s) of `app_config` are missing

dataset.app_config.grid_media_field = "foo"  # force mongoengine to detect a change
dataset.app_config.grid_media_field = "filepath"
dataset.save()

print(db.datasets.find_one({"name": "quickstart"})["app_config"])  # None
# {'grid_media_field': 'filepath'}

session = fo.launch_app(dataset)  # TODO: fix this
```

```shell
unset FIFTYONE_DATABASE_ADMIN
```

### DIFF
--- a/app/packages/core/src/__generated__/DatasetQuery.graphql.ts
+++ b/app/packages/core/src/__generated__/DatasetQuery.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<a2f8d4a86d8e3c6654edba23aa04eab9>>
+ * @generated SignedSource<<ada5cd8e34cacf34f90035e5f89a5932>>
  * @lightSyntaxTransform
  * @nogrep
  */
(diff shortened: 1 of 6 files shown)


brimoor:
@benjaminpkane can you help resolve the App-side TODOs here?

benjaminpkane:
Media fields are now optional. The first to do was already working for me. LMK if I missed this something.

brimoor:
LGTM, thanks @benjaminpkane! Can you approve? 😄

benjaminpkane (approved):
Pardon the delay. LGTM

PR #2289, opened 2022-11-10
App patches

benjaminpkane:
Fixes state updates when the router state is undefined and modal expansion when in a patches view of a grouped dataset.


brimoor (approved):
LGTM

PR #2295, opened 2022-11-11
Updating issue templates

brimoor:
pre-commit hooks are great and all, but in this case, it is better to ignore so that the issue templates are more readable in the github issue creation UI


kaixi-wang (approved):
lgtm

PR #2297, opened 2022-11-13
Update app.rst

oguz-hanoglu:

The explanation for 'all' option sidebar mode was not clear probably due to a typo.

### DIFF
--- a/docs/source/user_guide/app.rst
+++ b/docs/source/user_guide/app.rst
@@ -302,8 +302,7 @@ samples or fields, this may involve substantial computation.
 
 Therefore, the App supports three sidebar modes that you can choose between:
 
--   `all`: always compute counts for all fields and stats for all fields whose
-    filter tray is expanded
+-   `all`: always compute counts for all fields and stats
 -   `fast`: only compute counts and stats for fields whose filter tray is
     expanded
 -   `best` (*default*): automatically choose between `all` and `fast` mode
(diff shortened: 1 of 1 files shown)


docs/source/user_guide/app.rst
```diff
@@ -302,8 +302,7 @@ samples or fields, this may involve substantial computation.
 
 Therefore, the App supports three sidebar modes that you can choose between:
 
--   `all`: always compute counts for all fields and stats for all fields whose
-    filter tray is expanded
+-   `all`: always compute counts for all fields and stats
```
brimoor, line 305: What the previous version of the docs were trying to convey is that, in `all` mode:
- the count values will be shown for all fields that are visible in the sidebar
- In addition, the data inside the "filter tray" (eg the `id` and `label` widgets you see after clicking the down caret in the image below) will be computed whenever the sidebar loads

Do you have any suggestion for how to edit the docs in a different way to make that more clear? In your proposed version I think it may not be clear what "stats" means without some additional context

Screen Shot 2022-11-13 at 1 58 22 PM

brimoor (changes requested):

docs/source/user_guide/app.rst
oguz-hanoglu: OK. Thanks for detailed clarification.

After your explanation, I agree that my proposed version needs updating.

As an outsider reading the document(before your explanation), I did not know what "stats" is. Referring to it as 'stats' may be confusing when there is also 'stats' in the 'Statistics tabs' in the App.

My confusion led me think that there is a typo considering the redundant `for all fields` in 'all' bullet.

So, my new suggestion is:
- `all`: always compute counts _(for all fields -> this part can be removed)_ and stats _(-> this can be renamed)_ for all fields whose filter tray is expanded

brimoor: Sounds good 👍

brimoor (approved):
Thanks for your help on improving these docs! I'll add a quick edit to resolve the outstanding thread after merging

PR #2299, opened 2022-11-14
hide bookmark icon when no filters are applied

### LINKED ISSUE #2281: [FR] Hide bookmark icon when no filters are applied
The bookmark icon converts the current filters (the "extended view") into view stages in the view bar. Looks like the icon is currently always present (and furthermore is always orange) even when there are no filters currently applied. But if there are no filters applied, bookmarking would not do anything.

Therefore, to be consistent with our design, we should adopt one of the following when there are no filters applied:
1. Hide the bookmark icon completely
2. Show the icon with a non-orange background

I vote for 1. Cleanest.

Screen Shot 2022-11-09 at 7 15 05 PM

sashankaryal:

Fixes #2281

### Changelog
- Hide bookmark icon when no filters are applied
- Render anchor element instead of div in `ItemAction` when there's a href

On a dev build

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Actions/Common.tsx
+++ b/app/packages/core/src/components/Actions/Common.tsx
@@ -4,6 +4,8 @@ import { Launch } from "@mui/icons-material";
 import { useHighlightHover } from "./utils";
 import { ItemAction } from "./ItemAction";
 import { useExternalLink } from "@fiftyone/utilities";
+import ExternalLink from "@fiftyone/components/src/components/ExternalLink";
+import { useTheme } from "@fiftyone/components";
 
 type ActionOptionProps = {
   onClick?: (event?: Event) => void;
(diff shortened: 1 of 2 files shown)


app/packages/core/src/components/Actions/Common.tsx
```diff
@@ -35,6 +35,7 @@ export const ActionOption = React.memo(
     return (
       <ItemAction
         title={title ? title : text}
+        as={href ? "a" : undefined}
```
benjaminpkane, line 38: I'm not sure why I added an `href` prop to `ItemAction`, but it can probably  be removed. It's not used anywhere and `href` links aren't safe in the App because we distribute an Electron desktop app version and `a` links don't work there.

External links can be created with this [component](https://github.com/voxel51/fiftyone/blob/develop/app/packages/components/src/components/ExternalLink/ExternalLink.tsx).

app/packages/core/src/components/Actions/Common.tsx
```diff
@@ -38,12 +41,16 @@ export const ActionOption = React.memo(
...
         <span style={href ? { textDecoration: "underline" } : {}}>
-          {text}
-          {href && <Launch style={svgStyles} />}
+          {href ? (
+            <ExternalLink style={{ color: theme.text.primary }} href={href}>
+              {text}
+              <Launch style={svgStyles} />
+            </ExternalLink>
+          ) : (
+            `${text}`
```
sashankaryal, line 52: nit: superfluous template literal, will remove in some other pr

benjaminpkane (approved):
LGTM!

PR #2300, opened 2022-11-14
Storybook installation

ShawnOakley:

Adds storybook for use with base components.  

NOTE: Debatable whether we want examples included.  Typically I would say we shouldn't include them, but given this is an open-source app, we may want to include storybook examples for reference.

Tested by showing that storybook successfully renders.

## Release Notes

Installs storybook within app/component directory

-   [x ] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

-   [ X] App: FiftyOne application changes
-   [ X] Build: Build and test infrastructure changes
-   [ ] Other


ritch (changes requested):
I think we need to clean out all the storybook generated cruft before merging.

If we can do a story example of something simple like the dataset's typeahead or button that would be production code, then I think we can merge.

app/packages/components/src/components/Button/Button.tsx
```diff
@@ -3,12 +3,14 @@ import classnames from "classnames";
...
->> = ({ children, className, ...rest }) => {
+const Button: React.FC<
+  React.DetailedHTMLProps<
+    React.ButtonHTMLAttributes<HTMLButtonElement>,
+    HTMLButtonElement
+  >
+> = ({ children, className, ...rest }) => {
   const classNames = [button, className];
-
+  console.log("classNames", classNames);
```
ritch, line 13: Stray log?

app/packages/components/src/components/ThemeProvider/index.tsx
```diff
@@ -120,18 +120,19 @@ let theme = extendJoyTheme({
 });
 
 export const useTheme = () => {
-  return theme.colorSchemes[useRecoilValue(fos.theme)].palette;
+  return theme.colorSchemes["light"].palette;
```
ritch, line 123: Is this an intended change? This looks like it will break theme selection.

app/packages/components/src/components/ThemeProvider/index.tsx
```diff
@@ -120,18 +120,19 @@ let theme = extendJoyTheme({
...
 };
 
 const ThemeProvider: React.FC<
   React.PropsWithChildren<{ customTheme?: Theme }>
 > = ({ children, customTheme }) => {
-  if (customTheme) theme = customTheme;
-  const current = useRecoilValue(fos.theme);
+  // if (customTheme) theme = customTheme;
+  console.log("++++++++");
```
ritch, line 130: Stray log?

ritch (changes requested):
Needs some minor changes to avoid breaking themes in develop. Otherwise LGTM.

app/packages/components/src/components/ThemeProvider/index.tsx
ShawnOakley: Fixed

app/packages/components/src/components/ThemeProvider/index.tsx
ShawnOakley: Theme selection isn't working as intended since it's tightly coupled to recoil.  There was back and forth on how to successfully implement this, but I think the recoil integration should be a separate ticket.

ritch (approved):
LGTM 👍

Issue #2286, opened 2022-11-10
[BUG] Remove excluded media fields from App selector 


brimoor:
The App supports configuring [multiple media fields](https://voxel51.com/docs/fiftyone/user_guide/app.html#multiple-media-fields). However, the user can exclude non-default fields from their schema in various ways.

The App currently includes excluded media fields in the `Media field` selector under the settings menu. However, it would be more friendly if it would omit entries from that list that have been excluded from the current view's schema:

```py
import fiftyone as fo
import fiftyone.utils.image as foui
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")

foui.transform_images(
    dataset,
    size=(-1, 64),
    output_field="thumbnail_path",
    output_dir="/tmp/thumbnails",
)

dataset.app_config.media_fields = ["filepath", "thumbnail_path"]
dataset.app_config.grid_media_field = "thumbnail_path"
dataset.save()

session = fo.launch_app(dataset)

# Both of these views try and fail to access thumbnail_path
session.view = dataset.select_fields()
session.view = dataset.to_patches("ground_truth")
```

<img width="920" alt="Screen Shot 2022-11-10 at 1 11 50 AM" src="https://user-images.githubusercontent.com/25985824/201017437-9bdbe67e-7a7c-4563-a3c0-1c9ed2d21f83.png">

lanzhenw:
@brimoor Took a look at the issue, it looks like recoil is not getting the updated `mediaFields` from appConfig after the view changes. 

```
const datasetAppConfig = selector<State.DatasetAppConfig>({
  key: "datasetAppConfig",
  get: ({ get }) => get(dataset).appConfig,
});
```

I can overwrite the default mediaFields based on the view's field. Are `to-patches` and `select_fields` the only methods that can cause the view to exclude default media fields? 
Is it okay to assume the conditions are when the view has `select_field` or `field_names` in `kwargs` ?

These are the console of the view for the two bug cases:
<img width="417" alt="Screen Shot 2022-11-14 at 1 10 37 AM" src="https://user-images.githubusercontent.com/17770824/201599840-573ec6d3-5821-4c2a-a0d2-522e61b50127.png">
<img width="312" alt="Screen Shot 2022-11-14 at 9 02 07 AM" src="https://user-images.githubusercontent.com/17770824/201693195-218b85bd-85f4-4c66-a337-192c791038f8.png">

lanzhenw:
[...]

PR #2303, opened 2022-11-14
Add a filter that removes the mediaField item that does not exists in the selected fields

lanzhenw:
fixes [#2286](https://github.com/voxel51/fiftyone/issues/2286)

Added a filter that removes the mediaField item if it does not exists in the selected fields

Screen Shot 2022-11-14 at 9 52 23 AM

        notes for FiftyOne users.

-   [ ] Other


app/packages/core/src/components/Actions/Options.tsx
```diff
@@ -191,7 +191,6 @@ const MediaFields = ({ modal }) => {
   const [selectedMediaField, setSelectedMediaField] = useRecoilState(
     fos.selectedMediaField(modal)
   );
-
```
benjaminpkane, line 194: Can this be omitted?

lanzhenw: We need it, because we need the `setSelectedMediaField` method to update the `selectedMediaField` state on the modal.

benjaminpkane: I was only referring to line removal. But it looks like good cleanup. Disregard!

benjaminpkane (approved):
LGTM1

PR #2309, opened 2022-11-15
Increase selection modal width to avoid line wrap

### LINKED ISSUE #2256: [BUG] Increase selection modal width to avoid line wrap
`Current sample` line wrap = 👎 

Screen Shot 2022-11-08 at 11 01 20 AM

lanzhenw:
Fixes #2256 
The selection modal would be wider if there is longer text
Screen Shot 2022-11-15 at 10 02 47 AM
Screen Shot 2022-11-15 at 10 02 38 AM


benjaminpkane:
This is a reasonable solution. But maybe we can do CSS only solution? Adding `white-space: nowrap;` to this [component](https://github.com/voxel51/fiftyone/blob/f2d4e633bed5d8d43c635b320d7441a54277410a/app/packages/core/src/components/Actions/ItemAction.tsx#L4) should resolve the issue and avoid extra padding.

benjaminpkane (approved):
LGTM!

PR #2310, opened 2022-11-15
fiftyone vs pandas comparison notebook

jacobmarks:

(Details)

        notes for FiftyOne users.

-   [ ] Other


brimoor (approved):
LGTM, very thorough!

I made a quick editorial pass which you can see here: https://github.com/voxel51/fiftyone/pull/2310/commits/31452555406eae9f7db6e7687d448b6c3f4bc8f9. Feel free to revert anything if you disagree with it, but here's what I did:

Content
- Only one small thing: removed reference to `sample.field_names` because it doesn't actually return field names that are non-None on the individual sample, as implied by the text. It is always returns same thing as `get_field_schema()`

Formatting
- Added an entry at the bottom of `docs/source/tutorials/index.rst` so that this tutorial appears in the left sidebar of the docs, like other tutorials
- Tweaked the overview/card of the tutorial to emphasize how pandas expertise should convert to FiftyOne expertise
- Use `Sentence case` rather than `Title Case` in section headings, for consistency with other tutorials
- Updated some spacing in the code samples like `key=value` rather than `key = value` for function arguments, to match what our auto-formatter would have done if it were able to process `.ipynb` files

brimoor:
@jacobmarks just pointing out that you committed the image in two places in this PR, although its only referenced in one place. Intentional?

For context, once an image is committed to a repo, it is forever included in `git clone`, even if it is deleted (unless one specifically goes in and purges the image from the git history), so we try to be especially careful to avoid committing unnecessary images to the repo.

btw this repo bloat is one reason why we may want to split the docs into a separate repo. But haven't executed on that yet because I like having the docs in the same place as the code to encourage them to stay in sync.

PR #2318, opened 2022-11-16
lidar->camera conversion and 3d corners

jacobmarks:

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/kitti.py
+++ b/fiftyone/utils/kitti.py
@@ -983,13 +983,21 @@ def _roty(t):
     return np.array([[c, 0, s], [0, 1, 0], [-s, 0, c]])
 
 
+def _get_corners3d(h, w, l, R, t):
+    # Construct (x, y, z) coordinates of 3d box corners
+    corners3d = np.array(
+        [
+            [l / 2, l / 2, -l / 2, -l / 2, l / 2, l / 2, -l / 2, -l / 2],
+            [0, 0, 0, 0, -h, -h, -h, -h],
+            [w / 2, -w / 2, -w / 2, w / 2, w / 2, -w / 2, -w / 2, w / 2],
+        ]
+    )
+    corners3d = R @ corners3d + t[:, np.newaxis]
+
+    return corners3d
+
+
 def _proj_3d_to_right_camera(detections3d, calib, frame_size):
-    # Velodyne -> right camera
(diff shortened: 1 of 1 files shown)


fiftyone/utils/kitti.py
```diff
@@ -1033,6 +1037,29 @@ def _proj_3d_to_right_camera(detections3d, calib, frame_size):
...
+    p = np.array([x, y, z, 1])
+    return tuple((R0 @ V2C @ p)[0:3])
+
+
+def convert_box_lidar_to_camera(boxes, V2C, R0):
+    # (N, 7) -> (N, 7) x,y,z,h,w,l,r
+    ret = []
+    for box in boxes:
+        x, y, z, h, w, l, rz = box
+        (x, y, z), h, w, l, ry = (
```
brimoor, line 1051: Can we just write it like this for simplicity?

```py
x, y, z = _convert_point_lidar_to_camera((x, y, z), V2C, R0) 
ry = -rz - np.pi / 2
```

fiftyone/utils/kitti.py
```diff
@@ -1033,6 +1037,29 @@ def _proj_3d_to_right_camera(detections3d, calib, frame_size):
     return detections2d
 
 
+def _convert_point_lidar_to_camera(xyz, V2C, R0):
+    x, y, z = xyz
+    p = np.array([x, y, z, 1])
+    return tuple((R0 @ V2C @ p)[0:3])
+
+
+def convert_box_lidar_to_camera(boxes, V2C, R0):
```
brimoor, line 1046: Is this method used in a blog post? It's not used in this PR.

The reason I bring this up is that I prefer that all public functions have full docstrings, but this one feels out of place because the rest of the module doesn't provide any public function to access these required inputs.

fwiw, this could be written as a single line in whatever place you're using this operation:

```py
# Convert point LIDAR to camera coordinates
x, y, z = (R0 @ V2C @ np.array([x, y, z, 1]))[0:3]
```

brimoor (changes requested):
Can you provide some context on where this PR is used?

jacobmarks:
Hi @brimoor , these methods will be used in the 3d computer vision tutorial. When I finish the complete draft of the tutorial, we can revisit and see if it just makes sense for me to define in the context of the tutorial and not add to the code base.

jacobmarks:
I don't actually think this is necessary for my tutorial. Will just write the needed code in the jupyter notebook

Issue #2278, opened 2022-11-09
[FR] Compute sample and label tags counts separately in tag UI


brimoor:
It would be more performant to compute sample and label stats independently, only when their tab is loaded.

Currently both are computed even if the labels tab is never accessed, but (1) sample tags are more commonly used, and (2) label tags can be significantly more expensive to compute.

Also, for consistency with other loading states in the App right now, how about just animating the loading dots `...` rather than the spinner???

![2022-11-09 18 12 27](https://user-images.githubusercontent.com/25985824/200963440-5d3b0db5-6365-40d0-af41-c6bc19fb0c49.gif)

brimoor:
Also, for consistency with other loading states in the App right now, how about just animating the loading dots `...` rather than the spinner?

PR #2321, opened 2022-11-17
Use consistent animated Loading dots throughout the app

### LINKED ISSUE #2280: [FR] Show animated dots while sample and field counts are loading
For consistency with other stats loading states in the App, I think the static `...` that appears while field counts are loading in the App should be animated, similar to how `Loading sample/label tags [...]` is.

This isn't the most helpful GIF because the stats load fast, but, for larger datasets, stats could take longer and the user may wonder whether the App is "doing anything" or not:

Same goes for the sample count above the grid:

Screen Shot 2022-11-09 at 7 23 40 PM

lanzhenw:
Fixes  #2280  #2278

Show animated dots (...) for loading... 
and use consistent loading animation (... instead of circular spinner)

https://user-images.githubusercontent.com/17770824/204308330-fbb6f9f5-94ad-4b56-91d8-24035afb0068.mov


app/packages/core/src/components/Common/CountSubcount.tsx
```diff
@@ -23,7 +24,10 @@ const EntryCounts = ({
 
   if (typeof subcount !== "number") {
     return (
-      <span style={{ whiteSpace: "nowrap" }}>... {count.toLocaleString()}</span>
+      <span style={{ whiteSpace: "nowrap" }}>
+        {" "}
```
benjaminpkane, line 28: Can this be removed?

brimoor (changes requested):
Can we standardize on the stationary loading dots used in `Loading...`?

I don't like the new style in the sidebar where the `...` physically moves left and right. My goal is to provide the most visually *minimal* loading state that conveys that something is happening if the user really wants to know without drawing attention to the loading state. The already loaded content should draw the most focus.

The `Loading...` dots are very cramped though, due to Palanquin spacing. Increasing the spacing between the dots in an idiomatic way could be prudent.

lanzhenw:
https://user-images.githubusercontent.com/17770824/204937561-8487711a-e2e1-42a8-8287-0546701cbc9c.mov

lanzhenw:
Revert loading... back to original ...
https://user-images.githubusercontent.com/17770824/205159482-7767fe4b-41cc-478c-a682-884aa2673dea.mov

lanzhenw:
https://user-images.githubusercontent.com/17770824/205366863-2a50ef7e-ea52-4886-a4d2-c21ef826acff.mov

brimoor:
@lanzhenw here's a couple things I'm seeing:
- In the tagging UI, the font color should be the same secondary color it was prior to this PR. It is currently primary color (black in dark mode, rather than gray)
- The dots aren't very visible in light mode. Click the sun/moon icon in the upper right to see this. I think it is best to just revert to the simple 100% opaque animation that existed prior to this PR.

In other words, the scope of this PR should just be:
- Use animated loading dots for field counts in the sidebar (rather than static dots)
- Use dots instead of spinner in tagging UI

lanzhenw:
https://user-images.githubusercontent.com/17770824/205695263-a56cfe8f-37cc-4dc0-98a5-0651fa0c821d.mov




@brimoor Just updated the animation to properly inherit the primary color.

app/packages/components/src/components/Loading/LoadingDots.module.css
```diff
@@ -0,0 +1,23 @@
...
+  margin-right: 2px;
+}
+.loading::after {
+  display: inline-table;
+  white-space: pre;
+  content: "\A.\A..\A...";
+  animation: spin 1s steps(4) infinite;
+  margin-left: 1px;
+  margin-right: 2px;
+  letter-spacing: 0.1em;
```
brimoor, line 16: FYI- I used letter spacing to tweak the spacing of the dots a bit

brimoor (approved):
@lanzhenw LGTM, but please do update the text color in the tagging UI to be the secondary gray (before), rather than primary (after):

Before:
Screen Shot 2022-12-06 at 8 57 30 AM

After:
Screen Shot 2022-12-05 at 10 58 08 PM

app/packages/components/src/components/Loading/LoadingDots.module.css
lanzhenw: Thanks @brimoor

lanzhenw: Just updated the text color

https://user-images.githubusercontent.com/17770824/205979207-d359c95f-fb4b-4100-993d-639d2f0619ff.mov

PR #2327, opened 2022-11-18
Fix dataset routing

benjaminpkane:
`@fiftyone/dataset` has no router state, so the `useSetView` hook should only update a router state if it is present.


imanjra (approved):
lgtm

PR #2331, opened 2022-11-18
App unit tests and coverage reporting

benjaminpkane:
This work adds the Vitest testing framework and Recoil and Recoil Relay mocks for unit testing recoil selectors / data flow.

I've added  a **Testing** section to the App `README.md` which summarizes what I have set up, including a Codecov coverage reporting integration.

@brimoor @ritch  if you agree, we could add the `test` job from the `test-app` workflow as a required check for PRs.


app/vite.config.ts
```diff
@@ -0,0 +1,15 @@
...
+import relay from "vite-plugin-relay";
+
+export default defineConfig({
+  test: {
+    environment: "jsdom",
+    coverage: {
+      reporter: ["json"],
+      reportsDirectory: "./coverage",
+    },
+    exclude: [...configDefaults.exclude, "packages/plugins/**"],
```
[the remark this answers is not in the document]
ritch: I wonder if this is due to the one test in plugins? I'll try to convert that to a vitest compatible test.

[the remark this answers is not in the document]
benjaminpkane: Ok, cool. Yes, that's why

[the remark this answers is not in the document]
benjaminpkane: Resolved!

app/__mocks__/recoil.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+}
+
+export function selector<T extends unknown>(
+  options: Parameters<typeof recoil.selector<T>>[0]
+): { (): T; key: string } {
+  function resolver() {
+    return options.get({
+      get: getValue,
+      getCallback: (callback) => {
```
ritch, line 66: IIUC this will allow for testing selectors as they will resolve through this mock impl of recoil. Is there any way to know for sure that this implementation behaves like the real recoil?

.github/workflows/test-app.yml
```diff
@@ -0,0 +1,36 @@
+name: Test App
```
ritch, line 1: Shouldn't this be something like `App Unit Tests`?

ritch:
> @brimoor @ritch if you agree, we could add the test job from the test-app workflow as a required check for PRs.

100% agree

app/__mocks__/recoil.ts
benjaminpkane: The implementation is not correct, but I don't think it would be hard to implement. I've left it as a todo. Only `Snapshot` needs to be mocked to return mock values.

.github/workflows/test-app.yml
benjaminpkane: Sounds good

benjaminpkane:
@brimoor can you add `test-app` as a required check for PRs?

benjaminpkane:
Ok, I am calling complete. `all-tests` is still the only check. Workflows are much faster now, python and app coverage are both reported to codecov. Tests for app and python are conditionally run only if their respective source code changes. Dependency caching was improved, as well.

Installing python requirements still takes a couple minutes for some reason. But now all workflows should run in under 10 minutes now.

ritch (approved):

PR #2338, opened 2022-11-21
Resolved App React Spring error

### LINKED ISSUE #2330: [BUG] (App development) React Spring error
An unhandled error from React Spring, which breaks some development modes, is being thrown by one or more components with an incorrect animation value.

```
Unhandled Runtime Error
Error: Cannot animate between AnimatedValue and AnimatedString, as the "to" prop suggests
````

benjaminpkane:
Resolves #2330


ShawnOakley (approved):
LGTM

PR #2346, opened 2022-11-23
Saved views cleanup 

manivoxel51:

- cleanup / refactor
- add support for embed

(Details)

        notes for FiftyOne users.

-   [ ] Other


manivoxel51:
@kaixi-wang - let's get everything into a single branch and rebase the latest into that branch as we discussed with the team yesterday. That will help us run and test the embedded app easier hopefully. We can address bugs when we have more eyes on it.

app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
```diff
@@ -82,107 +72,107 @@ export default function ViewSelection(props: Props) {
...
+    [viewOptions, viewSearch]
+  );
 
-  useEffect(() => {
-    if (isEmptyView) {
-      setSelected(viewOptions[0]);
-      setSavedViewParam(null);
-    }
-  }, [loadedView]);
+  const loadedView = useRecoilValue<fos.State.Stage[]>(fos.view);
```
kaixi-wang, line 104: I think relying on state here may cause issues (like the page flashing I mentioned) if the session and relay get out of sync. We may need to refactor this section to use graphQL rather than states.

Do you think this is a concern @benjaminpkane?

kaixi-wang (approved):
Just one concern about state we maybe need to keep in mind, but looks good to merge!

app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
manivoxel51: good point on multiple tabs being open and some flashing if you switch between views. I'll work on fixing that today. any recommendation is welcomed of course!

manivoxel51: note that this code is already in your rebased branch - this branch a cleaner version

PR #2349, opened 2022-11-24
Adding support for importing annotations on grouped datasets

brimoor:
Previously annotations on grouped datasets would only be loaded for the active slice, whereas the consist thing to do would be for annotations for all slices to be loaded.

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-groups", max_samples=3)
view = dataset.select_group_slices(media_type="image")

view.annotate("test", label_field="ground_truth", launch_editor=True)

# Previously would only load annotations for the current `group_slice`
# Now all annotations are loaded
```

### DIFF
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -415,7 +415,7 @@ def _build_label_schema(
 
     _label_schema = {}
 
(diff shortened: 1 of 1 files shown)


benjaminpkane (approved):
LGTM

PR #2353, opened 2022-11-28
Provide custom task name for CVAT

### LINKED ISSUE #1753: [FR] Allow task names to be provided when annotating with CVAT
Currently, when annotating with the CVAT integration, the name for the tasks that get generated are hardcoded as `FiftyOne_{dataset_name}` which is not ideal when launching multiple annotation runs on the same dataset. There should be an argument allowing the user to provide one or more task names to use for the generated tasks.

Also, changing the default to `FiftyOne_{dataset_name}_{annotation_key}` would be helpful as well.

Rusteam:

Closes #1753 

1. Custom task name can be passed for labelling data in CVAT such as `dataset.annotate("anno_key", backend="cvat", task_name="Custom task name", ...)`
2. Default task name for CVAT to include an annotation key such as `FiftyOne_{dataset_name}_{annotation_key}`

added few lines to existing cvat unit-tests.

        notes for FiftyOne users.

A custom task name can be provided with CVAT annotation backend. Default CVAT task name now includes an annotation key.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -3060,6 +3060,7 @@ class CVATBackendConfig(foua.AnnotationBackendConfig):
             default, no project is used
         project_id (None): an optional ID of an existing CVAT project to which
             to upload the annotation tasks. By default, no project is used
+        task_name (None): an optional task name to use for the created CVAT task
         occluded_attr (None): an optional attribute name containing existing
             occluded values and/or in which to store downloaded occluded values
             for all objects in the annotation run
@@ -3091,6 +3092,7 @@ class CVATBackendConfig(foua.AnnotationBackendConfig):
         job_reviewers=None,
         project_name=None,
         project_id=None,
+        task_name=None,
         occluded_attr=None,
         group_id_attr=None,
         issue_tracker=None,
@@ -3109,6 +3111,7 @@ class CVATBackendConfig(foua.AnnotationBackendConfig):
         self.job_reviewers = job_reviewers
         self.project_name = project_name
         self.project_id = project_id
+        self.task_name = task_name
         self.occluded_attr = occluded_attr
         self.group_id_attr = group_id_attr
         self.issue_tracker = issue_tracker
@@ -4290,8 +4293,11 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
                     project_id = self.create_project(project_name, cvat_schema)
                     project_ids.append(project_id)
 
-                _dataset_name = samples_batch._dataset.name.replace(" ", "_")
-                task_name = "FiftyOne_%s" % _dataset_name
+                if config.task_name is None:
+                    _dataset_name = samples_batch._dataset.name.replace(" ", "_")
+                    task_name = "FiftyOne_%s" % _dataset_name
+                else:
+                    task_name = config.task_name
 
                 (
                     task_id,
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -6,6 +6,7 @@ Utilities for working with datasets in
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+import math
 from collections import defaultdict
 from copy import copy, deepcopy
 from datetime import datetime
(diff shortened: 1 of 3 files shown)


Rusteam:
@ehofesmann just have a couple of notes about this implementation:

1. I had to make a slight workaround to get an annotation key from `annotation_backend.upload_samples()` in order to get the latest annotation run key without changing the main api for all backends.
2. I was unable to build docs to check that the document update generated nicely because of an error in `bash generate_docs.bash`. Should I file an issue? Here is the traceback:
```
Generating model zoo listing page
  File "/Users/rustem.galiullin/Projects/fiftyone/docs/scripts/make_model_zoo_docs.py", line 17, in <module>
    import fiftyone.zoo as foz
  File "/Users/rustem.galiullin/Projects/fiftyone/fiftyone/zoo/__init__.py", line 9, in <module>
    from .datasets import *
  File "/Users/rustem.galiullin/Projects/fiftyone/fiftyone/zoo/datasets/__init__.py", line 20, in <module>
    import fiftyone.utils.data as foud
  File "/Users/rustem.galiullin/Projects/fiftyone/fiftyone/utils/data/__init__.py", line 10, in <module>
    from .base import *
  File "/Users/rustem.galiullin/Projects/fiftyone/fiftyone/utils/data/base.py", line 20, in <module>
  File "/Users/rustem.galiullin/Projects/fiftyone/fiftyone/core/fields.py", line 21, in <module>
  File "/Users/rustem.galiullin/Projects/fiftyone/fiftyone/core/odm/__init__.py", line 10, in <module>
    from .database import (
  File "/Users/rustem.galiullin/Projects/fiftyone/fiftyone/core/odm/database.py", line 64, in <module>
    class DatabaseConfigDocument(Document):
  File "/Users/rustem.galiullin/Projects/fiftyone/fiftyone/core/odm/database.py", line 70, in DatabaseConfigDocument
    version = fof.StringField()
AttributeError: partially initialized module 'fiftyone.core.fields' has no attribute 'StringField' (most likely due to a circular import)
```

fiftyone/utils/cvat.py
```diff
@@ -4290,8 +4293,18 @@ def upload_samples(self, samples, backend):
...
                     project_ids.append(project_id)
 
-                _dataset_name = samples_batch._dataset.name.replace(" ", "_")
-                task_name = "FiftyOne_%s" % _dataset_name
+                # use custom task name if provided, else use default name
+                if config.task_name is None:
+                    _dataset_name = samples_batch._dataset.name.replace(
+                        " ", "_"
+                    )
+                    latest_anno_key = _get_latest_anno_key(samples)
```
brimoor, line 4301: This is a clever workaround to get the `anno_key`, but now that `config.task_name` is supported, I think it would be best to just rely on the user putting their `anno_key` in the task name if that's what they want.

brimoor (changes requested):
In cases where there are more than one task in the run, what do you think about treating `tasks_name` as a base name and appending `1`, `2`, `3`, etc to it?

fiftyone/utils/cvat.py
brimoor: It is unfortunate that we didn't include `anno_key` in either the `upload_annotations()` method or in the `AnnotationBackendConfig`, because it's definitely reasonable to want to use it to generate default values like this 🤦

Rusteam: I guess it's enough for me that I can provide a custom task name to differentiate between tasks for the same dataset. 

Do you think that will require a lot of/breaking changes to pass `anno_key` to `upload_annotations`?

Rusteam:
Would not that require getting the count of annotation keys as well?

brimoor:
No I'm suggesting adding a separate enhancement to this PR: a single annotation run may create multiple tasks (eg 1 per video for video collections, or if the `task_size` parameter is provided for image collections). I'm suggesting that you update the logic so that the tasks within the run are named `task_name1`, task_name2`, etc.

Rusteam:
Oh yeah that can be done. 

How about the default task name? Do we include an annotation key or not?

brimoor:
> How about the default task name? Do we include an annotation key or not?

I don't think we should include annotation key in the default task name because there's no official way to get that information into `upload_samples()` right now without making a breaking change to the annotation API.

brimoor (approved):
LGTM, thanks for the contribution!

PR #2301, opened 2022-11-14
Adding support for on-disk segmentations

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1729

Adds support for storing `Segmentation` and `Heatmap` arrays on-disk. Any methods that require array pixels now accesses them via new `Segmentation.get_mask()` and `Heatmap.get_map()` methods.

Areas of the API that needed to be updated to be fully compatible with this change were:
- Using `get_mask()` in `evaluate_segmentations()`
- Handling in-memory versus on-disk exports when exporting in `ImageSegmentationDirectory` and `FiftyOneDataset` formats
- Adding support for storing `Segmentation` predictions generated by `apply_model()` on disk rather than in the database

Other goodies
- The media for fields declared in `dataset.app_config.media_fields` are now also included when exporting in `FiftyOneDataset` format

Example usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.labels as foul

dataset = foz.load_zoo_dataset(
    "quickstart",
    max_samples=10,
    seed=51,
    shuffle=True,
    dataset_name=fo.get_default_dataset_name(),
).select_fields().clone()

model = foz.load_zoo_model("deeplabv3-resnet101-coco-torch")
dataset.default_mask_targets = model.mask_targets

dataset.apply_model(model, "segmentations", output_dir="/tmp/segmentations")

dataset.export(
    export_dir="/tmp/seg-export",
    dataset_type=fo.types.ImageSegmentationDirectory,
    overwrite=True,
)

dataset.export(
    export_dir="/tmp/fo-export",
    dataset_type=fo.types.FiftyOneDataset,
    overwrite=True,
)

dataset2 = fo.Dataset.from_dir(
    dataset_dir="/tmp/seg-export",
    dataset_type=fo.types.ImageSegmentationDirectory,
    label_field="segmentations",
)

dataset3 = fo.Dataset.from_dir(
    dataset_dir="/tmp/fo-export",
    dataset_type=fo.types.FiftyOneDataset,
)

# All masks are stored on disk
for d in (dataset, dataset2, dataset3):
    assert d.count("segmentations.mask") == 0
    assert d.count("segmentations.mask_path") == len(d)

# App loads segmentations from disk!
session = fo.launch_app(dataset)
```

PR #2358, opened 2022-11-28
Add frontend support for loading overlay masks from disk

sashankaryal:

Frontend support for loading overlay masks from disk

Tested locally.

        notes for FiftyOne users.

Refer to #2301 

-   [ ] Other


app/packages/looker/src/util.ts
```diff
@@ -519,3 +519,9 @@ export const getMimeType = (sample: any) => {
 
 export const isFloatArray = (arr) =>
   arr instanceof Float32Array || arr instanceof Float64Array;
+
+export const getArrayBufferFromUrl = async (url: string) => {
+  const data = await fetch(url);
```
benjaminpkane, line 524: `getFetchFunction` in `@fiftyone/utilities` will need to be used here to support all environments.

sashankaryal: Makes sense, addressed in the next commit! Thanks for the review!

app/packages/looker/src/index.ts
```diff
@@ -642,9 +646,49 @@ export abstract class Looker<
     );
   }
 
+  /**
+   * Some label types (example: segmentation, heatmap) can have their mask datas stored in disks,
+   * we want to impute the `mask` property of those labels with what's stored in the disk
+   */
+  private async imputeOverlayMaskFromDisk(sample: Sample) {
```
benjaminpkane, line 653: This seems like something that should be done in a worker thread, no? Maybe this isn't blocking, but it seems like a bottle neck to load all this data into the main thread, only to transfer it.

app/packages/looker/src/index.ts
```diff
@@ -656,16 +700,19 @@ export abstract class Looker<
...
-    worker.postMessage({
-      method: "processSample",
-      coloring: this.state.options.coloring,
-      sample,
-      uuid: messageUUID,
+    labelsWorker.addEventListener("message", listener);
+
+    // todo: don't process overlays here: https://github.com/voxel51/fiftyone/issues/1344
+    this.imputeOverlayMaskFromDisk(sample).finally(() => {
+      labelsWorker.postMessage({
```
benjaminpkane, line 710: You will need to [transfer](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage) the array buffers that are created if they are created on the main thread.

app/packages/looker/src/index.ts
sashankaryal: Makes sense. Addressed in next commit.

benjaminpkane (approved):
Looks good! Left some stylistic suggestions. Regardless, I assume this will be refactored some in the async work

PR #2361, opened 2022-11-28
Using dataset as source of truth for existing frames when sampling frames with to_frames()

brimoor:
Fixes a bug when calling `to_frames(sample_frames=True)` on a view that filters the frames of the input dataset. The following would previously raise an error but now does not:

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.match_frames(F("frame_number") == 51)

frames1 = view.to_frames(sample_frames=True, sparse=True)
assert len(frames1) == 10
assert dataset.count("frames.filepath") == 10

frames2 = view.to_frames(sample_frames=True)
assert len(frames2) == 1279
assert dataset.count("frames.filepath") == 1279
```

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -5840,10 +5840,11 @@ class SampleCollection(object):
         populated will be omitted from the returned view.
 
         When ``sample_frames`` is True, this method samples each video in the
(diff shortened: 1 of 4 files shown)


benjaminpkane (approved):
LGTM

PR #2362, opened 2022-11-28
Add error handling to Looker

benjaminpkane:
Adds error handling UX to Looker.

### Changes
 * Frame streaming requests are now retried up to 2 times
 * Clicking on the red error icon will now reinstantiate the looker in the grid or expanded view (force retry)
 * Error info will sometimes be available with `Copy error info` 
Screen Shot 2022-11-23 at 6 24 37 PM
Screen Shot 2022-11-23 at 6 24 30 PM


app/packages/looker/src/elements/common/error.ts
```diff
@@ -25,6 +34,12 @@ export class ErrorElement<State extends BaseState> extends BaseElement<State> {
       this.errorElement.classList.add(lookerErrorPage);
       const errorImg = document.createElement("img");
       errorImg.src = errorIcon;
+      errorImg.title = "Click to reload";
```
brimoor, line 37: I haven't been able to trigger the error myself, but looks like here you've added a tooltip and changed the cursor style on hover for the `!` icon.

That's good, but I wonder if users still may not deduce that they can click to refresh. What if the image also changed from a `!` to a refresh icon on hover?

brimoor (approved):
LGTM in the sense that I wasn't able to trigger any errors by scrolling/mousing fast over the grid.

Full disclosure: I'm approving without reviewing the correctness of the retry logic.

app/packages/looker/src/elements/common/error.ts
benjaminpkane: Done

PR #2370, opened 2022-11-30
fix sidebar ordering in embed ctx

ritch:
Ensure the relay environment is created when the `Dataset` component is used apart for `useRouter()`.


app/packages/dataset/src/Dataset.tsx
```diff
@@ -28,6 +28,7 @@ import { DatasetQuery } from "@fiftyone/core";
 // built-in plugins
 import "@fiftyone/looker-3d";
 import "@fiftyone/map";
+import { setCurrentEnvironment } from "@fiftyone/state/src/hooks/useRouter";
```
benjaminpkane, line 31: Maybe formally export/import this function?

benjaminpkane (approved):
LGTM

PR #2371, opened 2022-11-30
Updating 3D visualizer docs

brimoor:
I recently learned from a conversation in the Slack community that our current 3D visualizer documentation is misleading. I believe this fixes it?

### DIFF
--- a/docs/source/user_guide/groups.rst
+++ b/docs/source/user_guide/groups.rst
@@ -595,6 +595,9 @@ Here's how a typical PCD file is structured:
     encoded in the `r` channel of the `rgb` field of the
     `PCD files <https://pointclouds.org/documentation/tutorials/pcd_file_format.html>`_.
 
+    When coloring by intensity in the App, the intensity values are
+    automatically scaled to use the full dynamic range of the colorscale.
+
 As usual, point cloud samples may contain any type and number of custom fields,
 including certain visualizable |Label| types as described below.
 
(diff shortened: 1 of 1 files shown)


ritch:
I would suggest using the render implementation terminology instead of "camera" / "object". Here is the literal structure in xml-ish format:

```jsx
<Canvas>
  <Scene>
     <PointCloud location={[0, 0, 0]} rotation={
     <Cube location={[x, y, z]} dimensions={[sizeX, sizeY, sizeZ]}  />
     <Line points={[vecA, vecB]} />
     <mesh id="a" location={[x2, y2, z2]}>
        <mesh id="b" location=[{1, 1, 1,}] />
     </mesh>
     <Camera location={defaultCameraPosition} />
  </Scene>
</Canvas>
```

**Canvas**

The canvas has its own coordinate space, which you can consider "screen space" - relative to the browser window. Eg. `[0, 0]` is the top left (2d, up=-y, down=y, left=-x, right=x). This is the coordinate space that the tooltips are rendered in.

**Scene**

This is the coordinate space that all objects are located in. [0, 0, 0] is the scene "origin". Each object in the scene has its own "origin". Eg. `mesh id=b` is positioned at `(mesh id=a).location + (mesh id=b).location`.

**Camera**

As the user interacts with the camera (drag) the camera is moving around the scene. An example of camera coordinates would be a `lookAt` vector. This vector would be relative to the cameras position, that way if the camera is moved it still has the same rotation. Another example (outside our scope) is camera coordinates in shaders, which can be used for things like generating depth passes.

Also note that the camera's lookAt vector is calculated using a target. Where `lookAt` is in camera coordinates, the `target` vector is in `Scene` coordinates, which is: `[0,0,0]`. This means that regardless of the `defaultCameraPosition` the camera will point at the target in the scene (which is the origin of the scene). This is not configurable at the moment.

**Cube**

Like any other object, the Cube is positioned within the `Scene` coordinates. It has one unique attribute "itemRotation" which works like this: `actualRotation = rotation.add(itemRotation)`. Right now the `y` coordinate of this position is actually based on the dimensions of the cube. This is very likely just a bug leftover from supporting the kitti format/dataset. I would like to remove this quirk and position the cube at the given scene coordinate.

**Line**

Note: `Line` does not currently support `itemRotation`.

ritch:
https://github.com/voxel51/fiftyone/pull/2377/files - did a pass and corrected everything to reflect what is currently released.

Although see my comments inline in the PR.

ritch:
I've moved this to https://github.com/voxel51/fiftyone/pull/2379/files

Issue #2335, opened 2022-11-21
[BUG] Cannot declare dynamic label attribute on patches view using set_values()


brimoor:
In the example below, manually calling `add_sample_field()` should not be required, but an error occurs without it because the implied `Classification` field is not recognized by `set_values()`:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")

# Must manually declare the field on the source dataset first
dataset.add_sample_field(
    "ground_truth.detections.color",
    fo.EmbeddedDocumentField,
    embedded_doc_type=fo.Classification,
)

view = dataset.to_patches("ground_truth")

# This should be allowed without first calling `add_sample_field()` above
view.set_values(
    "ground_truth.color",
    [fo.Classification(label="red") for _ in range(len(view))],
)
```

brimoor:
After digging a bit more, I believe the issue is that the schema expansion code in `set_values()` needs to return a `to_mongo` function in case an embedded document field is being set but wasn't added to the schema.

It would be nice to add a `dynamic=True` option to allow embedded attributes set via `set_values()` to be declared as well.

PR #2372, opened 2022-11-30
Upgrades to set_values() and a new set_label_values() method

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/2335

## Change log

- Adds an optional `set_values(..., dynamic=True)` flag that causes dynamic attributes to be declared on the dataset's schema
    - This is idiomatic, for consistency with other methods like `add_samples(..., dynamic=True)`, `from_dir(..., dynamic=True)`, etc
- Adds support for passing using `Label` instances to `set_values(..., values, ...)` in order to populate new dynamic attributes (with or without adding them to the schema)
    - Previously this was not possible because `set_values()` didn't deduce that it needed to serialize the `Label` instances if the field wasn't already declared on the dataset's schema
- Adds a new `set_label_values()` method with similar API to `set_values()` for cases where one has a dict mapping label IDs to attribute values of those labels that you wish to set
    - This interface differs from `set_values()`, which expects a list of values that must correspond 1-1 with the order of the labels in each sample field of the collection being processed, which may not be appropriate

## Example usage of `set_label_values()`

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.filter_labels("predictions", F("confidence") > 0.99)

label_ids = view.values("predictions.detections.id", unwind=True)
values = {_id: True for _id in label_ids}

# Set label attributes
dataset.set_label_values("predictions.detections.high_conf1", values)

# Set label attributes and add to schema
dataset.set_label_values("predictions.detections.high_conf2", values, dynamic=True)

print(dataset.count("predictions.detections"))  # 5620
print(len(label_ids))  # 279
print(dataset.count_values("predictions.detections.high_conf1"))  # {None: 5341, True: 279}
print(dataset.count_values("predictions.detections.high_conf2"))  # {None: 5341, True: 279}

assert "predictions.detections.high_conf1" not in dataset.get_field_schema(flat=True)
assert "predictions.detections.high_conf2" in dataset.get_field_schema(flat=True)
```

## Example usage of `set_values()`

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.filter_labels("predictions", F("confidence") > 0.99)

values = view.values("predictions.detections.label")
labels = [[fo.Classification(label=v) for v in vv] for vv in values]

# Add nested classifications
view.set_values("predictions.detections.label1", labels)

# Add nested classifications and add to schema
view.set_values("predictions.detections.label2", labels, dynamic=True)

print(dataset.count("predictions.detections"))  # 5620
print(view.count("predictions.detections"))  # 279
print(dataset.count("predictions.detections.label1"))  # 279
print(dataset.count("predictions.detections.label2"))  # 279

assert type(view.first().predictions.detections[0].label1) == fo.Classification
assert type(view.first().predictions.detections[0].label2) == fo.Classification
assert "predictions.detections.label1" not in dataset.get_field_schema(flat=True)
assert "predictions.detections.label2" in dataset.get_field_schema(flat=True)
```


fiftyone/core/collections.py
```diff
@@ -1894,86 +1909,245 @@ def set_values(
...
+        _root, _ = self._handle_frame_field(root)
+        is_list_field = issubclass(label_type, fol._LABEL_LIST_FIELDS)
+        _, label_id_path = self._get_label_field_path(label_field, "id")
+
+        id_map = {}
+
+        if len(values) <= 100000:
+            view = self.select_labels(ids=list(values), fields=label_field)
+        else:
+            view = self
```
benjaminpkane, line 1998: Can you add a note here? Clarifying intent might help readability

benjaminpkane (approved):
LGTM

fiftyone/core/collections.py
brimoor: Yeah good point. Added a note (it's just a heuristic optimization)

PR #2376, opened 2022-11-30
add flag for switching coordinate style

ritch:
Adds a new setting to the `3d` plugin.

**useLegacyCoordinates** - defaults to `false` - if you want to get the original style back, just set this to true

`true` - The y position for `Cuboids` will be `location.y + 0.5 * dimensions[1]`.
`false` - The y position for `Cuboids` will be `location.y`.

PR #2377, opened 2022-11-30
more 3d doc fixes

ritch:


docs/source/user_guide/groups.rst
```diff
@@ -618,14 +618,16 @@ detections represented as |Detection| instances with their `label`, `location`,
     # Object label
     label = "vehicle"
 
-    # The center location ``(x, y, z)`` in camera coordinates of the bottom (y=0)
-    # face of the cuboid
+    # The center location of the object ``(x, y, z)`` in scene coordinates
+    # NOTE: the y coordinate will be offset by half the objects y size.
```
ritch, line 622: This is essentially a bug. I think we should just fix this via: https://github.com/voxel51/fiftyone/pull/2376

ritch: This was originally added to correct the positioning of cuboids in the kitti data set, but it should have been done via a setting, as the kitti labels are rotated.

Something like `settings['objectOriginStyle'] = 'centeredYFloor'`
or apply the offset to the labels themselves on import.

docs/source/user_guide/groups.rst
```diff
@@ -1101,15 +1093,18 @@ shown below under the `plugins.3d` key of your
...
+                // The initial camera position in the 3D scene
                 "defaultCameraPosition": {"x": 0, "y": 0, "z": 0},
 
                 // Transformation from PCD -> reference coordinates
                 "pointCloud": {
                     // A rotation to apply to the PCD's coordinate system
                     "rotation": [0, 0, 0],
 
-                    // Don't render points below this z value
+                    // Define the lower z extent of the "color by height" range
```
brimoor, line 1104: To me, this first line makes it sound like points below `minZ` will still be rendered but will just have the same color as `minZ`. But later you clarify that this does indeed cull points below `minZ`.

ritch: Ah that is confusing. Removing that line.

ritch: See https://github.com/voxel51/fiftyone/pull/2379/files

ritch:
Closing in favor of https://github.com/voxel51/fiftyone/pull/2379/files

PR #2378, opened 2022-11-30
Override gridMeidaField and ModalMediaField to filepath 

lanzhenw:
#2303 

When gridMediaField and modalMediaField are set to a field that’s not in the current view’s schema, the App will override to `filepath` (which must always exist). This makes sure the grid always default render on the right path. 

Locally

        notes for FiftyOne users.

### What areas of FiftyOne doe
Screenshot 2022-11-30 at 11 53 15 AM
s this PR affect?

-   [ ] Other


benjaminpkane (approved):
LGTM 👍

Issue #2382, opened 2022-12-01
[BUG] modal_media_field setting is not respected by the App


brimoor:
On `develop`, when I run the code below, the media field in the modal defaults to `filepath`, not `thumbnail_path`, as I configured it:

```py
import fiftyone as fo
import fiftyone.utils.image as foui
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")

foui.transform_images(
    dataset,
    size=(-1, 64),
    output_field="thumbnail_path",
    output_dir="/tmp/thumbnails",
)

dataset.app_config.media_fields = ["filepath", "thumbnail_path"]
dataset.app_config.grid_media_field = "thumbnail_path"
dataset.app_config.modal_media_field = "thumbnail_path"
dataset.save()

session = fo.launch_app(dataset)
```

Issue #2365, opened 2022-11-28
[BUG] Plugin settings from App config are ignored


brimoor:
On `develop` right now, my 

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-geo")

print(fo.app_config.plugins)  # {'map': {'mapboxAccessToken': 'XXXX'}}
print(dataset.app_config["plugins"])  # {}

dataset.app_config["plugins"] = {"map": {"mapboxAccessToken": "bad-token"}}
dataset.save()

PR #2383, opened 2022-12-01
Fix modal media field and app config settings

benjaminpkane:
### Issues
* Resolves https://github.com/voxel51/fiftyone/issues/2365
* Resolves https://github.com/voxel51/fiftyone/issues/2382


brimoor (approved):
LGTM

PR #2390, opened 2022-12-02
Fix static directory server error

benjaminpkane:
Forces an "almost" empty `fiftyone/server/static` directory to be committed to the repository via a `.gitignore` file. This ensures the web server will start up without error, even if a built App has not been placed in the static directory via `yarn build`.


benjaminpkane:
Closing as this is probably a bad idea. Will interfere with development

PR #2393, opened 2022-12-02
Ensure sample is present when rendering an unloaded looker (Fix groups modal)

benjaminpkane:
Ensures the sample is loaded before rendering an unloaded state. Fixes expanding the group modal.

```py
import fiftyone.zoo as foz

session = fo.Session(dataset) # open the modal
```


ehofesmann (approved):
LGTM, tested and works now!

PR #2320, opened 2022-11-17
[WIP] Point cloud feature map generation utility

jacobmarks:

PR #2395, opened 2022-12-05
Birds eye view updates

brimoor:
Proposed refactor to https://github.com/voxel51/fiftyone/pull/2320 to make it more stylistically consistent with other utils that we provide in `fiftyone.utils.image` and `fiftyone.utils.video`.

Will discuss offline with @jacobmarks.


brimoor (commented):
@jacobmarks I added back support for accessing the raw feature maps. I also added some default behavior (allowing `size` to contain -1) that should avoid the need to support a DPI option that we discussed offline

brimoor:
These changes were manually integrated into https://github.com/voxel51/fiftyone/pull/2320 by @jacobmarks.

PR #2401, opened 2022-12-06
Map errors

benjaminpkane:
Fixes invalid token handling


brimoor (approved):
Confirmed that this works as expected, thanks 🙏

PR #2402, opened 2022-12-06
Fix refreshes

### LINKED ISSUE #2142: [BUG] View bar stops responding after session refresh
On `fiftyone==0.17.2`, the view bar stops responding after the following sequence of operations are done via copy + paste (not run in a script):
- Launch App
- Load view via `session.view = ...`
- Refresh session

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Do this manually, after the App loads
session.view = dataset.limit(10)

# After this, view bar stops responding
session.refresh()
```

benjaminpkane:
Resolves #2142


brimoor (changes requested):
This does prevent the issue of the view bar not responding, but I can see that the view bar is loading twice. Can you fix it without the double load?

I imagine the root issue is some kind of order of operations bug where the view bar refreshes before the new view is fully propagated in state?

https://user-images.githubusercontent.com/25985824/206560173-6c9dde7e-d4f1-4415-87fe-7a1fc40c12ec.mov

brimoor:
I'm not seeing this bug on `develop` anymore! 🎉

PR #2406, opened 2022-12-06
Default App retries

benjaminpkane:
Changes the default number of retries for all network requests to 2

### DIFF
--- a/app/packages/utilities/src/fetch.ts
+++ b/app/packages/utilities/src/fetch.ts
@@ -19,7 +19,7 @@ export interface FetchFunction {
     body?: A,
     result?: "json" | "blob",
     retries?: number,
-    retryDelay?: number
+    retryCodes?: number[]
   ): Promise<R>;
 }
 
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM

app/packages/utilities/src/fetch.ts
```diff
@@ -89,11 +89,11 @@ export const setFetchFunction = (
     const fetchCall = retries
       ? fetchRetry(fetch, {
           retries,
-          retryDelay,
+          0,
           retryOn: (attempt, error, response) => {
             if (
               error !== null ||
-              (response.status >= 400 && attempt < retries)
```
brimoor, line 96: Wait I'm confused; isn't this PR reducing the number circumstances in which retries are performed?

benjaminpkane: Yes. But why would we want to retry in all of those scenarios?

We retry on the event source indefinitely to check for liveliness, but we really shouldn't be retrying on other requests. An error means something isn't working...

brimoor: I was asking bc I didn't see how this PR was adding anything, but I missed that `retries` used to be 0

brimoor (approved):
LGTM

Issue #2405, opened 2022-12-06
[FR] Add support for negative matches via filter widgets


brimoor:
The App's filter widgets do not currently provide a way to do "negative matches".

For example with this dataset:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")
session = fo.launch_app(dataset)
```

and this filter state:

<img width="288" alt="Screen Shot 2022-12-06 at 12 11 43 PM" src="https://user-images.githubusercontent.com/25985824/205978550-3a379a31-ea91-4933-8b3a-c1eacf6e93c3.png">

what is happening is:
- Remove all person boxes, then omit all samples that have no boxes after filtering

which results in 199 of the 200 samples in the grid, many of which do actually contain people, it's just that the boxes have been removed.

This is *not* the same thing as the negative matching condition: 
- Only show samples that do not contain person boxes

which is not currently possible with the filtering UI we have created.

Other examples of negative matching that is not currently supported:
- Show me samples that *do not* have the given sample tag
- Show me samples whose `ListField` contains no values

One possible approach here is to add a `Show matches|Show non-matches` toggle below the `Select|Exclude` toggle. But perhaps the naming/UI can be improved.

brimoor:
See also:
- https://github.com/voxel51/fiftyone/issues/2277
- https://github.com/voxel51/fiftyone/issues/2404

brimoor:
@lanzhenw documenting our offline design discussion:

Here are all of the filtering operations that a user might want to perform:

```py
import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone import ViewField as F

dataset = foz.load_zoo_dataset("quickstart")

labels = ["cat", "dog"]

# Select ['cat', 'dog'] labels and omit empty samples
view1 = dataset.filter_labels("ground_truth", F("label").is_in(labels))
assert sorted(view1.distinct("ground_truth.detections.label")) == labels
assert view1.bounds(F("ground_truth.detections").length())[0] > 0

# Select ['cat', 'dog'] labels
view2 = dataset.filter_labels("ground_truth", F("label").is_in(labels), only_matches=False)
assert sorted(view2.distinct("ground_truth.detections.label")) == labels
assert view2.bounds(F("ground_truth.detections").length())[0] == 0

# Exclude ['cat', 'dog'] labels and omit empty samples
view3 = dataset.filter_labels("ground_truth", ~F("label").is_in(labels))
[...]

PR #2407, opened 2022-12-06
Resolve custom embedded list field in App filters

benjaminpkane:
Resolves filtering in the sidebar for custom embedded list fields, e.g.
```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.add_sample_field("ground_truth.detections.custom_tags", ftype=fo.ListField, subfield=fo.StringField)
sample = dataset.first()
sample.ground_truth.detections[0]["custom_tags"] = ["person", "standing", "resident"]
sample.save()

```

### DIFF
--- a/fiftyone/server/view.py
+++ b/fiftyone/server/view.py
@@ -355,6 +355,12 @@ def _is_label(field):
 
 def _make_scalar_expression(f, args, field):
     expr = None
+    if isinstance(field, fof.ListField):
+        return (
+            f.filter(_make_scalar_expression(F(), args, field.field)).length()
+            > 0
+        )
+
     if isinstance(field, fof.BooleanField):
         true, false = args["true"], args["false"]
         if true and false:


ehofesmann (commented):
This works for filtering by existing tags, but thigs start to get funky when filtering by None
image

I would expect to see all samples when filtering by `None` by by default it doesn't show any, then when I `exclude` I see the sample with this attribute populated. 

I get that this probably makes sense from the FiftyOne side since filtering by `None` likely implies "give me detections where this custom tag attribute is `[None]`" rather than `[]` or `None`"

brimoor:
@lanzhenw was just working on `None` handling in list fields in https://github.com/voxel51/fiftyone/pull/2403.

We decided to remove `None` as an option and replace its role with a more general FR for "negative matches" here: https://github.com/voxel51/fiftyone/issues/2405

brimoor:
I merged `develop` and so for now `None` doesn't appear in the list:

Screen Shot 2022-12-06 at 6 26 00 PM

brimoor (approved):
LGTM

PR #2410, opened 2022-12-06
Revert "Remove `none` as an option from other value options"

lanzhenw:
Reverts voxel51/fiftyone#2403

This can introduce bugs and issues, so i am reverting the pr.


ShawnOakley (approved):
LGTM

PR #2412, opened 2022-12-07
Remove none option (filter null value in rendering)

### LINKED ISSUE #2277: [BUG] Improve None-handling when filtering list fields in the App
Here I'm creating a `ListField` field in my dataset for which some samples have `[]` or `None` values:

```py
import fiftyone as fo
import fiftyone.zoo as foz

for tag in "abc":
    dataset.take(25).tag_samples(tag)

dataset.clone_sample_field("tags", "list_field")

```

On initial load, the filter widget exposes counts of `None` values:

Screen Shot 2022-11-09 at 5 09 34 PM

which makes me think I can filter based on the `None` value.

However, the results are nonsensical:

1. Here I've taken the bait and selected `None`, but instead of (presumably) seeing all samples with no values in `list_field`, I see nothing:
Screen Shot 2022-11-09 at 5 09 29 PM

2. Here I've filtered by another value and the `None` count is negative:
Screen Shot 2022-11-09 at 5 09 21 PM

`None` values are handled separately from non-None in other contexts. For example with this view:

```py

# Some confidences are numbers, some are None
some_nones = (F("confidence") < 0.5).if_else(None, F("confidence"))
view = dataset.set_field("predictions.detections.confidence", some_nones)

session.view = view
```

the widget lets you independently select (a) what valid data to include/exclude and then (b) whether to include or exclude `None`-valued data:

Screen Shot 2022-11-09 at 5 18 09 PM

List fields need similar consideration.

lanzhenw:
fixes #2277 

With this method, when rendering category options, we manually filter out the "null" count options, and do not change recoil results, as in upcoming tickets, we may still want to display the null count in other ways. 

Locally

        notes for FiftyOne users.

-   [ ] Other


lanzhenw:
Screenshot 2022-12-08 at 10 20 53 AM

`None` is removed from `ListField`. Boolean Field and String Field are not impacted.

brimoor (commented):
@benjaminpkane is this the right approach?

We don't want to hide counts for `None` values that are actually elements in a `ListField` (although it may not currently be possible because only typed list fields like `ListField(StringField)` are exposed in the App; but maybe we relax that in the future).

We simply do not want to include a separate `None` count that represents the number of empty/missing lists.

lanzhenw:
> @benjaminpkane is this the right approach?
> 
> We don't want to hide counts for `None` values that are actually elements in a `ListField` (although it may not currently be possible because only typed list fields like `ListField(StringField)` are exposed in the App; but maybe we relax that in the future).
> 
> We simply do not want to include a separate `None` count that represents the number of empty/missing lists.

i used `x.value !== null`. It will check not only for the text but also the type for `null`, so 'null' string is not treated the same as `null` type.

app/packages/core/src/components/Filters/CategoricalFilter.tsx
```diff
@@ -102,18 +102,25 @@ const Wrapper = ({
...
-  const counts = Object.fromEntries(results);
+  const filteredResults = schema?.ftype.includes("ListField")
+    ? results.filter(([v]) => v !== null)
+    : results;
+  const counts = Object.fromEntries(filteredResults);
   let allValues: V[] = selected.map<V>((value) => ({
     value,
     count: counts[String(value)] ?? 0,
   }));
+  allValues = schema?.ftype.includes("ListField")
```
benjaminpkane, line 113: I don't think this is necessary as these extra values represent items that have already been selected

app/packages/core/src/components/Filters/CategoricalFilter.tsx
```diff
@@ -102,18 +102,25 @@ const Wrapper = ({
   const selectedSet = new Set(selected);
   const setExcluded = excludeAtom ? useSetRecoilState(excludeAtom) : null;
   const sorting = useRecoilValue(fos.sortFilterResults(modal));
-  const counts = Object.fromEntries(results);
+  const filteredResults = schema?.ftype.includes("ListField")
```
benjaminpkane, line 105: I do think this is something that should happen in [`noneCount`](https://github.com/voxel51/fiftyone/blob/a40c2dd272b58b24d54760e84847ca1f508301ac/app/packages/state/src/recoil/aggregations.ts#L84) because this behavior should be shared by any other components. Defining in just this component could lead to other inconsistencies

benjaminpkane (approved):
LGTM! If you you have the cycles I'd encourage you to add a [unit test ](https://github.com/voxel51/fiftyone/tree/develop/app#testing-recoil-and-recoil-relay)or two before merging

PR #2415, opened 2022-12-07
Fix grid sync from event source

benjaminpkane:
Sometimes state updates that are pushed to the App are not handled correctly by the grid. This should fix the issue.


manivoxel51 (approved):
lgtm! 🍨

PR #2419, opened 2022-12-08
Graceful cleanup of multiple config docs

brimoor:
FiftyOne requires a single doc in the `config` collection, which is where it stores its config information. Although this *should not* happen, every now and then a user reports that they have multiple docs in their `config` collection, which currently causes `import fiftyone` to fail.

This PR gracefully detects and handles the multiple config docs scenario by automatically deleting all docs except the one with the latest `version` (seemed like a reasonable choice to me).

## Graceful cleanup in action

Add a duplicate config document:

```py

db = foo.get_db_conn()
config = foo.get_db_config()
d = dict(config.to_dict())
d.pop("_id")
db.config.insert_one(d)
```

Verify that extra doc is automatically cleaned up:

```py
import fiftyone as fo
# Unexpectedly found 2 documents in the 'config' collection; assuming the one with latest 'version' is the correct one
```

No warning this time:

```py
import fiftyone as fo
```

### DIFF
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -89,7 +89,7 @@ def get_db_config():
         cleanup_multiple_config_docs()
 
         # pylint: disable=no-member
-        config = DatabaseConfigDocument.objects.get()
+        config = DatabaseConfigDocument.objects.first()
 
     if config.version is None:
         #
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
LGTM

PR #2422, opened 2022-12-09
Fix none filter for numeric field

### LINKED ISSUE #2399: [BUG] Fix None filter for numeric fields
The `None` selector is not appearing and functioning correctly in all cases.

## Example

For example, below both the `predictions.detections.confidence` and `uniqueness` fields of `view` have some missing values:

```py
import fiftyone as fo
import fiftyone.zoo as foz

some_nones = lambda field: (F(field) < 0.5).if_else(None, F(field))
view = (
    dataset
    .set_field("predictions.detections.confidence", some_nones("confidence"))
    .set_field("uniqueness", some_nones("uniqueness"))
)

session = fo.launch_app(view)
```

**State A**: when I open the filter widget for `predictions.detections.confidence` initially, I see:
Screen Shot 2022-12-06 at 9 39 46 AM

**State B**: if I then drag the 0.5 slider to the right and then all the way back left, I then see:
Screen Shot 2022-12-06 at 9 39 57 AM

**C**: if I then uncheck the `None` field, I see this:
Screen Shot 2022-12-06 at 9 40 06 AM

## Option 1

One way to make the behavior consistent would be the following:
- In State A, `[x] None: 4056` should appear by default, since the grid is showing all 5620 boxes, including the 4056 that have `confidence == None`
- In State B, `[x] None` should not be checked, since we are only seeing `1,564 of 5,620` objects in the grid, which excludes all objects with `confidence == None`

The tricky thing here (and I believe the reason that things are inconsistent to begin with), is that, if the user adjusts the numeric slider's endpoints, then they probably do not want to see None-valued items too, so we were trying to save the user a click by automatically also omitting `None` values when the slider has been used. This leads me to second option:

## Option 2 (let's use this one?)

Only show the `[ ] None` filter if the slider's endpoints are in their default states (fully extended).

This would imply:
- If the slider is in the default location, allow the user to toggle whether `None` values are included or not, with the default being `[x] None`, since we don't filter anything by default and thus the `None` values are in fact included
- If the slider is in a non-default location, automatically omit `None` values, and visually convey this to the user by hiding the `[ ] None` filter altogether
    - This means that a user would *not* be able to ask for all values in `[a, b]` or `None`, but that's okay because they can do that through code if they really want to, and we also do not allow, for example, asking for values in `[a, b]` or `[c, d]` through the filter widget, for example

One additional detail: if the field contains zero `None` values, then hide the `[ ] None` filter altogether, even in the default state, since it would never offer any useful filtering capability (I think this is already the case, just confirming to keep it).

lanzhenw:
fixes #2399 

https://user-images.githubusercontent.com/17770824/206592972-a68968a7-990d-480a-8fca-b97dc1a1b14d.mov

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Filters/NumericFieldFilter.tsx
+++ b/app/packages/core/src/components/Filters/NumericFieldFilter.tsx
@@ -164,9 +164,7 @@ const NumericFieldFilter = ({
     (nonfinites.length === 1 && nonfinites[0][0] === "none")
   );
 
-  const hasNone =
-    nonfinites.some((x) => x[0] === "none") &&
-    nonfinites.find((x) => x[0] === "none")![1].count !== 0;
+  const hasNone = nonfinites.some((x) => x[0] === "none")
   const isSliderAtInitialPostion =
     bounds[0] === values[0] && bounds[1] === values[1];
 
--- a/app/packages/core/src/components/Filters/NumericFieldFilter.tsx
+++ b/app/packages/core/src/components/Filters/NumericFieldFilter.tsx
@@ -129,7 +129,14 @@ const NumericFieldFilter = ({
 
   const setFilter = useSetRecoilState(fos.filter({ modal, path }));
   const bounds = useRecoilValue(fos.boundsAtom({ path, defaultRange }));
-
+  const values = useRecoilValue(
+    fos.rangeAtom({
+      modal,
+      path,
+      defaultRange,
+      withBounds: true,
+    })
+  );
   const ftype = useRecoilValue(fos.fieldType({ path }));
   const field = useRecoilValue(fos.field(path));
   const hasDefaultRange = useRecoilValue(
@@ -156,6 +163,10 @@ const NumericFieldFilter = ({
     nonfinites.length === 0 ||
     (nonfinites.length === 1 && nonfinites[0][0] === "none")
   );
+
+  const hasNone = nonfinites.some((x) => x[0] === "none");
+  const isSliderAtInitialPostion =
+    bounds[0] === values[0] && bounds[1] === values[1];
 
   if (!hasNonfinites && !hasBounds && named) {
     return null;
@@ -238,9 +249,7 @@ const NumericFieldFilter = ({
           />
         ) : null}
         {(hasDefaultRange ||
-          nonfinites.some(
-            ([_, { value: v }]) => v !== nonfinites[0][1].value
-          ) ||
+          (hasNone && isSliderAtInitialPostion) ||
           !hasBounds) &&
           nonfinites.map(([key, props]) => (
             <Checkbox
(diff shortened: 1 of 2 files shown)


app/packages/core/src/components/Filters/NumericFieldFilter.tsx
```diff
@@ -157,6 +164,12 @@ const NumericFieldFilter = ({
     (nonfinites.length === 1 && nonfinites[0][0] === "none")
   );
 
+  const hasNone =
+    nonfinites.some((x) => x[0] === "none") &&
+    nonfinites.find((x) => x[0] === "none")![1].count !== 0;
+  const isSliderAtInitialPostion =
+    bounds[0] === values[0] && bounds[1] === values[1];
```
benjaminpkane, line 171: `isSliderAtInitialPosition` and `hasDefaultRange` should mean the same thing. Maybe you can fix the `isDefaultRange` selector?

app/packages/core/src/components/Filters/NumericFieldFilter.tsx
```diff
@@ -238,9 +251,7 @@ const NumericFieldFilter = ({
           />
         ) : null}
         {(hasDefaultRange ||
-          nonfinites.some(
-            ([_, { value: v }]) => v !== nonfinites[0][1].value
-          ) ||
+          (hasNone && isSliderAtInitialPostion) ||
```
benjaminpkane, line 254: If `isSliderAtInitialPosition` and `hasDefaultRange` have the same definition then this could become:
```js
        {(hasNone && hasDefaultRange) ||
```

app/packages/core/src/components/Filters/NumericFieldFilter.tsx
```diff
@@ -157,6 +164,12 @@ const NumericFieldFilter = ({
     (nonfinites.length === 1 && nonfinites[0][0] === "none")
   );
 
+  const hasNone =
+    nonfinites.some((x) => x[0] === "none") &&
+    nonfinites.find((x) => x[0] === "none")![1].count !== 0;
```
benjaminpkane, line 169: Based on the implementation of `useNonfinites` empty counts are already filtered:
```suggestion
  const hasNone = nonfinites.some((x) => x[0] === "none")
```

app/packages/core/src/components/Filters/NumericFieldFilter.tsx
lanzhenw: I will fix `hasDefaultRange` and name is `isDefaultRange`. I wasn't sure about `hasDefaultRange` so didn't touch it.

benjaminpkane (approved):
LGTM!

PR #2427, opened 2022-12-13
Only show supported eval keys

### LINKED ISSUE #2275: [BUG] Only include supported eval keys in App's patches UI
The patches menu in the App currently includes all evaluation keys, but only evaluations on `Detections` and `Polylines` support patch views. Evaluation keys for unsupported task types should not appear in the dropdown.

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("cifar10", split="test", max_samples=1000, shuffle=True)

def jitter(val):
    if random.random() < 0.10:
        return random.choice(classes)

    return val

predictions = [
]

dataset.evaluate_classifications("predictions", gt_field="ground_truth", eval_key="eval_class")

dataset.to_evaluation_patches("eval_class")
# ValueError: Invalid label field type <class 'fiftyone.core.labels.Classification'>. Extracting patches is only supported for the following types: (<class 'fiftyone.core.labels.Detections'>, <class 'fiftyone.core.labels.Polylines'>)

# Here's how to deduce whether the evaluation has a supported type
info = dataset.get_evaluation_info("eval_class")
label_type = dataset._get_label_field_type(info.config.gt_field)
is_supported = label_type in (fo.Detections, fo.Polylines)

```

Screen Shot 2022-11-09 at 3 16 42 PM

jwertherUM:
Resolves #2275. The current application will display any evaluation key in the Evaluations dropdown in the app's patches menu UI. This is problematic because evaluation types that are not supported will throw an error and cause the app to hang, yet showing them in the dropdown enables the user to pass the evaluation into to_evaluation_patches(). In this PR, rather than showing every key in the active session in the dropdown, I am filtering the keys based on whether or not the type of their corresponding evaluation is supported by the to_evaluation_patches() function. I did so by consulting the dataset schema, as performed in various other places in the Patcher.tsx file to extract the necessary field types.

This patch does not currently include an explicit test suite. I was unable to locate a test suite for the App's core components action directory, but can add a few cases asserting that only certain types get added to the dropdown. That being said, the logic to check the types in the test would be the exact same logic used to populate the dropdown. Circle back on this if possible.

        notes for FiftyOne users.

All evaluation keys present in the patches UI now support patch views. Previously, certain evaluation keys were erroneously left as options.

-   [X] App: FiftyOne application changes
-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Actions/Patcher.tsx
+++ b/app/packages/core/src/components/Actions/Patcher.tsx
@@ -69,7 +69,24 @@ export const clipsFields = selector<string[]>({
 const evaluationKeys = selector<string[]>({
   key: "evaluationKeys",
   get: ({ get }) => {
-    return get(fos.dataset).evaluations.map(({ key }) => key);
+    const paths = get(fos.labelFields({}));
+    const valid = paths.filter((p) =>
+      get(
+        fos.meetsType({
+          path: p,
+          ftype: EMBEDDED_DOCUMENT_FIELD,
+          embeddedDocType: PATCHES_FIELDS,
+        })
+      )
+    );
+    
+    const evals = get(fos.dataset).evaluations.filter(
+      e => valid.includes(e.config.predField) || valid.includes(e.config.gtField)
+    )
+
+    const keys = evals.map(({key}) => key);
+
+    return keys;
   },
 });
 


benjaminpkane (changes requested):
Well done @jwertherUM!

I've recently added test checks for the App. Do you think you could add a couple unit tests for this new implementation to guarantee behavior? A guide to testing in recoil is [here](https://github.com/voxel51/fiftyone/tree/develop/app#testing-recoil-and-recoil-relay).

benjaminpkane (approved):

PR #2429, opened 2022-12-13
[WIP] Virtual fields

### LINKED ISSUE #2186: [FR] Add dynamic fields based on view contents
There are many use cases for fields whose values can be dynamically updated based on the current contents of the view or the filters in the App. 

Example use case: Aafter running detection evaluation, each detection has an attribute populated with TP/FP/FN. Each sample has an `eval_tp`, `eval_fp`, and `eval_fn` field which shows he number of TP/FP/FN detections in that field. However, the sample-level values are based on the values of the view on which evaluation was run. It would be more useful for these values to be computed on each sample as the view/App filters are updated. So, for example, you can filter for specific labels or confidences, then see the corresponding `eval_X` counts update in real time. 

image

This can currently be done using
```python
session.view = dataset.set_field("eval_tp", F("predictions.detections").filter(F("eval") == "tp").length())
```
but only through the view bar, which is not usable for label filtering currently.

The syntax for this feature could potentially look something like this:
```python
sample["eval_tp"] = F("predictions.detections").filter(F("eval") == "tp").length()
```

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/2186

This PR adds support for defining **virtual fields** on datasets and views.

Unlike ordinary fields, virtual field values are not stored in the database; they are dynamically computed and attached when the collection is iterated/aggregated/etc.

Virtual fields are defined by simply including an `expr` when declaring them:

```py
# Declare a virtual field that counts the number of objects in the `ground_truth` field
dataset.add_sample_field(
    "num_objects",
    fo.IntField,
    expr=F("ground_truth.detections").length(),
)
```

Virtual fields may also be added to views via the `set_field()` view stage by including the type of the field as an additional parameter:

```py
# Add a virtual `num_objects` field to the view
view = dataset.set_field(
    "num_objects",
    F("ground_truth.detections").length(),
    ftype=fo.IntField,
)
```

Virtual view fields exist only on the view (and subsequent views derived from them); they are not added to the underlying dataset.

Note that the `ftype` information is required so that the new virtual field can be added to the view's schema. When `ftype` is omitted, `set_field()` can still be used to modify existing fields in-place or to populate dynamic embedded fields without explicitly adding them to the dataset's schema.

## Virtual fields on datasets

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.compute_metadata()

dataset.add_sample_field(
    "num_objects",
    fo.IntField,
    expr=F("ground_truth.detections").length(),
)

bbox_area = (
    F("$metadata.width") * F("bounding_box")[2] *
    F("$metadata.height") * F("bounding_box")[3]
)

dataset.add_sample_field(
    "ground_truth.detections.area_pixels",
    fo.FloatField,
    expr=bbox_area,
)
```

Virtual fields are included in the dataset's schema:

```py
assert "num_objects" in dataset.get_field_schema()
assert "ground_truth.detections.area_pixels" in dataset.get_field_schema(flat=True)

print(list(dataset.get_virtual_field_schema()))
# ['num_objects', 'ground_truth.detections.area_pixels']
```

Virtual fields are read-only on samples:

```py
sample = dataset.first()
print(sample.num_objects)  # 3

sample.num_objects = 4
# ValueError: Virtual fields cannot be edited
```

Virtual fields are applied *after* views:

```py
print(dataset.bounds("num_objects"))
# (0, 39)

print(dataset.bounds("ground_truth.detections.area_pixels"))
# (6.371400000000001, 353569.2344)

view = dataset.filter_labels("ground_truth", F("label") == "carrot")

print(view.bounds("num_objects"))
# (1, 13)

print(view.bounds("ground_truth.detections.area_pixels"))
# (155.3496, 118266.5256)
```

Use `materialize()` to inject virtual fields earlier so they can be filtered on:

```py
view = (
    dataset
    .filter_labels("ground_truth", F("label") == "person")
    .materialize("num_objects")
    .match(F("num_objects") > 10)
    .sort_by(F("num_objects"), reverse=True)
)

print(view.values("num_objects"))
# [14, 14, 14, 14, 14, 14, 14, 14, 14, 14, 13, 13, 13, 11]
```

Virtual fields can be selected/excluded from schemas as usual:

```py
fo.pprint(dataset._pipeline())
# Virtual fields are automatically attached here

view = dataset.exclude_fields("num_objects")
assert "num_objects" not in view.get_field_schema()

fo.pprint(view._pipeline())
# `num_objects` is no longer attached

view = dataset.exclude_fields("ground_truth")
assert "ground_truth" not in view.get_field_schema()

fo.pprint(view._pipeline())
# `ground_truth.detections.area_pixels` is no longer attached

view = dataset.exclude_fields("ground_truth.detections.area_pixels")
assert "ground_truth.detections.area_pixels" not in view.get_field_schema(flat=True)

fo.pprint(view._pipeline())
# `ground_truth.detections.area_pixels` is no longer attached
```

Virtual fields can be "materialized" into regular fields via `save()`:

```py
dataset.save(fields="num_objects")
# Skipping virtual field 'num_objects' when materialize=False

assert "num_objects" in dataset.get_virtual_field_schema()
assert dataset.get_field("num_objects").is_virtual == True

# Converts a virtual field into a regular field
dataset.save(fields="num_objects", materialize=True)

assert "num_objects" not in dataset.get_virtual_field_schema()
assert dataset.get_field("num_objects").is_virtual == False

# Converts all virtual fields into regular fields
dataset.save(materialize=True)

assert len(dataset.get_virtual_field_schema()) == 0
assert dataset.get_field("ground_truth.detections.area_pixels").is_virtual == False

# Fields are no longer virtual; the values are now stored in the database
print(dataset._pipeline())  # []
print(dataset.bounds("num_objects"))  # (0, 39)
print(dataset.bounds("ground_truth.detections.area_pixels"))  # (6.371400000000001, 353569.2344)
```

## Virtual fields on views

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.set_field(
    "num_objects",
    F("ground_truth.detections").length(),
    ftype=fo.IntField,
)

assert "num_objects" not in dataset.get_field_schema()
assert "num_objects" in view.get_field_schema()

print(view.bounds("num_objects"))
# (0, 39)

sample = view.first()
assert sample.num_objects == 3

sample.num_objects = 4
# ValueError: Virtual fields cannot be edited
```


benjaminpkane:
Very cool. Thinking out loud:
* Not sure if virtual fields on datasets are needed with saved views combined with `set_field`
* In an ideal, I always imagined transforming your schema would just work:
```py
box = F("bounding_box")
view = dataset.set_field("ground_truth.detections.area", box[2] * box[3])

assert isinstance(view.get_field_schema(flat=True)["ground_truth.detections.area"], fo.FloatField)
```
* Are these not fields from a view, and therefore could be named **View Fields**? View feels like the accepted terminology here (beyond FiftyOne, DBs in general)

brimoor:
I would love to make:

```py
view = dataset.set_field("ground_truth.detections.area", box[2] * box[3])
```

directly add a new virtual field to the schema, but the trouble is that our `ViewExpression` methods aren't typed, so we have no way to impute from the provided expression what `Field` type to use for the virtual field. For now it seems we have to rely on the user to tell us via `ftype`.

benjaminpkane:
> I would love to make:
> 
> ```python
> view = dataset.set_field("ground_truth.detections.area", box[2] * box[3])
> ```
> 
> directly add a new virtual field to the schema, but the trouble is that our `ViewExpression` methods aren't typed, so we have no way to impute from the provided expression what `Field` type to use for the virtual field. For now it seems we have to rely on the user to tell us via `ftype`.

Yes, I'm just proposing that we put that on the roadmap and work towards it

brimoor:
@benjaminpkane any chance you could help me get this PR across the finish line? I believe there's two small App tweaks that need to happen:

In the grid view:
- If a user filters by a FIELD that is virtual in the sidebar, a `Materialize(FIELD)` stage must be injected prior to the filter

In the modal:
- If the sample contains virtual fields, the virtual field's value must be recomputed every time the user applies filters in the sidebar (equivalently, the entire sample could be reloaded if that is preferable from an implementation standpoint)

benjaminpkane:
> If a user filters by a FIELD that is virtual in the sidebar, a Materialize(FIELD) stage must be injected prior to the filter

The first item (above) is straightforward and  can be worked on. On the second point:

> If the sample contains virtual fields, the virtual field's value must be recomputed every time the user applies filters in the sidebar (equivalently, the entire sample could be reloaded if that is preferable from an implementation standpoint)

It seems like a fair amount of complexity is underneath this. Filtering on `ground_truth` labels could change a sample-level virtual field which could create a disjoint set of results on some virtual label attribute. The loading states would be tricky, but I'm more concerned about breaking the invariant that filtering strictly results in a subset of values, e.g. slider bounds becoming larger because of filtering on some other field.

benjaminpkane:
I think the all virtual fields should probably be materialized before extending the view with sidebar filters. If not, the order in which sidebar filtering stages are added can change results. This would mean modal values would not update

brimoor:
@benjaminpkane hmm, my mental is the following:
- If no filters involve virtual fields, then they are statically computed at the very end
- Whenever a filter does involve a virtual field, a `materialize()` stage is injected prior to the filter. For all subsequent stages in the view, that virtual field behaves as a static field

Materializing all virtual fields before extending the view would reduce the usability of the feature in many cases. For example, the `num_objects` field in the example dataset above allows me to, for example, apply some sidebar filters and then dynamically look at the `num_objects` tag in the grid to see how many objects matched the filter.

This is a subtle feature though; I may be missing something. If you have any concrete problem cases that would help!

benjaminpkane:
I definitely see the value in this feature. I just want to make sure it's clear what the App should do if we add these implied dependency graphs for sample fields. 

A basic decision point that feels unclear shows up in the `num_objects` example.
```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.compute_metadata()

dataset.add_sample_field(
    "num_objects",
    fo.IntField,
    expr=F("ground_truth.detections").length(),
)
```

If a user first filters with the `num_obects` slider in the sidebar, and then filters any `ground_truth` field, the natural thing to do is to apply the `ground_truth` filtering first. So this means the extended view needs to know that `num_objects` is a virtual field that should materialize after `ground_truth`. In this case, the slider values need not change, but the data being filtered would change.

But if the virtual field was instead `has_objects` (a contrived string field example) and every sample had detections before filtering, then the filter would only materialize `yes` as a possible value. Once filtering occurs, `no` values could materialize which means the App may **have** to re-query values for virtual fields in the sidebar as other filters change to remain consistent with sample data.
```py
dataset.add_sample_field(
    "has_objects",
    fo.StringField,
    expr=(F("ground_truth.detections").length() > 1).if_else("yes", "no"),
)
```

Issue #1344, opened 2021-10-14
[FR] Disable expensive dense label rendering process when dense label fields are toggled off


brimoor:
As discussed in https://github.com/voxel51/fiftyone/issues/1229#issuecomment-942891173, currently the more expensive dense label rendering process required to display `Segmentation`/`Heatmap` data is enabled whenever the `DatasetView` contains any such fields, regardless of whether they have been toggled to visible in the App sidebar.

Since the dense rendering process is substantially more expensive than the basic `Classification`/`Detection` rendering, it would be nice if the dense rendering process would be turned off when all relevant fields are toggled off. Currently one must use `exclude_fields()` to remove them in order to recover the faster loading behavior.

PR #2439, opened 2022-12-14
[WIP] Feature/async overlay render

sashankaryal:

Refactor code such that we only operate on dense labels (deserialize, color, etc.) if they're toggled on.

Tested only for segmentations. Yet to be generalized for all dense labels. Want to get a buy-in first, it's possible this feature is not needed post #2301.

-   [ ] Other


brimoor:
Documenting an ad hoc test I ran: I compared this PR to current `develop` with a dataset of 1000 samples and 10 semantic segmentation fields (masks stored on disk), and didn't see much difference in grid loading time. So it seems this PR would take a bit more work to achieve the spirit of https://github.com/voxel51/fiftyone/issues/1344 (improvement in grid loading time when no masks need to be rendered).

Glad we have the PR to document the current status of this work though!

PR #2440, opened 2022-12-14
Remove label query from sample tagger data flow

### LINKED ISSUE #2433: Sample tag counts take unexpectedly long on video datasets?
I have a video dataset with 1500 samples in it (and many frame-level labels), and I observe that the `Sample tags` loading state takes longer than it does for an image dataset with 25000 samples in it. Why would this be? Sample tags should not involve frames, right?

https://user-images.githubusercontent.com/25985824/207427409-e4769669-4673-4c52-a380-0577c9c75a7a.mov

benjaminpkane:
Resolves #2433

Previously the `allTags` selector would query for both label and sample tags via a aggregation graphql call. Now they have separate dataflow.

### DIFF
--- a/app/packages/core/src/components/Actions/Tagger.tsx
+++ b/app/packages/core/src/components/Actions/Tagger.tsx
@@ -42,20 +42,6 @@ import {
   tagStats,
 } from "./utils";
 
-const IconDiv = styled.div`
-  position: absolute;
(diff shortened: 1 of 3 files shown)


brimoor (approved):
LGTM, thank you! 🚀

PR #2442, opened 2022-12-14
Plugin docs

ritch:
Adds a new section to the documentation:

image

This section is designed to enable plugin developers and core contributors to understand the inner working of the various APIs of the app. Of the past few major projects, most new app code has been built as plugins. We should continue this practice with the consideration of each new major piece of functionality being pluggable. That way we will test and document those new things in a way that new developers (core contributors or plugin develops) can understand, contribute to and use those new features.

Below are a few more screenshots of the docs.

image

image
image
image


ritch:
Fixed the incorrect type/param rendering as well as improved return type rendering when the return type is a function.

image

brimoor (approved):
LGTM! 🚀 

Notes:
- I converted the `index.md` to `index.rst` so that the `.. note::` directive could be used to generate nicer looking callouts
- I found that `jsx-lexer` was needed in my `docs/requirements.txt`

PR #2447, opened 2022-12-19
Filtering cheat sheet

jacobmarks:
Adds a filtering & matching cheat sheet that goes into detail on various common views that users create.


brimoor (changes requested):
I made a general editing pass and fixed some issues/TODOs/etc that I saw. One notable thing I did was to update the example data so that every section has a single code block that can be run to generate the data for that section. Previously there were some cases were multiple code blocks had to be run across sections, which might have led to errors when jumping to a specific section of interest.

FYI- I didn't actually run the code, so I definitely recommend running everything when you're finished revising to make sure everything works 😄

brimoor (approved):
LGTM!

Unfortunately upgrading to Sphinx 4.5 won't be trivial (cf #2530), so I think we'll have to ship with the App filter sections initially.

PR #2448, opened 2022-12-20
Feature/search zoo datasets

jacobmarks:
Adds the ability to list filtered zoo datasets via Python:

```py
import fiftyone as fo
import fiftyone.zoo as foz

# List all zoo datasets
names = foz.list_zoo_datasets()
print(names)

# List all zoo datasets with (both of) the specified tags
names = foz.list_zoo_datasets(tags=["image", "detection"])
print(names)

# List all zoo datasets available via the given source
names = foz.list_zoo_datasets(source="torch")
print(names)
```

analogous to the same functionality that exists via the CLI:

```shell
fiftyone zoo datasets list --names-only
fiftyone zoo datasets list --tags image,detection --names-only
fiftyone zoo datasets list --source torch --names-only
```


brimoor (changes requested):
@jacobmarks I forgot to mention offline that there is already a `fiftyone zoo datasets list` CLI command that provides functionality similar to this:

```
$ fiftyone zoo datasets list -h
usage: fiftyone zoo datasets list [-h] [-n] [-d] [-s SOURCE] [-t TAGS] [-b BASE_DIR]

List datasets in the FiftyOne Dataset Zoo.

    Examples::

        # List available datasets
        fiftyone zoo datasets list

        # List available datasets (names only)
        fiftyone zoo datasets list --names-only

        # List downloaded datasets
        fiftyone zoo datasets list --downloaded-only

        # List available datasets from the given source
        fiftyone zoo datasets list --source <source>

        # List available datasets with the given tag
        fiftyone zoo datasets list --tags <tag>

optional arguments:
  -h, --help            show this help message and exit
  -n, --names-only      only show dataset names
  -d, --downloaded-only
                        only show datasets that have been downloaded
  -s SOURCE, --source SOURCE
                        only show datasets available from the specified source
  -t TAGS, --tags TAGS  only show datasets with the specified tag or list,of,tags
  -b BASE_DIR, --base-dir BASE_DIR
                        a custom base directory in which to search for downloaded datasets
```

It is documented (along with the rest of the zoo API) here: 
https://voxel51.com/docs/fiftyone/user_guide/dataset_zoo/api.html

When I originally implemented this stuff, the primary use case I thought of was simply exploring/discovering what datasets are available, which is something I thought I would do either by (in this order):
1. Exploring the available datasets at https://voxel51.com/docs/fiftyone/user_guide/dataset_zoo/datasets.html
2. Listing the datasets via the CLI

I couldn't think of a use case for filtering zoo datasets via `list_zoo_datasets()` in Python itself. I likely wouldn't want to "download the first dataset smaller than 1GB", for example, because whatever I would write next would likely depend on a particular dataset being downloaded. Or I just want to do a quick test, which is what the various quickstart datasets are for.

Not sure if you knew about this CLI method, but, now that you do; what do you think: still want this filtering to be implemented in Python? or is CLI enough?

Either way, I could see value in wiring up the dataset sizes so that they're visible and potentially filterable via `fiftyone zoo datasets list`. One sticky point here is that the CLI method breaks things down on a per-split basis, but I don't think we have per split size data in all cases right now.

brimoor:
@jacobmarks just following up here: do you think `fiftyone zoo datasets list` is sufficient or would you like to have a way to discover datasets in Python itself?

Also, thoughts on this question?
> Either way, I could see value in wiring up the dataset sizes so that they're visible and potentially filterable via fiftyone zoo datasets list. One sticky point here is that the CLI method breaks things down on a per-split basis, but I don't think we have per split size data in all cases right now.

jacobmarks:
@brimoor thanks for following up. My thinking is that if someone is doing some exploration in a jupyter notebook, they may want to get a list of available datasets, filterable at least by size and tags, IN the notebook, without leaving the notebook. I know you can do CLI commands from a notebook, but it isn't as natural. 

The idea is not to give them full specs on the size of each split of the dataset, but rather to give them the _minimum_ download size so they know if it is worth downloading right then and playing around with. This is in line with the desire to add quickstart-classification, quickstart-pcd, etc. 

I also think if we go this direction, we might want to add a kwarg `minimum_download = False` to datasets, and if set to `True`, this would grab and download the smallest possible subset of the data (or course, for some datasets this may not make sense). This could be paired with a `minimum_download_size` (in Python and CLI) which would overcome the splits issue you alluded to.

Thoughts on this?

fiftyone/zoo/datasets/__init__.py
```diff
@@ -23,18 +23,72 @@
...
+        a sorted list of dataset names
     """
-    datasets = set()
+    if etau.is_str(source):
+        sources = [source]
+    elif source is not None:
+        sources = list(sources)
+    else:
+        sources, _ = _get_zoo_dataset_sources()
+
```
brimoor, line 70: It's important to rely on `_get_zoo_datasets()` to grab the available zoo datasets, because that method contains additional logic that respects the fact that users can [register additional zoo datasets](https://voxel51.com/docs/fiftyone/user_guide/dataset_zoo/api.html#adding-datasets-to-the-zoo).

brimoor (approved):
LGTM

fiftyone/zoo/datasets/__init__.py
jacobmarks: This makes sense. Thanks for adding this

PR #2450, opened 2022-12-20
CSV exporter

brimoor:
Example usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

for tag in ("train", "test", "validation"):
    dataset.take(100).tag_samples(tag)

dataset.export(
    labels_path="/tmp/labels.csv",
    fields=["filepath", "tags", "metadata.size_bytes", "uniqueness"],
    abs_paths=True,
)
```

### DIFF
--- a/fiftyone/utils/csv.py
+++ b/fiftyone/utils/csv.py
@@ -1,0 +1,215 @@
+"""
+CSV utilities.
+
+| Copyright 2017-2022, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+import csv
+
+import eta.core.utils as etau
+
+import fiftyone.core.utils as fou
+import fiftyone.utils.data as foud
+
+
+class CSVDatasetExporter(foud.BatchDatasetExporter, foud.ExportPathsMixin):
+    """A flexible CSV exporter that represents slice(s) of field values of a
+    dataset as columns of a CSV file.
+
+    See :ref:`this page <CSVDataset-export>` for exporting datasets of this
+    type.
+
+    Args:
+        export_dir (None): the directory to write the export. This has no
+            effect if ``data_path`` and ``labels_path`` are absolute paths
+        data_path (None): an optional parameter that enables explicit control
+            over the location of the exported media. Can be any of the
+            following:
+
+            -   a folder name like ``"data"`` or ``"data/"`` specifying a
+                subfolder of ``export_dir`` in which to export the media
+            -   an absolute directory path in which to export the media. In
+                this case, the ``export_dir`` has no effect on the location of
+                the data
+            -   a JSON filename like ``"data.json"`` specifying the filename of
+                the manifest file in ``export_dir`` generated when
+                ``export_media`` is ``"manifest"``
+            -   an absolute filepath specifying the location to write the JSON
+                manifest file when ``export_media`` is ``"manifest"``. In this
+                case, ``export_dir`` has no effect on the location of the data
+
+            If None, the default value of this parameter will be chosen based
+            on the value of the ``export_media`` parameter
+        labels_path (None): an optional parameter that enables explicit control
+            over the location of the exported labels. Can be any of the
+            following:
+
+            -   a filename like ``"labels.csv"`` specifying the location in
+                ``export_dir`` in which to export the labels
+            -   an absolute filepath to which to export the labels. In this
+                case, the ``export_dir`` has no effect on the location of the
+                labels
+
+            If None, the labels will be exported into ``export_dir`` using the
+            default filename
+        export_media (None): controls how to export the raw media. The
+            supported values are:
+
+            -   ``True``: copy all media files into the output directory
+            -   ``False``: don't export media
+            -   ``"move"``: move all media files into the output directory
+            -   ``"symlink"``: create symlinks to the media files in the output
+                directory
+            -   ``"manifest"``: create a ``data.json`` in the output directory
+                that maps UUIDs used in the labels files to the filepaths of
+                the source media, rather than exporting the actual media
+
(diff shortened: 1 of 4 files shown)


fiftyone/utils/csv.py
```diff
@@ -0,0 +1,215 @@
...
+            -   a JSON filename like ``"data.json"`` specifying the filename of
+                the manifest file in ``export_dir`` generated when
+                ``export_media`` is ``"manifest"``
+            -   an absolute filepath specifying the location to write the JSON
+                manifest file when ``export_media`` is ``"manifest"``. In this
+                case, ``export_dir`` has no effect on the location of the data
+
+            If None, the default value of this parameter will be chosen based
+            on the value of the ``export_media`` parameter
+        labels_path (None): an optional parameter that enables explicit control
```
swheaton, line 44: Recommend not calling it `labels_path` and default `labels.csv` since it works on all fields not just labels.

I would like to call it `data_path` but that convention is taken by where the media will go. `metadata.csv` is accurate but likely confusing with the `metadata` field.

It is a little grammatically awkward, but how about `fields_path` and `fields.csv`

fiftyone/utils/csv.py
```diff
@@ -0,0 +1,215 @@
...
+        rel_dir (None): an optional relative directory to strip from each input
+            filepath to generate a unique identifier for each media. When
+            exporting media, this identifier is joined with ``data_path`` to
+            generate an output path for each exported media. This argument
+            allows for populating nested subdirectories that match the shape of
+            the input paths. The path is converted to an absolute path (if
+            necessary) via :func:`fiftyone.core.utils.normalize_path`
+        abs_paths (False): whether to store absolute paths to the media in the
+            exported labels
+        media_path ("filepath"): the name of the field containing the media to
```
swheaton, line 79: Propose rename (refactor to whole file)
```suggestion
        media_path_field ("filepath"): the name of the field containing the media to
```

swheaton (approved):
Flexible and covers many edge cases. LGTM

swheaton:
Approved but with questions/suggestions

fiftyone/utils/csv.py
```diff
@@ -0,0 +1,215 @@
...
+        self._media_idx = None
+        self._include_media = None
+        self._needs_metadata = None
+
+    def setup(self):
+        self._media_exporter = foud.MediaExporter(
+            self.export_media,
+            export_path=self.data_path,
+            rel_dir=self.rel_dir,
+            supported_modes=(True, False, "move", "symlink"),
```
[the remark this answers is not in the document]
brimoor: Yeah good catch. I copied from an exporter that doesn't support manifests, but this one can support. I updated it

fiftyone/utils/csv.py
brimoor: The `data_path` and `labels_path` terminology is a convention that's used by almost every exporter. Perhaps I should not have, but I documented these on the main factory methods like `from_dir()` as well:
https://github.com/voxel51/fiftyone/blob/3c6d689d04605443492e3173c431d6768f842dee/fiftyone/core/dataset.py#L4589

btw there isn't a strict requirement that every exporter actually define and use `data_path` and `labels_paths`l if the user provides those values for an exporter that doesn't support them, I believe they'll see a warning message like "Ignoring unsupported parameter labels_path=XXX". My idea was just to standardize whenever possible so that the main factory method's docstring could be as descriptive as possible.

I agree that "labels" is not necessarily the correct description of what the user is exporting in some cases. But in other cases like [the CSV exporter recipe](https://voxel51.com/docs/fiftyone/recipes/custom_exporter.html), it is in fact "labels" being exported.

On balance, I think it may be best to lean towards consistency and use `labels_path` rather than inventing a new kwarg for this export type.

fiftyone/utils/csv.py
brimoor: I actually meant to use `media_field`. What do you think about that instead? If you grep the codebase (especially fiftyone-teams), you'll see a variety of uses of `media_field="filepath"` to allow the user to customize what field contains the media to export.

brimoor: > Approved but with questions/suggestions

Thanks for the review! FYI I need to add a couple more boilerplate things to this PR before it is mergable:
- Document the new export type here: https://github.com/voxel51/fiftyone/blob/develop/docs/source/user_guide/export_datasets.rst
- Add unit tests here: https://github.com/voxel51/fiftyone/blob/develop/tests/unittests/import_export_tests.py

fiftyone/utils/csv.py
swheaton: Hmm yeah I can see that point of view.
I think the CSV exporter recipe is slightly different because it's inheriting from `LabeledImageDatasetExporter` which I believe means there will be one line per label? Vs. `CSVExporter` would have one line per sample and make the list of labels into a single column. May be wrong though!

Regardless, OK, if we go with `labels_path`, let's at least make it clear in the docstring that we really mean any fields on the samples

fiftyone/utils/csv.py
swheaton: Yep `media_field` sounds good to me!

I can add in those other pieces? unless you have done them already 😉

fiftyone/utils/csv.py
brimoor: The CSV export recipe is for a classification task (one label per sample), so it also does have one row per sample.

Overall, the pattern followed by all importers/exporters is that it is possible to define a 1-1 correspondence between each media (if included in the export) and a chunk of corresponding label data.

swheaton: It supports only `fo.Classification` not, for example, `fo.Detections`, so that's why it's 1-1; I understand now. Thanks for explanation.

PR #2451, opened 2022-12-20
added exclude_groups() to mirror select_groups()

jacobmarks:
`dataset.exclude_groups()`, and the associated `ExcludeGroups()` stage. Syntax is as you'd expect:

```py
import fiftyone as fo
import fiftyone.zoo as foz

group_ids = dataset.take(10).values("group.id")

view = dataset.exclude_groups(group_ids)
view_group_ids = set(view.values("group.id"))
dataset_group_ids = set(dataset.values("group.id"))

assert view_group_ids == set(group_ids)
assert set() = view_group_ids.intersection(dataset_group_ids)
```

-   [ ] Other

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -3250,21 +3250,26 @@ class SampleCollection(object):
             dataset = foz.load_zoo_dataset("quickstart-groups")
 
             #
-            # Select some specific groups by ID
+            # Exclude some specific groups by ID
             #
 
-            group_ids = dataset.take(10).values("group.id")
+            view = dataset.take(2)
+            group_ids = view.values("group.id")
+            all_other_groups = dataset.exclude_groups(group_ids)
 
-            view = dataset.exclude_groups(group_ids)
+            #
+            # Verify set complements
+            #
 
-            view_group_ids = set(view.values("group.id"))
+            excluded_group_ids = set(group_ids)
+            all_other_group_ids = set(all_other_groups.values("group.id"))
             dataset_group_ids = set(dataset.values("group.id"))
 
-            assert view_group_ids == set(group_ids)
-            assert set() = view_group_ids.intersection(dataset_group_ids)
+            assert set() == all_other_group_ids.intersection(excluded_group_ids)
+            assert dataset_group_ids == all_other_group_ids.union(excluded_group_ids)
 
         Args:
-            groups_ids: the groups to select. Can be any of the following:
+            groups_ids: the groups to exclude. Can be any of the following:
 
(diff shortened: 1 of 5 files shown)


fiftyone/core/collections.py
```diff
@@ -3238,6 +3238,51 @@ def exclude_frames(self, frame_ids, omit_empty=True):
...
+
+        Examples::
+
+            import fiftyone as fo
+            import fiftyone.zoo as foz
+
+            dataset = foz.load_zoo_dataset("quickstart-groups")
+
+            #
+            # Select some specific groups by ID
```
brimoor, line 3253: nit: update docstrings to say "exclude" rather than "select" in a few places here

fiftyone/core/stages.py
```diff
@@ -897,6 +897,78 @@ def validate(self, sample_collection):
...
+
+    Examples::
+
+        import fiftyone as fo
+        import fiftyone.zoo as foz
+
+        dataset = foz.load_zoo_dataset("quickstart-groups")
+
+        #
+        # Select some specific groups by ID
```
brimoor, line 911: nit: update docstrings to say "exclude" rather than "select" in a few places here

fiftyone/core/stages.py
```diff
@@ -897,6 +897,78 @@ def validate(self, sample_collection):
...
+
+    @classmethod
+    def _params(cls):
+        return [
+            {
+                "name": "group_ids",
+                "type": "list<id>|id",
+                "placeholder": "list,of,group,ids",
+            }
+        ]
```
brimoor, line 969: would be nice to implement the `validate()` method here as done in `SelectGroups` to verify that the collection is grouped

brimoor (commented):
@jacobmarks thanks for adding this! Implementation looks good, just left a couple small nits.

Also, can you please:
- Add a unit test for this to `tests/unittests/group_tests.py`? Looks like I was lazy and didn't add a test for `select_groups()` but we should cover our bases with a quick test for both
- Consider mentioning `exclude_groups()` in the docs around here? https://voxel51.com/docs/fiftyone/user_guide/groups.html#selecting-groups

fiftyone/core/stages.py
jacobmarks: added

fiftyone/core/stages.py
jacobmarks: fixed

fiftyone/core/collections.py
jacobmarks: fixed

PR #2452, opened 2022-12-21
View stage cheat sheet

jacobmarks:
Adds a cheat sheet that summarizes the available view stages, categorized by type (select, exclude, filter, match, etc).


brimoor (approved):
LGTM! I made a quick editing pass and shored up a few things, but generally I think the categorization of select vs exclude vs filter vs match will definitely add value to readers 💪

PR #2460, opened 2022-12-22
extended view form for aggregate

imanjra:

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/relay/src/queries/countValues.ts
+++ b/app/packages/relay/src/queries/countValues.ts
@@ -1,11 +1,17 @@
 import { graphql } from "react-relay";
 
 export default graphql`
-  query countValuesQuery($dataset: String!, $view: BSONArray!, $path: String!) {
+  query countValuesQuery(
+    $dataset: String!
+    $view: BSONArray!
+    $path: String!
+    $form: ExtendedViewForm
+  ) {
     aggregate(
       datasetName: $dataset
       view: $view
       aggregations: [{ countValues: { field: $path } }]
+      form: $form
     ) {
       __typename
       ... on BoolCountValuesResponse {
--- a/app/packages/state/src/recoil/distributions.ts
+++ b/app/packages/state/src/recoil/distributions.ts
@@ -50,7 +50,7 @@ const extendedViewForm = selector({
       sampleIds: get(extendedSelection),
       filters: get(filters),
       slice: get(groupSlice(false)),
-      mixed: get(groupStatistics(false)) === "slice",
+      mixed: get(groupStatistics(false)) === "group",
     };
   },
 });
--- a/app/packages/state/src/recoil/distributions.ts
+++ b/app/packages/state/src/recoil/distributions.ts
@@ -12,7 +12,7 @@ import { RelayEnvironmentKey } from "./relay";
 
 import { datasetName } from "./selectors";
 import { view } from "./view";
-import { selectorFamily } from "recoil";
+import { selector, selectorFamily } from "recoil";
 import { expandPath, fieldPaths, field, labelFields, fields } from "./schema";
 import {
   BOOLEAN_FIELD,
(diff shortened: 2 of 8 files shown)


app/packages/state/src/recoil/distributions.ts
```diff
@@ -35,11 +35,26 @@ import {
...
 import { State } from "./types";
+import { extendedSelection } from "./atoms";
+import { filters } from "./filters";
+import { groupSlice, groupStatistics } from "./groups";
 
 export type AggregationResponseFrom<
   TAggregate extends { response: { aggregate: readonly unknown[] } }
 > = TAggregate["response"]["aggregate"][0];
 
+const extendedViewForm = selector({
```
kaixi-wang, line 46: Does this have any relation to the StateForm.extended?

app/packages/state/src/recoil/distributions.ts
```diff
@@ -35,11 +35,26 @@ import {
...
+  key: "extendedViewForm",
+  get: ({ get }) => {
+    return {
+      sampleIds: get(extendedSelection),
```
kaixi-wang, line 50: I'm making a selector for creating saved views that looks similar to this one, should we maybe make a form selector in `Selectors.ts` to reuse across the app? That way we don't have to check individual components when something is changed.
```
        send((session) =>
            saveView({
              onError,
              variables: {
                viewName: name,
                datasetName: datasetNameValue,
                viewStages: view,
                form: {
                  filters: snapshot.getLoadable(filters).contents,
                  sampleIds: [
                    ...snapshot.getLoadable(selectedSamples).contents,
                  ],
                  labels: snapshot.getLoadable(selectedLabelList).contents,
                  extended: snapshot.getLoadable(fos.extendedStages).contents,
                  slice:
                    snapshot.getLoadable(resolvedGroupSlice(false)).contents ||
                    null,
                },
                description,
                color,
                subscription,
                session,
              },

File: app/packages/relay/src/queries/countValues.ts [hunk 3]
```diff
@@ -1,11 +1,17 @@
 import { graphql } from "react-relay";
 
 export default graphql`
-  query countValuesQuery($dataset: String!, $view: BSONArray!, $path: String!) {
+  query countValuesQuery(
+    $dataset: String!
+    $view: BSONArray!
+    $path: String!
```
kaixi-wang, line 7: Probably a dumb question, but are the paths from the extended view not already calculated/handled)? I recall seeing some code in `core/aggregations.py` and `aggregations.ts` that seem to be constantly requerying aggregations, but I haven't looked into the exact parameters of the requests.
 
cc @benjaminpkane

kaixi-wang (approved):
lgtm and saw that you had already talked it over with ben 🚢

app/packages/state/src/recoil/distributions.ts
imanjra: Not too sure if it is related to the `extended` property, but there is much overlap between StateForm and ExtendedViewForm.

app/packages/state/src/recoil/distributions.ts
imanjra: Yeah, if this selector overlaps with the selector you're going to create then we can probably use a common selector for both.

imanjra: Currently, the aggregation form does not have `extended` but maybe we can a derived selector which is subset of above for form

app/packages/relay/src/queries/countValues.ts
benjaminpkane: Previously aggregations were grouped into a single request. But now they are issued separately/explicitly as a parameter

app/packages/state/src/recoil/distributions.ts
benjaminpkane: Yes, I would agree with a move to `selectors.ts` if you want to make that change @kaixi-wang

app/packages/state/src/recoil/distributions.ts
benjaminpkane: They are definitely similar. They may be the exact same, but an audit would have to be done to be sure.

benjaminpkane (approved):
LGTM!

PR #2463, opened 2022-12-23
[BUG] workflow for patching quickstart groups

jacobmarks:
I THINK this solves the rotation issue for `quickstart-groups`. The transformation makes the bounding boxes fit the objects really well. The fit isn't perfect, but I've compared against the KITTI multiview (and original quickstart groups) and the fit isn't perfect in those either. Take a look for yourself.

        notes for FiftyOne users.

Users should know that the detections are stored with different directions and numbers, so their quantitative results will be impacted if they use the same pipeline for the old and new versions of the dataset for anything that involves compute.

Undoing the "pointCloud", "overlay", and "itemRotation" from 3D visualizer, making the coordinates of the detections match the point cloud coordinates out of the box.

-   [x] Other -- zoo dataset

### DIFF
--- a/fiftyone/zoo/datasets/base.py
+++ b/fiftyone/zoo/datasets/base.py
@@ -6,12 +6,14 @@ FiftyOne Zoo Datasets provided natively by the library.
 |
 """
 import logging
+import numpy as np
 import os
 import shutil
 
 import eta.core.utils as etau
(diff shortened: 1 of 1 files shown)


brimoor:
Superseded by https://github.com/voxel51/fiftyone/pull/2474.

PR #2464, opened 2022-12-24
Saved view tweaks

brimoor:
Change log
- Adds saved view documentation (awaiting bugfixes and look & feel updates before actually adding the GIF though)
- Refactors the `no_dereference` work to avoid a leaky abstraction (comments inline)

### DIFF
--- a/docs/source/user_guide/app.rst
+++ b/docs/source/user_guide/app.rst
@@ -461,6 +461,44 @@ the App.
     :alt: app-views2
     :align: center
 
+.. _app-saving-views:
+
+Saving views
+____________
+
+You can use the menu in the upper-left of the App to save the current state of
+the App's view bar and filters sidebar as a new view into your dataset. Any
+views that you save are persisted on your dataset under a name of your choice
+so that you can quickly load them in a future session via this UI.
+
+Saved views are a convenient way to record semantically relevant subsets of a
+dataset, such as:
+
+-   samples in a particular state, e.g., with certain tag(s)
+-   a subset of a dataset that was used for a task, e.g., training a model
+-   samples that contain content of interest, e.g., object types or image
+    characteristics
+
+.. note::
+
+    Remember that saved views only store the rule(s) used to extract content
+    from the underlying dataset, not the actual content itself. Saving views
+    is cheap. Don't worry about storage space!
+
+    Keep in mind, though, that the contents of a saved view may change as the
+    underlying dataset is modified. For example, if a save view contains
+    samples with a certain tag, the view's contents will change as you
+    add/remove this tag from samples.
+
+.. note::
+
+    Did you know? You can also programmatically create, modify, and delete
+    saved views :ref:`via Python <saving-views>`!
+
+.. image:: /images/app/app-saved-views.gif
+    :alt: app-saved-views
+    :align: center
+
 .. _app-sample-view:
 
 Viewing a sample
--- a/fiftyone/core/odm/document.py
+++ b/fiftyone/core/odm/document.py
@@ -382,22 +382,9 @@ class MongoEngineBaseDocument(SerializableDocument):
         # pylint: disable=no-member
         return self._fields_ordered
 
-    def to_dict(self, extended=False, no_dereference=False):
+    def to_dict(self, extended=False):
         # pylint: disable=no-member
         d = self.to_mongo(use_db_field=True)
-        if no_dereference:
-            # Serialize the objects in dict and list reference fields instead of just the ObjectIds
-            for k, val in d.items():
-                if type(val) in [list, dict]:
-                    if type(val) == list and len(val) > 0:
-                        if isinstance(val[0], ObjectId):
-                            d[k] = [
-                                obj.serialize() for obj in getattr(self, k)
-                            ]
-                    elif type(val) == dict:
-                        for k2, v2 in val.items():
-                            if isinstance(v2, ObjectId):
-                                d[k][k2] = getattr(self, k)[k2].to_dict()
 
         if not extended:
             return d
--- a/fiftyone/core/odm/views.py
+++ b/fiftyone/core/odm/views.py
@@ -34,11 +34,3 @@ class SavedViewDocument(Document):
     created_at = DateTimeField()
     last_modified_at = DateTimeField()
     last_loaded_at = DateTimeField()
-
-    def serialize(self):
-        d = self.to_dict()
-
-        # TODO: remove nested get when ready to merge
-        d["id"] = str(d.get("_id", d.get("id", ObjectId())))
-        d["dataset_id"] = str(d.get("_dataset_id", d.get("dataset_id", "")))
-        return d
--- a/fiftyone/server/aggregate.py
+++ b/fiftyone/server/aggregate.py
@@ -117,7 +117,6 @@ class AggregateQuery:
     @gql.field
     async def aggregate(
         self,
-        *,
         dataset_name: str,
         view: t.Optional[BSONArray],
         aggregations: t.List[Aggregate],
--- a/fiftyone/server/mutation.py
+++ b/fiftyone/server/mutation.py
@@ -25,7 +25,7 @@ from fiftyone.server.events import get_state, dispatch_event
 from fiftyone.server.filters import GroupElementFilter, SampleFilter
 from fiftyone.server.query import Dataset, SidebarGroup, SavedView
 from fiftyone.server.scalars import BSON, BSONArray, JSON
-from fiftyone.server.view import get_dataset_view, extend_view
+from fiftyone.server.view import get_view, extend_view
 
 
 @gql.input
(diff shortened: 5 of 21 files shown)


fiftyone/core/odm/document.py
```diff
@@ -382,22 +382,9 @@ def _get_repr_fields(self):
         # pylint: disable=no-member
         return self._fields_ordered
 
-    def to_dict(self, extended=False, no_dereference=False):
```
brimoor, line 385: It would in principle make sense to add a `no_dereference` flag to this base class, but the implementation would need to be improved to handle any `ReferenceField`.

The problem with the implementation here is that it contains details specific to saved views/runs, like how `serialize()` is called on lists and `to_dict()` is called on dict keys. If we can't have something generic here, I think we should only add this flag on the `DatasetDocument` class.

fiftyone/core/odm/views.py
```diff
@@ -34,11 +34,3 @@ class SavedViewDocument(Document):
     created_at = DateTimeField()
     last_modified_at = DateTimeField()
     last_loaded_at = DateTimeField()
-
-    def serialize(self):
```
brimoor, line 38: @kaixi-wang I don't think this method makes sense here, because converting `_id (ObjectId)` to `id (str)` is a weird quirk of how you chose to build the GraphQL API. It's not something that other users of the `SavedViewDocument` class should ever need to use. Indeed, the intention is that `to_dict()` should always be used to serialize documents.

So, I moved this logic into the server.

docs/source/user_guide/app.rst
```diff
@@ -461,6 +461,44 @@ the App.
...
+    underlying dataset is modified. For example, if a save view contains
+    samples with a certain tag, the view's contents will change as you
+    add/remove this tag from samples.
+
+.. note::
+
+    Did you know? You can also programmatically create, modify, and delete
+    saved views :ref:`via Python <saving-views>`!
+
+.. image:: /images/app/app-saved-views.gif
```
brimoor, line 498: Hasn't been added yet. Awaiting L&F updates.

fiftyone/server/query.py
```diff
@@ -222,6 +224,14 @@ def modifier(doc: dict) -> dict:
         doc["evaluations"] = list(doc.get("evaluations", {}).values())
         doc["saved_views"] = doc.get("saved_views", [])
 
+        for d in doc["saved_views"]:
+            # @todo I think the `isinstance(d, dict)` check shouldn't be
+            # required, but currently data with no_dereference=False is getting
+            # here somehow... bug?
+            if isinstance(d, dict):
```
brimoor, line 231: @kaixi-wang bringing this to your attention. Without the `isinstance(d, dict)` check, I was getting an error here when launching the App, because a doc with `saved_views` as a list of ObjectIDs was appearing here.

That seems like a bug to me, because that's not what GraphQL expects. And it's not just saved views, runs with ObjectIDs as values were also appearing. IE a non-dereferenced doc is getting here, and I couldn't figure out why at an initial glance.

kaixi-wang: Was the bug consistent? I don't recall ever seeing this bug.

Was there anything unique about the dataset you were loading?
Was this with your change moving the no_dereference to DatasetDocument or before?

brimoor: The bug is consistent. Try loading any dataset with a saved view in the App on this branch with the `if` check commented out.

I don't see anything in this PR that would cause a non-dereferenced doc to get here; I think it's always been a problem but only now raises an error because the reformatting got moved here.

Can you investigate?

fiftyone/core/odm/views.py
brimoor: Removing entirely sounds great. Please do 👍

fiftyone/core/odm/document.py
brimoor: Making it generic may be tricky or at least expensive because you'd have to traverse the entire document tree and check for `ReferenceField` types. I think it is fine to just hard code `no_dereference` in `DatasetDocument` for now, as it is the only usage of reference fields.

fiftyone/server/query.py
kaixi-wang: looked into it and I believe the dereferenced saved_views are coming from the paginator query that directly queries the dataset collection. 
```    data = await foo.aggregate(collection, pipelines)
    results, total = data
    edges = []

    for doc in results:
        _id = doc["_id"]
        edges.append(Edge(node=from_db(doc), cursor=str(_id)))
```
`from_db` calls the Dataset.modifier without calling the `to_dict(no_dereference=True)` serializer, so the saved_views are never resolved from the dbrefs

fiftyone/core/odm/document.py
kaixi-wang: yeah that sounds good.

By the way, I had been looking into `mongoengine.context_managers.no_dereference` for a generic way to handle this, but I couldn't get it to obey flag. It seems like the way we wrap the mongo document types or maybe how we access the data prevents it from working as expected. 
http://docs.mongoengine.org/guide/querying.html#turning-off-dereferencing

kaixi-wang (approved):
lgtm!

fiftyone/server/aggregate.py
```diff
@@ -117,7 +117,6 @@ class AggregateQuery:
     @gql.field
     async def aggregate(
         self,
-        *,
```
brimoor, line 120: @kaixi-wang I removed a number of `*` args in method calls that were introduced by the saved view work. I don't see any reason why they are actually needed?

Was it because certain positional arguments to functions were renamed and you wanted to avoid an error being raised if a  `dataset_name` positional arg used to be called `name`?

I always try to enforce a strict convention that `key=value` syntax is never used when the function arg is positional, not keyword. So it should be allowed to rename positional args.

I'd like to avoid adding `*` in places where it isn't needed

fiftyone/server/view.py
```diff
@@ -213,19 +209,15 @@ def get_dataset_view(
         a :class:`fiftyone.core.view.DatasetView`
     """
     dataset = fod.load_dataset(dataset_name)
+    dataset.reload()
 
-    # Load a previously saved view
-    if view_name is not None and dataset.has_saved_view(view_name):
-        view = dataset.load_saved_view(view_name)
-        view.reload()
-        return
```
brimoor, line 221: @kaixi-wang apparently this code path wasn't ever being used? It wasn't actually returning the view...

fiftyone/server/mutation.py
```diff
@@ -97,7 +97,7 @@ async def set_sidebar_groups(
         sidebar_groups: t.List[SidebarGroupInput],
     ) -> bool:
         state = get_state()
-        view = get_dataset_view(dataset, stages=stages)
+        view = get_view(dataset, stages=stages)
```
brimoor, line 100: @kaixi-wang I didn't get a chance to review the saved view PRs where `get_view()` was renamed to `get_dataset_view()`, but I personally find that change to be unnecessary; it's a function that returns a view (saved or unsaved); no need to change it?

I discovered this rename had occurred in some other branches when merging in latest work. In general I think we'll minimize developer pain by not renaming public methods unless there's a really good reason, so mainly using this rename as an excuse to talk about that.

fiftyone/server/aggregate.py
kaixi-wang: I added them because it was hard to infer/recall what needed to be passed to the functions in what order. Since there aren't type annotations, more than a couple args, and overall many similarly named functions with slightly different signatures, it felt really cumbersome/precarious

brimoor: The `*` just forces callers to use `key=value` syntax, which is already something that all code we write should do for optional function parameters.

brimoor: This seems like more of a stylistic thing that we should either intentionally adopt everywhere, or nowhere.

brimoor:
@kaixi-wang LMK if you are okay with changes I pushed after you reviewed and if so I will merge

fiftyone/server/mutation.py
kaixi-wang: I agree it's generally better to avoid renaming public methods, but this method is only defined in and used in the server code, where `view` is often used interchangeably for a dataset view or a serialized view. I think the benefit of the clarity and knowing that you are dealing with the dataset view as opposed to just stages in this scope would actually be  minimizing developer pain.

fiftyone/server/view.py
kaixi-wang: oh good catch! this used to and is supposed to return the saved view

kaixi-wang:
> @kaixi-wang LMK if you are okay with changes I pushed after you reviewed and if so I will merge

I would prefer if we could start being more specific when discussing views because I think the clarity of knowing that you're working with a dataset view as opposed to just the view stages is worth the additional characters and is helpful for people unfamiliar with the code. However, that's just my preference. If you feel strongly about it and think it actually adds developer pain, then I'm fine with you changing it back.

kaixi-wang (approved):
lgtm. left a comment about the method renaming, but I trust your judgement

fiftyone/server/mutation.py
brimoor: Hmm I see. In my opinion, "view" should always mean `DatasetView`, so the other methods you're referring to (ones that work with "view stages") are the ones whose names are confusing.

PR #2466, opened 2022-12-27
filter | match toggle updates

### LINKED ISSUE #1161: [FR] Add support for toggling between "matching" and "selecting" labels by tags in the App
Currently, when labels are tagged, using the label tag filters in the App will create a `select_labels(tags=)` stage, which will only show exactly the labels with the specified tags and nothing else.

However, often one would like to instead see the *samples* containing the tagged labels, but continue to show ALL labels, so that the tagged labels can be compared to the other labels (eg if the tag is `duplicate`, one wants to be able to see what the other labels look like). This is what the `match_labels(tags=)` stage does.

We should expose the ability to toggle between selecting *only* tagged labels vs matching samples containing the tagged labels.

**Idea 1**
Add a `Label tag mode` setting to the App menu that allows the user to toggle between these two modes:
- `Select`: Use `select_labels(tags=)` to show only tagged labels
- `Match`: Use `match_labels(tags=)` to show samples containing the tagged labels with no filtering

The tooltip for the eye icons in the sidebar could change in response to this setting:
- `Select`: Only show labels with the "{$tag}" tag or other selected tags
- `Match`: Only show samples containing labels with the "{$tag}" tag or other selected tags

The same select/match logic can also be applied when using the eyeball icon in the expanded modal:
- `Select`: hide all labels that don't have the specified tag(s)
- `Match`: changes look-and-feel of tagged labels

**Idea 2**
Another interesting option could be to make the eye icon have three states: unselected -> match -> select -> unselected, which would be quite intuitive because the second click further narrows in the selection by hiding non-matching content

Screen Shot 2021-08-05 at 7 41 23 PM

Screen Shot 2021-08-05 at 7 43 20 PM

### LINKED ISSUE #2405: [FR] Add support for negative matches via filter widgets

lanzhenw:
Resolves #1161 
Resolves #2405 

Updated the `filter vs exclude` toggle to a dropdown with multiple filter options: 

For labels (nested ListField):
  - Filter Label: Filter labels, filter samples
  >  Python SDK: `dataset.filter_labels(field, expr, only_matches=True)`
  - Exclude label: Filter labels, do not filter samples
  > Python SDK: `dataset.filter_labels(field, ~expr, only_matches=False)`
  - Match Samples with Labels: Filter samples
   > Python SDK: `dataset.match_labels(fields=field, filter=expr)`
  - Negative Match Samples: Filter samples
  > Python SDK: `dataset.match_labels(fields=field, filter=~expr, bool=False)`

For other fields:
 - Match Samples
  > Python SDK: `dataset.match(expr)`
 - Negative Match Samples
> Python SDK: `dataset.match(~expr)`

Note: 
1. BooleanField will not have negative filter or negative match options
2. In the modal view, labels will not have match or negative match options
3. Keypoints should not have match options. 

Locally. 

- [x] nested ListField: numeric field, string field (filter, exclude, match, negative match)

- [x] others: booleanField, numeric field, string field, list field (match, negative match)

        notes for FiftyOne users.

- Users can do equivalent of `match_labels` selection in the UI. 
- Filter options are more explicit about what is selected and the corresponding SDK command.

-   [ ] Other


brimoor:
@lanzhenw did some testing just now with this dataset:

```py
import random
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

n = len(dataset)
labels = ["foo", "bar", "spam", "eggs"]

# Add top-level fields of each primitive type
dataset.set_values("bool_field", list(np.random.randn(n) > 0))
dataset.set_values("int_field", list(np.random.randint(10, size=n)))
dataset.set_values("float_field", list(np.random.randn(n)))
dataset.set_values("str_field", [random.choice(labels) for _ in range(n)])

# Add label attributes of each primitive type
p = len(patches)

dataset.add_sample_field("ground_truth.detections.bool_field", fo.BooleanField)
dataset.add_sample_field("ground_truth.detections.int_field", fo.IntField)
dataset.add_sample_field("ground_truth.detections.float_field", fo.FloatField)
dataset.add_sample_field("ground_truth.detections.str_field", fo.StringField)

patches.set_values("ground_truth.bool_field", list(np.random.randn(p) > 0))
patches.set_values("ground_truth.int_field", list(np.random.randint(10, size=p)))
patches.set_values("ground_truth.float_field", list(np.random.randn(p)))
patches.set_values("ground_truth.str_field", [random.choice(labels) for _ in range(p)])

```

Here's some bugs I saw:

1. In the dropdown pictured below, it should read `person        378` rather than `person378`.
Screen Shot 2023-01-10 at 10 00 57 PM

2. I get an error when trying to filter by the top-level `bool_field`:

```
ReferenceError: Cannot access 'Te' before initialization
```

3. The filter options are not appearing for int/float fields. I should see choices for configuring this match, right?

Screen Shot 2023-01-11 at 11 14 46 AM

Not errors but recommendations:

1. How about hiding `None` as an option from `BooleanField` widget when there are no `None` values? For consistency with how `None` is hidden in other contexts.

Screen Shot 2023-01-11 at 11 09 49 AM

With respect to design/look-and-feel, I also have a few suggestions:

1. I think it would be more consistent with rest of the design if you allowed the user to expand the filter options by clicking on the `Filter label` text. This is consistent with how, for example, you can click on `person` to toggle the checkbox to the left.
Screen Shot 2023-01-10 at 10 01 06 PM

2. Expanding on the previous point, I would like to avoid line wrap when the filter tray is in its non-expanded state. To achieve that, how about abbreviating `Show samples with person label` to `Show samples with ...` if it would cause line wrap. You could then show the full text in the tooltip (in place of the tooltip text that is currently shown, which attempts to provide a bit more information about what is actually being filtered, which I think we can discard).

5. What do you think about using the terminology listed below?

```
Show detections with label
Show detections w/o label
Show samples with label
Show samples w/o label
```

Here, `detections` would be computed as `field_type.lower() + "s"`.

I am abbreviating "without" as "w/o" to try to avoid line wrap, but perhaps abbreviations are not a good idea. Maybe you can make this dropdown a bit more compact? Less spacing around icons? Smaller font? Some line wrap here is probably inevitable if the user has chosen a very long field field.

Screen Shot 2023-01-10 at 10 01 12 PM

lanzhenw:
@brimoor the bugs are fixed. I implemented the look and feel changes, made the icons smaller and text tighter.

brimoor:
@lanzhenw here's an updated test dataset that includes list fields:

```py
import random
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

n = len(dataset)
labels = ["foo", "bar", "spam", "eggs"]

dataset.add_sample_field("list_field", fo.ListField, subfield=fo.StringField)

# Add top-level fields of each primitive type
dataset.set_values("bool_field", list(np.random.randn(n) > 0))
dataset.set_values("int_field", list(np.random.randint(10, size=n)))
dataset.set_values("float_field", list(np.random.randn(n)))
dataset.set_values("str_field", [random.choice(labels) for _ in range(n)])
dataset.set_values("list_field", [[random.choice(labels) for _ in range(2)] for __ in range(n)])

# Add label attributes of each primitive type
p = len(patches)

dataset.add_sample_field("ground_truth.detections.bool_field", fo.BooleanField)
dataset.add_sample_field("ground_truth.detections.int_field", fo.IntField)
dataset.add_sample_field("ground_truth.detections.float_field", fo.FloatField)
dataset.add_sample_field("ground_truth.detections.str_field", fo.StringField)
dataset.add_sample_field("ground_truth.detections.list_field", fo.ListField, subfield=fo.StringField)

patches.set_values("ground_truth.bool_field", list(np.random.randn(p) > 0))
patches.set_values("ground_truth.int_field", list(np.random.randint(10, size=p)))
patches.set_values("ground_truth.float_field", list(np.random.randn(p)))
patches.set_values("ground_truth.str_field", [random.choice(labels) for _ in range(p)])
dataset.set_values("ground_truth.list_field", [[random.choice(labels) for _ in range(2)] for __ in range(p)])

```

Here's some bugs I'm seeing on the latest version:

1. The tooltip for `int_field` and perhaps others are not correct. The `length()` syntax is only for `ListField`. For scalar fields, it can be simpler, say:

```
Show samples in the range: dataset.match(expr)
Omit samples in the range: dataset.match(~expr)
```

2. I get an error when trying to filter by the `list_field` from the grid view with the above dataset

3. Filtering by `bool_field` or `str_field` in the sample modal has no effect:

https://user-images.githubusercontent.com/25985824/211970452-d476f191-54fb-478c-8fd3-525e4a643e4f.mov

One design suggestion: can you switch cursor to `hand` when the user hovers over `Show samples with str_field` shown below (like it does when hovering on the filter icon):

Screen Shot 2023-01-11 at 10 51 42 PM

I would also like to suggest some more changes to the tooltips:

1. Rather than attempting to show additional information in the tooltip for the currently selected filtering option, I think it should just show the non-abbreviated version of the same text being shown in the main UI element. For example, in the screenshot below, just show:

```
Main UI: Select detections with bool_fie...
Tooltip: Select detections with bool_field
```

Screen Shot 2023-01-11 at 9 47 37 PM

2. For numeric fields, how about just using the same terminology from other fields? After all, the field may also contain `None` values, in which case the user's filter may not strictly be values within the range; `None` may also be included.

```
Select detections with float_field
Exclude detections with float_field
Show samples with float_field
Omit samples with float_field
```

lanzhenw:
@brimoor issues fixed and ready to be tested again.

app/packages/core/src/components/Filters/categoricalFilter/Wrapper.tsx
```diff
@@ -0,0 +1,173 @@
...
+    allValues = [
+      ...allValues,
+      ...results
+        .filter(([v]) => !selectedSet.has(v))
+        .map(([value, count]) => ({ value, count })),
+    ];
+  }
+
+  if (totalCount === 0) {
+    return (
```
benjaminpkane, line 70: This is not allowed. It will change the number hooks rendered.  The simplest fix will be to put after all hooks, i.e. `use*` calls though it is not the most performant.

Screen Shot 2023-01-13 at 9 06 58 AM

app/packages/core/src/components/Filters/categoricalFilter/Wrapper.tsx
```diff
@@ -0,0 +1,173 @@
...
+    );
+  }
+
+  allValues = [...new Set(allValues)];
+
+  // only show all four options the field is a nested ListField.
+  // pass down nestedField as a prop to generate options. (e.g. "detections")
+  const fieldPath = path.split(".").slice(0, -1).join(".");
+  const fieldSchema = useRecoilValue(fos.field(fieldPath));
+  const nestedField = Boolean(fieldSchema?.ftype.includes("ListField"))
```
benjaminpkane, line 90:
```suggestion
  const nestedField = fieldSchema?.ftype.includes("ListField")
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -0,0 +1,301 @@
+import React, { PropsWithChildren, useEffect } from "react";
+import styled from "styled-components";
+import { RecoilState, useRecoilState, useRecoilValue } from "recoil";
```
benjaminpkane, line 3:
```suggestion
import { RecoilState, useRecoilState } from "recoil";
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -0,0 +1,301 @@
...
+  modal: boolean,
+  isKeyPointLabel: boolean,
+  valueName: string
+) => {
+  //  feature requirements:
+  //  1) only nested ListField items should have the filter and negative filter options;
+  //  2) BooleanField should not have the negative filter or negative match options;
+  //  3) in expanded mode or keypoints field, do not show the match or negative match options;
+  const options: Option[] = [];
+  if (Boolean(nestedField)) {
```
benjaminpkane, line 52:
```suggestion
  if (nestedField) {
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -0,0 +1,301 @@
...
+      value: `Exclude ${nestedField} with ${valueName}`,
+      tooltip: "dataset.filter_labels(field, ~expr, only_matches=False)",
+    });
+  }
+  if (!modal && !isKeyPointLabel) {
+    options.push({
+      icon: "ImageIcon",
+      key: "match",
+      value: `Show samples with ${valueName}`,
+      tooltip: Boolean(nestedField)
```
benjaminpkane, line 73:
```suggestion
      tooltip: nestedField
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -0,0 +1,301 @@
...
+  const options: Option[] = [];
+  if (Boolean(nestedField)) {
+    options.push({
+      icon: "FilterAltIcon",
+      key: "filter",
+      value: `Select ${nestedField} with ${valueName}`,
+      tooltip: "dataset.filter_labels(field, expr, only_matches=True)",
+    });
+  }
+  if (Boolean(nestedField) && !shouldNotShowExclude) {
```
benjaminpkane, line 60:
```suggestion
  if (nestedField && !shouldNotShowExclude) {
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -0,0 +1,301 @@
...
+        ? "dataset.match_labels(fields=field, filter=expr)"
+        : "dataset.match(expr)",
+    });
+  }
+  if (!modal && !shouldNotShowExclude && !isKeyPointLabel) {
+    options.push({
+      icon: "HideImageIcon",
+      key: "negativeMatch",
+      value: `Omit samples with ${valueName}`,
+      tooltip: Boolean(nestedField)
```
benjaminpkane, line 83:
```suggestion
      tooltip: nestedField
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -0,0 +1,301 @@
...
+  excludeAtom,
+  onlyMatchAtom,
+  isMatchingAtom,
+}) => {
+  const [key, setKey] = React.useState<Key | null>(null);
+
+  const [open, setOpen] = React.useState(false);
+  const [excluded, setExcluded] = useRecoilState(excludeAtom);
+  const [onlyMatch, setOnlyMatch] = useRecoilState(onlyMatchAtom);
+  const [isMatching, setIsMatching] = useRecoilState(isMatchingAtom);
```
benjaminpkane, line 117:
```suggestion
  const setOnlyMatch = useSetRecoilState(onlyMatchAtom);
  const setIsMatching = useSetRecoilState(isMatchingAtom);
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -0,0 +1,301 @@
...
+    valueName
+  );
+
+  useEffect(() => {
+    // on initial load, if filter already exists, load exisiting filter for modal filters
+    // otherwise, show defaults: filter for nested listfield, match for other fields
+    if (key === null) {
+      // when filter initializes, isMatching is defaulted to false, but we need to use Boolean(nestedField) to set normal defaults to match
+      if (!excluded) {
+        Boolean(nestedField) ? setKey("filter") : setKey("match");
```
benjaminpkane, line 143:
```suggestion
        nestedField ? setKey("filter") : setKey("match");
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -0,0 +1,301 @@
...
+      } else {
+        Boolean(nestedField)
```
benjaminpkane, line 145:
```suggestion
        nestedField
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -0,0 +1,301 @@
...
+          ? setKey("negativeFilter")
+          : setKey("negativeMatch");
+      }
+    }
+  }, []);
```
benjaminpkane, line 150: nestedField and excluded are dependencies here. They probably won't change, but unforeseen changes my change that and cause issues
```suggestion
  }, [nestedField, excluded]);
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+  const [open, setOpen] = React.useState(false);
+  const [excluded, setExcluded] = useRecoilState(excludeAtom);
+  const [onlyMatch, setOnlyMatch] = useRecoilState(onlyMatchAtom);
+  const [isMatching, setIsMatching] = useRecoilState(isMatchingAtom);
+
+  const theme = useTheme();
+  const highlightedBGColor = Color(color).alpha(0.25).string();
+
+  const popoutRef = React.useRef();
```
benjaminpkane, line 122: Refs should be explicitly typed when they are initially undefined

```suggestion
  const popoutRef = React.useRef<HTMLDivElement>();
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+  font-size: 1rem;
+  margin: auto auto auto 5px;
+  ${({ theme }) => theme.text.secondary};
+`;
+
+const Item = React.memo(
+  React.forwardRef(
+    (
+      { icon, value, tooltip, color, highlightedBGColor, onClick }: ItemProp,
+      ref
```
benjaminpkane, line 29:
```suggestion
        ref: ForwardedRef<HTMLDivElement>
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+          case "filteraltofficon":
+            return <FilterAltOffIcon fontSize="small" />;
+          case "imageicon":
+            return <ImageIcon fontSize="small" />;
+          case "hideimageicon":
+            return <HideImageIcon fontSize="small" />;
+        }
+      };
+
+      const children: DetailedReactHTMLElement<any, HTMLElement> = (
```
benjaminpkane, line 59:
```suggestion
      const children = (
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+          <IconButton sx={{ color: color }}>{getIcon(icon!)}</IconButton>
+
+          <Text>{value}</Text>
+        </div>
+      );
+
+      return (
+        <StyledPanelItem>
+          {tooltip ? (
+            <Tooltip text={tooltip!} placement="right-start">
```
benjaminpkane, line 74:
```suggestion
            <Tooltip text={tooltip} placement="right-start">
```

app/packages/state/src/recoil/pathFilters/numeric.ts
```diff
@@ -240,6 +304,7 @@ export const numeric = selectorFamily<
         }
 
         if (nan && value === "nan") {
+          console.info(nan, exclude);
```
benjaminpkane, line 307: stray log

benjaminpkane (changes requested):
Great work! This is a code review and I only really saw one issue that will cause bugs. The rest are linting comments. I have some suggestions about consolidating UX / L&F that we can take offline.

brimoor:
Looking good @lanzhenw!

I defer to @benjaminpkane's review for code approval 🤗 

I just tested as well and see only a couple small things to mention:

- `False` values are not being rendered in the grid overlays for `BooleanField`s. It should render `True` and `False`, and only for `None` values should nothing render at all. `0` values in `IntegerField` are also not showing, so it seems you need to update some check in the client to render anything non-null, rather than only rendering things that are truth-y

Screen Shot 2023-01-13 at 10 15 47 AM

- Small thing: since the tooltips don't fit to the right of the filters sidebar in the sample modal, how about just hiding them completely? It's not as relevant to see the filter expression, for example, when it is just one sample, not the whole view

Screen Shot 2023-01-13 at 10 24 41 AM

brimoor (approved):
LGTM after last round of comments are addressed

lanzhenw:
Fixed the falsy display issue. 
Screenshot 2023-01-13 at 1 55 12 PM

PR #2468, opened 2022-12-28
Save extended view + additional bug fixes

kaixi-wang:

- [x] Fix frozen view bar and final state warnings
- [x] Fix clearing view bar error (https://voxel51.atlassian.net/browse/TEAMS-624)
- [x] Enable saving extended view without first tapping the bookmark button
- [x] Create shared selectorFamily form to reuse for view related mutations
- [x] Cleanup code for no_dereferencing

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Sidebar/Sidebar.tsx
+++ b/app/packages/core/src/components/Sidebar/Sidebar.tsx
@@ -736,71 +736,73 @@ const InteractiveSidebar = ({
           </Box>
         </Suspense>
       )}
-      <SidebarColumn
-        ref={container}
-        onScroll={({ target }) => {
-          if (start.current !== null) {
-            start.current += scroll.current - target.scrollTop;
-          }
-
-          scroll.current = target.scrollTop;
-          down.current && animate(last.current);
-        }}
-      >
-        <Container style={containerController.springs}>
-          {order.current.map((key) => {
-            const entry = items.current[key].entry;
-            if (entry.kind === fos.EntryKind.GROUP) {
-              group = entry.name;
-            }
-            const { shadow, cursor, ...springs } =
-              items.current[key].controller.springs;
-            const { children } = render(
-              key,
-              group,
-              entry,
-              items.current[key].controller,
-              trigger
-            );
-            const style = {};
-            if (entry.kind === fos.EntryKind.INPUT) {
-              style.zIndex = 0;
+      <Suspense>
+        <SidebarColumn
+          ref={container}
+          onScroll={({ target }) => {
+            if (start.current !== null) {
+              start.current += scroll.current - target.scrollTop;
             }
 
-            return (
-              <animated.div
-                onMouseDownCapture={() => {
-                  lastTouched.current = undefined;
-                  placeItems();
-                }}
-                key={key}
-                style={{
-                  ...springs,
-                  boxShadow: shadow.to(
-                    (s) => `rgba(0, 0, 0, 0.15) 0px ${s}px ${2 * s}px 0px`
-                  ),
-                  ...style,
-                }}
-              >
-                <div
-                  ref={(node) => {
-                    if (!items.current[key]) {
-                      return;
-                    }
+            scroll.current = target.scrollTop;
+            down.current && animate(last.current);
+          }}
+        >
+          <Container style={containerController.springs}>
+            {order.current.map((key) => {
+              const entry = items.current[key].entry;
+              if (entry.kind === fos.EntryKind.GROUP) {
+                group = entry.name;
+              }
+              const { shadow, cursor, ...springs } =
+                items.current[key].controller.springs;
+              const { children } = render(
+                key,
+                group,
+                entry,
+                items.current[key].controller,
+                trigger
+              );
+              const style = {};
+              if (entry.kind === fos.EntryKind.INPUT) {
+                style.zIndex = 0;
+              }
 
-                    items.current[key].el &&
-                      observer.unobserve(items.current[key].el);
-                    node && observer.observe(node);
-                    items.current[key].el = node;
+              return (
+                <animated.div
+                  onMouseDownCapture={() => {
+                    lastTouched.current = undefined;
+                    placeItems();
+                  }}
+                  key={key}
+                  style={{
+                    ...springs,
+                    boxShadow: shadow.to(
+                      (s) => `rgba(0, 0, 0, 0.15) 0px ${s}px ${2 * s}px 0px`
+                    ),
+                    ...style,
                   }}
                 >
-                  {children}
-                </div>
-              </animated.div>
-            );
-          })}
-        </Container>
-      </SidebarColumn>
+                  <div
+                    ref={(node) => {
+                      if (!items.current[key]) {
+                        return;
+                      }
+
+                      items.current[key].el &&
+                        observer.unobserve(items.current[key].el);
+                      node && observer.observe(node);
+                      items.current[key].el = node;
+                    }}
+                  >
+                    {children}
+                  </div>
+                </animated.div>
+              );
+            })}
+          </Container>
+        </SidebarColumn>
+      </Suspense>
     </Resizable>
   ) : null;
 };
--- a/app/packages/core/src/components/Sidebar/Sidebar.tsx
+++ b/app/packages/core/src/components/Sidebar/Sidebar.tsx
@@ -379,7 +379,7 @@ enum Direction {
 const SidebarColumn = styled.div`
   position: relative;
   max-height: 100%;
-  height: 100%;
+  height: 92%;
   width: 100%;
 
   overflow-y: scroll;
--- a/app/packages/core/src/components/Sidebar/ViewSelection/ViewDialog.tsx
+++ b/app/packages/core/src/components/Sidebar/ViewSelection/ViewDialog.tsx
@@ -17,7 +17,7 @@ import IconButton from "@mui/material/IconButton";
 import CloseIcon from "@mui/icons-material/Close";
 import DeleteIcon from "@mui/icons-material/Delete";
 import { useTheme } from "@fiftyone/components";
-import { selectedSavedViewState, viewDialogOpen } from ".";
+import { viewDialogOpen } from ".";
 import { DatasetViewOption } from "@fiftyone/components/src/components/Selection/Option";
 import * as fos from "@fiftyone/state";
 import {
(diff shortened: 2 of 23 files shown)


fiftyone/core/odm/dataset.py
```diff
@@ -463,3 +464,23 @@ class DatasetDocument(Document):
     annotation_runs = DictField(ReferenceField(RunDocument))
     brain_methods = DictField(ReferenceField(RunDocument))
     evaluations = DictField(ReferenceField(RunDocument))
+
+    def to_dict(self, extended=False, no_dereference=False):
```
brimoor, line 468: I think this should be updated to be implemented the way I did in https://github.com/voxel51/fiftyone/pull/2464, for two reasons:

1. `to_dict()` should be used to serialize dereferenced documents, not `to_mongo()`. This implementation will lead to subtle bugs if `to_dict()` contains logic not present in `to_mongo()`. For example, in your implementation here, `no_dereference` could be applied recursively if you'd use `to_dict()` instead of `to_mongo()`. And, it so happens that https://github.com/voxel51/fiftyone/pull/2429 is adding some new logic to `to_dict()` right now too

2. This implementation still feels like a hack to me. You are assuming that all lists of object IDs are reference fields, but it is allowable to have a regular `ListField(ObjectId)`. The way to properly implement this method in full abstraction would be to use the `_fields` dict to actually check the field types of each attribute. But, I don't think that is necessary. Better to just use the implementation from https://github.com/voxel51/fiftyone/pull/2464 for now

I should mention that `to_dict()` is a part of the code that we want to be as fast as we can possibly make it. It gets called millions of times when adding/saving documents on datasets, for example.

fiftyone/server/mutation.py
```diff
@@ -28,6 +28,22 @@
 from fiftyone.server.view import get_dataset_view, extend_view
 
 
+async def _build_result_view(result_view, form):
```
brimoor, line 31: no need for `async` decorator here

fiftyone/server/mutation.py
```diff
@@ -292,15 +294,25 @@ async def create_saved_view(
         # view arg required to be an instance of
         # `fiftyone.core.view.DatasetView`
         if use_state:
+            print("stateful create_saved_view dataset:", dataset)
```
brimoor, line 297: remove `print()`

fiftyone/server/query.py
```diff
@@ -135,7 +135,7 @@ def from_doc(cls, doc: SavedViewDocument):
         stage_dicts = [json_util.loads(x) for x in doc.view_stages]
         saved_view = from_dict(
             data_class=cls,
-            data=doc.serialize(),
+            data=doc.to_mongo(),
```
brimoor, line 138: Use `to_dict()` here. The goal of the `fiftyone.core.odm.XXXDocument` classes is to abstract the "mongo" part away from the rest of the code. Ideally in the future we could swap out mongo for something else and the code would continue to work. In order to achieve that, we should only be calling methods that are explicitly defined and documented on the `fiftyone.core.odm.XXXDocument` classes

brimoor (changes requested):

manivoxel51 (commented):
nice work Kacey! 🍨 

I left some comments about the FE changes. 
I'd also suggest testing your branch on
- single tab
- multi tab
- desktop app
- jupyter notebook - (this might have couple bugs as of now)

and go through our checklist in saved view pr https://github.com/voxel51/fiftyone/pull/2461

fiftyone/core/odm/dataset.py
kaixi-wang: Makes sense. Thanks for explaining! I've been unsure whether to use to_mongo() or to_dict(), so this clears it up.

Also, I was assuming you would merge your branch first and then I would just pull your version. That's the only reason why I haven't changed the hacky impl here yet

app/packages/core/src/components/Sidebar/ViewSelection/ViewDialog.tsx
```diff
@@ -132,7 +132,7 @@ export default function ViewDialog(props: Props) {
   }, [nameValue]);
 
   const onSaveView = useCallback(() => {
-    if (isCreating && view?.length) {
```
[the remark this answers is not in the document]
kaixi-wang: Gating by view.length doesn't work for supporting saving extended stages like filtering by label (https://voxel51.atlassian.net/browse/TEAMS-683). In those cases, unless users click the bookmark icon, the visible change in the grid isn't built into the view yet.

app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
```diff
@@ -134,7 +134,11 @@ export default function ViewSelection(props: Props) {
   }, [searchData, selected]);
 
   const loadedView = useRecoilValue<fos.State.Stage[]>(fos.view);
-  const isEmptyView = !loadedView?.length;
+  const [selectedSavedView] = useRecoilState<DatasetViewOption | null>(
+    selectedSavedViewState
+  );
+  const isEmptyView = false; //selectedSavedView.label !== "Unsaved view"; //
```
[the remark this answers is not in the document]
kaixi-wang: Same reason as above (to enable saving extended view). However, for the case of saving all samples, I've been debating whether it is worth checking all the additional atom states of the extended view just to prevent people from saving the full view.

On a semi-related note, it feels like there should be a way to get back to a full view from a saved view using the view selector, but right now, the only way to do it is to remove everything from the view bar. This isn't hard or complicated, but doesn't feel as intuitive

app/packages/state/src/hooks/useSavedViews.ts
```diff
@@ -47,37 +52,44 @@ export default function useSavedViews() {
...
+          );
+          send((session) =>
+            saveView({
+              onError,
+              variables: {
+                viewName: name,
+                datasetName: datasetNameValue,
+                viewStages: view,
+                form: snapshot.getLoadable(viewStateForm({ modal: false }))
+                  .contents,
```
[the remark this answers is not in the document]
kaixi-wang: we need the state of the selected labels/samples/etc to save the current view, which may not be limited to what is present in the view bar

app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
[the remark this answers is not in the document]
manivoxel51: I think it is ok to initially allow saving all samples and remove that check - we can add a guard later if needed. cc @brimoor to confirm

I agree there should be a better way to get to full view besides clearing the viewbar - we could think through a new option in our saved view selection 🤔

fiftyone/server/mutation.py
```diff
@@ -292,15 +294,25 @@ async def create_saved_view(
...
+            result_view = get_dataset_view(
```
manivoxel51, line 298: @kaixi-wang where is get_dataset_view defined? 🤔

brimoor: This needs to be renamed to `fiftyone.server.view.get_view()`. It was renamed in an upstream branch.

kaixi-wang: brian renamed it to get_view in this pr that merged (https://github.com/voxel51/fiftyone/pull/2464/files)

app/packages/core/src/components/Sidebar/Sidebar.tsx
```diff
@@ -736,71 +736,73 @@ const InteractiveSidebar = ({
...
-              key,
-              group,
-              entry,
-              items.current[key].controller,
-              trigger
-            );
-            const style = {};
-            if (entry.kind === fos.EntryKind.INPUT) {
-              style.zIndex = 0;
+      <Suspense>
```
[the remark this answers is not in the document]
manivoxel51: tagging @benjaminpkane since suspense came up in our chat earlier

app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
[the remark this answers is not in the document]
manivoxel51: remove the unused code

app/packages/core/src/components/ViewBar/viewBarMachine.ts
```diff
@@ -164,7 +164,7 @@ function setStages(ctx, stageInfo) {
   }
 }
 
-const viewBarMachine = Machine(
+const viewBarMachine = createMachine(
```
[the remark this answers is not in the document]
kaixi-wang: yeah let me know if this is bad. I changed it to fix the error highlighted in my IDE and looking at the docs/api, `Machine` isn't a function

for ref: https://xstate.js.org/docs/guides/machines.html

app/packages/state/src/hooks/useSetView.ts
```diff
@@ -104,41 +91,38 @@ const useSetView = (
...
             onCompleted: ({ setView: { dataset, view: value } }) => {
               const isDesktop = isElectron();
-              if (router.history.location.state) {
+              if (
+                router.history.location.state &&
+                router.history.location.state.state
+              ) {
                 const newState = {
-                  ...router.history.location.state,
+                  ...router.history.location.state.state,
```
kaixi-wang, line 119: @manivoxel51 I changed this because before we were creating increasingly nested state each time this code was triggered

app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
```diff
@@ -134,8 +135,8 @@ export default function ViewSelection(props: Props) {
   }, [searchData, selected]);
 
   const loadedView = useRecoilValue<fos.State.Stage[]>(fos.view);
-  const isEmptyView = !loadedView?.length;
-
+  const bookmarkIconOn = useRecoilValue(shouldToggleBookMarkIconOnSelector);
+  const isEmptyView = !bookmarkIconOn && !loadedView?.length;
```
manivoxel51, line 139: 👍

app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
```diff
@@ -134,8 +135,8 @@ export default function ViewSelection(props: Props) {
   }, [searchData, selected]);
 
   const loadedView = useRecoilValue<fos.State.Stage[]>(fos.view);
-  const isEmptyView = !loadedView?.length;
-
+  const bookmarkIconOn = useRecoilValue(shouldToggleBookMarkIconOnSelector);
```
manivoxel51, line 138: oh cool. this is how you check if it is an extended view basically

brimoor (approved):
LGTM

benjaminpkane (changes requested):
I'm seeing some behaviour/implementation issues. I have a PR we can discuss offline
* https://github.com/voxel51/fiftyone/pull/2494

PR #2469, opened 2022-12-28
Proposed refactor to embeddings plot backend

brimoor:
Updates the existing `post()` method to handle patches plots connected to non-patches views (and vice versa).

Also adds a proposed refactor that splits the above logic into three separate methods:
- `on_plot_load()`: logic needed to load an initial embeddings plot based on the current view
- `on_extended_view_update()`: logic needed to update the selection state of the points in an existing plot when the extended view changes
- `on_selection()`: logic needed to update the current extended view when a selection occurs in the embeddings plot

### DIFF
--- a/fiftyone/server/routes/embeddings.py
+++ b/fiftyone/server/routes/embeddings.py
@@ -16,6 +16,7 @@ import fiftyone.server.view as fosv
 
 
 MAX_CATEGORIES = 100
+_DATASET_CACHE = {}
 
 
 class Embeddings(HTTPEndpoint):
@@ -31,7 +32,7 @@ class Embeddings(HTTPEndpoint):
         extended_selection = data["extendedSelection"]
         stages = data["view"]
 
-        dataset = fo.load_dataset(dataset_name)
+        dataset = _load_dataset(dataset_name)
         results = dataset.load_brain_results(brain_key)
 
         # This is the view loaded in the view bar
@@ -111,7 +112,7 @@ class Embeddings(HTTPEndpoint):
 
         traces = {}
         for data in zip(curr_ids, curr_points, label_values, selected):
-            add_to_trace(traces, style, *data)
+            _add_to_trace(traces, style, *data)
 
         return {
             "traces": traces,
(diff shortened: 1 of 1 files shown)


fiftyone/server/routes/embeddings.py
```diff
@@ -5,131 +5,297 @@
...
+        if selected_ids is not None:
+            selected = [id in selected_ids for id in curr_ids]
+        else:
+            selected = itertools.repeat(True)
+
         style = (
-            "categorical"
-            if len(distinct_values) <= MAX_CATEGORIES
-            else "continuous"
+            "categorical" if values_count <= MAX_CATEGORIES else "continuous"
```
ritch, line 110: I know you didn't change this, but this isn't exactly right. Any suggestions? Eg. if all the values are `1` it is still technically numeric.

I was going to try something like `get_field_type(labels_field) == Number`

fiftyone/server/routes/embeddings.py
```diff
@@ -5,131 +5,297 @@
...
             "style": style,
             "values_count": values_count,
-            "selected_ids": None,
+            "index_size": index_size,
+            "available_count": available_count,
+            "missing_count": missing_count,
         }
 
+    @route
+    async def on_extended_view_update(self, request: Request, data: dict):
```
ritch, line 188: How are these exposed / expected to be used (the new `@route`s)?

fiftyone/server/routes/embeddings.py
```diff
@@ -5,131 +5,297 @@
...
+            # Samples plot and patches view
+            stage = fo.SelectBy("sample_id", selected_ids)
+
+        d = stage._serialize(include_uuid=False)
+        return {"_cls": d["_cls"], "kwargs": dict(d["kwargs"])}
+
+
+def _load_dataset(dataset_name):
+    dataset = fo.load_dataset(dataset_name)
+    _DATASET_CACHE[dataset_name] = dataset
```
ritch, line 285: Will these ever be cleaned up?

fiftyone/server/routes/embeddings.py
```diff
@@ -5,131 +5,297 @@
...
-        # No filter is applied, everything can be selected
-        return None
-    if selected_ids and extended_selection:
-        return selected_ids & set(extended_selection)
-    elif extended_selection:
-        return set(extended_selection)
-
-
-def add_to_trace(traces, selected_ids, id, points, label, style):
+        if not filters and not extended_stages and not extended_selection:
```
ritch, line 199: I think I can also prevent the api call on the front end in this case.

ritch (approved):

PR #2476, opened 2023-01-03
Publish steps for docs website

benjaminpkane:
Adds the publishing steps via `gsutil` to automate the release of docs updates. @brimoor is there a cloud storage service account secret we can use? Or I can create one. cc @findtopher


brimoor (commented):
@benjaminpkane is there risk that a docs build would fail in the action that would cause an incomplete/corrupted version of the docs to be synced to docs.voxel51.com? Or am I over-analyzing?

I recommend creating new GCS creds specifically for this workflow, just to be proper.

benjaminpkane:
> @benjaminpkane is there risk that a docs build would fail in the action that would cause an incomplete/corrupted version of the docs to be synced to docs.voxel51.com? Or am I over-analyzing?

I would consider it unlikely, but if we do want to roll back a re-upload should take less than a minute

.github/workflows/build-docs.yml
```diff
@@ -76,36 +76,16 @@ jobs:
...
-
-            voxel51/fiftyone@${{ github.sha }}
-          title: Automated fiftyone documentation update
-          body: " "
-          token: ${{ secrets.FIFTYONE_GITHUB_TOKEN }}
-          author: voxel51-bot <bot@voxel51.com>
-          labels: documentation
-          reviewers: brimoor, benjaminpkane
-          team-reviewers: developers
+          credentials_json: "${{ secrets.GCP_CREDENTIALS }}"
```
brimoor, line 87: Needs updating to `DOCS_GCP_CREDENTIALS`?

brimoor (approved):
LGTM

PR #2482, opened 2023-01-04
Upgrading the ffmpeg action 

benjaminpkane:
Should resolve rate limiting issues, noting the following:

* rate limiting can be resolved by supplying a token: https://github.com/FedericoCarboni/setup-ffmpeg/issues/4
* [v1 had no default value for token](https://github.com/FedericoCarboni/setup-ffmpeg/blob/v1/action.yml)
* [v2 correctly uses the github token for a default](https://github.com/FedericoCarboni/setup-ffmpeg/blob/v2/action.yml)


findtopher (approved):
Looks good

PR #2483, opened 2023-01-04
add support for rgb overlays

### LINKED ISSUE #1730: [FR] Add support for RGB Segmentation/Heatmap arrays
A common feature request is to allow [Segmentation](https://voxel51.com/docs/fiftyone/user_guide/using_datasets.html#semantic-segmentation) and [Heatmap](https://voxel51.com/docs/fiftyone/user_guide/using_datasets.html#heatmaps) data to be provided as RGB images, in which case these colors are directly used rather than using the [various strategies](https://voxel51.com/docs/fiftyone/user_guide/using_datasets.html#heatmaps) used to color 2D data.

This will be more feasible when https://github.com/voxel51/fiftyone/issues/1729 is implemented and arrays can be stored outside the main `Sample` document.

sashankaryal:
Resolves #1730.

## Change log

- Adds support for visualizing RGB segmentations in the App
- Adds support for RGB segmentations in `evaluate_segmentations()`
- Adds support for passing RGB mask targets to `objects_to_segmentations()`
- Adds a new `transform_segmentations()` utility to transform existing masks between grayscale <> RGB

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "coco-2017",
    split="validation",
    label_field="instances",
    max_samples=25,
    only_matching=True,
)

    dataset,
    "instances",
    "segmentations",
    mask_targets=mask_targets,
)

dataset.default_mask_targets = mask_targets

# Test in-database RGB segmentations in App

foul.export_segmentations(
)

# Test on-disk RGB segmentations in App
session.refresh()

# Test evaluating RGB segmentations
dataset.rename_sample_field("segmentations", "ground_truth")

results1 = dataset.evaluate_segmentations(
    "predictions", gt_field="ground_truth", mask_targets=mask_targets
)
results1.print_report()

results2 = dataset.evaluate_segmentations("predictions", gt_field="ground_truth")
results2.print_report()

# Convert to grayscale
targets_map = {"#FF6D04": 1, "#499CEF": 2, "#6D04FF": 3}
foul.import_segmentations(dataset, "ground_truth")
foul.transform_segmentations(dataset, "ground_truth", targets_map)
foul.transform_segmentations(
    dataset, "predictions", targets_map, update_mask_targets=True
)

# Edit grayscale values
foul.transform_segmentations(
    dataset, "predictions", {1: 70, 2: 140, 3: 210}, update_mask_targets=True
)

# Convert back to RGB
foul.transform_segmentations(
    dataset, "predictions", {70: "#FF6D04", 140: "#499CEF", 210: "#6D04FF"}
)
```


brimoor (changes requested):
@sashankaryal how's this PR coming? A couple things I noticed when testing just now:

- The tooltip does not render correctly when viewing a dataset without mask targets explicitly defined in the App (does not consistency appear; and when it does seems to be showing the wrong data?)

I also noticed that pixel values that do not appear in the provided `mask_targets` are currently being rendered. Is that intentional? Or are you planning to switch to the currently documented behavior that pixels not in `masks_targets` are not rendered? After testing just now and experiencing the current behavior, I'm thinking it could be fine to go ahead and always render all non-zero classes and just show `pixel` value when a `label` is not available.

I can see value both ways:
- If only `mask_targets` are painted, it would give a simple way to change what is rendered by editing `mask_targets`
- If all non-zero values are painted, then there is never any confusion why user's data is not appearing

sashankaryal:
> @sashankaryal how's this PR coming? A couple things I noticed when testing just now:
> 
> * The tooltip does not render correctly when viewing a dataset without mask targets explicitly defined in the App (does not consistency appear; and when it does seems to be showing the wrong data?)
> 
> I also noticed that pixel values that do not appear in the provided `mask_targets` are currently being rendered. Is that intentional? Or are you planning to switch to the currently documented behavior that pixels not in `masks_targets` are not rendered? After testing just now and experiencing the current behavior, I'm thinking it could be fine to go ahead and always render all non-zero classes and just show `pixel` value when a `label` is not available.
> 
> I can see value both ways:
> 
> * If only `mask_targets` are painted, it would give a simple way to change what is rendered by editing `mask_targets`
> * If all non-zero values are painted, then there is never any confusion why user's data is not appearing

Addressed in the new commits

app/packages/looker/src/overlays/segmentation.ts
```diff
@@ -130,29 +129,80 @@ export default class SegmentationOverlay<State extends BaseState>
...
-      type: "Segmentation",
-    };
+        field: this.field,
+        target,
+        type: "Segmentation",
+      };
+
+      if (!isRgbMaskTargets(maskTargets)) {
+        // getting color here is computationally inefficient, return no color for this edge case
+        return rgbSegmentationInfoWithoutColor;
```
benjaminpkane, line 154: Can you elaborate here? I'm not sure I understand

benjaminpkane (commented):
Generally the code looks good. Nice job!

app/packages/looker/src/overlays/segmentation.ts
sashankaryal: This is for the edge case when the mask is multi-channel (RGB) but the mask targets is not RGB (or is missing). In this scenario, we render the RGB mask as is but the tooltip will behave slightly differently in two respects:
1. The ribbon color won't reflect the segmentation color, it'll be static in the entire canvas (a shade of white, I think).
2. We won't display any "pixel" info.

This comment is re: (1); while it's possible to get the color of the underlying mask, it involves parsing the RGB buffer. In the happy path case, the RGB buffer is reduced into a targets lookup cable and we don't have this problem.

brimoor (approved):
LGTM

PR #2485, opened 2023-01-05
More RGB segmentation support

brimoor:
## Change log

- Adds support for RGB segmentations in `evaluate_segmentations()`
- Adds support for passing RGB mask targets to `objects_to_segmentations()`
- Adds a new `transform_segmentations()` utility to transform existing masks between grayscale <> RGB

## TODO

- Unit tests

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "coco-2017",
    split="validation",
    max_samples=25,
    label_field="instances",
)

    dataset,
    "instances",
    "segmentations",
    mask_targets=mask_targets,
    save_mask_targets=True,
)

# Test in-database RGB segmenations in App

foul.export_segmentations(
)

# Test on-disk RGB segmenations in App
session.refresh()

dataset.rename_sample_field("segmentations", "ground_truth")

results1 = dataset.evaluate_segmentations(
    "predictions", gt_field="ground_truth", mask_targets=mask_targets
)
results1.print_report()

results2 = dataset.evaluate_segmentations("predictions", gt_field="ground_truth")
results2.print_report()

# Convert to grayscale
targets_map = {"#FF6D04": 1, "#499CEF": 2, "#6D04FF": 3}
foul.import_segmentations(dataset, "ground_truth")
foul.transform_segmentations(dataset, "ground_truth", targets_map)
foul.transform_segmentations(
    dataset, "predictions", targets_map, update_mask_targets=True
)

# Edit grayscale values
foul.transform_segmentations(
    dataset, "predictions", {1: 70, 2: 140, 3: 210}, update_mask_targets=True
)

# Convert back to RGB
foul.transform_segmentations(
    dataset, "predictions", {70: "#FF6D04", 140: "#499CEF", 210: "#6D04FF"}
)
```


brimoor:
@sashankaryal I noticed some edge cases that the App doesn't seem to handle:

- The grid fails to render for me when it encounters RGB masks without `dataset.default_mask_targets` or `dataset.mask_targets` explicitly set
- In the example code above, after the `session` is first launched, if I open the expanded modal for some images, the modal crashes (I think?) as I mouse from certain regions with tooltips to regions without
- The App needs to gracefully handle the case where user has provided mismatched masks and mask targets (eg: RGB masks but mask targets that are int keys)

sashankaryal:
- (1) is an easy fix and will be addressed in the next commit.
- (2) will be addressed in the next commit as well.
- Re: (3), agree that it needs to be gracefully handled. I think a sensible default might be to render the segmentation overlay as-is, regardless of whether mask targets are present or not, or if they're mismatched. Only the tooltip is affected by the presence/integrity of mask targets. Let me know if you disagree.

sashankaryal:
Screenshot 2023-01-05 at 12 02 34 PM

Missing anything?

PR #2486, opened 2023-01-05
Adding 3D IoU support

jacobmarks:
Adds support for performing 3D IoU computation when `fiftyone.utils.iou.compute_ious` encounters 3D detections.


fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,204 @@
...
+    rect2 = [(corners2[i, 0], corners2[i, 2]) for i in range(3, -1, -1)]
+
+    _, inter_area = _convex_hull_intersection(rect1, rect2)
+    ymax = min(corners1[0, 1], corners2[0, 1])
+    ymin = max(corners1[4, 1], corners2[4, 1])
+    inter_vol = inter_area * max(0.0, ymax - ymin)
+
+    # area1 = _polygon_area(np.array(rect1)[:, 0], np.array(rect1)[:, 1])
+    # area2 = _polygon_area(np.array(rect2)[:, 0], np.array(rect2)[:, 1])
+    # iou_2d = inter_area / (area1 + area2 - inter_area)
```
brimoor, line 40: I commented this out because it wasn't being used.

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,204 @@
...
+
+    gt_vol = _cuboid_volume(corners1)
+    pred_vol = _cuboid_volume(corners2)
+
+    if gt_crowd:
+        union = pred_vol
+    else:
+        union = gt_vol + pred_vol - inter_vol
+
+    return min(etan.safe_divide(inter_vol, union), 1)
```
brimoor, line 50: Added `safe_divide()` like in 2D IoU computation.

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def _3d_corners(detection):
```
brimoor, line 53: I added this method. The old one is now called `_3d_corners_jacob()`. Isn't it important to include all three rotation angles in order to compute IoU correctly in general?

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,204 @@
...
+    rotation = np.array(detection.rotation)
+    location = np.array(detection.location)
+    x, y, z = np.array(detection.dimensions)
+    R = _rotation_matrix(rotation)
+
+    corners = np.vstack(
+        [
+            [x / 2, x / 2, -x / 2, -x / 2, x / 2, x / 2, -x / 2, -x / 2],
+            [z / 2, z / 2, z / 2, z / 2, -z / 2, -z / 2, -z / 2, -z / 2],
```
brimoor, line 62: I renamed `l, w, h` to `x, y, z` in this implementation, which makes it easier to see that you're swapping y and z coordinates here. Therefore, this method doesn't technically return 3D corners according to how we've defined the `dimensions`/`location`/`rotation`, right?

Since the unit tests are passing, I assume the rest of the implementation accounts for the swapped coordinates. But, I think it would be better to use `(x, y, z)` instead of `(x, z, y)`.

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,204 @@
...
+            [y / 2, -y / 2, -y / 2, y / 2, y / 2, -y / 2, -y / 2, y / 2],
+        ]
+    )
+
+    return (R @ corners + location[:, np.newaxis]).T
+
+
+def _rotation_matrix(rotation):
```
brimoor, line 70: This implementation matches what `o3d.geometry.get_rotation_matrix_from_xyz(rotation)` returns, which is what I used when generating the toy dataset in https://raw.githubusercontent.com/voxel51/fiftyone/develop/docs/source/user_guide/groups.rst.

Saying this because the toy dataset proves that this order of rotation matches how Looker3D renders 3D detections in the App.

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,204 @@
...
+        [[np.cos(ty), 0, np.sin(ty)], [0, 1, 0], [-np.sin(ty), 0, np.cos(ty)]]
+    )
+    Rz = np.array(
+        [[np.cos(tz), -np.sin(tz), 0], [np.sin(tz), np.cos(tz), 0], [0, 0, 1]]
+    )
+    return Rx @ Ry @ Rz
+
+
+# @todo: remove this?
+def _3d_corners_jacob(detection):
```
brimoor, line 85: Remove once we reconcile `(x, y, z)` vs `(x, z, y)` and rotation-y vs rotation-xyz.

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def _cuboid_volume(corners):
+    a = np.sqrt(np.sum((corners[0, :] - corners[1, :]) ** 2))
+    b = np.sqrt(np.sum((corners[1, :] - corners[2, :]) ** 2))
+    c = np.sqrt(np.sum((corners[0, :] - corners[4, :]) ** 2))
+    return a * b * c
+
+
+def _is_clockwise(p):
```
brimoor, line 201: I noticed this method isn't actually used. Remove?

tests/unittests/evaluation_tests.py
```diff
@@ -1496,6 +1497,109 @@ def test_load_evaluation_view_select_fields(self):
...
+        dataset = fo.Dataset()
+        dataset.add_samples(samples)
+        dataset.group_slice = "pcd"
+
+        sample = dataset.first()
+
+        # unit box at origin
+        dims = np.array([1, 1, 1])
+        loc = np.array([0, 0, 0])
+        rot = np.array([0, 0, 0])
```
brimoor, line 1532: Ah none of your tests use a non-zero rotation. Best to include that so we know the methods are working as expected.

brimoor (changes requested):

brimoor (approved):
LGTM!

PR #2491, opened 2023-01-05
Fix CVAT bug when checking if a project exists

ehofesmann:
There was a similar issue when checking if projects exists that we ran into for tasks in https://github.com/voxel51/fiftyone/pull/2070. The issue arises when too many projects exist in once CVAT organizaton.

This is also a similar solution in that it attempts to access the URL for the project and assumes it does not exist on a 404 error. This PR also adds a test for this issue that now passes.


brimoor (approved):
LGTM

PR #2494, opened 2023-01-06
Saved view proposals

benjaminpkane:
### Changes

* Removes `changing_saved_view` and `update` from state events
* Initializes routing and state through initial `getEventSource` connects
* Ensures `router` is used to avoid desktop issues
* Removes `ipython` specific checks
* Moves saved views fragment to the Dataset route query
* Adds the `queryParams` parameter to routing and the `view` search parameter for the Dataset route


app/packages/state/src/hooks/useStateUpdate.ts
```diff
@@ -52,6 +52,9 @@ const useStateUpdate = () => {
 
       if (state) {
         const view = get(viewAtoms.view);
+        if (dataset.stages && !state.view) {
+          state.view = dataset.stages;
```
kaixi-wang, line 56: Does this mean that If dataset.stages exist, then we are loading a saved view? 

In that case, would it make sense or be possible to set state.view_name here?

app/packages/core/src/Dataset.ts
```diff
@@ -8,13 +8,33 @@ import {
...
+
 export const DatasetNodeQuery = graphql`
   query DatasetQuery(
     $name: String!
     $view: BSONArray = null
     $viewName: String = null
   ) {
+    ...DatasetSavedViewsFragment
     dataset(name: $name, view: $view, viewName: $viewName) {
+      stages(slug: $viewName)
```
kaixi-wang, line 37: should `viewName` be `slug`? or vice versa? I think you use slug below.

app/packages/state/src/hooks/useStateUpdate.ts
benjaminpkane: Isn't `viewName` already set from routing? Is there specific behavior you are thinking about?

app/packages/core/src/Dataset.ts
benjaminpkane: Resolved, I believe. Thanks!

fiftyone/server/events.py
```diff
@@ -256,18 +256,53 @@ async def _initialize_listener(payload: ListenPayload) -> InitializedListener:
...
-            state.changing_saved_view = False
+                state.selected = []
+                state.selected_labels = []
+                state.view = None
+                state.view_name = None
+            else:
+                if payload.initializer.view:
+                    try:
+                        state.view = state.dataset.load_saved_view(
+                            payload.initializer.view
```
kaixi-wang, line 283: is the `payload.initializer` the url? and so would `payload.initializer.view` be the value of `view` in the search params?

fiftyone/server/events.py
```diff
@@ -256,18 +256,53 @@ async def _initialize_listener(payload: ListenPayload) -> InitializedListener:
...
+                        )
+                        state.selected = []
+                        state.selected_labels = []
+                        state.view_name = payload.initializer.view
+                    except:
+                        pass
+        elif (
+            payload.initializer.view
+            and payload.initializer.view != current_saved_view
```
kaixi-wang, line 292: if `payload.initializer.view` is from the url, its value is`view.slug`. Is `state.view_name` always going to be the slug? or only set here? If so, maybe rename it to slug to prevent setting it elsewhere in the code using `view.name`

fiftyone/server/view.py
```diff
@@ -161,7 +158,7 @@ def get_saved_view_stages(identifier):
     return list(
         db.views.aggregate(
             [
-                {"$match": {"slug": identifier}},
+                {"$match": {"name": identifier}},
```
kaixi-wang, line 161: the `name` is only unique relative to a dataset. to identify a specific view, either have to use `_id` or `name`+ `_dataset_id`

fiftyone/core/dataset.py
```diff
@@ -3137,14 +3137,13 @@ def save_view(
             raise ValueError("Cannot save view into a different dataset")
 
         view._set_name(name)
-        slug = self._validate_saved_view_name(name, overwrite=overwrite)
+        self._validate_saved_view_name(name, overwrite=overwrite)
```
kaixi-wang, line 3140: why are we removing slug?

fiftyone/core/dataset.py
```diff
@@ -3301,13 +3296,11 @@ def _get_saved_view_doc(self, name, pop=False):
...
 
     def _validate_saved_view_name(self, name, skip=None, overwrite=False):
-        slug = fou.to_slug(name)
-
         for view_doc in self._doc.saved_views:
             if view_doc is skip:
                 continue
 
-            if name == view_doc.name or slug == view_doc.slug:
+            if name == view_doc.name:
```
kaixi-wang, line 3303: I don't' think this will work. Unique names will resolve to the same slug 
image

kaixi-wang (changes requested):
I only looked at the python so far, but I don't think it's safe to just remove slug.

benjaminpkane:
Still reviewing with respect to encoding and view names

app/packages/state/src/hooks/useStateUpdate.ts
kaixi-wang: It wasn't previously. The view name is initially set in `set_view` and afterwards the url is updated in order to sync the other tabs.

fiftyone/core/dataset.py
benjaminpkane: I don't see the point of slug if it is not something made apparent to the user? What is the benefit?

fiftyone/core/dataset.py
benjaminpkane: Can you clarify? What is the issue?

fiftyone/server/events.py
benjaminpkane: It will come from the URL but will have already been decoded via [decodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent). The server only needs to know the name.

kaixi-wang:
haven't gotten a chance to test this branch and so not sure if your changes affect this, but if the view changes after someone loads a saved view (ie add/remove items from the view bar), it is no longer a saved view so the view selector should show unsaved view and the view param should be removed from the url.

fiftyone/server/view.py
benjaminpkane: I don't think the function is used anywhere. Can I remove?

fiftyone/server/events.py
benjaminpkane: This is a dataclass. It is defined [here](https://github.com/voxel51/fiftyone/blob/43a8752b3c48e86c7c9da8e9be05173ddf084adc/fiftyone/core/session/events.py#L98)

benjaminpkane:
I added encoding/decoding of view in the URL. Still need to test edge cases and notebooks in the morning.

app/packages/relay/src/mutations/setView.ts
```diff
@@ -6,8 +6,6 @@ export default graphql`
     $session: String
     $view: BSONArray!
     $viewName: String
-    $savedViewSlug: String
-    $changingSavedView: Boolean
```
manivoxel51, line 10: 👍 much simpler without these two

fiftyone/server/view.py
kaixi-wang: yeah it's no longer used because directly querying the db like this isn't more efficient that iterating on the saved views property on the dataset singleton. You can remove it.

manivoxel51 (approved):
Thank you for your work @benjaminpkane ! 💯 

I am testing this locally

app/packages/core/src/Dataset.ts
```diff
@@ -8,13 +8,33 @@ import {
...
+  fragment DatasetSavedViewsFragment on Query
+  @refetchable(queryName: "DatasetSavedViewsFragmentQuery") {
+    savedViews(datasetName: $name) {
+      id
+      datasetId
+      name
+      slug
+      description
+      color
+      viewStages
```
kaixi-wang, line 21: Since we don't actually set the view using `viewStages`, we can probably remove this from the fragment to avoid passing around potentially really long stage definitions that are not viewed.

app/packages/state/src/utils.ts
```diff
@@ -118,11 +119,11 @@ export type ResponseFrom<TQuery extends { response: unknown }> =
 export const getSavedViewName = (context: RoutingContext<any>): string => {
   const datasetName = getDatasetName(context);
   if (datasetName) {
-    const queryString = window.location.search;
+    const queryString = context.history.location.search;
     const params = new URLSearchParams(queryString);
     const viewName = params.get("view");
     if (viewName) {
-      return viewName;
+      return decodeURIComponent(viewName);
```
kaixi-wang, line 126: is `getSavedViewName` used where the name specifically (not the slug) is required? If so, then this might not work. For example if `viewName = "saved view"`, then `slug = "saved-view"`, and`decodeURIComponent(viewName)==slug` but `decodeURIComponent(viewName)!=viewName`

manivoxel51 (commented):
The selection doesn't seem to change when you click on them (even though the url is changing)

Screen Shot 2023-01-12 at 11 34 35 AM

manivoxel51 (commented):
tested this in browser, desktop, notebook and found 2 issues. the first issue seem to be related to slug vs. name changes and the fix is 

- The view selection is always changing to default - even if after you select a view

https://user-images.githubusercontent.com/109545780/212212119-c72bd2e0-ea3f-48f8-bcbd-181f8f721ae0.mov

I found that the issue is savedViewParam is being set to view's name/label where the equality to find the selected is based on slug so label !== slug nothing selected. note this works for single worded views

- Accessing a saved view directly is broken via view=VIEW_SLUG. might be related to slug changes too

https://user-images.githubusercontent.com/109545780/212212310-edcdfc06-ecc6-4ab8-803d-7f2e23f1e795.mov

no solution yet

app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
```diff
@@ -51,71 +45,63 @@ export type DatasetViewOption = Pick<
...
-  const fragments = usePreloadedQuery(DatasetSavedViewsQuery, queryRef);
-  const [data, refetch] = useRefetchableFragment(
-    DatasetSavedViewsFragment,
-    fragments
-  );
+  const fragmentRef = useContext(DatasetQueryRef);
 
-  const items =
-    (data as { savedViews: fos.State.SavedView[] })?.savedViews || [];
+  if (!fragmentRef) throw new Error("ref not defined");
```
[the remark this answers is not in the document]
benjaminpkane: It wouldn't. But it asserts the type

app/packages/app/src/index.tsx
```diff
@@ -162,7 +145,10 @@ const App: React.FC = ({}) => {
       },
       controller.signal,
       {
-        initializer: getDatasetName(contextRef.current),
+        initializer: {
+          dataset: getDatasetName(contextRef.current),
+          view: getSavedViewName(contextRef.current),
```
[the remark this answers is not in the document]
benjaminpkane: This inspects the URL and initializes the view slug through the event source

app/packages/state/src/utils.ts
benjaminpkane: This method was always assuming the slug, but I will rename.

benjaminpkane:
I fixed the `savedViewParam` issue. Thanks

I am not seeing any other issues, though.

manivoxel51:
> I fixed the `savedViewParam` issue. Thanks
> 
> I am not seeing any other issues, though.

The direct access to the url with view=x is broken in **some** scenarios. the first half of the video shows loading a dataset and immediately trying to load view=A (when view B is loaded), the potentialView does not change and content is still content B (although view param in url is now A) - second half shows when it works and changes the view properly.


https://user-images.githubusercontent.com/109545780/212383269-0c9d9a88-3394-4bd1-9de2-1d6e01dc1d24.mov

benjaminpkane:
This is a slug issue. I'll have to look at this weekend

app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
```diff
@@ -51,71 +45,63 @@ export type DatasetViewOption = Pick<
...
+  const searchData = useMemo(
     () =>
-      filter(
-        viewOptions,
-        ({ id, label, description, slug }: DatasetViewOption) =>
+      viewOptions.filter(
+        ({ id, label, description, slug }) =>
           id === DEFAULT_SELECTED.id ||
-          label.toLowerCase().includes(viewSearch) ||
+          label?.toLowerCase().includes(viewSearch) ||
```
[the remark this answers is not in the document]
kaixi-wang: is label the name?

kaixi-wang (approved):
just tested app, and everything looks good!

PR #2498, opened 2023-01-09
Documenting label conversion utils

brimoor:
Adds a section to the documentation that highlights the label conversion utils available via methods like `Detections.to_segmentations()` and the `fiftyone.utils.labels` module.

Questions on this topic are pretty common in the Slack community, so I figured I'd give us a place to link when fielding these questions in the future.


jacobmarks (approved):
This all looks good. Just two comments, neither of which are really specific to this code, but about this entire page of the user guide.

1. Are we 100% committed to using line numbers in all of the code blocks? There are definitely some scenarios, like when the last line of a method call is just ')' , when it looks a bit weird.
2. This page of the user guide is quite long. Combining this with the fact that the TOC preview on the right side is only 1 level deep, I would not have even been aware there was a subsection on label conversions until I happened upon it. If anything, I think labels could use their own user guide page. But this brings up another point - I'm fairly sure that soon we will have so much interesting content in each section, like tutorials and recipes, for it to be presented in the current structure the docs allow.

brimoor:
> This all looks good. Just two comments, neither of which are really specific to this code, but about this entire page of the user guide.
> 
> 1. Are we 100% committed to using line numbers in all of the code blocks? There are definitely some scenarios, like when the last line of a method call is just ')' , when it looks a bit weird.
> 2. This page of the user guide is quite long. Combining this with the fact that the TOC preview on the right side is only 1 level deep, I would not have even been aware there was a subsection on label conversions until I happened upon it. If anything, I think labels could use their own user guide page. But this brings up another point - I'm fairly sure that soon we will have so much interesting content in each section, like tutorials and recipes, for it to be presented in the current structure the docs allow.

I sometimes omit the line number directive for very short snippets. But in general, I usually do include line numbers, with the idea that someone might want to ask a question about line `n` of the snippet on page X.

WRT the last line of a method being just `)`, it is a bit strange, but I'm just matching the format that our linter [black](https://black.readthedocs.io/en/stable/) outputs.

I agree that many user guide pages are getting pretty long. Let's consider a refactor soon. My hesitation in doing that so far is that it will likely break links to specific sections that users may have in their notes.

PR #2499, opened 2023-01-09
Cleaning for SSR environments

benjaminpkane:
Cleaning PR that ensures `@fiftyone/state` can be imported in node (server) environments


app/packages/looker/src/elements/util.ts
```diff
@@ -57,7 +52,8 @@ export function createElementsTree<
   update: StateUpdate<State>,
   dispatchEvent: (eventType: string, details?: any) => void
 ): Element {
-  const element = new root.node(config, update, dispatchEvent);
+  const element = new root.node();
+  element.boot(config, update, dispatchEvent);
```
ritch, line 56: Whats the point of this change?

app/packages/plugins/src/index.ts
```diff
@@ -15,11 +15,13 @@ declare global {
   }
 }
 
-// required for plugins to use the same instance of React
-window.React = React;
-window.ReactDOM = ReactDOM;
-window.recoil = recoil;
-window.__fos__ = fos;
+if (typeof window !== "undefined") {
```
ritch, line 18: Unrelated to PR, but something I plan to do is add `components` here as well, hopefully thats compatible with SSR

app/packages/relay/src/resolve.ts
```diff
@@ -0,0 +1,9 @@
+export default <T extends unknown>(module: T): T => {
+  // @ts-ignore
```
ritch, line 2: For some of these random fixes can you add a comment so we remember why these things exist?

ritch (approved):

app/packages/dataset/package.json
```diff
@@ -19,21 +19,13 @@
         "README.md"
     ],
     "dependencies": {
-        "@fiftyone/core": "*"
+        "@fiftyone/core": "*",
+        "framer-motion": "^6.2.8"
```
sashankaryal: nit: is this a peer dependency for one of the libs, because I don't see it being imported anywhere

app/packages/relay/src/resolve.ts
sashankaryal: +1

sashankaryal (approved):
looks good

app/packages/looker/src/elements/util.ts
benjaminpkane: There is discrepancy between how transpilation happens between Teams and OSS. There may be a setting that avoids this, but I don't know of it.

The problem is with classes and property initialization. Perhaps this was an antipattern to begin with, but a subclass that does not define its own constructor will have property initialization happen after parent construction.
```ts
abstract class A {

  constructor() {
    this.init();
  }

  abstract init(): void
}

class B extends A {
  private value: number;

  init() {
    this.value = 0;
  }
}
  
if (B().value !== 0) throw new Error("woe is me")
```

app/packages/dataset/package.json
benjaminpkane: Hmm, I'm not sure it's showing this change. I don't see it locally.

PR #2505, opened 2023-01-10
ViewSelection and its subcomponents capable of rendering readonly or editable versions

manivoxel51:
This pr enables passing a canEdit property to ViewSelection. if true, this property would make ViewSelection an editable component. if false, disables the "save new view" and hides edit/delete options.

the changes for teams are in kacey's branch

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
+++ b/app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
@@ -67,7 +67,7 @@ export default function ViewSelection(props: Props) {
   const [selected, setSelected] = useRecoilState<DatasetViewOption | null>(
     selectedSavedViewState
   );
-
+  const canEditSavedViews = useRecoilValue<boolean>(fos.canEditSavedViews);
   const existingQueries = qs.parse(location.search, {
     ignoreQueryPrefix: true,
   });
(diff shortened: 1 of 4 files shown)


app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
```diff
@@ -206,6 +206,9 @@ export default function ViewSelection(props: Props) {
 
   useEffect(() => {
     const callback = (event: KeyboardEvent) => {
+      if (!canEditSavedViews) {
```
manivoxel51, line 209: disables keyboard shortcut

app/packages/dataset/src/Dataset.tsx
```diff
@@ -59,6 +59,7 @@ export interface DatasetProps {
   readOnly?: boolean;
   theme?: "dark" | "light";
   toggleHeaders?: () => void;
+  canEditSavedViews?: boolean;
```
manivoxel51, line 62: this file is not used but I am keeping it up to date with the one in teams in case

app/packages/state/src/recoil/atoms.ts
```diff
@@ -240,6 +240,11 @@ export const theme = atom<"dark" | "light">({
   ],
 });
 
+export const canEditSavedViews = atom({
+  key: "canEditSavedViews",
+  default: true,
```
manivoxel51, line 245: this will always be true in oss

kaixi-wang: how would this get set in teams? Instead of defaulting to true regardless, can we check client type is oss, then default to true?

kaixi-wang (approved):
Discussed offline and if we can gate the default by checking the config for teams, that would be great. However, it's 100% safe in oss and the default should be always overwritten in teams beforehand so looks good to me. Also, it's a low risk item even if exposed.

PR #2510, opened 2023-01-12
session support for spaces

imanjra:

(Details)

        notes for FiftyOne users.

-   [ ] Other


app/packages/core/src/components/SamplesContainer.tsx
```diff
@@ -134,27 +138,28 @@ function SamplesContainer() {
...
+  }, [spaces]);
 
   return (
     <Container>
-      {showSidebar && <Sidebar render={renderGridEntry} modal={false} />}
-      <SpacesRoot
-        id={`primary-${datasetName}`}
-        defaultState={defaultSpacesState}
-      />
+      {showSidebar && <Sidebar render={renderGridEntry} modal={false} />}{" "}
```
benjaminpkane, line 161: Assuming the quotes were unintentional
```suggestion
      {showSidebar && <Sidebar render={renderGridEntry} modal={false} />
```

app/packages/spaces/src/AppModules.tsx
```diff
@@ -113,6 +114,11 @@ export function useOutsideClick() {
   // do nothing
 }
 
+export function useDimensions() {
+  const ref = useRef();
+  return { ref };
+}
+
```
benjaminpkane, line 121: This feels like trivial code. Is it necessary?

app/packages/core/src/components/SamplesContainer.tsx
```diff
@@ -134,27 +138,28 @@ function SamplesContainer() {
...
+    if (!isMounted.current) {
+      isMounted.current = true;
+      return;
+    }
+    const serializedSpaces = spaces.toJSON();
+    if (!spaces.equals(sessionSpaces) && !spaces.equals(oldSpaces.current)) {
+      setSessionSpaces(serializedSpaces);
+    }
+    oldSpaces.current = serializedSpaces;
+  }, [spaces]);
```
benjaminpkane, line 157: I think the solution here is to move this update into `useStateUpdate` so updates are atomic. If it works, then this is ok but I think it would be good to leave a todo or make an issue.

fiftyone/server/mutation.py
```diff
@@ -400,3 +402,15 @@ async def update_saved_view(
             ),
             None,
         )
+
+    @gql.mutation
+    async def set_spaces(
+        self,
+        subscription: str,
+        session: t.Optional[str],
+        spaces: JSON,
```
benjaminpkane, line 411: This should be `BSON` if object ids need to be preserved

app/packages/core/src/components/SamplesContainer.tsx
imanjra: Yah, I believe formatter likely added those. Will remove

app/packages/spaces/src/AppModules.tsx
imanjra: This is for mock app of spaces. Mock spaces does not import actual module, instead it mocks them in this file. Can be ignored

app/packages/core/src/components/SamplesContainer.tsx
imanjra: The spaces package does not expose directly expose state atoms/selectors. It only exposes the hook which can be used to update the state. I attempted to use those hooks however it did not like that

imanjra:
Closing in favour of #2524

PR #2511, opened 2023-01-12
session support for spaces state

imanjra:

(Details)

        notes for FiftyOne users.

-   [ ] Other


fiftyone/core/state.py
```diff
@@ -149,6 +149,11 @@ def from_dict(cls, d, with_config=None):
 
         fo.config.timezone = d.get("config", {}).get("timezone", None)
 
+        spaces = d.get("spaces", None)
+
+        if spaces is not None:
+            spaces = Space.from_json(spaces)
```
brimoor, line 155: I added this so that `session.spaces` was available as a `Space` object rather than a string in Python

fiftyone/core/spaces.py
```diff
@@ -16,9 +23,11 @@ def _validate_children(children):
 
 
 class Component(EmbeddedDocument):
+    """Base class for App components."""
 
     meta = {"abstract": True, "allow_inheritance": True}
 
+    # @todo remove
```
brimoor, line 30: We can just remove this `id` prop, right? Deprecated by `component_id` I believe

fiftyone/core/state.py
imanjra: This is fixed in newest commit of PR https://github.com/voxel51/fiftyone/pull/2510
It's a  little different. `spaces` is coming as a string so I had to parse to JSON first

fiftyone/core/spaces.py
imanjra: Yes, will remove 👍

imanjra:
Closing in favour of #2524

PR #2513, opened 2023-01-12
Adding support for declaring dynamic fields on generated views via set_values()

brimoor:
It should ideally possible to declare a new dynamic label attribute by applying `set_values()` to a generated view (eg patches). Previously the values themselves would be added, but the field wouldn't be added to the underlying dataset's schema. This PR adds the schema piece.

```py
import fiftyone as fo
import fiftyone.zoo as foz

patches.set_values("ground_truth.foo", ["bar"] * len(patches), dynamic=True)

# previously False, now True
assert "ground_truth.detections.foo" in dataset.get_field_schema(flat=True)
```

### DIFF
--- a/fiftyone/core/clips.py
+++ b/fiftyone/core/clips.py
@@ -378,7 +378,7 @@ class ClipsView(fov.DatasetView):
         _, label_root = self._get_label_field_path(root)
         leaf = path[len(label_root) + 1 :]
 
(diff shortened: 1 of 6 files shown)


swheaton (approved):
Looks good if clips, patches, and video are the only view types impacted 👍🏼

PR #2515, opened 2023-01-13
fix tooltip bug in segmentation overlay

### LINKED ISSUE #2514: [BUG] Semantic segmentation tooltips do not reliably render for RGB masks w/o targets
When viewing a dataset with RGB segmentation masks but no mask targets stored, the tooltip does not reliably render for me in the sample modal. Sometimes it shows, usually it does not.

If I declare mask targets, the tooltip always renders, however.

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "coco-2017",
    split="validation",
    label_field="instances",
    max_samples=25,
    only_matching=True,
)

    dataset,
    "instances",
    "segmentations",
    mask_targets=mask_targets,
)

foul.export_segmentations(
)

# If I declare mask targets, the tooltip always renders
# dataset.default_mask_targets = mask_targets

```

sashankaryal:

Fixes #2514

### DIFF
--- a/app/packages/looker/src/worker.ts
+++ b/app/packages/looker/src/worker.ts
@@ -582,7 +582,7 @@ const UPDATE_LABEL = {
 
     const isMaskTargetsEmpty = Object.keys(maskTargets).length === 0;
 
-    const isRgbMaskTargetsCached = isRgbMaskTargets(maskTargets);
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM

PR #2521, opened 2023-01-13
Migrating cheat sheets to a dedicated section

jacobmarks:
Migrates the cheat sheets from the `Recipes` section to a new top-level `Cheat Sheets` section 👏


brimoor (approved):
LGTM

PR #2523, opened 2023-01-14
Unsafe module used in deserialization

openrefactory:
In file: ipc.py, a deserialization operation is performed by the unsafe Pickle Module (https://docs.python.org/3/library/pickle.html). The data that goes through the deserialization process comes from an untrusted source and it can not be trusted to be well formed (https://cwe.mitre.org/data/definitions/502.html). 

The deserialization operation is done with the safe PyYAML module. No fix created.


openrefactory:
It will be great to understand the use case for the Pickle module here. Could you please comment on that?

PR #2525, opened 2023-01-16
Complex YOLO tweaks

brimoor:
This PR should be logically equivalent to https://github.com/voxel51/fiftyone/pull/2333. I'm just updating the implementation to be more idiomatic.

TODO: test with actual feature maps, etc.

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.complex_yolo as fouy

model = foz.load_zoo_model("complex-yolo-v3-torch")

# TODO: compute feature maps

fouy.apply_model(dataset, model, ...)
```

### DIFF
--- a/fiftyone/utils/complex_yolo/__init__.py
+++ b/fiftyone/utils/complex_yolo/__init__.py
@@ -1,4 +1,5 @@
-"""Complex-YOLOv3 model utils.
+"""
+Complex-YOLOv3 model.
 
 | Copyright 2017-2022, Voxel51, Inc.
 | `voxel51.com <https://voxel51.com/>`_
@@ -7,9 +8,9 @@
 import types
 
 from .zoo import (
-    TorchComplexYOLOv3ModelConfig,
-    TorchComplexYOLOv3Model,
     apply_model,
+    TorchComplexYOLOv3Model,
+    TorchComplexYOLOv3ModelConfig,
 )
 
 # This enables Sphinx refs to directly use paths imported here
--- a/fiftyone/utils/complex_yolo/model.py
+++ b/fiftyone/utils/complex_yolo/model.py
@@ -1,50 +1,225 @@
 """
-Complex-YOLOv3 model adapted from https://github.com/ghimiredhikura/Complex-YOLOv3.
+Complex-YOLOv3 model adapted from
+https://github.com/ghimiredhikura/Complex-YOLOv3.
 
 | Copyright 2017-2022, Voxel51, Inc.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-from collections import OrderedDict
-from typing import Tuple, Union
-
 import math
 import numpy as np
-from shapely.geometry import Polygon
-import open3d as o3d
 
 import fiftyone.core.utils as fou
 
 fou.ensure_torch()
 import torch
-import torch.nn as nn
-import torch.nn.functional as nnF
-from torch.autograd import Variable
+from torch import nn
+import torch.nn.functional as F
+
+sg = fou.lazy_import(
+    "shapely.geometry", callback=lambda: fou.ensure_package("shapely")
+)
 
 
-class Upsample(nn.Module):
-    def __init__(self, scale_factor, mode="nearest"):
-        super(Upsample, self).__init__()
-        self.scale_factor = scale_factor
-        self.mode = mode
+class ComplexYOLOv3(nn.Module):
+    """Complex-YOLOv3 object detection model."""
 
-    def forward(self, x):
-        x = nnF.interpolate(x, scale_factor=self.scale_factor, mode=self.mode)
-        return x
+    def __init__(self, config_path):
+        super().__init__()
+        self.module_defs = parse_model_config(config_path)
+        self.hyperparams, self.module_list = create_modules(self.module_defs)
+        self.yolo_layers = [
+            layer[0]
+            for layer in self.module_list
+            if hasattr(layer[0], "metrics")
+        ]
+        self.seen = 0
+        self.header_info = np.array([0, 0, 0, self.seen, 0], dtype=np.int32)
 
+    def forward(self, x, targets=None):
+        img_dim = x.shape[2]
+        loss = 0
+        layer_outputs, yolo_outputs = [], []
+        for module_def, module in zip(self.module_defs, self.module_list):
+            if module_def["type"] in ["convolutional", "upsample", "maxpool"]:
+                x = module(x)
+            elif module_def["type"] == "route":
+                x = torch.cat(
+                    [
+                        layer_outputs[int(layer_i)]
+                        for layer_i in module_def["layers"].split(",")
+                    ],
+                    1,
+                )
+            elif module_def["type"] == "shortcut":
+                layer_i = int(module_def["from"])
+                x = layer_outputs[-1] + layer_outputs[layer_i]
+            elif module_def["type"] == "yolo":
+                x, layer_loss = module[0](x, targets, img_dim)
+                loss += layer_loss
+                yolo_outputs.append(x)
 
-class EmptyLayer(nn.Module):
-    """Placeholder for 'route' and 'shortcut' layers"""
+            layer_outputs.append(x)
 
-    def __init__(self):
-        super(EmptyLayer, self).__init__()
+        yolo_outputs = (torch.cat(yolo_outputs, 1)).detach().cpu()
+        return yolo_outputs if targets is None else (loss, yolo_outputs)
+
+    def _get_box_corners(self, x, y, w, l, yaw):
+        corners = np.zeros((4, 2), dtype=np.float32)
+        sy2, cy2 = np.sin(yaw) / 2, np.cos(yaw) / 2
+
+        # # front left
+        corners[0, 0] = x - w * cy2 - l * sy2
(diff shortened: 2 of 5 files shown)


fiftyone/utils/complex_yolo/__init__.py
```diff
@@ -1,4 +1,5 @@
-"""Complex-YOLOv3 model utils.
+"""
+Complex-YOLOv3 model.
```
brimoor, line 2: nit: other module docstrings start on newline

fiftyone/utils/complex_yolo/__init__.py
```diff
@@ -7,9 +8,9 @@
 import types
 
 from .zoo import (
-    TorchComplexYOLOv3ModelConfig,
-    TorchComplexYOLOv3Model,
     apply_model,
+    TorchComplexYOLOv3Model,
```
brimoor, line 12: nit: alphabetize imports

fiftyone/utils/complex_yolo/model.py
```diff
@@ -1,50 +1,225 @@
...
-import open3d as o3d
 
 import fiftyone.core.utils as fou
 
 fou.ensure_torch()
 import torch
-import torch.nn as nn
-import torch.nn.functional as nnF
-from torch.autograd import Variable
+from torch import nn
```
brimoor, line 16: removed unused imports, etc

fiftyone/utils/complex_yolo/model.py
```diff
@@ -1,50 +1,225 @@
...
+import torch.nn.functional as F
 
+sg = fou.lazy_import(
```
brimoor, line 19: By convention, when we write utils that use modules that aren't included as default dependencies of `fiftyone` package, we use the pattern below so that the user will get an informative error message if they don't have `shapely` installed.

You can see this in use in other parts of the codebase.

fiftyone/utils/complex_yolo/model.py
```diff
@@ -1,50 +1,225 @@
...
-class Upsample(nn.Module):
-    def __init__(self, scale_factor, mode="nearest"):
-        super(Upsample, self).__init__()
-        self.scale_factor = scale_factor
-        self.mode = mode
 
-    def forward(self, x):
-        x = nnF.interpolate(x, scale_factor=self.scale_factor, mode=self.mode)
-        return x
+class ComplexYOLOv3(nn.Module):
```
brimoor, line 24: nit: I think putting the "important" classes and functions at the top of the module is helpful to readers trying to understand why the module exists and what it does

fiftyone/utils/complex_yolo/model.py
```diff
@@ -1,50 +1,225 @@
...
+                keep_boxes += [detections[0]]
+                detections = detections[~invalid]
+
+            if keep_boxes:
+                output[image_i] = torch.stack(keep_boxes)
+
+        return output
+
+    def compute_height(
+        self, predictions, classes, class_heights=None, class_zs=None
```
brimoor, line 187: I updated this method's signature so that `class*` are explicitly passed in as arguments. Previously some private attributes of the instance were being secretly set by the user of the model, which is confusing for someone trying to figure out how the model functions in general and what this method does in particular.

fiftyone/utils/complex_yolo/zoo.py
```diff
@@ -7,34 +7,65 @@
...
-import numpy as np
-import open3d as o3d
 import torch
+from torch.autograd import Variable
 
-device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
-Tensor = (
-    torch.cuda.FloatTensor if torch.cuda.is_available() else torch.FloatTensor
-)
+o3d = fou.lazy_import("open3d", callback=lambda: fou.ensure_import("open3d"))
```
brimoor, line 25: Same as `shapely`; `open3d` isn't installed by default

fiftyone/utils/complex_yolo/zoo.py
```diff
@@ -7,34 +7,65 @@
...
 
-from .model import build_model
+from .model import ComplexYOLOv3
 
 
 logger = logging.getLogger(__name__)
 
 
+def apply_model(
```
brimoor, line 33: I updated this method's signature and the implementation of `predict_all()` below to work on non-grouped collections, because in future user should be able to directly pass a point cloud-only collection with feature maps stored in a supporting field.

fiftyone/utils/complex_yolo/zoo.py
```diff
@@ -7,34 +7,65 @@
...
+    samples,
+    model,
+    feature_map_field,
+    feature_group_slice=None,
```
brimoor, line 37: There were some hard-coded slice/field names in the previous implementation. When adding a generic utility that is intended to work on any user data, it is better to require the user to provide all relevant parameter names, since they could have picked anything.

There is an existing convention that `label_field="predictions"` is used by default in other parts of the code, which is why I've left it here.

fiftyone/utils/complex_yolo/zoo.py
```diff
@@ -44,250 +75,200 @@ class TorchComplexYOLOv3ModelConfig(
...
-        classes (None): an optional list of custom classes to use for zero-shot
-            prediction
+        nms_thresh (0.3) a non-maximum suppression threshold to use
     """
 
     def __init__(self, d):
         d = self.init(d)
         super().__init__(d)
 
+        self.config_url = self.parse_string(d, "config_url")
```
brimoor, line 89: Since you're subclassing from `TorchImageModel`, it was possible to reduce that amount of stuff explicitly declared here and instead rely on subclass implementation to parse the items.

fiftyone/utils/complex_yolo/zoo.py
```diff
@@ -44,250 +75,200 @@ class TorchComplexYOLOv3ModelConfig(
...
     Args:
         config: a :class:`TorchComplexYOLOv3ModelConfig`
     """
 
-    def __init__(self, config, args=dict()):
-        self._device = device
-        self.img_size = args.get("img_size", 608)
-        self.nms_thresh = args.get("nms_thresh", 0.3)
-        self.conf_thresh = args.get("conf_thresh", 0.9)
+    def __init__(self, config):
```
brimoor, line 117: `config` should define all parameters that the model can accept. No additional `args` here.

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -1742,31 +1742,28 @@
...
+            "base_filename": "yolov3_ckpt_epoch-298.pth",
             "version": null,
             "description": "Complex extension of YOLO object detection model <https://arxiv.org/abs/1803.06199>`_ ",
             "source": "https://github.com/ghimiredhikura/Complex-YOLOv3",
             "size_bytes": 353976522,
             "manager": {
                 "type": "fiftyone.core.models.ModelManager",
                 "config": {
-                    "url": "https://github.com/ghimiredhikura/Complex-YOLOv3/blob/master/config/complex_yolov3.cfg"
+                    "google_drive_id": "1e7PCqeV3tS68KtBIUbX34Uhy81XnsYZV"
```
brimoor, line 1753: By convention, `manager.config` is used to download the actual model weights.

Since we need to download an additional file in this case, which the existing infrastructure unfortunately doesn't support, we do still need to include `config_url` and `config_base_fiename` in the `default_deployment_config_dict.config`.

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -1742,31 +1742,28 @@
...
                     "config_base_filename": "complex_yolov3.cfg",
-                    "classes": "Car, Pedestrian, Cyclist",
                     "entrypoint_fcn": "",
-                    "output_processor_cls": "fiftyone.utils.torch.DetectorOutputProcessor",
-                    "labels_path": "{{eta-resources}}/ms-coco-labels.txt",
-                    "image_size": null,
-                    "image_min_dim": 224,
-                    "embeddings_layer": null
+                    "output_processor_cls": "fiftyone.utils.torch.NullOutputProcessor",
+                    "labels_string": "Car,Pedestrian,Cyclist",
```
brimoor, line 1763: Better to reuse the existing parameters declared by `TorchImageModelConfig` that we're subclassing from here.

fiftyone/zoo/models/manifest-torch.json
brimoor: For context, there are methods like `fiftyone.zoo.models. is_zoo_model_downloaded ()` which rely on using `manager.config` to download the model in order to correctly report whether the model is downloaded or not.

Ideally `is_zoo_model_downloaded() == True` would guarantee that no internet connection is required to use the model. But, until we get there, it's nice to let the builtin functionality handle the big part (model) and handle the small part (cfg) separately.

fiftyone/zoo/models/manifest-torch.json
brimoor: I added `NullOutputProcessor` to make it a bit more explicit that our subclass doesn't actually use the `OutputProcessor` interface as it is currently implemented.

We could put the output processing logic *into* a new `OutputProcessor` subclass, but this model doesn't use the builtin `predict_all()` implementation, so there's not much value in following the `OutputProcessor` interface currently.

PR #2527, opened 2023-01-16
Second pass over FiftyOne Teams docs

ehofesmann:
Reworked a few pages of the FiftyOne Teams docs. Specifically:

* Removed the old installation instructions with a higher level description and a note about getting more info after signing up
* Moved the cloud credential configuration to the install section
* Added lots of images everywhere
* Reformatted the FO vs Teams table
* Added more in depth descriptions for dataset permissions
* Gave more examples of Teams-specific cloud methods


docs/source/teams/install_overview.rst
```diff
@@ -0,0 +1,220 @@
...
+    necessary components, are made available to you by the Voxel51 team during the
+    onboarding process.
+
+
+.. _teams-python-sdk:
+
+Python SDK
+----------
+
+While the :ref:`FiftyOne Teams App <team-app>` allows for countless new App-centric
```
jacobmarks, line 31: I changed `team-app` to `teams-app`, which is what it should be I think

docs/source/teams/cloud_credentials.rst
```diff
@@ -1,168 +1 @@
-.. _cloud-credentials:
```
jacobmarks, line 1: I deleted this file, as I see you've moved this info to another place

docs/source/teams/overview.rst
```diff
@@ -11,60 +11,82 @@ FiftyOne Teams is designed to integrate naturally into your ML workflow. It inte
...
-       | private/public cloud
-     - ☑
-     - ☑
-     - ☑
-     - ☑
-     - ☑
-     - ☑
-     - | Unlimited data, flexible
-       | user-based licensing
+.. raw:: html
```
jacobmarks, line 14: This table is so much nicer! Thanks for improving the look and feel dramatically!

docs/source/teams/cloud_media.rst
```diff
@@ -37,6 +42,76 @@ You can get information about currently cached media files for a sample collecti
...
+
+* ``fiftyone.core.storage.upload_media(sample_collection,remote_dir,rel_dir=None,media_field="filepath",update_filepaths=False,overwrite=False,skip_failures=False,progress=False)``
+
+
+.. note::
+
+   Pro tip: Learn more about these and other FiftyOne methods by entering them
+   into an IPython cell followed by a question mark:
+
+   .. code-block:: python
```
jacobmarks, line 65: The docs build complains here because of the question mark (at least for me) and doesn't highlight the syntax as it should

docs/source/teams/cloud_media.rst
```diff
@@ -37,6 +42,76 @@ You can get information about currently cached media files for a sample collecti
...
+
+        import fiftyone.core.storage as fos
+
+        fos.upload_media?
+
+   .. code-block:: shell
```
jacobmarks, line 71: The docs build also complains for me here, I think because of the '``' in the block.

jacobmarks (changes requested):
Aside from the 4 notes (team-app --> teams-app, plus two syntax highlighting issues, and getting rid of the cloud_credentials file) it looks good to me! Thanks for taking a pass through all of this, and especially the images and table overhaul!

docs/source/teams/cloud_media.rst
ehofesmann: Hmm, yeah good catch. I'm not sure what can be done here. I'll just remove these two syntax notes and leave it without highlighting

PR #2397, opened 2022-12-06
Fix databricks integration

benjaminpkane:
Resolves #2197 

Testing requires a databricks account.

```py
!pip install fiftyone==0.18.1rc26

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")
session = fo.launch_app(dataset)
```

PR #2534, opened 2023-01-18
fix bug in fetch 

sashankaryal:
This was a small bug introduced by #2397 that caused fetch to not work in web workers.

Locally, on multiple datasets.

### DIFF
--- a/app/packages/utilities/src/fetch.ts
+++ b/app/packages/utilities/src/fetch.ts
@@ -33,24 +33,24 @@ export const getFetchHeaders = () => {
 
 export const getFetchOrigin = () => {
   // window is not defined in the web worker
-  if (typeof window !== "undefined" && window.FIFTYONE_SERVER_ADDRESS) {
+  if (hasWindow && window.FIFTYONE_SERVER_ADDRESS) {
     return window.FIFTYONE_SERVER_ADDRESS;
   }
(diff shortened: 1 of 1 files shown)


app/packages/utilities/src/fetch.ts
```diff
@@ -33,24 +33,24 @@ export const getFetchHeaders = () => {
...
-    typeof window !== "undefined" &&
-    typeof window.FIFTYONE_SERVER_PATH_PREFIX === "string"
-  ) {
+  if (hasWindow && typeof window.FIFTYONE_SERVER_PATH_PREFIX === "string") {
     return window.FIFTYONE_SERVER_PATH_PREFIX;
   }
 
-  return (
-    new URL(window.location.toString()).searchParams.get("proxy") ||
-    fetchPathPrefix
```
kaixi-wang, line 52: Is fetchPathPrefix set anywhere/is `fetchPathPrefix === ""` always?

sashankaryal: https://github.com/voxel51/fiftyone/blob/develop/app/packages/utilities/src/fetch.ts#L71

It depends on if it's a worker thread requesting fetch or main thread.

lanzhenw (approved):

PR #2537, opened 2023-01-19
Maintain Starlette route list

benjaminpkane:
For Teams routing, the OSS routes must be represented as a flat list of tuple definitions for proper construction.


imanjra (approved):
Tested changes locally and the in-app embeddings panel works as before. LGTM

PR #2539, opened 2023-01-19
update select sample with select label under condition

### LINKED ISSUE #2541: [BUG] tag sample bug in modal view when labels are selected
After opening the sample modal, click on a label, then try to tag the sample, it will apply the tag to all samples in the dataset.

image (10)
image (11)

The expected behavior should be the same as when no label is selected. 

1. Please fill in this template to ensure a timely and thorough response
2. Place an "x" between the brackets next to an option if it applies. For example:
    -   [x] Selected option
3. **Please delete everything above this line before submitting the issue**

### Commands to reproduce

As thoroughly as possible, please provide the Python and/or shell commands used to encounter the issue. Application steps can be described in the next section.

```
# commands here
```

### Describe the problem

Describe the problem clearly here. Include descriptions of the expected behavior and the actual behavior.

### Code to reproduce issue

Provide a reproducible test case that is the bare minimum necessary to generate the problem.

### Other info/logs

Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached. Please do not use screenshots for sharing text. Code snippets should be used instead when providing tracebacks, logs, etc.

-   [ ] `Core`: Core Python library issue

### Willingness to contribute

lanzhenw:
fixes #2541 

As a user, if I did not select any sample and I go into the sample modal view, then when I select labels on the sample, this sample will be added to the selected samples. This way, when I click add sample tag, the tag applies to the current sample, instead of all the samples in the dataset. 

Note: Select labels will not add current sample to selected sample, if the user already selected some samples. It only applies if selectedsample is a empty set. 

https://user-images.githubusercontent.com/17770824/213489763-fd8b088b-c39c-4e7a-bc55-9dd1605d9292.mov

See video.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Actions/ActionsRow.tsx
+++ b/app/packages/core/src/components/Actions/ActionsRow.tsx
@@ -196,7 +196,7 @@ const Selected = ({
 }) => {
   const [open, setOpen] = useState(false);
   const [loading, setLoading] = useState(false);
-  const [samples, setSamples] = useRecoilState(fos.selectedSamples);
+  const samples = useRecoilValue(fos.selectedSamples);
   const labels = useRecoilValue(fos.selectedLabelIds);
   const ref = useRef();
   useOutsideClick(ref, () => open && setOpen(false));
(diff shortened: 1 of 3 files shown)


brimoor:
@lanzhenw as discussed offline- please update so that the same behavior occurs (tagging a sample from the modal only applies to the current sample), but without explicitly checking the sample checkbox in the UI.

For context, if I open the modal with nothing selected at all, then I open the tag sample modal, any sample tag changes I perform are only applied to the current sample, without the sample's checkbox being selected.

brimoor (approved):
LGTM, works great! 🎉

PR #2545, opened 2023-01-19
Always run and upload reports

benjaminpkane:
Ensures both Python and App reports are uploaded and coverage deltas are accurate


brimoor (approved):
This somehow increases code coverage, but the CI passes now so LGTM(?)

PR #2546, opened 2023-01-19
[HOLD] Use dataset slug in App URL

kaixi-wang:

- Use the dataset's slug in the URL 
- Enable loading dataset by name or slug on the backend
- make dataset name required for `delete_saved_view` mutations
- change default of `SidebarGroup`/`SidebarGroupInput` gql types to `True` to match the schema

(Addresses issue in https://voxel51.atlassian.net/browse/TEAMS-386)

https://user-images.githubusercontent.com/30936736/213582497-d12dc8f7-72a2-4491-82a9-eec6d4564516.mov

Tested in app

## Release Notes

 - enable easy navigation to a dataset with spaces and other special characters in the name

-   [ ] Other


app/schema.graphql
```diff
@@ -330,7 +331,7 @@ type Mutation {
     subscription: String!
     session: String
     viewName: String!
-    datasetName: String
+    datasetName: String = null
```
brimoor, line 334: Shouldn't this be required?

app/schema.graphql
```diff
@@ -492,7 +493,7 @@ type SidebarGroup {
 input SidebarGroupInput {
   name: String!
   paths: [String!]
-  expanded: Boolean = true
+  expanded: Boolean = null
```
brimoor, line 496: Why was this changed?

fiftyone/core/dataset.py
```diff
@@ -6391,7 +6391,7 @@ def _load_clips_source_dataset(frame_collection_name):
     return load_dataset(doc["name"])
 
 
-def _load_dataset(obj, name, virtual=False):
+def _load_dataset(obj, name, virtual=False, slug=False):
```
brimoor, line 6394: I **don't** think we should allow datasets to be loaded by slug in the Python SDK.

There are quite a few places where dataset names are used as identifiers of datasets (eg listing, deleting) not captured here, so this change is incomplete, but regardless I don't see any benefit to allowing slug-based access; it will likely just lead to confusion.

Slug is not currently a user-facing concept, its just an internal way to generate human readable URLs.

brimoor (changes requested):

app/schema.graphql
kaixi-wang: I didn't change this. The schema is autogenerated from the code. Do you want me to change the mutation?

app/schema.graphql
kaixi-wang: I didn't change this. The schema is autogenerated from the code...

app/schema.graphql
kaixi-wang: right now it's technically not required input because we can just read it from the state

fiftyone/core/dataset.py
kaixi-wang: I wasn't planning on exposing it and forgot to remove this arg. 

Just to clarify, do you mean you don't want to include `slug` in the `find` query? Not sure I understand the reasoning why using slugs in the query would be problematic since slugs have to be unique so the lookup should only result in 1 dataset. If not, there is a problem elsewhere.

Alternatively, would you prefer me to write a separate function that first queries the db for a dataset name given a slug? And then load the dataset with that name? The issue is that given a direct link/slug to a dataset, there is no way of deriving the name without an additional lookup since the relationship between unique slugs and unique names is 1: many.

app/schema.graphql
brimoor: I'm just confused why this is appearing in this PR

app/schema.graphql
benjaminpkane: The default should be null. It is correct. It was never regenerated during sidebar updates

app/schema.graphql
benjaminpkane: This should be required. We shouldn't assume state in mutations.

app/packages/state/src/utils.ts
```diff
@@ -106,30 +106,30 @@ export const transformDataset = (dataset: any): Readonly<State.Dataset> => {
...
+      path: "/datasets/:slug",
       exact: true,
     },
     {},
     ""
   );
-
-  if (result) {
-    return decodeURIComponent(result.variables.name);
+  const slug = result?.variables?.slug;
```
benjaminpkane, line 119: This is guaranteed to be defined, no?

benjaminpkane:
If we are going to continue with this slug model, a lot of this will be simplified by adding a client side `toSlug` method.

app/schema.graphql
kaixi-wang: ok I'll make it required

@brimoor With no code change, the autogenerated schema would only be different if someone manually changed this file or changed the mutation without regenerating the schema

app/packages/state/src/utils.ts
kaixi-wang: yeah just took a look at `matchPath`

I'll just do the `if(result)` like before

kaixi-wang:
> If we are going to continue with this slug model, a lot of this will be simplified by adding a client side `toSlug` method.

not sure that the `toSlug` method would help since the slug is a field of the Dataset gql object. Plus, having the function defined in multiple places (frontend and backend) makes it possible and really easy to have bugs later on if someone updates one and not the other

brimoor:
@benjaminpkane @kaixi-wang let's talk offline about best practices for using names and slugs in interfaces.

PR #2550, opened 2023-01-20
Cleaning for builds

benjaminpkane:
Cleaning some code for builds to work in all environments


ritch (approved):

app/packages/core/src/components/ViewBar/ViewStage/ErrorMessage.tsx
```diff
@@ -1,9 +1,7 @@
 import { ReportProblem } from "@mui/icons-material";
 import { animated, config, useSpring } from "@react-spring/web";
 import { useService } from "@xstate/react";
-import React, { useEffect, useRef, useState } from "react";
```
kaixi-wang, line 4: These are necessary as well

PR #2553, opened 2023-01-20
[Saved views] clean up set_view mutation

kaixi-wang:

- clean up the set_view mutation

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/server/mutation.py
+++ b/fiftyone/server/mutation.py
@@ -186,12 +186,9 @@ class Mutation:
                 doc = ds._get_saved_view_doc(saved_view_slug, slug=True)
                 result_view = ds._load_saved_view_from_doc(doc)
             except:
-                logging.error(
-                    f"[mutation.py] ERROR: Unable to load saved view with "
-                    f"slug={saved_view_slug}"
-                )
+                pass
 
-        if not result_view:
+        if result_view is None:
             # Update current view with form parameters
             result_view = get_view(
                 dataset_name,
--- a/fiftyone/server/mutation.py
+++ b/fiftyone/server/mutation.py
@@ -5,6 +5,7 @@ FiftyOne Server mutations
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+import logging
 from dataclasses import asdict
 import strawberry as gql
 import typing as t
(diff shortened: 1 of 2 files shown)


fiftyone/server/mutation.py
```diff
@@ -178,31 +179,19 @@ async def set_view(
...
-        else:
-            state.view_name = None
-            state.saved_view_slug = None
+                logging.error(
+                    f"[mutation.py] ERROR: Unable to load saved view with "
+                    f"slug={saved_view_slug}"
+                )
 
-        if result_view is None:
+        if not result_view:
```
benjaminpkane, line 194: This will call [`__bool__`](https://github.com/voxel51/fiftyone/blob/7dd227cc52fe5000067c92363d57dca2584f76bd/fiftyone/core/collections.py#L181) which will in turn call `__len__` which is very bad. `is None` is best

fiftyone/server/mutation.py
```diff
@@ -178,31 +179,19 @@ async def set_view(
...
-                result_view = ds.load_saved_view(view_name)
-                view = result_view._serialize()  # serialized view stages
-                # Set view state
-                state.view = result_view
-                state.view_name = result_view.name
-                state.saved_view_slug = saved_view_slug
-        else:
-            state.view_name = None
-            state.saved_view_slug = None
+                logging.error(
```
benjaminpkane, line 189: I'm not sure I see the benefit to this logging. If anything, this should be an [expected error from GraphQL](https://strawberry.rocks/docs/guides/errors#expected-errors).

kaixi-wang: oh I meant for this to be logging.debug so that it can be enabled for developers and may be helpful. I didn't want to trigger an actual error (just wanted to log it in the console) that would surface to the end user.

I'll remove it though

fiftyone/server/mutation.py
kaixi-wang: oh good to know! For other things, sometimes the None specific check lets "" and [] get by, which can trigger errors downstream. However, it's safe here to use None because result_view is declared None at the beginning.

benjaminpkane (approved):
LGTM just be sure to use `is None` for checking views

PR #2555, opened 2023-01-23
Saved view tweaks

benjaminpkane:
Tweaks for embedded saved views

### DIFF
--- a/app/packages/dataset/src/Dataset.tsx
+++ b/app/packages/dataset/src/Dataset.tsx
@@ -56,7 +56,7 @@ export interface DatasetProps {
   theme?: "dark" | "light";
   toggleHeaders?: () => void;
   canEditSavedViews?: boolean;
-  savedViewSlug: string;
+  savedViewSlug?: string;
 }
 
 export const Dataset: React.FC<DatasetProps> = ({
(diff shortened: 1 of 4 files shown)


app/packages/dataset/src/Dataset.tsx
```diff
@@ -25,6 +25,7 @@ import styled from "styled-components";
 // built-in plugins
 import "@fiftyone/looker-3d";
 import "@fiftyone/map";
+import { useRouter } from "next/router";
```
kaixi-wang, line 28: Do we use the next router in oss?

benjaminpkane: No, I'll move

app/packages/dataset/src/Dataset.tsx
```diff
@@ -57,6 +56,7 @@ export interface DatasetProps {
   theme?: "dark" | "light";
   toggleHeaders?: () => void;
   canEditSavedViews?: boolean;
+  savedViewSlug: string;
```
kaixi-wang, line 59: should we make this optional?

benjaminpkane: good point

kaixi-wang (approved):
lgtm

PR #2559, opened 2023-01-24
fix sidebar issue

ritch:
**Related**

 - pull/2466
 - TEAMS-763

**Resolves**

"Rendered fewer hooks than expected" when dataset schema changes. Eg. when removing to_patches view stage or select stage.


brimoor (approved):
LGTM

PR #2568, opened 2023-01-26
Publish syntax fixes

benjaminpkane:
Fixes some syntax issues for workflow version regex when publishing from #2331


brimoor (approved):
LGTM

PR #2569, opened 2023-01-30
Adding support for querying by vectors/prompts

brimoor:
Adds support for two powerful new query options for similarity searches.

## Query by vector

You can now pass a manually-computed query vector to `sort_by_similarity()`:

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

fob.compute_similarity(dataset, model=model, brain_key="clip")

query = model.embed_prompt("kites high in the air")
view = dataset.sort_by_similarity(query, k=10, brain_key="clip")

session.view = view
```

## Query by prompt

You can now pass text prompts directly to `sort_by_similarity()` when the index was generated by a model like CLIP that supports prompt embeddings:

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

fob.compute_similarity(dataset, model="clip-vit-base32-torch", brain_key="clip")

prompt = "kites high in the air"
view = dataset.sort_by_similarity(prompt, k=10, brain_key="clip")

session.view = view
```


jacobmarks (approved):
👍 looks good to me!

swheaton (approved):

fiftyone/core/models.py
```diff
@@ -1958,6 +1958,48 @@ def embed_all(self, args):
...
+class PromptMixin(object):
+    """Mixin for :class:`Model` classes that can generate prompt embeddings.
+
+    This mixin allows for the possibility that only some instances of a class
+    are capable of generating prompt embeddings, per the value of the
+    :meth:`can_embed_prompts` property.
+    """
+
+    @property
+    def can_embed_prompts(self):
```
kaixi-wang, line 1970: `embed_prompts` is kind of misleading. It sound like we are saving the prompts on the model... can we use something like `text_search_enabled`

fiftyone/core/models.py
```diff
@@ -1958,6 +1958,48 @@ def embed_all(self, args):
...
+        """Whether this instance can generate prompt embeddings."""
+        raise NotImplementedError(
+            "subclasses must implement can_embed_prompts"
+        )
+
+    def embed_prompt(self, arg):
```
kaixi-wang, line 1976: similarly, since we aren't actually embedding anything, can we use something along the lines of `vectorize_prompt`, `encode_prompt`,  or `get_text_embeddings`

fiftyone/utils/clip/zoo.py
```diff
@@ -101,13 +128,10 @@ def _load_network(self, config):
 
         return build_model(model.state_dict()).to(self.device).float()
 
-    def _prepare_text_features(self):
+    def _embed_prompts(self, prompts):
```
kaixi-wang, line 131: `_text_features` make more sense than `embed_prompt`. Can we just use the current name?

fiftyone/utils/clip/zoo.py
```diff
@@ -139,13 +163,15 @@ def _prepare_text_features(self):
 
             text_features[i, : len(tokens)] = torch.tensor(tokens)
 
-        return text_features
+        with torch.no_grad():
+            return self._model.encode_text(text_features)
```
kaixi-wang, line 167: Can we just use `encode_text` like the model already does? For users who already have worked with text models previously, `encode_text` immediately allows them to know what the method is doing and is even guessable.

fiftyone/core/models.py
brimoor: I like `embed_prompt()` for consistency with the existing `embed()` method for the images

PR #2580, opened 2023-01-30
Improve no brain results message

### LINKED ISSUE #2573: [FR] Provide more informative message when no visualizations are available in embeddings panel
In the embeddings panel, when I have no brain runs at all, I see a message "No Brain Results Available".

However, if I have a `compute_similarity()` run but no `compute_visualization()` runs, I do not see the message, but the `Brain Result` dropdown is empty (only visualizations are allowed). It would be more helpful if the "No Brain Results Available" message always appeared when the selector is empty.

Additionally, it would be consistent with other cases if we linked to the documentation so the user could learn how to enable this feature. A good message could be:

```
No embeddings visualizations found. [Learn more](https://docs.voxel51.com/user_guide/brain.html#visualizing-embeddings) about using this feature.
```

Screen Shot 2023-01-30 at 10 38 47 AM

ritch:
Fixes #2573

Makes the wording / style of the embeddings visualization verbiage better and more consistent.

### DIFF
--- a/app/packages/components/src/components/ThemeProvider/index.tsx
+++ b/app/packages/components/src/components/ThemeProvider/index.tsx
@@ -101,6 +101,7 @@ let theme = extendJoyTheme({
         },
         primary: {
           main: "hsl(25, 100%, 51%)",
+          mainChannel: "none",
           plainColor: "hsl(25, 100%, 51%)",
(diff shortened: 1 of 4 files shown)


brimoor (approved):
LGTM

PR #2582, opened 2023-01-31
bugfix: update filter labels for listfield

### LINKED ISSUE #2508: [BUG] Add support for list attributes of Label fields in the sample modal
In the example below, I have declared custom list attributes of my `Detection` instances:

```py
import fiftyone as fo
import fiftyone.zoo as foz

filepath = foz.load_zoo_dataset("quickstart", max_samples=1).first().filepath
sample = fo.Sample(
    filepath=filepath,
    ground_truth=fo.Detections(
        detections=[
            fo.Detection(
                label="stop sign",
                bounding_box=[
                  0.20879166666666665,
                  0.18365625000000002,
                  0.5279791666666667,
                  0.42953125,
                ],
                color=["red", "white"],
                language=["english"],
            ),
            fo.Detection(label="foo", bounding_box=[0, 0, 0.1, 0.1]),
        ]
    )
)

dataset = fo.Dataset()
dataset.add_sample(sample, dynamic=True)

```

There are a couple of issues in the App's sample modal when working with these detections:

1. Filtering by a list attribute does not work correctly (boxes don't disappear; counts don't update as expected in sidebar):
Screen Shot 2023-01-11 at 9 55 51 AM

2. List attributes should be shown in the tooltip:
Screen Shot 2023-01-11 at 9 55 41 AM

lanzhenw:
fix  #2508 

1. filter dropdown should always show as long as totalcount is more than 0. 
2. added filter labels for listfields
3. tooltip will display listfields (except for bounding box) in modal view
Screen Shot 2023-01-30 at 11 08 40 PM

locally

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Modal/TooltipInfo.tsx
+++ b/app/packages/core/src/components/Modal/TooltipInfo.tsx
@@ -20,7 +20,7 @@ const TooltipDiv = animated(styled(ContentDiv)`
 const ContentItemDiv = styled.div`
   margin: 0;
   padding: 0;
-  max-width: 10rem;
+  max-width: 12rem;
 `;
 
 const ContentValue = styled.div`
--- a/app/packages/core/src/components/Modal/TooltipInfo.tsx
+++ b/app/packages/core/src/components/Modal/TooltipInfo.tsx
@@ -20,14 +20,16 @@ const TooltipDiv = animated(styled(ContentDiv)`
 const ContentItemDiv = styled.div`
   margin: 0;
   padding: 0;
-  max-width: 12rem;
-  word-wrap: break-word;
+  max-width: 10rem;
 `;
 
 const ContentValue = styled.div`
   font-size: 0.8rem;
   font-weight: bold;
   color: ${({ theme }) => theme.text.primary};
+  text-overflow: ellipsis;
+  overflow: hidden;
+  white-space: nowrap;
 `;
 
 const ContentName = styled.div`
--- a/app/packages/core/src/components/Modal/TooltipInfo.tsx
+++ b/app/packages/core/src/components/Modal/TooltipInfo.tsx
@@ -6,6 +6,7 @@ import styled from "styled-components";
 import * as fos from "@fiftyone/state";
 import { useRecoilValue } from "recoil";
 import { ContentDiv, ContentHeader } from "../utils";
+import { joinStringArray } from '../Filters/utils';
 
 const TooltipDiv = animated(styled(ContentDiv)`
   position: absolute;
@@ -19,7 +20,7 @@ const TooltipDiv = animated(styled(ContentDiv)`
 const ContentItemDiv = styled.div`
   margin: 0;
   padding: 0;
-  max-width: 10rem;
+  max-width: 12rem;
   word-wrap: break-word;
 `;
 
@@ -42,13 +43,14 @@ const ContentItem = ({
   style,
 }: {
   name: string;
-  value?: number | string;
+  value?: number | string | string[];
   style?: object;
 }) => {
-  if (typeof value === "object") {
+  debugger
+  if (typeof value === "object" && !value.length) {
     return null;
   }
-
+  
   return (
     <ContentItemDiv style={style}>
       <ContentValue>
@@ -60,6 +62,8 @@ const ContentItem = ({
               return value.length ? value : '""';
             case "boolean":
               return value ? "True" : "False";
+            case "object":
+              return joinStringArray(value);
             default:
               return "None";
           }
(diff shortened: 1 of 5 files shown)


app/packages/core/src/components/Modal/TooltipInfo.tsx
```diff
@@ -42,13 +43,14 @@ const ContentItem = ({
   style,
 }: {
   name: string;
-  value?: number | string;
+  value?: number | string | string[];
   style?: object;
 }) => {
-  if (typeof value === "object") {
+  debugger
```
brimoor, line 49: remove `debugger`

app/packages/core/src/components/Modal/TooltipInfo.tsx
```diff
@@ -170,7 +174,7 @@ const AttrInfo = ({ label, children = null }) => {
   );
 
   const other = entries.filter(
-    ([name]) => !["label", "confidence"].includes(name)
+    ([name]) => !["label", "confidence", "bounding_box"].includes(name)
```
brimoor, line 177: I think this will need to be generalized based on the `Label` type:
- `Detection`: don't include `bounding_box`/`mask` in tooltip
- `Classification`: don't include `logits` in tooltip
- `Keypoint`: don't include `points` in tooltip
- `Polyline`: don't include `points` in tooltip
- `Segmentation`: don't include `mask` in tooltip
- `Heatmap`: don't include `map` in tooltip

Based on the following rules:
- `bounding_box`/`points` are excluded because this data is already rendered spatially.
- `mask`/`map`/`logits` are excluded because they are binary (array) types, which aren't suitable for rendering (this may not need to be explicitly added, if binary field values are already implicitly excluded; I didn't check the implementation carefully)

brimoor (changes requested):
The behavior all looks great on the test dataset from https://github.com/voxel51/fiftyone/issues/2508!

One issue: I tested with a `Keypoints` field using the code snippet below, and realized that:
- (unique to this PR) the modal crashes when trying to hover over a keypoint in the image visualizer of the sample modal
- (also in `develop`, but probably relevant to the code you're editing here) an error occurs when trying to expand the filter tray for the `keypoints` field

```py
import random
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

keypoints = []
for _ in range(len(dataset)):
    kps = []
    for i in range(10):
        points = np.random.rand(18, 2).tolist()
        kps.append(fo.Keypoint(label=random.choice("abcdef"), points=points))

    keypoints.append(fo.Keypoints(keypoints=kps))

dataset.set_values("keypoints", keypoints)

```

app/packages/core/src/components/Modal/TooltipInfo.tsx
lanzhenw: @brimoor Does `Regression` type also have some label that we do not want to show?

app/packages/core/src/components/Modal/TooltipInfo.tsx
```diff
@@ -157,20 +162,20 @@ const useTarget = (field, target) => {
   return getTarget(field, target);
 };
 
-const AttrInfo = ({ label, children = null }) => {
+const AttrInfo = ({ label, infoType, children = null }) => {
```
brimoor, line 165: would `labelType` be a better name for the `infoType` parameter?

app/packages/core/src/components/Modal/TooltipInfo.tsx
```diff
@@ -157,20 +162,20 @@ const useTarget = (field, target) => {
...
+  const defaultLabels = ["label", "confidence"]
   const defaults = entries.filter(([name]) =>
-    ["label", "confidence"].includes(name)
+    defaultLabels.includes(name)
   );
-
+  
   const other = entries.filter(
-    ([name]) => !["label", "confidence"].includes(name)
+    ([name]) => !([...defaultLabels, ...HIDDENLABELS[infoType], "attributes"].includes(name))
```
brimoor, line 178: Ah I forgot to mention the `attributes` parameter in my first review. Only the `Detection`, `Keypoint` and `Polyline` classes have a builtin `attributes` parameter, so this logic could just be added to `HIDDEN_LABELS` below

app/packages/core/src/components/Modal/TooltipInfo.tsx
```diff
@@ -289,3 +295,13 @@ const OVERLAY_INFO = {
   Regression: RegressionInfo,
   Segmentation: SegmentationInfo,
 };
+
+const HIDDENLABELS = {
```
brimoor, line 299: nit: style guide would say to use `HIDDEN_LABELS` here

app/packages/core/src/components/Modal/TooltipInfo.tsx
brimoor: Ah I see 👍

brimoor (approved):
LGTM

PR #2584, opened 2023-01-31
looker refactor, dynamic pointsize, rgb pointclouds

sashankaryal:

1. **Feature:** dynamic point size controls
2. **Feature:** shader for supporting RGB pointclouds
3. **Readability / organization:** break down Looker3D into smaller modules
4. **Type safety:** address all "squiggles" enforced via ESLint rules
5. **Misc:** small performance improvements and minor bug fixes

### Notes to the reviewer
- There's not yet a clear consensus on UI control for configuring point size. In this implementation, we have a slider with bound values / current value hidden. The selected point size is persisted in local storage.
- For `colorBy` mode, we have `intensity`, `height`, and `none`. In this PR's implementation, when `none` is selected, there are two possibilities. If the PCD file has color attributes, that's what is rendered, if not, the points are rendered in white. **Does it make sense to add yet another option, nominally labeled `default`, in addition to`none`?** Or rename `none` with `default`?
- Pay special attention to the added `useMemo()` hooks, for instance in `fop.state::usePluginSettings()`.
- Pay attention to the `shaders` module. The RGB shader is a new one, but of the two existing ones, I've not changed much except add a `uniform` for point size, and **_updated how `gl_PointSize` was being calculated_**.

Smoke tested locally with a custom dataset with samples that contain either non-RGB pointclouds, or RGB pointclouds. Also tested with `quickstart-groups` dataset.

        notes for FiftyOne users.

-   [ ] Other


ritch:
I think if we are going to invest time in refactoring and making the code more maintainable we really need to add a few test suites for the hooks and shaders. This can come in a follow up PR.

We also need to do some looking into how intensity and default color are working as right now in this branch when I view "none" it actually shows the intensity data. Instead it should show just a static color.

When I choose the color field, and the color data includes the intensity data, its ok that I see that data as encoded. In other words we should support the color channel as described in the PCD spec and the R channel as intensity, but not mixed when possible.

I think we need to rework the color by actions:

	- Rename `Color by` => `Shade by`
	- New choices
		- Color
			- Grayed out if no color data provided
		- Intensity (title="uses the color.r value")
			- Grayed out if there is no color.r 
		- Height
		- None

ritch (approved):
I think this can land as is. If possible please incorporate the color by changes in the comments.

sashankaryal:
Since this is an additive PR and the changes have been verified, I'm going to merge this one and address the issues / comments in the imminent looker-3d PR.

PR #2594, opened 2023-02-01
Do not refresh grid when modal is open

manivoxel51:

The fix is to disable the background grid refresh every-time a tagging happens when the sample modal is open which prevents the extra refresh and also prevents the modal closing unexpectedly.

https://user-images.githubusercontent.com/109545780/216155993-9d17db6a-4758-46f2-a03b-32933bd79a08.mov

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Grid/Grid.tsx
+++ b/app/packages/core/src/components/Grid/Grid.tsx
@@ -99,7 +99,6 @@ const Grid: React.FC<{}> = () => {
       freeVideos();
     }),
     [
-      isModalOpen,
       stringifyObj(useRecoilValue(fos.filters)),
       useRecoilValue(fos.datasetName),
       useRecoilValue(fos.cropToContent(false)),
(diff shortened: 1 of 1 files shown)


brimoor (changes requested):
Thanks for the first pass @manivoxel51.

This does prevent the grid from refreshing, but I'm still seeing two undesirable things:
- The sidebar in the background still refreshes (that can actually be more expensive than the grid loading itself)
- Now the grid always refreshes when the modal is closed

Can you also prevent both of these things?

manivoxel51:
> Thanks for the first pass @manivoxel51.
> 
> This does prevent the grid from refreshing, but I'm still seeing two undesirable things:
> 
> * The sidebar in the background still refreshes (that can actually be more expensive than the grid loading itself)
> * Now the grid always refreshes when the modal is closed
> 
> Can you also prevent both of these things?
> 
> Not automatically refreshing the sidebar definitely presents the potential for users to be confused that tag stats aren't up-to-date, but I think it's something that we should deal with because refreshing the sidebar can be expensive.

we talked offline. for now, we are focused on getting the grid refresh to stop when 
1. the sample modal is open
2. the sample modal closes by either clicking outside of the modal || pressing esc key

future work to discuss with @benjaminpkane is to prevent refreshing the full sidebar but that needs more brainstorming before implementation.

@brimoor could you give this another test when you get a chance plz.

brimoor (approved):
LGTM

PR #2597, opened 2023-02-02
Add support for CVAT v2.4

### LINKED ISSUE #2565: Response format of CVAT job endpoint changed.
When I try to upload my dataset to the CVAT cloud instance, I get an error on the following line:

https://github.com/voxel51/fiftyone/blob/a49fd03ef9d5c98cbdf32f2ee8ac4f62d53c42a5/fiftyone/utils/cvat.py#L4141

The issue is reasonable, as the jobs may be created some time after the import happens (post request). It seems that if you want to request for the job IDs you must wait some time.
I don't know the best solution to fix that, but if you have an idea, I can create a pull request.

Furthermore, the response JSON format in my case is:
```
{'count': 1, 'next': None, 'previous': None, 'results': [{'url': 'https://app.cvat.ai/api/jobs/71332', 'id': 71332, 'task_id': 72946, 'project_id': 12856, ...}]}
```

The top layer keys are `['count', 'next', 'previous', 'results']` but in the line `[j["id"] for j in job_resp.json()]` the `id` is required to be in first layer. This causes the exception `string indices must be integer` as `j["id"]` is actually `"count"["id"]`.

I recommend changing it to `job_ids = [j["id"] for j in job_resp.json()["results"]]` as this resolves the second issue.

The results still might be empty right after the post request. There is still another fix required.

### LINKED ISSUE #2567: [BUG] Filetype CVAT issue
### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04): Linux 5.15.81-1-MANJARO
-   **Python version** (`python --version`): Python 3.10.8
-   **FiftyOne version** (`fiftyone --version`): FiftyOne v0.19.0, Voxel51, Inc.

### Commands to reproduce

```python
import fiftyone as fo

dataset = fo.Dataset()
sample = fo.Sample(filepath='some/image/without/filetype')
dataset.add_samples([sample])

dataset.annotate(
  'mytask',
  label_field="new_field",
  label_type="detections",
  classes=["dog", "cat", "person"]
)
```

### Describe the problem
When trying to create a task in CVAT, the job creation will fail to create due to a missing file type extension and the process will indefinitely wait without warning that the filename is invalid, or explanation of the error.
```
 100% |███████████████| 1/1 [8.6ms elapsed, 0s remaining, 117.5 samples/s] 
Computing metadata...
 100% |███████████████| 1/1 [87.8ms elapsed, 0s remaining, 11.4 samples/s] 
Uploading samples to CVAT...
```

If the filename is changed, to have `.jpg` at the end, then the CVAT upload succeeds.

While it seems silly to provide an image without a filetype extension, the user should be informed that the job failed to create due to this.

### Code to reproduce issue

Run the script listed above.

### Other info/logs

The program hangs due to a while loop in the `cvat.py` util file as it cannot locate the job.
```python
# @todo is this loop really needed?
job_ids = []
while not job_ids:
    job_resp = self.get(self.jobs_url(task_id))
    job_ids = [j["id"] for j in job_resp.json()]
```

### Willingness to contribute

-   [x] No. I cannot contribute a bug fix at this time

### LINKED ISSUE #2572: [DOCS] CVAT Annotation Tutorial: CVAT v2.4 API breaking change
### URL(s) with the issue
https://docs.voxel51.com/tutorials/cvat_annotation.html

### Description of proposal (what needs changing)
CVAT v2.4 has a breaking API change that causes the annotation code in this tutorial to error out if using the CVAT server on cvat.ai.

Proposed workaround is to downgrade to CVAT v2.3 if possible.

ehofesmann:
Resolves #2572
Resolves #2567
Resolves #2565 

Thanks to @tjaeuth for the recommendation!

The CVAT API response for the jobs query has been modified slightly which broke our integration for v2.4 servers. This PR adds support for v2.4 servers and updates our tests. All tests pass on both v2.3 and v2.4 now.

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -4139,7 +4139,11 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
         job_ids = []
         while not job_ids:
             job_resp = self.get(self.jobs_url(task_id))
-            job_ids = [j["id"] for j in job_resp.json()]
+            job_resp_json = job_resp.json()
+            if "results" in job_resp_json:
+                job_resp_json = job_resp_json["results"]
+
+            job_ids = [j["id"] for j in job_resp_json]
 
         if job_assignees is not None:
             num_assignees = len(job_assignees)
--- a/tests/intensive/cvat_tests.py
+++ b/tests/intensive/cvat_tests.py
@@ -441,7 +441,10 @@ class CVATTests(unittest.TestCase):
                 self.assertEqual(task_json["name"], f"{task_name}_{idx + 1}")
                 if user is not None:
                     self.assertEqual(task_json["assignee"]["username"], user)
-                for job in api.get(api.jobs_url(task_id)).json():
+                jobs_json = api.get(api.jobs_url(task_id)).json()
+                if "results" in jobs_json:
+                    jobs_json = jobs_json["results"]
+                for job in jobs_json:
                     job_json = api.get(job["url"]).json()
                     if user is not None:
                         self.assertEqual(


swheaton (approved):
lgtm

PR #2600, opened 2023-02-02
[bug fix] update grid to cropped items after clicking patches button

kaixi-wang:

- Currently, patches views don't crop to the bounding box: 
image

- with this change:
image

(Details)

        notes for FiftyOne users.

-   [ ] Other


brimoor:
@kaixi-wang I tested this locally and it runs fine, but I also do not see the error you're fixing without this change locally either? So question is: does it make sense to you that this change is required for Teams but not OSS?

PR #2605, opened 2023-02-03
Adding support for importing/exporting generic media

brimoor:
Adds a `MediaDirectory` type for importing/exporting arbitrary (homogeneous) directories of media (images, videos, point clouds, etc).

```py
import fiftyone as fo
import fiftyone.zoo as foz

# dataset = foz.load_zoo_dataset("quickstart-video")  # also works

dataset.export(
    export_dir="/tmp/media",
    dataset_type=fo.types.MediaDirectory,
)

    dataset_dir="/tmp/media",
    dataset_type=fo.types.MediaDirectory,
)
```


swheaton:
I think the change itself looks good.

Note for future
- We could probably refactor to remove some duplicated code by re-using media exporter concepts as a base for image and video.

swheaton (approved):

brimoor:
Yeah I agree. I regret building the dedicated `Image*` and `Video*` interfaces with method names like` requires_image_metadata` because it could have been unified under a single `Media*` interface.

But, I hesitate to modify the existing interfaces because it's possible for users to write their own custom subclasses for use with `from_dir()`, `export()`, etc.

PR #2609, opened 2023-02-03
[bug fix] rm duplicate appConfig

kaixi-wang:

- consolidate appConfig fields and remove duplicate field on query

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/Dataset.ts
+++ b/app/packages/core/src/Dataset.ts
@@ -49,12 +49,14 @@ export const DatasetNodeQuery = graphql`
       appConfig {
         gridMediaField
         mediaFields
+        modalMediaField
         plugins
         sidebarGroups {
           expanded
           paths
           name
         }
+        sidebarMode
       }
       sampleFields {
         ftype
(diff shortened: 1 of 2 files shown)


brimoor (dismissed):

app/packages/core/src/Dataset.ts
```diff
@@ -132,17 +134,6 @@ export const DatasetNodeQuery = graphql`
       viewCls
       viewName
       savedViewSlug
-      appConfig {
```
imanjra, line 135: Isn't this different from appConfig in dataset?

kaixi-wang: not sure what appConfig you are referring to? `appConfig` here is defined twice in the same query. Is that intensional?

benjaminpkane: You are right @kaixi-wang that this is duplicate code. A query like this is actually bad, and should be broken out into smaller fragments for better consumption. Not important right now

But there is a global app config in the GraphQL schema. It is actually not used by the App (yet) because it is initialized from the event source (subscription) which is not yet GraphQL

benjaminpkane (approved):
LGTM

PR #2611, opened 2023-02-04
fix spaces and view state update conflicts

imanjra:

Changes in this PR fixes a state conflict between spaces and view. 

## How is this patch tested? If it is not, please explain `why.`

Tested in the app and using the session to ensure state changes do not override other non-modified state

        notes for FiftyOne users.

-   [ ] Other


kaixi-wang (changes requested):
I just tested this locally and it breaks view/saved views in the non-stateful app:
-the url and selector get out of sync
-the view doesn't clear when navigating away and back to the dataset
image
2.

kaixi-wang (approved):
image

PR #2612, opened 2023-02-04
[Bug fix] Enforce unique slugs for new saved views

kaixi-wang:

- add method to convert name to slug 
- check if slug already exists

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Common/utils.tsx
+++ b/app/packages/core/src/components/Common/utils.tsx
@@ -109,8 +109,8 @@ export const toSlug = (name: string) => {
   if (name.length < 1) {
     return "";
   }
-  const valid_chars = new RegExp("[a-z0-9._+-]", "g");
-  const replace_symbols = new RegExp("[-._+]+", "g");
+  const valid_chars = new RegExp("[a-z0-9._+ -]", "g");
+  const replace_symbols = new RegExp("[-._+ ]+", "g");
   const trim = new RegExp("-?(?<slug>[0-9a-z][0-9a-z-]*?)-?$");
 
   let slug = name.toLowerCase();
--- a/app/packages/core/src/components/Common/utils.tsx
+++ b/app/packages/core/src/components/Common/utils.tsx
@@ -94,3 +94,34 @@ export const getStep = (
 
   return step;
 };
+
+export const toSlug = (str: string) => {
+  /**  Returns the URL-friendly slug for the given string.
+   *
+   * The following strategy is used to generate slugs:
+   *   (based on fiftyone.core.utils `to_slug` function)
+   *   -   The characters ``A-Za-z0-9`` are converted to lowercase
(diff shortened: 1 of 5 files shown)


app/packages/core/src/components/Common/utils.tsx
```diff
@@ -94,3 +94,36 @@ export const getStep = (
 
   return step;
 };
+
+export const toSlug = (name: string) => {
```
manivoxel51, line 98: hm... we have this method in the backend and I believe we are using it to make sure the dataset with the same name is not getting saved. I am not sure we should have this logic in the frontend and backend. If one change, the other can diverge. 🤔 

https://user-images.githubusercontent.com/109545780/217056381-6acb05c1-1990-46d7-b6c1-6db19f2ac5df.mov

manivoxel51 (commented):
thanks for your work @kaixi-wang! could you add more comments as to what this PR fixes

app/packages/core/src/components/Common/utils.tsx
manivoxel51: I have to double check the code but the better way could be call the api with debounce and make sure the slug is available or not (it might be doing that already). what do you think?

kaixi-wang: yeah I had the same thoughts about wanting to reuse the backend method, but then it felt like it was making too many server requests for something that can be derived client side

kaixi-wang: however, if we wanted to do the check on click of save view to guard against submitting the view (but not for disabling the button), it would make more sense to call the api...

Also, it may make sense to make it a rest endpoint rather than graphql query

benjaminpkane (approved):
LGTM. I think having `toSlug` on the client will be good because we can write the GraphQL API using that is the identifier

app/packages/core/src/components/Common/utils.tsx
benjaminpkane: I think having two fairly simple functions on the backend/frontend is better than the API consequences of not having a common identifier

app/packages/core/src/components/Common/utils.tsx
```diff
@@ -94,3 +94,36 @@ export const getStep = (
...
+    return "";
+  }
+  const valid_chars = new RegExp("[a-z0-9._+ -]", "g");
+  const replace_symbols = new RegExp("[-._+ ]+", "g");
+  const trim = new RegExp("-?(?<slug>[0-9a-z][0-9a-z-]*?)-?$");
+
+  let slug = name.toLowerCase();
+  let matches = [];
+  let match;
+  while ((match = valid_chars.exec(name.toLowerCase())) !== null) {
```
imanjra, line 119: Suggestion (may ignore): replace `name.toLowerCase()` with `slug`

imanjra (approved):

PR #2614, opened 2023-02-05
Always use IDs in plots and brain results

brimoor:
Updates methods like `scatterplot()` and the in-App embeddings backend to always rely on sample/label IDs when pulling data for plots.

Previously, these implementations assumed that the `view` on which embeddings/etc were computed has not changed in any way. Now, these methods will gracefully handle added/deleted data.

New tests are added to `tests/intensive/plot_tests.py` to verify that the plotting methods function as expected.

### DIFF
--- a/fiftyone/core/plots/utils.py
+++ b/fiftyone/core/plots/utils.py
@@ -5,46 +5,339 @@ Plotting utils.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-import os
+import itertools
+import warnings
 
 import numpy as np
+import sklearn.linear_model as skl
+import sklearn.metrics as skm
 
-import eta.core.image as etai
 import eta.core.serial as etas
 import eta.core.utils as etau
 
+import fiftyone.core.expressions as foe
+import fiftyone.core.labels as fol
 
-# Original images from https://www.iconfinder.com
-BUTTON_ICONS = {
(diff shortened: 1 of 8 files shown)


fiftyone/server/routes/embeddings.py
```diff
@@ -42,20 +42,41 @@ async def post(self, request: Request, data: dict) -> dict:
...
                 _, root = dataset._get_label_field_path(patches_field)
                 leaf = label_field[len(root) + 1 :]
-                _, label_field = curr_view._get_label_field_path(
+                _, label_field = view._get_label_field_path(
                     patches_field, leaf
                 )
 
-            labels = curr_view.values(label_field, unwind=True)
-            field = curr_view.get_field(label_field)
+            labels = view._get_values_by_id(
```
brimoor, line 75: Updates embeddings backend to always pull color-by data using IDs

fiftyone/core/plots/utils.py
```diff
@@ -5,46 +5,336 @@
...
+
+    if etau.is_str(values) or isinstance(values, foe.ViewExpression):
+        if samples is None:
+            raise ValueError(
+                "You must provide `samples` in order to extract field values "
+                "for the `%s` parameter" % parameter
+            )
+
+        if ids is not None and not is_frames:
+            values = samples._get_values_by_id(
```
brimoor, line 255: All of the refactoring in the plotting methods was to achieve this line: when the user has provided IDs and is asking to pull values from a dataset via path/expression, use `_get_values_by_id()` to ensure that the correct values are pulled, even if `samples` doesn't correspond 1-1 with other data that may have been provided.

tests/intensive/plot_tests.py
```diff
@@ -0,0 +1,349 @@
...
+    dataset = foz.load_zoo_dataset("quickstart")
+
+    view = dataset.take(51, seed=51)
+    points = np.random.randn(len(view), 2)
+    ids = view.values("id")
+
+    plot = fo.scatterplot(
+        points=points,
+        samples=dataset,
+        ids=ids,
```
brimoor, line 97: This is an example of something that wouldn't previously work. User provided `ids` + `points` corresponding to a subset of the `samples=dataset` argument that they provided, but they provided paths/expressions for `labels` and `sizes` arguments.

Previously this would fail because `dataset.values()` would naively be used (which would result in too many labels/sizes). Now the `ids` argument is used to lookup the correct labels/sizes in the correct order corresponding to points/ids.

ritch (approved):
Embeddings.py changes LGTM - Other plot changes look fine but I'm less familiar with those.

PR #2615, opened 2023-02-05
Use TTL caches when accessing datasets in App server

brimoor:
Updates all usages of `load_dataset()` in the App server to use a TTL cache so that recently used dataset objects will not be garbage collected between async calls.

```py
@cachetools.func.ttl_cache(maxsize=10, ttl=900)  # ttl in seconds
def load_dataset(name):
    """Loads the dataset with the given name.

    Dataset objects are singletons, but we use a TTL + LRU cache here to ensure
    that references to recently used datasets exist in memory so that dataset
    objects aren't garbage collected between async calls.

    Args:
        name: the dataset name

    Returns:
        a :class:`fiftyone.core.dataset.Dataset`
    """
    return fod.load_dataset(name)
```

### DIFF
--- a/fiftyone/server/routes/embeddings.py
+++ b/fiftyone/server/routes/embeddings.py
@@ -7,17 +7,24 @@ FiftyOne Server ``/embeddings`` route.
 """
 import itertools
 
-import cachetools.func
 from starlette.endpoints import HTTPEndpoint
 from starlette.requests import Request
 
-import fiftyone as fo
+import fiftyone.core.fields as fof
+import fiftyone.core.stages as fos
+
 from fiftyone.server.decorators import route
+import fiftyone.server.utils as fosu
 import fiftyone.server.view as fosv
 
 
 MAX_CATEGORIES = 100
-COLOR_BY_TYPES = (fo.StringField, fo.BooleanField, fo.IntField, fo.FloatField)
+COLOR_BY_TYPES = (
+    fof.StringField,
+    fof.BooleanField,
+    fof.IntField,
+    fof.FloatField,
+)
 
 
 class OnPlotLoad(HTTPEndpoint):
@@ -29,7 +36,7 @@ class OnPlotLoad(HTTPEndpoint):
         stages = data["view"]
         label_field = data["labelField"]
 
-        dataset = _load_dataset(dataset_name)
+        dataset = fosu.load_dataset(dataset_name)
         results = dataset.load_brain_results(brain_key)
 
         view = fosv.get_view(dataset_name, stages=stages)
@@ -55,7 +62,7 @@ class OnPlotLoad(HTTPEndpoint):
 
             labels = curr_view.values(label_field, unwind=True)
             field = curr_view.get_field(label_field)
-            if isinstance(field, fo.FloatField):
+            if isinstance(field, fof.FloatField):
                 style = "continuous"
             else:
                 if len(set(labels)) <= MAX_CATEGORIES:
@@ -119,7 +126,7 @@ class EmbeddingsSelection(HTTPEndpoint):
         if not filters and not extended_stages and not extended_selection:
             return {"selected": None}
 
-        dataset = _load_dataset(dataset_name)
+        dataset = fosu.load_dataset(dataset_name)
         results = dataset.load_brain_results(brain_key)
 
         view = fosv.get_view(dataset_name, stages=stages)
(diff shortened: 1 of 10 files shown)


fiftyone/core/singletons.py
```diff
@@ -24,19 +24,22 @@ def __new__(metacls, *args, **kwargs):
         return cls
 
     def __call__(cls, name=None, _create=True, *args, **kwargs):
+        instance = cls._instances.pop(name, None)
```
brimoor, line 27: @kaixi-wang I believe this integrates your idea from https://github.com/voxel51/fiftyone/pull/2634 in a more holistic way: apparently it is possible in multithread/multiprocess scenarios for `load_dataset()` to encounter a `Dataset` object with `deleted==False` but `name==None`. So, here we detect that and decide to reload the dataset from the database in that case.

fiftyone/server/routes/embeddings.py
```diff
@@ -252,12 +258,6 @@ async def post(self, request: Request, data: dict) -> dict:
 ]
 
 
-def _load_dataset(dataset_name):
```
brimoor, line 255: Note how the current implementation is stored a *permanent* reference to datasets that are used in the embeddings routes. Some version of this PR needs to be checked in to resolve that!

fiftyone/server/routes/sort.py
```diff
@@ -25,31 +24,22 @@ async def post(self, request: Request, data: dict):
...
-        if dist_field and not dataset.get_field(dist_field):
-            dataset.add_sample_field(dist_field, fof.FloatField)
-            changed = True
-
-        fosv.get_view(dataset_name, stages=stages, filters=filters)
-
-        state = fose.get_state().copy()
         view = fosv.get_view(dataset_name, stages=stages, filters=filters)
-        state.dataset = view._dataset
+        dataset = view._dataset
```
brimoor, line 29: Unrelated to the rest of this PR, while inspecting server code, it seemed to me that this route's current implementation was broken...

fiftyone/server/routes/sort.py
```diff
@@ -25,31 +24,22 @@ async def post(self, request: Request, data: dict):
...
 
-        if isinstance(view, fov.DatasetView):
-            state.view = view
-        else:
-            view = None
+        state = fose.get_state().copy()
```
kaixi-wang, line 31: I hadn't noticed we create a shallow copy here. Why is it better to do that for sort?

fiftyone/server/routes/sort.py
benjaminpkane: Copying the state and modifying has no effect

fiftyone/server/routes/sort.py
```diff
@@ -25,31 +24,22 @@ async def post(self, request: Request, data: dict):
...
-        return {
-            "dataset": await serialize_dataset(
+        if dist_field and not dataset.has_field(dist_field):
+            dataset.add_sample_field(dist_field, fof.FloatField)
+            _dataset = await serialize_dataset(
                 dataset_name=dataset_name,
-                serialized_view=stages,
-                view_name=view.name,
+                serialized_view=view._serialize(),
+                saved_view_slug=fou.to_slug(view.name) if view.name else None,
```
kaixi-wang, line 40: good catch on the change in param names, but won't `view.name` here always be null since the view is built from `fosv.get_view(dataset_name, stages=stages, filters=filters)`

fiftyone/core/singletons.py
kaixi-wang: If somehow instance.name is None here, isn't this just creating an instance with no name

fiftyone/server/routes/embeddings.py
```diff
@@ -30,7 +36,7 @@ async def post(self, request: Request, data: dict) -> dict:
         stages = data["view"]
         label_field = data["labelField"]
 
-        dataset = _load_dataset(dataset_name)
+        dataset = fosu.load_and_cache_dataset(dataset_name)
```
kaixi-wang, line 39: Since datasets with embeddings are explicitly cached, won't we get an error if we delete a dataset after plotting it and recreate a dataset with the same name as before?

fiftyone/core/singletons.py
brimoor: If `instance.name` is None, this will remove that instance from the dict, which is good.

`name=None` is possible. That's why `name = instance.name` is done later. `Dataset()` internally generates a unique name when called with no `name` arg.

fiftyone/server/routes/embeddings.py
brimoor: Well based on you and I's earlier experiments, what may previously happen in that case is that `load_dataset()` would return an instance with `name=None`. Hence the other change in this PR: now `load_dataset()` should detect that case and go ahead and reload the dataset from mongo, which is what we want.

fiftyone/server/routes/sort.py
benjaminpkane: Yes it will always be null

fiftyone/server/routes/sort.py
benjaminpkane: I don't recall. Removing it

benjaminpkane (approved):
LGTM

PR #2616, opened 2023-02-06
Add support for importing CSV datasets

swheaton:
Adds support for importing datasets stored as CSV files.

## Example usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("cifar10")

dataset.export(
    labels_path="/tmp/cifar10.csv",
    fields={
        "filepath": "filepath",
        "tags": "tags",
        "ground_truth.label": "ground_truth",
    },
    abs_paths=True,
)

    labels_path="/tmp/cifar10.csv",
    fields={
        "filepath": None,  # load as strings
        "tags": lambda v: v.strip("").split(","),
        "ground_truth": lambda v: fo.Classification(label=v) if v else None,
    },
)

assert set(dataset.values("filepath")) == set(dataset2.values("filepath"))
assert dataset.count_sample_tags() == dataset2.count_sample_tags()
assert dataset.count_values("ground_truth.label") == dataset2.count_values("ground_truth.label")
```

### DIFF
--- a/docs/source/user_guide/dataset_creation/datasets.rst
+++ b/docs/source/user_guide/dataset_creation/datasets.rst
@@ -264,6 +264,9 @@ refer to the corresponding dataset format when reading the dataset from disk.
     | :ref:`BDDDataset <BDDDataset-import>`                                                 | A labeled dataset consisting of images and their associated multitask predictions  |
     |                                                                                       | saved in `Berkeley DeepDrive (BDD) format <https://bdd-data.berkeley.edu>`_.       |
     +---------------------------------------------------------------------------------------+------------------------------------------------------------------------------------+
+    | :ref:`CSVDataset <CSVDataset-import>`                                                 | A labeled dataset consisting of images or videos and their associated field values |
+    |                                                                                       | stored as columns of a CSV file.                                                   |
+    +---------------------------------------------------------------------------------------+------------------------------------------------------------------------------------+
     | :ref:`DICOMDataset <DICOMDataset-import>`                                             | An image dataset whose image data and optional properties are stored in            |
     |                                                                                       | `DICOM format <https://en.wikipedia.org/wiki/DICOM>`_.                             |
     +---------------------------------------------------------------------------------------+------------------------------------------------------------------------------------+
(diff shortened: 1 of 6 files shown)


fiftyone/utils/csv.py
```diff
@@ -214,3 +219,246 @@ def _parse_value(value):
...
+            samples = []
+            with fou.ProgressBar(total=self.max_samples) as pb:
+                # Rip through csv rows
+                for csv_row in pb(labels_csv):
+                    media_loc = csv_row.pop(self.media_field)
+                    # Not absolute path (either local or remote)
+                    if not os.path.isabs(media_loc) and not re.match(
+                        r"^\w+://", media_loc
+                    ):
+                        media_loc = os.path.join(self.data_path, media_loc)
```
swheaton, line 435: Is this right? Do we want to reconstitute a relative path to an absolute path, relative to `self.data_path`? What if there's no media included?

brimoor:
Notes on v1 of this implementation:

## Potential TODO

- I don't love the `class Parsers` that I ended up with. Open to suggestions for improvement. The good thing about it is it's not strict so any custom function or lambda can be passed as a parser for a field.
- Split CSV read so we can do some row parsing in parallel?
- Any corner cases I may have missed?
- Add more common `Parsers` to add? (May get much more complicated if we try for detections because that is nested lists

Create CSVDatasetImporter complementary to CSVDatasetExporter.

Most unique aspect ( the field parsing ) should be called out, and it's described in the docstring which I've pasted here.

```
fields (None): An optional parameter that specifies fields to read and parse
    from the CSV file. Can be any of the following:
    -   `None`. Will use all fields in the CSV and parse with default
        string parser.
    -   `List[str]` which are the field names to be read from
        the CSV. All data will be parsed with default string parser.
    -   `Dict[str, Optional[fiftyone.utils.csv.Parsers]]` where the key
        specifies the field name to be read and the value specifies the
        parser to be used, e.g., `StringParser` or `IntParser`. If `None`,
        default string parser will be used. `ListParser` can be uniquely
        used by passing in sub parser type to use for each data field.
        For example:
        {
            "float_field": Parsers.FloatParser,
            "list_field": Parsers.ListParser(Parsers.StringParser),
            "classification_field": Parsers.ClassificationParser,
            "string_field": None # uses DefaultParser
        }
```

docs/source/user_guide/dataset_creation/datasets.rst
```diff
@@ -4204,6 +4207,163 @@ directory containing the corresponding media files by providing the
     If the `name` key of your labels contains absolute paths to the source
     media, then you can omit the `data_path` parameter from the example above.
 
+.. _CSVDataset-import:
```
brimoor, line 4210: Added documentation for importer

fiftyone/utils/csv.py
```diff
@@ -6,13 +6,230 @@
...
+    @property
+    def has_sample_field_schema(self):
+        return False
+
+    @property
+    def has_dataset_info(self):
+        return False
+
+    def setup(self):
+        media_paths_map = self._load_data_map(self.data_path, recursive=True)
```
brimoor, line 174: Our builtin importers are generally implemented using this `_load_data_map()` function. I'll explain offline when we have some free time.

fiftyone/utils/csv.py
```diff
@@ -6,13 +6,230 @@
...
+                reader = csv.DictReader(f)
+                if media_field not in reader.fieldnames:
+                    raise ValueError(
+                        "Media field '%s' not found in '%s'"
+                        % (media_field, self.labels_path)
+                    )
+
+                for row in reader:
+                    filename = row.pop(media_field, None)
+                    if filename is None or os.path.isabs(filename):
```
brimoor, line 191: Allows CSV file to contain absolute paths, or `paths.jpg`, or `rel/ative/paths.jpg` in `data/`.

fiftyone/utils/csv.py
```diff
@@ -6,13 +6,230 @@
...
+                            )
+
+                    rows_map[filepath] = row
+
+        filepaths = set(rows_map.keys())
+
+        if self.include_all_data:
+            filepaths.update(media_paths_map.values())
+
+        filepaths = self._preprocess_list(sorted(filepaths))
```
brimoor, line 211: By convention, `(max_samples=n, shuffle=True)` samples from entire dataset. This implementation achieves that.

fiftyone/utils/csv.py
```diff
@@ -78,8 +295,14 @@ class CSVDatasetExporter(foud.BatchDatasetExporter, foud.ExportPathsMixin):
...
         media_field ("filepath"): the name of the field containing the media to
             export for each sample
-        fields (None): a field or iterable of fields to include as columns in
-            the exported CSV file. By default, only the ``media_field`` is used
+        fields (None): an optional argument specifying the fields or
+            ``embedding.field.names`` to include as columns in the exported
+            CSV. Can be:
+
+            -   a field or iterable of fields
+            -   a dict mapping field names to column names
```
brimoor, line 303: Added support for passing `fields={"weather.label": "weather"}` to extract `Classification.label` values and store them in a `weather` column in the CSV file. See comment on unit tests for details.

tests/unittests/import_export_tests.py
```diff
@@ -2104,17 +2104,19 @@ def _make_dataset(self):
...
-                fo.Classification(label=value) if value is not None else None
-            ),
-            "filepath": None,  # default parser
-            "tags": lambda value: value.strip("").split(","),
-            "float_field": lambda value: float(value),
+        export_fields = {
+            "filepath": "filepath",
+            "tags": "tags",
+            "float_field": "float_field",
+            "weather.label": "weather",
```
brimoor, line 2111: Uses the `export_fields=dict` syntax to be able to export a nested field value and then reconstruct it in the same format.

For future work: we should probably add a way for `CSVDatasetImporter` to handle columns names that contain `.`.

brimoor (approved):
LGTM. I took this across the finish line to avoid distracting from other more important features

PR #2617, opened 2023-02-06
Faceted aggregation bug fixes

brimoor:
Fixes a couple bugs with the faceted aggregation logic in `aggregate()`:
- Proper handling of multiple aggregations with the same field name and type
- Proper handling of expressions applied to nested list fields

The unit tests added fail on `develop` but pass on this branch.


benjaminpkane (approved):
LGTM

PR #2619, opened 2023-02-06
Tweak Similarity Menu

lanzhenw:
As a user, if I submitted a similarity, but no samples are currently selected, 
when I open the similarity menu, 
I should see doc_link and reset button only. 

https://user-images.githubusercontent.com/17770824/216999066-fe6e969c-f568-475b-bc1d-20ad0d717c59.mov

(Details)


brimoor (approved):
LGTM 💪

PR #2622, opened 2023-02-06
Relay invariant comment on scalar hydration

benjaminpkane:
I hope this PR will start a larger discussion on "JSON" scalars in FiftyOne, which is essentially schema-less data used largely for sample data and some inputs. Admittedly, this is a shortcoming of the Relay implementation that I did not anticipate.

The issue at hand can be found [here](https://github.com/facebook/relay/issues/91) and it seems unlikely a proper fix will be provided. A practical solution would be to always serialize "JSON" scalars so edge cases don't have to be accounted for.

### DIFF
--- a/app/packages/state/src/recoil/groups.ts
+++ b/app/packages/state/src/recoil/groups.ts
@@ -170,6 +170,8 @@ export const pinnedSliceSample = graphQLSelector<
   mapResponse: (response) => {
     const actualRawSample = response?.sample?.sample;
 
+    // This value may be a string that needs to be deserialized
+    // Only occurs after calling useUpdateSample for pinned sample
     if (actualRawSample && typeof actualRawSample === "string") {
       return {
         ...response.sample,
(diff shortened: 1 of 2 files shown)


benjaminpkane:
Not stringifying the JSON in `useUpdateSample` results in an invariant error.
```
Invariant Violation: RelayRecordProxy#setValue(): Expected a scalar or array of scalars, got:
```
Screen Shot 2023-02-06 at 2 02 00 PM

app/packages/state/src/recoil/groups.ts
```diff
@@ -168,8 +168,11 @@ export const pinnedSliceSample = graphQLSelector<
     };
   },
   mapResponse: (response) => {
+    console.log(response?.sample?.sample);
```
ritch, line 171: Stray log

benjaminpkane: Thanks 👍

brimoor (approved):
LGTM

PR #2623, opened 2023-02-06
fix plugins require fiftyone key

ritch:
Before this patch, if you added a plugin without a `fiftyone` key in its package.json, all plugin loading would fail.


benjaminpkane (approved):
LGTM

PR #2624, opened 2023-02-06
Ensure viewCls for state proxy

benjaminpkane:
In a stateless context, `viewCls` must be set in the state proxy in `useSetView` to ensure default behavior like crop to content in patches views work correctly.


kaixi-wang (approved):
lgtm

PR #2626, opened 2023-02-07
Store patch embeddings directly on label objects

brimoor:
Updates the `compute_patch_embeddings()` method to store patch embeddings directly on `Label` objects when the `embeddings_field` argument is provided.

Previously a `num_objects x num_dims` array would be stored as a top-level sample/frame field, which is less desirable because in that case there's no simple way to, for example, use `filter_labels()` + `values()` to extract the patch embeddings for a particular class.


benjaminpkane (approved):
LGTM

PR #2628, opened 2023-02-07
Fix matching single label fields

benjaminpkane:
On `develop`, fixes matching via the sidebar in the App for categorical values on single label fields like `ground_truth` in `cifar100`

```py
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("cifar100")
session = fo.Session(dataset)
```

Filter `ground_truth.label`  in sidebar


brimoor (approved):
LGTM

PR #2629, opened 2023-02-07
Add extended selection before converting to patches when pinning state

benjaminpkane:
Extended selection should come before patches and other schema converting stages when pinning state.

Screen Shot 2023-02-07 at 3 20 25 PM


brimoor (approved):
LGTM

PR #2631, opened 2023-02-07
Select slice when pinning

benjaminpkane:
Current when pinnning/saving filters with selected samples in a group view, the following server error is encountered
Screen Shot 2023-02-07 at 3 32 16 PM

It seems this error also exist in `0.18.0`


brimoor (approved):
LGTM

brimoor (changes requested):
I think the proper fix here would be to use `SelectGroups(...)` when converting the current selection to a view stage for grouped collections.

In the meantime though, rather than the bookmark icon doing this:

```
dataset.select_group_slices().select(...)
```

I think it can just do this:

```
dataset.select(...)
```

which is what the "show only selected samples" action does anyway.

brimoor (approved):
LGTM

PR #2632, opened 2023-02-08
Fix pinned sample sidebar

benjaminpkane:
Fixes sample data state in the sidebar, etc. when the pinned sample is active in group datasets, e.g. `quickstart-groups`


brimoor (approved):
LGTM

PR #2633, opened 2023-02-08
Search by text prompt in the App

lanzhenw:
Together with #2569 , this allows the user to use text prompt to sort their samples by similarity. Also updates the control pop to be more condensed. 

- new condensed design of similarity search control by selected image/text prompt
- allow sort samples by text prompt query

https://user-images.githubusercontent.com/17770824/217593748-40c55260-ad87-48bb-addb-a47a89d947b7.mp4

If in textprompt search mode, brain key exists but does not support prompts, 
Screenshot 2023-02-08 at 10 03 23 AM

        notes for FiftyOne users.

-   [ ] Other


brimoor:
Great work @lanzhenw! 🙌 

I have some suggested edits to the wording in the UI:

## Text similarity menu

Magnifying glass tooltip
- Current: Sort by text similarity
- Proposed: Search by text similarity

Search box
- Current: Search by text
- Proposed:Type anything!

Gear tooltip
- Current: Advanced Settings
- Proposed:Advanced settings

## Image similarity menu

Photo icon tooltip
- Current: Sort by image similarity
- Proposed:Search by image similarity

Button
- Current: Apply Sort
- Proposed:Show similar samples

## Advanced settings (both menus)

Current:

```
Advanced Settings:
Find the [10] most similar samples,
[ ] in reverse order
with this brain key:
[ ] clip
Save the float field that stores the distance
of each example to the speicified query as:
dist_field (default = None)
```

Proposed:

```
Advanced settings
Show the [10] most similar samples
[ ] show least similar instead
using this brain key
[ ] clip
Optional: store the distance between each
sample and the query in this field:
[dist_field (default=None)]
```

brimoor:
I did find two small bugs when testing:

1. Simple: clicking the (i) icon should open a new tab, but currently I believe it opens a new window instead
2. Select a sample, then select "Clear selected samples" under the checkmark menu. Now nothing is selected, but the image similarity icon shows instead of the text similarity icon. I think you need to add some extra logic to ensure that any means of clearing the current selection always switches back to the text similarity menu

brimoor:
One additional thought: rather than doing this:
```
Show the [10] most similar samples
[ ] show least similar instead
```

maybe you could instead do this:

```
Show the [10] [most] similar samples
```

where `[most]` is a selector that can be changed to `[least]` instead. This would toggle `reverse=True|False`. What do you think about that?

brimoor:
One final thought: we might consider increasing the default `k` a bit. Maybe `k=25`?

lanzhenw:
Update implemented:
https://user-images.githubusercontent.com/17770824/218007970-bade3275-4a03-4770-ae1a-d09e95654804.mov

brimoor:
@lanzhenw edits look great 💪

After a bit more testing, I'm seeing a couple additional bugs:

1. Do the following steps:

- Select a sample
- Click the image icon to open the image similarity menu
- Click outside the menu to close it
- Click on the checkmark icon and select "Clear selected samples"
- Now click on the magnifying icon to open the text similarity menu

The modal is in a broken state where the button says `Show similar samples` (from the image similarity menu) but this is the text similarity menu.

2. Try loading multiple similarity runs, only one of which supports text prompts:

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

fob.compute_similarity(dataset, model="clip-vit-base32-torch", brain_key="clip")

fob.compute_similarity(dataset, model=model, brain_key="clip2")

```

With these two similarity runs, only `clip` supports text prompts, not `clip2`, since text prompts currently require passing the *string name* of the model, not the model itself.

With this setup, the menus incorrectly show the `Search by XXX similarity [^]` help text, which should only show when there are zero compatible brain runs:

Screen Shot 2023-02-10 at 8 58 49 AM
Screen Shot 2023-02-10 at 8 58 54 AM

Also, the advanced settings for text similarity should not show `clip2`, because it doesn't support text prompts:

Screen Shot 2023-02-10 at 8 58 41 AM

app/packages/core/src/components/Actions/Similar.tsx
```diff
@@ -67,39 +86,52 @@ const useSortBySimilarity = (close) => {
...
+              view,
+              subscription,
+              filters: await snapshot.getPromise(fos.filters),
+              extended: toSnakeCase(combinedParameters),
+            }
+          );
+
+          // update selectedSamples atom to new set.
+          // modifying data, otherwise useStateUpdate will set the selectedSamples again
+          data.state.selected = [];
```
benjaminpkane, line 118: This is not allowed. You cannot mutate state on the client like this. I think you need to merge `develop` to get some recent changes I made

app/packages/core/src/components/Actions/Similar.tsx
```diff
@@ -198,35 +244,43 @@ interface SortBySimilarityProps {
...
-      ({ reset }) =>
-        () =>
-          reset(fos.similarityParameters)
-    );
     const hasSimilarityKeys =
       useRecoilValue(availableSimilarityKeys(modal)).length > 0;
-
     const choices = useRecoilValue(currentSimilarityKeys(modal));
     const sortBySimilarity = useSortBySimilarity(close);
     const hasSelectedSamples = [...selectedSamples].length > 0;
```
benjaminpkane, line 272: Unused

app/packages/core/src/components/Actions/Similar.tsx
```diff
@@ -237,91 +291,201 @@ const SortBySimilarity = React.memo(
       current && setState(current);
     }, [current]);
 
+    useEffect(() => {
+      if (!hasNoSelectedSamples) {
+        setIsImageSearch(true);
+      }
+    }, [hasNoSelectedSamples]);
```
benjaminpkane, line 298: This should be a selector. Right now this causes an unnecessary re-render

app/packages/core/src/components/Actions/Similar.tsx
```diff
@@ -237,91 +291,201 @@ const SortBySimilarity = React.memo(
...
+
+    const popoutStyle = { minWidth: 280 };
+
+    const onInfoLink = () => {
+      window.open(SORT_BY_SIMILARITY, "_blank");
```
benjaminpkane, line 303: This is not allowed. You'll have to use `useExternalLink` to ensure it will work in the desktop app.

benjaminpkane (commented):
Feature looks great! Left some comments. I do think this component should be broken out into a couple new components/files. 250 lines for a functional component is a lot of complexity

app/packages/core/src/components/Actions/Similar.tsx
```diff
@@ -237,91 +291,201 @@ const SortBySimilarity = React.memo(
...
+                  validator={(value) =>
+                    value === "" || /^[0-9\b]+$/.test(value)
+                  }
+                  value={state.k === null ? "" : String(state.k)}
+                  setter={(value) => {
+                    setParameter("k", value === "" ? null : Number(value));
+                  }}
+                />
+              </div>{" "}
+              {"   "}
```
benjaminpkane, line 452: Spaces like this are never necessary

app/packages/core/src/components/Actions/Similar.tsx
```diff
@@ -237,91 +291,201 @@ const SortBySimilarity = React.memo(
...
-            {state.brainKey && (
-              <>
-                <PopoutSectionTitle></PopoutSectionTitle>
+          </>
+        )}
+        {open && hasValidKeys && (
+          <div>
+            <div>
+              Find the{" "}
+              <div style={{ width: "30px", display: "inline-block" }}>
```
benjaminpkane, line 440: Please try using flex positioning

app/packages/core/src/components/Actions/Similar.tsx
```diff
@@ -237,91 +291,201 @@ const SortBySimilarity = React.memo(
...
+                onClick={() => {
+                  close();
+                  reset();
+                }}
+                style={{
+                  margin: "0.25rem",
+                  height: "2rem",
+                  borderRadius: 2,
+                  textAlign: "center",
+                  width: "185px",
```
benjaminpkane, line 365: Flex

app/packages/core/src/components/Actions/Similar.tsx
lanzhenw: This is not needed, either. I am stripping it.

app/packages/core/src/components/Actions/Similar.tsx
lanzhenw: will update this

app/packages/core/src/components/Actions/Similar.tsx
lanzhenw: Here, 30px is exactly the size we want. And we don't want the input to stretch. The size of the Popout is fixed, too. 
Screenshot 2023-02-10 at 12 12 07 PM

app/packages/core/src/components/Actions/Similar.tsx
```diff
@@ -237,91 +291,201 @@ const SortBySimilarity = React.memo(
...
+              {"  "} similar samples using this brain key
+              <RadioGroup
+                choices={choices.choices}
+                value={state.brainKey}
+                setValue={(v) => setParameter("brainKey", v)}
+              />
+            </div>
+            Optional: store the distance between each sample and the query in
+            this field
+            <div style={{ width: "60%" }}>
```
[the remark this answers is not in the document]
lanzhenw: Any particular reason for avoiding percentage and px sizing?

manivoxel51:
curious: can a user search by text and save the results as a saved view?

swheaton:
Very cool. Why the magic `k` ?

lanzhenw:
@manivoxel51 Yes, right now this is added to the extended stage, like other filters. You can then bookmark it to create a view it. 
@swheaton Trying to be consistent with the python SDK command `sort_by_similarity(prompt, k, brain_key)`

brimoor:
@manivoxel51 @lanzhenw I didn't test on this branch, but ideally the user doesn't have to click the bookmark icon before creating a saved view; the idea behind "save current filters" in the saved view menu is that anything in the extended view (sidebar filters, embeddings plot selections, selected samples in the grid, and similarity searches) should be automatically included when the user saves a view.

app/packages/core/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
```diff
@@ -108,11 +108,11 @@ export const PARSER = {
     },
   },
   "list<field>": {
-    castFrom: (value) => value.join(","),
-    castTo: (value) => value.split(","),
-    parse: (value) => value.replace(/[\s\'\"\[\]]/g, ""),
+    castFrom: (value) => value?.join(","),
```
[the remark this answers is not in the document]
lanzhenw: There is a bug with the validate method in the `PARSER` with all the list fields. It won't catch the cases that aren't array, so some edge case broke the render before. I fixed the list<str> part for this feature, but they all need to be fixed. I will create a new bug ticket.

manivoxel51:
> @manivoxel51 @lanzhenw I didn't test on this branch, but ideally the user doesn't have to click the bookmark icon before creating a saved view; the idea behind "save current filters" in the saved view menu is that anything in the extended view (sidebar filters, embeddings plot selections, selected samples in the grid, and similarity searches) should be automatically included when the user saves a view.

Right! right now, you can create saved views that are extended so I would imagine this feature would also work the same. we should verify it by saving a search result as saved view

app/packages/core/src/components/Actions/similar/Similar.tsx
```diff
@@ -0,0 +1,230 @@
...
+      icon: "InfoIcon",
+      arialLabel: "information",
+      tooltipText: "Learn more about sorting by similarity",
+      onClick: () => externalLink(SORT_BY_SIMILARITY),
+      sx: { color: "#0288d1" },
+    },
+    {
+      icon: "SettingsIcon",
+      arialLabel: "Advanced settings",
+      tooltipText: "advanced seettings",
```
brimoor, line 105: typo: seettings -> settings

brimoor (dismissed):
@lanzhenw the icons are backwards right now: it shows the photo icon for text similarity searches and the magnifying glass for image similarity searches

brimoor:
@lanzhenw a couple other things:

- The info icon doesn't link anywhere currently. It should link to https://docs.voxel51.com/user_guide/app.html#app-similarity
- On current branch, I don't see the default `k` value populated by default when the expanded settings are opened:

Screen Shot 2023-02-27 at 11 31 40 AM

brimoor:
@lanzhenw the App's state is not fully updating when a similarity search is provided. The grid updates, but `200 samples` is not correct, and the sidebar has not updated:

Screen Shot 2023-02-27 at 11 35 47 AM

app/packages/core/src/components/Actions/similar/Similar.tsx
lanzhenw: Thanks Brian!

brimoor (changes requested):
@lanzhenw Almost there! Here's a last round of small feedback:

- Tooltip for image similarity icon incorrectly says "Search by text similarity". It should be contextual based on image vs text similarity
- Can you add a loading state in the magnifying glass icon while a search is being performed? It's not clear anything is happening currently, which is especially important for the first search, which takes a bit longer because the model has to be loaded

brimoor (approved):
I defer to @benjaminpkane for code-related approval, but the LGTM from a usage standpoint. Nice work! 🙌

benjaminpkane (approved):
LGTM

PR #2634, opened 2023-02-08
Only update dataset entry if name exists

kaixi-wang:

- when loading a dataset, only update the last loaded at for datasets with non null names, otherwise will trigger an error for saving documents with duplicate keys (equal to none)

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -285,8 +285,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         self._deleted = False
 
         if not _virtual:
-            if name:
-                self._update_last_loaded_at()
+            self._update_last_loaded_at()
 
     def __eq__(self, other):
         return type(other) == type(self) and self.name == other.name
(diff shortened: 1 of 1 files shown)


fiftyone/core/dataset.py
```diff
@@ -6204,6 +6204,11 @@ def _serialize(self):
         return self._doc.to_dict(extended=True)
 
     def _update_last_loaded_at(self):
+        if self.name is None:
```
kaixi-wang, line 6207: @brimoor since this update is called every time a dataset is accessed, I think having this check is better than not. While it doesn't solve the root of the problem, it will prevent writing no-name entries to the collection for bugs we may not be aware of

brimoor: `_update_last_loaded_at` is only called in three cases:
1. in `Dataset()`
2. in `DatasetSingleton`
3. in `Dataset.reload()`

#2615 ensures that 1 and 2 will never find a `Dataset` with `name=None`.

brimoor: Only thing I think we could do is add something to `Dataset.reload()` that will mark the dataset as deleted if it reloads its data and finds that `name=None`. And *not* update `last_loaded_at`.

brimoor:
https://github.com/voxel51/fiftyone/pull/2615 achieves the same outcome in a more general way.

PR #2635, opened 2023-02-08
Include all group slices in media size stat

brimoor:
Previously the media size computation in `Dataset.stats()` only accounted for the current slice of a grouped dataset. Now it includes all slices, as expected.

```py
import fiftyone as fo
import fiftyone.zoo as foz

stats1 = dataset.stats(include_media=True)
stats2 = dataset.select_group_slices(_allow_mixed=True).stats(include_media=True)

# previously failed; now succeeds
assert stats1["total_bytes"] == stats2["total_bytes"]
```


kaixi-wang (approved):
ltgm

PR #2636, opened 2023-02-09
select all groups in unlabeled media dataset exporter

swheaton:
fixes media directory export in a group type dataset by selecting all slices first

Tested - before only get 200 images, after get 400 images plus pcd files
```
dataset=fo.load_dataset('quickstart-groups')
dataset.export(dataset_type=fo.types.MediaDirectory)
```


brimoor (commented):
Per offline discussion, I think this logic is best inserted by callers of `export()`, to maintain consistency with other formats like `export(..., fo.types.ImageDirectory)`.

PR #2641, opened 2023-02-09
Fix Python client session initialization

benjaminpkane:
Currently, Python client's cannot connect properly because the initializer payload is malformed.
```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.limit(10)

session = fo.launch_app(view)
```


brimoor (approved):
LGTM

PR #2642, opened 2023-02-09
Ensure notebook context and screenshots

benjaminpkane:
Fixes screenshotting by using the `notebook` URL search flag to check for context


app/packages/state/src/hooks/useTo.ts
```diff
@@ -2,7 +2,10 @@ import { useCallback, useContext, useTransition } from "react";
 import { RouterContext, RoutingContext } from "../routing";
 
 const goTo = (router: RoutingContext, path: string, state: any) => {
-  router.history.push(path, state);
+  const searchParams = new URLSearchParams(window.location.search);
+  searchParams.delete("view");
```
brimoor, line 6: I assume you considered whether there are any consequences of this in non-notebook-screenshot contexts?

brimoor (approved):
LGTM

PR #2645, opened 2023-02-10
[HOLD] Index sample tags by default

brimoor:
Indexes the sample `tags` field by default.

This is probably long overdue, considering the prominent real estate in the App we give to tags and the builtin methods like `match_tags()`, `count_sample_tags()`, etc.

## UPDATE

Hmmm, so apparently indexed list fields may no longer guarantee that samples in a `[{"$match": ...}]` pipeline will maintain insertion order, which is something that I imagine users are implicitly assuming today... 🤔 

```py
import fiftyone as fo

dataset = fo.Dataset()
dataset.add_samples(
    [
        fo.Sample(filepath="image1.png", tags=["train"], i=1),
        fo.Sample(filepath="image2.png", tags=["test"], i=2),
        fo.Sample(filepath="image3.png", tags=["train", "test"], i=3),
        fo.Sample(filepath="image4.png", i=4),
    ]
)

# This is true with and without an index
view = dataset.match_tags("test")
assert view.values("i") == [2, 3]

# This is true with no index, but the order is [4, 1] with an index :(
view = dataset.match_tags("test", bool=False)
assert view.values("i") == [1, 4]
```

## Benchmarking

Benchmarks on `develop` verify that `match_tags()` views are ~3.7x faster with indexed `tags` field on a dataset with 60K samples and 10 tags per sample:

```py
import string
import random

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("cifar10")

all_tags = list(string.ascii_lowercase)
tags = [random.sample(all_tags, 10) for _ in range(len(dataset))]
dataset.set_values("tags", tags)
print(dataset.count_sample_tags())  # yep, lots of tags

counts1 = []
with fo.ProgressBar(iters_str="ops") as pb:
    for tag in pb(all_tags):
        counts1.append(len(dataset.match_tags(tag)))

dataset.create_index("tags")

counts2 = []
with fo.ProgressBar(iters_str="ops") as pb:
    for tag in pb(all_tags):
        counts2.append(len(dataset.match_tags(tag)))
```

```
100% |██████████████████████| 26/26 [1.4s elapsed, 0s remaining, 18.9 ops/s]
100% |██████████████████████| 26/26 [371.9ms elapsed, 0s remaining, 69.9 ops/s]
```

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -8162,7 +8162,7 @@ class SampleCollection(object):
             return []
 
         if self._is_patches:
-            names = ["id", "filepath", "sample_id"]
+            names = ["id", "filepath", "tags", "sample_id"]
             if self._is_frames:
                 names.extend(["frame_id", "_sample_id_1_frame_number_1"])
 
@@ -8172,14 +8172,15 @@ class SampleCollection(object):
             return [
                 "id",
                 "filepath",
+                "tags",
                 "sample_id",
                 "_sample_id_1_frame_number_1",
             ]
 
         if self._is_clips:
-            return ["id", "filepath", "sample_id"]
+            return ["id", "filepath", "tags", "sample_id"]
 
-        return ["id", "filepath"]
+        return ["id", "filepath", "tags"]
 
     def reload(self):
         """Reloads the collection from the database."""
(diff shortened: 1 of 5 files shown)


kaixi-wang:
you may have to also have the initial order in the index to have it preserve the order

PR #2651, opened 2023-02-12
Present error message for embeddings plot if no results

kaixi-wang:

Currently throws an error: 
image

Proposal:
Handle fetch failure:
image

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/embeddings/src/useViewChangeEffect.tsx
+++ b/app/packages/embeddings/src/useViewChangeEffect.tsx
@@ -39,6 +39,12 @@ export function useViewChangeEffect() {
     fetchPlot({ datasetName, brainKey, view, labelField })
       .catch((err) => setLoadingPlotError(err))
       .then((res) => {
+        if (!res || !res.index_size) {
+          warnings.add(
+            `The current view does not have any corresponding embeddings.`
+          );
+          return;
+        }
         const notUsed = res.index_size - res.available_count;
         const missing = res.missing_count;
         const total = res.index_size;
--- a/app/packages/embeddings/src/useViewChangeEffect.tsx
+++ b/app/packages/embeddings/src/useViewChangeEffect.tsx
@@ -40,9 +40,14 @@ export function useViewChangeEffect() {
       .catch((err) => setLoadingPlotError(err))
       .then((res) => {
         if (!res || !res.index_size) {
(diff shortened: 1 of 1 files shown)


app/packages/embeddings/src/useViewChangeEffect.tsx
```diff
@@ -47,6 +47,12 @@ export function useViewChangeEffect() {
     fetchPlot({ datasetName, brainKey, view, labelField })
       .catch((err) => setLoadingPlotError(err))
       .then((res) => {
+        if (!res || !res.index_size) {
+          warnings.add(
+            `Error fetching embeddings corresponding to the current view.`
```
brimoor, line 52: nit: I'd recommend:

```
Failed to fetch embeddings for the current view.
```

app/packages/embeddings/src/useViewChangeEffect.tsx
```diff
@@ -47,6 +47,12 @@ export function useViewChangeEffect() {
     fetchPlot({ datasetName, brainKey, view, labelField })
       .catch((err) => setLoadingPlotError(err))
       .then((res) => {
+        if (!res || !res.index_size) {
```
brimoor, line 50: Shouldn't `res.index_size == 0` have a different warning? Maybe the user loaded `dataset.limit(0)` view. Something like:
```
The index contains no embeddings for the current view.
```

brimoor (commented):
I defer to @ritch for review but this looks like a nice addition to me (although fwiw I haven't been able to get my App in a state where `!res`)

app/packages/embeddings/src/useViewChangeEffect.tsx
kaixi-wang: `res.index_size == 0` is handled below. This catches if there is no `index_size` on the result object, which was the case I hit. Instead it contained an error message (didn't throw an error that would have been caught by the catch above)

ritch: Index size === 0 is being caught here since we return and !0 === !undefined

kaixi-wang: updated to return a different message when index is 0 vs no index at all

ritch (approved):

PR #2656, opened 2023-02-14
Orthographic projection backend

sashankaryal:

Adds `compute_orthographic_projection_images()` API to the `fo::utils3d` module that can be used to generate orthographic projections for datasets with point clouds.

Example usage:

```python
import fiftyone as fo
import fiftyone.zoo as foz

zoo_dataset = foz.load_zoo_dataset("quickstart-groups", max_samples=6)
pcd_only_view = zoo_dataset.select_group_slices("pcd")
pcd_only_view.delete_sample_field("group")

fou3d.compute_orthographic_projection_images(pcd_only_view, (-1, 512), "/tmp/projection")

print(pcd_only_view.first())

--

<SampleView: {
    'id': '63eba9e36d88da1af48f7520',
    'media_type': 'point-cloud',
    'filepath': '/Users/sashankaryal/fiftyone/datasets/quickstart-groups/data/003037.pcd',
    'tags': ['train'],
    'metadata': <Metadata: {'size_bytes': 1870104, 'mime_type': 'application/octet-stream'}>,
    'group': <Group: {'id': '63b70d2018577ca18b763c1e', 'name': 'pcd'}>,
    'ground_truth': <Detections: {
        'detections': [
            <Detection: {
                'id': '63b70d2018577ca18b763c27',
                'attributes': {},
                'tags': [],
                'label': 'Car',
                'bounding_box': [],
                'mask': None,
                'confidence': None,
                'index': None,
                'dimensions': [1.66, 4.49, -1.67],
                'location': [-3.81, 30.32, -0.885],
                'rotation': [0, 0, 0.010796326794896505],
            }>,
            <Detection: {
                'id': '63b70d2018577ca18b763c28',
                'attributes': {},
                'tags': [],
                'label': 'Car',
                'bounding_box': [],
                'mask': None,
                'confidence': None,
                'index': None,
                'dimensions': [1.56, 3.09, -1.72],
                'location': [-20.2, 46.74, -1.21],
                'rotation': [0, 0, 3.140796326794897],
            }>,
            <Detection: {
                'id': '63b70d2018577ca18b763c29',
                'attributes': {},
                'tags': [],
                'label': 'Car',
                'bounding_box': [],
                'mask': None,
                'confidence': None,
                'index': None,
                'dimensions': [1.46, 4.12, -1.59],
                'location': [-4.19, 50.44, -0.9349999999999999],
                'rotation': [0, 0, -0.009203673205103513],
            }>,
            <Detection: {
                'id': '63b70d2018577ca18b763c2a',
                'attributes': {},
                'tags': [],
                'label': 'Car',
                'bounding_box': [],
                'mask': None,
                'confidence': None,
                'index': None,
                'dimensions': [1.57, 3.32, -1.57],
                'location': [-20.21, 64.9, -1.2349999999999999],
                'rotation': [0, 0, 3.1207963267948964],
            }>,
            <Detection: {
                'id': '63b70d2018577ca18b763c2b',
                'attributes': {},
                'tags': [],
                'label': 'Car',
                'bounding_box': [],
                'mask': None,
                'confidence': None,
                'index': None,
                'dimensions': [1.57, 5.24, -1.2],
                'location': [-20.1, 70.64, -1.3399999999999999],
                'rotation': [0, 0, 0.0007963267948964958],
            }>,
            <Detection: {
                'id': '63b70d2018577ca18b763c2c',
                'attributes': {},
                'tags': [],
                'label': 'Car',
                'bounding_box': [],
                'mask': None,
                'confidence': None,
                'index': None,
                'dimensions': [1.72, 4.32, -1.66],
                'location': [-4.26, 69.13, -0.91],
                'rotation': [0, 0, 0.010796326794896505],
            }>,
            <Detection: {
                'id': '63b70d2018577ca18b763c2d',
                'attributes': {},
                'tags': [],
                'label': 'Car',
                'bounding_box': [],
                'mask': None,
                'confidence': None,
                'index': None,
                'dimensions': [1.62, 4.5, -1.63],
                'location': [-0.27, 34.18, -0.825],
                'rotation': [0, 0, 3.140796326794897],
            }>,
        ],
    }>,
    'orthographic_projection_metadata': <OrthographicProjectionMetadata: {
        'filepath': '/tmp/map/003037-2.png',
        'min_bound': array([-56.78701019, -76.05518341]),
        'max_bound': array([79.1415863 , 76.70161438]),
        'width': 512,
        'height': 456
    }>,
}>
```

Unit testing, golden testing

        notes for FiftyOne users.

This PR is accompanied by [another PR_link_todo](http://todo) for frontend which contains the release notes.

-   [ ] Other


sashankaryal:
note: the unit tests fail because the specs need to be updated after the most recent commit on rotation

fiftyone/server/samples.py
```diff
@@ -44,7 +44,7 @@ class ImageSample(Sample):
 
 @gql.type
 class PointCloudSample(Sample):
-    pass
+    aspect_ratio: float
```
benjaminpkane, line 47: `aspect_ratio` could be moved to the `Sample` interface now that all three types require it

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,394 @@
...
+# Reference: https://math.stackexchange.com/q/180418
+def _rotation_matrix_from_vectors(vec1, vec2):
+    """Returns the rotation matrix that aligns vec1 to vec2."""
+    a = (np.asarray(vec1) / np.linalg.norm(vec1)).reshape(3)
+    b = (np.asarray(vec2) / np.linalg.norm(vec2)).reshape(3)
+    v = np.cross(a, b)
+    c = np.dot(a, b)
+    s = np.linalg.norm(v)
+    K = np.array([[0, -v[2], v[1]], [v[2], 0, -v[0]], [-v[1], v[0], 0]])
+    return np.eye(3) + K + K.dot(K) * ((1 - c) / (s**2))
```
jacobmarks, line 360: This will not be needed

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,394 @@
...
+import fiftyone.core.validation as fov
+from fiftyone.core.odm import DynamicEmbeddedDocument
+
+o3d = fou.lazy_import("open3d", callback=lambda: fou.ensure_package("open3d"))
+
+logger = logging.getLogger(__name__)
+
+Vec2Tuple = Tuple[int, int]
+Vec3Tuple = Tuple[int, int, int]
+
```
jacobmarks, line 30: Typically const vars are given all CAPS names, so SHADING_GRADIENT_MAP. Also, why the specific values 0.111, 0.222, etc.? What does 1/9 have to do with shading?

jacobmarks (changes requested):
It's coming along well! Main thing to change is how the orthographic projections are done. I'd also add more examples in the docstring that show how the various arguments come into play.

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,394 @@
...
+        shading_mode: an optional field to specify how to shade the points.
+            "intensity" and "rgb" are only valid if the header contains
+            "rgb" flag, or else it'll default to "none" mode. In "none" mode,
+            all points are shaded white.
+        rel_dir (None): an optional relative directory to strip from each input
+            filepath to generate a unique identifier that is joined with
+            ``output_dir`` to generate an output path for each image. This
+            argument allows for populating nested subdirectories in
+            ``output_dir`` that match the shape of the input paths
+        projection_normal (None): (experimental) the normal vector of the plane
```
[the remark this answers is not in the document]
brimoor: I added the `projection_normal`. The use case is what if user's data isn't oriented z-up? They can provide `projection_normal=[0, 1, 0]` to project along y-axis instead.

The way it's implemented is to rotate `projection_normal -> [0, 0, 1]` so that the rest of the code can assume z-up.

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,394 @@
...
+        min_bound=min_bound, max_bound=max_bound
+    )
+
+    # crop bounds and translate so that min bound is at the origin
+    pc = pc.crop(bbox).translate((-min_bound[0], -min_bound[1], -min_bound[2]))
+
+    # this strategy is arbitrarily chosen and might need to be tuned more
+    # we might also want to expose this as a parameter, since there are
+    # multiple subsampling strategies possible (voxel-based, uniform)
+    if len(np.asarray(pc.points)) > 1e6:
```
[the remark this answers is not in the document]
brimoor: How about no subsampling by default and just expose subsampling params as optional?

fiftyone/utils/utils3d.py
brimoor: See previous comment. I think `projection_normal` is useful.

fiftyone/utils/utils3d.py
brimoor: Side comment: Ben started adding type hints to App server code, but none of `fiftyone.core` or `fiftyone.utils` have type hints currently, so I'd rather just omit them from here for consistency.

brimoor: @sashankaryal wrt the gradient map, we have [this function](https://github.com/voxel51/fiftyone/blob/dfb81b6fffb8ef9bb36bcb0d534f357e664ccfc8/fiftyone/core/config.py#L377) that can generate a colormap with any number of discrete colors and can support a variety of custom colorscales as documented there. I'd recommend leveraging that so user can customize the colorscale in a familiar way.

brimoor: Also, can you update so this method return images represented as `np.uint8` arrays in `[0, 255]`. In general, throughout the library, if we say we return an "image", we generally mean uint8, rather than float, especially when you're already discretizing.

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,394 @@
...
+        raise ValueError("%s has no point cloud slices" % type(samples))
+
+    slice_name = next(iter(point_cloud_slices))
+
+    if len(point_cloud_slices) > 1:
+        logger.warning(
+            "Found multiple point cloud slices; using '%s'", slice_name
+        )
+
+    return slice_name
```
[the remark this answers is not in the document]
brimoor: That's already documented.

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,394 @@
...
+            intensities_normalized_t = (colors[:, 0] - min_intensity) / (
+                max_intensity - min_intensity
+            )
+            # map intensity value to RGB
+            rgb_refs = _clamp_to_discrete(
+                intensities_normalized_t, list(shading_gradient_map.keys())
+            )
+            rgbs = np.array([shading_gradient_map[v] for v in rgb_refs])
+            image_map[np.int_(points[:, 0]), np.int_(points[:, 1]), :] = rgbs
+    elif shading_mode == "height":
```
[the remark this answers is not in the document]
brimoor: Per earlier comment, data will always be in z-up format after `projection_normal` is applied.

fiftyone/utils/utils3d.py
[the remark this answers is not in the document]
sashankaryal: Will parametrize it, thanks!

fiftyone/server/metadata.py
```diff
@@ -53,10 +53,21 @@ async def get_metadata(
         metadata dict
     """
     filepath = sample["filepath"]
-    is_video = media_type == fom.VIDEO
     metadata = sample.get("metadata", None)
     urls = _create_media_urls(collection, sample, url_cache)
 
+    has_orthographic_projection_metadata = (
```
brimoor, line 59: Here you're handling the case where the collection's `media_type` is `point-cloud`, but we could also now handle the case where the collection's `media_type` is `group` but the user has selected a point cloud slice to visualize in the grid via the group slice selector.

Currently, the group slice selector only includes image/video slices, but that logic could be implemented to also include any point cloud slices if the dataset has an `OrthographicProjectionMetadata` field.

This can be a separate PR if appropriate.

fiftyone/server/metadata.py
```diff
@@ -53,10 +53,21 @@ async def get_metadata(
...
+        media_type == fom.POINT_CLOUD
+        and "orthographic_projection_metadata" in sample
```
brimoor, line 61: I think this should be implemented by checking for a field of type `OrthographicProjectionMetadata` rather than specifically checking for the field name that the util in this PR happens to use.

There is special logic to prevent users from renaming default fields like `filepath` and `metadata`, but there is no logic protecting the `orthographic_projection_metadata` field name right now (and I don't think we need to add that), so we should be prepared that a user might rename the field to something shorter.

brimoor (changes requested):

fiftyone/server/metadata.py
brimoor: FYI I directly exposed this field as a `metadata_field` parameter of the util to make it easier for users to pick a different name if they want.

If you have concerns about type-based rather than name-based logic here, LMK

tests/unittests/utils3d_tests.py
```diff
@@ -0,0 +1,282 @@
...
+
+class BaseOrthographicProjectionTests(unittest.TestCase):
+    def setUp(self) -> None:
+        self.temp_dir = tempfile.TemporaryDirectory()
+        self.test_pcd_path = os.path.join(self.temp_dir.name, "test_pcd.pcd")
+
+    def tearDown(self) -> None:
+        self.temp_dir.cleanup()
+
+    def assertValidProjection(self, metadata_field, expected_image_path=None):
```
brimoor, line 30: Some of these tests fail locally for me with the following error:

```
FileNotFoundError: [Errno 2] No such file or directory: 'specs/3d/100x100_seed_1_none.png'
```

Seems like an absolute vs relative path issue?

fiftyone/utils/utils3d.py
```diff
@@ -0,0 +1,422 @@
...
+            cloud data. Only applicable if ``samples`` is a grouped collection.
+            If ``samples`` is a grouped collection and this parameter is not
+            provided, the first point cloud slice will be used
+        out_group_slice (None): the name of a group slice to which to add new
+            samples containing the feature images/maps. Only applicable if
+            ``samples`` is a grouped collection
+        metadata_field ("orthographic_projection_metadata"): the name of the
+            field in which to store :class:`OrthographicProjectionMetadata`
+            instances for each projection
+        shading_mode (None): an optional shading algorithm for the points.
```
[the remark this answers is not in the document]
sashankaryal: Shading looks interesting in 3D, in 2D projections, no shading offers the best contrast.

brimoor (approved):
LGTM

PR #2657, opened 2023-02-14
fix the viewbar bug

lanzhenw:

Fix the issue of ViewbarMachine parser not validating non-array type correctly. 

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
+++ b/app/packages/core/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
@@ -120,7 +120,8 @@ export const PARSER = {
         array = stripped.split(",");
       }
       return (
-        Array.isArray(value) &&
+        typeof value !== "string" &&
+        Array.isArray(array) &&
         array.every((e) => PARSER.field.validate(e, fields))
       );
     },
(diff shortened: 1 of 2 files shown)


app/packages/core/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
```diff
@@ -152,7 +152,7 @@ export const PARSER = {
       } catch {
         array = stripped.split(",");
       }
-      return Array.isArray(array) && array.every((e) => PARSER.str.validate(e));
+      return Array.isArray(value) && array.every((e) => PARSER.str.validate(e));
```
manivoxel51, line 155: Lanny, not sure we need to change these. the change causing an issue where the value disappears when I use MatchTags() 🤔 

my 2cents as to why we don't need these changes
User can pass an object to the field that could potentially be parsed to an array ex: `'["dd", "cc"]';` -> ["dd", "cc"] // type is array

User can also pass an object to the field that could be parsed into an object (hence not an array). which the validation should fail and it would in `Array.isArray(array)` part

what do you think?

manivoxel51 (commented):
good start Lanny! 💯 

left a thought - we can chat irl as well

app/packages/core/src/components/ViewBar/ViewStage/viewStageParameterMachine.ts
lanzhenw: Previously when value is a string ("kite in the sky"), it can pass the `Array.isArray(array)` check and will then gets treated like a str<list>, then cause the build to break, as it doesn't have the array method. 

I see what you are pointing out. What I did on "similarity-textprompt" branch earlier was to add an additional condition to make sure value is not string type. I will revert to that.

PR #2667, opened 2023-02-16
Check slug instead of the entire dataset before saving views

manivoxel51:

@lanzhenw recently noticed that saving views fails (intermittently) raising `raise ValueError("Cannot save view into a different dataset")`

Digging into it, It fails every other time or so after trying to save a view (whatever the view be). For some reason (ideas are welcomed) the `view._root_dataset is not self:` condition started passing/failing depending on how many times you try to save a view. 

Fix: making the condition check based on slug is more efficient and seems to fix the issue.

https://user-images.githubusercontent.com/109545780/219476413-f7375aab-221d-4149-b69f-2678b10b961c.mov

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -3226,7 +3226,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
             overwrite (False): whether to overwrite an existing saved view with
                 the same name
         """
-        if view._root_dataset.slug != self.slug:
+        if view._root_dataset._doc.id != self._doc.id:
             raise ValueError("Cannot save view into a different dataset")
 
         view._set_name(name)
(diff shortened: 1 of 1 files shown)


kaixi-wang:
at the end of the video, does the error get triggered by using the sidebar? (the video stops before any error is shown)

manivoxel51:
> at the end of the video, does the error get triggered by using the sidebar? (the video stops before any error is shown)

The end is the fix :p The beginning shows how it breaks, then the code change, and the fix. I had to cut the video - size limit has changed I believe

fiftyone/core/dataset.py
```diff
@@ -3226,7 +3226,7 @@ def save_view(
             overwrite (False): whether to overwrite an existing saved view with
                 the same name
         """
-        if view._root_dataset is not self:
+        if view._root_dataset.slug != self.slug:
```
kaixi-wang, line 3229:
```suggestion
        if view._root_dataset._doc.id != self._doc.id:
```

kaixi-wang (commented):
Can we compare id's instead? I want to avoid potential issues that can come into play with renaming.

kaixi-wang (approved):
image

PR #2670, opened 2023-02-17
bugfix: file already exist when exporting with media

oddeirikigland:

When exporting a dataset with export_media=True, it fails if you have exported the dataset earlier. The copy file method does not handle if the file already exists.   

A local run with version 0.18.0. Added some new images to the dataset, and tried to export, but failed on the images already exported.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/data/exporters.py
+++ b/fiftyone/utils/data/exporters.py
@@ -1204,7 +1204,8 @@ class MediaExporter(object):
                 uuid = self._get_uuid(outpath)
 
             if self.export_mode == True:
-                etau.copy_file(media_path, outpath)
+                if not os.path.exists(outpath):
+                    etau.copy_file(media_path, outpath)
             elif self.export_mode == "move":
                 etau.move_file(media_path, outpath)
             elif self.export_mode == "symlink":


brimoor (changes requested):
I can't reproduce this bug. `etau.copy_file()` just uses `shutil.copy()`, and the latter's [documentation](https://docs.python.org/3/library/shutil.html#shutil.copy) states that it should always overwrite existing files:

```
shutil.copy(src, dst, *, follow_symlinks=True)
Copies the file src to the file or directory dst. src and dst should be [path-like objects](https://docs.python.org/3/glossary.html#term-path-like-object) or strings. If dst specifies a directory, the file will be copied into dst using the base filename from src. If dst specifies a file that already exists, it will be replaced. Returns the path to the newly created file.
```

lawchingman (approved):

lawchingman (approved):

lawchingman (approved):

oddeirikigland:
> I can't reproduce this bug. `etau.copy_file()` just uses `shutil.copy()`, and the latter's [documentation](https://docs.python.org/3/library/shutil.html#shutil.copy) states that it should always overwrite existing files:
> 
> ```
> shutil.copy(src, dst, *, follow_symlinks=True)
> Copies the file src to the file or directory dst. src and dst should be path-like objects
> or strings. If dst specifies a directory, the file will be copied into dst using the base
> filename from src. If dst specifies a file that already exists, it will be replaced.
> Returns the path to the newly created file.
> ```

Inside shutil.copy in the copyfile method, the SameFileError is raised for my example.
```
def copyfile(src, dst, *, follow_symlinks=True):
    """Copy data from src to dst in the most efficient way possible.

    If follow_symlinks is not set and src is a symbolic link, a new
    symlink will be created instead of copying the file it points to.

    """
    sys.audit("shutil.copyfile", src, dst)

    if _samefile(src, dst):
        raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
```

brimoor:
I see so looks like your issue is not overwriting *any* existing file but rather overwriting the exact same file with itself:
https://stackoverflow.com/questions/57083594/disable-samefileerror-exception-in-shutil-copy

oddeirikigland:
> I see so looks like your issue is not overwriting _any_ existing file but rather overwriting the _exact same file_ with itself: https://stackoverflow.com/questions/57083594/disable-samefileerror-exception-in-shutil-copy

Yes. So when updating an exported dataset, with what I suggest, you skip those files already exported to the export dataset folder.

brimoor:
Okay, I went ahead and patched this issue in https://github.com/voxel51/eta/pull/583. which will be included in the next release `fiftyone==0.20.0`, scheduled to come out March 15th.

We don't want to *always* skip the copy when the destination file already exists, because it should be allowed to overwrite an existing file with a different file.

PR #2678, opened 2023-02-22
Fix docstring for Dataset.from_dir()

ehofesmann:
Update `export_dir` to `dataset_dir` in the `fo.Dataset.from_dir()` docstring.


brimoor (approved):
LGTM

PR #2679, opened 2023-02-22
Operators

ritch:
Extend the FiftyOne App with Python or Typescript Operators!


imanjra (approved):

imanjra:
Tested and reviewed changes over a screen share due to large amount of changes. Minor fixes and changes will be part of a separate PR

PR #2680, opened 2023-02-22
externalize plugin deps

ritch:
Allow for plugins to load components and utilities from runtime instead of compiling their own, with its own state at runtime. This fixes issues where the fetch function will point to the wrong URL and cuts down on build times for plugins that use components.

### DIFF
--- a/app/packages/plugins/src/index.ts
+++ b/app/packages/plugins/src/index.ts
@@ -4,6 +4,8 @@ import * as _ from "lodash";
 import React, { FunctionComponent, useEffect, useMemo, useState } from "react";
 import ReactDOM from "react-dom";
 import * as recoil from "recoil";
+import * as foc from "@fiftyone/components";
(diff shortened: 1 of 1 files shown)


app/packages/plugins/src/index.ts
```diff
@@ -4,16 +4,30 @@ import * as _ from "lodash";
...
     React: any;
     ReactDOM: any;
     recoil: any;
     __fos__: any;
+    __foc__: any;
+    __fou__: any;
   }
 }
 
+if (typeof window !== "undefined") {
```
imanjra, line 21: Should we move these to identical if block below and de-duplicate?

imanjra (approved):

app/packages/plugins/src/index.ts
ritch: I'll deduplicate before merging...

benjaminpkane (approved):
LGTM

PR #2683, opened 2023-02-23
Skipping downward migrations for future-but-compatible versions

brimoor:
Suppose an SDK client with version `X` and `FIFTYONE_DATABASE_ADMIN=true` encounters a dataset created by a different client with version `Y` such that `Y > X` but `Y` is in the compatibility range of `X`.

Previously, when `FIFTYONE_DATABASE_ADMIN=true`, an error would be raised when trying to load the dataset using client `X`, because the migration runner would complain that `Y > X` and therefore it had no way to know whether any migrations exist for the `Y --> X` downgrade path.

Now, as long as `Y` is in the compatibility range of `X`, the dataset will be loaded without migration.

This also applies to databases themselves. So, for example, since the compatibility range of `0.19.1` is `>=0.19,<0.20`, it will now be possible for a client running `0.19.1` with `FIFTYONE_DATABASE_ADMIN=true` to connect to a database with version `0.19.2`. Previously, an error would be raised in that case upon `import fiftyone`.

### DIFF
--- a/fiftyone/migrations/runner.py
+++ b/fiftyone/migrations/runner.py
@@ -115,7 +115,7 @@ def migrate_database_if_necessary(destination=None, verbose=False):
     default_destination = destination is None
 
     if default_destination:
-        if not fo.config.database_admin and _is_compatible_version(head):
+        if _can_skip_migration(head):
             return
 
         destination = foc.VERSION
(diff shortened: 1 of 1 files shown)


swheaton (approved):
lgtm

PR #2685, opened 2023-02-23
Adding tensorflow-macos requirement option

brimoor:


jacobmarks (approved):
Didn't try every single model, but tried a few and it seems to be working for me! LGTM 👍

PR #2686, opened 2023-02-23
Add support for negative integer mask targets

brimoor:
Previous versions of FiftyOne technically allowed negative integer mask targets (although the docs recommend only using non-negative integers), but [this update](https://github.com/voxel51/fiftyone/commit/171071da26e432b6ea2d831b589d09f6fe79b7e1) changed the implementation in such a way that existing datasets that *do* happen to contain mask targets with negative integers would no longer be loadable.


swheaton (approved):
weird but ok lgtm

sashankaryal (approved):

PR #2687, opened 2023-02-24
Documenting on-the-fly custom embedded document creation

brimoor:
Documents a previously-possible-but-undocumented pattern for creating custom embedded documents:
- declare an empty `DynamicEmbeddedDocument` field
- add dynamic attributes to it as desired

## Example usage

```py
import fiftyone as fo

dataset = fo.Dataset()

# Define an empty embedded document field
dataset.add_sample_field(
    "camera_info",
    fo.EmbeddedDocumentField,
)

# Declare a new `camera_id` attribute
dataset.add_sample_field("camera_info.camera_id", fo.StringField)
assert "camera_info.camera_id" in dataset.get_field_schema(flat=True)

# Includes a new `quality` attribute
sample1 = fo.Sample(
    filepath="/path/to/image1.jpg",
    camera_info=fo.DynamicEmbeddedDocument(
        camera_id="123456789",
        quality=51.0,
    ),
)

sample2 = fo.Sample(
    filepath="/path/to/image2.jpg",
    camera_info=fo.DynamicEmbeddedDocument(camera_id="123456789"),
)

dataset.add_samples([sample1, sample2], dynamic=True)
assert "camera_info.quality" in dataset.get_field_schema(flat=True)

# Populate a new `description` attribute
dataset.set_values("camera_info.description", ["foo", "bar"], dynamic=True)
assert "camera_info.description" in dataset.get_field_schema(flat=True)
```

### DIFF
--- a/fiftyone/core/odm/mixins.py
+++ b/fiftyone/core/odm/mixins.py
@@ -1552,7 +1552,11 @@ def _add_field_doc(field_docs, field_or_doc):
             field_docs[i] = new_field_doc
             return
 
-    field_docs.append(new_field_doc)
+    try:
+        field_docs.append(new_field_doc)
+    except ReferenceError:
+        # mongoengine seems to lose references to things... it's okay
+        pass
 
 
 def _update_field_doc(field_docs, field_name, field):
--- a/docs/source/user_guide/using_datasets.rst
+++ b/docs/source/user_guide/using_datasets.rst
@@ -3377,29 +3377,128 @@ predefined types and optional default values, while the
 users to populate arbitrary custom fields at runtime, like FiftyOne's
 :ref:`builtin label types <using-labels>`.
 
-To use this feature, simply define some custom embedded document classes in a
-`foo.bar` module, using the appropriate types from the
+.. _defining-custom-documents-on-the-fly:
+
+Defining custom documents on-the-fly
+------------------------------------
+
+The simplest way to define custom embedded documents on your datasets is to
+declare empty |DynamicEmbeddedDocument| field(s) and then incrementally
+populate new :ref:`dynamic attributes <dynamic-attributes>` as needed.
+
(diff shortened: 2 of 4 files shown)


fiftyone/core/odm/mixins.py
```diff
@@ -1552,7 +1552,11 @@ def _add_field_doc(field_docs, field_or_doc):
             field_docs[i] = new_field_doc
             return
 
-    field_docs.append(new_field_doc)
+    try:
+        field_docs.append(new_field_doc)
+    except ReferenceError:
+        # mongoengine seems to lose references to things... it's okay
```
swheaton, line 1558: uh what? are we sure this isn't hiding some other problem?

brimoor: It's hiding something, but it's some weird mongoengine garbage collection issue that I don't understand.

Sit down for this one... without the `try-except` I added, the following code works if you paste the whole thing into ipython at once, but it raises an error if you wait one second between the `add_sample_field()` calls:

```py
import fiftyone as fo

dataset = fo.Dataset()

dataset.add_sample_field(
    "camera_info",
    fo.EmbeddedDocumentField,
)

# This fails if you wait one second between running this and the above code
dataset.add_sample_field("camera_info.camera_id", fo.StringField)
```

brimoor: I'm confident that everything works as expected if we just ignore this error. The last thing that happens at the end of `add_sample_field()` is a call to `dataset._reload()`, which recreates the relevant mongoengine classes and should paint over any weird issues due to mongoengine's `append()` implementation failing.

The unit tests I added verify that the dataset functions as intended.

brimoor: Thanks for the deep dive!

PR #2693, opened 2023-02-27
[WIP] Change look and feel of selected label tags

### LINKED ISSUE #2445: [FR] Change look and feel of selected labels
It would be helpful if the look-and-feel of any boxes with selected label tags would change to indicate that they are the ones with the specified tag (especially important when in "match" mode, but would also be fine in "select" mode).

A natural thing would be to change their box color to match the color of the field assigned to that label tag. In addition, for spatial label types like Detections, perhaps increasing the thickness of the boxes would further help emphasize the selected labels. Or adopt a similar dotted border used when physically selecting the labels (but we need to avoid confusion between these selection operations).

Screen Shot 2021-08-05 at 7 41 23 PM

Screen Shot 2021-08-05 at 7 43 20 PM

lanzhenw:
fix #2445 

label tag color is applied to filtered detections:

Screenshot 2023-02-27 at 11 03 19 AM

When label tags are selected in filter (filter/match mode), change their box color to match the color of the field assigned to that label tag. 

Highlight the label box color if they are tagged and the tags are active in the filter. 

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/looker/src/elements/common/tags.ts
+++ b/app/packages/looker/src/elements/common/tags.ts
@@ -60,7 +60,7 @@ export class TagsElement<State extends BaseState> extends BaseElement<State> {
   renderSelf(
     {
       config: { fieldSchema },
-      options: { activePaths, coloring, timeZone },
+      options: { activePaths, coloring, timeZone, selectedLabelTags, ...props },
       playing,
     }: Readonly<State>,
     sample: Readonly<Sample>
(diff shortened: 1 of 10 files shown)


lanzhenw:
selected label tags color is applied to polylines area and lines:
Screenshot 2023-03-09 at 12 20 43 PM
Screenshot 2023-03-09 at 12 20 35 PM

brimoor (changes requested):
I played around with this feature a bit just now and noticed a few things:

## Bugs

- In the grid screenshot below, shouldn't the `person: X` tags be orange to match the label field color? Not sure where the red is coming from (UPDATE: oh each tag value is getting its own color. there's a lot of colors, kind of confusing)
- In the modal screenshot below, shouldn't the `person` label also be in orange?

## Usability feedback/discussion points

- In the grid view, I think showing both `person: X` and changing the color of the box is duplicative. Perhaps we should consider adding a toggle in the settings menu that configures whether to show the `person: X` OR change the label's color. Currently with both options applied together, the scene is a bit cluttered
- In the modal, there's no way to show all labels while also showing only the tagged labels in the different color. If you want to see the tagged labels in a different color, you have to filter the labels, which hides all other labels and thus defeats the purpose of changing the color in the first place. In order to support this, we could add the "show samples with selected labels" to the filter tray (as we use in the grid). This terminology is a bit weird, of course, since it's always the same sample. Perhaps we could instead update so that whenever the top-level `label tags` checkbox is selected in the modal, it immediately changes the color of all tagged labels to the label tag color? But then if there were label tags `A,B,C` in the scene, there wouldn't be a way to show only labels with tag `A` in a different color without also hiding other labels. So it's almost like there needs to be a different mode where the box colors are changed rather than actually filtering the boxes 🤔  

## Screenshots

Screen Shot 2023-03-09 at 9 33 27 PM

Screen Shot 2023-03-09 at 9 35 01 PM

PR #2696, opened 2023-02-28
Renaming and saving run results

brimoor:
Change log
- Adds support for renaming runs via `rename_annotation_run()`, `rename_evaluation()`, and `rename_brain_run()`
- Adds an optional `type` argument to `list_annotation_runs()`, `list_evaluations()` and `list_brain_runs()` to retrieve runs of specific type (eg similarity indexes or visualizations within the full set of brain keys)
- Adds `save()` method to all `RunResults` objects that allows for saving updates to results to the database


docs/source/user_guide/brain.rst
```diff
@@ -989,45 +989,18 @@ demonstrated empirical value in many model training exercises.
...
-When you run a brain method on a dataset, the run is recorded on the dataset,
-allowing you to retrive information about it later, delete it (along with any
-modifications to your dataset that were performed by it), or even retrieve the
-view into your dataset that you processed.
-
-Brain method runs can be accessed later by their `brain_key`:
-
-.. tabs::
-
-    .. tab:: Visualizations
```
swheaton, line 1001: Why are we deleting all this? Could still be useful as an FYI

swheaton (commented):
just one comment about repeated code.
lgtm other than that

fiftyone/core/runs.py
```diff
@@ -293,18 +312,77 @@ def _validate_fields_match(self, key, field_name, existing_info):
...
+            type = etau.get_class(type)
+
+        if type is not None:
+            keys = []
+            for key in run_docs.keys():
+                run_info = cls.get_run_info(samples, key)
+                if issubclass(run_info.config.run_cls, type):
+                    keys.append(key)
+        else:
+            keys = run_docs.keys()
```
kaixi-wang, line 339: if you only list the keys without the class, is it possible you end up with a list like ["train","train", ...]? Also, how useful is it to know keys without the class (if the key's themselves aren't descriptive)?

fiftyone/utils/eval/regression.py
```diff
@@ -163,12 +164,33 @@ def get_fields(self, samples, eval_key):
 
         return fields
 
+    def rename(self, samples, eval_key, new_eval_key):
```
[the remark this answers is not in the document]
brimoor: I was following along with how the `cleanup()` and `get_fields()` methods are implemented explicitly on each evaluation protocol.

The implementations for each protocol are currently similar, but there's not an inherent reason why they would always be the same, so I think some repeated code is okay in this case.

fiftyone/core/runs.py
brimoor: I'm not following the `["train", "train", ...]` comment. `run_docs` is a dict keyed by run keys.

wrt the second point, adding the `type` arg is a way to get keys of a particular type, which I agree might be valuable sometimes.

docs/source/user_guide/brain.rst
brimoor: I'll add it back.

fiftyone/utils/eval/regression.py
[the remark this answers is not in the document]
swheaton: See what you're saying but can't say I agree with the logic. I feel like this is why when we add something to OSS backend we will have multiple places we have to add it to. Which is probably fine if you're making the edit but anyone else might easily miss a spot.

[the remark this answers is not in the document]
brimoor: As it currently is, there are `XXXEvaluation` base classes for each type of evaluation protocol: classification, detection, regression, segmentation. 

The abstraction to remove duplicate code would be "this class of evaluation protocol supports either sample-level or frame-level fields. When processing sample-level fields, it populates a single field whose name matches the `eval_key`. When processing frame-level fields, it populates a frame-level and sample-level fields whose name matches `eval_key`." For that class of methods, we could abstract `get_fields()`, `rename()`, and `cleanup()` into a mixin and share the code.

It doesn't seem that useful to me though. Sometimes when there's a bunch of fancy subclass implementations it becomes less clear how to implement a new instance of the interface (eg a new label type to evaluate) because it seems like you have to follow how the base class does some things rather than just doing whatever you need for your case.

PR #2711, opened 2023-03-02
Add dataset ID to all documents

brimoor:
Adds the `_dataset_id` to all sample/frame documents in datasets.

This value is not currently used by the data model; it is primarily added in case users wish to have a sharded database where all information associated with a given dataset is stored on the same shard.


fiftyone/migrations/revisions/v0_20_0.py
```diff
@@ -0,0 +1,61 @@
+"""
+FiftyOne v0.20.0 revision.
+
+| Copyright 2017-2023, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
+
+def up(db, dataset_name):
```
swheaton, line 10: Is this going to be a slow upgrade process for large datasets? I have a limited concept of how fast adding a field to all samples and frame samples would be

swheaton (approved):
Where code is added, it LGTM.
One thing I don't feel I can comment on is whether all cases are covered where a document is created and we want to add dataset_id

fiftyone/migrations/revisions/v0_20_0.py
brimoor: Hand-wavy guess: on the order of a few seconds per million documents being updated. Depends on database configuration, load, etc of course.

It's definitely a more expensive operation

PR #2719, opened 2023-03-04
Add support for customizing progress bar rendering at method level

brimoor:
## Change log

- Adds an optional `progress` parameter to all methods that support rendering progress bars that allows for customizing progress bar rendering at the method level, overriding the default global `fo.config.show_progress_bars` setting
- Adds support for passing a callback function via the `progress` parameter rather than rendering the builtin progress bar (see below for example usages)

### Explicit progress control

```py
import fiftyone as fo
import fiftyone.zoo as foz

print(fo.config.show_progress_bars)  # True

# Don't render progress bars, despite global config default
dataset = foz.load_zoo_dataset("quickstart", progress=False)
dataset.compute_metadata(progress=False)
```

### Custom progress callbacks

```py
import fiftyone as fo
import fiftyone.zoo as foz

def print_progress(pb):
    if pb.complete:
        print("COMPLETE")
    else:
        print("PROGRESS: %0.3f" % pb.progress)

# Print progress at 10 equally-spaced increments
progress = fo.report_progress(print_progress, n=10)
dataset.compute_metadata(progress=progress)

# Print progress every 0.5 seconds
progress = fo.report_progress(print_progress, dt=0.5)
dataset.compute_metadata(progress=progress, overwrite=True)
```

## Usage in delegated operations

Screen Shot 2024-01-02 at 10 00 56 PM

### DIFF
--- a/docs/source/plugins/developing_plugins.rst
+++ b/docs/source/plugins/developing_plugins.rst
@@ -1198,6 +1198,26 @@ using the pattern show below:
 
         ctx.dataset.add_images_dir(images_dir, progress=progress)
 
+You can also use the builtin
+:class:`ProgressHandler <fiftyone.operators.ProgressHandler>` class to
+automatically forward logging messages to
+:meth:`set_progress() <fiftyone.operators.executor.ExecutionContext.set_progress>`
+as `label` values using the pattern shown below:
+
+.. code-block:: python
+    :linenos:
+
+    import logging
+    import fiftyone.operators as foo
+    import fiftyone.zoo as foz
+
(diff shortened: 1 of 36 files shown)


benjaminpkane (approved):
LGTM

brimoor (commented):
This was a user-contributed PR and I need to do an audit to make sure all instances of `progress` were correctly updated before merging.

brimoor:
Merged into `release/v0.23.3` instead in https://github.com/voxel51/fiftyone/pull/3979.

PR #2729, opened 2023-03-06
fix hardcoded orthographic_projection_metadata bug in app

### LINKED ISSUE #2721: [BUG] App assumes hard-coded orthographic_projection_metadata field name
It appears the App is hard-coded to assume the `orthographic_projection_metadata` field name, but the user should be allowed to pick a different name for the `OrthographicProjectionMetadata` field, as shown below:

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.select_group_slices("pcd")

fou3d.compute_orthographic_projection_images(
    view,
    (-1, 512),
    "/tmp/proj",
    metadata_field="opm",  # everything works with this omitted
)

session = fo.launch_app(view)
```

App loads, but images are not found:

Screen Shot 2023-03-04 at 2 39 27 PM

sashankaryal:
Fixes #2721

### DIFF
--- a/app/packages/state/src/hooks/useCreateLooker.ts
+++ b/app/packages/state/src/hooks/useCreateLooker.ts
@@ -75,11 +75,20 @@ export default <T extends FrameLooker | ImageLooker | VideoLooker>(
         constructor = ImageLooker;
       }
 
-      const sampleMediaFilePath =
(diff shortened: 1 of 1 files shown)


benjaminpkane (approved):
LGTM!

PR #2734, opened 2023-03-06
Empty samples page with reset button

manivoxel51:
Screen Shot 2023-03-08 at 3 55 09 PM
Screen Shot 2023-03-08 at 3 55 15 PM

Shows an Empty component for when there are no samples in a dataset or there are no samples in the given view.

Screen Shot 2023-03-08 at 3 54 52 PM
Screen Shot 2023-03-08 at 3 54 57 PM

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/EmptySamples.tsx
+++ b/app/packages/core/src/components/EmptySamples.tsx
@@ -5,21 +5,24 @@ import * as fos from "@fiftyone/state";
 import { SamplesHeader } from "./ImageContainerHeader";
 import { Typography } from "@mui/material";
 import ClearAll from "@mui/icons-material/ClearAll";
-import { Button } from "@fiftyone/components";
+import { Button, useTheme } from "@fiftyone/components";
+import ErrorImg from "../images/error.svg";
+import { ExternalLink } from "../utils/generic";
 
 const BaseContainer = styled.div`
   display: flex;
   align-items: center;
   justify-content: center;
-  flex-direction: row;
+  height: 100%;
 `;
 
 const Container = styled(SamplesHeader)`
-  height: 30vh;
+  height: auto;
   display: flex;
   align-items: center;
   justify-content: center;
   flex-direction: column;
+  min-width: 300px;
 `;
 
 const TextContainer = styled(Container)`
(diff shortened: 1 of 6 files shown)


app/packages/core/src/components/EmptySamples.tsx
```diff
@@ -0,0 +1,74 @@
...
+`;
+
+const ActionContainer = styled(TextContainer)`
+  justify-content: start;
+  display: flex;
+  position: relative;
+`;
+
+function EmptySamples() {
+  const loadedView = useRecoilValue<fos.State.Stage[]>(fos.view);
```
sashankaryal, line 39: nit: type parameter implicitly derived

app/packages/core/src/components/EmptySamples.tsx
```diff
@@ -0,0 +1,74 @@
...
+    fos.count({ path: "", extended: false, modal: false })
+  );
+  const setView = fos.useSetView();
+
+  const showEmptySamples = useMemo(
+    () => loadedView?.length && !totalSamples,
+    [totalSamples, loadedView]
+  );
+
+  if (!showEmptySamples) return null;
```
sashankaryal, line 50: nit: newline before return

sashankaryal (approved):
lgtm but i might not have the broadest context in terms of specs

brimoor (changes requested):
Thanks for adding this!

**Functional comment:** can you also handle the case where it is an empty *dataset* rather than a view? When the dataset itself is empty, there's no need for a "clear view" button, the user just needs to know that their dataset is empty:

```py
import fiftyone as fo

dataset = fo.Dataset()
```

**For fun:** what do you think of showing the 404 image from Teams instead? Potential layout:

```
[image]
Your view is empty
[Clear view button]
```

Screen Shot 2023-03-06 at 9 04 15 PM

manivoxel51:
> Thanks for adding this!
> 
> **Functional comment:** can you also handle the case where it is an empty _dataset_ rather than a view? When the dataset itself is empty, there's no need for a "clear view" button, the user just needs to know that their dataset is empty:
> 
> ```python
> import fiftyone as fo
> 
> dataset = fo.Dataset()
> session = fo.launch_app(dataset)
> ```
> 
> **For fun:** what do you think of showing the 404 image from Teams instead? Potential layout:
> 
> ```
> [image]
> Your view is empty
> [Clear view button]
> ```
> 
> Screen Shot 2023-03-06 at 9 04 15 PM

good idea!
I am thinking having both use the same design (our 404 page in teams) make sense too. I'll adapt the pr to
1. show empty page with image + `Your view is empty` title as text and `clear view` as button when there are view stages
2. show empty page with image + `Your dataset is empty` as title, no clear button. also, maybe add a code snippet here to show how to bring data into an existing dataset? or just leave the code snippet out

@brimoor what do you think?

brimoor:
Rather than including a code snippet for the empty dataset case, you could just link to this page on the topic:
https://docs.voxel51.com/user_guide/dataset_creation/index.html

brimoor (approved):
LGTM

PR #2736, opened 2023-03-06
bugfix: use wildcard pattern matching in app

sashankaryal:

Currently we use regex matching for string fields (like `id`, `filepath`, etc.). It breaks the app when the user enters invalid regex, example: `*`. This PR includes a fix such that we use wildcard matching instead of regex matching.

Tested locally. No automated tests.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/server/routes/values.py
+++ b/fiftyone/server/routes/values.py
@@ -5,14 +5,14 @@ FiftyOne Server /values route
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+import regex as re
 from starlette.endpoints import HTTPEndpoint
 from starlette.requests import Request
 
 import fiftyone.core.aggregations as foa
-
 import fiftyone.server.constants as foc
-from fiftyone.server.decorators import route
 import fiftyone.server.view as fosv
+from fiftyone.server.decorators import route
 
 
 class Values(HTTPEndpoint):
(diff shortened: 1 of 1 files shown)


kaixi-wang (commented):
The search arg only appears in the `CountValues(Aggregation)` class. I think it might be just used to match the paths in GraphQL to populate the sidebar counts (ie internal only but required to match user-named fields), but not something exposed to users... 

cc @benjaminpkane

fiftyone/server/routes/values.py
```diff
@@ -36,13 +36,20 @@ async def post(self, request: Request, data: dict):
 
         sort_by = "count" if count else "_id"
 
+        # escape special characters, add support for wilcard pattern matching
+        regex_safe_search = (
+            re.escape(search).replace(r"\*", ".*").replace(r"\?", ".")
```
kaixi-wang, line 41: I think this replacement doesn't work if a user starts with valid regex. 

https://user-images.githubusercontent.com/30936736/223288932-1ac61c29-ef1e-40d1-bb7b-751dcc61b5e0.mov

sashankaryal: In the frontend, we're supporting wildcard pattern matching, not regex matching. With wildcard pattern matching, the text you specified `.*kacey.*` would only match strings like `.ABCDkacey.ABCDEF`.

brimoor (approved):
LGTM, thanks!

PR #2743, opened 2023-03-07
Fix SelectGroupSlices params

### LINKED ISSUE #2720: [BUG] View bar crashes when loading SelectGroupSlices view stages
On `develop`, the view bar crashes when trying to load the view below, which includes a `select_group_slices("pcd")` stage:

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.select_group_slices("pcd")

fou3d.compute_orthographic_projection_images(view, (-1, 512), "/tmp/proj")

session = fo.launch_app(view)
```

Screen Shot 2023-03-04 at 2 37 36 PM

benjaminpkane:
Resolves #2720


fiftyone/core/stages.py
```diff
@@ -4092,7 +4092,7 @@ def _params(cls):
         return [
             {
                 "name": "slices",
-                "type": "NoneType|list<str>|str",
+                "type": "NoneType|str|list<str>",
```
brimoor, line 4095: There are a handful of other instances of `list<str>|str` in other view stages, which I thought was required so that `a,b` is parsed as `["a", "b"]` rather than being treated as a string.

Isn't there another way to address this issue that preserves the possibility of providing a single string or `list,of,string`?

brimoor (changes requested):

ritch (approved):

fiftyone/core/stages.py
benjaminpkane: Yes, fixed

brimoor (approved):
LGTM

PR #2744, opened 2023-03-07
Include ftfy and regex as default packages

brimoor:
## Change log

- Adds a missing `regex` requirement introduced in https://github.com/voxel51/fiftyone/pull/2736
- Adds `ftfy` as a default requirement too, so that the CLIP model from the zoo only requires that `torch + torchvision` be manually installed, which is nice because CLIP will be heavily used by folks wanting to leverage [text similarity in the App](https://github.com/voxel51/fiftyone/pull/2633)


sashankaryal (approved):
lookin good

PR #2745, opened 2023-03-07
Show group count in fast mode

### LINKED ISSUE #2602: [BUG] Group count doesn't show when stats=group and sidebar=fast
When the stats mode is "group" and sidebar mode is "fast", it would be preferable/consistent if the group count gets computed and rendered above the grid. Currently, it's only computed when sidebar mode is "all".

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Switch to statistics mode "group" and sidebar mode "fast" in the settings menu
```

Fast mode (group count not shown):
Screen Shot 2023-02-02 at 6 04 18 PM

All mode (group count is shown):
Screen Shot 2023-02-02 at 6 04 25 PM

benjaminpkane:
Resolves #2602


imanjra (approved):
lgtm 👍

PR #2749, opened 2023-03-07
fixes #2472

### LINKED ISSUE #2472: [BUG] Shift key conflict with grouped datasets in App
In the expanded modal for grouped datasets, pressing the `shift` key has two different meanings:
- Looker2D: hide labels (*only* while shift key is held down)
- Looker3D: switch to pan mode rather than rotate mode when mouse click/drag in 3D space

As the video below shows, if Looker2D has focus but my mouse is hovering over Looker3D, pressing and holding shift will simultaneously hide the labels in Looker2D while allowing me to click + drag to pan in Looker3D. However, this causes Looker2D to enter an unintended state where labels are permanently hidden even though the shift key is no longer pressed.

Ways to resolve this:
- Ensure that Looker2D show/hide labels toggle always resets to "show" when Looker2D loses focus (I believe the issue with the above is that `shift` is only released once focus has shifted to Looker3D, so there is never an on-shift-release event triggered)
- Update logic so that `shift` hotkey only applies to the visualizer under in which the mouse currently has focus
- Change the `shift` hotkey to something else in Looker2D (I don't like this option)
- Something else?

https://user-images.githubusercontent.com/25985824/210190666-8f7631b0-c7d9-4a40-9728-504b6cd9ff39.mov

ritch:
Fixes #2472  by fixing existing support for this feature - only handle events when the looker has mouse over.

### DIFF
--- a/app/packages/core/src/components/Modal/Looker.tsx
+++ b/app/packages/core/src/components/Modal/Looker.tsx
@@ -64,9 +64,6 @@ const Looker = ({ lookerRef, onClose }: LookerProps) => {
   const [id] = useState(() => uuid());
 
   const sampleData = useRecoilValue(fos.modal);
-  if (!sampleData) {
-    throw new Error("bad");
-  }
   const { sample } = sampleData;
 
   const theme = useTheme();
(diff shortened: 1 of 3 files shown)


imanjra (approved):
lgtm

PR #2753, opened 2023-03-08
looker3d improvements

sashankaryal:

- Add support for grid at origin which is always perpendicular to the up vector
  - `ToggleGridHelper.tsx`
  - `Environment.tsx`
  - `Looker3d.tsx`
- Display arrow helper when up vector is not orthonormal
  - `Environment.tsx`
- Camera top position is no longer z-biased
  - `renderables/pointcloud/index.tsx`
- Add support for non-attenuated points. This should improve how pointclouds look since all points will now have constant size. This also fixes the shadow-y artifact looker3D had when viewed from certain angles
  - `renderables/pointcloud/index.tsx`
  - `shaders.tsx`
- Correct normalization bug when shading by height
  - `renderables/pointcloud/index.tsx`
  - `shaders.tsx`
- Minor bug fixes and refactoring

Screenshot 2023-03-08 at 11 53 28 AM

Above, is a screenshot of a pointcloud with non-standard up = `[0,1,1]`. The yellow arrow helper vector points in the direction of the up vector whereas the grid passes through the origin normal to the up vector.

Locally. No automated testing.

        notes for FiftyOne users.

-   [ ] Other


app/packages/looker-3d/src/renderables/pointcloud/shaders.tsx
```diff
@@ -131,7 +136,7 @@ const ShadeByIntensityShaders = {
 
     vec4 mvPosition = modelViewMatrix * vec4(pos, 1.0);
 
-    gl_PointSize = pointSize * (1.0 / length(mvPosition.xyz));
+    gl_PointSize = pointSize * (isPointSizeAttenuated ? (1.0 / length(mvPosition.xyz)) : 1.0);
```
ritch, line 139: Not sure I fully understand this - IIUC this means that customizing the point size disables the distance based sizing - or in other words a point near the camera and far from the camera will appear the same size.

app/packages/state/src/hooks/useCreateLooker.ts
```diff
@@ -56,7 +56,10 @@ export default <T extends FrameLooker | ImageLooker | VideoLooker>(
 
       const mimeType = getMimeType(sample);
 
-      if (mimeType !== null) {
+      // checking for pcd extension instead of media_type because this also applies for group slices
+      if (urls.filepath.endsWith(".pcd")) {
```
ritch, line 60: This seems a bit contradictory to the plugin "activator" function. Eg. does this mean that plugin devs cannot provide their own pcd visualizer?

app/packages/looker-3d/src/renderables/pointcloud/index.tsx
```diff
@@ -72,20 +74,21 @@ export const PointCloudMesh = ({
     }
   }, [boundingBox, pointsGeometry, points, onLoad]);
 
-  if (minZ === undefined) {
+  if (!minZ) {
```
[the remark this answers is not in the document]
sashankaryal: Ah, good point! I should have checked for `null` + `undefined` since that's what was causing the issue:
```
  const minZ = _.get(settings, "pointCloud.minZ", null);
```

ritch (approved):

PR #2754, opened 2023-03-08
Fix Map Panel Crash when Extended Selection is Patch ids

ritch:
image

When lassoing patch emeddings, instead of crashing, we show an existing "No Data" message (the same if no samples match).


benjaminpkane (approved):
LGTM 👍

PR #2762, opened 2023-03-10
bugfix: fix the nonetype bug from adding type

lanzhenw:

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/server/query.py
+++ b/fiftyone/server/query.py
@@ -478,9 +478,12 @@ async def serialize_dataset(
             data.group_slice = collection.group_slice
 
         for brain_method in data.brain_methods:
-            setattr(
-                brain_method.config, "type", brain_method.config.type().value
+            value = (
+                brain_method.config.type().value
+                if brain_method.config.type() is not None
+                else None
             )
+            setattr(brain_method.config, "type", value)
 
         return data
 


fiftyone/server/query.py
```diff
@@ -478,9 +478,12 @@ def run():
...
 
         for brain_method in data.brain_methods:
-            setattr(
-                brain_method.config, "type", brain_method.config.type().value
+            value = (
+                brain_method.config.type().value
+                if brain_method.config.type() is not None
+                else None
             )
+            setattr(brain_method.config, "type", value)
```
kaixi-wang, line 486: I'm a little confused here. You're reading the value from brain_method and setting the same value on brain method?

lanzhenw: We added the `type` property in the graphQL on query.py. Brain_method.config does not have`type`. 
When the user submits the similarity run, `/sort` route would refresh the dataset with `serialize_dataset`. And brain_method.config does not have `type`. That's why we are adding `type` to `brain_method.config` here.

benjaminpkane, line 486:
```suggestion
            value = brain_method.config.type()
            setattr(brain_method.config, "type", value.value if value else None)
```

benjaminpkane (approved):
LGTM!

PR #2765, opened 2023-03-10
Fix frames support streaming

benjaminpkane:
Fixes flipped logic for adding a support in the lookup pipeline for `/frames`. Currently, a BSON limit is encountered for viewing long/large video samples, which is a regression.


allenleetc (approved):
I built this and it fixed the harderror on my test dataset/movie

PR #2767, opened 2023-03-10
Fixing an issue with dynamic embedded list schemas

brimoor:
Fixes a bug with `dynamic=True` expansion for list fields that contain embedded documents.


benjaminpkane (approved):
LGTM

PR #2770, opened 2023-03-11
Adding validation to set_values()

brimoor:
The `set_values()` method previously did not validate the provided values against the type of existing fields. This PR fixes that.


benjaminpkane (approved):
LGTM

PR #2773, opened 2023-03-12
fixes app crashing while filtering keypoints labels (view.py) & keypoints missing metrics

akshitpriyesh:
1. Changes in fiftyone/server/view.py --> As the previous code looks only for numeric classes so when the user is filtering wrt. labels the app was crashing . Made the changes in func "_make_keypoint_list_filter" now code will able to handle those cases.
2. Added keypoints support where it was missing.
3. Created test metrics for keypoints.

its thoroughly tested on thousands of images.

## Release Notes
Keypoints filter bug fix and extended keypoints support.

Yes

-   [ x] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

Fixes app crashing while filtering keypoints labels & extended keypoints missing metrics. This change is specific to keypoints function only. The changes are made only on keypoints functions it is not interfering other functions. 

-   [x ] App: FiftyOne application changes
-   [x ] Core: Core `fiftyone` Python library changes
-   [ x] Documentation: FiftyOne documentation changes
-   [ ] Other

### DIFF
--- a/fiftyone/core/patches.py
+++ b/fiftyone/core/patches.py
@@ -23,7 +23,7 @@ import fiftyone.core.validation as fova
 import fiftyone.core.view as fov
 
 
-_PATCHES_TYPES = (fol.Detections, fol.Polylines)
+_PATCHES_TYPES = (fol.Detections, fol.Polylines, fol.Keypoints)
 _NO_MATCH_ID = ""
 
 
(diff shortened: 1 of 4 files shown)


brimoor (changes requested):
@akshitpriyesh thanks for the PR! It looks like you're fixing a bug with keypoint filtering and also separately adding support for evaluating keypoints. Can you split these two items into separate PRs so they can be reviewed separately? Thanks! 🤗

PR #2774, opened 2023-03-12
fixes app crashing while filtering keypoints labels (view.py)

akshitpriyesh:
Changes in fiftyone/server/view.py --> As the previous code looks only for numeric classes so when the user is filtering wrt. labels the app was crashing. Made the changes in func "_make_keypoint_list_filter" now code will able to handle those cases.

its thoroughly tested on thousands of images.

## Release Notes
Keypoints filter bug fix to avoid app crashing while filtering categorical variables.

yes

        notes for FiftyOne users.

Fixes app crashing while filtering keypoints labels. This change is specific to keypoints function only. The changes are made only on keypoints functions it is not interfering with other functions.

-   [x ] App: FiftyOne application changes
-   [ ] Other

### DIFF
--- a/fiftyone/server/view.py
+++ b/fiftyone/server/view.py
@@ -640,6 +640,15 @@ def _make_keypoint_list_filter(args, view, path, field):
 
             return {"filter": expr}
 
+    if isinstance(field.field, (fof.StringField)):
+        f = F(name)
+        if 'values' in args:
+            val = args["values"]
+            expr = f.is_in(val)
+            if args["exclude"]:
+                expr = ~expr
+            return {"filter": expr}
+
 
 def _apply_others(expr, f, args, is_label):
     is_matching = args.get("isMatching", False)
--- a/fiftyone/server/view.py
+++ b/fiftyone/server/view.py
@@ -626,12 +626,19 @@ def _make_keypoint_list_filter(args, view, path, field):
 
     if isinstance(field.field, (fof.FloatField, fof.IntField)):
         f = F(name)
-        mn, mx = args["range"]
-        expr = (f >= mn) & (f <= mx)
-        if args["exclude"]:
-            expr = ~expr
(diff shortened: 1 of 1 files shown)


fiftyone/server/view.py
```diff
@@ -626,12 +626,19 @@ def _make_keypoint_list_filter(args, view, path, field):
 
     if isinstance(field.field, (fof.FloatField, fof.IntField)):
         f = F(name)
-        mn, mx = args["range"]
-        expr = (f >= mn) & (f <= mx)
-        if args["exclude"]:
-            expr = ~expr
+        #adding support if filter will applied on categorical variable i.e. 'labels'
```
lanzhenw, line 629: Thanks for catching this. Is 'labels' field intField or stringField?

When the field.field is FloatField or IntField, it will be rendered with the `numericFilter`as a range slide rather than the categorical filter. We can add a separate function to handle StringField. 

I would suggest:
```
    if isinstance(field.field, (fof.FloatField, fof.IntField)):
        f = F(name)
        mn, mx = args["range"]
        expr = (f >= mn) & (f <= mx)
        if args["exclude"]:
            expr = ~expr

        return {"filter": expr}
    
    if isinstance(field.field, (fof.StringField)):
        f = F(name)
        if 'values' in args:
            val = args["values"]
            expr = f.is_in(val)
            if args["exclude"]:
                expr = ~expr
            return {"filter": expr}
```

akshitpriyesh: Yes, it was for string fields . I have updated accordingly.

akshitpriyesh:
added according to [lanzhenw](https://github.com/lanzhenw)

benjaminpkane (approved):
LGTM. Thanks @akshitpriyesh! 🥇 

It looks like boolean fields are missing too. I will have a follow up PR

lanzhenw (approved):
LGTM! Thanks @akshitpriyesh

Issue #2257, opened 2022-11-08
[BUG] Select visible option ignores active sample in grouped contexts


brimoor:
When working with grouped datasets in the App, the options in the selection toolbar such as `Select visible (current sample)` always apply to the sample in Looker2D, but should switch to the point cloud sample when Looker3D is active:

<img width="2253" alt="Screen Shot 2022-11-08 at 11 25 48 AM" src="https://user-images.githubusercontent.com/25985824/200620899-9526d87c-d23d-47b7-873a-d12bfde28b40.png">

brimoor:
https://github.com/voxel51/fiftyone/pull/2782

PR #2782, opened 2023-03-13
Fix group selection target

ritch:
See #2257 for bug details.

https://user-images.githubusercontent.com/462228/224837371-ab31c720-d5b5-4827-85bb-e094f1d5015f.mov

### DIFF
--- a/app/packages/state/src/routing/RouterContext.ts
+++ b/app/packages/state/src/routing/RouterContext.ts
@@ -263,6 +263,7 @@ async function fetchGraphQL(
   );
 
   if ("errors" in data && data.errors) {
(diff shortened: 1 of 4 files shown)


app/packages/state/src/routing/RouterContext.ts
```diff
@@ -263,6 +263,7 @@ async function fetchGraphQL(
   );
 
   if ("errors" in data && data.errors) {
+    console.error(data);
```
ritch, line 266: Added intentionally as the error I was seeing was not actually displayed in the logs.

sashankaryal (approved):
lgtm

PR #2783, opened 2023-03-14
Fix empty state loading

ritch:
Empty state was reloading and returning null even while loading. This caused several issues. 

 - empty state would not update/display when filters caused 0 results
 - when selecting samples in embeddings empty state would re-render in the loading state and cause issues with spaces (closing the embeddings tab)

### DIFF
--- a/app/packages/core/src/components/EmptySamples.tsx
+++ b/app/packages/core/src/components/EmptySamples.tsx
@@ -43,6 +43,9 @@ export default function EmptySamples() {
   const totalSamples = useRecoilValue(
     fos.count({ path: "", extended: true, modal: false })
   );
+
+  console.log({ totalSamples });
+
   const theme = useTheme();
   const setView = fos.useSetView();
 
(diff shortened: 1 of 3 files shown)


app/packages/core/src/components/EmptySamples.tsx
```diff
@@ -109,4 +109,4 @@ function EmptySamples() {
   );
 }
 
-export default React.memo(EmptySamples);
+// export default React.memo(EmptySamples);
```
imanjra, line 112: Can this be removed?

imanjra (approved):

manivoxel51 (approved):
Thank you for adding suspense to our new empty state 🍨

brimoor (dismissed):
@ritch I think this reveals(?) an existing issue where the state is temporarily set to empty before loading the new state (see below).

If it only reveals this, can you tweak so the empty state doesn't render? Or maybe it causes this issue, in which case would be good to fix it.

https://user-images.githubusercontent.com/25985824/224881333-a1304b5f-0937-4402-91a2-7c5bf7dd0106.mov

app/packages/core/src/components/EmptySamples.tsx
ritch: Yes

brimoor (approved):
LGTM

PR #2784, opened 2023-03-14
Use dedicated media field for orthographic projection filepath

sashankaryal:

Use dedicated media field for orthographic projection filepath

Locally, with quickstart-groups dataset, both with and without projections computed.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/server/metadata.py
+++ b/fiftyone/server/metadata.py
@@ -11,6 +11,8 @@ import shutil
 import struct
 import typing as t
 
+from functools import reduce
+
 import asyncio
 import aiofiles
 import strawberry as gql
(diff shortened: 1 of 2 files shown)


brimoor (approved):
LGTM

fiftyone/server/metadata.py
```diff
@@ -87,10 +85,6 @@ async def get_metadata(
                     aspect_ratio=width / height,
                     frame_rate=frame_rate,
                 )
-        elif has_orthographic_projection_metadata:
```
brimoor, line 90: I think this is still needed: when the projection images are being rendered in the grid, you need to use them to compute the aspect ratio so that the grid tiling algorithm works properly.

As currently implemented, the tiling is not correct in the grid.

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "quickstart-groups",
    dataset_name="test-quickstart-groups",
    max_samples=30,
)
dataset.group_slice = "pcd"

fou3d.compute_orthographic_projection_images(dataset, (-1, 512), "/tmp/projections")

```

brimoor (changes requested):

fiftyone/server/metadata.py
sashankaryal: Yeah, I got up this morning with the same thought in my head 😅  pushed a fix in the new commit

benjaminpkane (approved):
LGTM

brimoor (approved):
LGTM

PR #2785, opened 2023-03-14
Fix group aggregations

kaixi-wang:

- add `_id` field to match `AppSample` and fix group count query
- fix `mixed` param getting set for groups 

https://user-images.githubusercontent.com/30936736/224942590-cd233f4b-7971-4c14-a5d3-331c43ba7095.mov

Selecting labels in the pcd in the modal view:

https://user-images.githubusercontent.com/30936736/224945800-7655e9b0-625b-41f5-a248-b15a810a679b.mov

(Details)

        notes for FiftyOne users.

-   [ ] Other


kaixi-wang:
@sashankaryal 
The label counts are the same across the groups I checked, is this expected? The queries now look correct, so if it's wrong then I think that means the paths included in the mongo aggregation are wrong...

image

image

app/packages/looker/src/state.ts
```diff
@@ -23,6 +23,7 @@ export interface Coloring {
 }
 
 export type OrthogrpahicProjectionMetadata = {
+  _id: string;
```
brimoor, line 26: [OrthographicProjectionMetadata](https://github.com/voxel51/fiftyone/blob/130cf6022260809a02c269561cd2f154709b7d54/fiftyone/utils/utils3d.py#L413-L430) doesn't have an `id` attribute, so we should not add this

brimoor (changes requested):

app/packages/looker/src/state.ts
kaixi-wang: we have to have it because it gives the specific id which is expected by downstream constants and is also what enables filtering and such to work

brimoor: Is this done for the regular `Metadata` classes too? They don't have `id` attributes either

kaixi-wang: It's not called Metadata, but this is it's equivalent for non-ortho proj samples: https://github.com/voxel51/fiftyone/blob/develop/app/packages/state/src/recoil/atoms.ts#L9

brimoor: 👍

benjaminpkane (approved):
There are some subtleties that we can clean up in the future. I pushed a quick tweak to ensure both `slice` and `group` modes behave correctly. Notice the counts in the screen recording

https://user-images.githubusercontent.com/19821840/225106991-8a922e97-e957-4dde-b345-c5f847a52b28.mov

sashankaryal (approved):
LGTM!

PR #2786, opened 2023-03-14
Fix input interference in modal

benjaminpkane:
Prevents the document keydown listener in the modal from interfering with input interactions


sashankaryal (approved):
looks good!!

PR #2789, opened 2023-03-14
respect FIFTYONE_DEFAULT_APP_PORT env var

sashankaryal:
respect FIFTYONE_DEFAULT_APP_PORT env var

### DIFF
--- a/app/packages/core/src/components/Setup.tsx
+++ b/app/packages/core/src/components/Setup.tsx
@@ -47,7 +47,7 @@ const Code = styled.pre`
 
 const port = (() => {
   if (isElectron()) {
-    return parseInt(process.env.FIFTYONE_SERVER_PORT) || 5151;
+    return parseInt(process.env.FIFTYONE_DEFAULT_APP_PORT) || 5151;
   }
 
   if (typeof window !== "undefined" && window.location.port !== undefined) {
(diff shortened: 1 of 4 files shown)


app/packages/core/src/components/Setup.tsx
```diff
@@ -47,7 +47,7 @@ const Code = styled.pre`
 
 const port = (() => {
   if (isElectron()) {
-    return parseInt(process.env.FIFTYONE_SERVER_PORT) || 5151;
```
benjaminpkane, line 50: This is used by the Desktop App and should be kept (with priority)
* https://github.com/voxel51/fiftyone/blob/10cf8844dee0a9cea716c745916d80ed901074ec/fiftyone/core/service.py#L415

app/packages/utilities/src/fetch.ts
```diff
@@ -178,7 +178,7 @@ export const getAPI = () => {
   }
   return isElectron()
     ? `http://${process.env.FIFTYONE_SERVER_ADDRESS || "localhost"}:${
-        process.env.FIFTYONE_SERVER_PORT || 5151
+        process.env.FIFTYONE_DEFAULT_APP_PORT || 5151
```
benjaminpkane, line 181: Same comment

app/packages/core/src/components/Setup.tsx
sashankaryal: Don't know why we're using a different env var for electron apps, but I bet there's a reason. Addressed in next commit.

benjaminpkane (approved):
LGTM

PR #2790, opened 2023-03-14
Additional error handling for plot loading

ritch:
Handle both then load_brain_result() returns None or throws.


app/packages/embeddings/src/useViewChangeEffect.tsx
```diff
@@ -37,7 +37,10 @@ export function useViewChangeEffect() {
     setOverrideStage(null);
     setLoadingPlot(true);
     fetchPlot({ datasetName, brainKey, view, labelField })
-      .catch((err) => setLoadingPlotError(err))
+      .catch((err) => {
+        setLoadingPlotError(err);
+        // setBrainKey(null);
```
ritch, line 42: This causes the empty state to load, which would remove the error.

brimoor (approved):
LGTM

PR #2801, opened 2023-03-16
fix delete saved view crash

manivoxel51:

deleting saved views crashes since the property does not exist in state
https://voxel51.atlassian.net/browse/TEAMS-1029

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/server/mutation.py
+++ b/fiftyone/server/mutation.py
@@ -322,7 +322,11 @@ class Mutation:
         # If the current view is deleted, set the view state to the full
         # dataset view
         state = get_state()
-        if view_name and state.view and state.view.name == view_name:
+        if (
+            view_name
+            and state.view is not None
+            and state.view.name == view_name
+        ):
             state.view = dataset.view()
             state.view_name = None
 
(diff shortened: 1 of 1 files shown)


fiftyone/server/mutation.py
```diff
@@ -322,7 +322,7 @@ async def delete_saved_view(
         # If the current view is deleted, set the view state to the full
         # dataset view
         state = get_state()
-        if view_name and state.view_name == view_name:
+        if view_name and state.view and state.view.name == view_name:
```
brimoor, line 325: `state.view` is expensive because `DatasetView.__bool__` computes `len(view) > 0`. I think you should explicitly use `state.view is not None` instead.

brimoor (changes requested):

fiftyone/server/mutation.py
manivoxel51: good to know 👍

brimoor: 👍 I defer to an App code owner for approval from here

ritch (approved):

PR #2817, opened 2023-03-20
Fix values search

benjaminpkane:
Fixes values search for group datasets and a pipeline issue in `CountValues` when both `_selected` and `_search` are provided


sashankaryal (approved):
looks good

PR #2818, opened 2023-03-20
Point-E Tutorial

jacobmarks:
Tutorial showcasing our new 3D visualizer's capabilities by building a 3D self-driving dataset with Point-E

        notes for FiftyOne users.

Build a 3D self-driving point cloud dataset from scratch with OpenAI's Point-E and FiftyOne!

-   [ ] Other


ehofesmann (approved):
Awesome tutorial @jacobmarks ! This is a really unique workflow that we don't really have anything like yet.

This looks good to me, just two points that you can choose to act on or not:
* Would it be possible to use a different example than the rubber duck? It's a bit hard to make out what the object is supposed to be and it would be good to show Point-E in a positive light so people are incentivized to read to the end
* Not necessary for this tutorial if it doesn't exist, but it seems like the orthographic projections can be a bit difficult to see in the grid view of the App, being able to specify the point size for `compute_orthographic_projection_images()` could be useful

sashankaryal:
Looking good! 🥳 Loved following through the tutorial, too.

sashankaryal (approved):

PR #2820, opened 2023-03-20
Tagging fixes

benjaminpkane:
Fixes tagging and tagging statistics for groups, notably
 * Tagging with respect to the currently pinned sample in the modal
 * Tagging selected labels regardless of currently pinned slice in the modal


sashankaryal (approved):
skimmed through, looks good to me

Issue #2192, opened 2022-10-24
[BUG] Datasets cannot have 'pk' fields


brimoor:
Datasets cannot currently have a `pk` field:

```py
import fiftyone as fo

sample = fo.Sample(filepath="image.jpg", pk=123)

dataset = fo.Dataset()
dataset.add_sample(sample)

dataset.set_values("foo", ["bar"])
# TypeError: id must be an instance of (bytes, str, ObjectId), not <class 'int'>
```

I believe `pk` should be a reserved keyboard, IE these should be updated to raise errors:

```py
sample = fo.Sample(filepath="image.jpg")
sample["pk"]  # ValueError: 'pk' is a reserved keyword

sample = fo.Sample(filepath="image.jpg", pk=123)    # ValueError: 'pk' is a reserved keyword
```

There may be other such keywords that need similar handling, btw.

I haven't investigated the source of the error. It might be possible to instead patch things so that `pk` is allowable.

nebulae:
some investigation - mongoengine considers 'pk' an alias for an id field, so it seems to want to treat is with the same constraints as it would any ObjectId: 

https://github.com/MongoEngine/mongoengine/blob/master/mongoengine/base/document.py#L1110

PR #2822, opened 2023-03-21
[HOLD] check for reserved keywords on sample, raise ValueError on reserved keyword

nebulae:

Checking supplied fields against a list of reserved keywords, currently the list contains a single keyword: `'pk'`, which is an alias for the `_id` field. We may want to add more of these reserved keywords in the future, so created this as a list.

Also, support for checking all the supplied fields on sample instantiation, and returning a comprehensive error message if multiple reserved keywords are attempting to be used (`'pk' is a reserved keyword` vs `'pk', 'foo' are reserved keywords`)

added a test in the existing `unittests/sample.py`

        notes for FiftyOne users.

original issue here: https://github.com/voxel51/fiftyone/issues/2192

-   [ ] Other

### DIFF
--- a/fiftyone/core/sample.py
+++ b/fiftyone/core/sample.py
@@ -57,6 +57,7 @@ class _SampleMixin(object):
         if self.media_type == fomm.VIDEO and fofu.is_frame_number(field_name):
             return self.frames[field_name]
 
+        self._check_reserved_keywords([field_name])
         return super().__getitem__(field_name)
 
     def __setitem__(self, field_name, value):
@@ -64,6 +65,7 @@ class _SampleMixin(object):
             self.frames[field_name] = value
             return
 
+        self._check_reserved_keywords([field_name])
         self._secure_media(field_name, value)
         super().__setitem__(field_name, value)
 
@@ -474,6 +476,24 @@ class _SampleMixin(object):
 
         return fields, frame_fields, omit_fields, omit_frame_fields
 
+    def _check_reserved_keywords(self, fields):
+        """Validates the provided fields.
+
+        Raises a :class:`ValueError` if any of the fields are considered
+        reserved keywords.
+
+        Currently only checks for 'pk', which is reserved for the
+        primary key as an alias for '_id'
+        """
+        reserved_fields = {"pk"}.intersection(fields)
+        if reserved_fields:
+            reserved = ", ".join([f"'{x}'" for x in reserved_fields])
+            raise ValueError(
+                f"{reserved} is a reserved keyword"
+                if len(reserved_fields) == 1
+                else f"{reserved} are reserved keywords"
+            )
+
 
 class Sample(_SampleMixin, Document, metaclass=SampleSingleton):
     """A sample in a :class:`fiftyone.core.dataset.Dataset`.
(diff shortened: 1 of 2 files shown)


fiftyone/core/sample.py
```diff
@@ -57,13 +57,15 @@ def __getitem__(self, field_name):
         if self.media_type == fomm.VIDEO and fofu.is_frame_number(field_name):
             return self.frames[field_name]
 
+        self._check_reserved_keywords([field_name])
```
brimoor, line 60: I don't think we need validation on `__getitem__`. If the data is there, go ahead and let the user access it.

fiftyone/core/sample.py
```diff
@@ -57,13 +57,15 @@ def __getitem__(self, field_name):
...
         return super().__getitem__(field_name)
 
     def __setitem__(self, field_name, value):
         if self.media_type == fomm.VIDEO and fofu.is_frame_number(field_name):
             self.frames[field_name] = value
             return
 
+        self._check_reserved_keywords([field_name])
```
brimoor, line 68: The `Sample` class inherits from `fiftyone.core.document._Document`:
https://github.com/voxel51/fiftyone/blob/75acb79de8362efaf02d79fe2697e5338106606c/fiftyone/core/document.py#L19

which maintains a `_doc` property that uses the `fiftyone.core.odm.document.MongoEngineBaseDocument` interface to actually store the field values:
https://github.com/voxel51/fiftyone/blob/75acb79de8362efaf02d79fe2697e5338106606c/fiftyone/core/odm/document.py#L319

I believe that's the source of the `pk` issue, so it would be the appropriate place to do the reserved keyword check, as this same issue would apply to all users of this class, including other public-facing classes like `fiftyone.core.frame.Frame` that aren't currently covered by this PR.

tests/unittests/sample_tests.py
```diff
@@ -858,6 +858,24 @@ def test_vector_array_fields(self):
...
+
+        sample1 = fo.Sample(
+            filepath="img.png",
+        )
+
+        with self.assertRaises(ValueError):
+            sample1.pk = 321
+
+        with self.assertRaises(ValueError):
+            sample1["pk"] = 321
```
brimoor, line 874: It's been awhile since I've thought about this issue, but back when I created it, I thought to myself that there's likely a way to allow `sample["xxx"]` to work for *any* possible field name without restrictions.

`sample.xxx` will never be possible in many situations, even if we prevent `sample.pk` in particular. For example, a user could never use `Sample.__getattr__` to access a field they added that happens to match any of the public method names on the `Sample` class, either.

brimoor (changes requested):
Thanks for the PR! Left a couple comments for your consideration. Happy to discuss offline too if needed.

Full disclosure: in my opinion this issue isn't that important. However, I'm glad you looked into it because it is an instructive way to get familiar with how data storage/presentation works with the `Document` interface.

Based on my comments below, I think there's two possible paths for this PR:
- Expand its scope to provide a more general treatment of reserved keyword names
- Decide that further work is not worth it at this time

tests/unittests/sample_tests.py
brimoor: Internally, the mongoengine `Document` class upon which the entire FO data model is built maintains the actual values in a `_data` dict. `__getattr__` always has to compete with other public symbols of the `Document` class, but `__getitem__` can directly access the values from the `_data` dict and therefore is always a "safe" way to access arbitrarily-named fields.

benjaminpkane:
Closing for now

PR #2829, opened 2023-03-22
Ignore spaces when sorting by similarity

benjaminpkane:
Prevents spaces from resetting when sorting by similarity


brimoor (approved):
Can't comment on implementation, but tested and works 👍

lanzhenw (approved):
LGTM

Issue #2836, opened 2023-03-24
[BUG] stats() is inaccurate on a DatasetView with a SelectFieldsStage


swheaton:
### Commands to reproduce

When you exclude fields the document size does not contain those fields but for some reason when you select fields the document size includes all fields.

I have verified that simply adding a "$project" stage with the given fields before the "$bsonSize" stage generated in `SampleCollection._get_samples_bytes()` gives the expected result, but that should be exactly what the `SelectFieldsStage` is doing. So it needs to be looked into more.

As a workaround if one needed it, one could exclude all but the wanted (and default) fields and it seemingly works.
```python
import fiftyone.zoo as foz
dataset=foz.load_zoo_dataset('quickstart')

# Fails, sizes are equal
select_view = dataset.select_fields()
assert select_view.stats()["total_bytes"] < dataset.stats()["total_bytes"]

# Works as expected
exclude_view = dataset.exclude_fields(["ground_truth", "predictions"])
assert exclude_view.stats()["total_bytes"] < dataset.stats()["total_bytes"]
```

### What areas of FiftyOne does this bug affect?

-   [ ] `App`: FiftyOne application issue
-   [x] `Core`: Core Python library issue
-   [ ] `Server`: FiftyOne server issue

### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04): Mac
-   **Python version** (`python --version`): 3.10.9
-   **FiftyOne version** (`fiftyone --version`): 0.20.0
-   **FiftyOne installed from** (pip or source): pip

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the FiftyOne codebase?

-   [x] Yes. I can contribute a fix for this bug independently
-   [ ] Yes. I would be willing to contribute a fix for this bug with guidance from the FiftyOne community
-   [ ] No. I cannot contribute a bug fix at this time

brimoor:
I can't reproduce the bug you're reporting here, at least on `develop`. This succeeds:

```py
import fiftyone.zoo as foz

dataset=foz.load_zoo_dataset('quickstart')
select_view = dataset.select_fields()
assert select_view.stats()["total_bytes"] < dataset.stats()["total_bytes"]
```

brimoor:
The `mongo()` stage can be leveraged to avoid inclusion of default fields that `select_fields()` performs:

```py
def force_select_fields(sample_collection, fields):
[...]

PR #2840, opened 2023-03-24
workaround mongodb bug in _get_samples_bytes()

swheaton:
Closes https://github.com/voxel51/fiftyone/issues/2836 where `SampleCollection.stats()` was being reported as the full document size total even if a SelectField stage was present.

This bug was due to a change in behavior of $bsonSize within a $group stage after MongoDB ~ 5.2. It's believed because of the usage of the Slot Based Execution query engine for $group stage in 5.2.
A [MongoDB bug ticket has been filed](https://jira.mongodb.org/browse/SERVER-75267) but that's about all we can do since there appears to be no way to turn off the SDE engine selectively.

The workaround is to grab document sizes first then sum them, in 2 stages instead of 1. This may be slightly more inefficient but likely not by much.

Added unit test exposing bug in MongoDB >= 5.2. Passes now.

```python
# connect to MongoDB 6.0.X ...
import fiftyone.zoo as foz
dataset=foz.load_zoo_dataset('quickstart')

def get_stats(collection):
    return collection.stats()["total_bytes"]

select_view = dataset.select_fields()
select_view2 = dataset.select_fields("uniqueness")
select_view3 = dataset.select_fields(["uniqueness", "ground_truth"])
assert get_stats(select_view) < get_stats(select_view2) < get_stats(select_view3) < get_stats(dataset)
```

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -594,13 +594,23 @@ class SampleCollection(object):
 
     def _get_samples_bytes(self):
         """Computes the total size of the sample documents in the collection."""
+
+        # Impl note: We *could* do group directly without projecting $bsonSize
+        #   first, however we found in Mongo 5.2 a new execution engine
+        #   (Slot Based Execution) optimizes this away and incorrectly sums
+        #   the full document size, thus ignoring selected fields. A ticket
+        #   has been opened to resolve this bug.
+        #   https://jira.mongodb.org/browse/SERVER-75267
+        #   But until then we must do this in 2 stages (project-size, group-sum)
+        #   rather than 1 (group-sum-size)
         pipeline = [
+            {"$project": {"size": {"$bsonSize": "$$ROOT"}}},
             {
                 "$group": {
                     "_id": None,
-                    "size_bytes": {"$sum": {"$bsonSize": "$$ROOT"}},
+                    "size_bytes": {"$sum": "$size"},
                 }
-            }
+            },
         ]
 
         results = self._aggregate(pipeline=pipeline)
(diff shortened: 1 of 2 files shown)


tests/unittests/view_tests.py
```diff
@@ -2098,7 +2104,21 @@ def test_exclude_fields(self):
             with self.assertRaises(AttributeError):
                 sample.exclude_fields_field1
 
-            self.assertIsNone(sample.exclude_fields_field2)
+            self.assertEqual(sample.exclude_fields_field2, 1)
```
brimoor, line 2107: Suggestion: move the `stats()`-specific tests into separately-named test methods? 

This bug is strictly the fault of interplay between `$bsonSize + $project`, which could occur in more than just the `select_fields()` and `exclude_fields()` code paths. This is very minor, of course.

brimoor (approved):
LGTM, thanks for getting to the bottom of this! 🥇 

Left one small suggestion for your consideration

fiftyone/core/collections.py
```diff
@@ -594,13 +594,23 @@ def stats(self, include_media=False, compressed=False):
 
     def _get_samples_bytes(self):
         """Computes the total size of the sample documents in the collection."""
+
+        # Impl note: We *could* do group directly without projecting $bsonSize
+        #   first, however we found in Mongo 5.2 a new execution engine
+        #   (Slot Based Execution) optimizes this away and incorrectly sums
+        #   the full document size, thus ignoring selected fields. A ticket
+        #   has been opened to resolve this bug.
+        #   https://jira.mongodb.org/browse/SERVER-75267
```
kaixi-wang, line 603: Did some reading on slot based execution, and it's possible this is the expected behavior.

Previous engine outputted documents at each stage, so $$ROOT would refer to the document of the projected fields in that stage, but with slot based, the $$ROOT is now the original document (no intermediate document created)

> Hey! Query Execution engineer here, my team and I are working on SBE. Currently, query execution engine works in terms of documents. For example, in the aggregation language, every stage produces a document to be consumed by the next stage. SBE goes one step lower in abstraction layers and works in terms of values. Aggregation stages executed through SBE pass values to each other directly, without extra document materialization. This allows us to avoid many unnecessary computations and execute queries faster. As of now, we only use SBE for certain queries with $group and $lookup, but the intent is to use it across the board in the future.
https://www.reddit.com/r/mongodb/comments/v7uaso/what_does_the_slotbased_query_execution_engine_do/

swheaton: An execution engine should not change functionality unless documented (it is not), so either way it is a bug IMO.
They may come back (if at all) and say yeah not gonna change that. In which case their documented way of getting total document size is just broken

tests/unittests/view_tests.py
swheaton: Sure thing

swheaton: Done

PR #2855, opened 2023-03-28
Fixing a bug in dynamic schema expansion

brimoor:
The unit test added in this PR fails on `develop`, but now passes as expected.


manivoxel51 (approved):
nice! 🍨

PR #2861, opened 2023-03-29
Bugfix: a user can save a float/datetime object in field.info and see that in the tootlip

lanzhenw:

- update json dump to be able to save a date object in field
- display datetime and float type in field.info table in the tooltip
To test this change, you can run the following SDK code: 
```
In [1]: import fiftyone as fo
    ...: from datetime import datetime
    ...:
    ...: dataset = foz.load_zoo_dataset("quickstart")
    ...:
    ...: field = dataset.get_field("ground_truth")
    ...: field.info = {"created_at": datetime.utcnow(), "sample": float(InF)}
    ...: field.save()
    ...: dataset.save()
    ...: session = fo.launch_app(dataset)
```

Screenshot 2023-03-28 at 7 17 30 PM
Screenshot 2023-03-28 at 7 14 56 PM
Screenshot 2023-03-28 at 7 15 15 PM

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/FieldLabelAndInfo/index.tsx
+++ b/app/packages/core/src/components/FieldLabelAndInfo/index.tsx
@@ -73,8 +73,6 @@ function useFieldInfo(field, nested, { expandedPath, color }) {
     setOpen(selectedField === instanceId);
   }, [selectedField]);
 
-  console.info("field", field);
-
   return {
     open,
     hoverTarget,
@@ -509,8 +507,9 @@ function keyValueIsRenderable([key, value]) {
     case "number":
     case "boolean":
       return true;
+    case "object":
+      if (value._cls === "DateTime") return true;
     default:
-      console.info(typeof value);
       return false;
   }
 }
@@ -518,6 +517,12 @@ function toRenderValue([key, value]): [string, string] {
   switch (typeof value) {
     case "boolean":
       return [key, value ? "True" : "False"];
+    case "object":
+      if (value._cls === "DateTime" && value.datetime) {
+        return [key, new Date(value.datetime).toUTCString()];
+      } else {
+        return [key, ""];
+      }
     default:
       return [key, value];
   }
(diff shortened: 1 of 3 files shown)


app/packages/core/src/components/FieldLabelAndInfo/index.tsx
```diff
@@ -500,12 +501,13 @@ function FieldInfoTable({ info, type, collapsed, subfield, description }) {
...
 function keyValueIsRenderable([key, value]) {
   if (value === undefined || value === null) return true;
-
   switch (typeof value) {
     case "string":
     case "number":
     case "boolean":
       return true;
+    case "object":
+      if (value.$date || value.$numberDouble) return true;
```
manivoxel51, line 510: nit: `return value.$date || value.$numberDouble;`

app/packages/core/src/components/FieldLabelAndInfo/index.tsx
```diff
@@ -514,6 +516,14 @@ function toRenderValue([key, value]): [string, string] {
   switch (typeof value) {
     case "boolean":
       return [key, value ? "True" : "False"];
+    case "object":
+      if (value.$date) {
+        return [key, new Date(value.$date).toUTCString()];
```
manivoxel51, line 521: nice! I'd think local time makes sense for the user looking at it in the UI. cc @brimoor 
```.toLocaleString()```

**Review by manivoxel51 — APPROVED** [review]
Nice work @lanzhenw! 🍨 

I tested locally and it is showing the date field. Left minor comments

File: app/packages/core/src/components/FieldLabelAndInfo/index.tsx (hunk 2 above)
Comment (brimoor) [reply app/packages/core/src/components/FieldLabelAndInfo/index.tsx:521]: When rendering `DateTimeField`s in the sidebar, the App respects the user's timezone setting: 
https://docs.voxel51.com/user_guide/config.html#configuring-a-timezone

It would be nice to respect that in the field tooltip too.

**Review by manivoxel51 — COMMENTED** [review]
Thanks @benjaminpkane for adding the timezone! The actual time is not getting rendered although date is correct. Maybe It is useful to see the local time the field was created it was created 🤔 
Screen Shot 2023-03-30 at 11 09 08 AM
```
from datetime import datetime
dataset = ds
field.info = {"created_at": datetime.utcnow(), "sample": float(0)}
field.save()
```

PR #2863, opened 2023-03-29
Adding support for dynamically inferring fields on embedded lists and documents

brimoor:
The `get_dynamic_field_schema()` and `add_dynamic_sample_fields()` methods did not previously handle either of these cases:
- fields that contain lists of embedded documents 
- dynamic embedded documents (eg `Label` fields)

Now it does!

```py
import fiftyone as fo

sample = fo.Sample(
    filepath="/Users/Brian/Desktop/test.png",
    tasks=[
        fo.DynamicEmbeddedDocument(
            annotator="eric",
            labels=fo.Classifications(
                classifications=[
                    fo.Classification(label="cat", mood="surly"),
                    fo.Classification(label="dog"),
                ]
            ),
        ),
        fo.DynamicEmbeddedDocument(
            annotator="mani",
            labels=fo.Classifications(
                classifications=[
                    fo.Classification(label="rabbit"),
                    fo.Classification(label="squirrel", age=51),
                ]
            ),
        ),
    ]
)

dataset = fo.Dataset()
dataset.add_sample(sample)

fo.pprint(dataset.get_field_schema(flat=True))
fo.pprint(dataset.get_dynamic_field_schema())

fo.pprint(dataset.get_field_schema(flat=True))
fo.pprint(dataset.get_dynamic_field_schema())
```

### DIFF
--- a/fiftyone/core/aggregations.py
+++ b/fiftyone/core/aggregations.py
@@ -1841,12 +1841,13 @@ class Schema(Aggregation):
         field_or_expr,
         expr=None,
         dynamic_only=False,
+        _doc_type=None,
         _include_private=False,
     ):
         super().__init__(field_or_expr, expr=expr)
         self._dynamic_only = dynamic_only
         self._include_private = _include_private
-        self._doc_type = None
+        self._doc_type = _doc_type
 
     def _kwargs(self):
         return [
@@ -1919,16 +1920,17 @@ class Schema(Aggregation):
 
     def to_mongo(self, sample_collection, context=None):
         field_name = self._field_name
-        doc_type = None
 
-        if self._expr is None:
-            field_type = _get_field_type(
-                sample_collection, field_name, unwind=True
-            )
-            if isinstance(field_type, fof.EmbeddedDocumentField):
-                doc_type = field_type
+        if self._doc_type is None:
+            doc_type = None
+            if self._expr is None:
+                field_type = _get_field_type(
+                    sample_collection, field_name, unwind=True
+                )
+                if isinstance(field_type, fof.EmbeddedDocumentField):
+                    doc_type = field_type
 
-        self._doc_type = doc_type
+            self._doc_type = doc_type
 
         path, pipeline, _, _, _ = _parse_field_and_expr(
             sample_collection,
--- a/fiftyone/core/aggregations.py
+++ b/fiftyone/core/aggregations.py
@@ -9,10 +9,12 @@ from collections import defaultdict, OrderedDict
 from copy import deepcopy
 from datetime import date, datetime
 import inspect
+import logging
 import reprlib
 import uuid
 
 import numpy as np
+from mongoengine.base import get_document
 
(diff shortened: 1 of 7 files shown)


fiftyone/core/collections.py
```diff
@@ -1198,6 +1198,9 @@ def _get_dynamic_field_schema(self, schema, prefix, fields=None):
         aggs = []
         paths = []
         for name, field in schema.items():
+            if isinstance(field, fof.ListField):
```
benjaminpkane, line 1201: In theory, this should be a `while` loop

fiftyone/core/aggregations.py
```diff
@@ -1890,6 +1896,16 @@ def parse_result(self, d):
 
             if name in doc_fields:
                 field = doc_fields[name]
+            elif "." in _type:
+                _type, _cls = _type.split(".", 1)
```
kaixi-wang, line 1900: do we do any validation on field names? What if a user tries to add a dynamic field with a `.` in the name?

brimoor: If a user tried to set a field like `foo.bar` without first declaring `foo`, they'd see an error like this:

```
ValueError: Cannot infer an appropriate type for non-existent sample field 'foo' while defining embedded field 'foo.bar'
```

manivoxel51 (commented):
testing this now

manivoxel51 (approved):
Thanks for adding this Brian 🍨 

Tested a top level list and works well!

not sure how much of a use case the following is - but noticed it is not handled in the PR. Its a list of dynamic documents that have another nested list in them. 
```
sample = ds.head()[0]
sample["tasks_2"] = [
  fo.DynamicEmbeddedDocument(age=2, level_1_list=. 
  [fo.DynamicEmbeddedDocument(level_2_list=["tag1"]), 
  fo.DynamicEmbeddedDocument(level_2_list=["tag2"])]),
fo.DynamicEmbeddedDocument(age=2, level_1_list=. 
  [fo.DynamicEmbeddedDocument(level_2_list=["tag11"]), 
  fo.DynamicEmbeddedDocument(level_2_list=["tag22"])]),
]
sample.save()
```
I see
```
    'tasks_2': <fiftyone.core.fields.ListField object at 0x7fa3d1dba940>,
    'tasks_2.level_1_list': <fiftyone.core.fields.ListField object at 0x7fa3d1dba700>,
    'tasks_2.age': <fiftyone.core.fields.IntField object at 0x7fa3e559e520>,
```
Should we be seeing one more field?
```
    'tasks_2.level_1_list.level_2_list': <fiftyone.core.fields.StringField object at 0x7fa3d1dba700>,
```

brimoor:
@manivoxel51 hmm yeah, the `Schema()` aggregation does not currently try to infer the type of values inside `ListField`. It just declares it as a `ListField` with no sub `field` declared. I'll take a quick look to see if it is possible to make this implementation recursive 🤔

PR #2872, opened 2023-04-03
add support for jpg segmentations / heatmaps

sashankaryal:

Only png images were supported for `mask_path` / `map_path` for on-disk segmentations / heatmaps, this PR adds support for jpg as well.

Locally

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -47,5 +47,8 @@
     "workspaces": [
         "packages/*"
     ],
-    "packageManager": "yarn@3.2.1"
+    "packageManager": "yarn@3.2.1",
+    "dependencies": {
+        "jpeg-js": "^0.4.4"
+    }
 }
--- a/app/packages/looker/src/worker/index.ts
+++ b/app/packages/looker/src/worker/index.ts
@@ -14,6 +14,7 @@ import {
   VALID_LABEL_TYPES,
 } from "@fiftyone/utilities";
 import { decode as decodePng } from "fast-png";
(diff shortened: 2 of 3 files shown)


manivoxel51 (approved):
Thank you! 🍨 

curious if there are other types we need to support for later

brimoor:
@sashankaryal I tested this using the `Heatmap` example from the docs:

```py
import os
import numpy as np
import fiftyone as fo
import fiftyone.zoo as foz

def random_kernel(metadata):
    h = metadata.height // 2
    w = metadata.width // 2
    sign = np.sign(np.random.randn())
    x, y = np.meshgrid(np.linspace(-1, 1, w), np.linspace(-1, 1, h))
    x0, y0 = np.random.random(2) - 0.5
    kernel = sign * np.exp(-np.sqrt((x - x0) ** 2 + (y - y0) ** 2))
    return fo.Heatmap(map=kernel, range=[-1, 1])

dataset.compute_metadata()

for sample in dataset:
    heatmap = random_kernel(sample.metadata)

    # Convert to on-disk
    map_path = os.path.join("/tmp/heatmaps", os.path.basename(sample.filepath))
    heatmap.export_map(map_path, update=True)

    sample["heatmap"] = heatmap
    sample.save()

```

and I see this:

Screen Shot 2023-04-03 at 11 45 40 PM

`Heatmap` images should always be single channel. Looks like the code is injecting three channels erroneously.

PR #2877, opened 2023-04-03
Fix client serialization initialization

benjaminpkane:
Post data in `fiftyone.core.session.client.Client.run` initialization  the same way `post_event` does to fix `datetime` serialization when initializing a`Session` with a non-`None` dataset value

```py
import fiftyone as fo
import fiftyone.zoo as foz
from datetime import datetime

field.info = {"created_at": datetime.utcnow(), "sample": float("inf")}
field.save()

session = fo.launch_app(dataset) # previously would fail
```

### DIFF
--- a/fiftyone/core/session/client.py
+++ b/fiftyone/core/session/client.py
@@ -69,17 +69,22 @@ class Client:
                         "Accept": "text/event-stream",
                         "Content-type": "application/json",
                     },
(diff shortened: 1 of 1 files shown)


manivoxel51 (approved):
thank you @benjaminpkane ! 🍨

PR #2878, opened 2023-04-04
Removes supports prompts warning from server

benjaminpkane:
Creating a brain config does not require the `kwarg`


brimoor (approved):
LGTM

PR #2886, opened 2023-04-06
[draft/wip] github importer

kaixi-wang:

Load external packages/modules from fiftyone
ex plugins:

```
import fiftyone.zoo.utils.github as gh_importer
sample_exporter = gh_importer.load('export',
                                   'https://github.com/voxel51/fiftyone-plugins/tree/operators/packages')
sample_exporter.register()
```

```
module =  gh_importer.load('dummy_dataset', 'kaixi-wang/dummy_dataset')
module.hello()
```

Unsure about what will run, just print `__code__`:
```
print(sample_exporter.__code__)

import fiftyone as fo
class ExportSamples(foo.DynamicOperator):
    def __init__(self):
        super().__init__(
            "export_samples",
            "Export Samples",
        )
        
    def resolve_input(self, ctx):
        show_estimate = False
        inputs = types.Object()
        view = types.View(label="Export samples")
        inputs.define_property("style", types.Enum(["direct_download", "cloud_export"]), default="direct_download")
        
        if (ctx.params.get("style", None) == "cloud_export"):
            inputs.define_property("cloud_storage_path", types.String())
        inputs.define_property("view", types.Enum(["current_filters"]), default="current_filters")
        inputs.define_property("data", types.Enum([
            "media_and_labels",
            "media_only",
            "labels_only",
            "filepaths_and_tags",
            "filepaths_only"
        ]))
        cur_data = ctx.params.get("data", None)
        if (cur_data == "media_and_labels" or cur_data == "labels_only"):
            inputs.define_property("label_format", types.Enum([
                "COCO",
                "CSV Dataset",
                "CVAT image",
                "FiftyOne Dataset",
                "FiftyOne image detections",
                "Image segmentation directory",
                "KITTI",
                "TF object detections",
                "VOC",
                "YOLO v4",
                "YOLO v5"
            ]))
            cur_label_format = ctx.params.get("label_format", None)
            needs_field = cur_label_format != "FiftyOne Dataset"
            if (needs_field):
                single_field = cur_label_format != "CSV Dataset"
                if (single_field):
                    inputs.define_property("field", types.Enum(get_label_fields(ctx.dataset)))
                else:
                    # TODO: add support for multiple selections
                    inputs.define_property("fields", types.Enum(ctx.dataset.get_field_schema().keys()))
        if show_estimate:
            inputs.define_property("estimate", types.String(), default="Estimated export size: TODO MB")
        return types.Property(inputs, view=view)
        
op = None
def register():
    op = ExportSamples()
    foo.register_operator(op)
def unregister():
    foo.unregister_operator(op)
def get_label_fields(dataset):
    return list(dataset.get_field_schema(embedded_doc_type=fo.Label).keys())

```

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/zoo/utils/github.py
+++ b/fiftyone/zoo/utils/github.py
@@ -1,0 +1,397 @@
+import io
+import json
+import logging
+import os
+import re
+import sys
+import tarfile
+import types
+from typing import Optional
+import zipfile
+from contextlib import contextmanager
+import eta.core.web as etaw
+
+import requests
+
+__GH_CONTENT_URL = {
+    "url": "https://{domain}/{user}/{repo}/{branch}{path}",
+    "domain": "raw.githubusercontent.com",
+}
+# https://gist.githubusercontent.com/kaixi-wang/ed4eb4f4dd29e944a1a5de762a9511a0/raw/fd04658f92b124798dabaa99d03ea434a5849113/coco-2017.py
+__GIST_CONTENT_URL = {
+    "url": "https://{domain}/{user}/{gist_id}/raw/{file}/",
+    "domain": "gist.githubusercontent.com",
+}
+# ====================== Configuration ======================
+
+__LOCAL_HOME_DIR = os.path.expanduser("~")
+# ====================== Logging ======================
+
+log_level = logging.DEBUG
+log_format = "%(message)s"
+logger = logging.getLogger(__name__)
+logger.setLevel(log_level)
+log_handler = logging.StreamHandler()
+log_handler.setLevel(log_level)
+log_formatter = logging.Formatter(log_format)
+log_handler.setFormatter(log_formatter)
+logger.addHandler(log_handler)
+
+# ====================== HTTP abstraction ======================
+
+
+def http(url, method="get", headers=None):
+    """Wraps HTTP/S calls in one place
+    Args:
+        url (str):
+        headers (dict):
+        method (str):
+    Returns:
+        dict: A dict containing 'code', 'headers', 'body' of HTTP response
+    """
+    if not headers:
+        headers = {"User-Agent": "fiftyone"}
+    try:
+        req = getattr(requests, method.lower())
+        resp = req(url, headers=headers)
+        resp.raise_for_status()
+        if resp.status_code == 200:
+            return {
+                "code": resp.status_code,
+                "headers": resp.headers,
+                "body": resp.text,
+            }
+    except requests.exceptions.HTTPError as he:
+        return {"error": he, "code": resp.status_code}
+    except Exception as e:
+        raise ValueError(f"Uncaught Error: {e}")
+
+
+def _create_paths(module_name, file_types=["py"]):
+    """Returns possible paths where a module/package could be located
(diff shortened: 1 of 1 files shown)


fiftyone/zoo/datasets/__init__.py
```diff
@@ -200,6 +309,12 @@ def load_zoo_dataset(
     :class:`fiftyone.core.dataset.Dataset` will be returned.
 
     Args:
+        name: the name of an existing zoo dataset or one to use for a
```
brimoor, line 312: I don't think we need both `name` and `gh_repo` for the case of loading a zoo dataset stored in a GH repo. My preference would be that the `name` argument should be extended to allow a github URL.

The existing `dataset_name` argument can be used if the user wants to control the name of the `Dataset` instance that is returned.

And I propose that, for github repos, `<user>/<name>` be used as the unique identifier for the zoo dataset, meaning that the dataset would be downloaded to `dataset_zoo_dir/<user>/<name>`.

fiftyone/zoo/datasets/__init__.py
```diff
@@ -176,8 +182,111 @@ def download_zoo_dataset(
...
+    # zoo_dataset, dataset_dir = _parse_dataset_details(
+    #     dataset_name, dataset_dir
+    # )
+    # split = splits[0] if len(splits) == 1 else None
+    if not name:
+        name = dataset_name
+    # dataset_module = httpimport.load(name,
+    #                               repo_or_url)
+    print("user", user, "dataset_name", dataset_name)
+    with httpimport.github_repo(user, dataset_name, ref="main"):
```
brimoor, line 273: I think this should be implemented more simply:
1. Download the github repo to disk 
2. Import the module that contains the `download_and_prepare()` function, then execute it

Prior art
- https://pytorch.org/docs/stable/hub.html#loading-models-from-hub
- https://pytorch.org/docs/stable/_modules/torch/hub.html#load

Our spec proposes the standard of having a `.fiftyone` file in the repository that, in turn, contains any metadata about the zoo dataset and could contain the name of the entrypoint`.py` file that contains the `download_and_prepare.py` function.

brimoor: It's important that the github repository be explicitly downloaded in step 1, because that's what plugins expect: the source code living in some directory on disk.

brimoor (changes requested):

fiftyone/zoo/datasets/__init__.py
```diff
@@ -176,8 +182,111 @@ def download_zoo_dataset(
     )
 
 
+def load_dataset_from_gh(
+    repo_or_url,
+    name: Optional[str] = None,
```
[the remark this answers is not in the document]
kaixi-wang: whoops this file shouldn't be here

fiftyone/zoo/datasets/__init__.py
kaixi-wang: this file was an old file I just referenced to copy some of the url parsing. please disregard

PR #2898, opened 2023-04-13
filter fields by description or info 

nebulae:

add meta_filter to select_fields

added additional unit tests, tested manually.

        notes for FiftyOne users.

an optional parameter `meta_filter` added to `dataset.select_fields` which allows for a string or dict to be passed along, which will be used to filter to the fields that have a match to this filter in their description or info. 

to select only fields that have a description that contains the string "my description", you can use `meta_filter="my description"`. To select only fields that have a specific key in their info, you can use `meta_filter=dict(key="value")`
 
-   [ ] Other


fiftyone/core/collections.py
```diff
@@ -5430,7 +5430,9 @@ def select_by(self, field, values, ordered=False):
         )
 
     @view_stage
-    def select_fields(self, field_names=None, _allow_missing=False):
+    def select_fields(
+        self, field_names=None, _allow_missing=False, meta_filter=None
```
brimoor, line 5434: nit: put `meta_filter` before `_allow_missing` so that all public kwargs come before private ones. 

(by convention, we're okay with changing the order of kwargs; we assume users use `key=value` for all optional args)

fiftyone/core/stages.py
```diff
@@ -5205,26 +5220,122 @@ class SelectFields(ViewStage):
             May contain ``embedded.field.name`` as well
     """
 
-    def __init__(self, field_names=None, _allow_missing=False):
+    def __init__(
+        self, field_names=None, _allow_missing=False, _meta_filter=None
```
brimoor, line 5224: By convention, arg names for `ViewStage` constructor should match the corresponding `@view_stage` method on `SampleCollection`.

So here, go ahead and use `meta_filter`. And copy the docstring changes over here 😄

fiftyone/core/stages.py
```diff
@@ -5319,6 +5430,7 @@ def _kwargs(self):
         return [
             ["field_names", self._field_names],
             ["_allow_missing", self._allow_missing],
+            ["_meta_filter", self._meta_filter],
```
brimoor, line 5433: In addition to adding new parameters to `_kwargs()`, they also need to be added to `_params()`. The latter declares the type(s) of the arguments so that the App can properly render instances of the stage in the view bar.

fiftyone/core/stages.py
```diff
@@ -5205,26 +5220,122 @@ class SelectFields(ViewStage):
...
+        """The meta filters to apply, as a string or dict."""
+        return self._meta_filter or {}
+
     def get_selected_fields(self, sample_collection, frames=False):
         if frames:
             return self._get_selected_frame_fields(sample_collection)
 
         return self._get_selected_fields(sample_collection)
 
+    def get_meta_filtered_fields(self, schema):
```
brimoor, line 5251: Are you open to converting this to a private method and then simply calling it from inside the existing `get_selected_fields()` method?

From the perspective of users, I don't think it matters whether this stage is selecting fields based on an explicit list of `field_names` or a dynamic set of field names inferred based on a provided `meta_filter`.

With this change, we wouldn't need to add the new `get_meta_filtered_fields()` method to the `ViewStage` interface.

For context, the only reason `get_selected_fields()` and `get_excluded_fields()` were added to the interface is that it is important for the rest of the library to know if a view omits fields from the full dataset, as this has implications on what kind of database operations can be used when saving edits to samples. But it's not important whether the selected fields were chosen in a fancy way.

fiftyone/core/collections.py
```diff
@@ -5497,11 +5499,25 @@ def select_fields(self, field_names=None, _allow_missing=False):
...
+                that have a specific key in their info, you can use
+                ``meta_filter=dict(key="value")``.
+
         Returns:
             a :class:`fiftyone.core.view.DatasetView`
         """
+
         return self._add_view_stage(
-            fos.SelectFields(field_names, _allow_missing=_allow_missing)
+            fos.SelectFields(
```
brimoor, line 5516: We've generally tried to maintain parity between `SelectFields` and `ExcludeFields`. Can you also add the `meta_filter` parameter to `ExcludeFields` so that users can *exclude* fields based on a dynamic criteria if they want?

tests/unittests/dataset_tests.py
```diff
@@ -4833,6 +4833,203 @@ def test_select_exclude_dynamic_fields(self):
         self.assertFalse(frame.has_field("field"))
         self.assertFalse(frame.predictions.detections[0].has_field("field"))
 
+    @drop_datasets
+    def test_filter_fields(self):
```
brimoor, line 4837: Thanks for the tests! Can you move these to `tests/unittests/view_tests.py`? There should already be some tests for the existing `SelectFields` syntax over there.

tests/unittests/dataset_tests.py
```diff
@@ -4833,6 +4833,203 @@ def test_select_exclude_dynamic_fields(self):
...
+            "owner": "joe",
+            "test": True,
+            "other": "this is a unique info value",
+        }
+        field_3.info = {"one": {"two": {"three": "test123"}}}
+
+        # doesn't return anything on empty string
+        view = dataset.select_fields(meta_filter="")
+        fields = view.get_field_schema(flat=True)
+        assert "field_1" not in fields
```
brimoor, line 4891: nit: can you follow the existing convention of using methods like `self.assertNotIn`?

tests/unittests/dataset_tests.py
```diff
@@ -4833,6 +4833,203 @@ def test_select_exclude_dynamic_fields(self):
...
+                "e_2": {"f_3": "oo", "g_3": {"h_4": {"i_5": {"j_6": "nope"}}}}
+            },
+        }
+        field_2.info = {
+            "list": [1, 2, 3],
+            "owner": "joe",
+            "test": True,
+            "other": "this is a unique info value",
+        }
+        field_3.info = {"one": {"two": {"three": "test123"}}}
```
brimoor, line 4886: Strictly speaking you should call `field.save()` after modifying a field's info/description to ensure that it is persisted to the database.

You're getting away with it here because the only uses of this info are done using info that's in-memory on the `dataset` singleton. But if a call to `dataset.reload()` were inserted for some reason (which ordinarily should be harmless) the tests would start failing because the in-memory data on the `dataset` would be replaced with the previous values from the database.

brimoor (changes requested):
This implementation is looking on the right track to me 💪 I left some stylistic comments for ya

tests/unittests/dataset_tests.py
```diff
@@ -4833,6 +4833,203 @@ def test_select_exclude_dynamic_fields(self):
...
+            meta_filter=dict(one=dict(two=dict(three="test123")))
+        )
+        fields = view.get_field_schema(flat=True)
+        assert "field_1" not in fields
+        assert "field_2" not in fields
+        assert "field_3" in fields
+        assert "ground_truth" not in fields
+
+        # match recursively located key:value
+        view = dataset.select_fields(meta_filter=dict(three="test123"))
```
brimoor, line 4958: Just confirming: with the current design, there's no way to say you want to match *only* on `description` or on `info`, right?

fiftyone/core/stages.py
```diff
@@ -788,6 +830,124 @@ def validate(self, sample_collection):
...
+        return
+
+    # meta_filter = self._meta_filter
+    if isinstance(meta_filter, str):
+        str_filter = meta_filter
+        meta_filter = {}
+    else:
+        str_filter = None
+
+    description_filter = meta_filter.pop("description", None)
```
kaixi-wang, line 844: you're removing name, description, and info here

tests/unittests/dataset_tests.py
nebulae: there is now, and also name

manivoxel51 (approved):
nice work! 🍨 

your test cases were very helpful. I'll have to hook this up to the UI next

brimoor (changes requested):
@nebulae FYI I implemented one necessary update to this implementation here: https://github.com/voxel51/fiftyone/pull/2939. I recommend merging that before making any further changes to avoid merge conflicts.

I have two other enhancement requests:
1. Can you support `meta_filter={"info.key": "value"}` with dot notation as shorthand for `meta_filter=dict(info=dict(key="value"))`? There's an existing convention of using `embedded.field.name` notation throughout the SDK when referring to nested fields. In fact, it would be fine with me to *require* the dot notation and not support nested dicts at all (for user-facing syntax), but if you want to keep nested dicts as an undocumented syntax, that would be fine with me
2. Currently `meta_filter` is only applied to top-level fields, but we could expand to nested fields as well

Perhaps we should discuss 2 offline with @manivoxel51. In the UI he's building, in manual field selection mode, there's a toggle to control whether nested fields are included or not. When using the search UI (`meta_filter`), there could be a similar toggle to control whether nested fields are included. For example, multiple users may have populated custom attributes on the `Detection` instances within a `Detections` field. Including nested fields in `meta_filter` would allow for only selecting the attributes that I added while excluding attributes that someone else added. But it might be the case that we need to expose this toggle on the view stages themselves to avoid undesirable behavior when one only *intends* to process top-level fields.

nebulae:
> 2. Currently `meta_filter` is only applied to top-level fields, but we could expand to nested fields as well
> 
> Perhaps we should discuss 2 offline with @manivoxel51. 

@brimoor for this one, is it safe to say that if we have a toggle to include nested fields, then those fields are filtered and returned in this way: 
- if none of the nested fields have a match, return them all (if the parent matched the filter and is being returned)
- if one of the nested fields has a match, only return that one, regardless of if the parent is being returned

or, would the toggle to include nested fields lead the user to believe that the nested fields will be returned / included regardless of if there is a match, if the parent is being returned?

brimoor:
> @brimoor for this one, is it safe to say that if we have a toggle to include nested fields, then those fields are filtered and returned in this way:
> 
> * if none of the nested fields have a match, return them all (if the parent matched the filter and is being returned)
> * if one of the nested fields has a match, only return that one, regardless of if the parent is being returned
> 
> or, would the toggle to include nested fields lead the user to believe that the nested fields will be returned / included regardless of if there is a match, if the parent is being returned?

discussed offline, we believe we can wire up an `include_nested_fields=True/False` param to `get_field_schema(flat=include_nested_fields)` and it will... just work! 🤞

fiftyone/core/collections.py
```diff
@@ -3632,14 +3634,40 @@ def exclude_fields(self, field_names, _allow_missing=False):
...
+            meta_filter (None): a filter that dynamically excludes fields in
+                the collection's schema according to the specified rule, which
+                can be matched against the field's ``name``, ``type``,
+                ``description``, and/or ``info``. For example:
+
+                -   Use ``meta_filter="2023"`` or
+                    ``meta_filter={"any": "2023"}`` to exclude fields that have
+                    the string "2023" anywhere in their name, type,
+                    description, or info
+                -   Use ``meta_filter={"type": "StringField"}`` or
```
brimoor, line 3648: FYI I removed `"type": fo.StringField` from the docstring because `ViewStage` kwargs need to be JSON serializable in order to be saved, loaded in the App, etc. 

I didn't touch the actual logic in the `type_filter` though, so the syntax is still available as an undocumented option.

Note: other `ViewStage` classes accept non JSON values, but they are careful to convert the relevant data into JSON serializable values internally in such a way that they can be serialized/deserialized.

PR #2750, opened 2023-03-08
FiftyOne V2 GraphQL Boilerplate

benjaminpkane:
Adds boilerplate package code and build/publish workflow. Also updates `strawberry-graphql` in the code package.

PR #2910, opened 2023-04-20
Add stack to GraphQL error extensions

benjaminpkane:
Adds back `stack` to GraphQL exceptions ensuring error boundaries work. Unfortunately, this regression was introduced by the unexpected upgrade then downgrade of strawberry caused by #2750


kaixi-wang (approved):
lgtm

PR #2913, opened 2023-04-20
refactor plugin listing

ritch:


fiftyone/plugins/definitions.py
```diff
@@ -0,0 +1,162 @@
...
+
+def load_plugin_definition(metadata_file):
+    """Loads the plugin definition from the given metadata_file."""
+    with open(metadata_file, "r") as f:
+        metadata_dict = yaml.load(f, Loader=yaml.FullLoader)
+        module_dir = os.path.basename(os.path.dirname(metadata_file))
+        definition = PluginDefinition(module_dir, metadata_dict)
+        if not definition.js_bundle:
+            # check if package.json exists
+            package_json_path = os.path.join(module_dir, "package.json")
```
kaixi-wang, line 117: I thought the decision was to use yml for all packages. Are we going to also check package.json?

PR #2914, opened 2023-04-20
Updating docs header/footer

brimoor:
Updating header/footer to match current https://voxel51.com.


findtopher (approved):
yep, looks right to me

Issue #2915, opened 2023-04-21
[BUG] Scale AI export 


fnobis:
### Describe the problem

I am experimenting with the export to scale API. For this, I created an image dataset without any labels. Even though I do not have hypothesis labels available, I would like to use the export to scale. Since I want use the fo import function, relying on the same UUIDs from the scale annotations.  When I try to export the image files to send to annotation, I get the following error:

```
File [/usr/local/lib/python3.10/dist-packages/fiftyone/utils/scale.py:476]
.
.
.
--> [476](file:///usr/local/lib/python3.10/dist-packages/fiftyone/utils/scale.py?line=475)         "hypothesis": anno_dict,
    [477](file:///usr/local/lib/python3.10/dist-packages/fiftyone/utils/scale.py?line=476)     }
    [479](file:///usr/local/lib/python3.10/dist-packages/fiftyone/utils/scale.py?line=478) etas.write_json(labels, json_path)

UnboundLocalError: local variable 'anno_dict' referenced before assignment
```

I could fix this locally and export the images without labels by creating the anno_dict =  {} variable in [line 430](https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/scale.py#L430)

### Documentation link outdated

Additionally, I noticed that the documentation links at scale changed [here](https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/scale.py#L263) and further locations.
E.g. the "#" got replaced with "/"
https://docs.scale.com/reference#general-image-annotation -> https://docs.scale.com/reference/general-image-annotation

### Code to reproduce issue

```
import fiftyone.utils.scale as fous
fous.export_to_scale(collection_to_export_wo_labels, exported_file_name)
```

### Other info/logs

Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached. Please do not use screenshots for sharing text. Code snippets should be used instead when providing tracebacks, logs, etc.

### What areas of FiftyOne does this bug affect?

-   [ ] `App`: FiftyOne application issue
-   [x] `Core`: Core Python library issue
-   [ ] `Server`: FiftyOne server issue

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the FiftyOne codebase?

-   [ ] Yes. I can contribute a fix for this bug independently
[...]

PR #2916, opened 2023-04-21
Fixing Scale docs links + unlabeled annotation tasks

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/2915.


jacobmarks (approved):
Mostly changing links. Overall LGTM 👍 !

PR #2918, opened 2023-04-21
Handling missing label fields when deleting labels

brimoor:
`delete_labels()` did not previously handle the case where a sample with `None`-valued field was involved.

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.take(10, seed=51).set_field("ground_truth", None).save()

label_id = dataset.exists("ground_truth").first().ground_truth.detections[0].id
view = dataset.select_labels(ids=label_id, fields="ground_truth")

assert len(view) == 1

# Previously would fail; now succeeds
dataset.delete_labels(ids=label_id, fields="ground_truth")

assert len(view) == 0
```

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -2999,11 +2999,12 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
             ops = []
             if issubclass(label_type, fol._LABEL_LIST_FIELDS):
                 array_field = field + "." + label_type._LABEL_LIST_FIELD
+                query = {array_field: {"$exists": True}}
 
                 if view_ids is not None:
                     ops.append(
                         UpdateMany(
-                            {},
+                            query,
                             {
                                 "$pull": {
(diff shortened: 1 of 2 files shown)


fiftyone/core/dataset.py
```diff
@@ -2999,11 +2999,12 @@ def delete_labels(
             ops = []
             if issubclass(label_type, fol._LABEL_LIST_FIELDS):
                 array_field = field + "." + label_type._LABEL_LIST_FIELD
+                query = {array_field: {"$exists": True}}
```
kaixi-wang, line 3002: not sure if this is applicable without looking at the rest of the aggregation/knowing the root, but shouldn't the schema be the same? (field should exist but may be null).

brimoor: Not sure what you're asking. These updates are looking for documents in a nested list field and deleting them by ID. So there needs to be a `query` to skip documents that are missing the top-level field, or else an error will occur.

benjaminpkane (approved):
LGTM

PR #2925, opened 2023-04-25
Relax starlette dependency

benjaminpkane:
Relaxes the `starlette` dependency per a use request. All minor versions were verified to work


sashankaryal (approved):
lgtm

PR #2926, opened 2023-04-25
add support for indexed encoding in pngs

### LINKED ISSUE #2920: [BUG] Cannot visualise/show RGB semantic segmentations masks
### Instructions

Hi,
I'm trying to visualise imported RGB masks (.png) on 51, but while the masks looks like they are loaded, for some reason the image doesn't show up at all.

### System information

-   **OS Platform and Distribution**: `Windows 11, 22H2`, Build: `22621.1555`) using WSL2: `Ubuntu-20.04`
-   **Python version** (`python --version`): `3.10.9`
-   **FiftyOne version** (`fiftyone --version`): 
-   **FiftyOne installed from** (pip or source): `pip`, version: `22.3.1`
-   **Virtual Env Conda**:  `22.9.0`
-   **Chrome Version**: `112.0.5615.138`

### Commands to reproduce

Steps:
1. Create a new virtual env with the specks above
2. Update the dataset paths to point to where the image and mask is stored on your local env.
3. Run the code below on a Jupyter env on VS Code
4. Open 51 on browser using Chrome on `http://localhost:5151`

### Describe the problem

Fiftyone doesn't display correctly RGB masks stored as png image files.
The only way this can work is by first converting the mask to greyscale, save it, and then load that mask.

### Code to reproduce issue
RGB code:
```python
import fiftyone  as fo

dataset_rgb = fo.Dataset.from_dir(
    data_path="path/to/images",
    labels_path="path/to/masks_rgb",
    dataset_type=fo.types.ImageSegmentationDirectory,
    label_field="segmentations",
)

foul.segmentations_to_detections(
    dataset,
    "segmentations",
    "instances",
    mask_targets={
        "#000000": "__background__",
        "#32ff32": "big_dog",
        "#ff3232": "small_dog"
    },
    mask_types="thing"
)
dataset.save()
```
Greyscale mask code:
```python
import fiftyone  as fo

    data_path="path/to/images",
    labels_path="path/tomasks",
    dataset_type=fo.types.ImageSegmentationDirectory,
    label_field="segmentations",
)

foul.segmentations_to_detections(
    dataset,
    "segmentations",
    "instances",
    mask_targets={111: "big_dog", 170: "small_dog"},
    mask_types="thing"
)
dataset.save()

```

### Other info/logs
While converted masks to detections seem to have happenned correctly, the image doesn't load:
image
Example of detection:
image

-   [ ] `Core`: Core Python library issue

### Willingness to contribute

-   [x] Yes. I would be willing to contribute a fix for this bug with guidance from the FiftyOne community

### Data used for this use case/but.
dogs
dogs
dogs

Any suggestions, advice, or guidance on how to tackle this is welcomed :).
Thanks

sashankaryal:

fixes #2920, essentially adding support for color type 3 PNGs (https://www.w3.org/TR/PNG-Chunks.html)
Screenshot 2023-04-25 at 11 34 06 AM
Screenshot 2023-04-25 at 11 33 56 AM

Locally.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/looker/src/worker/indexed-png-decoder.ts
+++ b/app/packages/looker/src/worker/indexed-png-decoder.ts
@@ -1,43 +1,39 @@
+export type ValidPngBitDepth = 1 | 2 | 4 | 8;
+export type ColorPalette = number[][];
+
 export const indexedPngBufferToRgb = (
-  data: Uint8Array,
-  bitDepth: 1 | 2 | 4 | 8,
-  colorPalette: number[][]
+  inputData: Uint8Array,
+  bitDepth: ValidPngBitDepth,
+  colorPalette: ColorPalette
 ) => {
   const indicesPerByte = 8 / bitDepth;
-  const numIndexes = data.length * indicesPerByte;
-  const resultSize = numIndexes * 3;
+  const numIndices = inputData.length * indicesPerByte;
+  const resultSize = numIndices * 3;
+  const newRgbArray = new Uint8Array(resultSize);
 
-  const rgbArray = new Uint8Array(resultSize);
-
-  let rgbIdx = 0;
-  let paletteIdx = 0;
-
-  const colorIndices = new Uint8Array(numIndexes);
+  let rgbOffset = 0;
 
   // Determine the initial bit mask based on the bit depth:
   const initialBitMask = ((1 << bitDepth) - 1) << (8 - bitDepth);
 
   // extract color indexes from data based on bit depth
-  for (const byte of data) {
+  for (const byte of inputData) {
     let currentBitMask = initialBitMask;
-    let bitsToShift = 8 - bitDepth;
+    let remainingBits = 8;
 
-    while (currentBitMask) {
-      colorIndices[paletteIdx++] = (byte & currentBitMask) >> bitsToShift;
+    while (remainingBits >= bitDepth) {
+      const index = (byte & currentBitMask) >> (remainingBits - bitDepth);
+      const color = colorPalette[index];
+      if (!color) {
+        throw new Error("Incorrect index of palette color");
+      }
+      newRgbArray.set(color, rgbOffset);
+      rgbOffset += 3;
+
+      remainingBits -= bitDepth;
       currentBitMask >>= bitDepth;
-      bitsToShift -= bitDepth;
     }
   }
 
-  // map color indices to rgb values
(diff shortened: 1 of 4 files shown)


app/packages/looker/src/worker/indexed-png-decoder.ts
```diff
@@ -0,0 +1,39 @@
...
+  const resultSize = numIndices * 3;
+  const newRgbArray = new Uint8Array(resultSize);
+
+  let rgbOffset = 0;
+
+  // Determine the initial bit mask based on the bit depth:
+  const initialBitMask = ((1 << bitDepth) - 1) << (8 - bitDepth);
+
+  // extract color indexes from data based on bit depth
+  for (const byte of inputData) {
```
benjaminpkane, line 20: I think this will be slow. Can this be changed to a regular for loop?

benjaminpkane (changes requested):
Looks good! I just think the for loop should be tweaked for performance

app/packages/looker/src/worker/indexed-png-decoder.ts
benjaminpkane: It could be more than just the for loop. Just thinking some benchmarking could be done. In my experience a bunch of garbage collection / memory allocation can happen if these for loops over large arrays aren't kept simple.

sashankaryal: - Good observation, thanks, I changed it to a regular for loop in the next commit.
- So I did some benchmarking...some notes:

1. This input array, because it uses color type 3 encoding, is naturally terse. We're "unpacking" it. The one used in the example has length 115248.
2. I did some poor-man benchmarking in the browser (in the app code) with bit depth = 2 (a middle ground with four colors supported) with 1e5, 1e6, and 1e7 len arrays. Report:

```
Sample size: 115248, average elapsed time: 12.40 ms (REALISTIC INPUT)
Sample size: 1e5, average elapsed time: 11.60 ms (REALISTIC INPUT)
Sample size: 1e6, average elapsed time: 127.20 ms
Sample size: 1e7, average elapsed time: 1779.10 ms
```

Changing to regular for-loop had almost no noticeable impact, perhaps v8's optimization.

Bit depth 1 would be slower but has no practical usage, and bit depths 4 and 8 would be significantly faster.

PR #2941, opened 2023-05-01
Document New Teams Management SDK

swheaton:
Documentation of FO Teams Management SDK

### DIFF
--- a/docs/source/teams/index.rst
+++ b/docs/source/teams/index.rst
@@ -68,6 +68,12 @@ pages on this site apply to Teams deployments as well.
     :button_text: Migrate to Teams
     :button_link: migrations.html
 
+.. customcalloutitem::
+    :header: Management SDK
+    :description: Learn how to manage your Teams deployment via code with the Management SDK.
+    :button_text: Manage your Teams deployment with code
+    :button_link: management_sdk.html
+
 .. End callouts ---------------------------------------------------------------
 
 .. raw:: html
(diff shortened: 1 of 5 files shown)


swheaton:
Copying relevant comment over from https://github.com/voxel51/fiftyone-teams/pull/222

> It's going to be a bit awkward to get the API reference docs into docs.voxel51.com. Probably deserves an offline discussion to brainstorm ideas.
> 
> Given the way things are currently organized, I think my ideal would be:
> 
> Overview
> ...
> User Guide
> FiftyOne Teams
>     Overview
>     Installation
>     Cloud-backed media
>     Roles and permissions
>     FiftyOne Teams App
>     Migrations
>     API Reference   <-- new section that contains Teams-only docs
> Integrations
> ...
> API Reference   <-- OSS-only 
> ...
> In order to make that happen, we'll likely need some symlinking similar to how that is done for the fiftyone-brain docs here:
> https://github.com/voxel51/fiftyone/blob/develop/docs/generate_docs.bash
> 
> I think having auto-generated API reference docs appear should be doable for the net-new fiftyone.management module.
> 
> I've currently done something very hacky to document the Teams-specific methods that are scattered throughout the main fiftyone namespace...
> https://docs.voxel51.com/teams/cloud_media.html#api-reference
> 
> ...I manually copied them here: https://github.com/voxel51/fiftyone/blob/develop/docs/source/teams/cloud_media.rst#api-reference
> 
> We may be able to get those to appear in the Teams-specific API Reference section above by manually writing an .rst file that defines what to generate docstrings for...

swheaton:
I also plan to add API Key generation documentation (UI), as that is currently lacking.

swheaton:
Closing in favor of https://github.com/voxel51/fiftyone/pull/3068

PR #2948, opened 2023-05-03
Remove unused loading text for tags

benjaminpkane:
This update prevents the `tags` aggregation from being executed when it is not needed


brimoor (changes requested):
@benjaminpkane just tested this locally and I'm seeing `samples` rather than `XXX samples` above the grid. Count aggregation no longer running?

PR #2956, opened 2023-05-04
Allow config files to not exist

brimoor:


kaixi-wang (commented):
Are these changes necessary? I'd think removing these statements and checks would lead to errors downstream and make it hard debugging why

brimoor:
> Are these changes necessary? I'd think removing these statements and checks would lead to errors downstream and make it hard debugging why

I think these `locate_XXX()` methods should always return paths on disk. The methods already document that the file might not actually exist. But for some reason I cannot deduce/remember, if the user customized the config location via env var, it will switch to raising an error if the file doesn't exist.

Since the caller must already check for existence when the config location is default, I think it is better to never raise an error  in these methods.

benjaminpkane (approved):
LGTM

PR #2965, opened 2023-05-05
Sync map style to session storage

benjaminpkane:
Per user feedback, this PR adds map style as a browser storage persistent setting when its value is not a theme value (i.e. `Dark` and `Light`)


sashankaryal (approved):
lgtm

PR #2967, opened 2023-05-05
add run_doc check before results check

kaixi-wang:
prevent throwing an error that prevents loading a dataset when run doc doesn't exist

### DIFF
--- a/fiftyone/core/runs.py
+++ b/fiftyone/core/runs.py
@@ -582,7 +582,7 @@ class Run(Configurable):
 
         run_doc = cls._get_run_doc(samples, key)
 
-        if not run_doc.results:
+        if not run_doc or not run_doc.results:
             return None
 
         # Load run config


fiftyone/core/runs.py
```diff
@@ -582,7 +582,7 @@ def load_run_results(
 
         run_doc = cls._get_run_doc(samples, key)
 
-        if not run_doc.results:
+        if not run_doc or not run_doc.results:
```
brimoor, line 585: Given how `_get_run_doc()` is implemented, how would `not run_doc` be true?

https://github.com/voxel51/fiftyone/blob/90b26829cfb09ff88ef8896a2e83559c3307505a/fiftyone/core/runs.py#L745-L755

brimoor (changes requested):

fiftyone/core/runs.py
kaixi-wang: the current blocking error is that in some cases
`run_docs.get(key, None)` will return a nonexistant ObjectId, so when mongoengine tries to derefernce it, it fails so it becomes a `Dbref` that presents as None

brimoor: Okay I see. I think we'll need to take a more comprehensive approach to addressing this issue. There are other methods besides `load_run_results()` that rely on `_get_run_doc()` actually returning a doc.

brimoor: If `_get_run_doc()` can't find a doc, then I think we should log a warning telling the user that they can either:
- Run a util we'll add to `fiftyone.core.odm.database` that attempts to locate the orphan run doc
- use the relevant `delete_XXX()` method to delete the bad run (`delete_run()` will have to be tweaked to not raise an error on bad DBRefs)

brimoor: I implemented the utils here: https://github.com/voxel51/fiftyone/pull/2970

brimoor:
Closing in favor of https://github.com/voxel51/fiftyone/pull/2970.

PR #2968, opened 2023-05-05
Add .github/workflows/e2e.yml

aidavoxel51:

Added e2e.yml to run e2e/* tests in pull requests to develop.

Test job run https://github.com/voxel51/fiftyone/actions/runs/4887646675/jobs/8724476236

        notes for FiftyOne users.

-   [ ] Other


e2e/index.ts
```diff
@@ -29,31 +26,13 @@ const runCypress = async () => {
...
-  if (!options.headed) {
-    throw new Error(
-      "Cypress is running in headless mode, but screenshotting doesn't work as expected in this mode with looker. See https://github.com/cypress-io/cypress/issues/15605 for more details."
-    );
-  }
+  // if (!options.headed) {
+  //   throw new Error(
+  //     "Cypress is running in headless mode, but screenshotting doesn't work as expected in this mode with looker. See https://github.com/cypress-io/cypress/issues/15605 for more details."
+  //   );
+  // }
```
sashankaryal, line 33: could you uncomment these?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,80 @@
+name: Test e2e
+
+on:
```
benjaminpkane, line 3: Following the existing pattern, can we have this workflow run via `on: workflow_call`?

The idea would be we add the workflow call to the PR and Publish workflows.
* https://github.com/voxel51/fiftyone/blob/develop/.github/workflows/pr.yml
* https://github.com/voxel51/fiftyone/blob/develop/.github/workflows/publish.yml

And add it to the `needs` lists so the workflows fail/reject when e2e tests fail
* https://github.com/voxel51/fiftyone/blob/327789be148f1f6c7fb0cb847db4c4f1b3f4447c/.github/workflows/pr.yml#L18
* https://github.com/voxel51/fiftyone/blob/327789be148f1f6c7fb0cb847db4c4f1b3f4447c/.github/workflows/publish.yml#L19
* https://github.com/voxel51/fiftyone/blob/327789be148f1f6c7fb0cb847db4c4f1b3f4447c/.github/workflows/publish.yml#L43

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Configure
+        id: test_config
+        run: |
+          python tests/utils/setup_config.py
+          python tests/utils/github_actions_flags.py
+
+      - name: FFmpeg
+        uses: FedericoCarboni/setup-ffmpeg@v2
+
+      - name: Run e2e tests
```
benjaminpkane, line 62: The test step is currently passing when tests fail. Ideally this will fail when tests fail so the workflow calls can then fail

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,80 @@
...
+        run: |
+          FIFTYONE_DATABASE_NAME=cypress python ../fiftyone/server/main.py --address 0.0.0.0 --port 8787 &
```
sashankaryal, line 64: @benjaminpkane Is this supposed to replace the fake server? 🥳  that's cool, but `yarn start` will no longer start cypress correctly locally, no?

sashankaryal: I guess we could add something like:
```
"cy:run": "FIFTYONE_DATABASE_NAME=cypress python ../fiftyone/server/main.py --address 0.0.0.0 --port 8787 & cypress run"
```

benjaminpkane: Yes, I'll tweak it

e2e/index.ts
benjaminpkane: Yeah, headed is required. Reverted

.github/workflows/e2e.yml
aidavoxel51: I saw you made the changes, were you able to test it?

.github/workflows/e2e.yml
benjaminpkane: I'm proposing the background server is started manually by the developer when testing locally with `yarn start-test-server`. I updated the readme. Trying to avoid creating orphaned processes

benjaminpkane (approved):
LGTM. I'm comfortable merging. I did end up removing some of the existing cypress subprocess code in favor of a blocking approach to the Python subprocess execution. @sashankaryal @aidavoxel51

PR #2970, opened 2023-05-08
Adding utils to patch broken saved view/run docs

brimoor:
Formalized version of https://github.com/voxel51/fiftyone/pull/2967.

Note: in a follow-on PR https://github.com/voxel51/fiftyone/pull/2971, I add hooks into the main codebase to warn users when running these utils is necessary.

When mongoengine encounters a ReferenceField that contains an ObjectId with no matching document, the doc will be a DBRef. We're not currently sure why this happens... but we've observed that ObjectIDs in `dataset_doc` can somehow get out of sync with their corresponding saved view/run documents. The new utilities added in this PR can patch this issue and bring the dataset back to a usable state with no data loss.

Example usage:

```py
from bson import ObjectId, DBRef

import fiftyone as fo
import fiftyone.zoo as foz

db = foo.get_db_conn()

dataset = foz.load_zoo_dataset("quickstart", max_samples=10, dataset_name="zzz")
dataset.save_view("test", dataset.limit(10))

# Break some references
dataset_dict = db.datasets.find_one({"name": dataset.name})
dataset_dict["saved_views"] = [ObjectId()]
dataset_dict["evaluations"] = {"eval": ObjectId()}
db.datasets.replace_one({"name": dataset.name}, dataset_dict)

# Bad IDs present as DBRefs
dataset.reload()
assert isinstance(dataset._doc.saved_views[0], DBRef)
assert isinstance(dataset._doc.evaluations["eval"], DBRef)

foo.patch_saved_views(dataset.name)
# Purging 1 bad saved view ID(s) {ObjectId('6458f353efee59d28b1cefde')} from dataset
# Adding 1 misplaced saved view(s) [(ObjectId('6458f346efee59d28b1cefdd'), 'test')] back to dataset

foo.patch_evaluations(dataset.name)
# Purging 1 evaluation(s) {('eval', ObjectId('6458f353efee59d28b1cefdf'))} from dataset
# Adding 1 misplaced evaluation(s) {('eval', ObjectId('6458f346efee59d28b1cefda'))} to dataset

# Now things are fixed
dataset.reload()
assert isinstance(dataset._doc.saved_views[0], foo.SavedViewDocument)
assert isinstance(dataset._doc.evaluations["eval"], foo.RunDocument)
```


fiftyone/core/odm/database.py
```diff
@@ -864,6 +1030,102 @@ def delete_dataset(name, dry_run=False):
...
+
+    _logger.info(
+        "Deleting %d saved views from dataset '%s'",
+        len(del_ids),
+        dataset_name,
+    )
+
+    if not dry_run:
+        dataset_dict["saved_views"] = []
+        conn.datasets.replace_one({"name": dataset_name}, dataset_dict)
```
kaixi-wang, line 1124: why `replace` instead of `update`?

fiftyone/core/odm/database.py
```diff
@@ -795,6 +795,172 @@ def list_datasets():
...
+            num_missing_runs,
+            run_str,
+            missing_runs,
+        )
+        runs_dict.update(missing_runs)
+        made_changes = True
+
+    if made_changes:
+        dataset_dict[runs_field] = runs_dict
+        conn.datasets.replace_one({"name": dataset_name}, dataset_dict)
```
kaixi-wang, line 961: since we are updating only a single field, why are we replacing the entire doc?

brimoor: Good call, will switch to `update_one()`

kaixi-wang (commented):
latest development in the saga, run.results objectId isn't pointing to an existing object...

I've been attempting to implement a more generic solution that would hook into/address any ReferenceField, but the way the doc classes are defined and the auto-dereferencing/referencing on access makes it much more complicated than I had anticipated

kaixi-wang (approved):
logic looks correct to me

kaixi-wang (approved):
manually checked that inserting the objectId into the list works

PR #2971, opened 2023-05-08
Log warnings that hint how to patch broken saved views and runs

brimoor:
If broken saved view/run references are encountered, log warnings that describe how to run the utils introduced in https://github.com/voxel51/fiftyone/pull/2970 to patch the issue *and* omit the broken references from all operations that would raise an error when encountering broken refs.

Example usage:

```py
from bson import ObjectId, DBRef

import fiftyone as fo
import fiftyone.zoo as foz

db = foo.get_db_conn()

dataset = foz.load_zoo_dataset("quickstart", max_samples=10, dataset_name="zzz")
dataset.save_view("test", dataset.limit(10))

# Break some references
dataset_dict = db.datasets.find_one({"name": dataset.name})
dataset_dict["saved_views"] = [ObjectId()]
dataset_dict["evaluations"] = {"eval": ObjectId()}
db.datasets.replace_one({"name": dataset.name}, dataset_dict)

dataset.reload()
dataset.list_saved_views()  # []
# This dataset's saved view references are corrupted. Run fiftyone.core.odm.database.patch_saved_views('zzz') and dataset.reload() to resolve
dataset.list_evaluations()  # []
# This dataset's evaluation references are corrupted. Run fiftyone.core.odm.database.patch_evaluations('zzz') and dataset.reload() to resolve

foo.patch_saved_views(dataset.name)
# Purging 1 bad saved view ID(s) {ObjectId('6458f353efee59d28b1cefde')} from dataset
# Adding 1 misplaced saved view(s) [(ObjectId('6458f346efee59d28b1cefdd'), 'test')] back to dataset

foo.patch_evaluations(dataset.name)
# Purging 1 evaluation(s) {('eval', ObjectId('6458f353efee59d28b1cefdf'))} from dataset
# Adding 1 misplaced evaluation(s) {('eval', ObjectId('6458f346efee59d28b1cefda'))} to dataset

# Now things are fixed
dataset.reload()
dataset.list_saved_views()  # ['test']
dataset.list_evaluations()  # ['eval']
```

### DIFF
--- a/fiftyone/core/odm/dataset.py
+++ b/fiftyone/core/odm/dataset.py
@@ -5,6 +5,10 @@ Documents that track datasets and their sample schemas in the database.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+import logging
+
+from bson import DBRef
+
 import eta.core.utils as etau
 
 from fiftyone.core.fields import (
@@ -26,14 +30,23 @@ import fiftyone.core.utils as fou
 
 # import fiftyone.core.colorscheme as foc
 
+from .database import (
+    patch_saved_views,
+    patch_annotation_runs,
+    patch_brain_runs,
+    patch_evaluations,
+)
 from .document import Document
 from .embedded_document import EmbeddedDocument
 from .runs import RunDocument
-from .views import SavedViewDocument
 from .utils import create_field
+from .views import SavedViewDocument
 
 fol = fou.lazy_import("fiftyone.core.labels")
 fom = fou.lazy_import("fiftyone.core.metadata")
+
+
+logger = logging.getLogger(__name__)
 
 
 class SampleFieldDocument(EmbeddedDocument):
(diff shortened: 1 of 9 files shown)


kaixi-wang (commented):
I really don't think we should be deleting these automatically. It hides the issue and will make it even harder to debug

fiftyone/core/odm/dataset.py
```diff
@@ -480,21 +493,109 @@ class DatasetDocument(Document):
...
+        # When mongoengine encounters a ReferenceField that contains an
+        # ObjectId with no matching document, the doc is a falsey DBRef.
+        #
+        # We're not currently sure why this happens... but we've observed that
+        # ObjectIDs in DatasetDocument can somehow get out of sync with their
+        # corresponding view documents. This attempts to patch on-the-fly.
+        #
+        if any(isinstance(doc, DBRef) for doc in saved_views):
+            try:
+                patch_saved_views(self.name)
```
kaixi-wang, line 509: wrapping a potential db write in the read method might lead to teams bugs

brimoor:
> 

It *fixes* the issue. It brings the dataset doc back into sync with the corresponding documents in the `views`/`runs` collections. If there are no docs there, then the view/run is gone forever.

fiftyone/core/odm/dataset.py
brimoor: > I really don't think we should be deleting these automatically. It hides the issue and will make it even harder to debug


It *fixes* the issue, not *hides*. These utils bring the dataset doc back into sync with the corresponding documents in the views/runs collections. If there are no docs in the external collections, there's no point in having an orphan `ObjectId` in the dataset doc.

brimoor:
> I really don't think we should be deleting these automatically. It hides the issue and will make it even harder to debug

It *fixes* the issue, not hides. These utils bring the dataset doc back into sync with the corresponding documents in the views/runs collections. If there are no docs in the external collections, there's no point in having an orphan `ObjectId` in the dataset doc.

fiftyone/core/odm/dataset.py
brimoor: Hmm, its a valid point that this introduces the possibility of an edit operation being attempted when a user is only trying to do a read operation...

brimoor: I can update to instead log a warning telling the user they should manually run the relevant `patch_XXX()` utils.

brimoor: ^updated

fiftyone/core/odm/dataset.py
```diff
@@ -480,21 +493,81 @@ class DatasetDocument(Document):
     brain_methods = DictField(ReferenceField(RunDocument))
     evaluations = DictField(ReferenceField(RunDocument))
 
+    def get_saved_views(self):
+        saved_views = []
+        for view_doc in self.saved_views:
+            if not isinstance(view_doc, DBRef):
+                saved_views.append(view_doc)
+            else:
+                logger.warning(
```
kaixi-wang, line 502:
```suggestion
                oid = getattr(ref, "id", "")
                col = getattr(ref, "collection", "")
                created = getattr(ref.id, "generation_time", "") if oid else ""
                logger.error(
                    f"Could not resolve ObjectId('{oid}') from '{col}' collection created at {created}"
                )
`````

Can we include something more informative like the above in the logging message? I think it will help with diagnosing the problem

kaixi-wang (approved):
image

fiftyone/core/odm/dataset.py
brimoor: The user could just pass `dry_run=True` to the relevant patch method to see the offending IDs. Let's just stick with that for now

PR #2981, opened 2023-05-08
improve listing perf of plugins

ritch:
Before:

```
Time to run test_core_list_plugins: 47.207579135894775 seconds
```

```
Time to run test_core_list_plugins: 0.10054922103881836 seconds
```

The output is the same for all values of `FIFTYONE_PLUGINS_DIR` I've used. The numbers above are for my `~/Project/voxel51` dir, which includes all of fiftyone and other voxel51 source.


fiftyone/plugins/core.py
```diff
@@ -31,6 +31,61 @@
...
+    if not plugins_dir:
+        return []
+
+    normal_locations = _find_files(
+        plugins_dir,
+        _PLUGIN_METADATA_FILENAME,
+        _PLUGIN_METADATA_EXTENSIONS,
+        _MAX_PLUGIN_SEARCH_DEPTH,
+    )
+    node_modules_locations = _find_files(
```
ritch, line 76: These might not really be needed

PR #2993, opened 2023-05-10
Safely serialize reference fields

kaixi-wang:

- Enable loading a dataset even when referenced objects cannot be resolved. 
- Doesn't address the root cause of why this is happening though, but best not to alter/delete anything permanently until it's identified and correctly patched.

### DIFF
--- a/fiftyone/core/odm/dataset.py
+++ b/fiftyone/core/odm/dataset.py
@@ -7,6 +7,7 @@ Documents that track datasets and their sample schemas in the database.
 """
 import eta.core.utils as etau
 from bson import DBRef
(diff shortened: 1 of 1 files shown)


brimoor (commented):
@kaixi-wang I like the idea of omitting broken references 👍

This PR only patches one place where bad references would cause an error. I folded this idea into https://github.com/voxel51/fiftyone/pull/2971, where *all* cases are handled (plus the warning messages explaining how to fix).

brimoor:
We decided to go with https://github.com/voxel51/fiftyone/pull/2970 and https://github.com/voxel51/fiftyone/pull/2971 instead.

Issue #2737, opened 2023-03-06
[INSTALL] Could not import 'fiftyone' in Ubuntu 22.04.1 LTS


stevezkw1998:
### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04): Linux Ubuntu 22.04.1 LTS
-   **Python version** (`python --version`): Python 3.10.6
-   **FiftyOne version** (`fiftyone --version`): '0.19.1'
-   **FiftyOne installed from** (pip or source): `pip install fiftyone`

### Commands to reproduce

```python
In [1]: import fiftyone as fo
```

### Describe the problem

My local system is linux ubuntu 22.04.1 LTS, I install fiftyone using `pip install fiftyone`,  then enter command ipython3 into ipython
And then run this line
```python
In [1]: import fiftyone as fo
```
Then it has bug

### Other info/logs

```python
kewenzhu@kewenzhu:~/code/data-auto-analysis-toolkit-backend$ ipython3
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.31.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import fiftyone as fo
Subprocess ['/home/kewenzhu/.local/lib/python3.10/site-packages/fiftyone/db/bin/mongod', '--dbpath', '/home/kewenzhu/.fiftyone/var/lib/mongo', '--logpath', '/home/kewenzhu/.fiftyone/var/lib/mongo/log/mongo.log', '--port', '0', '--nounixsocket'] exited with error 127:
/home/kewenzhu/.local/lib/python3.10/site-packages/fiftyone/db/bin/mongod: error while loading shared libraries: libcrypto.so.1.1: cannot open shared object file: No such file or directory
---------------------------------------------------------------------------
ServiceListenTimeout                      Traceback (most recent call last)
<ipython-input-1-8c082ccbb59f> in <module>
----> 1 import fiftyone as fo

~/.local/lib/python3.10/site-packages/fiftyone/__init__.py in <module>
     23 __version__ = _foc.VERSION
     24 
---> 25 from fiftyone.__public__ import *
     26 
     27 import fiftyone.core.uid as _fou

~/.local/lib/python3.10/site-packages/fiftyone/__public__.py in <module>
     13 app_config = _foc.load_app_config()
     14 
---> 15 _foo.establish_db_conn(config)
     16 
[...]

Issue #1803, opened 2022-05-28
[BUG] FiftyOne import fails, Ubuntu 22.02 LTS


Nils-ChristianIseke:
### System information

-   **OS Platform and Distribution**: Ubuntu 22.04 LTS
-   **FiftyOne installed from (pip or source)**: pip
-   **FiftyOne version (run `fiftyone --version`)**: Not applicable, due to failing import:
-   **Python version**: Python 3.10.4

### Commands to reproduce

```
conda create -n PDSM pip
conda activate PDSM
pip install --upgrade pip setuptools wheel
pip install fiftyone
python3
import fiftyone

```

### Describe the problem

```
python
Python 3.10.4 (main, Mar 31 2022, 08:41:55) [GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import fiftyone
Subprocess ['/home/nils/anaconda3/envs/PDSM/lib/python3.10/site-packages/fiftyone/db/bin/mongod', '--dbpath', '/home/nils/.fiftyone/var/lib/mongo', '--logpath', '/home/nils/.fiftyone/var/lib/mongo/log/mongo.log', '--port', '0', '--nounixsocket'] exited with error 127:
/home/nils/anaconda3/envs/PDSM/lib/python3.10/site-packages/fiftyone/db/bin/mongod: error while loading shared libraries: libcrypto.so.1.1: cannot open shared object file: No such file or directory
Uncaught exception
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/nils/anaconda3/envs/PDSM/lib/python3.10/site-packages/fiftyone/__init__.py", line 25, in <module>
    from fiftyone.__public__ import *
  File "/home/nils/anaconda3/envs/PDSM/lib/python3.10/site-packages/fiftyone/__public__.py", line 15, in <module>
    _foo.establish_db_conn(config)
  File "/home/nils/anaconda3/envs/PDSM/lib/python3.10/site-packages/fiftyone/core/odm/database.py", line 155, in establish_db_conn
    port = _db_service.port
  File "/home/nils/anaconda3/envs/PDSM/lib/python3.10/site-packages/fiftyone/core/service.py", line 276, in port
    return self._wait_for_child_port()
  File "/home/nils/anaconda3/envs/PDSM/lib/python3.10/site-packages/fiftyone/core/service.py", line 170, in _wait_for_child_port
    return find_port()
[...]

PR #2994, opened 2023-05-10
Adding fiftyone-db build for Ubuntu 22.04

brimoor:
Partially addresses https://github.com/voxel51/fiftyone/issues/1803 and  https://github.com/voxel51/fiftyone/issues/2737

According to the [platform support matrix](https://www.mongodb.com/docs/manual/administration/production-notes/#platform-support-matrix), MongoDB 6.0.4+ now supports Ubuntu 22.04.

This PR adds a `fiftyone-db-ubuntu2204` build with MongoDB 6.0.5 for Ubuntu 22.04.

### DIFF
--- a/.github/workflows/build-db.yml
+++ b/.github/workflows/build-db.yml
@@ -49,6 +49,13 @@ jobs:
         run: |
           cd package/db
           python setup.py bdist_wheel --plat-name mac-arm64
+      - name: Build wheel (Ubuntu 22.04 x86_64)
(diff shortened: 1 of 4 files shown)


benjaminpkane (approved):
LGTM. Unit tests pass in a 22.04 VM. I've uploaded the [new package](https://pypi.org/project/fiftyone-db-ubuntu2204/) so it is now available. I also updated the install script to use 6.0.5 for all systems, but feel free to revert that.

PR #3002, opened 2023-05-11
[HOLD] Avoid and log db inconsistencies

kaixi-wang:

- Limit the db writes to only the update the changed fields to avoid inconsistencies that might arise from replacing the entire doc while a version of it is being accessed by another client 
- add logging/throw errors for incomplete single delete operations for debugging to prevent pipelines from continuing despite failure

(Details)

        notes for FiftyOne users.

-   [ ] Other


fiftyone/core/odm/database.py
```diff
@@ -1345,11 +1345,26 @@ def _delete_run(dataset_name, run_key, runs_field, run_str, dry_run=False):
 
     if not dry_run:
         _logger.info("Deleting %s doc '%s'", run_str, run_id)
-        conn.runs.delete_one({"_id": run_id})
-        conn.datasets.update_one(
-            {"name": dataset_name},
-            {"$set": {runs_field: runs}},
+        # first remove the run from the dataset
```
kaixi-wang, line 1348: in case of interruption, delete from the dataset before deleting the run itself to minimize the possibility of referencing runs already deleted

brimoor (changes requested):
These utils aren't used by the SDK; they're just internal utils to help fix problems when datasets are corrupted. So I don't think any changes are needed here

PR #3004, opened 2023-05-11
allow plugins to resolve paths

ritch:
This is needed to support placement icons in all form factors


imanjra (approved):

ritch (commented):
Ibrahim's commit LGTM

PR #3009, opened 2023-05-11
[HOLD] Fix CVAT 'import annotations()'

kaixi-wang:

- don't delete annotation key when download succeeds
- throw an error if there is an error while downloading for project_id, but enable attempting to download all tasks and cleanup failed ones at the end in case the problem is with a specific task

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/runs.py
+++ b/fiftyone/core/runs.py
@@ -241,7 +241,7 @@ class Run(Configurable):
         run_info = run_info_cls(
             key, version=version, timestamp=timestamp, config=self.config
         )
-        self.save_run_info(samples, run_info)
+        self.save_run_info(samples, run_info, overwrite)
 
     def validate_run(self, samples, key, overwrite=True):
         """Validates that the collection can accept this run.
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -59,6 +59,7 @@ def import_annotations(
     occluded_attr=None,
     group_id_attr=None,
     backend="cvat",
+    anno_key=None,
     **kwargs,
 ):
     """Imports annotations from the specified CVAT project or task(s) into the
@@ -124,6 +125,9 @@ def import_annotations(
         group_id_attr (None): an optional attribute name in which to store the
             group id for labels
         backend ("cvat"): the name of the CVAT backend to use
+        anno_key (None): an optional annotation key to use to store the
+            imported annotations. By default, a new key is generated of the form
+            ``"cvat_import_<project_or_task_id>"``
         **kwargs: CVAT authentication credentials to pass to
             :class:`CVATBackendConfig`
     """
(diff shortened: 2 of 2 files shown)


fiftyone/core/runs.py
```diff
@@ -241,7 +241,7 @@ def register_run(self, samples, key, overwrite=True):
         run_info = run_info_cls(
             key, version=version, timestamp=timestamp, config=self.config
         )
-        self.save_run_info(samples, run_info)
+        self.save_run_info(samples, run_info, overwrite)
```
brimoor, line 244: This change isn't needed. `validate_run()` already handles the `overwrite` parameter

fiftyone/utils/cvat.py
```diff
@@ -124,6 +125,9 @@ def import_annotations(
         group_id_attr (None): an optional attribute name in which to store the
             group id for labels
         backend ("cvat"): the name of the CVAT backend to use
+        anno_key (None): an optional annotation key to use to store the
```
brimoor, line 128: The fact that this method uses an annotation run behind the scenes to import the annotations is not intended to be leaked to the user. It's just a convenient way to load the annotations from CVAT.

fiftyone/utils/cvat.py
```diff
@@ -270,12 +291,23 @@ def import_annotations(
...
                     label_types,
                     anno_backend,
-                    anno_key,
+                    task_anno_key,
                     **kwargs,
                 )
-    finally:
-        anno_backend.delete_run(dataset, anno_key)
-        api.close()
+            except:
```
brimoor, line 297: I am okay with logging warnings for failed tasks rather than terminating, but please keep the existing code structure, eg calling `api.close()` in a `finally` block so that it always runs, even if the user ctrl + C's or something.

brimoor (changes requested):
So the goal here is to gracefully continue if task 51/100 fails? That's fwm, but I believe this PR should be rebranded as an enhancement because there's not a "bug", it's just a different behavior.

fiftyone/utils/cvat.py
kaixi-wang: I put the arg so that a user can specify what key to use... 
If you call this method to import existing annotations, how would you later access the results?

brimoor: The annotations are placed onto your dataset. No need to access the `results` object. If you want directional workflows between a CVAT task and FO; you need to use `annotate()` + `load_annotations()`.

kaixi-wang: > So the goal here is to gracefully continue if task 51/100 fails? That's fwm, but I believe this PR should be rebranded as an enhancement because there's not a "bug", it's just a different behavior.

No, the bug is that imported annotations are getting downloaded and saved with a key like `tmp_` and then automatically getting deleted regardless of success. And when multiple tasks are downloaded, the results of each will overwrite the others. 

But perhaps I don't really understand what an annotation_run's results contain...

brimoor: `delete_annotation_run()` doesn't delete the labels from the dataset. It just deletes the run metadata from the `runs` collection

brimoor:
This is working as expected on `develop`:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = (
    foz.load_zoo_dataset("quickstart", max_samples=2)
    .select_fields()
    .clone()
)

results = dataset.annotate(
    "test",
    label_field="new_field",
    label_type="detections",
    classes=["person"],
    launch_editor=True,
)

# Add 1 box in CVAT

# Manually download annotations from CVAT
new_dataset = fo.Dataset()
fouc.import_annotations(
    new_dataset,
    task_ids=results.task_ids,
    data_path="/tmp/quickstart-cvat",
    label_types={"detections": "new_field"},
    insert_new=True,
    download_media=True,
)
print(new_dataset.count("new_field.detections"))
assert new_dataset.count("new_field.detections") == 1

# Download annotations via integration
print(dataset.count("new_field.detections"))
assert dataset.count("new_field.detections") == 1

# Deleting the run doesn't delete the annotations
dataset.delete_annotation_run("test")
assert dataset.count("new_field.detections") == 1
```

PR #3010, opened 2023-05-11
video dataset should have access to similarity run

lanzhenw:
```py
import numpy as np

import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

embeddings = np.random.randn(len(dataset), 512)

fob.compute_similarity(
    dataset,
    embeddings=embeddings,
    model="clip-vit-base32-torch",  # hack so we can test text similarity
    brain_key="sim",
)

```


brimoor (approved):
Tested, LGTM

PR #3019, opened 2023-05-12
Add segment-anything models (SAM) to model zoo 

Rusteam:

Add three [segment-anything](https://segment-anything.com/) models to the model zoo.

Added an extra intensive model test for sam backbone.

        notes for FiftyOne users.

Segment-anything model is now available in the model zoo. It can perform automatic mask generation and guided mask generation with prompts.

-   [ ] Other

### DIFF
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -363,13 +363,9 @@ def _apply_image_model_data_loader(
 
     samples = samples.select_fields(extra_fields)
     # we need to make sure classes are available in the samples
-    if (
-        extra_fields
-        and model.field_type == "detections"
-        and not samples.classes
-    ):
+    if extra_fields:
         class_labels = samples.values(
-            f"{model.needs_fields[0]}.detections.label"
+            f"{model.needs_fields[0]}.{model.field_type}.label"
         )
         class_labels = list(
             set(one for labels in class_labels for one in labels)
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -234,6 +234,7 @@ def apply_model(
                 num_workers,
                 skip_failures,
                 filename_maker,
+                **kwargs,
             )
 
         if batch_size is not None:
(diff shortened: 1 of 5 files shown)


Rusteam:
@brimoor  Need your input on the following issue. I'm following this [tutorial](https://towardsdatascience.com/see-what-you-sam-4eea9ad9a5de) to add SAM to the model zoo, so far AMG mode worked fine. In order to implement prompting SAM with points and boxes I want to let users specify keypoint/detection label field while calling `dataset.apply_mode(points_from=<keypoint_label_field>, boxes_from=None)` and then pull up points and bounding boxes from a sample collection. However in order to achieve that, I have to parse points and boxes before the `model.predict_all()` method [here](https://github.com/voxel51/fiftyone/blob/963418d1d90bcf4c930dcb3c1ff4e2caeee72e7e/fiftyone/core/models.py#L368) and then pass them as parameters to that call. This should be backward-compatible but looks like a major change. What do you think is the best scenario to achieve that?

timmermansjoy:
also a idea and idk if it would bloat this PR. but if you would integrate groundingDINO first you would also add a parameter of "objects" or so and then just prompt the model for objects.

Rusteam:
Thanks for the suggestion @timmermansjoy. I am myself not familiar with grounding Dino and that's why I want to limit the scope to segment-anything for the moment. After it is integrated, we can think about adding grounding Dino.

brimoor:
@Rusteam hmm, I see what you mean. Perhaps there should be a way for a `Model` instance to declare that it needs the sample objects to be supplied along with the images during inference?

```py
prediction = model.predict(img, sample)
predictions = model.predict_all(imgs, samples)
```

One stylistic way to achieve this would be to implement a `SampleMixin` class (named similarly to `PromptMixin` and `EmbeddingsMixin`).

Note that this line:
https://github.com/voxel51/fiftyone/blob/963418d1d90bcf4c930dcb3c1ff4e2caeee72e7e/fiftyone/core/models.py#L312

will need to be excluded for models that inherit from `SampleMixin`, as they'll need access to all fields in principle (the `select_fields()` was added to minimize database read bandwidth, since, until now, inference never required anything beyond the image pixels).

Alternatively there could be a `SampleMixin.needs_fields` property that returns a list of sample fields that should be included:

```py
if isinstance(model, SampleMixin):
    samples = samples.select_fields(model.needs_fields)
```

Rusteam:
I've slightly changed the behavior of Keypoints from the @jacobmarks blogpost. In this implementation, each Keypoint object will be processed individually with all points passed as positive labels (and no background labels). Also a user can select which mask (0,1, or 2) to use. I guess this requires a short tutorial, which I can work on after we complete this integration.

Rusteam:
@brimoor hi, this is now ready for testing and integration.

Rusteam:
@brimoor hey any updates?

brimoor (approved):
@Rusteam LGTM (sorry for the delay! 😅)

I'm going to merge this into a feature branch and add a couple additional enhancements. Then we'll merge into `develop` and it will be included in the next release 📈

PR #3024, opened 2023-05-12
schema upgrades bug fixes batch 2/3

manivoxel51:

bug fixes for schema upgrades + UI polish

Screen Shot 2023-05-11 at 8 42 46 AM

Screen Shot 2023-05-11 at 8 42 53 AM

Screen Shot 2023-05-11 at 8 42 40 AM

(Details)

        notes for FiftyOne users.

-   [ ] Other


app/packages/core/src/components/Sidebar/Sidebar.tsx
```diff
@@ -22,6 +27,7 @@ import ViewSelection from "./ViewSelection";
 import { resizeHandle } from "./Sidebar.module.css";
 import SchemaSettings from "../Schema/SchemaSettings";
 import { graphql, useFragment } from "react-relay";
+import { selectedPathsState } from "@fiftyone/state/src/hooks/useSchemaSettings";
```
benjaminpkane, line 30: nit: src import

app/packages/state/src/hooks/useSchemaSettings.ts
```diff
@@ -35,14 +37,16 @@ const skipFiled = (path: string, ftype: string) => {
...
   return (
     path === "tags" ||
     path === "filepath" ||
     (ftype === OBJECT_ID_FIELD && path === "id") ||
     path.startsWith("metadata") ||
     path.endsWith("frames.frame_number") ||
-    ftype === FRAME_NUMBER_FIELD
+    ftype === FRAME_NUMBER_FIELD ||
+    path === "group" ||
+    path === "frames.group"
```
benjaminpkane, line 49: I don't think this actually covers the general case case of group datasets. I suppose dynamic groups (as views) complicates things more. Ideally there is a way to inspect a required group field in the schema cc @brimoor

app/packages/state/src/hooks/useSchemaSettings.ts
```diff
@@ -76,23 +80,31 @@ export const schemaSearchRestuls = atom<string[]>({
...
-            path &&
-            (viewSchema?.[path]?.ftype || schema?.[path]?.ftype) &&
-            !skipFiled(path, viewSchema[path].ftype)
-        );
+        const greenPaths = [...newPaths]
+          .filter(
+            (path) =>
+              path &&
+              (viewSchema?.[
+                path.startsWith("frames.") ? path.replace("frames.", "") : path
```
benjaminpkane, line 88: This could break image datasets with a `frames` field. There are probably existing issues with a `frames` field on image datasets but just sharing some context. Ok with me until a more formal discussion happens

benjaminpkane (changes requested):
I think group datasets with arbitrary field naming may need to be tested still.
* https://docs.voxel51.com/user_guide/groups.html#creating-grouped-datasets

`dataset.add_group_field()` can take any name for a group field

app/packages/state/src/hooks/useSchemaSettings.ts
manivoxel51: Got it fixed thanks to you! it is looking at the dataset.group_field now instead

app/packages/state/src/hooks/useSchemaSettings.ts
manivoxel51: do you mean custom defined `frames` field on image datasets? videos are the only ones with `frames` prefix I thought 🤔

benjaminpkane: I always developed the App to in theory support adding an arbitrary `frames` field, e.g.
```py
dataset.add_sample_field("frames", fo.IntField)
```

But it could very likely not be supported in the SDK / in general. It's just not officially documented. I digress

brimoor: The SDK is careful to allow `frames` fields for image datasets. The `frames` keyword only has special meaning for datasets that contain videos.

app/packages/state/src/hooks/useSchemaSettings.ts
```diff
@@ -30,19 +33,22 @@ const skipFiled = (path: string, ftype: string) => {
   return (
     ftype === DICT_FIELD ||
```
benjaminpkane, line 34: [`fieldsReducer`](https://github.com/voxel51/fiftyone/blob/adc271617adb47e062b41736610d222513126a94/app/packages/state/src/recoil/sidebar.ts#L147) and/or [`UNSUPPORTED_FILTER_TYPES`](https://github.com/voxel51/fiftyone/blob/adc271617adb47e062b41736610d222513126a94/app/packages/utilities/src/index.ts#L318) could be leveraged here

app/packages/state/src/hooks/useSchemaSettings.ts
```diff
@@ -30,19 +33,22 @@ const skipFiled = (path: string, ftype: string) => {
...
     ftype === JUST_FIELD ||
+    ftype === VECTOR_FIELD ||
     path.includes(".logits") ||
```
benjaminpkane, line 37: `logits` is a vector field so it shouldn't be necessary

app/packages/state/src/hooks/useSchemaSettings.ts
```diff
@@ -30,19 +33,22 @@ const skipFiled = (path: string, ftype: string) => {
...
     path.endsWith(".index") ||
     path.endsWith(".bounding_box")
```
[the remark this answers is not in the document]
manivoxel51: to remedy, I added a check to the current's path's parent's embedded document. We skip if it is Detection type and path ends with`.bounding_box`.

app/packages/state/src/hooks/useSchemaSettings.ts
```diff
@@ -26,23 +33,28 @@ export const TAB_OPTIONS_MAP = {
...
     path === "tags" ||
     path === "filepath" ||
     (ftype === OBJECT_ID_FIELD && path === "id") ||
+    ftype === FRAME_NUMBER_FIELD ||
+    groupField === path ||
+    path === "sample_id" ||
     path.startsWith("metadata") ||
-    path.endsWith("frames.frame_number") ||
-    ftype === FRAME_NUMBER_FIELD
+    path.endsWith("frames.frame_number")
```
benjaminpkane, line 57: Using the [`FrameNumberField`](https://github.com/voxel51/fiftyone/blob/c9f37ad3f8ea52bb69d80e19e466d7cc9591dfb7/fiftyone/core/fields.py#L1033) might be cleaner here

benjaminpkane (approved):
Code LGTM! 👍

app/packages/state/src/hooks/useSchemaSettings.ts
manivoxel51: good point. I'll double check in quick followup PR

PR #3028, opened 2023-05-12
Add ArrayField and VectorField to unsupported OTHER sidebar group

benjaminpkane:
```py
import fiftyone.zoo as foz

dataset.add_sample_field("vector", fo.VectorField)
dataset.add_sample_field("array", fo.ArrayField)
```

Screen Shot 2023-05-12 at 2 44 07 PM

### DIFF
--- a/app/packages/utilities/src/index.ts
+++ b/app/packages/utilities/src/index.ts
@@ -3,11 +3,11 @@ import _ from "lodash";
 import mime from "mime";
 import { isElectron } from "./electron";
 
+export * from "./Resource";
 export * from "./color";
 export * from "./electron";
 export * from "./errors";
 export * from "./fetch";
-export * from "./Resource";
 export * from "./styles";
 
 interface O {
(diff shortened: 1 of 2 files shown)


app/packages/utilities/src/index.ts
```diff
@@ -312,6 +314,15 @@ export const VALID_NUMERIC_TYPES = [
   INT_FIELD,
 ];
 
+// list fields may not have a subfield type, so null, undefined is included
+export const UNSUPPORTED_FILTER_TYPES = [
+  ARRAY_FIELD,
```
manivoxel51, line 319: Just making sure that array ARRAY_FIELD here does not exclude any custom array of primitive strings from being rendered as a regular field (as opposed to a field under `Other` group )

manivoxel51 (approved):
🍨

app/packages/utilities/src/index.ts
benjaminpkane: Yes, `ArrayField` is binary data. Distinct from `ListField`

PR #3038, opened 2023-05-15
Plugin updates

brimoor:
## Change log

Main updates
- Added `fiftyone plugins create` command to CLI
- Show plugin enabled status in `fiftyone plugins list`
- Adds optional `outdir` parameter to `creeate_plugin()` to create plugins in custom directories, if desired
- Adds a `plugin_names` parameter to `download_plugin()` to download specific plugins by name
- Entire interface now allows for `fiftyone.yaml` or `fiftyone.yml`
- Move `GitHubRepository` to `fiftyone.utils.github` as it is a general purpose utility
- Exposed all relevant function parameters in CLI
- Added CLI documentation for all new methods

Other things I encountered while making above changes:
- Fixed bugs
- Remove deprecated/unused/duplicate methods
- General implementation cleanup

## Example usages

```shell
fiftyone plugins list

# Create a plugin
fiftyone plugins create test
fiftyone plugins list

# Disable a plugin
fiftyone plugins disable test
fiftyone app config
fiftyone plugins list

# Enable a plugin
fiftyone plugins enable test
fiftyone app config
fiftyone plugins list

# Delete a plugin
fiftyone plugins delete test
fiftyone app config
fiftyone plugins list

# Download a plugin from a private repo
export GITHUB_TOKEN=XXXXXXXX
fiftyone plugins download https://github.com/voxel51/fiftyone-gpt
fiftyone plugins list

# Download specific plugins from a GitHub repository
fiftyone plugins download \
    -n '@voxel51/examples'
fiftyone plugins list
```

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -316,14 +316,16 @@ def find_files(root_dir, patt, max_depth=1):
 
     Exammples::
 
+        import fiftyone.core.utils as fou
+
         # Find .txt files in `/tmp`
-        find_files("/tmp", "*.txt")
+        fou.find_files("/tmp", "*.txt")
 
         # Find .txt files in subdirectories of `/tmp` that begin with `foo-`
-        find_files("/tmp/foo-*", "*.txt")
+        fou.find_files("/tmp/foo-*", "*.txt")
 
         # Find .txt files in `/tmp` or its subdirectories
-        find_files("/tmp", "*.txt", max_depth=2)
+        fou.find_files("/tmp", "*.txt", max_depth=2)
 
     Args:
         root_dir: the root directory
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -11,6 +11,7 @@ from collections import defaultdict
 from contextlib import contextmanager
 from copy import deepcopy
 from datetime import date, datetime
+import glob
 import hashlib
 import importlib
 import inspect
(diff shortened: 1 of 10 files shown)


fiftyone/plugins/core.py
```diff
@@ -49,498 +48,473 @@ def __repr__(self):
...
+            if existing_dir is not None:
+                if not overwrite:
+                    logger.debug(f"Skipping existing plugin '{plugin.name}'")
+                    continue
 
+                logger.debug(f"Overwriting existing plugin '{plugin.name}'")
+                etau.delete_dir(existing_dir)
+                plugin_dir = existing_dir
+            else:
+                plugin_dir = _recommend_plugin_dir(plugin)
```
kaixi-wang, line 242: Can we keep it as before and use the url as the directory name? This helps separate locally developed and downloaded under most cases, enables checking if the archive needed to be redownloaded, and also keeps track of download source

fiftyone/plugins/__init__.py
```diff
@@ -1,45 +1,28 @@
...
     list_enabled_plugins,
     find_plugin,
     download_plugin,
     create_plugin,
 )
-from .definitions import list_plugins
-
-
-@deprecated.deprecated(reason="Use list_plugins() instead")
-def list_all_plugins():
```
kaixi-wang, line 25: I included this because @ritch had a plugin that used it. Also ritchie mentioned that there are some users that have already been developing plugins, and just in case they used this method, I didn't want their plugin to break...

fiftyone/plugins/__init__.py
```diff
@@ -1,45 +1,28 @@
...
-
 
 # This enables Sphinx refs to directly use paths imported here
-# and ignore functions marked as depreciated
 __all__ = [
     k
     for k, v in globals().items()
-    if not k.startswith("_")
-    and not (type(v) in [wrapt.decorators.AdapterWrapper, types.ModuleType])
+    if not k.startswith("_") and not isinstance(v, types.ModuleType)
```
kaixi-wang, line 27: I believe I had an issue with Sphinx including imported modules when using `isinstance` so I used `type`

ritch (approved):

fiftyone/core/utils.py
```diff
@@ -1488,6 +1529,23 @@ def __exit__(self, *args):
         logging.disable(logging.NOTSET)
 
 
+class add_sys_path(object):
```
ritch, line 1532: Where is this used? Also why is this class using `this_casing_style`?

brimoor: Something like this will need to be used by plugins that aren't Python packages (like GPT) until we find a better solution for local module imports.

fiftyone/plugins/__init__.py
brimoor: This same code block is inserted elsewhere in the repo, so I'm keeping it the same for consistency. 

FWIW the docs build fine for me.

fiftyone/plugins/core.py
brimoor: It's complicated because if the user redownloads and `overwrite=True`, the current logic will update the existing directory in-place. If we keep the download hash in the folder tree in `PLUGINS_DIR`, that will be confusing because the new version may not be from that hash.

brimoor: An alternative approach would be to inject some metadata within the downloaded directory that notes when the download happened, where it came from, etc.

PR #3041, opened 2023-05-15
Plugin py docs fixups

ritch:


fiftyone/operators/__init__.py
```diff
@@ -4,6 +4,12 @@
 |
 """
 
+import types
+import wrapt
```
imanjra, line 8: Is this used?

fiftyone/operators/__init__.py
```diff
@@ -4,6 +4,12 @@
...
 from .operator import Operator, OperatorConfig
 from .registry import register_operator
 from .executor import execute_operator
+
+# This enables Sphinx refs to directly use paths imported here
+# and ignore functions marked as depreciated
+__all__ = [k for k, v in globals().items() if not k.startswith("_")]
```
ritch, line 15: Not actually certain this is needed. Let me try again without it.

imanjra (approved):

fiftyone/operators/__init__.py
ritch: image

This is without.

ritch: With:

image

fiftyone/operators/__init__.py
ritch: Removed

PR #3050, opened 2023-05-16
Fix plugin route

ritch:
Fixes recent refactor regressed the contract for `GET /plugins` which should return `{plugins: []}`.

Note: this is not an issue in other form factors (teams).


brimoor (approved):
LGTM

PR #3055, opened 2023-05-16
Dynamic groups action cleanup

benjaminpkane:
The current rest route causes unwanted blocking on the server. I've moved to backend calls to GraphQL, and moved aggregation calls to occur only once the action popout is visible.

I'm not sure if we should be using `count` as a filter for the group fields at all. Filtering by count for all candidate fields will be slow on large datasets and will prevent usage.


sashankaryal (approved):
I added a commit to make closing behavior more consistent with other pill buttons, but other than that, LGTM, thanks for the refactor!

PR #3056, opened 2023-05-16
[HOLD] Raise error for no dataset name

kaixi-wang:

Raise an error for attempts to load a Dataset without providing a name to prevent any further operations that may lead to upserting a dataset document with no name.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -257,8 +257,11 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         _virtual=False,
         **kwargs,
     ):
-        if name is None and _create:
-            name = get_default_dataset_name()
+        if name is None:
+            if _create:
+                name = get_default_dataset_name()
+            else:
+                raise ValueError("Dataset `name` is required")
 
         if overwrite and dataset_exists(name):
             delete_dataset(name)


brimoor (changes requested):
I've seen you add this before but don't understand why, because this already raises an informative error:

```py
import fiftyone as fo

dataset = fo.load_dataset(None)
# ValueError: Dataset 'None' not found
```

brimoor:
Also, in the unfortunate scenario where a user winds up with a dataset named `None`, it would be nice for `fo.load_dataset(None)` to work.

kaixi-wang:
> I've seen you suggest this before but don't understand why, because this already raises an informative error:
> 
> ```python
> import fiftyone as fo
> 
> dataset = fo.load_dataset(None)
> # ValueError: Dataset 'None' not found
> 
> fo.Dataset(None, _create=False)
> # ValueError: Dataset 'None' not found
> ```

Since all datasets need names, Dataset with name =n `None` is in an error state. We would not/should not allow the dataset to continue to "work"

brimoor:
If a `None`-named dataset does happen somehow, this PR would prevent the user from loading it. I think it is better to let them load it so they can fix the problem with `dataset.name = "valid-name"`.

PR #3057, opened 2023-05-16
Plugin loader error handling tweaks

kaixi-wang:

Prevent attempting to registering and loading a plugin that has errors
image

(Details)

-   [ ] Other

### DIFF
--- a/fiftyone/operators/loader.py
+++ b/fiftyone/operators/loader.py
@@ -90,15 +90,21 @@ def register_module(plugin_definition, mod):
     Returns:
         the :class:`PluginContext` for the plugin
     """
+    pctx = None
     try:
         pctx = PluginContext(plugin_definition)
-        if pctx.errors:
-            # Don't register if there are errors
-            return pctx
         mod.register(pctx)
-        KNOWN_PLUGIN_CONTEXTS[pctx.name] = pctx
     except Exception as e:
-        pctx.errors.append(traceback.format_exc())
+        if pctx:
+            pctx.errors.append(traceback.format_exc())
+    finally:
+        if pctx and pctx.name:
+            KNOWN_PLUGIN_CONTEXTS[pctx.name] = pctx
+        else:
+            # If plugin context is None, then there is no plugin name to
+            # register so just silently fail
+            pass
+
     return pctx
 
 
@@ -137,7 +143,6 @@ def load_from_dir():
             try:
                 pctx = exec_module_from_dir(module_dir, plugin_definition)
                 plugin_contexts.append(pctx)
-                KNOWN_PLUGIN_CONTEXTS[plugin_definition.name] = pctx
             except ValueError as e:
                 print("Error loading plugin from %s" % module_dir)
                 pass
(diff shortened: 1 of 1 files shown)


fiftyone/operators/loader.py
```diff
@@ -89,11 +92,13 @@ def register_module(plugin_definition, mod):
     """
     try:
         pctx = PluginContext(plugin_definition)
-        KNOWN_PLUGIN_CONTEXTS[pctx.name] = pctx
+        if pctx.errors:
+            # Don't register if there are errors
+            return pctx
         mod.register(pctx)
```
ritch, line 98: This should be called even with errors

fiftyone/operators/loader.py
```diff
@@ -89,11 +92,13 @@ def register_module(plugin_definition, mod):
     """
     try:
         pctx = PluginContext(plugin_definition)
-        KNOWN_PLUGIN_CONTEXTS[pctx.name] = pctx
+        if pctx.errors:
```
ritch, line 95: This isn't needed because we actually want to register the errors.

fiftyone/operators/loader.py
```diff
@@ -157,6 +162,15 @@ def exec_module_from_dir(module_dir, plugin_definition):
         )
     spec = importlib.util.spec_from_file_location(mod_dir, mod_filepath)
     mod = importlib.util.module_from_spec(spec)
+    try:
+        pctx = register_module(plugin_definition, mod)
```
ritch, line 166: We don't want to call `register_module()` twice.

fiftyone/operators/loader.py
```diff
@@ -157,6 +162,15 @@ def exec_module_from_dir(module_dir, plugin_definition):
...
+    except Exception as e:
+        raise e
+    if pctx.has_errors():
+        raise ValueError(
+            f"Plugin {plugin_definition.name} has errors: {pctx.errors}"
+        )
+    # only add to sys.modules if there are no errors
```
ritch, line 173: This seems like the only changed needed - to only add to sys.modules if there are not errors, although I'm not sure you can attempt to load `exec_module` without it added, I've not tried that.

fiftyone/operators/loader.py
kaixi-wang: updated and put in `finally` so whether or not an error is raised, it gets added

fiftyone/operators/loader.py
kaixi-wang: yeah I forgot to remove the previous call. Fixed now

fiftyone/operators/loader.py
```diff
@@ -87,13 +90,21 @@ def register_module(plugin_definition, mod):
...
     except Exception as e:
-        errors = [traceback.format_exc()]
-        pctx.errors = errors
+        if pctx:
+            pctx.errors.append(traceback.format_exc())
+    finally:
+        if pctx and pctx.name:
+            KNOWN_PLUGIN_CONTEXTS[pctx.name] = pctx
+        else:
+            # If plugin context is None, then there is no plugin name to
```
ritch, line 104: We should print the plugin_definition.directory

kaixi-wang: added logging

PR #3060, opened 2023-05-17
Plugins support symlinks

ritch:
This is needed to fully support symlinks for plugin directories. Note: this also requires min version of 0.24.0 of starlette.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ INSTALL_REQUIRES = [
     "setuptools",
     "sseclient-py>=1.7.2,<2",
     "sse-starlette>=0.10.3,<1",
-    "starlette>=0.20.4,<0.27",
+    "starlette>=0.24.0,<0.27",
     "strawberry-graphql==0.138.1",
     "tabulate",
     "xmltodict",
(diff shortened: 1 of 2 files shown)


imanjra (approved):

setup.py
```diff
@@ -67,7 +67,7 @@ def get_version():
     "setuptools",
     "sseclient-py>=1.7.2,<2",
     "sse-starlette>=0.10.3,<1",
-    "starlette>=0.20.4,<0.27",
+    "starlette>=0.24.0,<0.27",
```
brimoor, line 70: @benjaminpkane is this okay?

brimoor (commented):
Symlink support is 💯

setup.py
brimoor: I guess its just a tightening, so no downstream consequences huh?

brimoor (approved):

setup.py
benjaminpkane: Yes, we recently had the request to increase upper bound. We may have users dependent on the lower bound but it's hard to say. I'm ok with it

PR #3062, opened 2023-05-17
Welcome message!

brimoor:
Logs a welcome message the first time the App is launched on a machine with a new FiftyOne version

Screen Shot 2023-05-22 at 10 40 51 AM


fiftyone/core/uid.py
```diff
@@ -99,3 +114,35 @@ def send_import_event():
...
+    """
+    if not _allow_logging():
+        return
+
+    try:
+        last_version = etas.load_json(foc.WELCOME_PATH)["version"]
+    except:
+        last_version = None
+
+    if foc.VERSION == last_version:
```
swheaton, line 131: More a product decision than code but in case you hadn't considered it...
You could consider printing this every time except if a config is set. Sanic prints below every time it starts, for example, unless you have a setting telling it not to.

The tradeoff being annoyance level vs. repetition of call to action I suppose.

```
[2023-05-17 08:59:57 -0400] [60595] [INFO] 
  ┌────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │                                             Sanic v22.12.0                                             │
  │                                    Goin' Fast @ http://0.0.0.0:8000                                    │
  ├───────────────────────┬────────────────────────────────────────────────────────────────────────────────┤
  │                       │        mode: debug, single worker                                              │
  │     ▄███ █████ ██     │      server: sanic, HTTP/1.1                                                   │
  │    ██                 │      python: 3.10.11                                                           │
  │     ▀███████ ███▄     │    platform: macOS-12.3-arm64-arm-64bit                                        │
  │                 ██    │ auto-reload: enabled                                                           │
  │    ████ ████████▀     │    packages: sanic-routing==22.8.0, sanic-testing==22.12.0, sanic-ext==22.12.0 │
  │                       │                                                                                │
  │ Build Fast. Run Fast. │                                                                                │
  └───────────────────────┴────────────────────────────────────────────────────────────────────────────────┘
```
Lake FS prints this every time, with no way to tell it not to, as far as I can tell with quick searching....
```
lakeFS 0.93.0 - Up and running (^C to shutdown)...


     ██╗      █████╗ ██╗  ██╗███████╗███████╗███████╗
     ██║     ██╔══██╗██║ ██╔╝██╔════╝██╔════╝██╔════╝
     ██║     ███████║█████╔╝ █████╗  █████╗  ███████╗
     ██║     ██╔══██║██╔═██╗ ██╔══╝  ██╔══╝  ╚════██║
     ███████╗██║  ██║██║  ██╗███████╗██║     ███████║
     ╚══════╝╚═╝  ╚═╝╚═╝  ╚═╝╚══════╝╚═╝     ╚══════╝

│
│ If you're running lakeFS locally for the first time,
│     complete the setup process at http://127.0.0.1:8000/setup
│

│
│ For more information on how to use lakeFS,
│     check out the docs at https://docs.lakefs.io/quickstart/repository
│

│
│ For support or any other question,
│     join our Slack channel https://docs.lakefs.io/slack
│
```

findtopher:
This all gets disabled in the Teams `FIFTYONE_INTERNAL_SERVICE` context, right?  I'd really like to avoid spamming the server logs with this if we can.  (It won't be seen by folks who will heed the call to action and it could make tracking down real problems an issue...)

brimoor:
@findtopher services will likely not launch the App, but yes good call, I will make sure `FIFTYONE_INTERNAL_SERVICE` is explicitly checked in Teams, too.

fiftyone/core/uid.py
brimoor: Sweet thanks for the prior art 👍

fiftyone/core/uid.py
```diff
@@ -14,14 +15,37 @@
...
+██║     ██║██║        ██║      ██║   ╚██████╔╝██║ ╚████║███████╗
+╚═╝     ╚═╝╚═╝        ╚═╝      ╚═╝    ╚═════╝ ╚═╝  ╚═══╝╚══════╝ v{0}
+
+If you find FiftyOne helpful for your work, here's how you can get involved:
+
+    ⭐⭐⭐ Give the project a star on GitHub ⭐⭐⭐
+    https://github.com/voxel51/fiftyone
+
+    🚀🚀🚀 Join the FiftyOne Slack community 🚀🚀🚀
+    https://join.slack.com/t/fiftyone-users/shared_invite/zt-s6936w7b-2R5eVPJoUw008wP7miJmPQ
```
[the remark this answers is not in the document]
brimoor: yep, done ✔️

PR #3067, opened 2023-05-17
Remove redundant dataset serialization in server events

benjaminpkane:
Dataset serialization is already handled by state description serialization in the[ `dict_factory`'s](https://github.com/voxel51/fiftyone/blob/4bc4cf46bd3ccbe435cb8de3305573e8bf2a8d5c/fiftyone/core/session/events.py#L134) call to `StateDescription.serialize()`


benjaminpkane:
Stale

PR #3068, opened 2023-05-17
Documentation of management SDK

swheaton:
**!!NEW!!** now with 100% more plugins!

Contains a new doc under FiftyOne Teams:`management_sdk.rst` 

This contains some preface text but then is mostly autodoc components of the various sections of management SDK (connection, api keys, dataset perms, plugins, users)

The reason I made this manually instead of using `sphinx-apidoc` is I needed more fine grained inclusion/exclusion capabilities than I could find with just using that tool, plus it was not really possible to keep it separate from the mainline API documentation.

How to build documentation with management SDK properly built:

- Build from `fiftyone-teams`
OR
- Build from `fiftyone` but pass `fiftyone-teams` repo path to generate_docs script which links and unlinks necessary dirs
```
bash generate_docs.bash -t <PATH_TO_FIFTYONE_TEAMS_REPO>
```

Before merging:
- [x] change api key/uri setup configuration text to link to what @j053y is cooking up.


swheaton:
Example screenshots of what the page looks like, so you can see samples without having to build docs yourself.

image
image
image

swheaton:
This PR now also includes `plugins.rst` documentation!

docs/source/teams/management_sdk.rst
```diff
@@ -0,0 +1,78 @@
...
+| `api_uri`                     | `FIFTYONE_API_URI`                  | `None`                        | The URI where the FiftyOne Teams API (`teams-api` container) is exposed. Check         |
+|                               |                                     |                               | with your Teams admin for the value of this field for your deployment.                 |
++-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
+| `api_key`                     | `FIFTYONE_API_KEY`                  | `None`                        | The FiftyOne Teams API key to use for authentication with the API. This key is         |
+|                               |                                     |                               | unique to each user. See <blah> for information on creating an API key.                |
++-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
+
+
+.. _teams-sdk-api-reference:
```
brimoor, line 34: So unfortunately the docs-build CI workflow isn't currently going to have access to a `fiftyone-teams` instance to populate these with.

As a temporary workaround, how about splitting the Teams SDK reference build into a separate script that outputs the `.rst` to stdout/tmp file:

```shell
bash docs/scripts/generate_teams_docs.bash
```

Then, for now, whenever we want to update this section, we run that script and commit the output here?

brimoor (changes requested):
@swheaton looking good, thanks for adding! 🙇 

LMK what you think of my suggestion re. teams API reference docs. Happy to chat offline if needed.

docs/source/teams/management_sdk.rst
swheaton: Yeah I feared as much.

The problem with that is that the `.. automodule::` is evaluated at build time so what's here _is_ the `.rst`. We would have to commit the HTML file `build/...management_sdk.html` to this repo

swheaton: Will take offline

brimoor (approved):
LGTM

PR #3072, opened 2023-05-17
Markdown view

ritch:


app/packages/core/src/plugins/SchemaIO/components/MarkdownView.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import { Typography, Box, Link } from "@mui/material";
+import ReactMarkdown from "react-markdown";
+import CodeView from "./CodeView";
+import { d } from "vitest/dist/index-5aad25c1";
```
imanjra, line 5: Is this import intentional?

app/packages/core/src/plugins/SchemaIO/components/MarkdownView.tsx
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export default function MarkdownView(props) {
+  const { schema, data } = props;
+  const { view = {} } = schema;
+  const { label, description } = view;
+
+  return (
+    <Box>
+      <h3>{label}</h3>
```
imanjra, line 39: Can use `HeaderView` here:

```js
<Box>
  <HeaderView {...props} />
 ...
 
```

imanjra (approved):
Is `.chroma` directory include intentional?

ritch:
> @ritch can you purge `.chroma` from the PR? Would be nice to avoid the extra MBs of data sticking around in clones forever.

I amended the commit without the file so I think it should not be in the history.

PR #3080, opened 2023-05-18
Markdown view support

ritch:


app/packages/core/src/plugins/SchemaIO/components/MarkdownView.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { HeaderView } from ".";
+import styled from "styled-components";
+
+const InlineCode = styled.span`
+  background: ${({ theme }) => theme.background.level1};
+  color: ${({ theme }) => theme.voxel[500]};
+  border-radius: 3px;
+  padding: 0.2em 0.4em;
+  font-size: 85%;
+  font-family: Monaco, Consolas, "Andale Mono", "DejaVu Sans Mono", monospace;
```
imanjra, line 14: I believe we have used `Roboto Mono, monospace` for `CodeBlock`. Should we use same here?

imanjra (approved):

PR #3089, opened 2023-05-20
Schema field visibility bug fixes for datasets with frames and more

manivoxel51:

- [x] Fixes when you enable show-nested-fields toggle and (un)check top level fields. before it was not unselecting - now is
- [x] Fixes when you try to re-select a `frames.ANYTHING` field for datasets that have frames - before clicking on it would not select it
- [x] Fixes search for frame fields that result in field path suffixed with `._cls` for some reason
- [x] embeddedDocument types that are top level and have dynamic embedDocType are disappearing when unselected and applied.  
- [x] fix search for frames

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/state/src/hooks/useSchemaSettings.ts
+++ b/app/packages/state/src/hooks/useSchemaSettings.ts
@@ -32,7 +32,7 @@ import {
   JUST_FIELD,
   OBJECT_ID_FIELD,
 } from "@fiftyone/utilities";
-import { keyBy } from "lodash";
+import { isEmpty, keyBy } from "lodash";
 import { useCallback, useContext, useEffect, useMemo } from "react";
 import { useMutation, useRefetchableFragment } from "react-relay";
 import {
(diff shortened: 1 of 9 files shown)


app/packages/state/src/hooks/useSchemaSettings.ts
```diff
@@ -252,29 +263,31 @@ export const excludedPathsState = atomFamily({
...
+                    [EMBEDDED_DOCUMENT_FIELD, LIST_FIELD].includes(
+                      combinedSchema[path]?.ftype
+                    ) &&
                     [
-                      EMBEDDED_DOCUMENT_FIELD,
                       DYNAMIC_EMBEDDED_DOCUMENT_FIELD,
-                      LIST_FIELD,
-                    ].includes(combinedSchema[path]?.subfield) &&
-                    isVideo
+                      DYNAMIC_EMBEDDED_DOCUMENT_FIELD_V2,
```
lanzhenw, line 279:
```
 const DYNAMIC_EMBEDDED_DOCUMENT_FIELD =
"fiftyone.core.fields.DynamicEmbeddedDocumentField";
const DYNAMIC_EMBEDDED_DOCUMENT_FIELD_V2 =
"fiftyone.core.odm.embedded_document.DynamicEmbeddedDocument";
```
Wonder what's the difference between the two?

lanzhenw (approved):
Looking good in general. 
niw: both color scheme and scheme selection used the conversion of stripping `frames.` when necessary from the path in many occasions. Shall we add a util function to simplify it? Also would make it easier to search where we converted the path for videos.

PR #3090, opened 2023-05-22
bugfix: color scheme instance mask, refresh with dataset, color picker UI and etc

lanzhenw:
This bugfix fixes the following bugs: 

- [x] On teams and in the UI toggling between different tabs, there was a regression about applying dataset default setting. 
- [x]  instance mask should update with bounding boxes as settings changed
- [x] embeddedDoc filter item color should update
- [x] When user changes datasets in the App, the session.color_scheme needs to update
- [x] JSON editor: colorByAttribute should not be null when using default "label" setting
- [x] Line spacing
- [x] Color picker: simple picker values should sync; fancy color picker had slide and drag issue with selection. 

Instance Mask:
https://github.com/voxel51/fiftyone/assets/17770824/0b3bdad1-802b-40a2-8fc5-7a03b2a9e494

Changing dataset on UI updates session.color_scheme:
https://github.com/voxel51/fiftyone/assets/17770824/852fbfc1-f623-4e25-8454-93307bad9e0e

Chrome color picker:
https://github.com/voxel51/fiftyone/assets/17770824/abe6437d-07f3-437d-ab6c-05199c70bcc0

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/ColorModal/FieldSetting.tsx
+++ b/app/packages/core/src/components/ColorModal/FieldSetting.tsx
@@ -18,6 +18,7 @@ import { TwitterPicker } from "react-color";
 import { useRecoilValue } from "recoil";
 import Checkbox from "../Common/Checkbox";
 import Input from "../Common/Input";
+import { resetColor } from "./ColorFooter";
 import {
   FieldCHILD_STYLE,
   FieldColorSquare,
@@ -28,7 +29,6 @@ import AttributeColorSetting from "./colorPalette/AttributeColorSetting";
 import { colorPicker } from "./colorPalette/Colorpicker.module.css";
 import ColorAttribute from "./controls/ColorAttribute";
 import ModeControl from "./controls/ModeControl";
-import { resetColor } from "./ColorFooter";
 
 type Prop = {
   prop: { field: Field; expandedPath: string };
@@ -40,7 +40,8 @@ type State = {
 };
 
 const FieldSetting: React.FC<Prop> = ({ prop }) => {
-  const colorContainer: React.RefObject<HTMLDivElement> = React.createRef();
+  const wrapperRef: React.RefObject<HTMLDivElement> = React.createRef();
+  const pickerRef: React.RefObject<TwitterPicker> = React.createRef();
   const { field, expandedPath } = prop;
   const path = field?.path;
   const { colorPool, fields } = useRecoilValue(fos.sessionColorScheme);
@@ -51,6 +52,7 @@ const FieldSetting: React.FC<Prop> = ({ prop }) => {
 
   const [showFieldPicker, setShowFieldPicker] = useState(false);
   const [input, setInput] = useState(color);
+  const [colors, setColors] = useState(colorPool);
   const [state, setState] = useState<State>({
     useLabelColors: Boolean(
       setting?.valueColors && setting.valueColors.length > 0
@@ -94,9 +96,12 @@ const FieldSetting: React.FC<Prop> = ({ prop }) => {
   const onValidateColor = useCallback(
     (input) => {
       if (isValidColor(input)) {
-        const hexColor = colorString.to.hex(colorString.get(input).value);
+        const hexColor = colorString.to.hex(
+          colorString.get(input)?.value ?? []
+        );
         onChangeFieldColor(hexColor);
         setInput(hexColor);
+        setColors([...new Set([...colors, hexColor])]);
       } else {
         // revert input to previous value
         setInput("invalid");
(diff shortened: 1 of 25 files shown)


manivoxel51 (commented):
🍨 

question: Is there a way to refactor all the outside click logic into its own hook or re-use and existing one?

app/packages/core/src/components/ColorModal/FieldSetting.tsx
```diff
@@ -40,7 +40,8 @@ type State = {
 };
 
 const FieldSetting: React.FC<Prop> = ({ prop }) => {
-  const colorContainer: React.RefObject<HTMLDivElement> = React.createRef();
+  const wrapperRef: React.RefObject<HTMLDivElement> = React.createRef();
+  const pickerRef: React.RefObject<TwitterPicker> = React.createRef();
```
sashankaryal, line 44: Why not `useRef()`? Is it intentional that you want these refs to not be memoized between renders?

app/packages/core/src/components/ColorModal/FieldSetting.tsx
```diff
@@ -155,6 +160,21 @@ const FieldSetting: React.FC<Prop> = ({ prop }) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [useRecoilValue(resetColor)]);
 
+  useEffect(() => {
+    function handleClickOutside(event) {
+      if (wrapperRef.current && !wrapperRef.current.contains(event.target)) {
+        setShowFieldPicker(false);
+      }
+    }
```
sashankaryal, line 168: Could you use `useOutsideClick()` hook here?

app/packages/core/src/components/ColorModal/colorPalette/ColorPalette.tsx
```diff
@@ -74,10 +74,30 @@ const ColorPalette: React.FC<ColorPaletteProps> = ({
...
+      }
+    }
+
+    // Bind the event listener
+    document.addEventListener("mousedown", handleClickOutside);
+    return () => {
+      // Unbind the event listener on clean up
+      document.removeEventListener("mousedown", handleClickOutside);
+    };
+  }, [wrapperRef]);
```
sashankaryal, line 91: Same comment as above, why not `useOutsideClick()`?

app/packages/looker/src/lookers/shared.ts
```diff
@@ -21,10 +23,17 @@ export const LookerUtils = {
     next: Readonly<Optional<BaseState["options"]>>
   ): boolean => {
     let reloadSample = false;
+
     if (next.coloring && current.coloring.seed !== next.coloring.seed) {
       reloadSample = true;
     } else if (next.coloring && next.coloring.by !== current.coloring.by) {
       reloadSample = true;
+    } else if (!_.isEmpty(_.xor(next.coloring.pool, current.coloring.pool))) {
```
sashankaryal, line 31: cool

app/packages/looker/src/util.ts
```diff
@@ -547,3 +548,31 @@ export const getMimeType = (sample: any) => {
...
+function compareObjectArr(arr1: object[], arr2: object[]): boolean {
+  const sortedArr1 = arr1
+    .map((el) => JSON.stringify(el, Object.keys(el).sort()))
+    .sort();
+  const sortedArr2 = arr2
+    .map((el) => JSON.stringify(el, Object.keys(el).sort()))
+    .sort();
+
+  return JSON.stringify(sortedArr1) === JSON.stringify(sortedArr2);
+}
```
sashankaryal, line 578: nit: use lodash? something like...
```
function compareObjectArr(arr1: object[], arr2: object[]): boolean {
  const sortedArr1 = sortBy(arr1, (el) => JSON.stringify(el, Object.keys(el).sort()));
  const sortedArr2 = sortBy(arr2, (el) => JSON.stringify(el, Object.keys(el).sort()));

  return isEqual(sortedArr1, sortedArr2);
}
```

app/packages/state/src/recoil/atoms.ts
```diff
@@ -410,3 +410,8 @@ export const sessionColorScheme = atom<ColorSchemeSetting>({
     fields: [],
   },
 });
+
+export const isUsingSessionColorScheme = atom<boolean>({
+  key: "isUsingSessionColorScheme",
+  default: false,
```
sashankaryal, line 416: Is this something we'd want to persist in browser storage?

sashankaryal (changes requested):
Some minor comments, but looks good overall!

Requesting changes only because I'm not sure if I see color options for different targets for segmentations the way it's shown in the video. Will clarify offline.

sashankaryal:
> 🍨
> 
> question: Is there a way to refactor all the outside click logic into its own hook or re-use and existing one?

We already have a hook for that, it's called `useOutsideClick()`.

app/packages/state/src/recoil/atoms.ts
lanzhenw: No, this is an indicator to determine whether to apply `dataset.appConfig.colorScheme` in loading. (when isUsingSessionColorScheme is false, apply dataset setting). It's set when we load the session state in `useStateUpdate`.

app/packages/core/src/components/ColorModal/FieldSetting.tsx
lanzhenw: changing to useRef()

sashankaryal (approved):
LGTM! 🥳

PR #3091, opened 2023-05-22
Fixing color scheme bugs

brimoor:
Change log
- `launch_app()` with no `dataset` was previously raising an error due to color scheme parsing, but should be allowed
- `session.color_scheme` should always be presented as a `ColorScheme` object. Previously it was presented as a string when updates were made in the App
- By convention, `session.color_scheme = None` should be allowed as shorthand for reverting to the default color scheme
- `session.color_scheme` was not updating when `session.dataset` is modified

```py
import fiftyone as fo
import fiftyone.zoo as foz

session = fo.launch_app()
print(session.color_scheme)  # default

dataset.app_config.color_scheme = fo.ColorScheme(
    color_pool=["#ff0000", "#00ff00", "#0000ff", "pink", "yellowgreen"],
    fields=[
        {
            "path": "ground_truth",
            "fieldColor": "#ff00ff",
            "colorByAttribute": "label",
            "valueColors": [{"value": "dog", "color": "yellow"}],
        }
    ]
)
dataset.save()

session.dataset = dataset
print(session.color_scheme)  # now is custom scheme above

session.dataset = fo.Dataset()
print(session.color_scheme)  # default again

session.view = dataset.limit(10)
print(session.color_scheme)  # custom scheme above once again

session.dataset = None
print(session.color_scheme)  # default again
```

### DIFF
--- a/fiftyone/core/odm/dataset.py
+++ b/fiftyone/core/odm/dataset.py
@@ -172,7 +172,14 @@ class ColorScheme(EmbeddedDocument):
         # Store a custom color scheme for a dataset
         dataset.app_config.color_scheme = fo.ColorScheme(
             color_pool=["#ff0000", "#00ff00", "#0000ff", "pink", "yellowgreen"],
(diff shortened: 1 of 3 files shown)


lanzhenw (approved):
@brimoor Thanks for the session related fix. 
LGTM

Issue #3016, opened 2023-05-12
[BUG] ValueError: Dataset name 'dataset' is not available


NevilleMthw:
### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04): Linux Ubuntu
-   **Python version** (`python --version`): 3.10.0
-   **FiftyOne version** (`fiftyone --version`): 0.20.1
-   **FiftyOne installed from** (pip or source): pip

### Commands to reproduce

As thoroughly as possible, please provide the Python and/or shell commands used to encounter the issue. Application steps can be described in the next section.

```
None
```

### Describe the problem

The problem I am facing is quite strange, I get an error which says ```ValueError: Dataset name 'dataset' is not available```. But the dataset name is already available. However, it gets fixed when I edit the name a second time and then run the code, no errors pop up again.

### Code to reproduce issue

```python
dataset_dir = "/file/path/"
name = "dataset"

# Create the dataset
dataset = fo.Dataset.from_dir(
    dataset_dir=dataset_dir,
    dataset_type=fo.types.ImageClassificationDirectoryTree, # Or ImageClassificationDirectoryTree for classification dataset
    name=name,
)

print(dataset)
```
### Other info/logs

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 2
      1 # Create the dataset
----> 2 dataset = fo.Dataset.from_dir(
      3     dataset_dir=dataset_dir,
      4     dataset_type=fo.types.ImageClassificationDirectoryTree, # Or ImageClassificationDirectoryTree for classification dataset
      5     name=name,
      6 )
      8 print(dataset)

File [~/anaconda3/envs/fiftyone/lib/python3.10/site-packages/fiftyone/core/dataset.py:5199](https://file+.vscode-resource.vscode-cdn.net/home/intern/Desktop/fiftyone-scripts/~/anaconda3/envs/fiftyone/lib/python3.10/site-packages/fiftyone/core/dataset.py:5199), in Dataset.from_dir(cls, dataset_dir, dataset_type, data_path, labels_path, name, label_field, tags, dynamic, **kwargs)
   5097 @classmethod
   5098 def from_dir(
   5099     cls,
   (...)
   5108     **kwargs,
   5109 ):
   5110     """Creates a :class:`Dataset` from the contents of the given directory.
   5111 
   5112     You can create datasets with this method via the following basic
   (...)
   5197         a :class:`Dataset`
[...]

PR #3095, opened 2023-05-22
Add persistent and overwrite arguments to dataset factory functions

swheaton:

I propose adding `persistent (default False)` and `overwrite (default False)` arguments to each of the Dataset factory functions (`fo.Dataset.from_*`), to mirror the parameters available in the `fo.Dataset` constructor.

Makes the solution for https://github.com/voxel51/fiftyone/issues/3016 a little more elegant

For reference, here is the args docstring for `class Dataset`
```
    Args:
        name (None): the name of the dataset. By default,
            :func:`get_default_dataset_name` is used
        persistent (False): whether the dataset should persist in the database
            after the session terminates
        overwrite (False): whether to overwrite an existing dataset of the same
            name
```

- added to `DatasetFactoryTests`

```py
import fiftyone as fo
datadir = "/path/to/images"
dataset_name = "dataset"

    datadir,
    dataset_type=fo.types.ImageDirectory,
    name=dataset_name,
    persistent=True
)
assert dataset.persistent

# raises ValueError because name "dataset" already taken
fo.Dataset.from_dir(
    datadir,
    dataset_type=fo.types.ImageDirectory,
    name=dataset_name,
    persistent=True
)

# Succeeds
    datadir,
    dataset_type=fo.types.ImageDirectory,
    name=dataset_name,
    persistent=True,
    overwrite=True,
)
assert dataset.persistent
assert dataset2.name == dataset_name
```

-   [ ] Other

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -5109,11 +5109,11 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         data_path=None,
         labels_path=None,
         name=None,
+        persistent=False,
+        overwrite=False,
         label_field=None,
         tags=None,
         dynamic=False,
-        persistent=False,
-        overwrite=False,
         **kwargs,
     ):
         """Creates a :class:`Dataset` from the contents of the given directory.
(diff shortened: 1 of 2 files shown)


brimoor (approved):
LGTM

PR #3100, opened 2023-05-23
Filter out flattened and redundant sub fields to avoid overcounting label paths, failed tagging, etc.

lanzhenw:
Fix 1
The sample modal was not successfully tagging - unrelated to this work - @sashankaryal fixed that!

Fix 2
small backend tweak where dynamic fields with labels were coming in as `custom_document.detections` and getting suffixed with another detections causing DB filed error

Fix 3
With the recent introduction of allowing labels in embedded fields to be visualized/filtered, some redundant paths were added to the count and tagging paths. the schema is now flattened so that these dynamic fields can be recognized and visualized, etc. 

the flat schema is now bringing some unwanted/redundant label fields into labelPaths which eventually powers the tag label count and gets sent to /tag route when submitting a tag, causing a crash.

ex: labelPaths = [custom_document.detections, custom_document.detections.detections]
is wrong and will cause unwanted issues with anything dependent on the labelPaths.

where,
labelPaths = [custom_document.detections]
works as expected with label count (when Tagger is open) and submitting any tags for labels now success - before crashing if embedded document type exists

Screenshot 2023-05-23 at 10 45 10 AM

Screenshot 2023-05-23 at 10 45 26 AM

Label tag counts are 2x reality

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/state/src/recoil/schema.ts
+++ b/app/packages/state/src/recoil/schema.ts
@@ -1,3 +1,5 @@
+import _ from "lodash";
+
 import { atomFamily, RecoilState, selector, selectorFamily } from "recoil";
 
 import {
@@ -19,7 +21,6 @@ import {
 } from "@fiftyone/utilities";
 
 import { Sample } from "@fiftyone/looker/src/state";
-import _ from "lodash";
 import * as atoms from "./atoms";
 import { State } from "./types";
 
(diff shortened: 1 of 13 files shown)


app/packages/state/src/recoil/schema.ts
```diff
@@ -220,14 +212,17 @@ export const fieldPaths = selectorFamily<
...
+        fieldSchema({ space: State.SPACE.FRAME, flat: true })
+      );
+
+      const sampleLabels = Object.keys(sampleFields)
         .filter((l) => !l.startsWith("_"))
         .sort();
-      const frameLabels = Object.keys(
-        get(fieldSchema({ space: State.SPACE.FRAME, flat: true }))
-      )
+      const frameLabels = Object.keys(frameFields)
```
lanzhenw, line 225: Nice, This is easier to follow.
We should add a comment about do not remove 'flat:true'

manivoxel51 (approved):
lgtm! 🍨 

Should fix tagging.

two things pending
- classification bubbles don't show for embedded docs - Lanny looking into
- Classifications can't be tagged - db error 

will work on right after this.

PR #3107, opened 2023-05-24
fix urls transformation bug

sashankaryal:

`useCreateLooker()` expects `urls` as an object. But because of erroneous upstream transformation, it's sometimes received as an array of objects. This PR adds a guard so that urls passed as arrays are transformed into objects. This was previously done in `Looker.tsx` but there're other modules that also invoke `useCreateLooker()`, like `Grid.tsx` or flashlight, and it makes more sense for this transformation logic to be in `useCreateLooker()`.

Locally.

        notes for FiftyOne users.

-   [ ] Other


lanzhenw (approved):
Looking good

PR #3113, opened 2023-05-25
build-docs gh action checkout teams

swheaton:
Proposal for how to include management SDK in `build-docs` github action.

How it works:
1. Checks out `fiftyone` as normal
2. Checks out `main` branch of `fiftyone-teams` under `fiftyone/fiftyone-teams`
    - needs deployment key secret to checkout private repo (see below)
    - harder to correspond branches dev->dev or releaseX->releaseY, so for now just use `main` which should be latest release.
4. Uses `-t fiftyone-teams` option to `generate_docs.bash` to link relevant folders in before building.

What would be needed to test and finalize it:
1. Create an ssh key
2. Upload private key as a secret to fiftyone, called TEAMS_SSH_PRIVATE_KEY
3. Upload public key as a [deployment key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys#deploy-keys) to `fiftyone-teams` with read only access.
4. Delete generated ssh key from local machine

I think it is safe because:
- deployment key only has read only access to the single repo
- private ssh key is contained within github secret and nowhere else


brimoor (approved):
LGTM

swheaton:
Alright I fixed and tested it - now it will build properly with `fiftyone-teams/main` linked in

(it will probably keep failing until fiftyone-teams release branch is merged into `main`)

benjaminpkane (approved):

PR #3127, opened 2023-05-28
Only match .txt files when reading YOLO labels

democat3457:

Only load files if they end in `.txt`. This will ensure non-text files that are in the directory will not get loaded by accident.

Related to #1975, but takes changes from suggestion here: https://github.com/voxel51/fiftyone/pull/1975#issuecomment-1219818368.

Pass a folder with both `.txt` and non-text files like `.png` into `add_yolo_labels()`.

        notes for FiftyOne users.

-   [ ] Other


brimoor (approved):
LGTM, thanks for the contribution!

PR #3153, opened 2023-06-01
allow configuring max thread pool workers

imanjra:

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -35,6 +35,7 @@ import types
 from xml.parsers.expat import ExpatError
 import zlib
 from matplotlib import colors as mcolors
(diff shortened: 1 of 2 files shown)


fiftyone/core/utils.py
```diff
@@ -1901,6 +1902,16 @@ def to_slug(name):
 
 _T = t.TypeVar("_T")
 
+sync_task_executor = None
```
ritch, line 1905: One improvement would be to handle cleanup of the singleton when not needed.

In generally we want 1 or 0 thread pools. But we don't need an idle pool unless it is being used. I believe this can be accomplished using `with` instead of just a function call.

That being said we could add that later I think.

ritch (approved):

PR #3154, opened 2023-06-01
[1.3.1 Bug fixes]  Field visibility improvements

manivoxel51:

- [x] Field visibility selection mode is the default
- [x] schema search help text removing stray #s
- [x] use selectFields(metadata="foo") when in filter rule mode (continue using exclude_fields() in selection mode)
- [x] remove line number from search help text
- [x] Refactored schema controls
- [x] Add documentation icon/link to custom colors and field visibility modals
- [x] uncheck subpaths when parent path is unchecked even if the subpaths are disabled

Screen Shot 2023-06-06 at 1 22 08 PM

Screen Shot 2023-06-06 at 1 21 32 PM

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Schema/SchemaSelectControls.tsx
+++ b/app/packages/core/src/components/Schema/SchemaSelectControls.tsx
@@ -2,7 +2,6 @@ import React, { useMemo } from "react";
 import { Box, FormControlLabel, FormGroup, Switch } from "@mui/material";
 
 import { useSchemaSettings } from "@fiftyone/state";
-import { TAB_OPTIONS_MAP } from "@fiftyone/state/src/hooks/useSchemaSettings";
 import styled from "styled-components";
 
 const ContainerBox = styled(Box)`
@@ -19,22 +18,21 @@ export const SchemaSelectionControls = () => {
     setShowNestedFields,
     allFieldsChecked,
     setAllFieldsChecked,
-    selectedTab,
+    isFilterRuleActive,
     showMetadata,
     setShowMetadata,
     searchResults,
     includeNestedFields,
     setIncludeNestedFields,
   } = useSchemaSettings();
-  const isFilterRuleMode = selectedTab === TAB_OPTIONS_MAP.FILTER_RULE;
-  const showMetadataVisibile = !!!(isFilterRuleMode && !searchResults.length);
-  const includeNestedVisible = !!(isFilterRuleMode && searchResults.length);
+  const showMetadataVisible = !!!(isFilterRuleActive && !searchResults.length);
+  const includeNestedVisible = !!(isFilterRuleActive && searchResults.length);
 
   const controlList = useMemo(() => {
     return [
       {
         label: "Show metadata",
-        isVisible: showMetadataVisibile,
+        isVisible: showMetadataVisible,
         value: showMetadata,
         checked: showMetadata,
         onChange: () => setShowMetadata(!showMetadata),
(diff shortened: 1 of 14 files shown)


app/packages/core/src/components/Schema/SchemaSelectControls.tsx
```diff
@@ -15,23 +12,61 @@ const ContainerBox = styled(Box)`
...
-    selectedTab,
+    isFilterRuleActive,
     showMetadata,
     setShowMetadata,
     searchResults,
     includeNestedFields,
     setIncludeNestedFields,
   } = useSchemaSettings();
-  const isFilterRuleMode = selectedTab === TAB_OPTIONS_MAP.FILTER_RULE;
+  const showMetadataVisible = !!!(isFilterRuleActive && !searchResults.length);
```
benjaminpkane, line 28:
```suggestion
  const showMetadataVisible = !(isFilterRuleActive && !searchResults.length);
```

benjaminpkane (changes requested):
Looks good! Left some high level comments and suggestions. The one case statement that does not return is the only critical thing from what I saw. I recommend adding eslint to your editor, and looking at some of the dependency warnings in `useSchemaSettings`. It looks like some dependencies may be missing.

It would be interesting see how the `useSchemaSettings` hook could be broken out, and perhaps (in part) moved into recoil data flow. But not important now

app/packages/state/src/hooks/useSchemaSettings.ts
```diff
@@ -70,11 +72,17 @@ const disabledField = (
...
-  const { ftype } = currField;
+  const { ftype, embeddedDocType } = currField;
 
   const parentPath = path.substring(0, path.lastIndexOf("."));
   const parentField = combinedSchema?.[parentPath];
   const parentEmbeddedDocType = parentField?.embeddedDocType;
+  const pathSplit = path.split(".");
+  const embeddedDocTypeSplit = embeddedDocType?.split(".");
+  const hasDynamicEmbeddedDocument = [
+    DYNAMIC_EMBEDDED_DOCUMENT_FIELD,
```
lanzhenw, line 83: Do we still need DYNAMIC_EMBEDDED_DOCUMENT_FIELD? Don't think we are still using dynamic embedded document as a field.

lanzhenw:
Looking good to me. Only one minor thing with this the json formatter. It doesn't seem to still format. We can remove the format button if it's not a required feature.
https://github.com/voxel51/fiftyone/assets/17770824/4426ff72-aa8c-4e70-ba4d-f25d49327828

app/packages/core/src/components/Schema/SchemaSelection.tsx
```diff
@@ -53,30 +63,17 @@ export const SchemaSelection = () => {
...
             );
           case "boolean":
-            <CodeBlock
-              showLineNumbers={false}
-              text={fInfo ? "True" : "False"}
-              fontSize={12}
-              key={`${fInfo}`}
-            />;
+            const boolLabel: string = fInfo ? "True" : "False";
+            <MetaInfoBlock key={path + boolLabel}>{boolLabel}</MetaInfoBlock>;
```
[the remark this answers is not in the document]
manivoxel51: good catch. should return

app/packages/state/src/hooks/useSchemaSettings.ts
manivoxel51: just to confirm you are saying we don't need to check for the first one anymore?
```
export const DYNAMIC_EMBEDDED_DOCUMENT_FIELD =
  "fiftyone.core.fields.DynamicEmbeddedDocumentField";
export const DYNAMIC_EMBEDDED_DOCUMENT_FIELD_V2 =
  "fiftyone.core.odm.embedded_document.DynamicEmbeddedDocument";
```

lanzhenw (approved):
Tested. LGTM

app/vite.config.ts
```diff
@@ -8,6 +8,14 @@ export default defineConfig({
     coverage: {
       reporter: ["json", "lcov"],
       reportsDirectory: "./coverage",
+      enabled: true,
+      all: true,
+      exclude: [
```
benjaminpkane, line 13: Nice

benjaminpkane (approved):
LGTM 🚀

PR #3156, opened 2023-06-02
Add basic index usage to group matching

benjaminpkane:
Adds basic index usage to group matching in `dataset._pipeline()` calls and `SelectGroupSlices`. Lookup pipelines could also be used, but `ViewExpression` use would need to be updated or replaced.

### DIFF
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -4467,10 +4467,10 @@ class SelectGroupSlices(ViewStage):
         slices = self._get_slices(sample_collection)
 
         if etau.is_container(slices):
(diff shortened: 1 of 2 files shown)


fiftyone/core/stages.py
```diff
@@ -4501,7 +4501,6 @@ def _make_pipeline(self, sample_collection):
         ]
 
     def get_media_type(self, sample_collection):
-        group_field = sample_collection.group_field
```
benjaminpkane, line 4504: unused

benjaminpkane:
Stale

PR #3164, opened 2023-06-06
move sample bar to modal from looker in dynamic groups

sashankaryal:

Move sample bar to modal from looker in dynamic groups. This also makes it possible to maximize carousel / image in dynamic groups modal.

All E2E tests green locally.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Modal/Group/DynamicGroup/UnorderedDynamicGroup.tsx
+++ b/app/packages/core/src/components/Modal/Group/DynamicGroup/UnorderedDynamicGroup.tsx
@@ -1,19 +1,71 @@
-import React from "react";
+import { Bar } from "@fiftyone/components";
+import { AbstractLooker } from "@fiftyone/looker";
+import * as fos from "@fiftyone/state";
+import React, { useRef } from "react";
+import { useRecoilValue } from "recoil";
+import styled from "styled-components";
+import { ModalActionsRow } from "../../../Actions";
 import Sample from "../../Sample";
 import { useGroupContext } from "../GroupContextProvider";
 import { DynamicGroupCarousel } from "./carousel/DynamicGroupCarousel";
 
+const RootContainer = styled.div`
+  height: 100%;
+  width: 100%;
+`;
+
+const ElementsContainer = styled.div`
+  width: 100%;
+  height: 100%;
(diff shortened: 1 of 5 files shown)


imanjra (approved):

app/packages/core/src/components/Modal/Group/DynamicGroup/UnorderedDynamicGroup.tsx
```diff
@@ -1,19 +1,72 @@
...
+            lookerRef={lookerRef}
+            hideSampleBar
+          />
+        )}
+      </ElementsContainer>
+    </RootContainer>
+  );
+};
+
+const UnorderedDynamicGroupBar = ({
```
ritch, line 55: Consider:

```typescript
interface UnorderedDynamicGroupBarProps {
  lookerRef: React.RefObject<AbstractLooker>;
}

const UnorderedDynamicGroupBar: React.FC<UnorderedDynamicGroupBarProps> = ({
  lookerRef,
}) => {
  // ...
};
```

ritch: `React.MutableRefObject<AbstractLooker | undefined>; could be replaced with React.RefObject<AbstractLooker> because MutableRefObject is expected to be initialized when it's created.

e2e/cypress/support/e2e.ts
```diff
@@ -9,6 +9,12 @@ compareSnapshotCommand({
   errorThreshold: 0,
 });
 
+Cypress.on("uncaught:exception", (err, runnable) => {
```
ritch, line 12: We probably shouldn't be ignoring all exceptions. Is there specific one you were trying to avoid causing a faliure? if so, add it here.

ritch (approved):
Added a few comments. Overall LGTM.

e2e/cypress/support/e2e.ts
sashankaryal: Right now we indiscriminately return false which means we're not ignoring any exceptions. I just wanted to keep it there in case we wanted to catch any known exceptions that could cause us trouble.

PR #3169, opened 2023-06-06
Server patches

benjaminpkane:
### Changes

* In `fiftyone.server.events`, copy the `StateUpdate` event as it is a shared object. Ideally, `StateDescription` can be updated to a `dataclass` for a cleaner solution
* In `fiftyone.core.state`, recursively convert `BaseDict` and `BaseList` data as they interferes with `dataclasses.asdict`

### DIFF
--- a/fiftyone/core/state.py
+++ b/fiftyone/core/state.py
@@ -5,13 +5,13 @@ Defines the shared state between the FiftyOne App and backend.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-import json
+
 import logging
 import typing as t
 
 from bson import json_util
 from dataclasses import asdict
-from mongoengine.base import BaseDict
+from mongoengine.base import BaseDict, BaseList
 import strawberry as gql
 
 import eta.core.serial as etas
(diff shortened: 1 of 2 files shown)


lanzhenw (approved):

fiftyone/core/state.py
```diff
@@ -239,10 +239,20 @@ def serialize_fields(schema: t.Dict) -> t.List[SampleField]:
...
                     else field.info,
                 )
             )
 
     return data
+
+
+def _convert_mongoengine_data(data):
+    if isinstance(data, BaseDict):
+        return {k: _convert_mongoengine_data(v) for k, v in data.items()}
```
kaixi-wang, line 253: this will make ordered dicts into unordered dicts. not sure what implications if any that has, but just wanted to call it out in case

kaixi-wang (approved):

fiftyone/core/state.py
lanzhenw: shouldn't be an issue for python3.7 and after

benjaminpkane: JSON data is not ordered, and this is only for sending payload data

PR #3174, opened 2023-06-08
use $addFields and $project in aggregation pipelines

brimoor:
Replaces all usages of `$set` and `$unset` in aggregation pipelines with `$addFields` and `$project: {XXX: False}`, respectively.

This is being done as part of an effort to broaden the compatibility of FiftyOne with older MongoDB versions and other technologies like DocumentDB.

### DIFF
--- a/fiftyone/core/clips.py
+++ b/fiftyone/core/clips.py
@@ -757,12 +757,12 @@ def _write_support_clips(
 
     if is_list:
         pipeline.extend(
-            [{"$unwind": "$support"}, {"$set": {"_rand": {"$rand": {}}}}]
+            [{"$unwind": "$support"}, {"$addFields": {"_rand": {"$rand": {}}}}]
         )
 
     pipeline.extend(
         [
-            {"$set": {"_dataset_id": dataset._doc.id}},
+            {"$addFields": {"_dataset_id": dataset._doc.id}},
             {"$out": dataset._sample_collection_name},
         ]
     )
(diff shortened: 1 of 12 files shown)


fiftyone/core/clips.py
```diff
@@ -887,13 +890,14 @@ def _write_trajectories(dataset, src_collection, field, other_fields=None):
...
-                {"$unset": _tmp_field},
+                {"$project": {_tmp_field: False}},
                 {"$out": dataset._sample_collection_name},
             ]
         )
     finally:
-        cleanup_op = {"$unset": {_tmp_field: ""}}
-        src_dataset._sample_collection.update_many({}, cleanup_op)
+        src_dataset._sample_collection.update_many(
+            {}, {"$unset": {_tmp_field: ""}}
```
swheaton, line 899: Did you mean to use $project here also?

fiftyone/core/clips.py
```diff
@@ -964,8 +968,9 @@ def _write_manual_clips(dataset, src_collection, clips, other_fields=None):
...
```
swheaton, line 972: Same here

brimoor: `$set` and `$unset` are okay in update operations, they just cannot be used in aggregation pipelines in DocumentDB/older Mongo versions

swheaton: Ok I did not notice it was an update op

swheaton (approved):
Looks good to me!

For future readers, here's the [documentDB compatibility page](https://docs.aws.amazon.com/documentdb/latest/developerguide/mongo-apis.html)

Side note for future, another stage we use that is unsupported by Document DB is `$merge`. But that one will be much harder to replace than `$set` and `$unset` I imagine.

swheaton:
I mean, I think it's the main effect right? No other reason to replace those stages.

There are also many ViewExpression functions that are incompatible such as `trim` and `ceil`. Probably not as big of a deal for it to just not work in doc db

Anddraca:
I figured the main push was for compatibility with the older mongo versions. 

If we're going to target documentDB as a first-class second-class citizen, I think it would be worth it to abstract the db access out further so swapping is a lot easier....which brings up the mongoengine conversation, as well. That's why I was curious if it was a targeted goal from a larger perspective

PR #3181, opened 2023-06-10
Add CVAT support for frame_start frame_stop and frame_step

ehofesmann:
CVAT has [three arguments](https://opencv.github.io/cvat/docs/manual/basics/create_an_annotation_task/) that let you define which portions of videos are uploaded: `frame_start`, `frame_stop`, and `frame_step` 

This PR adds support to these through keyword arguments when calling `dataset.annotate()` with the CVAT backend as well as docs and tests.

@brimoor I'm a bit confused why this is required: https://github.com/voxel51/fiftyone/blob/8aca7cb29cbb9ddcb12a54340dffb36f2736c5b9/fiftyone/utils/cvat.py#L3218-L3222
These args are being loaded as mongoengine versions of lists/dicts if you load this config from the DB at some future time. I'm not sure if this is the right way to handle this or if this is indicative of other issues with serialization/deserialization of these args in the config.

Docs:

- frame_start (None): nonnegative integer(s) defining the first frame of videos to upload when creating video tasks. Supported types are:
  - integer: the first frame to upload for each video
  - list: a list of first frame integers corresponding to videos in the given samples. If fewer frame_start values are provided than there are videos in the given samples, they will be reassigned with a round-robin strategy
  - dict: a dictionary mapping sample filepath to the first frame integer to use for the corresponding video

  Note: This argument is only supported for videos.

- frame_stop (None): nonnegative integer(s) defining the last frame of videos to upload when creating video tasks. Supported types are:

  - integer: the last frame to upload for each video

  - list: a list of last frame integers corresponding to videos in the given samples. If fewer frame_stop values are provided than there are videos in the given samples, they will be reassigned with a round-robin strategy

  - dict: a dictionary mapping sample filepath to the last frame integer to use for the corresponding video

  Note: This argument is only supported for videos.

- frame_step (None): positive integer(s) defining which frames to sample when creating video tasks. For example, a frame step of 25 will include frames 1, 26, 51 and so on. Supported types are:

  - integer: the frame step to apply to each video task

  - list: a list of frame step integers corresponding to videos in the given samples. If fewer frame_step values are provided than there are videos in the given samples, they will be reassigned with a round-robin strategy

  - dict: a dictionary mapping sample filepath to the frame step integer to use for the corresponding video

  Note: This argument is only supported for videos and does not support uploading existing tracks when provided.

image

### DIFF
--- a/docs/source/integrations/cvat.rst
+++ b/docs/source/integrations/cvat.rst
@@ -505,6 +505,48 @@ provided:
     or when using `task_size` and generating multiple tasks
 -   **organization** (*None*): the name of the organization to use when sending
     requests to CVAT
+-   **frame_start** (*None*): nonnegative integer(s) defining the first frame of
+    videos to upload when creating video tasks.
+    Supported types are:
+
+    - `integer`: the first frame to upload for each video
+    - `list`: a list of first frame integers corresponding to
+      videos in the given samples. If fewer `frame_start` values are
+      provided than there are videos in the given samples, they will be
+      reassigned with a round-robin strategy
+    - `dict`: a dictionary mapping sample filepath to the first frame
+      integer to use for the corresponding video
+
+    Note: This argument is only supported for videos.
+-   **frame_stop** (*None*): nonnegative integer(s) defining the last frame of
+    videos to upload when creating video tasks.
+    Supported types are:
+
+    - `integer`: the last frame to upload for each video
+    - `list`: a list of last frame integers corresponding to
+      videos in the given samples. If fewer `frame_stop` values are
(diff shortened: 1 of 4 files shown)


brimoor:
@ehofesmann is there a reason you're casting from `BaseDict` and `BaseList` to `dict` and `list`?

brimoor:
> @ehofesmann is there a reason you're casting from `BaseDict` and `BaseList` to `dict` and `list`?

oh I see you just need to use `issubclass(d, dict)` rather than `type(d) == dict`. I'll update.

brimoor (approved):
LGTM

PR #3182, opened 2023-06-11
Bumping compatibility version

brimoor:
The compatibility version wasn't bumped with the 0.21 release.

This likely wasn't detected because `database_admin=True` by default in OSS, so compatibility version doesn't come into play (the compatibility version *was* bumped in Teams, where `database_admin=False` by default).


benjaminpkane (approved):
LGTM

Issue #3185, opened 2023-06-12
[BUG] Concurrently adding new fields from different processes is unstable


brimoor:
Executing the steps below in the order indicated raises errors.

The root problem is multiple processes adding elements to `DatasetDocument.sample_fields` concurrently. Rather than both adding elements using array updates, they are both be `$set`-ing the entire field, as `mongoengine` does not support intelligent granular updates of array fields.

## TERMINAL A

```py
import fiftyone as fo
import fiftyone.zoo as foz

# step 1
dataset = foz.load_zoo_dataset("quickstart")
dataset.persistent = True

# step 3
sample = ds.first()
sample["foo"] = "bar"

# step 5
sample.save()

# step 7: ERROR
dataset.reload()
```

## TERMINAL B

```py
import fiftyone as fo

# step 2
dataset = fo.load_dataset("quickstart")

# step 4
sample = dataset.first()
sample["spam"] = "eggs"

# step 6
sample.save()

# step 8: ERROR
dataset.reload()
```

PR #3186, opened 2023-06-12
[WIP] Improved robustness of concurrent schema updates

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/3185

As explained in the `@todo`, this is a bit of a hack to specifically avoid issues when concurrently modifying a dataset's schema. However, the underlying problem still exists whenever list fields other than `DatasetDocument.sample_fields` and `DatasetDocument.frame_fields` are concurrently edited without first reloading.

I think it makes sense to go ahead and merge this particular patch because schema updates are by far the most likely case where concurrent list edits may arise, since in many workflows `dataset` objects may be held in-memory for long periods of time without reloading them. Unlike samples, which are generally only loaded + modified on-demand.


allenleetc (dismissed):

allenleetc:
did not test but LGTM

brimoor:
Closing in favor of https://github.com/voxel51/fiftyone/pull/3308.

PR #3200, opened 2023-06-14
fix annoying pcd selector disappearing bug

sashankaryal:

Fix selector closing inconveniently as a result of container.onmouseleave triggering when looker3d re-renders. Removing `onMouseLeave` from `Container` has no real consequence, as `onMouseOver` subsumes what it intends to do.

Before:
before

After:
after

Locally.

        notes for FiftyOne users.

-   [ ] Other


lanzhenw (approved):
LGTM

PR #3212, opened 2023-06-16
add methods for finding and deleting duplicates

kaixi-wang:

- [x] enable finding duplicate plugin directories and deleting them easily from the sdk (helpful in resolving situations where you manually move files into the plugin dir or rename to an existing name while developing plugins)

- [x] optimize filtering by using generators

- [x] add tests for new functions


ritch (approved):

brimoor (changes requested):
I would rather not expose delete_plugin_dir() because it may cause unnecessary confusion about what the difference between a plugin and a plugin directory actually is.

I’m fine with a utility to identify duplicates, but, since that should be a rare occurrence, better to just leave the handling of that case to the user.

fiftyone/plugins/__init__.py
```diff
@@ -11,12 +11,14 @@
...
     disable_plugin,
     delete_plugin,
+    delete_plugin_dir,
     list_plugins,
     list_disabled_plugins,
     list_downloaded_plugins,
     list_enabled_plugins,
     get_plugin,
     find_plugin,
+    find_duplicates,
```
brimoor, line 21: for consistency with other method names, suggest `find_duplicate_plugins` here

fiftyone/plugins/core.py
```diff
@@ -564,18 +589,20 @@ def _iter_plugin_metadata_files():
 
 
 def _get_plugin(name, enabled=None, check_for_duplicates=True):
+    """Find plugin packages matching name."""
```
brimoor, line 592: I think it is confusing for `_get_plugin(name)` to always be a generator. The default arg of  `check_for_duplicates=True` implies that it will either return a single plugin or raise an error if there are duplicates.

A solution would be to add an optional syntax that returns a list rather than a single plugin:

```py
# Returns a list of one or more plugins
plugins = _get_plugin(name, include_duplicates=True)
```

brimoor (commented):
I believe that all plugin management methods should be made available via the CLI. So, if we're going to merge this PR, I propose exposing the duplicate plugin capability via the following syntax:

```
fiftyone plugins list --include-duplicates
```

If `--include-duplicates` flag is passed, then an extra column `duplicate=✓` is added to the output table that indicates whether a given plugin is a "duplicate", ie shadowed by another plugin with the same name.

```
$ fiftyone plugins list
plugin                version    enabled    directory
--------------------  ---------  ---------  ----------------------------------------------------------------------
@voxel51/io           1.0.0      ✓          /Users/Brian/fiftyone/__plugins__/fiftyone-plugins/plugins/io
@voxel51/annotation   1.0.0      ✓          /Users/Brian/fiftyone/__plugins__/fiftyone-plugins/plugins/annotation
@voxel51/hello-world  1.0.0                 /Users/Brian/fiftyone/__plugins__/fiftyone-plugins/plugins/hello-world
@voxel51/examples     1.0.0                 /Users/Brian/fiftyone/__plugins__/fiftyone-plugins/plugins/examples
@voxel51/python-view  1.0.0                 /Users/Brian/fiftyone/__plugins__/fiftyone-plugins/plugins/python-view
@voxel51/voxelgpt     1.0.0      ✓          /Users/Brian/fiftyone/__plugins__/voxelgpt
```

Since the table contains the plugin directory, it would be easy to see what directory you need to manually deal with in order to resolve the duplicate plugin issue.

fiftyone/plugins/core.py
```diff
@@ -90,16 +90,27 @@ def disable_plugin(plugin_name):
...
     Args:
         plugin_name: the plugin name
     """
-    plugin = _get_plugin(plugin_name)
+    plugin = next(_get_plugin(plugin_name))
     _update_plugin_settings(plugin_name, delete=True)
     etau.delete_dir(plugin.path)
 
 
+def delete_plugin_dir(plugin_dir):
```
brimoor, line 103: I would rather not have a `delete_plugin_dir()` method because it may cause unnecessary confusion about what the difference between a plugin and a plugin directory actually is.

If you do have duplicate plugins, it may be because you have the same plugin on your path twice, but it also could be that you're developing a plugin whose name clashes with another plugin, in which case the solution is not to delete either plugin but instead rename one plugin. Or maybe that's not the solution, but point is that dealing with duplicate plugins is a manual workflow.

kaixi-wang: I tried to use the docstring to make it clear what the use case is and why it's different from deleting a plugin.

I agree that dealing with duplicate plugins is manual, so that's why I didn't make a delete duplicate plugin method. Although a user can easily delete a directory in other ways, I think this is just nice to have to fully resolve a duplicate plugin error without using another library or going into the gui filesystem

fiftyone/plugins/core.py
brimoor: I’m aware. Comment still stands

kaixi-wang:
> Since the table contains the plugin directory, it would be easy to see what directory you need to manually deal with in order to resolve the duplicate plugin issue.

I can add a cli method, but I don't think it's useful to fold it in the list method. Plugin name collisions cause runtime errors, and should be addressed specifically. Regular plugin methods like list should always throw an error.

fiftyone/plugins/__init__.py
kaixi-wang: updated

PR #3222, opened 2023-06-26
Session proxy url

### LINKED ISSUE #2995: [FR] Serving FiftyOne behind a reverse Proxy
Hello and thank you for maintaining this project!

Our data scientists love working with FiftyOne, however, we currently cannot use it with our infrastructure.

I'd assume that we're not the only one with this issue, as our setup should be pretty common.

## Problem

We are serving Notebooks through Kubeflow. Kubeflow itself is exposed through an Ingress and the notebooks are available through a dedicated URL, that forwards traffic to one port (8888 in our case).

Thus, only **one** port to the container can be opened. This port is used by Jupyter. 
There is no publicly available port left to open for fiftyone, which is an issue because fiftyone launches a server that binds to a port.

This setup should be pretty common. 

There is a Kubeflow dedicated issue already. However, it doesn't propose a solution and has become stale: https://github.com/voxel51/fiftyone/issues/1901

## Our Solution Approach

We've set up a reverse proxy in our containers (see files below). Nginx listens on port 8888, Jupyter on port 8889, and fiftyone on port 5151.
All requests to `/fiftyone` (and other endpoints, such as `/assets`) are proxied to fiftyone, while all other requests are by default received by Jupyter.

This works pretty well and we're lucky here that Jupyter and FiftyOne have no overlapping paths. 
Opening `http://url/fiftyone` in the browser works fine and displays FiftyOne, while `http://url` opens Jupyter.

## Limitations

While the setup works fine, the `launch_app` isn't flexible enough.

The following code launches it on the right port, however, this is not the same port that is accessible to the user.
```python
import fiftyone

fiftyone.launch_app(port=5151, address="0.0.0.0")
```

The user can only connect to FiftyOne on the nginx port and path /fiftyone. 

Do you have any ideas how this could be solved?

Thank you for any kind of advice and help!

## The Test Case 

I've attached a simple example. It build a Jupyter docker image that contains the nginx reverse proxy and fiftyone.

### Launch Command

```bash
docker build -t fiftyone:dev . && docker run --rm -it -p 8888:8888 fiftyone:dev 
# You may add -p 5151:5151 to access fiftyone directly on port 5151 
```

You may then open `http://localhost:8888` to open Jupyter and `http://localhost:8888/fiftyone` for the FiftyOne UI.

### Dockerfile
```Dockerfile
FROM jupyter/datascience-notebook

RUN pip install fiftyone 

USER 0

ENV UID=1000

RUN \
    # Install nginx 
    apt-get update --yes && \ 
    apt-get install --yes nginx && \
    # Install libraries required for fiftyone
    apt-get install --yes libcurl4 && \
    wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2.18_amd64.deb && \
    dpkg -i libssl1.1_1.1.1f-1ubuntu2.18_amd64.deb && \
    rm libssl1.1_1.1.1f-1ubuntu2.18_amd64.deb

## add permissions for nginx user 
RUN mkdir /var/cache/nginx && \ 
    touch /var/run/nginx.pid && \ 
    chown -R $UID:$UID /var/cache/nginx && \ 
    chown -R $UID:$UID /var/log/nginx/ && \ 
    chown -R $UID:$UID /var/lib/nginx/ && \ 
    chown -R $UID:$UID /var/run/nginx.pid 

COPY ./nginx.conf /etc/nginx/nginx.conf 
COPY ./default.conf /etc/nginx/conf.d/default.conf 

ENV \
    # Have Jupyter run on port 8889
    JUPYTER_PORT=8889 \
    # Used by entrypoint script: disable token authentication for testing
    NOTEBOOK_ARGS="--LabApp.token='' --LabApp.allow_origin='*'"

USER $UID

ENTRYPOINT ["/bin/bash"]
CMD ["-c", "\
    python3 -c 'import fiftyone; fiftyone.launch_app(port=5151, address=\"0.0.0.0\"); print(\"Started FiftyOne\"); import time; time.sleep(9999);' & \
    nginx && echo 'Started nginx' && \
    start-notebook.sh"\
]
```

### default.conf

```conf
map $http_upgrade $connection_upgrade { 
    default upgrade; 
    '' close; 
} 

server { 
    listen 8888; 
    server_name localhost; 
    
    location /fiftyone { 
        proxy_pass http://127.0.0.1:5151/; 
    } 
    location /plugins { 
        proxy_pass http://127.0.0.1:5151/plugins; 
    } 
    location /assets { 
        proxy_pass http://127.0.0.1:5151/assets; 
    } 
    location /events { 
        proxy_pass http://127.0.0.1:5151/events; 
    } 
    location /graphql { 
        proxy_pass http://127.0.0.1:5151/graphql; 
    } 
    
    location / { 
        proxy_pass http://127.0.0.1:8889; 
        # Configure reverse proxy for jupyter: https://jupyterhub.readthedocs.io/en/stable/reference/config-proxy.html 
        proxy_set_header X-Real-IP $remote_addr; 
        proxy_set_header Host $host; 
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for; 
        
        # websocket headers 
        proxy_http_version 1.1; 
        proxy_set_header Upgrade $http_upgrade; 
        proxy_set_header Connection $connection_upgrade; 
        proxy_set_header X-Scheme $scheme; 
        
        proxy_buffering off; 
    } 
}
```

### nginx.conf
```conf
# This file is taken from /etc/nginx/nginx.conf

# Only change: commented following line
# user www-data;

worker_processes auto;
pid /run/nginx.pid;
include /etc/nginx/modules-enabled/*.conf;

events {        
    worker_connections 768;        
    # multi_accept on;
}

http {        
    ##        
    # Basic Settings        
    ##        
    
    sendfile on;        
    tcp_nopush on;       
    tcp_nodelay on;        
    keepalive_timeout 65;        
    types_hash_max_size 2048;        
    # server_tokens off;        
    
    # server_names_hash_bucket_size 64;        
    # server_name_in_redirect off;        
    
    include /etc/nginx/mime.types;        
    default_type application/octet-stream;        
    
    ##        
    # SSL Settings        
    ##        
    
    ssl_protocols TLSv1 TLSv1.1 TLSv1.2 TLSv1.3; # Dropping SSLv3, ref: POODLE        
    ssl_prefer_server_ciphers on;        
    
    ##        
    # Logging Settings        
    ##        
    
    access_log /var/log/nginx/access.log;        
    error_log /var/log/nginx/error.log;        
    
    ##        
    # Gzip Settings        
    ##        gzip on;        
    
    # gzip_vary on;        
    # gzip_proxied any;        
    # gzip_comp_level 6;        
    # gzip_buffers 16 8k;        
    # gzip_http_version 1.1;        
    # gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;        
    
    ##        
    # Virtual Host Configs        
    ##        
    
    include /etc/nginx/conf.d/*.conf;        
    include /etc/nginx/sites-enabled/*;
}
```

benjaminpkane:
Resolves #2995 

Adds `proxy_url` to the `AppConfig` allowing for an override of the server URL. This only affect how the session resolves `session.url` for browser interactions, e.g. opening a tab or displaying a notebook cell

### DIFF
--- a/fiftyone/core/session/notebooks.py
+++ b/fiftyone/core/session/notebooks.py
@@ -72,7 +72,7 @@ def display_ipython(
     iframe = IPython.display.IFrame(
         focx.get_url(
             cell.address,
-            os.environ.get("FIFTYONE_APP_CLIENT_PORT", cell.port),
+            cell.port,
             notebook=True,
             proxy_url=proxy_url,
             subscription=cell.subscription,
--- a/fiftyone/core/session/notebooks.py
+++ b/fiftyone/core/session/notebooks.py
@@ -47,7 +47,10 @@ def capture(cell: NotebookCell, data: fose.CaptureNotebookCell) -> None:
 
 
 def display(
-    client: Client, cell: NotebookCell, reactivate: bool = False
+    client: Client,
+    cell: NotebookCell,
+    proxy_url: str = None,
+    reactivate: bool = False,
 ) -> None:
     """Displays a running FiftyOne instance."""
     funcs = {
@@ -56,12 +59,13 @@ def display(
         focx._DATABRICKS: display_databricks,
     }
     fn = funcs[focx._get_context()]
-    fn(client, cell, reactivate)
+    fn(client, cell, proxy_url=proxy_url, reactivate=reactivate)
 
 
 def display_ipython(
     client: Client,
     cell: NotebookCell,
+    proxy_url: str = None,
     reactivate: bool = False,
     **kwargs: t.Dict[str, t.Union[str, bool]],
 ) -> None:
@@ -70,6 +74,7 @@ def display_ipython(
             cell.address,
             os.environ.get("FIFTYONE_APP_CLIENT_PORT", cell.port),
             notebook=True,
+            proxy_url=proxy_url,
             subscription=cell.subscription,
             **kwargs,
         ),
@@ -83,7 +88,10 @@ def display_ipython(
 
 
 def display_colab(
-    client: Client, cell: NotebookCell, reactivate: bool = False
+    client: Client,
+    cell: NotebookCell,
+    proxy_url: str = None,
+    reactivate: bool = False,
 ) -> None:
     """Display a FiftyOne instance in a Colab output frame.
 
@@ -134,7 +142,10 @@ def display_colab(
 
 
 def display_databricks(
-    client: Client, cell: NotebookCell, reactivate: bool = False
+    client: Client,
+    cell: NotebookCell,
+    proxy_url: str = None,
+    reactivate: bool = False,
 ):
     """Display a FiftyOne instance in a Databricks output frame.
 
@@ -144,6 +155,7 @@ def display_databricks(
     display_ipython(
         client,
         cell,
-        reactivate=reactivate,
         polling=True,
+        proxy_url=proxy_url,
+        reactivate=reactivate,
     )
--- a/docs/source/user_guide/config.rst
+++ b/docs/source/user_guide/config.rst
@@ -650,6 +650,9 @@ The FiftyOne App can be configured in the ways described below:
 +---------------------------+----------------------------------------+-----------------------------+-------------------------------------------------------------------------------------------+
 | `notebook_height`         | `FIFTYONE_APP_NOTEBOOK_HEIGHT`         | `800`                       | The height of App instances displayed in notebook cells.                                  |
 +---------------------------+----------------------------------------+-----------------------------+-------------------------------------------------------------------------------------------+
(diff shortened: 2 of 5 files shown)


fiftyone/core/session/notebooks.py
```diff
@@ -56,20 +59,22 @@ def display(
...
     cell: NotebookCell,
+    proxy_url: str = None,
     reactivate: bool = False,
     **kwargs: t.Dict[str, t.Union[str, bool]],
 ) -> None:
     iframe = IPython.display.IFrame(
         focx.get_url(
             cell.address,
-            os.environ.get("FIFTYONE_APP_CLIENT_PORT", cell.port),
+            cell.port,
```
imanjra, line 75: Is this change intentional?

PR #3230, opened 2023-06-27
Allow coloring embeddings plots by list fields

brimoor:
Allows embeddings plots to be colored by list fields, eg sample or label tags.

This is useful, for example, in situations where each sample contains only one tag. IE train/test/val split. If there are multiple tags, the first one is used.

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

fob.compute_visualization(dataset, brain_key="img_sim")
fob.compute_visualization(dataset, patches_field="ground_truth", brain_key="gt_sim")

for tag in "abcd":
    dataset.take(100).tag_samples(tag)
    dataset.take(100).tag_labels(tag, label_fields="ground_truth")

```

https://github.com/voxel51/fiftyone/assets/25985824/81c9ee70-bcf4-4e5a-acb2-091344ffb1e3

### DIFF
--- a/fiftyone/server/routes/embeddings.py
+++ b/fiftyone/server/routes/embeddings.py
@@ -25,6 +25,7 @@ COLOR_BY_TYPES = (
     fof.BooleanField,
     fof.IntField,
     fof.FloatField,
+    fof.ListField,
 )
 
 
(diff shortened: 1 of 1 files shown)


manivoxel51 (approved):
cool! 🍨

PR #3231, opened 2023-06-27
Resolve saved views when serializing datasets

benjaminpkane:
Fixes the `/sort` route and other occurrences of serializing a dataset outside of graphql


manivoxel51 (approved):
thank you! 🍨

PR #3237, opened 2023-06-29
Fix reset color

lanzhenw:

- Bugfix: `reset` button failed to set `can view` permission user setting back to dataset app_config setting on TEAMS. 
- Previously x-direction scroller shows up unnecessarily on color modal. 

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/ColorModal/ColorFooter.tsx
+++ b/app/packages/core/src/components/ColorModal/ColorFooter.tsx
@@ -38,6 +38,11 @@ const ColorFooter: React.FC = () => {
     return true;
   }, [savedSettings]);
 
+  // set to be true only for teams
+  const isTeams = useRecoilValue(fos.compactLayout);
+  const datasetDefault =
+    useRecoilValue(fos.datasetAppConfig)?.colorScheme ?? null;
+
   if (!activeColorModalField) return null;
 
   return (
@@ -47,7 +52,7 @@ const ColorFooter: React.FC = () => {
           text={"Reset"}
           title={`Clear session settings and revert to default settings`}
           onClick={() => {
-            setColorScheme(false, null);
+            setColorScheme(false, isTeams ? datasetDefault : null);
             setReset((prev) => prev + 1);
           }}
           style={BUTTON_STYLE}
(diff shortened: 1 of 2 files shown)


app/packages/core/src/components/ColorModal/ColorFooter.tsx
```diff
@@ -38,6 +38,11 @@ const ColorFooter: React.FC = () => {
     return true;
   }, [savedSettings]);
 
+  // set to be true only for teams
+  const isTeams = useRecoilValue(fos.compactLayout);
```
benjaminpkane, line 42: This is logic/naming something we should avoid. `compactLayout` means `isTeams` by coincidence only

app/packages/core/src/components/ColorModal/ColorFooter.tsx
```diff
@@ -47,7 +52,7 @@ const ColorFooter: React.FC = () => {
           text={"Reset"}
           title={`Clear session settings and revert to default settings`}
           onClick={() => {
-            setColorScheme(false, null);
+            setColorScheme(false, isTeams ? datasetDefault : null);
```
benjaminpkane, line 55: I'm not sure I understand the behavior here. Reset means two different things based on the context? How does a Teams user nullify a dataset default? Ideally we add a `selector` that resolves to the reset value (resolution should be opaque/decoupled)

benjaminpkane (approved):
LGTM for the release

app/packages/core/src/components/ColorModal/ColorFooter.tsx
lanzhenw: Reset always means reset to dataset default (if there is one). We have something else to clear dataset default setting. 

By setting colorScheme to `null`, OSS will load dataset default or color pool from `fo.app_config` or `session.config` from the session. But teams can't update through the session. Therefore, I am pointing the setting to datasetDefault here for TEAMS.

benjaminpkane: Ok. In my routing work, the color scheme will update through a `session` regardless of what the means to the context (Teams, OSS, or a hypothetical). I'm just advocating that this doesn't leak into OSS and component logic

PR #3244, opened 2023-06-30
[DOCS] Compound indexes for group datasets

benjaminpkane:
Adds a compound indexes example for App filtering index usage


ehofesmann (approved):
LGTM!

PR #3248, opened 2023-07-03
Split field visibility and filter in sidebar

lanzhenw:
Feature: split sidebar filter and visibility control
- Added filter/visibility mode pill button to sidebar filter control (field search, visibility pill button), which is now sticky in the sidebar
- Implemented visibility and filter separation in the sidebar. The two features are independent now.

Example of label tags (all labels in samples with people are tagged "people", go to the station sample, untag the train): 

https://github.com/voxel51/fiftyone/assets/17770824/afe7054f-7998-4512-b8af-490f47e745d8

Example of nested ListField: 

https://github.com/voxel51/fiftyone/assets/17770824/0f583ce9-bc26-4f4e-b77e-170e4d76bc8d

Example of primitive fields:
Boolean, Number, String fields
https://github.com/voxel51/fiftyone/assets/17770824/88eb5f49-b7aa-47f8-8aec-e789ec4e6f04

ListFields
https://github.com/voxel51/fiftyone/assets/17770824/c7963237-8e66-471b-924d-3818f4acfd77

## How to test this?
Use the following code create different types of primitive and label fields:
```
import random
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

n = len(dataset)
labels = ["foo", "bar", "spam", "eggs"]

# Add top-level fields of each primitive type
dataset.set_values("bool_field", list(np.random.randn(n) > 0))
dataset.set_values("int_field", list(np.random.randint(10, size=n)))
dataset.set_values("float_field", list(np.random.randn(n)))
dataset.set_values("str_field", [random.choice(labels) for _ in range(n)])

# Add label attributes of each primitive type
p = len(patches)

dataset.add_sample_field("ground_truth.detections.bool_field", fo.BooleanField)
dataset.add_sample_field("ground_truth.detections.int_field", fo.IntField)
dataset.add_sample_field("ground_truth.detections.float_field", fo.FloatField)
dataset.add_sample_field("ground_truth.detections.str_field", fo.StringField)

patches.set_values("ground_truth.bool_field", list(np.random.randn(p) > 0))
patches.set_values("ground_truth.int_field", list(np.random.randint(10, size=p)))
patches.set_values("ground_truth.float_field", list(np.random.randn(p)))
patches.set_values("ground_truth.str_field", [random.choice(labels) for _ in range(p)])

n = len(dataset)
labels = ["foo", "bar", "spam", "eggs"]

dataset.add_sample_field("list_field", fo.ListField, subfield=fo.StringField)

# Add top-level fields of each primitive type
dataset.set_values("bool_field", list(np.random.randn(n) > 0))
dataset.set_values("int_field", list(np.random.randint(10, size=n)))
dataset.set_values("float_field", list(np.random.randn(n)))
dataset.set_values("str_field", [random.choice(labels) for _ in range(n)])
dataset.set_values("list_field", [[random.choice(labels) for _ in range(2)] for __ in range(n)])

# Add label attributes of each primitive type
p = len(patches)

dataset.add_sample_field("ground_truth.detections.bool_field", fo.BooleanField)
dataset.add_sample_field("ground_truth.detections.int_field", fo.IntField)
dataset.add_sample_field("ground_truth.detections.float_field", fo.FloatField)
dataset.add_sample_field("ground_truth.detections.str_field", fo.StringField)
dataset.add_sample_field("ground_truth.detections.list_field", fo.ListField, subfield=fo.StringField)

patches.set_values("ground_truth.bool_field", list(np.random.randn(p) > 0))
patches.set_values("ground_truth.int_field", list(np.random.randint(10, size=p)))
patches.set_values("ground_truth.float_field", list(np.random.randn(p)))
patches.set_values("ground_truth.str_field", [random.choice(labels) for _ in range(p)])
dataset.set_values("ground_truth.list_field", [[random.choice(labels) for _ in range(2)] for __ in range(p)])

```

(Details)

        notes for FiftyOne users.

-   [ ] Other


app/packages/core/src/components/Common/Checkbox.tsx
```diff
@@ -58,6 +58,7 @@ const Checkbox = <T extends unknown>({
     () => (typeof count === "number" ? constSelector(count) : null),
     [count]
   );
+  const isFilterMode = useRecoilValue(fos.isSidebarFilterMode);
```
manivoxel51, line 61: nit: what do you think about passing this down as prop so we keep this Checkbox component as generic as possible and configure it through props.

app/packages/core/src/components/Filters/categoricalFilter/Wrapper.tsx
```diff
@@ -105,7 +101,8 @@ const Wrapper = ({
       </>
     );
   }
-
+  console.info("V", visibility);
+  console.info("F", filter);
```
manivoxel51, line 105: ✂️

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -137,19 +169,26 @@ const FilterOption: React.FC<Props> = ({
...
+        ? nestedField || isLabelTag
+          ? "filter"
+          : "match"
+        : nestedField || isLabelTag
+        ? "negativeFilter"
+        : "negativeMatch"
+      : !excluded
+      ? "visible"
+      : "notVisible";
   });
```
manivoxel51, line 192: nit: imo the if/else is more readable 
```
if (isFilterMode) {
  if (!excluded) {
    return nestedField || isLabelTag ? "filter" : "match";
  } else {
    return nestedField || isLabelTag ? "negativeFilter" : "negativeMatch";
  }
} else {
  return !excluded ? "visible" : "notVisible";
}
```

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -208,32 +253,95 @@ const FilterOption: React.FC<Props> = ({
...
+  };
+
+  const onSelectNotVisible = () => {
+    !excluded && setExcluded(true);
+    onlyMatch && setOnlyMatch(false);
+  };
+
+  useEffect(() => {
+    // when swtiching the mode, update the key to reflect the current mode selection
+    if (isFilterMode && ["visible", "notVisible"].includes(key)) {
```
manivoxel51, line 314: nit: constants for "visible", "notVisible"

app/packages/core/src/components/Sidebar/Entries/GroupEntries.tsx
```diff
@@ -61,6 +66,35 @@ const numGroupFieldsFiltered = selectorFamily<
     },
 });
 
+const numGroupFieldsVisible = selectorFamily<
```
manivoxel51, line 69: could add unit testing for this - separate PR is cool with me

app/packages/core/src/components/Sidebar/Entries/GroupEntries.tsx
```diff
@@ -194,7 +228,7 @@ const useClearFiltered = (modal: boolean, group: string) => {
   return useRecoilCallback(
     ({ set, snapshot }) =>
       async () => {
-        let paths = await snapshot.getPromise(
+        const paths = await snapshot.getPromise(
```
manivoxel51, line 231: is this change needed?

app/packages/core/src/components/Sidebar/Entries/GroupEntries.tsx
```diff
@@ -210,6 +244,26 @@ const useClearFiltered = (modal: boolean, group: string) => {
   );
 };
 
+const useClearVisibility = (modal: boolean, group: string) => {
```
manivoxel51, line 247: unit test here would be good too

app/packages/core/src/components/Sidebar/Entries/utils.tsx
```diff
@@ -8,6 +8,33 @@ export const InputDiv = styled.div`
...
   user-select: none;
   padding-top: 2px;
+  margin: 0 0.25rem 0 1rem;
+
+  display: flex;
+  justify-content: space-between;
+  & > input {
+    color: ${({ theme }) => theme.text.secondary};
+    font-size: 14px !important;
+    font-size: 1rem;
```
manivoxel51, line 18: same properties

app/packages/state/src/recoil/fieldVisibility.ts
```diff
@@ -0,0 +1,95 @@
+import { atom, selectorFamily } from "recoil";
```
manivoxel51, line 1: this file also very unit testable

app/packages/state/src/recoil/fieldVisibility.ts
```diff
@@ -0,0 +1,95 @@
...
+import { expandPath, fields } from "./schema";
+import { State } from "./types";
+import { hiddenLabelIds } from "./selectors";
+
+export const modalFieldVisibility = atom<State.Filters>({
+  key: "modalFieldVisibility",
+  default: {},
+});
+
+export const fieldVisibility = atom<State.Filters>({
```
manivoxel51, line 14: the naming is very very similar to our Original Field Visibility feature. :)

manivoxel51 (commented):
Nice progress Lanny! 🍨

I am almost half way done with the code review. have not tested extensively yet.

- just confirmed that the saved views seem to respect the new filters as well which is great.

- We need to make this work with Field Visibility feature. For example, when you use the new filter cat `predictions`, then click on the cog icon to open field visibility modal and deselect `predictions`. click `Apply` and app crashes. we can brainstorm how to approach this offline.

Also another point is what should the `Reset` button do in Field visibility modal? should it clear all filters applied through this new mechanism. 🤔  this is more minor.

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
```diff
@@ -208,32 +253,95 @@ const FilterOption: React.FC<Props> = ({
...
+    excluded && setExcluded(false);
+    onlyMatch && setOnlyMatch(false);
+  };
+
+  const onSelectNotVisible = () => {
+    !excluded && setExcluded(true);
+    onlyMatch && setOnlyMatch(false);
+  };
+
+  useEffect(() => {
```
benjaminpkane, line 312: Can this initialization effect be avoided? Ideally this state has a default automatically, and is updated through explicit user interaction.

lanzhenw:
> Nice progress Lanny! 🍨
> 
> I am almost half way done with the code review. have not tested extensively yet.
> 
> * just confirmed that the saved views seem to respect the new filters as well which is great.
> * We need to make this work with Field Visibility feature. For example, when you use the new filter cat `predictions`, then click on the cog icon to open field visibility modal and deselect `predictions`. click `Apply` and app crashes. we can brainstorm how to approach this offline.
> 
> Also another point is what should the `Reset` button do in Field visibility modal? should it clear all filters applied through this new mechanism. 🤔 this is more minor.

Modifying the field visibility panel may make current filters not valid any more after applying the changes. 
We should give the user a warning ("Your current sidebar filter/visibility settings will be cleared.") and clear the sidebar filter/visibility settings before applying/resetting field changes.

app/packages/state/src/recoil/fieldVisibility.ts
lanzhenw: The sidebar features are working on the attribute level, so I just updated the names to attributeVisibility

app/packages/core/src/components/Common/Checkbox.tsx
lanzhenw: will setup an `omitAggregation` prop as the flag

benjaminpkane: Omitting the `countAtom` prop has the same effect

app/packages/core/src/components/Filters/NumericFieldFilter.tsx
```diff
@@ -167,6 +168,9 @@ const NumericFieldFilter = ({
   });
 
   const isFiltered = useRecoilValue(fos.fieldIsFiltered({ modal, path }));
+  const hasVisibilitySetting = useRecoilValue(
```
benjaminpkane, line 171: Letting visibility setting be part of `fos.fieldIsFiltered` flag selector would avoid any changes to this component. Might be cleaner

app/packages/core/src/components/Filters/categoricalFilter/ResultComponent.tsx
```diff
@@ -1,7 +1,10 @@
+import * as fos from "@fiftyone/state";
 import React from "react";
+import { useRecoilValue } from "recoil";
 import { V } from "./CategoricalFilter";
 
 const ResultComponent = ({ value: { value, count } }: { value: V }) => {
+  const isFilterMode = useRecoilValue(fos.isSidebarFilterMode);
```
benjaminpkane, line 7: I would propose making `V["count"]` an optional type to avoid using this hook here. Just a suggestion

app/packages/core/src/components/Filters/categoricalFilter/filterOption/FilterOption.tsx
benjaminpkane: This hook always runs on on initial render and changes state. It was part of the reason for the bug fixed by https://github.com/voxel51/fiftyone/pull/3249

lanzhenw: Ah, I see what you mean. I can fix it.

app/packages/core/src/components/Filters/NumericFieldFilter.tsx
lanzhenw: Screenshot 2023-07-13 at 2 26 59 PM
Thought about this, but `fieldIsFiltered` atom is also used in the `numGroupFieldsFiltered` selectorFamily to calculate these status icons. We want to have separate active filter and visibility counts, soI kept separate atoms

brimoor:
@lanzhenw nice work! 🥇 (cc @manivoxel51 @benjaminpkane as well for thoughts)

Here's some feedback after using the feature a bit:

### Suggestions

- Remove ability to filter sidebar by typing in FILTER|VISIBILITY text area
    - We have the field visibility modal for that now; don't need two ways to do the same thing
- Add tooltips to FILTER|VISIBILITY to explain what the sidebar does and how the two modes differ. Here's some suggested text
    - FILTER: Use the controls below to create filtered views into your data
    - VISIBILITY: Use the controls below to toggle the visibility of fields in the grid
- FILTER|VISIBILITY modes do the exact same thing in the modal, only visibility mode doesn't show counts, so it's strictly less useful. I think what we really want is the following hybrid of the two modes:
    - Don't show FILTER|VISIBILITY toggle at the top. It's all the same
    - DO show label counts in the sidebar
    - Use Show/hide terminology in the filter widgets, rather than "select|exclude"

### Questions
- Looks like you changed the default for label list filters from "Select detections with label" to "Show samples with label". Was that intentional?

### Bugs
- Something is wrong with the filter mode selector in the modal; it says "match" with huge text. If I manually select a valid filter mode from the dropdown, then things work

Screen Shot 2023-07-13 at 5 04 50 PM

lanzhenw:
@brimoor Thanks for catching the bug with modal view bug. I fixed that, reverted default filter option back to "filter mode" for nested list field and implemented suggested changes. 
https://github.com/voxel51/fiftyone/assets/17770824/60438d4a-3bdc-4467-83fb-0bca5554bb35

brimoor:
@lanzhenw nice 👍 

Can you try left-aligning the `FILTER/VISIBILITY` text in grid mode like it used to be? Also, what do you think about making it so that clicking this text would also toggle between modes? I'm trying to think of ways to make the different modes as easy to discover as possible.

lanzhenw:
@brimoor Yes, that makes sense. Just updated. Also removed the top toggle from modal view. 

https://github.com/voxel51/fiftyone/assets/17770824/a39c3fcc-6236-4ae9-878b-bd52d0628744

app/packages/state/src/recoil/pathFilters/boolean.ts
```diff
@@ -20,6 +22,28 @@ const getFilter = (
...
   modal: boolean,
   path: string
+): BooleanFilter => {
+  // nested listfield, label tag and modal use "isMatching: false" default
+  const fieldPath = path.split(".").slice(0, -1).join(".");
+  const fieldSchema = get(schemaAtoms.field(fieldPath));
+  const isNestedfield = fieldSchema?.ftype.includes("ListField");
+  const defaultToFilterMode = isNestedfield || modal || path === "_label_tags";
+  return {
+    _CLS: "bool",
```
benjaminpkane, line 32: `_CLS` is no longer needed

app/packages/core/src/components/Schema/SchemaSettings.tsx
```diff
@@ -107,7 +110,7 @@ const SchemaSettings = () => {
     setSelectedPaths({ [datasetName]: new Set(lastAppliedPaths.selected) });
     setExcludedPaths({ [datasetName]: new Set(lastAppliedPaths.excluded) });
   };
-
+  console.info("filters", filters);
```
benjaminpkane, line 113: Stray log

app/packages/core/src/components/Filters/NumericFieldFilter.tsx
```diff
@@ -242,6 +246,7 @@ const NumericFieldFilter = ({
             disabled={true}
             name={"No results"}
             setValue={() => {}}
+            omitAggregation={!isFilterMode}
```
benjaminpkane, line 249: This should be remove, right?

app/packages/core/src/components/Schema/SchemaSettings.tsx
```diff
@@ -218,8 +221,8 @@ const SchemaSettings = () => {
                 borderRadius: "4px",
               }}
               onClick={() => {
+                resetAttributeFilters();
```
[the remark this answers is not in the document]
manivoxel51: I will create a bug and tackle this separately - edge case

app/packages/core/src/components/Filters/categoricalFilter/CategoricalFilter.tsx
```diff
@@ -259,6 +264,24 @@ const CategoricalFilter = <T extends V = V>({
...
+              onSelect={onSelect}
+              inputStyle={{
+                color: theme.text.secondary,
+                fontSize: "1rem",
+                width: "100%",
+              }}
+              containerStyle={{ borderBottomColor: color, zIndex: 1000 }}
+              toKey={({ value }) => String(value)}
+            />
+          )}
```
benjaminpkane, line 284: This looks like it is duplicated from the above but only differs in its `placeholder` value? Can the `placeholder` just prop just be changed?

```ts
placeholder={`+ ${isFilterMode ? "filter by" : "set visibility"} by ${name}`}
```

app/packages/looker/src/state.ts
```diff
@@ -132,6 +132,7 @@ interface BaseOptions {
   coloring: Coloring;
   customizeColorSetting: CustomizeColor[];
   selectedLabels: string[];
+  attributeVisibility: object;
```
benjaminpkane, line 135: This should have a type. Can it be added, or a todo comment?

benjaminpkane: Conceptually, visibility could be built into the `filter` function and this option wouldn't be necessary. This seems like it could be decoupled from `Looker`

app/packages/state/src/hooks/useGroupEntries.ts
```diff
@@ -0,0 +1,129 @@
...
+        set(
+          fos.activeFields({ modal }),
+          active.filter((p) => !paths.includes(p))
+        );
+      },
+    [modal, group]
+  );
+};
+
+export const useClearFiltered = (modal: boolean, group: string) => {
```
manivoxel51, line 91: nice!

app/packages/state/src/recoil/pathFilters/boolean.ts
```diff
@@ -240,41 +293,123 @@ export const booleanSelectedValuesAtom = selectorFamily<
...
+    : values.every((v) => v == value) ||
+      (noneValue && NONE.has(value) && Boolean(value));
 
-      return (value) => {
-        if (isMatching) {
-          return true;
+  return r;
+};
+
+export const generateBooleanSelectorFamily = (key) =>
```
manivoxel51, line 321: nit: large selector... can we break it down to smaller functions

app/packages/state/src/recoil/pathFilters/string.ts
```diff
@@ -135,51 +199,169 @@ export const string = selectorFamily<
   get:
     (params) =>
     ({ get }) => {
-      if (!get(filterAtoms.filter(params))) {
-        return (value) => true;
+      // Extract common properties
+      const filter = get(filterAtoms.filter(params));
```
manivoxel51, line 203: nit: might be room to refactor

app/packages/state/tests/hooks/useGroupEntries.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+  modalAttributeVisibility,
+  sidebarGroupsDefinition,
+} from "../../src";
+import {
+  useClearFiltered,
+  useClearVisibility,
+} from "../../src/hooks/useGroupEntries";
+import { filters, modalFilters } from "../../src/recoil/filters";
+
+const Root1: React.FC<React.PropsWithChildren<{}>> = ({ children }) => {
```
manivoxel51, line 16: Molto bene! 👍

manivoxel51 (commented):
@lanzhenw awesome work! 🍨 

tested locally and really like the separation. Also, your selectors are looking clean and sharp! 🦈 easily unit-testable ;) 

non-blocking thought that came to mind while testing the feature was that it is difficult to clear all the filters or the visibility choices at once. I believe this is because we don't have a top level clear button. might be worth adding one in later time. A related design thought is how we would coordinate this new "clear all" button with the field visibility's clear button 🤔

app/packages/looker/src/state.ts
lanzhenw: @benjaminpkane  We are actually using the `filter` function to determine which tag/overlay to render. I just need a flag to determine whether to reload the overlays for visibility changes because filters always trigger grids reload. That's why I added `attributeVisibility` separately.

benjaminpkane: Ok I read the code in `tags.ts`. I see why it's used. At some point we will need to move `tags.ts` to canvas to avoid equality checks and formalize rendering as an overlay only feature.

manivoxel51 (approved):
thank you! 🍨 

we will be writing e2e for couple of scenarios in this PR in a separate ongoing work

PR #3275, opened 2023-07-10
Suppress errors due to none fields

kaixi-wang:
Avoid throwing hard errors that prevent loading a dataset in the app due to a missing field name


benjaminpkane (approved):
LGTM

PR #3293, opened 2023-07-13
Label conversion bugfix/tests

brimoor:
Fixes a bug when passing an RGB hex string to `Detection.to_segmentation(target=...)`.

Also adds some unit tests for Detection <> Polyline <> Segmentation conversions.


benjaminpkane (approved):
LGTM

PR #3317, opened 2023-07-19
Dependency cleanup

### LINKED ISSUE #3298: [INSTALL] Some `setup.py` dependencies are not used.
### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04): Ubuntu 22.04
-   **Python version** (`python --version`): 3.9.17
-   **FiftyOne version** (`fiftyone --version`): 0.21
-   **FiftyOne installed from** (pip or source): pip/poetry

### Describe the problem

My main issue was related to `eventlet` library, so I started looking where it's used. When I didn't find a place where it's used I decided to go through `setup.py` to find other unused ones. Here's the list with comments:

- `eventlet` - not used in the project but causes issues when it's installed; for example, `FastAPI` crashes during file downloading with a traceback pointing to a file descriptor used by a `socket` library (which is basically `asyncio` internals), and that makes sense because `eventlet` is just a hack to make sync code async. Also I saw the same issues with `s3fs` library.
- `future` - this library is for Python2/3 compatibility; not sure if this is needed since the minimum version of Python this project supports is 3.7.
- `kaleido` - `kaleido` after installation exports the same symbol `plotly` which I believe is being overwritten with the `plotly` library.

**Question**: Am I missing something or these dependencies are really not used in the project? If so is it possible to remove them? (I can make a PR)

benjaminpkane:
Resolves #3298


imanjra (approved):

requirements/common.txt
```diff
@@ -4,14 +4,12 @@ boto3==1.17.36
...
 dacite==1.6.0
 Deprecated==1.2.11
-eventlet==0.31.0
 Flask==1.1.2
 ftfy==6.1.1
 httpx==0.23.0
 hypercorn==0.13.2
 importlib-metadata==1.4.0; python_version<"3.8"
 Jinja2==3.0.3
-kaleido==0.2.1
```
brimoor, line 14: `kaleido` is required to export plotly plots as static images:
https://github.com/voxel51/fiftyone/blob/3c0cd9d9ab596ad18506bfeb02a3328638eb8832/fiftyone/core/plots/plotly.py#L1583

Reference
https://plotly.com/python/static-image-export/

brimoor (changes requested):

PR #3320, opened 2023-07-20
Fixed warn condition plugin component registration

Sa-Schmi:

Changed the ok-condition from equal to not equal. As `warn()` prints the message if the ok-condition is false and we want the ok-condition to be false if `registration.type` equals `PluginComponentType.Plot`.

Compiled from source and warning `_______ is a Plot Plugin Component. This is deprecated. Please use "Panel" instead.`
Is gone for plugins which are not a Plot Plugin Component.

-   [ x ] No. 

-   [ x ] App: FiftyOne application changes
-   [ ] Other

### DIFF
--- a/app/packages/plugins/src/index.ts
+++ b/app/packages/plugins/src/index.ts
@@ -318,7 +318,7 @@ class PluginComponentRegistry {
       `${name} is already a registered Plugin Component`
     );
     warn(
-      registration.type === PluginComponentType.Plot,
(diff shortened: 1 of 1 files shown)


imanjra (approved):

imanjra:
LGTM. Thank you @Sa-Schmi!

Sa-Schmi:
My bad, I updated my fork...

PR #3323, opened 2023-07-20
Resolve errors with Labelbox integration

### LINKED ISSUE #2654: [FR] Update Labelbox integration to handle deprecation of datasets
Labelbox is moving away from the "dataset" methodology for structuring data and focusing on batches. Our integration heavily uses labelbox datasets and should be updated to handle this migration: https://docs.labelbox.com/reference/migrating-to-workflows#how-do-these-changes-affect-me

ehofesmann:
Labelbox has now fully deprecated "Queue mode" in favor of "batch mode". Previously, attempting to use the Labelbox integration would result in an error, this PR updates the integration to use batch mode and brings the integration back to its previous state.

Resolves #2654

### DIFF
--- a/fiftyone/utils/labelbox.py
+++ b/fiftyone/utils/labelbox.py
@@ -486,7 +486,7 @@ class LabelboxAnnotationAPI(foua.AnnotationAPI):
         """Uploads the media for the given samples to Labelbox.
 
         This method uses ``labelbox.schema.dataset.Dataset.create_data_rows()``
-        to add data in batches, and sets the external ID of each DataRow to the
+        to add data in batches, and sets the global key of each DataRow to the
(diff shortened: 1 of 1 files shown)


fiftyone/utils/labelbox.py
```diff
@@ -594,7 +610,9 @@ def download_annotations(self, results):
             metadata = self._get_sample_metadata(project, sample_id)
             if metadata is None:
                 logger.warning(
-                    "Skipping sample '%s' with no metadata", sample_id
+                    "Skipping sample '%s' with no metadata, likely due to no "
```
brimoor, line 613: typo in the message here

brimoor (approved):
Thanks for patching! ⚒️🏅

PR #3326, opened 2023-07-21
Embeddings only color-by supported list fields

brimoor:
## Before

`ground_truth.detections` shouldn't be shown as a choice:

Screen Shot 2023-07-20 at 9 37 05 PM

## After

Now the only allowed list field is `tags`, which is valid:

Screen Shot 2023-07-20 at 9 36 27 PM

### DIFF
--- a/fiftyone/server/routes/embeddings.py
+++ b/fiftyone/server/routes/embeddings.py
@@ -25,7 +25,6 @@ COLOR_BY_TYPES = (
     fof.BooleanField,
     fof.IntField,
     fof.FloatField,
-    fof.ListField,
 )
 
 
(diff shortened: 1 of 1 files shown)


fiftyone/server/routes/embeddings.py
```diff
@@ -25,7 +25,6 @@
     fof.BooleanField,
     fof.IntField,
     fof.FloatField,
```
lanzhenw, line 27: we don't support customize color by FloatField through ColorScheme in the app yet. Do we want to be consistent here?

brimoor: Coloring by float is useful (and fully supported) by the embeddings panel. It will show a continuous colorbar.

I believe the colorscale is pulled from dataset.app_config.colorscale, which is also used by Heatmap fields

lanzhenw (approved):

PR #2843, opened 2023-03-27
adding the ability to use any device trough parameter + apple's MPS automatic check

timmermansjoy:
## What changes are proposed in this pull request?

added the functionality to do inference on the GPU on apple devices with the [metal backend in pytorch](https://pytorch.org/blog/introducing-accelerated-pytorch-training-on-mac/)

(Please fill in changes proposed in this fix)

changed up the `_using_gpu` so it also checks for cuda and mps and then after also make the images go to `self._device` instead of going to `cuda()` directly

## How is this patch tested? If it is not, please explain why.

tested on some models in the zoo, I can not test if it conflicts on other devices 

(Details)

## Release Notes

for Mac users have the ability to utilise more of their available hardware by utilising the MPS backend

### Is this a user-facing change that should be mentioned in the release notes?

-   [x] No. You can skip the rest of this section.
-   [ ] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

(Details in 1-2 sentences. You can just refer to another PR with a description
if this PR is part of a larger change.)

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

PR #3328, opened 2023-07-21
Only use MPS when manually specified

brimoor:
#2843 added MPS support. Unfortunately MPS does not yet support certain ops, so some zoo models like `clip-vit-base32-torch` don't support it.

This PR reverts to the previous behavior of using CUDA if available, else falling back to CPU.

MPS can still be requested by manually passing `device="mps"`:

```py
import fiftyone.zoo as foz

# Use default device (CUDA if available, else CPU)
model = foz.load_zoo_model("...")

# Force a specific device
model = foz.load_zoo_model("...", device="mps")
```


jacobmarks (approved):
LGTM and code works for me

PR #3329, opened 2023-07-21
Add media factory for e2e 

sashankaryal:

Allows creation of blank images and videos for testing purposes. This way we don't have to load a zoo dataset over network.

Locally.

        notes for FiftyOne users.

-   [ ] Other


benjaminpkane:
Adding a media factory for pcd files would be valuable. Is that feasible?

sashankaryal:
> Adding a media factory for pcd files would be valuable. Is that feasible?

Is Is!
I'll add something.

PR #3333, opened 2023-07-24
Sparse groups and e2e coverage

benjaminpkane:
Adding support and graceful handling of group datasets where only the current grid slice is guaranteed to be present when navigating through samples in the modal.


benjaminpkane:
Multiple media types will be a separate PR

e2e-pw/src/oss/fixtures/loader.ts
```diff
@@ -122,14 +122,22 @@ export class OssLoader extends AbstractFiftyoneLoader {
       }
     };
 
-    if (!datasetName) {
-      await forceDatasetFromSelector();
+    if (savedView) {
```
manivoxel51, line 125: 👍 soon we will have other params. should we append param instead

e2e-pw/src/oss/poms/modal.ts
```diff
@@ -29,15 +41,57 @@ export class ModalPom {
...
+    return this.waitForSampleToLoad(allowErrorInfo);
+  }
+
+  async navigateSlice(
+    groupField: string,
+    slice: string,
+    allowErrorInfo = false
+  ) {
+    const currentSlice = await this.sidebarPom.getSidebarEntryText(
+      `sidebar-entry-${groupField}.name`
```
manivoxel51, line 66: nit: should this be `sidebar-entry-${groupField}-name` with a dash

e2e-pw/src/oss/specs/groups/sparse-groups-scenes.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+            )
+    dataset.add_samples(samples)
+    dataset.save_view("dynamic", dataset.group_by("dynamic", order_by="id"))`);
+});
+
+test.beforeEach(async ({ page, fiftyoneLoader }) => {
+  await fiftyoneLoader.waitUntilLoad(page, datasetName);
+});
+
+test(`ego default group slice transitions`, async ({ grid, modal }) => {
```
manivoxel51, line 56: Very nice!

manivoxel51 (approved):
Nice work @benjaminpkane ! 🍨 

lots of tests 🎉

app/packages/state/src/recoil/groups.ts
```diff
@@ -37,114 +47,125 @@ export const isGroup = selector<boolean>({
...
+      .map(({ name }) => name);
+  },
+});
+
+export const allNonPcdSlices = selector<string[]>({
+  key: "allNonPcdSlices",
+  get: ({ get }) => {
+    return get(groupMediaTypes)
+      .filter(
+        ({ mediaType }) => !["point-cloud", "point_cloud"].includes(mediaType)
```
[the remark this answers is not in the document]
benjaminpkane: I will remove `point-cloud`. No longer valid

e2e-pw/src/oss/poms/modal.ts
benjaminpkane: Ok, we would need to follow that formatting in the react component. But that would create collisions, though this is only for testing

Issue #3334, opened 2023-07-24
[BUG] Issue with cloning group datasets


benjaminpkane:
Selecting fields and cloning seems to copy unwanted fields into the clone. This issue only applies to group datasets
```py
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")
dataset = dataset.select_fields().clone("image-clone")
dataset.first() # Works

dataset = foz.load_zoo_dataset("quickstart-groups")
dataset = dataset.select_fields().clone("group-clone")
dataset.first() # Unexpected ground truth data
```

Traceback
```
Traceback (most recent call last):
  File "/Users/benjaminkane/code/fiftyone/fiftyone/core/dataset.py", line 6406, in _sample_dict_to_doc
    return self._sample_doc_cls.from_dict(d, extended=False)
  File "/Users/benjaminkane/code/fiftyone/fiftyone/core/odm/document.py", line 452, in from_dict
    return cls._from_son(d)
  File "/Users/benjaminkane/venvs/fo/lib/python3.10/site-packages/mongoengine/base/document.py", line 837, in _from_son
    obj = cls(__auto_convert=False, _created=created, **data)
  File "/Users/benjaminkane/venvs/fo/lib/python3.10/site-packages/mongoengine/base/document.py", line 99, in __init__
    raise FieldDoesNotExist(msg)
mongoengine.errors.FieldDoesNotExist: The fields "{'ground_truth'}" do not exist on the document "samples.64bea2e5c505526444f4d0c0"

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/benjaminkane/code/scratch/mult-pcd.py", line 5, in <module>
    dataset.first()
  File "/Users/benjaminkane/code/fiftyone/fiftyone/core/dataset.py", line 1114, in first
    return super().first()
  File "/Users/benjaminkane/code/fiftyone/fiftyone/core/collections.py", line 712, in first
    return next(iter(self))
  File "/Users/benjaminkane/code/fiftyone/fiftyone/core/dataset.py", line 2183, in iter_samples
    for sample in samples:
  File "/Users/benjaminkane/code/fiftyone/fiftyone/core/dataset.py", line 2199, in _iter_samples
    sample = make_sample(d)
  File "/Users/benjaminkane/code/fiftyone/fiftyone/core/dataset.py", line 2212, in make_sample
    doc = self._sample_dict_to_doc(d)
[...]

PR #3336, opened 2023-07-25
Reapply schema changes after selecting group slices

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/3334.

`select_group_slices()` emits unfiltered samples, since it applies a fresh `$lookup` to retrieve slice data. However, the current implementation does *intend* maintain any schema changes (select/exclude fields) that were applied prior to the `select_group_slices()` stage. https://github.com/voxel51/fiftyone/issues/3334 identified a gap in the enforcement of this convention, which this PR patches.

## Notes

This PR is pretty ugly... which is a symptom of the tension between the inconsistency of *maintaining* schema changes while *not maintaining* field filters when selecting group slices.

Here's what it would look like to enforce a convention that `select_group_slices()` does *not* maintain schema changes (for consistency with how it does *not* maintain any field filters):

```py

### Case 1

# Presumably this should *not* contain fields that were selected above
dataset2 = view.clone()

### Case 2

# If flat_view does *not* maintain the schema change...
flat_view = view.select_group_slices()

# ... then this dataset *would* contain all original fields
dataset3 = flat_view.clone()
```

The trouble with this is that Case 1 and Case 2 have different behaviors, but `view.select_group_slices(_allow_mixed=True)` is used internally to implement Case 1...

An alternative attempt was made in https://github.com/voxel51/fiftyone/pull/3337, but it was closed because it was going to take more work to implement and was not necessarily what users would want anyway.

### DIFF
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -4518,7 +4518,7 @@ class SelectGroupSlices(ViewStage):
         elif slices is not None:
             expr &= F(name_field) == slices
 
-        return [
+        pipeline = [
             {"$project": {group_field: True}},
             {
                 "$lookup": {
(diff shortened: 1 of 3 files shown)


benjaminpkane (approved):
LGTM

PR #3338, opened 2023-07-25
App supports bazel

mys007:

When fiftyone is installed under bazel environment and `fo.launch_app()` is called, the App does not launch and only 404 is displayed. 

This is because bazel creates virtual package installations, where everything - including `fiftyone/server/static` - is just a symbolic link to a central, shared location. The story with bazel is also mentioned in https://github.com/encode/starlette/issues/1083 and was one of the motivation for introducing `StaticFiles(follow_symlink=True)` for this purpose.

This PR turns on the support for symlinks for the `static` directory, so that fiftyone works with bazel.

Tested locally.

        notes for FiftyOne users.

-   [ ] Other


benjaminpkane (approved):
LGTM! Thanks @mys007!

PR #3345, opened 2023-07-26
Delegated operators CLI

brimoor:
## Change log

- Adds a `fiftyone delegated` CLI command that supports a variety of useful actions related to delegated operators
- Adds a `foo.execute_operator()` method that supports programmatically execution of operators

## CLI examples

Launch a local service that will run all delegated operations:

```shell
fiftyone delegated launch
```

List delegated operations:

```shell
# List all operations
fiftyone delegated list

# List specific operations
fiftyone delegated list \
    --dataset quickstart \
    --operator @voxel51/io/export_samples \
    --state COMPLETED \
    --sort-by COMPLETED_AT \
    --limit 2
```

Cleanup delegated operations that you no longer want to keep records for:

```shell
# Delete all failed operations associated with a given dataset
fiftyone delegated cleanup --dataset quickstart --state FAILED

# Delete all delegated operations associated with non-existent datasets
fiftyone delegated cleanup --orphan

# Print information about operations rather than actually deleting them
fiftyone delegated cleanup --orphan --dry-run        
```

## Programmatically run an operation

```py
import fiftyone as fo
import fiftyone.zoo as foz

ctx = dict(dataset=dataset)

params = dict(
    target="DATASET",
    export_dir="/tmp/coco",
    export_type="LABELS_ONLY",
    dataset_type="COCO",
    label_field="ground_truth",
)

foo.execute_operator("@voxel51/io/export_samples", ctx, params)
```

### DIFF
--- a/fiftyone/core/cli.py
+++ b/fiftyone/core/cli.py
@@ -2777,7 +2777,8 @@ class DelegatedListCommand(Command):
         fiftyone delegated list \\
             --dataset quickstart \\
             --operator @voxel51/io/export_samples \\
-            --status COMPLETED \\
+            --state COMPLETED \\
+            --sort-by COMPLETED_AT \\
             --limit 10
     """
 
--- a/fiftyone/core/cli.py
+++ b/fiftyone/core/cli.py
@@ -2903,15 +2903,14 @@ def _print_delegated_list(ops):
 
     rows = []
     for op in ops:
-        state = _parse_state(op.run_state)
         rows.append(
             {
                 "id": op.id,
                 "operator": op.operator,
                 "dataset": op.context.request_params.get("dataset_name", None),
                 "queued_at": op.queued_at,
-                "state": op.run_state,
-                "completed": state == ExecutionRunState.COMPLETED,
+                "state": op.run_state.value,
+                "completed": op.run_state == ExecutionRunState.COMPLETED,
             }
         )
 
@@ -3058,8 +3057,7 @@ def _cleanup_delegated(operator=None, dataset=None, state=None, dry_run=False):
 
     del_ids = set()
     for op in ops:
-        state = _parse_state(op.run_state)
-        if state != ExecutionRunState.RUNNING:
+        if op.run_state != ExecutionRunState.RUNNING:
             del_ids.add(op.id)
 
     num_del = len(del_ids)
--- a/fiftyone/core/cli.py
+++ b/fiftyone/core/cli.py
@@ -30,6 +30,7 @@ import fiftyone.core.session as fos
 import fiftyone.core.utils as fou
 import fiftyone.migrations as fom
 import fiftyone.operators as foo
+import fiftyone.operators.delegated as food
 import fiftyone.plugins as fop
 import fiftyone.utils.data as foud
 import fiftyone.utils.image as foui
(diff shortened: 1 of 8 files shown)


fiftyone/core/cli.py
```diff
@@ -2702,6 +2704,372 @@ def _print_operator_info(operator_uri):
...
+        parser.print_help()
+
+
+class DelegatedLaunchCommand(Command):
+    """Launches a service for running delegated operations.
+
+    Examples::
+
+        # Launch a local service
+        fiftyone delegated launch --type local
```
brimoor, line 2729: @nebulae my thinking was to set this up so that something like this could be possible in the future:

```
fiftyone delegated launch --type airflow
```

fiftyone/core/cli.py
```diff
@@ -2702,6 +2704,372 @@ def _print_operator_info(operator_uri):
...
+
+        _print_delegated_list(ops)
+
+
+def _parse_state(state):
+    from fiftyone.operators.executor import ExecutionRunState
+
+    if state is not None:
+        try:
+            state = ExecutionRunState[state.upper()]
```
brimoor, line 2856: @nebulae I had to add this boilerplate for `run_state`, `sort_by`, `sort_direction` to convert strings to enums. Also elsewhere in the code there has to be `.value` append to enums when the string is desired.

It's old school, but what about implementing "enums" just as classes:

```python
class ExecutionRunState(object):
    RUNNING = "running"
    COMPLETED = "completed"
    ...
```

so that these methods and run docs themselves could just directly use strings?

fiftyone/factory/__init__.py
```diff
@@ -8,26 +8,28 @@
...
 
 
-class DelegatedOpPagingParams(object):
-    class SortByField(Enum):
-        QUEUED_AT = "queued_at"
-        COMPLETED_AT = "completed_at"
-        STARTED_AT = "started_at"
-        FAILED_At = "failed_at"
-        OPERATOR_NAME = "operator"
+class SortByField(Enum):
```
brimoor, line 11: Needed access to these to write CLI methods

fiftyone/factory/__init__.py
```diff
@@ -8,26 +8,28 @@
...
 
+class DelegatedOpPagingParams(object):
     def __init__(
         self,
-        skip: int = 0,
-        limit: int = 10,
         sort_by: SortByField = SortByField.QUEUED_AT,
         sort_direction: SortDirection = SortDirection.DESCENDING,
+        skip: int = 0,
+        limit: int = 10,
```
brimoor, line 30: @nebulae what about making these be `None` by default? The fact that App usages of these params will likely want paging doesn't mean that programmatic usages will necessarily want paging

fiftyone/factory/repos/delegated_operation.py
```diff
@@ -181,12 +180,15 @@ def list_operations(
...
 
         if paging:
-            docs = (
-                self._collection.find(query)
-                .skip(paging.skip)
-                .limit(paging.limit)
-                .sort(paging.sort_by.value, paging.sort_direction.value)
-            )
+            docs = self._collection.find(query)
+            if paging.sort_by.value:
```
brimoor, line 184: sort by should happen first, then skip/limit

PR #3347, opened 2023-07-27
Use weights parameter instead of deprecated pretrained for torchvision models

helioshe4:

- Use weights parameter when loading torchvision models for people with torchvision >= 0.13.0
- Use pretrained=true parameter for users with torchvision < 0.13.0.  
- Backwards compatible for users with older versions of torchvision.

Downloaded various torchvision models using torchvision==0.15.2 and then using torchvision==0.12.0.
Tested model with no entrypoint_args.

        notes for FiftyOne users.

-   [ ] Other


fiftyone/zoo/models/torch.py
```diff
@@ -93,7 +93,12 @@ def _download_model(self, config):
 
     def _load_network(self, config):
         entrypoint = etau.get_function(config.entrypoint_fcn)
-        kwargs = config.entrypoint_args or {}
+
+        if version.parse(torchvision.__version__) < version.parse("0.13.0"):
+            kwargs = {"pretrained": True} if config.entrypoint_args else {}
```
brimoor, line 98: For extra security, how about only pop-ing `weights'` and replacing it with `pretrained`?

brimoor (commented):
@helioshe4 this is great, thanks! I left one comment for your consideration.

Also, could you update the destination branch for the PR to `release/v0.21.5` so we can include this patch in the next release?? 😄

fiftyone/zoo/models/torch.py
helioshe4: Sure, just pushed the new changes, hopefully I understood correctly.  I've also changed thee destination branch now.  Thanks for the feedback

helioshe4: Created new PR off destination branch `release/v0.21.5` at #3348

PR #3348, opened 2023-07-27
Use weights parameter instead of deprecated pretrained for torchvision models

helioshe4:

- Use weights parameter when loading torchvision models for people with torchvision >= 0.13.0
- Use pretrained=true parameter for users with torchvision < 0.13.0.  
- Backwards compatible for users with older versions of torchvision.

Downloaded various torchvision models using torchvision==0.15.2 and then using torchvision==0.12.0.
Tested model with no entrypoint_args.

        notes for FiftyOne users.

-   [ ] Other


brimoor (approved):
LGTM, thanks for the contribution! 🎉

PR #3349, opened 2023-07-27
Ensure has_field() supports dynamic fields

### LINKED ISSUE #3346: [BUG] Document `has_field` does not work for non-default field names
### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04): Ubuntu 20.04.3 LTS
-   **Python version** (`python --version`): Python 3.10.2
-   **FiftyOne version** (`fiftyone --version`): 0.21.4

### Commands to reproduce

Running the following in an `ipython` shell should demonstrate the problem with `has_field`. 

```
import fiftyone as fo
# Create an empty detection
det = fo.Detection()
# Add a custom field to the detection
det['asdf'] = 1.0
# Check with has_field
print(f"Has asdf: {det.has_field('asdf')}")
# Check with `in`
print(f"Has asdf (in): {'asdf' in det}")
```

### Describe the problem

In previous versions, custom fields on a `Detection` would work with `has_field`.  Since updating to version 0.21.4, this no longer works for detections regardless of saved status.

### Code to reproduce issue

See above.  Can be demonstrated with the quickstart dataset as well after saving samples

### Willingness to contribute

-   [x ] Yes. I would be willing to contribute a fix for this bug with guidance from the FiftyOne community

brimoor:
Resolves #3346

### DIFF
--- a/fiftyone/core/odm/document.py
+++ b/fiftyone/core/odm/document.py
@@ -297,16 +297,15 @@ class MongoEngineBaseDocument(SerializableDocument):
         return self._get_field_names(include_private=False)
 
     def has_field(self, field_name):
-        # pylint: disable=no-member
-        chunks = field_name.split(".", 1)
-        if len(chunks) > 1:
-            field = self._fields.get(chunks[0], None)
-            try:
-                return field.has_field(chunks[1])
-            except AttributeError:
-                return False
+        try:
+            chunks = field_name.split(".", 1)
+            if len(chunks) > 1:
+                return self._get_field(chunks[0]).has_field(chunks[1])
 
-        return field_name in self._fields
+            _ = self._get_field(field_name)
+            return True
+        except (KeyError, AttributeError):
+            return False
 
     def get_field(self, field_name):
         chunks = field_name.split(".", 1)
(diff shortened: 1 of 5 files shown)


fiftyone/core/odm/document.py
```diff
@@ -297,16 +297,15 @@ def field_names(self):
...
-        if len(chunks) > 1:
-            field = self._fields.get(chunks[0], None)
-            try:
-                return field.has_field(chunks[1])
-            except AttributeError:
-                return False
+        try:
+            chunks = field_name.split(".", 1)
+            if len(chunks) > 1:
+                return self._get_field(chunks[0]).has_field(chunks[1])
```
swheaton, line 303:
```suggestion
```
I think you can rely on your new `self._get_field(field_name)` to do this initial split for you.

swheaton (changes requested):
lgtm and seems to work as advertised!
Just one minor comment

fiftyone/core/odm/document.py
brimoor: Good call, will add 👍

swheaton (approved):
lgtm

PR #3351, opened 2023-07-27
Add support for viewing group slices within embedding panel

ritch:
Uses the current group slice to properly build a view for loading plot data for the embeddings panel. With this change you will now see corresponding samples that match the following:

1. the currently selected group slice
2. exist within the currently selected brain key (visualization)

Note: even with this change there is no simple "1 click" ux for showing multiple slices, even if they meet criteria 2.

### DIFF
--- a/fiftyone/server/routes/embeddings.py
+++ b/fiftyone/server/routes/embeddings.py
@@ -60,8 +60,6 @@ class OnPlotLoad(HTTPEndpoint):
             ) % brain_key
             return {"error": msg}
 
-        print(stages)
-
         view = fosv.get_view(
             dataset_name,
             stages=stages,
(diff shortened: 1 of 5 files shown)


ritch:
Reminder: before merging this will need some regression testing against non-group embeddings plots.

fiftyone/server/routes/embeddings.py
```diff
@@ -39,7 +45,9 @@ def _post_sync(self, data):
         dataset_name = data["datasetName"]
         brain_key = data["brainKey"]
         stages = data["view"]
+        filters = data.get("filters", None)
```
ritch, line 48: this can be removed - it will always be `None`

PR #3353, opened 2023-07-28
[WIP] histogram crashes (group dataset) when statistics display option opens and group is selected

manivoxel51:

histogram crashes (group dataset) when statistics display option opens and group is selected
- [x] fix the python side - one liner
- [x] FE e2e test for this exact scenario
  - user opens a new panel, click on histogram, clicks back on Samples panel and clicks on Display Options (cog icon) - no crash should occure
  - [ ] Finalize the test with @sashankaryal's help. I am using the actionRows Pom that was created by the test stalls when calling `openDisplayOptions`. we can sync tomorrow

(Details)

        notes for FiftyOne users.

-   [ ] Other


e2e-pw/src/oss/specs/display-options/display-options.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    await panel.openNew();
+    await panel.open();
+    await actionsRow.openDisplayOptions();
+    await actionsRow.displayActions.setSidebarStatisticsMode("group");
+    await panel.open("histograms");
+
+    await panel.distributionContainer().waitFor({ state: "visible" });
+    await panel.distributionContainer().click();
+
+    await panel.assert.histogramLoaded();
```
benjaminpkane, line 44: Thinking out loud, adding an assertion (screenshot or HTML assertions) to test the contents of the distributions would be valuable. If more work is done on histograms in the future. I definitely think this captures the error, though 🚀

e2e-pw/src/oss/poms/panel.ts
```diff
@@ -0,0 +1,50 @@
+import { Locator, Page, expect } from "src/oss/fixtures";
+
+export class PanelPom {
```
benjaminpkane, line 3: 🚀

benjaminpkane (approved):
LGTM! 🚀

app/packages/spaces/src/types.ts
```diff
@@ -14,6 +14,7 @@ export type AddPanelItemProps = {
   Icon?: React.ComponentType;
   onClick?: () => void;
   spaceId: string;
+  dataCy?: string;
```
sashankaryal, line 17: same comment as above re: `dataCy` as prop

sashankaryal (approved):
looks good, minor comments

PR #3354, opened 2023-07-28
Plugin secrets

kaixi-wang:

(Details)

        notes for FiftyOne users.

-   [ ] Other


fiftyone/plugins/definitions.py
```diff
@@ -13,6 +13,8 @@
 
 import fiftyone as fo
 
+PLUGINS_DIR = fo.config.plugins_dir
```
brimoor, line 16: I think this redeclaration should be deleted and `fo.config.plugins_dir` should continue to be directly used everywhere:
- opens up the possibility that `fo.config.plugins_dir` can be modified in Python mid-session
- consistency with how all other config settings are used

brimoor (changes requested):

fiftyone/plugins/definitions.py
kaixi-wang: ok I can change it back... but wouldn't modifying the plugins_dir mid session be something we would want to prevent? It feels like it would introduce more opportunity for things to go wrong.

brimoor: If we wanted `fo.config` to be read-only then we would need to raise an error when this is attempted: `fo.config.plugins_dir = "new_value"`. Creating a secret copy is confusing.

The plugin registry is currently rebuilt from scratch every time it is requested, so pretty sure you can in fact modify it live.

Examples of config settings where dynamic modification is explicitly docmented:
https://docs.voxel51.com/user_guide/config.html#modifying-your-config-in-code

PR #3355, opened 2023-07-28
User can open color panel on sample modal view

lanzhenw:

User can access color panel from sample modal view through sidebar or color button.

https://github.com/voxel51/fiftyone/assets/17770824/a2af4543-8efd-440b-a265-d94647c5d1fc

(Details)

        notes for FiftyOne users.

-   [ ] Other


benjaminpkane (changes requested):
This is a nice addition. There are a few historically pieces here that need to be reconciled though. Historically, (before color schemes were added) the coloring settings in the modal were initialized with the current grid coloring settings. But they had their own state that could be modified without grid coloring changes.

It now seems we are heading to a global color scheme settings cc @brimoor. If so, some cleanup and reorganization needs to happen. I haven't tested thoroughly, but shuffle colors and label opacity don't apply to the modal until after closing the modal again and reopening.

Note that this is because the modal sidebar and looker(s) are still using the `modal=true` version of the recoil selectors/atoms. Pointing out some code, here is how (historical) color settings are currently initialized in `useExpandSample`
* https://github.com/voxel51/fiftyone/blob/29ae0a04024ec9754d612eff37bcb42babf7e003/app/packages/state/src/hooks/useExpandSample.ts#L26


In general, all of the coloring recoil state, i.e. path color, color seed, etc. should be updated to not be parameterized by the `modal` boolean to avoid tech debt

lanzhenw:
Updated the code. Now opacity, shuffle color, use multicolor keypoints and skeleton settings are consistent in the modal and grid view. 
https://github.com/voxel51/fiftyone/assets/17770824/9c007aa1-fb48-45b5-914f-903681c64eba

app/packages/state/src/hooks/useExpandSample.ts
```diff
@@ -34,9 +34,9 @@ export default <T extends Lookers>(store: LookerStore<T>) => {
             schemaAtoms.activeFields({ modal: false }),
           ],
           [atoms.cropToContent(true), atoms.cropToContent(false)],
-          [atoms.colorSeed(true), atoms.colorSeed(false)],
+          [atoms.colorSeed, atoms.colorSeed],
```
benjaminpkane, line 37: `[atoms.colorSeed, atoms.colorSeed]` and `[atoms.alpha, atoms.alpha]` can be removed. They have no effect

benjaminpkane (approved):
LGTM. I updated some coloring state so the modal parameter is no longer present to avoid future confusion

manivoxel51 (approved):
nice work! 🍨 

Couple of thoughts
- Should hovering over `Sample Tags` give us the color tooltip in the modal - similar to how tooltip shows in the grid.

In modal hover over Sample Tags
Screen Shot 2023-08-01 at 1 18 24 PM

In Grid you see tooltip
Screen Shot 2023-08-01 at 1 18 38 PM

- Super minor: When you hover over `predictions` and click on Custom Color icon, the modal opens correctly but the `predictions` tab (on the left) is not active/highlighted

PR #3358, opened 2023-07-28
various delegated operator improvements

ritch:


app/packages/operators/src/types.ts
```diff
@@ -942,6 +942,27 @@ export class MarkdownView extends View {
   }
 }
 
+/**
+ * Operator class for rendering a prompt.
+ */
+export class PromptView extends View {
+  constructor(
```
imanjra, line 949: For consistency, should we use options as the only param here?

imanjra (approved):

nebulae (approved):

PR #3375, opened 2023-08-02
Handle flashlight page errors

benjaminpkane:
#3333 introduces the shared pagination hook for samples (grid and carousel) `useFlashlightPager`. Error handling for pagination requests is missing in this new hook.

Group datasets do not have defined slices until samples are added, so I've added an e2e test that asserts an expected error from the grid when paginating over the undefined default slice.
```py
import fiftyone as fo
dataset = fo.Dataset("empty-group-dataset")
dataset.add_group_field("group", default="empty")
```
Screen Shot 2023-08-02 at 10 39 10 AM

### DIFF
--- a/e2e-pw/src/oss/fixtures/loader.ts
+++ b/e2e-pw/src/oss/fixtures/loader.ts
@@ -110,7 +110,12 @@ export class OssLoader extends AbstractFiftyoneLoader {
     throw new Error("Method not implemented.");
   }
 
-  async waitUntilLoad(page: Page, datasetName: string, savedView?: string) {
+  async waitUntilLoad(
+    page: Page,
+    datasetName: string,
+    savedView?: string,
+    withGrid = true
+  ) {
     const forceDatasetFromSelector = async () => {
       await page.goto("/");
       await page.getByTestId(`selector-Select dataset`).click();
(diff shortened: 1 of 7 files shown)


e2e-pw/src/oss/fixtures/loader.ts
```diff
@@ -145,8 +150,11 @@ export class OssLoader extends AbstractFiftyoneLoader {
       }
     }
 
-    await page.waitForSelector("[data-cy=flashlight-section]", {
-      state: "visible",
-    });
+    await page.waitForSelector(
```
manivoxel51, line 153: nit: I think you can use page.getByTestId('TEST_ID').waitFor({ state: 'visible'}) - not sure if it makes a difference

e2e-pw/src/oss/poms/panel.ts
```diff
@@ -39,12 +39,20 @@ export class PanelPom {
...
-  constructor(private readonly gridPom: PanelPom) {}
+  constructor(private readonly panelPom: PanelPom) {}
 
   async histogramLoaded() {
-    await expect(this.gridPom.distributionContainer()).toBeVisible();
+    await expect(this.panelPom.distributionContainer()).toBeVisible();
+  }
+
+  async hasError() {
+    await expect(this.panelPom.errorBoundary()).toBeVisible();
```
manivoxel51, line 56: nice

manivoxel51 (approved):
🍨

e2e-pw/src/oss/fixtures/loader.ts
benjaminpkane: Right, I saw that too. I hesitated because I thought @sashankaryal had a reason for this approach

sashankaryal: Appreciate the benefit of the doubt, Ben!
Let's just say I was learning Playwright. Feel free to change both of these to `getByTestId::waitFor`

PR #3376, opened 2023-08-02
Fix optmized sample pagination

benjaminpkane:
#3333 introduced the shared `useFlashlightPager` hook but omitted the `paginationData` flag for reduced sample sizes in the grid.

Note the `pcdSamples` selector requires the full sample because it is utilized by `Looker3d` and not grid or carousel elements.


manivoxel51 (approved):
🍨

PR #3382, opened 2023-08-03
Update pytube requirement to fix zoo YouTube download error

ehofesmann:
Some video datasets in the zoo like ActivityNet and Kinetics will optionally download the videos directly from the sources on YouTube using `pytube`. However, with PyTube versions <15, the video downloads would fail with this error:
```
Failed to download video 'https://www.youtube.com/watch?v=-9l1Rh10bO8': 'streamingData' 
```

[Updating the minimum supported Pytube version to >=15 resolves this issue.](https://stackoverflow.com/a/76268388/14277626)
This also lets us get rid of the patching code we were using for pytube v11.0.2.

There is also a new error that pytube throws which was causing issues with the error parsing in the zoo datasets, the error handling of kinetics/activitynet was updated to properly log this error now.

### DIFF
--- a/fiftyone/utils/activitynet.py
+++ b/fiftyone/utils/activitynet.py
@@ -371,7 +371,7 @@ class ActivityNetDatasetManager(object):
         return self.info.existing_split_sample_ids(split)
 
     def split_sample_ids(self, split):
-        return self.info.split_sample_ids[split]
+        return self.info.split_sample_ids(split)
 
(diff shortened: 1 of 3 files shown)


brimoor (approved):
LGTM 🥇

brimoor:
@ehofesmann can you update destination branch to `release/v0.21.5` instead?

PR #3384, opened 2023-08-03
PCD tagging and group modal resets

benjaminpkane:
Fixes and/or adds support for tagging all visible (selected) PCD slices. Also ensures modal state is configured correctly on expansion. Currently on `release/v0.21.5`, opening a group dataset sample and then closing, and switching to non-group dataset and _then_ opening the modal will crash the App.

Example multi-pcd slice group dataset
```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = fo.Dataset("pcd-only-sparse")
dataset.add_group_field("group", default="shared")
dataset.persistent = True

source = foz.load_zoo_dataset("quickstart-groups", max_samples=9)
source.group_slice = "pcd"

first = fo.Group()
one = source.first().copy()
one.group = first.element("shared")

two = source.skip(1).first().copy()
two.group = first.element("unique")

second = fo.Group()
three = source.skip(2).first().copy()
three.group = second.element("shared")

dataset.add_samples([one, two, three])
```

### DIFF
--- a/app/packages/core/src/components/Actions/Tagger.tsx
+++ b/app/packages/core/src/components/Actions/Tagger.tsx
@@ -1,13 +1,7 @@
 import { PopoutSectionTitle, useTheme } from "@fiftyone/components";
 import { FrameLooker, ImageLooker, VideoLooker } from "@fiftyone/looker";
 import * as fos from "@fiftyone/state";
-import {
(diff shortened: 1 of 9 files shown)


sashankaryal (approved):
lgtm

PR #3386, opened 2023-08-04
bugfix: sample tag and label tag visibility and filter issue

lanzhenw:

fix the sample tag and label tag bubble visibility setting was not applied correctly.

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/state/src/recoil/pathFilters/index.ts
+++ b/app/packages/state/src/recoil/pathFilters/index.ts
@@ -120,28 +120,27 @@ export const pathFilter = selectorFamily<PathFilterSelector, boolean>({
                 return () => true;
               }
 
-              if (value[0]) {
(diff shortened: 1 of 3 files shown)


app/packages/state/src/recoil/pathFilters/index.ts
```diff
@@ -180,7 +179,7 @@ const matchesLabelTags = (
   if (!filter && visibility) {
     const { values, exclude } = visibility;
 
-    const contains = value.tags.some((tag) => values.includes(tag));
+    const contains = value.tags?.some((tag) => values.includes(tag));
```
manivoxel51, line 182: nit: curious why we are optionally picking now using `?`

manivoxel51 (approved):
Thank you for fixing this! 🍨 

We should add a note in our e2e effort to make a test for this if possible.

PR #3388, opened 2023-08-04
Fix dynamic groups modal of a view

benjaminpkane:
Will follow up with e2e.

Dynamic grouping of a group dataset requires pagination over the dynamically grouped samples. These results require providing the current view while replacing the `GroupBy` stage with the matching stage and filtering limit and skip stages post matching stage.
```py
import fiftyone as fo
import fiftyone.zoo as foz

ds = foz.load_zoo_dataset('quickstart-groups',dataset_name='qsg-test')

scene_id = ['scene1','scene2','scene3','scene4']*50
frame = range(len(ds))
for slice in ds.group_slices:
    ds.group_slice = slice
    ds.set_values('scene_id',scene_id)
    ds.set_values('frame',frame)

v = ds.limit(10)
view = v.group_by('scene_id', order_by='frame')
ds.save_view('dynamic-limit', view)

session = fo.launch_app(view)
```
Screen Shot 2023-08-04 at 10 34 05 AM


sashankaryal (approved):
lgtm

Issue #3392, opened 2023-08-05
[?] How to export open-images-v7 in YOLOv5Dataset format to train YOLOv8 


glenn-jocher:
Hello, I'm the author of Ultralytics [YOLOv8](https://github.com/ultralytics/ultralytics) and am exploring using fiftyone for training some of our datasets, but there seems to be a bug.

Note that for our use case YOLOv5Dataset works fine, though also please be aware that we've updated the Ultralytics YOLOv3/5/8 data.yaml formats to use a class dictionary rather than a `names` list and `nc` class count. See https://docs.ultralytics.com/datasets/detect/

<img width="1427" alt="Screenshot 2023-08-05 at 13 34 51" src="https://github.com/voxel51/fiftyone/assets/26833433/da29d556-229e-438a-afc7-84df1add353f">

### System information

-   **OS Platform and Distribution**:  macOS Ventura 13.4.1)
-   **Python version** (`python --version`):  3.11
-   **FiftyOne version** (`fiftyone --version`): 0.21.4 (latest)
-   **FiftyOne installed from** (pip or source): pip install fiftyone

### Commands to reproduce

As thoroughly as possible, please provide the Python and/or shell commands used to encounter the issue. Application steps can be described in the next section.

```python
import shutil
from pathlib import Path
import fiftyone as fo

# Load Open Images dataset
dataset = fo.zoo.load_zoo_dataset("open-images-v7",
                                  #splits=["train", "validation"],
                                  label_types=["detections"],
                                  max_samples=10)

# Export to YOLO format
save_dir = "./datasets/open-images-v7"
if Path(save_dir).exists():
    shutil.rmtree(save_dir)
dataset.export(export_dir=save_dir, dataset_type=fo.types.YOLOv5Dataset)
```

Output:
```
Downloading split 'train' to '/Users/glennjocher/fiftyone/open-images-v7/train' if necessary
Necessary images already downloaded
Existing download of split 'train' is sufficient
Downloading split 'validation' to '/Users/glennjocher/fiftyone/open-images-v7/validation' if necessary
Necessary images already downloaded
Existing download of split 'validation' is sufficient
Loading 'open-images-v7' split 'train'
 100% |█████████| 10/10 [38.4ms elapsed, 0s remaining, 260.1 samples/s]     
Loading 'open-images-v7' split 'validation'
 100% |█████████| 10/10 [30.6ms elapsed, 0s remaining, 327.2 samples/s]    
[...]

PR #3393, opened 2023-08-06
Update Ultralytics YOLO `dataset.yaml` format

glenn-jocher:

@brimoor PR per https://github.com/voxel51/fiftyone/issues/3392. 

If we could squeeze this into `fiftyone==0.21.5` that would be awesome!

Replaced `nc` integer and `names` list with a `names` dictionary per the new format in https://docs.ultralytics.com/datasets/detect/. 

This new format works for all Ultralytics YOLO repos (v3/v5/v8) and has been in place for about a year, so this should be widely compatible with most forks and copies in the wild.

The old format is being maintained through 2023 and will be deprecated in 2024.

Not tested.

        notes for FiftyOne users.

> Replaced deprecated `nc` integer and `names` list with a `names` dictionary per the new format in https://docs.ultralytics.com/datasets/detect/. This new format works for all Ultralytics YOLO repos (v3/v5/v8) and has been in place for about a year, so this should be widely compatible with most forks and copies in the wild.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/yolo.py
+++ b/fiftyone/utils/yolo.py
@@ -944,8 +944,9 @@ class YOLOv5DatasetExporter(
             d["path"] = os.path.dirname(self.yaml_path)
 
         d[self.split] = _make_yolo_v5_path(self.data_path, self.yaml_path)
-        d["nc"] = len(classes)
-        d["names"] = classes
+
+        # New data.yaml format https://docs.ultralytics.com/datasets/detect/
+        d["names"] = dict(enumerate(classes))  # class names dictionary
 
         _write_yaml_file(d, self.yaml_path, default_flow_style=False)
 


brimoor (approved):
LGTM thanks @glenn-jocher! 💪  we'll get this into `fiftyone==0.21.5`

glenn-jocher:
Awesome thanks!

AUTUMN1124:
Hello, I'd like to ask why YOLOv8 training was normal yesterday, but today it has been training for over two hours with no results?T_T

brimoor:
Sounds like an issue with the `ultralytics` package, not `fiftyone`

AUTUMN1124:
> Sounds like an issue with the `ultralytics` package, not `fiftyone`

I've solved the issue; it was because I downloaded the wandb library for tuning but forgot to apply for the API

PR #3405, opened 2023-08-07
Fixing bug when selecting nested fields within already-selected root fields

brimoor:


benjaminpkane (approved):
Static review but LGTM. Tests are great 👍

manivoxel51 (approved):
🍨

PR #3406, opened 2023-08-08
Adding a fiftyone.core.storage module

brimoor:
Adds a `fiftyone.core.storage` module that provides a common interface for filesystem I/O


brimoor:
Pulled these commits into https://github.com/voxel51/fiftyone-teams/pull/338.

ritch (approved):

Issue #1821, opened 2022-06-03
[FR] Add support for rotated bounding boxes


HamLam:
### Proposal Summary

Could you please provide support for the visualization of rotated bounding boxes for coco dataset? 
The bounding box parameters for rotated boxes would include a fifth entry (rotated angle in radian) instead of the usual four for coco formatted dataset. Thank you! 

### Motivation

-   What is the use case for this feature?
Rotated bounding boxes are commonly bound on coco dataset in object detection. 

-   Why is this use case valuable to support for FiftyOne users in general?
It is a valuable feature because it expands the capability of FiftyOne

### What areas of FiftyOne does this feature affect?

-   [ *] App: FiftyOne application
-   [* ] Core: Core `fiftyone` Python library
-   [ ] Server: FiftyOne server

### Details

(Use this section to include any additional information about the feature. If
you have a proposal for how to implement this feature, please include it here.)

### Willingness to contribute

The FiftyOne Community encourages new feature contributions. Would you or
another member of your organization be willing to contribute an implementation
of this feature?

-   [ ] Yes. I can contribute this feature independently.
-   [ ] Yes. I would be willing to contribute this feature with guidance from
        the FiftyOne community.
-   [ *] No. I cannot contribute this feature at this time.

ehofesmann:
I definitely agree that rotated bounding boxes would be valuable to have in FiftyOne. While it shouldn't be too much of a lift to add, it will require updates to the App to support rotated bbox visualization, and we also need to decide on how the rotation angle will be stored in the dataset. It seems like the best options would be either:
1) add the rotation angle into the `bounding_box` list allowing it to either be 4 or 5 elements in length
2) add a new dedicated attribute `rotation_angle` to the `Detection`

radao:
Any plans to add rotated boxes in the near term?

robmarkcole:
This is an increasingly common format for aerial datasets and its inclusion would benefit this entire domain

brimoor:
Note that one can already visualize this data by storing it as [polylines](https://voxel51.com/docs/fiftyone/user_guide/using_datasets.html#polylines-and-polygons).

Perhaps it would be sufficient to just add a `Polyline.from_rotated_box()` factory method?


```py
import cv2
import numpy as np

import fiftyone as fo
import fiftyone.core.labels as fol


def from_rotated_box(xc, yc, w, h, theta, **kwargs):
[...]

PR #3409, opened 2023-08-12
Use frame size when constructing rotated boxes

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/1821


benjaminpkane (approved):
LGTM

PR #3425, opened 2023-08-17
add sample bar in sample3d

sashankaryal:


app/packages/core/src/components/Modal/Sample3d.tsx
```diff
@@ -37,10 +39,14 @@ const Looker3dPluginWrapper = () => {
 };
 
 export const Sample3d = () => {
+  const lookerRef = useRef<AbstractLooker | undefined>(undefined);
```
benjaminpkane, line 42: suggestion: I've been typing with `Lookers` which is union type

benjaminpkane (approved):
LGTM!

PR #3426, opened 2023-08-17
group tests refactor

sashankaryal:

- execute python code only once in group tests
- skip pcd for now because slice selection behavior is different from other modalities


benjaminpkane (approved):
LGTM! Thank you

PR #3429, opened 2023-08-18
Fix OpenLABEL importer when given incomplete keypoint skeletons

ehofesmann:
The OpenLABEL importer is designed to set `None` values for missing list attributes of keypoints. 
The issue is that setting None values like that for these attributes:
```python

keypoint = fo.Keypoint(
    label="rectangle",
    points=[
        (0.3, 0.3),
        (float("nan"), float("nan")),  # use nan to encode missing points
        (0.7, 0.7),
        (0.3, 0.7),
    ],
    my_occluded_attr=[
        "partially",
        None,
        "fully",
        "partially",
    ]
)
```
Will result in errors when trying to add the attributes to the dataset schema (`dataset.add_dynamic_sample_fields()`)

This PR updates this logic to set them to be `""` instead for strings, `0` or `0.0` for numeric types, and `False` for booleans.

Also updated the OpenLABEL tests to check for this so it would have failed previously and now passes.

### DIFF
--- a/fiftyone/utils/openlabel.py
+++ b/fiftyone/utils/openlabel.py
@@ -1170,6 +1170,10 @@ class OpenLABELPoint(OpenLABELShape):
     """An OpenLABEL keypoint."""
 
     @classmethod
+    def _get_empty_value_for_type(cls, example_type):
+        return example_type()
+
+    @classmethod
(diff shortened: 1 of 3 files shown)


allenleetc (approved):
LGTM

PR #3432, opened 2023-08-21
[HOLD] run results shared between dataset clones

swheaton:

Proposal is to have dataset clones share run result documents in GridFS through the use of reference counts on the metadata of the file object.

#### Why?
It saves a marginal amount of space in MongoDB by not duplicating run result documents in GridFS for fully-cloned datasets.

However the primary motivator is to set this baseline usage so that snapshots in FiftyOneTeams work seamlessly as well as clones.

#### Implementation Details
1. dataset1 create run with results with id X
2. dataset2 is cloned from dataset1. Instead of copying the results contents to a new file, it increments the reference counter.
   1. **Note** changing the metadata of a GridFS file after it's written is technically not a part of the public API. However they are just normal collections in most ways, so if we only edit our reference count field atomically and change nothing else, we should be OK.
3. If one of these datasets delete the run result (either through deletion or saving a new one), the reference count is decremented.
4. File is deleted after reference count hits 0.

This is done through creation of `SharedFileField` and `ReferenceCountingGridFSProxy` which subclass `mongoengine.fields.FileField` and `mongoengine.fields.GridFSProxy`, respectively. It wraps the functionality of `GridFSProxy` with adding the `reference_count` field to the file's metadata. We try to atomically test-and-update this counter whenever possible to avoid race conditions.

- Added unit tests for `SharedFileField` directly in `odm_tests.py`
- Added a test to `evaluation_tests.py` to test sharing documents works from the higher level
- Manual testing like the below example

``` python
import fiftyone as fo
dataset = fo.zoo.load_zoo_dataset("quickstart")
results = dataset.evaluate_detections("predictions", gt_field="ground_truth", eval_key="eval")
original_results = results.serialize()
results_id = dataset._doc.evaluations["eval"].results.grid_id

clone = dataset.clone()
results_id2 = clone._doc.evaluations["eval"].results.grid_id

# Sharing same result object
assert results_id == results_id2

# Deleting original evaluation should not delete clone evaluation
dataset.delete_evaluation("eval")
clone_results = clone.load_evaluation_results("eval", cache=False)
assert clone_results.serialize() == original_results
```

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -7335,12 +7335,9 @@ def _clone_run(run_doc):
     _run_doc.id = ObjectId()
     _run_doc.results = None
 
-    # Unfortunately the only way to copy GridFS files is to read-write them...
-    # https://jira.mongodb.org/browse/TOOLS-2208
+    # Use copy_from to make a shallow copy
     if run_doc.results:
-        run_doc.results.seek(0)
-        results_bytes = run_doc.results.read()
-        _run_doc.results.put(results_bytes, content_type="application/json")
+        _run_doc.results.copy_from(run_doc.results)
 
     return _run_doc
 
--- a/fiftyone/core/odm/fields.py
+++ b/fiftyone/core/odm/fields.py
@@ -1,0 +1,88 @@
+"""
+Dataset sample fields.
+
+| Copyright 2017-2023, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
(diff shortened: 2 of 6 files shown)


swheaton:
I don't see why it would be a surprise, they should never be looking at the run results directly anyway.

In any case, in an offline thread, it was discussed more, and I am considering retracting this PR

swheaton:
Closing this PR as it was decided we not go forward with this, it's too risky for little gain.

PR #3435, opened 2023-08-21
Omit `close_app` in `launch_app`

benjaminpkane:
When launching the via `fiftyone.core.session.session.launch_app()`, the current behavior is to close the existing global `_session` if it is defined.

This change proposes to omit that step for faster launches when `launch_app` is used multiple times. The previous `session` will be cleaned up when the main process exits.


brimoor (approved):
LGTM! This is much better for me as I often call `launch_app()` multiple times in one ipython session 😅

PR #3436, opened 2023-08-21
Unique ID for full sample JSON in modal

### LINKED ISSUE #3431: [BUG] Group sample in modal missing attributes
In `0.21.6`, a group sample in the modal will be missing attribute data unless the fields are added via `add_dynamic_sample_fields()` or `add_dynamic_frame_fields()`. This is due a subtle detail in the `relay` cache, which is straightforward to resolve.

```py
import fiftyone as fo

for sample in dataset:
    for detection in sample.ground_truth.detections:
        detection["my_attr"] = "value"
    sample.save()

# tooltip will not contain `my_attr` on detections


# will force `my_attr` to appear
```

benjaminpkane:
Resolves #3431 

In `0.21.6`, modal sample data can be overwritten by the group carousel due to globally unique id store in relay. Adding a `-modal` suffix to sample IDs when they are requested without pagination data resolves the issue. The `-modal` naming is tentative.

The best way to add an e2e test is to screenshot the JSON panel in a group dataset, if it is practical.

### DIFF
--- a/app/packages/core/src/components/Actions/Tagger.tsx
+++ b/app/packages/core/src/components/Actions/Tagger.tsx
@@ -364,7 +364,9 @@ const useTagCallback = (
               ids.add(sample.sample._id);
             });
           });
-          updateSamples(Array.from(ids).map((id) => [id, undefined]));
+          updateSamples(
+            Array.from(ids).map((id) => [id.split("-")[0], undefined])
+          );
         } else if (samples) {
           set(fos.refreshGroupQuery, (cur) => cur + 1);
           updateSamples(samples.map((sample) => [sample._id, sample]));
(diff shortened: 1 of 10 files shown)


manivoxel51:
@benjaminpkane today I tried two ways to test this and failed, 

- I tried screenshotting JSON viewer but quickly realized new attributes added to samples are not available in JSON viewers at all so I stopped.
- I confirmed that new attributes are visible in canvas if you hover over a label with attribute. But, canvas testing does not look straight forward so far...

@sashankaryal what do you advise?

manivoxel51:
If it turns out uneasy to test, I vote for merging it since it is a low risk code path,

benjaminpkane:
> I tried screenshotting JSON viewer but quickly realized new attributes added to samples are not available in JSON viewers at all so I stopped

The attributes should be present in JSON. I don't think this is true.

I think adding JSON viewer method(s) to the Modal POMs would be valuable, but it doesn't have to be for this PR. I'm fine with merging without tests.

PR #3443, opened 2023-08-22
Minor e2e refactoring

sashankaryal:

- Add `getWaitForGridRefreshPromise()` method to the grid POM
- Minor linting

- Local E2E green

### DIFF
--- a/e2e-pw/src/oss/poms/grid/index.ts
+++ b/e2e-pw/src/oss/poms/grid/index.ts
@@ -11,7 +11,7 @@ export class GridPom {
   readonly locator: Locator;
 
   constructor(
-    private readonly page: Page,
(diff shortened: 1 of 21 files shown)


benjaminpkane (approved):
Looks good! Relying on flashlight events is much cleaner

e2e-pw/src/oss/poms/grid/index.ts
```diff
@@ -54,19 +55,34 @@ export class GridPom {
   }
 
   async getNthFlashlightSection(n: number) {
-    return this.page.getByTestId("flashlight-section").nth(0);
+    return this.page.getByTestId("flashlight-section").nth(n);
```
lanzhenw, line 58: Thanks for the fix.

PR #3447, opened 2023-08-23
fix dynamic groups e2e bug

sashankaryal:

- This fixes the flaky behavior of some e2e tests related to dynamic groups caused by selecting a group-by field before it's populated.

- Locally


benjaminpkane (approved):
LGTM 🥇  I should have updated these when I added the Selector POM

PR #3461, opened 2023-08-29
early return for empty dir and ignore hidden files

kaixi-wang:
Fix `ValueError: max() arg is an empty sequence` for when a plugin dir exists but is empty

### DIFF
--- a/fiftyone/operators/decorators.py
+++ b/fiftyone/operators/decorators.py
@@ -6,6 +6,8 @@ FiftyOne operator decorators.
 |
 """
 import asyncio
+import glob
+
 from cachetools.keys import hashkey
 from contextlib import contextmanager
 from functools import wraps
@@ -85,6 +87,8 @@ def plugins_cache(func):
 def dir_state(dirpath):
     if not os.path.isdir(dirpath):
         return None
-    return max(
-        os.path.getmtime(os.path.join(dirpath, f)) for f in os.listdir(dirpath)
-    )
+    # use glob instead of os.listdir to ignore hidden files (eg .DS_STORE)
+    dirs = glob.glob(os.path.join(dirpath, "*"))
+    if not dirs:
+        return None
+    return max(os.path.getmtime(f) for f in dirs)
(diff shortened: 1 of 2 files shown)


fiftyone/operators/decorators.py
```diff
@@ -85,6 +87,8 @@ def wrapper(*args, **kwargs):
 def dir_state(dirpath):
     if not os.path.isdir(dirpath):
         return None
-    return max(
-        os.path.getmtime(os.path.join(dirpath, f)) for f in os.listdir(dirpath)
-    )
+    # use glob instead of os.listdir to ignore hidden files (eg .DS_STORE)
+    dirs = glob.glob(os.path.join(dirpath, "*"))
```
brimoor, line 91: I defer to @ritch's judgement on this hidden file change. This isn't necessary to fix the bug, and while it would be strange to use hidden files for important things, it *could* be the case that a plugin has a hidden file whose update *should* trigger a plugin reload.

brimoor (commented):
I defer to @ritch for approval after his thoughts on the hidden file change

fiftyone/operators/decorators.py
```diff
@@ -85,6 +87,8 @@ def wrapper(*args, **kwargs):
...
+    if not dirs:
+        return None
+    return max(os.path.getmtime(f) for f in dirs)
```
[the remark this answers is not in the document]
kaixi-wang: I don't think adding None works: 
```
max([],None)
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
TypeError: '>' not supported between instances of 'NoneType' and 'list'
```

fiftyone/operators/decorators.py
kaixi-wang: the only place it is used is for the plugin_cache and finding the latest updated timestamp in the plugins dir. Any hidden files updated by a plugin should be inside the plugin's directory but not the fo.config.plugins_dir right?

fiftyone/operators/decorators.py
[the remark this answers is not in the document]
brimoor: Sorry I meant `max(..., default=None)`

fiftyone/operators/decorators.py
brimoor: Oh right I forgot we aren't recursively listing. Just hoping that the directory's `mtime` is sufficient. Either way, defer to Ritchie's review

ritch (approved):

PR #3462, opened 2023-08-29
[HOLD] Reload derived collections after changing label tags

benjaminpkane:
Proposal to reload `ClipsView`, `FramesView`, `EvaluationPatchesView` and `PatchesView` collections after tagging labels in the App. Server tests are in progress.


brimoor:
The only reason a `view.reload()` would be necessary after label tag changes are made is specifically if the tag changes should result in a different number of patches in the generated dataset (which would only happen specifically if the patches view contains a stage that matches on label tags). The actual tag changes themselves should be immediately reflected on both the source dataset and the patches view.

Suggest continuing the discussion on this [here](https://voxel51.atlassian.net/browse/CS-354).

PR #3483, opened 2023-09-02
Generalize _is_docker() to is_container()

ledkarlsson:

Generalize _is_docker() to is_container(), in order for same behavior using docker and podman. 
The is_docker() method allowed a host to connect to Fiftyone running in a docker container without extra config. Using podman, extra config was needed.

Tested by running is_container() using podman, docker, and without container. The function returns true for both podman and docker, and false without running as container.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/session/session.py
+++ b/fiftyone/core/session/session.py
@@ -358,7 +358,7 @@ class Session(object):
             )
 
         if address is None:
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
LGTM! Thank you for the contribution @ledkarlsson

PR #3492, opened 2023-09-07
Adding support for listing datasets with tags

brimoor:
```py
import fiftyone as fo

dataset = fo.Dataset("tmp-auto")
dataset.tags = ["tmp", "automotive"]
dataset.persistent = True

fo.list_datasets(tags="automotive")
fo.list_datasets(tags=["automotive", "healthcare"])
```

```shell
fiftyone datasets list --tags automotive
fiftyone datasets list --tags automotive healthcare

fiftyone datasets info --tags automotive
fiftyone datasets info --tags automotive healthcare

fiftyone datasets delete tmp-auto
```

### DIFF
--- a/docs/source/cli/index.rst
+++ b/docs/source/cli/index.rst
@@ -324,7 +324,7 @@ List FiftyOne datasets.
 
 .. code-block:: text
 
-    fiftyone datasets list [-h] [-p PATT]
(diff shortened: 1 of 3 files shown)


allenleetc (approved):
LGTM

PR #3507, opened 2023-09-11
limit refresh on embeddings panel when clearing selection

manivoxel51:

watch proposed fix below. 

- [x] Don't cause refresh the embedding panel when clearing embeddibngs
- [x] Confirm fix

[Updated] Proposed fix 

https://github.com/voxel51/fiftyone/assets/109545780/b9814216-65fc-4830-a283-8173b4757858

(Details)

        notes for FiftyOne users.

-   [ ] Other


brimoor:
@manivoxel51 sorry looks like we had a miscommunication. The sample grid *does* need to refresh every time the selection is changed (it always needs to stay in sync with the extended view, which changes whenever embeddings selection changes). The specific thing that is unnecessary is that the embeddings panel should not re-render (which can take awhile and causes things like zoom/pan to be reset unnecessarily). 

On `develop` right now, changing from `n` to `m != n` points selected does not cause the embeddings panel to fully reset, but going from `n` to `0` does cause it to refresh. The latter thing is the specific thing we want to avoid.

manivoxel51:
> @manivoxel51 sorry looks like we had a miscommunication. The sample grid _does_ need to refresh every time the selection is changed (it always needs to stay in sync with the extended view, which changes whenever embeddings selection changes). The specific thing that is unnecessary is that the embeddings panel should not re-render (which can take awhile and causes things like zoom/pan to be reset unnecessarily).
> 
> On `develop` right now, changing from `n` to `m != n` points selected does not cause the embeddings panel to fully reset (good), but going from `n` to `0` does cause it to refresh (bad). The latter thing is the specific thing we want to avoid.

cool makes sense! that should be doable

benjaminpkane:
The change is working for me 🚀 . I see that clearing the filters was added (or moved) to the panel close effect. So this means sidebar filters are reset when the panel is closed (see recording)? Is that expected behavior?

https://github.com/voxel51/fiftyone/assets/19821840/bfc98a57-5c71-452d-abac-a0c931933884

app/packages/embeddings/src/usePlotSelection.tsx
```diff
@@ -38,7 +37,6 @@ export function usePlotSelection() {
     resetExtendedSelection();
     setPlotSelection(null);
     setSelectedSamples(new Set());
-    setFilters({});
```
manivoxel51, line 41: this

benjaminpkane (approved):
LGTM 🚀 I am fine with skipping e2e as this fixes a pending (loading) state. May be hard (or overkill) to test that at this stage

PR #3512, opened 2023-09-12
Update operator asyc execution check

kaixi-wang:
Make the check for any awaitable, not just asyncio.iscoroutinefunction which might have some issues/will be depreciated: https://github.com/python/cpython/issues/94912

### DIFF
--- a/fiftyone/operators/executor.py
+++ b/fiftyone/operators/executor.py
@@ -6,6 +6,7 @@ FiftyOne operator execution.
 |
 """
 import asyncio
+import inspect
 import traceback
 import types as python_types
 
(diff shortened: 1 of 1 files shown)


fiftyone/operators/executor.py
```diff
@@ -199,7 +200,7 @@ async def execute_or_delegate_operator(operator_uri, request_params):
         try:
             raw_result = await (
                 operator.execute(ctx)
-                if asyncio.iscoroutinefunction(operator.execute)
+                if inspect.isawaitable(operator.execute)
```
ritch, line 203: This will always return false. These are not functionally equivelent.

ritch: image

ritch: image

ritch (changes requested):

PR #3515, opened 2023-09-13
Compound group-by and better index creation control

brimoor:
Enhancements
- Adds support for compound group-by operations via `dataset.group_by(["field1", "field2"])`
    - This was previously possible _only_ via a more verbose syntax `dataset.group_by(E([F("field1"), F("field2")]))`
- Adds a `create_index=False` option for `sort_by()` and `group_by()` in case users do *not* want to create indexes

Bug fixes
- Fixes a bug where default indexes for grouped datasets were not created via `dataset.clone()` and `dataset.merge_samples()`

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -4623,6 +4623,7 @@ class SampleCollection(object):
         flat=False,
         match_expr=None,
         sort_expr=None,
+        create_index=True,
     ):
         """Creates a view that groups the samples in the collection by a
         specified field or expression.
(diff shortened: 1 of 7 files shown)


benjaminpkane:
@brimoor are you hoping to include App support in the release? I don't believe the frontend needs some small changes

```py
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("cifar10", max_samples=200)

for sample in dataset:
    sample["label_copy"] = sample.ground_truth.label
    sample.save()

grouping = dataset.group_by(["ground_truth.label", "label_copy"])

session = fo.launch_app(view)
```
Screen Shot 2023-09-13 at 2 41 05 PM

brimoor:
@benjaminpkane hmm I didn't think to try launching the App because `dataset.group_by(E([F("field1"), F("field2")]))` was already something that could be done. `dataset.group_by(["field1", "field2"])` is just a more concise way to define the same view.

Is the App issue specific to the new syntax? Or is there an issue with any dynamic group that's not a single field?

brimoor:
@benjaminpkane can we approve + merge this now? Definitely want to include #3529 but since the App already doesn't support `group_by(expr)` I think it's okay to merge this now and not support `group_by([field1, field2])` in the App for a hot minute either

benjaminpkane (approved):
Sounds good 👍 LGTM

PR #3516, opened 2023-09-13
SetProgress operator

brimoor:
Example usage in the `execute()` method of a generator operator:

```py
yield ctx.trigger("set_progress", dict(progress=0.51, label="Processed 51 of 100"))
```


app/packages/operators/src/built-in-operators.ts
```diff
@@ -613,6 +613,42 @@ class ShowOutput extends Operator {
...
+    return new types.Property(inputs);
+  }
+  useHooks(ctx: ExecutionContext): {} {
+    return {
+      io: useShowOperatorIO(),
+    };
+  }
+  async execute({ params, hooks: { io } }: ExecutionContext) {
+    const loading = new types.Object();
+    const progressView = new types.ProgressView({ label: params.label });
```
ritch, line 640: We should support all options of the ProgressView.

PR #3517, opened 2023-09-13
treat torch models as feature extractors when feature_extractor=True

Rusteam:

New parameter `feature_extractor=False` added to torch models. 
If set to True, then the model output is treated as features without
forward hooks. 

When using `dino` model with default forward hooks, its output
becomes 3-demensional and can't be used as embeddings. With 
this fix the output is expected 2-dimensional. 

Also, if `embeddings_layer=xxxx` is provided, then the model
is replaced with a feature extractor for that layer using torchvision.

Finally, added a check on `embeddings_layer` making sure it's
not an `Identity()` layer is it does not return anything with 
forward hooks.

Added a test with `dino` model from torch hub.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -252,14 +252,14 @@ class TorchEmbeddingsMixin(fom.EmbeddingsMixin):
         layer_name (None): the name of the embeddings layer whose output to
             save, or ``None`` if this model instance should not expose
             embeddings. Prepend ``"<"`` to save the input tensor instead
-        feature_extractor (False): whether this model instance should
+        as_feature_extractor (False): whether this model instance should
             operate as a feature extractor, in which case ``layer_name`` if provided
             is used to create a feature extractor. If ``layer_name`` is not provided,
             the model is used as-is for feature extraction.
     """
 
-    def __init__(self, model, layer_name=None, feature_extractor=False):
-        if feature_extractor:
+    def __init__(self, model, layer_name=None, as_feature_extractor=False):
+        if as_feature_extractor:
             if layer_name:
                 # create a torchvision feature extractor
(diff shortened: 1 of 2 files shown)


fiftyone/utils/torch.py
```diff
@@ -251,32 +252,50 @@ class TorchEmbeddingsMixin(fom.EmbeddingsMixin):
         layer_name (None): the name of the embeddings layer whose output to
             save, or ``None`` if this model instance should not expose
             embeddings. Prepend ``"<"`` to save the input tensor instead
+        feature_extractor (False): whether this model instance should
```
kaixi-wang, line 255: since this is a bool, can you rename this to `extract_features` or `as_feature_extractor` to make it more clear

fiftyone/utils/torch.py
```diff
@@ -251,32 +252,50 @@ class TorchEmbeddingsMixin(fom.EmbeddingsMixin):
...
+            the model is used as-is for feature extraction.
     """
 
-    def __init__(self, model, layer_name=None):
-        if layer_name is not None:
+    def __init__(self, model, layer_name=None, feature_extractor=False):
+        if feature_extractor:
+            if layer_name:
+                # create a torchvision feature extractor
+                self._model = create_feature_extractor(
```
kaixi-wang, line 265: can this be used with all torchmodels or just dino?

Rusteam: with all torch models

fiftyone/utils/torch.py
Rusteam: renamed!

brimoor (approved):
LGTM, makes sense 👍

PR #3518, opened 2023-09-13
exhaust the generator before returning

nebulae:

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/operators/delegated.py
+++ b/fiftyone/operators/delegated.py
@@ -301,5 +301,6 @@ class DelegatedOperationService(object):
 
             if not is_generator:
                 return raw_result
-            else:
-                all(raw_result)
+
+            for _ in raw_result:
+                pass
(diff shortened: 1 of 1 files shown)


fiftyone/operators/delegated.py
```diff
@@ -301,3 +301,5 @@ async def _execute_operator(self, doc):
 
             if not is_generator:
                 return raw_result
+            else:
+                all(raw_result)
```
brimoor, line 305: I'm not familiar with this code but is it safe to assume that the generator always returns something truthy? `all()` would short circuit if it encounters something falsey.

Just use this for safety?

```py
for _ in raw_result:
    pass
```

brimoor (changes requested):

fiftyone/operators/delegated.py
```diff
@@ -301,3 +301,6 @@ async def _execute_operator(self, doc):
 
             if not is_generator:
                 return raw_result
+
+            for _ in raw_result:
```
ritch, line 305: comment here would be good as this code looks very weird

PR #3519, opened 2023-09-13
Always log stack traces in server

benjaminpkane:
In starlette, the `debug` flag for an `App` is documented as:
```
Boolean indicating if debug tracebacks should be returned on errors.
```
which is always useful 🚀


lanzhenw (approved):
LGTM

PR #3520, opened 2023-09-13
[Proposal] Require e2e in PR and Publish workflows

benjaminpkane:
e2e tests are stable now, so I propose we include them as required checks (open to discussion, they are slower than other checks)


sashankaryal (approved):
🥇

manivoxel51 (approved):
thank you! 🕺

PR #3528, opened 2023-09-15
Optimize list datasets info

brimoor:


swheaton:
This is a lot faster especially in Teams through API. Was basically unusable before.
Do you think anyone relies on / would miss the `num_samples` part of the info?

swheaton:
Right. A little better than loading every dataset completely (which is like `n * 4` db calls), but still not ideal.

swheaton (approved):
lgtm. tested and works well also

PR #3545, opened 2023-09-19
Fix group modal when grid filters are present

benjaminpkane:
This change removes the `filters` variable from the main sample query selector and limits inherited filters when expanding a sample or group to only include labels and label tags settings. Grid filters on primitive values should not exclude slices in the expanded view. This is a subtle detail so we can also discuss offline @manivoxel51 @lanzhenw 

e2e has been added with the example of filtering by the `left` slice in the grid in `quickstart-groups`. See recordings for before and after behavior

Before:
https://github.com/voxel51/fiftyone/assets/19821840/dd7fd961-b714-410d-8a98-1a60622bb721

After:
https://github.com/voxel51/fiftyone/assets/19821840/81e8919a-5701-4dbf-8362-e98c438ecddd

### DIFF
--- a/app/packages/state/src/hooks/useExpandSample.ts
+++ b/app/packages/state/src/hooks/useExpandSample.ts
@@ -1,7 +1,7 @@
 import { FlashlightConfig } from "@fiftyone/flashlight";
 import { get } from "lodash";
 import { useRelayEnvironment } from "react-relay";
-import { RecoilState, useRecoilCallback } from "recoil";
+import { CallbackInterface, RecoilState, useRecoilCallback } from "recoil";
 import * as atoms from "../recoil/atoms";
 import * as filterAtoms from "../recoil/filters";
 import * as groupAtoms from "../recoil/groups";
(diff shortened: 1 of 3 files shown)


app/packages/state/src/hooks/useExpandSample.ts
```diff
@@ -14,70 +14,81 @@ import * as viewAtoms from "../recoil/view";
...
 
+const setModalFilters = async ({ snapshot, set }: CallbackInterface) => {
+  const paths = await snapshot.getPromise(
+    schemaAtoms.labelPaths({ expanded: false })
+  );
+  const filters = await snapshot.getPromise(filterAtoms.filters);
+  const modalFilters = Object.fromEntries(
+    Object.entries(filters).filter(
+      ([path]) =>
+        paths.some((p) => path.startsWith(p)) || path === "_label_tags"
```
manivoxel51, line 25: I think this can use a nice unit test to make sure we lock down the logic and kind of document one-offs like `_lable_tags`

manivoxel51 (approved):
Thank you 🍨 

I tested locally image, video, and group datasets and looks good!

I left a comment but that is a good-to-have

app/packages/state/src/hooks/useExpandSample.ts
benjaminpkane: I will follow up this week

PR #3549, opened 2023-09-20
Typo in componentProps

marijnl:

`fo.operators.types.View` refers to `componentProps` but this property is not used in the frontend.
The correct property name is `componentsProps`.

If the user sets `types.Notice(.., description="", componentsProps={"description": ""})` it will work because all arguments to `Notice` are passed to the frontend when calling `to_json`.

        notes for FiftyOne users.


imanjra (approved):

imanjra:
Thank you for the contribution @marijnl! 🎉

PR #3556, opened 2023-09-21
[Proposal] Add lint step to e2e

benjaminpkane:
We have added a lot poms and specs recently 🚀 . Proposing we add a lint stage to `e2e` to catch changes to poms, etc. I find myself tweaking constructors and methods a lot, so I think it's good to have the automation catch these changes so specs don't fall behind


sashankaryal (approved):
Let's go!

Issue #3557, opened 2023-09-22
[BUG] ctx.selected_labels is empty in resolve_input and resolve_output


msamadony:
### System information

-   **OS Platform and Distribution**: Ubuntu 20.04
-   **Python version**: 3.8.10
-   **FiftyOne version**: 0.22.0
-   **FiftyOne installed from**: pip

### Commands to reproduce

See the sample operator in the code section below.

### Describe the problem

The `ctx.selected_labels` always return an empty list when it is read from the context of `resolve_input` and `resolve_output` of an operator in python. The `ctx.selected_labels` returns the correct information only when it is read from the `execute` function. It is worth noting that the value of `ctx.selected` works correctly in the three functions.

The below screenshots shows the input and output panels of the below referenced sample operator.

Operator Input:
![image](https://github.com/voxel51/fiftyone/assets/224718/17337a15-3832-4c60-9ca6-d48d9da62dec)

Operator Output:
![image](https://github.com/voxel51/fiftyone/assets/224718/768d7485-c4b3-46b7-90e3-fd98acc24c22)

Having a look at the frontend code, it seems that the `selectedLabels` state is missing from `useExecutionContext`:

https://github.com/voxel51/fiftyone/blob/7ee780bc7a4c854371dc9c5facd181517a718d9b/app/packages/operators/src/state.ts#L113-L132

### Code to reproduce issue

Below is a sample operator to reproduce the behavior:
```
class SelectedLabelsOperator(foo.Operator):
    @property
    def config(self):
        return foo.OperatorConfig(
            name="selected_labels_operator",
            label="Selected Labels Operator",
            dynamic=True,
        )

    def resolve_input(self, ctx):
        inputs = types.Object()

        inputs.message(
            "Number of Selected Labels",
            label=f"Number of Selected Label: {len(ctx.selected_labels)}",
        )

        inputs.message(
            "Number Selected Samples",
            label=f"Number Selected Samples: {len(ctx.selected)}",
        )

        return types.Property(inputs)

    def execute(self, ctx):
        return {
            "num_selected_labels": len(ctx.selected_labels),
            "num_selected_samples": len(ctx.selected),
        }

    def resolve_output(self, ctx):
        outputs = types.Object()

        outputs.message(
            "Number of Selected Labels",
            label=f"Number of Selected Label: {len(ctx.selected_labels)}",
        )

        outputs.message(
            "Number Selected Samples",
            label=f"Number Selected Samples: {len(ctx.selected)}",
[...]

PR #3558, opened 2023-09-22
fix issue 3557

ritch:
Resolves https://github.com/voxel51/fiftyone/issues/3557

This is tested with this [new example operator](https://github.com/voxel51/fiftyone-plugins/pull/58/files).

### DIFF
--- a/app/packages/operators/src/state.ts
+++ b/app/packages/operators/src/state.ts
@@ -112,8 +112,15 @@ const currentContextSelector = selectorFamily({
 
 const useExecutionContext = (operatorName, hooks = {}) => {
   const curCtx = useRecoilValue(currentContextSelector(operatorName));
-  const { datasetName, view, extended, filters, selectedSamples, params } =
-    curCtx;
+  const {
+    datasetName,
+    view,
+    extended,
+    filters,
+    selectedSamples,
+    params,
+    selectedLabels,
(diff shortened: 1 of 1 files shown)


imanjra (approved):
LGTM

app/packages/operators/src/state.ts
```diff
@@ -123,6 +130,7 @@ const useExecutionContext = (operatorName, hooks = {}) => {
         extended,
         filters,
         selectedSamples,
+        selectedLabels,
       },
       hooks
     );
```
msamadony, line 136: (Github doesn't allow me to comment on the actual line below)

Shouldn't the line below (the second argument of `useMemo`) also be adapted to include `selectedLabels` as well?

i.e. change it from:
```
}, [params, datasetName, view, extended, filters, selectedSamples, hooks]);
```

to
```
}, [params, datasetName, view, extended, filters, selectedSamples, selectedLabels, hooks]);
```

ritch: I’ll make this change. Curious that the example plugin still seems to work. Passing the arrays here may not function as expected.

msamadony: Yeah, I also wonder why. Probably one of the other dependencies gets updated too.

PR #3559, opened 2023-09-23
Only process instance masks with width and height

### LINKED ISSUE #1750: [BUG] Thumbnail view of images does not render all images
Several images are not rendered for some reason in the FiftyOne App:
image

Clicking on the image it seems the file exists and the property (image size) was read by Fifty One successfully:

image

I can also double check the image exists and can be displayed in Python:

image

benjaminpkane:
Closes #1750 

It is already the case that segmentation and heatmap overlays safely avoid processing image data without defined dimensions, but instance masks are missing this check.

Without this check, users have been encountering the this error when their browser is creating the mask `ImageData`
```
Uncaught DOMException: Failed to construct 'ImageData': The source width is zero or not a number
```

### DIFF
--- a/app/packages/looker/src/overlays/detection.ts
+++ b/app/packages/looker/src/overlays/detection.ts
@@ -43,6 +43,11 @@ export default class DetectionOverlay<
 
     if (this.label.mask) {
       const [height, width] = this.label.mask.data.shape;
+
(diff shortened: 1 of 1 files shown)


sashankaryal (approved):
LGTM! 🚀

PR #3567, opened 2023-09-25
Update Playwright .env.dev.template

benjaminpkane:
Updating the playwright `.env.dev.template` to us `FIFTYONE_ROOT_DIR`, which has replaced `PYTHONPATH`


sashankaryal (approved):
lgtm, thanks for the cleanup

PR #3577, opened 2023-09-26
EmbeddedDocument field with a label type (ex: ground_truth of type Detections) should not have other typed fields of type primitives.

manivoxel51:

An embedded document field with a label list type (ex: ground_truth of type Detections) should not have other typed fields (ground_truth.annotator for example) of type primitives.

Note this can be extended to more use cases but a good safeguard for now.

```
import fiftyone as fo

dataset = fo.load_datase("quickstart")
sample = dataset.first()
sample["ground_truth.annotator"] = "foo"
sample.save()

dataset.save()

// launch app
```

before fix

Screen Shot 2023-09-26 at 2 38 10 PM

after fix

Screen Shot 2023-09-26 at 2 08 50 PM

(Details)

        notes for FiftyOne users.

-   [ ] Other


lkaneda:
in your screenshot after fix, the field annotator got moved to live at the Detection level. did you change your code to make that happen? or are you showing it nested incorrectly?

manivoxel51:
> in your screenshot after fix, the field annotator got moved to live at the Detection level. did you change your code to make that happen? or are you showing it nested incorrectly?

good point! the two shouldn't really be compared. I added a nested annotator to detections field (in the second screenshare) which is ok to show I believe. The fix is to remove the annotator as in the first screenshare

lkaneda:
By showing it though, doesn't it imply that it can be visualized like the other fields (and that it would appear in the side bar). This is how I understand it.

manivoxel51:
> By showing it though, doesn't it imply that it can be visualized like the other fields (and that it would appear in the side bar). This is how I understand it.

right! the annotator within `ground_truth.detections` is showing in the sidebar and in field visibility no?
Screen Shot 2023-09-26 at 4 06 43 PM

lkaneda:
sweet - thanks for clarifying. lgtm!

benjaminpkane (changes requested):
I made a tweak to this because I think we only need to include the list field itself. This ensures all fields are excluded, includings list fields, dict fields, etc.

For example
```py
first = dataset.first()
first.detections["other_list"] = ["data"]
first.save()
```
We want `other_list` to be omitted from schema settings, which should be handled now.

I am seeing an issue with frame fields (see screenshot). The nested detection fields are missing `detections.detections` in the path. Is there an easy fix for that?
Screenshot 2023-09-28 at 9 04 12 AM

manivoxel51:
> I made a tweak to this because I think we only need to include the list field itself. This ensures all fields are excluded, includings list fields, dict fields, etc.
> 
> For example
> 
> ```python
> dataset = foz.load_zoo_dataset("quickstart")
> first = dataset.first()
> first.detections["other_list"] = ["data"]
> first.save()
> ```
> 
> We want `other_list` to be omitted from schema settings, which should be handled now.
> 
> I am seeing an issue with frame fields (see screenshot). The nested detection fields are missing `detections.detections` in the path. Is there an easy fix for that? Screenshot 2023-09-28 at 9 04 12 AM

thanks for testing Ben! I'll address your comment out later today

benjaminpkane (approved):
LGTM, last comment is unrelated

PR #3580, opened 2023-09-27
Only run one delegated operation at a time

brimoor:
Updating the `fiftyone delegated launch` command to only run one operation at a time.

This is functionality equivalent to the previous implementation (with a small delay between consecutive jobs) but is more parallel-friendly. Previously if a new service were launched while the first service was halfway through a batch of queued runs, the new service would start executing the same jobs and the original service would not realize that they are no longer in queued and re-execute them.


brimoor:
Replaced by https://github.com/voxel51/fiftyone/pull/3615.

PR #3581, opened 2023-09-27
Gracefully handle runs that are already migrated when importing data in FiftyOneDataset format

brimoor:
Fixes a bug where datasets that contains runs (evaluation, brain, annotation) exported in `FiftyOneDataset` format using `fiftyone<0.19` would not be importable using `fiftyone>=0.19`.

The issue was that the `fiftyone==0.19` migration script was not smart enough to handle cases where datasets need to be migrated upwards but already have runs stored in the `runs` collection.

## venv A

In a venv with `fiftyone==0.22` installed:

```shell
fiftyone migrate -v 0.18
```

## venvB

A fresh venv:

```
pip install ipython fiftyone==0.18
```

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.evaluate_detections(
    "predictions",
    gt_field="ground_truth",
    eval_key="eval",
)

dataset.export(
    export_dir="/tmp/quickstart",
)
```

## venvA

```
fiftyone migrate
```

```py
import fiftyone as fo

# This would previously raise an error but now succeeds
    dataset_dir="/tmp/quickstart",
)

assert len(dataset.list_evaluations()) == 1
```

### DIFF
--- a/fiftyone/migrations/revisions/v0_19_0.py
+++ b/fiftyone/migrations/revisions/v0_19_0.py
@@ -67,11 +67,15 @@ def _up_runs(db, dataset_dict, runs_field):
 
     _runs = {}
     for key, run_doc in runs.items():
-        _id = ObjectId()
-        run_doc["_id"] = _id
-        run_doc["_dataset_id"] = dataset_dict["_id"]
-        _runs[key] = _id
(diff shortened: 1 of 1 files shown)


lkaneda (approved):
LGTM!

PR #3582, opened 2023-09-27
Adding a delete_frame() method

brimoor:
By convention, FO exposes all dunder methods like `del sample.frames[1]` as a public method like `sample.frames.delete_frame(1)`. The latter was missing so I added it.

### DIFF
--- a/fiftyone/core/frame.py
+++ b/fiftyone/core/frame.py
@@ -105,12 +105,7 @@ class Frames(object):
         self.add_frame(frame_number, frame)
 
     def __delitem__(self, frame_number):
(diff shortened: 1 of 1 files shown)


swheaton (approved):
lgtm
reminder to self to add this (and similar dunder methods here and elsewhere) as mutators in my readonly dataset PR

brimoor:
@swheaton lol I added this *because* of the read-only PR. This was the only dunder method that didn't internally call some public method.

swheaton:
lol I looked at that one after commenting here, should've reversed order of checking email 🤣

PR #3595, opened 2023-09-28
Release patch

benjaminpkane:
Adding `develop` delta to `release/v0.22.1`. There are already features such as #3496 in `release/v0.22.1`, and the only notable new additions are small dependency changes which are low-risk

PR #3603, opened 2023-10-02
Merge/0 22 1 to develop 20231002

findtopher:
Daily merge of `release/v0.22.1` into `develop`

The following files had conflicts which I resolved, please pay particular attention to them:
```
        both modified:   app/packages/core/src/components/Actions/ActionsRow.tsx
        both modified:   app/packages/core/src/components/ColorModal/ColorFooter.tsx
        both modified:   app/packages/core/src/components/ResourceCount.tsx
        both modified:   fiftyone/utils/coco.py
```

### DIFF
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -2157,9 +2157,19 @@ def _coco_segmentation_to_mask(segmentation, bbox, frame_size):
     width, height = frame_size
 
     if isinstance(segmentation, list):
+        # Filter out empty lists. For polygons of 4 points (1 pixel), duplicate
+        # to convert to valid polygon.
+        new_segmentation = []
+        for seg in segmentation:
+            if len(seg) == 0:
+                continue
+            if len(seg) == 4:
+                seg = seg * 4
+            new_segmentation.append(seg)
+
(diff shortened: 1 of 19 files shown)


fiftyone/utils/coco.py
```diff
@@ -2157,9 +2157,19 @@ def _coco_segmentation_to_mask(segmentation, bbox, frame_size):
     width, height = frame_size
 
     if isinstance(segmentation, list):
+        # Filter out empty lists. For polygons of 4 points (1 pixel), duplicate
```
brimoor, line 2160: This is undoing https://github.com/voxel51/fiftyone/commit/4658b6ecfd7747a40f30751e2936bc9b2984fe98 which is in `develop` but wasn't included in https://github.com/voxel51/fiftyone/pull/3595 because it only included merge commits.

I think we need to revisit the task of pulling changes from `develop` into `release/v0.22.1` before doing any merges in the other direction.

brimoor (changes requested):

fiftyone/utils/coco.py
findtopher: Yeah, I had a 50/50 shot on this one - didn't do enough research and picked the wrong one.

Should be fixed now.

PR #3609, opened 2023-10-02
Databricks update

benjaminpkane:
Fixes databricks notebook URLs. An effort was made to add support for interactive notebooks,  but it seems plotly support is also limited like in Google's Colab. Interactive `ipywidgets` with callbacks do not seem to be supported, and the officially recommend using plotly [offline](https://docs.databricks.com/en/visualizations/plotly.html)

Screenshot 2023-10-02 at 2 40 41 PM


sashankaryal (approved):
lgtm!

PR #3612, opened 2023-10-03
Plugin utils

brimoor:
Adds a `fiftyone.plugins.utils` module with a few helpful utilities for parsing plugin info from GitHub repositories. All are used by https://github.com/voxel51/fiftyone-plugins/pull/66.

### DIFF
--- a/fiftyone/plugins/utils.py
+++ b/fiftyone/plugins/utils.py
@@ -20,7 +20,7 @@ logger = logging.getLogger(__name__)
 
 
 def list_zoo_plugins(info=False):
-    """Returs a list of available plugins registered in the
+    """Returns a list of available plugins registered in the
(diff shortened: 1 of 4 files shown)


fiftyone/plugins/utils.py
```diff
@@ -0,0 +1,217 @@
...
+
+from fiftyone.utils.github import GitHubRepository
+from fiftyone.plugins.core import PLUGIN_METADATA_FILENAMES
+
+
+logger = logging.getLogger(__name__)
+
+
+def list_zoo_plugins(info=False):
+    """Returs a list of available plugins registered in the
```
benjaminpkane, line 23: small typo
```suggestion
    """Returns a list of available plugins registered in the
```

brimoor: ty!

PR #3613, opened 2023-10-03
Adds a Snackbar alert when onSetView fails.

manivoxel51:

Adds a Snackbar alert when onSetView fails.
- [x] Add error handling
- [x] Add Snackbar alert

Before change

https://github.com/voxel51/fiftyone/assets/109545780/ad4ce15e-a5bb-466a-ab9d-b16a6407c1ea

After change

https://github.com/voxel51/fiftyone/assets/109545780/369e115d-e6f7-4f2e-aec6-7f9b0e83245b

@benjaminpkane @imanjra and anyone else, can you advise on the followings please,
- Is there a better way? I had to bypass ErrorBoundary because it was refreshing the page and getting tricky to show Alert and Children in ErrorBoundary without this refresh.
- How do I stop the loading state when onSetView --> onComplete returns data.errors? The top loader stays visible and the event call never gets cancelled. I tried setting AppState to Close in useEventSource.ts but no success.

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/app/src/Renderer.tsx
+++ b/app/packages/app/src/Renderer.tsx
@@ -12,10 +12,21 @@ import {
 import { Queries } from "./makeRoutes";
 import { Entry, useRouterContext } from "./routing";
 
-export const pendingEntry = atom<boolean>({
-  key: "pendingEntry",
-  default: false,
-});
+export const [pendingEntry, setPending] = (() => {
+  let set: (value: boolean) => void;
+  return [
+    atom<boolean>({
+      key: "pendingEntry",
+      default: false,
+      effects: [
+        ({ setSelf }) => {
+          set = setSelf;
+        },
+      ],
+    }),
+    (value: boolean) => set(value),
+  ];
+})();
 
 export const entry = atom<Entry<Queries> | null>({
   key: "Entry",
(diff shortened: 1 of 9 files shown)


benjaminpkane:
Hmm, I'm not opposed to the change. But the `SortBy` stage does accept an field name or an expression, which is why `json` is allowed. I think there is some error handling that needs to be added. It looks like the `setView` mutation is returning an error. This [change](https://github.com/voxel51/fiftyone/compare/bug/sortby-crash-stage...bug/sortby-crash-stage-2) could be used, or even handling the error gracefully by showing a toast?

manivoxel51:
> Hmm, I'm not opposed to the change. But the `SortBy` stage does accept a field name or an expression, which is why `json` is allowed. I think there is some error handling that needs to be added. It looks like the `setView` mutation is returning an error. This [change](https://github.com/voxel51/fiftyone/compare/bug/sortby-crash-stage...bug/sortby-crash-stage-2) could be used, or even handling the error gracefully by showing a toast?

Thanks Ben!
I merged your branch to this. Two things I'll have to add

1. Toast for when onSetView fails
2. The event seems to be pending in loading state after sortBy:non-existing-field. Let me know if you know more on how to cancel this event on failure.

Screen Shot 2023-10-04 at 11 43 24 AM

benjaminpkane (approved):
LGTM! Works great 🚀 

Regarding the `/events` request, that is ok that it is still open. It is an SSR request that stays open to receive events, it never closes unless the server disconnects. What I did add is a small hack function `rollbackViewBar` to get the view bar back to the current state. That way the page stays consistent.

PR #3614, opened 2023-10-03
fix groups aggregation problem resulting in unstable ordering of documents

### LINKED ISSUE #3421: [BUG] dynamic groups in video-only datasets returns unstable ordering of samples
### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04): macOS 13.4.1
-   **Python version** (`python --version`): 3.10.8
-   **FiftyOne version** (`fiftyone --version`): 0.21.5
-   **FiftyOne installed from** (pip or source): source

### Commands to reproduce

```
import fiftyone as fo
import fiftyone.zoo as foz

```

### Describe the problem

When creating unordered dynamic groups from the app on image or pcd only datasets, it returns a stable sample collection. But with video datasets, the returned sample collection is unstable and changes in every page refresh. 

### Willingness to contribute

-   [x] Yes. I can contribute a fix for this bug independently

sashankaryal:

This PR fixes #3421.

#### Previous Aggregation Pipeline

The previous pipeline (somewhat inexplicably) outputs an unstable ordering of documents. While the `group` stage doesn't guarantee a stable sort, the absence of the `lookup` stage, which adds the sorted `frames` object from the frames collection, results in a stable order of documents. This might be due to the presence of a nested field object, such as `frames`, outputted by the `lookup` stage, which could be affecting how the `group` stage operates.

```
[
  {
    $lookup: {
      from: "frames.samples.6421c8af194325086c697069",
      let: {
        sample_id: "$_id",
      },
      pipeline: [
        {
          $match: {
            $expr: {
              $eq: ["$$sample_id", "$_sample_id"],
            },
          },
        },
        {
          $sort: {
            frame_number: 1,
          },
        },
      ],
      as: "frames",
    },
  },
  {
    $sort: {
      "metadata.duration": 1,
      "metadata.size_bytes": 1,
    },
  },
  {
    $group: {
      _id: "$metadata.duration",
      doc: {
        $first: "$$ROOT",
      },
    },
  },
  {
    $replaceRoot: {
      newRoot: "$doc",
    },
  },
]
```

#### New Aggregation Pipeline

This PR solves the unstable sort problem by applying the `sort` stage _after_ the `group` stage.

```
[
  {
    $lookup: {
      from: "frames.samples.6421c8af194325086c697069",
      let: {
        sample_id: "$_id",
      },
      pipeline: [
        {
          $match: {
            $expr: {
              $eq: ["$$sample_id", "$_sample_id"],
            },
          },
        },
        {
          $sort: {
            frame_number: 1,
          },
        },
      ],
      as: "frames",
    },
  },
  {
    $group: {
      _id: "$metadata.duration",
      doc: {
        $first: "$$ROOT",
      },
    },
  },
  {
    $replaceRoot: {
      newRoot: "$doc",
    },
  },
    {
    $sort: {
      "metadata.duration": 1,
      "metadata.size_bytes": 1,
    },
  },
]
```

Locally.

## Release Notes

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -3437,6 +3437,10 @@ class GroupBy(ViewStage):
 
         pipeline = []
 
+        # sort so that first document in each group comes from a sorted list
+        if self._sort_stage is not None:
+            pipeline.extend(self._sort_stage.to_mongo(sample_collection))
+
         pipeline.extend(
             [
                 {"$group": {"_id": group_expr, "doc": {"$first": "$$ROOT"}}},
@@ -3444,8 +3448,13 @@ class GroupBy(ViewStage):
             ]
         )
 
-        if self._sort_stage is not None:
-            pipeline.extend(self._sort_stage.to_mongo(sample_collection))
+        # add a sort stage so that we return a stable ordering of groups
+        # sort by _id to preserve insertion order
+        pipeline.extend(
+            [
+                {"$sort": {"_id": 1}},
+            ]
+        )
 
         return pipeline
 
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -3437,15 +3437,15 @@ class GroupBy(ViewStage):
 
         pipeline = []
 
-        if self._sort_stage is not None:
-            pipeline.extend(self._sort_stage.to_mongo(sample_collection))
-
         pipeline.extend(
             [
                 {"$group": {"_id": group_expr, "doc": {"$first": "$$ROOT"}}},
                 {"$replaceRoot": {"newRoot": "$doc"}},
             ]
         )
+
+        if self._sort_stage is not None:
+            pipeline.extend(self._sort_stage.to_mongo(sample_collection))
 
         return pipeline
 
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -3450,11 +3450,7 @@ class GroupBy(ViewStage):
 
         # add a sort stage so that we return a stable ordering of groups
         # sort by _id to preserve insertion order
-        pipeline.extend(
-            [
-                {"$sort": {"_id": 1}},
-            ]
-        )
+        pipeline.append({"$sort": {"_id": 1}})
 
         return pipeline
 


brimoor (changes requested):
The reason that sorting comes before grouping in the current implementation is that it’s trying to guarantee that the $first doc that is chosen to represent each group is the lowest wrt `order_by` within each group.

For example in 

```py
view = dataset.group_by(“sample_id”, order_by=“frame_number”)
```

it’s trying to guarantee that when you load `view` in the sample grid you see frame number 1 for each group.

I guess the problem you’re trying to solve is that the order of the sample IDs in `view` may not be deterministic when the $sort stage is not the first in the pipeline? The user could fix that by appending a `sort_by(“sample_id”)` stage if they wanted. But maybe you *need* the order to always be deterministic for App navigation?

sashankaryal:
> The reason that sorting comes before grouping in the current implementation is that it’s trying to guarantee that the $first doc that is chosen to represent each group is the lowest wrt `order_by` within each group.
> 
> For example in
> 
> ```python
> view = dataset.group_by(“sample_id”, order_by=“frame_number”)
> ```
> 
> it’s trying to guarantee that when you load `view` in the sample grid you see frame number 1 for each group.
> 
> I guess the problem you’re trying to solve is that the order of the sample IDs in `view` may not be deterministic when the $sort stage is not the first in the pipeline? The user could fix that by appending a `sort_by(“sample_id”)` stage if they wanted. But maybe you _need_ the order to always be deterministic for App navigation?

Thanks for flagging that. I added a new sort stage at the end of the pipeline to guarantee a deterministic behavior.

Note: We need to guarantee a deterministic ordering because we implement pagination with a new aggregation per query with `.skip()` as opposed to reusing the same cursor.

fiftyone/core/stages.py
```diff
@@ -3447,6 +3448,14 @@ def _make_grouped_pipeline(self, sample_collection):
             ]
         )
 
+        # add a sort stage so that we return a stable ordering of groups
+        # sort by _id to preserve insertion order
+        pipeline.extend(
```
brimoor, line 3453: not: just use append() directly, not extend() with one element list

brimoor (approved):

PR #3623, opened 2023-10-04
Add build system to pyproject.toml

benjaminpkane:
Due to [PEP 517](https://peps.python.org/pep-0517/), installing `fiftyone` from source occurs in an isolated build environment. For Python 3.7, this requires the `importlib-metadata` package. This build requirement has been added to the `pyproject.toml` per [PEP 518](https://peps.python.org/pep-0518/).

I've also updated the `MANIFEST.in` because unwanted directories are currently included in a build (`e2e`, etc.). More could be done here, but this update removes most of the unnecessary files

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,8 +3,11 @@ global-exclude *.py[cod]
 
 prune .*
 prune app
+prune e2e
+prune e2e-pw
 prune eta
 prune package
+prune requirements
 prune tools
 prune tests
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,5 @@
+[build-system]
+requires = ["importlib-metadata; python_version<'3.8'", "setuptools", "wheel"]
 [tool.black]
 line-length = 79
 include = '\.pyi?$'


findtopher (approved):
Looks good - builds good with 3.7 and 3.10

:shipit:

findtopher:
It is pretty interesting that the `develop` build package is smaller than the `bug/pep517` build...

```
python3.10 bug/pep517 - make python -o app  9.49s user 1.35s system 95% cpu 11.350 total

❯ du -sk fiftyone-0.22.1a2-py3-none-any.whl
7572    fiftyone-0.22.1a2-py3-none-any.whl

python3.10 - develop - make python -o app  9.66s user 1.35s system 99% cpu 11.083 total

❯ du -sk fiftyone-0.22.0a1-py3-none-any.whl
7536    fiftyone-0.22.0a1-py3-none-any.whl
```

Entirely unimportant with relation to the PR, just interesting to me.

benjaminpkane:
The app build is definitely smaller now. In part due to #2742

PR #3629, opened 2023-10-05
improve fileview

ritch:
Improvements for fileview.

Usage example here: https://github.com/voxel51/fiftyone-plugins/pull/68


brimoor:
@ritch can we include in `relase/v0.22.1` please?

app/packages/core/src/plugins/SchemaIO/components/FileView.tsx
```diff
@@ -11,33 +11,49 @@ export default function FileView(props) {
...
+            onChange={async (files, clear) => {
+              if (files?.length === 0) {
+                return onChange(path, "");
+              }
               const [file] = files;
               const { error, result } = await fileToBase64(file);
-              if (error) return; // todo: handle error
+              if (error) {
+                clear();
+                showError("Error reading file");
```
imanjra, line 37: Should we also include the error returned here?

app/packages/core/src/plugins/SchemaIO/components/FileView.tsx
```diff
@@ -68,3 +82,16 @@ function fileToBase64(
     fileReader.onerror = (error) => resolve({ error });
   });
 }
+
+function humanReadableBytes(bytes: number): string {
```
imanjra, line 86: Can we move this to a util. This is duplicated in FileTable.tsx

fiftyone/operators/types.py
```diff
@@ -971,8 +971,34 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
 
+class UploadedFile(dict):
```
imanjra, line 974: Can we add mapping similar to https://github.com/voxel51/fiftyone/pull/3586/files#diff-136b627e19a5efadd63b2ecebf3d40f59685b433ac4ab576c052c913531951a0R46

app/packages/core/src/plugins/SchemaIO/components/FileView.tsx
ritch: Yep

app/packages/core/src/plugins/SchemaIO/components/FileView.tsx
```diff
@@ -1,43 +1,65 @@
...
-            onChange={async (files) => {
-              if (files?.length === 0) return onChange(path, "");
+            onChange={async (files, clear) => {
+              if (files?.length === 0) {
+                return onChange(path, "");
+              }
               const [file] = files;
               const { error, result } = await fileToBase64(file);
-              if (error) return; // todo: handle error
+              if (error) {
```
brimoor, line 36: nit: when `FileView` is used with an `Object` property and the value is `None`, the component currently shows "Invalid object" error message. Ideally it would show "Required property" instead.

brimoor (changes requested):

brimoor:
@ritch can we include in `release/v0.22.2` please?

ritch:
> @ritch can we include in `release/v0.22.2` please?

yep!

fiftyone/operators/types.py
ritch: This isn't a `BaseType` - but I could convert it to one. The auto mapping does infer that its a `types.Object` or `types.UploadedFile` though doesn't it? When the actual support is `string | types.Object | types.UploadedFile`...?

app/packages/core/src/plugins/SchemaIO/components/FileView.tsx
imanjra: This is due to the default value being an empty string. Fixed

imanjra (approved):
LGTM! 🔥 
I will make the changes suggested by me as a follow-up as changes in PR https://github.com/voxel51/fiftyone/pull/3652 are needed for it

ritch:
Closed in favor of https://github.com/voxel51/fiftyone/pull/3679

Issue #3457, opened 2023-08-28
[BUG] Error while uploading to labelbox for annotations


ashish-dahal:
### System information

-   **OS Platform and Distribution**: Ubuntu 20.04.6 LTS (Databricks)
-   **Python version** (`python --version`): 3.9.5
-   **FiftyOne version** (`fiftyone --version`): 0.21.6
-   **FiftyOne installed from** (pip or source): pip

### Describe the problem

While using Dataset.annotate() with backend='labelbox', the labelbox server responds with a 'Failed to upload, message: File upload error - invalid MIME type' error for .bmp images.

Reference:
[Relevant code block in FiftyOne](https://github.com/voxel51/fiftyone/blob/9ff834d53416117e622e3803bcc1a17d5d8789c7/fiftyone/utils/annotations.py#L64)

### Code to reproduce issue
```python
import requests
from PIL import Image
from io import BytesIO
import fiftyone as fo

img_path = "example.bmp"
response = requests.get('https://filesamples.com/samples/image/bmp/sample_640%C3%97426.bmp')
img = Image.open(BytesIO(response.content))
img.save(img_path)

LABELBOX_API_KEY = "<your labelbox api key>"
classes = {
    "class1",
    "class2"
}

sample = fo.Sample(filepath="example.bmp")
dataset = fo.Dataset("test")
dataset.add_sample(sample)
dataset.annotate(anno_key="test", backend="labelbox", api_key=LABELBOX_API_KEY, label_field="ground_truth", label_type="detections", classes=classes)
```

### What areas of FiftyOne does this bug affect?

-   [ ] `App`: FiftyOne application issue
-   [x] `Core`: Core Python library issue
-   [ ] `Server`: FiftyOne server issue

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the FiftyOne codebase?

-   [x] Yes. I can contribute a fix for this bug independently
-   [ ] Yes. I would be willing to contribute a fix for this bug with guidance from the FiftyOne community
-   [ ] No. I cannot contribute a bug fix at this time

ashish-dahal:
[...]

PR #3635, opened 2023-10-06
Fix issue  #3457 :  Error while uploading to labelbox for annotations

ashish-dahal:

Fix issue  #3457  (minor change)

        notes for FiftyOne users.

-   [ ] Other


brimoor:
@ashish-dahal I see that you closed the underlying issue https://github.com/voxel51/fiftyone/issues/3457. Does that mean that this PR is no longer needed?

fiftyone/core/annotation.py
```diff
@@ -1,69 +1,2569 @@
 """
-Annotation runs framework.
+Annotation utilities.
```
brimoor, line 2: There should not be any changes to `fiftyone.core.annotations`.

fiftyone/utils/annotations.py
```diff
@@ -145,6 +146,12 @@ def annotate(
             one of the supported methods. For existing label fields, if classes
             are not provided by this argument nor ``label_schema``, the
             observed labels on your dataset are used
+        content_type (None): specified the mime type of the provided samples.
```
brimoor, line 149: `Sample.filepath` is always available, so the only reason I could see needing to manually specify `content_type` is if you had a BMP-formatted image stored with a `.jpg` extension for something like that.

brimoor (changes requested):

ashish-dahal:
> @ashish-dahal I see that you closed the underlying issue #3457. Does that mean that this PR is no longer needed?

Yes, I found a simple fix: The Labelbox SDK uses the Python [`mimetypes`](https://docs.python.org/3/library/mimetypes.html#module-mimetypes) to infer image mime types. This, in turn, looks up mime types from `/etc/mime.types` on Linux. On the Databricks cluster I was using, the mime type for BMP images was `image/x-ms-bmp` instead of the expected `image/bmp`, causing Labelbox to throw an incorrect mime error. Solution was to simply update the mime type entry for BMP images in `/etc/mime.types`  to the correct value.

brimoor:
Thanks for closing the loop on this! Will close the PR since a solution has been found 👍

PR #3643, opened 2023-10-06
Adding more robust proxy support

benjaminpkane:
The FiftyOne App can be served through a reverse proxy, but currently some edge cases exist with changing the dataset route when a `pathname` is present in the proxy.

Tested with an nginx proxy on a Mac:
```
brew install nginx
```

configure the http server in `/opt/homebrew/etc/nginx/nginx.conf` with
```
        location /proxy/ {
            proxy_pass   http://localhost:5151/;
        }
```

```
brew services start nginx
```

After building the App with `yarn build` in `./app`, and starting the server with `python fiftyone/server/main.py`, http://localhost:8080/?proxy=/proxy should load the App

Notes:
* Also fixes search param persistence in view and dataset setters
* The `proxy` search param is required in order for routing to resolve fetch URLs correctly
* Introduces consistent usage of the new `resolveURL` function to ensure search and pathname behavior is guaranteed


sashankaryal:
Is `proxy` search param not undefined in `http://localhost:8080/proxy?=/proxy`?

benjaminpkane:
Included in #3746

PR #3645, opened 2023-10-09
Add polling to Colab session URL

benjaminpkane:
This change ensures colab notebooks provide a functional URL when accessing the `session.url` property by including the required `polling` search param, i.e. `https://LONG_ID-5151-colab.googleusercontent.com/?polling=true`


sashankaryal (approved):
lgtm

PR #3649, opened 2023-10-10
Move ensureColorScheme to state and fix minor color related bugs

lanzhenw:

- move `ensureColorScheme` to `fiftyone/state`
- fix minor bugs related to colorBy global changes

(Details)

        notes for FiftyOne users.

-   [ ] Other


benjaminpkane (approved):
Thanks for raising this PR @lanzhenw! This will make future changes a lot safer. Right now, the only place where I think we need a fallback object is in `Sync.tsx`, which can provide the `appConfig` which I think is a good contract (for Teams as well)

Also, thanks for the other fix as well. LGTM!

PR #3650, opened 2023-10-10
bumping versions

brimoor:


fiftyone/constants.py
```diff
@@ -41,7 +41,7 @@
 # This setting may be ``None`` if this client has no compatibility with other
 # versions
 #
-COMPATIBLE_VERSIONS = ">=0.19,<0.22"
+COMPATIBLE_VERSIONS = ">=0.19,<0.23"
```
brimoor, line 44: Looks like this was not bumped for `fiftyone=0.22`, which could cause issues for users that are trying to connect with different client versions to the same database (usually in Teams context).

package/desktop/setup.py
```diff
@@ -16,7 +16,7 @@
 import shutil
 
 
-VERSION = "0.30.2"
+VERSION = "0.30.3"
```
brimoor, line 19: I noticed that this was already bumped on `develop`, so shouldn't it be here too?

setup.py
```diff
@@ -16,7 +16,7 @@
 from setuptools import setup, find_packages
 
 
-VERSION = "0.22.1"
+VERSION = "0.22.2"
```
brimoor, line 19: I noticed that this was already bumped on `develop`, so shouldn't it be here too?

setup.py
```diff
@@ -113,7 +113,7 @@ def get_install_requirements(install_requires, choose_install_requires):
     return install_requires
 
 
-EXTRAS_REQUIREMENTS = {"desktop": ["fiftyone-desktop>=0.29,<0.30"]}
+EXTRAS_REQUIREMENTS = {"desktop": ["fiftyone-desktop>=0.30,<0.31"]}
```
brimoor, line 116: This was not bumped for `fiftyone==0.22`. Unsure of consequences but definitely seems wrong

package/desktop/setup.py
findtopher: it should be 0.31.0 on `develop` and `0.30.3` here, yes.

setup.py
findtopher: it should be 0.23.0 on develop and 0.22.2 here, sure

setup.py
findtopher: This definitely results in the `0.29` package being installed

```
root@b1f5e165669a:/# fiftyone migrate --info
Client version: 0.22.1
Compatible versions: >=0.19,<0.22

Database version: 0.22.1
root@b1f5e165669a:/# pip install fiftyone[desktop]
[stuff]
Successfully installed fiftyone-desktop-0.29.0
```

findtopher (approved):
👍

fiftyone/constants.py
findtopher: Seems to work fine on the lower bound:
```
❯ fiftyone --version
  File "/home/topher/.virtualenvs/fiftyone/bin/fiftyone", line 5, in <module>
    from fiftyone.core.cli import main
  File "/home/topher/.virtualenvs/fiftyone/lib/python3.10/site-packages/fiftyone/__init__.py", line 34, in <module>
    _fom.migrate_database_if_necessary()
  File "/home/topher/.virtualenvs/fiftyone/lib/python3.10/site-packages/fiftyone/migrations/runner.py", line 147, in migrate_database_if_necessary
    runner = MigrationRunner(head, destination)
  File "/home/topher/.virtualenvs/fiftyone/lib/python3.10/site-packages/fiftyone/migrations/runner.py", line 300, in __init__
    raise EnvironmentError(
OSError: You must have fiftyone>=0.22.1 installed in order to migrate from v0.22.1 to v0.18.0, but you are currently running fiftyone==0.18.0.

See https://voxel51.com/docs/fiftyone/getting_started/install.html#downgrading-fiftyone for information about downgrading FiftyOne.
```

brimoor: The problem is more specific: setting `COMPATIBLE_VERSIONS = ">=0.19,<0.23"` would allow for a user running `fiftyone==0.22.2` to connect to a database that has been upgraded to a hypothetical future version of say 0.22.51.

For example, since we didn't bump `COMPATIBLE_VERSIONS` for 0.22.0, a user running `fiftyone==0.22.0` would not be able to connect to a database that's been upgraded to 0.22.X.

brimoor: Perhaps db <> client version mismatches are only ever allowed/relevant when `FIFTYONE_DATABASE_ADMIN=false`. I forget the details.

PR #3651, opened 2023-10-10
Adding support for directly exporting object patches

brimoor:
This syntax for exporting object patches was previously not possible, but should have been:

```py
import fiftyone as fo
import fiftyone.zoo as foz

patches.export(
    export_dir="/tmp/patches",
    dataset_type=fo.types.ImageDirectory,
)
```

### DIFF
--- a/fiftyone/utils/patches.py
+++ b/fiftyone/utils/patches.py
@@ -23,12 +23,13 @@ class ImagePatchesExtractor(object):
 
     Args:
         samples: a :class:`fiftyone.core.collections.SampleCollection`
-        patches_field: the name of the field defining the image patches in each
-            sample to extract. Must be of type
+        patches_field (None): the name of the field defining the image patches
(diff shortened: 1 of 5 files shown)


fiftyone/utils/patches.py
```diff
@@ -45,12 +46,21 @@ class ImagePatchesExtractor(object):
...
-        patches_field,
+        patches_field=None,
         include_labels=False,
         force_rgb=False,
         force_square=False,
         alpha=None,
     ):
+        if patches_field is None:
+            if samples._is_patches:
+                patches_field = samples._label_fields[0]
```
benjaminpkane, line 57: In the situation where other label fields are copied into the patches view, will this always assign the correct patches field?

brimoor: yep `_label_fields` is a property specific to patches views that just contains the field(s) that defined the view.

PR #3655, opened 2023-10-11
Fix dataset deletion/recreation across processes

benjaminpkane:
Dataset singletons are currently keyed by name. The consequence of this is that when a dataset is deleted and recreated in one process, any other processes will raise a dataset is deleted exception if the original dataset singleton is still in memory. This causes the App server to remain in a persisted error state if the session is attached to the original (now deleted) dataset.

This is a sensitive change, but references to dataset singleton instances seem minimal, and keying by an immutable value seems more robust.

The below now works with this change.
```py
import fiftyone as fo
import fiftyone.zoo as foz

# wait for app to launch

# delete and recreate quickstart
dataset.delete()

# refreshing the App page loads the new quickstart, currently broken on `release/v0.22.2`
```

Screenshot of existing error
Screenshot 2023-10-11 at 10 29 40 AM

### DIFF
--- a/fiftyone/core/singletons.py
+++ b/fiftyone/core/singletons.py
@@ -36,7 +36,11 @@ class DatasetSingleton(type):
             instance.__init__(name=name, _create=_create, *args, **kwargs)
             name = instance.name  # `__init__` may have changed `name`
         else:
-            instance._update_last_loaded_at()
+            try:
+                instance._update_last_loaded_at()
+            except ValueError:
+                instance._deleted = True
+                return cls.__call__(name=name, _create=False, *args, **kwargs)
 
         cls._instances[name] = instance
 
--- a/fiftyone/core/singletons.py
+++ b/fiftyone/core/singletons.py
@@ -40,7 +40,9 @@ class DatasetSingleton(type):
                 instance._update_last_loaded_at()
             except ValueError:
                 instance._deleted = True
-                return cls.__call__(name=name, _create=False, *args, **kwargs)
+                return cls.__call__(
+                    name=name, _create=_create, *args, **kwargs
+                )
 
         cls._instances[name] = instance
 
--- a/tests/unittests/synchronization_tests.py
+++ b/tests/unittests/synchronization_tests.py
@@ -9,6 +9,7 @@ import os
 import unittest
 
 import fiftyone as fo
+import fiftyone.core.odm as foo
 
 from decorators import drop_datasets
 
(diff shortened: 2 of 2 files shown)


fiftyone/core/singletons.py
```diff
@@ -24,7 +26,13 @@ def __new__(metacls, *args, **kwargs):
         return cls
 
     def __call__(cls, name=None, _create=True, *args, **kwargs):
-        instance = cls._instances.pop(name, None)
+        instance = None
+        if name:
+            doc = foo.get_db_conn().datasets.find_one(
```
brimoor, line 31: Just calling out that this adds another db connection into the process of loading a dataset. Not impactful for direct mongo connections but would add more overhead to API connections.

brimoor (commented):
Just brainstorming- is there a way to detect and handle deletion/recreation without changing how singletons are keyed?

It's tricky because the database *has* to be involved somewhere; the singleton cache is simply outdated. I guess the only hope is to detect this rare case only upon failure, not checking a priori for it.

benjaminpkane:
> Just brainstorming- is there a way to detect and handle deletion/recreation without changing how singletons are keyed?
> 
> It's tricky because the database _has_ to be involved somewhere; the singleton cache is simply outdated. I guess the only hope is to detect this rare case only upon failure, not checking a priori for it.

Yeah, I can try catching the `instance._update_last_loaded_at()` attempt. That is where the exception occurs

fiftyone/core/singletons.py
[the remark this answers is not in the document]
benjaminpkane: Yes, it's certainly a low-usage code path, but I will try Brian's suggestion

benjaminpkane:
Ok comments addressed. The test added is a best attempt to simulate a the multiprocessing with orphaning mongodb collections/documents

fiftyone/core/singletons.py
```diff
@@ -36,7 +36,11 @@ def __call__(cls, name=None, _create=True, *args, **kwargs):
             instance.__init__(name=name, _create=_create, *args, **kwargs)
             name = instance.name  # `__init__` may have changed `name`
         else:
-            instance._update_last_loaded_at()
+            try:
+                instance._update_last_loaded_at()
+            except ValueError:
+                instance._deleted = True
+                return cls.__call__(name=name, _create=False, *args, **kwargs)
```
brimoor, line 43: nit: `_create=_create` might be slightly more readable to convey that you're just re-invoking the method?

brimoor (approved):
LGTM, nice solution to this! 💪

brimoor (approved):
LGTM

PR #3657, opened 2023-10-11
[Operators] unit tests for coroutine timeout

kaixi-wang:
just adding unit tests


swheaton:
And why do we want to prevent cancellation?

kaixi-wang:
> And why do we want to prevent cancellation?

annotation plugin throwing cancellation errors. still want/need to test locally with the plugin though

swheaton:
I feel like that's from the multiprocessing and we should fix that not hide it

kaixi-wang:
on second thought, yeah your right. Shielding would just make debugging harder... removed the shield, but it wouldn't hurt to keep the tests.

ritch (approved):

PR #3662, opened 2023-10-11
Fix error with Esc key

kaixi-wang:

Fix error when pressing Esc button while in grid view with and without anything selected 
- reset to new set not array
- check selected.size exists 
image

run locally. no error when press escape

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Actions/ActionsRow.tsx
+++ b/app/packages/core/src/components/Actions/ActionsRow.tsx
@@ -234,7 +234,7 @@ const Selected = ({
       setLoading(e.detail)
     );
 
-  if (samples.size < 1 && !modal) {
+  if (!samples.size || (samples.size < 1 && !modal)) {
     return null;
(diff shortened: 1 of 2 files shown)


app/packages/core/src/components/Actions/ActionsRow.tsx
```diff
@@ -234,7 +234,7 @@ const Selected = ({
       setLoading(e.detail)
     );
 
-  if (samples.size < 1 && !modal) {
+  if (!samples.size || (samples.size < 1 && !modal)) {
```
benjaminpkane, line 237: This check unfortunately hides the component in the modal when there are no selected samples. But the button should always be visible in the modal
```suggestion
  if (samples.size < 1 && !modal) {
```

benjaminpkane (approved):
LGTM! 🚀

PR #3671, opened 2023-10-14
remove unneeded Flask dep

swheaton:
Dependabot complains about Flask but I don't think we even use it anywhere, so maybe we can just get rid of it. It's not included in the fiftyone package, only when built from source. Dep line last touched 3 yrs ago. @benjaminpkane confirm?

https://github.com/voxel51/fiftyone/security/dependabot/69


benjaminpkane (approved):
Yeah, stale req. LGTM

PR #3674, opened 2023-10-16
bump desktop

benjaminpkane:
Bumping desktop to 0.31 to ensure compatibility

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,7 +113,7 @@ def get_install_requirements(install_requires, choose_install_requires):
     return install_requires
 
 
-EXTRAS_REQUIREMENTS = {"desktop": ["fiftyone-desktop>=0.31,<0.32"]}
+EXTRAS_REQUIREMENTS = {"desktop": ["fiftyone-desktop~=0.31"]}
 
 
 with open("README.md", "r") as fh:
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -113,7 +113,7 @@ def get_install_requirements(install_requires, choose_install_requires):
     return install_requires
 
 
-EXTRAS_REQUIREMENTS = {"desktop": ["fiftyone-desktop>=0.30,<0.31"]}
+EXTRAS_REQUIREMENTS = {"desktop": ["fiftyone-desktop>=0.31,<0.32"]}
```
findtopher, line 116: couldn't we just make this `fiftyone-desktop~=0.31`?  It would allow for any patch versions of `0.32`, limit it to `0.32` or higher and exclude `0.33`...

Means only bumping one version number in the requirement and feels less prone to mistakes?  (plus we could write a test that makes sure `package/desktop/setup.py` `VERISON` and that requirement version contain the same string...

Just a thought.

findtopher (dismissed):
looks legit to me

setup.py
benjaminpkane: I wasn't aware of that syntax. Makes sense to me

findtopher (approved):
I like, I love it, I want some more of it!

👍

PR #3679, opened 2023-10-17
File obj support rel

ritch:
Improvements for fileview.

Usage example here: https://github.com/voxel51/fiftyone-plugins/pull/68


imanjra (approved):
LGTM

PR #3683, opened 2023-10-18
Fix browser cache issues when upgrading

benjaminpkane:
I would like to test the upgrade process between the last two releases to confirm the issue, but based on user reports, it seems that an App bug is often encountered after upgrading because the browser is using the cached version from the previously installed release.

This change ensures the current `index.html` file is always returned which will load the correct App assets.

### DIFF
--- a/fiftyone/server/app.py
+++ b/fiftyone/server/app.py
@@ -57,9 +57,7 @@ class Static(StaticFiles):
         )
         if response.path.endswith("index.html"):
             response.headers["cache-control"] = "no-store"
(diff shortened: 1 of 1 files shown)


benjaminpkane:
I've tested on starlette 0.24 which is the minimum version requirement, and the change works.

Note this only adds the Cache-Control `no-store` response header only for the `index.html` response. All other App assets take advantage of caching.

imanjra (approved):
NIce! LGTM

PR #3685, opened 2023-10-18
primitive types validation enhancements

imanjra:

When using the `String` for a field, the empty value `""` will now be considered invalid by default.

You may opt-out of this default behaviour by setting the `allow_empty` option to `True`:

```py
    def resolve_input(self, ctx):
        inputs = types.Object()

        # when using Object.str method
        inputs.str("my_str", allow_empty=True, required=True)

        # when using String constructor directly
        inputs.define_property("my_str", types.String(allow_empty=True), required=True)

        return types.Property(inputs)
```

Using a test operator in the app

-   [] No. You can skip the rest of this section.
        notes for FiftyOne users.

See above

-   [ ] Other

### DIFF
--- a/app/packages/operators/src/types.ts
+++ b/app/packages/operators/src/types.ts
@@ -237,10 +237,21 @@ export class Property {
  * Operator type for representing a string value for operator input/output.
  */
 class OperatorString extends BaseType {
+  allowEmpty?: boolean;
+
+  /**
+   * Construct operator type for string values
+   * @param options options for defining constraints on a string value
+   * @param options.allowEmpty allow an empty string value
+   * number
+   */
+  constructor(options: { allowEmpty?: boolean } = {}) {
+    super();
+    this.allowEmpty = options.allowEmpty;
+  }
+
   static fromJSON(json: any) {
-    const Type = this;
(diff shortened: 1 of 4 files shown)


brimoor (changes requested):
@imanjra can you also ensure that empty inputs are not allowed for other property types?

For example in the code below if I remove the default values from the int and float fields, there is no validation error, but ideally I would see `Required property` for each of them.

```py
inputs.str(
    "str",
    label="String",
    required=True,
)

inputs.int(
    "int",
    label="Integer",
    default=1,
    required=True,
)

inputs.float(
    "float",
    label="Float",
    default=5.1,
    required=True,
)
```

brimoor (approved):
Works like a charm!

I suspect this might be a larger underlying issue (and if it is, feel free to go ahead and merge this PR), but note in the screen recording below that when I switch to another tab and back, I'm able to get into a state where my *required* primitive properties are empty but they are not being marked as invalid like they were *before* I switched tabs:

https://github.com/voxel51/fiftyone/assets/25985824/2b1287fe-f957-4a7f-be49-8f13592256ed

PR #3689, opened 2023-10-18
[Proposal] Remove unused e2e

benjaminpkane:
`./e2e-pw` has become the actively developed end to end project and has been integrated into ci. There **are** some tests that contain valuable coverage in the old `./e2e` project, in particular the dynamic group tests. But because the project is no longer developed, we might as well remove it. The project is still in the repository history if it needs to be referred to


findtopher:
For context - we're getting pings on vulnerable packages in this configuration that can't be remediated (packages that are no longer being maintained that have dependencies that need to be upgraded)

Removing this unused directory structure would be very helpful in preparing for the ISO audit.

benjaminpkane:
@sashankaryal Is there any other cleanup that should be done? Old documentation references?

findtopher (approved):
if @sashankaryal says it's good, then it's good!

sashankaryal (approved):
lgtm

Issue #3686, opened 2023-10-18
[BUG] Error when downloading dinov2_* zoo models


swheaton:
### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04): Mac OSX
-   **Python version** (`python --version`): 3.8.10
-   **FiftyOne version** (`fiftyone --version`): 0.22.1
-   **FiftyOne installed from** (pip or source): pip

### Describe the problem

I get an error `ValueError: Layer 'head' is an Identity layer. Use previous layer.` when downloading DINOv2 zoo models.. Below is one, I tried most of the others and got the same error.

```
>>> m=fo.zoo.load_zoo_model('dinov2-vits14-torch')
Downloading: "https://dl.fbaipublicfiles.com/dinov2/dinov2_vits14/dinov2_vits14_pretrain.pth" to /Users/stuart/.cache/torch/hub/checkpoints/dinov2_vits14_pretrain.pth
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 84.2M/84.2M [00:01<00:00, 49.3MB/s]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/stuart/dev/fiftyone_venv2/lib/python3.8/site-packages/fiftyone/zoo/models/__init__.py", line 253, in load_zoo_model
    model = fom.load_model(config_dict, model_path=model_path, **kwargs)
  File "/Users/stuart/dev/fiftyone_venv2/lib/python3.8/site-packages/fiftyone/core/models.py", line 1777, in load_model
    return config.build()
  File "/Users/stuart/dev/fiftyone_venv2/lib/python3.8/site-packages/eta/core/learning.py", line 296, in build
    return self._model_cls(self.config)
  File "/Users/stuart/dev/fiftyone_venv2/lib/python3.8/site-packages/fiftyone/utils/torch.py", line 549, in __init__
    TorchEmbeddingsMixin.__init__(
  File "/Users/stuart/dev/fiftyone_venv2/lib/python3.8/site-packages/fiftyone/utils/torch.py", line 271, in __init__
    embeddings_layer = SaveLayerTensor(model, layer_name)
  File "/Users/stuart/dev/fiftyone_venv2/lib/python3.8/site-packages/fiftyone/utils/torch.py", line 1010, in __init__
    self._setup(model, layer_name)
[...]

PR #3690, opened 2023-10-19
Fixing #3686

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/3686

This was an unintentional regression of some extra (supposed) error checking introduced in https://github.com/voxel51/fiftyone/pull/3517.

The example code from https://docs.voxel51.com/user_guide/model_zoo/models.html#dinov2-vits14-torch now works again.


swheaton (approved):
nice, didnt realize it was that simple 😃

PR #3692, opened 2023-10-19
Improve operator registry and plugin loading

kaixi-wang:
- reduce the number of times plugins are reloaded by listing operators from a global registry
- add method for forcing refresh of the registry
- for delegated operations, only run enabled operations
- disable plugins if missing requirements to run

This should prevent errors like: 

```
Delegated operation service running

To exit, press ctrl + c
Failed to register operators for plugin @jacobmarks/twilio_automation

Running operation 6529bbd8de97d1438e00bd89 (@jacobmarks/fuzzy_search/create_text_index)
Failed to register operators for plugin @jacobmarks/twilio_automation
 100% |███████████████████████████████████████████████████████████████████████████████████████████████| 193/193 [2.6m elapsed, 0s remaining, 0.8 samples/s]
Failed to register operators for plugin @jacobmarks/twilio_automation
Operation 6529bbd8de97d1438e00bd89 complete
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
Failed to register operators for plugin @jacobmarks/twilio_automation
```

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/operators/registry.py
+++ b/fiftyone/operators/registry.py
@@ -9,16 +9,20 @@ from .builtin import BUILTIN_OPERATORS
 import fiftyone.plugins.context as fopc
 
 
-def get_operator(operator_uri):
+_operator_registry = None
+
+
+def get_operator(operator_uri, include_disabled=False):
     """Gets the operator with the given URI.
 
     Args:
         operator_uri: the operator URI
+        enabled: whether to include only enabled operators (True) or any
 
     Returns:
         an :class:`fiftyone.operators.Operator`
     """
-    registry = OperatorRegistry(enabled="all")
+    registry = OperatorRegistry(enabled="all" if include_disabled else True)
     operator = registry.get_operator(operator_uri)
     if operator is None:
         raise ValueError(f"Operator '{operator_uri}' not found")
@@ -36,8 +40,16 @@ def list_operators(enabled=True):
     Returns:
         a list of :class:`fiftyone.operators.Operator` instances
     """
-    registry = OperatorRegistry(enabled=enabled)
-    return registry.list_operators(include_builtin=enabled != False)
+    global _operator_registry
+    if not _operator_registry:
+        _operator_registry = OperatorRegistry()
+    return _operator_registry.list_operators(include_builtin=enabled != False)
+
+
+def reload_registry():
+    """Reloads the operator registry."""
+    global _operator_registry
+    _operator_registry = OperatorRegistry()
 
 
 class OperatorRegistry(object):
@@ -50,6 +62,7 @@ class OperatorRegistry(object):
     def __init__(self, enabled=True):
         self.plugin_contexts = fopc.build_plugin_contexts(enabled=enabled)
 
+    # @plugins_cache
     def list_operators(self, include_builtin=True):
         """Lists the available FiftyOne operators.
 
--- a/fiftyone/operators/registry.py
+++ b/fiftyone/operators/registry.py
@@ -17,7 +17,7 @@ def get_operator(operator_uri, include_disabled=False):
 
     Args:
         operator_uri: the operator URI
-        enabled: whether to include only enabled operators (True) or any
+        include_disabled (False): whether to include disabled operators
 
     Returns:
         an :class:`fiftyone.operators.Operator`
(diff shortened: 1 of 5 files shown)


fiftyone/plugins/context.py
```diff
@@ -136,6 +136,14 @@ def register_all(self):
             sys.modules[module.__name__] = module
             spec.loader.exec_module(module)
             module.register(self)
+        except ModuleNotFoundError as e:
```
brimoor, line 139: I don't think we should disable the plugin here, for a few reasons:
- Maybe other components of the plugin will still function without the Python operators
- We don't disable the plugin a few lines below this if things fail for other reasons. Prefer consistency

FWIW I'm not a fan of automatically disabling the users stuff in either case. Import failures are already non-fatal (we log a warning), so better to raise the error every time so the user either manually disables the plugin or fixes the issue.

fiftyone/operators/registry.py
```diff
@@ -36,8 +40,22 @@ def list_operators(enabled=True):
     Returns:
         a list of :class:`fiftyone.operators.Operator` instances
     """
-    registry = OperatorRegistry(enabled=enabled)
-    return registry.list_operators(include_builtin=enabled != False)
+    global _operator_registry
```
brimoor, line 43: There are lots of places where `OperatorRegistry()` is called. If we're going to consider making it a singleton (which folks will hate, I can see it now) then that would need to be done holistically, not just here.

But either way, as willing as I am to introduce singletons, I am hesitant to do that for plugins for fear of losing the ability to dynamically edit a plugin's source and immediately see the results in a running App.

brimoor: Also `OperatorRegistry()` calls `build_plugin_contexts()` which already has a `@plugins_cache` decorator that I believe was added to optimize reloads.

brimoor (changes requested):

brimoor:
@kaixi-wang I believe the reason for the spammed message in the PR description is right here:
https://github.com/voxel51/fiftyone/blob/267180d2afb3ddfc4adb13c13c1a5352bd56ee83/fiftyone/factory/repos/delegated_operation.py#L284

`list_operations()` currently uses `OperatorRegistry(enabled="all")` but it should just use `OperatorRegistry()`. The only reason that a registry is used there is to retrieve the operator's label. It probably used `enabled="all"` thinking that it should be nice and retrieve as many labels as possible. But that was a bad choice as it means that disabled operators are imported. So, better to just use `OperatorRegistry()` and not have labels for disabled operators.

This will all be a moot point in `fiftyone==0.23` anyway as operations have their own labels. The use of `OperatorRegistry()` in the code above is already gone.

fiftyone/operators/registry.py
kaixi-wang: > fear of losing the ability to dynamically edit a plugin's source and immediately see the results in a running App.

I think there is a better way of handling this by importing the operator when it's actually being called. It can still be registered beforehand. I don't think we should reimport every operator every time list_operators is called (current behavior).

kaixi-wang:
> `list_operations()` currently uses `OperatorRegistry(enabled="all")` but it should just use `OperatorRegistry()`
I don't see list_operations using OperatorRegistry() at all on devellop...

brimoor:
> > `list_operations()` currently uses `OperatorRegistry(enabled="all")` but it should just use `OperatorRegistry()`
> 
> I don't see list_operations using OperatorRegistry() at all on develop...

Exactly, I said that:

> This will all be a moot point in fiftyone==0.23 anyway as operations have their own labels. The use of OperatorRegistry() in the code above is already gone in develop.

fiftyone/plugins/context.py
kaixi-wang: I disagree. If a module is missing, the plugin shouldn't work at all/will not work and it's failure should not keep spamming the logs. Also for the delegated operators case, it'll keep trying to run an operator that cannot be imported.

PR #3694, opened 2023-10-19
Delegated operation tweaks

brimoor:
- Fixes the spammed "Failed to register operators for plugin..." message discussed in https://github.com/voxel51/fiftyone/pull/3692#issue-1951043003
- Adds a `foo.operator_exists()` utility

### DIFF
--- a/fiftyone/factory/repos/delegated_operation.py
+++ b/fiftyone/factory/repos/delegated_operation.py
@@ -281,7 +281,7 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
         if paging.limit:
             docs = docs.limit(paging.limit)
 
-        registry = OperatorRegistry(enabled="all")
+        registry = OperatorRegistry()
         return [
             DelegatedOperationDocument().from_pymongo(doc, registry=registry)
             for doc in docs
--- a/fiftyone/factory/repos/delegated_operation_doc.py
+++ b/fiftyone/factory/repos/delegated_operation_doc.py
@@ -92,7 +92,7 @@ class DelegatedOperationDocument(object):
         # generated fields:
         try:
             if registry is None:
-                registry = OperatorRegistry(enabled="all")
+                registry = OperatorRegistry()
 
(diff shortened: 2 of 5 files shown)


ritch (approved):
LGTM

fiftyone/factory/repos/delegated_operation.py
```diff
@@ -281,7 +281,7 @@ def list_operations(
         if paging.limit:
             docs = docs.limit(paging.limit)
 
-        registry = OperatorRegistry(enabled="all")
+        registry = OperatorRegistry()
```
kaixi-wang, line 284: this is gone on develop and I don't think you should filter here as I think you should be able to see/list previous runs of disabled operators

fiftyone/factory/repos/delegated_operation_doc.py
```diff
@@ -92,7 +92,7 @@ def from_pymongo(self, doc: dict, registry: OperatorRegistry = None):
         # generated fields:
         try:
             if registry is None:
-                registry = OperatorRegistry(enabled="all")
+                registry = OperatorRegistry()
```
kaixi-wang, line 95: again I don't think filtering should be on the doc model

fiftyone/operators/registry.py
```diff
@@ -40,6 +45,21 @@ def list_operators(enabled=True):
     return registry.list_operators(include_builtin=enabled != False)
 
 
+def operator_exists(operator_uri, enabled=True):
```
kaixi-wang, line 48: I don't know why we need this. There already is `registry.operator_exists`

kaixi-wang (changes requested):

fiftyone/operators/registry.py
brimoor: You needed it in https://github.com/voxel51/fiftyone/pull/3692. You wrote:

```py
if op.operator in foo.list_operators():
```

which would be nicer to read as:

```py
if foo.operator_exists(op.operator):
```

PR #3695, opened 2023-10-19
Add support for custom runs

brimoor:
## Change log

- Adds support for running custom evaluation methods via `evaluate_XXX(..., method=YourCustomConfig, ...)`
- If a custom evaluation is not available when the results are loaded, gracefully fallback to the appropriate base class
    - This allows for running analyses like `results.print_report()` that are implemented on the base class even if the classes associated with the custom evaluation protocol are not available anymore/on the current machine
- Adds support for "properly" configuring custom evaluation classes via a new `fo.evaluation_config`
    - This is exactly analogous to how custom annotation/similarity backends can be configured
- Adds support for storing arbitrary custom runs on datasets via a new API 
    - This is an extension of the annotation/evaluation/brain API for arbitrary user data

## Custom evaluation example

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.eval.coco as fouc

class CustomDetectionConfig(fouc.COCOEvaluationConfig):
    pass

class CustomDetection(fouc.COCOEvaluation):
    pass

dataset.persistent = True

dataset.evaluate_detections(
    "predictions",
    gt_field="ground_truth",
    method=CustomDetectionConfig,
    eval_key="custom",
)
```

Later in another session:

```py
import fiftyone as fo

print(dataset.get_evaluation_info("custom"))
results = dataset.load_evaluation_results("custom")
results.print_report()
```

## Custom run example

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.persistent = True

config = dataset.init_run()
config.foo = "bar"  # add as many key-value pairs as you need

# Also possible
# config = fo.RunConfig(foo="bar")

dataset.register_run("custom", config)

results = dataset.init_run_results("custom")
results.spam = "eggs"  # add as many key-value pairs as you need

# Also possible
# results = fo.RunResults(dataset, config, "custom", spam="eggs")

dataset.save_run_results("custom", results)
```

Later in another session:

```py
import fiftyone as fo

print(dataset.get_run_info("custom"))
results = dataset.load_run_results("custom")
print(results)
```


allenleetc (commented):
LGTM :rocket:. I tested this as a user. Not sure if you want to merge immediately or await developments

brimoor:
@allenleetc sweet thanks! Let's wait to merge until we've had more design discussion. Specially around how to invoke custom evaluation methods from operators like `@voxel51/evaluation/evaluate_model`.

PR #3698, opened 2023-10-19
fix key for group dataset statistics

manivoxel51:

Regression caused by [this PR](https://github.com/voxel51/fiftyone/commit/491c237b4f9327ae16ecad25a85460e6db85ae98). The reason is a wrong key in `groupStatistics -> graphQLSyncFragmentAtomFamily` which was `dataset`. 

before fix
Screen Shot 2023-10-12 at 8 17 27 PM

after fix

https://github.com/voxel51/fiftyone/assets/109545780/e946d725-c879-4d17-a94a-dc362103a316

(Details)

        notes for FiftyOne users.

-   [ ] Other


benjaminpkane (commented):
Good catch here! This was an oversight by me. It might not make sense to keep this as a graphql sync atom. On `main`, this setting is a regular atom value. We could probably revert it to that. Though, this setting only applies to datasets/views when the `hasGroupSlices` selector is `true`

manivoxel51:
> Good catch here! This was an oversight by me. It might not make sense to keep this as a graphql sync atom. On `main`, this setting is a regular atom value. We could probably revert it to that. Though, this setting only applies to datasets/views when the `hasGroupSlices` selector is `true`

@benjaminpkane could you check the latest commit - I don't have a full context as to why these were changed but revert to simple atom seems to work.

app/packages/state/src/recoil/groups.ts
```diff
@@ -334,16 +334,10 @@ export const activeModalSample = selector({
...
->(
-  { fragments: [datasetFragment], keys: ["dataset"], default: "slice" },
-  {
-    key: "groupStatistics",
-  }
-);
+export const groupStatisticsState = atom<"group" | "slice">({
+  key: "groupStatisticsState",
+  default: "slice",
+});
```
benjaminpkane, line 340: This should be reverted to original boolean a atom family, or else toggling the value in the modal will cause unwanted grid updates
```suggestion
export const groupStatistics = atomFamily<"group" | "slice">({
  key: "groupStatistics",
  default: "slice",
});
```

benjaminpkane (approved):
Tested. LGTM

PR #3699, opened 2023-10-19
strip b64 prefix from uploaded files

ritch:
Cleanup the `UploadedFile` to now only included the base64 data and not the type prefix (if it exists).

### DIFF
--- a/app/packages/core/src/plugins/SchemaIO/components/FileView.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/FileView.tsx
@@ -50,15 +50,16 @@ export default function FileView(props) {
                 return;
               }
               setCurrentError(null);
+              const resultStripped = stripBase64Prefix(result);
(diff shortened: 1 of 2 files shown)


app/packages/core/src/plugins/SchemaIO/components/FileView.tsx
```diff
@@ -86,3 +87,7 @@ function fileToBase64(
     fileReader.onerror = (error) => resolve({ error });
   });
 }
+
+function stripBase64Prefix(data: string): string {
+  return data.replace(/^data:image\/[a-z]+;base64,/, "");
```
brimoor, line 92: This could be other MIME types too, like `data:application/json;base64,XXX`.

In general I think splitting on the first occurrence of `base64,` is sufficient

imanjra (approved):
LGTM. ty!

Issue #3257, opened 2023-07-06
[BUG] FiftyoneDataset path of media fields are not merged


Sa-Schmi:
### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04): Ubuntu 22.04
-   **Python version** (`python --version`): 3.10.8
-   **FiftyOne version** (`fiftyone --version`): 0.21.0
-   **FiftyOne installed from** (pip or source): pip

### Describe the problem

If the `rel_dir` argument is used for exporting a `fo.types.FiftyOneDataset` as mentioned [here](https://docs.voxel51.com/user_guide/export_datasets.html#fiftyonedataset) the `rel_dir` is removed from all media field paths (Segmentation, Heatmap etc.). [Source](https://github.com/voxel51/fiftyone/blob/1a4c3cac247494bb4c6dbbb841d937c26c87ad4d/fiftyone/utils/data/exporters.py#L2194)

When importing such a dataset, the `rel_dir `is not used to fix the paths for these media fields. I do not see an easy fix, as there is no information which media fields exists in a sample. [Source](https://github.com/voxel51/fiftyone/blob/1a4c3cac247494bb4c6dbbb841d937c26c87ad4d/fiftyone/utils/data/importers.py#L1848)

Quick and dirty fix:
```python
for key, value in sd.items():
    try:
        if value.get("_cls") is not None:
            for subkey, _ in value.items():
                if "path" in subkey:
                    sd[key][subkey] = os.path.join(rel_dir, sd[key][subkey])
    except Exception as ex: 
        print(ex)
```

### What areas of FiftyOne does this bug affect?

-   [ ] `App`: FiftyOne application issue
-   [x] `Core`: Core Python library issue
-   [ ] `Server`: FiftyOne server issue

marian-beszedes:
Same experience. FiftyOne v1.0.1

Export dataset (images are stored in `workspaces/images`) :

`export_dir = "/workspaces/datasets/my_dataset"
rel_dir="/workspaces/"

valid_objects_view.export(
    export_dir=export_dir,
    dataset_type=fo.types.FiftyOneDataset,
    export_media=False,
    rel_dir=rel_dir
)`

Import dataset:

`
export_dir = "/workspaces/datasets/my_dataset"
rel_dir="/workspaces/"
dataset_name = "my_dataset"

cleaned_labels_dataset = fo.Dataset.from_dir(
    dataset_dir=export_dir,
[...]

PR #3704, opened 2023-10-20
Expand relative `mask_path` and make paths valid for used OS (#3257)

denkrau:

When importing a dataset relative `mask_path` keys get expanded as well.
This even works with nested structures.
Additionally, `filepath` and `mask_path` path separators get converted based on the OS.

Importing a dataset in Ubuntu (which was exported in Windows) with relative `mask_path` and check if path is absolute and valid now.

        notes for FiftyOne users.

-   [x] Other

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -1368,6 +1368,26 @@ def safe_relpath(path, start=None, default=None):
     return relpath
 
 
+def expand_mask_paths(sample, rel_path):
+    """Finds all `mask_path` keys inside a serialized sample recursively and expand relative paths.
+    It is assumed that `mask_path` can only exist in `dict`.
+    
+    Args:
+        sample: a serialized sample
+        rel_path: part of path to make relative paths absolute
+    """
+    if isinstance(sample, dict):
+        for key, value in sample.items():
+            if key == "mask_path" and not os.path.isabs(key):
+                # Expand relative path
+                sample[key] = os.path.join(rel_path, value)
+            elif isinstance(value, (dict, list, tuple)):
+                expand_mask_paths(value, rel_path)
+    elif isinstance(sample, (list, tuple)):
+        for item in sample:
+            expand_mask_paths(item, rel_path)
+
+
 def compute_filehash(filepath, method=None, chunk_size=None):
     """Computes the hash of the given file.
 
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -1379,13 +1379,28 @@ def expand_mask_paths(sample, rel_path):
     if isinstance(sample, dict):
         for key, value in sample.items():
             if key == "mask_path" and not os.path.isabs(key):
-                # Expand relative path
-                sample[key] = os.path.join(rel_path, value)
+                expanded = os.path.join(rel_path, value)
+                sample[key] = make_path_os_safe(expanded)
             elif isinstance(value, (dict, list, tuple)):
                 expand_mask_paths(value, rel_path)
     elif isinstance(sample, (list, tuple)):
         for item in sample:
             expand_mask_paths(item, rel_path)
+
+
+def make_path_os_safe(path):
+    """Convert path delimiters depending on OS.
+    
+    Args:
+        path: path to make OS safe
+        
+    Returns:
+        valid path
+    """
+    if sys.platform == "win32":
+        return path.replace("/", "\\")
+    else:
+        return path.replace("\\", "/")
 
 
 def compute_filehash(filepath, method=None, chunk_size=None):
--- a/fiftyone/utils/data/importers.py
+++ b/fiftyone/utils/data/importers.py
@@ -1856,6 +1856,10 @@ class FiftyOneDatasetImporter(BatchDatasetImporter):
                 _parse_media_fields(sd, media_fields, rel_dir)
 
             sd["_dataset_id"] = dataset_id
+
+            # Find all mask_path keys recursively and extend relative paths
+            fou.expand_mask_paths(sd, rel_dir)
+            
             return sd
 
         sample_ids = foo.insert_documents(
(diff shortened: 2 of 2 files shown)


brimoor (changes requested):
Hi @denkrau thanks for the PR!

Can you explain a bit more about your use case? The existing implementation of the `FiftyOneDataset` format was designed to handle relative -> absolute path conversions for you, but it sounds like it's not working in your case?

The example below demonstrates this working as expected:

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "coco-2017",
    split="validation",
    label_types="segmentations",
    max_samples=10,
)

# Create a `Segmentation` field with masks stored on disk 
    dataset,
    "ground_truth",
    "segmentations",
    output_dir="/tmp/segmentations",
    overwrite=True,
)

# Export in FiftyOneDataset format
dataset.export(
    export_dir="/tmp/fod",
)
```

The directory structure of `/tmp/fod` is like this:

```
/tmp/fod/
    data/
        000000001000.jpg
        ...
    fields/
        segmentations/
            000000001000.png
            ...
    metadata.json
    samples.json
```

And here's how the `samples.json` stores the mask paths (as relative to this JSON file):

```
$ python -m json.tool /tmp/fod/samples.json

            ...
            "segmentations": {
                "_id": {
                    "$oid": "6532854e5f4d8589d27625e8"
                },
                "_cls": "Segmentation",
                "tags": [],
                "mask_path": "fields/segmentations/000000001000.png"
            }
        }
    ]
}
```

When I then import this dataset:

```py
    dataset_dir="/tmp/fod",
)

print(dataset2.first())
```

the absolute mask paths are correctly constructed:

```
    ...
    'segmentations': <Segmentation: {
        'id': '6532854e5f4d8589d27625df',
        'tags': [],
        'mask': None,
        'mask_path': '/tmp/fod/fields/segmentations/000000000139.png',
    }>,
}>
```

denkrau:
Hey @brimoor, thanks for the quick reply.

Indeed your sample is working, so this got me wondering.

**tl;dr: It's about `export_media` parameter during export. When it is `None` or `True` as in your example everything works fine, otherwise it does not.** 

In my case I parse samples from (a network) drive and create a dataset. Data copying must be avoided, this means `export_media=False` if I understand correctly. I was able to reproduce both the error and success with a dummy dataset by setting that parameter accordingly. Somehow, when data is copied the new path of the segmentation mask gets correctly expanded.

But you can test it with your code example as well. Just set `fo.config.dataset_zoo_dir = "/tmp/fiftyone_zoo"` to have a common prefix, add `rel_dir="/tmp"` to the export and import call, and set `export_media=False`. This will give an unexpanded `mask_path`. See the following summary.

| Keys                          | With `export_media=None` or `True`                                                                                                                                                                     | With `export_media=False` and `rel_dir="/tmp"`                  |
|-------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------|
| `samples.json`: `filepath`    | `data/000000000139.jpg`                                                                                                                                                                                | `fiftyone_zoo/coco-2017/validation/data/000000000139.jpg`      |
| `samples.json`: `mask_path`   | `fields/segmentations/000000000139.png`                                                                                                                                                                | `segmentations/000000000139.png`                               |
| File structure                | /tmp/fod/<br>&nbsp;&nbsp;data/<br>&nbsp;&nbsp;&nbsp;&nbsp;000000001000.jpg<br>&nbsp;&nbsp;&nbsp;&nbsp;...<br>&nbsp;&nbsp;fields/<br>&nbsp;&nbsp;&nbsp;&nbsp;segmentations/<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;000000001000.png<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;...<br>&nbsp;&nbsp;metadata.json<br>    &nbsp;&nbsp;samples.json | /tmp/fod/<br>&nbsp;&nbsp;metadata.json<br>&nbsp;&nbsp;samples.json             |
| Imported dataset: `filepath`  | `/tmp/fod/data/000000000139.jpg`                                                                                                                                                                       | `/tmp/fiftyone_zoo/coco-2017/validation/data/000000000139.jpg` |
| Imported dataset: `mask_path` | `/tmp/fod/fields/segmentations/000000000139.png`                                                                                                                                                       | `segmentations/000000000139.png`                               |

brimoor:
@denkrau your example with `export_media=False` and `rel_dir="/tmp"` is working as intended.

The `rel_dir` parameter is used to strip a common prefix from each filepath. When `export_media=False`, this results in exporting relative filepaths rather than absolute ones. When `export_media=True`, it causes the media to be exported in a tree structure that matches the relative paths (rather than just exporting all media in a flat directory based on the filenames of all media in the collection you're exporting).

denkrau:
> Generally you wouldn't combine export_media=False with the rel_dir option.

Thanks for the clarification. So there is no way to avoid copying/restructuring and to have all relative paths expanded? Is such a feature even desired from the project's point of view? If not, could it be possible to use file links with `export_media=True`, toggled by an additional parameter?

I'm just trying to figure out how to proceed with this or another PR.

brimoor:
Can you clarify what you're trying to achieve that's not possible? I'm not sure 😅

denkrau:
Sure, thanks for your help, here we go.

There are several raw datasets available on a shared network drive. Multiple users access them. The mount point for each user might vary. Now for some raw datasets FiftyOne dataset versions shall be created (i.e. exported) additionally, so users can import them on demand. The raw datasets cannot be moved/restructured, because some users work with this structure. Also I would like to avoid copying raw datasets for disk space and maintaining reasons.
In conclusion I need no data restructuring, no data copying and correctly expanded relative paths when importing FiftyOne datasets.
Does this make it more clear?

In my understanding of FiftyOne so far with the insights you gave me, such a scenario is not possible to achieve. Because it would require to expand all relative paths from `samples.json` (including the mask paths), while keeping the original file structure. Right?

PR #3707, opened 2023-10-20
Bump sphinx to 3.5.4

benjaminpkane:
Bumps the Sphinx docs requirement to 3.5.4. #2742 broke sphinx builds using 2.4.4 (only in Python 3.9)

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -7,6 +7,6 @@ jupyter-client==6.1.3
 myst-parser==0.13.7
 nbsphinx==0.8.8
 sphinx-tabs==1.2.1
-Sphinx==2.4.4
+Sphinx==3.5.4
 sphinxcontrib-napoleon==0.7
 sphinx-copybutton==0.4.0


brimoor:
@benjaminpkane have you successfully built the docs using Sphinx 3.5.4? I believe I previously tried to upgrade (maybe to a different version, can't remember) but it broke certain layout aspects of the docs and I had to revert to Sphinx 2.4.4

swheaton:
Trying...

benjaminpkane:
I am checking the artifact build now. Can be downloaded [here](https://github.com/voxel51/fiftyone/suites/17457717092/artifacts/997843041)

swheaton:
Built locally as well, seems normal to me?

swheaton:
Order of classes and functions is swapped for some reason? dunno if that's a big deal

brimoor (approved):
I agree, automated and local builds are looking good! 💪

PR #3708, opened 2023-10-20
[Proposal] Automate release branch syncing

benjaminpkane:
I'm not sure this workflow is correct yet, but the concept is to create (or update) a PR from any release branch into develop when a merge into a release branch occurs.


benjaminpkane:
Not quite there. Closing until I can make this functional

PR #3706, opened 2023-10-20
Merge/main to develop

findtopher:
merge of `main` to `develop` after the v0.22.2 release

PR #3710, opened 2023-10-20
Fix App on develop

benjaminpkane:
Changes from https://github.com/voxel51/fiftyone/pull/3706 seem to require that plugins are loaded after the page query and session info are ready.

This change ensures plugins are loaded after the App is initialized


imanjra (approved):
Tested locally. LGTM

PR #3714, opened 2023-10-23
bugfix: last used brain key should be loaded by default in similarity sort

lanzhenw:

Fixed the regression that last used brain key was not loaded by default.
Updated the unique identifier of dataset when we store the key in local storage

https://github.com/voxel51/fiftyone/assets/17770824/339f86af-4135-4f3c-b432-a07f97bebb37

        notes for FiftyOne users.

-   [ ] Other


benjaminpkane (approved):
Good catch! `id` can change unfortunately, and the name + createdAt definitely solved the issue. But just to ensure an immutable key (even after dataset rename) is always available, I added `datasetId` to the dataset object. So we can use this for other features too 🚀

lanzhenw:
Thanks for adding the immutable key!

PR #3718, opened 2023-10-23
sample (de)select keyboard shortcut for modal and Grid

manivoxel51:

Shortcuts for selecting and deselecting samples quicker
- [x] when sample modal open, pressing `x` will toggle current sample selection
- [x] when in Grid, holding down `Meta` or `Control` key and pressing `x` toggles targeted sample selection
- [x] Add shortcut to help panel

https://github.com/voxel51/fiftyone/assets/109545780/3fbf35fe-4d6a-47b3-8d64-d37f0653c97f

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Actions/ActionsRow.tsx
+++ b/app/packages/core/src/components/Actions/ActionsRow.tsx
@@ -357,7 +357,6 @@ const SaveFilters = () => {
       async () => {
         const loading = await snapshot.getPromise(fos.savingFilters);
         const selected = await snapshot.getPromise(fos.selectedSamples);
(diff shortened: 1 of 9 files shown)


benjaminpkane (approved):
LGTM! I think using the native `event` data is a more self-contained solution so I added the `MouseEvent` to the `onItemClick` flashlight callback. This way the event handlers rely on less external state.

Also saw an opportunity to update some of the `useRecoilCallback` details. Inside those callbacks, we have access to `set`, `reset`, and. `snapshot` so there is no need to declare external `useRecoil...` dependencies

PR #3720, opened 2023-10-24
bugfix: boolean value cannot display correctly in the sample modal view

lanzhenw:
fix the issue that boolean values are not displayed in the sample modal view. 

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/utils/generic.tsx
+++ b/app/packages/core/src/utils/generic.tsx
@@ -29,7 +29,7 @@ export const prettify = (
     );
   }
 
-  return v;
+  return result;
 };
 
 export const genSort = (a, b, asc) => {


app/packages/core/src/utils/generic.tsx
```diff
@@ -29,7 +29,7 @@ export const prettify = (
     );
   }
 
-  return v;
+  return result;
```
manivoxel51, line 32: ah I see. makes sense! the pretty() utility function is taking care of converting it into string ("True" or "False") in the following code

```
export const prettify = (
  v: boolean | string | null | undefined | number | number[]
): URL | string => {
  if (typeof v === "string") {
    if (isURL(v)) {
      try {
        return new URL(v);
      } catch {}
    }

    return v;
  } else if (typeof v === "number") {
    return Number(v.toFixed(3)).toLocaleString();
  } else if (v === true) {
    return "True";
  } else if (v === false) {
    return "False";```

manivoxel51 (approved):
🍨

PR #3726, opened 2023-10-27
Bugfix: out of sync metadata.json on export

swheaton:
Reload dataset document before exporting metadata.json in case it has been updated by another process.
Samples are exported straight from database backing so we must reload dataset doc otherwise metadata.json may not match samples.json.

Note there could still be a race condition between 2 processes but the window for that is quite small.

Scripts below written by a reporting user. Thanks.

Run first script then run second script. Make sure `field` shows up in metadata.json as well as samples.json

```python
# main1.py
import time

import fiftyone as fo

DATASET_NAME = "quickstart"
EXPORT_PATH = "export-test"

dataset: fo.Dataset = fo.zoo.load_zoo_dataset(DATASET_NAME)
dataset.export(
    export_dir=EXPORT_PATH,
    export_media=False,
    overwrite=True,
)

print("SLEEPING, RUN YOUR SECOND SCRIPT")
time.sleep(30)

dataset: fo.Dataset = fo.load_dataset(DATASET_NAME)
dataset.export(
    export_dir=EXPORT_PATH,
    export_media=False,
    overwrite=True,
)
```
```python
# main2.py
import fiftyone as fo

DATASET_NAME = "quickstart"
dataset: fo.Dataset = fo.load_dataset(DATASET_NAME)

dataset.add_sample(
    fo.Sample(
        "blah.jpg",
        field1=fo.DynamicEmbeddedDocument(some_field="abc")
    ),
    dynamic=True
)
dataset.save()
```

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/data/exporters.py
+++ b/fiftyone/utils/data/exporters.py
@@ -2118,6 +2118,7 @@ class FiftyOneDatasetExporter(BatchDatasetExporter):
             )
 
         dataset = sample_collection._dataset
+        dataset._doc.reload()
         dataset_dict = dataset._doc.to_dict()
         dataset_dict["saved_views"] = []
         dataset_dict["annotation_runs"] = {}


ehofesmann (commented):
Would it make sense to add this reload to the base [DatasetExporter.log_collection()](https://github.com/voxel51/fiftyone/blob/2876d846664298aefb6532a236e5cab70ede973c/fiftyone/utils/data/exporters.py#L1312) method and the [BatchDatasetExporter.export_samples()](https://github.com/voxel51/fiftyone/blob/2876d846664298aefb6532a236e5cab70ede973c/fiftyone/utils/data/exporters.py#L1377) method?

It seems like it would be generally useful for all exporters to first reload the dataset doc before performing the export.

swheaton:
The pattern of classes does not easily allow for this because there's no code that's actually run in BatchDatasetExporter. It's more like an interface than a base class.
Lemme think if we can put it somewhere common

```python
    def export_samples(self, sample_collection):
        """Exports the given sample collection.

        Args:
            sample_collection: a
                :class:`fiftyone.core.collections.SampleCollection`
        """
        raise NotImplementedError("subclass must implement export_samples()")
```

brimoor (approved):
LGTM

swheaton:
ok sgtm, lets merge it then

PR #3728, opened 2023-10-27
add LazyFieldView to OperatorIO

imanjra:

Add a new LazyFieldView to OperatorIO which only applies changes in the input field when a save button within an input field is clicked or optionally (enabled by default) when you blur the input field. When a field is unsaved, reset button (to go back to the previously applied or initial state) and save button (to apply changes) will appear

Preview (unsaved state of a LazyFieldView):
image

Usage:

```
def resolve_input(self, ctx):
    inputs = types.Object()
    inputs.str("url", view=types.LazyFieldView(save_on_blur=False))
    return types.Property(inputs)
```

Using a test operator in the app

        notes for FiftyOne users.

```
- Added a new :class:`LazyFieldView <fiftyone.operators.types.LazyFieldView>` type that extends capabilities of :class:`FieldView <fiftyone.operators.types.FieldView >` to allow applying user changes in an input field only with a click of a save button within an input field; or, optionally (enabled by default) with a input field blur event on an input field `#3728<https://github.com/voxel51/fiftyone/pull/3728>`_
```

-   [ ] Other

### DIFF
--- a/fiftyone/operators/types.py
+++ b/fiftyone/operators/types.py
@@ -1378,6 +1378,28 @@ class FieldView(View):
         super().__init__(**kwargs)
 
 
+class LazyFieldView(View):
+    """Displays a lazy text input which only apply input field changes on blur
+    or when user clicks the save button within the field.
+
+    .. note::
+
+        Must be used with :class:`String` or :class:`Number` properties.
+
+    Args:
+        save_on_blur (True): when set to False, changes in input field will not
+            be automatically applied when user moves mouse out of the changed
+            field. To apply changes, user must click the save button.
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.save_on_blur = kwargs.get("save_on_blur", True)
+
+    def to_json(self):
+        return {**super().to_json(), "save_on_blur": self.save_on_blur}
+
+
 class DropdownView(Dropdown):
     """Displays a dropdown selector input."""
 
--- a/app/packages/core/src/plugins/SchemaIO/components/LazyFieldView.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/LazyFieldView.tsx
@@ -1,0 +1,92 @@
+import { Replay, Save } from "@mui/icons-material";
+import { Box, IconButton, Stack } from "@mui/material";
+import { merge } from "lodash";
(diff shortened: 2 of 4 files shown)


fiftyone/operators/types.py
```diff
@@ -1378,6 +1378,23 @@ def __init__(self, **kwargs):
...
+    """Displays a lazy text input which only apply input field changes on blur
+    or when user clicks the save button within the field.
+
+    .. note::
+
+        Must be used with :class:`String` or :class:`Number` properties.
+
+    Args:
+        disable_save_on_blur (None): prevents applying field changes on blur.
+        When set to True, user must click the save button to apply the changes.
```
brimoor, line 1391: For Python docstrings, I believe Sphinx requires lines 2 and beyond of multiline parameter descriptions to be indented 4 spaces in order to be correctly rendered. For example:
https://github.com/voxel51/fiftyone/blob/48b050ba52de95dd7b8501d807485632f24ca172/fiftyone/core/dataset.py#L2864

fiftyone/operators/types.py
```diff
@@ -1378,6 +1378,23 @@ def __init__(self, **kwargs):
...
+class LazyFieldView(View):
+    """Displays a lazy text input which only apply input field changes on blur
+    or when user clicks the save button within the field.
+
+    .. note::
+
+        Must be used with :class:`String` or :class:`Number` properties.
+
+    Args:
+        disable_save_on_blur (None): prevents applying field changes on blur.
```
brimoor, line 1390: nit: no strong preference but could consider `save_on_blur=False` to be more concise, with default being `save_on_blur=True`

brimoor (commented):
I defer to code owners for review but LGTM! Just left a small comment

fiftyone/operators/types.py
swheaton: Just lurking unwanted here but @imanjra as a plug-in dev i have no idea what blur means in this context, would be helpful to explain in the docstring

ritch (approved):
LGTM. Please add an example to the fiftyone-plugins examples.

imanjra:
@ritch Thank you for the review! Added an example operator https://github.com/voxel51/fiftyone-plugins/pull/89

PR #3730, opened 2023-10-27
Remove max_depth constraint from plugin listing/caching

brimoor:
- update `@plugin_cache` so that it always changes whenever any addition/edit/removal of plugins occurs (including `touch PLUGINS_DIR`)
- remove `max_depth` flag from `download_plugin()`, for consistency with how `list_plugins()` searches infinitely deep, if necessary, to find plugins

### DIFF
--- a/docs/source/cli/index.rst
+++ b/docs/source/cli/index.rst
@@ -1202,11 +1202,7 @@ formats:
 
 .. code-block:: text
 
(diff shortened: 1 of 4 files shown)


brimoor:
Closing in favor of https://github.com/voxel51/fiftyone/pull/3732

PR #3734, opened 2023-10-29
fix pcd labels projection bug

sashankaryal:

Add `location`, `dimensions`, and `rotation` to the `Detection` label. While these fields are only relevant to the 3D label, not having them in the schema means they get omitted when `fo.SelectFields` view stage is applied (for instance in `_project_pagination_paths`, which runs as part of `paginateSamplesQuery` in the grid), which in turn can have unintended side effects like 3D labels not being projected in the grid.

Since the impetus for this PR is 3D labels not being projected in the grid, an alternative approach - if we're opposed to adding these fields in the schema - is to add a conditional logic in `_project_pagination_paths` that ~passes `meta_filters` to  `fo.SelectFields()` if the media type for the current view is `point-cloud` so that `location`, `dimensions` and `rotation` fields aren't excluded.~

Yet another approach is to use a different label class, nominally `Detection3D`, for 3D detections.

before: 
Screenshot 2023-10-29 at 6 31 19 PM

after:
Screenshot 2023-10-29 at 1 01 18 PM

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -945,16 +945,6 @@ class SampleCollection(object):
                 use_db_fields=use_db_fields,
             )
 
-            if self.media_type == fom.POINT_CLOUD and issubclass(
-                field.document_type, fol.Detection
-            ):
-                field_names = (
-                    *field_names,
-                    "location",
-                    "dimensions",
-                    "rotation",
-                )
-
             return tuple(path + "." + f for f in field_names)
 
         field_names = self._dataset._sample_doc_cls._get_default_fields(
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -945,6 +945,16 @@ class SampleCollection(object):
                 use_db_fields=use_db_fields,
             )
 
+            if self.media_type == fom.POINT_CLOUD and issubclass(
+                field.document_type, fol.Detection
+            ):
+                field_names = (
+                    *field_names,
+                    "location",
+                    "dimensions",
+                    "rotation",
+                )
+
             return tuple(path + "." + f for f in field_names)
 
         field_names = self._dataset._sample_doc_cls._get_default_fields(
--- a/app/packages/looker/src/worker/threed-label-processor.ts
+++ b/app/packages/looker/src/worker/threed-label-processor.ts
@@ -78,6 +78,10 @@ const getInferredParamsForUndefinedProjection = (sample: Readonly<Sample>) => {
 
     if (label._cls === DETECTIONS) {
       for (const detection of label.detections as DetectionLabel[]) {
+        if (!detection.location || !detection.dimensions) {
+          continue;
+        }
+
         const [x, y] = detection.location;
         const [lx, ly] = detection.dimensions;
 
@@ -87,6 +91,10 @@ const getInferredParamsForUndefinedProjection = (sample: Readonly<Sample>) => {
         maxY = Math.max(maxY, y + ly / 2);
       }
     } else if (label._cls === "Detection") {
+      if (!label.location || !label.dimensions) {
+        continue;
+      }
+
       const [x, y] = label.location as DetectionLabel["location"];
       const [lx, ly] = label.dimensions as DetectionLabel["dimensions"];
 
@@ -144,6 +152,10 @@ const PainterFactory3D = (
     } = orthographicProjectionParams;
     const [xmin, ymin] = min_bound;
     const [xmax, ymax] = max_bound;
+
+    if (!label.location || !label.dimensions) {
+      return;
+    }
 
     const [x, y, z] = label.location; // centroid of bounding box
     const [lx, ly, lz] = label.dimensions; // length of bounding box in each dimension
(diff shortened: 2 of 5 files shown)


brimoor (changes requested):
Hmm, could we find another solution to this?

The downside of this approach is that every dataset will have three extra attributes on every `Detection` field, even if they don't contain 3D data. This will add extra overhead to importing/exporting samples in general (not sure how significant), and perhaps more significant will cause the App sidebar to have to run more aggregations to render field filters (most of which will find all `None`s and thus get hidden).

From an SDK standpoint, it's definitely "wrong" that `dataset.select_fields("ground_truth.detections")` on a 3D dataset will omit attributes that are required to visualize 3D detections. However, this syntax is probably quite niche, the more common thing to do is to select/exclude *top-level* fields.

Is this bug arising now because of a change in App server implementation? Perhaps we could update the following method to detect that `Detection` fields on datasets that contain 3D data should treat a few extra fields as default?

https://github.com/voxel51/fiftyone/blob/60d3c2c0334777a7bc73e2e0641902ac2dbdb227/fiftyone/core/collections.py#L943

Having a dedicated `Detection3D` type for 3D labels would be the most natural. The problem is (was) that grouped datasets share the same schema, so it wouldn't be easy to have a dataset whose `ground_truth` field contains `Detections` for image slices and `Detections3D` for point cloud slices.

swheaton:
Don't know the context but plus one for separate 3d objects types, if we could make that work

sashankaryal:
@brimoor Good points! And thank you for the added context. I updated the `_get_default_sample_fields` method.

@swheaton separate 3d objects types would be nice, but Brian raises an important point:
> The problem is (was) that grouped datasets share the same schema, so it wouldn't be easy to have a dataset whose ground_truth field contains Detections for image slices and Detections3D for point cloud slices.

The following gif makes it clear:


Without same field for detection, I suppose we'd either have to offer an API to "link" the two detections, or change the app UI and offer multiple toggles for 2D and 3D.

fiftyone/core/collections.py
```diff
@@ -945,6 +945,16 @@ def _get_default_sample_fields(
                 use_db_fields=use_db_fields,
             )
 
+            if self.media_type == fom.POINT_CLOUD and issubclass(
```
brimoor, line 948: We'll need to generalize the `self.media_type == fom.POINT_CLOUD` check a bit to handle the different cases that can arise involving grouped datasets. Basically we need the same logic as [`_contains_videos()`](https://github.com/voxel51/fiftyone/blob/60d3c2c0334777a7bc73e2e0641902ac2dbdb227/fiftyone/core/collections.py#L9484) for `POINT_CLOUD` media type rather than `VIDEO`.

I'd recommend adding a new `_contains_media_type()` method like so:

```py
def _contains_media_type(self, media_type, any_slice=False):
   # generalize the existing _contains_videos() here

def _contains_videos(self, any_slice=False):
    return self._contains_media_type(fom.VIDEO, any_slice=any_slice)
```

For the purposes of this PR, the check we need is then `self._contains_media_type(fom.POINT_CLOUD, any_slice=True)`.

Also, for generality, the logic we're discussing here should be added to `_get_default_frame_fields()` as well, for when users can have playback of 3D scenes 😄

brimoor: Side note: I considered suggesting that we move this logic inside `EmbeddedDocumentField._get_default_fields()`. This would cause the rest of the SDK to know that these fields are required, which would prevent users from doing certain things like:

```py
# Delete the field
dataset.delete_sample_field("ground_truth.detections.location")

# Remove the field from the dataset's schema without clearing the values
dataset.remove_dynamic_sample_field("ground_truth.detections.location")
```

on a dataset that contains 3D data.

For context, since these 3D-specific fields are not formally declared on the dataset's schema by default, if users want them to be available in the App then they have to manually declare them like so:

```py
dataset.add_sample_field("ground_truth.detections.location", fo.ListField, subfield=fo.FloatField)
dataset.add_sample_field("ground_truth.detections.dimensions", fo.ListField, subfield=fo.FloatField)
dataset.add_sample_field("ground_truth.detections.rotation", fo.ListField, subfield=fo.FloatField)
```

For reversibility, it's probably best to let them be able to undeclare/delete these fields from their schema if they want to.

**tldr**: I think it is best to *NOT* move the logic from this PR inside `EmbeddedDocumentField._get_default_fields()`.

benjaminpkane:
I think the [initial changes](https://github.com/voxel51/fiftyone/pull/3734/commits/efb8e56d6c254241a27dc57b36774e2cf2176e3a) adding the fields to the label type are reasonable. If it affects import/export significantly, perhaps it is not viable. But in terms of the sidebar and unwanted aggregations, we already exclude `bounding_box` [here](https://github.com/voxel51/fiftyone-teams/blob/ec9207be0ec4459b8997b947affc1ed58b455d9d/app/packages/core/src/components/Sidebar/Entries/FilterablePathEntry.tsx#L66). The same can be done for these new fields

brimoor:
Good point about how the App excludes `bounding_box` 👍 this would be an analogous case where we have more list fields that aren't appropriate for in App filtering.

Maybe I'm being overly cautious but I was also hoping to avoid `print(fo.Detection())` showing three more attributes that aren't applicable in many cases...

The ideal would really be a `Detection3D` class, just don't know how to achieve it...

brimoor:
@sashankaryal @benjaminpkane can we go with https://github.com/voxel51/fiftyone/pull/3761 instead to fix this?

sashankaryal:
good with me

> @sashankaryal @benjaminpkane can we go with #3761 instead to fix this?

Works for me.

PR #3739, opened 2023-10-30
Fix active field on view change

benjaminpkane:
Fixes active (checked) fields in the sidebar on view change and persists the value to session storage

https://github.com/voxel51/fiftyone/assets/19821840/d4c46813-b19e-4038-b93f-62fc9e5f6a9b


sashankaryal (approved):
lgtm

PR #3741, opened 2023-10-30
Respect overwrite param in _merge_lists

kaixi-wang:
currently, passing overwrite in `_merge_lists` does nothing

### DIFF
--- a/fiftyone/core/document.py
+++ b/fiftyone/core/document.py
@@ -847,9 +847,13 @@ class DocumentView(_Document):
 
 
 def _merge_lists(dst, src, overwrite=True):
-    if src is None:
-        return
-
+    if overwrite:
+        # in cases where __eq__ has been defined such that two
+        # instances without the exact same property values returns as
+        # equal, we want to make sure we persist src properties over dst
+        dst, src = src, dst
+    if not dst:
+        dst = []
     dst.extend(v for v in src if v not in dst)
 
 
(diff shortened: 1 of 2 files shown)


fiftyone/core/document.py
```diff
@@ -847,9 +847,13 @@ def _reload_parents(self):
...
-    if src is None:
-        return
-
+    if overwrite:
+        # in cases where __eq__ has been defined such that two
+        # instances without the exact same property values returns as
+        # equal, we want to make sure we persist src properties over dst
+        dst, src = src, dst
+    if not dst:
+        dst = []
```
brimoor, line 856: You can't create a new array here because this function processes `dst` in place. It doesn't return a new reference

fiftyone/core/document.py
```diff
@@ -847,9 +847,13 @@ def _reload_parents(self):
...
 
 def _merge_lists(dst, src, overwrite=True):
-    if src is None:
-        return
-
+    if overwrite:
+        # in cases where __eq__ has been defined such that two
+        # instances without the exact same property values returns as
+        # equal, we want to make sure we persist src properties over dst
+        dst, src = src, dst
```
brimoor, line 854: This is subtle, but for core methods subtlety is important:

This implementation can cause the order of the elements in the destination list to change when `overwrite=True`, as the source list is used as the starting point.

Conversely, in the implementation of `_merge_labels()` below, I was careful to do it in such a way that any overwrites are done in-place, and any new elements are strictly appended to the end. So that the order of elements in the destination list will not change.

For example in `Keypoint` labels there is a convention that the `confidence` field contains a list of values that correspond 1-1 with the shapes defined in the `points` field.

brimoor (changes requested):

Issue #3740, opened 2023-10-30
[BUG] ctx.selected_labels is still empty in resolve_input in v0.22.1 and v0.22.2


msamadony:
### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04): Ubuntu 20.04
-   **Python version** (`python --version`): Ubuntu 20.04
-   **FiftyOne version** (`fiftyone --version`): 0.22.1 and 0.22.2
-   **FiftyOne installed from** (pip or source): pip

### Describe the problem

With reference to a previously reported issue #3557, the ctx.selected_labels still returns an empty list in v0.22.1 and v0.22.2, even though the issue is marked as resolved in the v0.22.1 release notes.

The App shows the number of selected labels correctly:
![image](https://github.com/voxel51/fiftyone/assets/224718/865f0d3b-a832-42d5-ad97-8f6b58dc1e46)

However, the operator `resolve-type` request has an empty `selectedLabels` field:
![image](https://github.com/voxel51/fiftyone/assets/224718/42f70e61-b678-4eb6-a57e-a5b40c884bae)

I checked the code again, and it seems that another commit has removed the `selectedLabels` from the `globalContextSelector`. I suspect that this might be the root cause (I haven't tested it though).

https://github.com/voxel51/fiftyone/blob/4ad742a75d3f931198e0ec49576f22f7e804a7a6/app/packages/operators/src/state.ts#L78-L89

Commit Reference:
https://github.com/voxel51/fiftyone/commit/7d6de86eadba791f8483ffa05c3663fe4c4af2d9#diff-c4910bb1f5e42c2989667f7c357d1a520608050168897d95429e1ac0bb9eda4f

### Code to reproduce issue

Same as #3557

### Other info/logs

Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached. Please do not use screenshots for sharing text. Code snippets should be used instead when providing tracebacks, logs, etc.

### What areas of FiftyOne does this bug affect?

-   [x] `App`: FiftyOne application issue
-   [ ] `Core`: Core Python library issue
-   [ ] `Server`: FiftyOne server issue

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the FiftyOne codebase?

-   [ ] Yes. I can contribute a fix for this bug independently
[...]

PR #3742, opened 2023-10-30
fix placements flash and remove duplicate resolve

imanjra:

- Fix disappearance and re-appearance of placements when placements are re-resolved on context changes
- Fix multiple duplicate calls to resolve placements on the initial app load

Interactively within the app using an example operator and with network profiling

        notes for FiftyOne users.

```
- Fixed disappearance and re-appearance of placements when placements are re-resolved on context changes `#3742 <https://github.com/voxel51/fiftyone/pull/3742 >`_
- Fixed multiple duplicate calls to resolve placements on the initial app load `#3742 <https://github.com/voxel51/fiftyone/pull/3742 >`_
```

-   [ ] Other

### DIFF
--- a/app/packages/operators/src/hooks.ts
+++ b/app/packages/operators/src/hooks.ts
@@ -1,0 +1,38 @@
+import { useEffect, useRef, useState } from "react";
+import { useRecoilValue, useSetRecoilState } from "recoil";
+import { isEqual } from "lodash";
+import {
+  ExecutionContext,
+  fetchRemotePlacements,
+  resolveLocalPlacements,
+} from "./operators";
+import { operatorPlacementsAtom, operatorThrottledContext } from "./state";
+
+export function useOperatorPlacementsResolver() {
+  const context = useRecoilValue(operatorThrottledContext);
+  const setOperatorPlacementsAtom = useSetRecoilState(operatorPlacementsAtom);
+  const [resolving, setResolving] = useState(false);
+  const lastContext = useRef(null);
+
+  useEffect(() => {
+    async function updateOperatorPlacementsAtom() {
+      setResolving(true);
+      try {
+        const ctx = new ExecutionContext({}, context);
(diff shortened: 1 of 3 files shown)


app/packages/operators/src/state.ts
```diff
@@ -805,7 +802,14 @@ export function useOperatorPlacements(place: Place) {
...
       selectedSamples,
     });
-  }, [datasetName, view, extendedStages, filters, selectedSamples]);
+  }, [
+    setThrottledContext,
+    datasetName,
+    view,
+    extendedStages,
+    filters,
+    selectedSamples,
```
benjaminpkane, line 811: Can `selectedLabels` be added here? It seems like the fix for https://github.com/voxel51/fiftyone/issues/3740

imanjra: Yah. However, it would not fix #3740 by adding here as this context is only used for resolving placement. 

I will look into #3740

app/packages/operators/src/hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+        const remotePlacements = await fetchRemotePlacements(ctx);
+        const localPlacements = await resolveLocalPlacements(ctx);
+        const placements = [...remotePlacements, ...localPlacements];
+        setOperatorPlacementsAtom(placements);
+      } catch (error) {
+        console.error(error);
+      }
+      setResolving(false);
+    }
+    if (!isEqual(lastContext.current, context) && context?.datasetName) {
```
benjaminpkane, line 31: I think this equality check may not pass with `_uuid` values in view stages. There is a `viewsAreEqual` function I use sometimes for view equality checks.

I don't think anything needs to change here. Just flagging a sutble detail about serialized views

benjaminpkane (approved):
Code LGTM! Did a little testing and it's definitely reduced the calls to /resolve-placements 🚀

PR #3743, opened 2023-10-31
Smarter deletion of temp datasets

kaixi-wang:
Changes:
  - add a flag for not deleting datasets referenced in saved views 
  - don't add deleted datasets to singletons list

Todo:
  - add tests

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -36,7 +36,6 @@ import fiftyone.core.groups as fog
 import fiftyone.core.labels as fol
 import fiftyone.core.media as fom
 import fiftyone.core.metadata as fome
-from fiftyone.core.odm.dataset import SampleFieldDocument
 from fiftyone.core.odm.dataset import DatasetAppConfig
 import fiftyone.migrations as fomi
 import fiftyone.core.odm as foo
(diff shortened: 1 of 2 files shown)


brimoor (changes requested):
Is there a specific problem you're trying to solve? There should be another way to solve it.

Views like `ToPatches()` that use temporary datasets are designed to automatically regenerate their backing datasets if they're ever deleted:
https://github.com/voxel51/fiftyone/blob/440704307d8a480eefa99d102939fa576e5c99e5/fiftyone/core/stages.py#L7453

In fact it is *preferable* that these temporary datasets do get deleted; if the old collection still exists, then loading a saved patches view would just use it, even though the root dataset may have changed significantly since the view was saved. The patches view would then be out of sync, which is counter to the goal of dataset views: they're supposed to be a virtual concept that's always in sync (yes, creating temporary collections is an expensive operation; but I didn't have a better idea at the time to implement these views).

The way to force a patches view to "resync" with the underlying dataset is to call `view.reload()`, which, case and point, deletes the temporary dataset, because this is a safe thing to do:
https://github.com/voxel51/fiftyone/blob/440704307d8a480eefa99d102939fa576e5c99e5/fiftyone/core/patches.py#L335

brimoor:
Tracking whether a collection is in-use or not is a relevant problem that connects with work that @swheaton is doing on versioning. But the relevant definition of "in use" is whether a user is literally using the dataset at that moment, not whether a saved view that's not currently being used stores a reference to a dataset.

kaixi-wang:
> Is there a specific problem you're trying to solve? There should be another way to solve it.
> 
> Views like `ToPatches()` that use temporary datasets are designed to automatically regenerate their backing datasets if they're ever deleted:
> 
> https://github.com/voxel51/fiftyone/blob/440704307d8a480eefa99d102939fa576e5c99e5/fiftyone/core/stages.py#L7453
> 
> In fact it is _preferable_ that these temporary datasets do get deleted; if the old collection still exists, then loading a saved patches view would just use it, even though the root dataset may have changed significantly since the view was saved. The patches view would then be out of sync, which is counter to the goal of dataset views: they're supposed to be a virtual concept that's always in sync (yes, creating temporary collections is an expensive operation; but I didn't have a better idea at the time to implement these views).
> 
> The way to force a patches view to "resync" with the underlying dataset is to call `view.reload()`, which, case and point, deletes the temporary dataset, because this is a safe thing to do:
> 
> https://github.com/voxel51/fiftyone/blob/440704307d8a480eefa99d102939fa576e5c99e5/fiftyone/core/patches.py#L335

regenerating the view is expensive and can take time, so when loading saved views in the app, if the temp dataset doesn't already exist, the app will throw an error.

brimoor:
Regenerating the temp dataset only when the root dataset is updated would have ideal properties from a reuse standpoint, but we don't have a way to track `Dataset.last_updated_at` just yet. It does connect with work that @swheaton is doing on `Sample.last_updated_at` though. Any efforts to reuse, retain, or otherwise change the behavior of temporary dataset cleanup should really be done in conjunction with dataset versioning, as they both use "hidden datasets" and have similar cleanup needs.

If loading a saved view in the App builds the view multiple times, I can see how that could lead to temporary patch datasets getting generated multiple times concurrently, which is wasteful from a computation standpoint. Sounds like we should try to remove duplicate view load calls, for efficiency but also just for cleanliness.

Can you provide a reproducible example of seeing a duplicate index error? Can't guess how/why that would be happening.

manivoxel51:
> If loading a saved view in the App builds the view multiple times,

@kaixi-wang can you share a video on how to confirm we are loading a view **multiple times** when saved view loads? I just checked the network tab and I only see a set_view call once. maybe you mean in the SDK side?

PR #3744, opened 2023-10-31
Update App events

benjaminpkane:
Update for GA4. Removes Python imports as an event as it is not actively measured


app/packages/app/package.json
```diff
@@ -41,6 +41,7 @@
         "react-dom": "18.1.0",
         "react-error-boundary": "^3.1.4",
         "react-ga": "^3.1.2",
```
manivoxel51, line 43: ✂️

manivoxel51 (approved):
nice! 🍨 

did not test

manivoxel51 (approved):

PR #3748, opened 2023-10-31
Expand only labels and tags sidebar groups by default

benjaminpkane:
Screenshot 2023-10-31 at 2 27 04 PM

### DIFF
--- a/app/packages/state/src/recoil/sidebar.ts
+++ b/app/packages/state/src/recoil/sidebar.ts
@@ -761,13 +761,7 @@ export const groupShown = selectorFamily<
       const data = get(sidebarGroupMapping({ modal, loading }))[group];
 
       if ([null, undefined].includes(data.expanded)) {
-        if (["tags"].includes(group)) {
-          return null;
-        }
-        return (
-          !data.paths.length ||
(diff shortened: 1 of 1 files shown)


lanzhenw (dismissed):
LGTM 🚀

manivoxel51:
> Screenshot 2023-10-31 at 2 27 04 PM

Looks good 🍨 

There are some connections with other features that we may have to consider.

For example, if a field visibility is applied, do we keep all the fields collapsed or should we expand them? How would you know which collapsed field has field visibility applied to? one idea is to add an icon to every field with a count of fields hidden in that top field. but that seems like a overkill
There might be consequences to the work @lanzhenw has done with field filter/visibility feature too? thinking out loud

manivoxel51:
Perhaps, this could be a setting in our settings dropdown too

benjaminpkane:
> For example, if a field visibility is applied, do we keep all the fields collapsed or should we expand them? How would you know which collapsed field has field visibility applied to? one idea is to add an icon to every field with a count of fields hidden in that top field. but that seems like a overkill

Good points. I think I will hold off on merging this until we can discuss more

PR #3749, opened 2023-10-31
Fix grid selection and e2e

benjaminpkane:
Fixes a grid selection bug introduced by me in #3718. Also adds back `e2e` as a required check in pull requests to help prevent these regressions


manivoxel51 (approved):
🍨 thank you for fixing this

PR #3752, opened 2023-11-01
[DOCS] Proposal to update github bug report template

swheaton:
I'm proposing some updates to make the bug report template easier and clearer for both writer and reader.
I will comment on the specific reasons for each change inline

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report_template.md
+++ b/.github/ISSUE_TEMPLATE/bug_report_template.md
@@ -30,22 +30,15 @@ Describe the problem clearly here. Include descriptions of the expected behavior
 
 ### Code to reproduce issue
 
-Provide a reproducible test case that is the bare minimum necessary to generate the problem.
+Provide a reproducible test case that is the bare minimum necessary to generate the problem. Include a short video or screenshot if the bug is in the app.
 
 ```
-# commands here
+# commands and/or screenshots here
 ```
 
 ### Other info/logs
 
 Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached. Please do not use screenshots for sharing text. Code snippets should be used instead when providing tracebacks, logs, etc.
-
-### What areas of FiftyOne does this bug affect?
-
--   [ ] `App`: FiftyOne application issue
--   [ ] `Core`: Core Python library issue
--   [ ] `Server`: FiftyOne server issue
--   [ ] `Zoo`: FiftyOne Dataset or Model Zoo issue
 
 ### Willingness to contribute
 
--- a/.github/ISSUE_TEMPLATE/bug_report_template.md
+++ b/.github/ISSUE_TEMPLATE/bug_report_template.md
@@ -12,9 +12,14 @@ assignees: ''
 Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/voxel51/fiftyone/blob/develop/ISSUE_POLICY.md) for information on what types of issues we address.
 
 1. Please fill in this template to ensure a timely and thorough response
-2. Place an "x" between the brackets next to an option if it applies. For example:
+2. Remove the section instructions but leave the section header
+3. Place an "x" between the brackets next to an option if it applies. For example:
     -   [x] Selected option
-3. **Please delete everything above this line before submitting the issue**
+4. **Please delete everything above this line before submitting the issue**
+
+### Describe the problem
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report_template.md
```diff
@@ -12,9 +12,14 @@ assignees: ''
 Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/voxel51/fiftyone/blob/develop/ISSUE_POLICY.md) for information on what types of issues we address.
 
 1. Please fill in this template to ensure a timely and thorough response
-2. Place an "x" between the brackets next to an option if it applies. For example:
+2. Remove the section instructions but leave the section header
```
swheaton, line 15: Try to make sure the report is cleaner ... we want reporter to remove the descriptions

.github/ISSUE_TEMPLATE/bug_report_template.md
```diff
@@ -12,9 +12,14 @@ assignees: ''
...
+3. Place an "x" between the brackets next to an option if it applies. For example:
     -   [x] Selected option
-3. **Please delete everything above this line before submitting the issue**
+4. **Please delete everything above this line before submitting the issue**
+
+### Describe the problem
```
swheaton, line 20: Move description to the top so it's the first thing you see on the page or github email.

.github/ISSUE_TEMPLATE/bug_report_template.md
```diff
@@ -12,43 +12,34 @@ assignees: ''
...
 
-### Describe the problem
-
-Describe the problem clearly here. Include descriptions of the expected behavior and the actual behavior.
-
-### Code to reproduce issue
+### System information
 
-Provide a reproducible test case that is the bare minimum necessary to generate the problem.
+-   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04):
```
swheaton, line 34: Move system info to after description and reproducing steps

.github/ISSUE_TEMPLATE/bug_report_template.md
```diff
@@ -12,43 +12,34 @@ assignees: ''
...
+-   **Python version** (`python --version`):
+-   **FiftyOne version** (`fiftyone --version`):
+-   **FiftyOne installed from** (pip or source):
 
 ### Other info/logs
 
 Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached. Please do not use screenshots for sharing text. Code snippets should be used instead when providing tracebacks, logs, etc.
 
-### What areas of FiftyOne does this bug affect?
```
swheaton, line 46: I don't think this section provides any valuable information to us. It will likely be obvious to maintainers by the description and reproduction steps. And the user might not know distinction between app and server or core and server, for example.
If we wanted to bucket issues into categories then we should use labels.

.github/ISSUE_TEMPLATE/bug_report_template.md
```diff
@@ -12,43 +12,34 @@ assignees: ''
...
-### System information
+### Describe the problem
 
--   **OS Platform and Distribution** (e.g., Linux Ubuntu 16.04):
--   **Python version** (`python --version`):
--   **FiftyOne version** (`fiftyone --version`):
--   **FiftyOne installed from** (pip or source):
+Describe the problem clearly here. Include descriptions of the expected behavior and the actual behavior.
 
-### Commands to reproduce
```
swheaton, line 26: It's not clear to me that the distinction between "Commands to reproduce" and "Code to reproduce issue" is all that meaningful. So let's just make it simpler by having description above, space for reproducing steps (code, screenshots/videos, instructions, etc.), and additional info section down below if needed.

.github/ISSUE_TEMPLATE/bug_report_template.md
swheaton: Now, you have to scroll a ways down sometimes (especially on mobile) to get to the crux of the issue, you've got the system info upfront which is relevant but relatively unimportant.

swheaton:
@brimoor bug report template improvements for your consideration, at lowest priority of course.

benjaminpkane (approved):
Less is more! I like it. Only things I added was a comment about reproduction code. And 16.04 -> 22.04.

If you think there is more to remove/improve I'm all for it!

swheaton:
thanks!

PR #3753, opened 2023-11-01
fix sidebar re-arranging bug

imanjra:

fixed sidebar group re-arranging issue where sample tags jumps to the moving group when the moving group is taken above tags group

By attempting to drag any of the group in sidebar above the sticky tags group

        notes for FiftyOne users.

```
- Fixed sample tags jumping to sidebar group being re-arranged issue `#3753 <https://github.com/voxel51/fiftyone/pull/3753>`_
```

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Sidebar/Sidebar.tsx
+++ b/app/packages/core/src/components/Sidebar/Sidebar.tsx
@@ -535,6 +535,10 @@ const InteractiveSidebar = ({
       entry.kind !== fos.EntryKind.INPUT
     );
 
+    if (after === undefined) {
+      return lastOrder.current;
+    }
+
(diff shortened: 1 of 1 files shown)


manivoxel51 (approved):
nice work! 🍨 

did a quick test before and after your fix and looks good

before

https://github.com/voxel51/fiftyone/assets/109545780/33af3eb7-c6aa-4dc7-b931-010d3e5f389e

after - which seems to be locking the top tag group and stop the jump of sample tags field

Uploading Screen Recording 2023-11-01 at 7.13.39 PM.mov…

will defer to @benjaminpkane for final green light in case there are other side effects

PR #3757, opened 2023-11-01
Operator list is behind the modal

lanzhenw:

fix the zIndex conflict between modals. 

Operator List modal uses a z-index of 1300. I set the sample modal and color modal between 1300 to fix the problem. 

Screenshot 2023-11-01 at 3 52 58 PM

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/ColorModal/ShareStyledDiv.ts
+++ b/app/packages/core/src/components/ColorModal/ShareStyledDiv.ts
@@ -6,7 +6,7 @@ export const ModalWrapper = styled.div`
   left: 0;
   width: 100%;
   height: 100%;
-  z-index: 100000000 !important;
+  z-index: 1000; // do not set more than 1300 (operator panel)
   align-items: center;
   display: flex;
   justify-content: center;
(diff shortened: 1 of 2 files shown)


app/packages/core/src/components/ColorModal/ShareStyledDiv.ts
```diff
@@ -6,7 +6,7 @@ export const ModalWrapper = styled.div`
   left: 0;
   width: 100%;
   height: 100%;
-  z-index: 100000000 !important;
+  z-index: 1000; // do not set more than 1300 (operator panel)
```
sashankaryal, line 9: We should consider having one authoritative file that contains all z-index values. soon ™️

sashankaryal (approved):
lgtm

app/packages/core/src/components/ColorModal/ShareStyledDiv.ts
lanzhenw: Let's talk about this offline.

PR #3758, opened 2023-11-01
Reset stage states in cloned saved views

kaixi-wang:

There's a problem with saved_views of cloned datasets, specifically ones with stages involving generated/temp datasets.  The name of the generated view in the saved view of the cloned dataset is the same as the original dataset. However, after a dataset is cloned, it can be modified independently. This means that the generated temp dataset also must be different. Otherwise changes on the generated view will be persisted to both the clone and the original, which may lead to errors and undesirable effects. 

Example:
```
# Load dataset with saved views that have stages with _state:
dataset=fo.load_dataset('quickstart-sim-sort-tests')
cloned = dataset.clone()
orig_views= dataset._doc.saved_views
cloned_views = cloned._doc.saved_views
for (o,c) in zip(orig_views, cloned_views):
    print("original:{}\ncloned:{}\n".format(o.view_stages, c.view_stages))
```

```original:['{"_cls": "fiftyone.core.stages.ToPatches", "kwargs": [["field", "predictions"], ["config", null], ["_state", {"dataset": "quickstart", "stages": [], "field": "predictions", "config": null, "name": "2023.09.29.14.22.14"}]]}']
cloned:['{"_cls": "fiftyone.core.stages.ToPatches", "kwargs": [["field", "predictions"], ["config", null], ["_state", {"dataset": "quickstart", "stages": [], "field": "predictions", "config": null, "name": "2023.09.29.14.22.14"}]]}']
```


brimoor (changes requested):
Consider this example:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.save_view("test", patches)

dataset2 = dataset.clone()
patches2 = dataset2.load_saved_view("test")

assert patches._serialize() != patches2._serialize()
```

When `patches2` is loaded, the cached `_state` from `dataset` that has been carried over to `dataset2` is invalidated because `dataset2.name != dataset.name`:
https://github.com/voxel51/fiftyone/blob/ccf77ffd8aaa7c32bac180234465c508343bea43/fiftyone/core/stages.py#L7441

So, although the cached `_state` is not valid, there's not any direct risk of accidentally editing the wrong source dataset as you are suggesting.

brimoor:
One thing I *wish* we did was use `dataset.id` rather than `dataset.name` in the cached state: changing a dataset's name should not invalidate the cache.

We could make that change, but it would cause any saved views with `_state` to be forever cache-invalid because their saved `_state` will contain names, not IDs. However, that's not really a big deal because if temporary datasets are cleaned up reasonably often, the saved view will also be cache-invalid because it's temporary dataset will no longer exist.

brimoor:
I don't see value in going further down this rabbit hole for now. Unless there is a specific reproducible error report.

We've already decided that once `Dataset.last_updated_at` is updated we'll make improvements here.

kaixi-wang:
> When `patches2` is loaded, the cached `_state` from `dataset` that has been carried over to `dataset2` is invalidated because `dataset2.name != _state["dataset"]`:

I didn't see this check. Where is it?

kaixi-wang:
Also, I don't see any harm this change would do. As a data-centric company, we should value persisting accurate/valid data. Cloning shouldn't introduce inconsistencies in the first place that may or may not be resolved depending on if a user calls a certain sdk method, especially since there are app only users

PR #3759, opened 2023-11-01
Plugin secrets available in resolve_input

kaixi-wang:

Secrets should be available via `ctx.secrets` when `resolve_input` is called.


brimoor (approved):
LGTM

PR #3761, opened 2023-11-02
Ensure that 3D detection fields are treated as default by SelectFields

brimoor:
Alternative solution to https://github.com/voxel51/fiftyone/pull/3734 that should fix the App issue targeted there while also having the benefit of ensuring that 3D fields are not excluded by things like `select_fields("ground_truth.detections.label")`, which should never exclude fields that are required to visualize 3D data.


sashankaryal (approved):
lgtm! 🥇

benjaminpkane (approved):
LGTM

PR #3767, opened 2023-11-03
Sample and Frame created_at field

swheaton:

Add `created_at` field to all samples and frames. It's set to now when added to the dataset, and reset back to `None` if removed from the dataset.
We could set `created_at` if the sample gets created outside of the dataset first, but it's slightly easier to do it this way and there's not likely to be a meaningful distinction between when a sample was created and when added to dataset.

Locations to add `created_at` were identified using as a template [adding dataset id to sample and frames docs](https://github.com/voxel51/fiftyone/pull/2711) 

### Notes
- Samples get a creation time only when they're added to a dataset.
- Followup PR will be sample/dataset `last_updated_at`
- Merge for existing samples was a little tricky but if we first `addField` but then add it to `delete_fields` when we create the `whenMatched` pipeline, we'll make sure that it will be added on insert but ignored on merge.
- Not sure what happens with generated datasets, or if it even matters. But I'll check on that.
- Writes to this field through `set_field`, `__setattr__`, `set_values()`, `clear_sample_field`, etc., are prohibited.
   -  I attempted to achieve this by creating a `readonly` flag for the `Field`, akin to the `is_virtual` property in [virtual fields PR](https://github.com/voxel51/fiftyone/pull/2429).

- Added tests for each case in both image and video, and ran manually. Created_at is set on creation but not again. Distinct cases below:
   - `add_samples()`
   - `clone()`
   - `add_collection(new_ids=True)`
   - `merge_samples()  # both existing and new samples`
   - `fo.Dataset.from_dir()`
- Also then tested that any manner of modifying the field throws an error.
- If you want to test manually, following along with the test cases is best rather than pasting full here.

        notes for FiftyOne users.

Add `created_at` default `DateTimeField` to samples and frames, which contains the time the sample was added to the dataset.

-   [ ] Other

### DIFF
--- a/fiftyone/core/video.py
+++ b/fiftyone/core/video.py
@@ -5,6 +5,7 @@ Video frame views.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+import datetime
 from collections import defaultdict
 from copy import deepcopy
 import logging
@@ -691,12 +692,17 @@ def make_frames_dataset(
 
     pipeline.extend(
         [
-            {"$addFields": {"_dataset_id": dataset._doc.id}},
+            {
+                "$addFields": {
+                    "_dataset_id": dataset._doc.id,
+                    "created_at": datetime.datetime.utcnow(),
+                }
+            },
             {
                 "$merge": {
                     "into": dataset._sample_collection_name,
                     "on": ["_sample_id", "frame_number"],
-                    "whenMatched": "merge",
+                    "whenMatched": "merge",  # TODO STW what do when existing is merge with created_at
                     "whenNotMatched": "discard",
                 }
             },
@@ -870,6 +876,7 @@ def _init_frames(
             missing_docs = None
             missing_fps = None
 
+        now = datetime.datetime.utcnow()
         # Create necessary frame documents
         for fn in doc_frame_numbers:
             if is_clips:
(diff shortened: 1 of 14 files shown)


swheaton:
Figured out the snags I was having with mongoengine fields and so I added in readonly protection for created_at. lmk if you like the pattern or not - it's not foolproof because we use mongoengine for some things and direct mongo for others, but I think I've gotten all cases covered?

swheaton:
Wondering if we should have frame created_at in addition to sample? Seems potential overkill since it adds 2 extra datetimes for every frame. Discussing offline

fiftyone/core/video.py
```diff
@@ -691,12 +692,17 @@ def make_frames_dataset(
...
+                    "_dataset_id": dataset._doc.id,
+                    "created_at": datetime.datetime.utcnow(),
+                }
+            },
             {
                 "$merge": {
                     "into": dataset._sample_collection_name,
                     "on": ["_sample_id", "frame_number"],
-                    "whenMatched": "merge",
+                    "whenMatched": "merge",  # TODO STW what do when existing is merge with created_at
```
swheaton, line 705: Remaining TODO

fiftyone/core/video.py
swheaton, line 1: Actually I think we can probably ignore generated datasets. This would include frames dataset like here but also patches, clips, and trajectories.
It's not super important that the times be right, unless we're trying to use them to decide what to sync back or something.

swheaton:
Drafting because we're going to do last_updated_at along with created_at

swheaton:
Superseded by https://github.com/voxel51/fiftyone/pull/3805

PR #3772, opened 2023-11-03
Move Snackbar to @fiftyone/core

benjaminpkane:
Moves the new Snackbar component from #3613 to `@fiftyone/core` for reusability and adds the component to the `IndexPage`

https://github.com/voxel51/fiftyone/assets/19821840/ae6b8605-155f-4f49-937c-5bb6a33405ab


manivoxel51 (approved):
just pulled and tested! nice 🍨

Issue #3778, opened 2023-11-04
[BUG] Wrong image dimensions when image has any rotation in the exif orientation entry


Fraccalo:
### Describe the problem

See discussion on Slack: https://fiftyone-users.slack.com/archives/C0189A476EQ/p1698408138572119

When importing an image that has an orientation entry in its exif, fiftyone imports the wrong image dimension, and consequently the annotations are displayed in the wrong position.  

### Code to reproduce issue

```
import fiftyone as fo
from fiftyone.utils.coco import add_coco_labels

ds = fo.Dataset()
ds.add_images_dir("dataset")

add_coco_labels(sample_collection = ds,
                label_field = "test",
                labels_or_path = "anno.json",
                classes = ["0","test"],
                label_type = "detections")

fo.launch_app(ds)
```

Here a sample image can be used for the test:
![img_rot](https://github.com/voxel51/fiftyone/assets/33780601/9c0fe485-2cf8-4b29-bd6b-2ad30cbf84fa)
While the annotation file is simply:
```
{"images": [{"id": 1, "width": 1200, "height": 841, "file_name": "img_rot.jpg"}], 
 "annotations": [{"id": 1, "iscrowd": 0, "image_id": 1, "category_id": 1, "extra": {}, "bbox": [641, 100, 200, 100], "area": 20000}], 
 "categories": [{"id": 1, "name": "test"}]}
```

This is what happens when loading the annotation into fiftyone:
<img width="959" alt="image" src="https://github.com/voxel51/fiftyone/assets/33780601/73f04b65-67cb-40a6-999b-20df8e9fc60f">
you can see that the image dimension is w:841, h:1200, while it should be the other way around (the width of the image is bigger than its height).
Also, the bounding box right end coincides with the image right edge, but the annotation is x:641, w:200, so it should arrive at 841, which shouldn't be the edge of the image (as the real width is 1200).

This is due to PIL library that messes up with the exif orientation entry. In fact, if I try to load the image with PIL it is actually loaded in a rotated fashion:
<img width="622" alt="image" src="https://github.com/voxel51/fiftyone/assets/33780601/f1c05e24-4786-44e7-aae3-0283845f730f">

This can be fixed by using ImageOps from the PIL library to rotate images according to the exif:
```ImageOps.exif_transpose(Image.open("dataset/img_rot.jpg"))```

[...]

PR #3779, opened 2023-11-04
Add PIL image rotation according to exif info

Fraccalo:

Introduce exif orientation support when importing images into a dataset - see https://github.com/voxel51/fiftyone/issues/3778

Tested on the code posted in the Issue linked above.

Before the fix image dimensions are flipped and annotation is displayed in the wrong position:
image

After the fix dimensions are correct and annotation is displayed in the correct position:
image

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/operators/builtin.py
+++ b/fiftyone/operators/builtin.py
@@ -187,6 +187,40 @@ class RenameSampleField(foo.Operator):
         ctx.trigger("reload_dataset")
 
 
+class ClearSampleField(foo.Operator):
+    @property
+    def config(self):
+        return foo.OperatorConfig(
+            name="clear_sample_field",
+            label="Clear sample field",
+            dynamic=True,
+        )
+
+    def resolve_input(self, ctx):
+        inputs = types.Object()
+        fields = ctx.dataset.get_field_schema(flat=True)
+        field_keys = list(fields.keys())
+        field_selector = types.AutocompleteView()
+        for key in field_keys:
+            field_selector.add_choice(key, label=key)
+
+        inputs.enum(
+            "field_name",
+            field_keys,
+            label="Field to clear",
+            view=field_selector,
+            required=True,
+        )
+
+        return types.Property(
+            inputs, view=types.View(label="Clear sample field")
+        )
+
+    def execute(self, ctx):
+        ctx.dataset.clear_sample_field(ctx.params.get("field_name", None))
+        ctx.trigger("reload_dataset")
+
+
 class DeleteSelectedSamples(foo.Operator):
     @property
     def config(self):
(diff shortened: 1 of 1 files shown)


swheaton:
@Fraccalo 🙌🏼 awesome! Was just thinking about this issue the other day. Thanks for contributing a fix and giving the detailed bug description - and a first time contributor too! 🥇  

I was able to see what you're talking about with the rotated width/height. I'll try the full annotation import also.

swheaton:
Before merging, let's do the following.
1. Add test for that function. I can help with this since it doesn't seem like we have an existing test.
2. Let this marinate and we'll see if anyone has any further insights. This is my first experience with EXIF.

fiftyone/core/metadata.py
```diff
@@ -283,7 +283,7 @@ def get_image_info(f):
     Returns:
         ``(width, height, num_channels)``
     """
-    img = Image.open(f)
+    img = ImageOps.exif_transpose(Image.open(f))
```
swheaton, line 286: Since this is an extremely core function call, I would just want to be careful to not add any latency if not necessary.
```suggestion
    img = Image.open(f)
    if 'exif' in img.info:
        img = ImageOps.exif_transpose(img)
```

swheaton: Guess I never hit submit on that comment

Fraccalo: That's a good call, thanks for the change. It's an extremely busy period for me so I haven't been able to look at adding a test function, not sure when I'll be able to look into that but hopefully within a couple of weeks I can give a look at that :)

swheaton: No worries I'm working on it now so we can get this out with the next release :)

brimoor (changes requested):
Can we find a way to handle exif without actually loading the image?

[exif_transpose()](https://github.com/python-pillow/Pillow/blob/da59ad000d1405eaecd557175e29083a87d19f7c/src/PIL/ImageOps.py#L608) calls `Image.load()`, which means the entire image must be read just to deduce the correct width/height.

I don't think we can afford that much computation here. This method is called for every image in the App grid and also called by things like `compute_metadata()` which we've taken great pains to optimize as much as possible.

If it helps, I see no point in supporting orientations other than 0,90,180,270.

swheaton:
ya working on it @brimoor, re: thread a few weeks ago.
You really only need to swap width and height when orientation is 90 or 270 which maps to Exif Orientation 5,6,7,8

swheaton:
did not mean to close it, was trying to push to fraccalo/develop. i'll figure it out

swheaton:
I apologize @Fraccalo I'm unsure how to undo what I did easily. I opened a new PR https://github.com/voxel51/fiftyone/pull/3824
Your commit is still in the history so I hope it gives credit still.

PR #3781, opened 2023-11-06
Update Labelbox cleanup to support batch mode

ehofesmann:
The Labelbox integration was updated some time ago to migrate from their dataset to queue mode for creating projects. This PR updates the cleanup method and deletion of Labelbox projects to not only delete the project itself, but also the associated data rows and ontologies created by FiftyOne. 

The API method to delete the project makes the deletion of data row batches and ontologies configurable kwargs, but the cleanup method will automatically delete batches and ontologies as the purpose of it is to automatically clean up fiftyone-generated artifacts.


brimoor (approved):
LGTM

PR #3783, opened 2023-11-06
Resolve secret if ctx.secret[key] accessed but not resolved

kaixi-wang:

Rather than only resolve secrets when the operator is prepared, attempt to resolve a secret when accessed via`ctx.secret(key)` if not already resolved, making it's value accessible on demand at any stage of the operation

(Details)

        notes for FiftyOne users.

-   [ ] Other


fiftyone/operators/executor.py
```diff
@@ -430,20 +430,25 @@ def results(self):
         return self.request_params.get("results", {})
 
     @property
-    def secrets(self) -> dict:
+    def secrets(self) -> list[str]:
         """The dict of secrets available to the operation (if any)."""
```
ritch, line 434: Update the type to match

fiftyone/operators/executor.py
```diff
@@ -430,20 +430,25 @@ def results(self):
         return self.request_params.get("results", {})
 
     @property
-    def secrets(self) -> dict:
-        """The dict of secrets available to the operation (if any)."""
-        return self._secrets
+    def secrets(self) -> list[str]:
```
brimoor, line 433: We've already released and documented `ctx.secrets` as a dict. Why do we need a breaking change here?

brimoor: - https://docs.voxel51.com/plugins/using_plugins.html#plugin-secrets
- https://docs.voxel51.com/teams/secrets.html#using-secrets
- https://docs.voxel51.com/plugins/developing_plugins.html#accessing-secrets

kaixi-wang: To avoid confusion on why a secret is None or potential errors raised from using ctx.secrets[key] and encourage users to use ctx.secret(key)

PR #3784, opened 2023-11-06
Add sample and frame counts to GraphQL Dataset object

benjaminpkane:
Adds efficient `estimated_sample_count` and `estimated_frame_count` fields to the `Dataset` object via MongoDB's [estimatedDocumentCount](https://www.mongodb.com/docs/manual/reference/method/db.collection.estimatedDocumentCount/).

Corresponding recoil selectors have been added to `@fiftyone/state`, named `datasetSampleCount` and `datasetFrameCount` for the `DatasetPage`. `datasetFrameCount` is `null` when a `frame_collection` is not defined.


fiftyone/core/odm/database.py
```diff
@@ -388,14 +395,16 @@ def get_db_conn():
     return _apply_options(db)
 
 
-def get_async_db_client():
+def get_async_db_client(use_global=False):
     """Returns an async database client.
 
+    Args:
+        reuse: whether to use the global client singleton
```
imanjra, line 402: Should this be `use_global`?

imanjra (approved):
Tested locally. Works great! Ty! LGTM.

fiftyone/core/odm/database.py
benjaminpkane: Yes, Thanks!

PR #3786, opened 2023-11-07
loosen OperatorConfig constructor signature

imanjra:

Loosen OperatorConfig constructor signature for future extensibility without the need to break backward compatibility

Using a test operator with an extra config kwarg

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/operators/operator.py
+++ b/fiftyone/operators/operator.py
@@ -48,6 +48,7 @@ class OperatorConfig(object):
         icon=None,
         light_icon=None,
         dark_icon=None,
+        **kwargs
     ):
         self.name = name
         self.label = label or name
@@ -61,6 +62,7 @@ class OperatorConfig(object):
         self.icon = icon
         self.dark_icon = dark_icon
         self.light_icon = light_icon
+        self.kwargs = kwargs  # unused, placeholder for future extensibility
 
     def to_json(self):
         return {


fiftyone/operators/operator.py
```diff
@@ -61,6 +62,7 @@ def __init__(
         self.icon = icon
         self.dark_icon = dark_icon
         self.light_icon = light_icon
+        self.kwargs = kwargs  # unused, placeholder for future extensibility
```
kaixi-wang, line 65: do you actually want to access them via self.kwargs[field] or do you want to be able to use them like attributes like self.field? ie something like
```
for k, v in kwargs.items():
    setattr(self, k, v)
```

imanjra: Currently, we don't have a use case for additional configurations. This enhancement is primarily to support adding a new kwarg in future without breaking backward compatibility of an operator with prior fo version

ritch: Don't forget why I added this in the first place: so that config was not just a `dict`. This to an extent removes why I added it in the first place. Which is fine, but we should do so knowingly.

`setattr` is interesting but also just a fancy `dict` if we add that. The benefit of the original style is the explicitness of the arguments.

ritch (approved):

PR #3787, opened 2023-11-07
Better App group slice persistence

benjaminpkane:
### Changes

* Fixes the `useSetGroupSlice` subscription event in `@fiftyone/app/useEvents`
* Fixes `groupSlice` persistence in multi-tab contexts and across page reloads/refreshes

The underlying goal of this change is to keep the group slice setting across page loads even when a view without a `group` media type is present

### Before

https://github.com/voxel51/fiftyone/assets/19821840/3c3c670b-bf3f-4cc1-8b16-022441c4b547

### After

https://github.com/voxel51/fiftyone/assets/19821840/b695e480-b3db-4c6c-8220-1922b6be0bd9

### DIFF
--- a/app/packages/app/src/useEvents/utils.ts
+++ b/app/packages/app/src/useEvents/utils.ts
@@ -1,5 +1,5 @@
-import { ColorSchemeInput } from "@fiftyone/relay";
-import { State, ensureColorScheme, useSessionSetter } from "@fiftyone/state";
+import { ColorSchemeInput, subscribeBefore } from "@fiftyone/relay";
+import { Session, State, ensureColorScheme } from "@fiftyone/state";
 import { env, toCamelCase } from "@fiftyone/utilities";
 import { atom } from "recoil";
 import { DatasetPageQuery } from "../pages/datasets/__generated__/DatasetPageQuery.graphql";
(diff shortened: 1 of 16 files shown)


fiftyone/core/state.py
```diff
@@ -75,6 +76,9 @@ def __init__(
         self.spaces = spaces
         self.color_scheme = color_scheme or build_color_scheme()
         self.field_visibility_stage = field_visibility_stage
+        if group_slice is None and view is not None:
+            group_slice = view.group_slice
+        self.group_slice = group_slice
```
benjaminpkane, line 81: We want to keep the `group_slice` setting, even when the view, e.g. a patches view, has no group slices

app/packages/app/src/useEvents/utils.ts
```diff
@@ -12,24 +12,28 @@ export const appReadyState = atom<AppReadyState>({
...
-  setter("sessionGroupSlice", state.group_slice);
-  setter("selectedSamples", new Set(state.selected));
-  setter(
-    "selectedLabels",
-    toCamelCase(state.selected_labels) as State.SelectedLabel[]
-  );
-  state.spaces && setter("sessionSpaces", state.spaces);
-  setter("fieldVisibilityStage", state.field_visibility_stage || undefined);
+
+  const unsubscribe = subscribeBefore<DatasetPageQuery>(({ data }) => {
```
benjaminpkane, line 22: Moves state update changes to the page change recoil transaction

fiftyone/core/state.py
```diff
@@ -113,7 +117,7 @@ def serialize(self, reflective=True):
                 ]
 
                 view = self.view if self.view is not None else self.dataset
-                if view.media_type == fom.GROUP:
+                if view.group_slice:
```
manivoxel51, line 120: This was breaking the two tab issue I was talking about. I had to bring it back and modify to view.group_slice for the listing tab to get the updated groupSlice through the session

manivoxel51 (approved):
lgtm! 🍨

fiftyone/core/state.py
benjaminpkane: Ok, adding this will break another edge case. Let's discuss offline tomorrow

benjaminpkane: I reverted this for another edge case, but let's follow up offline so we can fix

PR #3789, opened 2023-11-08
Use non-default browser to launch the app

Rusteam:

Fiftyone works best in Chrome, while users might prefer other browsers for surfing the internet.

Add a `--browser <browser_name>` parameter to `fiftyone app launch` to use a non-default browser to open the App.

Running these commands:
```
fiftyone app launch -b chrome
fiftyone app launch --browser safari
fiftyone app launch --browser oper  # fails (no browser registered with this name)
 fiftyone app launch --browser opera  # fails (non-executable found)
```

        notes for FiftyOne users.

Launch the app in the non-default browser. 

-   [ ] Other

### DIFF
--- a/docs/source/cli/index.rst
+++ b/docs/source/cli/index.rst
@@ -1864,7 +1864,7 @@ Launch the FiftyOne App.
 
 .. code-block:: text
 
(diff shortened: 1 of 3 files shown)


benjaminpkane (approved):
Awesome! Great addition. Looks like all the details are here 🙌

PR #3792, opened 2023-11-08
Add ctx.current_sample

ritch:
See [an example](https://github.com/voxel51/fiftyone-plugins/pull/83/files) operator that shows the usage of this new field.


benjaminpkane (approved):
Tested. LGTM 🚀 
Screenshot 2023-11-09 at 10 11 21 AM

PR #3766, opened 2023-11-02
new empty pages

imanjra:
#### When there are no datasets in FiftyOne
image
image

#### When there are datasets in FiftyOne, but none is selected
image
image

#### When a dataset is selected, but there are no samples in the dataset
image
image

## How is this patch tested? If it is not, please explain why.

See preview above

### Is this a user-facing change that should be mentioned in the release notes?

-   [ ] No. You can skip the rest of this section.
-   [x] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

See above

### What areas of FiftyOne does this PR affect?

-   [x] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

PR #3793, opened 2023-11-09
sync placement loading with native actions

imanjra:

Sync loading of placements with native actions

PR depends on #3766

Interactively in the app by making sure app load waits for placement by introduce a loading delay temporarily 

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/plugins/src/index.ts
+++ b/app/packages/plugins/src/index.ts
@@ -1,3 +1,4 @@
+import { useOperators } from "@fiftyone/operators";
 import * as fos from "@fiftyone/state";
 import * as fou from "@fiftyone/utilities";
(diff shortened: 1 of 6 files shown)


app/packages/plugins/src/index.ts
```diff
@@ -151,13 +156,14 @@ export function usePlugins() {
       })
       .then(() => {
         setState("ready");
+        setPluginsInitialized(true);
```
ritch, line 159: Seems redundant to `setState('ready')`

ritch (approved):

app/packages/plugins/src/index.ts
imanjra: Good point. I should've just externalized the local state instead. Will fix this

PR #3799, opened 2023-11-09
Update Labelbox integration to handle existing data rows

ehofesmann:
In our Labelbox integration, each sample that is annotated is uploaded as a Labelbox data row with a global key that is equal to the FiftyOne sample id. However, these are globally unique, so if the same samples are being annotated twice, they cannot and should not be uploaded again.

This PR adds logic that checks if existing samples already exists in Labelbox by referencing existing global keys and avoids reuploading them if they are found.


brimoor (approved):
LGTM

PR #3801, opened 2023-11-10
Optimize metadata computation

brimoor:
Optimizes `SampleCollection.compute_metadata()` in two ways:
1. Use thread pool instead of process pool when retrieving metadata
2. Use `set_values()` rather than `sample.save()` to save metadata to the DB

Benefits of 1
- Avoids unnecessary use of multiprocessing, which is not suitable for async environments (App server) and other situations (eg S3 Lambda) or on Windows in general
- This thread pool implementation has similar performance to the prior process pool implementation. I've tested both when the media are local and when the media are URLs that require HTTP requests

Benefits of 2
- Greatly improves performance in situations when DB connections have higher overhead, as requests are batched into single `values()` and  `set_values()` calls

### DIFF
--- a/fiftyone/core/metadata.py
+++ b/fiftyone/core/metadata.py
@@ -295,7 +295,7 @@ def get_image_info(f):
     return (img.width, img.height, len(img.getbands()))
 
 
-def _compute_metadata(sample_collection, overwrite=False):
+def _compute_metadata(sample_collection, overwrite=False, batch_size=1000):
     if not overwrite:
         sample_collection = sample_collection.exists("metadata", False)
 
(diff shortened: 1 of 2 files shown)


fiftyone/core/metadata.py
```diff
@@ -291,14 +299,27 @@ def _compute_metadata(sample_collection, overwrite=False):
...
+    inputs = zip(ids, filepaths, media_types)
+    values = {}
+
+    try:
+        with fou.ProgressBar(total=num_samples) as pb:
+            for args in pb(inputs):
+                sample_id, metadata = _do_compute_metadata(args)
+                values[sample_id] = metadata
+    finally:
+        sample_collection.set_values("metadata", values, key_field="id")
```
ehofesmann, line 322: Would it be possible to `set_values()` in batches of some preconfigured size? Users often need to compute metadata on very large datasets and I'm worried that they could run into memory issues storing the `Metadata` objects in a dictionary for millions of samples.

brimoor: FYI the size of `values` in-memory is ~1.7GB for 10M samples, so it would likely take O(100M) samples to run out of RAM:

```py
import sys
import eta.core.utils as etau

import fiftyone as fo
import fiftyone.zoo as foz

dataset.compute_metadata()

size = 0
for i, m in zip(*dataset.values(["id", "metadata"])):
    size += sys.getsizeof(i)
    size += sys.getsizeof(m)

n = 10000000
size_n = size * (n / len(dataset))

print("Size for %d samples: %s" % (n, etau.to_human_bytes_str(size_n)))
# Size for 10000000 samples: 1.7GB
```

brimoor: That said, I do agree some batching would be prudent here. No reason not to 👍

brimoor: Added batching of 1000 samples 👍

PR #3803, opened 2023-11-11
Exhaust generator operators when executing directly

brimoor:
Fixes a bug where `foo.execute_operator()` would not exhaust generator operators. While fixing this, took the opportunity to remove code duplication, since the same logic was already in-place when executing delegated operations.

```py
import fiftyone as fo
import fiftyone.zoo as foz

ctx = {
    "dataset": dataset,
    "params": {
        "num_workers": 1,
        "delegate": False,
    }
}

# Previously did no computation, now works as expected
foo.execute_operator("@voxel51/utils/compute_metadata", ctx)
```

### DIFF
--- a/fiftyone/operators/delegated.py
+++ b/fiftyone/operators/delegated.py
@@ -6,17 +6,14 @@ FiftyOne delegated operations.
 |
 """
 import asyncio
-import collections
-import inspect
 import logging
 import traceback
-import types as python_types
-import fiftyone.core.utils as fou
 
 from fiftyone.factory.repo_factory import RepositoryFactory
 from fiftyone.factory import DelegatedOperationPagingParams
 from fiftyone.operators.executor import (
     prepare_operator_executor,
+    do_execute_operator,
     ExecutionResult,
     ExecutionRunState,
     ExecutionProgress,
@@ -42,7 +39,8 @@ class DelegatedOperationService(object):
         Args:
             operator: the operator name
             delegation_target (None): an optional delegation target
-            label (None): an optional label for the operation (will default to the operator if not supplied)
+            label (None): an optional label for the operation (will default to
+                the operator if not supplied)
             context (None): an
                 :class:`fiftyone.operators.executor.ExecutionContext`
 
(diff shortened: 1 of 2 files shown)


fiftyone/operators/delegated.py
```diff
@@ -362,30 +367,8 @@ async def _execute_operator(self, doc, log=False, run_link=None):
             set_progress=self.set_progress,
         )
 
-        # if a validation error happened during preparation,
-        # only an ExecutionResult with an error is returned.
-        # Raise it so the delegated operation is marked as a failure.
```
brimoor, line 367: In the previous implementation, validation errors would result in `set_failed()` being called without `set_running()` ever being called, which seemed weird to me? Now all operations will have a running time logged before possibly failing.

fiftyone/operators/delegated.py
```diff
@@ -362,30 +367,8 @@ async def _execute_operator(self, doc, log=False, run_link=None):
...
         if isinstance(prepared, ExecutionResult):
             raise prepared.to_exception()
-        else:
-            operator, _, ctx = prepared
 
-            if log:
-                logger.info("Running operator %s", operator_uri)
```
brimoor, line 374: This was a duplicate log

fiftyone/operators/executor.py
```diff
@@ -262,11 +264,26 @@ async def prepare_operator_executor(
...
+    result = await (
+        operator.execute(ctx)
+        if asyncio.iscoroutinefunction(operator.execute)
+        else fou.run_sync_task(operator.execute, ctx)
     )
 
+    if not exhaust:
+        return result
+
+    if inspect.isgenerator(result):
```
ritch, line 277: Not a fan of swallowing the messages the operator is sending via the generator. We should probably allow for the callee to pass in a function that processes these. Otherwise there is no path to handling messages when calling `execute_operator()` directly.

ritch: Although in this could be the behavior of `exhaust` which would be the default. I'm thinking mostlyu about adding `execute_operator(message_reader=MyMessageReader())` or similar.

ritch (approved):

fiftyone/operators/executor.py
brimoor: Sounds potentially useful to me 👍 

This PR just unifies how `execute_operator()` work with generators with how delegated operator execution works (which already "swallows" generators)

PR #3805, opened 2023-11-14
Sample/frame created_at & Dataset/Sample/Frame last_updated_at

swheaton:

Add `created_at` field to all samples and frames. It's set to "now" when added to the dataset, and reset back to `None` if removed from the dataset.

Add `last_updated_at` field to all samples and frames. It's set to now when added to the dataset, , and reset back to `None` if removed from the dataset.

Add `last_updated_at` to Dataset document which is updated whenever the document is saved.

Add `Dataset.last_updated_at` property which is equal to this pseudocode:

```python
max(DatasetDocument.last_updated_at, max(sample.last_updated_at for sample in dataset.samples))
```

### TODO
- [ ] Ensure generated views syncing back to base dataset updates last_updated_at. Resolve in-code TODO surrounding video dataset merge.
- [ ] When a frame is updated independently of a sample, make sure sample update time is separately updated.
- [ ] Check SaveContext
- [ ] Test for merge_python
- [ ] Complete audit of public mutator functions to ensure all base mutators are covered (possibly after `develop` cut)

### Notes
- Locations to add `created_at` were identified using as a template [adding dataset id to sample and frames docs](https://github.com/voxel51/fiftyone/pull/2711) 
- Samples get a creation time only when they're added to a dataset. We could set `created_at` if the sample gets created outside of the dataset first, but it's easier to do it this way and there's not likely to be a meaningful distinction between when a sample was created and when added to dataset.
- Merge for existing samples was a little tricky but if we first `addField` but then add it to `delete_fields` when we create the `whenMatched` pipeline, we'll make sure that it will be added on insert but ignored on merge.
- Writes to these fields through `set_field`, `__setattr__`, `set_values()`, `clear_sample_field`, etc., are prohibited.
   -  I attempted to achieve this by creating a `readonly` flag for the `Field`, akin to the `is_virtual` property in [virtual fields PR](https://github.com/voxel51/fiftyone/pull/2429).
- In a long `add_samples()` call, each batch of samples will get the same `created_at` time.

- Added tests for each case in both image and video, and ran manually. Created_at is set on creation but not again. Distinct cases below:
   - `add_samples()`
   - `clone()`
   - `add_collection(new_ids=True)`
   - `merge_samples()  # both existing and new samples`
   - `fo.Dataset.from_dir()`
- Then tested that any manner of modifying the field throws an error.
- Updating a sample in any of myriad of ways updates its `last_updated_at` time, and `dataset.last_updated_at` is equal.
- Updating Dataset metadata updates its `last_updated_at` time.
- If you want to test manually, following along with the test cases is best rather than pasting full here.

        notes for FiftyOne users.

Add `created_at` default `DateTimeField` to samples and frames, which contains the time the sample was added to the dataset.

-   [ ] Other

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -5,6 +5,7 @@ Interface for sample collections.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+import datetime
 from collections import defaultdict
 from copy import copy
 import fnmatch
@@ -1503,6 +1504,8 @@ class SampleCollection(object):
 
     def _edit_sample_tags(self, update):
         ops = []
+        now = datetime.datetime.utcnow()
+        update.update({"$set": {"last_updated_at": now}})
         for ids in fou.iter_batches(self.values("_id"), 100000):
             ops.append(UpdateMany({"_id": {"$in": ids}}, update))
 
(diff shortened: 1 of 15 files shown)


swheaton:
superseded by https://github.com/voxel51/fiftyone/pull/4597

PR #3806, opened 2023-11-14
Synchronously resolve secrets via ctx.secret

kaixi-wang:
- Currently, secrets are only resolved in the ExecutionContext via the `prepare_operator_executor()` method. This limits secret resolution to the execution phase when running an operator. 
- Adding sync methods for querying secrets enables on demand access to secrets anywhere in user defined code/plugin operator code via `ctx.secret(<KEY>)`
- added custom dictionary class `SecretsDictionary` for `ctx.secrets`:
  -  enables `ctx.secrets[<key>]` to also resolve plugin secrets automatically
  - adds some security by disabling copying or printing all values via .items() or .values()
  
- **Note**: `ctx.secret(<KEY>)` will only resolve a secret if the `KEY` is included in the secrets section of the plugin's `fiftyone.yml`. Will return None even if the value is set, but the key is not listed in the plugin's definition. 
image

- add sync request methods
- add unit tests

- add unit tests
- tested locally

        notes for FiftyOne users.

Enables on demand access to secrets anywhere in user defined code/plugin operator code via `ctx.secret(<KEY>)` or `ctx.secrets[<KEY>]`

-   [ ] Other

### DIFF
--- a/fiftyone/internal/secrets/providers/iprovider.py
+++ b/fiftyone/internal/secrets/providers/iprovider.py
@@ -19,10 +19,27 @@ class ISecretProvider(abc.ABC):
         """
         ...
 
+    @abc.abstractmethod
+    def get_sync(self, key: str, **kwargs) -> Optional[ISecret]:
+        """
+        Get a secret by key in a synchronous context.
+        """
+        ...
+
+    @abc.abstractmethod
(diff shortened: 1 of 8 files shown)


ritch (approved):
LGTM

PR #3811, opened 2023-11-16
Prevent receding loading state on empty page transition

benjaminpkane:
Removes a small pixelating state flicker when navigating from the empty dataset page. Wondering if this change is ok @imanjra?

## Before

https://github.com/voxel51/fiftyone/assets/19821840/4a81fdc4-e385-4a70-a388-c45d9818aae9

## After

https://github.com/voxel51/fiftyone/assets/19821840/6887fa01-560f-41ad-a3b0-94b186fdd797

### DIFF
--- a/app/packages/app/src/Sync.tsx
+++ b/app/packages/app/src/Sync.tsx
@@ -31,7 +31,8 @@ export const SessionContext = React.createContext<Session>(SESSION_DEFAULT);
 
 const Plugins = ({ children }: { children: React.ReactNode }) => {
   const plugins = usePlugins();
-  if (plugins.isLoading) return <Loading>Pixelating...</Loading>;
+
+  if (plugins.state === "loading") return <Loading>Pixelating...</Loading>;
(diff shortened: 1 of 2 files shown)


imanjra:
@benjaminpkane Unfortunately, this will undo the synchronization of placement loading (https://github.com/voxel51/fiftyone/pull/3793). There are couple of other options I considered as an alternative to global Pixelating state. One of them was moving the pixelating state downward to Spaces container where sidebar would load and sample would say Pixelating... until placements are loaded. 


_Unrelated: thoughts on replacing `Pixelating...` screen with actual pixel like sample panel?_

benjaminpkane:
Adding the dataset operators to the GraphQL page query would be ideal. But in the meantime, I will add the operators request to the page query `subscribe`  function in `@fiftyone/relay` in this PR for initialization. Converting to draft for now

Falling pixels could be cool 👍 can take offline

sashankaryal (approved):
lgtm

benjaminpkane:
Closing for now

PR #3746, opened 2023-10-31
Officially add SageMaker Notebooks support

benjaminpkane:
Adds SageMaker support with new `amzn` distribution support in `fiftyone-db` and improved proxy handling. Updates documentation FAQ, troubleshooting, and environments pages

### Testing

In SageMaker, [create a notebook instance](https://docs.aws.amazon.com/sagemaker/latest/dg/gs-setup-working-env.html)

Open the notebook and install the RCs with
```py
!RELEASE_VERSION=0.5.0rc1 pip install fiftyone-db==0.5.0rc1 fiftyone==0.23.0rc2
```

Launch the App with a configured proxy
```py
import fiftyone as fo
fo.app_config.proxy_url = "https://<notebook-instance>.notebook.us-east-2.sagemaker.aws/proxy/5151"

dataset, session = fo.quickstart()
```

The changes can also be tested in Colab, but only `session.url` is functional. See #3644 

```py
!RELEASE_VERSION=0.5.0rc1 pip install fiftyone-db==0.5.0rc1 fiftyone==0.23.0rc2
import fiftyone as fo

# in colab, proxy_url is not needed
dataset, session = fo.quickstart()
```
Screenshot 2023-11-06 at 4 05 28 PM

PR #3819, opened 2023-11-17
fiftyone-db 1.0

benjaminpkane:
For v0.23.0, I propose we bump the `fiftyone-db`package to 1.0 to indicate general support for Linux distributions


benjaminpkane:
Closing in favor of #3746

PR #3820, opened 2023-11-18
Allow iterating ctx.secrets.items()

kaixi-wang:

Re-enable iterating through secrets like a dict (TEAMS-1980)

### DIFF
--- a/fiftyone/plugins/secrets.py
+++ b/fiftyone/plugins/secrets.py
@@ -5,12 +5,11 @@ Plugin secrets resolver.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-# from __future__ import annotations
 import logging
-import re
+import traceback
 import typing
-from collections.abc import Mapping
 from typing import Optional
+
 from ..internal import secrets as fois
 
 
@@ -104,7 +103,7 @@ def _get_secrets_client():
     return client()
 
 
-class SecretsDictionary(Mapping[str, str]):
+class SecretsDictionary:
     """
     A more secure dictionary for accessing plugin secrets in
     operators that will attempt to resolve missing plugin secrets upon access.
@@ -112,17 +111,26 @@ class SecretsDictionary(Mapping[str, str]):
 
     def __init__(
         self,
-        secrets_dict,
-        operator_uri=None,
-        resolver_fn=None,
-        required_keys=None,
+        secrets_dict: typing.Dict[str, str],
+        operator_uri: str = None,
+        resolver_fn: typing.Callable = None,
+        required_keys: typing.List[str] = None,
     ):
-        self.__secrets = secrets_dict
-        self.__required_keys = required_keys
-        self._operator_uri = operator_uri
-        if resolver_fn:
-            self._resolver = resolver_fn
+        self.__frozen = False
+        if required_keys:
+            self.__required_keys = frozenset(required_keys)
+            self.__secrets = secrets_dict
+            self._operator_uri = operator_uri
+            if resolver_fn:
+                self._resolver = resolver_fn
+            else:
+                self._resolver = None
+
+            self.__frozen = True
         else:
+            self.__required_keys = []
+            self.__secrets = {}
+            self._operator_uri = None
             self._resolver = None
 
     def __len__(self):
@@ -139,17 +147,44 @@ class SecretsDictionary(Mapping[str, str]):
     def __getitem__(self, key):
         # Override __getitem__ to suppress KeyError and attempt to resolve
         # plugin secrets if not yet resolved
+        if not self.__frozen:
+            return None
+        if key not in self.__required_keys:
+            return None
         val = self.__secrets.get(key, None)
         if self._resolver and val is None:
             val = self._resolver(key=key, operator_uri=self._operator_uri)
             if val:
-                self.__secrets[key] = val
+                self.__secrets[key] = val.value
+            else:
+                # If the secret is not found, set it to an empty string to
+                # prevent continually trying to resolve it
+                self.__secrets[key] = ""
         return val
 
     def __setattr__(self, key, value):
-        if re.search("_SecretsDictionary__(secrets|required_keys)$", key):
-            raise KeyError("Cannot mutate hidden properties")
-        else:
+        if not getattr(
+            self, "_SecretsDictionary__frozen", False
+        ) or not hasattr(self, key):
+            super().__setattr__(key, value)
+            return
+
+        if key == "_SecretsDictionary__secrets" and value != {}:
+            is_allowed = all(v in self.__required_keys for v in value.keys())
+            if not is_allowed:
+                raise KeyError(
+                    "Cannot access secrets not defined in the "
+                    "plugin's definition"
+                )
+        elif key == "_SecretsDictionary__required_keys" and self.__frozen:
+            raise AttributeError("Cannot mutate plugin secrets requirement")
+        elif (
+            key == "_SecretsDictionary__frozen"
+            and len(self.__required_keys) > 0
+        ):
+            raise AttributeError("Cannot mutate plugin secrets requirement")
+
+        elif hasattr(super(), key):
             super().__setattr__(key, value)
 
     def __setitem__(self, key, value):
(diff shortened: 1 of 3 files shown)


fiftyone/plugins/secrets.py
```diff
@@ -118,6 +125,12 @@ def __init__(self, secrets_dict, operator_uri=None, resolver_fn=None):
     def __eq__(self, other):
         return self.__secrets == other
 
+    def __setattr__(self, key, value):
+        if key == "__secrets" or key == "__required_keys":
```
brimoor, line 129: This isn't necessary. Double underscore attributes already aren't accessible via `d.__secrets` from outside the class. Python does weird name mangling that makes them available as `d._SecretsDictionary__secrets`.

Either way, don't waste effort trying to prevent things from happening if users modify private variables. It's Python. If the user really wants to modify private variables then they can mess things up infinitely worse than just overriding secrets.

fiftyone/plugins/secrets.py
```diff
@@ -155,5 +168,7 @@ def values(self):
         raise RuntimeError("Iteration through values is not allowed")
 
     def items(self):
-        # Override items() to prevent iterating through plaintext values
-        raise RuntimeError("Iteration through items is not allowed.")
+        # Override items() to use __getitem__ to automatically resolve
```
brimoor, line 171: `values()` should be allowed too. It's a documented as a dict, so users should be allowed to do dict-like things

fiftyone/plugins/secrets.py
```diff
@@ -155,5 +168,7 @@ def values(self):
...
+        # missing secrets upon iteration
+        for key in self.keys():
+            yield key, self[key]
```
brimoor, line 174: IDK how the resolver is implemented in teams, but need to verify: is calling `self[key]` in a loop performant? IE will it only make one API call to pull in all secrets?

There are plugins that declare 10+ secrets, so 10 API requests would be 😢

brimoor (changes requested):
Inheriting from `Mapping` would give you a clean way to implement a read-only dict interface:

```py
from collections.abc import Mapping

class SecretsDictionary(Mapping):

    def __init__(self, secrets):
        self._secrets = secrets

    def __getitem__(self, key):
        # resolve here
        return self._secrets[key]

    def __len__(self):
        return len(self._secrets)

    def __iter__(self):
        return iter(self._secrets)

d = SecretsDictionary({"foo": "bar", "spam": "eggs"})
```

```py
print(d["foo"])  # bar
print(d["spam"])  # eggs
print(len(d))  # 2
print(list(d))  # ['foo', 'spam']
print(list(d.keys()))  # ['foo', 'spam']
print(list(d.values()))  # ['bar', 'eggs']
print(list(d.items()))  # [('foo', 'bar'), ('spam', 'eggs')]

d["foo"] = "hi"  # TypeError: 'SecretsDictionary' object does not support item assignment
```

fiftyone/plugins/secrets.py
kaixi-wang: this is the best option right now. Anyways, the secrets are only potentially unresolved in resolve_input(). Any usage of the secrets in execute or resolve_ouput() should already have batch resolved the secrets. Moreover, self[key] will not only resolve it, but add it to the secrets dict so the maximum number of requests is the number of unique secrets. If performance is an issue, then it is likely the number of times the context is regenerated/operators are initialized.

fiftyone/plugins/secrets.py
kaixi-wang: I'm against this, but fine

fiftyone/plugins/secrets.py
kaixi-wang: I don't see it as wasted effort because at the end of the day if they mess it up, it comes back to me to debug. I'd rather prevent mutation from the start rather than allow it.

ritch (approved):

ritch:
I've tested this locally and can mutate the secrets and required_keys.

```
d = SecretsDictionary({"foo": "bar"}, required_keys=["foo"])
d._SecretsDictionary__required_keys.append('bar')
d._SecretsDictionary__secrets = {'bar': 'bat'}

d.keys()
# => ['bar']
```

ritch (changes requested):
Please fix the ability to mutate `__secrets` and `__required_keys`.

kaixi-wang:
> Please fix the ability to mutate `__secrets` and `__required_keys`.

@brimoor requested I remove that part...

kaixi-wang:
Disabled setting the value. Attempts will throw an error 
```
Error loading operators from server

     ...
    fiftyone/plugins/secrets.py", line 223, in __setattr__
    raise KeyError("Cannot mutate hidden properties")
KeyError: 'Cannot mutate hidden properties'
```

fiftyone/plugins/secrets.py
```diff
@@ -5,8 +5,11 @@
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+# from __future__ import annotations
```
brimoor, line 8: please remove commented code

fiftyone/plugins/secrets.py
```diff
@@ -5,8 +5,11 @@
...
 import logging
+import re
 import typing
+from collections.abc import Mapping
```
brimoor, line 12: nit: style guide says to alphabetize imports

fiftyone/plugins/secrets.py
```diff
@@ -101,20 +104,35 @@ def _get_secrets_client():
     return client()
 
 
-class SecretsDictionary:
+class SecretsDictionary(Mapping[str, str]):
```
brimoor, line 107: just checking: is this syntax valid in Python 3.7? We still promise support for it

fiftyone/plugins/secrets.py
```diff
@@ -101,20 +104,35 @@ def _get_secrets_client():
...
         if resolver_fn:
             self._resolver = resolver_fn
         else:
             self._resolver = None
 
+    def __len__(self):
+        """Returns the number of secrets defined in the plugin definition"""
+        return len(self.__required_keys)
+
+    def __iter__(self):
```
brimoor, line 132: For dicts, `__iter__` should only return keys, not values

fiftyone/plugins/secrets.py
```diff
@@ -128,32 +146,36 @@ def __getitem__(self, key):
...
+        logging.warning("Copying the SecretsDictionary values is not allowed.")
         return self.__deepcopy__()
 
     def keys(self):
         return [k for k, v in self.__secrets.items() if v is not None]
 
     def values(self):
-        # Override values() to prevent iterating through plaintext values
-        raise RuntimeError("Iteration through values is not allowed")
+        # Override values() to ensure that resolvable secrets are always returned upon iteration
```
brimoor, line 174: My suggestion to inherit from `Mapping` had two benefits:
- it automatically handles throwing errors when mutation methods like `__setitem__` and `update()` are called
- you only have to implement `__getitem__`, `__len__`, and `__iter__` as described [here](https://github.com/voxel51/fiftyone/pull/3820#pullrequestreview-1740197017) and then other methods like `keys()`, `values()`, and `items()` will automatically use those in the correct ways.

In other words, once you fix your `__iter__` implementation I believe `keys()`, `values()`, and `items()` can just be deleted.

fiftyone/plugins/secrets.py
```diff
@@ -128,32 +146,36 @@ def __getitem__(self, key):
                 self.__secrets[key] = val
         return val
 
+    def __setattr__(self, key, value):
+        if re.search("_SecretsDictionary__(secrets|required_keys)$", key):
```
brimoor, line 150: Still don't see a point in preventing `__setattr__` here when the user can still do other things like `d._SecretsDictionary__secrets["bad"] = "thing"`.

fiftyone/plugins/secrets.py
```diff
@@ -128,32 +146,36 @@ def __getitem__(self, key):
...
+            super().__setattr__(key, value)
+
     def __setitem__(self, key, value):
         raise RuntimeError("Setting values is not allowed")
 
     def __deepcopy__(self, memodict={}):
-        logging.warning(
-            "Copying the SecretDictionary values is not " "allowed."
-        )
+        logging.warning("Copying the SecretsDictionary values is not allowed.")
```
brimoor, line 159: Why log a warning that something is now allowed but let it happen anyway?

Also why is this method implemented? It returns a regular dict so is not actually copying the object, it returns a different type of object.

Why not let the default implementations of `copy()` and `__deepcopy__` do their things?

brimoor (approved):
Approving to unblock.

The `__iter__` implementation change needs to happen, and I don't understand why `__deepcopy__` is implemented.

Other comments would cleanup code.

fiftyone/plugins/secrets.py
kaixi-wang: currently, there will be no error if you manually set the secret, however, it will not show up when you iterate through ctx.secrets.items()

fiftyone/plugins/secrets.py
kaixi-wang: removed

fiftyone/plugins/secrets.py
kaixi-wang: removed

fiftyone/plugins/secrets.py
brimoor: Fixing this is important because users often iterate over dicts like this:

```py
for key in ctx.secrets:
    value = ctx.secrets[key]
```

fiftyone/plugins/secrets.py
kaixi-wang: To discourage copying the secret values and passing them around and encourage usage exclusively via ctx.secret/s both for security reasons and to avoid potential issues with copying the value to an object and expecting that the object have the secrets when called in another part of the code due to the execution context potentially being reset inbetween

fiftyone/plugins/secrets.py
kaixi-wang: fixed

fiftyone/plugins/secrets.py
ritch: Would overriding `__getattr__` be the solution to reading/mutating the underlying dictionary?

ritch (approved):
LGTM

PR #3827, opened 2023-11-21
[WIP] Allow reloading database connection

brimoor:
Marked as draft for discussion.

Inspired by a Slack community question, I wanted to see what it would look like to allow changing databases mid-Python session by calling a `fo.reload()` method.

I believe this PR is fully functional for programmatic usage, but there are a couple snags when using the App:
- The App runs in a subprocess which currently re-reads its `fo.config` from environment variables, so you have to change database via `os.environ` rather than just editing `fo.config` in order for `launch_app()` to work after the database change in the example below
- If I `launch_app()` before _and_ after the database change, the second App does not connect to the new database

```py
import os
import fiftyone as fo
import fiftyone.zoo as foz

# do stuff with default database
print(fo.list_datasets())

#
# Change database
#

"""
os.environ["FIFTYONE_DATABASE_URI"] = "mongodb://localhost:12345"
os.environ["FIFTYONE_DATABASE_DIR"] = "/path/to/new/db"
fo.reload(hard=True)
"""

fo.config.database_uri = "mongodb://localhost:12345"
fo.config.database_dir = "/path/to/new/db"
fo.reload()

# do stuff with new database
print(fo.list_datasets())

# Only works with `fo.reload(hard=True)` since subprocess reads config from env vars
```

### DIFF
--- a/fiftyone/__public__.py
+++ b/fiftyone/__public__.py
@@ -6,12 +6,39 @@ FiftyOne's public interface.
 |
 """
 
-import fiftyone.core.config as _foc
+config = None
+annotation_config = None
+evaluation_config = None
+app_config = None
 
-config = _foc.load_config()
-annotation_config = _foc.load_annotation_config()
-evaluation_config = _foc.load_evaluation_config()
-app_config = _foc.load_app_config()
+
+def reload(hard=False):
+    """Reloads the current database connection.
+
+    Args:
+        hard (False): whether to reconnect using the current in-memory config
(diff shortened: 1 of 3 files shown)


kaixi-wang:
unsure how/if changing the mongo connection would affect pymongo, but through aliased connections (https://docs.mongoengine.org/guide/connecting.html#switch-database), you could maybe hack a way to overcome the snags you mentioned?

PR #3834, opened 2023-11-24
Update to read intensity and visualization

Leo-B-Y:

1. update three to support `intensity` field parse in pcd
2. Determine if `intensity` field exists, if not then use `color.r` for _ShadeByIntensity_

test local with pcd file which has `intensity` field
```
# .PCD v0.7 - Point Cloud Data file format
VERSION 0.7
FIELDS x y z intensity timestamp ring
SIZE 4 4 4 4 8 2
TYPE F F F F F U
COUNT 1 1 1 1 1 1
WIDTH 201191
HEIGHT 1
VIEWPOINT 0 0 0 1 0 0 0
POINTS 201191
DATA binary_compressed
```

image

        notes for FiftyOne users.

There is almost no impact on existing users, if pcd doesn't have an intensity field it will continue to use color.r for rendering, maybe the documentation section needs to be updated a bit

-   [ ] Other

### DIFF
--- a/app/packages/looker-3d/src/renderables/pointcloud/shaders.tsx
+++ b/app/packages/looker-3d/src/renderables/pointcloud/shaders.tsx
@@ -9,6 +9,7 @@ export type ShaderProps = {
   max: number;
   pointSize: number;
   isPointSizeAttenuated: boolean;
+  useIntensity: boolean;
 };
 
 const useGradientMap = (gradients: Gradients) => {
@@ -120,9 +121,11 @@ const ShadeByIntensityShaders = {
   uniform float min;
   uniform float pointSize;
   uniform bool isPointSizeAttenuated;
+  uniform bool useIntensity;
 
   varying vec2 vUv;
   varying float hValue;
+  attribute float intensity;
   attribute vec3 color;
 
   float remap ( float minval, float maxval, float curval ) {
@@ -132,7 +135,8 @@ const ShadeByIntensityShaders = {
   void main() {
     vUv = uv;
     vec3 pos = position;
-    hValue = remap(min, max, color.r);
+    float effectiveIntensity = useIntensity ? intensity : color.r;
+    hValue = remap(min, max, effectiveIntensity);
 
     vec4 mvPosition = modelViewMatrix * vec4(pos, 1.0);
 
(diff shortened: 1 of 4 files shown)


app/packages/looker-3d/src/renderables/pointcloud/shaders.tsx
```diff
@@ -132,7 +135,8 @@ const ShadeByIntensityShaders = {
   void main() {
     vUv = uv;
     vec3 pos = position;
-    hValue = remap(min, max, color.r);
+    float effectiveIntensity = useIntensity ? intensity : color.r;
+    hValue = remap(min, max, effectiveIntensity);
```
ritch, line 139: Its been a while since I've worked on this but I believe the `hValue` mapping will changed when `useIntensity` is `true`. I think the `min` and `max` values will need to be derived in one way or another from the intensity range (possibly `(0, 1)` ?).

ritch: To clarify `min` and `max` should probably be renamed `colorMin` and `colorMax` which would be unrelated to the `intensity` min/max.

ritch (changes requested):
A couple things are needed for us to the land this change and release it. I think we can help with some of these:

1. if you can fix the issue with remapping the intensity (or propose an alternative)
2. we need to add this new support to the documentation, we can help with this
3. regression test the upgrade of three.js
4. regression test this change with other PCD files

Thanks for the PR!

ritch:
We might be able to help with all the items I've mentioned in my comments. Thanks again for the contribution.

app/packages/looker-3d/src/renderables/pointcloud/shaders.tsx
Leo-B-Y: In general, it's common for `intensity` values in PCD files to be mapped to an 8-bit integer range, typically ranging from 0 to 255. However, to accommodate potential exceptions and ensure broader compatibility, I have utilized the `setColorMinMax(computeMinMaxForColorBufferAttribute(intensityAttribute))` in my code. This function dynamically determines the range of intensity values, which is then passed to the ShadeByIntensityShaders. This method ensures that our shading logic remains flexible and accurate, regardless of the specific range of intensity values encountered in different PCD files.

Leo-B-Y:
I apologize for the build failure, I didn't run the yarn command in codespace, the `yarn.lock` file has been updated.

app/packages/looker-3d/src/renderables/pointcloud/shaders.tsx
ritch: Awesome that looks right to me!

PR #3841, opened 2023-11-27
Set color scheme tweaks

benjaminpkane:
Moves `removeRgbProperty` to `@fiftyone/state` for reuse, and ensures an color scheme `id` in the `setColorScheme` mutation

handle scenarios that `ensureColorscheme` takes `null` as the first argument.


lanzhenw (approved):
Thanks, ben! Added https://github.com/voxel51/fiftyone/pull/3826 here, as it previously use the wrong base branch.

PR #3842, opened 2023-11-28
[WIP] Secrets interface cleanup

brimoor:
Change log
- Avoids usage of `operator._plugin_secrets` outside of the `Operator` class
- All secret resolution now happens in `ExecutionContext`, as opposed to being partly in `ExecutionContext` and partly in `SecretsDictionary`

Notes
- Ideally (to me) it would be the case that all secret resolution happens when `ctx.resolve_secret_values()` is called. Currently that is not the case, as it also appears to be necessary for `ctx.secrets` to dynamically resolve secrets. That behavior is maintained in this PR (although all secrets are resolved in a batch the first time `ctx.secrets` is accessed, rather than resolving one-by-one).

### DIFF
--- a/fiftyone/operators/operator.py
+++ b/fiftyone/operators/operator.py
@@ -120,12 +120,13 @@ class Operator(object):
         # Plugin names are populated when the operators are registered
         plugin_name = BUILTIN_OPERATOR_PREFIX if _builtin else None
 
-        self._builtin = _builtin
-        self._plugin_secrets = None
         self.plugin_name = plugin_name
         self.definition = Object()
         self.definition.define_property("inputs", Object())
         self.definition.define_property("outputs", Object())
+
+        self._builtin = _builtin
+        self._secrets = None
 
     @property
     def name(self):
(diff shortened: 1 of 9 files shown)


fiftyone/operators/operator.py
```diff
@@ -141,6 +142,27 @@ def config(self):
...
+    def has_secrets(self):
+        """Whther this operator has any secrets registered."""
+        return bool(self._secrets)
+
+    @property
+    def secrets(self):
+        """The list of secrets known to the operator, or None."""
+        return self._secrets
+
+    def register_secrets(self, secrets):
```
kaixi-wang, line 155: doesn't this allow registering secrets that are not in the plugin definition?

kaixi-wang (changes requested):

fiftyone/plugins/secrets.py
```diff
@@ -103,123 +103,17 @@ def _get_secrets_client():
...
-            else:
-                self._resolver = None
-
-            self.__frozen = True
-        else:
-            self.__required_keys = []
-            self.__secrets = {}
-            self._operator_uri = None
-            self._resolver = None
+class SecretsDictionary(Mapping):
```
kaixi-wang, line 106: Why are you removing this?!?

fiftyone/operators/operator.py
kaixi-wang: The current implementation doesn't facilitate users to register unlisted secrets

brimoor: All I did was rename the `add_secrets()` method to `register_secrets()`

ritch (commented):
needs rebasing or closing

fiftyone/plugins/secrets.py
```diff
@@ -26,72 +29,68 @@ def __new__(cls):
...
         if not self._instance:
             self._instance = self.__new__(self.__class__)
         return self._instance.client
 
-    async def get_secret(
-        self, key: str, operator_uri: str, **kwargs
-    ) -> Optional[fois.ISecret]:
-        """
-        Get the value of a secret.
+    async def get_secret(self, key, operator_uri, **kwargs):
```
kaixi-wang, line 40: Why are we removing type annotations? This feels like it's just creating more tech debt.

PR #3843, opened 2023-11-28
Lightning tweaks

benjaminpkane:
### Changes

* Adds support for creating global wildcard indexes in the SDK `ds.create_index("$**")` and `ds.create_index("frames.$**")`
* Adds lightning threshold input in Options popout (grid only)
* Adds documentation links for lightning mode (separate PR) and sidebar mode in the Options popout (see recording)
* Updates sidebar mode resolution to match [documentation](https://docs.voxel51.com/user_guide/app.html#sidebar-mode) (best, fast, all) 
* Boolean filter and counts fixes
* Adds `No results` below `see more...` when no granular fields have data (see recording)

https://github.com/voxel51/fiftyone/assets/19821840/21359007-2e55-4e05-9b75-735ec7554dcf


app/packages/state/src/recoil/sidebar.ts
```diff
@@ -101,12 +101,12 @@ export const readableTags = selectorFamily<
 
       return get(
         group === "label tags"
-          ? aggregationAtoms.cumulativeValues({
+          ? cumulativeValues({
```
benjaminpkane, line 104: import fixes

manivoxel51 (approved):
nice work! 🍨 

two nit suggestions

- Changing dataset currently keeps the threshold of the last dataset. Should we clear it? since thresholds are really dataset-specific depending on sample size.

- It would be useful if user could "clear" the currently selected threshold. one idea is to have a clear button with lightning icon + "clear threshold" tooltip right next to the Settings gear icon

manivoxel51:
one more suggestion came from TT is to add a clear button to the threshold input

lanzhenw (dismissed):
lgtm

app/packages/state/src/recoil/aggregations.ts
```diff
@@ -131,10 +132,20 @@ export const aggregation = selectorFamily({
...
+          const n = get(schemaAtoms.isNumericField(p));
+          return numeric ? n : !n;
+        });
+      }
+
       return get(
         aggregations({
           ...params,
-          paths: get(schemaAtoms.filterFields(path)),
+          mixed: params.modal && numeric,
```
lanzhenw, line 147: 👍🏼

lanzhenw (approved):
Code looks good. 🚀

lanzhenw (approved):

PR #3846, opened 2023-11-28
Lightning Documentation

benjaminpkane:
Adds new **Lightning mode** section to the App user guide. Also fixes the `see more...` icon when lightning mode is not unlocked, keypoint labels, and filtering by `id`

### Keypoints

```py
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("coco-2017", split="validation").clone()

dataset.default_skeleton = model.skeleton

view = (
    dataset
    .match_labels(fields="ground_truth", filter=F("label") == "person")
    .limit(2)
)

# Add some custom attributes
for sample in view:
    for keypoint in sample["predictions_keypoints"].keypoints:
        keypoint["occluded"] = [np.random.randn() > 0 for _ in range(len(keypoint.points))]
        keypoint["age"] = np.random.randint(51)

    sample.save()
```
Screenshot 2023-12-01 at 3 11 04 PM

### DIFF
--- a/app/packages/components/src/components/TabOption/TabOption.tsx
+++ b/app/packages/components/src/components/TabOption/TabOption.tsx
@@ -74,7 +74,7 @@ const TabOption = ({
 
   return (
     <TabOptionDiv
-      style={style}
+      style={{ ...style, minWidth: `${6 * options.length || 1}rem` }}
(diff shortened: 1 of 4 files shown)


manivoxel51 (approved):
lgtm! 🍨

PR #3847, opened 2023-11-29
colorscheme bugfix

lanzhenw:

- implemented design changes: add an info icon to named colorscale that links to guidelines
- fixed the following bugs:

- [x] On global setting tab, toggling between name/list, and modify values, colorscale jumps back to use previous value. 

- [x] If `dataset.app_config.colorscheme` is `null`, colorscheme cannot be saved to DB successfully.

- [x] `setDatasetColorScheme` cannot update the linked record successfully.

tested locally

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/ColorModal/ColorFooter.tsx
+++ b/app/packages/core/src/components/ColorModal/ColorFooter.tsx
@@ -4,7 +4,6 @@ import * as fos from "@fiftyone/state";
 import React, { useEffect, useMemo } from "react";
 import { useMutation } from "react-relay";
 import { useRecoilState, useRecoilValue } from "recoil";
-import { v4 as uuid } from "uuid";
 import { ButtonGroup, ModalActionButtonContainer } from "./ShareStyledDiv";
 import { activeColorEntry } from "./state";
 
@@ -73,7 +72,7 @@ const ColorFooter: React.FC = () => {
                 datasetName,
                 colorScheme: {
                   ...colorScheme,
-                  id: datasetDefault?.id ?? colorScheme.id ?? uuid(),
+                  id: datasetDefault?.id ?? null,
                   colorBy: colorScheme.colorBy ?? "field",
                   colorPool: colorScheme.colorPool ?? [],
                   colorscales: colorScheme.colorscales ? newColorscales : [],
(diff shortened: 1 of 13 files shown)


app/packages/core/src/components/ColorModal/colorPalette/Colorscale.tsx
```diff
@@ -217,7 +228,20 @@ const Colorscale: React.FC = () => {
           />
           {tab === "name" && (
             <div>
-              Use a named plotly colorscale:
+              <Guide>
+                Use a named colorscale
+                <a
+                  href="https://plotly.com/python/colorscales/"
```
benjaminpkane, line 234: Maybe move to ./utils/links.ts?

benjaminpkane (approved):
LGTM!

app/packages/core/src/components/ColorModal/ColorFooter.tsx
```diff
@@ -100,7 +104,7 @@ const ColorFooter: React.FC = () => {
                     );
 
                   const record = store.get(fragment.id);
-                  record && config?.setLinkedRecord(record, "colorScheme");
+                  record && config!.setLinkedRecord(record, "colorScheme");
```
manivoxel51, line 107: does this mean we want to ignore the fact that `config` can be undefined? maybe we should return on line 97

app/packages/core/src/components/ColorModal/colorPalette/Colorscale.tsx
```diff
@@ -233,7 +257,7 @@ const Colorscale: React.FC = () => {
           )}
           {tab === "list" && (
             <div>
-              Define a custom colorscale:
+              Define a custom colorscale (range between 0 and 1):
```
manivoxel51, line 260: 👍

manivoxel51 (dismissed):
Nice work! 🍨 

tested locally. left two inline comments, 

one UX bug (unrelated to this PR but we should probably address)

- The manual input in the color selector disappears after entering 3 characters and seems to autocomplete the input. Can we have it stay open for the whole string?

https://github.com/voxel51/fiftyone/assets/109545780/f6b691dc-20e8-4a5b-8416-22f96dfa6040

UX issue 

- the tooltip on `select an attribute to color` dropdown is blocking the selection option. Can we have the tooltip open on the top instead

nit feedback - we can discuss offline

- I believe having a multi select dropdown for the color by value input would be useful. I personally always forget if the value was capitalized or not and what label values are available.

app/packages/core/src/components/ColorModal/ColorFooter.tsx
lanzhenw: `config` cannot be undefined. I will add a return.

benjaminpkane (approved):

PR #3848, opened 2023-11-29
snapshot archive docs

imanjra:

Documentation for new snapshot archive/unarchive flows in UI

N/A

        notes for FiftyOne users.

-   [ ] Other


docs/source/teams/dataset_versioning.rst
```diff
@@ -3,6 +3,8 @@
 Dataset Versioning
 ==================
 
+.. versionadded:: Teams 1.4.0
```
brimoor, line 6: Oh does this work??? I vaguely remember not being able to use this command in the past

brimoor (approved):
LGTM, thank you good sirs 🙇

docs/source/teams/dataset_versioning.rst
imanjra: Cool! I am going note this down. I had seen this somewhere awhile back and was looking for it just couple days but couldn't remember what it was called. 🥇

brimoor: So it does render, but it doesn't get any kind of special CSS, just plain text. Which is annoying 😭 

I decided to remove it on https://github.com/voxel51/fiftyone/pull/3870 since we don't have a precedent for marking version added anywhere else in the docs yet.

swheaton: Agree it doesn't look the best but that can be changed.

> I decided to remove it on https://github.com/voxel51/fiftyone/pull/3870 since we don't have a precedent for marking version added anywhere else in the docs yet.

Disagree with this logic - I think it is a useful thing to have in docs and if you don't start setting a precedent then there won't be a precedent. It's easier to do it now on release than try to go back and remember when each feature was added or changed to retroactively insert

swheaton: image

Mucked around with adapting CSS from another sphinx theme I found and got these examples.
Not a stylist but this shows one possibility.

PR #3850, opened 2023-11-29
add ExecutionOptions to execute_operator()

brimoor:
As currently implemented, if a user passes `request_delegation=True` but the operator doesn't support delegated execution, it will log a warning and just execute immediately instead:

```py
foo.execute_operator(operator_uri, ctx=ctx, request_delegation=True)
# This operation does not support delegated execution; it will be executed immediately
```

We could decide to raise an exception instead though.

### DIFF
--- a/fiftyone/operators/executor.py
+++ b/fiftyone/operators/executor.py
@@ -12,8 +12,6 @@ import inspect
 import logging
 import os
 import traceback
-import types as python_types
-import typing
 
 import fiftyone as fo
 import fiftyone.core.dataset as fod
@@ -26,6 +24,9 @@ from fiftyone.plugins.secrets import PluginSecretsResolver, SecretsDictionary
 from .decorators import coroutine_timeout
 from .registry import OperatorRegistry
 from .message import GeneratedMessage, MessageType
+
+
+logger = logging.getLogger(__name__)
 
 
 class ExecutionRunState(object):
@@ -56,16 +57,15 @@ class InvocationRequest(object):
         }
 
 
-class ExecutionProgress:
+class ExecutionProgress(object):
     """Represents the status of an operator execution.
 
-    at least one of progress or label must be provided
     Args:
         progress (None): an optional float between 0 and 1 (0% to 100%)
         label (None): an optional label to display
     """
 
-    def __init__(self, progress: float = None, label: str = None):
+    def __init__(self, progress=None, label=None):
         self.progress = progress
         self.label = label
         self.updated_at = None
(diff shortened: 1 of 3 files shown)


fiftyone/operators/executor.py
```diff
@@ -516,24 +513,26 @@ def current_sample(self):
 
     @property
     def delegated(self):
-        """Whether the operation's execution was delegated to an orchestrator.
-
-        This property is only available for methods that are invoked after an
-        operator is executed, e.g. :meth:`resolve_output`.
-        """
+        """Whether the operation's execution was delegated to an orchestrator."""
```
ritch, line 516: Now this doc is incorrect

fiftyone/operators/executor.py
```diff
@@ -157,14 +151,13 @@ def execute_operator(operator_uri, ctx, params=None):
...
-    view = ctx.get("view", None)
-    selected = ctx.get("selected", None)
-    selected_labels = ctx.get("selected_labels", None)
+def _parse_ctx(ctx=None, **kwargs):
+    if ctx is None:
+        ctx = {}
 
-    if params is None:
-        params = ctx.get("params", {})
+    ctx = {**ctx, **kwargs}  # don't modify input `ctx` in-place
```
brimoor, line 158: @ritch this is the other key change: it allows for `request_delegation` and `delegation_target` to be correctly passed through to the `ExecutionContext` using either of these syntaxes:

```py
foo.execute_operator(..., ctx={"request_delegation": True, ...})
foo.execute_operator(..., request_delegation=True)
```

fiftyone/operators/operator.py
```diff
@@ -176,43 +182,48 @@ def resolve_definition(self, resolve_dynamic, ctx):
...
         Returns:
             a :class:`fiftyone.operators.executor.ExecutionOptions` instance
         """
         from .executor import ExecutionOptions
 
+        # pylint: disable=assignment-from-none
         resolved_delegation = self.resolve_delegation(ctx)
-        if resolved_delegation is None:
-            # legacy behavior
+        if resolved_delegation is not None:
```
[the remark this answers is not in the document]
ritch: Excellent 👍

fiftyone/operators/executor.py
ritch: Great - and I'm assuming this is backwards compatible. Doesn't appear to be but I would be surprised if you missed that.

ritch (dismissed):

fiftyone/operators/executor.py
brimoor: yep, backwards compatible 👍

PR #3851, opened 2023-11-29
InferredView fixes and enhancements

imanjra:

Fix exceptions for unknown value type and ensure defaults are propagated correctly

Using an operator with InferredView rendering a deeply nested object

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/plugins/SchemaIO/utils/generate-schema.ts
+++ b/app/packages/core/src/plugins/SchemaIO/utils/generate-schema.ts
@@ -1,11 +1,11 @@
 // Auto generate SchemaIO compatible schema based on a value
 
-import { isPrimitiveType } from "@fiftyone/utilities";
+import { isNullish, isPrimitiveType } from "@fiftyone/utilities";
 
 // todo: add support for OneOfView, TupleView, and MapView
-export function generateSchema(value, options?) {
+export function generateSchema(value: any, options?: GenerateSchemaOptions) {
   const type = getType(value);
-  const { label, readOnly } = options;
+  const { label, readOnly } = options || {};
   if (type === "array") {
     const dominantType = getDominantType(value);
     const isValueTable = isTable(value);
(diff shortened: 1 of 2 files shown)


app/packages/core/src/plugins/SchemaIO/utils/generate-schema.ts
```diff
@@ -16,10 +16,9 @@ export function generateSchema(value, options?) {
           label,
           component: "TableView",
           columns: getTableColumns(value),
-          readOnly: true,
         },
       };
-    } else if (isPrimitiveType(dominantType) && readOnly) {
+    } else if (isPrimitiveType(dominantType || "") && readOnly) {
```
ritch, line 21: This is probably the crux of the fix, but I find it very hard to follow. Its odd that `getDominantType()` returns null(ish) values? A variable name or comment indicating that the `||` is due to it being an empty array (isn't that the only case where this happens?

Also I realize we're using existing utils here, but I'd rather have metadata be in object form than functions, the data is already there so its a bit odd to call functions to get at it:

```typescript
const value = "foo";
const typeInfo = getTypeInfoForValue(value);
// typeInfo.isArray => false
// typeInfo.dominantType => null (rename to typeInfo.arrayItemDominantType)
// typeInfo.isPrimitive => true
// typeInfo.value => "foo"
// typeInfo.isNullish => false
// typeInfo.type => string
// etc
```

Both of these a minor comments, but thought they might be useful feedback.

imanjra: Yah, that's right. dominantType will be unknown for an empty array. Planning to use something like one of view in future for unknown type but it can potentially be a nested object.

Not too sure about the second half of the comment. The `generate-schema` (I should probably rename to `inferSchemaUsingValue(value)` is only used for `InferredView` to generate a `SchemaIO` compatible schema based only on a value.

PR #3852, opened 2023-11-29
selectGroupSlices stage should 🏃 before Select samples stage

manivoxel51:
### problem
The Select samples stage is running before selectGroupSlices and causing the sample modal to crash (not open).

### solution
Make sure selectGroupSlices stage is computed before Select samples stage when making optimized view function is called.

#### Before

https://github.com/voxel51/fiftyone/assets/109545780/5e8f12d0-985e-47a4-bf79-19599883a7e8

#### After

https://github.com/voxel51/fiftyone/assets/109545780/586f6c23-b25d-418c-b767-3b70b01915c2

- unit test making sure select group slices is computed before Select stage
- test script 

```
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.select_group_slices(["left", "right"])

session = fo.launch_app(view)
```

- Try clicking on the second sample (right slice) and it should open modal
- Try tagging the second sample or its labels and it should work
- Try tagging the first sample (left slice) and it should work

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Actions/utils.tsx
+++ b/app/packages/core/src/components/Actions/utils.tsx
@@ -77,6 +77,7 @@ export const tagStatistics = selectorFamily<
               ? {
                   id: modal ? get(groupId) : null,
                   slices: get(fos.currentSlices(modal)),
+                  slice: get(fos.currentSlice(modal)),
                   mode: get(groupStatistics(modal)),
                 }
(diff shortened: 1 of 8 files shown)


benjaminpkane (approved):
LGTM! 👏

PR #3855, opened 2023-11-30
Fix Take seed in App

benjaminpkane:
Fixes Take seed changes in the App.

### Changes
* Fix comma handling
* Process view changes from the bar and exclude `_rand` kwarg when seed changes

https://github.com/voxel51/fiftyone/assets/19821840/2692b77e-090e-43ab-bb75-83ebd2f6cb83

https://github.com/voxel51/fiftyone/assets/19821840/fcf5e003-2253-465a-9106-0812d3efa256


manivoxel51 (approved):
thank you! 🍨

PR #3863, opened 2023-12-01
Bugfix - video as frames string cast

sashankaryal:

This fixes a critical bug in frames/videos feature where `totalFramesCount` was not being incorrectly set when group by video keys were not by default string values.

Pending E2E tests. Local smoke testing.

```
import fiftyone as fo
import fiftyone.zoo as foz

i = 0
for sample in dataset.iter_samples(autosave=True, progress=True):
    sample["num"] = i // 10
    sample["num2"] = i % 10
    i += 1
```


app/packages/looker/src/elements/imavid/frame-count.ts
```diff
@@ -12,7 +12,9 @@ export class FrameCountElement extends BaseElement<ImaVidState> {
   }
 
   renderSelf({ currentFrameNumber, config }: Readonly<ImaVidState>) {
-    this.element.innerHTML = `${currentFrameNumber} / ${config.frameStoreController.totalFrameCount}`;
+    this.element.innerHTML = `${currentFrameNumber} / ${
+      config.frameStoreController.totalFrameCount || 1
```
benjaminpkane, line 16: Can `config.frameStoreController.totalFrameCount` be undefined?

benjaminpkane (approved):
LGTM 👍

PR #3864, opened 2023-12-01
Bugfix Orthographic projection

sashankaryal:

This PR fixes a bug where orthographic projections weren't showing because `sampleMediaFilePath` was being set after sample config was already instantiated.

Screenshot 2023-12-01 at 10 18 31 PM


benjaminpkane (approved):
Tested! LGTM!

PR #3866, opened 2023-12-01
regression fix: RGB segmentation corruption

lanzhenw:
In `painter.ts`, commented out `targets[i]=r`, as this corrupts rgb masks. 
```
 if (isRgbMaskTargets_) {
          targets[i] = (maskTargets as RgbMaskTargets)[
            currentHexCode
          ].intTarget;
        } else {
          // assign an arbitrary uint8 value here; this isn't used anywhere but absence of it affects tooltip behavior
          // having this line caused the mask to corrupt
          **// targets[i] = r;**
        }
```

After the fix: 

https://github.com/voxel51/fiftyone/assets/17770824/c5e449c4-24df-44fc-9506-73860f11eb51

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/looker/src/worker/painter.ts
+++ b/app/packages/looker/src/worker/painter.ts
@@ -279,7 +279,8 @@ export const PainterFactory = (requestColor) => ({
           ].intTarget;
         } else {
           // assign an arbitrary uint8 value here; this isn't used anywhere but absence of it affects tooltip behavior
-          targets[i] = r;
+          // having this line caused the mask to corrupt
+          // targets[i] = r;
         }
       }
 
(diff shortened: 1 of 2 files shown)


app/packages/looker/src/worker/painter.ts
```diff
@@ -254,6 +254,7 @@ export const PainterFactory = (requestColor) => ({
     const isRgbMaskTargets_ = isRgbMaskTargets(maskTargets);
 
     if (maskData.channels > 2) {
+      console.log("rendering rgb mask");
```
sashankaryal: let's remove this `console.log`

sashankaryal (commented):
Nice! Will wait for the tests to pass and approve it. Got to catch an incidental bug that was way more nefarious because of the corruption 😜

sashankaryal (approved):
lgtm

PR #3868, opened 2023-12-04
Bugfix dynamic groups

sashankaryal:

Fixes a bug introduced by #3863 that was causing dynamic groups tests to fail.

All E2E now green.

### DIFF
--- a/app/packages/core/src/components/Actions/Options.tsx
+++ b/app/packages/core/src/components/Actions/Options.tsx
@@ -281,12 +281,7 @@ const Options = ({ modal, anchorRef }: OptionsProps) => {
   const view = useRecoilValue(fos.view);
 
   return (
-    <Popout
-      style={{ minWidth: "250px" }}
-      modal={modal}
-      fixed
-      anchorRef={anchorRef}
-    >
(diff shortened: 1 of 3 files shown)


app/packages/core/src/components/Actions/Options.tsx
```diff
@@ -281,7 +281,12 @@ const Options = ({ modal, anchorRef }: OptionsProps) => {
   const view = useRecoilValue(fos.view);
 
   return (
-    <Popout modal={modal} fixed anchorRef={anchorRef}>
+    <Popout
+      style={{ minWidth: "250px" }}
```
benjaminpkane, line 285: Is there a reason the width can't be auto?

lanzhenw (approved):
locally tested, it fixed the dynamic group carousel and pagination bug.

app/packages/core/src/components/Actions/Options.tsx
sashankaryal: Was being lazy. Changed it in `TabOption` in the next commit

lanzhenw (approved):

PR #3869, opened 2023-12-04
ImaVid zoom/pan, loop, and disabled overlays

benjaminpkane:
Enhancements PR for consistency with `VideoLooker`. Very easy to add these updates 🚀 

### Changes

* Adds support for `loop` setting
* Adds zoom/pan (mousewheel) and crop to content (z)
* Disable overlay selection and tooltips during playback

The loop behavior sometimes misaligns when returning to first frame. Possibly could improve. I also removed unused variables/properties but am happy to revert those changes if we want them for future work


app/packages/looker/src/elements/imavid/index.ts
```diff
@@ -276,14 +269,29 @@ export class ImaVidElement extends BaseElement<ImaVidState, HTMLImageElement> {
...
+          requestAnimationFrame(() => {
+            const next = frameNumberToDraw + 1;
+
+            if (next > this.framesController.totalFrameCount) {
+              this.update(({ options: { loop } }) => {
+                if (loop) {
+                  this.drawFrame(1);
+                  return {
+                    playing: true,
+                    currentFrameNumber: 1,
```
benjaminpkane, line 281: @sashankaryal could this be an arbitrary value? Is "first" frame key that should be used?

app/packages/looker/src/elements/common/canvas.ts
```diff
@@ -123,6 +123,7 @@ export class CanvasElement<State extends BaseState> extends BaseElement<
 
               const xs = (x - px) / scale;
               const ys = (y - py) / scale;
+              console.log(zoomPad);
```
sashankaryal, line 126: remove?

sashankaryal (approved):
Looking good! Thanks for the cleanup! 🚀

ritch (approved):

PR #3872, opened 2023-12-05
cloud creds docs

ritch:
Documenting Cloud Credentials through the UI, including bucket-prefixed credentials


swheaton (approved):
made some rearranging / copy / prettification edits but nothing major.

Did not do too much to the bucket prefixes section as I ran out of time, but it reads fine to me. Possibly a bit too verbose where a simple diagram would serve better.

PR #3874, opened 2023-12-05
fiftyone-db 1.0

benjaminpkane:
To not break existing installations, general support for Linux distributions should via a source distribution should be released a `fiftyone-db==1.0`


findtopher (approved):
This is sane and reasonable.

PR #3877, opened 2023-12-05
Always show lightning search input 

benjaminpkane:
A regression from #3846, lightning search input should always be shown. Also improves Options popout sizing.
Before
Screenshot 2023-12-05 at 2 01 55 PM

After
Screenshot 2023-12-05 at 2 01 40 PM

### DIFF
--- a/app/packages/components/src/components/TabOption/TabOption.tsx
+++ b/app/packages/components/src/components/TabOption/TabOption.tsx
@@ -27,6 +27,7 @@ const TabOptionDiv = animated(styled.div`
   & > div > div {
     display: flex;
     justify-content: center;
+    padding: 0 1rem;
   }
 `);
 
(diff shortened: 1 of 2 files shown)


benjaminpkane:
This was the cause of the regression
```
diff --git a/app/packages/core/src/components/Filters/StringFilter/useSelected.ts b/app/packages/core/src/components/Filters/StringFilter/useSelected.ts
index 4ca1321f9..552122780 100644
--- a/app/packages/core/src/components/Filters/StringFilter/useSelected.ts
+++ b/app/packages/core/src/components/Filters/StringFilter/useSelected.ts
@@ -50,7 +50,9 @@ export default function (
   const results =
     resultsLoadable.state === "hasValue" ? resultsLoadable.contents : null;
   const length = results?.results?.length ?? 0;
-  const shown = showSearch || length > CHECKBOX_LIMIT;
+  const shown =
+    resultsLoadable.state !== "loading" &&
+    (showSearch || length > CHECKBOX_LIMIT);
 
   return {
     results,
```

ritch (approved):

PR #3881, opened 2023-12-05
Allow 0.4 DB installs

benjaminpkane:


findtopher (approved):
I like backwards compatibility!

PR #3886, opened 2023-12-06
Fix imavid tagging

sashankaryal:

This PR fixes tagging for video-frames (imavid) feature.

Smoke test locally.


sashankaryal:
Subsumed by #3893

PR #3887, opened 2023-12-06
fix imavid regression in buffering indicator

sashankaryal:

Fixes imavid regression


benjaminpkane:
This fixes the regression I caused! My mistake. But `disableOverlays` is meant to hide the tooltip and and prevent label selection which is consistent with the video looker.

We could allow label selection and tooltips for both video style lookers. But the tooltip would need to update during playback event when the mouse isn't moving.

https://github.com/voxel51/fiftyone/assets/19821840/23db82ca-76b7-4ca9-a016-d83f1e6dfaeb

PR #3890, opened 2023-12-06
Fix `session.dataset = None`

benjaminpkane:
In 0.23.0, `session.dataset = None` does not return the App to the index page. Updates and adds unit tests to `resolveURL` to ensure this fix.

```py
import fiftyone as fo

# wait for App to load dataset...

# unset dataset
session.dataset = None
```
https://github.com/voxel51/fiftyone/assets/19821840/81599ee9-1ca9-4bc9-9cd8-787e6305634d


imanjra (approved):
Tested with several flows locally. LGTM

PR #3900, opened 2023-12-07
Fixing bug when inferring doubly-nested dynamic list field types

brimoor:
Fixes a very subtle issue when inferring dynamic schemas for doubly-nested list fields. The new unit test fails on `develop` but passes on this branch.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1249,8 +1249,9 @@ class SampleCollection(object):
             else:
                 fields = list(fields)
 
+        unwind_cache = []
         dynamic_schema = self._do_get_dynamic_field_schema(
-            schema, frames=frames, fields=fields
+            schema, unwind_cache, frames=frames, fields=fields
         )
 
         # Recurse into new dynamic fields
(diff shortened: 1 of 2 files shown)


fiftyone/core/collections.py
```diff
@@ -1322,6 +1331,12 @@ def _do_get_dynamic_field_schema(
                 if is_list_field and not _path.endswith("[]"):
                     _path += "[]"
 
+                    # Cache the manual unwind in case we need it recursively
+                    # Insert rather than append so that nested paths are found
+                    # first when using this cache
+                    _clean_path = _path.replace("[]", "")
+                    unwind_cache.insert(0, (_clean_path, _path))
+
```
swheaton, line 1339: I do not entirely understand what's going on here so just making sure it's intentional that this logic only applies when `_path.endswith("[]")` is `False`

brimoor: Yeah this is correct as written. The only way that `[]` gets into the path is inside this `if`, which is the time when we need to cache an undeclared list field that we've found.

PR #3902, opened 2023-12-07
Enable loading dataset via id for ExecutionContext.Dataset

kaixi-wang:

- load dataset by id rather than name if possible in case a dataset has been renamed (TEAMS-2042)
- pass dataset_id in request params when delegated/queued operation executes
- don't call `load_dataset` to load the dataset without triggering migration and logging last loaded at
- make sure the ctx.view is a view of ctx.dataset

Tested locally:
1. Schedule a delegated operation to run (without launching delegation)
2. renaming dataset
3. launching local delegation
4. verify run starts/completes without error due to missing dataset 
5. verify delegated_op doc in db shows the previous/no longer existing dataset name

Also tested rerunning a failed operation after renaming.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -262,8 +262,11 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         overwrite=False,
         _create=True,
         _virtual=False,
+        _doc=None,
         **kwargs,
     ):
+        if _doc:
+            name = _doc.name
         if name is None and _create:
             name = get_default_dataset_name()
 
@@ -276,7 +279,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
             )
         else:
             doc, sample_doc_cls, frame_doc_cls = _load_dataset(
-                self, name, virtual=_virtual
+                self, name, virtual=_virtual, doc=_doc
             )
 
         self._doc = doc
@@ -6808,11 +6811,13 @@ def _load_clips_source_dataset(frame_collection_name):
     return load_dataset(doc["name"])
 
 
-def _load_dataset(obj, name, virtual=False):
+def _load_dataset(obj, name, virtual=False, doc=None):
     if not virtual:
         fomi.migrate_dataset_if_necessary(name)
 
     try:
+        if doc:
+            return _do_load_dataset_from_doc(obj, doc)
         return _do_load_dataset(obj, name)
     except Exception as e:
         try:
@@ -6836,6 +6841,13 @@ def _do_load_dataset(obj, name):
         dataset_doc = foo.DatasetDocument.objects.get(name=name)
     except moe.DoesNotExist:
         raise ValueError("Dataset '%s' not found" % name)
+
+    return _do_load_dataset_from_doc(obj, dataset_doc)
+
+
+def _do_load_dataset_from_doc(obj, dataset_doc):
+    if isinstance(dataset_doc, dict):
+        dataset_doc = foo.DatasetDocument(**dataset_doc)
 
     sample_collection_name = dataset_doc.sample_collection_name
     frame_collection_name = dataset_doc.frame_collection_name
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -31,6 +31,9 @@ import timeit
 import types
 from xml.parsers.expat import ExpatError
 import zlib
+
+from bson import ObjectId
+from bson.errors import InvalidId
 from matplotlib import colors as mcolors
 from concurrent.futures import ThreadPoolExecutor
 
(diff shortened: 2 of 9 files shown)


fiftyone/operators/executor.py
```diff
@@ -468,13 +494,20 @@ def dataset_id(self):
...
-        _ = self.dataset
+        # Always derive the view from the context's dataset
+        dataset = self.dataset
+        if not dataset:
+            raise RuntimeError(
+                "Cannot create a view without setting the dataset in the context first."
+            )
 
-        dataset_name = self.request_params.get("dataset_name", None)
+        dataset_name = dataset.name
```
ritch, line 510: One optimization (not sure if worthwhile) would be to modify `fosv.get_view()` to take `dataset` as a param, since its calling `load_dataset()` internally redundantly.

fiftyone/operators/executor.py
```diff
@@ -443,11 +446,34 @@ def dataset(self):
...
+            dataset_name = self.request_params.get("dataset_name", None)
+        else:
+            # set the dataset_name in the context to the name resolved from
+            # the db so downstream calls to load_dataset will work
+            self.request_params["dataset_name"] = dataset_name
+
+        if dataset_name:
+            # loading the dataset like this prevents triggering db
+            # migration check and logging last_loaded_at
+            self._dataset = fod.Dataset(
```
brimoor, line 474: Why are you preventing migrations/last loaded updates? This should just `load_dataset()` like any other usage of FO does

fiftyone/operators/executor.py
```diff
@@ -443,11 +446,34 @@ def dataset(self):
...
             return self._dataset
-
-        dataset_name = self.request_params.get("dataset_name", None)
-
-        self._dataset = fod.load_dataset(dataset_name)
-        self._dataset.reload()
+        dataset_name = None
+        # since the dataset may have been renamed, if we have the
+        # dataset_id, get the current dataset_name from the db
+        if self.request_params.get("dataset_id", None) is not None:
```
brimoor, line 452: How about a utility function like `load_dataset(id=None, name=None)` that loads the dataset by ID if provided, else name?

The idea of loading a dataset by ID is likely useful elsewhere

fiftyone/operators/executor.py
```diff
@@ -468,13 +494,20 @@ def dataset_id(self):
...
+        # since the context.dataset may have been changed, we should only
+        # return the current view if the current DatasetView's root dataset
+        # matches the dataset in the context
+        if self._view and self._view._root_dataset.name == self.dataset.name:
             return self._view
 
-        # Forces a dataset reload if necessary
-        _ = self.dataset
+        # Always derive the view from the context's dataset
+        dataset = self.dataset
```
brimoor, line 504: You could avoid needing any changes or complex logic in the `view` property if you just update `request_params["dataset_name"]` within the `dataset` property, if the name changed.

Line 504 ensures that `dataset` is accessed before `get_view()` is called

brimoor (changes requested):

fiftyone/operators/executor.py
kaixi-wang: Calling 'load_dataset' leads to excessive/noisy logging and preventing migrations avoids issues of hanging executions due to accidental/automatic db migrations/migration failures happening without direct user interaction or in a separate process.. To execute/queue the operation, presumably the user must have loaded the dataset at some point, which is when the logging/migration should happen.

brimoor: There is nothing special about delegated execution; it should use the same public interface as everything else. Maybe the task has been queued for 1 day and an admin upgraded the deployment in that time and the orchestrator is now running a newer SDK version. The dataset needs to be migrated

fiftyone/operators/executor.py
kaixi-wang: that is intentional to ensure the view matches the dataset... it follows the pattern the app uses to keep the state.view compatible with state.dataset

brimoor: In this class, `self.dataset` does the work of ensuring that the dataset is properly reloaded (including loading it under it's current name, if it has changed), so the only change needed in this function is to do this:

```py
dataset_name = self.dataset.name
```

instead of referencing `request_params["dataset_name"]` as was previously done (your idea to use `dataset.name` is good here 👍)

fiftyone/core/dataset.py
```diff
@@ -262,8 +262,11 @@ def __init__(
         overwrite=False,
         _create=True,
         _virtual=False,
+        _doc=None,
```
kaixi-wang, line 265: it seemed redundant to query the database to get the name to just then query the same document using the name, so I thought it would be nice to be able to init the dataset with the dataset doc just read from the db

fiftyone/core/utils.py
```diff
@@ -1992,3 +1995,26 @@ def validate_hex_color(value):
...
 fos = lazy_import("fiftyone.core.storage")
+
+
+def load_dataset(identifier):
+    """Loads the dataset from the database with the given name or id."""
+    import fiftyone.core.odm as foo
+    import fiftyone.core.dataset as fod
+
+    db = foo.get_db_conn()
+    try:
```
kaixi-wang, line 2006: querying using id or name, but not both because it's possible after the dataset is renamed, the previous name is used for a different dataset

PR #3908, opened 2023-12-09
Add get method to secretsdict 

kaixi-wang:

Add get method for backwards compat


brimoor (approved):
LGTM

PR #3909, opened 2023-12-11
V7 doc updates

allenleetc:
Update v7 integration docs per current state


allenleetc:
There may be small tweaks if there is feedback from v7 but I don't expect anything significant

brimoor (approved):

PR #3912, opened 2023-12-11
bugfix: similarity search only works on default_group_slice

lanzhenw:

Pass in the active session group slice as a parameter for similarity sort requests to update samples
Fixes a bug: similarity search only works on default_group_slice

https://github.com/voxel51/fiftyone/assets/17770824/264345d7-63cf-47e7-9fed-f20a942b224e

```
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.brain as fob

ds = foz.load_zoo_dataset('quickstart-groups',max_samples=50)
print(ds.default_group_slice) # prints 'left'
view = ds.select_group_slices('right')
fob.compute_similarity(view,model='clip-vit-base32-torch',brain_key='sim_right')

# now Open Dataset in FOT App. Select 'right' group slice. 
# Select an image in the grid or type in "mountian/road/anything" in similarity search prompt to run an Image Sim Search. 

# this should be working as expected
```

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/state/src/recoil/selectors.ts
+++ b/app/packages/state/src/recoil/selectors.ts
@@ -8,14 +8,14 @@ import {
   datasetFragment$key,
   graphQLSyncFragmentAtom,
 } from "@fiftyone/relay";
-import { fieldVisibilityStage } from "@fiftyone/state";
+import { currentSlice, fieldVisibilityStage, isGroup } from "@fiftyone/state";
 import { toSnakeCase } from "@fiftyone/utilities";
 import { DefaultValue, atomFamily, selector, selectorFamily } from "recoil";
 import { v4 as uuid } from "uuid";
 import * as atoms from "./atoms";
 import { config } from "./config";
 import { dataset as datasetAtom } from "./dataset";
-import { currentModalSample, modalSample } from "./modal";
+import { currentModalSample, isModalActive, modalSample } from "./modal";
 import { pathFilter } from "./pathFilters";
 import { State } from "./types";
 import { isPatchesView } from "./view";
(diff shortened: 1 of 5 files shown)


benjaminpkane (dismissed):
LGTM 🚀

manivoxel51 (commented):
@lanzhenw this is great work! 🥇 

I see three issues here and I think you have fixed the main one but two pending.

1. when a current active slice doesn't have a similarity computed, we still show the "Type anything" input. This causes confusion as the user can type "car" for example and the loading spinner will spin forever and a network error occurs without being surfaced to the client. The reason is that current slice is "left" for example but only "right" slice has similarity computed. 

I think I know a fix, will push something for it.

2. If you compute similarity for "right slice". then search for "car". then switch back to left slice, app crashes.

@lanzhenw Can you look into this when you get a chance. end of video shows the crash

Video is app after your fix and my fix (1) above

https://github.com/voxel51/fiftyone/assets/109545780/54d170e8-eaa6-4682-ab8b-11a78739399d

app/packages/state/src/recoil/selectors.ts
```diff
@@ -384,7 +384,24 @@ export const similarityMethods = selector<{
...
+          if (_cls === "fiftyone.core.stages.SelectGroupSlices") {
+            const sliceValue = kwargs.filter(
+              (kwarg: string[]) => kwarg[0] === "slices"
+            )?.[0]?.[1];
+            if (sliceValue && sliceValue !== activeSlice) return false;
+          }
+          return true;
+        });
+      });
+    }
```
manivoxel51, line 404: @benjaminpkane @lanzhenw added this filters out a method if the activeSlice is not the one one with the similarity computed. 

Shows help message for the active slice without similarity computed

https://github.com/voxel51/fiftyone/assets/109545780/bb2100b0-68f1-41b7-992d-80c72b444f71

manivoxel51 (approved):
Thank you! 🍨

PR #3920, opened 2023-12-12
Load datasets by ID when executing delegated operations

kaixi-wang:

- load dataset by id in the execution context (particularly for delegated operations)  if possible in case a dataset has been renamed (TEAMS-2042)
- pass dataset_id in request params when delegated/queued operation executes
- make sure the ctx.view is a view of ctx.dataset

Tested locally:
1. Schedule a delegated operation to run (without launching delegation)
2. renaming dataset
3. launching local delegation
4. verify run starts/completes without error due to missing dataset 
5. verify delegated_op doc in db shows the previous/no longer existing dataset name

Also tested rerunning a failed operation after renaming.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -276,9 +276,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
             )
         else:
             doc, sample_doc_cls, frame_doc_cls = _load_dataset(
-                self,
-                name,
-                virtual=_virtual,
+                self, name, virtual=_virtual
             )
 
         self._doc = doc
@@ -6810,11 +6808,7 @@ def _load_clips_source_dataset(frame_collection_name):
     return load_dataset(doc["name"])
 
 
-def _load_dataset(
-    obj,
-    name,
-    virtual=False,
-):
+def _load_dataset(obj, name, virtual=False):
     if not virtual:
         fomi.migrate_dataset_if_necessary(name)
 
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -262,11 +262,8 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         overwrite=False,
         _create=True,
         _virtual=False,
-        _doc=None,
         **kwargs,
     ):
-        if _doc:
-            name = _doc.name
         if name is None and _create:
             name = get_default_dataset_name()
 
@@ -279,7 +276,9 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
             )
         else:
             doc, sample_doc_cls, frame_doc_cls = _load_dataset(
-                self, name, virtual=_virtual, doc=_doc
+                self,
+                name,
+                virtual=_virtual,
             )
 
         self._doc = doc
@@ -6811,13 +6810,15 @@ def _load_clips_source_dataset(frame_collection_name):
     return load_dataset(doc["name"])
 
 
-def _load_dataset(obj, name, virtual=False, doc=None):
+def _load_dataset(
+    obj,
+    name,
+    virtual=False,
+):
     if not virtual:
         fomi.migrate_dataset_if_necessary(name)
 
     try:
-        if doc:
-            return _do_load_dataset_from_doc(obj, doc)
         return _do_load_dataset(obj, name)
     except Exception as e:
         try:
@@ -6841,13 +6842,6 @@ def _do_load_dataset(obj, name):
         dataset_doc = foo.DatasetDocument.objects.get(name=name)
     except moe.DoesNotExist:
         raise ValueError("Dataset '%s' not found" % name)
-
-    return _do_load_dataset_from_doc(obj, dataset_doc)
-
-
-def _do_load_dataset_from_doc(obj, dataset_doc):
-    if isinstance(dataset_doc, dict):
-        dataset_doc = foo.DatasetDocument(**dataset_doc)
 
     sample_collection_name = dataset_doc.sample_collection_name
     frame_collection_name = dataset_doc.frame_collection_name
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -262,8 +262,11 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         overwrite=False,
         _create=True,
         _virtual=False,
+        _doc=None,
         **kwargs,
     ):
+        if _doc:
+            name = _doc.name
         if name is None and _create:
             name = get_default_dataset_name()
 
@@ -276,7 +279,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
             )
         else:
             doc, sample_doc_cls, frame_doc_cls = _load_dataset(
-                self, name, virtual=_virtual
+                self, name, virtual=_virtual, doc=_doc
             )
 
         self._doc = doc
@@ -6808,11 +6811,13 @@ def _load_clips_source_dataset(frame_collection_name):
     return load_dataset(doc["name"])
 
 
-def _load_dataset(obj, name, virtual=False):
+def _load_dataset(obj, name, virtual=False, doc=None):
     if not virtual:
         fomi.migrate_dataset_if_necessary(name)
 
     try:
+        if doc:
+            return _do_load_dataset_from_doc(obj, doc)
         return _do_load_dataset(obj, name)
     except Exception as e:
         try:
@@ -6836,6 +6841,13 @@ def _do_load_dataset(obj, name):
         dataset_doc = foo.DatasetDocument.objects.get(name=name)
     except moe.DoesNotExist:
         raise ValueError("Dataset '%s' not found" % name)
+
+    return _do_load_dataset_from_doc(obj, dataset_doc)
+
+
+def _do_load_dataset_from_doc(obj, dataset_doc):
+    if isinstance(dataset_doc, dict):
+        dataset_doc = foo.DatasetDocument(**dataset_doc)
 
     sample_collection_name = dataset_doc.sample_collection_name
     frame_collection_name = dataset_doc.frame_collection_name
--- a/fiftyone/core/odm/utils.py
+++ b/fiftyone/core/odm/utils.py
@@ -673,10 +673,9 @@ def load_dataset(identifier):
         query_key = "name"
         query_val = identifier
 
-    res = db.datasets.find_one({query_key: query_val})
+    res = db.datasets.find_one({query_key: query_val}, {"name": True})
     if not res:
         raise ValueError(
             f"Dataset with {query_key}={query_val} does not " f"exist"
         )
-    dataset_doc = foo.DatasetDocument.from_dict(res)
-    return fod.Dataset(_doc=dataset_doc, _create=False, _virtual=True)
+    return fod.Dataset(res.get("name"), _create=False)
--- a/fiftyone/core/odm/utils.py
+++ b/fiftyone/core/odm/utils.py
@@ -657,3 +657,26 @@ class DocumentRegistryError(Exception):
 
 
 _document_registry = DocumentRegistry()
+
+
+def load_dataset(identifier):
+    """Loads the dataset from the database with the given name or id."""
+    import fiftyone.core.odm as foo
+    import fiftyone.core.dataset as fod
+
+    db = foo.get_db_conn()
+    try:
+        query_key = "_id"
+        query_val = ObjectId(identifier)
+
+    except:
+        query_key = "name"
+        query_val = identifier
+
+    res = db.datasets.find_one({query_key: query_val})
+    if not res:
+        raise ValueError(
+            f"Dataset with {query_key}={query_val} does not " f"exist"
+        )
+    dataset_doc = foo.DatasetDocument.from_dict(res)
+    return fod.Dataset(_doc=dataset_doc, _create=False, _virtual=True)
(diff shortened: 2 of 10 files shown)


brimoor (changes requested):
Sorry this cannot be merged as is. Will leave comments when I can

fiftyone/core/dataset.py
```diff
@@ -6808,11 +6811,13 @@ def _load_clips_source_dataset(frame_collection_name):
     return load_dataset(doc["name"])
 
 
-def _load_dataset(obj, name, virtual=False):
+def _load_dataset(obj, name, virtual=False, doc=None):
```
brimoor, line 6814: Unfortunately we cannot load the dataset doc outside of `Dataset()` and pass it in because a key step of the workflow is line 6813, which ensures that any migrations happen when loading existing datasets.

Datasets may not necessarily be migrated to the current database version until the first time they are loaded in a new database version environment. This is also why you cannot pass `_virtual=True` to the `Dataset` constructor as this PR does, even if the rest of the PR were fine. That kwarg is only used in a couple special circumstances where we do not wish to trigger dataset migrations.

fiftyone/core/utils.py
```diff
@@ -1992,3 +1995,26 @@ def validate_hex_color(value):
 
 
 fos = lazy_import("fiftyone.core.storage")
+
+
+def load_dataset(identifier):
```
brimoor, line 2000: By convention, things in `fiftyone.core.utils` are things that users may directly call. I would argue that users should not be loading datasets by `ObjectID` since this is not information that we publicly expose to them.

Can you please move this utility to the `fiftyone.core.odm.utils` module? That's a better place for internal-only utilities.

fiftyone/core/utils.py
```diff
@@ -1992,3 +1995,26 @@ def validate_hex_color(value):
...
+    """Loads the dataset from the database with the given name or id."""
```
brimoor, line 2001: I propose updating the signature of this method to:

```py
def load_dataset(id=None, name=None):
```

where the user can provide either the `name` or `id` of the dataset as a kwarg.

Although it is an edge case, the current implementation does not allow for the possibility that the user assigns a dataset name like `name=str(ObjectId())`, as it would be assumed that is the `_id`, not the `name` of the dataset.

As for the implementation of the method, since you cannot directly load the `DatasetDocument` outside of the `Dataset` constructor, if the user provides an `id`, you'll just have to lookup the `name` and then load the dataset using the name in a second step. It's unfortunate that it requires two database connections, but it would take more effort to avoid that given the migration requirements and also the fact that `Dataset` objects are singletons keyed by `name`, not `id`.

fiftyone/operators/executor.py
```diff
@@ -468,19 +468,23 @@ def dataset_id(self):
     @property
     def view(self):
         """The :class:`fiftyone.core.view.DatasetView` being operated on."""
-        if self._view is not None:
+        if self._view:
```
brimoor, line 471: It's critical that this remain unchanged. `bool(DatasetView)` checks if `len(view) > 0`, which is not the same thing as `view is not None`.

fiftyone/server/view.py
```diff
@@ -84,7 +84,8 @@ def get_view(
     """Gets the view defined by the given request parameters.
 
     Args:
-        dataset_name: the dataset name
+        dataset: the dataset name or :class:`fiftyone.core.dataset.Dataset`
+        instance
```
brimoor, line 88: nit: indent second, third etc lines of docstrings as done elsewhere. This is required for Sphinx docstrings to render properly

brimoor (changes requested):
I would suggest renaming this PR to reflect the main goal (like the other PR) executing delegated operations by ID rather than by name. Because the load-dataset-by-ID is an internal utility to support that

fiftyone/core/utils.py
kaixi-wang: sure I didn't intend this to be user facing but I didn't know which util file to use

fiftyone/core/dataset.py
swheaton: I think dataset migration should not occur in a delegated operation. However this should be enforced by convention of never having `FIFTYONE_DATABASE_ADMIN=True` in the delegated runner, so that an appropriate error will be raised to the user.
So I agree that it shouldn't quietly ignore the migration check.

brimoor: There's actually a distinction between database and dataset migrations. The `FIFTYONE_DATABASE_ADMIN=true` setting is not required in order to migrate a dataset to the *current* version of the database.

For example, it's possible that an admin runs:

```
# Update database version but do not migrate any datasets
fiftyone migrate -v X.XX
```

rather than 

```
# Update database version and migrate all datasets
fiftyone migrate -v X.XX --all
```

and relies on the automatic dataset-migration-to-HEAD to occur as (potentially non-admin) users actually access each dataset.

kaixi-wang: Like you said below, this isn't a user facing function and I believe this is a circumstance where you don't want to trigger migrations. Migration should not be triggered when the datasets data is being accessed from an external worker (like an orchestrator). Migrations should be deliberate. Also failed migrations happening in a separate process will cause the app to hang, and it won't be obvious to end users why fiftyone isn't working.

Also, without passing the doc, then every resolution of dataset id requires 2 queries (which also triggers 2 mutations to that dataset to update last loaded at). Being able to directly pass in a doc would reduce the number of queries would benefit performance (especially when thinking about the serialization/deserialization roundtrips via the proxy)

brimoor: Here's the aforementioned logic:
https://github.com/voxel51/fiftyone/blob/4787b76e676c35598b58c25ec627bb484ab10f95/fiftyone/migrations/runner.py#L229-L246

brimoor: The two queries is unfortunate, but it's the only straightforward path here without doing more work to explicitly allow directly loading datasets by ID, which would require dealing with the fact that datasets are singletons keyed by `name`, not `id`. It's possible, but not the goal of this PR.

There will not be two `last_updated_at` updates. In your new `load_dataset()` util the first query will be a direct query to `find()` the `name` for the `id`.

kaixi-wang: whatever is triggered regarding the migration is often a problem when upgrading fiftyone and requires restarting the process

In general, if we ever want to make the sdk usable in teams, we need to be able to not have all these automatic mutations happening under the hood that add overhead and opportunity for unanticipated/concurrency errors.

fiftyone/core/utils.py
kaixi-wang: How does it take more effort? The current implementation accomplishes this.

kaixi-wang: Again, like you said above, this is NOT user facing. It can be used for debugging or internally when you need to access the dataset in some intermediate call.

fiftyone/core/dataset.py
kaixi-wang: although the dataset would ideally be already migrated if necessary when a user loaded a dataset in the app, such additional checks is killing any hopes of every making the proxy usable

kaixi-wang: Also, maybe I'm missing something, but can you explain what the problem would be if the dataset isn't migrated in the delegated context? I can see issues if the app failed to migrate the dataset and the UI would throw errors, but mongo is supposed to be schema-less and be flexible.

fiftyone/core/utils.py
kaixi-wang: Also loading the dataset via id uses pymongo directly to get the dataset doc from the db, but to just read the name from that query to then load the full dataset via mongoengine doc.objects, you are then attempting to load a cached dataset doc from the mongoengine cache, which may be causing errors like "dataset-name doesn't exist" even though you just created it or creating corrupted datasets/missing data when mixing local sdk usage with UI/in app changes.

For ref:
```
As of MongoEngine 0.8 the querysets utilise a local cache. So iterating it multiple times will only cause a single query.
```
https://docs.mongoengine.org/guide/querying.html

kaixi-wang:
@brimoor 
skipping the migration check and logging improves loading times significantly as well. 3.5sec vs 700ms:
image
image

kaixi-wang:
@swheaton @brimoor 
removed loading datasets via doc without logging and migration

fiftyone/core/odm/utils.py
```diff
@@ -657,3 +657,28 @@ class DocumentRegistryError(Exception):
...
+
+    if name:
+        return fod.Dataset(name, _create=False)
+
+    if not _id:
+        raise ValueError("Must provide either id or name")
+
+    db = foo.get_db_conn()
+    try:
+        uid = ObjectId(_id)
```
brimoor, line 675: This has the benefit of allowing the passed `id` to be either a string or `ObjectId`, which is a nice touch 👍

fiftyone/core/odm/utils.py
```diff
@@ -657,3 +657,28 @@ class DocumentRegistryError(Exception):
...
 _document_registry = DocumentRegistry()
+
+
+def load_dataset(_id=None, name=None):
+    """Loads the dataset from the database by _id or name."""
+    import fiftyone.core.odm as foo
+    import fiftyone.core.dataset as fod
+
+    if name:
+        return fod.Dataset(name, _create=False)
```
brimoor, line 668: would be nice to use `fod.load_dataset(name)` here and below in this function rather than `Dataset(name, _create=False)` to minimize the number of times that private kwargs are used

fiftyone/core/odm/utils.py
```diff
@@ -657,3 +657,28 @@ class DocumentRegistryError(Exception):
 
 
 _document_registry = DocumentRegistry()
+
+
+def load_dataset(_id=None, name=None):
```
brimoor, line 662: Can you please use `id` instead of `_id` as the parameter name here? For consistency with how other methods name kwargs that refer to IDs. And since the implementation of this function allows the provided value to be either an `Objectid` or a `str` because you cast to `ObjectId` internally (which is 👍).

For context, *within* methods, I often use `_id` when the value is an `ObjectId` and `id` when the value is a `str`.

As for the "public" interface of functions, kwargs that start with `_` are undocumented parameters that are intended to convey to other developers that hey, if you're using this parameter, you're doing something slightly subtle/unintended, there's probably a better way.

brimoor: Also can you please fill out the docstring for this method with `Args` and `Returns` like the other methods in this module?

By convention, every non-private function in `fiftyone.core` should have a full docstring, even if the function is only intended to be called by developers, not end users.

fiftyone/core/odm/utils.py
```diff
@@ -657,3 +657,28 @@ class DocumentRegistryError(Exception):
...
+        uid = ObjectId(_id)
+    except:
+        # Although _id is an ObjectId by default, it's possible to set it to
+        # something else
+        uid = _id
+
+    res = db.datasets.find_one({"_id": uid}, {"name": True})
+    if not res:
+        raise ValueError(f"Dataset with _id={uid} does not exist")
+    return fod.Dataset(res.get("name"), _create=False)
```
brimoor, line 684: prefer `load_dataset()` here

fiftyone/operators/executor.py
```diff
@@ -471,16 +472,20 @@ def view(self):
         if self._view is not None:
             return self._view
 
-        # Forces a dataset reload if necessary
```
brimoor, line 474: note here how we were relying on the fact that a `dataset.reload()` is triggered

fiftyone/operators/executor.py
```diff
@@ -443,12 +438,18 @@ def dataset(self):
         """The :class:`fiftyone.core.dataset.Dataset` being operated on."""
         if self._dataset is not None:
             return self._dataset
-
-        dataset_name = self.request_params.get("dataset_name", None)
-
-        self._dataset = fod.load_dataset(dataset_name)
-        self._dataset.reload()
```
brimoor, line 450: The `reload()` call is still necessary here because `Dataset` objects are singletons and therefore we are protecting against situations where the dataset may have been previously operated on by this process but has since had a schema change that the in-memory object doesn't know about. 

The change you made in https://github.com/voxel51/fiftyone/pull/3922 removed an unintended type of caching that I didn't realize existed, but it is still by design (somewhat regrettable, but nonetheless an intentional choice) that dataset objects are cached in-process by [DatasetSingleton](https://github.com/voxel51/fiftyone/blob/4787b76e676c35598b58c25ec627bb484ab10f95/fiftyone/core/singletons.py#L12):

```py

# No db connection here
also_dataset = fo.load_dataset(name)

assert dataset is also_dataset
```

fiftyone/operators/executor.py
```diff
@@ -471,16 +472,20 @@ def view(self):
         if self._view is not None:
             return self._view
 
-        # Forces a dataset reload if necessary
-        _ = self.dataset
+        # Always derive the view from the context's dataset
+        dataset = self.dataset
+        if not dataset:
```
brimoor, line 477: This error check is unnecessary: if a valid dataset could not be loaded an error would have already been raised when `self.dataset` is called.

brimoor (changes requested):
@kaixi-wang basic implementation is looking good now, thanks for the changes! 💪 

Left some small cleanup requests then this is ready to go

fiftyone/operators/executor.py
brimoor: One optimization we could make that we should *not* group into this PR is adding a syntax like this:

```py
dataset = fod.load_dataset(name, reload=True)
```

in place of:

```py
dataset = fod.load_dataset(name)
dataset.reload()
```

that would reload the dataset from the db even if it exists in the singleton cache. This would avoid what is currently a double load of the dataset the first time it is accessed in a process.

kaixi-wang: ok added the reload() with a TODO in the comment

brimoor (dismissed):
Perfect thanks 🙇

PR #3922, opened 2023-12-13
Remove loading from mongoengine cache

kaixi-wang:

- Current dataset loading relies on getting the DatasetDocument from the local mongoengine queryset cache, which is problematic when using fiftyone in different processes
- This change should prevent errors like "Dataset does not exist" immediately after creating a dataset or "Dataset already exists" after deleting a dataset, especially when using the python sdk locally but connecting to a remote app.

```
As of MongoEngine 0.8 the querysets utilise a local cache. 
```
link to mongoengine doc: https://docs.mongoengine.org/guide/querying.html

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -6831,12 +6831,12 @@ def _load_dataset(obj, name, virtual=False):
 
 
 def _do_load_dataset(obj, name):
-    try:
-        # pylint: disable=no-member
-        db = foo.get_db_conn()
-        dataset_doc = db.datasets.find_one({"name": name})
-    except moe.DoesNotExist:
+    # pylint: disable=no-member
+    db = foo.get_db_conn()
+    res = db.datasets.find_one({"name": name})
+    if not res:
         raise ValueError("Dataset '%s' not found" % name)
+    dataset_doc = foo.DatasetDocument.from_dict(res)
 
     sample_collection_name = dataset_doc.sample_collection_name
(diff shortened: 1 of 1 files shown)


fiftyone/core/dataset.py
```diff
@@ -6831,11 +6831,12 @@ def _load_dataset(obj, name, virtual=False):
 
 
 def _do_load_dataset(obj, name):
-    try:
-        # pylint: disable=no-member
-        dataset_doc = foo.DatasetDocument.objects.get(name=name)
-    except moe.DoesNotExist:
+    # pylint: disable=no-member
```
brimoor, line 6834: I think the pylint disablement is no longer necessary?

brimoor (approved):
LGTM, thanks for this🥇

I didn't realize that Mongoengine's `objects` cached things... this could definitely resolve some pesky concurrency issues!!

brimoor:
@kaixi-wang can you update the destination branch to `release/v0.23.2` so we can get this out ASAP?

swheaton:
Do we know the performance implications (if any) of this?

kaixi-wang:
> Do we know the performance implications (if any) of this?

Connecting to a local mongodb, comparing `load_dataset` times on for this branch against release/0.23.2, there doesn't seem to be any significant performance difference:
image
image

This makes sense since mongodb also has a local cache, which is much more reliable since it isn't client side

brimoor (approved):
LGTM

PR #3923, opened 2023-12-13
Toggle dynamic batcher for add_samples(), others

swheaton:

- Add `StaticBatcher` as an option to use in place of `DynamicBatcher`. The latter attempts to optimize for a nice flowing progress bar, whereas the former optimizes for predictability and is better in cases where latency between machine and MongoDB is higher.
- Add `BSONSizeDynamicBatcher` as an option. Instead of using latency between calls as a target, attempts to achieve a target BSON content size. The caller must register "backpressure" with the batcher so it can adjust the batch size - by giving it an iterable of BSON-able items that would approximate the size you want to optimize for. In the main case, it would be `list[fo.Sample]` in the batches and corresponding `list[dict]` provided back to the batcher.
- Add configuration to allow switching from the default "latency" to new "static" and "size" batchers. Also database batcher size (for static), target latency (for default 'latency' batcher) and target content size (for 'size' batcher) are all configurable.

- By default, add_samples() should do the same thing.
- Only if you configure it to use `StaticBatcher` will it change behavior.
- Added some unit tests for batch factory and static batcher

```python
import fiftyone as fo

dataset1 = fo.zoo.load_zoo_dataset("quickstart")

dataset2 = fo.Dataset()

# Can set this or not, default is 'latency'
#fo.config.default_batcher = "latency"

# Should see irregular batch size with progress bar flowing smoothly,
#  approximately 5 updates per second.
fo.config.batcher_target_latency = 0.2
dataset2.add_samples(dataset1)

# Should see regular batch size of 10 in the progress bar
fo.config.default_batcher = "static"
fo.config.batcher_static_size = 10
dataset2.add_samples(dataset1)

# Should see irregular batch size, where the sample dict size is approximately 32 KiB,
#  or 6-8 samples of "quickstart" dataset.
fo.config.default_batcher = "size"
fo.config.batcher_target_size = 2 ** 15
dataset2.add_samples(dataset1)

dataset1.delete()
dataset2.delete()
```

        notes for FiftyOne users.

Added new batcher choices (static; target-content-size-based) for greater predictability in some batched database operations such as `add_samples()`.

-   [ ] Other

### DIFF
--- a/docs/source/user_guide/config.rst
+++ b/docs/source/user_guide/config.rst
@@ -74,9 +74,10 @@ FiftyOne supports the configuration options described below:
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
 | `batcher_static_size`         | `FIFTYONE_BATCHER_STATIC_SIZE`      | `100`                         | Fixed size of batches. Ignored if `default_batcher` is not `static`.                   |
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
-| `batcher_target_size`         | `FIFTYONE_BATCHER_TARGET_SIZE`      | `2 ** 20`                     | Target content size of batches, in bytes. Ignored if `default_batcher` is not `size`.  |
+| `batcher_target_size_bytes`   | `FIFTYONE_BATCHER_TARGET_SIZE_BYTES`| `2 ** 20`                     | Target content size of batches, in bytes. Ignored if `default_batcher` is not `size`.  |
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
-| `batcher_target_latency`      | `FIFTYONE_BATCHER_TARGET_LATENCY`   | `0.2`                         | Target latency between batches. Ignored if `default_batcher` is not `latency`.         |
+| `batcher_target_latency`      | `FIFTYONE_BATCHER_TARGET_LATENCY`   | `0.2`                         | Target latency between batches, in seconds. Ignored if `default_batcher` is not        |
+|                               |                                     |                               | `latency`.                                                                             |
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
 | `default_sequence_idx`        | `FIFTYONE_DEFAULT_SEQUENCE_IDX`     | `%06d`                        | The default numeric string pattern to use when writing sequential lists of             |
 |                               |                                     |                               | files.                                                                                 |
--- a/docs/source/user_guide/config.rst
+++ b/docs/source/user_guide/config.rst
@@ -55,7 +55,10 @@ FiftyOne supports the configuration options described below:
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
 | `default_batch_size`          | `FIFTYONE_DEFAULT_BATCH_SIZE`       | `None`                        | A default batch size to use when :ref:`applying models to datasets <model-zoo-apply>`. |
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
-| `bulk_write_batch_size`       | `FIFTYONE_BULK_WRITE_BATCH_SIZE`    | `100_000`                     | Batch size to use for bulk writing MongoDB operations; cannot be > 100,000             |
+| `bulk_write_batch_size`       | `FIFTYONE_BULK_WRITE_BATCH_SIZE`    | `100_000`                     | Batch size to use for bulk writing MongoDB operations; cannot be > 100,000.            |
+|                               |                                     |                               |                                                                                        |
+|                               |                                     |                               | Default changes to `10_000` for                                                        |
+|                               |                                     |                               | :ref:`Teams SDK in API connection mode <teams-api-connection>`.                        |
 +-------------------------------+-------------------------------------+-------------------------------+----------------------------------------------------------------------------------------+
 | `default_batcher`             | `FIFTYONE_DEFAULT_BATCHER`          | `latency`                     | Batching implementation to use in some batched database operations such as             |
 |                               |                                     |                               | :meth:`add_samples() <fiftyone.core.dataset.Dataset.add_samples>`. Supported values    |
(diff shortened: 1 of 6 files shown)


j053y (commented):
When would the exposed backpressure mechanism be relevant? Shouldn't it always be internal to the batcher?

j053y (approved):
I see, thanks for the explanation. That makes sense to me. Thanks for this!

swheaton:
rebased

fiftyone/core/utils.py
```diff
@@ -1144,10 +1144,317 @@ def _compute_batch_size(self):
...
+        if self._last_time is not None:
+            time_delta = current_time - self._last_time
+
+        self._last_time = current_time
+        return time_delta
+
+
+# Define this for backwards compatibility in case someone was using this
+# batcher directly
+DynamicBatcher = LatencyDynamicBatcher
```
brimoor, line 1247: nice touch 💪

fiftyone/core/utils.py
```diff
@@ -1144,10 +1144,317 @@ def _compute_batch_size(self):
...
+
+
+class BsonSizeDynamicBatcher(BaseDynamicBatcher):
```
brimoor, line 1250: nit: elsewhere in the codebase we've historically used `BSONSizeDynamicBatcher`, ie all acronym letters are capitalized. EG `TorchCLIPModel`.

fiftyone/core/utils.py
```diff
@@ -1144,10 +1144,317 @@ def _compute_batch_size(self):
...
+            ``progress=True``. If not provided, it is computed via
+            ``len(iterable)``, if possible
+    """
+
+    manual_backpressure = True
+
+    def __init__(
+        self,
+        iterable,
+        target_size=2**15,
```
brimoor, line 1319: nit: I see 2**15 here but in `fo.config` the default is `2**20`. Standardize on `2**20`?

docs/source/user_guide/config.rst
```diff
@@ -55,6 +55,21 @@ FiftyOne supports the configuration options described below:
...
+|                               |                                     |                               | :meth:`add_samples() <fiftyone.core.dataset.Dataset.add_samples>`. Supported values    |
+|                               |                                     |                               | are `latency`, `size`, and `static`.                                                   |
+|                               |                                     |                               |                                                                                        |
+|                               |                                     |                               | `latency` is the default, which uses a dynamic batch size to achieve a target latency  |
+|                               |                                     |                               | of `dynamic_batch_target_latency` between calls. `size` is the default for the         |
+|                               |                                     |                               | FiftyOne Teams SDK, which targets a size of `dynamic_batch_target_size` bytes for      |
```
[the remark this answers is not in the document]
brimoor: FWIW I think its fine to leave as-is

fiftyone/core/config.py
```diff
@@ -154,6 +154,30 @@ def __init__(self, d=None):
             env_var="FIFTYONE_DEFAULT_BATCH_SIZE",
             default=None,
         )
+        self.default_batcher = self.parse_string(
+            d,
+            "default_batcher",
+            env_var="FIFTYONE_DEFAULT_BATCHER",
+            default="latency",
+        )
+        self.database_static_batch_size = self.parse_int(
```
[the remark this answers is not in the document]
swheaton: guilty ... I messed with the names so they would fit in the documentation table 😬 
seems fine to me

docs/source/user_guide/config.rst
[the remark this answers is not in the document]
swheaton: I will mention it

fiftyone/core/utils.py
swheaton: ok sure, just a style choice. In this case, it's equally as readable.
https://stackoverflow.com/questions/2853531/how-do-you-pep-8-name-a-class-whose-name-is-an-acronym

fiftyone/core/utils.py
swheaton: relic of testing different values

fiftyone/core/config.py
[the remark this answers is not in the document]
swheaton: Part of it was there already exists `FIFTYONE_DEFAULT_BATCH_SIZE` but that is only for model application. Then we have a separate mechanism for `bulk_write()` ... probably nice to clean this up at some point but fine for now.

swheaton:
Rebased after `bulk_write` config merge, and made final suggested changes.

docs/source/user_guide/config.rst
```diff
@@ -55,7 +55,28 @@ FiftyOne supports the configuration options described below:
...
+| `default_batcher`             | `FIFTYONE_DEFAULT_BATCHER`          | `latency`                     | Batching implementation to use in some batched database operations such as             |
+|                               |                                     |                               | :meth:`add_samples() <fiftyone.core.dataset.Dataset.add_samples>`. Supported values    |
+|                               |                                     |                               | are `latency`, `size`, and `static`.                                                   |
+|                               |                                     |                               |                                                                                        |
+|                               |                                     |                               | `latency` is the default, which uses a dynamic batch size to achieve a target latency  |
+|                               |                                     |                               | of `dynamic_batch_target_latency` between calls. `size` is the default for the         |
```
brimoor, line 68: @swheaton the settings in the description column here need updating too 😄

swheaton: crap

fiftyone/core/config.py
```diff
@@ -160,6 +160,30 @@ def __init__(self, d=None):
...
+            env_var="FIFTYONE_DEFAULT_BATCHER",
+            default="latency",
+        )
+        self.batcher_static_size = self.parse_int(
+            d,
+            "batcher_static_size",
+            env_var="FIFTYONE_BATCHER_STATIC_SIZE",
+            default=100,
+        )
+        self.batcher_target_size = self.parse_int(
```
brimoor, line 175: okay one final idea, take it or leave it: you might consider `batcher_target_size_bytes` here just to clarify that "size" means bytes here, not the same as `batcher_static_size` or `default_batch_size`. As is, the different semantics could be confusing.

brimoor (approved):
LGTM. Saw one documentation bug and one last naming suggestion for consideration

fiftyone/core/config.py
swheaton: agree. and it _just_ fits in the table too!

swheaton:
ok last time 🤞🏼

fiftyone/core/config.py
brimoor: lol we're never increasing the width of this table again so we're now limited to 26^current_width different config settings 🤣

swheaton: I have some ideas for configuration upgrades that could fix this but it is not this day

j053y (approved):

PR #3927, opened 2023-12-15
ensure line breaks in embeddings legend

nebulae:

ensure line breaks in the embedding legend so the legend doesn't stretch too far into the plot when the values are very long strings 

image

locally

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/embeddings/src/tracesToData.tsx
+++ b/app/packages/embeddings/src/tracesToData.tsx
@@ -84,7 +84,7 @@ export function tracesToData(
 }
 
 const addLineBreaks = ([key, trace]) => {
-  // split the key into chunks of <maxLegendLineLength>> characters or less, respecting word boundaries.
+  // split the key into chunks of <maxLegendLineLength> characters or less, respecting word boundaries.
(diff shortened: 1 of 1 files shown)


app/packages/embeddings/src/tracesToData.tsx
```diff
@@ -78,6 +83,24 @@ export function tracesToData(
     });
 }
 
+const addLineBreaks = ([key, trace]) => {
+  // split the key into chunks of <maxLegendLineLength>> characters or less, respecting word boundaries.
```
benjaminpkane, line 87:
```suggestion
  // split the key into chunks of <maxLegendLineLength> characters or less, respecting word boundaries.
```

benjaminpkane (approved):
It seems like a string with whitespace splits that include 35+ length words might still have long formatting?

Maybe an ellipses after 35+ chars and tooltip is more maintainable? But that may be hard to configure in plotly

kaixi-wang (approved):
reapproving :)

PR #3929, opened 2023-12-15
evaluate detection in case polylines is forming polygon

akshitpriyesh:

(For the cases where polylines is just the group of points on straight lines their existing evaluate detection method will not works as it tries to create a polygon using shapely and calculates iou.)

(This Patch is tested on multiple datasets having thousands of images. Here we are calculating IOU using gaussian distance.)
Not necessary

-   [ x] No. You can skip the rest of this section.
        notes for FiftyOne users.

For the cases where polylines is just the group of points on straight lines their existing evaluate detection method will not works as it tries to create a polygon using shapely and calculates iou. Here we are calculating IOU using gaussian distance.

-   [ x] App: FiftyOne application changes
-   [ ] Other

### DIFF
--- a/fiftyone/utils/iou.py
+++ b/fiftyone/utils/iou.py
@@ -495,28 +495,40 @@ def _compute_bbox_ious(preds, gts, iscrowd=None, classwise=False):
 
     return ious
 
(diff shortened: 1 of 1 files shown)


akshitpriyesh:
Another pr

PR #3930, opened 2023-12-15
Update iou.py

ArvindKumarS:

For the cases where polylines is just the group of points on straight lines (not closed polygon) their existing evaluate detection method will not works as it tries to create a polygon using shapely and calculates iou.

We modified _compute_polyline_ious function such that if its a closed polyline --> The usual computation happens to compute iou. 
If it is not closed and just straight lines: Then we compute normalized euclidian distance the same way as key point evaluation and take the average. 

We tested this on our sample dataset with a few hundred samples. 

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/iou.py
+++ b/fiftyone/utils/iou.py
@@ -638,7 +638,7 @@ def _compute_polyline_ious(
 ):
     is_symmetric = preds is gts
 
-    are_all_filled = np.sum([p["filled"] for p in preds])
+    are_all_filled = np.sum([gt["filled"] for gt in gts])
     if are_all_filled > 0:
         # are_all_filled = True for atleast one polyline instance, i.e., polygons/solid shape
         ious = polygon_iou_matrix(
--- a/fiftyone/utils/iou.py
+++ b/fiftyone/utils/iou.py
@@ -495,81 +495,141 @@ def _compute_bbox_ious(preds, gts, iscrowd=None, classwise=False):
 
     return ious
 
+def polygon_iou_matrix(preds, gts, error_level, is_symmetric, iscrowd=None, classwise=False, gt_crowds=None):
+    with contextlib.ExitStack() as context:
+            # We're ignoring errors, so suppress shapely logging that occurs when
+            # invalid geometries are encountered
+            if error_level > 1:
+                # pylint: disable=no-member
+                context.enter_context(
+                    fou.LoggingLevel(logging.CRITICAL, logger="shapely")
+                )
 
+            num_pred = len(preds)
+
+            pred_polys = _polylines_to_shapely(preds, error_level)
+            pred_labels = [pred.label for pred in preds]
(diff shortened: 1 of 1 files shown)


brimoor (changes requested):
@ArvindKumarS thanks for the contribution! 🥇 

Can you please install the pre-commit hooks as described in the [developer guide](https://github.com/voxel51/fiftyone/blob/develop/CONTRIBUTING.md#developer-guide) and run them on this file you edited? Thanks!

fiftyone/utils/iou.py
```diff
@@ -495,81 +495,141 @@ def _compute_bbox_ious(preds, gts, iscrowd=None, classwise=False):
...
+        each_dist.append(np.exp(-(dists**2) / (2 * (scale**2))))
+    return np.mean(each_dist)
+    
+def _compute_polyline_ious(
+    preds, gts, error_level, iscrowd=None, classwise=False, gt_crowds=None
+):
+    is_symmetric = preds is gts
 
-        return ious
+    are_all_closed = np.sum([p ["closed"] for p in preds])
```
brimoor, line 624: Whether a `Polyline` object represents a polyline or a polygon is slightly undefined in our current data model. There are two parameters:

```
        closed (False): whether the shapes are closed, i.e., and edge should
            be drawn from the last vertex to the first vertex of each shape
        filled (False): whether the polyline represents polygons, i.e., shapes
            that should be filled when rendering them
```

But this leaves open the possibility that the user set `closed=False` and instead duplicated the first node at the end of the `points` list to manually "close" the shape.

Arguably checking for `filled=True` is a more suitable check for polygons, since if the user provides `filled=False` the shape will be rendered as lines rather than filled polygons in the App. However, since we have not been strict about what determines a polygon vs a polyline, potentially a user could have a closed shape that they intend to represent a polygon but for whatever reason they set `filled=False` because they prefer for it not to be filled in when visualizing in the App.

brimoor: Actually I'm now looking at the documentation of the main `compute_ious()` method on line 40 and I see:

```
    For polylines, IoUs are computed assuming the shapes are solid (filled),
    regardless of their ``filled`` attributes.
```

so I recommend using `filled=True/False` to distinguish between polygons and polylines, and correspondingly update the documentation on line 40 to read:

```
    For :class:`fiftyone.core.labels.Polyline` objects, IoUs are computed
    as solid shapes when ``filled=True` and are computed using
    `object keypoint similarity <https://cocodataset.org/#keypoints-eval>`_
    when ``filled=False``.
```

ArvindKumarS: @brimoor  fixed in commit https://github.com/voxel51/fiftyone/pull/3930/commits/17665146f562bd8e9936d700c7cdc7be06a3257e

brimoor (approved):
@ArvindKumarS thanks for your contribution! We'll get this out in `fiftyone==0.23.2`!

PR #3932, opened 2023-12-15
Update selected sample operators

benjaminpkane:
Updates the the built-in operators that update the current sample selection. Due to routing updates, it is unnecessary to use the `useSetSelected` and update the `selectedSamples` atom as they are equivalent.

Tested locally


imanjra (approved):
LGTM

PR #3940, opened 2023-12-18
add parsing of string version list of sample ids in _parse_sample_ids

nebulae:

added check & conversion for comma delimited string of sample ids in _parse_sample_ids

locally & with unit test

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/tests/unittests/view_tests.py
+++ b/tests/unittests/view_tests.py
@@ -4268,6 +4268,20 @@ class ViewStageTests(unittest.TestCase):
         ]
         self.assertEqual(optimized_view._all_stages, expected_stages)
 
+    def test_make_optimized_select_view_group_dataset_with_strings_and_lists(
+        self,
+    ):
+        dataset, sample_ids = self._make_group_dataset()
+
+        optimized_view = fov.make_optimized_select_view(
+            dataset, [sample_ids[0], sample_ids[0]], flatten=True
+        )
+        expected_stages = [
+            fosg.SelectGroupSlices(),
+            fosg.Select(f"{sample_ids[0]},{sample_ids[0]}"),
+        ]
+        self.assertEqual(optimized_view._all_stages, expected_stages)
+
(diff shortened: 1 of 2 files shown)


tests/unittests/view_tests.py
```diff
@@ -4268,6 +4268,20 @@ def test_make_optimized_select_view_group_dataset(self):
...
+        self,
+    ):
+        dataset, sample_ids = self._make_group_dataset()
+
+        optimized_view = fov.make_optimized_select_view(
+            dataset, [sample_ids[0], sample_ids[0]], flatten=True
+        )
+        expected_stages = [
+            fosg.SelectGroupSlices(),
+            fosg.Select(f"{sample_ids[0]},{sample_ids[0]}"),
```
brimoor, line 4281: I don't believe we've publicly documented anywhere that you can pass a comma-separated list of IDs to `Select()`. Where did you encounter that?

brimoor: I see. Actually this line:

https://github.com/voxel51/fiftyone/blob/ad17d5e7c5bd080a00d4bec623d67c35222e5707/fiftyone/core/stages.py#L5648

is supposed to tell the App to parse `"list,of,str"` into `["list", "of", "str"]`, so it seems that is no longer working? cc @benjaminpkane can you help track down the root cause?

The `type=list<>` syntax is used in other view stages too, so I think that's the place to fix the string parsing.

benjaminpkane: I've included a view bar fix in #3941 with https://github.com/voxel51/fiftyone/pull/3941/commits/c45be8ac761c12533a18deb07b633fc7c93aaee3 as it also contains ObjectId client-side validation

brimoor:
Closing as the issue has been resolved by https://github.com/voxel51/fiftyone/pull/3940#discussion_r1439920957.

PR #3942, opened 2023-12-19
Optimize reload docs when performing bulk writes

brimoor:
When performing `dataset._bulk_write()`, only reload the docs that were involved in the mutations, if known.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -2642,7 +2642,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
 
         return d
 
-    def _bulk_write(self, ops, frames=False, ordered=False):
+    def _bulk_write(self, ops, ids=None, frames=False, ordered=False):
         if frames:
             coll = self._frame_collection
         else:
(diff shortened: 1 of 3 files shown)


swheaton:
Test snippet
```python
import time
import fiftyone as fo
ds = fo.zoo.load_zoo_dataset('coco-2017', splits='validation')
ds.clear_sample_field("metadata")
_ = [s for s in ds.iter_samples()]

t1 = time.time()
ds.compute_metadata()
t2 = time.time()
print(t2-t1)
```

Timings
1. 1.5s: clear singleton Sample instances first
2. 64s: On `release/v0.23.2`
3. 10s: On this branch

The speedup gets more pronounced as number of samples gets bigger though, so this change is definitely worth it

fiftyone/core/dataset.py
```diff
@@ -7586,7 +7588,7 @@ def _add_collection_with_new_ids(
         for old_id, new_id in zip(old_ids, new_ids)
     ]
 
-    dataset._bulk_write(ops, frames=True)
+    dataset._bulk_write(ops, ids=[], frames=True)
```
swheaton, line 7591: Nice; they're new samples so we don't need to reload any singleton samples

swheaton (approved):
rationale discussed offline prior, so agreed on that.
implementation lgtm; changes needed, see prior comment

PR #3943, opened 2023-12-19
[WIP] Adding config setting to disable singleton cache

brimoor:
## Example usage

```py
import fiftyone as fo

fo.config.singleton_cache = False

dataset = fo.Dataset()
sample = fo.Sample(filepath="image.jpg")

dataset.add_sample(sample)
print(sample.in_dataset)  # True

also_sample = dataset.first()
print(sample is also_sample)  # False

also_dataset = fo.load_dataset(dataset.name)
print(also_dataset is dataset)  # False
```

## TODO

When singleton caching is disabled, there are a variety of places where patterns like this are used where the `reload()` is now always unnecessary:

```py
dataset.reload()  # reload in case we got a singleton rather than a refresh load from the database
```

We should probably refactor those calls into a single method call so that unnecessary reloads can be skipped:

```py
# Reload the dataset (only) if a singleton instance was returned
dataset = fo.load_dataset(name, reload=True)
```


ritch (commented):
🙌

swheaton:
FYI This PR points at release branch currently but is not intended for release.

PR #3945, opened 2023-12-20
[WIP] standardizing reload usage

brimoor:
Untested attempts to standardized `dataset.reload()` usage such that unnecessary reloads are not be performed when `load_dataset()` does *not* use a cached singleton reference.

Cam: I am picking up this PR plz ignore until I move it into review

TODO:
- Test that these changes actually work and don't break anything
- Look for new instances of load or reload dataset that might be impacted by these changes
- Discuss the server changes more in depth with ben and sash


CamronStaley:
Plz ignore this PR

PR #3957, opened 2023-12-30
Add support for configuring whether run cleanup happens

brimoor:
No change in default behavior; just adds support for a new syntax:

```py
run.register_run(dataset, key, overwrite=True, cleanup=False)
```

which overwrites an existing run without running the existing run's `cleanup()` method.


brimoor:
Merging into `release/v0.23.3` instead in https://github.com/voxel51/fiftyone/pull/3978

PR #3958, opened 2024-01-02
Fix sort by similarity for datasets with a color scheme

benjaminpkane:
The `/sort` route is currently used only to ensure a `dist_field` is correctly populated if the parameter is provided. Before routing updates, this meant a new dataset scheme was needed in the response. Omitting the response prevents serialization errors when a dataset has a `color_scheme`


sashankaryal (approved):
sounds good 🐐

PR #3964, opened 2024-01-04
Group mode re-filtering of aggregations

benjaminpkane:
The `group` statistics mode in the App requires re-applying sidebar filters after the group collection is "flattened" through a self-referencing lookup. Adds corresponding GraphQL tests

### Before

https://github.com/voxel51/fiftyone/assets/19821840/362e9af6-07af-47ec-a707-384f977de381

### After

https://github.com/voxel51/fiftyone/assets/19821840/2645cac0-3c59-41c7-9186-dd6bf3817e67


sashankaryal (dismissed):
aggregation count tests look 🤌

sashankaryal (approved):
🚀

PR #3965, opened 2024-01-04
fix keyboard navigation for selectors

imanjra:

- fix keyboard navigation for selectors
- add tweak to auto-scroll to active selector option when navigating to hidden options

Using the dataset selector in the app with both keyboard and mouse navigation

        notes for FiftyOne users.

fix keyboard navigation for selectors in the app

-   [ ] Other


benjaminpkane (dismissed):
LGTM ⌨️ 

Historically, an `undefined` active value functioned as a pivot and default value. I tweaked things for that behavior because I don't think we can assume the current value, e.g. in the dataset selector, is present in a subset of search results. Hope that makes sense! 😅

imanjra:
Thanks for the tweaks @benjaminpkane! 🙂

benjaminpkane (approved):

PR #3666, opened 2023-10-12
Patches group slice fix

benjaminpkane:
Fixes converting to patches in the App in group datasets when sidebar filters are present

* [x] e2e test
* [x] `fiftyone.server.view.get_view` unit test

PR #3966, opened 2024-01-04
Use mixed mode in select_group_slices

sashankaryal:

When a group dataset is created that has a video and image slice and frame fields (like frame detections) are added to the video slice, then images cause an error in the App when you try to open them in the sample modal. This PR fixes this bug.

This bug is a potential regression introduced by #3666.

Smoke tested with the following test code.

```
import fiftyone as fo

dataset = fo.Dataset('group-frame-fields-bug, persistent=True)

video_sample = fo.Sample(filepath='/Users/abc/video.mp4', group=fo.Group().element('video'))
dataset.add_sample(video_sample)
dataset.add_frame_field('boxes', fo.EmbeddedDocumentField, embedded_doc_type=fo.Detections)

thumbnail_sample = fo.Sample(
    filepath='/Users/abc/image.jpg',
    group=fo.Group().element('thumbnail')
)
dataset.add_sample(thumbnail_sample)
```

`to_patches` was also verified to be working. This was tested against `quickstart` dataset's `predictions.detections` field.

-   [ ] Other

### DIFF
--- a/fiftyone/server/mutation.py
+++ b/fiftyone/server/mutation.py
@@ -217,6 +217,17 @@ class Mutation(SetColorScheme):
                 else None,
             )
 
+            is_to_patches_in_stages = False
+
+            if form.add_stages:
+                is_to_patches_in_stages = any(
+                    stage.get("_cls") == "fiftyone.core.stages.ToPatches"
+                    for stage in form.add_stages
+                )
+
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
LGTM 🐛

PR #3976, opened 2024-01-05
fix url construction

sashankaryal:
This PR fixes the edge case in fetch URL construction where fetch `origin` is already encapsulated in `path`, in which case `origin` can be omitted.


benjaminpkane (approved):
LGTM. I tweaked the solution after doing some more investigation. `getSampleSrc` and the fetch function are both doing prefix handling because `getSampleSrc` was initially only used for media URLs that don't use `fetch`.

Future work could clean this up, but avoiding any path changes if a valid URL avoids the redundant prefix.

PR #3977, opened 2024-01-05
Apply YOLO Models Directly to FiftyOne Datasets

jacobmarks:

Adds support for passing Ultralytics YOLO models directly to `sample_collection.apply_model()`!

Example Usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz
from ultralytics import YOLO

yolov5n = YOLO('yolov5n.pt')
yolov8n = YOLO('yolov8n.pt')
yolov8seg = YOLO('yolov8n-seg.pt')
yolov8npose = YOLO('yolov8n-pose.pt')

## YOLOv5 detection model, which we previously supported
dataset.apply_model(yolov5n, label_field="yolov5n")

## New: YOLOv8 — we get it for free!
dataset.apply_model(yolov8n, label_field="yolov8n")

## Instance segmentation model
dataset.apply_model(yolov8seg, label_field="yolov8seg")

## Pose estimation model
dataset.apply_model(yolov8npose, label_field="yolov8npose")
```

Additionally, I found that the previous YOLO instance segmentation conversion was incorrect, so this patches that!

## Additional Notes:
It would be pretty easy for us to do something similar for CLIP, SAM, OpenCLIP, YOLO-NAS, and potentially even HF models! Of course, too many and we would likely want to reorganize the code, but I think this will make life easier for users :)

-   [x] Other

### DIFF
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -122,19 +122,6 @@ def apply_model(
     try:
         if isinstance(model, ultralytics.YOLO):
             model = fouu._convert_yolo_model(model)
-            return apply_model(
-                samples,
-                model,
-                label_field=label_field,
-                confidence_thresh=confidence_thresh,
-                store_logits=store_logits,
-                batch_size=batch_size,
-                num_workers=num_workers,
-                skip_failures=skip_failures,
-                output_dir=output_dir,
-                rel_dir=rel_dir,
-                **kwargs,
-            )
     except ImportError:
         pass
 
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -118,7 +118,7 @@ def apply_model(
         )
 
     if _is_ultralytics_model(model):
-        model = fouu.convert_model(model)
+        model = fouu.convert_ultralytics_model(model)
 
     if not isinstance(model, Model):
         raise ValueError("Unsupported model type: %s" % type(model))
@@ -788,7 +788,7 @@ def compute_embeddings(
         )
 
     if _is_ultralytics_model(model):
-        model = fouu.convert_model(model)
+        model = fouu.convert_ultralytics_model(model)
 
     if not isinstance(model, Model):
         raise ValueError("Unsupported model type: %s" % type(model))
@@ -1272,12 +1272,10 @@ def compute_patch_embeddings(
             missing or ``None`` values to indicate uncomputable embeddings
     """
     if _is_ultralytics_model(model):
-        model = fouu.convert_model(model)
+        model = fouu.convert_ultralytics_model(model)
 
     if not isinstance(model, Model):
-        raise ValueError(
-            "Model must be a %s instance; found %s" % (Model, type(model))
-        )
+        raise ValueError("Unsupported model type: %s" % type(model))
 
     if not model.has_embeddings:
         raise ValueError(
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -59,22 +59,21 @@ def apply_model(
     rel_dir=None,
     **kwargs,
 ):
-    """Applies the :class:`FiftyOne model <Model>` or
-    :class:`Lightning Flash model <flash:flash.core.model.Task>` to the samples
-    in the collection.
+    """Applies the model to the samples in the collection.
 
     This method supports all of the following cases:
 
     -   Applying an image :class:`Model` to an image collection
     -   Applying an image :class:`Model` to the frames of a video collection
     -   Applying a video :class:`Model` to a video collection
+    -   Applying an ``ultralytics.YOLO`` model to an image collection
     -   Applying a :class:`flash:flash.core.model.Task` to an image or video
         collection
 
     Args:
         samples: a :class:`fiftyone.core.collections.SampleCollection`
-        model: a :class:`Model` or :class:`flash:flash.core.model.Task` or
-            `ultralytics.YOLO` model
+        model: a :class:`Model`, ``ultralytics.YOLO`` model, or
+            :class:`flash:flash.core.model.Task`
         label_field ("predictions"): the name of the field in which to store
             the model predictions. When performing inference on video frames,
             the "frames." prefix is optional
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -117,19 +117,11 @@ def apply_model(
             **kwargs,
         )
 
-    try:
-        import ultralytics
-
-        if isinstance(model, ultralytics.YOLO):
-            model = fouu.convert_model(model)
-    except ImportError:
-        pass
+    if _is_ultralytics_model(model):
+        model = fouu.convert_model(model)
 
     if not isinstance(model, Model):
-        raise ValueError(
-            "Model must be a %s or %s; found %s"
-            % (Model, _BASE_FLASH_TYPE, type(model))
-        )
+        raise ValueError("Unsupported model type: %s" % type(model))
 
     if samples.media_type == fom.IMAGE:
         fov.validate_image_collection(samples)
@@ -279,15 +271,22 @@ def apply_model(
         )
 
 
-_BASE_FLASH_TYPE = "flash.core.model.Task"
-
-
 def _is_flash_model(model):
     for cls in inspect.getmro(type(model)):
-        if etau.get_class_name(cls) == _BASE_FLASH_TYPE:
+        if etau.get_class_name(cls) == "flash.core.model.Task":
             return True
 
     return False
+
+
+def _is_ultralytics_model(model):
+    try:
+        # pylint: disable=import-error
+        import ultralytics
+
+        return isinstance(model, ultralytics.YOLO)
+    except:
+        return False
 
 
 def _apply_image_model_single(
@@ -725,8 +724,7 @@ def compute_embeddings(
     **kwargs,
 ):
     """Computes embeddings for the samples in the collection using the given
-    :class:`FiftyOne model <Model>` or
-    :class:`Lightning Flash model <flash:flash.core.model.Task>`.
+    model.
 
     This method supports all the following cases:
 
@@ -789,11 +787,11 @@ def compute_embeddings(
             **kwargs,
         )
 
+    if _is_ultralytics_model(model):
+        model = fouu.convert_model(model)
+
     if not isinstance(model, Model):
-        raise ValueError(
-            "Model must be a %s or %s; found %s"
-            % (Model, _BASE_FLASH_TYPE, type(model))
-        )
+        raise ValueError("Unsupported model type: %s" % type(model))
 
     if not model.has_embeddings:
         raise ValueError(
@@ -1207,7 +1205,7 @@ def compute_patch_embeddings(
     skip_failures=True,
 ):
     """Computes embeddings for the image patches defined by ``patches_field``
-    of the samples in the collection using the given :class:`Model`.
+    of the samples in the collection using the given model.
 
     This method supports all the following cases:
 
@@ -1273,6 +1271,9 @@ def compute_patch_embeddings(
             ``True`` and any errors are detected, this nested dict will contain
             missing or ``None`` values to indicate uncomputable embeddings
     """
+    if _is_ultralytics_model(model):
+        model = fouu.convert_model(model)
+
     if not isinstance(model, Model):
         raise ValueError(
             "Model must be a %s instance; found %s" % (Model, type(model))
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -280,13 +280,7 @@ def _is_flash_model(model):
 
 
 def _is_ultralytics_model(model):
-    try:
-        # pylint: disable=import-error
-        import ultralytics
-
-        return isinstance(model, ultralytics.YOLO)
-    except:
-        return False
+    return type(model).__module__.startswith("ultralytics.")
 
 
 def _apply_image_model_single(
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -34,7 +34,6 @@ foup = fou.lazy_import("fiftyone.utils.patches")
 fout = fou.lazy_import("fiftyone.utils.torch")
 fouu = fou.lazy_import("fiftyone.utils.ultralytics")
 
-ultralytics = fou.lazy_import("ultralytics")
 
 logger = logging.getLogger(__name__)
 
@@ -120,8 +119,10 @@ def apply_model(
         )
 
     try:
+        import ultralytics
+
         if isinstance(model, ultralytics.YOLO):
-            model = fouu._convert_yolo_model(model)
+            model = fouu.convert_model(model)
     except ImportError:
         pass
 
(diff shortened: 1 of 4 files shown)


fiftyone/utils/ultralytics.py
```diff
@@ -270,3 +290,84 @@ def _to_detections(self, df):
...
+class YOLOSegmentationModel(Model):
+    def __init__(self, config):
+
+        self.model = config.get("model", None)
+
+    @property
+    def media_type(self):
+        return "image"
+
+    def predict(self, args):
```
brimoor, line 322: nit: `arg` not `args`. `predict()` is for a single image

fiftyone/utils/ultralytics.py
```diff
@@ -270,3 +290,84 @@ def _to_detections(self, df):
...
+        image = Image.fromarray(args)
+        predictions = self.model(image, verbose=False)
+        return to_instances(predictions[0])
+
+    def predict_all(self, samples, args):
```
brimoor, line 327: Shouldn't this signature just be `def predict_all(self, args):`?

brimoor (changes requested):

fiftyone/core/models.py
```diff
@@ -72,7 +74,8 @@ def apply_model(
 
     Args:
         samples: a :class:`fiftyone.core.collections.SampleCollection`
-        model: a :class:`Model` or :class:`flash:flash.core.model.Task`
+        model: a :class:`Model` or :class:`flash:flash.core.model.Task` or
+            `YOLO` model from `ultralytics
```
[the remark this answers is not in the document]
jacobmarks: Makes sense!

fiftyone/utils/ultralytics.py
```diff
@@ -9,9 +9,15 @@
...
 import numpy as np
+from PIL import Image
 
+import fiftyone.core.utils as fou
 import fiftyone.core.labels as fol
 import fiftyone.utils.torch as fout
+from fiftyone.core.models import Model
+
+ultralytics = fou.lazy_import("ultralytics")
+from ultralytics import YOLO
```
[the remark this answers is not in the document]
jacobmarks: Gotcha. I was just thinking of efficiency in loading/importing, not considering the crash.

fiftyone/core/models.py
```diff
@@ -116,6 +119,25 @@ def apply_model(
             **kwargs,
         )
 
+    try:
+        if isinstance(model, ultralytics.YOLO):
+            model = fouu._convert_yolo_model(model)
+            return apply_model(
```
brimoor, line 125: Lines 125-137 are unnecessary, right?

fiftyone/utils/ultralytics.py
```diff
@@ -270,3 +289,107 @@ def _to_detections(self, df):
...
+    def transforms(self):
+        return None
+
+    @property
+    def preprocess(self):
+        return False
+
+    def predict(self, args):
+        raise NotImplementedError(
+            "Subclass must implement `predict` or `predict_all`"
```
brimoor, line 323: nit: remove `predict_all` from this error message. Subclasses **have** to implement `predict`

fiftyone/core/models.py
jacobmarks: Ah yep, you're right! That's nice indeed.

fiftyone/utils/ultralytics.py
```diff
@@ -270,3 +289,105 @@ def _to_detections(self, df):
...
+
+    @property
+    def preprocess(self):
+        return False
+
+    def predict(self, args):
+        raise NotImplementedError("Subclass must implement `predict`")
+
+    def predict_all(self, args):
+        return self.predict(args)
```
brimoor, line 325: I don't think this actually works as written, as `predict()` calls `Image.fromarray()`which  cannot handle a list of images or `NHWC` tensor.

Since you set `ragged_batches == True` in the FO model definition, I believe `apply_model(..., batch_size>1)` will raise an error complaining that the model does not support batch inference. So it may be fine to just delete this implementation of `predict_all()` and let the default implementation from `fom.Model` be used.

brimoor: One of the reasons I didn't initially write these wrapper classes is that `apply_model()` doesn't currently have a way to know that a model can directly accept filepaths for inference, like Ultralytics model objects support. It currently always loads the images into numpy arrays, possibly using a data loader if it is a Torch model.

In the Ultralytics integration docs, I showed how batch inference can be done like so:
https://docs.voxel51.com/integrations/ultralytics.html#batch-inference

I did not investigate the Ultralytics implementation to see what is happening inside: does it generate a 4D tensor and do a single forward pass (presumably more efficient) or does it just process each image serially (presumably less efficient)? And is it more efficient to pass filepaths directly to the Ultralytics model rather than loading them yourself?

fiftyone/utils/ultralytics.py
```diff
@@ -270,3 +289,105 @@ def _to_detections(self, df):
...
+
+def _convert_yolo_model(model):
+    if isinstance(model.model, ultralytics.nn.tasks.SegmentationModel):
+        return _convert_yolo_segmentation_model(model)
+    elif isinstance(model.model, ultralytics.nn.tasks.PoseModel):
+        return _convert_yolo_pose_model(model)
+    elif isinstance(model.model, ultralytics.nn.tasks.DetectionModel):
+        return _convert_yolo_detection_model(model)
+
+    return model
```
brimoor, line 393: Better to raise an error here saying that we cannot convert this model to an FO `Model`?

fiftyone/utils/ultralytics.py
```diff
@@ -270,3 +289,105 @@ def _to_detections(self, df):
                 for row in df.itertuples()
             ]
         )
+
+
+class FiftyOneYOLOModel(Model):
```
brimoor, line 294: Strictly speaking, the `fom.Model` interface requires that the `config` parameter to be an instance of a subclass of `fiftyone.core.config.Config` whose name is `<ModelClass>Config`.

So it would look more like this:

```py
from fiftyone.core.config import Config
from fiftyone.core.models import Model

class YOLOConfig(Config):
    def __init__(self, d):
        self.model = self.parse_raw(d, "model", default=None)

class YOLO(Model):
    def __init__(self, config):
        self.config = config
        self.model = config.model

config = YOLOConfig({"model": model})
model = YOLO(config)
```

Why do it like this when your current implementation works as written for your use case? The idea in general is that the `Config` class is supposed to be a serializable (ie something that can be expressed as JSON via `config.serialize()`) so that, for example, everything necessary to reload the model in another process can be stored in a file like [manifest-torch.json](https://github.com/voxel51/fiftyone/blob/develop/fiftyone/zoo/models/manifest-torch.json).

However, in this use case, the parameter is a preloaded `model` instance, so in the code above `config.serialize()` would not actually yield something that can be written to a JSON file.

brimoor: tldr, I'm mostly just taking the opportunity to explain the design behind the `fom.Model` interface's "normal" use case.

In this case, I think it's fine to keep `config` as a dict if you want for now because these models are not yet usable in JSON form in either approach.

fiftyone/core/models.py
```diff
@@ -72,7 +74,8 @@ def apply_model(
 
     Args:
         samples: a :class:`fiftyone.core.collections.SampleCollection`
-        model: a :class:`Model` or :class:`flash:flash.core.model.Task`
+        model: a :class:`Model` or :class:`flash:flash.core.model.Task` or
```
brimoor, line 77: @jacobmarks can you also update the docstring of `SampleCollection.apply_model()` to match?

fiftyone/utils/ultralytics.py
brimoor: I wasn't sure if numpy arrays could be passed to Ultralytics models (how would it know if any necessary preprocessing had already been performed?) or if their API *required* filepaths to be provided in general

jacobmarks: Did a quick timing test for YOLO models to see how filepath vs PIL Image and batching impact efficiency.

```
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.brain as fob

import timeit
from PIL import Image
from ultralytics import YOLO

yolov5n = YOLO('yolov5n.pt')

def run_yolov5n_filepath(filepath):
    yolov5n(filepath, verbose=False);

print(timeit.timeit(lambda: run_yolov5n_filepath(dataset.first().filepath), number=100))
## 6.987219458999988

def run_yolov5_image(filepath):
    yolov5n(Image.open(filepath), verbose=False);

print(timeit.timeit(lambda: run_yolov5_image(dataset.first().filepath), number=100))
## 7.138490875000002

def run_yolov5_images(filepaths):
    ims = [Image.open(filepath) for filepath in filepaths]
    yolov5n(ims, verbose=False);

print(timeit.timeit(lambda: run_yolov5_images([dataset.first().filepath]*10), number=10))
## 4.4615429579999955

def run_yolov5_filepaths(filepaths):
    yolov5n(filepaths, verbose=False);

print(timeit.timeit(lambda: run_yolov5_filepaths([dataset.first().filepath]*10), number=10))
## 4.528784499999972

def run_yolov5_numpy(filepaths):
    ims = [np.array(Image.open(filepath)) for filepath in filepaths]
    yolov5n(ims, verbose=False);

print(timeit.timeit(lambda: run_yolov5_numpy([dataset.first().filepath]*10), number=10))
4.5486639160000095
```

Obviously this is just one model size and one image, but roughly speaking it appears that passing in filepaths vs PIL Images vs numpy arrays is rather inconsequential, while utilizing the model's internal batching is important, if/when it is possible.

fiftyone/core/models.py
jacobmarks: Yep :) Done!

fiftyone/utils/ultralytics.py
jacobmarks: Totally agree!

fiftyone/utils/ultralytics.py
jacobmarks: Totally understand. I would love to keep things consistent, especially if we are potentially going to add more integrations like this moving forward. What do you think about this approach:

It looks like YOLO models have a `ckpt_path` property, and this can be used to instantiate the model in the future, with `YOLO(checkpoint_path)` — this works for me. 

We don't want the model to be reloaded from disk just for the sake of serializability, but one option might be to allow the `YOLOConfig` to accept either a `YOLO` model, *or* a `checkpoint_path`, and then override the config's `serialize()` method to only pass along the `checkpoint_path`, after it is converted to absolute filepath. 

It might look something like this:

```
from fiftyone.core.config import Config
from fiftyone.core.models import Model

class YOLOConfig(Config):
    def __init__(self, d):
        self.checkpoint_path = self.parse_raw(d, "checkpoint_path", default=None)
        self.model = self.parse_raw(d, "model", default=None)
        
        if self.model and not self.checkpoint_path:
                 self.checkpoint_path = self.model.ckpt_path
        elif self.checkpoint_path and not self.model:
                self.model = YOLO(self.checkpoint_path)

        self.checkpoint_path = os.path.abspath(self.checkpoint_path)

   def serialize(self, reflective=False):
        """Serializes the object into a dictionary.

        Serialization is applied recursively to all attributes in the object,
        including element-wise serialization of lists and dictionary values.

        Args:
            reflective: whether to include reflective attributes when
                serializing the object. By default, this is False

        Returns:
            a JSON dictionary representation of the object
        """
        d = self._prepare_serial_dict(reflective)
        for a in self.attributes():
             if a == "model":
                continue
            d[a] = _recurse(getattr(self, a), reflective)
        return d

```

It seems like the serializability is encoded here: https://github.com/voxel51/eta/blob/ceb853415d7c468b553b5d49bbcffc1fb6216921/eta/core/serial.py#L406C4-L406C43

fiftyone/utils/ultralytics.py
brimoor: nice! so this should mean that you can just return `ragged_batches == False` in the model's definition and then `apply_model()` will allow a `batch_size > 1` to be provided

fiftyone/utils/ultralytics.py
brimoor: Nice, `checkpoint_path` would be a great alternative to support!

WRT serialization, I was a bit strong with my wording: the most important thing is that the model can be *loaded* from JSON, at least in one possible configuration (`checkpoint_path` fulfills this here, so it's a 💯 addition); it's not actually necessary that all possible configurations actually be _serializable_. 

For example, [`TorchImageModelConfig`](https://github.com/voxel51/fiftyone/blob/acf3a8f886505d852903e320d057057813261993/fiftyone/utils/torch.py#L364) uses the same strategy of providing an optional `model` parameter where you can directly provide a `torch.nn.Module` in the case where you are constructing a Torch model via code. But I didn't bother modifying `serialize()` to manually omit that parameter.

Here's how I'd recommend implementing this (inspired by [TorchImageModel](https://github.com/voxel51/fiftyone/blob/acf3a8f886505d852903e320d057057813261993/fiftyone/utils/torch.py#L771)):

```py
class YOLOConfig(Config):
    def __init__(self, d):
        self.model = self.parse_raw(d, "model", default=None)
        self.checkpoint_path = self.parse_string(d, "checkpoint_path", default=None)

class YOLO(Model):
    def __init__(self, config):
        self.config = config
        self._model = self._load_model(config)

    def _load_model(self, config):
        if config.model is not None:
            return config.model

        return YOLO(config.checkpoint_path)
```

The reason is that the `Config` is _intended_ to be a static record of the input parameters that constructed the model. Any logic about how the parameters may be used is best done within the `Model` class so that you could check `model.config` to see how the model was defined.

Doesn't necessarily have any impact on the behavior of the model, just explaining my idea behind the `Model` + `ModelConfig` design 🤓

fiftyone/utils/ultralytics.py
jacobmarks: Okay, `ragged_batches` is now `False`, and running `apply_model()` works. Is this what you had in mind, or do I need to define the `predict_all()` method as well to get the batching speedups?

brimoor: I'm confused how `predict_all()` could work right now: it just forwards the list of numpy arrays to `predict()`, but does `Image.fromarray()` actually support a list of arrays as input?

jacobmarks: Fixed! Now batching happens properly for `FiftyOneYOLODetectionModel` and `FiftyOneYOLOPoseModel`, while batching is skipped for segmentation models because masks get padded.

fiftyone/utils/ultralytics.py
jacobmarks: Changes implemented :)

jacobmarks:
Ready to roll!

brimoor (approved):
LGTM 🥇 

@jacobmarks note that I updated the integration docs to document that `apply_model()` can now be used. Hopefully we have blogs, etc coming for this integration, but also want to make sure that the official integration docs cover all the important stuff!

PR #3981, opened 2024-01-09
Fix sorting by similarity with MongoDB backend

benjaminpkane:
Fixes sorting by similarity when a using a [MongoDB similarity index ](https://docs.voxel51.com/integrations/mongodb.html) via the App.

### Changes

Due to an implementation detail, similarity search should occur before any other extended view stages are added


sashankaryal (approved):
I see. Looks good!!

PR #3983, opened 2024-01-09
Directly apply transformers models

jacobmarks:
Following the same pattern as https://github.com/voxel51/fiftyone/pull/3977, this PR adds support for applying Hugging Face transformer models to FiftyOne datasets.

In particular:
- Adds `transformers` utils for converting any `transformers.models.<submod1>.<submod2>...ForImageClassification` and any `transformers.models.<submod1>.<submod2>...ForObjectDetection` models to FiftyOne `fo.Model` objects
- Allows these models to be entered directly into a sample collection's `apply_model()`, `compute_embeddings()` and `compute_patch_embeddings()` methods, and will "auto-magically work".

## Example Usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=10)
```

### detection

```py
from transformers import DetrForObjectDetection

detr = DetrForObjectDetection.from_pretrained(
    "facebook/detr-resnet-50", revision="no_timm"
)
dataset.apply_model(detr, label_field="detr", batch_size=10)
```

### classification

```py
from transformers import BeitForImageClassification

beit = BeitForImageClassification.from_pretrained(
    "microsoft/beit-base-patch16-224"
)
dataset.apply_model(beit, label_field="beit")

# embeddings
embeddings = dataset.compute_embeddings(beit)
print(embeddings.shape)
```

### semantic segmentation

```py
from transformers import SegformerForSemanticSegmentation

segformer = SegformerForSemanticSegmentation.from_pretrained(
    "nvidia/segformer-b0-finetuned-ade-512-512"
)
dataset.apply_model(segformer, label_field="segformer")
```

The FiftyOne models generated from `transformers` classification and detection models also expose embeddings, which means that they can be used within `compute_embeddings()` and `compute_patch_embeddings()`. They will also be usable directly in Brain methods such as `compute_similarity()` and `compute_visualization()`.

## Classification models

This should work without any modification for any of the models listed [here](https://huggingface.co/docs/transformers/tasks/image_classification#:~:text=The%20task%20illustrated%20in%20this%20tutorial%20is%20supported%20by%20the%20following%20model%20architectures%3A)

## Detection models

Here is a subset of the detection models it should work for without modification:

- https://huggingface.co/facebook/detr-resnet-50
- https://huggingface.co/facebook/detr-resnet-101
- https://huggingface.co/docs/transformers/model_doc/deta
- https://huggingface.co/docs/transformers/model_doc/conditional_detr
- https://huggingface.co/docs/transformers/model_doc/deformable_detr
- https://huggingface.co/hustvl/yolos-tiny
- https://huggingface.co/hustvl/yolos-small
- https://huggingface.co/docs/transformers/model_doc/table-transformer

## Semantic Segmentation

Here is a subset of the segmentation models it should work for without modification:

- https://huggingface.co/docs/transformers/main/en/model_doc/mask2former
- https://huggingface.co/docs/transformers/main/en/model_doc/maskformer
- https://huggingface.co/docs/transformers/main/en/model_doc/segformer
- https://huggingface.co/docs/transformers/main/en/model_doc/mobilevit
- https://huggingface.co/docs/transformers/main/en/model_doc/mobilenet_v2
- https://huggingface.co/docs/transformers/main/en/model_doc/beit
- https://huggingface.co/docs/transformers/main/en/model_doc/dpt

## Notes

There are of course many ways to extend this. Here are the next steps I would propose:

- Support loading from name with HF's `from_pretrained` via a custom syntax
- Add depth estimation (but only 2 models here so far). They also have `ForDepthEstimation`, which is something cool that we don't have much content for yet, to my knowledge.

        notes for FiftyOne users.

-   [x] Other


brimoor (commented):
@jacobmarks awesome! The example snippets in the PR description work for me 💯 

I went ahead and rebased this branch onto `apply-yolo-models`, so that merge is taken care of.

What do you think about adding a Hugging Face integration page that documents this new superpower, analogous to https://docs.voxel51.com/integrations/ultralytics.html?

fiftyone/core/models.py
```diff
@@ -781,7 +790,9 @@ def compute_embeddings(
             **kwargs,
         )
 
-    if _is_ultralytics_model(model):
+    if _is_transformers_model(model):
```
brimoor, line 793: I went ahead and added this conversion call to `compute_embeddings()` and `compute_patch_embeddings()` so that if our HF model wrappers expose embeddings then these methods will "just work"

fiftyone/utils/transformers.py
```diff
@@ -0,0 +1,270 @@
...
+        return _get_image_processor(self.model)
+
+    def _load_model(self, config):
+        raise NotImplementedError("Subclass must implement _load_model()")
+
+    def predict(self, arg):
+        raise NotImplementedError("Subclass must implement predict()")
+
+
+class FiftyOneTransformerForImageClassification(FiftyOneTransformer):
```
brimoor, line 189: @jacobmarks think you could implement the `fiftyone.core.models.EmbeddingsMixin` interface here so that HF classification models would support embeddings?

Here's how that was implemented for Torch models:
https://github.com/voxel51/fiftyone/blob/d16ed6946b3abaccf0ba1a7c975c25bc1b365df6/fiftyone/utils/torch.py#L247

fiftyone/core/models.py
jacobmarks: Awesome! There are going to be a few subtleties if we want to expose this functionality — which I think we should.

- Currently the conversion requires that the model is a detection of classification model, but for embeddings this is not necessary. [Here](https://huggingface.co/docs/transformers/model_doc/deit#transformers.DeiTModel)'s an example where `DeiTModel` can generate the embeddings, and doesn't need the `ForObjectDetection` etc.
- If we are exposing for embeddings, does this mean we are exposing for `compute_similarity()` as well?

fiftyone/utils/transformers.py
jacobmarks: Absolutely! Can do

jacobmarks:
> @jacobmarks awesome! The example snippets in the PR description work for me 💯
> 
> I went ahead and rebased this branch onto `apply-yolo-models`, so that merge is taken care of.
> 
> What do you think about adding a Hugging Face integration page that documents this new superpower, analogous to https://docs.voxel51.com/integrations/ultralytics.html?

Yep, absolutely I can do this. Will write the integration doc once the code is finalized :)

fiftyone/core/models.py
jacobmarks: Looks like `ultralytics.YOLO` models don't support extracting embeddings so I'm going to remove them from this.

brimoor: It's okay to leave this code in here. Maybe we add support for additional Ultralytics models that do expose embeddings in the future. 

As written, if you pass an `ultralytics.YOLO` model to `compute_embeddings()` it will convert it but then you'll get an error saying that the model doesn't expose embeddings, which is exactly the error we want.

brimoor: Any `fom.Model` that implements `fom.EmbeddingsMixin` will automatically become available for use with `compute_similarity()`, yes.

jacobmarks: Fair enough!

fiftyone/utils/transformers.py
```diff
@@ -0,0 +1,388 @@
...
+    """
+
+    def _load_model(self, config):
+        if config.model is not None:
+            return config.model
+
+        return transformers.AutoModelForSemanticSegmentation.from_pretrained(
+            config.name_or_path
+        )
+        self.mask_targets = config.model.id2label
```
brimoor, line 327: @jacobmarks line 327 is unreachable. Did you intend for this to happen? Or just delete it?

brimoor (changes requested):
@jacobmarks embeddings and semantic segmentation support are looking 🔥 

FYI- I updated the example code in the PR description to show how to use a HF classification model with `compute_embeddings()`; the code runs, but the resulting `embeddings` is a 1D vector when I would expect a 2D array. Is this user error or implementation error?

fiftyone/utils/transformers.py
jacobmarks: I can take another look at `compute_embeddings()`, but in order to make it compatible with batching as well as single samples, and both image embeddings and patch embeddings, I could have sworn this was the best solution

jacobmarks: @brimoor and for `mask_targets`, I don't see a way for `apply_model()` to propagate `mask_targets` from a model to the dataset. I put this here in the hopes that we could figure something out, as I think this would be useful for other segmentation models as well, unless I'm missing something

brimoor: Line 327 doesn't get executed as currently written, it is after the `return` statement

brimoor: Having `model.mask_targets` is totally fine, here's an existing example of that:

```py
dataset.default_mask_targets = model.mask_targets
```

jacobmarks: ahhh I see. Thank you!

jacobmarks:
@brimoor integration docs added :)

PR #3984, opened 2024-01-10
fix missing default/data propagation for operator views

imanjra:

- Fixes an issue where data is unset due to missing default in the output of an operator
- Fixes an issue where the default for the `List` type is not initialized when rendered as non-composite views

Using an example operator with rendering both scenarios above

        notes for FiftyOne users.

See changes proposed above

-   [ ] Other

### DIFF
--- a/app/packages/core/src/plugins/SchemaIO/components/DynamicIO.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/DynamicIO.tsx
@@ -1,9 +1,9 @@
 import { PluginComponentType, useActivePlugins } from "@fiftyone/plugins";
-import { isPrimitiveType } from "@fiftyone/utilities";
-import { get } from "lodash";
+import { isNullish } from "@fiftyone/utilities";
+import { get, isEqual } from "lodash";
 import React, { useEffect } from "react";
-import { getComponent, getErrorsForView } from "../utils";
 import { isPathUserChanged } from "../hooks";
+import { getComponent, getErrorsForView, isCompositeView } from "../utils";
 
 export default function DynamicIO(props) {
   const { data, schema, onChange, path, parentSchema, relativePath } = props;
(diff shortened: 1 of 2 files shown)


app/packages/core/src/plugins/SchemaIO/components/DynamicIO.tsx
```diff
@@ -19,9 +19,10 @@ export default function DynamicIO(props) {
   // todo: need to improve initializing default value in state
   useEffect(() => {
     if (
-      isPrimitiveType(type) &&
-      data !== defaultValue &&
-      !isPathUserChanged(path)
+      !isCompositeView(schema) &&
```
benjaminpkane, line 22: IIUC composite views should not call onChange because their children will?

app/packages/core/src/plugins/SchemaIO/components/DynamicIO.tsx
```diff
@@ -19,9 +19,10 @@ export default function DynamicIO(props) {
...
+      !isEqual(data, defaultValue) &&
+      !isPathUserChanged(path) &&
+      !isNullish(defaultValue)
```
benjaminpkane, line 25: `onChange` should never be called if `defaultValue` is nullish?

benjaminpkane (approved):

app/packages/core/src/plugins/SchemaIO/components/DynamicIO.tsx
imanjra: Yah, that's right. At the moment, the initialization of default value provided by resolved schema is done through onChange on component mount (I do want to optimize this in future)

app/packages/core/src/plugins/SchemaIO/components/DynamicIO.tsx
imanjra: This `onChange` is only used to initialize the default value (i.e. when component initially mount of when default value in the schema is updated externally). When a user makes a change, this flow is not used.

PR #3991, opened 2024-01-12
[TUTORIAL] Monocular Depth Estimation

jacobmarks:
Adds a monocular depth estimation workflow to the tutorials section of the docs! A static pdf of the tutorial in the docs is attached :)

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [x] Documentation: FiftyOne documentation changes
-   [ ] Other

[mde_workflow_static.pdf](https://github.com/voxel51/fiftyone/files/13915090/mde_workflow_static.pdf)

PR #3990, opened 2024-01-11
Integration of YOLO-NAS into FiftyOne Model Zoo #3925

AdonaiVera:
Hi FiftyOne Team 👋 

We're excited to announce the successful integration of YOLO-NAS into the FiftyOne Model Zoo, a collaborative effort with @jacobmarks. This addition enriches our model offerings with advanced vision capabilities from YOLO-NAS.

## Integration Highlights:
YOLO-NAS, sourced from [SuperGradients' GitHub](https://github.com/Deci-AI/super-gradients), is now seamlessly integrated into FiftyOne, offering users its state-of-the-art vision capabilities within the familiar FiftyOne environment.
This integration is designed for smooth compatibility and user-friendly interaction with existing FiftyOne features.

## Interactive Demonstration:
I've developed an informative Jupyter notebook to illustrate YOLO-NAS's functions in FiftyOne, focusing on its practical applications in vision tasks. Check out the [FiftyOne + YOLO-NAS Demo](https://colab.research.google.com/drive/1wKdBfPk66JcZQ_omV4ji_06_FO2EMWyF?usp=sharing) for a hands-on experience.

Besides, you can import the model directly from super_gradients library or Fiftyone zoo models

## Example Usage 

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=10)
```

### Detection with super_gradients

```py
from super_gradients.training import models

model = models.get("yolo_nas_m", pretrained_weights="coco")
dataset.apply_model(model, label_field="yolo-nas", confidence_thresh=0.7)
```

### Detection with FiftyOne zoo model

```py
model = foz.load_zoo_model("yolo-nas-torch)
dataset.apply_model(model, label_field="yolo-nas", confidence_thresh=0.7)
```

## YOLO-NAS Model Variants

The integration includes a variety of YOLO-NAS models, each with its specific application potential, detailed in the demo notebook.

Your insights and suggestions are invaluable. Please feel free to share your thoughts or seek clarification regarding this integration.

Warm regards,
Adonai Vera

PR #3986, opened 2024-01-10
Adding gpu support for openclip models

danielgural:
## What changes are proposed in this pull request?

Added gpu support in the new open_clip models by adding checks for config.device

## How is this patch tested? If it is not, please explain why.

Running openclip model on gpu 

model = foz.load_zoo_model("open-clip-torch", device="cuda")

### Is this a user-facing change that should be mentioned in the release notes?

GPU now supported with OpenClip models

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ X] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

PR #4004, opened 2024-01-15
Cherry-picking ML squad work for v0.23.3

brimoor:
Cherry-picks the following PRs from `develop` onto `release/v0.23.3`:
- https://github.com/voxel51/fiftyone/pull/3977
- https://github.com/voxel51/fiftyone/pull/3983
- https://github.com/voxel51/fiftyone/pull/3986
- https://github.com/voxel51/fiftyone/pull/3990
- https://github.com/voxel51/fiftyone/pull/3991
- https://github.com/voxel51/fiftyone/pull/4003


docs/source/integrations/index.rst
jacobmarks, line 1: I like this change. Thoughts on changing `model-zoo` to `models`?

jacobmarks (dismissed):
LGTM!

One note: I think we're missing batch inference for YOLO-NAS — at least I wasn't able to get it to work for me. Aside from that, everything works as it should. Just made some updates to the transformers docs :)

docs/source/integrations/index.rst
brimoor: 💯 made the change 👍

jacobmarks (approved):
LGTM 👍

PR #4009, opened 2024-01-16
Proxy URL fix

benjaminpkane:
Fixes App router path matching when a `proxy_url` is configured. Adds corresponding unit tests


sashankaryal (approved):
lgtm

PR #4015, opened 2024-01-17
Add dynamic batching to bulk writes

brimoor:
## Change log

- use `get_default_batcher()` inside `bulk_write()`
- removes now-unused `bulk_write_batch_size` config setting
- add optional ability to render progress bars on bulk write operations like `set_values()` and `set_label_values()`

### set_values()

```py
import fiftyone as fo

dataset = fo.Dataset()
_ = dataset.add_samples([fo.Sample(filepath=f"{i}.jpg") for i in range(10000)])

dataset.set_values("int", [1] * len(dataset))
dataset.set_values("int", [1] * len(dataset), progress=True)
```

### set_label_values()

```py
import fiftyone as fo
import fiftyone.zoo as foz

values = {_id: True for _id in dataset.values("predictions.detections.id", unwind=True)}

dataset.set_label_values("predictions.detections.bool", values)
dataset.set_label_values("predictions.detections.bool", values, progress=True)
```

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -2569,6 +2569,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
                 sample.frames.save()
 
         if batcher is not None and batcher.manual_backpressure:
+            # @todo can we infer content size from insert_many() above?
             batcher.apply_backpressure(dicts)
 
         return [str(d["_id"]) for d in dicts]
@@ -2617,7 +2618,11 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
                 d.pop("_id", None)
                 ops.append(InsertOne(d))  # adds `_id` to dict
 
-        foo.bulk_write(ops, self._sample_collection, ordered=False)
+        try:
+            self._sample_collection.bulk_write(ops, ordered=False)
+        except BulkWriteError as bwe:
+            msg = bwe.details["writeErrors"][0]["errmsg"]
+            raise ValueError(msg) from bwe
 
         for sample, d in zip(samples, dicts):
             doc = self._sample_dict_to_doc(d)
@@ -2627,6 +2632,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
                 sample.frames.save()
 
         if batcher is not None and batcher.manual_backpressure:
+            # @todo can we infer content size from bulk_write() above?
             batcher.apply_backpressure(dicts)
 
     def _make_dict(self, sample, include_id=False):
@@ -2641,13 +2647,15 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
 
         return d
 
-    def _bulk_write(self, ops, ids=None, frames=False, ordered=False):
+    def _bulk_write(
+        self, ops, ids=None, frames=False, ordered=False, progress=False
+    ):
         if frames:
             coll = self._frame_collection
         else:
             coll = self._sample_collection
 
-        foo.bulk_write(ops, coll, ordered=ordered)
+        foo.bulk_write(ops, coll, ordered=ordered, progress=progress)
 
         if frames:
             fofr.Frame._reload_docs(self._frame_collection_name, frame_ids=ids)
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -777,6 +777,7 @@ def insert_documents(docs, coll, ordered=False, progress=None, num_docs=None):
                 coll.insert_many(batch, ordered=ordered)
                 ids.extend(b["_id"] for b in batch)
                 if batcher.manual_backpressure:
+                    # @todo can we infer content size from insert_many() above?
                     batcher.apply_backpressure(batch)
 
     except BulkWriteError as bwe:
@@ -786,18 +787,30 @@ def insert_documents(docs, coll, ordered=False, progress=None, num_docs=None):
     return ids
 
 
-def bulk_write(ops, coll, ordered=False):
+def bulk_write(ops, coll, ordered=False, progress=False):
     """Performs a batch of write operations on a collection.
 
     Args:
         ops: a list of pymongo operations
         coll: a pymongo collection
         ordered (False): whether the operations must be performed in order
+        progress (False): whether to render a progress bar (True/False), use
+            the default value ``fiftyone.config.show_progress_bars`` (None), or
+            a progress callback function to invoke instead
     """
+    batcher = fou.get_default_batcher(ops, progress=progress)
+
     try:
-        batch_size = fo.config.bulk_write_batch_size
-        for ops_batch in fou.iter_batches(ops, batch_size):
-            coll.bulk_write(list(ops_batch), ordered=ordered)
+        with batcher:
+            for batch in batcher:
+                batch = list(batch)
+                coll.bulk_write(batch, ordered=ordered)
+                if batcher.manual_backpressure:
+                    # @todo can we infer content size from bulk_write() above?
+                    # @todo do we need a more accurate measure of size here?
+                    content_size = sum(len(str(b)) for b in batch)
+                    batcher.apply_backpressure(content_size)
+
     except BulkWriteError as bwe:
         msg = bwe.details["writeErrors"][0]["errmsg"]
         raise ValueError(msg) from bwe
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -1594,6 +1594,30 @@ def get_default_batcher(iterable, progress=False, total=None):
         )
 
 
+def recommend_batch_size_for_value(value, alpha=0.9, max_size=None):
+    """Computes a recommended batch size for the given value type such that a
+    request involving a list of values of this size will be less than
+    ``alpha * fo.config.batcher_target_size_bytes`` bytes.
+
+    Args:
+        value: a value
+        alpha (0.9): a safety factor
+        max_size (None): an optional max batch size
+
+    Returns:
+         a recommended batch size
+    """
+    # Even if ``fo.config.default_batcher != "size"``, it's still reasonable to
+    # use the size threshold to limit the size of individual requests
+    target_size = fo.config.batcher_target_size_bytes
+    value_bytes = sys.getsizeof(value, 40)  # 40 is size of an ObjectId
+    batch_size = int(alpha * target_size / value_bytes)
+    if max_size is not None:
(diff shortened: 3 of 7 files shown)


swheaton:
I had started this but semi abandoned it since we decided to just require nginx content body size config be increased.
However it's good to have this ability in any case.
I'll take a look shortly.

swheaton:
> I did some quick benchmarking last night and, as suspected, we're well into the "flat" part of the performance curve when deciding between batches of 1000 and 100k bulk writes. The "latency"-based batcher in OSS has served us well for add_samples() for a long time now, so I'm confident that it will serve us well for bulk_write() too.

Agree. It might be kind of annoying for bulk writes (docs are very likely to be smaller than those for add_samples, and I would argue has a stronger expectation of completing faster) that it starts at 1 and scales up rather than some more reasonable starting place.

Re: heuristic - yeah this is what I was mentioning to you the other day. In Teams we can get the real size because there is a single serialization point unlike with what we've seen in pymongo. `str(op)` would be an upper bound I think, except there are some extra default parameters that get thrown in there depending on the op. Should be OK on net

fiftyone/core/dataset.py
```diff
@@ -2617,7 +2618,11 @@ def _upsert_samples_batch(
                 d.pop("_id", None)
                 ops.append(InsertOne(d))  # adds `_id` to dict
 
-        foo.bulk_write(ops, self._sample_collection, ordered=False)
+        try:
+            self._sample_collection.bulk_write(ops, ordered=False)
+        except BulkWriteError as bwe:
+            msg = bwe.details["writeErrors"][0]["errmsg"]
+            raise ValueError(msg) from bwe
```
swheaton, line 2626: good call, I did this also in my local impl so that we can avoid the further batching in `foo.bulk_write()` and not overcomplicate that function.

fiftyone/core/dataset.py
```diff
@@ -2569,6 +2569,7 @@ def _add_samples_batch(
                 sample.frames.save()
 
         if batcher is not None and batcher.manual_backpressure:
+            # @todo can we infer content size from insert_many() above?
```
swheaton, line 2572: What do you mean, we can't do much better than getting the sum of BSON representations here.

fiftyone/core/dataset.py
```diff
@@ -2627,6 +2632,7 @@ def _upsert_samples_batch(
                 sample.frames.save()
 
         if batcher is not None and batcher.manual_backpressure:
+            # @todo can we infer content size from bulk_write() above?
```
swheaton, line 2635: Same as above.

> What do you mean, we can't do much better than getting the sum of BSON representations here.

fiftyone/core/odm/database.py
```diff
@@ -777,6 +777,7 @@ def insert_documents(docs, coll, ordered=False, progress=None, num_docs=None):
                 coll.insert_many(batch, ordered=ordered)
                 ids.extend(b["_id"] for b in batch)
                 if batcher.manual_backpressure:
+                    # @todo can we infer content size from insert_many() above?
```
swheaton, line 780: Same as above.

> What do you mean, we can't do much better than getting the sum of BSON representations here.

fiftyone/core/utils.py
```diff
@@ -1562,7 +1567,7 @@ def get_default_batcher(iterable, progress=False, total=None):
             iterable,
             target_latency=target_latency,
             init_batch_size=1,
-            max_batch_beta=2.0,
+            max_batch_beta=8.0,
```
brimoor, line 1570: @swheaton I bumped this to 8.0 to match the value you chose for the size-based batcher. Agree that it's a nice touch to get up to 1k batches in 4 << 10 steps

fiftyone/core/dataset.py
brimoor: Just hoping/wondering if somehow the request size could be *inferred* from the `insert_many()` call above rather than manually recomputing it here 🤷 

Serializing large docs can be a significant cost

fiftyone/core/odm/database.py
```diff
@@ -786,18 +787,30 @@ def insert_documents(docs, coll, ordered=False, progress=None, num_docs=None):
...
-        batch_size = fo.config.bulk_write_batch_size
-        for ops_batch in fou.iter_batches(ops, batch_size):
-            coll.bulk_write(list(ops_batch), ordered=ordered)
+        with batcher:
+            for batch in batcher:
+                batch = list(batch)
+                coll.bulk_write(batch, ordered=ordered)
+                if batcher.manual_backpressure:
+                    # @todo can we infer content size from bulk_write() above?
+                    # @todo do we need a more accurate measure of size here?
```
swheaton, line 810: As detailed in a mainline comment, I think this is likely to be good enough. Alternatively we could add size of BSON representation of `_filter` and `_doc` together (I think it is going to send the command to MongoDB as BSON not the string repr which will be slightly different), though not a great idea to use protected attributes.

fiftyone/core/utils.py
swheaton, line 1570: Just want to make sure this is intended. 2.0 was what was used originally.

I changed it to 8.0 for the `BSONSizeDynamicBatcher` because I figured we're going to be more confident with the size based estimate than the latency based estimate (less variables), as well as the target batch size will likely need to be much larger than with the latency version. So we'd rather converge on the "proper" size faster and with less calls to MongoDB (going from batch size 1 to 10k takes 6 calls as opposed to 15)

fiftyone/core/utils.py
```diff
@@ -1467,10 +1468,14 @@ def __init__(
...
-        batch_content_size = sum(
-            len(json_util.dumps(obj)) for obj in bsonable_batch
-        )
+    def apply_backpressure(self, batch_or_size):
+        if isinstance(batch_or_size, numbers.Number):
+            batch_content_size = batch_or_size
+        else:
+            batch_content_size = sum(
+                len(json_util.dumps(obj)) for obj in batch_or_size
+            )
```
swheaton, line 1477: If we allow this, we should probably just change the name of this class from `BSONSizeDynamicBatcher` to `ContentSizeDynamicBatcher`

fiftyone/core/dataset.py
swheaton: Ah yes unfortunately not sure that we can, the return result doesn't give much information.
We could sample the docs instead of summing all of them.

But the thinking was that if you need to use the content size batcher it's because you have a higher overhead on each call (Teams use case), and therefore the cost of serializing the docs twice instead of once is dwarfed by the overhead of the call itself so it's sort of negligible.

fiftyone/core/utils.py
swheaton: ok sgtm just wanted to double check

fiftyone/core/dataset.py
brimoor: > But the thinking was that if you need to use the content size batcher it's because you have a higher overhead on each call (Teams use case), and therefore the cost of serializing the docs twice instead of once is dwarfed by the 
overhead of the call itself so it's sort of negligible

Yep agree with this 100%, which is why I left the @todo rather than actually trying to address them now. But if there's minimal hope of achieving the @todo dreams we can just delete.

Adding optional support for random sampling to `ContentSizeDynamicBatcher` would be a nice addition 👍

swheaton:
Oops I pushed here instead of to a new branch, oh well. If the content size limit is indeed something like 1MB, it's also important to limit the individual operation size otherwise even a single op in the batch will be too big.

Check it out - https://github.com/voxel51/fiftyone/pull/4015/commits/bee48c162d190e3b792f544cd603122704d642be

fiftyone/core/dataset.py
swheaton: I'm thinking there's minimal hope but sure we can leave it in, in case someone has an epiphany later (or pymongo adds info to return types)

fiftyone/core/odm/database.py
brimoor: Yeah I think its fine as-is now

fiftyone/core/utils.py
brimoor: I made this change: https://github.com/voxel51/fiftyone/pull/4015/commits/fbccc7d49385c1628917d3d45c65db832ce8d5ad

swheaton:
Another (maybe future) idea, we could "look ahead" to try to default the starting batch size to something more reasonable than 1 in the case that you have like 20k small updates and we can do them all in 2 batches of 10k instead of 7 (1, 8, 64, 512, 4_096, 10_000, 5_319).

brimoor:
> Check it out - [bee48c1](https://github.com/voxel51/fiftyone/commit/bee48c162d190e3b792f544cd603122704d642be)

ah good call, how do you feel about this as an abstraction of that logic? https://github.com/voxel51/fiftyone/pull/4015/commits/9f5f3a4781fbaf1af45dff2f73a35a0e63402356

```py
from bson import ObjectId

fou.recommend_batch_size_for_value(ObjectId())  # 23592
```

swheaton:
> > Check it out - [bee48c1](https://github.com/voxel51/fiftyone/commit/bee48c162d190e3b792f544cd603122704d642be)
> 
> ah good call, how do you feel about this as an abstraction of that logic? [9f5f3a4](https://github.com/voxel51/fiftyone/commit/9f5f3a4781fbaf1af45dff2f73a35a0e63402356)
> 
> ```python
> from bson import ObjectId
> import fiftyone.core.utils as fou
> 
> fou.recommend_batch_size_for_value(ObjectId())  # 23592
> ```

sure

swheaton:
Another use case to consider (similar to edit sample/label tags) where there's a list of individually-large operations is setting an array such as embeddings. I determined the over-the-wire size (Teams) of a single sample update of n=512 embeddings vector is 6500 (unknown amount larger in OSS) - if the static batch size of 10k is used, it will easily be above 50MB, much too large.

(This PR should cover this use case, just documenting).

swheaton (dismissed):
This lgtm as is. There are a couple other nice-to-haves for the size batcher, noted in this PR, but they should not hold up the feature.

brimoor:
> Another use case to consider (similar to edit sample/label tags) where there's a list of individually-large operations is setting an array such as embeddings. I determined the over-the-wire size (Teams) of a single sample update of n=512 embeddings vector is 6500 (unknown amount larger in OSS) - if the static batch size of 10k is used, it will easily be above 50MB, much too large. cc @allenleetc
> 
> (This PR should cover this use case, just documenting).

yep exactly, another reason that bulk writes really needed the full dynamic treatment. This should unlock yet more successful usages of `set_values()` 👍

PR #4016, opened 2024-01-17
Include sample/frame singletons when clearing dataset cache

brimoor:
Primary intended usage here is situations like the following:

```py
dataset = fo.Dataset()
dataset.add_samples(lots_of_samples)

# clear sample cache so that operation below does NOT reload n docs
dataset.clear_cache()

dataset.set_values("field", lots_of_values)
```


swheaton (approved):
Lgtm

PR #4018, opened 2024-01-17
Fix similarity action k requirement

benjaminpkane:
Fixes `k` requirement logic in the App similarity logic


lanzhenw (approved):
the previous useEffect missed the brainConfig dependency. Thanks for simplifying the logic and fix the bug

PR #3821, opened 2023-11-19
Set pyproject.toml

bigstones:
## What changes are proposed in this pull request?

(Please fill in changes proposed in this fix)
I think it easier to configure the environment with poetry.
I created a pyproject.toml file based on my setup.py file. 

## How is this patch tested? If it is not, please explain why.

(Details)

$ poetry shell
$ poetry update
$ poetry install
$ python

    import fiftyone as fo
    import fiftyone.zoo as foz
    dataset = foz.load_zoo_dataset("quickstart")
    session = fo.launch_app(dataset)

### Is this a user-facing change that should be mentioned in the release notes?

-  [ ] No. You can skip the rest of this section.

-  [X] Yes. Give a description of this change to be included in the release

    notes for FiftyOne users.

(Details in 1-2 sentences. You can just refer to another PR with a description

if this PR is part of a larger change.)

If you want to set up a development environment, you can do it easily with poetry.

### What areas of FiftyOne does this PR affect?

-  [ ] App: FiftyOne application changes

-  [X] Build: Build and test infrastructure changes

-  [ ] Core: Core `fiftyone` Python library changes

-  [ ] Documentation: FiftyOne documentation changes

-  [ ] Other

PR #4019, opened 2024-01-17
Move to pyproject.toml package file

benjaminpkane:
Building on top of #3821, this PR moves the main `fiftyone` package description from a setup.py file to the now recommended [pyproject.toml](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) file.

### Changes

[Hatchling](https://pypi.org/project/hatchling/) is now used as the build backend for the project. This requires Python 3.8+, so this change will implicitly drop support for Python 3.7 when installing from source

The OpenCV dependency is still dynamically chosen (see `dependencies.py`) when installing from source. See https://github.com/voxel51/fiftyone/issues/1494

Builds themselves have been improved with the `hatch.toml` file, which is more prescriptive about what files to include. Specifically, the App static files, and documentation recipes and tutorials.

Other packages can be transitioned to `pyproject.toml` descriptions if this change is accepted!


swheaton:
@benjaminpkane appreciate the constant eye toward modernization.
I trust your knowledge of python packaging, mine is limited. Wheel? Egg? setuptools.py?? 😭 

- Are we using hatch as well as hatchling? [Or is hatch just kind of a superset that may include hatchling](https://hatch.pypa.io/latest/why/)?
- Were there any alternatives to hatch/hatchling? Why'd you choose it over others if so? In their docs I only see comparison to setuptools and no other alternatives.
- What potential gotchas do you foresee with this? As we have seen, poetry for example, sounds nice but has some snags we've tripped on over time.

swheaton:
These may or may not be related useful references.
https://sinoroc.gitlab.io/kb/python/packaging_tools_comparisons.html#build-back-ends
https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#modernize-setup-py-project

As to python3.8 requirement, it does not seem so bad since it's only for building from source which doesn't happen so often. We'll still want to float it around a bit before committing though.

benjaminpkane:
I'd be in favor of officially dropping 3.7 with this change. Still need to fix documentation builds, so leaving it as a draft for now.

> @benjaminpkane appreciate the constant eye toward modernization. I trust your knowledge of python packaging, mine is limited. Wheel? Egg? setuptools.py?? 😭
> 
> * Are we using hatch as well as hatchling? [Or is hatch just kind of a superset that may include hatchling](https://hatch.pypa.io/latest/why/)?
> * Were there any alternatives to hatch/hatchling? Why'd you choose it over others if so? In their docs I only see comparison to setuptools and no other alternatives.
> * What potential gotchas do you foresee with this? As we have seen, poetry for example, sounds nice but has some snags we've tripped on over time.

1. We are only using `hatchling` which is a the [build backend](https://peps.python.org/pep-0517/) component of `hatch`. I experimented with `hatch` for developer installs, but seems unnecessary
2. `hatchling` is what the [packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/#choosing-build-backend) currently uses as a default. setuptools is older / more flexible but also harder to maintain. The single `hatch.toml` simplifies the build description.
3. The only gotchas beyond what is described above is ensuring the contents of the distribution (now defined by `hatch.toml`) do not break installations [current v0.23.2 dist](https://pypi.org/project/fiftyone/0.23.2/#files) and this PR's [latest dist artifact](https://github.com/voxel51/fiftyone/actions/runs/7562127204/artifacts/1176570314)

Issue #3975, opened 2024-01-05
[INSTALL] The mongodb binaries aren't downloaded and installed on Archlinux


jakubhejhal:
### System information

-   **OS Platform and Distribution** - `ArchLinux x86_64`:
-   **Python version** - `3.10.13`:
-   **FiftyOne version** - `0.23.2`:
-   **FiftyOne installed from** - `pip`:

### Commands to reproduce

```
$ fiftyone
```

### Describe the problem

The `mongodb` database binaries which are supposed to be shipped with `fiftyone` by the `fiftyone-db` auxilliary package aren't downloaded and installed:

```
$> ls .venv/lib/python3.10/site-packages/fiftyone/db/
__init__.py  __pycache__/
```

Thus running fiftyone fails:

```
$> fiftyone
Traceback (most recent call last):
  File "/home/user/test/.venv/lib/python3.10/site-packages/fiftyone/core/odm/database.py", line 200, in establish_db_conn
    _db_service = fos.DatabaseService()
  File "/home/user/test/.venv/lib/python3.10/site-packages/fiftyone/core/service.py", line 80, in __init__
    self.start()
  File "/home/user/test/.venv/lib/python3.10/site-packages/fiftyone/core/service.py", line 209, in start
    super().start()
  File "/home/user/test/.venv/lib/python3.10/site-packages/fiftyone/core/service.py", line 118, in start
    + self.command,
  File "/home/user/test/.venv/lib/python3.10/site-packages/fiftyone/core/service.py", line 238, in command
    DatabaseService.find_mongod(),
  File "/home/user/test/.venv/lib/python3.10/site-packages/fiftyone/core/service.py", line 286, in find_mongod
    raise ServiceExecutableNotFound("Could not find `mongod`")
fiftyone.core.service.ServiceExecutableNotFound: Could not find `mongod`

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/user/test/.venv/bin/fiftyone", line 5, in <module>
    from fiftyone.core.cli import main
  File "/home/user/test/.venv/lib/python3.10/site-packages/fiftyone/__init__.py", line 25, in <module>
    from fiftyone.__public__ import *
  File "/home/user/test/.venv/lib/python3.10/site-packages/fiftyone/__public__.py", line 16, in <module>
    _foo.establish_db_conn(config)
[...]

PR #4025, opened 2024-01-19
[`fiftyone-db`] Fix /etc/os-release files with empty lines

benjaminpkane:
Resolves an `/etc/os-release` parsing issue mentioned in https://github.com/voxel51/fiftyone/issues/3975#issuecomment-1900421186


findtopher (approved):
works for me!

PR #4044, opened 2024-01-25
Fix multi PCD groups in the modal

benjaminpkane:
Fixes non-default pcd slice rendering in the modal
```py
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-groups", max_samples=3)
dataset.persistent = True
dataset.group_slice = "pcd"
extra = dataset.first().copy()
extra.group.name = "extra"
dataset.add_sample(extra)
```
### Before

https://github.com/voxel51/fiftyone/assets/19821840/ea703ee5-cfd9-4193-a53a-f745a4a866f3

### After

https://github.com/voxel51/fiftyone/assets/19821840/6ab0b721-9a43-4054-af8d-9c0638d3d018


sashankaryal (approved):
👏 for adding tests too

PR #4050, opened 2024-01-29
Update Open CLIP to use parsed Torch device

danielgural:

Fixes when a user does not define a device, sometimes the model will be gpu is some but not all places due to config.device not being updated

Run openclip models on gpu without specifying the gpu. Can check with things like model._using_gpu and checking tensor devices.


danielgural:
Tested and LGTM

PR #4054, opened 2024-01-30
Plugin Utils for getting, setting, and listing views

jacobmarks:
Implements 3 utility functions for plugins:

1. `list_target_views()`: adds target view choices to inputs
2. `get_target_view()`: decodes user's choice from `list_target_views()`
3. `set_view()`: wrapper for `ctx.trigger()` with `set_view` operator, utilizing serialization

Usage:

```py

import fiftyone as fo
from fiftyone.operators import types
import fiftyone.plugins.utils as fopu

class MyOperator(foo.Operator):
    ....
 
    def resolve_input(self, ctx):
       inputs = types.Object()
       ....
       fopu.list_target_views(ctx, inputs)
       ....

    def execute(self, ctx):
       ....
       target_view = fopu.get_target_view(ctx, target)
       ....
       view = ...
       fopu.set_view(ctx, view)
```

        notes for FiftyOne users.

-   [x] Other

### DIFF
--- a/fiftyone/plugins/utils.py
+++ b/fiftyone/plugins/utils.py
@@ -13,6 +13,7 @@ from bs4 import BeautifulSoup
 import yaml
 
 import fiftyone.core.utils as fou
+from fiftyone.operators import types
 from fiftyone.utils.github import GitHubRepository
 from fiftyone.plugins.core import PLUGIN_METADATA_FILENAMES
 
@@ -219,3 +220,51 @@ def _do_get_plugin_info(task):
             spec = gh_repo
 
         logger.debug("Failed to retrieve plugin info for %s: %s", spec, e)
+
+
+def _get_target_view(ctx, target):
+    if target == "SELECTED_SAMPLES":
+        return ctx.view.select(ctx.selected)
+
+    if target == "DATASET":
+        return ctx.dataset
+
+    return ctx.view
+
+
+def _list_target_views(ctx, inputs):
+    has_view = ctx.view != ctx.dataset.view()
+    has_selected = bool(ctx.selected)
+    default_target = "DATASET"
+    if has_view or has_selected:
+        target_choices = types.RadioGroup()
+        target_choices.add_choice(
+            "DATASET",
+            label="Entire dataset",
+            description="Run model on the entire dataset",
+        )
+
+        if has_view:
+            target_choices.add_choice(
+                "CURRENT_VIEW",
+                label="Current view",
+                description="Run model on the current view",
+            )
+            default_target = "CURRENT_VIEW"
+
+        if has_selected:
+            target_choices.add_choice(
+                "SELECTED_SAMPLES",
+                label="Selected samples",
+                description="Run model on the selected samples",
+            )
+            default_target = "SELECTED_SAMPLES"
+
+        inputs.enum(
+            "target",
+            target_choices.values(),
+            default=default_target,
+            view=target_choices,
+        )
+    else:
+        ctx.params["target"] = "DATASET"
--- a/fiftyone/plugins/utils.py
+++ b/fiftyone/plugins/utils.py
@@ -5,11 +5,13 @@ FiftyOne plugin utilities.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+from bson import json_util
 import logging
 import multiprocessing.dummy
 import os
 
 from bs4 import BeautifulSoup
+import json
 import yaml
 
 import fiftyone.core.utils as fou
@@ -268,3 +270,14 @@ def _list_target_views(ctx, inputs):
         )
     else:
         ctx.params["target"] = "DATASET"
+
+
+def _serialize_view(view):
+    return json.loads(json_util.dumps(view._serialize()))
+
+
+def _set_view(ctx, view):
+    ctx.trigger(
+        "set_view",
+        params=dict(view=_serialize_view(view)),
+    )
--- a/fiftyone/plugins/utils.py
+++ b/fiftyone/plugins/utils.py
@@ -225,6 +225,19 @@ def _do_get_plugin_info(task):
 
 
 def get_target_view(ctx, target):
+    """
+    Get the target view for the current context. This is a helper function that
+    returns the view based on the `target` parameter. To be used in conjunction
+    with `list_target_views()`.
+
+    Args:
+        ctx: the current context
+        target: the target view
+
+    Returns:
+        the target view
+
+    """
     if target == "SELECTED_SAMPLES":
         return ctx.view.select(ctx.selected)
 
@@ -235,6 +248,21 @@ def get_target_view(ctx, target):
 
 
 def list_target_views(ctx, inputs):
+    """
+    List the available target views for the current context. This is a helper
+    function that sets the `target` parameter based on the current context. Use
+    it in the `inputs` method of a plugin to allow the user to select the target
+    view from selected samples, the current view, or the entire dataset. To be
+    used in conjunction with `get_target_view()`.
+
+    Args:
+        ctx: the current context
+        inputs: the inputs object
+
+    Returns:
+        None
+
+    """
     has_view = ctx.view != ctx.dataset.view()
     has_selected = bool(ctx.selected)
     default_target = "DATASET"
@@ -277,6 +305,19 @@ def _serialize_view(view):
 
 
 def set_view(ctx, view):
+    """
+    Set the view of the current context. This is a helper function that
+    triggers the `set_view` event. It performs serialization of the view
+    before triggering the event.
+
+    Args:
+        ctx: the current context
+        view: the view to set
+
+    Returns:
+        None
+
+    """
     ctx.trigger(
         "set_view",
         params=dict(view=_serialize_view(view)),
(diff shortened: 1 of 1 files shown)


fiftyone/plugins/utils.py
```diff
@@ -219,3 +222,103 @@ def _do_get_plugin_info(task):
             spec = gh_repo
 
         logger.debug("Failed to retrieve plugin info for %s: %s", spec, e)
+
+
+def get_target_view(ctx, target):
```
ritch, line 227: Instead of this using a string as a param I would prefer it using the enum property you referred to elsewhere.

fiftyone/plugins/utils.py
```diff
@@ -219,3 +222,103 @@ def _do_get_plugin_info(task):
...
+
+        if has_selected:
+            target_choices.add_choice(
+                "SELECTED_SAMPLES",
+                label="Selected samples",
+                description="Run on the selected samples",
+            )
+            default_target = "SELECTED_SAMPLES"
+
+        inputs.enum(
```
ritch, line 293: This creates a `foo.types.Property` - instead of a utility function we should create a type for this. That along side a utility method on the `types.Object` class is how we've made these sorts of utilities in the past. I'd like to keep that consistent.

fiftyone/plugins/utils.py
```diff
@@ -219,3 +222,103 @@ def _do_get_plugin_info(task):
...
+        )
+    else:
+        ctx.params["target"] = "DATASET"
+
+
+def _serialize_view(view):
+    return json.loads(json_util.dumps(view._serialize()))
+
+
+def set_view(ctx, view):
```
ritch, line 307: I had a plan for these built in operators. The usage I want to include is:

```
ctx.ops.set_view(view)
```

In general I would like to organize things around the `ExecutionContext` and `operator.types`.

ritch (changes requested):
Let me know if you want to discuss the changes. I'd also be willing to make them myself.

fiftyone/plugins/utils.py
brimoor: Alternatively, what if we just supported this?

```py
ctx.trigger("set_view", view=view)
```

with the idea that the builtin serialization logic inside the core operator framework would check for `DatasetView` objects and automatically serialize them via `json.loads(json_util.dumps(view._serialize()))`?

jacobmarks: That works for me, but is it okay to just accept a non-serialized object in one specific case?

fiftyone/plugins/utils.py
brimoor: In the [`@voxel51/utils`](https://github.com/voxel51/fiftyone-plugins/blob/6d1de93dee2cd32d04b1f3071bbbb844697c00c5/plugins/utils/__init__.py#L1708) plugin it also made sense for me to include `SELECTED_LABELS`:

```py
def _get_target_view(ctx, target):
    if target == "SELECTED_LABELS":
        return ctx.view.select_labels(labels=ctx.selected_labels)

    if target == "SELECTED_SAMPLES":
        return ctx.view.select(ctx.selected)

    if target == "DATASET":
        return ctx.dataset

    return ctx.view
```

The way I see it, there's two common patterns:
1. The plugin only cares about media, in which case the only relevant choices are `DATASET/VIEW/SELECTED_SAMPLES`
2. The plugin cares about labels, in which case the relevant choices also include `SELECTED_LABELS`

So IDK what the best way to build a reusable utility is that covers both cases.

fiftyone/plugins/utils.py
```diff
@@ -219,3 +222,103 @@ def _do_get_plugin_info(task):
...
+    """
+    has_view = ctx.view != ctx.dataset.view()
+    has_selected = bool(ctx.selected)
+    default_target = "DATASET"
+    if has_view or has_selected:
+        target_choices = types.RadioGroup()
+        target_choices.add_choice(
+            "DATASET",
+            label="Entire dataset",
+            description="Run on the entire dataset",
```
brimoor, line 274: A couple things that prevented me from building a utility like this myself:
1. I found myself preferring slightly different descriptions for each option depending on what operation is being performed
2. I wanted to change descriptions from this UI element to indicate whether the user is processing a dataset or a view
3. I need access to the chosen target view so I can do things like getting schemas on the correct collection (although this particular point could be addressed by just calling `get_target_view()` immediately after `list_target_views()`)

For example, here's a more sophisticated usage of target views that I wrote:
https://github.com/voxel51/fiftyone-plugins/blob/6d1de93dee2cd32d04b1f3071bbbb844697c00c5/plugins/utils/__init__.py#L1525

When considering writing a utility like this, I though to myself: well, if I'm calling a utility but then have to specify which view types to choose between and also the descriptions for each of them, how much effort am I really saving the plugin developer here?

Food for thought 🍪

fiftyone/plugins/utils.py
```diff
@@ -5,14 +5,17 @@
...
 import logging
 import multiprocessing.dummy
 import os
 
 from bs4 import BeautifulSoup
+import json
 import yaml
 
 import fiftyone.core.utils as fou
+from fiftyone.operators import types
```
brimoor, line 18: These are all operator-specific utilities, so if we go ahead with any of them, I think they belong in `fiftyone.operators.utils` instead.

fiftyone/plugins/utils.py
ritch: I don't want `trigger()` to be more magical. `ctx.ops` is a great place for these things (eg preparing to a `ctx.trigger()` call). It also removes the "set_view" magic string.

fiftyone/plugins/utils.py
brimoor: Ah yeah, I think `types.Object` is actually the right tool for the job here 💡

fiftyone/plugins/utils.py
ritch: No need to re-invent the wheel. A new built in type and utility would make all of this possible.

```py
# override the defaults
inputs.target_view_selection("target_view", label="Select a View", description="Custom description", other_option="something else")
# or just use the defaults
inputs.target_view_selection("target_view")
```

fiftyone/plugins/utils.py
ritch: Please do not create a utils module, instead as I've suggested elsewhere, update the `ExecutionContext` and `types` following the same patterns that already exist.

fiftyone/plugins/utils.py
brimoor: In the simplest case it might look like this:


```py
inputs.target_view(
    "target_view",
    label="Target view",
    description="Choose what samples to process",
)

target_view = ctx.params["target_view"]
```

And if you wanted to customize labels/descriptions and which types of views are allowed, perhaps it would look like this:

```py
from fiftyone.operators import types

choices = types.RadioGroup()
choices.add_choice(
    types.ViewTypes.DATASET,
    label="Entire dataset",
    description="Generate thumbnails for the entire dataset",
)
choices.add_choice(
    types.ViewTypes.VIEW,
    label="Current view",
    description="Generate thumbnails for the current view",
)
choices.add_choice(
    types.ViewTypes.SELECTED_SAMPLES,
    label="Selected samples",
    description="Generate thumbnails for the selected samples",
)

inputs.target_view(
    "target_view",
    label="Target view",
    description="Choose what samples to process",
    choices=choices,
)

# Allow access to target enum so plugin dev can conditionally do stuff below based on the specific choice that the user made?
d = ctx.params["target_view"]
assert type(d["view"]) == fo.DatasetView
assert type(d["target"]) == types.ViewTypes.SELECTED_SAMPLES
```

Starts to get a bit verbose, eh?

ritch:
Here's the proposed usage that I'd like to see:

```py
from fiftyone.operators import types

class MyOperator(foo.Operator):
    def resolve_input(self, ctx):
       inputs = types.Object()
       inputs.select_view(ctx)

    def execute(self, ctx):
       selected_view = ctx.get_selected_view()
       ctx.ops.set_view(view)
```

fiftyone/plugins/utils.py
brimoor: I think that `set_view()` in particular may in fact be best suited on `ExecutionContext`, but potentially there could be other things that are not important enough to clutter the `ExecutionContext` but still merit inclusion in a util file somewhere

fiftyone/plugins/utils.py
brimoor: Isn't `inputs.select_view` referenced [here](https://github.com/voxel51/fiftyone/pull/4054#issuecomment-1918056225) actually shorthand for this?

```py
inputs.add_property("target_view", prop)
```

and therefore the way you should access the corresponding parameter would be:

```py
target_view = ctx.params["target_view"]
```

?

brimoor: Here's a similar but different use case:
https://github.com/voxel51/fiftyone-plugins/blob/6d1de93dee2cd32d04b1f3071bbbb844697c00c5/plugins/io/__init__.py#L1069

where the src collection may be the current dataset, current view, or another dataset (or potentially a view into another dataset).

Perhaps that's too niche and doesn't need to be covered by any builtin op here; just pointing out the use case.

ritch:
Closed in favor of https://github.com/voxel51/fiftyone/pull/4076

PR #4068, opened 2024-02-06
dynamic groups bug fixes

sashankaryal:
This PR contains a number of dynamic groups bug fixes and improvements.
1. Only render dynamic groups configuration options when relevant
2. The "suspend" logic previously implemented in `sidebarSampleId` was causing a lot of stray aggregations query to fire, and it has been converted into something simpler. This makes imavid looker feel much more smooth.
3. Reset dynamic groups view mode to pagination after coming back from video mode.
4. Fix e2e.

## Testing
All groups e2e tests are green locally.


benjaminpkane:
e2e ran to completion 🆒 🤔

benjaminpkane (approved):
Tested. LGTM! I like the aggregation calls change 🚀

benjaminpkane (approved):

PR #4070, opened 2024-02-07
Bugfix: YOLO contour labels were boxes of inadequate shape

sergiev:
contour label support

- Another common YOLO label format is not boxes as `x y w h`, but polygons as `x1 y1 ... xn yn`. 
- Before this patch, they were misinterpreted, so user (me in particular 😬 ) may stuck with such questions as "why evaluation scores are in the neighbourhood of zero" or "why my GT labels have confidence" or "why groundtruth labels are not fitting in the image"
- This patch reduces such stuckness: polygon labels still will be converted to boxes, but as proper envelope, so you can use them in visualization, evaluation, et cetera.

On custom datasets with both kinds of labels.

        notes for FiftyOne users.

I think this change deserves a little mention that users don't need change their labels outside of the library to reach compatibility with it.

-   [x] Other

### DIFF
--- a/fiftyone/utils/yolo.py
+++ b/fiftyone/utils/yolo.py
@@ -9,8 +9,8 @@ import logging
 import os
 import warnings
 
-import yaml
 import numpy as np
+import yaml
 
 import eta.core.utils as etau
 
@@ -27,7 +27,7 @@ def add_yolo_labels(
     label_field,
     labels_path,
     classes,
-    include_missing=False
+    include_missing=False,
 ):
     """Adds the given YOLO-formatted labels to the collection.
 
@@ -1025,14 +1025,13 @@ def load_yolo_annotations(txt_path, classes):
     Args:
         txt_path: the path to the annotations TXT file
         classes: the list of class label strings
-        
+
     Returns:
         a :class:`fiftyone.core.detections.Detections`
     """
     detections = []
     for row in _read_file_lines(txt_path):
         detection = _parse_yolo_row(row, classes)
-        c = detection.confidence
         detections.append(detection)
 
     return fol.Detections(detections=detections)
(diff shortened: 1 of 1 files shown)


fiftyone/utils/yolo.py
```diff
@@ -1082,24 +1083,30 @@ def _get_yolo_v5_labels_path(image_path):
...
+    confidence = None
+    if len(row_vals) < 7: # box case
+        xc, yc, w, h = map(float, row_vals[1:5])
+        minx = xc - 0.5 * w
+        miny = yc - 0.5 * h
+
+        if len(row_vals) > 5:
+            confidence = float(row_vals[5])
+
+    else: # polygon case
```
brimoor, line 1096: @sergiev since this case is polygons, we could just return `fo.Polylines()` instances here instead of `fo.Detections()` and then there would be no data loss at all

brimoor (changes requested):
@sergiev thanks for the PR! I didn't realize people stored polygons in YOLO format. FiftyOne [supports polylines](https://docs.voxel51.com/user_guide/using_datasets.html#polylines-and-polygons) so would it be more appropriate to just return `Polylines` rather than `Detections` in this case so there's no data loss for the user?

fiftyone/utils/yolo.py
sergiev: Thank you sincerely, didn't know that! I'll rewrite this part to polylines and post update here soon after the test.

sergiev: @brimoor sorry, but looks like Polylines and Detections differences too big to fit the scope of this PR.
When I return Polylines instead of Detections the non-interchangeability comes clear.
image
I'll be happy to fix it, but need guidance: which level (module, file, method) will be the best place to propose such changes?

brimoor (approved):
@sergiev thanks for looking into migrating to `Polylines` here; I'll go ahead and merge the PR as-is and finish the `Polylines` support as it will require a couple small changes in other places

PR #4073, opened 2024-02-07
Fix video, and video slice support for Labelbox

ehofesmann:
Resolves https://github.com/voxel51/fiftyone/issues/1654

Reintroduces support for video datasets and adds support for video slices of grouped datasets. Tested classification and detection annotation runs with flat video, grouped video with a selected slice, flat image, and grouped image with a selected slice datasets.

Additionally fixes a bug when annotating checkbox list attributes of the form:
```python
label_schema = {
    "field": {
        "type": "detections",
        "classes": ["TEST"],
        "attributes": {
            "type": "checkbox",
            "values": ["testval"],
        }
    }
}
```

### DIFF
--- a/fiftyone/utils/labelbox.py
+++ b/fiftyone/utils/labelbox.py
@@ -151,7 +151,7 @@ class LabelboxBackend(foua.AnnotationBackend):
 
     @property
     def supports_video_sample_fields(self):
-        return False  # @todo resolve FiftyOne bug to allow this to be True
+        return True
 
     @property
     def requires_label_schema(self):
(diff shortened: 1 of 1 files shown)


fiftyone/utils/labelbox.py
```diff
@@ -641,8 +644,13 @@ def download_annotations(self, results):
 
         project = self._client.get_project(project_id)
         labels_json = self._download_project_labels(project=project)
-        is_video = results._samples.media_type == fomm.VIDEO
-
+        is_video = (results._samples.media_type == fomm.VIDEO) or (
+            results._samples.media_type == fomm.GROUP
+            and results._samples.group_media_types[
```
benjaminpkane, line 649: I think `results._samples.group_media_types` will be `None` for dynamic groups. But `_parent_media_type` will be defined.

Not sure if dynamic groups should be supported. There is also dynamically grouped groups dataset 😅

ehofesmann: Good catch! I should have been more clear, but annotating any kind of grouped datasets themselves is NOT supported today: https://github.com/voxel51/fiftyone/blob/21ff9a49a93115210cd5e4d34a8653c3ad0b3c24/fiftyone/utils/labelbox.py#L113

This is specifically if a slice has been selected from a group dataset that has had `select_group_slices()` called on it, or some other way to get an Image/Video view out of the group.

benjaminpkane: Ok, and this happens only on download 👍  💂

PR #4076, opened 2024-02-09
Add target view utilities to operators

ritch:
Enables this usage for adding the ability to select the source of the "target view`. This is a pattern that exists in many plugins.

```py
# common usage
class MyOperator(foo.Operator):
    def resolve_input(self, ctx):
       inputs = types.Object()
       inputs.select_view(ctx)

    def execute(self, ctx):
       selected_view = ctx.target_view()
       ctx.ops.set_view(view)

# advanced usage
view_target_property = ViewTargetProperty(ctx, view_type=Dropdown)
view_target_property.choices.entire_dataset.label = "My Custom Label"
view_target_property.choices.selected_samples.include = False # hide an option
inputs.define_property("view_target", view_target_property)
view = ctx.target_view("view_target")
```

Tested via [this example operator](https://github.com/voxel51/fiftyone-plugins/pull/108/files#diff-836176c6defe9a0b797ef30b54545233875a1ef48cbef6c5a3d28d45f76adc64R796).

### DIFF
--- a/fiftyone/operators/executor.py
+++ b/fiftyone/operators/executor.py
@@ -508,12 +508,25 @@ class ExecutionContext(object):
         """The target :class:`fiftyone.core.view.DatasetView` for the operator
         being executed.
         """
-        target = self.request_params.get(param_name, None)
+        target = self.params.get(param_name, None)
         if target == "SELECTED_SAMPLES":
             return self.view.select(self.selected)
         if target == "DATASET":
             return self.dataset
         return self.view
+
+    @property
+    def has_custom_view(self):
+        """Whether the operator has a custom view."""
+        stages = self.request_params.get("view", None)
+        filters = self.request_params.get("filters", None)
+        extended = self.request_params.get("extended", None)
+        has_stages = stages is not None and stages != [] and stages != {}
+        has_filters = filters is not None and filters != [] and filters != {}
+        has_extended = (
+            extended is not None and extended != [] and extended != {}
+        )
+        return has_stages or has_filters or has_extended
 
     @property
     def selected(self):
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
Tested. LGTM

fiftyone/operators/executor.py
```diff
@@ -501,6 +504,30 @@ def view(self):
 
         return self._view
 
+    def target_view(self, param_name="view_target"):
```
imanjra, line 507: Should we call it `param_name="VIEW"` for consistency?

fiftyone/operators/types.py
```diff
@@ -723,7 +757,11 @@ class Choices(View):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.choices = kwargs.get("choices", [])
```
imanjra, line 726: I wonder if this is a breaking change

fiftyone/operators/executor.py
ritch: Thats the value - this is choosing which param name - `ctx.params[param_name]`

fiftyone/operators/types.py
ritch: I tried to make this entirely backwards compatible - a choice is only included now if it has a `include = True` - which is the default.

fiftyone/operators/types.py
```diff
@@ -723,7 +757,11 @@ class Choices(View):
...
+        self._choices = kwargs.get("choices", [])
+
+    @property
+    def choices(self):
```
ritch, line 763: This replaces the old `choices` attribue

fiftyone/operators/executor.py
imanjra: Ohh, yah true. Never mind 🙈

PR #4079, opened 2024-02-12
docs for cloud client extra kwargs

swheaton:
Documenting extra kwargs allowed to be passed into cloud provider clients for Teams users.


benjaminpkane (approved):
lgtm!

swheaton:
Think it missed the cut, I'll wait for the next one.
Anyone that needs this can easily be instructed how to use it (only 1 known case so far).

PR #4094, opened 2024-02-20
Adding support for YOLO contours

brimoor:
Finishes the work started in https://github.com/voxel51/fiftyone/pull/4070 to add formal support for YOLO contour labels


benjaminpkane (approved):
lgtm

PR #4097, opened 2024-02-21
Fix pinning sidebar filters for group datasets

benjaminpkane:

Adding sidebar label filters to the current view in group datasets currently omits label filtering that one might expect. This was due to filtering occurring before group slices were selected, causing the filtering rules to be overwritten by the group lookup stage

https://github.com/voxel51/fiftyone/assets/19821840/3486169f-60de-4a93-8278-a9bf70ab19c3

In progress

## Release Notes

* Fixed label filtering after adding sidebar filters to the current view for group datasets

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced filtering capabilities in the data view interface.
- **Refactor**
	- Improved the logic for setting slice views based on conditions related to form parameters.
	- Refactored the handling of sample filter groups for better selection of group slices and IDs.
	- Updated the creation of saved views to exclude unnecessary sample filter conditions.
- **Tests**
	- Updated unit tests to accommodate changes in view functions and optimizations.


fiftyone/core/stages.py
```diff
@@ -941,8 +942,8 @@ def _get_meta_filtered_fields(sample_collection, meta_filter, frames=False):
             base, leaf = key.split(".", 1)
             info_filter[leaf] = val
 
-    matcher = (
-        lambda q, v: q.lower() in v.lower()
+    matcher = lambda q, v: (
```
swheaton, line 945: I find this lambda incredibly confusing. I recommend refactoring to a named function instead
```python

  def matcher(query, value):
    if isinstance(query, str):
      if isinstance(value, str):
        return query.lower() in value.lower()
      elif isinstance(value, dict):
        return query.lower() in str(value).lower()
    return query == value
```

fiftyone/core/stages.py
```diff
@@ -951,8 +952,8 @@ def _get_meta_filtered_fields(sample_collection, meta_filter, frames=False):
         )
     )
 
-    type_matcher = (
-        lambda query, field: (
+    type_matcher = lambda query, field: (
```
swheaton, line 955: Same, I find this lambda incredibly confusing. I recommend refactoring to a named function instead

```python
  def type_matcher(query, field):
    if isinstance(query, str):
      if isinstance(field, EmbeddedDocumentField):
        return type(field.document_type).__name__ == query or field.document_type.__name__ == query
      else:
        return type(field).__name__ == query
    elif isinstance(field, EmbeddedDocumentField):
      return isinstance(field.document_type, query) or field.document_type == query
    else:
      return isinstance(field, query)
```

fiftyone/server/aggregations.py
```diff
@@ -132,7 +132,26 @@ async def aggregate_resolver(
...
+            extended_stages=form.extended_stages,
+            sample_filter=SampleFilter(
+                group=(
+                    GroupElementFilter(
+                        id=form.group_id, slice=form.slice, slices=[form.slice]
+                    )
+                    if not form.sample_ids
+                    else None
+                )
+            ),
```
swheaton, line 150: did you intend for slice view to not be awaitable like above view is?

fiftyone/server/aggregations.py
```diff
@@ -132,7 +132,26 @@ async def aggregate_resolver(
         awaitable=True,
     )
 
-    slice_view = view if form.mixed and "" in form.paths else None
+    slice_view = (
+        fosv.get_view(
```
swheaton, line 136: this is almost exactly the same as the call above. 1) is that right? 2) can we re-use it somehow to remain DRY?

fiftyone/server/view.py
```diff
@@ -121,18 +121,43 @@ def run(dataset):
...
         if sample_filter is not None:
             if sample_filter.group:
-                if sample_filter.group.slice:
-                    view.group_slice = sample_filter.group.slice
-
-                if sample_filter.group.id:
-                    view = fov.make_optimized_select_view(
-                        view, sample_filter.group.id, groups=True
-                    )
+                unselected = all(
```
swheaton, line 124: large code block here, would appreciate some comments as I don't know exactly what's happening - let alone future me.

swheaton (changes requested):

fiftyone/core/stages.py
benjaminpkane: The diff here and the one above are only due to formatting. I can update, but it has no affect on the focus of the PR

swheaton: I know. Your choice on whether you want to refactor in this PR or not. Just calling it out.

fiftyone/server/aggregations.py
benjaminpkane: It should be awaitable. Thanks 👍

swheaton (approved):
lgtm after changes

PR #4100, opened 2024-02-21
Update Label Studio documentation for local storage feature

ehofesmann:
Documents https://github.com/voxel51/fiftyone/pull/3969. To be merged alongside that PR.

image

PR #3969, opened 2024-01-05
allow local files in LS

bonlime:
## What changes are proposed in this pull request?

Closes #3911 
Implements support for LS local storage

## How is this patch tested? If it is not, please explain why.
I couldn't create any way to test rather than manual....

### Is this a user-facing change that should be mentioned in the release notes?

-   [ ] No. You can skip the rest of this section.
-   [x] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

Label Studio integration now support using local storage

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [x] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

PR #4103, opened 2024-02-24
Cherry-picking Label Studio local storage for release

brimoor:
Cherry-picks https://github.com/voxel51/fiftyone/pull/3969 and https://github.com/voxel51/fiftyone/pull/4100 for release.


swheaton (approved):
rubber stamp since eric already approved

PR #4105, opened 2024-02-26
Allow infinitely iterable batchers instead of over a fixed iterable

swheaton:

Allow the batching utilities to be an infinite iterator rather than creating batches over a fixed iterable.
This is useful for reusing the backpressure/adjustment mechanism within the dynamic batchers.

Unit tests added. Tested within Teams context also, passing next batch size to server based on perceived size of current batch.


brimoor (approved):
LGTM

PR #4113, opened 2024-02-28
[BUGFIX] fix bug with operator target view utility when no view or selection

swheaton:

When there's no view or sample selection, the target view utility creates a radio group with no values. This results in the error in screenshot below because the default of `DATASET` isn't an enum in the radio group.

Proposed fix is to always put `DATASET` into the enum of options.
This will cause the radio group to appear with only one option though which is silly. So if there is only 1 option, we make it a `HiddenView`.

Also did some code tidying while I was here.

image

1. Use an operator that uses target view utility such as jacobmarks/image-quality-issues compute aspect ratio operator.
2. No view or selection: no target view input exists but execution is allowed and occurs over the entire dataset.
3. View or selection: target view input exists and offers entire dataset or view/selection - execution is allowed and occurs on the appropriate sample collection based on input selection.

### DIFF
--- a/fiftyone/operators/types.py
+++ b/fiftyone/operators/types.py
@@ -1722,27 +1722,33 @@ class ViewTargetProperty(Property):
     """
 
     def __init__(self, ctx, view_type=RadioGroup, **kwargs):
-        has_custom_view = ctx.has_custom_view
-        has_selected = bool(ctx.selected)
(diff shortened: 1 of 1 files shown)


benjaminpkane (approved):
Tested before/after. LGTM. Nice cleanup

PR #4115, opened 2024-02-29
Include indexes when cloning datasets/views

brimoor:
`dataset.clone()` includes saved views and brain runs, so for consistency it would make sense that it would also include any custom indexes too (users have asked for this behavior). Now it does.


swheaton (approved):
lgtm.
Independently tested, including select-fields and exclude-fields view clones.

PR #4119, opened 2024-03-01
More doc typos

swheaton:

Ran [codespell](https://github.com/codespell-project/codespell) on python and docs source to find more typos. Could be some false negatives left, but this should find most typos.
In the future we can add codespell as a precommit or github action check, which we have already done in other repos.

No testing, all comments and docs.


findtopher (approved):
🔤

brimoor (approved):
Good grief, please add codespell as a pre-commit hook! 😬

swheaton:
I can! The list of exceptions gets a little long due to the fo* convention and some other random things, but it's not too bad.

PR #4123, opened 2024-03-04
Replace 'python setupy.py' with 'python -m build' for better compatibility with modern installers

### LINKED ISSUE #4117: [BUG] fiftyone-db cannot be installed anymore with poetry and pip 24
### Describe the problem

Realted issue on pypa : https://github.com/pypa/pip/issues/12550

### Code to reproduce issue

```console
pip wheel --no-cache-dir --use-pep517 fiftyone-db
```

will fail on linux and pip version 24.

This is most likely related with the fact that no pre-built wheel exists for linux. pip will then pull the source and try to build it itself. However, pip's default build creates a virtual environment solely for the build, which no longer includes `wheel` since version 24, while `fiftyone-db`'s setpu.py does import it. See related PR that was included in version 24 : https://github.com/pypa/pip/pull/12449

I asked pip devs how to solve the issue, but a working (albeit not necessarily the cleanest) solution is to add a small `pyroject.toml` file at the root of the `db` folder with the following info :

```toml
[build-system]
# Minimum requirements for the build system to execute.
requires = ["setuptools", "wheel"]  # PEP 508 specifications.
```

Unless something better is suggested, I can make a PR to add it

### System information

-   **OS Platform and Distribution**: ubuntu 22.04
-   **Python version** (`python --version`): 3.11
-   **FiftyOne version** (`fiftyone --version`): N/A

### Willingness to contribute

FiftyOne codebase?

-   [X] Yes. I can contribute a fix for this bug independently
        from the FiftyOne community

ClementPinard:
Fixes #4117

Each mention of the `python setup.py sdist` or `python setup.py bdist-wheel` is replaced with `python -mI build`. The rationale is that with the `build` package, we can construct the sdist and the wheel. If we wanted the wheel, we're good, if we wanted the sdist, the wheel building is a way to test that the sdist is sufficiently self contained to be built in an isolated environment.

Additionally, since this new test would make the the `fiftyone-db` fail, a `pryproject.toml` file was added to include `wheel` to `build-requires`, since it's not included by default anymore (see https://github.com/pypa/pip/pull/12449 )

Tested directly on github actions
More info to be added soon here

        notes for FiftyOne users.

Build packages (wheels and dist) with `build` package instead of calling `setup.py` directly, for more compatibility with modern installers, especially pip>=24

-   [X] Build: Build and test infrastructure changes
-   [ ] Other


ClementPinard:
@benjaminpkane Here is an implementation proposition of the fix proposed in #4117 , feel free to comment or to make another PR if you prefer to do it another way.

benjaminpkane:
Thanks @ClementPinard! These changes look correct to me! 🙌  I will test things locally and try to debug that issue as well

ClementPinard:
Finally figured out the way to go. Not really well documented, but to pass build options like `--plat-name=whatever` you need to pass the option `-C="--build-option=--plat-name=whatever"` instead of `-C="--plat-name=whatever` See the related code : https://github.com/pypa/setuptools/blob/dbc6471dbacb5a43ed60625cda8b9fb4d8f0aa34/setuptools/build_meta.py#L283

You can rerun the CI and it should succeed. See the corresponding CI in my own repo : https://github.com/ClementPinard/fiftyone/pull/1

ClementPinard:
So the ci is failing but it looks like it's a credentials problem for building the docs rather than anything else, so I suppose it's not related to these changes.

Don't hesitate to ping if you want me to apply a fix for this

benjaminpkane:
Thanks @ClementPinard! The failure in `build-docs` is expected. Would you mind just removing the `test-envs` directory altogether? They are no longer used, might as well remove them now.

Dockerfile
```diff
@@ -88,7 +88,7 @@ RUN apt -y update \
 #   pydicom: DICOM images
 #
 
-RUN pip --no-cache-dir install --upgrade pip setuptools wheel ipython
+RUN pip --no-cache-dir install --upgrade pip setuptools wheel build ipython
```
benjaminpkane, line 91: I don't believe `build` is necessary here, as it expects a wheel is already build in `dist`
```suggestion
RUN pip --no-cache-dir install --upgrade pip setuptools wheel ipython
```

.github/workflows/build-db.yml
```diff
@@ -43,27 +43,27 @@ jobs:
...
         run: |
           cd package/db
-          FODB_MACHINE=x86_64 FODB_SYSTEM=Windows python setup.py bdist_wheel
+          FODB_MACHINE=x86_64 FODB_SYSTEM=Windows python -Im build
       - name: Build sdist
         if: ${{ matrix.platform == 'sdist' }}
         run: |
           cd package/db
-          python setup.py sdist
+          python -Im build
```
benjaminpkane, line 66: Small detail, we only want `sdist` built here so linux machines are able to download the correct MongoDB.
```suggestion
          python -Im build --sdist # linux machines must build from an sdist
```

ClementPinard:
> Thanks @ClementPinard! The failure in `build-docs` is expected. Would you mind just removing the `test-envs` directory altogether? They are no longer used, might as well remove them now.

Do I also need to remove the mention of it in the README ?
https://github.com/voxel51/fiftyone/tree/develop/package#testing-on-other-linux-distributions

benjaminpkane:
> > Thanks @ClementPinard! The failure in `build-docs` is expected. Would you mind just removing the `test-envs` directory altogether? They are no longer used, might as well remove them now.
> 
> Do I also need to remove the mention of it in the README ? https://github.com/voxel51/fiftyone/tree/develop/package#testing-on-other-linux-distributions

Yes, that section can be removed

.github/workflows/build-db.yml
benjaminpkane: Yes, the smoke test is valuable. The only thing that can't happen is an upload of a Linux wheel. In the latest `fiftyone-db` [release](https://pypi.org/project/fiftyone-db/#files) you can see that Mac and Windows wheels are published, but not Linux. This is required so various Linux distributions (e.g. Ubuntu, RHEL, etc.) are supported.

So passing `--sdist` avoids building the Linux wheel. It is also ok to omit `--sdist` to smoke test that the Linux wheel can be built. But we just don't want to publish it to PyPI in the final [workflow step](https://github.com/voxel51/fiftyone/blob/233f7c6c87c7caf5f16f5445fd75483b02c37f0e/.github/workflows/build-db.yml#L120)

ClementPinard: Indeed you are right, it would upload the created wheel as well. Also, the smoke test is actually already done for other distributions, so if the sdist build is not good, the other builds will fail.
Done

webknjaz: FWIW you should look into `cibuildwheel` — it is able to cover all platform-specific wheels, including GNU/Linux (it makes standards-compliant manylinux and musllinux variants). That's what the wider community defaults to nowadays. Shipping sdists and forcing the end-users to compile on install is considered a hack.

package/db/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = ["setuptools", "wheel"]  # PEP 508 specifications.
```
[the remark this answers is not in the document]
ClementPinard: Not sure what you meant here.
The initial issue was about the missing `wheel` dependency on the `steup.py` file for this very package. Without touching the `setup.py` file, the build would fail because it defines a custom wheel class  and thus imports the `wheel` package. when building metadata.

A solution you proposed in the original issue was to simply provide a backup import with try except so that the `setup.py` can be run at the meta_build phase, but it sounded a bit hacky to me.

Another solution would be to either completely rewrite the `setup.py` in order to never import the wheel class, and provide the wanted specific features by pure python code, or to use a whole another build backend altogether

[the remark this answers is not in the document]
webknjaz: Yeah, I think I confused several files you have here. I propose that a comment is added here to state that this entry should be removed as soon as `setup.py` stops importing `wheel`..

PR #4125, opened 2024-03-04
fix on_startup operator not executing on startup

imanjra:

Fix `on_startup` not executing on startup due to race condition where on_startup operator are executed before invocation queue is initialized

Using the VoxelGPT plugin and verifying that the VoxelGPT panel launches on startup 

## Release Notes

Fix `on_startup` not executing on app startup

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/operators/src/StartupOperatorExecutor.tsx
+++ b/app/packages/operators/src/StartupOperatorExecutor.tsx
@@ -1,0 +1,15 @@
+import { useEffect } from "react";
+import { executeStartupOperators } from "./operators";
+
+let startupOperatorsExecuted = false;
+
+export default function StartupOperatorExecutor() {
+  useEffect(() => {
+    if (!startupOperatorsExecuted) {
+      executeStartupOperators();
+      startupOperatorsExecuted = true;
+    }
+  }, []);
+
+  return null;
+}
--- a/app/packages/operators/src/loader.tsx
+++ b/app/packages/operators/src/loader.tsx
@@ -4,21 +4,15 @@ import { useEffect, useState } from "react";
 import { useRecoilValue, useSetRecoilState } from "recoil";
 import { registerBuiltInOperators } from "./built-in-operators";
 import { useOperatorPlacementsResolver } from "./hooks";
-import { executeStartupOperators, loadOperatorsFromServer } from "./operators";
+import { loadOperatorsFromServer } from "./operators";
 import {
   availableOperatorsRefreshCount,
   operatorsInitializedAtom,
 } from "./state";
 
-let startupOperatorsExecuted = false;
-
 async function loadOperators(datasetName: string) {
   registerBuiltInOperators();
   await loadOperatorsFromServer(datasetName);
-  if (!startupOperatorsExecuted) {
-    executeStartupOperators();
-    startupOperatorsExecuted = true;
-  }
 }
 
 /**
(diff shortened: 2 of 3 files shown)


app/packages/operators/src/loader.tsx
```diff
@@ -4,21 +4,15 @@ import { useEffect, useState } from "react";
...
 
-let startupOperatorsExecuted = false;
-
 async function loadOperators(datasetName: string) {
   registerBuiltInOperators();
   await loadOperatorsFromServer(datasetName);
-  if (!startupOperatorsExecuted) {
-    executeStartupOperators();
-    startupOperatorsExecuted = true;
-  }
```
benjaminpkane, line 21: I believe the doc string below is incorrect now? `OperatorCore` now queues start-up operators?

benjaminpkane (commented):
I'm not sure I see the race condition. For edification purposes, is there specific logic in `useOperators` that causes this? `getInvocationRequestQueue` creates the queue if needed

imanjra:
`getInvocationRequestQueue` indeed creates a queue, however, the one of the executor (the only one atm) for the queue is initialized by `OperatorInvocationRequestExecutor`. The `OperatorInvocationRequestExecutor` uses `useInvocationRequestQueue` which initializes a subscriber for the queue (primarily updating `requests` state to be picked up by `OperatorInvocationRequestExecutor` reactively). However, the notifying to subscriber is not reactive but rather procedural and is invoked after queue is modified. If a subscriber is added with items already in queue (i.e. OperatorInvocationRequestExecutor) they are not notified of the items in the queue. There are alternate ways to go about solving this by notifying the subscriber on add or subscriber checking the queue when adding itself as a subscriber to the queue.

In summary (race-condition):

1. `loadOperators` populates the queue (and [notifies existing subscribers](https://github.com/voxel51/fiftyone/blob/develop/app/packages/operators/src/operators.ts#L778)) with start-up operators and blocks the mount of `OperatorCore` until it's done
2. Subscriber is initialized on the mount of OperatorInvocationRequestExecutor but does not get notified as the queue is already populated
3. Any further call to `queue.add` will execute

app/packages/operators/src/loader.tsx
imanjra: Yah, good point. I will update the doc string below.

benjaminpkane (approved):
LGTM!

app/packages/operators/src/StartupOperatorExecutor.tsx
```diff
@@ -0,0 +1,15 @@
+import { useEffect } from "react";
+import { executeStartupOperators } from "./operators";
+
+let startupOperatorsExecuted = false;
+
+export default function StartupOperatorExecutor() {
```
ritch, line 6: This seems like a workaround to the actual problem. How can we make the following possible regardless of where or when it happens:

```ts
const queue = getInvocationRequestQueue();
queue.add(new InvocationRequest());
```

ritch (changes requested):
PTAL a look at comment. We can discuss offline if needed.

app/packages/operators/src/StartupOperatorExecutor.tsx
ritch: > There are alternate ways to go about solving this by notifying the subscriber on add or subscriber checking the queue when adding itself as a subscriber to the queue.

This

ritch: Its important that `executeOperator()` is essentially possible from any state, thats why we introduced the queue in the first place.

imanjra:
Closing in favour of #4136

PR #4130, opened 2024-03-06
Adding Places dataset to the zoo

rohis06:

Added Places dataset to the zoo. It can be loaded from the zoo with `dataset = foz.load_zoo_dataset('places')`. 

Local validation of all the supported splits by loading the dataset from the zoo.

        notes for FiftyOne users.

The addition of Places to the dataset zoo.

-   [ ] Other


brimoor (approved):
@rohis06 implementation looks perfect! thanks so much for contributing! 🙌 

I'll make sure this is included in the next release, which might be `fiftyone==0.24` (targeting 2 weeks for this) or potentially earlier if we do a `fiftyone==0.23.7` patch release before then.

PR #4134, opened 2024-03-07
Add support for executing operators programmatically in notebooks

brimoor:
`foo.execute_operator()` currently does not work in notebooks. Upon investigation, the reason is that notebooks internally use an async event loop to do their thing, and `foo.execute_operator()` uses `asyncio.run()`, which is not allowed when there is already an event loop running.

Using `asyncio.get_running_loop().create_task(...)` in place of `asyncio.run(...)` seems to do the trick when running in notebooks 💡 

https://github.com/voxel51/fiftyone/assets/25985824/47bf4c76-9e8a-48b9-834b-d265d2a75e58

### DIFF
--- a/fiftyone/operators/executor.py
+++ b/fiftyone/operators/executor.py
@@ -13,6 +13,7 @@ import os
 import traceback
 
 import fiftyone as fo
+import fiftyone.core.context as foc
 import fiftyone.core.dataset as fod
 import fiftyone.core.odm.utils as focu
 import fiftyone.core.utils as fou
@@ -146,12 +147,18 @@ def execute_operator(operator_uri, ctx=None, **kwargs):
     """
     request_params = _parse_ctx(ctx=ctx, **kwargs)
 
-    result = asyncio.run(
-        execute_or_delegate_operator(
-            operator_uri, request_params, exhaust=True
-        )
+    coroutine = execute_or_delegate_operator(
+        operator_uri, request_params, exhaust=True
     )
-    result.raise_exceptions()
+
+    if foc.is_notebook_context():
+        # Notebooks already have event loops running, so we must execute there
+        loop = asyncio.get_event_loop()
+        loop.create_task(coroutine)
+        result = None  # @todo can we await result here?
+    else:
+        result = asyncio.run(coroutine)
+        result.raise_exceptions()
 
     return result
 
--- a/fiftyone/operators/executor.py
+++ b/fiftyone/operators/executor.py
@@ -138,7 +138,8 @@ def execute_operator(operator_uri, ctx=None, **kwargs):
             as keyword arguments rather than including them in ``ctx``
 
     Returns:
-        an :class:`ExecutionResult`
+        an :class:`ExecutionResult`, or an ``asyncio.Task`` if you run this
+        method in a notebook context
 
     Raises:
         ExecutionError: if an error occurred while immediately executing an
@@ -157,11 +158,10 @@ def execute_operator(operator_uri, ctx=None, **kwargs):
         loop = None
 
     if loop is not None:
-        # @todo how can we await result here?
+        # @todo is it possible to await result here?
         # Sadly, run_until_complete() is not allowed in Jupyter notebooks
         # https://nocomplexity.com/documents/jupyterlab/tip-asyncio.html
-        loop.create_task(coroutine)
-        result = None
+        result = loop.create_task(coroutine)
     else:
         result = asyncio.run(coroutine)
         result.raise_exceptions()
--- a/fiftyone/operators/executor.py
+++ b/fiftyone/operators/executor.py
@@ -13,7 +13,6 @@ import os
 import traceback
 
 import fiftyone as fo
-import fiftyone.core.context as foc
 import fiftyone.core.dataset as fod
 import fiftyone.core.odm.utils as focu
 import fiftyone.core.utils as fou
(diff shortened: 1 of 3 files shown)


fiftyone/operators/executor.py
```diff
@@ -146,12 +147,18 @@ def execute_operator(operator_uri, ctx=None, **kwargs):
...
+    coroutine = execute_or_delegate_operator(
+        operator_uri, request_params, exhaust=True
     )
-    result.raise_exceptions()
+
+    if foc.is_notebook_context():
+        # Notebooks already have event loops running, so we must execute there
+        loop = asyncio.get_event_loop()
+        loop.create_task(coroutine)
+        result = None  # @todo can we await result here?
```
brimoor, line 158: @benjaminpkane any asyncio wizardry here? Ideally we want to be able to await the tasks completion so we can return its `result()` to the user.

Some operators may perform computation and the user may want to access the data from the returned result object.

benjaminpkane: Will take a look today 👍

swheaton: I think you could do
```python
    loop = asyncio.get_event_loop()
    result = loop.run_until_complete(coroutine)
# later ...
result.raise_exceptions()
return result
```

fiftyone/operators/executor.py
```diff
@@ -146,12 +147,18 @@ def execute_operator(operator_uri, ctx=None, **kwargs):
...
-    result = asyncio.run(
-        execute_or_delegate_operator(
-            operator_uri, request_params, exhaust=True
-        )
+    coroutine = execute_or_delegate_operator(
+        operator_uri, request_params, exhaust=True
     )
-    result.raise_exceptions()
+
+    if foc.is_notebook_context():
```
swheaton, line 154: are there other cases that an event loop is existing? instead of being notebook specific we could do:

```python
try:
    loop = asyncio.get_running_loop()
    result = loop.run_until_complete(coro)
except RuntimeError:
    result = asyncio.run(coro)
```

swheaton: since this is a less common case, maybe we don't want to rely on try-except

fiftyone/operators/executor.py
swheaton: awaiting ben's comments as a more async savvy person though 👍🏼

fiftyone/operators/executor.py
benjaminpkane:
```python
try:
    loop = asyncio.get_running_loop()
    result = loop.run_until_complete(coro)
except RuntimeError:
    result = asyncio.run(coro)
``` 
I was going to recommend the above as well. `get_event_loop` is deprecated I believe. The above handles any situation where an event loop is running, and falls back to `run` if not

brimoor: I love the idea of `try-except` around `asyncio.get_running_loop()` rather than specifically checking for notebook contexts 👍 

Sadly, `run_until_complete()` is not allowed in Jupyter notebooks 😭 
https://nocomplexity.com/documents/jupyterlab/tip-asyncio.html

benjaminpkane: Hmm, perhaps the best practice is the user should await the coroutine in their notebook. From [stackoverflow](https://stackoverflow.com/questions/47518874/how-do-i-run-python-asyncio-code-in-a-jupyter-notebook/53581739#53581739)

brimoor: Yeah I arrived at the same conclusion and updated the PR. It's fine to just return the task to the caller and then they can `await` it in the notebook.

Will update the docs accordingly too 👍

swheaton: > Sadly, run_until_complete() is not allowed in Jupyter notebooks 😭
😭 😠

fiftyone/operators/executor.py
```diff
@@ -157,11 +158,10 @@ def execute_operator(operator_uri, ctx=None, **kwargs):
         loop = None
 
     if loop is not None:
-        # @todo how can we await result here?
+        # @todo is it possible to await result here?
```
benjaminpkane, line 161: Returning the coroutine is also an option, I think. Some discussion on [coroutines and tasks](https://stackoverflow.com/questions/57966935/asyncio-task-vs-coroutine)

```python
try:
    # Some contexts like notebooks already have a running event loop
    # return the coroutine for the user to await or schedule, if so
    loop = asyncio.get_running_loop()
    return coroutine
except:
    pass

result = asyncio.run(coroutine)
result.raise_exceptions()
return result
```

fiftyone/operators/executor.py
```diff
@@ -138,20 +138,33 @@ def execute_operator(operator_uri, ctx=None, **kwargs):
...
+        loop = None
+
+    if loop is not None:
+        # @todo is it possible to await result here?
+        # Sadly, run_until_complete() is not allowed in Jupyter notebooks
+        # https://nocomplexity.com/documents/jupyterlab/tip-asyncio.html
+        result = loop.create_task(coroutine)
+    else:
+        result = asyncio.run(coroutine)
+        result.raise_exceptions()
```
benjaminpkane, line 167: Returning the coroutine is also an option, I think. Some discussion on [coroutines and tasks](https://stackoverflow.com/questions/57966935/asyncio-task-vs-coroutine)
```python
try:
    # Some contexts like notebooks already have a running event loop
    # return the coroutine for the user to await or schedule, if so
    loop = asyncio.get_running_loop()
    return coroutine
except:
    pass

result = asyncio.run(coroutine)
result.raise_exceptions()
return result
```

brimoor: Creating the task has the benefit that you don't *have* to `await` in the notebook. Empirically what I see is that the task starts immediately and everything works exactly as I would want.

If the user does need the `ExecutionResult` from the operator, then they just have to use this pattern, which isn't too bad IMO:

```py
task = foo.execute_operator(...)

# Only necessary if you want the ExecutionContext
await task
result = task.result()
```

benjaminpkane: Makes sense 👍

PR #4136, opened 2024-03-07
notify new subscriber of pending operator invocation requests

imanjra:

Fix `on_startup` not executing on startup due to race condition where on_startup operators are executed before the invocation queue is initialized. With the change in the PR, when a subscriber is added, it will be notified of the queue if there are any pending invocation requests in the queue

Using the VoxelGPT plugin and verifying that the VoxelGPT panel launches on startup 

        notes for FiftyOne users.

Fix `on_startup` not executing on app startup

-   [ ] Other


ritch:
I like this solution a lot better 👍

PR #4137, opened 2024-03-07
add ability to auto-execute operator on opening a dataset

imanjra:

Add a new boolean OperatorConfig param `on_dataset_open` (`onDatasetOpen` in JS) to indicate if an operator should execute on opening a dataset. When set to `true`, the operator will execute for cases below:
- When a dataset is loaded on launching the app
- When a dataset is switched from dataset selector in the app
- When a dataset is switching using `session.dataset`
- When a dataset is switched using a trigger in the execution of an operator
- When a dataset is opened/switced by using a direct URL in the browser

Using cases listed above in the app

        notes for FiftyOne users.

Add a new boolean OperatorConfig param `on_dataset_open` (`onDatasetOpen` in JS) to indicate if an operator should execute on opening a dataset.

-   [ ] Other

### DIFF
--- a/app/packages/operators/src/index.ts
+++ b/app/packages/operators/src/index.ts
@@ -5,7 +5,6 @@ export {
   registerOperator,
   Operator,
   OperatorConfig,
-  executeStartupOperators,
   executeOperator,
   abortOperationsByExpression,
   abortOperationsByURI,
--- a/app/packages/operators/src/loader.tsx
+++ b/app/packages/operators/src/loader.tsx
@@ -4,7 +4,7 @@ import { useEffect, useState } from "react";
 import { useRecoilValue, useSetRecoilState } from "recoil";
 import { registerBuiltInOperators } from "./built-in-operators";
 import { useOperatorPlacementsResolver } from "./hooks";
-import { executeStartupOperators, loadOperatorsFromServer } from "./operators";
+import { executeOperatorsForEvent, loadOperatorsFromServer } from "./operators";
 import {
   availableOperatorsRefreshCount,
   operatorsInitializedAtom,
@@ -15,8 +15,9 @@ let startupOperatorsExecuted = false;
 async function loadOperators(datasetName: string) {
   registerBuiltInOperators();
   await loadOperatorsFromServer(datasetName);
+  executeOperatorsForEvent("onDatasetOpen");
   if (!startupOperatorsExecuted) {
-    executeStartupOperators();
+    executeOperatorsForEvent("onStartup");
     startupOperatorsExecuted = true;
   }
 }
--- a/app/packages/operators/src/operators.ts
+++ b/app/packages/operators/src/operators.ts
@@ -735,16 +735,22 @@ export class InvocationRequestQueue {
     this._queue = [];
   }
   private _queue: QueueItem[];
-  private _subscribers: ((queue: InvocationRequestQueue) => void)[] = [];
+  private _subscribers: InvocationRequestQueueSubscriberType[] = [];
   private _notifySubscribers() {
     for (const subscriber of this._subscribers) {
-      subscriber(this);
+      this._notifySubscriber(subscriber);
     }
   }
-  subscribe(subscriber: (queue: InvocationRequestQueue) => void) {
-    this._subscribers.push(subscriber);
+  private _notifySubscriber(subscriber: InvocationRequestQueueSubscriberType) {
+    subscriber(this);
   }
-  unsubscribe(subscriber: (queue: InvocationRequestQueue) => void) {
+  subscribe(subscriber: InvocationRequestQueueSubscriberType) {
+    this._subscribers.push(subscriber);
+    if (this.hasPendingRequests()) {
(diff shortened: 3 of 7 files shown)


brimoor:
@imanjra I updated the `load_default_view` operator proposed in https://github.com/voxel51/fiftyone-plugins/pull/115 to use the new `on_dataset_open` hook! 🎉 

It currently seems to work when I switch datasets via the selector in the App and when setting `session.dataset = dataset`. However, it does not currently trigger for me when I initially call `launch_app()`. Or at least the default view does not wind up being loaded.

The `load_default_view` operator is a good representative use case for this new hook. Importantly, note that what the user really wants is for the default view to be loaded *instead* of the dataset. For example, loading the full dataset may be expensive and the default view may load much quicker. Currently, it appears that the full dataset first loads and then another request to load the default view happens afterwards, which seems to lead to some race condition that causes it to take longer than usual to actually load the default view. Ideally the default view would be directly loaded and there would not be any attempt to load the full dataset at all. Analogous to calling `session.view = view` where `view` is a view into a different dataset than you are currently working with in the App.

It occurs to me that we might(?) need separate notions of *pre-load* and *post-load* hooks? Perhaps some operations might require the dataset to be loaded first while in the `load_default_view` use case we actually want to intercept the initial dataset load and load something else instead.

imanjra:
> However, it does not currently trigger for me when I initially call launch_app(). Or at least the default view does not wind up being loaded.

I believe this is due to the `on_startup` bug which is fixed in PR #4136. I'll rebase this branch to include those changes.

> It occurs to me that we might(?) need separate notions of pre-load and post-load hooks? Perhaps some operations might require the dataset to be loaded first while in the load_default_view use case we actually want to intercept the initial dataset load and load something else instead.

Yah, good point. The loading of the operators indeed blocks dataset loading so it should be possible and is worth a spike.  Need to verify to make sure the view set prior to initial loading of dataset uses the view to generate the route query

imanjra:
@brimoor Sorry, I accidentally overwrote the commit you had added updating the docs with rebase and force-push. Unfortunately, I do not have that commit locally (force pushed before pull 💀). Do you mind re-committing that commit?

brimoor:
> @brimoor Sorry, I accidentally overwrote the commit you had added updating the docs with rebase and force-push. Unfortunately, I do not have that commit locally (force pushed before pull 💀). Do you mind re-committing that commit?

No worries, I added the documentation commit back on the branch: https://github.com/voxel51/fiftyone/pull/4137/commits/e199a8a35da4f1148315d724c10cd8e0142c767e

brimoor:
@imanjra I'm using the example code in the description of https://github.com/voxel51/fiftyone-plugins/pull/115 to test this feature, and I'm seeing two issues:

1. still seeing the previously reported behavior where the `on_dataset_load` hook does not seem to be triggered upon initial `launch_app()`
2. `ctx.trigger("set_view", {"name": name})` fails with the following error:

Screenshot 2024-03-16 at 8 01 58 PM

imanjra:
@brimoor thank you for testing! Since the base was changed from `develop` to `release/v0.23.7` and the PR https://github.com/voxel51/fiftyone/pull/4136 is in developed, the on_startup is broken. I will cherry-pick https://github.com/voxel51/fiftyone/pull/4136 to this PR. Regarding saved view crash, I believe this can happen if you attempt set a saved view that does not exist on the dataset.

brimoor:
@imanjra ah I see, confirmed that the hooks are executing upon initial `launch_app()` now, thanks ✅ 

Would you mind running the example code in the description of https://github.com/voxel51/fiftyone-plugins/pull/115 to see what happens for you? I'm still getting the error I mentioned above, regardless of whether I set `on_dataset_open=True` or whether I manually trigger the operator, and the saved view in question definitely exists on the dataset:

```py
# fails
ctx.trigger("set_view", params={"name": default_view})

# works
view = ctx.dataset.load_saved_view(default_view)
ctx.trigger("set_view", params={"view": serialize_view(view)})
```

brimoor (approved):

imanjra:
> Would you mind running the example code in the description of https://github.com/voxel51/fiftyone-plugins/pull/115 to see what happens for you? I'm still getting the error I mentioned above, regardless of whether I set on_dataset_open=True or whether I manually trigger the operator, and the saved view in question definitely exists on the dataset:

Ohh interesting. I will give it a try and follow-up with a fix if needed. I wonder if it's specific to a particular saved-view and  if the same error occur when selecting the saved view from the saved-view selector.

brimoor:
> One thing I have run into (and could be possible cause) is if I use the non-sluggified saved view name, it does not work. It has to be the slug of a saved-view

@imanjra ah that would do it! I was testing with a saved view of name `collected_on`, but the sluggified name is `collected-on`.

Here's how I was attempting to use the operator:

```py
if ctx.dataset.has_saved_view(name):
    ctx.trigger("set_view", params={"name": name})
```

This feels like a very natural use case so I definitely think that the `set_view` operator should be operated to internally sluggify the provided `name` if necessary.

PR #4138, opened 2024-03-07
Don't recommend wildcard indexes by default

brimoor:
Wildcard indexes should not be the default recommendation for lightning mode.


swheaton (approved):
lgtm

PR #3973, opened 2024-01-05
Include coinciding filters in sidebar search results

benjaminpkane:
Adds coinciding filters to search results in the sidebar (not applicable to lightning filters)

### Before

https://github.com/voxel51/fiftyone/assets/19821840/efd5ad70-2989-4fd3-8050-2632ad5ffa85

### After

https://github.com/voxel51/fiftyone/assets/19821840/18329da9-f677-49ea-a423-04b97918d3db

PR #4139, opened 2024-03-07
Fix indexed BooleanFields in Lightning Mode

benjaminpkane:

Fixes a regression introduced by #3973 where indexed `BooleanField`s will throw an error in the `pathSearchResults` selector

```python
import fiftyone.zoo as foz

dataset.create_index("$**")
sample = dataset.first()
sample["bool"] = True
sample.save()
``` 

Opening the filter currently raises an error (see recording)

https://github.com/voxel51/fiftyone/assets/19821840/04383ecd-7e41-46b5-b675-605b04d96848

## Release Notes

* Fix indexed BooleanFields in Lightning Mode

-   [ ] Other

### DIFF
--- a/app/packages/state/src/recoil/pathData/lightningBoolean.ts
+++ b/app/packages/state/src/recoil/pathData/lightningBoolean.ts
@@ -15,7 +15,9 @@ export const lightningBooleanResults = selectorFamily<
       const [data] = get(lightningQuery([{ path: params.path }]));
 
       if (data.__typename !== "BooleanLightningResult") {
-        throw new Error("unexpected");
+        throw new Error(
+          `unexpected result ${data.__typename} for path ${params.path}`
+        );
       }
 
       return {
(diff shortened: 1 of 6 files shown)


app/packages/state/src/recoil/pathData/lightningBoolean.ts
```diff
@@ -15,7 +15,9 @@ export const lightningBooleanResults = selectorFamily<
       const [data] = get(lightningQuery([{ path: params.path }]));
 
       if (data.__typename !== "BooleanLightningResult") {
-        throw new Error("unexpected");
+        throw new Error(
+          `unexpected result ${data.__typename} for path ${params.path}`
```
kaixi-wang, line 19: Maybe in the message include what was expected in addition to what type it is?

findtopher (changes requested):
Feels like this will require a Teams release to go along with the OSS release.  That hasn't been set up or planned.

This should probably be targetted for the 3/20 Teams 1.5.8 / OSS 0.23.8 release

kaixi-wang (approved):
lgtm thanks!

PR #4143, opened 2024-03-08
Avoid re-exporting media unnecessarily

brimoor:
Small optimization to the `MediaExporter` class to avoid re-exporting the same media file multiple times, which would previously happen in situations like below where we are exporting a collection that contains multiple samples with the same `filepath`:

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.to_patches("ground_truth")

assert len(view.distinct("filepath")) == 1

# Now only exports the common image once
view.export(
    export_dir="/tmp/test",
    label_field="ground_truth",
)

    dataset_dir="/tmp/test",
    label_field="ground_truth",
)

assert len(dataset2.distinct("filepath")) == 1
```

### DIFF
--- a/fiftyone/utils/data/exporters.py
+++ b/fiftyone/utils/data/exporters.py
@@ -1221,6 +1221,7 @@ class MediaExporter(object):
             outpath = fos.normalize_path(outpath)
 
         if etau.is_str(media_or_path):
+            seen = False
             media_path = fos.normalize_path(media_or_path)
 
             if outpath is not None:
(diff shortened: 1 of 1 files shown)


findtopher (dismissed):
release/v0.23.7 was intended to be an OSS-only release

These changes will require a Teams release and should be targetted for the 3/20 release schedule.

sashankaryal (approved):
lgtm

PR #4150, opened 2024-03-11
Group dataset validation

brimoor:
Adds validation to nudge users to use `select_group_slices()` if they are passing a grouped collection to a method that was not specifically designed to handle grouped datasets/views.

This is intended to avoid confusion where users might expect that all slices of a grouped dataset will "automatically" be processed, when in fact the method would only process the active slice.

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -127,6 +127,16 @@ def import_annotations(
         **kwargs: CVAT authentication credentials to pass to
             :class:`CVATBackendConfig`
     """
+    if sample_collection.media_type == fom.GROUP:
+        if insert_new:
+            raise ValueError(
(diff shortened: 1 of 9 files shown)


fiftyone/utils/cvat.py
```diff
@@ -127,6 +127,16 @@ def import_annotations(
         **kwargs: CVAT authentication credentials to pass to
             :class:`CVATBackendConfig`
     """
+    if sample_collection.media_type == fom.GROUP:
+        if insert_new:
+            raise ValueError(
+                "insert_new=True is not support for grouped collections"
```
swheaton, line 133:
```suggestion
                "insert_new=True is not supported for grouped collections"
```

swheaton (approved):
lgtm i think

PR #4151, opened 2024-03-11
[DOCS] Document GCS and AWS env var credential options

swheaton:
Adds documentation of environment variables FOT recognizes for GCS and AWS, which was missing.

### DIFF
--- a/docs/source/teams/installation.rst
+++ b/docs/source/teams/installation.rst
@@ -221,7 +221,10 @@ You can do this in any of the following ways:
 
 1. Configure/provide AWS credentials in accordance with the
 `boto3 <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials>`_
-python library.
+python library, such as via environment variable, AWS config file, or IAM role.
+
+For example, by setting the environment variables: ``AWS_ACCESS_KEY_ID`` and
+``AWS_SECRET_ACCESS_KEY``.
 
 2. Permanently register AWS credentials on a particular machine by adding the
 following keys to your :ref:`media cache config <teams-media-cache-config>`:
@@ -232,6 +235,15 @@ following keys to your :ref:`media cache config <teams-media-cache-config>`:
         "aws_config_file": "/path/to/aws-config.ini",
         "aws_profile": "default"  # optional
     }
+
+3. Provide AWS credentials on a per-session basis by setting the following
+environment variables to point to your AWS credentials on disk:
+
+    .. code-block:: shell
+
+        export AWS_CONFIG_FILE="/path/to/aws-config.ini"
+        export AWS_PROFILE=default  # optional
+
 
 In the above, the `.ini` file should use the syntax of the
 `boto3 configuration file <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-a-configuration-file>`_.
@@ -257,7 +269,7 @@ here is an example configuration:
             "responseHeader": ["*"],
             "maxAgeSeconds": 86400
         }
-    ] 
+    ]
 
 .. _teams-google-cloud:
 
@@ -269,14 +281,23 @@ you simply need to provide
 `service account credentials <https://cloud.google.com/iam/docs/service-accounts>`_
 to your Teams client with read access to the relevant objects and buckets.
 
-You can register GCP credentials on a particular machine by adding the
-following key to your :ref:`media cache config <teams-media-cache-config>`:
+You can do this in any of the following ways:
+
+1. Permanently register GCS credentials on a particular machine by adding the
+following keys to your :ref:`media cache config <teams-media-cache-config>`:
 
 .. code-block:: json
 
     {
         "google_application_credentials": "/path/to/gcp-service-account.json"
     }
+
+2. Provide GCS credentials on a per-session basis by setting the following
+environment variables to point to your GCS credentials on disk:
+
+    .. code-block:: shell
+
+        export GOOGLE_APPLICATION_CREDENTIALS="/path/to/gcp-service-account.json"
 
 If you need to `configure CORS on your GCP buckets <https://cloud.google.com/storage/docs/using-cors>`_,
 here is an example configuration:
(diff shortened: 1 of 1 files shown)


docs/source/teams/installation.rst
```diff
@@ -221,7 +221,10 @@ You can do this in any of the following ways:
 
 1. Configure/provide AWS credentials in accordance with the
 `boto3 <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials>`_
-python library.
+python library, such as via environment variable, AWS config file, or IAM role.
+
+For example, by setting the environment variables: ``AWS_ACCESS_KEY_ID`` and
```
brimoor, line 226: For context, we used to explicitly mention options like `AWS_ACCESS_KEY_ID` here, but at some point we removed them and instead just linked to `boto3`, with the *goal* of conveying to the reader that they can configure credentials in their environment in *any way* that boto3 supports, which includes access key, config file, web identity token, IAM role, and probably other concoctions they've invented since then.

Re-reading this documentation now, I can see why something caused you to revisit this; the way it currently reads it doesn't feel like I'm being drawn to click on the `boto3` link hard enough... And also we *do* explicitly list all the supported ways to work with Azure and MinIO below.

My concern is that we should either list every combination of environment variables, or none of them.

A failure mode is that customers make feature requests for us to support things like `AWS_WEB_IDENTITY_TOKEN_FILE`, which we do in fact already support.

For consistency with Azure/MinIO, perhaps we should just bring back the explicit list of credential types that are supported, together with a `note::` stating that there could be more?

brimoor (changes requested):

docs/source/teams/installation.rst
swheaton: It was an internal ask by @lkaneda that env vars be documented. I added the credentials path variables for consistency but then realized she might also have meant she wanted `AWS_ACCESS_KEY_ID`. So put that with qualifier language like "such as ...", "for example ..." 

> For consistency with Azure/MinIO, perhaps we should just bring back the explicit list of credential types that are supported, together with a note:: stating that there could be more?

Sure sounds reasonable.

docs/source/teams/installation.rst
```diff
@@ -269,15 +281,24 @@ you simply need to provide
...
     {
         "google_application_credentials": "/path/to/gcp-service-account.json"
     }
 
+2. Provide GCS credentials on a per-session basis by setting the following
+environment variables to point to your GCS credentials on disk:
+
+    .. code-block:: shell
+
+        export GOOGLE_APPLICATION_CREDENTIALS="/path/to/gcp-service-account.json"
```
brimoor, line 300: Not sure if this is the only way to provide credentials to [gcs.Client()](https://github.com/voxel51/eta/blob/eed37df10661d504485fd06ff68274ad203f682a/eta/core/storage.py#L1893). Would need some research to produce an exhaustive list.

Note that `google_application_credentials` above is something that *we invented* to allow a service account path to be configured permanently rather than relying on the potentially various different env var configurations or IAM roles that GCS clients support.

docs/source/teams/installation.rst
brimoor: Here's a list of ways we used to document here:

```
export AWS_ACCESS_KEY_ID=...
export AWS_SECRET_ACCESS_KEY=...
export AWS_SESSION_TOKEN=... (if applicable)
export AWS_DEFAULT_REGION=...

export AWS_ROLE_ARN=...
export AWS_WEB_IDENTITY_TOKEN_FILE=...
export AWS_ROLE_SESSION_NAME... (if applicable)
export AWS_DEFAULT_REGION=...

export AWS_SHARED_CREDENTIALS_FILE="/path/to/aws-credentials.ini"
export AWS_PROFILE=default  # optional

export AWS_CONFIG_FILE="/path/to/aws-config.ini"
export AWS_PROFILE=default  # optional
```

docs/source/teams/installation.rst
swheaton: I don't think so, this appears to be a valid environment variable for google. We also happen to intercept it first.
https://cloud.google.com/docs/authentication/application-default-credentials

swheaton: Added reference to the other options - attaching a service account and using gcloud CLI

image

docs/source/teams/installation.rst
swheaton: Added the above with a note that it is just for convenience and subject to change.

image

docs/source/teams/installation.rst
```diff
@@ -223,6 +223,32 @@ You can do this in any of the following ways:
 `boto3 <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials>`_
 python library.
 
+For convenience, below are the environment variables that can be set to
+use some of the AWS-supported credentialing methods. This list is subject to
+change; please see the `boto3` link above for the up-to-date reference.
+
+.. code-block:: bash
+
+    # Access Key
```
brimoor, line 232: Can you put these in separate `.. code-block:: bash` as done in the Azure section just to help clarify that these are distinct options (ie you don't have to provide all of them)?

docs/source/teams/installation.rst
```diff
@@ -223,6 +223,32 @@ You can do this in any of the following ways:
...
+    export AWS_ROLE_ARN=...
+    export AWS_WEB_IDENTITY_TOKEN_FILE=...
+    export AWS_ROLE_SESSION_NAME... #if applicable
+    export AWS_DEFAULT_REGION=...
+
+    # Shared Credentials File
+    export AWS_SHARED_CREDENTIALS_FILE="/path/to/aws-credentials.ini"
+    export AWS_PROFILE=default  # optional
+
+    # AWS Config File
```
brimoor, line 248: This is now listed twice. Looks like in the Azure section we split raw creds and `.ini` files into separate numbered sections. Do the same here for consistency?

brimoor (changes requested):
@swheaton can you retarget this at `release/v0.23.7` please?

brimoor (approved):
I will merge and handle my comments

swheaton:
Cool thanks

PR #4153, opened 2024-03-12
Add YOLOv8, YOLOv9, and YOLO-World to FiftyOne Model Zoo

jacobmarks:
With our Ultralytics integration, it is currently possible to load a YOLOv8, YOLOv9, or YOLO-World model from Ultralytics and then apply it to a FiftyOne dataset. This PR goes a step further and directly adds these models to the model zoo.

One upside of this approach is that model location is managed by FiftyOne, as opposed to loading directly from Ultralytics, which results in the checkpoint being downloaded to your current working directory.

The PR updates the Ultralytics integration docs to reflect these new changes.

(Details)

Adds YOLOv8, YOLOv9, and YOLO-World models directly to the zoo

        notes for FiftyOne users.

-   [x] Other

### DIFF
--- a/docs/source/integrations/ultralytics.rst
+++ b/docs/source/integrations/ultralytics.rst
@@ -21,7 +21,7 @@ following packages:
 
 .. code-block:: shell
 
-    pip install ultralytics "torch>=1.8"
+    pip install "ultralytics>=8.1.0" "torch>=1.8"
 
 .. _ultralytics-inference:
 
@@ -70,6 +70,10 @@ You can directly pass Ultralytics YOLO detection models to
     # model = YOLO("yolov5l.pt")
     # model = YOLO("yolov5x.pt")
 
+    # YOLOv9
+    # model = YOLO('yolov9c.pt')
+    # model = YOLO('yolov9e.pt')
+
     dataset.apply_model(model, label_field="boxes")
 
     session = fo.launch_app(dataset)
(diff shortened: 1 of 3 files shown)


fiftyone/utils/ultralytics.py
```diff
@@ -282,7 +287,7 @@ def _to_keypoints(result, confidence_thresh=None):
     return fol.Keypoints(keypoints=keypoints)
 
 
-class FiftyOneYOLOConfig(Config):
+class FiftyOneYOLOConfig(fout.TorchImageModelConfig, fozm.HasZooModel):
```
brimoor, line 290: Inheriting from `HasZooModel` is good here, because it gives you a `download_model_if_necessary()` method that you can call from within your `Model` class to ensure that:
- the model is downloaded
- a `config.model_path` property is populated pointing to the model weights

I believe you're inheriting from `TorchImageModelConfig` so that you have access to the `entrypoint_fcn` and `entrypoint_args` parameters that it defines. However, this is a bit misleading because these models are not directly `torch.nn.Module`s and therefore many of the parameters defined by `TorchImageModelConfig` don't apply to these Ultralytics models.

Instead, it should be straightforward to delete the `load_yolo_model()` function you added and instead put that logic inside the `FiftyOneYOLOModel` class. The flow inside `FiftyOneYOLOModel`'s constructor will just be:
- call `config.download_model_if_necessary()` to populate 
- call `ultralytics.YOLO(config.model_path)`

fiftyone/utils/ultralytics.py
```diff
@@ -433,3 +496,22 @@ def _to_detections(self, df):
                 for row in df.itertuples()
             ]
         )
+
+
+def load_yolo_model(**kwargs):
```
brimoor, line 501: recommend deleting per my above comment

brimoor (changes requested):
@jacobmarks thanks for this! LMK what you think about my proposed refactor which should simplify things 👍

fiftyone/utils/ultralytics.py
jacobmarks: Took your advice. The one thing I'm not entirely satisfied with here is the `model_name` is not passing through properly, just the `model_path`. This is sufficient to load the appropriate checkpoint, but it would be nice to keep this additional information intact. Maybe there's something I'm doing wrong 🤷

jacobmarks:
@brimoor just updated to reflect this approach!

docs/source/integrations/ultralytics.rst
```diff
@@ -230,6 +220,77 @@ manually convert Ultralytics predictions to :ref:`FiftyOne format <keypoints>`:
    :alt: ultralytics-keypoints
    :align: center
 
+.. _ultralytics-open-vocabulary-object-detection:
```
brimoor, line 223: moved this to the end of the section because the previous sections all use the same dataset defined in the `Setup` section

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -2348,9 +2348,361 @@
...
+            "size_bytes": 6534387,
+            "manager": {
+                "type": "fiftyone.core.models.ModelManager",
+                "config": {
+                    "url": "https://github.com/ultralytics/assets/releases/download/v8.1.0/yolov8n.pt"
+                }
+            },
+            "default_deployment_config_dict": {
+                "type": "fiftyone.utils.ultralytics.FiftyOneYOLODetectionModel",
+                "config": {}
```
brimoor, line 2368: FYI actually no parameters are required here. `load_zoo_model()` automatically downloads the model and injects the path into the `model_path` of the config

brimoor (approved):
LGTM! I simplified the implementation a bit (and made sure that all the integration examples still run before approving 😄)

PR #4157, opened 2024-03-13
Declaring compatibility with python 3.11 and (partially) 3.12

swheaton:

Declare support for python 3.11 and python 3.12.

With caveat that 3D in python 3.12 is not yet supported until `open3d` [supports 3.12](https://github.com/isl-org/Open3D/issues/6433)

For version in python3.11, python3.12 ...
- Unit tests passing
- Launched app with quickstart and did various cursory manual testing
- Read through documented release notes / breaking changes / deprecations to identify potential issues.

There is probably something I missed (if unit tests don’t cover). But mostly it seems to just work.

        notes for FiftyOne users.

Officially support python 3.11 and 3.12

-   [ ] Other

### DIFF
--- a/docs/source/getting_started/install.rst
+++ b/docs/source/getting_started/install.rst
@@ -20,9 +20,11 @@ Prerequisites
 -------------
 
 You will need a working Python installation. FiftyOne currently requires
-**Python 3.7 - 3.10**.
+**Python 3.7 - 3.12** (3D features may not be available in 3.12 until
+`open3d <https://pypi.org/project/open3d/>`_ supports it)
 
-On Linux, we recommended installing Python through your system package manager
+
+On Linux, we recommend installing Python through your system package manager
 (APT, YUM, etc.) if it is available. On other platforms, Python can be
 downloaded `from python.org <https://www.python.org/downloads>`_. To verify that
 a suitable Python version is installed and accessible, run `python --version`.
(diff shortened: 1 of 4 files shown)


docs/source/getting_started/install.rst
```diff
@@ -20,9 +20,11 @@ Prerequisites
 -------------
 
 You will need a working Python installation. FiftyOne currently requires
-**Python 3.7 - 3.10**.
+**Python 3.7 - 3.12** (3D features may not be available in 3.12 until
```
brimoor, line 23: IMO we can just remove the `open3d` note here. It's not even installed by default and if you don't have it you'll get a nice message saying "this feature required open3d to be installed". Then if you try to follow orders on Python 3.12, you'll figure out that it is either available or not.

Plus, are we really going to remember to update this when 3D is supported? 🤣

brimoor (approved):
LGTM 👍

brimoor:
@swheaton can you retarget this at `release/v0.23.7` please?

docs/source/getting_started/install.rst
swheaton: I subscribed to that issue so I would get notifications but yeah you're right, was just trying to save someone a few minutes if they ran into this

swheaton:
ok rebased and retargeted 0.23.7, removed open3d note

brimoor (approved):
LGTM

PR #4160, opened 2024-03-14
Adding Non-Maximum Suppression (NMS) utility to fiftyone.utils.labels through perform_nms()

rohis06:

Adds Non-Maximum Suppression (NMS) utility to fiftyone.utils.labels through `perform_nms()`! :tada:

It can be loaded as follows:
```python
import fiftyone.utils.labels as fl

    "coco-2017",
    split="validation",
    dataset_name="detector-recipe",
)

predictions_view = dataset.take(10, seed=51)

fl.perform_nms(sample_collection=predictions_view, in_field="predictions", out_field="nms_predictions", conf_threshold=0.65, iou_threshold=0.4, progress=True)
```

It has been thoroughly tested, as seen in this [nms_test.pdf](https://github.com/voxel51/fiftyone/files/14607543/nms_test.pdf)

        notes for FiftyOne users.

Addition of perform_nms() utility to fiftyone.utils.labels

-   [ ] Other

### DIFF
--- a/fiftyone/utils/labels.py
+++ b/fiftyone/utils/labels.py
@@ -5,6 +5,7 @@ Label utilities.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+import eta.core.serial as etas
 import eta.core.utils as etau
 
 import fiftyone.core.labels as fol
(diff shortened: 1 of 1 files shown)


fiftyone/utils/labels.py
```diff
@@ -673,3 +674,116 @@ def classifications_to_detections(
...
+
+    NMS is a post-processing technique used in object detection to eliminate
+    duplicate detections and select the most relevant detected objects. This
+    helps reduce false positives.
+
+    Args:
+        sample_collection: a
+            :class:`fiftyone.core.collections.SampleCollection`
+        in_field: the name of the :class:`fiftyone.core.labels.Detections` field
+        out_field: the name of the :class:`fiftyone.core.labels.Detections`
```
brimoor, line 698: How about making this optional: if not specified, the input field is updated in-place?

fiftyone/utils/labels.py
```diff
@@ -673,3 +674,116 @@ def classifications_to_detections(
...
+    """
+    fov.validate_collection_label_fields(
+        sample_collection, in_field, fol.Detections
+    )
+
+    samples = sample_collection.select_fields(in_field)
+
+    with fou.ProgressBar(progress=progress) as pb:
+        for sample in pb(samples):
+            detections_data_json_str = sample[in_field].to_json()
```
brimoor, line 722: Can you implement this by processing `sample[in_field].detections` as a list of `Detection` objects rather than serializing to JSON? We generally try to avoid converting to other formats if not necessary.

Note that you can call `sample[in_field].copy()` to get a copy of the detections so that the existing detections aren't modified in-place

fiftyone/utils/labels.py
```diff
@@ -673,3 +674,116 @@ def classifications_to_detections(
...
+    while len(detections) > 0:
+        # Pick the detection with highest confidence
+        selected_detection = detections[0]
+        nms_detections.append(selected_detection)
+        del detections[0]
+
+        # Compare with other detections for NMS
+        for d in detections:
+            if d["label"] == selected_detection["label"]:
+                iou = _calculate_iou(
```
brimoor, line 756: Can you use the existing `fiftyone.utils.iou.compute_ious()` utility here to avoid re-implementing this calculation?

brimoor (changes requested):
@rohis06 overall the method looks great, just have a few suggestions on implementation that would clean things up a bit if you're open to them! 😄

rohis06:
> @rohis06 overall the method looks great, just have a few suggestions on implementation that would clean things up a bit if you're open to them! 😄

Thanks a lot for reviewing this, @brimoor. I really appreciate your feedback! 😄
I'm definitely open to suggestions, and I'll incorporate them to clean things up.

fiftyone/utils/labels.py
rohis06: Sure, that sounds like a good idea to me.
I'll make it optional.

fiftyone/utils/labels.py
rohis06: Oh, definitely! I will check out the `.copy()` method.

fiftyone/utils/labels.py
rohis06: Sure, I'll incorporate this. It's much better always to use an existing utility! 😄

rohis06:
> > @rohis06 overall the method looks great, just have a few suggestions on implementation that would clean things up a bit if you're open to them! 😄
> 
> Thanks a lot for reviewing this, @brimoor. I really appreciate your feedback! 😄 I'm definitely open to suggestions, and I'll incorporate them to clean things up.

@brimoor, I have addressed the review comments.
Could you let me know if this clears things up? 😄

brimoor (approved):
@rohis06 thanks for your contribution! 🎉

rohis06:
> @rohis06 thanks for your contribution! 🎉

Anytime! 😄

PR #4055, opened 2024-01-30
WIP: add operations object to ctx

ritch:
Adds a new collection of methods to the `ctx` via `ctx.ops`. This should eventually contain all built in operators - both JS and python.

PR #4163, opened 2024-03-15
[Proposal] add util to execution context with a set_view util

imanjra:

Add a util to execution context with a util to set_view without the need to serialize in operator. Altough, the utili is only being added with `set_view`, the idea is this would contain common utilities that can be reusable. 

Using a test operator

        notes for FiftyOne users.

Add a util to execution context with a util to set_view without the need to serialize in operator.

-   [ ] Other

### DIFF
--- a/fiftyone/operators/executor.py
+++ b/fiftyone/operators/executor.py
@@ -5,12 +5,14 @@ FiftyOne operator execution.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+
 import asyncio
 import collections
 import inspect
 import logging
 import os
 import traceback
+import json
 
 import fiftyone as fo
 import fiftyone.core.dataset as fod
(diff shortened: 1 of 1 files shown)


brimoor (commented):
Having this utility available will be great!

@imanjra @ritch curious what you're thinking about making this (and other utils as Ibrahim points out) available as `ctx.set_view()` versus `ctx.utils.set_view()`.

https://github.com/voxel51/fiftyone/pull/4076 added `ctx.target_view()` directly to the `ExecutionContext` and it seems like we should standardize on a single pattern for this.

ritch:
I had thought about this before and I feel strongly that it should be `ctx.ops.set_view()`. `ops` would be an instance of `ContextualOperations` (or similar name). The idea is that each method on that class is syntactic sugar for calling a built in operator. This would be the mechanism that we would document built in operators.

With that in mind I would suggest we finish/merge this instead: https://github.com/voxel51/fiftyone/pull/4055

imanjra:
Oh, cool. Didn't realize that existed. Sounds good. I will close this in favour of #4055 and continue there.

PR #4164, opened 2024-03-15
Contextual operations

ritch:
Adds a new collection of methods to the `ctx` via `ctx.ops`. This should eventually contain all built in operators - both JS and python.

### DIFF
--- a/fiftyone/operators/operations.py
+++ b/fiftyone/operators/operations.py
@@ -1,40 +1,46 @@
+"""
+FiftyOne operator execution.
+
+| Copyright 2017-2023, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
 import json
+
 from bson import json_util
 
 
 class Operations(object):
-    def __init__(self, ctx):
-        """
-        Initialize the Operations class with a :class:`fiftyone.operators.ExecutionContext`.
+    """Interface to all builtin operations for an execution context._ctx
 
-        Args:
-            ctx: The :class:`fiftyone.operators.ExecutionContext` to use.
-        """
+    Args:
+        ctx: an :class:`fiftyone.operators.ExecutionContext`
+    """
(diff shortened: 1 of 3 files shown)


fiftyone/operators/operations.py
```diff
@@ -0,0 +1,323 @@
...
+        """Rename a sample field to a new field name.
+
+        Args:
+            field_name: the name of the field to rename
+            new_field_name: the new name for the field
+        """
+        return self._ctx.trigger(
+            "rename_sample_field",
+            params={
+                "field_name": field_name,
```
brimoor, line 56: @ritch just checking: does `resolve_input()` get called when triggering operations via `ctx.trigger()`?

Asking because in the vast majority of existing operations, parameter validation is _only_ performed in `resolve_input()`, and `execute()` assumes that all provided data is valid. If this method of triggering operations bypasses validation, then all kinds of bad things could happen.

brimoor (approved):
LGTM - added docs for ya

fiftyone/operators/operations.py
imanjra: `ctx.trigger` at the moment does not prompt user for input even if input is required for an operator. However, we do perform validation Python operator prior to execution. However, it seems like we do not do the same for JavaScript operator so we should add execution time validation for JavaScript operators. The failure due to validation should get logged in the console for Python operators. The follow-up improvement unrelated to the PR could be:

- Add JS execution time validation (if indeed missing)
- Show validation error (and other errors) in a palette even when invoked with trigger

imanjra (approved):
Awesome! LGTM 🚀

fiftyone/operators/operations.py
brimoor: @imanjra thank you! I've created a separate Jira ticket for your proposed follow-up improvements 👍

PR #4170, opened 2024-03-18
Validate JS Operators on ctx.trigger()

ritch:
Consider the following execution request from the example `OpenHistogramPanel` python operator:

```py
         ctx.trigger(
            "open_panel",
            params=dict(),
        )
```

Prior to this PR, this would silently fail to execute in the browser. After this change validation is enforced regardless of how the JS operator was executed. In cases where the operator prompt was used, this would be redundant, but this change doesn't attempt to prevent that since the validation is essentially free in the browser context.

With this change, when a JS operator is executed and the inputs are not valid the following error is shown to the user:

image

...with these errors in the console:

image

### DIFF
--- a/app/packages/operators/src/operators.ts
+++ b/app/packages/operators/src/operators.ts
@@ -2,6 +2,7 @@ import { getFetchFunction, isNullish, ServerError } from "@fiftyone/utilities";
 import { CallbackInterface } from "recoil";
 import * as types from "./types";
 import { stringifyError } from "./utils";
+import { ValidationContext, ValidationError } from "./validation";
 
 class InvocationRequest {
   constructor(public operatorURI: string, public params: any = {}) {}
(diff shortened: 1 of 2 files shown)


app/packages/operators/src/operators.ts
```diff
@@ -539,6 +550,19 @@ export async function executeOperatorWithContext(
...
+    const [vctx, errors] = await validateOperatorInputs(
+      operator,
+      ctx,
+      resolvedInputs
+    );
+    if (vctx.invalid) {
+      console.error(`Invalid inputs for operator ${operatorURI}:`);
+      console.error(errors);
+      throw new Error(
+        `Failed to execute operator ${operatorURI}. See console for details.`
```
ritch, line 563: This is the message shown to the user

app/packages/operators/src/operators.ts
```diff
@@ -484,6 +485,16 @@ export async function executeOperator(operatorURI, params: any = {}) {
   queue.add(request);
 }
 
+export async function validateOperatorInputs(
```
imanjra, line 488: should we use this in useOperatorPrompt as well?

ritch: I was suprised not to see `Operator#validate()` if we normalize it should probably be there. I think we can do that in another PR though.

PR #4175, opened 2024-03-19
np.int -> int in labels.py

kemaleren:
`numpy.int` is deprecated. It was an alias for `int`, so this change yields identical behavior and fixes the `AttributeError` that gets thrown with newer versions of NumPy.

The release notes for NumPy 1.20.0 list all the builtin aliases that were deprecated:

https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

Replace `np.int` with `int`.

`np.int` was just an alias for `int`, so this change should yield identical behavior.

        notes for FiftyOne users.

-   [ ] Other


swheaton (approved):
Thanks for the contribution @kemaleren !

PR #4178, opened 2024-03-19
add support for setting view by name and slug

imanjra:

add support for setting view by name and slug

### Usage: setting a view using the slug of a saved view
```py
def execute(self, ctx):
    ctx.trigger("set_view", {"name": "my-saved-view-name"})
```
### Usage: setting a view using the name of a saved view
```py
def execute(self, ctx):
    ctx.trigger("set_view", {"name": "My Saved View Name"})
```

Using a test operator

        notes for FiftyOne users.

add support for setting view by name and slug

-   [ ] Other

### DIFF
--- a/app/packages/operators/src/built-in-operators.ts
+++ b/app/packages/operators/src/built-in-operators.ts
@@ -20,6 +20,8 @@ import {
 } from "./operators";
 import { useShowOperatorIO } from "./state";
 import { useSetRecoilState } from "recoil";
+import useRefetchableSavedViews from "../../core/src/hooks/useRefetchableSavedViews";
+import { toSlug } from "@fiftyone/utilities";
 
 //
 // BUILT-IN OPERATORS
@@ -458,7 +460,10 @@ class SetView extends Operator {
     });
   }
   useHooks(ctx: ExecutionContext): {} {
+    const refetchableSavedViews = useRefetchableSavedViews();
+
     return {
+      refetchableSavedViews,
       setView: fos.useSetView(),
       setViewName: useSetRecoilState(fos.viewName),
     };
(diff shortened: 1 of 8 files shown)


app/packages/operators/src/built-in-operators.ts
```diff
@@ -467,13 +471,24 @@ class SetView extends Operator {
...
+      hooks.setView(view);
+    } else if (slug) {
+      hooks.setViewName(slug);
+    } else if (name) {
+      const savedViews = hooks.refetchableSavedViews?.[0]?.savedViews;
+      const savedView =
+        Array.isArray(savedViews) &&
+        savedViews.find((view) => name === view.name);
+      if (!savedView) {
+        throw new Error(`Saved view with name "${name}" does not exist.`);
```
brimoor, line 489: @imanjra do we have a JS implementation of the [to_slug()](https://github.com/voxel51/fiftyone/blob/fea54b1fb0d23f1a58853204f835f4c4877438e5/fiftyone/core/utils.py#L2373) utility?

Asking because, if so, then you could just do `name = to_slug(name)` here and then `name` could effectively be either a name or slug. The public interface validates that any new saved view name must be unique both as a `name` and a `slug`, so it wouldn't cause any views to be "unreachable".

One consequence of such an approach is that the error message could only read:

```
`Saved view with name or slug "${name}" does not exist`
```

imanjra: Good idea. Yeah, we do have it. I'll use that util here

brimoor (approved):
LGTM, tested by running https://github.com/voxel51/fiftyone-plugins/pull/115 and also tested a case where a saved view didn't exist and I saw the error notification as expected. Very nice!

PR #4180, opened 2024-03-19
MongoDB enhancements in the App server

benjaminpkane:

The goal of this PR is to ensure calls to `pymongo` through the SDK (as opposed to `motor`) on the App server do not occur in the running event loop. This prevents unnecessary server lock-ups due to synchronous IO calls.

e2e, locally

## Release Notes

* Optimized App server calls to MongoDB

-   [ ] Other


imanjra (approved):
LGTM. Smoke tested the app locally. Seems to be working normally 🚀

fiftyone/core/odm/document.py
```diff
@@ -644,9 +645,6 @@ def _save(
                 "Cannot save an abstract document"
             )
 
-        if self._meta.get("auto_create_index", True):
-            self.ensure_indexes()
-
```
swheaton, line 649: Agree with the sentiment, ensure indexes every time is really annoying.
Will this have any negative externalities? I don't think so but just checking

fiftyone/core/dataset.py
```diff
@@ -6809,7 +6809,10 @@ def _reload_docs(self, hard=False):
     def _serialize(self):
         return self._doc.to_dict(extended=True)
 
-    def _update_last_loaded_at(self):
+    def _update_last_loaded_at(self, force=False):
```
swheaton, line 6812: There is probably many other cases where we do not need to do this like in API. But good start here.

Is there never a case where we want the server to update last loaded at though? I'm thinking of dataset selection from the dropdown, etc.
I see the `force` parameter but it does not seem to be used in this PR (passed from `serialize_dataset(update_last_loaded_at)` which is default `False` and not overridden)

swheaton: nvm, ctrl-f fail

benjaminpkane: Yeah, the server should still update when loading a new dataset/view. Definitely more work/refinement to do here

fiftyone/core/odm/document.py
benjaminpkane: From what I have observed, it shouldn't as indexes are ensured on import for the dataset collection. But let me check dataset creation. Unfortunately this change impacts sample/frame collections as well

swheaton: yeah and like every other document type too right? Although the others don't have non-default (id) indexes I think so should be fine. Also I traced back the code stack and I don't really see where it's doing anything (`index_specs` not found in `fiftyone`) - but empirically I see index creation *all* the time so it must be doing something

swheaton: dataset creation definitely it happens, as well as cloning

benjaminpkane: Yes, I'm not seeing that as well 👍

PR #4185, opened 2024-03-21
Speed up segmentation conversion and `_transform_mask()`.

kemaleren:
`_segmentation_to_detections()`, `_segmentations_to_polylines()`, and `_transform_mask()` create a new full-sized mask for each object in the mask, which scales badly. In our use case we have images with dimension about 22,000 x 9,000, with tens of thousands of objects in them. Looping through each object and creating a new full-sized mask object in each iteration is extremely slow.

The proposed changes scale exponentially better. In a 10,000 x 10,000 image with 10,000 objects, the old method would have taken almost three hours, whereas the new method took only 8 seconds.

Here are timing comparisons with a mask of shape 10,000 x 10,000 on a MacBook M2 Air.

`_segmentation_to_detections()` and `_segmentations_to_polylines()` - similar numbers for both:
| method  | n objects | time (seconds) |
|--- | --- | ---|
| old | 10   | 11.3   |
| | 100 | 101.2 |
| | 1,000 | 1,022.5 |
| | 10,000 | ~10,000 (skipped) |
| new | 10 | 2.0 |
| | 100 | 2.0 |
| | 1,000 | 2.11 |
| | 10,000 | 8.26 |

`transform_mask()`:
| method  | n objects | time (seconds) |
|--- | --- | ---|
| old | 10   | 0.4   |
| | 100 | 3.5 |
| | 1,000 | 35.7 |
| | 10,000 | 352.2 |
| new | 10 | 2.0 |
| | 100 | 1.9 |
| | 1,000 | 2.0 |
| | 10,000 | 7.4 |

- Implement a new function `_find_slices()` to preprocess a mask and find the extent of each unique object. This function is the key to speeding up the others.

- Use that function to speed up `_segmentations_to_detections()`, `_segmentations_to_polylines()`, and `_transform_mask()`.

- Keep track of offsets and overall segmentation mask size, to ensure the new detections have correct relative bounding boxes and the new polylines have the correct coordinates. This includes switching from using `_get_polygons()` to `skimage.measure.label()` in `_parse_thing_instances()`, which avoids the problem of polygon tolerance being off by one or more pixels.

- Combine the core functionality of `_segmentations_to_detections()` and `_segmentations_to_polylines()` into a single function, `_convert_segmentations()`. This eliminates code duplication and allows future segmentation conversions to reuse this code.

- In `_convert_segmentations()`, convert RGB masks to integer masks, matching the way it is done in `_transform_mask()`. This eliminates a number of if/then branches and makes the rest of the function easier to reason about.

- Add unit tests to ensure the changed functions behave as expected.

Updated and added new unit tests to check correctness.

Running times comparisons appear above.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/labels.py
+++ b/fiftyone/core/labels.py
@@ -7,10 +7,14 @@ Labels stored in dataset samples.
 """
 import itertools
 import warnings
+from functools import partial
 
 from bson import ObjectId
 import cv2
 import numpy as np
+from scipy.ndimage import find_objects
+from skimage.segmentation import relabel_sequential
+from skimage.measure import label
 
 import eta.core.frameutils as etaf
 import eta.core.image as etai
@@ -758,7 +762,9 @@ class Polyline(_HasAttributesDict, _HasID, Label):
         return sg.MultiLineString(points)
 
     @classmethod
-    def from_mask(cls, mask, label=None, tolerance=2, **attributes):
+    def from_mask(
+        cls, mask, label=None, tolerance=2, round_coords=False, **attributes
+    ):
         """Creates a :class:`Polyline` instance with polygons describing the
         non-zero region(s) of the given full image mask.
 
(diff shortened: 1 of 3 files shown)


kemaleren:
After digging into `eta` to see exactly how `_mask_to_polygons()` worked, I realized that `round_coords` was a bad idea. There are ways to get the behavior I wanted here (basically to undo the interpolation done by `skimage.measure.find_contours()`, but simple rounding isn't the way to do it. It happened to work for the simple cases in the unit tests, but not in general.

brimoor (approved):
Awesome speed, thanks so much @kemaleren! 💪 

This will be included in `fiftyone==0.23.8` and FiftyOne Teams 1.5.9, coming to a package manager near you in a couple weeks 🚀

PR #4193, opened 2024-03-25
Hugging Face Hub Integration!

jacobmarks:
[Hugging Face Integration — FiftyOne 0.24.0 documentation.pdf](https://github.com/voxel51/fiftyone/files/14808654/Hugging.Face.Integration.FiftyOne.0.24.0.documentation.pdf)

## Overview

This integration introduces two functions: `load_from_hub()` and `push_to_hub()`.

The major architectural changes from the previous version are as follows:

1. Use configs `fiftyone.yml`, `fiftyone.yaml`, or a custom local yaml config file (can be extended to github as well) specifying the format of the dataset to be converted into FiftyOne. This is in contrast to the previous version of the integration, which had a default converter that tried its best but was not guaranteed to succeed, and custom Python loading script execution, which we can still add in at some point, but is not necessarily the lowest barrier to entry
2. As opposed to using HF's `datasets` library, loading the dataset with their `load_dataset()` function, and then converting — which was limiting our flexibility, requiring massive downloads in many cases, and resulting in duplication of a ton of files, this version uses the HF [Datasets Server API](https://huggingface.co/docs/datasets-server/index) to directly request and load the relevant data within needing to go through the `datasets` Python library. This gives a few key additional advantages, as I will document below.
3. Config info can be specified exclusively via kwargs, removing the need to explicitly create a `fiftyone.yml` file locally.
4. Supports loading datasets from gated repos

## Additional improvements:

- Added Logging during the loading & conversion process
- `description`, `license`, and `repo_id` from the HF config file are added to the FiftyOne dataset's `info` dictionary, and all tags listed in the config are added to the dataset's `tags`. 
- Support for thumbnails in the sample grid, additional media fields, and segmentation mask labels
- Includes FiftyOne version validation, ensuring that the dataset to be loaded is compatible with the user's current version of `fiftyone`

## Loading from the Hub

The `load_from_hub()` utility in the `hf_hub` utils allows you to load datasets from the Hugging Face Hub that are in either:
1. Parquet format — compatible with the Hugging Face `datasets` library, and accessible via the Datasets Server API
2. Any FiftyOne dataset type, from those listed [here](https://docs.voxel51.com/user_guide/dataset_creation/datasets.html#supported-import-formats)

When you use `load_from_hub()`, you *must* specify the `repo_id`, which specifies the organization and repo on Hugging Face Hub where the dataset can be found. This is the only positional argument.

The loading config also needs to be specified in one of three ways:
1. Via a `fiftyone.yml` or `fiftyone.yaml` file in the Hugging Face repo itself
2. Using the `config_file` keyword argument to specify the location (locally) of the config file to use
3. By passing the config params directly into the `load_from_hub()` call via keywords.

The only *required* element is a `format` specifier. For Parquet datasets, you can use `format="ParquetFilesDataset"` or `format="parquet"` for short. For FiftyOne formats, use the name of the class. For instance, for a dataset in the format `fiftyone.types.dataset_types.COCODetectionDataset`, use `format="COCODetectionDataset"`.

### Loading Arguments
Additionally, the user can specify options:

- `revision`: the revision (or version commit) or the dataset to load
- `split` or `splits`: which of the available splits they want to load.
- `subset` or `subsets`: which of the available subsets they want to load. Many datasets on the hub have multiple subsets. As an example, check out the [newyorker_caption_contest](https://huggingface.co/datasets/jmhessel/newyorker_caption_contest), which has 30 subsets.
- `max_samples`: the maximum number of samples per <split, subset> pair to load. This can be useful if you want to rapidly get a feel for the dataset without downloading 100s of GBs of data.
- `batch_size`: the batch size to use when requesting data from the datasets server and adding samples to the FiftyOne dataset
- `num_workers`: thread pool workers to use when downloading media
- `overwrite`: whether to overwrite existing documents for the dataset
- `persistent`: whether to persist the loaded dataset to disk
- `name`: a name to use for the dataset. If included, this will override any name present in the config file.

### Example Usage

To illustrate the power, flexibility, and simplicitly of this approach, here are a few examples with popular datasets on the Hugging Face Hub.

For all of these examples, we will use the following imports:

```python
import fiftyone as fo
import fiftyone.utils.hf_hub as fouh
```

#### [mnist](https://huggingface.co/datasets/mnist)

Load the test split of MNIST dataset:

```python
dataset = fouh.load_from_hub(
   "mnist", 
   split="test", 
   format="parquet", 
   classification_fields="label"
)

```

Here, "mnist" is the repo id, and we are using `classification_fields="label"` to specify that the feature called "label" in the Hugging Face dataset should be converted into a FiftyOne `Classification` label.

#### [coyo-700m](https://huggingface.co/datasets/kakaobrain/coyo-700m)

Load the first 1,000 samples from the COYO-700M dataset:

```python
dataset = fouh.load_from_hub(
   "kakaobrain/coyo-700m", 
   format="parquet", 
   max_samples=1000
)

```

Here we use `max_samples` to specify that we only want the first 1,000.

#### [cppe-5](https://huggingface.co/datasets/cppe-5)

Load the CPPE-5 dataset and persist it to database:

```python
dataset = fouh.load_from_hub(
   "cppe-5", 
   format="parquet", 
   detection_fields="objects",
   persistent=True
)

```

Here we use `detection_fields="objects"` to specify that the feature "objects" should be converted into a FiftyOne `Detections` label field.

#### [scene_parse150](https://huggingface.co/datasets/scene_parse_150)

Just load the test split from the `scene_parsing` subset:

```python
dataset = fouh.load_from_hub(
   "scene_parse150", 
   format="parquet", 
   subset="scene_parsing",
   split="test",
   classification_fields="scene_category",
   mask_fields="annotation"
)

```

Here we are using the `"split"` and `"subset"` keyword arguments to specify what we want to download. Also note that we are converting multiple features from the Hugging Face dataset into FiftyOne label fields. The segmentation masks are saved to disk.

**Documentation**: For comprehensive coverage of all of these options, supported datasets, and more, see the PDF version of the integration docs, attached.

## Pushing to the Hub

If you are working with a dataset in FiftyOne and you want to quickly share it with others, you can do so via the `push_to_hub()` function, which takes two positional arguments: the FiftyOne sample collection (a `Dataset` or `DatasetView`), and the `repo_name`, which will be combined with your username/org to construct the `repo_id` where the dataset will be uploaded.

When you push to the hub, a few things happen:
1. The dataset and its media files are exported/uploaded in a specified format. By default, this format is `fiftyone.types.dataset_types.FiftyOneDataset`, but you can specify the format you want via the `dataset_type` keyword argument.
2. A `fiftyone.yml` config file for the dataset is generated and uploaded, which contains all of the necessary information so that the dataset can be loaded with `load_from_hub()`.
3. A Hugging Face [Dataset Card](https://huggingface.co/docs/hub/en/datasets-cards) for the dataset is auto-generated, providing tags, metadata, license info, and a code snippet illustrating how to load the dataset from the hub.

When you push to the hub, you can specify any/all of the following dataset card and config file attributes:
- `description`
- `license`
- `tags`

`push_to_hub` supports the following Hugging Face API arguments:
- `private`: whether to upload the dataset as private or public
- `exist_ok`: whether to throw an error if the repo already exists

### Example Usage

- Upload the Quickstart dataset to a private repo:

```python
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.hf_hub as fouh

fouh.push_to_hub(dataset, "quickstart")
```

- Upload this dataset as a public `COCODetectionDataset` with an MIT license:

```python
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.hf_hub as fouh
import fiftyone.types as fot

fouh.push_to_hub(
    dataset,
    "quickstart-coco",
    dataset_type=fot.dataset_types.COCODetectionDataset,
    private=False,
    license="mit",
    label_fields="*" ### convert all label fields, not just ground truth
 )
```

- Upload the first 3 samples of a video dataset:

```python
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.hf_hub as fouh
import fiftyone.types as fot

fouh.push_to_hub(
    dataset[:3],
    "video-test",
    private=True,
 )
```

## Wish List
- Support different detection formats (like VOC, COCO, etc)
- Extend support for specifying the location of the config file at remote urls, such as github repos
- Extend this "config" based approach to loading from Github, and to working with datasets, models, and plugins
- URL redirects for anchor links from previous version of the integration which just had transformers
- Add robustness to the download process — currently, sometimes when sending requests to the Datasets Server it hangs. This is not great.
- Add support for converting directly from the Hugging Face dataset, assuming it was downloaded with `datasets.load_dataset()`
- Add functionality to extend an existing dataset with the next batch of samples. In other words, if you used `load_from_hub()` with `max_samples=1000`, but now you want the remaining samples, it shouldn't need to query the server for the first 1000. Currently, it doesn't re-download media files, but it does re-query the server.
- Handle FiftyOne version compatibility: currently, `push_to_hub()` sets the required FiftyOne version to the user's version on upload, but this is too restrictive...

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced utilities for working with Hugging Face, enabling pushing FiftyOne datasets to the Hugging Face Hub, loading datasets from the Hub into FiftyOne, and managing dataset repositories.
	- Introduced functionalities for handling dataset metadata, field conversions, and configurations based on Hugging Face Hub dataset configurations.
	- Added options for loading datasets with splits, subsets, and batch loading.


fiftyone/utils/hf_hub.py
```diff
@@ -0,0 +1,723 @@
...
+            if not os.path.exists(filepath):
+                Image.open(requests.get(url, stream=True).raw).save(filepath)
+        except:
+            logger.warning(f"Failed to download image from {url}")
+
+    return convert_media_field
+
+
+def _get_image_shape(image_path):
+    with Image.open(image_path) as img:
```
brimoor, line 426: please use this instead to avoid reimplementation (also it handles some tricky cases like EXIF tags):

```py
import fiftyone.core.metadata as fom

def _get_image_shape(image_path):
    metadata = fom.ImageMetadata.build_for(image_path)
    return (metadata.width, metadata.height)
```

fiftyone/utils/hf_hub.py
```diff
@@ -0,0 +1,723 @@
...
+    """
+    Load a dataset from the Hugging Face Hub into FiftyOne.
+
+    Args:
+        repo_id: the Hugging Face Hub identifier of the dataset
+        revision: the revision of the dataset to load
+        **kwargs: optional keyword arguments
+            split: the split of the dataset to load
+            splits: the splits of the dataset to load
+            subset: the subset of the dataset to load
```
brimoor, line 51: just curious: how does HF use "subset" compared to "split"?

fiftyone/utils/hf_hub.py
```diff
@@ -0,0 +1,723 @@
...
+
+def load_from_hub(repo_id, revision=None, **kwargs):
+    """
+    Load a dataset from the Hugging Face Hub into FiftyOne.
+
+    Args:
+        repo_id: the Hugging Face Hub identifier of the dataset
+        revision: the revision of the dataset to load
+        **kwargs: optional keyword arguments
+            split: the split of the dataset to load
```
brimoor, line 49: nit: would recommend converting these to named parameters in the function's definition, all defaulting to `None`

fiftyone/utils/hf_hub.py
```diff
@@ -0,0 +1,723 @@
...
+            splits: the splits of the dataset to load
+            subset: the subset of the dataset to load
+            subsets: the subsets of the dataset to load
+            max_samples: the maximum number of samples to load
+            batch_size: the batch size to use when loading samples
+            overwrite (True): whether to overwrite an existing dataset with the same name
+            persistent (False): whether the dataset should be persistent
+            name: the name of the dataset to create
+            config_file: the path to a config file to use to load the dataset,
```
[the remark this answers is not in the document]
jacobmarks: Yes! Was thinking about this as well, and for simple cases this would definitely be the better option, but I'd also love to allow config files stored in github repos (for the complicated cases). What do you think about this? I.e. allowing either of the following:

- passing the kwargs directly
- using config from a url

fiftyone/utils/hf_hub.py
jacobmarks: They often use "subset" for things where the schema is slightly different, whereas split is for traditional train test validation.

A good example of subsets are the [SVHN](https://huggingface.co/datasets/svhn) dataset, which has a "cropped_digits" subset that has classification labels, and "full_numbers" subset which has detection labels.

fiftyone/utils/hf_hub.py
jacobmarks: Can do!

fiftyone/utils/hf_hub.py
jacobmarks: Can do!

fiftyone/utils/hf_hub.py
```diff
@@ -0,0 +1,723 @@
...
+        if feature["_type"] == "Image":
+            url = row_content[media_field_name]["src"]
+        else:
+            url = row_content[media_field_name]
+
+        sample_dict[media_field_key] = filepath
+        try:
+            ## create the file if it doesn't exist
+            if not os.path.exists(filepath):
+                Image.open(requests.get(url, stream=True).raw).save(filepath)
```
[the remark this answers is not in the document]
jacobmarks: Would definitely love help with this — I think this is the source of slowness right now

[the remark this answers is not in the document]
jacobmarks: Think I have this working — may still not be the optimal solution, but I've managed to get a pretty decent speedup

fiftyone/utils/hf_hub.py
```diff
@@ -0,0 +1,1109 @@
+from concurrent.futures import ThreadPoolExecutor
+import json
+import logging
+import os
+from packaging.requirements import Requirement
+from PIL import Image
+import requests
+import shutil
+
+import huggingface_hub as hfh
```
brimoor, line 10: Can you use this import pattern?

```py
hfh = fou.lazy_import(
    "huggingface_hub", callback=lambda: fou.ensure_package("huggingface_hub")
)
```

This will give the user an informative error message if they don't have `huggingface_hub` installed, but also doesn't prevent the module from being imported in case the user is trying to do something that doesn't actually require the `huggingface_hub` package to the installed.

fiftyone/utils/hf_hub.py
```diff
@@ -0,0 +1,1109 @@
+from concurrent.futures import ThreadPoolExecutor
```
brimoor, line 1: Please add our standard module docstring to the top. Something like this:

```
"""
Utilities for working with `Hugging Face <https://huggingface.co>`_.

| Copyright 2017-2024, Voxel51, Inc.
|
"""
```

brimoor: As an aside, how would you feel about renaming the module to `fiftyone.utils.huggingface`? I'm thinking that interacting with HF Hub is a pretty core aspect of how FiftyOne and HF integrate, so it would be nice for the workflows to have "prime" branding:

```py
import fiftyone.utils.huggingface as fouh

dataset = fouh.load_from_hub(...)
fouh.push_to_hub(...)
```

Noting of course that the Transformers stuff has its own `fiftyone.utils.transformers` module.

In truth we probably should have started nesting things one level further `fiftyone.utils.huggingface.transformers` and `fiftyone.utils.huggingface.hf_hub` and then imported the important stuff in `fiftyone/utils/huggingface/__init__.py`, but that ship has already sailed for `transformers`.

If we use `fiftyone.utils.huggingface` for this PR, we could move stuff to a submodule in the future and import the public functions one-level higher in a backwards-compatible way, if we decided that this module is getting too big.

fiftyone/utils/hf_hub.py
```diff
@@ -0,0 +1,1109 @@
...
+            - a dictionary mapping frame label field names to keys to use when
+              constructing the frame label dictionaries to pass to the exporter
+        data_card_kwargs: additional keyword arguments to pass to the
+            `DatasetCard` constructor
+    """
+    if dataset_type is None:
+        dataset_type = fot.FiftyOneDataset
+
+    ### export the dataset to a temp local dir
+    tmp_dir = f"/tmp/{repo_name}"
```
brimoor, line 99: We have a `TempDir` context manager for cases like this:

```py
import eta.core.utils as etau

with etau.TempDir() as tmp_dir:
    dataset.export(...)

    # other stuff

# tmp_dir is automatically deleted when the context exits
```

The context manager ensures that things will always get cleaned up, even if there's an error or the user cancels the operation. And it handles macOS vs Windows path considerations for you.

brimoor (commented):
@jacobmarks this is looking very nice! definitely have my 👍 to add docs to the HF integration page so we can ship this 🚢

fiftyone/utils/hf_hub.py
jacobmarks: Yeah I'm with you. I ran into this when doing the first iteration of docs for the hub integration, realizing a lot of headers potentially needed to be changed. Moving to this paradigm for the future.

fiftyone/utils/hf_hub.py
```diff
@@ -0,0 +1,1109 @@
...
+    "float32",
+    "float64",
+    "bool",
+    "string",
+)
+
+logger = logging.getLogger(__name__)
+
+
+def get_cache():
```
brimoor, line 45: @jacobmarks ah I missed this before; can you refactor to avoid using `globals`? You can likely do this by passing a list between your various functions and append new paths to it.

Using globals is not recommended in general as it can cause issues with concurrency. Also in this implementation, if a user presses ctrl+c during execution, I believe the global won't be cleared and then the next time they try to download a dataset they'll get the old files downloaded as an unintended surprise!

fiftyone/utils/hf_hub.py
```diff
@@ -0,0 +1,1109 @@
...
+                with open(filepath, "wb") as f:
+                    for chunk in r.iter_content(chunk_size=8192):
+                        f.write(chunk)
+                # Optionally, open and save the image to ensure it's valid
+                Image.open(filepath).save(filepath)
+    except Exception as e:
+        logger.warning(f"Failed to download image from {url}: {e}")
+
+
+def _download_images_in_batch(urls_and_filepaths):
```
brimoor, line 742: @jacobmarks can you add a `num_workers` parameter to `load_from_hub()` that gets passed through to here that lets the user configure the thread pool size if desired? As done here, for example:
https://github.com/voxel51/fiftyone/blob/df0b63b64c56bdb3d51d19137d1b182ed8a57431/fiftyone/core/metadata.py#L253

Note that there's a utility to help parse `num_workers` and apply `fo.config` limits if needed:

```py
num_workers = fou.recommend_thread_pool_workers(num_workers)
```

Also, by convention we typically add a separate code path for `num_workers <= 1` that avoids the use of thread pools altogether. It's admittedly less important for threads but this is important when using process pools, because some systems (eg Windows) and some environments (eg when this code runs in a plugin) don't like/support multiprocessing. Here's an example of this:
https://github.com/voxel51/fiftyone/blob/df0b63b64c56bdb3d51d19137d1b182ed8a57431/fiftyone/core/metadata.py#L283C5-L283C25

fiftyone/utils/hf_hub.py
jacobmarks: Done!

docs/source/images/integrations/hf_push_advanced_example.jpg
davanstrien, line 1: nit, but it would be nice if the screenshot showed the datasets viewer once the data has loaded successfully.

docs/source/integrations/huggingface.rst
```diff
@@ -891,3 +901,1025 @@ model:
...
+To push datasets to and load datasets from the `Hugging Face Hub <https://huggingface.co/docs/hub/index>`_,
+you will need the `Hugging Face Hub Python client <https://github.com/huggingface/huggingface_hub>`_,
+which you can install via PyPI:
+
+.. code-block:: shell
+
+    pip install -U huggingface_hub
+
+
+To push a dataset to the Hub, and in many cases to even access a dataset on
```
davanstrien, line 932:
```suggestion
To push a dataset to the Hub, and in some cases, to access a dataset on
```

docs/source/integrations/huggingface.rst
```diff
@@ -891,3 +901,1025 @@ model:
...
+
+
+When you run this code, a few things happen:
+
+- The dataset and its media files are exported to a temporary directory and
+  uploaded to the specified Hugging Face repo.
+- A `fiftyone.yml` config file for the dataset is generated and uploaded to
+  the repo, which contains all of the necessary information so that the dataset
+  can be loaded with :func:`load_from_hub() <fiftyone.utils.huggingface.load_from_hub>`.
+- A Hugging Face `Dataset Card <https://huggingface.co/docs/hub/en/datasets-cards>`_
```
davanstrien, line 1003: super nice!

docs/source/integrations/huggingface.rst
```diff
@@ -891,3 +901,1025 @@ model:
...
+
+
+Your dataset will be available on the Hub at the following URL:
+
+.. code-block:: shell
+
+    https://huggingface.co/datasets/<your-username-or-org-name>/my-quickstart-dataset
+
+
+Pushing a |DatasetView| to the Hub works in exactly the same way. For example,
```
davanstrien, line 1015: This is a really nice feature, getting a bit ahead of myself but I think this would be very good to communicate about in comms for this integration.

docs/source/integrations/huggingface.rst
```diff
@@ -891,3 +901,1025 @@ model:
...
+
+    import fiftyone.utils.huggingface as fouh
+
+    dataset = fouh.load_from_hub("<username-or-org>/my-private-dataset")
+
+
+.. note::
+
+    As long as you have an environment variable `HF_TOKEN` set with your Hugging
+    Face token (with read access), you can load gated datasets that you have
```
davanstrien, line 1178:
```suggestion
    Face token (with read access), you can load private or gated datasets that you have
```

davanstrien: We have private and gated datasets that need approval/agreement of some terms before accessing. The token will work the same either way but this might be clearer for end users.

docs/source/integrations/huggingface.rst
```diff
@@ -891,3 +901,1025 @@ model:
...
+        subsets="full_numbers",
+        max_samples=1000,
+        batch_size=50,
+        num_workers=4
+    )
+
+    session = fo.launch_app(dataset)
+
+
+**Loading Gated Datasets**:
```
davanstrien, line 1905:
```suggestion
**Loading Private or Gated Datasets**:
```

docs/source/integrations/huggingface.rst
```diff
@@ -891,3 +901,1025 @@ model:
...
+
+Like public datasets, you can also load gated datasets from the Hub, as long as
```
davanstrien, line 1907:
```suggestion
Like public datasets, you can also load private or gated datasets from the Hub, as long as
```

davanstrien (changes requested):
This is looking really nice! I've made some small suggestions focusing on the docs for now. I'll try out the current PR tomorrow and post anything about usage.

docs/source/images/integrations/hf_push_advanced_example.jpg
jacobmarks: ah yes can absolutely change it! Great find

brimoor (approved):
LGTM! 🤗

docs/source/images/integrations/hf_push_advanced_example.jpg
jacobmarks: Note: because the dataset is uploaded in FiftyOne format rather than Parquet, the dataset viewer does not know how to process it. I think perhaps the better thing to do would be to something integrate FiftyOne datasets on the hub with a public deployment like try.fiftyone.ai so people can "visualize with FiftyOne"

PR #3985, opened 2024-01-10
Add 3d media type / mesh support

sashankaryal:
## Changes
- Added a new media type 3d to support 3D visualizations.
- Implemented classes for `Scene`, `Material`, `Mesh`, `Shape`, `PointCloud`, and `Light`, along with their respective serialization and deserialization logic.
- Add rendering logic tailored for FiftyOne 3d scenes.
- Add support for exporting/importing fo3d files

## Testing
- Unit tests for newly introduced SDK classes
- Currently working on Playwright tests to cover end-to-end workflows

## Summary by CodeRabbit

- **New Features**
	- Added support for 3D datasets in the application, including new functionalities for 3D visualization.
	- Introduced configuration updates for language settings and review workflows.
	- Enhanced the app with chat auto-reply functionality.
	- Added installation of specific dependencies for building documentation.
	- Comprehensive updates to support rendering of both "3d" and "point_cloud" media types, including new components for 3D visualization.
	- Updated documentation to reflect support for 3D datasets and new visualization functionalities.
- **Refactor**
	- Refactored the configuration loading process for more flexibility in plugin loading.
	- Modified functionality related to 3D slices visibility and introduced new logic for conditional rendering based on 3D slice presence.
	- Updated import paths and condition checks to improve consistency and support new 3D functionalities.
- **Bug Fixes**
	- Fixed issues related to 3D media type handling in GraphQL queries and state management.
- **Tests**
	- Updated end-to-end tests to reflect changes in attribute checks and support for new 3D functionalities.
- **Chores**
	- Enhanced utility functions and introduced new test cases for improved functionality.

PR #4206, opened 2024-03-27
Fix 0º and 180º singularity cases in _rotation_matrix_from_vectors

### LINKED ISSUE #4203: [BUG] Issue with Orthographic Projection when supplied projection_normal=(0, 0, -1)
### Describe the problem

The method [`_rotation_matrix_from_vectors`](https://github.com/voxel51/fiftyone/blob/7bea0c2b4aae5cad19d84c1ac49e0baa65fbe817/fiftyone/utils/utils3d.py#L777-L786) is ill-defined and returns `NaN` when the angle between `vec1` and `vec2` is 180º. 

From a geometrical standpoint the issue is being cause by the fact that there is no longer a single "shortest" rotation that transforms `vec1` to `vec2` There are now infinite perpendicular axes to `vec1` and `vec2` that can be used to rotate `vec1` onto `vec2`.

From a numerical perspective, what is happening is that `s` in this [line](https://github.com/voxel51/fiftyone/blob/7bea0c2b4aae5cad19d84c1ac49e0baa65fbe817/fiftyone/utils/utils3d.py#L786) is 0. And the rotation matrix returned becomes `NaN`.

A solution would would be to pick a random axis perpendicular to `vec1` and `vec2` and apply rotation 180º around it. 

### Code to reproduce issue

I triggered the issue when calling  `fiftyone.utils.utils3d.compute_orthographic_projection_images`, with 

```python
compute_orthographic_projection_images(
    ...
    projection_normal=(0, 0, -1),
)
```

But you can shortcut all the stack call and simply call 

```python
fiftyone.utils.utils3d. _rotation_matrix_from_vectors((0, 0, -1), (0, 0, 1))
```

In general, it would be ideal if one could specify one not only the projection normal, but an actual full rotation matrix to fully define the camera orientation. As things are, the pose still has some degrees of freedom.

### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 22.04): macOS Sonoma 14.4 (23E214)
-   **Python version** (`python --version`): 3.10.14
-   **FiftyOne version** (`fiftyone --version`): v0.23.7

### Other info/logs

n.a

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community

SergioRAgostinho:

Fixes #4203.

The method [_rotation_matrix_from_vectors](https://github.com/voxel51/fiftyone/blob/7bea0c2b4aae5cad19d84c1ac49e0baa65fbe817/fiftyone/utils/utils3d.py#L777-L786) is ill-defined and returnes `NaN` when the angle between `vec1` and `vec2` was 0º or 180º.

From a geometrical standpoint the issue is being cause by the fact that there is no longer a single "shortest" rotation that transforms vec1 to vec2 There are now infinite perpendicular axes to vec1 and vec2 that can be used to rotate vec1 onto vec2.

This PR addresses that. 

With unit tests validating exactly the cases described above, as well as the original cases.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/utils3d.py
+++ b/fiftyone/utils/utils3d.py
@@ -5,6 +5,7 @@
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+
 import itertools
 import logging
 import warnings
@@ -774,16 +775,38 @@ def _clamp_to_discrete(arr, discrete):
     return clamp_list[np.clip(idx, 0, len(clamp_list) - 1)]
 
 
-# Reference: https://math.stackexchange.com/q/180418
+def _skew_symmetric_matrix(vec):
+    """Returns the skew-symmetric matrix of a 3D vector."""
+    return np.array(
+        [[0, -vec[2], vec[1]], [vec[2], 0, -vec[0]], [-vec[1], vec[0], 0]]
+    )
+
+
+# References:
+# 1. https://math.stackexchange.com/q/180418
+# 2. https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula#Matrix_notation
 def _rotation_matrix_from_vectors(vec1, vec2):
     """Returns the rotation matrix that aligns vec1 to vec2."""
     a = (np.asarray(vec1) / np.linalg.norm(vec1)).reshape(3)
     b = (np.asarray(vec2) / np.linalg.norm(vec2)).reshape(3)
     v = np.cross(a, b)
     c = np.dot(a, b)
-    s = np.linalg.norm(v)
-    K = np.array([[0, -v[2], v[1]], [v[2], 0, -v[0]], [-v[1], v[0], 0]])
-    return np.eye(3) + K + K.dot(K) * ((1 - c) / (s**2))
+
+    if abs(c + 1) > np.finfo(c.dtype).eps:
+        K = _skew_symmetric_matrix(v)
+        return np.eye(3) + K + K.dot(K) / (1 + c)
+
+    # 180º rotation singular case
+    # we need to pick a random perpendicular vector to `a`
+    for axis in [(1, 0, 0), (0, 1, 0), (0, 0, 1)]:
+        # `v` will be perpendicular to `a`, thus rotating `a` 180º along `v`
+        # will bring it to the opposite direction
+        v = np.cross(a, axis)
+        if np.linalg.norm(v) > np.finfo(v.dtype).eps:
+            break
+
+    K = _skew_symmetric_matrix(v)
+    return np.eye(3) + 2 * K.dot(K)
 
 
 def _parse_size(size, bounds):
--- a/tests/unittests/utils/utils3d_tests.py
+++ b/tests/unittests/utils/utils3d_tests.py
@@ -1,0 +1,34 @@
+"""
+| Copyright 2017-2024, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
+import numpy as np
+import pytest
+
+import fiftyone.utils.utils3d as fou3d
+
+
+@pytest.mark.parametrize(
+    ("vec1", "vec2"),
+    [
+        ((0, 1, 0), (0, 0, 1)),  # normal 90º rotation
+        ((0, 0, 1), (0, 0, 1)),  # singular case of 0º rotation
+        ((0, 0, -1), (0, 0, 1)),  # singular case of 180º rotation
+        ((0.5, 0, 2), (2, 1, 0)),  # random vector
+    ],
+)
+def test_rotation_matrix_from_vectors(vec1, vec2):
+    """Validates the returned rotation matrix transforms rotates vec1 to vec2."""
+    R = fou3d._rotation_matrix_from_vectors(vec1, vec2)
+
+    vec1_rot = R @ vec1
+    vec1_rn = vec1_rot / np.linalg.norm(vec1_rot)
+    vec2_n = vec2 / np.linalg.norm(vec2)
+
+    # The returned matrix is a rotation matrix
+    np.testing.assert_allclose(np.eye(3), R.T @ R, atol=1e-15)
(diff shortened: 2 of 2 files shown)


tests/unittests/utils/utils3d_tests.py
swheaton, line 1: We already have `tests/unittests/utils3d_tests.py`. But I think this structure makes more sense, we should probably refactor to do that consistently at some point...
Please first do `git mv tests/unittests/utils3d_tests.py tests/unittests/utils/utils3d_tests.py` (don't want to lose git history) and then add your test after - obviously save to a temp location first so you have it.

SergioRAgostinho:
Regarding this.

> | Objective                                                             | Addressed | Explanation                                                                                           |
> |----------------------------------------------------------------------|-----------|-------------------------------------------------------------------------------------------------------|
> | Allow specifying a full rotation matrix for camera orientation in orthographic projection (#4203) |     ❌     | The changes focus on fixing the NaN issue but do not address the flexibility in specifying a full rotation matrix. |

It's better to handle it an a separate PR because it requires touching a public facing API and it needs discussion on what to do with current way of specifying the camera orientation.

swheaton:
> Regarding this.
> 
> > Objective
> > Addressed
> > Explanation
> > 
> > 
> > 
> > 
> > Allow specifying a full rotation matrix for camera orientation in orthographic projection (#4203)
> > ❌
> > The changes focus on fixing the NaN issue but do not address the flexibility in specifying a full rotation matrix.
> 
> It's better to handle it an a separate PR because it requires touching a public facing API and it needs discussion on what to do with current way of specifying the camera orientation.

Def agree! Code Rabbit keeping us honest

fiftyone/utils/utils3d.py
```diff
@@ -774,16 +775,38 @@ def _clamp_to_discrete(arr, discrete):
...
     b = (np.asarray(vec2) / np.linalg.norm(vec2)).reshape(3)
     v = np.cross(a, b)
     c = np.dot(a, b)
-    s = np.linalg.norm(v)
-    K = np.array([[0, -v[2], v[1]], [v[2], 0, -v[0]], [-v[1], v[0], 0]])
-    return np.eye(3) + K + K.dot(K) * ((1 - c) / (s**2))
+
+    if abs(c + 1) > np.finfo(c.dtype).eps:
+        K = _skew_symmetric_matrix(v)
+        return np.eye(3) + K + K.dot(K) / (1 + c)
```
swheaton, line 797: Nice, I had seen this simplification also

tests/unittests/utils/utils3d_tests.py
```diff
@@ -0,0 +1,34 @@
...
+        ((0, 0, 1), (0, 0, 1)),  # singular case of 0º rotation
+        ((0, 0, -1), (0, 0, 1)),  # singular case of 180º rotation
+        ((0.5, 0, 2), (2, 1, 0)),  # random vector
+    ],
+)
+def test_rotation_matrix_from_vectors(vec1, vec2):
+    """Validates the returned rotation matrix transforms rotates vec1 to vec2."""
+    R = fou3d._rotation_matrix_from_vectors(vec1, vec2)
+
+    vec1_rot = R @ vec1
```
swheaton, line 26: whoa I never knew this operator existed ... (named as `def __matmul__`)

fiftyone/utils/utils3d.py
```diff
@@ -5,6 +5,7 @@
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+
```
SergioRAgostinho, line 8: Oops. An external linter kicked unintentionally. Will revert.

tests/unittests/utils/utils3d_tests.py
swheaton: thanks for the tests though! Historically we have used `unittest.TestCase` but I personally prefer pytest as, for example, I'm not sure if you can replicate `pytest` parametrization in `unittest`.

SergioRAgostinho: Heads up: There are some hard coded file paths loading assets in those tests that will need to change. Let me know if I should also move these assets or just update the paths.
 https://github.com/voxel51/fiftyone/blob/7bea0c2b4aae5cad19d84c1ac49e0baa65fbe817/tests/unittests/utils3d_tests.py#L177

swheaton: Ok forget it just add your new test to the existing test file in `tests/unittests/utils3d_tests.py`

fiftyone/utils/utils3d.py
sashankaryal: Plug for https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.align_vectors.html

I use this `Rotation` module a lot myself in #3985

SergioRAgostinho: If `scipy` is a valid dependency for this project (didn't even consider) than there's no point in having `_rotation_matrix_from_vectors` at all. We should just use `Rotation.align_vectors`. Less code to maintain and this PR can be reduced to a 1-2 LoC diff.

swheaton: `scipy` was added as a [dependency](https://github.com/voxel51/fiftyone/blob/develop/setup.py#L69) very very recently, and is not in any release yet. But it's already been implicitly required for a while, as we require `scikit-learn` and likely others that require `scipy`.

I would definitely be a big fan of using a module of its stature rather than rolling our own.
Please make this change, and if there are any other functions/operations that would more appropriately be replaced by `scipy`, let us know or add to this PR!

swheaton (changes requested):
hold for resolution of comments

swheaton (approved):
Sorry @SergioRAgostinho I forgot to circle back on this. LGTM, thanks for added tests 👍🏼 

Congrats on your first contribution, welcome to the FiftyOne family!! Looking forward to more contribs from you 😃

PR #4221, opened 2024-04-02
Revert "bump version for v0.23.8"

imanjra:
This reverts commit 993629235a888589f267200f026d0e9589ef1f88.

I believe the version bump needs to come later


brimoor:
Actually I think its fine to bump the version already!

PR #4223, opened 2024-04-03
HF hub integration tweaks

brimoor:
- single progress bar when downloading parquet files datasets
- pass `token` to all relevant methods in `push_to_hub()`

### DIFF
--- a/fiftyone/utils/huggingface.py
+++ b/fiftyone/utils/huggingface.py
@@ -30,13 +30,13 @@ hfh = fou.lazy_import(
     "huggingface_hub", callback=lambda: fou.ensure_package("huggingface_hub")
 )
 
+
 DATASETS_SERVER_URL = "https://datasets-server.huggingface.co"
 DEFAULT_MEDIA_TYPE = "image"
 DATASET_METADATA_FILENAMES = ("fiftyone.yml", "fiftyone.yaml")
 DATASETS_MAX_BATCH_SIZE = 100
 DEFAULT_IMAGE_FILEPATH_FEATURE = "image"
 FIFTYONE_BUILTIN_FIELDS = ("id", "filepath", "tags", "metadata")
-REPO_TYPE = "dataset"
 SUPPORTED_DTYPES = (
     "int8",
     "int16",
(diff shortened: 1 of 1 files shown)


fiftyone/utils/huggingface.py
```diff
@@ -109,6 +111,10 @@ def push_to_hub(
     else:
         tags = _get_dataset_tags(dataset)
 
+    # do this now in case HF login fails before we do anything expensive
+    hf_username = hfh.whoami(token=token)["name"]
```
brimoor, line 115: needed to pass `token` here or else `HF_TOKEN` is strictly required

fiftyone/utils/huggingface.py
```diff
@@ -945,53 +928,57 @@ def _add_parquet_subset_to_dataset(dataset, config, split, subset, **kwargs):
...
+        f"Downloading {num_rows} images from {config.name} ({split}, {subset})..."
     )
 
     tags = [split]
     if subset != "default" and subset != config._repo_id:
         tags.append(subset)
 
-    for start_idx in range(0, num_rows, batch_size):
-        urls_and_filepaths = []
+    with fou.ProgressBar(total=num_rows) as pb:
```
brimoor, line 949: single progress bar tracking the entire operation

fiftyone/utils/huggingface.py
brimoor: @jacobmarks FYI I had an older version of `huggingface_hub` installed that didn't seem to support the `HF_TOKEN` environment variable. Can you track down the min version that supports it and update the `ensure_package()` call above to let the user know the min version that they need?

jacobmarks: Looks like this is where they added support in Google Colab at least: https://github.com/huggingface/huggingface_hub/pull/1895
👉 https://github.com/huggingface/huggingface_hub/releases/tag/v0.20.0
so `>=0.20.0`

jacobmarks (approved):
LGTM! 👍  Love the progress bar upgrade.

Will change the `huggingface_hub` version req

PR #4236, opened 2024-04-05
Lazily connect to database when needed

### LINKED ISSUE #182: Lazily start DB service to reduce import time?
Currently, running `fiftyone config`, which simply loads and prints one's FO config, takes ~2 seconds to execute on my machine, because `import fiftyone` currently triggers a DB service to be started, among other things.

Can we adopt a lazy initialization strategy where the DB service is only started when absolutely necessary?

There are many valid use cases, like the CLI, or utilizing many of the utility methods in `fiftyone.utils`, where no DB connection is needed.

### LINKED ISSUE #1804: [FR] Reduce FiftyOne import time
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Macos 12.0.1
-   **FiftyOne version (run `fiftyone --version`)**: v0.15.1
-   **Python version**: 3.10.4

### Commands to reproduce

`fiftyone`
or
`fiftyone --version`
or
`python -c "import fiftyone"`

### Describe the problem

Importing fiftyone in general is VERY slow often 4-6 seconds.
`import fiftyone`

This is not desirable if fiftyone is needed for command line applications.

The main culprits seem to be:
- submodules in fiftyone reference back to the public api which means importing submodules imports everything. Eg. `fiftyone.core.dataset` contains `import fiftyone`
- In `fiftyone/__public__.py` the following files seem to be the culprits:
    + **core.config**: 548.17ms
    + **core.odm**: 1019.60ms
    + **core.plots**: 620.60ms
    + core.session: 54.95ms
    + load_config(): 214.52ms
    + **establish_db_conn()**: 3761.12ms

### Code to reproduce issue

`import fiftyone`

`from fiftyone.core.dataset import Dataset`
- Should be noted that this seems to have back references to `fiftyone.__public__`. This should bot be the case across the library. Imports in submodules should make direct references to the components they need to speed up importing. Importing this should be expected to be faster than importing the entire public interface.

### Other info / logs

Output of running: `python3 -c "import fiftyone"` (With modifications given below)

Custom logs for each import from `fiftyone/__public__.py`
```bash
N/A: 0.00ms
core.config: 548.17ms
core.odm: 1019.60ms
load_config(): 214.52ms
load_annotation_config(): 0.19ms
load_app_config(): 0.07ms
N/A: 0.00ms
core.config: 278.28ms
core.odm: 582.77ms
load_config(): 58.93ms
load_annotation_config(): 0.11ms
load_app_config(): 0.04ms
establish_db_conn(): 0.01ms
core.aggregations: 2.35ms
core.config: 0.00ms
core.dataset: 122.92ms
core.expressions: 0.01ms
core.fields: 0.01ms
core.frame: 0.00ms
core.labels: 0.01ms
core.metadata: 0.00ms
core.models: 0.00ms
core.odm: 0.00ms
core.plots: 620.60ms
core.sample: 0.00ms
core.stages: 2.01ms
core.session: 54.95ms
core.utils: 0.01ms
core.view: 0.00ms
utils.eval.classification: 12.39ms
utils.eval.detection: 0.00ms
utils.eval.segmentation: 0.00ms
utils.eval.quickstart: 25.54ms
establish_db_conn(): 3761.12ms
core.aggregations: 2.75ms
core.config: 0.01ms
core.dataset: 68.54ms
core.expressions: 0.01ms
core.fields: 0.01ms
core.frame: 0.00ms
core.labels: 0.01ms
core.metadata: 0.00ms
core.models: 0.00ms
core.odm: 0.01ms
core.plots: 344.27ms
core.sample: 0.00ms
core.stages: 1.26ms
core.session: 31.29ms
core.utils: 0.01ms
core.view: 0.00ms
utils.eval.classification: 6.16ms
utils.eval.detection: 0.00ms
utils.eval.segmentation: 0.00ms
utils.eval.quickstart: 13.43ms
```

<details>
<summary>Code Modification of `__public__.py` Used To Generate the Above</summary>
<br>
<p>
import time

_t = time.time()
def print_elapsed(msg: str = None):
    global _t
    t = time.time() - _t
    print(f'{msg}: {t*1000:.2f}ms' if msg else f'{t*1000:.2f}ms')
    _t = time.time()

print_elapsed('N/A')
import fiftyone.core.config as _foc
print_elapsed('core.config')
import fiftyone.core.odm as _foo
print_elapsed('core.odm')

config = _foc.load_config()
print_elapsed('load_config()')
annotation_config = _foc.load_annotation_config()
print_elapsed('load_annotation_config()')
app_config = _foc.load_app_config()
print_elapsed('load_app_config()')

_foo.establish_db_conn(config)
print_elapsed('establish_db_conn()')

from .core.aggregations import (
    Bounds,
    Count,
    CountValues,
    Distinct,
    HistogramValues,
    Mean,
    Std,
    Sum,
    Values,
)
print_elapsed('core.aggregations')
from .core.config import AppConfig
print_elapsed('core.config')
from .core.dataset import (
    Dataset,
    list_datasets,
    dataset_exists,
    load_dataset,
    delete_dataset,
    delete_datasets,
    delete_non_persistent_datasets,
    get_default_dataset_name,
    make_unique_dataset_name,
    get_default_dataset_dir,
)
print_elapsed('core.dataset')
from .core.expressions import (
    ViewField,
    ViewExpression,
    VALUE,
)
print_elapsed('core.expressions')
from .core.fields import (
    ArrayField,
    BooleanField,
    ClassesField,
    DateField,
    DateTimeField,
    DictField,
    EmbeddedDocumentField,
    EmbeddedDocumentListField,
    Field,
    FrameNumberField,
    FrameSupportField,
    FloatField,
    GeoPointField,
    GeoLineStringField,
    GeoPolygonField,
    GeoMultiPointField,
    GeoMultiLineStringField,
    GeoMultiPolygonField,
    IntField,
    IntDictField,
    KeypointsField,
    ListField,
    ObjectIdField,
    PolylinePointsField,
    StringField,
    TargetsField,
    VectorField,
)
print_elapsed('core.fields')
from .core.frame import Frame
print_elapsed('core.frame')
from .core.labels import (
    Label,
    Attribute,
    BooleanAttribute,
    CategoricalAttribute,
    NumericAttribute,
    ListAttribute,
    Regression,
    Classification,
    Classifications,
    Detection,
    Detections,
    Polyline,
    Polylines,
    Keypoint,
    Keypoints,
    Segmentation,
    Heatmap,
    TemporalDetection,
    TemporalDetections,
    GeoLocation,
    GeoLocations,
)
print_elapsed('core.labels')
from .core.metadata import (
    Metadata,
    ImageMetadata,
    VideoMetadata,
)
print_elapsed('core.metadata')
from .core.models import (
    apply_model,
    compute_embeddings,
    compute_patch_embeddings,
    load_model,
    Model,
    ModelConfig,
    EmbeddingsMixin,
    TorchModelMixin,
    ModelManagerConfig,
    ModelManager,
)
print_elapsed('core.models')
from .core.odm import KeypointSkeleton
print_elapsed('core.odm')
from .core.plots import (
    plot_confusion_matrix,
    plot_pr_curve,
    plot_pr_curves,
    plot_roc_curve,
    lines,
    scatterplot,
    location_scatterplot,
    Plot,
    ResponsivePlot,
    InteractivePlot,
    ViewPlot,
    ViewGrid,
    CategoricalHistogram,
    NumericalHistogram,
)
print_elapsed('core.plots')
from .core.sample import Sample
print_elapsed('core.sample')
from .core.stages import (
    Concat,
    Exclude,
    ExcludeBy,
    ExcludeFields,
    ExcludeFrames,
    ExcludeLabels,
    Exists,
    FilterField,
    FilterLabels,
    FilterKeypoints,
    Limit,
    LimitLabels,
    GeoNear,
    GeoWithin,
    GroupBy,
    MapLabels,
    Match,
    MatchFrames,
    MatchLabels,
    MatchTags,
    Mongo,
    Shuffle,
    Select,
    SelectBy,
    SelectFields,
    SelectFrames,
    SelectLabels,
    SetField,
    Skip,
    SortBy,
    SortBySimilarity,
    Take,
    ToPatches,
    ToEvaluationPatches,
    ToClips,
    ToFrames,
)
print_elapsed('core.stages')
from .core.session import (
    close_app,
    launch_app,
    Session,
)
print_elapsed('core.session')
from .core.utils import (
    pprint,
    pformat,
    ProgressBar,
)
print_elapsed('core.utils')
from .core.view import DatasetView
print_elapsed('core.view')
from .utils.eval.classification import (
    evaluate_classifications,
    ClassificationResults,
    BinaryClassificationResults,
)
print_elapsed('utils.eval.classification')
from .utils.eval.detection import (
    evaluate_detections,
    DetectionResults,
)
print_elapsed('utils.eval.detection')
from .utils.eval.segmentation import (
    evaluate_segmentations,
    SegmentationResults,
)
print_elapsed('utils.eval.segmentation')
from .utils.quickstart import quickstart
print_elapsed('utils.eval.quickstart')
</p>
</details>

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community.

### LINKED ISSUE #1964: [FR] Make fiftyone --version work even if FiftyOne can't be imported?
Currently, `fiftyone --version` requires `import fiftyone as fo` to be runnable, which in turn requires, among other things, the ability to successfully connect to the database.

This presents a problem in the following situation:
- I install an incompatible (eg old) version of `fiftyone` for my database
- I try to `import fiftyone` and get an error (as expected)
- Unsure what the problem is, I want to create a GitHub issue to get some help
- As instructed by FiftyOne's issue template, I try to run `fiftyone --version` to report my package version
- I get the same error!

Thus the FR: it would be ideal if `fiftyone --version` were runnable even for a greater range of "broken" installations. So at least I can obtain my basic system information.

One way to retrieve this info is:

```py
import pkg_resources

print(pkg_resources.get_distribution("fiftyone").version)
```

But it may be better to continue to access it via `fiftyone.constants.VERSION`, but in such a way that database connection issues don't interfere with the process...

### LINKED ISSUE #3189: [FR] Do not create a database connection during import
### Proposal Summary

Do not setup a database connection during import. 

### Motivation

Setting up a database connection can be time consuming and error prone. Doing it implicitly during an import is unexpected (for me). It leads to errors and creates global state (that we all hate). This is a bad code smell. 

There is a very good alternative: having a client object that represents the connection. 

### Details

We including FiftyOne in our multi-gpu training code and it crashed because of it. Multiple processes tried to connect to the same port. This is an unwanted side-effect of doing an import in a file that is not used during the training. 

### Willingness to contribute

The FiftyOne Community welcomes contributions! Would you or another member of your organization be willing to contribute an implementation of this feature?

-   [ ] Yes. I can contribute this feature independently
-   [ ] Yes. I would be willing to contribute this feature with guidance from the FiftyOne community
-   [x] No. I cannot contribute this feature at this time

swheaton:
Closes #182
Closes #1964
Closes #1804 
Closes #3189 

Lazily connect to database so you can import `fiftyone` without a database connection.
Most of the work was testing.

### Lazy connection testing
A sampling of things users might want to be able to do without connecting to a database
#### Version
BEFORE
```shell
$ FIFTYONE_DATABASE_URI=invalid.com:27017 fiftyone --version
...
  File "/Users/stuart/dev/fiftyone/fresh_venv/bin/fiftyone", line 5, in <module>
    from fiftyone.core.cli import main
  File "/Users/stuart/dev/fiftyone/fiftyone/__init__.py", line 25, in <module>
    from fiftyone.__public__ import *
  File "/Users/stuart/dev/fiftyone/fiftyone/__public__.py", line 17, in <module>
    _foo.establish_db_conn(config)
  File "/Users/stuart/dev/fiftyone/fiftyone/core/odm/database.py", line 216, in establish_db_conn
    _validate_db_version(config, _client)
  File "/Users/stuart/dev/fiftyone/fiftyone/core/odm/database.py", line 301, in _validate_db_version
    raise ConnectionError("Could not connect to `mongod`") from e
ConnectionError: Could not connect to `mongod`
```
AFTER
```shell
$FIFTYONE_DATABASE_URI=invalid.com:27017 fiftyone --version
FiftyOne v0.24.0, Voxel51, Inc.
```
#### Config
BEFORE
```shell
$FIFTYONE_DATABASE_URI=invalid.com:27017 fiftyone config
...
ConnectionError: Could not connect to `mongod`
```
AFTER
```shell
$FIFTYONE_DATABASE_URI=invalid.com:27017 fiftyone config
FiftyOne v0.24.0, Voxel51, Inc.
```
#### Zoo
BEFORE
```shell
$FIFTYONE_DATABASE_URI=invalid.com:27017 fiftyone zoo datasets download quickstart
...
ConnectionError: Could not connect to `mongod`
```

AFTER
```shell
$FIFTYONE_DATABASE_URI=invalid.com:27017 fiftyone zoo datasets download quickstart
Dataset already downloaded
```
#### Data model
BEFORE
```python
import fiftyone as fo

...
ConnectionError: Could not connect to `mongod`
```
AFTER
```python
import fiftyone as fo

test_sample = fo.Sample(
    "/path/to/sample.png",
    ground_truth=fo.Detections(
        detections=[
            fo.Detection(label="thing", bounding_box=[0.5, 0.5, 0.5, 0.5])
        ]
    ),
)
assert test_sample.ground_truth.detections[0].label == "thing"
```

#### Actually requires DB stuff
BOTH
```python
import fiftyone as fo

ds = fo.Dataset()

...
ConnectionError: Could not connect to `mongod`
```

### Correctness Testing
Made sure things still work for normal operation.
This is where I'll likely need some input / ideas.

Works with both database_uri set and mongodb spun up for me.
1. Unit tests pass
2. `fiftyone datasets list`
3. `fiftyone app launch` and launch app from shell
   1. Load dataset
   2. Manipulate view via session
   3. Save view
   4. Filter via sidebar
   5. Patches
   6. Compute/view embeddings+viz
   7. simple dataset-using operator (example_target_view)
4. Debug mode: run `server/main.py` and `yarn dev` separately
5. Make sure migration gets checked/run on first DB-usage also
6. mongoengine documents can be saved before other operations where connection established

Other
1. Spins up 1 mongodb which is shared by other fo's (that are using DB). And it's not killed until last user is done.

        notes for FiftyOne users.

<pre>
Importing ``fiftyone`` doesn't attempt a database connection, rather it's deferred to when actually required.
</pre>

-   [ ] Other

## Summary by CodeRabbit

- **Refactor**
	- Improved database migration process based on environment variable configuration.
	- Enhanced database connection management in the repository factory module.

- **Tests**
	- Optimized unit tests for plugin secrets management with streamlined code.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -75,6 +75,7 @@ autodoc_default_options = {
     "inherited-members": True,
     "member-order": "bysource",
     "autosummary": True,
+    "exclude-members": "objects",
 }
 autodoc_inherit_docstrings = True
 autoclass_content = "class"
--- a/fiftyone/__init__.py
+++ b/fiftyone/__init__.py
@@ -7,8 +7,8 @@ See https://voxel51.com/fiftyone for more information.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+
 from pkgutil import extend_path as _extend_path
-import os as _os
 
 #
 # This statement allows multiple `fiftyone.XXX` packages to be installed in the
--- a/fiftyone/__init__.py
+++ b/fiftyone/__init__.py
@@ -26,8 +26,6 @@ from fiftyone.__public__ import *
 
 import fiftyone.core.logging as _fol
 
-_fol.init_logging()
-
 # The old way of doing things, migrating database on import. If we
 #   REALLY need to do this, for example doc build, we can.
 if (
@@ -37,3 +35,5 @@ if (
     import fiftyone.migrations as _fom
 
     _fom.migrate_database_if_necessary()
+
+_fol.init_logging()
--- a/fiftyone/__init__.py
+++ b/fiftyone/__init__.py
@@ -26,6 +26,8 @@ from fiftyone.__public__ import *
 
 import fiftyone.core.logging as _fol
 
+_fol.init_logging()
+
 # The old way of doing things, migrating database on import. If we
 #   REALLY need to do this, for example doc build, we can.
 if (
@@ -35,5 +37,3 @@ if (
     import fiftyone.migrations as _fom
 
     _fom.migrate_database_if_necessary()
-
-_fol.init_logging()
--- a/fiftyone/__init__.py
+++ b/fiftyone/__init__.py
@@ -25,9 +25,6 @@ __version__ = _foc.VERSION
 from fiftyone.__public__ import *
 
 import fiftyone.core.logging as _fol
-import fiftyone.migrations as _fom
+
 
 _fol.init_logging()
-
-if _os.environ.get("FIFTYONE_DISABLE_SERVICES", "0") != "1":
-    _fom.migrate_database_if_necessary()
--- a/fiftyone/__init__.py
+++ b/fiftyone/__init__.py
@@ -26,14 +26,5 @@ from fiftyone.__public__ import *
 
 import fiftyone.core.logging as _fol
 
-# The old way of doing things, migrating database on import. If we
-#   REALLY need to do this, for example doc build, we can.
-if (
-    _os.environ.get("FIFTYONE_DISABLE_SERVICES", "0") != "1"
-    and "_FIFTYONE_FORCE_DB_CONNECT_ON_IMPORT" in _os.environ
-):
-    import fiftyone.migrations as _fom
-
-    _fom.migrate_database_if_necessary()
 
 _fol.init_logging()
--- a/fiftyone/__init__.py
+++ b/fiftyone/__init__.py
@@ -26,5 +26,14 @@ from fiftyone.__public__ import *
 
 import fiftyone.core.logging as _fol
 
+# The old way of doing things, migrating database on import. If we
+#   REALLY need to do this, for example doc build, we can.
+if (
+    _os.environ.get("FIFTYONE_DISABLE_SERVICES", "0") != "1"
+    and "_FIFTYONE_FORCE_DB_CONNECT_ON_IMPORT" in _os.environ
+):
+    import fiftyone.migrations as _fom
+
+    _fom.migrate_database_if_necessary()
 
 _fol.init_logging()
--- a/fiftyone/__public__.py
+++ b/fiftyone/__public__.py
@@ -5,7 +5,6 @@ FiftyOne's public interface.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-import os as _os
 
 import fiftyone.core.config as _foc
 
@@ -13,13 +12,6 @@ config = _foc.load_config()
 annotation_config = _foc.load_annotation_config()
 evaluation_config = _foc.load_evaluation_config()
 app_config = _foc.load_app_config()
-
-# The old way of doing things, connecting to database on import. If we
-#   REALLY need to do this, for example doc build, we can.
-if "_FIFTYONE_FORCE_DB_CONNECT_ON_IMPORT" in _os.environ:
-    import fiftyone.core.odm as _foo
-
-    _foo.establish_db_conn(config)
 
 from .core.aggregations import (
     Aggregation,
--- a/fiftyone/__public__.py
+++ b/fiftyone/__public__.py
@@ -7,14 +7,11 @@ FiftyOne's public interface.
 """
 
 import fiftyone.core.config as _foc
-import fiftyone.core.odm as _foo
 
 config = _foc.load_config()
 annotation_config = _foc.load_annotation_config()
 evaluation_config = _foc.load_evaluation_config()
 app_config = _foc.load_app_config()
-
-_foo.establish_db_conn(config)
 
 from .core.aggregations import (
     Aggregation,
--- a/fiftyone/__public__.py
+++ b/fiftyone/__public__.py
@@ -5,6 +5,7 @@ FiftyOne's public interface.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+import os as _os
 
 import fiftyone.core.config as _foc
 
@@ -12,6 +13,13 @@ config = _foc.load_config()
 annotation_config = _foc.load_annotation_config()
 evaluation_config = _foc.load_evaluation_config()
 app_config = _foc.load_app_config()
+
+# The old way of doing things, connecting to database on import. If we
+#   REALLY need to do this, for example doc build, we can.
+if "_FIFTYONE_FORCE_DB_CONNECT_ON_IMPORT" in _os.environ:
+    import fiftyone.core.odm as _foo
+
+    _foo.establish_db_conn(config)
 
 from .core.aggregations import (
     Aggregation,
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -72,7 +72,8 @@ class DatabaseConfigDocument:
     type: str
 
     def __init__(self, conn, version=None, type=None, *args, **kwargs):
-        # Create our own __init__ so we can ignore extra kwargs
+        # Create our own __init__ so we can ignore extra kwargs / unknown
+        #   fields from other versions
         self._conn = conn
         self.version = version
         self.type = type
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -27,7 +27,7 @@ import eta.core.utils as etau
 
 import fiftyone as fo
 import fiftyone.constants as foc
-import fiftyone.migrations as fom
+
 from fiftyone.core.config import FiftyOneConfigError
 import fiftyone.core.fields as fof
 import fiftyone.core.service as fos
@@ -35,6 +35,7 @@ import fiftyone.core.utils as fou
 
 from .document import Document
 
+fom = fou.lazy_import("fiftyone.migrations")
 foa = fou.lazy_import("fiftyone.core.annotation")
 fob = fou.lazy_import("fiftyone.core.brain")
 fod = fou.lazy_import("fiftyone.core.dataset")
@@ -228,7 +229,14 @@ def establish_db_conn(config):
             % (db_config.type, foc.CLIENT_TYPE)
         )
 
-    if os.environ.get("FIFTYONE_DISABLE_SERVICES", "0") != "1":
+    # Migrate the database if necessary after establishing connection.
+    #   If we are forcing immediate connection upon import though, migration
+    #   is being performed at that time, and we cannot do it here as well,
+    #   or else it'll be a circular import.
+    if (
+        os.environ.get("FIFTYONE_DISABLE_SERVICES", "0") != "1"
+        and "_FIFTYONE_FORCE_DB_CONNECT_ON_IMPORT" not in os.environ
+    ):
         fom.migrate_database_if_necessary()
 
 
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -6,6 +6,7 @@ Database utilities.
 |
 """
 import atexit
+import dataclasses
 from datetime import datetime
 import logging
 from multiprocessing.pool import ThreadPool
@@ -15,7 +16,6 @@ import asyncio
 from bson import json_util, ObjectId
 from bson.codec_options import CodecOptions
 from mongoengine import connect
-import mongoengine.errors as moe
 import motor.motor_asyncio as mtr
 
 from packaging.version import Version
@@ -29,11 +29,8 @@ import fiftyone as fo
 import fiftyone.constants as foc
 import fiftyone.migrations as fom
 from fiftyone.core.config import FiftyOneConfigError
-import fiftyone.core.fields as fof
 import fiftyone.core.service as fos
 import fiftyone.core.utils as fou
-
-from .document import Document
 
 foa = fou.lazy_import("fiftyone.core.annotation")
 fob = fou.lazy_import("fiftyone.core.brain")
@@ -55,26 +52,35 @@ _db_service = None
 #
 # All past and future versions of FiftyOne must be able to deduce the
 # database's current version and type from the `config` collection without an
-# error being raised so that migrations can be properly run and, if necsssary,
+# error being raised so that migrations can be properly run and, if necessary,
 # informative errors can be raised alerting the user that they are using the
 # wrong version or type of client.
 #
 # This is currently guaranteed because:
 #   - `DatabaseConfigDocument` is declared as non-strict, so any past or future
 #     fields that are not currently defined will not cause an error
-#   - All declared fields are optional and we have promised ourselves that
+#   - All declared fields are optional, and we have promised ourselves that
 #     their type and meaning will never change
 #
 
 
-class DatabaseConfigDocument(Document):
+@dataclasses.dataclass(init=False)
+class DatabaseConfigDocument:
     """Backing document for the database config."""
 
-    # strict=False lets this class ignore unknown fields from other versions
-    meta = {"collection": "config", "strict": False}
+    version: str
+    type: str
 
-    version = fof.StringField()
-    type = fof.StringField()
+    def __init__(self, conn, version=None, type=None, *args, **kwargs):
+        # Create our own __init__ so we can ignore extra kwargs
+        self._conn = conn
+        self.version = version
+        self.type = type
+
+    def save(self):
+        self._conn.config.replace_one(
+            {}, dataclasses.asdict(self), upsert=True
+        )
 
 
 def get_db_config():
@@ -83,19 +89,18 @@ def get_db_config():
     Returns:
         a :class:`DatabaseConfigDocument`
     """
+    conn = get_db_conn()
     save = False
-
-    try:
-        # pylint: disable=no-member
-        config = DatabaseConfigDocument.objects.get()
-    except moe.DoesNotExist:
-        config = DatabaseConfigDocument()
+    config_docs = list(conn.config.find())
+    if not config_docs:
         save = True
-    except moe.MultipleObjectsReturned:
-        cleanup_multiple_config_docs()
-
-        # pylint: disable=no-member
-        config = DatabaseConfigDocument.objects.first()
+        config = DatabaseConfigDocument(conn)
+    elif len(config_docs) > 1:
+        config = DatabaseConfigDocument(
+            conn, **cleanup_multiple_config_docs(conn, config_docs)
+        )
+    else:
+        config = DatabaseConfigDocument(conn, **config_docs[0])
 
     if config.version is None:
         #
@@ -129,39 +134,25 @@ def get_db_config():
     return config
 
 
-def cleanup_multiple_config_docs():
+def cleanup_multiple_config_docs(conn, config_docs):
     """Internal utility that ensures that there is only one
     :class:`DatabaseConfigDocument` in the database.
     """
 
-    # We use mongoengine here because `get_db_conn()` will not currently work
-    # until `import fiftyone` succeeds, which requires a single config doc
-    # pylint: disable=no-member
-    docs = list(DatabaseConfigDocument.objects)
-    if len(docs) <= 1:
-        return
+    if not config_docs:
+        return {}
+    elif len(config_docs) <= 1:
+        return config_docs[0]
 
     logger.warning(
         "Unexpectedly found %d documents in the 'config' collection; assuming "
         "the one with latest 'version' is the correct one",
-        len(docs),
+        len(config_docs),
     )
+    keep_doc = max(config_docs, key=lambda d: Version(d.get("version", 0)))
 
-    versions = []
-    for doc in docs:
-        try:
-            versions.append((doc.id, Version(doc.version)))
-        except:
-            pass
-
-    try:
-        keep_id = max(versions, key=lambda kv: kv[1])[0]
-    except:
-        keep_id = docs[0].id
-
-    for doc in docs:
-        if doc.id != keep_id:
-            doc.delete()
+    conn.config.delete_many({"_id": {"$ne": keep_doc["_id"]}})
+    return keep_doc
 
 
 def establish_db_conn(config):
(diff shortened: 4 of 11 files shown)


swheaton:
Looks like I need to fix doc build, should be easy enough

swheaton:
Fixing docs not as easy as I thought, it goes and pokes every single member of every class which ends up making an undesirable call within mongoengine. Various attempts to fix it unsuccessful so far... stay tuned

swheaton:
Ok trying something to see if it fixes doc build. Due to the way that autodoc works, I'm not sure how else to fix this except for what I did which is optionally do the old way in this specific circumstance.

fiftyone/__init__.py
```diff
@@ -25,9 +25,15 @@
 from fiftyone.__public__ import *
 
 import fiftyone.core.logging as _fol
-import fiftyone.migrations as _fom
 
-_fol.init_logging()
+# The old way of doing things, migrating database on import. If we
+#   REALLY need to do this, for example doc build, we can.
+if (
+    _os.environ.get("FIFTYONE_DISABLE_SERVICES", "0") != "1"
```
brimoor, line 32: @swheaton can you explain why the docs build is causing us problems here? Would love to avoid needing `_FIFTYONE_FORCE_DB_CONNECT_ON_IMPORT`

swheaton: Yeah I really hate it also 😫😫

Basically it is a mongoengine problem.

Autodoc tries to go in and touch every member and inherited member of each class. Which includes a field on the mongoengine document called "objects" which has a getter that requires the db connection 🤬

So it goes in, the connection has not been made, the "objects" getter fails, and then doc build fails.

I tried to play with autodoc settings including to ignore inherited members or specifically skip "objects", but didn't come up with anything that worked.

I'll try to think of other solutions, but brainstorming welcome

brimoor: I see. One way to clean up a bit would be to only call `migrate_database_if_necessary()` inside `establish_db_conn()`

brimoor: you tried added `exclude-members` [to the conf](https://github.com/voxel51/fiftyone/blob/85339fcecf03978435fed6fd94565d1c63f58dd6/docs/source/conf.py#L73
)? 


```py
autodoc_default_options = {
    ...
    "exclude-members": "objects"
}
```

fiftyone/__public__.py
```diff
@@ -5,16 +5,21 @@
...
 
 config = _foc.load_config()
 annotation_config = _foc.load_annotation_config()
 evaluation_config = _foc.load_evaluation_config()
 app_config = _foc.load_app_config()
 
-_foo.establish_db_conn(config)
+# The old way of doing things, connecting to database on import. If we
+#   REALLY need to do this, for example doc build, we can.
+if "_FIFTYONE_FORCE_DB_CONNECT_ON_IMPORT" in _os.environ:
```
brimoor, line 19: If the env var needs to stay, you could turn it into more of a feature by adding it to the [FiftyOneConfig](https://github.com/voxel51/fiftyone/blob/85339fcecf03978435fed6fd94565d1c63f58dd6/fiftyone/core/config.py#L63):

```py
        self._lazy_database_connection = self.parse_bool(
            d,
            "_lazy_database_connection",
            env_var="_FIFTYONE_LAZY_DATABASE_CONNECTION",
            default=True,
        )
```

Private parameters are not included in `print(fo.config)` or `$ fiftyone config` so it's both more cleanly organized and hidden by default.

Having said that, perhaps we should just make this a fully-documented public setting. It would have the benefit of allowing users to switch back to immediate db connections in the event that we discover some issue with lazy connections.

fiftyone/__init__.py
swheaton: Ok perhaps I do not understand sphinx directives like I thought I did, but adding to `conf.py` works where directly to the automodule directive did not.
I thought I had tried that too but oh well glad it works!

fiftyone/__public__.py
```diff
@@ -7,15 +7,12 @@
...
 config = _foc.load_config()
 annotation_config = _foc.load_annotation_config()
 evaluation_config = _foc.load_evaluation_config()
 app_config = _foc.load_app_config()
 
-_foo.establish_db_conn(config)
-
 from .core.aggregations import (
     Aggregation,
     Bounds,
```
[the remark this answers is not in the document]
swheaton: These imports are necessary to place the wanted methods and classes in the top level `fiftyone` module namespace

docs/source/conf.py
```diff
@@ -75,6 +75,7 @@
     "inherited-members": True,
     "member-order": "bysource",
     "autosummary": True,
+    "exclude-members": "objects",
```
brimoor, line 78: @swheaton one alternative here would be to exclude `mongoengine` subpackages altogether:
https://github.com/rdb/sphinx-autopackagesummary/issues/4#issuecomment-630211238

This would be nice because we intend to remove it as a dependency and therefore any inherited members really should not be documented anyway.

Also slight benefit that we can have `objects` members on classes if we want 🤓

brimoor: oh this is a different package than `sphinx.ext.autodoc`

brimoor: ah it is available for autodoc as well:
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports

```
autodoc_mock_imports = ["mongoengine"]
```

swheaton: tried it already, doesn't work, even if I try `fiftyone.core.odm.dataset.mongoengine` :(

When I researched why, I think I found that it only mocks if the package is NOT available, otherwise uses the available package. Which is not what we want in our case.

brimoor:
@swheaton speaking of `objects`, I did a quick grep of the codebase and we're still using mongoengine `objects` to access the database in one place:
https://github.com/voxel51/fiftyone/blob/85339fcecf03978435fed6fd94565d1c63f58dd6/fiftyone/core/odm/database.py#L87-L97

Can you update this to use pymongo? Necessary to remove `mongoengine` anyway and better that we explicitly control all database accesses.

docs/source/conf.py
swheaton: Even if you tell it `inherited-members=False`, it still tries to access inherited members ...

swheaton:
> @swheaton speaking of `objects`, I did a quick grep of the codebase and we're still using mongoengine `objects` to access the database in one place:
> 
> https://github.com/voxel51/fiftyone/blob/85339fcecf03978435fed6fd94565d1c63f58dd6/fiftyone/core/odm/database.py#L87-L97
> 
> Can you update this to use pymongo? Necessary to remove `mongoengine` anyway and better that we explicitly control all database accesses.

aha good find!

fiftyone/factory/repo_factory.py
```diff
@@ -15,11 +15,19 @@
...
+def _get_db():
+    global _db
+    if _db is None:
+        _db = foo.get_db_conn()
+    return _db
 
 
 class RepositoryFactory(object):
     repos = {}
+    db = None
```
benjaminpkane, line 30: nit: is this used?

benjaminpkane:
> @swheaton speaking of `objects`, I did a quick grep of the codebase and we're still using mongoengine `objects` to access the database in one place:
> 
> https://github.com/voxel51/fiftyone/blob/85339fcecf03978435fed6fd94565d1c63f58dd6/fiftyone/core/odm/database.py#L87-L97
> 
> Can you update this to use pymongo? Necessary to remove `mongoengine` anyway and better that we explicitly control all database accesses.

`fiftyone.core.odm.database.cleanup_multiple_config_docs` is also relying on `objects`, I believe.
* https://github.com/voxel51/fiftyone/blob/85339fcecf03978435fed6fd94565d1c63f58dd6/fiftyone/core/odm/database.py#L139

fiftyone/factory/repo_factory.py
swheaton: no I think this was a remnant of a different approach. I really didn't like how this class was/is...

tests/no_wrapper/multiprocess_tests.py
```diff
@@ -15,14 +15,18 @@
 
 class MultiprocessTest(unittest.TestCase):
     def test_multiprocessing(self):
-        with multiprocessing.Pool(1, _check_process) as pool:
-            for _ in pool.imap(_check_process, [None]):
+        food.establish_db_conn(fo.config)
+        port = food._connection_kwargs["port"]
+        with multiprocessing.Pool(2, _check_process, [port]) as pool:
```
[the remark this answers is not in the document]
swheaton: No the database connection is part of the test so if it fails I want the test to fail.

swheaton:
> > @swheaton speaking of `objects`, I did a quick grep of the codebase and we're still using mongoengine `objects` to access the database in one place:
> > https://github.com/voxel51/fiftyone/blob/85339fcecf03978435fed6fd94565d1c63f58dd6/fiftyone/core/odm/database.py#L87-L97
> > 
> > Can you update this to use pymongo? Necessary to remove `mongoengine` anyway and better that we explicitly control all database accesses.
> 
> `fiftyone.core.odm.database.cleanup_multiple_config_docs` is also relying on `objects`, I believe.
> 
> * https://github.com/voxel51/fiftyone/blob/85339fcecf03978435fed6fd94565d1c63f58dd6/fiftyone/core/odm/database.py#L139

Yes I am looking into that one as well ... there are circular reference pitfalls I have to navigate between databaseconfig, database, and document. May have to end up keeping DatabaseConfig out of mongoengine.

benjaminpkane:
> @benjaminpkane thanks for looking at this. Do you have any ideas of additional use cases to test? I'm keeping the description updated with all the things I've tried.

Description is very thorough! I am not seeing any issues in my testing. Much simpler than expected!

> For one, due to the above, I've found that simply saving a document does not initialize the connection first. Idk if there's a legit use case to do this, but I don't want to break any behavior that previously worked.

Interesting, I will have to try that

> Maybe something around notebooks / Databricks / etc. ?

I don't believe there should be anything unique about notebooks / Databricks. But testing in a Teams deployment might be wise before merging

swheaton:
Latest commit a9a840888df5ef41fcc9322f7117c128a67319f7 removes the references to `objects` described above, as well as allows our mongoengine documents (`fiftyone.core.odm.document.Document`) to be saved without a db connection first - now it ensures connection first, which is a no-op unless it's the very first time.

This didn't work previously.
```python
import fiftyone as fo
ws = fo.core.odm.WorkspaceDocument(name="blahhhhh", slug="blahhhhh", child=fo.core.odm.Space())
ws.save()
ws.delete()
```

tests/unittests/utils_tests.py
```diff
@@ -444,16 +444,14 @@ def test_multiple_config_cleanup(self):
...
 
         # Ensure that duplicate documents are automatically cleaned up
         config = foo.get_db_config()
 
         self.assertEqual(len(list(db.config.aggregate([]))), 1)
-        self.assertEqual(config.id, orig_config.id)
+        self.assertEqual(config, orig_config)
 
 
 from bson import ObjectId
```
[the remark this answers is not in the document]
swheaton: you're right but i don't want to touch that in this PR

fiftyone/core/odm/database.py
```diff
@@ -128,39 +134,25 @@ def get_db_config():
...
+    elif len(config_docs) <= 1:
+        return config_docs[0]
 
     logger.warning(
         "Unexpectedly found %d documents in the 'config' collection; assuming "
         "the one with latest 'version' is the correct one",
-        len(docs),
+        len(config_docs),
     )
+    keep_doc = max(config_docs, key=lambda d: Version(d.get("version", 0)))
```
benjaminpkane, line 152: This seems dangerous to potentially revert to a version `0`. It's never happened to my knowledge, but currently if no valid config document is found, a hard exception occurs

benjaminpkane (commented):
Changes LGTM! For peace of mind, what are your thoughts on adding extra coverage to `cleanup_multiple_config_docs` asserting a hard exception occurs when >=1 config docs are present, but none have a valid `version` value?

fiftyone/core/odm/database.py
swheaton: Hmm well if there's a hard error then that would prevent migration for the version to be fixed. So you'd be stuck.

I will test it, I think it should be treated the same as no config doc (v0.14.4)

swheaton: If there's no config document it assumes v0.14.4 but doesn't hard error.

benjaminpkane: Right, but if there is >=1 docs, all without a version value, the behavior has changed. But it's and odd situation

benjaminpkane (approved):
LGTM 🚀 🐎

fiftyone/core/odm/database.py
swheaton: Just tested on `develop` and behavior has not changed, what I said above is true for this case.
However I will push a commit because an integer is no bueno for `Version`, so we will use `"0.0"`.

swheaton: fixed/pushed

fiftyone/core/odm/database.py
```diff
@@ -54,26 +52,35 @@
...
+        self._conn = conn
+        self.version = version
+        self.type = type
 
-    version = fof.StringField()
-    type = fof.StringField()
+    def save(self):
+        self._conn.config.replace_one(
+            {}, dataclasses.asdict(self), upsert=True
+        )
```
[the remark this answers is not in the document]
swheaton: I do not want the `__init__` autogenerated for the reason in the comment
```
# Create our own __init__ so we can ignore extra kwargs
```

fiftyone/core/odm/database.py
```diff
@@ -54,26 +52,36 @@
...
```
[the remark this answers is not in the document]
swheaton: This is a duplicate comment that was already resolved when you reviewed [59e8444](https://github.com/voxel51/fiftyone/pull/4236/commits/59e8444c6b4946e2f77246603cea6036a8f65620)
Please do not repeat comments you have previously made.

PR #4244, opened 2024-04-08
Export/model inference cleanup (for v0.23.8)

brimoor:
## Change log

- Autosave contexts now respect the default batching strategy and can be configured to use content size-based batching
- All SDK methods now use `autosave=True` or its equivalent rather than calling `sample.save()` in a loop
- Renames `export_media` parameter to `write_clips` in video sample parsers, for clarity
- Centralizes the application of `select_fields()` in `compute_embeddings()` and `compute_patch_embeddings()`, for consistency with how this is handled in the `apply_model()` utility.
- Adds support for sample collections to `fou.iter_slices()`
- Adds a `SampleCollection._resolve_media_field()` utility
- Adds a `fos.read_files()` method to efficiently read from multiple files in a threadpool
- Adds a warning when uploading large amounts of data to CVAT
- Removes unneeded `pylint: disable=no-member`

### DIFF
--- a/docs/source/release-notes.rst
+++ b/docs/source/release-notes.rst
@@ -9,6 +9,15 @@ FiftyOne Teams 1.5.9
 
 Includes all updates from :ref:`FiftyOne 0.23.8 <release-notes-v0.23.8>`, plus:
 
+- :ref:`Download contexts <teams-cloud-media-python>` now support batching
+  based on content size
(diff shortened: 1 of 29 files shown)


swheaton (approved):
no test instructions but i combed through code - lot of applying same thing to different situations. Wish it were simpler but it is what it is. approve

PR #4249, opened 2024-04-09
Support Labelbox v2 export APIs and formats

ehofesmann:
At the end of April 2024, Labelbox is removing support for their v1 exports which the FiftyOne integration utilized: https://docs.labelbox.com/docs/migration-guide-export-v1-to-export-v2

This PR adds support to the v2 export format and API. Specifically, the main changes include:
* Removal of `project.export_labels()`
* Removal of `batch.export_data_rows()`
* Support for parsing v2 formatted label exports
     * This specifically was done by reformatting much of the code that converts Labelbox label formats to FiftyOne label formats. Adding the relevant methods into `_LabelboxExportToFiftyOneConverterV1` and `_LabelboxExportToFiftyOneConverterV2`
     * It also created the `_FiftyOneToLabelboxExportConverterV1` that converts FiftyOne to Labelbox labels, but the v2 of this was not implemented since the integration doesn't support uploading labels from FiftyOne to Labelbox anyway
         * The only methods where this is relevant are the utilities `import_from_labelbox`, `export_to_labelbox`, and `convert_labelbox_export_to_import`, but for now these just have a deprecation warning added to them since they are not part of the core integration.

To test:
* Go to app.labelbox.com and create an account/sign in
* Create a labelbox API key through their App
* `export FIFTYONE_LABELBOX_API_KEY=...` and `pip install labelbox`
* Test the integration:
```python
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=3).clone()
label_schema = {
    "new_field": {
        "type": "detections", 
        "classes": ["dog", "cat"],
        "attributes": {"test": {"type": "checkbox", "values": ["testattr"]}},
    }
}
results = dataset.annotate(
    "anno_key",
    label_schema=label_schema,
    backend="labelbox",
    launch_editor=True,
)

# Edit in Labelbox

dataset.load_annotations("anno_key", cleanup=True)
```

This was tested by manually running through the examples in [the integration docs](https://docs.voxel51.com/integrations/labelbox.html) as well as testing other edge cases. Found and fixed an additional issue with the `select` attribute type having been deprecated.
I also ran into [this bug](https://github.com/voxel51/fiftyone/issues/4255) which can be fixed in a follow up.

## Summary by CodeRabbit

- **Documentation**
	- Updated integration guide for Labelbox: removed note about using Model Assisted Labeling with a paid Labelbox account, changed attribute types from `select` to `checkbox` and `radio`.
	- Added information on editing existing labels, and clarified instructions on deleting projects and datasets in Labelbox.

- **Bug Fixes**
	- Corrected the environment variable name in the API key setup prompt message.

### DIFF
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -2279,7 +2279,7 @@ class AnnotationAPI(object):
             prefix = "FIFTYONE_%s_" % backend.upper()
             logger.info(
                 "Please enter your API key.\nYou can avoid this in the future "
-                "by setting your `%sKEY` environment variable",
+                "by setting your `%sAPI_KEY` environment variable",
                 prefix,
             )
 
--- a/docs/source/integrations/labelbox.rst
+++ b/docs/source/integrations/labelbox.rst
@@ -743,7 +743,7 @@ The meaning of the `mutable` attribute is defined as follows:
 -   `True` (default): the attribute is dynamic and can have a different value
     for every frame in which the object track appears
 -   `False`: the attribute is static and is the same for every frame in which
-    the object track appears
+    the object track appears (**Not yet supported**)
 
 .. _labelbox-loading-annotations:
 
--- a/docs/source/integrations/labelbox.rst
+++ b/docs/source/integrations/labelbox.rst
@@ -487,6 +487,8 @@ can also be provided:
     top level of the editor (False) or whether to show the label field at the
     top level and annotate the class as a required attribute of each object
     (True)
+-   **export_version** (*"v2"*): the Labelbox export format and API version to
+    use. Suppoted values are `("v1", "v2")`
 
 .. note::
 
@@ -680,8 +682,6 @@ For Labelbox, the following `type` values are supported:
 -   `text`: a free-form text box. In this case, `values` is unused
 -   `radio`: a radio button list UI. In this case, `values` is required
 -   `checkbox`: a list of checkboxes. In this case, `values` is required
-
-(The `select` type has been deprecated and removed by Labelbox)
 
 When you are annotating existing label fields, the `attributes` parameter can
 take additional values:
@@ -1481,10 +1481,6 @@ or
 methods to delete specific Labelbox project(s) associated with an annotation
 run.
 
-Note that if setting `delete_batches=True` when deleting projects, then the
-corresponding datarows will no longer be deleted from Labelbox since the
-transition to Labelbox export v2.
-
 .. code:: python
     :linenos:
 
@@ -1522,3 +1518,9 @@ transition to Labelbox export v2.
     # Delete all projects and datasets from your Labelbox account
     api.delete_projects(project_ids)
     api.delete_datasets(dataset_ids)
+
+.. note::
+
+    Note that passing `delete_batches=True` when deleting projects will not
+    delete the corresponding data rows from Labelbox when using the V2 export
+    API (the default).
--- a/docs/source/integrations/labelbox.rst
+++ b/docs/source/integrations/labelbox.rst
@@ -488,7 +488,7 @@ can also be provided:
     top level and annotate the class as a required attribute of each object
     (True)
 -   **export_version** (*"v2"*): the Labelbox export format and API version to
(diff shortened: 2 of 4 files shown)


allenleetc:
Testing WIP, a couple notes/questions

- To confirm, label attributes that aren’t specified in LB are just not added at all, vs added with a `None` value
- To confirm, label attributes aren't added to the dataset's dynamic field schema
- Polylines that aren't closed in LB are closed when loaded back into FO

fiftyone/utils/annotations.py
```diff
@@ -2279,7 +2279,7 @@ def _prompt_api_key(self, backend):
             prefix = "FIFTYONE_%s_" % backend.upper()
             logger.info(
                 "Please enter your API key.\nYou can avoid this in the future "
-                "by setting your `%sKEY` environment variable",
+                "by setting your `%sAPI_KEY` environment variable",
```
brimoor, line 2282: @ehofesmann this is a universal bug, not just labelbox, eh? Proposing to pull it into `release/v0.23.8` immediately just in case here: https://github.com/voxel51/fiftyone/pull/4256

allenleetc: Believe universal yes for annotation integrations

brimoor: ah you added this @allenleetc, thanks 💪

ehofesmann: Yes, universal for all API key-based backends it seems (everything but CVAT). Good catch Allen!

ehofesmann:
Thanks for the detailed testing @allenleetc !!

> Testing WIP, a couple notes/questions
> 
> * To confirm, label attributes that aren’t specified in LB are just not added at all, vs added with a `None` value

* Right, this is expected and also how other annotation integrations work (CVAT at least). The label attributes just get parsed directly from what is returned by the annotation backend, there is no additional check done against the original label schema to populate `None` values.

> * To confirm, label attributes aren't added to the dataset's dynamic field schema

* Correct, that could cause a lot of unwanted clutter in the dataset, so adding attributes to the field schema should be left up to the user.


> * Polylines that aren't closed in LB are closed when loaded back into FO



* Good catch! That seems to have been the case the entire time, but I agree that polylines really should be open since there is an explicit polygon annotation tool. Will update so that polylines are open and not filled always.

allenleetc:
A little more testing

- Video datasets, does anything special have to be done on to populate `.index` for new objects/tracks when labeling videos? Given the “track” interface in LB i thought `.index` would be auto-populated
- In the same vein the [video-level attributes ](https://docs.voxel51.com/integrations/labelbox.html#video-label-attributes) didn't seem to have an effect when labeling, I seemed to be able to change immutable attributes

allenleetc:
@ehofesmann added some test cases I used to `tests/intensive/labelbox_tests.py`, please feel free to accept/reject/edit etc.

allenleetc (approved):
Did a decent amount of testing, LGTM

ehofesmann:
@allenleetc 
> A little more testing
> 
> * Video datasets, does anything special have to be done on to populate `.index` for new objects/tracks when labeling videos? Given the “track” interface in LB i thought `.index` would be auto-populated
> * In the same vein the [video-level attributes ](https://docs.voxel51.com/integrations/labelbox.html#video-label-attributes) didn't seem to have an effect when labeling, I seemed to be able to change immutable attributes
> 
> My testing is done, I'll add a commit with some testpoints but otherwise @ehofesmann do we still need a code review?

Right, there is no notion of tracks at the moment when loading back video labels from Labelbox through this integration, though I believe there may be a unique ID returned by [Labelbox video annotations]() that we could use to link them to the same index in the future.

As for the video attributes, I'm going to update that section, you're right that the `mutable` attribute is not actually implemented/used at the moment. And I'm only skeptical that Labelbox actually supports these [object-level scopes for immutable attributes](https://docs.labelbox.com/docs/video-annotations#supported-annotation-types).

The tests you added are super helpful! Thank you!

ehofesmann:
Closing in favor of: https://github.com/voxel51/fiftyone/pull/4260

PR #4256, opened 2024-04-11
Fix API key env var hint

brimoor:
Pulls in a universal fix ID'd by @ehofesmann in https://github.com/voxel51/fiftyone/pull/4249


ehofesmann (approved):
LGTM!

PR #4262, opened 2024-04-12
python panel fixes and enhancements

imanjra:

- Fix Python panel layout bounds
- add loading state
- Persist dynamically registered panel across session refresh
- Fix MarkdownView not rendering default value

Using a python panel operator

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/spaces/src/hooks.ts
+++ b/app/packages/spaces/src/hooks.ts
@@ -1,6 +1,10 @@
-import { PluginComponentType, useActivePlugins } from "@fiftyone/plugins";
+import {
+  PluginComponentType,
+  subscribeToRegistry,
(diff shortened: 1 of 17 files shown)


app/packages/spaces/src/hooks.ts
```diff
@@ -100,6 +104,17 @@ export function usePanel(name: SpaceNodeType) {
   return panels.find((panel) => panel.name === name);
 }
 
+export function useReactivePanel(name: SpaceNodeType) {
```
ritch, line 107: Add comment explaining why this is needed because the `count` variable here is going to throw us off int he future.

ritch (approved):

PR #4281, opened 2024-04-17
Linking and New Grid

benjaminpkane:

Adds a new `@fiftyone/spotlight` package and sample/group linking

-   [ ] Other


fiftyone/core/session/session.py
```diff
@@ -304,10 +315,15 @@ class Session(object):
...
-            defining a space configuration to load
-        color_scheme (None): an optional :class:`fiftyone.core.odm.dataset.ColorScheme`
-            defining a custom color scheme to use
+        group_id (None): an optional expanded
+            :class:`fiftyone.core.groups.Group` id
+        sample_id (None): an optional expanded
+            :class:`fiftyone.core.sample.Sample` id
+        spaces (None): an optional :class:`fiftyone.core.odm.workspace.Space`
+            instance defining a space configuration to load
+        color_scheme (None): an optional
```
benjaminpkane, line 324: alphabetize

e2e-pw/src/oss/poms/url.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  constructor(
+    private readonly page: Page,
+    private readonly eventUtils: EventUtils
+  ) {
+    this.assert = new UrlAsserter(this);
+  }
+
+  get groupSlice() {
+    return this.searchParams.get("groupSlice");
```
benjaminpkane, line 15:
```suggestion
    return this.searchParams.get("slice");
```

app/packages/core/src/components/Grid/useAt.ts
```diff
@@ -0,0 +1,62 @@
...
+          at: {
+            description,
+            offset: snapshot.getLoadable(gridOffset).getValue(),
+          },
+          key,
+        };
+      },
+    [getKey]
+  );
+  const set = useRecoilTransaction_UNSTABLE(
```
benjaminpkane, line 51: Scroll end events trigger a new stateful grid location. This enables resizing/tagging/spacing/zooming to return to the last grid location

app/packages/core/src/components/Grid/useRefreshers.ts
```diff
@@ -0,0 +1,59 @@
+import { subscribe } from "@fiftyone/relay";
+import * as fos from "@fiftyone/state";
+import { useEffect, useMemo } from "react";
+import uuid from "react-uuid";
+import { useRecoilValue } from "recoil";
+import { gridPage } from "./recoil";
+
+export default function useRefreshers() {
```
benjaminpkane, line 8: A refresh resets either the layout or the page _and_ layout

app/packages/core/src/components/Grid/useSpotlightPager.ts
```diff
@@ -0,0 +1,129 @@
...
+  const refresher = useRecoilValue(fos.refresher);
+  useEffect(() => {
+    refresher;
+    const current = records.current;
+    return () => {
+      commitLocalUpdate(fos.getCurrentEnvironment(), (store) => {
+        for (const id of Array.from(current.keys())) {
+          store.get(id)?.invalidateRecord();
+        }
+        current.clear();
```
benjaminpkane, line 121: On refresher changes, e.g. from tagging, invalidate relay records so new network requests for samples are issued

app/packages/relay/src/Writer.tsx
```diff
@@ -10,14 +10,15 @@ import { datasetQuery } from "./queries";
 import { SelectorEffectContext, Setter } from "./selectorWithEffect";
 
 export interface PageQuery<T extends OperationType> {
+  event?: "modal";
```
benjaminpkane, line 13: New `event` attribute signaling a modal page change has occurred

app/packages/core/src/components/Grid/Grid.tsx
```diff
@@ -1,191 +1,165 @@
...
+import useEscape from "./useEscape";
+import useRefreshers from "./useRefreshers";
+import useSelect from "./useSelect";
+import useSelectSample from "./useSelectSample";
+import useSpotlightPager from "./useSpotlightPager";
+import useThreshold from "./useThreshold";
+
+function Grid() {
+  const id = useMemo(() => uuid(), []);
+  const lookerStore = useMemo(() => new WeakMap<ID, Lookers>(), []);
```
sashankaryal, line 27: Can take it offline and flag for later, but I'm curious about the usage of `WeakMap` here with ID has the key...I feel like we aren't getting much out of it, and if anything, losing access to iterators, etc.

IIRC, weak maps/sets are useful because they don't add strong refs to the _keys_ (not values), and when all refs to the keys are lost, the _key_ is garbage collected along with the value (assuming no remaining strong refs to the value, which in this case is unlikely. I thought we'd be concerned about GC of looker instances, not IDs). 

```
const looker = new Looker();

const desc = {description};

lookerStore.add(desc, looker);
looker = null; // no remaining refs to looker except in the weakmap
// the following still yields looker instance
// because we still have a strong ref to looker in the weakmap (because it's a value).
// if this is the intended effect, then ignore my comment 😅 
lookerStore.get(desc); 
```

app/packages/core/src/components/Grid/recoil.ts
```diff
@@ -7,21 +7,43 @@ export const defaultGridZoom = selector<number>({
...
+  key: "gridOffset",
+  default: 0,
+  effects: [],
+});
+
+export const gridSpacing = atom({
+  key: "gridSpacing",
+  default: 3,
+  effects: [
+    fos.getBrowserStorageEffectForKey("gridSpacing", { valueClass: "number" }),
```
sashankaryal, line 30: nit: tooltip needs update too ;)

app/packages/core/src/components/Grid/useRefreshers.ts
sashankaryal: Would love to have these comments in the code itself...doesn't have to be polished and formal.

app/packages/core/src/components/Grid/useSpotlightPager.ts
sashankaryal: same comment as above, would love to see these comments in the code

app/packages/spotlight/src/closest.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Copyright 2017-2024, Voxel51, Inc.
+ */
+
+import { ONE, TWO, ZERO } from "./constants";
+
+export function closest<I>(
+  rows: I[],
```
sashankaryal, line 8: 🎨 
this one could probably use some unit tests

e2e-pw/src/oss/poms/url.ts
```diff
@@ -0,0 +1,73 @@
+import { expect, Page } from "src/oss/fixtures";
+import { EventUtils } from "src/shared/event-utils";
+
+export class UrlPom {
+  readonly assert: UrlAsserter;
```
sashankaryal, line 5: 🧞

sashankaryal (approved):
looks good, tested locally and works great!!! would be nice if we could sprinkle in more documentation though, or else it'll look esoteric in the future. 😅  

one minor comment re: weak maps, but most likely I'm wrong there myself and def not high priority for now.

🚀 🚀 🚀 🚀

app/packages/app/src/useEvents/useSetSample.ts
```diff
@@ -0,0 +1,58 @@
...
+  search.set("id", id);
+};
+
+const useSetSample: EventHandlerHook = ({ router }) => {
+  const setModalState = useSetModalState();
+  const setter = useSessionSetter();
+
+  return useCallback(
+    (payload) => {
+      setModalState().then(() => {
```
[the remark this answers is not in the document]
benjaminpkane: Yes, will do 💯

app/packages/core/src/components/Grid/Grid.tsx
benjaminpkane: So I will admit there are not tests that assert this setup. And it is an initial foray into this pattern...

The intention of the design is that only `@fiiftyone/spotlight` data structures, e.g. a `Row`, contain strong references to `ID` values. And the implication is rows are garbage collected when a `Spotlight` contracts, e.g. drops its backward section, mutates its forward section as the new backward section, and creates a fresh (empty) forward section.

So, in theory, the `ID`s are garbage collected through scrolling and `Looker` instances (and sample data in the other `WeakMap`) are garbage collected, as well

Though adding tests/better assurances around this is the next step 🚀

PR #4282, opened 2024-04-17
Validate 'tags' argument of match_tags()

ehofesmann:

Adds validation to the `SampleCollection.match_tags()` method to ensure the `tags` argument is a string or iterable of strings. Otherwise, it raises an obscure error like this if you accidentally pass in, say, a `DatasetView`:
```
InvalidDocument: cannot encode object: <SampleView: {
    'id': '662001ac9379aa4cf021c06e',
    'media_type': 'image',
    'filepath': '/path/to/image.jpg',
    'tags': [],
    'metadata': None,
}>, of type: <class 'fiftyone.core.sample.SampleView'>
```

Tested by calling match_tags() with various types of input for the `tags` argument.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced validation for the `tags` argument to ensure it is either a string or an iterable of strings, improving data integrity.

### DIFF
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -5356,7 +5356,7 @@ class MatchTags(ViewStage):
             tags = [tags]
         else:
             tags = list(tags)
-            if not builtins.all([etau.is_str(t) for t in tags]):
+            if not builtins.all(etau.is_str(t) for t in tags):
                 raise ValueError(
                     "The `tags` argument must be a string or iterable of "
                     "strings."
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -6,6 +6,7 @@ View stages.
 |
 """
 
+import builtins
 from collections import defaultdict, OrderedDict
 import contextlib
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


fiftyone/core/stages.py
```diff
@@ -5355,6 +5356,11 @@ def __init__(self, tags, bool=None, all=False):
             tags = [tags]
         else:
             tags = list(tags)
+            if not builtins.all([etau.is_str(t) for t in tags]):
```
swheaton, line 5359: you should not have to import `builtins` like this. See `SortBy.validate` for a similar validation. Unless the `all` property on this class is causing issues?
also minor but all() takes any iterator and so does not need to be a list comprehension
```suggestion
            if not all(etau.is_str(t) for t in tags):
```

ehofesmann: The problem is that the MatchTags has an `all` kwarg that overwrites the builtin. The fact that it can take an iterator is good to know though!

swheaton: ah ok didnt see that kwarg. minor suggestion then
```suggestion
            if not builtins.all(etau.is_str(t) for t in tags):
```

swheaton (approved):
lgtm. minor suggestion take or leave

fiftyone/core/stages.py
ehofesmann: I like it! Thanks Stuart

PR #4288, opened 2024-04-18
Hugging Face Hub Integration Upgrades

jacobmarks:
Mainly makes the uploading and downloading of media files in `push_to_hub()` and `load_from_hub()` much more efficient, smart, and informative. For compatibility with Hugging Face repos, this PR creates a new dataset type in FiftyOne called `FiftyOneChunkedDataset` which allows you to split media files into subdirectories of specified size rather than putting all files directly into one `data` directory.

(Details)

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

- **New Features**
  - Added the ability to specify a `chunk_size` parameter for dataset chunking, nested directories, and efficient management of large datasets.
  - Introduced batch download mechanism for media files to improve performance.

- **Enhancements**
  - Improved dataset configuration and management for `FiftyOneChunkedDataset`.
  - Renamed dataset category from `semantic-segmentation` to `image-segmentation` for clarity.

- **Refactor**
  - Optimized data uploading methods to efficiently handle media files.

### DIFF
--- a/docs/source/integrations/huggingface.rst
+++ b/docs/source/integrations/huggingface.rst
@@ -1021,10 +1021,10 @@ and then passing that in to
 
     import fiftyone.utils.huggingface as fouh
 
-    # create view with high confidence predictions
+    # Create view with high confidence predictions
     view = dataset.filter_labels("predictions", F("confidence") > 0.95)
 
-    # push view to the Hub as a new dataset
+    # Push view to the Hub as a new dataset
     fouh.push_to_hub(view, "my-quickstart-high-conf")
 
 When you do so, note that the view is exported as a new dataset, and other 
@@ -1047,7 +1047,8 @@ either a relative or absolute path to the preview file on your local machine:
     dataset = foz.load_zoo_dataset("quickstart")
 
     session = fo.launch_app(dataset)
-    # Screenshot and save the preview image to a file
+
+    # Screenshot and save the preview image to a file...
 
     fouh.push_to_hub(
         dataset,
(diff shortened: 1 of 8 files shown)


brimoor (changes requested):
How about instead just adding an optional `chunk_size` parameter to `FiftyOneDatasetExporter` that enables the batched `data/` folder that you're after?

This way of exporting media would be equally applicable to all dataset formats, not just `FiftyOneDataset`. I'd just pass the `chunk_size` parameter through to the existing `MediaExporter` class so that any other data format could support nested subfolders just by passing this `chunk_size` parameter through as well. 

As you noticed, `FiftyOneDatasetImporter` already "knows" how to import media regardless of how it is organized within `data/`, so it's really still "FiftyOneDataset" format. This also applies to all other builtin importers: the way they are generally implemented, they'll gracefully handle nested subdirectories too.

jacobmarks:
> How about instead just adding an optional `chunk_size` parameter to `FiftyOneDatasetExporter` that enables the batched `data/` folder that you're after?
> 
> This way of exporting media would be equally applicable to all dataset formats, not just `FiftyOneDataset`. I'd just pass the `chunk_size` parameter through to the existing `MediaExporter` class so that any other data format could support nested subfolders just by passing this `chunk_size` parameter through to their `MediaExporter` instance as well.
> 
> As you noticed, `FiftyOneDatasetImporter` already "knows" how to import media regardless of how it is organized within `data/`, so it's really still "FiftyOneDataset" format. This also applies to all other builtin importers: the way they are generally implemented, they'll gracefully handle nested subdirectories too.

Good idea. Let me try this :)

brimoor (approved):
LGTM 🚀

PR #4299, opened 2024-04-22
Fix dynamically grouping group datasets in the App

benjaminpkane:

Fixes a regression in 0.23.8 due to https://github.com/voxel51/fiftyone/pull/4097. The behavior of filtering was improved https://github.com/voxel51/fiftyone/pull/4097, but dynamic groups was overlooked 😞 

In light of this! A new `groupView` selector has been added that supplies the correct representation of a view for group requests, i.e. all slices of group in **group dataset**. The only case for modification presently is when the group dataset is _dynamically grouped_, and the `GroupBy` stage is filtered.

In reviewing the functionality, I took the opportunity to move **dynamic** group state in `@fiftyone/state` to a new `dynamicGroups` module, as it is distinct from **group dataset** state (yet they _can_ interact as features, e.g. the above)

### Test code

```python
import fiftyone as fo
import fiftyone.zoo as foz

step = 25
for slice in dataset.group_slices:
    dataset.group_slice = slice
    scene_id = 0
    order_by = 0
    for sample in dataset:
        sample.set_field("scene_id", scene_id // step)
        sample.set_field("timestamp", order_by % step)
        sample.save()
        scene_id += 1
        order_by += 1

dataset.save()
``` 

Will also review offline with @sashankaryal 

In progress

## Release Notes

* Fixed dynamic groups of group datasets in the App

-   [ ] Other

### DIFF
--- a/app/packages/core/src/components/Modal/Group/DynamicGroup/carousel/DynamicGroupsFlashlightWrapper.tsx
+++ b/app/packages/core/src/components/Modal/Group/DynamicGroup/carousel/DynamicGroupsFlashlightWrapper.tsx
@@ -30,7 +30,7 @@ const pageParams = selector({
 
     const params = {
       dataset,
-      view: get(fos.dynamicGroupViewQuery({})),
+      view: get(fos.dynamicGroupViewQuery(null)),
     };
 
     return (page: number, pageSize: number) => {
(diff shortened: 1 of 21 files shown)


sashankaryal:
I was doing some functional testing and noticed something... In dynamic groups of groups, only the first (G)roup behaves expectedly. Intra-(G)roup navigation seems to be broken starting the second big group. I see the right network requests but the UI doesn't update.

benjaminpkane:
> I was doing some functional testing and noticed something... In dynamic groups of groups, only the first (G)roup behaves expectedly. Intra-(G)roup navigation seems to be broken starting the second big group. I see the right network requests but the UI doesn't update.

Ahh ok, I will make a pass on that. Thanks! 🙏

sashankaryal:
Verified working, thanks!

sashankaryal (approved):
lgtm

PR #4304, opened 2024-04-23
Fix to load on Linux filepaths generated on Windows

Buligon:

When importing datasets, change the '\' to '/' in the file path so it will load properly on Linux. Currently, Windows generate those paths, that can't be opened on Linux.

I've tested it doing the following process on Windows and Linux:
1. Import the dataset;
2. See how the images are loaded on the app;
3. Change the importers.py so it changes the slashes;
4. Import the dataset again and see how the images are loaded.

On Linux, during the step 2, the images weren't loaded properly due to the path, but after the change they were.

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved file path handling in sample import function to ensure compatibility across different operating systems.

### DIFF
--- a/fiftyone/utils/data/importers.py
+++ b/fiftyone/utils/data/importers.py
@@ -1904,7 +1904,7 @@ class FiftyOneDatasetImporter(BatchDatasetImporter):
 
         def _parse_sample(sd):
             if not os.path.isabs(sd["filepath"]):
-                sd["filepath"] = os.path.join(rel_dir, sd["filepath"])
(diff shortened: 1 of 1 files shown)


brimoor (approved):
Good catch, we should definitely handle localizing file separators to the current OS 👍 

I'm going to add a follow-on upgrade here to use `fos.normpath()` of `replace()`, which is a bit more robust 😄

PR #4309, opened 2024-04-25
fix: coco category ids can now be not sequential - avoiding memory leak

### LINKED ISSUE #4293: [BUG] Memory leak when importing CocoDetection Dataset with random category_id into FiftyOne
### Describe the problem

When adding a dataset into fiftyone with the following code, the cell run until it crashes the notebook because of an out-of-memory error.
This happens if the dataset contains random category ids (instead of ids ranging from `0` to `num_categories-1`).
Fiftyone should either handle this case (random category ids), or write a clear error message (and definitely not eat all the memory until it gets killed).

### Code to reproduce issue

```python
ds = fo.Dataset.from_dir(
    name=f"name",
    overwrite=True, 
    persistent=True, 
    data_path='path_to_images',  
    labels_path='path_to_json', 
    label_types=['segmentations'],
    use_polylines=False, 
    include_id=True,
)
```

Example of a json file that will lead to this issue (you have to have an image with filename `image.jpg`):
```json
{
  "categories": [
    {
      "id": 74673216,
      "name": "Solar Panel",
      "supercategory": "root"
    }
  ],
  "images": [
    {
      "id": 789,
      "file_name": "image.jpg",
      "height": 1024,
      "width": 1024,
      "extra": {}
    }
  ],
  "annotations": [
       {
      "id": 15567,
      "image_id": 789,
      "category_id": 74673216,
      "segmentation": [
        [
          445.6344437096958,
          354.34683544084845,
          452.1216184704017,
          348.23038026732564,
          411.9939701135346,
          303.93243513679033,
          406.06287473360334,
          310.2342501039047
        ]
      ],
      "area": 521.9080653650577,
      "bbox": [
        406.06287473360334,
        303.93243513679033,
        46.05874373679836,
        50.41440030405812
      ],
      "iscrowd": 0,
      "extra": {}
    }
  ]
}
```

### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 22.04): Ubuntu 20.04.6
-   **Python version** (`python --version`): 3.11.6
-   **FiftyOne version** (`fiftyone --version`): 0.23.7 

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community

patrontheo:
Closes #4293.

`parse_coco_categories` returns `classes_map` instead of `classes`. It is a dict mapping class IDs to class labels. In areas where the list of classes are needed, it it obtained through `list(classes_map.values())`.

The caveat I see here is that the public facing `add_coco_labels` function takes classes as an argument (the list of class labels). To avoid changing the public interface, I did not modify that and built `classes_map`using `classes_map = {i: label for i, label in enumerate(classes)}`. This should work but requires that in `labels_or_path` files, the category IDs are sequential (starting from zero). This was already the case before this change.
If we want to allow non-sequential category IDs here also, we need to change the public interface and require `classes_map`instead of `classes`. Or do you see another solution ? Should we change the public interface or leave it like that ?

Not thoroughly. I am opening a draft PR to run the tests, as explained in [Contributing](https://github.com/voxel51/fiftyone/blob/develop/CONTRIBUTING.md).

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced label handling in COCO dataset utilities by introducing a mapping of class IDs to labels for improved accuracy and flexibility.

### DIFF
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -129,7 +129,8 @@ def add_coco_labels(
             will be created if necessary
         labels_or_path: a list of COCO annotations or the path to a JSON file
             containing such data on disk
-        classes: the list of class label strings
+        classes: the list of class label strings or a dict mapping class IDs to
+            class labels
         label_type ("detections"): the type of labels to load. Supported values
             are ``("detections", "segmentations", "keypoints")``
         coco_id_field (None): this parameter determines how to map the
@@ -194,7 +195,10 @@ def add_coco_labels(
     view.compute_metadata()
     widths, heights = view.values(["metadata.width", "metadata.height"])
 
-    classes_map = {i: label for i, label in enumerate(classes)}
+    if isinstance(classes, dict):
+        classes_map = classes
+    else:
+        classes_map = {i: label for i, label in enumerate(classes)}
 
     labels = []
     for _coco_objects, width, height in zip(coco_objects, widths, heights):
@@ -1433,7 +1437,7 @@ def parse_coco_categories(categories):
     Returns:
         a tuple of
 
-        -   classes_map: a dict mapping class ids to labels
+        -   classes_map: a dict mapping class IDs to labels
         -   supercategory_map: a dict mapping class labels to category dicts
     """
     classes_map = {
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -194,6 +194,8 @@ def add_coco_labels(
     view.compute_metadata()
     widths, heights = view.values(["metadata.width", "metadata.height"])
 
+    classes_map = {i: label for i, label in enumerate(classes)}
+
     labels = []
     for _coco_objects, width, height in zip(coco_objects, widths, heights):
         frame_size = (width, height)
@@ -202,7 +204,7 @@ def add_coco_labels(
             _labels = _coco_objects_to_detections(
                 _coco_objects,
                 frame_size,
-                classes,
+                classes_map,
                 None,
                 False,
                 include_annotation_id,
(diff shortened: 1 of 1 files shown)


fiftyone/utils/coco.py
```diff
@@ -597,7 +603,7 @@ def setup(self):
             license_map = None
 
         self._info = info
-        self._classes = classes
+        self._classes = classes_map
```
brimoor, line 606: could rename this to `self._classes_map` for consistency with other naming conventions (renaming private instance variables is not a concern from a backwards-compatibility standpoint)

brimoor (commented):
@patrontheo thanks for taking this on! I'll defer to @swheaton for functional review.

With respect to your comment about `add_coco_labels()`, perhaps you could update so that its `classes` argument can be either a list (current) or a dict (new format)?

fiftyone/utils/coco.py
patrontheo: yes, absolutely!

patrontheo: done

patrontheo:
> With respect to your comment about `add_coco_labels()`, perhaps you could update so that its `classes` argument can be either a list (current) or a dict (new format)?

Yes, good idea, I did that.

For the Exporter, I haven't tried yet to use existing IDs as you suggested (in #4293), but I'll see if I get a bit of time to look at it.

patrontheo:
@swheaton  any input ?

brimoor (approved):
LGTM, I'll handle my suggestion to retain class mappings when exporting in a follow-up commit 👍

swheaton:
@patrontheo sorry been very busy, prepping to go on leave.

Thank you for the contribution!!

Issue #4322, opened 2024-04-30
[BUG] Mongoengine: using pop() to update dict fields is dangerous and should be avoided


brimoor:
As of this writing, the `update_run_key()` and `delete_run()` methods currently use `pop()` to modify the `runs` dict:
- https://github.com/voxel51/fiftyone/blob/bdfc97f446370631bc141b3bfbc8d171bcc1243e/fiftyone/core/runs.py#L504
- https://github.com/voxel51/fiftyone/blob/bdfc97f446370631bc141b3bfbc8d171bcc1243e/fiftyone/core/runs.py#L825

Unfortunately, it appears that `mongoengine` treats `pop()`-based dict edits with a very heavy hammer: the corresponding update operations that are sent to `pymongo` set the **entire** `runs` key of the dataset's backing document:

```py
import fiftyone as fo

dataset = fo.Dataset("zzz")

config = dataset.init_run(foo="bar")
dataset.register_run("test1", config)

dataset.rename_run("test1", "still_test1")
# {'$set': {'runs': {'still_test1': ObjectId('663043c40e1de06966c90021')}}}

dataset.delete_run("still_test1")
# {'$unset': {'runs': 1}}
```

Why is this bad? If another run was added in a separate process, it could be accidentally obliterated by performing a valid update/deletion of a different run, as demonstrated below: 

### Process 1

```py
import fiftyone as fo

dataset = fo.Dataset("zzz")

config = dataset.init_run(foo="bar")
dataset.register_run("test1", config)

print(dataset.list_runs())
# ['test1']
```

### Process 2

```py
import fiftyone as fo

dataset = fo.load_dataset("zzz")

config = dataset.init_run(spam="eggs")
dataset.register_run("test2", config)

print(dataset.list_runs())
# ['test1', 'test2']
```

### Back in Process 1

```py
dataset.delete_run("test1")
```

### Back in Process 2

```py
dataset.reload()

# uh oh, 'test2' is also gone, but we never deleted it!
print(dataset.list_runs())
# []
```

brimoor:
As per https://github.com/voxel51/fiftyone/pull/4323, the solution is (annoyingly) to replace `d.pop(key)` with `del d[key]`. Both are equivalent Python operations, but apparently `mongoengine` converts `del` to **nested** pymongo updates, which are robust to situations where a concurrent process has added/deleted other runs.

PR #4323, opened 2024-04-30
Fix concurrency bug when updating/deleting runs

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/4322

Fixes the bug pointed out in https://github.com/voxel51/fiftyone/pull/4320 with root cause analysis detailed in https://github.com/voxel51/fiftyone/issues/4322.

The solution is (annoyingly) to replace `d.pop(key)` with `del d[key]`. Both are equivalent Python operations, but apparently `mongoengine` converts `del` to **nested** pymongo updates, which are robust to situations where a concurrent process has added/deleted other runs.

A unit test is added that verifies that concurrent run updates will work going forward.

### Process 1

```py
import fiftyone as fo

dataset = fo.Dataset("zzz")

config = dataset.init_run(foo="bar")
dataset.register_run("test1", config)

print(dataset.list_runs())
# ['test1']
```

### Process 2

```py
import fiftyone as fo

config = dataset.init_run(spam="eggs")
dataset.register_run("test2", config)

print(dataset.list_runs())
# ['test1', 'test2']
```

### Back in Process 1

```py
dataset.rename_run("test1", "still_test1")
"""
BEFORE
{'$set': {'runs': {'still_test1': ObjectId('663043c40e1de06966c90021')}}}

AFTER
{
    '$set': {'runs.still_test1': ObjectId('6630446149a8040d98c51068')},
    '$unset': {'runs.test1': 1},
}
"""

dataset.delete_run("still_test1")
"""
BEFORE
{'$unset': {'runs': 1}}

AFTER
{'$unset': {'runs.still_test1': 1}}
"""
```

### Back in Process 2

```py
dataset.reload()
print(dataset.list_runs())
"""
BEFORE
[] <-- data loss!!!

AFTER
['test2'] <-- no data loss
"""
```

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the handling of run documents in the application to avoid using the `pop()` method, improving reliability and error management.

- **Tests**
	- Added a new test to ensure concurrent updates on runs across multiple datasets function correctly, enhancing the robustness of run operations.

### DIFF
--- a/fiftyone/core/runs.py
+++ b/fiftyone/core/runs.py
@@ -501,11 +501,13 @@ class BaseRun(Configurable):
 
         # Update run doc
         run_docs = getattr(dataset._doc, cls._runs_field())
-        run_doc = run_docs.pop(key)
+        # DON'T use pop()! https://github.com/voxel51/fiftyone/issues/4322
+        run_doc = run_docs[key]
+        del run_docs[key]
         run_doc.key = new_key
         run_docs[new_key] = run_doc
         run_doc.save()
-        dataset._doc.save()
+        dataset.save()
 
         # Update results cache
         results_cache = getattr(dataset, cls._results_cache_field())
@@ -822,7 +824,11 @@ class BaseRun(Configurable):
 
         # Delete run from dataset
         run_docs = getattr(dataset._doc, cls._runs_field())
-        run_docs.pop(key, None)
+        try:
+            # DON'T use pop()! https://github.com/voxel51/fiftyone/issues/4322
+            del run_docs[key]
+        except KeyError:
+            pass
         results_cache = getattr(dataset, cls._results_cache_field())
         run_results = results_cache.pop(key, None)
         if run_results is not None:
--- a/tests/unittests/run_tests.py
+++ b/tests/unittests/run_tests.py
@@ -129,6 +129,47 @@ class RunTests(unittest.TestCase):
         runs = dataset.list_runs(method="test")
         self.assertListEqual(runs, ["custom2"])
 
+    @drop_datasets
+    def test_concurrent_run_updates(self):
+        dataset = fo.Dataset()
+        kwargs = {"foo": "bar", "spam": "eggs"}
+
+        config1 = dataset.init_run(**kwargs)
+        dataset.register_run("custom1", config1)
+
+        results1 = dataset.init_run_results("custom1", **kwargs)
+        dataset.save_run_results("custom1", results1)
+
+        # Don't reuse singleton; we want to test concurrent edits here
+        dataset._instances.pop(dataset.name)
+        also_dataset = fo.load_dataset(dataset.name)
+        self.assertIsNot(dataset, also_dataset)
+
+        config2 = also_dataset.init_run(**kwargs)
+        also_dataset.register_run("custom2", config2)
+
+        results2 = also_dataset.init_run_results("custom2", **kwargs)
+        also_dataset.save_run_results("custom2", results2)
+
+        self.assertListEqual(dataset.list_runs(), ["custom1"])
+
+        dataset.rename_run("custom1", "still_custom1")
+        also_dataset.reload()
+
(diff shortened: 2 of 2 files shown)


swheaton:
For some reason with `pop()` it tries to be smart in the case that the dict goes from 1 object to 0. If there are more than 1 runs then `pop(key)` has the same behavior as `del runs[key]`. So I think this may have contributed to the elusiveness of seeing this issue.

I'm concerned that this is brittle and an update to mongoengine could randomly break this too if they decide to add this "optimization". I think we should also reload the dataset doc just in case, like we did in https://github.com/voxel51/fiftyone/pull/3726

Also I believe we have a similar problem with saved views and workspaces. In this case it always writes the full list.

Shell 1
```python
ds=fo.Dataset("zzz")
ds.save_view("test1", ds.limit(1))
```
Shell 2
```python
ds=fo.load_dataset("zzz")
ds.save_view("test2", ds.limit(2))
ds.delete_saved_view("test2")
```
Back in shell 1
```python
ds.reload()
ds.list_saved_views()  # [] --- oopsie where'd my saved view go??
```

fiftyone/core/runs.py
```diff
@@ -501,11 +501,13 @@ def update_run_key(cls, samples, key, new_key):
 
         # Update run doc
```
swheaton, line 502:
```suggestion
        # Update run doc
        dataset._doc.reload()
```

swheaton:
same with tags and other lists. We added a doc reload before schema additions in https://github.com/voxel51/fiftyone/pull/3308.

Notably,
> As explained in the @todo, this is a bit of a hack to specifically avoid issues when concurrently modifying a dataset's schema. However, the underlying problem still exists whenever list fields other than DatasetDocument.sample_fields and DatasetDocument.frame_fields are concurrently edited without first reloading.

swheaton:
The nuclear option is to get rid of mongoengine LOL.
The ballistic arms option is to create a shim class that actually does list differencing in the `delta` aspect of mongoengine. That way we could use something like:
```
{"$pull": {"list_field": {"$eq": "value"}}}
```

swheaton:
So the reason `pop()` doesn't work but `del` does, is that in the former, `mongoengine` marks the whole dict as "dirty" whereas in the latter `my_dict.the_key` is [marked](https://github.com/MongoEngine/mongoengine/blob/master/mongoengine/base/datastructures.py#L90) as dirty.

Same with lists - some operations such as `__setitem__` mark `my_list.1` as dirty but do not optimize for `append()` for example and just set the whole list as dirty.

Additionally (this is something I've discovered before but didn't know exactly why), the `clear()` method on `list` is not wrapped (aka changes not tracked) so if you call `ds.tags.clear(); ds.save()` for example, it is not saved!

The wrapping mechanism is somewhat clever but is incomplete/suboptimal. Outside of forking or rewriting it, I'm not sure yet the best way forward

swheaton:
@brimoor Yeah that makes sense. In talking with Eric and Ben, I called it the reload salt shaker approach haha. Definitely not ideal.

Another fallout of the mongoengine impl (that isn't fixed by reload shaker) is that if you do
```python
sample.predictions.detections.append(fo.Detection("one-more-detection"))
sample.save()
```
it will send the entire contents of `sample.predictions.detections`. Not *so* bad with only a few per sample but in many use cases there could be dozens, hundreds, thousands, etc. !

So that being said, I think it is ultimately necessary to take more of this burden in-house to ensure correctness and performance. I am looking into a proposal for doing this (without replacing mongoengine .. yet) but it does not seem as simple as I'd hoped.

In the meantime, we may want to do this for dataset doc mutations only. Nothing we can really do about user-made mutations (`dataset.tags.append("blah")`).
> reloading the minimum possible set of fields when performing mutations

swheaton (approved):
lgtm as it stands.
targeted reload() can be considered in follow on

benjaminpkane (approved):

PR #4334, opened 2024-05-01
Fix concurrency bug when updating/deleting runs

brimoor:
Alternative to https://github.com/voxel51/fiftyone/pull/4323 that uses `reload()` rather than replacing `pop` with `del`.

IMO this is slightly less ideal than https://github.com/voxel51/fiftyone/pull/4323 because:
- this version is not fully protected against concurrent edits (although now the window of concurrency issues is the time it takes to execute a few lines of Python code and then make a pymongo API call)
- `reload()` is an extra API call
- a breaking change in how `mongoengine` implements pop/del deltas is unlikely, given that the package is not being actively developed (only like 1-2 releases in last 12 months)

## Summary by CodeRabbit

- **New Features**
	- Enhanced dataset run management to support concurrent edits and reliable updates.
- **Bug Fixes**
	- Improved the reliability of run deletions from datasets.
- **Tests**
	- Added a new test for concurrent run updates to ensure robustness in multi-user environments.

### DIFF
--- a/fiftyone/core/runs.py
+++ b/fiftyone/core/runs.py
@@ -501,11 +501,13 @@ class BaseRun(Configurable):
 
         # Update run doc
         run_docs = getattr(dataset._doc, cls._runs_field())
-        run_doc = run_docs.pop(key)
+        # DON'T use pop()! https://github.com/voxel51/fiftyone/issues/4322
+        run_doc = run_docs[key]
+        del run_docs[key]
         run_doc.key = new_key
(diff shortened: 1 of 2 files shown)


swheaton:
Agree I prefer the other one. After reading code to understand how the delta works, I can say it is near zero chance that del has the same problem as pop in the future - it would basically have to be actively sabotaged lol

brimoor:
Closing in favor of https://github.com/voxel51/fiftyone/pull/4323.

PR #4340, opened 2024-05-02
refactor(threed): push args in docstring to child class

swheaton:

In service of `Explicit is better than implicit`, I like putting all inherited args in the child class docstring and init up front. This way you see exactly how to create an instance and don't have to hop back and forth in the family tree.

The flip side is that there is a lot of duplication and if an arg is added to `Object3D` then it has to be pushed to all children. I would argue this is an acceptable few-time cost in exchange for more readable code and user documentation.

`**kwargs` is interesting if the args can truly be dynamic, in my opinion.

Open for discussion.

unit tests still fine

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Improved lighting customization with added parameters for visibility, position, scale, and orientation in various light classes.
  - Introduced opacity control for different material types, enabling users to adjust transparency.
  - Updated 3D objects and shapes to include customizable visibility, position, scale, and orientation settings.

- **Refactor**
  - Enhanced constructor parameter lists for lights, materials, meshes, point clouds, and 3D shapes to provide advanced customization options.

### DIFF
--- a/fiftyone/core/threed/pointcloud.py
+++ b/fiftyone/core/threed/pointcloud.py
@@ -29,7 +29,6 @@ class PointCloud(Object3D):
             :class:`fiftyone.core.threed.Scene` can have at most one asset
             flagged for orthographic projection. Defaults to ``False``. If
             multiple assets are flagged, the first one will be chosen
-                name: the name of the object
         visible (True): default visibility of the point cloud in the scene
         position (None): the position of the object in point cloud space
         quaternion (None): the quaternion of the point cloud in object space
--- a/fiftyone/core/threed/pointcloud.py
+++ b/fiftyone/core/threed/pointcloud.py
@@ -10,6 +10,7 @@ from typing import Optional
 
 from .material_3d import PointCloudMaterial
 from .object_3d import Object3D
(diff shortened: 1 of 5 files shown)


fiftyone/core/threed/pointcloud.py
```diff
@@ -28,8 +29,11 @@ class PointCloud(Object3D):
             :class:`fiftyone.core.threed.Scene` can have at most one asset
             flagged for orthographic projection. Defaults to ``False``. If
             multiple assets are flagged, the first one will be chosen
-        **kwargs: keyword arguments for the
-            :class:`fiftyone.core.threed.Object3D` parent class
+                name: the name of the object
```
brimoor, line 32: `name` is defined twice in this docstring. I think this one needs to be deleted (also indentation is borked)

brimoor (approved):
LGTM after fixing the one `name` issue I flagged.

And +1 for explicit parameter documentation! Tedious but very useful for end users 🤗 

@sashankaryal should probably approve too before merging as code owner

sashankaryal (approved):
lgtm!! ⚡

PR #4341, opened 2024-05-02
Relax pydantic requirement to allow 1 or 2

swheaton:
Relax pydantic requirement in `setup.py` to allow v1 or v2.

Our usage of pydantic is pretty minimal but could be used more in the future so I think requiring pydantic is fine.
That being said, we **must not** break any user setups that must have pydantic v1.

So this change just makes our use of pydantic compatible with both v1 and v2 by using the deprecated `validator` instead of the newer `field_validator`. We can stick with this for a bit then re-evaluate if we do need the features of `field_validator`.

unit tests, creating objects and seeing validation still work

```python
import fiftyone.core.threed as fo3d
obj = fo3d.Object3D(name="blah", position=1)
obj.position # (1.0, 1.0, 1.0)
obj.position = "blah"  # validation error
obj.position = 2.0  # (2.0, 2.0, 2.0)
```

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Refactor**
	- Enhanced stability by updating import statements in the 3D module.
- **Chores**
	- Ensured broader compatibility by modifying package dependencies.

### DIFF
--- a/fiftyone/core/threed/validators.py
+++ b/fiftyone/core/threed/validators.py
@@ -28,6 +28,6 @@ def normalize_to_vec3(v: Optional[Vec3UnionType]) -> Union[Vector3, None]:
 
 
 def vec3_normalizing_validator(field: str) -> classmethod:
-    decorator = field_validator(field)
+    decorator = field_validator(field, allow_reuse=True)
     validator_class_method = decorator(normalize_to_vec3)
     return validator_class_method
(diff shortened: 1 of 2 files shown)


swheaton:
> Hmm, `Object3D` doesn't use the validator. Only `PerspectiveCamera`, which doesn't seem to behave as expected
> 
> ```python
> import fiftyone.core.threed as fo3d
> obj = fo3d.PerspectiveCamera(position=1) # error
> ```

It does but just uses it as a function. It is definitely odd that `PerspectiveCamera` is a dataclass and `Object3D` is not, and that the behavior of `position` is different. Reconciling offline

swheaton:
closing, we will do something else

PR #4349, opened 2024-05-03
Debug3546

### LINKED ISSUE #3546: [BUG] Clearing the tags field and then tagging samples raises error
To reproduce:
```python
import fiftyone.zoo as foz

dataset.clear_sample_field("tags")
dataset.tag_samples("test")

# ValueError: Cannot apply $addToSet to non-array field. Field named 'tags' has non-array type null
```

This is happening because `clear_sample_field` sets the values of the field to `None` for each sample, but `tag_samples()` then tries to add a value to what it expects is a list.

We should probably add a check to `tag_samples()` to verify that the field isn't `None` before adding samples. Alternatively, we could have `clear_sample_field("tags")` check if you are clearing the tags field and set it to `[]` instead of `None`, but that field could still be set to `None` in other ways, so it seems the first approach would be preferred.

tina1092:

this commit is the solution for bug3546

closes #3546 

create a test case and run successfully

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced tag management by preventing the addition of tags to samples without a defined tag type and logging an error instead.
	
- **New Features**
	- Updated behavior for clearing fields in datasets to differentiate between `list[]` type samples and other types, ensuring appropriate clearing actions are taken.

- **Tests**
	- Added a new test to verify the functionality of tagging and untagging samples in dataset views, ensuring tags are cleared and counts are updated correctly.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1600,6 +1600,8 @@ class SampleCollection(object):
 
     def tag_samples(self, tags):
         """Adds the tag(s) to all samples in this collection, if necessary.
+            Stop adding tags and log the error if there is a sample's tags
+            type is None.
 
         Args:
             tags: a tag or iterable of tags
@@ -1630,6 +1632,13 @@ class SampleCollection(object):
         view._edit_sample_tags(update)
 
     def _edit_sample_tags(self, update):
+        """Applies the update operation to the 'tags' field of given
+            samples in this collection.
+
+        Args:
+            update (dict): A dictionary specifying the
+            MongoDB update operations.
+        """
         ids = []
         ops = []
         batch_size = fou.recommend_batch_size_for_value(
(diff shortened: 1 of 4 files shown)


brimoor:
@tina1092 thanks for taking a shot at resolving this!

One of the design principles of FiftyOne is that all non-required fields should be nullable, so even if we were to take the proposed approach here of updating `clear_sample_field("tags")` to store `[]` rather than `None`, it is still valid for a user to do something like this:

```py
dataset.set_field("tags", None).save()
```

As a result, a more robust solution is to patch the `tag_samples()`, `untag_samples()` and related methods so that they can gracefully handle null-valued fields. I've taken that approach in https://github.com/voxel51/fiftyone/pull/4351!

brimoor:
Resolved in https://github.com/voxel51/fiftyone/pull/4351!

PR #4350, opened 2024-05-05
Fix concurrency bugs when updating views/workspaces/group-slices

swheaton:

Improve concurrent edits of saved views and workspaces by getting out our reload salt shaker and sprinkling it around.

Improve concurrent edits of group slices by shaking it some more AND converting `dict.pop(key)` to `del dict[key]` (see https://github.com/voxel51/fiftyone/pull/4323/). As opposed to 4323, we do need to do the targeted reload also here because of the `default_group_slice` field that should be in sync. The last part of the added groups test exercises this.

a la https://github.com/voxel51/fiftyone/pull/3308
As explained in the todo, this is a bit of a hack to specifically avoid issues when concurrently modifying a dataset's schema.

added targeted reload to:
- save new
- delete (circa `get()` with `pop` parameter)
- delete all

dont need for:
- update info - this just edits the view/workspace doc and so there's no changes to the list in the dataset doc.

I don't think there is any other critical place where we need to do this immediately. Tags maybe but that's hard because it's a user setting it. The rest can wait for better list delta support.

Unit tests were added that verifies that concurrent saved view / workspace updates will work going forward, for add/delete/delete-all.
Also added for concurrent updates to group slices and default_group_slice field.

### Process 1

```py
import fiftyone as fo

dataset = fo.Dataset("zzz")

print(dataset.list_saved_views()())
# []
```

### Process 2

```py
import fiftyone as fo

dataset.save_view("view2", dataset.limit(1))
print(dataset.list_runs())

# ['test2']
```

### Back in Process 1

```py
dataset.save_view("view1", dataset.limit(1))
"""
BEFORE
{'$set': {'saved_views': [ObjectId('663781cfd7c7ab4e1c83c6dd')]}}

AFTER
{'$set': {'saved_views': [ObjectId('6637818ee0bc8c516066763a'), ObjectId('663781cfd7c7ab4e1c83c6dd')]}}
"""
```

### Back in Process 2

```py
dataset.reload()
print(dataset.list_saved_views())
"""
BEFORE
['view1'] <-- data loss!!!

AFTER
['view2', 'view1] <-- no data loss
"""
```

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced the dataset management system to support better concurrency safety, allowing for more reliable updates and interactions with saved views and workspaces.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -3590,7 +3590,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
 
         # Targeted reload of saved views for better concurrency safety.
         # @todo improve list field updates in general so this isn't necessary
-        self._doc.reload(["saved_views"])
+        self._doc.reload("saved_views")
 
         self._doc.saved_views.append(view_doc)
         self.save()
@@ -3710,7 +3710,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
 
         # Targeted reload of saved views for better concurrency safety.
         # @todo improve list field updates in general so this isn't necessary
-        self._doc.reload(["saved_views"])
+        self._doc.reload("saved_views")
 
         for view_doc in self._doc.saved_views:
             if isinstance(view_doc, DBRef):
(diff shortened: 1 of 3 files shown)


brimoor (approved):
LGTM, thanks for tracking these down 💪

PR #4366, opened 2024-05-07
add translate on shift + left mouse

sashankaryal:

New camera controls doesn't have translate on shift + left click by default. Instead, it uses right mouse, or two-finger move / three-finger move on trackpads. This isn't consistent with our prior CX, which used shift + click for trucking. This PR fixes that.

Manually.

### DIFF
--- a/app/packages/looker-3d/src/action-bar/ViewHelp.tsx
+++ b/app/packages/looker-3d/src/action-bar/ViewHelp.tsx
@@ -7,7 +7,7 @@ export const LOOKER3D_HELP_ITEMS = [
   { shortcut: "Left mouse drag", title: "Rotate", detail: "Rotate the camera" },
   { shortcut: "Wheel", title: "Zoom", detail: "Zoom in and out" },
   {
-    shortcut: "Middle mouse drag",
+    shortcut: "Shift + drag",
     title: "Translate",
     detail: "Translate the camera",
   },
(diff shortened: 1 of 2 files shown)


app/packages/looker-3d/src/action-bar/ViewHelp.tsx
```diff
@@ -4,15 +4,19 @@ import { ActionItem } from "../containers";
...
 export const LOOKER3D_HELP_ITEMS = [
-  { shortcut: "Left mouse drag", title: "Rotate", detail: "Rotate the camera" },
-  { shortcut: "Wheel", title: "Zoom", detail: "Zoom in and out" },
+  { shortcut: "Left drag", title: "Rotate", detail: "Rotate the camera" },
+  { shortcut: "T", title: "Top-down", detail: "Reset camera to top-down view" },
   {
-    shortcut: "Middle mouse drag",
-    title: "Translate",
-    detail: "Translate the camera",
+    shortcut: "Ctrl + left drag / Wheel",
```
brimoor, line 10: Perhaps I don't fully appreciate the left/right clicks as a single button mouse user, but would this more concise set of instructions suffice?

```
Wheel           Zoom        Zoom in and out
Shift + drag    Translate   Translate the camera
```

brimoor (commented):
Tested and confirmed that shift + drag translation is back! 🎉

Not sure if this was intentional or unavoidable, but clicking on the XYZ nodes on the axes widget in the upper left no longer changes the view. I found that feature useful, but if I can only have that or `shift + drag`, I'd rather have `shift + drag` ✅ 

Also, completely unrelated but I've been meaning to request this: could you convert the `Inspect` button into an icon like the other actions? The button doesn't feel "polished"

Two ideas for this:
1. Convert to a `{}` icon in the header of the 3D pop-out menu, eg to the left of the search icon
2. Convert to an icon in the menu bar to the right with the other actions. Maybe a... `AccountTree` or something that captures the graph structure?
Screenshot 2024-05-07 at 5 33 04 PM

ritch (approved):
Tested locally and all looks in order. 👍

sashankaryal:
@brimoor good idea. How do you like this? Also added a hotkey `I` for viewing FO3D.

PR #4367, opened 2024-05-07
include location, dimensions, and rotation in fo3d

sashankaryal:
PCD detection fields are treated as default by `SelectFields`, but that should also apply to FO3D. One of the symptoms of it not being included is that labels are not painted in the grid for fo3d files. This PR fixes that.

Screenshot 2024-05-07 at 3 55 06 PM

Screenshot 2024-05-07 at 3 55 34 PM

## Summary by CodeRabbit

- **New Features**
	- Enhanced media type handling to include 3D models in addition to point clouds for field name determination in collections management.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1041,7 +1041,7 @@ class SampleCollection(object):
                 use_db_fields=use_db_fields,
             )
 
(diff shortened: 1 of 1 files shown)


brimoor (approved):
Tested and LGTM (added the same fix to frame fields as well) 🚀

PR #4368, opened 2024-05-07
3D export fixes and tests

swheaton:
Checkpointing my work in case I don't get to finish it.

1. added tests for at least 4 different configurations of fo3d data layout + relative vs. absolute + flatten vs. rel_dir.
2. Fixed a few bugs caught by tests
   - .mtl file in ObjMesh wasn't getting rewritten (it was only checking first asset path for each node)
   - use input to output asset path map to determine when we need to rewrite asset path in the fo3d file. fixes the case when flattening (no `rel_dir`) causes a file to be renamed from, say, `pcd.pcd` to `pcd-2.pcd`

The tests are absolutely horrendous on the eyes, I'm so sorry!! They could be refactored to something nicer but didn't have time right now.

What I would like to do if I have time:
- allow `asset_dir`
- make changes work with Teams which includes `fos.resolve()` and a cloud friendly `os.path.relpath()` if it doesn't already exist. Actually the latter probably just works but we would want to double check.
- make a batch media exporter that does this in bulk instead of one at a time. Then re-use that for `fos.upload_media()` and `download_media()` in Teams so we don't have the same fragile code implemented twice. With the side benefit that export should be faster.

## Summary by CodeRabbit

- **New Features**
	- Introduced a method to update asset paths in 3D scenes, enhancing the management of asset locations based on user-defined mappings.
- **Tests**
	- Added new tests for 3D media handling, ensuring robustness and reliability in asset path updates and scene integrity.

### DIFF
--- a/fiftyone/core/threed/scene_3d.py
+++ b/fiftyone/core/threed/scene_3d.py
@@ -248,12 +248,56 @@ class Scene(Object3D):
 
         Args:
             include_self: whether to include the current node in the traversal
+
+        Yields:
+            :class:`Object3D`
+
         """
         if include_self:
             yield self
 
         for child in self.children:
             yield from child.traverse(include_self=True)
+
+    def update_asset_paths(self, asset_rewrite_paths: dict):
+        """Update asset paths in this scene according to an input dict mapping.
+
+        Asset path is unchanged if it does not exist in ``asset_rewrite_paths``
+
+        Args:
+            asset_rewrite_paths: ``dict`` mapping asset path to new asset path
+
+        Returns:
+            ``True`` if the scene was modified.
+        """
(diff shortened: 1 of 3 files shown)


swheaton:
> make a batch media exporter that does this in bulk instead of one at a time. Then re-use that for fos.upload_media() and download_media() in Teams so we don't have the same fragile code implemented twice. With the side benefit that export should be faster.

Ok I am seeing that this exists, sort of, in Teams when a cloud path is involved. I am unable to take this refactor on right now so perhaps I'll have to leave it alone for now.

brimoor (approved):
LGTM

swheaton:
@coderabbitai review

PR #4371, opened 2024-05-08
Fix DateField for GMT+ users

### LINKED ISSUE #4365: Wrong DateField value loaded from MongoDB
### Describe the problem

I load a persistent dataset from MongoDB with the instruction:
`dataset = fo.load_dataset(dataset_name)`

This is the print from the first sample:
```
print(f"{sample['filepath']}\n{sample['date']} - {sample['datetime']}") 
/root/fiftyone/global/2023/2023-09-08/d30adad1335f04da4bf928c0f023768374fafeca15d94dc94f0265c8a73de8630e3c37fcf972a43059d4b5897061fdff82605d2f03675d8e90862b6cee94415e.jpg
2023-09-07 - 2023-09-08 11:36:17
```
where the date value 2023-09-07 is wrong. It must be 2023-09-08.

Checking the same sample directly into MongoDB I get these values that seems correct:
```
{
    "_id": {
        "$oid": "662266926a533ff7fe32ddef"
    },
    "filepath": "/root/fiftyone/global/2023/2023-09-08/d30adad1335f04da4bf928c0f023768374fafeca15d94dc94f0265c8a73de8630e3c37fcf972a43059d4b5897061fdff82605d2f03675d8e90862b6cee94415e.jpg",
    "tags": [],
    "_media_type": "image",
    "_rand": 0.9997939277596051,
    "ground_truth": {
        "_id": {
            "$oid": "662266916a533ff7fe32ddee"
        },
        "_cls": "Classification",
        "tags": [],
        "label": "Lifted Electrode"
    },
    "datetime": {
        "$date": {
            "$numberLong": "1694172977000"
        }
    },
    "year": 2023,
    "date": {
        "$date": {
            "$numberLong": "1694131200000"
        }
    },
    "_dataset_id": {
        "$oid": "662266916a533ff7fe32dded"
    }
}
```

### Code to reproduce issue

I don't know if my localization in Rome could be a possible problem.

### System information

-   Linux Ubuntu 22.04
-   Python version 3.9.16
-   FiftyOne version 0.23.8

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community
-   [x] No. I cannot contribute a bug fix at this time

swheaton:

Closes #4365

When the DB is not made timezone aware, it is not UTC by default but rather timezone agnostic. So when you call `astimezone(pytz.utc)` it assumes local timezone in order to do the conversion.
When timezone is GMT-2, the stored date (2023, 9, 8, 0, 0) becomes (2023, 9, 8, 2, 0) which results in the same date so you don't see a difference.
When timezone is GMT+2, like the reporting user, (2023, 9, 8, 0, 0) becomes (2023, 9, 7, 22, 0) which is a different date (2023/09/07) than expected!

1. Use `db.with_options()` to apply timezone awareness of "UTC" if that is specified in `fo.config`. A datetime with timezone UTC is not the same as a datetime with no timezone.
   - Would love to change the default behavior as well but it would be a potential behavior change whereas I believe the above is strictly a bug fix.
2. Remove conversion to UTC in `Date` field which will do the wrong thing in the case that the datetime is not timezone aware. I don't know why we wanted to convert to UTC in the first place. See inline PR comments.

I am not sure how to test because you have to fake localized timezone. There's probably a way to do that.

        notes for FiftyOne users.

```
Fixed :class:`fiftyone.core.fields.DateField` for users in GMT+ timezones when `fo.config.timezone` is unset.
```

-   [ ] Other

## Summary by CodeRabbit

- **Refactor**
    - Simplified datetime to date conversion for improved reliability and performance.
    - Enhanced timezone handling logic for consistent behavior across user environments.

### DIFF
--- a/fiftyone/core/fields.py
+++ b/fiftyone/core/fields.py
@@ -514,10 +514,7 @@ class DateField(mongoengine.fields.DateField, Field):
         if value is None:
             return None
 
-        # Explicitly converting to UTC is important here because PyMongo loads
-        # everything as `datetime`, which will respect `fo.config.timezone`,
-        # but we always need UTC here for the conversion back to `date`
-        return value.astimezone(pytz.utc).date()
+        return value.date()
 
     def validate(self, value):
         if not isinstance(value, date):
--- a/fiftyone/core/fields.py
+++ b/fiftyone/core/fields.py
@@ -514,6 +514,15 @@ class DateField(mongoengine.fields.DateField, Field):
         if value is None:
             return None
 
+        # Explicitly converting to UTC is important here because PyMongo loads
+        # everything as `datetime`, which will respect `fo.config.timezone`,
+        # but we always need UTC here for the conversion back to `date` because
+        # we write to the DB in UTC time.
+        # If value is timezone unaware, then do not convert because it will be
+        # assumed to be in local timezone and date will be wrong for GMT+
+        # localities.
+        if value.tzinfo is not None:
+            value = value.astimezone(pytz.utc)
         return value.date()
 
     def validate(self, value):
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -422,7 +422,7 @@ def get_async_db_conn(use_global=False):
 def _apply_options(db):
     timezone = fo.config.timezone
 
-    if not timezone or timezone.lower() == "utc":
+    if not timezone:
         return db
 
     if timezone.lower() == "local":
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -423,7 +423,7 @@ def _apply_options(db):
     timezone = fo.config.timezone
 
     if not timezone:
-        timezone = "utc"
+        return db
 
     if timezone.lower() == "local":
         tzinfo = datetime.now().astimezone().tzinfo
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -423,7 +423,7 @@ def _apply_options(db):
     timezone = fo.config.timezone
 
     if not timezone:
-        return db
+        timezone = "utc"
 
     if timezone.lower() == "local":
         tzinfo = datetime.now().astimezone().tzinfo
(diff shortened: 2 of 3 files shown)


fiftyone/core/fields.py
```diff
@@ -514,10 +514,7 @@ def to_python(self, value):
         if value is None:
             return None
 
-        # Explicitly converting to UTC is important here because PyMongo loads
-        # everything as `datetime`, which will respect `fo.config.timezone`,
-        # but we always need UTC here for the conversion back to `date`
-        return value.astimezone(pytz.utc).date()
+        return value.date()
```
swheaton, line 517: I did not understand this comment, why would we want it to be in UTC to get the date?
Removing this conversion fixes the issue described in PR.

fiftyone/core/odm/database.py
```diff
@@ -422,7 +422,7 @@ def get_async_db_conn(use_global=False):
 def _apply_options(db):
     timezone = fo.config.timezone
 
-    if not timezone or timezone.lower() == "utc":
+    if not timezone:
```
swheaton, line 425: If you specify `fo.config.timezone="utc"`, we should apply timezone awareness with `utc` timezone explicitly.
Otherwise the `datetime` coming from DB will have no timezone attached which is not the same as having UTC timezone. For example (see other file), `astimezone()` assumes local time in order to convert to UTC so changes the datetime value.

brimoor (dismissed):
LGTM. I have no explanation/defense for the comment in `DateField.to_python()`. Apparently I was pretty sure of myself at the time but it doesn't make sense to me now, either 🤷 

The key thing is that dates always need to be **stored** in the DB as UTC

swheaton:
> LGTM. I have no explanation/defense for the comment in `DateField.to_python()`. Apparently I was pretty sure of myself at the time but it doesn't make sense to me now, either 🤷
> 
> The key thing is that dates always need to be **stored** in the DB as UTC

Hmm I see. Well I think what is intended then is `replace()`? From datetime [docs](https://docs.python.org/3/library/datetime.html). You want to just assume the date is in UTC in DB, ignoring whatever timezone options were applied to the DB. I think this jives with the original comment and the (now-failing) unit test.
> If you merely want to attach a time zone object tz to a datetime dt without adjustment of date and time data, use dt.replace(tzinfo=tz).

swheaton:
@brimoor i think i fixed it for real this time and added an exposing test case by mocking the "local" timezone.

brimoor (approved):
LGTM

PR #4379, opened 2024-05-09
Add new permission flags for tagging, changing sidebar groups, creating new field for similarity search

lanzhenw:

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Enhanced permissions system with new capabilities:
    - Modify sidebar groups
    - Create new fields
    - Tag samples or labels
    - Edit custom colors
    - Edit saved views
  
- **Bug Fixes**
  - Improved logic for enabling/disabling actions based on permissions.

- **Refactor**
  - Replaced `readOnly` state with specific permission checks across multiple components for better modularity and clarity.

- **Usability Improvements**
  - Dynamic titles and disabled states for better user feedback based on permissions.

### DIFF
--- a/app/packages/core/src/components/Actions/ActionsRow.tsx
+++ b/app/packages/core/src/components/Actions/ActionsRow.tsx
@@ -163,8 +163,7 @@ const Tag = ({
   const labels = useRecoilValue(fos.selectedLabelIds);
   const samples = useRecoilValue(fos.selectedSamples);
   const canTag = useRecoilValue(fos.canTagSamplesOrLabels);
-  const readOnly = useRecoilValue(fos.readOnly);
-  const isReadOnly = readOnly || !canTag;
+  const isReadOnly = canTag.enabled !== true;
 
   const selected = labels.size > 0 || samples.size > 0;
   const tagging = useRecoilValue(fos.anyTagging);
@@ -182,7 +181,9 @@ const Tag = ({
 
   const baseTitle = `Tag sample${modal ? "" : "s"} or labels`;
   const title = isReadOnly
-    ? `Can not ${baseTitle.toLowerCase()} in read-only mode.`
+    ? `Can not ${baseTitle.toLowerCase()} in read-only mode` + canTag.message
+      ? `: ${canTag.message}`
+      : ""
     : baseTitle;
 
   return (
--- a/app/packages/core/src/components/Actions/ActionsRow.tsx
+++ b/app/packages/core/src/components/Actions/ActionsRow.tsx
@@ -162,8 +162,9 @@ const Tag = ({
   const [available, setAvailable] = useState(true);
   const labels = useRecoilValue(fos.selectedLabelIds);
   const samples = useRecoilValue(fos.selectedSamples);
-  const readOnly =
-    useRecoilValue(fos.readOnly) || !useRecoilValue(fos.canTagSamplesOrLabels);
+  const canTag = useRecoilValue(fos.canTagSamplesOrLabels);
+  const readOnly = useRecoilValue(fos.readOnly);
+  const isReadOnly = readOnly || !canTag;
 
   const selected = labels.size > 0 || samples.size > 0;
   const tagging = useRecoilValue(fos.anyTagging);
@@ -180,7 +181,7 @@ const Tag = ({
     useEventHandler(lookerRef.current, "pause", () => setAvailable(true));
 
   const baseTitle = `Tag sample${modal ? "" : "s"} or labels`;
-  const title = readOnly
+  const title = isReadOnly
     ? `Can not ${baseTitle.toLowerCase()} in read-only mode.`
     : baseTitle;
 
@@ -188,7 +189,7 @@ const Tag = ({
     <ActionDiv ref={ref}>
       <PillButton
         style={{
-          cursor: readOnly
+          cursor: isReadOnly
             ? "not-allowed"
             : disabled || !available
             ? "default"
@@ -196,7 +197,7 @@ const Tag = ({
         }}
         icon={tagging ? <Loading /> : <LocalOffer />}
         open={open}
-        onClick={() => !disabled && available && !readOnly && setOpen(!open)}
+        onClick={() => !disabled && available && !isReadOnly && setOpen(!open)}
         highlight={(selected || open) && available}
         title={title}
         data-cy="action-tag-sample-labels"
--- a/app/packages/core/src/components/Actions/ActionsRow.tsx
+++ b/app/packages/core/src/components/Actions/ActionsRow.tsx
@@ -162,9 +162,7 @@ const Tag = ({
   const [available, setAvailable] = useState(true);
   const labels = useRecoilValue(fos.selectedLabelIds);
   const samples = useRecoilValue(fos.selectedSamples);
-  const canTag = useRecoilValue(fos.canTagSamples);
-  const isSnapshot = useRecoilValue(fos.readOnly);
-  const readOnly = isSnapshot || !canTag;
+  const readOnly = useRecoilValue(fos.readOnly) || !useRecoilValue(fos.canTagSamples);
 
   const selected = labels.size > 0 || samples.size > 0;
   const tagging = useRecoilValue(fos.anyTagging);
--- a/app/packages/core/src/components/Actions/ActionsRow.tsx
+++ b/app/packages/core/src/components/Actions/ActionsRow.tsx
@@ -162,7 +162,8 @@ const Tag = ({
   const [available, setAvailable] = useState(true);
   const labels = useRecoilValue(fos.selectedLabelIds);
   const samples = useRecoilValue(fos.selectedSamples);
-  const readOnly = useRecoilValue(fos.readOnly) || !useRecoilValue(fos.canTagSamples);
+  const readOnly =
+    useRecoilValue(fos.readOnly) || !useRecoilValue(fos.canTagSamplesOrLabels);
 
   const selected = labels.size > 0 || samples.size > 0;
   const tagging = useRecoilValue(fos.anyTagging);
--- a/app/packages/core/src/components/Actions/ActionsRow.tsx
+++ b/app/packages/core/src/components/Actions/ActionsRow.tsx
@@ -162,7 +162,9 @@ const Tag = ({
   const [available, setAvailable] = useState(true);
   const labels = useRecoilValue(fos.selectedLabelIds);
   const samples = useRecoilValue(fos.selectedSamples);
-  const readOnly = useRecoilValue(fos.readOnly);
+  const canTag = useRecoilValue(fos.canTagSamples);
+  const isSnapshot = useRecoilValue(fos.readOnly);
+  const readOnly = isSnapshot || !canTag;
 
   const selected = labels.size > 0 || samples.size > 0;
   const tagging = useRecoilValue(fos.anyTagging);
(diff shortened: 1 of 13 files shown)


lanzhenw:
@coderabbitai review

app/packages/core/src/components/Actions/ActionsRow.tsx
```diff
@@ -162,7 +162,9 @@ const Tag = ({
   const [available, setAvailable] = useState(true);
   const labels = useRecoilValue(fos.selectedLabelIds);
   const samples = useRecoilValue(fos.selectedSamples);
-  const readOnly = useRecoilValue(fos.readOnly);
+  const canTag = useRecoilValue(fos.canTagSamples);
+  const isSnapshot = useRecoilValue(fos.readOnly);
```
imanjra, line 166: I would suggest keeping it `readOnly` instead of `isSnapshot` so it can be used for other readOnly use cases in future

app/packages/core/src/components/Actions/ActionsRow.tsx
```diff
@@ -162,7 +162,9 @@ const Tag = ({
   const [available, setAvailable] = useState(true);
   const labels = useRecoilValue(fos.selectedLabelIds);
   const samples = useRecoilValue(fos.selectedSamples);
-  const readOnly = useRecoilValue(fos.readOnly);
+  const canTag = useRecoilValue(fos.canTagSamples);
```
imanjra, line 165: should we rename `canTagSamples` to `canTag` or `canTagSamplesOrLabels`?

app/packages/core/src/components/Sidebar/Entries/Draggable.tsx
```diff
@@ -19,8 +19,10 @@ const Draggable: React.FC<
   const theme = useTheme();
   const [hovering, setHovering] = useState(false);
   const [dragging, setDragging] = useState(false);
-  const isReadOnly = useRecoilValue(readOnly);
-  const isFieldVisibilityApplied = useRecoilValue(isFieldVisibilityActive);
+  const isSnapShot = useRecoilValue(fos.readOnly);
+  const canAddSidebarGroup = useRecoilValue(fos.canAddSidebarGroup);
```
imanjra, line 23: Should we use a different one here (i.e. `canUpdateSidebarOrder`) or rename it to something more general `canEditSidebar`?

imanjra (commented):
Outside of the scope of this PR, so feel free to follow-up or defer:
- Can we add support for restricting [renaming sidebar group](https://github.com/voxel51/fiftyone/blob/develop/app/packages/core/src/components/Sidebar/Entries/GroupEntries.tsx#L211)?
- Can we add support for restricting [deleting sidebar group](https://github.com/voxel51/fiftyone/blob/develop/app/packages/core/src/components/Sidebar/Entries/GroupEntries.tsx#L231)?

app/packages/core/src/components/Actions/ActionsRow.tsx
lanzhenw: Yes, let's go with the longer version to differentiate it from tagging datasets.

app/packages/core/src/components/Sidebar/Entries/Draggable.tsx
lanzhenw: How about`canModifySidebarGroups`?

lanzhenw:
> Outside of the scope of this PR, so feel free to follow-up or defer:
> 
> * Can we add support for restricting [renaming sidebar group](https://github.com/voxel51/fiftyone/blob/develop/app/packages/core/src/components/Sidebar/Entries/GroupEntries.tsx#L211)?
> * Can we add support for restricting [deleting sidebar group](https://github.com/voxel51/fiftyone/blob/develop/app/packages/core/src/components/Sidebar/Entries/GroupEntries.tsx#L231)?

I updated the flag name to `canModifySidebarGroups` to encapsulate the permission to change sidebar groups, including renaming, creating and deleting.

app/packages/state/src/session.ts
```diff
@@ -42,6 +45,9 @@ export const SESSION_DEFAULT: Session = {
   canEditCustomColors: true,
   canEditSavedViews: true,
   canEditWorkspaces: true,
+  canCreateNewField: true,
+  canModifySidebarGroup: true,
+  canTagSamplesOrLabels: true,
```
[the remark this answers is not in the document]
lanzhenw: No, that's not the designed default or fallback behavior.

app/packages/core/src/components/Actions/ActionsRow.tsx
```diff
@@ -162,7 +162,9 @@ const Tag = ({
   const [available, setAvailable] = useState(true);
   const labels = useRecoilValue(fos.selectedLabelIds);
   const samples = useRecoilValue(fos.selectedSamples);
+  const canTag = useRecoilValue(fos.canTagSamplesOrLabels);
   const readOnly = useRecoilValue(fos.readOnly);
+  const isReadOnly = readOnly || !canTag;
```
manivoxel51, line 167: minor - feels confusing having `isReadOnly` deriving from `readOnly` and another condition.

Can we instead bake in the `!canTag` as part of `readOnly`? so this line will become
```
const isReadOnly = readOnly;
```
or can be completely and readOnly instead being used.

manivoxel51 (commented):
looking great @lanzhenw. I had one inline comment 🍨

app/packages/core/src/components/Sidebar/Entries/GroupEntries.tsx
```diff
@@ -132,6 +132,9 @@ const GroupEntry = React.memo(
     const canCommit = useRef(false);
     const theme = useTheme();
     const notify = fos.useNotification();
+    const readOnly = useRecoilValue(fos.readOnly);
+    const canModifySidebarGroup = useRecoilValue(fos.canModifySidebarGroup);
+    const canModify = !readOnly && canModifySidebarGroup;
```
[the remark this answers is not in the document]
lanzhenw: The name has already been updated to `canModifySidebarGroup` in previous commits.

app/packages/core/src/components/Actions/ActionsRow.tsx
lanzhenw: Ibrahim suggested that we keep the `readOnly` name from reading in the state. Meanwhile, `!canTag` is also necessary in this `isReadOnly`, because `readOnly` right now reflects `isSnapshot` (in the future it will possibly capture more).

imanjra: Good point. I believe in some parts of the app where the combination of readOnly and permission atom is used, we use `canEdit` so that's a possible alternative. For example:

```js
const canEdit = !readOnly && canTag;
// Then, we would invert all the usage below
```

app/packages/core/src/components/Actions/ActionsRow.tsx
```diff
@@ -179,23 +181,23 @@ const Tag = ({
     useEventHandler(lookerRef.current, "pause", () => setAvailable(true));
 
   const baseTitle = `Tag sample${modal ? "" : "s"} or labels`;
-  const title = readOnly
+  const title = isReadOnly
     ? `Can not ${baseTitle.toLowerCase()} in read-only mode.`
```
imanjra, line 185: Thoughts on splitting this message? For example:

- Has permission and is not read-only: Tag samples or labels
- Does not have permission: You do not have permission to tag samples or labels
- Readonly mode: Can not tag samples or labels in read-only mode.

imanjra (approved):
LGTM. Nice work! 🎉 

Left a comment which can be follow-up if needed

manivoxel51 (approved):
Nice work! 🍨 

thanks @imanjra for helping us out! 💯

benjaminpkane (changes requested):
Can we move the business logic to the internal App?

The below doesn't need to be in OSS. The `fos.canTagSamplesOrLabels` session atom can capture the condition
```typescript
const isReadOnly = readOnly || !canTag;
```

lanzhenw:
> Can we move the business logic to the internal App?
> 
> The below doesn't need to be in OSS. The `fos.canTagSamplesOrLabels` session atom can capture the condition
> 
> ```ts
> const isReadOnly = readOnly || !canTag;
> ```

This is still following the same pattern as before. We decoupled "!canEdit" from `readOnly`, so `readOnly` right now refers to "isSnapshot" (but will likely expend in future per @imanjra ).

app/packages/core/src/components/Actions/ActionsRow.tsx
lanzhenw: Just saw this. Yes, I will implement it in the messages.

app/packages/core/src/components/Actions/ActionsRow.tsx
```diff
@@ -162,13 +162,14 @@ const Tag = ({
   const [available, setAvailable] = useState(true);
   const labels = useRecoilValue(fos.selectedLabelIds);
   const samples = useRecoilValue(fos.selectedSamples);
-  const readOnly = useRecoilValue(fos.readOnly);
+  const canTag = useRecoilValue(fos.canTagSamplesOrLabels);
```
manivoxel51, line 165: cool, this should work even after introducing the TAG permission since TAG < EDIT. so no feature flagging for dynamic collaborator is required on teams side.

app/packages/core/src/components/Actions/ActionsRow.tsx
```diff
@@ -162,13 +162,14 @@ const Tag = ({
...
+  const disableTag = canTag.enabled !== true;
```
manivoxel51, line 166:
```suggestion
  const disableTag = !canTag.enabled;
```
assuming enabled is a required property and not being set means should not be enabled

app/packages/core/src/components/Actions/similar/Similar.tsx
```diff
@@ -93,7 +93,9 @@ const SortBySimilarity = ({
     []
   );
   const isLoading = useRecoilValue(fos.similaritySorting);
-  const isReadOnly = useRecoilValue(fos.readOnly);
+  const canCreateNewField = useRecoilValue(fos.canCreateNewField);
+  const disabled = canCreateNewField.enabled !== true;
```
manivoxel51, line 97: same here

app/packages/core/src/components/Sidebar/Entries/GroupEntries.tsx
```diff
@@ -207,7 +209,7 @@ const GroupEntry = React.memo(
                   }
                 }}
               />
-              {hovering && !editing && setValue && (
+              {hovering && !editing && setValue && !disabled && (
```
manivoxel51, line 212: should we disable it instead of hiding it? and show the message passed down? `canModifySidebarGroup.message`

app/packages/core/src/components/Sidebar/Entries/GroupEntries.tsx
```diff
@@ -228,7 +230,7 @@ const GroupEntry = React.memo(
                 </span>
               )}
               {pills}
-              {onDelete && !editing && (
+              {onDelete && !editing && !disabled && (
```
manivoxel51, line 233: same here

app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
```diff
@@ -145,7 +143,7 @@ export default function ViewSelection() {
 
   useEffect(() => {
     const callback = (event: KeyboardEvent) => {
-      if (!canEdit) {
+      if (!disabled) {
```
manivoxel51, line 146:
```suggestion
      if (disabled) {
```
right?

manivoxel51 (commented):
🍨 

left couple of comments - one major, rest minor

benjaminpkane (approved):

PR #4393, opened 2024-05-10
Adding a clean step to Makefile

minhtuev:

Adding a clean step to Makefile so that we don't bundle existing dist content. I observed that running `make python` repeatedly right now will increase the size of the archived files because we are also bundling the dist content as well :) 

Tested locally and verified that the archived file no longer contains the dist folder.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Chores**
	- Updated the packaging process to exclude the `dist` directory, ensuring cleaner distributions.
	- Enhanced the build process with a `clean` target to remove old distributions before creating new ones.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -10,6 +10,7 @@ prune package
 prune requirements
 prune tools
 prune tests
+prune dist
 
 global-exclude .*
 
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,10 @@
 app:
 	@cd app && yarn && yarn build && cd ..
 
-python: app
+clean:
+	@rm -rf ./dist/*
+
+python: app clean
 	@python -Im build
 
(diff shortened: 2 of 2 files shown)


findtopher:
I would suggest adding `prune dist` to the `MANIFEST.in` file too - that will prevent it from being included as well.

I wonder how many times we've shipped out wheels that had wheels in them 😊

minhtuev:
> I would suggest adding `prune dist` to the `MANIFEST.in` file too - that will prevent it from being included as well.
> 
> I wonder how many times we've shipped out wheels that had wheels in them 😊

Wheel within wheel within wheel, sounds like a Dune reference :)

benjaminpkane (approved):
Great catch! LGTM 🙇

PR #4395, opened 2024-05-13
misc 3d bug fixes

sashankaryal:

- Make name arg explicit in lights constructors
- Fix camera args bug where undefined | null aspect ratio causes canvas to hang

## Summary by CodeRabbit

- **New Features**
	- Enhanced lighting options in 3D environments by allowing users to assign custom names to different types of lights for better organization and manipulation.
	- Improved handling of camera properties in the 3D media type component to enhance the viewing experience.

### DIFF
--- a/fiftyone/core/threed/lights.py
+++ b/fiftyone/core/threed/lights.py
@@ -7,7 +7,7 @@ Lights definition for 3D visualization.
 """
 
 from math import pi as PI
-from typing import Optional
+from typing import Optional, Union
 
 from .object_3d import Object3D
 
@@ -35,7 +35,7 @@ class Light(Object3D):
 
     def __init__(
         self,
-        name: str,
+        name: Union[str, None] = None,
         color: str = COLOR_DEFAULT_WHITE,
         intensity: float = 1.0,
         visible=True,
@@ -44,7 +44,7 @@ class Light(Object3D):
         quaternion: Optional[Quaternion] = None,
     ):
         super().__init__(
-            name=name,
+            name=name or self.__class__.__name__,
             visible=visible,
             position=position,
             scale=scale,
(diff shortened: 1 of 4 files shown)


fiftyone/core/threed/lights.py
```diff
@@ -43,7 +44,7 @@ def __init__(
         quaternion: Optional[Quaternion] = None,
     ):
         super().__init__(
-            name=self.__class__.__name__,
+            name=name,
```
minhtuev, line 47: and `name=name or self.__class__.__name__`?

minhtuev:
Added a small comment, if we can pass in the class name in the parent class, we don't have to pass in the individual class name :)

ritch (approved):

fiftyone/core/threed/lights.py
```diff
@@ -68,6 +69,7 @@ class AmbientLight(Light):
     This light globally illuminates all objects in the scene equally.
 
     Args:
+        name ("AmbientLight"): the name of the light
```
minhtuev, line 72: since we are setting the default value to the class name in the base `Light` class, we do not need to explicitly set this `name` value for any child class anymore :)

fiftyone/core/threed/lights.py
```diff
@@ -126,6 +131,7 @@ class DirectionalLight(Light):
 
     def __init__(
         self,
+        name: str = "DirectionalLight",
```
minhtuev, line 134: same!

fiftyone/core/threed/lights.py
sashankaryal: We don't, but omitting this means `name` arg isn't explicit in the docstring / API. This way you don't have to hop back in the inheritance tree (eg: `PointLight -> Light -> Object3D`). We do this for other attributes like `position`, `quaternion`, etc.

PR #4386, opened 2024-05-10
Teams integration

benjaminpkane:
## What changes are proposed in this pull request?

Automated integration into the Teams SDK

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [x] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced a new job to trigger a workflow in another repository with specific configurations.
  - Enhanced the existing test job to include the new workflow job as a dependency.

- **Chores**
  - Updated the release workflow to improve checkout efficiency with a `fetch-depth: 0` parameter.

PR #4405, opened 2024-05-21
[Proposal] Automatic PRs into `develop`

benjaminpkane:

Proposal to add automatic PRs into `develop` when `main` or `release/*` branches are pushed to

-   [x] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced a GitHub Actions workflow "Push Release" to automate merging changes from `main` or `release` branches into `develop`.

### DIFF
--- a/.github/workflows/push-release.yml
+++ b/.github/workflows/push-release.yml
@@ -11,6 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: develop
       - run: git pull origin ${{ github.ref_name }} --no-rebase
       - uses: peter-evans/create-pull-request@v6
         with:
--- a/.github/workflows/push-release.yml
+++ b/.github/workflows/push-release.yml
@@ -1,0 +1,22 @@
+name: Push Release
+
+on:
+  push:
+    branches:
+      - main
+      - release/v[0-9]+.[0-9]+.[0-9]+
+
+jobs:
+  pr:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


findtopher (approved):
This looks like it will take all the `release/v*` and all the `develop` merges and create PRs for them, which is I think what you're looking to do...

Obviously we'll have to put process around managing those PRs once they're created, but it feels right.

I'm not entirely sure how to manage a mapping of `fiftyone:release/v0.24.0` to `fiftyone-teams:release/v1.7.0` to "close the loop" and ensure that release PRs wind up going to Teams releases...  🤔 

It's possible that we need to make a different job that has hard-coded branches in it and update those branches as part of throwing the release branches?  Will need to consider it... (because we know we'll wind up in a situation with two or three release branches at the same time...)

Those are all just thoughts for the future - this looks good today.

benjaminpkane:
> This looks like it will take all the `release/v*` and all the `develop` merges and create PRs for them, which is I think what you're looking to do...
> 
> Obviously we'll have to put process around managing those PRs once they're created, but it feels right.
> 
> I'm not entirely sure how to manage a mapping of `fiftyone:release/v0.24.0` to `fiftyone-teams:release/v1.7.0` to "close the loop" and ensure that release PRs wind up going to Teams releases... 🤔
> 
> It's possible that we need to make a different job that has hard-coded branches in it and update those branches as part of throwing the release branches? Will need to consider it... (because we know we'll wind up in a situation with two or three release branches at the same time...)
> 
> Those are all just thoughts for the future - this looks good today.

Thanks! Keeping these thoughts in mind in #4386

PR #4408, opened 2024-05-22
fix panel state changes

ritch:
Prior to this change the following would display the previous choice:

```python
    def render(self, ctx):
        # ...
        choices = types.Choices()
        choices.add_choice(1, label="One")
        choices.add_choice(2, label="Two")
        choices.add_choice(3, label="Three")
        grid.enum('choice', values=choices.values(), on_change=self.on_change_choice)
        return types.Property(panel)

    def on_change_choice(self, ctx):
        print("Choice changed", ctx.panel.state)
        # this would show the previously selected choice! this PR fixes that issue
        ctx.ops.notify(f"Choice changed to {ctx.panel.state.grid.get('choice', 'No choice')}")
```

## Summary by CodeRabbit

- **Refactor**
  - Improved performance and reliability of context change events in custom panels.
  - Removed unnecessary console logs for cleaner code execution.

### DIFF
--- a/app/packages/operators/src/useCustomPanelHooks.ts
+++ b/app/packages/operators/src/useCustomPanelHooks.ts
@@ -44,12 +44,8 @@ export interface CustomPanelHooks {
 
 function useCtxChangePanelEvent(panelId, value, operator) {
   const triggerCtxChangedEvent = usePanelEvent();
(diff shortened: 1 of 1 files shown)


imanjra (approved):
LGTM

PR #4414, opened 2024-05-23
add panel state deep get and set methods

ritch:
Added several new ways of reading and writing panel state and data. Here's a summary of how you can interact with a `Panel` via `PanelRef` after these changes:

```python
# Setting and getting panel state
panel.set_state("user.settings.language", "en")
print(panel.get_state("user.settings.language"))  # => "en"

# Simple properties can be set via dot syntax
panel.state.foo = "bar"

# Can also read panel.state with dot syntax
print(panel.state.foo) # => "bar"

# Setting panel data
panel.set_data("foo.bar", {"bat": "baz"})

# Trying to read panel data fails
panel.data.foo # throws WriteOnlyError

# Clearing panel state and data
panel.state.clear()
panel.data.clear()

# Closing the panel
panel.close()
```

### DIFF
--- a/fiftyone/operators/panel.py
+++ b/fiftyone/operators/panel.py
@@ -9,6 +9,8 @@ FiftyOne operators.
 import fiftyone.operators.types as types
 from fiftyone.operators.operator import OperatorConfig, Operator
 
+import pydash
+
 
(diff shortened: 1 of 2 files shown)


imanjra (approved):
LGTM

Issue #4428, opened 2024-05-27
[BUG]


Anton-Cherepkov:
### Describe the problem

When using [a sample code from the FiftyOne's docs](https://docs.voxel51.com/integrations/super_gradients.html) for visualizing the predictions of a super-gradients `yolo_nas_m` model, the detections do not appear on the FiftyOne's Web UI.

### Code to reproduce issue

Provide a reproducible test case that is the bare minimum necessary to generate
the problem. Please avoid sharing code that relies on your local data or
datasets. Include a short video or screenshot if the bug is in the App.

```
import fiftyone as fo
import fiftyone.zoo as foz

from super_gradients.training import models

dataset = foz.load_zoo_dataset("quickstart", max_samples=5)
dataset.select_fields().keep_fields()

model = models.get("yolo_nas_m", pretrained_weights="coco")

dataset.apply_model(model, label_field="yolo_nas", confidence_thresh=0.7)

session = fo.launch_app(dataset)
```

![image](https://github.com/voxel51/fiftyone/assets/11422372/dff2dc38-283c-43ca-bd85-c48e4d424aee)

### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 22.04): Linux Ubuntu 22.04
-   **Python version** (`python --version`): Python 3.10.14
-   **FiftyOne version** (`fiftyone --version`): FiftyOne v0.23.8, Voxel51, Inc.
-   **FiftyOne installed from** (pip or source): pip

### Other info/logs

The only warning in the console was:
```
/app/src/super_gradients/training/utils/media/image.py:64: UserWarning: It seems you are using predict() with batch input
  warnings.warn("It seems you are using predict() with batch input")
```

My investigation showed that:
- `labels_batch` object [here](https://github.com/voxel51/fiftyone/blob/v0.23.8/fiftyone/core/models.py#L437) holds correct detections

![image](https://github.com/voxel51/fiftyone/assets/11422372/34a75654-ede2-451d-83ac-8253c77d9f1d)

- `labels` object [here](https://github.com/voxel51/fiftyone/blob/v0.23.8/fiftyone/core/models.py#L444) has type `str` and does not contain any detections

![image](https://github.com/voxel51/fiftyone/assets/11422372/cdfbe6b6-eda8-462c-9475-b53cf3313749)

My conclusion: the following iteration construction `for label in labels_batch` does not work properly.

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another
[...]

PR #4429, opened 2024-05-27
Fix YOLO-NAS inference

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/4428

## Context

@jacobmarks as far as I can tell, `TorchYoloNasModel` was previously broken because its `_predict_all()` method returned a single `Detections` instance, but the `TorchImageModel` contract requires `_predict_all()` to return a list of `Detections`.

Note that, as currently implemented, `TorchYoloNasModel` does not support batching because it produces a model with `ragged_batches == True`:

```py
from super_gradients.training import models
import fiftyone.utils.super_gradients as fousg

sg_model = models.get("yolo_nas_m", pretrained_weights="coco")
fo_model = fousg._convert_yolo_nas_detection_model(sg_model)

print(fo_model.ragged_batches)  # True
```

which means that `apply_model()` will not allow a `batch_size` to be provided: 
https://github.com/voxel51/fiftyone/blob/6faa5d72f61a566f58b85c4a8dbd8df39b9f4e18/fiftyone/core/models.py#L1807

If the underlying SuperGradients model instance does in fact support batching, then the `TorchYoloNasModel` wrapper would need to be tweaked to allow this to occur.

In the meantime, this PR fixes inference by implementing `predict()` and `predict_all()` correctly.

## Example usage

The following now works as expected:

```
pip install super-gradients
```

```py
import fiftyone as fo
import fiftyone.zoo as foz

from super_gradients.training import models

dataset.select_fields().keep_fields()

model = models.get("yolo_nas_m", pretrained_weights="coco")

dataset.apply_model(model, label_field="yolo_nas", confidence_thresh=0.7)

```

## Summary by CodeRabbit

- **Refactor**
  - Simplified configuration setup for model conversion.
  - Improved model loading logic to handle GPU availability more effectively.
  - Updated prediction methods to support both single and multiple image predictions.

### DIFF
--- a/fiftyone/utils/super_gradients.py
+++ b/fiftyone/utils/super_gradients.py
@@ -49,13 +49,7 @@ def convert_super_gradients_model(model):
 
 
 def _convert_yolo_nas_detection_model(model):
-    config_model = {
-        "model": model,
-        "labels_path": "{{eta-resources}}/ms-coco-labels.txt",
-        "output_processor_cls": "fiftyone.utils.torch.ClassifierOutputProcessor",
-        "raw_inputs": True,
-    }
-
+    config_model = {"model": model, "raw_inputs": True}
     config = TorchYoloNasModelConfig(config_model)
     return TorchYoloNasModel(config)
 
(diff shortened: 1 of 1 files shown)


jacobmarks (approved):
LGTM 👍 and works on my end. Will need to check into the underlying model at some point. For the sake of the upcoming release, this works.

PR #4423, opened 2024-05-24
fix dataset group slice is not being used on launch_app

imanjra:
## What changes are proposed in this pull request?

Fixes an issue where `dataset.group_slice` is not being used in the app on `launch_app`

## How is this patch tested? If it is not, please explain why.

Using snippet below in py shell and ensuring app is opened with `pcd` slice selected

```py
import fiftyone as fo
dataset = fo.load_dataset("quickstart-groups")
dataset.group_slice = "pcd"
session = fo.launch_app(dataset)
```

### Is this a user-facing change that should be mentioned in the release notes?

-   [ ] No. You can skip the rest of this section.
-   [x] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

Fix launch_app not using `dataset.group_slice`

### What areas of FiftyOne does this PR affect?

-   [x] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Enhanced session initialization to support group slicing when a dataset is present.

PR #4434, opened 2024-05-28
Avoid slow session launch

benjaminpkane:

Follow up to https://github.com/voxel51/fiftyone/pull/4423 to avoid `len(dataset)` calls which can be slow. Also adds `view` group slice handling

```py
import fiftyone as fo
dataset = fo.load_dataset("quickstart-groups")
dataset.group_slice = "pcd"

# view
view = dataset.view()
view.group_slice = "right"
session = fo.launch_app(view)
```


imanjra (approved):
LGTM

PR #4442, opened 2024-05-30
3D Scene I/O optimization and bugfixes

brimoor:
## Change log

Optimizations
- Added a cache to `compute_metadata()` to avoid recomputing metadata for repeated 3D asset paths

Bugfixes
- Handle partial maps when updating Scene asset paths
    - Unit test added would previously fail but now succeeds as expected

## Summary by CodeRabbit

- **New Features**
  - Introduced `include_assets` parameter in various `fo.Dataset` methods to handle 3D scene assets.
  - Added `download_scenes` method to the `fo.Dataset` class.

- **Bug Fixes**
  - Enhanced asset path handling in `SceneMetadata` and `Scene` classes to ensure correct path resolution and scene modification.
  - Updated `Exporter` class to handle different export modes based on `self.export_mode`.

- **Tests**
  - Added `test_update_asset_paths` method to verify asset path updates in scenes.
  - Adjusted unit tests to align with new path handling logic.

### DIFF
--- a/fiftyone/core/storage.py
+++ b/fiftyone/core/storage.py
@@ -282,8 +282,8 @@ def join(a, *p):
 
 
 def resolve(path):
-    """Resolves path to absolute, resolving symlinks and relative path
-        indicators such as `.` and `..`.
+    """Resolves the given path to absolute, resolving symlinks and relative
+    path indicators such as ``.`` and ``..``.
 
     Args:
         path: the filepath
--- a/fiftyone/core/storage.py
+++ b/fiftyone/core/storage.py
@@ -281,8 +281,8 @@ def join(a, *p):
     return os.path.join(a, *p)
 
 
-def resolve(path):
-    """Resolves the given path to absolute, resolving symlinks and relative
+def realpath(path):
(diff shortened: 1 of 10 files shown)


findtopher:
should this be targeting `release/v0.24.1` to go out with `release/v1.7.1` or is this a `develop`-only bug?

fiftyone/core/storage.py
```diff
@@ -281,9 +281,9 @@ def join(a, *p):
     return os.path.join(a, *p)
 
 
-def resolve(path):
-    """Resolves path to absolute, resolving symlinks and relative path
-        indicators such as `.` and `..`.
+def realpath(path):
```
benjaminpkane, line 284: Just noting this is a public API change

brimoor: Yes true. My thinking was that this was only introduced in `fiftyone==0.24.0` and the spirit of the `fiftyone.core.storage` module is to exactly match the `os` package's naming convention for the methods that it wraps, so this felt consistent.

benjaminpkane (approved):
LGTM 🚀

PR #4444, opened 2024-05-31
Fixing anomaly detection tutorial title level

brimoor:
## Summary by CodeRabbit

- **Documentation**
  - Updated the header format in the "Anomaly Detection with FiftyOne and Anomalib" tutorial for improved readability.

PR #4416, opened 2024-05-23
Allow generated datasets to be directly created

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/4397

The `make_{patches|frames|clips}_dataset()` methods were originally designed to be called internally by the `to_{patches|frames|clips}()` methods but not called directly. For example, as observed in https://github.com/voxel51/fiftyone/issues/4397, they currently return a dataset that behaves like a view in certain ways. Also, directly calling `make_clips_dataset()` is *especially* pernicious, as it currently returns a dataset that silently reuses the same frames collection as the input dataset.

However, there is a good use case for directly calling `make_{patches|frames|clips}_dataset()` to convert a collection into an independent dataset.

So, this PR maintains the behavior of `to_{patches|frames|clips}()` while tweaking the default behavior of the `make_{patches|frames|clips}_dataset()` methods so that they do return a completely normal dataset.

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.core.patches as fop
import fiftyone.core.clips as foc
import fiftyone.core.video as fov
from fiftyone import ViewField as F

#
# Patches dataset
#

dataset = foz.load_zoo_dataset("quickstart")

patches_view = dataset.to_patches("ground_truth")
patches_dataset = fop.make_patches_dataset(dataset, "ground_truth")

assert patches_view._is_generated == True
assert patches_dataset._is_generated == False
assert patches_dataset._sample_collection_name != dataset._sample_collection_name
assert patches_dataset._frame_collection_name is None
assert patches_dataset.count() == patches_view.count()
assert patches_dataset.count() == dataset.count("ground_truth.detections")

#
# Frames dataset
#

dataset = foz.load_zoo_dataset("quickstart-video")

frames_view = dataset.to_frames(sample_frames=True)
frames_dataset = fov.make_frames_dataset(dataset, sample_frames=True)

assert frames_view._is_generated == True
assert frames_dataset._is_generated == False
assert frames_dataset._sample_collection_name != dataset._sample_collection_name
assert frames_dataset._frame_collection_name is None
assert frames_dataset.count() == frames_view.count()
assert frames_dataset.count() == dataset.count("frames")

#
# Clips dataset
#

dataset = foz.load_zoo_dataset("quickstart-video")

expr = F("detections.detections").length() > 10
clips_view = dataset.to_clips(expr)
clips_dataset = foc.make_clips_dataset(dataset, expr)

assert clips_view._is_generated == True
[...]

PR #4398, opened 2024-05-19
add ultralytics yolov8 open images v7 pretrained models into zoo

fcakyon:
## What changes are proposed in this pull request?

Update model zoo metadata to include Open Images v7 pretrained Ultralytics YOLOv8 model weights.

Usage:

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.brain as fob
from fiftyone import ViewField as F
dataset = foz.load_zoo_dataset("quickstart", max_samples=10)

model = foz.load_zoo_model('yolov8m-oiv7-torch')
dataset.apply_model(model, label_field="oi")
```

Fixes https://github.com/voxel51/fiftyone/issues/4399.

What do you think about the PR @jacobmarks?

### Is this a user-facing change that should be mentioned in the release notes?

-   [ ] No. You can skip the rest of this section.
-   [x] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

Add Ultralytics YOLOv8 weights pretrained on Open Images v7 dataset with 600 target classes into Fiftyone Model Zoo.

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [x] Core: Core `fiftyone` Python library changes
-   [x] Documentation: FiftyOne documentation changes
-   [ ] Other

PR #4392, opened 2024-05-10
Support loading annotations for large CVAT tasks with many jobs

ehofesmann:
## What changes are proposed in this pull request?

Optimized loading annotations from the CVAT backend. Annotations are now loaded from individual jobs instead of entire tasks which allows for importing annotations from much larger task sizes. There is one task in the internal CVAT deployment with 10k samples, in 200 jobs of 50 samples. Previously, trying to load this task would make a single request to the CVAT server to load all annotations from the task at once, this crashes the CVAT server. Now, annotations from each job are loaded sequentially which resolves this problem.

## How is this patch tested? If it is not, please explain why.

Unit tests pass:
```
export FIFTYONE_CVAT_URL=...
export FIFTYONE_CVAT_USERNAME=...
export FIFTYONE_CVAT_PASSWORD=...

pytest /path/to/fiftyone/tests/intensive/cvat_tests.py
```

Also task 159 on the internal CVAT test deployment containing bdd100k-validation now imports properly. It is recommended you have bdd100k validation images available locally on disk as it makes this easier:
```python
import fiftyone as fo
import fiftyone.utils.cvat as fouc
import os

cvat_url = "..."
cvat_username = "..."
cvat_password = "..."

bdd_path = "/path/to/bdd100k-validation/"
filepaths = os.list_dir(bdd_path)
data_map = {fp: os.path.join(bdd_path, fp) for fp in fps}

dataset = fo.Dataset()
# WARNING: Only run this on this branch, this will crash the CVAT deployment if run on `develop`
fouc.import_annotations(dataset, task_ids=[159], data_path=data_map, url=cvat_url, username=cvat_username, password=cvat_password)
```

### Is this a user-facing change that should be mentioned in the release notes?

-   [ ] No. You can skip the rest of this section.
-   [x] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

Optimized loading annotations from the CVAT backend. Annotations are now loaded from individual jobs instead of entire tasks which allows for importing annotations from much larger task sizes.

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [x] Other: Annotation integrations

## Summary by CodeRabbit

- **New Features**
	- Added a method to generate URLs for job annotations in the CVAT tool.
- **Refactor**
[...]

PR #4455, opened 2024-06-05
Cherry picking PRs into v0.24.1

brimoor:
These PRs were included in the release notes for 0.24.0 but somehow appear to have not actually been released (I've moved their associated release notes into v0.24.1 in https://github.com/voxel51/fiftyone/pull/4456):
- https://github.com/voxel51/fiftyone/pull/4398
- https://github.com/voxel51/fiftyone/pull/4416

Also cherry-picks these bug fixes:
- https://github.com/voxel51/fiftyone/pull/4442
- https://github.com/voxel51/fiftyone/pull/4444

And the following PR which just missed the cutoff from v0.24.0 but has tests:
- https://github.com/voxel51/fiftyone/pull/4392

## Summary by CodeRabbit

- **New Features**
  - Introduced methods for downloading scenes and 3D assets with configurable options.
  - Added parameters to control dataset persistence and creation in various dataset functions.

- **Enhancements**
  - Improved handling of 3D scene metadata and asset paths.
  - Enhanced CVAT integration with refined parameter descriptions and logic updates.
  - Refactored dataset exporter logic for better efficiency.

- **Bug Fixes**
  - Corrected logic for counting file sizes and asset counts in scene metadata tests.
  - Fixed path resolution issues in 3D scene and asset handling.

- **Tests**
  - Added new test methods for patches, scenes, and video datasets.
  - Updated existing tests to reflect new parameters and logic changes.

### DIFF
--- a/fiftyone/core/patches.py
+++ b/fiftyone/core/patches.py
@@ -558,6 +558,8 @@ def make_patches_dataset(
     other_fields=None,
     keep_label_lists=False,
     name=None,
+    persistent=False,
+    _generated=False,
 ):
     """Creates a dataset that contains one sample per object patch in the
     specified field of the collection.
@@ -586,6 +588,8 @@ def make_patches_dataset(
             fields of the same type as the input collection rather than using
             their single label variants
         name (None): a name for the dataset
+        persistent (False): whether the dataset should persist in the database
+            after the session terminates
 
     Returns:
         a :class:`fiftyone.core.dataset.Dataset`
@@ -606,7 +610,12 @@ def make_patches_dataset(
         sample_collection, field, keep_label_lists
     )
 
-    dataset = fod.Dataset(name=name, _patches=True, _frames=is_frame_patches)
+    dataset = fod.Dataset(
+        name=name,
+        persistent=persistent,
+        _patches=_generated,
+        _frames=is_frame_patches and _generated,
+    )
     dataset.media_type = fom.IMAGE
     dataset.add_sample_field("sample_id", fof.ObjectIdField)
     dataset.create_index("sample_id")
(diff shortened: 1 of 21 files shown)


fiftyone/core/patches.py
```diff
@@ -720,6 +734,8 @@ def make_evaluation_patches_dataset(
             -   ``True`` to include all other fields
             -   ``None``/``False`` to include no other fields
         name (None): a name for the dataset
+        persistent (False): whether the dataset should persist in the database
```
kaixi-wang, line 737: if patches (or any generated view) are now persistent, this might trigger unexpected dataset permissioning issues in teams/might require a hub change 

cc @j053y

kaixi-wang (commented):
Slightly concerned making generated datasets persistent may cause unexpected view permission issues... is there an expectation that if someone has view access to a dataset, they would also have view access to the persisted patches/frames dataset?

fiftyone/core/patches.py
brimoor: The point of the PR being cherry-picked here (which is already in `develop`) is to allow these utilities to be called directly, in which case they return ordinary datasets.

They are *not* the fancy generated datasets that occur when the undocumented `_generated=True` flag is pased.

findtopher: I guess the question is, is this a _patch_ or a new functionality?

Seems like something new to me.  I would suggest this should be included in, and tested in, 1.8.0, in my super uninformed analysis of new Other Fields.

brimoor: We already announced two of these things and recognized a community contribution in Slack, so this is a patch to correct the omission.

findtopher: Sure, new functionality instead of just a minor release seems fine.

Why do we object to putting this in a minor release and releasing it this week, and apologizing for our omission since this PR is happening in a public repo where anybody who cares already knows it's happened?

Semver means things.

findtopher (approved):
sure, why not?  hate to stand in the way of progress.

benjaminpkane (dismissed):

PR #4469, opened 2024-06-07
Refetch saved views on dataset change

sashankaryal:
## Summary by CodeRabbit

- **New Features**
  - Enhanced `ViewSelection` component to automatically refetch data when the dataset name changes.

- **Bug Fixes**
  - Improved state management for dataset group slices to handle exceptions more gracefully.

### DIFF
--- a/app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
+++ b/app/packages/core/src/components/Sidebar/ViewSelection/index.tsx
@@ -81,6 +81,10 @@ export default function ViewSelection() {
   );
 
   useEffect(() => {
+    refetch({ name: datasetName });
+  }, [datasetName]);
+
+  useEffect(() => {
     if (
       selected &&
       selected?.id !== fos.DEFAULT_SELECTED.id &&
--- a/fiftyone/server/events/initialize.py
+++ b/fiftyone/server/events/initialize.py
@@ -132,6 +132,7 @@ def handle_dataset_change(
     """
     try:
         state.dataset = fod.load_dataset(initializer.dataset)
+        state.group_slice = state.dataset.group_slice
         state.selected = []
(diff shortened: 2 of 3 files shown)


brimoor:
@sashankaryal @benjaminpkane confirmed that saved views are refetched as expected when navigating between datasets via the dataset selector in the App 👍👍 

However, I do notice the following issues when navigating by editing the URL directly:
1. If I try to remove `?view=VIEW` from the URL and press ENTER, it is added back to the URL
2. If I try to navigate from `http://localhost:5151/datasets/DATASET1?view=VIEW` directly to `http://localhost:5151/datasets/DATASET2`, it instead tries to load `http://localhost:5151/datasets/DATASET2?view=VIEW`, which causes an error 

Is URL-based navigation an easy patch as well?

benjaminpkane (dismissed):
🙇

benjaminpkane:
> @sashankaryal @benjaminpkane confirmed that saved views are refetched as expected when navigating between datasets via the dataset selector in the App 👍👍
> 
> However, I do notice the following issues when navigating by editing the URL directly:
> 
> 1. ~If I try to remove `?view=VIEW` from the URL and press ENTER, it is added back to the URL~
>    
>    * According to [fix workspace persistence issue #4471 (comment)](https://github.com/voxel51/fiftyone/pull/4471#issuecomment-2156081313), this point is a feature, not a bug
> 2. If I try to navigate from `http://localhost:5151/datasets/DATASET1?view=VIEW` directly to `http://localhost:5151/datasets/DATASET2`, it instead tries to load `http://localhost:5151/datasets/DATASET2?view=VIEW`, which causes an error
> 
> Is URL-based navigation an easy patch as well?

I will push a tweak 👍

benjaminpkane (approved):

PR #4492, opened 2024-06-13
Don't mutate expressions passed to to_clips() in-place

brimoor:
```py
import fiftyone as fo
import fiftyone.zoo as foz

vehicles = F("detections.detections").filter(F("label") == "vehicle")
clips = dataset.to_clips(vehicles.length() >= 2)

dataset.save_view("clips", clips)

also_clips = dataset.load_saved_view("clips")

# Bugfix: `ToClips._state` should not change, so existing `_dataset` is now reused
assert clips is not also_clips
assert clips._dataset.name == also_clips._dataset.name  # previously failed
```

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method to handle expressions with a specific prefix.
  
- **Bug Fixes**
  - Improved handling of expressions to prevent in-place modifications, ensuring data integrity during operations.

- **Tests**
  - Added tests to verify correct behavior of view creation and serialization with frozen expressions.

### DIFF
--- a/fiftyone/core/clips.py
+++ b/fiftyone/core/clips.py
@@ -974,9 +974,12 @@ def _write_expr_clips(
         _, path = src_collection._get_label_field_path(expr)
         leaf, _ = src_collection._handle_frame_field(path)
         expr = F(leaf).length() > 0
-
-    if isinstance(expr, dict):
+    elif isinstance(expr, dict):
         expr = foe.ViewExpression(expr)
+    else:
+        # map() modifies the expression in-place and we don't want to cause
+        # side effects to the caller
+        expr = deepcopy(expr)
 
     frame_numbers, bools = src_collection.values(
         ["frames.frame_number", F("frames").map(expr)]
--- a/fiftyone/core/expressions.py
+++ b/fiftyone/core/expressions.py
@@ -257,11 +257,6 @@ class ViewExpression(object):
     def _freeze_prefix(self, prefix):
         _do_freeze_prefix(self, prefix)
 
-    def _with_prefix(self, prefix):
-        expr = deepcopy(self)
-        expr._freeze_prefix(prefix)
-        return expr
-
     @property
     def is_frozen(self):
         """Whether this expression's prefix is frozen."""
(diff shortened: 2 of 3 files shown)


sashankaryal:
`tests/unittests/aggregation_tests.py::DatasetTests::test_nan_inf` and `tests/unittests/aggregation_tests.py::DatasetTests::test_quantiles` are failing. In both cases, it has to with with `dataset.quantiles()` and looking at the call stack, it has to do with view expressions.

```
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/fiftyone/core/collections.py:7882: in quantiles
    return self._make_and_aggregate(make, field_or_expr)
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/fiftyone/core/collections.py:9844: in _make_and_aggregate
    return self.aggregate(make(args))
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/fiftyone/core/collections.py:9537: in aggregate
    _results = foo.aggregate(self._dataset._sample_collection, pipelines)
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/fiftyone/core/odm/database.py:347: in aggregate
    result = collection.aggregate(pipelines[0], allowDiskUse=True)

```

```
pymongo.errors.OperationFailure: PlanExecutor error during aggregation :: caused by :: TypeError: array is null :
@:1:20
, full error: {'ok': 0.0, 'errmsg': 'PlanExecutor error during aggregation :: caused by :: TypeError: array is null :\n@:1:20\n', 'code': 139, 'codeName': 'JSInterpreterFailure'}
```

Rings a bell?

brimoor:
> Rings a bell?

@sashankaryal ah, turns out my first solution breaks other ways that view expressions are used. I've updated the implementation apply a more localized fix. I also verified that this same fix is not needed for other generated views `to_patches()` and `to_frames()`

sashankaryal (approved):
Lgtm!!! Thank you.

PR #4498, opened 2024-06-14
fix ctx event race condition

ritch:
Ensure ctx panel events aren't triggered until the panel is loaded. This allows the panel implementation in python to set initial state without being overridden by ctx changes.


imanjra (approved):
🚀

Issue #4508, opened 2024-06-17
TorchImageModel Ragged batches cfg value 


dimidagd:
Problem:  Ragged_batches parameter should be loaded by cfg instead of hard coded logic.

I am passing my own transforms to the TorchImageModel. After failing to run batched predictions, I figured out that the `ragged_batches` value is hard coded to True when passing custom transforms to the constructor, instead of calling `_build_transform`.

https://github.com/voxel51/fiftyone/blob/ae4c865ce86fd154f9ab5164bb87c7c3ca38bddb/fiftyone/utils/torch.py#L792

The `ragged_batches` attribute is used to determine the collation function used in the pytorch dataloader.
https://github.com/voxel51/fiftyone/blob/ae4c865ce86fd154f9ab5164bb87c7c3ca38bddb/fiftyone/core/models.py#L716

In the end of the day, I can not run batched inference, f.x. with `compute_embeddings` , instead it runs serially with a warning that the `model does not support batching`
https://github.com/voxel51/fiftyone/blob/ae4c865ce86fd154f9ab5164bb87c7c3ca38bddb/fiftyone/core/models.py#L1808

PR #4509, opened 2024-06-17
Fix/ragged-batches

dimidagd:

Resolves  https://github.com/voxel51/fiftyone/issues/4508

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Improvements**
  - The transformation process in the app now uses configurable settings for handling ragged batches, making it more flexible for diverse user requirements.

### DIFF
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -342,6 +342,8 @@ class TorchImageModelConfig(foc.Config):
             # Use the `image_XXX` parameters defined below to build a transform
             transforms = build_transforms(image_XXX, ...)
 
+            # Set `transforms_args.ragged_batches` to `True` or `False` depending on the case
+
     3.  OutputProcessor::
 
             # Directly provide an OutputProcessor
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -789,7 +789,7 @@ class TorchImageModel(
         return model
 
     def _build_transforms(self, config):
-        ragged_batches = True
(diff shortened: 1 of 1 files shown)


brimoor (approved):
@dimidagd thanks for the PR! I see the issue and agree that users should have the ability to explicitly configure `ragged_batches` when using custom transforms.

I think it would be best to add a new (optional) `ragged_batches` parameter rather than pulling it from `transforms_args["ragged_batches"]`, as the latter is supposed to be usable in conjunction with `transforms_fcn` like this:

```py
transforms = transforms_fcn(**transforms_args)
```

I'll go ahead and merge this PR into a feature branch so you get credit for your contribution(!) and will follow up with a commit that migrates the new parameter were the above.

dimidagd:
> @dimidagd thanks for the PR! I see the issue and agree that users should have the ability to explicitly configure `ragged_batches` when using custom transforms.
> 
> I think it would be best to add a new (optional) `ragged_batches` parameter rather than pulling it from `transforms_args["ragged_batches"]`, as the latter is supposed to be usable in conjunction with `transforms_fcn` like this:
> 
> ```python
> transforms = transforms_fcn(**transforms_args)
> ```
> 
> I'll go ahead and merge this PR into a feature branch so you get credit for your contribution(!) and will follow up with a commit that migrates the new parameter per the above.

This requires changes in other parts of the codebase as I've understood.

brimoor:
> This requires changes in other parts of the codebase as I've understood.

This is all that's required: https://github.com/voxel51/fiftyone/pull/4512

PR #4477, opened 2024-06-11
Add support for annotating frame views

brimoor:
Adds support for passing frame views to `annotate()`.

No changes are necessary in the annotation backends themselves to support frame views. From the backend's standpoint it's just like any other image collection is being annotated; the extra bookkeeping is handled automatically via the core SDK 🎉 

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video")

frames = dataset.to_frames(fps=1, sample_frames=True)

frames.annotate(
    "test",
    label_field="detections",
    label_type="detections",
    classes=["vehicle"],
    launch_editor=True,
)

# Add, edit, and delete vehicle labels...

dataset.load_annotations("test")

session = fo.launch_app(dataset)
```

## Summary by CodeRabbit

- **New Features**
  - Introduced improvements to the annotation process for frames views, streamlining user interactions and data management.

- **Refactor**
  - Enhanced ID mapping and handling of frame fields.
  - Improved management of unexpected labels and annotation runs.
  - Updated methods for merging scalars and labels, as well as saving and updating ID maps.

- **Tests**
  - Added new test cases to validate the functionality of frame annotations, ensuring accurate upload, creation, and data integrity.

These changes significantly enhance the annotation functionality and user experience.

PR #4511, opened 2024-06-18
Add support for annotating clip views

brimoor:
Adds support for passing clip views to `annotate()`.

So far only the CVAT backend supports clip views.

Unlike [frame views](https://github.com/voxel51/fiftyone/pull/4477), there may be a minor amount of backend-specific work required to support clips rather than full videos. For example, in the case of CVAT, a touch of work was required when uploading the videos (as frames) and then when mapping CVAT frame IDs to the correct frame numbers during annotation download.

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Three clips into the same sample
clips = dataset.limit(1).to_clips([[(10, 20), (30, 40), (50, 60)]])

# Three clips into different samples
# clips = dataset.limit(3).to_clips([[(10, 20)], [(30, 40)], [(50, 60)]])

anno_key = "test"
clips.annotate(
    anno_key,
    backend="cvat",
    label_type="detections",
    classes=["vehicle"],
    project_name=anno_key,
    launch_editor=True,
)

# Add, edit, and delete vehicle labels...

```

### DIFF
--- a/fiftyone/utils/labelbox.py
+++ b/fiftyone/utils/labelbox.py
@@ -696,9 +696,7 @@ class LabelboxAnnotationAPI(foua.AnnotationAPI):
         labels_json = self._download_project_labels(project=project)
 
         dataset = results.samples._root_dataset
-        is_video = dataset._contains_videos(any_slice=True) and any(
-            dataset._is_frame_field(f) for f in label_schema.keys()
-        )
+        is_video = any(dataset._is_frame_field(f) for f in label_schema.keys())
 
         annotations = {}
 
--- a/fiftyone/utils/labelbox.py
+++ b/fiftyone/utils/labelbox.py
@@ -620,10 +620,13 @@ class LabelboxAnnotationAPI(foua.AnnotationAPI):
         project_name = config.project_name
         members = config.members
         classes_as_attrs = config.classes_as_attrs
-        is_video = (samples.media_type == fomm.VIDEO) or (
-            samples.media_type == fomm.GROUP
-            and samples.group_media_types[samples.group_slice] == fomm.VIDEO
-        )
+        is_video = samples.media_type == fomm.VIDEO
+
+        if samples._is_clips:
+            raise ValueError(
+                "The Labelbox backend does not yet support annotating clip "
+                "views"
+            )
 
         for label_field, label_info in label_schema.items():
             if label_info["existing_field"]:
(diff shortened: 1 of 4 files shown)


fiftyone/utils/labelbox.py
```diff
@@ -620,10 +620,13 @@ def upload_samples(self, samples, anno_key, backend):
         project_name = config.project_name
         members = config.members
         classes_as_attrs = config.classes_as_attrs
-        is_video = (samples.media_type == fomm.VIDEO) or (
```
brimoor, line 623: @ehofesmann Checking for grouped datasets at *upload time* is not necessary since the Labelbox backend's `supported_media_types` property already forces the user to pass in a `select_group_slices()` view when annotating video datasets.

This was acknowledged in https://github.com/voxel51/fiftyone/pull/4073 but the extra code was still added.

fiftyone/utils/labelbox.py
```diff
@@ -691,13 +694,10 @@ def download_annotations(self, results):
 
         project = self._client.get_project(project_id)
         labels_json = self._download_project_labels(project=project)
-        is_video = (results._samples.media_type == fomm.VIDEO) or (
```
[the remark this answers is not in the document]
brimoor: Note that `dataset.load_annotations()` works independent of the current `dataset.group_slice` for grouped datasets. It flattens the entire collection and loads data for any necessary samples:

https://github.com/voxel51/fiftyone/blob/d095b222ca15b13bae69a5bdace9a81a443e4ffd/fiftyone/utils/annotations.py#L1486-L1493

fiftyone/utils/labelbox.py
ehofesmann: Ah yup, you're right. I must have misunderstood what ben was saying, but this should just be checked when downloading samples

fiftyone/utils/labelbox.py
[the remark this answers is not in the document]
ehofesmann: Ah, good call, that makes sense

fiftyone/utils/labelbox.py
```diff
@@ -620,10 +620,13 @@ def upload_samples(self, samples, anno_key, backend):
...
-            samples.media_type == fomm.GROUP
-            and samples.group_media_types[samples.group_slice] == fomm.VIDEO
-        )
+        is_video = samples.media_type == fomm.VIDEO
+
+        if samples._is_clips:
+            raise ValueError(
+                "The Labelbox backend does not yet support annotating clip "
+                "views"
+            )
```
ehofesmann, line 629: What do you think about moving this into `annotations.py` and adding a `supports_clip_views` property to each annotation backend that defaults to `False` (With CVAT returning `True`). I'm just thinking about the darwin and other non builtin integrations which would be good to prevent from needing to be manually updated to prevent annotation of clip views.

brimoor: Great idea! Just added ✅

PR #4528, opened 2024-06-24
Python Panels

ritch:
Python Panels


requirements/common.txt
```diff
@@ -18,6 +18,8 @@ plotly==5.17.0
 pprintpp==0.4.0
 psutil>=5.7.0
 pymongo>=3.12
+pydantic==2.6.4
```
imanjra, line 21: Stuard had removed `pydantic` in the past. Is bringing back here intentional/needed?

imanjra, line 21: Stuart had removed `pydantic` in the past. Is bringing back here intentional/needed?

app/packages/core/src/plugins/SchemaIO/components/MediaPlayerView.tsx
```diff
@@ -0,0 +1,41 @@
...
+export default function MediaPlayerView(props) {
+  const { schema, data } = props;
+  const { default: defaultValue, playerProps = {} } = schema;
+  const actualData = data ?? defaultValue;
+  const mediaUrl = actualData?.url;
+
+  const handleEvent =
+    (event) =>
+    (...args) => {
+      console.log(`Video event: ${event}`, ...args);
```
imanjra, line 16: Looks like this needs to be wired up or removed

imanjra (approved):
We can continue to test and address comment as a follow-up. Good to check this in  early as it's a large PR. LGTM 🚀

benjaminpkane (changes requested):
Can we please have passing e2e before merging? 🙏

requirements/common.txt
benjaminpkane: Including `pydantic` has longterm implications. Would be good to discuss offline

benjaminpkane: Looks like it's not used anywhere
```suggestion
```

fiftyone/operators/types.py
```diff
@@ -1757,6 +1960,41 @@ def options(self):
...
+        Must be used with :class:`Object` properties.
+
+    Args:
+        orientation ("horizontal"): the orientation of the stack. Can be either
+            ``"horizontal"`` or ``"vertical"``
+        gap (1): the gap between the components
+        align_x ("left"): the alignment of the components. Can be either ``"left"``, ``"center"``,
+            or ``"right"``
+        align_y ("top"): the alignment of the components. Can be either ``"top"``, ``"center"``,
+            or ``"bottom"``
```
benjaminpkane, line 1978:
```suggestion
        align_x ("left"): the alignment of the components. Can be either
            ``"left"``, ``"center"``, or ``"right"``
        align_y ("top"): the alignment of the components. Can be either
            ``"top"``, ``"center"``, or ``"bottom"``
```

app/packages/components/package.json
```diff
@@ -46,6 +46,7 @@
         "@textea/json-viewer": "^3.4.1",
         "classnames": "^2.3.1",
         "framer-motion": "^6.2.7",
+        "material-icons": "^1.13.12",
```
sashankaryal, line 49: Why are we bringing it in when we already have `@mui/icons-material`? This will cause significant bloating otherwise.

imanjra: This is font-based and it's to support dynamic icon by name. In a separate PR, we should probably replace all the icons from `@mui/icons-material` to this instead. It will reduce build size/time

PR #4540, opened 2024-06-27
event path

ritch:
Add "path" param for plot and button events...

```python
class MyPanel(foo.Panel):
    def render():
        # ...
        panel.plot('plot1', on_click=self.on_click)
        panel.plot('plot2', on_click=self.on_click)
    
    def on_click(ctx):
        # which plot was clicked?
        print(ctx.params.get('path')) # "plot1" or "plot2"
```

## Summary by CodeRabbit

- **New Features**
  - Enhanced `ButtonView` and `PlotlyView` components to support a new `path` property in their props.
  - Improved logging and parameter handling in `PlotlyView` to include the `path` property.

### DIFF
--- a/app/packages/core/src/plugins/SchemaIO/components/PlotlyView.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/PlotlyView.tsx
@@ -9,7 +9,7 @@ import { HeaderView } from ".";
 import { getComponentProps } from "../utils";
 
 export default function PlotlyView(props) {
-  const { data, schema } = props;
+  const { data, schema, path } = props;
   const { view = {} } = schema;
   const { config = {}, layout = {} } = view;
   const theme = useTheme();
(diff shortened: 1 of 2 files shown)


app/packages/core/src/plugins/SchemaIO/components/PlotlyView.tsx
```diff
@@ -9,14 +9,15 @@ import { HeaderView } from ".";
...
-  const { data, schema } = props;
+  const { data, schema, path } = props;
   const { view = {} } = schema;
   const { config = {}, layout = {} } = view;
   const theme = useTheme();
   const panelId = usePanelId();
   let range = [0, 0];
   const triggerPanelEvent = usePanelEvent();
   const handleEvent = (event?: string) => (e) => {
+    console.log(props);
```
imanjra, line 20: ✂️

imanjra (approved):
LGTM

PR #4541, opened 2024-06-28
More robust handling of multiple config docs

brimoor:
Under heavy concurrent SDK usage, some users have reported seeing the following error when connecting to the database:

```
OSError: Cannot connect to database v0.14.4 with client v0.24.0 (compatibility >=0.19,<0.25). See https://docs.voxel51.com/user_guide/config.html#database-migrations for more information
```

For context, the `config` collection is supposed to have exactly one document. However, if it were empty for some reason, then this error would occur. How could it be empty? I suspect that, under heavy concurrent usage, this function:
https://github.com/voxel51/fiftyone/blob/a5d2209903d3699be3d310a55ab1e518f858050a/fiftyone/core/odm/database.py#L101

could be subject to a race condition that accidentally deletes all config docs.

This PR makes two improvements to protect against such a race condition:
1. When there are multiple config docs, keep the newest one rather than the one with the latest `version`. Also, for extra protection, only delete the duplicate docs if `database_admin=True`
    - This is more robust to race conditions because `_id` are universally unique while the duplicate config docs likely have the same `version`, so the `max()` logic used to select which doc to keep may not be deterministic, thus causing concurrent sessions to accidentally delete *all* the config docs
    - Note that sorting by `_id` is equivalent to sorting by insertion date
2. When generating a new database config doc, directly insert the current `foc.VERSION` rather than `0.14.4`
    - `0.14.4` was previously being inserted to correctly handle migrating <0.14.4 databases up to 0.14.4 (when the `config` collection was introduced). However, the `0.14.4` release is now over 2 years old, so let's assume that everyone has already migrated up. Why? Because...
    - ... the benefit of this change is that *even if* the `config` collection still somehow becomes empty, the new doc will assume the current client version and thus the compatibility error that we're trying to avoid will no longer occur

I don't know for certain that this race condition is actually the reason that users are seeing the compatibility error, but this PR doesn't have any side effects so we might as well give it a shot 🚀 

## Summary by CodeRabbit

- **New Features**
  - Improved handling of database configuration, including better management of multiple config documents and setting default values for version and type.
  
- **Bug Fixes**
  - Enhanced selection of the newest configuration document to ensure more reliable behavior.

- **Tests**
  - Updated unit tests to ensure robust handling of new database configuration logic.

### DIFF
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -114,6 +114,7 @@ def get_db_config():
         # this information is required in order to run migrations...
         #
         config.version = foc.VERSION
+        if db_config.type != foc.CLIENT_TYPE:
         save = True
 
     if config.type is None:
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -136,9 +136,6 @@ def get_db_config():
 
 
 def _handle_multiple_config_docs(conn, config_docs):
-    # Use the newest one
-    keep_doc = max(config_docs, key=lambda d: d["_id"])
-
     if fo.config.database_admin:
         logger.warning(
             "Unexpectedly found %d documents in the 'config' collection; "
@@ -152,7 +149,12 @@ def _handle_multiple_config_docs(conn, config_docs):
             [{"$sort": {"_id": -1}}, {"$limit": 1}, {"$out": "config"}]
         )
 
-    return keep_doc
+        config_doc = next(iter(conn.config.find()))
+    else:
+        # Use the newest one
+        config_doc = max(config_docs, key=lambda d: d["_id"])
+
+    return config_doc
 
 
 def establish_db_conn(config):
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -91,29 +91,29 @@ def get_db_config():
         a :class:`DatabaseConfigDocument`
     """
     conn = get_db_conn()
-    save = False
+
     config_docs = list(conn.config.find())
-    if not config_docs:
-        save = True
-        config = DatabaseConfigDocument(conn)
-    elif len(config_docs) > 1:
-        config = DatabaseConfigDocument(
-            conn, **cleanup_multiple_config_docs(conn, config_docs)
-        )
+    if config_docs:
+        if len(config_docs) > 1:
+            config_doc = _cleanup_multiple_config_docs(conn, config_docs)
+        else:
+            config_doc = config_docs[0]
+
+        config = DatabaseConfigDocument(conn, **config_doc)
+        save = False
     else:
-        config = DatabaseConfigDocument(conn, **config_docs[0])
+        config = DatabaseConfigDocument(conn)
+        save = True
 
     if config.version is None:
         #
-        # The database version was added to this config in v0.15.0, so if no
-        # version is available, assume the database is at the preceding
-        # release. It's okay if the database's version is actually older,
-        # because there are no significant admin migrations prior to v0.14.4.
+        # If the database has no version, then assume the version of the client
+        # that is currently connecting to it.
         #
         # This needs to be implemented here rather than in a migration because
         # this information is required in order to run migrations...
         #
-        config.version = "0.14.4"
+        config.version = foc.VERSION
         save = True
 
     if config.type is None:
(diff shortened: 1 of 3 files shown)


fiftyone/core/odm/database.py
```diff
@@ -91,29 +91,29 @@ def get_db_config():
...
-        # release. It's okay if the database's version is actually older,
-        # because there are no significant admin migrations prior to v0.14.4.
+        # If the database has no version, then assume the version of the client
+        # that is currently connecting to it.
         #
         # This needs to be implemented here rather than in a migration because
         # this information is required in order to run migrations...
         #
-        config.version = "0.14.4"
+        config.version = foc.VERSION
```
kaixi-wang, line 116: Is it possible that someone connects with a greater version and could cause the database version to upgrade and thus brick the connection for everyone else until they upgrade as well? Or require that the datasets be migrated before they can be accessed by everyone else?

brimoor: Yes what you're saying is true and could be bad. However, if the `config` collection is empty, then something has already gone terribly wrong. So the most important contribution of this PR is to (hopefully) remove the possibility that all `config` docs could ever get deleted in the first place.

Using `foc.VERSION` instead of `0.14.4` here has another benefit. Currently, when you connect to a brand new database for the very first time, you see:

```py
import fiftyone as fo; fo.list_datasets()
# Migrating database to v0.24.1
```

because the config version was previously getting set to `0.14.4` and then is immediately migrated up to `foc.VERSION`, which is a pointless migration because, by construction, the database is completely empty. Directly setting the version to `foc.VERSION` prevents the log message, which is cleaner.

kaixi-wang: I kinda mentioned it in the meeting but didn't want to waste everyone's time talking about implementation detail, but I was thinking that instead of automatically choosing the max, it would be safer to set the return_document to before on the update command and after the update check the value of the config version and if the before document's version is greater than the one used for the update command (0.14.4) immediately save the config again using the before docs values. Since the before doc is from the db, even if an error in the local client initially read the value as none, the returned doc will always reflect what was overwritten, so should prevent resetting the value accidentally without simultaneously assuming the connected client has the right version.

kaixi-wang: Alternatively it might just be better to throw an error and force the user to reissue the command as something is wrong with the connection at this point, putting into question whether the results of any subsequent queries in the chain can be accurate

brimoor: I'm not quite understanding your suggestion in the first comment.

Personally I'm still struggling to understand how the `config` collection could become empty in the first place. As explained in the PR description, I could see it happening *IF* the `config` collection contained multiple documents, but... I don't have an explanation for how it could come to contain >1 document, so the mystery still isn't fully solved.

We could start questioning low-level aspects of pymono, like supposing it fails to read document fields correctly and accidentally reads `version=None`, but if that could happen then every aspect of FO could fail, which isn't happening and isn't helpful to think about.

kaixi-wang: To be clear, I don't think there is ever more or less than one config saved to the db. I think it's the one config getting overridden.

brimoor: How do you think the one doc is getting overridden with `version=None`?

fiftyone/core/odm/database.py
```diff
@@ -135,26 +135,23 @@ def get_db_config():
     return config
 
 
-def cleanup_multiple_config_docs(conn, config_docs):
-    """Internal utility that ensures that there is only one
-    :class:`DatabaseConfigDocument` in the database.
-    """
+def _handle_multiple_config_docs(conn, config_docs):
+    # Use the newest one
+    keep_doc = max(config_docs, key=lambda d: d["_id"])
```
swheaton, line 140: I think we should not do this anymore. This function should accept `conn` only. Then re-query for the config document that survived at the end. Avoids another potential concurrency issue

swheaton (changes requested):
small change requested.
overall lgtm - I also am not sure the exact case where config collection can get messed up but I agree with the change here.

fiftyone/core/odm/database.py
brimoor: Updated 👍

swheaton (approved):
lgtm

PR #4545, opened 2024-07-01
Updated snippets for exporting dataset to use `load_dataset`

minhtuev:

1. Currently, snippets for exporting dataset use `fo.Dataset(...)`, which is incorrect because this is the syntax for creating a new dataset. To load an existing dataset, use `fo.load_dataset(...)` instead.

Example snippet:
```
import fiftyone as fo
dataset_or_view = fo.Dataset("oi-v7full-2mm")
export_dir = "/path/for/export"
label_field = "ground_truth"
```

Error:
```
File ~/workspace/fiftyone-teams/fiftyone/core/dataset.py:117, in _validate_dataset_name(name, skip)
    115 clashing_name = clashing_name_doc["name"]
    116 if clashing_name == name:
--> 117     raise ValueError(f"Dataset name '{name}' is not available")
    118 else:
    119     raise ValueError(
    120         f"Dataset name '{name}' is not available: slug '{slug}' "
    121         f"in use by dataset '{clashing_name}'"
    122     )

ValueError: Dataset name 'oi-v7full-2mm' is not available
```

Correct snippet should be:
```
import fiftyone as fo
dataset_or_view = fo.load_dataset("oi-v7full-2mm")
```

2. Add a parameter `create_if_necessary ` to `load_dataset` to return an empty dataset if none currently exists.

1. Documentation should render correctly.
2. Passing `create_if_necessary ` to `load_dataset` should create a new dataset if an existing one does not exist.

Testing:
* Manual testing (✅ )
* Unit tests (✅ )

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Improved error handling with the introduction of `DatasetNotFoundError` for missing datasets.
- **Bug Fixes**
  - Enhanced error messaging by replacing `ValueError` with `DatasetNotFoundError` when datasets are not found.
- **Tests**
  - Added new tests validating dataset loading, creation, re-loading, and exception handling scenarios.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -54,6 +54,13 @@ foud = fou.lazy_import("fiftyone.utils.data")
 logger = logging.getLogger(__name__)
 
 
+class MissingDatasetError(ValueError):
+    """Exception raised when a dataset is missing."""
+    def __init__(self, name):
+        self._dataset_name = name
+        super().__init__("Dataset '%s' not found" % name)
+
+
 def list_datasets(glob_patt=None, tags=None, info=False):
     """Lists the available FiftyOne datasets.
 
@@ -123,7 +130,7 @@ def _validate_dataset_name(name, skip=None):
     return slug
 
 
-def load_dataset(name):
+def load_dataset(name, create_if_missing=False):
     """Loads the FiftyOne dataset with the given name.
 
     To create a new dataset, use the :class:`Dataset` constructor.
@@ -136,11 +143,18 @@ def load_dataset(name):
 
     Args:
         name: the name of the dataset
+        create_if_missing (False): if no dataset exists, create empty one
 
     Returns:
         a :class:`Dataset`
     """
-    return Dataset(name, _create=False)
+    try:
+        return Dataset(name, _create=False)
+    except MissingDatasetError as ex:
+        if create_if_missing:
+            return Dataset(name)
+        else:
+            raise ex
 
 
 def get_default_dataset_name():
@@ -7591,7 +7605,7 @@ def _do_load_dataset(obj, name):
     db = foo.get_db_conn()
     res = db.datasets.find_one({"name": name})
     if not res:
-        raise ValueError("Dataset '%s' not found" % name)
+        raise MissingDatasetError(name)
     dataset_doc = foo.DatasetDocument.from_dict(res)
 
     sample_collection_name = dataset_doc.sample_collection_name
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -58,7 +58,7 @@ class MissingDatasetError(ValueError):
     """Exception raised when a dataset is missing."""
     def __init__(self, name):
         self._dataset_name = name
-        super().__init__("Dataset '%s' not found" % name)
+        super().__init__(f"Dataset {name} not found")
 
 
 def list_datasets(glob_patt=None, tags=None, info=False):
@@ -148,13 +148,12 @@ def load_dataset(name, create_if_missing=False):
     Returns:
         a :class:`Dataset`
     """
-    try:
+    if dataset_exists(name):
         return Dataset(name, _create=False)
-    except MissingDatasetError as ex:
-        if create_if_missing:
-            return Dataset(name)
-        else:
-            raise ex
+    elif create_if_missing:
+        return Dataset(name)
+    else:
+        raise MissingDatasetError(name)
 
 
 def get_default_dataset_name():
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -153,10 +153,10 @@ def load_dataset(name, create_if_necessary=False):
         a :class:`Dataset`
     """
     try:
-        return Dataset(name, _create=create_if_necessary)
+        return Dataset(name, _create=False)
     except DatasetNotFoundError as ex:
         if create_if_necessary:
-            return Dataset(name)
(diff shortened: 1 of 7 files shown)


minhtuev:
Thanks @manivoxel51 and @allenleetc for helping me to debug my export/import snippet 💯

swheaton:
Hmm I think this is partially just user learning curve. On the other side of the coin, if the dataset did not exist then calling `load_dataset()` would fail too.

The intention is that it's not a fully executable code snippet, but that `dataset_or_view` should be set to a new dataset, loaded dataset, or any view into a dataset. From there, the remaining snippet lines would work.

Perhaps there is another way to denote that we aren't dictating how to load the dataset or view in the snippet but the focus is on what happens afterwards.

minhtuev:
@swheaton : since the tutorial is about exporting dataset, I think it is safe to assume that the dataset exists. The same code snippet is also shared when clicking the export button on Fiftyone as well.

image

fiftyone/core/dataset.py
```diff
@@ -136,11 +143,18 @@ def load_dataset(name):
...
         a :class:`Dataset`
     """
-    return Dataset(name, _create=False)
+    try:
+        return Dataset(name, _create=False)
+    except MissingDatasetError as ex:
+        if create_if_missing:
+            return Dataset(name)
+        else:
+            raise ex
```
sashankaryal, line 157: I'd use `fo.dataset_exists()` over try-except.

sashankaryal (changes requested):

fiftyone/core/dataset.py
minhtuev: Good idea, done!

fiftyone/core/dataset.py
```diff
@@ -123,7 +130,7 @@ def _validate_dataset_name(name, skip=None):
     return slug
 
 
-def load_dataset(name):
+def load_dataset(name, create_if_missing=False):
```
brimoor, line 133: @minhtuev I'm onboard with this new parameter 👍 

Can we call it `create_if_necessary` for consistency with the nomenclature used in other places, such as `load_zoo_dataset(name, download_if_necessary=True)`? 🤓

fiftyone/core/dataset.py
```diff
@@ -54,6 +54,13 @@
 logger = logging.getLogger(__name__)
 
 
+class MissingDatasetError(ValueError):
```
brimoor, line 57: nit: suggest `DatasetNotFoundError` for consistency with how the error message is worded

"Missing" implies that I need to go looking for it 😉

fiftyone/core/dataset.py
minhtuev: Done!

fiftyone/core/dataset.py
minhtuev: Good idea!

fiftyone/core/dataset.py
swheaton: I disagree with @sashankaryal 's comment. While it might look nicer, this causes an extra database lookup for the happy path.
We should assume it exists and fall back to creating rather than checking for existence.

swheaton:
@minhtuev so that actually is coming from the Teams app which is in `voxel-hub` repo.
The snippet you're seeing is specifically in `voxel-hub/app/packages/app/pages/datasets/\[slug\]/components/ExportView/CodeExport.tsx` (backslashes are mine to make it work in unix)

fiftyone/core/dataset.py
minhtuev: Hmm, how much do we want to optimize database lookup? I don't think the users will be calling `load_dataset` very often, probably once at the beginning of the script.

swheaton: you might be surprised. especially when going through pymongo proxy (dont wanna get into it lol). where each call to mongodb must be absolutely necessary haha.

minhtuev: I see, @sashankaryal what do you think?

minhtuev: Chatted with @sashankaryal, let's go with @swheaton's suggestion

fiftyone/core/dataset.py
```diff
@@ -136,11 +143,22 @@ def load_dataset(name):
...
+    Raises:
+        DatasetNotFoundError: if the dataset does not exist and
+            `create_if_necessary` is False
 
     Returns:
         a :class:`Dataset`
     """
-    return Dataset(name, _create=False)
+    try:
+        return Dataset(name, _create=create_if_necessary)
```
swheaton, line 156: can't do this, we need to try and load it first then create, otherwise we'll get an error.
```suggestion
        return Dataset(name, _create=False)
```

fiftyone/core/dataset.py
```diff
@@ -136,11 +143,22 @@ def load_dataset(name):
...
+    except DatasetNotFoundError as ex:
+        if create_if_necessary:
+            return Dataset(name)
```
swheaton, line 159: Suggest being explicit with the `_create` argument for clarity's sake.
```suggestion
            return Dataset(name, _create=True)
```

swheaton (changes requested):

fiftyone/core/dataset.py
minhtuev: Sure

sashankaryal (approved):
New `load_dataset` LGTM!!

swheaton (approved):
great thanks

PR #4551, opened 2024-07-04
order imports and modify style guide

findtopher:

It turns out that we weren't actually following our style guide, and it would be very difficult to do so without an automated too.

There _is_ however a section of PEP8 that defines that there are _three_ sections of imports, which works nicely because that's also what the automated formatters want to use.

This actually alphabetizes our imports and groups them according to the PEP8 rules.  It does _not_ rearrange our 'not at the top' imports that happen after other testing occurs.

I tried `isort` and I tried `ruff` and Ruff was both much faster and didn't do strange things to parenthetical imports, so I went with it.  We should look at using ruff to handle formatting (it takes all the black arguments) and linting, but that's a problem for tomorrow.

This also adds Ruff as a pre-commit hook to keep things orderly going forward.

```
make python
pip install dist/*.whl
python -c 'import fiftyone as fo'
```

I assume our unit tests will catch any other issues, though I'm not sure how we'd have any...

        notes for FiftyOne users.

-   [x] Other


STYLE_GUIDE.md
```diff
@@ -78,15 +78,14 @@ generic:
...
--   Voxel51-authored non-FiftyOne package dependencies
 -   FiftyOne modules (least generic)
 
 For core FiftyOne imports, we import modules as `fox`, where `x` is the first
 letter of the module imported. If necessary, we use `foxy` to disambiguate
 between two modules that start with the same letter.
 
 Within each import group, imports should be sorted alphabetically by full
-package path, ignoring `from` and `import`.
+package path, with direct `import`s first and `from` imports after.
```
findtopher, line 88: because this is how all the automated tools handle it, and we weren't really doing a good job of following this guidance anyway

swheaton: [configurable](https://pycqa.github.io/isort/docs/configuration/options.html#force-sort-within-sections)

swheaton:
It places `eta` within the third party modules which is technically correct but confusing / a bit disingenuous.
We can make a [custom section](https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html) for `eta` before `FIRSTPARTY` which would make it follow the style guideline as written.

For example I think we can do this?

```python
known_eta=eta
sections=FUTURE,STDLIB,THIRDPARTY,ETA,FIRSTPARTY,LOCALFOLDER
```

pyproject.toml
```diff
@@ -8,3 +8,11 @@ exclude = '''
   | \.git
 )/
 '''
+
+[tool.ruff]
+line-length = 79
+
+[tool.ruff.lint.isort]
+lines-after-imports = 2
+force-sort-within-sections = false
```
swheaton, line 17: @findtopher did you mean to make this `true`? Since you resolved my comment without disputing it.
[ignore direct vs from imports](https://github.com/voxel51/fiftyone/pull/4551/commits/b1a41ba300832130938e42edde9557168085b036)

swheaton:
Btw thanks @findtopher for contributing to code quality 💪🏼 
You must've been super bored or on a style guideline kick 😜 
If `ruff` works out better we should add it to our other repos which some use black and isort.

findtopher:
You know, if I can get the tests compass 🤪

findtopher:
Maybe I'll try again some day when I have time to debug the import errors 🤷

PR #4561, opened 2024-07-10
Ensure db connection for unpickled Dataset

swheaton:

Actual customer issue was due to using a parallel processing engine that basically created a `Dataset` object by deserializing, thereby bypassing any calls that make the DB connection in `Dataset.__init__()`.
So just like how we call `ensure_connection()` before `Document._save()` it should also be fine to call it before any `reload()` of a `Document`. If we already have a connection this is a no-op.

Easiest way to show the issue. 
```
import bson

import fiftyone as fo

doc = fo.core.odm.DatasetDocument(id=bson.ObjectId())
doc.reload()  # ConnectionFailure before, no error now
```

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of the reload functionality by ensuring a database connection is established before reloading documents.

### DIFF
--- a/fiftyone/core/odm/document.py
+++ b/fiftyone/core/odm/document.py
@@ -589,6 +589,7 @@ class Document(BaseDocument, mongoengine.Document):
         Args:
             *fields: an optional args list of specific fields to reload
         """
+        ensure_connection()
         super().reload(*fields, **kwargs)
 
     def save(


brimoor (approved):
LGTM

PR #4575, opened 2024-07-16
Add test for frame view annotation

ehofesmann:
Adds a test to the CVAT intensive tests for annotation of frame views implemented in https://github.com/voxel51/fiftyone/pull/4477


brimoor (approved):
LGTM, thank you! 🙌

PR #4579, opened 2024-07-17
Windows fo3d bug fix for finding asset files

CamronStaley:

Fixing error which was occurring on Windows machines when reading asset paths for fo3d.
Animation

Manual testing on windows computer and on mac laptop (to ensure it didn't break anything).

Used the dataset on local machine in the example and also used quickstart-3D.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of file paths with backslashes in `getFo3dRoot` function.
  - Adjusted `LogContainer` component to prevent unwanted scrolling behavior.

- **Tests**
  - Added test case to verify `getFo3dRoot` function with Windows-style file paths.

### DIFF
--- a/app/packages/looker-3d/src/fo3d/utils.ts
+++ b/app/packages/looker-3d/src/fo3d/utils.ts
@@ -125,7 +125,7 @@ export const getMediaPathForFo3dSample = (
 
 export const getFo3dRoot = (fo3dPath: string) => {
   // remove filename and the last slash to get the root
-  const root = fo3dPath.replace(/\/[^/]*\.fo3d$/, "/");
+  const root = fo3dPath.replace(/(\/[^/]*\.fo3d$|\\[^\\]*\.fo3d$)/, "/");
 
   return root;
 };
--- a/app/packages/looker-3d/src/Logs.tsx
+++ b/app/packages/looker-3d/src/Logs.tsx
@@ -13,7 +13,7 @@ const LogContainer = styled.div`
   margin-left: 0.25em;
   display: flex;
   align-content: center;
-  overflow: scroll;
+  overflow: hidden;
   align-items: center;
 `;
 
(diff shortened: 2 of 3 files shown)


app/packages/looker-3d/src/fo3d/utils.ts
```diff
@@ -125,7 +125,7 @@ export const getMediaPathForFo3dSample = (
 
 export const getFo3dRoot = (fo3dPath: string) => {
```
benjaminpkane, line 126: Nice!! This function is missing TLC without a unit tests I think 😄 

A good example of App unit tests can be found [here](https://github.com/voxel51/fiftyone/blob/develop/app/packages/looker/src/overlays/classifications.test.ts). Tests can be colocated, e.g. `app/packages/looker-3d/src/fo3d/utils.test.ts`

App unit tests can be run in `./app` with `yarn test`

swheaton:
Awesome 👏 congrats on your first 🎉 

Delta Looks good to me but I wonder if there is a better filepath utility in typescript?

Also should the replacement string of "/" be "\" for windows or does it not matter?

app/packages/looker-3d/src/fo3d/utils.ts
sashankaryal: This one has unit tests, but could use more coverage:

https://github.com/voxel51/fiftyone/blob/develop/app/packages/looker-3d/src/fo3d/tests/get-fo3d-root.test.ts

CamronStaley:
> Awesome 👏 congrats on your first 🎉
> 
> Delta Looks good to me but I wonder if there is a better filepath utility in typescript?
> 
> Also should the replacement string of "/" be "" for windows or does it not matter?

CamronStaley:
> Awesome 👏 congrats on your first 🎉
> 
> Delta Looks good to me but I wonder if there is a better filepath utility in typescript?
> 
> Also should the replacement string of "/" be "" for windows or does it not matter?

This doesn't seem to matter for windows since you can search files with either forwards or backwards. I tried both locally by updating the replace function and it seems to make no difference regardless it's able to find the files.

app/packages/looker-3d/src/fo3d/utils.ts
CamronStaley: I can add a couple tests with backslashes as well

PR #4586, opened 2024-07-19
Update fiftyone-teams doc on roles and permissions

lanzhenw:

Update fiftyone-teams doc: 
- add "can tag" permission descriptions and update related details
- update dataset "default access level" to "implicit access level" to for consistency.

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Documentation**
	- Clarified roles and permissions for members, collaborators, and guests in dataset management.
	- Expanded collaborator permissions to include tagging capabilities.
	- Updated default access levels to include tagging options for various user roles.
	- Enhanced guidance on user roles and access conditions for improved understanding of dataset access management.


docs/source/images/teams/admin_roles_page.png
brimoor, line 1: Can you update the description of `Dataset Access Level` in the App as follows? (and then update this screenshot)

```
**Dataset Access Level**
This determines the permission source(s) by which users may be granted access to datasets.

**EXPLICIT**: these users have access any datasets to which they (or a group to which they belong) have been granted [specific access](https://docs.voxel51.com/teams/roles_and_permissions.html#teams-specific-access)
**IMPLICIT**: these users can be granted explicit access to datasets. In addition, they receive the dataset's [default access](https://docs.voxel51.com/teams/roles_and_permissions.html#default-access), if one is configured
**ALL**: these users have Can Manage permissions on all datasets
```

docs/source/teams/roles_and_permissions.rst
```diff
@@ -166,22 +171,23 @@ following sources:
...
 
 .. image:: /images/teams/dataset_default_access.png
-   :alt: default-access
+   :alt: implicit-access
    :align: center
 
 .. note::
 
-   Default access level only applies to Members. Guests and Collaborators must
+   Implicit access level only applies to Members. Guests and Collaborators must
```
brimoor, line 190: Can we please change the `People and groups with access` heading back to `Specific access` below?

Here I'm proposing to revert a change that was made when user groups were introduce so that the table of contents matches the jargon in the App.

brimoor (changes requested):
In my opinion, we should revert the edits here that convert "default" to "implicit`. I believe the App itself still uses the "Default access level" terminology on the `Manage` tab, for example.

Instead, I suggest we just update the Roles page as I've proposed to make it more clear what "IMPLICIT" and "EXPLICIT" mean to admins who encounter these terms in the App.

In other words, think of "default" and "specific" access as end-user-facing concepts, and think of "implicit" and "explicit" as admin-only jargon that's explained in the App and links to the user-facing concepts available in the public docs.

lanzhenw:
Yes, that would make a lot more sense.

brimoor (approved):
LGTM

PR #4590, opened 2024-07-23
Add ability to dynamically alter 3D labels line width

CamronStaley:

https://voxel51.atlassian.net/browse/TEAMS-3117

Adding sliders in 3d datasets that allow us to increase / decrease the polylines width (both lines and cuboids)

These changes required us to move away from the default implementation of ["Lines" offered by three.js](https://threejs.org/docs/#api/en/materials/LineBasicMaterial.linewidth) because if you look in their notes line width changes will not be reflected in browsers running openGL due to a limitation (this applies to most all browsers). So instead we are using "Lines2" now which uses polygons to allow for lines to be resized.

Launched local server using dataset quickstart-3d:
```
yarn dev
```

Added polylines to the first object in quickstart-3d so that I could test those using this script:

```
import fiftyone as fo

dataset = fo.load_dataset("quickstart-3d")
dataset.persistent = True

sample = dataset.first()

print(sample)

# Object label
label = "lane"

# A list of lists of `[x, y, z]` points in scene coordinates describing
# the vertices of each shape in the polyline
points3d = [[[-5, -99, -2], [-8, 99, -2]], [[4, -99, -2], [1, 99, -2]]]

# A set of semantically related 3D polylines
polyline = fo.Polyline(label=label, points3d=points3d)

sample["polylines"] = fo.Polylines(polylines=[polyline]);
sample.save()

print(sample)
```

        notes for FiftyOne users.

When working with 3D datasets you can now change polyline width within the dropdown under labels.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced rendering capabilities for the Cuboid and Line components, improving visual fidelity with advanced line techniques.
	- Introduced dynamic line width controls for cuboids and polylines using state management.
	- Added new constant for panel ordering to improve component organization.
	- Expanded functionality of the modal component by integrating new interactions and verification methods.

- **Bug Fixes**
	- Resolved issues related to line rendering and performance.

- **Documentation**
	- Updated configurations and usage guidelines for the new line rendering and state management features.


swheaton:
whoa code rabbit is making seq diagrams now lol

app/packages/looker-3d/src/labels/cuboid.tsx
```diff
@@ -49,11 +70,25 @@ export const Cuboid = ({
...
-
   useCursor(isCuboidHovered);
 
-  if (!location || !dimensions) return null;
+  const geometry = useMemo(() => {
+    return new LineSegmentsGeometry().fromLineSegments(
+      new THREE.LineSegments(new THREE.EdgesGeometry(geo))
+    );
+  }, [geo]);
+  const material = useMemo(
```
sashankaryal, line 80: nit: newline here

app/packages/looker-3d/src/labels/index.tsx
```diff
@@ -34,10 +47,49 @@ export const ThreeDLabels = ({ sampleMap }: ThreeDLabelsProps) => {
...
+    cuboidLabelLineWidthAtom
+  );
+  const [polylineWidth, setPolylineWidth] = useRecoilState(
+    polylineLabelLineWidthAtom
+  );
   const selectedLabels = useRecoilValue(fos.selectedLabelMap);
   const tooltip = fos.useTooltip();
   const labelAlpha = colorScheme.opacity;
 
+  const constLabelLevaContrals = {
```
sashankaryal, line 60: nit: typo. `Contrals` -> `Controls`

app/packages/looker-3d/src/labels/line.tsx
```diff
@@ -1,6 +1,23 @@
...
+extend({ Line2, LineMaterial, LineGeometry });
+
+declare global {
+  namespace JSX {
+    interface IntrinsicElements {
+      line2: ReactThreeFiber.Node<Line2, typeof Line2>;
+      lineGeometry: ReactThreeFiber.Node<LineGeometry, typeof LineGeometry>;
+    }
+  }
+}
```
sashankaryal, line 20: can we put these declaration merging / global augmentation code in a separate file? maybe `three-d-labels.d.ts`, or even better `index.d.ts` in looker-3d package root?

app/packages/looker-3d/src/labels/line.tsx
```diff
@@ -29,10 +47,20 @@ export const Line = ({
     return tooltip.getMeshProps(label);
   }, [tooltip, label]);
 
+  const geometry = React.useMemo(() => {
+    return new LineGeometry().fromLine(new THREE.Line(geo));
+  }, [geo]);
+  const material = React.useMemo(() => {
```
sashankaryal, line 53: nit: linebreak

sashankaryal (commented):
Looking good. Not the most trivial of things, but a good opportunity (both in terms of learning + adding coverage for 3D labels) to add e2e tests.

swheaton:
@CamronStaley great work tracking down what turned into a bigger task than initially expected!

Couple feature questions before looking at the code:
1. 
> So instead we are using "Lines2" now which uses polygons to allow for lines to be resized.

Will this impact performance or number of lines we can support? I'm imagining number of points quadrupling - 2 points in a line to 8 points for a long cuboid.

2.
Where does the line width setting live? Is it for the current scene only or the whole session? Stored in the fo3d file?
Users that care about line width likely will want to set it once somewhere and have it persist.

CamronStaley:
@swheaton 
> Will this impact performance or number of lines we can support? I'm imagining number of points quadrupling - 2 points in a line to 8 points for a long cuboid.

I think this could impact performance in cases where there are tons of lines / cuboids, but I can't say for certain how much it will hurt it. Since it's [an extension of mesh](https://github.com/mrdoob/three.js/blob/master/examples/jsm/lines/LineSegments2.js) I would assume it's similar efficiency to that, which was already used for cuboids. 

One potential solution could be if we are seeing performance issues with >X amount of lines we could disable the option to increase line thickness and default back to the old implementation to save performance in those cases.

> Where does the line width setting live? Is it for the current scene only or the whole session? Stored in the fo3d file? Users that care about line width likely will want to set it once somewhere and have it persist.

Currently it's only stored in state so it won't persist between sessions, but it will persist between samples. We can discuss if we think it should persist between sessions in some way since I agree users who want it to always remain at X size should have that option. 

This could be a bigger discussion about settings as a whole since there are other settings which could probably live in the user's settings such as lighting for 3d scenes.

swheaton:
> I think this could impact performance in cases where there are tons of lines / cuboids, but I can't say for certain how much it will hurt it. Since it's [an extension of mesh](https://github.com/mrdoob/three.js/blob/master/examples/jsm/lines/LineSegments2.js) I would assume it's similar efficiency to that, which was already used for cuboids.

I'm not sure exactly how to measure performance of a web drawing tool but you could try this? Add a zillion lines to one of the scenes until the approach in this PR slows down. Then try same scene with the old lines approach and note any differences.

> This could be a bigger discussion about settings as a whole since there are other settings which could probably live in the user's settings such as lighting for 3d scenes.

lighting settings can be persisted in the FO3D file

sashankaryal:
We can persist line width settings in the browser storage for now. Add `getBrowserStorageEffectForKey` effect to the atom and you should be all set.

app/packages/looker-3d/src/state.ts
```diff
@@ -127,3 +127,15 @@ export const activeNodeAtom = atom<FoSceneNode>({
   key: "fo3d-activeNode",
   default: null,
 });
+
+export const cuboidLabelLineWidthAtom = atom<number>({
```
benjaminpkane, line 131: Type is implicit via the default value :)
```suggestion
export const cuboidLabelLineWidthAtom = atom({
```

app/packages/looker-3d/src/state.ts
```diff
@@ -127,3 +127,15 @@ export const activeNodeAtom = atom<FoSceneNode>({
...
+  key: "fo3d-cuboidLabelLineWidth",
+  default: 3,
+  effects: [getBrowserStorageEffectForKey("fo3d-cuboidLabelLineWidth")],
```
benjaminpkane, line 134: I think `valueClass` needs to be added when not a string
```suggestion
  effects: [getBrowserStorageEffectForKey("fo3d-cuboidLabelLineWidth", { valueClass: "number" })],
```

app/packages/looker-3d/src/state.ts
```diff
@@ -127,3 +127,15 @@ export const activeNodeAtom = atom<FoSceneNode>({
...
+});
+
+export const polylineLabelLineWidthAtom = atom<number>({
+  key: "fo3d-polylineLabelLineWidth",
+  default: 3,
+  effects: [getBrowserStorageEffectForKey("fo3d-polylineLabelLineWidth")],
```
benjaminpkane, line 140: Same as above :)

e2e-pw/src/oss/poms/modal/leva.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.locator = page.locator("#fo-leva-container");
+    this.assert = new LevaAsserter(this);
+  }
+}
+
+class LevaAsserter {
+  constructor(private readonly modalLevaPom: ModalLevaPom) {}
+
+  async verifyDefaultFolders() {
+    const defaultFolderNames = ["Visibility", "Labels", "Lights"];
```
benjaminpkane, line 20: Minor suggestion, but I find it useful to constants as define as a module global  after import, e.g. 
```typescript
const DEFAULT_FOLDER_NAMES = ["Visibility", "Labels", "Lights"]
```

app/packages/looker-3d/src/labels/cuboid.tsx
```diff
@@ -22,6 +30,7 @@ export const Cuboid = ({
   color,
   useLegacyCoordinates,
 }: CuboidProps) => {
+  const [lineWidth] = useRecoilState(cuboidLabelLineWidthAtom);
```
benjaminpkane, line 33: nit: `useRecoilValue` can be used when read only
```suggestion
  const lineWidth = useRecoilValue(cuboidLabelLineWidthAtom);
```

benjaminpkane (commented):
Awesome addition! Looks like a tricky configuration 🙇 

Just leaving a few linting comments

e2e-pw/src/oss/poms/modal/leva.ts
```diff
@@ -0,0 +1,36 @@
+import { Locator, Page, expect } from "src/oss/fixtures";
+import { ModalPom } from ".";
+
+export class ModalLevaPom {
+  readonly page: Page;
+  readonly locator: Locator;
+  readonly assert: LevaAsserter;
+
+  constructor(page: Page, private readonly modal: ModalPom) {
```
benjaminpkane, line 9: The `Page` is only used by the constructor for creating a `Locator`. Maybe a locator of the `ModalPom` can be used to find `"#fo-leva-container"`? Also, the `dataCy` plus `getByTestId` pattern is preferred over an HTML `#id` e.g. [here](https://github.com/voxel51/fiftyone/blob/59e6e6e1973edb7da246018e111c44267dfe9591/e2e-pw/src/oss/poms/selector.ts#L17)

@sashankaryal is the e2e guru, but this has been my observation :)

e2e-pw/src/oss/specs/3d/fo3d-pcd-stl.spec.ts
```diff
@@ -79,9 +89,34 @@ test.describe("fo3d", () => {
...
+      {
+        mask,
+        animations: "allow",
+      }
+    );
+
+    await modal.leva.maxSlider("Polyline Line Width");
+    await modal.leva.maxSlider("Cuboid Line Width");
+    await expect(modal.modalContainer).toHaveScreenshot(
+      "max-line-width-scene.png",
```
sashankaryal, line 115: beautiful

sashankaryal (approved):
LGTM, love it!
Feel free to address the remaining comments or not, to your discretion.

e2e-pw/src/oss/poms/modal/leva.ts
```diff
@@ -0,0 +1,76 @@
...
+        )
+      )
+    );
+  }
+
+  async verifyAssetFolders(assetNames: string[]) {
+    await Promise.all(
+      assetNames.map((assetName) =>
+        expect(this.modalLevaPom.getFolder(assetName)).toHaveCount(2)
+      )
```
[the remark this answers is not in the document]
CamronStaley: this is because the leva component has two divs containing that assets name (the asset which was added to the scene. One is the visibility dropdown and the other is the dropdown named after that assets allowing you to edit it.

e2e-pw/src/oss/poms/modal/leva.ts
```diff
@@ -0,0 +1,76 @@
...
+  constructor(page: Page, private readonly modal: ModalPom) {
+    this.page = page;
+    this.locator = page.locator("#fo-leva-container");
+    this.assert = new LevaAsserter(this);
+  }
+
+  getFolder(folderName: string) {
+    return this.page.locator('div[style*="--leva-colors-folderWidgetColor"]', {
+      has: this.page.locator(`text="${folderName}"`),
+    });
```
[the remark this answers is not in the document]
CamronStaley: yeah it would just hang for whatever reason I tried a few different combinations and could only get this one working properly.

app/packages/looker-3d/src/labels/line.tsx
```diff
@@ -1,6 +1,14 @@
...
 import * as THREE from "three";
 import { OverlayProps } from "./shared";
+import { useRecoilValue } from "recoil";
+import { polylineLabelLineWidthAtom } from "../state";
+import { Line2 } from "three/examples/jsm/lines/Line2";
+import { LineMaterial } from "three/examples/jsm/lines/LineMaterial";
+import { LineGeometry } from "three/examples/jsm/lines/LineGeometry";
+import { extend } from "@react-three/fiber";
+
+extend({ Line2, LineMaterial, LineGeometry });
```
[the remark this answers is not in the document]
CamronStaley: Yeah Line2 requires type of "LineMaterial" as its material attribute

PR #4592, opened 2024-07-24
upgrade react-use version

sashankaryal:
This is a minor version upgrade that gets rid of the problematic `fast-loops` dependency which has a known vulnerability.


benjaminpkane (approved):
🦺

PR #4596, opened 2024-07-25
Checking for null fields in dataset schema

minhtuev:

Some customers reported that if a field in `sample_fields` list is None, it causes Dataset not to load. We can filter out `None` value field to address this issue.

* Manual test
* Unit test

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced a new filtering method to enhance data integrity by removing `None` values from lists and dictionaries during document serialization.
	- Added a method to convert MongoDB data into Python objects while filtering out `None` values.

- **Bug Fixes**
	- Improved logging capabilities to issue warnings when `None` values are detected and ignored during data filtering.

### DIFF
--- a/fiftyone/core/odm/document.py
+++ b/fiftyone/core/odm/document.py
@@ -8,6 +8,7 @@ Base classes for documents that back dataset contents.
 
 from copy import deepcopy
 import json
+import logging
 
 from bson import json_util, ObjectId
 import mongoengine
@@ -20,6 +21,8 @@ import fiftyone.core.utils as fou
 from .database import ensure_connection
 from .utils import serialize_value, deserialize_value
 
+
+logger = logging.getLogger(__name__)
 
 class SerializableDocument(object):
     """Mixin for documents that can be serialized in BSON or JSON format."""
@@ -439,7 +442,12 @@ class MongoEngineBaseDocument(SerializableDocument):
                 d[k] = cls._simple_filter(v)
         elif isinstance(d, list):
             # Pymongo cannot create a document with a list that contains None
-            d = [v for v in d if v is not None]
+            r = [cls._simple_filter(v) for v in d if v is not None]
+            if len(r) != len(d):
+                msg = (f"Detected {len(d)-len(r)} None values in list,"
+                       f" ignoring them.")
+                logger.warning(msg)
+            d = r
         return d
 
     @classmethod
--- a/fiftyone/core/odm/document.py
+++ b/fiftyone/core/odm/document.py
@@ -433,6 +433,16 @@ class MongoEngineBaseDocument(SerializableDocument):
         return json.loads(json_util.dumps(d))
 
     @classmethod
+    def _simple_filter(cls, d):
+        if isinstance(d, dict):
+            for k, v in d.items():
+                d[k] = cls._simple_filter(v)
+        elif isinstance(d, list):
+            # Pymongo cannot create a document with a list that contains None
+            d = [v for v in d if v is not None]
+        return d
+
+    @classmethod
     def from_dict(cls, d, extended=False):
         if not extended:
             try:
@@ -440,6 +450,7 @@ class MongoEngineBaseDocument(SerializableDocument):
                 # extended form
 
                 # pylint: disable=no-member
+                d = cls._simple_filter(d)
                 return cls._from_son(d)
             except Exception:
                 pass
@@ -447,7 +458,7 @@ class MongoEngineBaseDocument(SerializableDocument):
         # Construct any necessary extended JSON components like ObjectIds
         # @todo can we optimize this?
         d = json_util.loads(json_util.dumps(d))
-
+        d = cls._simple_filter(d)
         # pylint: disable=no-member
         return cls._from_son(d)
 
--- a/fiftyone/core/odm/document.py
+++ b/fiftyone/core/odm/document.py
@@ -8,7 +8,6 @@ Base classes for documents that back dataset contents.
 
 from copy import deepcopy
 import json
-import logging
 
 from bson import json_util, ObjectId
 import mongoengine
@@ -21,8 +20,6 @@ import fiftyone.core.utils as fou
 from .database import ensure_connection
 from .utils import serialize_value, deserialize_value
 
-
-logger = logging.getLogger(__name__)
 
 class SerializableDocument(object):
     """Mixin for documents that can be serialized in BSON or JSON format."""
@@ -436,21 +433,6 @@ class MongoEngineBaseDocument(SerializableDocument):
         return json.loads(json_util.dumps(d))
 
     @classmethod
-    def _simple_filter(cls, d):
-        if isinstance(d, dict):
-            for k, v in d.items():
-                d[k] = cls._simple_filter(v)
-        elif isinstance(d, list):
-            # Pymongo cannot create a document with a list that contains None
-            r = [cls._simple_filter(v) for v in d if v is not None]
-            if len(r) != len(d):
-                msg = (f"Detected {len(d)-len(r)} None values in list,"
-                       f" ignoring them.")
-                logger.warning(msg)
-            d = r
-        return d
-
-    @classmethod
     def from_dict(cls, d, extended=False):
         if not extended:
             try:
@@ -458,7 +440,6 @@ class MongoEngineBaseDocument(SerializableDocument):
                 # extended form
 
                 # pylint: disable=no-member
-                d = cls._simple_filter(d)
                 return cls._from_son(d)
             except Exception:
                 pass
@@ -466,7 +447,7 @@ class MongoEngineBaseDocument(SerializableDocument):
         # Construct any necessary extended JSON components like ObjectIds
         # @todo can we optimize this?
         d = json_util.loads(json_util.dumps(d))
-        d = cls._simple_filter(d)
+
         # pylint: disable=no-member
         return cls._from_son(d)
 
--- a/fiftyone/core/fields.py
+++ b/fiftyone/core/fields.py
@@ -7,6 +7,7 @@ Dataset sample fields.
 """
 from copy import deepcopy
 from datetime import date, datetime
+import logging
 import numbers
 import re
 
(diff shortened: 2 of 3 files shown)


benjaminpkane:
Nice! Is it possible to add a test that asserts the fix? The change looks correct to me, just a sanity check.

brimoor (changes requested):
@minhtuev @benjaminpkane I don't think this is quite the error that users have reported. I believe they are seeing cases where `dataset._doc.sample_fields[i] is None`.

Here we're attempting to gracefully continue in cases where `dataset._doc.samples_fields[i].fields[j] is None`.

I think the first step here is to manage to reproduce how the corruption is occurring in the first place. For example, updating this PR to gracefully omit `dataset._doc.sample_fields[i]` if None would not actually do much good. It would allow `load_dataset()` to work, but then I bet the user would immediately get a "found unknown field 'foo'" error upon calling `sample = dataset.first()` because whatever the `None` value was we omitted from the schema was probably *supposed* to be defining an actual field.

Note that last year we made a fix in this area: 
https://github.com/voxel51/fiftyone/blob/53c446629ad0a9e50187425969fb39c341ebe436/fiftyone/core/odm/mixins.py#L251-L253

As mentioned in the comments, this was done to prevent issues that could arise when concurrently modifying a dataset's schema in multiple processes: https://github.com/voxel51/fiftyone/issues/3185

My bet is that those `_reload_fields()` we added are not fully solving the issue.

Either way, I think we must find proactive fixes here, not reactive ones.

minhtuev:
@brimoor : oh nested fields, very interesting... @ehofesmann reported in the ticket that the customers circumvented this issue by going into the DB and deleted null items under `sample_fields`, so Dataset can load again and there was no further problem (?) So I assume that it was a first-order issue of loading from MongoDB, not a second-order problem of concurrency. There is a set of scripts that the customers provide on how they were doing the actual ingestion of data into the DB. Happy to jump on a call to discuss this further!

brimoor:
A first order problem of loading from MongoDB feels unlikely to me. That could happen for any field, anywhere, at any time, right!? Noting special about the `datasets` collection in that regard. Therefore we'd be equally likely to see issue reports like "why is `filepath` None for this dataset?" or "why is my sample missing a tag I assigned to it?"

I do still think that "going into the DB and deleted null items under `sample_fields`" means deleting `dataset._doc.sample_fields[i]`, not deleting `dataset._doc.sample_fields[i].fields[j]`. But @ehofesmann can please set us straight 😄

minhtuev:
@brimoor: makes sense, I was able to reproduce this issue by adding "null" values to `sample_fields` in the Dataset record in MongoDB. Not sure how these null values were in the customer's dataset record in the first place, but as a patch, we can try to gracefully handle `null` value for `sample_fields[i]`.

Any thought on this @ehofesmann ? :)

minhtuev:
Full traceback:

```
In [1]: import fiftyone as fo
   ...:

In [2]: dataset2 = fo.load_dataset("open-images-v7-validation-200")
InvalidDocumentError                      Traceback (most recent call last)
Cell In[2], line 1
----> 1 dataset2 = fo.load_dataset("open-images-v7-validation-200")

File ~/workspace/fiftyone-teams/fiftyone/core/dataset.py:203, in load_dataset(name, snapshot)
    200     head_name = dataset_doc["name"]
    201     return _load_snapshot_dataset(name, head_name, snapshot)
--> 203 return Dataset(name, _create=False)

File ~/workspace/fiftyone-teams/fiftyone/core/singletons.py:36, in DatasetSingleton.__call__(cls, name, _create, *args, **kwargs)
     29 if (
     30     _create
     31     or instance is None
     32     or instance.deleted
     33     or instance.name is None
     34 ):
     35     instance = cls.__new__(cls)
---> 36     instance.__init__(name=name, _create=_create, *args, **kwargs)
     37     name = instance.name  # `__init__` may have changed `name`
     38 else:

File ~/workspace/fiftyone-teams/fiftyone/core/dataset.py:371, in Dataset.__init__(self, name, persistent, overwrite, _create, _virtual, _head_name, _snapshot_name, **kwargs)
    365 else:
    366     self.__permission = (
    367         dataset_permissions.get_dataset_permissions_for_current_user(
    368             name
    369         )
    370     )
--> 371     doc, sample_doc_cls, frame_doc_cls = _load_dataset(
    372         self, name, virtual=_virtual
    373     )
    375 self._doc = doc
    376 self._sample_doc_cls = sample_doc_cls

File ~/workspace/fiftyone-teams/fiftyone/core/dataset.py:7778, in _load_dataset(obj, name, virtual)
   7771 if version != focn.VERSION:
   7772     raise ValueError(
   7773         "Failed to load dataset '%s' from v%s using client v%s. "
   7774         "You may need to upgrade your client"
   7775         % (name, version, focn.VERSION)
   7776     ) from e
-> 7778 raise e

File ~/workspace/fiftyone-teams/fiftyone/core/dataset.py:7764, in _load_dataset(obj, name, virtual)
   7761     fomi.migrate_dataset_if_necessary(name)
   7763 try:
-> 7764     return _do_load_dataset(obj, name)
   7765 except Exception as e:
   7766     try:

File ~/workspace/fiftyone-teams/fiftyone/core/dataset.py:7787, in _do_load_dataset(obj, name)
   7785 if not res:
   7786     raise ValueError("Dataset '%s' not found" % name)
-> 7787 dataset_doc = foo.DatasetDocument.from_dict(res)
   7789 sample_collection_name = dataset_doc.sample_collection_name
   7790 frame_collection_name = dataset_doc.frame_collection_name

File ~/workspace/fiftyone-teams/fiftyone/core/odm/document.py:454, in MongoEngineBaseDocument.from_dict(cls, d, extended)
    449 d = json_util.loads(json_util.dumps(d))
    451 # from pdb import set_trace; set_trace()
    452
    453 # pylint: disable=no-member
--> 454 return cls._from_son(d)

File /opt/anaconda3/envs/vteam310/lib/python3.10/site-packages/mongoengine/base/document.py:830, in BaseDocument._from_son(cls, son, _auto_dereference, created)
    825     errors = "\n".join([f"Field '{k}' - {v}" for k, v in errors_dict.items()])
    826     msg = "Invalid data to create a `{}` instance.\n{}".format(
    827         cls._class_name,
    828         errors,
    829     )
--> 830     raise InvalidDocumentError(msg)
    832 # In STRICT documents, remove any keys that aren't in cls._fields
    833 if cls.STRICT:

InvalidDocumentError: Invalid data to create a `DatasetDocument` instance.
Field 'sample_fields' - 'NoneType' object has no attribute 'get'
```

minhtuev:
@brimoor @benjaminpkane : verified that simply filtering out null values from the list fixes the issue ~

@coderabbitai review

fiftyone/core/odm/document.py
```diff
@@ -440,14 +458,15 @@ def from_dict(cls, d, extended=False):
...
+                d = cls._simple_filter(d)
                 return cls._from_son(d)
             except Exception:
                 pass
 
         # Construct any necessary extended JSON components like ObjectIds
         # @todo can we optimize this?
         d = json_util.loads(json_util.dumps(d))
-
+        d = cls._simple_filter(d)
```
brimoor, line 469: In general, list fields can contain `None` values. For example:

```py
import fiftyone as fo

sample = fo.Sample(filepath="image.jpg", list_field=[None, None])

dataset = fo.Dataset()
dataset.add_sample(sample)

sample.reload()

assert sample.list_field == [None, None]
```

So we can't change `MongoEngineBaseDocument` as this is the base class for all documents, eg sample and frame documents, not just `DatasetDocument`.

A more robust patch here would be to gracefully catch validation errors related to `sample_fields` or `frame_fields` being None in a `DatasetDocument.from_dict()` super method, and, if found, immediately update these field(s) in the database so this patch doesn't get triggered permanently thereafter.

Ideally we'd still have a hypothesis for why this is happening and confirmation that this won't just lead to a downstream error as I described earlier:

> For example, I bet that updating this PR to gracefully omit dataset._doc.sample_fields[i] if None would not actually do much good. It would allow load_dataset() to work, but then I bet the user would immediately get a "found unknown field 'foo'" error upon calling sample = dataset.first() because whatever the None value was we omitted from the schema was probably supposed to be defining an actual field.

minhtuev: I see, now it all makes sense after digging into this code! We can certainly do both, with both patching for graceful handling and trying to figure out why `dataset._doc.sample_fields[i]` contains None value sometimes. 

@ehofesmann : do you know if the customers encounter any other problem after removing `null` values from `sample_fields`? Do they do any additional update to the record?

fiftyone/core/odm/document.py
```diff
@@ -432,6 +435,21 @@ def to_dict(self, extended=False):
         # @todo can we optimize this?
         return json.loads(json_util.dumps(d))
 
+    @classmethod
+    def _simple_filter(cls, d):
+        if isinstance(d, dict):
+            for k, v in d.items():
+                d[k] = cls._simple_filter(v)
+        elif isinstance(d, list):
+            # Pymongo cannot create a document with a list that contains None
```
kaixi-wang, line 444: I don't think this is true.

kaixi-wang (changes requested):
I don't think we should automatically be filtering out None fields as I've seen multiple instances where users filter by None to be able to identify groups of samples

fiftyone/core/odm/document.py
kaixi-wang: I don't think it's correct to assume that a none value field is a bug. This problem should be handled so None doesn't throw an error/is handled appropriately where the error is being thrown and not messing with the schema

minhtuev: @kaixi-wang agreed with the argument from Brian above that filtering out None value is not the solution. In general, do we have a sense when `None` value is acceptable as a value and when it is not? Pymongo throws an exception here because it is not accepting `None` as a value, so in this case the conversion fails.

minhtuev:
@kaixi-wang : we were discussing doing a try-catch exception for `sample_fields` since `None` conversion fails specifically for this field, but it would be helpful to know if there is a general rule for `None` conversion :)

minhtuev:
@coderabbitai full review

minhtuev:
Closing this PR since we discussed a different approach in chat

PR #4597, opened 2024-07-26
Adding `created_at`, `last_modified_at`, and read-only fields

brimoor:
## Change log

- Adds support for marking fields as read-only
- Adds auto-populated (and read-only) `created_at` fields to `Sample` and `Frame`
- Adds auto-populated (and read-only) `last_modified_at` fields to `Sample` and `Frame`
- Adds auto-populated `Dataset.last_modified_at` property
- Adds auto-populated `Field.created_at` property
- Adds auto-populated `created_at` property to all fields

## Implementation details

As currently implemented, `Dataset.last_modified_at` is not automatically updated in response to changes to its child samples, and `Sample.last_modified_at` of video samples is not automatically updated in response to modifications to its child frames. Instead, the user must cause this to happen (if they want it) by calling:

```py
dataset.sync_last_modified_at()
```

I chose this path because:
- `created_at` and `last_modified_at` indexes are created by default (with the assumption that sorting by these fields will be popular), so computing `$max` across these collections is efficient
- automatically updating upstream `last_modified_at` values upon every downstream update would be a significant performance penalty as it would 2x or 3x (for frame modifications) the number of update operations that must be performed.

## Migration considerations

The upward migration declares `created_at` and `last_modified_at` fields on all datasets and populates the `created_at` values via `{"$toDate": "$_id"}`, taking advantage of the fact that `ObjectId`s encode the original timestamp when they were generated. Since we do not know when the samples were actually last modified, though, `last_modified_at` is left as `None` (okay we could *also* set this to `{"$toDate": "$_id"}` as well since we technically edited the sample by setting `created_at`, but that's not any more informative than leaving it as None). I ensured these `None` values will not cause problems when edits are later made or `sync_last_modified_at()` is later called.

The downward migration does **not** delete the `created_at` and `last_modified_at` fields, since they are still valid fields, they will just no longer be default + read-only fields and the `last_modified_at` field will no longer be automatically updated. However, the assumption is that the user will either want to re-upgrade soon, at which point they'll be glad to have their `last_modified_at` values retained, or else they can just use `delete_sample_field()` to manually delete these fields.

### `created_at`/`last_modified_at`

```py
import fiftyone as fo
import fiftyone.zoo as foz

print(dataset.created_at)
# 2024-07-26 04:45:56.062000

print(dataset.last_modified_at)
# 2024-07-26 04:45:56.373000

print(dataset.bounds("created_at"))
# (datetime.datetime(2024, 7, 26, 4, 45, 56, 386000),
#  datetime.datetime(2024, 7, 26, 4, 45, 59, 901000))

print(dataset.bounds("last_modified_at"))
# (datetime.datetime(2024, 7, 26, 4, 45, 56, 386000),
#  datetime.datetime(2024, 7, 26, 4, 45, 59, 901000))

dataset.set_values("foo", ["bar"] * len(dataset))

print(dataset.bounds("last_modified_at"))
# (datetime.datetime(2024, 7, 26, 4, 46, 0, 36000),
#  datetime.datetime(2024, 7, 26, 4, 46, 0, 36000))

for sample in dataset.select_fields():
    sample["spam"] = "eggs"
    sample.save()

print(dataset.bounds("last_modified_at"))
# (datetime.datetime(2024, 7, 26, 4, 46, 0, 107000),
#  datetime.datetime(2024, 7, 26, 4, 46, 2, 140000))

dataset.sync_last_modified_at()

print(dataset.last_modified_at)
# 2024-07-26 04:46:02.140000
```

```py
from datetime import datetime

sample = dataset.first()

sample.created_at = datetime.utcnow()
# ValueError: Cannot edit read-only field 'created_at'

sample.last_modified_at = datetime.utcnow()
# ValueError: Cannot edit read-only field 'last_modified_at'
```

### Read-only fields

```py
sample = dataset.first()

field1 = dataset.get_field("filepath")
field1.read_only = True
field1.save()

field2 = dataset.get_field("ground_truth.detections.label")
field2.read_only = True
field2.save()

sample.filepath = "no.jpg"
# ValueError: Cannot edit read-only field 'filepath'

sample.ground_truth.detections[0].label = "no"
sample.save()
# ValueError: Cannot edit read-only field 'ground_truth.detections.label'

field1.read_only = False
field1.save()

field2.read_only = False
field2.save()

sample.filepath = "yes.jpg"
sample.ground_truth.detections[0].label = "yes"
sample.save()  # success
```


fiftyone/core/odm/document.py
```diff
@@ -649,63 +657,112 @@ def _save(
...
+            ops = self._insert(doc, deferred=deferred)
         else:
             # Update existing document
             updates = {}
-            sets, unsets = self._delta()
 
-            if sets:
-                updates["$set"] = sets
+            try:
+                # OPTIMIZATION: we monkey patch `to_mongo()` and
```
brimoor, line 686: Apologies for this, I couldn't help myself...

This significantly improves the efficiency of `sample.save()` if a sample contains many large fields but you've only made a simple edit to one field.

fiftyone/migrations/revisions/v0_26_0.py
```diff
@@ -0,0 +1,110 @@
...
+                "read_only": True,
+            }
+        )
+
+    return not found_created_at
+
+
+def _add_created_at(db, dataset_name, collection_name):
+    try:
+        pipeline = [{"$set": {"created_at": {"$toDate": "$_id"}}}]
```
benjaminpkane, line 103: Pulling the timestamp from the id. Very cool 🚀

fiftyone/core/odm/utils.py
```diff
@@ -325,6 +337,23 @@ def get_field_kwargs(field):
     return kwargs
 
 
+def _parse_field_str(field_str):
+    chunks = field_str.strip().split("(", 1)
```
benjaminpkane, line 341: Fancy! Maybe a comment for readability and future eyes 👀 
```suggestion
    # e.g. fiftyone.core.fields.EmbeddedDocumentField(fiftyone.core.labels.Detections)
    chunks = field_str.strip().split("(", 1)
```

fiftyone/core/collections.py
```diff
@@ -567,6 +568,100 @@ def summary(self):
         """
         raise NotImplementedError("Subclass must implement summary()")
 
+    def sync_last_modified_at(self, include_frames=True):
+        """Syncs the ``last_modified_at`` property(s) of the dataset.
+
+        Updates the :attr:`last_modified_at` property of the dataset if
+        necessary to incorporate any modification timestamps to its samples.
+
+        If ``include_frames==True``, the the ``last_modified_at`` property of
```
benjaminpkane, line 577:
```suggestion
        If ``include_frames==True``, the ``last_modified_at`` property of
```

benjaminpkane (approved):
Nuts and bolts LGTM. I like the new `mode` parameter 🚀

fiftyone/core/stages.py
```diff
@@ -1785,6 +1825,16 @@ def only_matches(self):
         """Whether to only include samples that match the filter."""
         return self._only_matches
 
+    def get_edited_fields(self, sample_collection, frames=False):
```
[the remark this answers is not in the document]
brimoor: Yeah there is code duplication here it's also true for some of the other `ViewStage` interface methods so I was just following along with the existing broken windows 🙈

fiftyone/core/odm/utils.py
brimoor: 👍 added

fiftyone/core/odm/document.py
```diff
@@ -649,63 +657,112 @@ def _save(
...
+                if validate and changed_fields:
+                    self._validate_updates(
+                        roots,
+                        paths,
+                        clean=clean,
+                        enforce_read_only=enforce_read_only,
+                    )
+
+                # OPTIMIZATION: apply monkey patch
+                doc = self.to_mongo(fields=roots)
```
swheaton, line 706: See https://github.com/voxel51/fiftyone/pull/4729 for full explanation of this super weird behavior. But even though that PR fixes the actual FO usage where this might rear its head, I'm proposing we be doubly safe and follow along with their bad code :( 
Making sure that the doc wouldn't be pruned in this case where we're not fully tracking changes.

```suggestion
                roots = roots if hasattr(self, "_changed_fields") else []
                doc = self.to_mongo(fields=roots)
```

brimoor: I'm still not fully understanding the monoengine weirdness here. If `roots` might not contain all fields, then `_validate_updates()` is not going to successfully validate that other fields contain proper values, right?

Do I need to just rip out this optimization?

brimoor: At minimum, happy to refactor the optimization into a separate PR for consideration.

swheaton: ah yes we would want the altered `roots` to be passed there too.

Basically the weirdness can be summarized by: "if doc does not have `_changed_fields` attribute then it is interpreted the same as `[]` meaning all fields, no matter what `doc._get_changed_fields()` returns"

I think putting into a separate PR for discussion there would be ideal 👍🏼

swheaton:
test failures 👀

swheaton:
forgot that PRs into other branches don't run the tests 🙃

fiftyone/core/odm/document.py
brimoor: Okay I updated this to handle docs with no `_changed_fields` property (in which case all fields are validated, just to be safe).

I could still move the monkey patch optimization to another PR, but the no `_changed_fields` property code still needs to be here, so the monkey patch is okay to ship IMO.

swheaton: ok

swheaton (approved):
lgtm - @brimoor are we ready to merge into develop?

brimoor:
@swheaton I've still been pondering in background mode whether we should ship it like this or automatically update `Dataset.last_modified_at`/`Sample.last_modified_at` when samples/frames are updated, respectively. Any thoughts about that?

Side note: I'm looking at https://github.com/voxel51/fiftyone/pull/4765 now.

brimoor:
Alright, code freeze approaches, this is fine as-is 🚀

swheaton:
great thanks :shipit:

swheaton:
> I proceeded from the PoV that we construct the dataset from the raw data produced in lab conditions when we control the metadata. Yes, the lab settings do not cover all data acquisition cases, take the public datasets for example.
> 
> Indeed, `Sample.created_at.created_at` does not distinguish samples, now I see.
> 
> My point about drawbacks was that `Sample.created_at` is almost equal to `Dataset.created_at` and cannot be sorted in the case of a small dataset and a single iteration of samples ingestion. This is a narrow subset of the rest cases I have not considered.
> 
> Thanks for the explanation.

No problem. Doing something sensible for all or almost all possible use cases is hard! That's why we add flexibility to cater it to your own situation with (hopefully) minimal extra code.

PR #4600, opened 2024-07-30
update dockerfile to use python base image

findtopher:

Update the Dockerfile to be based on `python:3.11-slim` (because we're deprecating python 3.8 and using the python base image seems to make more sense...)

```
make docker
docker run -p 5151:5151 -it local/fiftyone
import fiftyone as fo
import fiftyone.zoo as foz

```

results in `http://localhost:5151` displaying the quickstart dataset

        notes for FiftyOne users.

Updates the default docker base image to be `python:3.11-slim` to avoid deprecation notices using python 3.8

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Updated the Docker environment to a Python 3.11 base for improved package installation efficiency.
	- Introduced a multi-stage build process to streamline Docker image creation.

- **Bug Fixes**
	- Removed unnecessary system package installations, reducing complexity.

- **Chores**
	- Changed Docker image tag to `local/fiftyone` for local development focus.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,81 +1,45 @@
-# Dockerfile for building an image with a source FiftyOne install atop a
-# Debian-based Linux distribution.
+# Copyright 2017-2024, Voxel51, Inc.
+# voxel51.com
 #
-# By default, Ubuntu 20.04 and Python 3.8 are used, but these can be customized
-# via ARGs.
+# Dockerfile for building an image with source FiftyOne atop a Python 3.11
+#  base image
 #
 # ARGs::
 #
-#   BASE_IMAGE (ubuntu:20.04): The Debian-based image to build from
-#   PYTHON_VERSION (3.8): The Python version to install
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):

Dockerfile
```diff
@@ -1,81 +1,45 @@
...
 #       -p 5151:5151 \
-#       -it voxel51/fiftyone
-#
-# Copyright 2017-2022, Voxel51, Inc.
-# voxel51.com
+#       -it local/fiftyone
 #
 
-# The base image to build from; must be Debian-based (eg Ubuntu)
-ARG BASE_IMAGE=ubuntu:20.04
```
benjaminpkane, line 32: Is it worth bumping Ubuntu while we are here? 😄

benjaminpkane: Oh I see a Python based image is now used, SGTM

PR #4602, opened 2024-07-30
Update unit and integration test to validate value for `EmbeddedDocumentListField`

minhtuev:

Added a unit test for validating value for EmbeddedDocumentListField

* Unit test (✅ )
* Integration test (✅ )

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced a validation method for embedded document list fields to prevent `None` entries, enhancing data integrity.

- **Bug Fixes**
	- Addressed potential issues with invalid states in embedded document list fields by ensuring they do not accept `None` values.

- **Tests**
	- Added comprehensive unit tests for the validation logic of the EmbeddedDocumentListField, ensuring both valid and invalid inputs are correctly handled.
	- Enhanced test coverage for sample and frame field validations in datasets, confirming proper handling of null assignments.

### DIFF
--- a/tests/unittests/dataset_tests.py
+++ b/tests/unittests/dataset_tests.py
@@ -25,7 +25,6 @@ import fiftyone as fo
 import fiftyone.core.fields as fof
 import fiftyone.core.odm as foo
 import fiftyone.utils.data as foud
-import fiftyone.zoo as foz
 from fiftyone import ViewField as F
 
 from decorators import drop_datasets, skip_windows
@@ -1031,15 +1030,16 @@ class DatasetTests(unittest.TestCase):
 
     @drop_datasets
     def test_validate_sample_fields(self):
-        dataset = foz.load_zoo_dataset("quickstart")
-        dataset._doc.sample_fields[-1] = None
+        dataset = fo.Dataset()
+        dataset._doc.sample_fields.append(None)
         with self.assertRaises(ValidationError):
             dataset.save()
 
     @drop_datasets
     def test_validate_frame_fields(self):
-        dataset = foz.load_zoo_dataset("quickstart-video")
-        dataset._doc.frame_fields[-1] = None
+        dataset = fo.Dataset()
+        dataset.media_type = "video"
+        dataset._doc.frame_fields.append(None)
         with self.assertRaises(ValidationError):
             dataset.save()
 
--- a/tests/unittests/dataset_tests.py
+++ b/tests/unittests/dataset_tests.py
@@ -25,6 +25,7 @@ import fiftyone as fo
 import fiftyone.core.fields as fof
 import fiftyone.core.odm as foo
 import fiftyone.utils.data as foud
+import fiftyone.zoo as foz
 from fiftyone import ViewField as F
 
 from decorators import drop_datasets, skip_windows
@@ -1027,6 +1028,20 @@ class DatasetTests(unittest.TestCase):
             dataset.validate_field_type(
                 "spam", embedded_doc_type=fo.Detections
             )
+
+    @drop_datasets
+    def test_validate_sample_fields(self):
+        dataset = foz.load_zoo_dataset("quickstart")
+        dataset._doc.sample_fields[-1] = None
+        with self.assertRaises(ValidationError):
+            dataset.save()
+
+    @drop_datasets
+    def test_validate_frame_fields(self):
+        dataset = foz.load_zoo_dataset("quickstart-video")
+        dataset._doc.frame_fields[-1] = None
+        with self.assertRaises(ValidationError):
+            dataset.save()
 
     @drop_datasets
     def test_frame_field_schemas(self):
--- a/tests/unittests/documents.py
+++ b/tests/unittests/documents.py
@@ -1,0 +1,38 @@
+import unittest
+from mongoengine import EmbeddedDocument
+from mongoengine.errors import ValidationError
+from fiftyone.core.fields import EmbeddedDocumentListField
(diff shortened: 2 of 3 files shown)


minhtuev:
```
In [6]: dataset._doc.sample_fields.append(None)

In [7]: dataset._doc.save()
ValidationError                           Traceback (most recent call last)
Cell In[7], line 1
----> 1 dataset2._doc.save()

...

ValidationError: ValidationError (DatasetDocument:66a410180f828fad92885ee7) (Embedded document list fields should not contain None values.: ['sample_fields'])
```

benjaminpkane:
I think this change has the right intention. My concern is it is a breaking change for existing list fields with None values 🤔

brimoor (changes requested):
I just ran this on `develop`:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset._doc.sample_fields[-1] = None

dataset.save()
# ValidationError: ValidationError (DatasetDocument:66abb08fbe55cb66cf57fe2e) (9.Invalid embedded document instance provided to an EmbeddedDocumentField: ['sample_fields'])
```

So it appears that `fiftyone.core.fields.EmbeddedDocumentListField` **already** inherits validation from `mongoengine.fields.EmbeddedDocumentListField` that prevents `None` values from being saved via a "normal" pathway like setting an entry to `None` and then calling `save()`.

brimoor:
@minhtuev I appreciate the detective work here but this rabbit hole feels too deep, especially since we can't yet reproduce the error.

minhtuev:
@brimoor : oh nice, I should have checked that first haha! I am a little stumped, I am not sure how the DB was corrupted if we already have a validation method. We can keep the unit test here so to prevent future regression. I have checked with the customers, and it seems that they have moved on from the issue. I will continue to monitor the Slack channel in case the error arises again :)

minhtuev:
The other theory that we were discussing yesterday with @benjaminpkane is that the issue is caused by a concurrency error when Pymongo tries to update the database. We can revisit this theory again when the issue resurfaces.

benjaminpkane:
My take is we either migrate `sample_fields` and `frame_fields` to dict fields, or write our own odm array update logic to better handle concurrency in general. I prefer the latter

Longer term we should add transactions for real concurrency guarantees when users (customers) for want them

tests/unittests/documents.py
```diff
@@ -0,0 +1,36 @@
...
+from fiftyone.core.fields import EmbeddedDocumentListField
+from fiftyone.core.odm.dataset import SampleFieldDocument
+
+
+class MockEmbeddedDocument(EmbeddedDocument):
+    pass
+
+
+class TestEmbeddedDocumentListField(unittest.TestCase):
+    def test_validate(self):
```
brimoor, line 13: @minhtuev can you also add an integration test here to clarify why we're testing this in the first place?

```py
dataset = fo.Dataset()
dataset._doc.sample_fields[0] = None

# sample field docs cannot be None
with self.assertRaises(ValidationError):
    dataset.save()
```

brimoor:
> My take is we either migrate `sample_fields` and `frame_fields` to dict fields, or write our own odm array update logic to better handle concurrency in general. I prefer the latter

Yeah +1 to fixing what is likely the real issue here: concurrent updates of array fields. There's nothing special about `sample_fields`. I'll make a ticket for it.

tests/unittests/documents.py
minhtuev: Done ✅

tests/unittests/dataset_tests.py
```diff
@@ -1028,6 +1029,20 @@ def test_field_schemas(self):
                 "spam", embedded_doc_type=fo.Detections
             )
 
+    @drop_datasets
```
minhtuev, line 1032: These two tests live in `DatasetTests` class :)

tests/unittests/dataset_tests.py
```diff
@@ -1028,6 +1029,20 @@ def test_field_schemas(self):
...
+    def test_validate_sample_fields(self):
+        dataset = foz.load_zoo_dataset("quickstart")
```
brimoor, line 1034: We need to use `fo.Dataset()` here because we don't want unit tests to download datasets from the zoo (too large of a side effect)

tests/unittests/dataset_tests.py
```diff
@@ -1028,6 +1029,20 @@ def test_field_schemas(self):
...
+        dataset._doc.sample_fields[-1] = None
+        with self.assertRaises(ValidationError):
+            dataset.save()
+
+    @drop_datasets
+    def test_validate_frame_fields(self):
+        dataset = foz.load_zoo_dataset("quickstart-video")
```
brimoor, line 1041: ditto: use this to avoid downloading a zoo dataset:

```py
dataset = fo.Dataset()
dataset.media_type = "video"
```

brimoor (changes requested):

tests/unittests/dataset_tests.py
minhtuev: Done ✅

tests/unittests/documents.py
```diff
@@ -0,0 +1,36 @@
+import unittest
```
brimoor, line 1: nit: add the normal copyright header, eg:

```py
"""
FiftyOne document-related unit tests.

| Copyright 2017-2024, Voxel51, Inc.
|
"""
```

brimoor (approved):
LGTM

PR #4606, opened 2024-08-01
close more items of AdaptiveMenu on pressing ESC

imanjra:

Close "More Items" popover of adaptive menu when ESC is pressed

Using ActionRow in split panel view

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced keyboard shortcut support to close the menu using the "Escape" key.
	- Added a new `useKeyDown` hook for managing keyboard events, improving interactive capabilities.

- **Improvements**
	- Enhanced performance by optimizing the menu closure function.
	- Improved responsiveness of components by incorporating refresh capabilities based on state changes.

### DIFF
--- a/app/packages/state/src/hooks/useKeyDown.tsx
+++ b/app/packages/state/src/hooks/useKeyDown.tsx
@@ -1,0 +1,22 @@
+import { DependencyList, useEffect } from "react";
+
+/**
+ * a react hook that calls the handler when a given key is down
+ */
+export default function useKeyDown(
+  key: KeyboardEvent["key"],
+  handler: (down: boolean, e: KeyboardEvent) => void,
+  deps: DependencyList = []
+) {
+  useEffect(() => {
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key === key) {
+        handler(true, e);
+      }
+    };
+    document.addEventListener("keydown", handleKeyDown);
+    return () => {
+      document.removeEventListener("keydown", handleKeyDown);
+    };
+  }, [key, handler, ...deps]); // eslint-disable-line react-hooks/exhaustive-deps
+}
--- a/app/packages/components/src/components/AdaptiveMenu/index.tsx
+++ b/app/packages/components/src/components/AdaptiveMenu/index.tsx
@@ -9,7 +9,7 @@ import React, {
   useMemo,
   useRef,
 } from "react";
-import { MoveEvent, ReactSortable } from "react-sortablejs";
+import { MoveEvent, ReactSortable, SortableEvent } from "react-sortablejs";
 import { useSetRecoilState } from "recoil";
 import PillButton from "../PillButton";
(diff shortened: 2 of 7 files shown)


app/packages/state/src/hooks/useKeyDown.tsx
```diff
@@ -0,0 +1,22 @@
...
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key === key) {
+        handler(true, e);
+      }
+    };
+    document.addEventListener("keydown", handleKeyDown);
+    return () => {
+      document.removeEventListener("keydown", handleKeyDown);
+    };
+  }, [key, handler, deps]);
```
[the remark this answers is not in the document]
imanjra: I believe this is not necessary. As long as the array is of observable values, the effect should get invoked when any of the observable items changes

app/packages/state/src/hooks/useKeyDown.tsx
```diff
@@ -0,0 +1,22 @@
+import { DependencyList, useEffect } from "react";
+
+/**
+ * a react hook that calls the handler when a given key is down
+ */
+export default function useKeyDown(
```
benjaminpkane, line 6: In the future it might be nice to generalize native event hooks, so we don't have to enumerate each one 🤔 

Also, we should try to consolidate the `useEventHandler` hook in `hooks-utils.ts` so the pattern is clear. Maybe since these hooks aren't really stateful we create a new `@fiftyone/hooks` package

benjaminpkane (commented):
Works! I'm noticing that `ESC` now clears sample selection and closes the action row popout (first recording)

Maybe out of scope, but I'm also seeing some wrapping occur in certain actions when creating dynamic groups.

https://github.com/user-attachments/assets/f1ec665b-9d3b-4236-b223-59cc4d79b630

https://github.com/user-attachments/assets/a7b719ef-dbc9-43a3-a967-7e791b58d7f9

app/packages/state/src/hooks/useKeyDown.tsx
imanjra: +1 for `@fiftyone/hooks`, right now, they are all in state, so I added this one to state as well.

imanjra:
> Works! I'm noticing that ESC now clears sample selection and closes the action row popout (first recording)
This will be interesting to solve. Thinking out loud, we would need a common stack for key-down handler and as you press, things are popped off the stack as they handled

> Maybe out of scope, but I'm also seeing some wrapping occur in certain actions when creating dynamic groups.
Nice catch! I wonder if we should instead use the three loading dots instead of that loading text... Will check it out

imanjra:
@benjaminpkane fixed both wrapping of dynamic group button's loading state and escape clearing selected samples when adaptive menu's more items popout is opened.

benjaminpkane (approved):
Nice!! 🙇

PR #4612, opened 2024-08-02
rename PanelOperatorConfig to PanelConfig

ritch:
PanelOperatorConfig => PanelConfig

## Summary by CodeRabbit

- **New Features**
	- Updated class naming for improved clarity, renaming `PanelOperatorConfig` to `PanelConfig`.
	- Introduced an alias for backward compatibility, ensuring existing references to the old class name remain functional.
	- Enhanced the module interface by making `PanelConfig` available in the operators module.

- **Bug Fixes**
	- No changes affecting the control flow or functionality; existing features remain intact.

### DIFF
--- a/e2e-pw/src/shared/assets/plugins/e2e/__init__.py
+++ b/e2e-pw/src/shared/assets/plugins/e2e/__init__.py
@@ -97,7 +97,7 @@ class E2EProgress(foo.Operator):
 class E2ECounterPythonPanel(foo.Panel):
     @property
     def config(self):
-        return foo.PanelOperatorConfig(
+        return foo.PanelConfig(
             name="e2e_counter_python_panel",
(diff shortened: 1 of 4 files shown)


imanjra (approved):

brimoor (changes requested):
@ritch you'll also need to update `fiftyone/operators/__init__.py` to import both `PanelOperatorConfig` and `PanelConfig` symbols

brimoor (changes requested):
@ritch actually I think it would be best to **not** add the alias. The only reason I proposed it was to avoid disrupting panel prototyping, but in fact we should publish all panels with `PanelConfig`, and the best way to ensure that is to **force** all existing panels to rename.

brimoor (approved):
LGTM

PR #4624, opened 2024-08-06
Adapt index creation to support both sync and async mode

minhtuev:

Index creation can take a long time, especially for large datasets. This change allows `dataset.create_index` to support an async mode when `using_async` is true.

* Manual test (✅ )
* Unit test (in progress)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced asynchronous capabilities for the `create_index` method, allowing for non-blocking index creation.
	- Added new asynchronous properties for accessing sample and frame collections, enhancing non-blocking data retrieval.

- **Bug Fixes**
	- Improved functionality for handling asynchronous operations in collections and datasets, optimizing performance.


minhtuev:
@coderabbitai generate unit testing code for this file

fiftyone/core/collections.py
```diff
@@ -9175,12 +9211,22 @@ def create_index(self, field_or_spec, unique=False, **kwargs):
...
-        if is_frame_index:
-            coll = self._dataset._frame_collection
+        if using_async:
+            logging.info("Creating index '%s' asynchronously...", index_name)
+            coll = (
+                self._dataset._frame_collection_async
+                if is_frame_index
+                else self._dataset._sample_collection_async
+            )
+            name = await coll.create_index(index_spec, unique=unique, **kwargs)
```
brimoor, line 9221: What happens if this is never `await`ed? Will the index still be created?

brimoor (commented):
@minhtuev is there a way to support this without async? IE so user can do this in a REPL?

```py
dataset.create_index(...)

# do something else even though index is still being created
```

fiftyone/core/collections.py
minhtuev: Motor's version of `create_index` is a co-routine, so it must be awaited. 

Te wrapper should help with `dataset.create_index(...)` for both sync and async case

`dataset.create_index(...)` should function as normal, synchronously
`dataset.creata_index(..., using_async=True)` returns a co-routine, which should be awaited

minhtuev:
@brimoor usage example

Synchronous execution:
```
IPython 8.24.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import fiftyone as fo
   ...:

In [2]: import logging

In [3]: logging.basicConfig(
   ...:     level=logging.INFO,  # Set the logging level
   ...:     format='%(asctime)s - %(levelname)s - %(message)s',  # Set the format of the log message
   ...:     datefmt='%Y-%m-%d %H:%M:%S'  # Set the format of the date/time
   ...: )

In [4]: dataset = fo.load_dataset("dave-quickstart")
2024-08-07 13:04:14 - INFO - {"message": "Waiting for suitable server to become available", "selector": "Primary()", "operation": "buildinfo", "topologyDescription": "<TopologyDescription id: 66b3d33e902ba2daae4614da, topology_type: Unknown, servers: [<ServerDescription ('staging-mongodb.fiftyone.ai', 29017) server_type: Unknown, rtt: None>]>", "clientId": {"$oid": "66b3d33e902ba2daae4614da"}, "remainingTimeMS": 29}

In [5]: dataset.create_index("ground_truth.detections.bounding_box", using_async=False)
Out[5]: 'ground_truth.detections.bounding_box'

In [6]: dataset.create_index("ground_truth.detections.index", using_async=False)
2024-08-07 13:08:10 - INFO - Creating index 'ground_truth.detections.index' synchronously...
Out[6]: 'ground_truth.detections.index'

In [7]: dataset.create_index("ground_truth.detections.index", using_async=True)
2024-08-07 13:08:38 - INFO - Running async, checking for running event loop
Out[7]: <coroutine object SampleCollection.create_index at 0x102a5b8b0>

In [8]: exit
```

Async execution:
```
(voxel310) minhtuevo@minh-tues-mbp fiftyone % apython
Python 3.10.14 (main, May  6 2024, 14:42:37) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
---
This console is running in an asyncio event loop.
It allows you to wait for coroutines using the 'await' syntax.
Try: await asyncio.sleep(1, result=3)
---
>>> import fiftyone as fo
>>> dataset = fo.load_dataset("dave-quickstart")
>>> dataset.create_index("ground_truth.detections.index", using_async=True)
<coroutine object SampleCollection.create_index at 0x29c6ba8f0>
>>> import logging
>>> logging.basicConfig(level=logging.INFO)
>>> await dataset.create_index("ground_truth.detections.index", using_async=True)
'ground_truth.detections.index'
>>> await dataset.create_index("ground_truth.detections.area", using_async=True)
'ground_truth.detections.area'
```

fiftyone/core/collections.py
brimoor: My question is slightly different: what happens if I do this:

```py
dataset.create_index(..., using_sync=True)
```

and don't await it? Will the index be created or not?

minhtuev: The index won't be created as the coroutine is not executed

minhtuev: Something like this:

```
In [4]: dataset = fo.load_dataset("dave-quickstart")
2024-08-07 15:43:55 - INFO - {"message": "Waiting for suitable server to become available", "selector": "Primary()", "operation": "buildinfo", "topologyDescription": "<TopologyDescription id: 66b3f8abf5462878245b98b9, topology_type: Unknown, servers: [<ServerDescription ('staging-mongodb.fiftyone.ai', 29017) server_type: Unknown, rtt: None>]>", "clientId": {"$oid": "66b3f8abf5462878245b98b9"}, "remainingTimeMS": 29}

In [5]: dataset.list_indexes()
Out[5]:
['id',
 'filepath',
 'ground_truth.detections.label',
 'ground_truth.detections.confidence',
 'ground_truth.detections.tags',
 'predictions.detections.tags',
 'predictions.detections.label',
 'ground_truth.detections.bounding_box',
 'ground_truth.detections.index',
 'ground_truth.detections.area']

In [6]: dataset.drop_index("ground_truth.detections.label")

In [7]: dataset.list_indexes()
Out[7]:
['id',
 'filepath',
 'ground_truth.detections.confidence',
 'ground_truth.detections.tags',
 'predictions.detections.tags',
 'predictions.detections.label',
 'ground_truth.detections.bounding_box',
 'ground_truth.detections.index',
 'ground_truth.detections.area']

In [8]: dataset.create_index("ground_truth.detections.label", using_async=True)
2024-08-07 15:46:29 - INFO - Running async, checking for running event loop
Out[8]: <coroutine object SampleCollection.create_index at 0x1037def10>

In [9]: dataset.list_indexes()
Out[9]:
['id',
 'filepath',
 'ground_truth.detections.confidence',
 'ground_truth.detections.tags',
 'predictions.detections.tags',
 'predictions.detections.label',
 'ground_truth.detections.bounding_box',
 'ground_truth.detections.index',
 'ground_truth.detections.area']

In [10]: dataset.list_indexes()
Out[10]:
['id',
 'filepath',
 'ground_truth.detections.confidence',
 'ground_truth.detections.tags',
 'predictions.detections.tags',
 'predictions.detections.label',
 'ground_truth.detections.bounding_box',
 'ground_truth.detections.index',
 'ground_truth.detections.area']

In [11]: dataset.create_index("ground_truth.detections.label")
2024-08-07 15:46:50 - INFO - Creating index 'ground_truth.detections.label' synchronously...
Out[11]: 'ground_truth.detections.label'

In [12]: dataset.list_indexes()
Out[12]:
['id',
 'filepath',
 'ground_truth.detections.confidence',
 'ground_truth.detections.tags',
 'predictions.detections.tags',
 'predictions.detections.label',
 'ground_truth.detections.bounding_box',
 'ground_truth.detections.index',
 'ground_truth.detections.area',
 'ground_truth.detections.label']
```

minhtuev: Another solution is to keep `dataset.create_index(...)` as it is; I will refactor so that we have a completely different async method, `dataset.create_index_async(...)` which is identical, but in async mode instead :) In this way, we can avoid the complicated wrapper around `create_index(..., using_async=True)`.

brimoor: Okay my goal here is not to make `create_index()` async friendly per-se, but rather to see if there's a way for the code to complete (whether sync or async) and return while the index creation is still ongoing.

If there is a way, then we don't need to require [this code](https://github.com/voxel51/fiftyone-plugins/blob/2fe3f6bc59339f85ef199c432a97442dc040f12c/plugins/indexes/__init__.py#L38-L57) to be run via delegated execution; users could run the operator in immediate execution mode and their API request will complete although the index is still being built.

If you can't find a way for the index to be created without `await` or without synchronously waiting for `create_index()` to return, then we'll need to require users to run the operator I linked in delegated mode.

Make sense?

kaixi-wang:
Unless you issue additional commands to drop the index, once you call create index synchronously, can you just crl+c to avoid waiting for it to complete? (Never tried it) Or just call it in another window?

Note that for large index builds, calling create index could impact everyone using that cluster, and you probably shouldn't continue working on that dataset while an index is being built so not quite following the need/desire to wrap this method.

fiftyone/core/collections.py
minhtuev: Yes, we discussed this on Monday and I have been trying to find a way to send the command to build indexes via Mongo DB cluster but without success so far :) that's why I thought we can try utilizing the async Motor client since it is the official MongoDB async driver. I have been searching the documentation for RESTful CRUD APIs but it doesn't seem to be very popular; we can try to dive-deeper into Pymongo and try to pull up the URI and the command but I feel it was a little bit hacky.

Another approach would be to add a time to the sync `dataset.create_index(...)` so that it can terminate after a short period of time (a few seconds), but it might be tricky to time it since the client itself might take a while to open a connection to the DB cluster.

```
2024-08-07 13:04:14 - INFO - {"message": "Waiting for suitable server to become available", "selector": "Primary()", "operation": "buildinfo", "topologyDescription": "<TopologyDescription id: 66b3d33e902ba2daae4614da, topology_type: Unknown, servers: [<ServerDescription ('staging-mongodb.fiftyone.ai', 29017) server_type: Unknown, rtt: None>]>", "clientId": {"$oid": "66b3d33e902ba2daae4614da"}, "remainingTimeMS": 29}
```

minhtuev:
@kaixi-wang : I think once the cluster has successfully received the command it is okay to terminate, but it is a little bit tricky to time the execution code since the client might take sometimes to successfully open a connection to the DB cluster :D I wish they just have a background early return mode haha!

```
In [4]: dataset = fo.load_dataset("dave-quickstart")
2024-08-07 13:04:14 - INFO - {"message": "Waiting for suitable server to become available", "selector": "Primary()", "operation": "buildinfo", "topologyDescription": "<TopologyDescription id: 66b3d33e902ba2daae4614da, topology_type: Unknown, servers: [<ServerDescription ('staging-mongodb.fiftyone.ai', 29017) server_type: Unknown, rtt: None>]>", "clientId": {"$oid": "66b3d33e902ba2daae4614da"}, "remainingTimeMS": 29}
```

benjaminpkane (commented):
One option is a background thread with a queue for non-blocking requests. By default, maintaining blocking behavior

A lightning test is failing because the decorator does not block for index creation, I think. The lightning tests expect blocking index creation.

Outside of Python, I found this note on index creation in MongoDB interesting.
* https://www.mongodb.com/docs/manual/core/index-creation/#behavior

minhtuev:
@brimoor @benjaminpkane @kaixi-wang : here is a different take on this, adding a timeout decorator to `dataset.create_index(...)` so that it will timeout after `x` seconds. We can also add some retry logic so that it will do a longer retry, capping at `max_retry` after verifying that the index is not created for some reasons :)

https://github.com/voxel51/fiftyone/pull/4641

minhtuev:
Closing in favor of https://github.com/voxel51/fiftyone/pull/4624

PR #4587, opened 2024-07-19
Enable GPU inference for transformers models

mwoodson1:
## What changes are proposed in this pull request?

Automatically detect if transformers model is on the GPU and if so perform inference on same device.

## How is this patch tested? If it is not, please explain why.

Tested in collab with and without GPU, results are within precision difference of each other and no changes on CPU.

### Is this a user-facing change that should be mentioned in the release notes?

-   [ ] No. You can skip the rest of this section.
-   [ ] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

(Details in 1-2 sentences. You can just refer to another PR with a description
if this PR is part of a larger change.)

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [x] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced tensor operations by introducing device management, allowing for efficient use of GPU or CPU based on model parameters.
	- Input tensors are now automatically transferred to the designated device before processing, improving performance and compatibility.

PR #4634, opened 2024-08-08
Fix bug when loading HF transformers from zoo

brimoor:
This started failing when https://github.com/voxel51/fiftyone/pull/4587 was merged, now succeeds again:

```py
import fiftyone as fo
import fiftyone.zoo as foz
   
model = foz.load_zoo_model(
    "zero-shot-detection-transformer-torch",
    name_or_path="google/owlvit-base-patch32",
    classes=["cat", "dog", "bird", "fish", "turtle"],
)
```

## Summary by CodeRabbit

- **Enhancements**
	- Improved the initialization logic for model and device assignment, leading to more reliable performance during setup.

### DIFF
--- a/fiftyone/utils/transformers.py
+++ b/fiftyone/utils/transformers.py
@@ -452,7 +452,7 @@ class FiftyOneTransformer(TransformerEmbeddingsMixin, Model):
         self.config = config
         self.model = self._load_model(config)
         self.device = (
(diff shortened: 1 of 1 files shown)


mwoodson1 (approved):
Surprised we didn't catch this when we tested before 🤔 But lgtm

PR #4637, opened 2024-08-08
Python panel docs

brimoor:
### TODO
- [x] Update `Quick examples` section
    - [x] Convert HelloWorld JS panel to a python panel
    - [x] Add `Example Python panel` subsection
- [x] Add subsections to the `Developing panels` section that show short code blocks that demonstrate common patterns, such as:
    - [x] menu: a menu with a dropdown and action buttons with callbacks
    - [x] buttons: that use `ctx.trigger` and `ctx.prompt` to trigger operators
    - [x] plots: a histogram with callback that updates current view on clicking a bar
    - [x] "tutorial-style": vstack + markdown + videos + next/prev buttons (like [try_panels](https://github.com/voxel51/try_panels), but ensuring that all updated best practices are used <-- panels may use some hacks that were later converted to builtins)
- [x] Add links in appropriate places to example python panels in https://github.com/voxel51/fiftyone-plugins

## Summary by CodeRabbit

- **Documentation Enhancements**
	- Updated and clarified documentation for developing FiftyOne plugins, including detailed sections on Python and JavaScript plugins.
	- Introduced a new section on developing panels, highlighting their capabilities and interactivity within the app.
	- Improved docstrings and descriptions across various classes and methods to enhance usability and clarity.

- **Command-Line Interface Updates**
	- Added new optional arguments to the `fiftyone operators list` command, allowing users to filter output for operators or panels specifically. 

- **Functionality Improvements**
	- Enhanced operator listing functionality with the addition of a `type` parameter to filter between panels and operators in the `list_operators` method. 
    - Introduced a new function to consolidate operator and panel retrieval, improving code organization and maintainability.


fiftyone/operators/operations.py
```diff
@@ -299,35 +303,54 @@ def register_panel(
...
+        """Registers a panel with the given name and lifecycle callbacks.
 
         Args:
-            name: the name of the panel to register
-            icon (None): the icon to display in the panel tab
-            dark_icon (None): the icon to display in the panel tab in dark mode of app
-            light_icon (None): the icon to display in the panel tab in light mode of app
+            name: the name of the panel
+            label: the display name of the panel
+            icon (None): the icon to show in the panel's tab
```
ritch, line 311: For all icon settings: This isn't actually implemented.

fiftyone/operators/operations.py
```diff
@@ -299,35 +303,54 @@ def register_panel(
...
+            light_icon (None): the icon to show in the panel's tab when the App
+                is in light mode
+            dark_icon (None): the icon to show in the panel's tab when the App
+                is in dark mode
             on_load (None): an operator to invoke when the panel is loaded
             on_unload (None): an operator to invoke when the panel is unloaded
-            on_change (None): an operator to invoke when the panel state changes
-            allow_duplicates (False): whether to allow multiple instances of the panel
+            on_change (None): an operator to invoke when the panel state
```
ritch, line 318: These events are not supported here. Need to remove these:

```
            on_load (None): an operator to invoke when the panel is loaded
            on_unload (None): an operator to invoke when the panel is unloaded
            on_change (None): an operator to invoke when the panel state
                changes
            on_change_ctx (None): an operator to invoke when the panel
                execution context changes
            on_change_view (None): an operator to invoke when the current view
                changes
            on_change_dataset (None): an operator to invoke when the current
                dataset changes
            on_change_current_sample (None): an operator to invoke when the
                current sample changes
            on_change_selected (None): an operator to invoke when the current
                selected samples changes
            on_change_selected_labels (None): an operator to invoke when the
                current selected labels changes
            on_change_extended_selection (None): an operator to invoke when the
                current extended selection changes
```

Instead - these are supported as class methods only. They are enabled when the class method is defined.

fiftyone/operators/panel.py
```diff
@@ -1,36 +1,48 @@
...
-
 
 class PanelConfig(OperatorConfig):
-    """A configuration for a panel operator."""
+    """Configuration for a panel.
+
+    Args:
+        name: the name of the panel
+        label: the display name for the panel
+        icon (None): the icon to show in the panel's tab
```
ritch, line 21: This isn't supported currently (we should add it though)

fiftyone/operators/panel.py
```diff
@@ -118,39 +141,38 @@ class WriteOnlyError(Exception):
...
 
         Args:
-            key (str): The key.
-            value (any): The value.
+            key: a key, ``"nested.key.path"``, or dict mapping multiple
+                possibly-nested keys to values
+            value (None): the value, if key is a string
         """
-        pydash.set_(self._data, key, value)
+        d = {key: value} if value is not None else key
```
ritch, line 163: This should be something like:

```python
d = key if isinstance(key, dict) else {key: value}
```

Since `set('foo', None)` is valid.

fiftyone/operators/panel.py
```diff
@@ -174,25 +196,31 @@ def __getattr__(self, key):
...
 
         Args:
-            key (str): A dot delimited path.
-            value (any): The state value.
+            key: a key, ``"nested.key.path"``, or dict mapping multiple
+                possibly-nested keys to values
+            value (None): the value, if key is a string
         """
-        super().set(key, value)
+        d = {key: value} if value is not None else key
```
ritch, line 217: Same comment as above

fiftyone/operators/panel.py
```diff
@@ -118,39 +141,38 @@ class WriteOnlyError(Exception):
...
 
     def get(self, key, default=None):
-        """
-        Gets the value from the dictionary.
+        """Gets a value from the dictionary.
 
         Args:
-            key (str): The key.
-            default (any): The default value if key is not found.
+            key: a key
```
ritch, line 171: This should be the same description as the others. And we should match [the pydash description](https://pydash.readthedocs.io/en/latest/deeppath.html).

> A deep path string is used to access a nested data structure of arbitrary length. Each level is separated by a "." and can be used on both dictionaries and lists. If a "." is contained in one of the dictionary keys, then it can be escaped using "\". For accessing a dictionary key that is a number, it can be wrapped in brackets like "[1]".

ritch: If theres a way to define this once and refer to it for all `key` params... that would be ideal.

ritch (changes requested):
Overall awesome edits 🙇 ! The changes to get|set need a fix. Key could use a better description.

fiftyone/operators/panel.py
brimoor: ah good catch!

fiftyone/operators/panel.py
brimoor: Let's not encourage people to use numeric keys or keys that contain `.`. That will break other things like the ability to use attribute access syntax `ctx.panel.state.key`.

Will update this `key` docstring to match the others though 👍

fiftyone/operators/panel.py
```diff
@@ -1,62 +1,79 @@
...
-        self.dark_icon = dark_icon
         self.light_icon = light_icon
+        self.dark_icon = dark_icon
         self.allow_multiple = allow_multiple
         self.unlisted = True
         self.on_startup = True
         self.kwargs = kwargs  # unused, placeholder for future extensibility
 
     def to_json(self):
-        d = super().to_json()
```
brimoor, line 40: @ritch just confirming: is it okay to remove `OperatorConfig` attributes from `PanelConfig.to_json()` here? 

Doing this makes `fiftyone operators info my/panel` show only the panel-relevant properties, which is cleaner.

fiftyone/operators/panel.py
```diff
@@ -1,62 +1,79 @@
...
             "label": self.label,
             "icon": self.icon,
-            "dark_icon": self.dark_icon,
             "light_icon": self.light_icon,
+            "dark_icon": self.dark_icon,
             "allow_multiple": self.allow_multiple,
         }
 
 
-# Alias for backwards compatibility
```
brimoor, line 52: @ritch just confirming that I'm deleting `PanelOperatorConfig` here as discussed

fiftyone/plugins/definitions.py
```diff
@@ -86,7 +86,9 @@ def fiftyone_requirement(self):
     @property
     def operators(self):
         """The operators of the plugin."""
-        return self._metadata.get("operators", [])
+        return self._metadata.get("operators", []) + self._metadata.get(
```
brimoor, line 89: @ritch just confirming: this adds support for `panels` as an alias of `operators` in `fiftyone.yml` files

fiftyone/operators/registry.py
```diff
@@ -86,6 +93,15 @@ def list_operators(self, include_builtin=True):
         if include_builtin:
             operators.extend(BUILTIN_OPERATORS)
 
+        if type == "panel":
+            operators = [op for op in operators if isinstance(op, Panel)]
+        elif type == "operator":
+            operators = [op for op in operators if not isinstance(op, Panel)]
```
imanjra, line 99: Looks like a type `isinstance(op, Panel)`. Will need to import and use op here

fiftyone/operators/panel.py
```diff
@@ -49,9 +49,7 @@ def __init__(
         self.kwargs = kwargs  # unused, placeholder for future extensibility
 
     def to_json(self):
-        d = super().to_json()
```
imanjra, line 40: We should keep this for things. I believe `on_startup` will break

imanjra (approved):
LGTM. Let's merge this one in. We can follow-up further changes.

ritch (approved):
👍

PR #4633, opened 2024-08-07
Add `indexes_in_progress` field to dataset stats

minhtuev:
## What changes are proposed in this pull request?

Added a field to return the list of dataset indices that are in the progress of being built

## How is this patch tested? If it is not, please explain why.

* Manual testing (✅ )
* Unit test (✅ )

### Is this a user-facing change that should be mentioned in the release notes?

-   [x] No. You can skip the rest of this section.
-   [ ] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

(Details in 1-2 sentences. You can just refer to another PR with a description
if this PR is part of a larger change.)

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [x] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Enhanced index information retrieval with the addition of an `include_progress` parameter for more comprehensive reporting.
  - New metrics added to statistics, including `indexes_in_progress` and `index_sizes` for improved dataset index state visibility.

- **Bug Fixes**
  - Improved handling and reporting of index build status, ensuring accurate representation of index states.

- **Tests**
  - Added new tests to validate the behavior of dataset indexing, specifically checking the status of index builds.

PR #4641, opened 2024-08-08
Simple implementation for timeout decorator with specific graceful exception

minhtuev:

`dataset.create_index(...)` can timeout after `x` seconds

* Manual testing (✅ )
* Unit tests

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced a new exception class for handling timeouts more effectively.
	- Added a timeout decorator to manage function execution duration and improve error handling.
	- Enhanced the `create_index` method to prevent indefinite blocking during index creation.

These updates enhance the application's robustness and user experience when dealing with long-running operations.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -70,6 +70,13 @@ def _make_registrar():
     return registrar
 
 
+class SuccessTimeout(Exception):
+    def __init__(self, seconds):
+        message = "Succeeded after %s seconds" % seconds
+        logging.info(message)
+        super().__init__(message)
+
+
 class GracefulTimeoutError(Exception):
     def __init__(self, seconds):
         message = "Graceful timeout after %s seconds" % seconds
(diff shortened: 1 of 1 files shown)


brimoor (commented):
@minhtuev I appreciate the creative solutions here with `async` and now `@timeout` 🧠 

My goal here is to understand whether `create_index()` can be called in an operator that executes immediately.

In https://github.com/voxel51/fiftyone/pull/4624 we learned that `async` alone is not a solution, as the coroutine must still be `await`ed to trigger the index build.

IMO, this approach with `@timeout` is a bit too hacky to ship for two reasons:
- 10 seconds is too long to wait if the goal is to provide a solution for immediate execution
- as you mentioned offline, if we decrease substantially below 10 seconds then maybe the initial request never succeeds and the index build is never triggered

I haven't looked at the implementation of pymongo's `create_index()`, but it was theorized and/or confirmed that, under the hood, it's doing a build request + polling. If true, then directly calling the build request seems way less hacky than relying on network timing.

If the above doesn't work and there's no other ideas, then feels like we just throw in the towel and require index builds to be done via delegated operations because apparently that's the way the pymongo overlords intended it.

minhtuev:
@brimoor: let me look into Pymongo's `create_index(...)` again to see if we can pull something out of it that will just send the build command to the cluster without waiting for a confirmation.

But to add to the spirit of hacking, I have added an implementation for early stopping that includes an early stopping check if the new index is successfully in the building process by the cluster :)

fiftyone/core/collections.py
```diff
@@ -66,6 +70,83 @@ def registrar(func):
...
+                else TimeoutError(
+                    "Function call timed out after %s seconds" % time_count
+                )
+            )
+
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            dataset = args[0]
+            index_name = args[1] if len(args) > 1 else kwargs.get("index_name")
```
minhtuev, line 127: We have an implementation for more robust retrieving index-in-progress information here:
https://github.com/voxel51/fiftyone/pull/4633

minhtuev:
Closing in favor of https://github.com/voxel51/fiftyone/pull/4624

PR #4681, opened 2024-08-14
Make to_(generated view) explicit about params it can use

swheaton:
`Explicit is better than implicit.`

In service of this, explicitly pass parameters in `to_xxx()` that can be used by `make_xxx_dataset()` functions. They're typically documented anyways so this just rounds the square.

Tests pass

## Summary by CodeRabbit

- **New Features**
	- Enhanced data processing methods with additional parameters for greater flexibility.
	- Users can now customize operations in `to_patches`, `to_evaluation_patches`, `to_clips`, `to_trajectories`, and `to_frames` methods.
	- Improved control over clip generation and trajectory customization.

These updates empower users to tailor their data handling processes more effectively within the FiftyOne framework.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -6835,7 +6835,11 @@ class SampleCollection(object):
 
     @view_stage
     def to_patches(
-        self, field, other_fields=None, keep_label_lists=False, **kwargs
+        self,
+        field,
+        other_fields=None,
+        keep_label_lists=False,
+        **kwargs,
     ):
         """Creates a view that contains one sample per object patch in the
         specified field of the collection.
@@ -7087,7 +7091,12 @@ class SampleCollection(object):
 
     @view_stage
     def to_trajectories(
-        self, field, other_fields=None, tol=0, min_len=0, **kwargs
+        self,
+        field,
+        other_fields=None,
+        tol=0,
+        min_len=0,
+        **kwargs,
     ):
         """Creates a view that contains one clip for each unique object
         trajectory defined by their ``(label, index)`` in a frame-level field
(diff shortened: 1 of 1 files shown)


benjaminpkane (commented):
I like it 😄

brimoor (commented):
@swheaton so there was actually _some_ method to the madness behind why `to_XXX()` uses `**kwargs` in the method signature but then documents all the parameters explicitly in the docstring. As usual, it has to do with _saved views_.

If we make the proposed change here, every possible `kwarg` will be persisted in the view stage's `_config` and thus will be explicitly injected into the `make_XXX_dataset()` method henceforward whenever the saved view needs to be regenerated, eg here:
https://github.com/voxel51/fiftyone/blob/c0bc6b0e7020720acde533cf09e28249d42c52f5/fiftyone/core/stages.py#L7480

This means that the serialized size of saved generated views will get a bit bigger. Okay, not a big deal as its just <= 10 parameters. But, it also means that we'd need to run a database migration if we ever wanted to remove/rename one of the more obscure parameters, otherwise folks' saved views would start failing to load just because we stored a default value of an obscure unused parameter in the saved view's definition. 

The current behavior optimizes a bit more for flexibility: unless the user specifically set a non-default value for a certain parameter when defining the view stage, we are free to change/deprecate parameters without needing to worry about database migrations (and if the user actually did explicitly set the value of a parameter that we changed, then it's for the best that the view fails to load in newer versions anyway).

brimoor:
@swheaton net-net, I'd lean towards _not_ making this change to support your other motives, as it does actually have some side effects.

swheaton:
Could make it work without changes, but not worth it. Will withdraw

PR #4684, opened 2024-08-15
SAM2 tweaks

brimoor:


docs/scripts/make_model_zoo_docs.py
```diff
@@ -98,15 +101,16 @@
...
     dataset = foz.load_zoo_dataset("quickstart-video", max_samples=2)
 
     # Only retain detections in the first frame
-    for sample in dataset:
-        for frame_idx, frame in sample.frames.items():
-            if frame_idx >= 2:
-                frame.detections = None
-        sample.save()
+    (
+        dataset
```
brimoor, line 109: Idiomatic way to clear frames > 1 without reading anything into memory 🤓

fiftyone/utils/sam2.py
```diff
@@ -23,19 +24,19 @@
 fou.ensure_torch()
 import torch
 
-sam2 = fou.lazy_import("sam2")
-smvid = fou.lazy_import("sam2.sam2_video_predictor")
-smutil = fou.lazy_import("sam2.utils.misc")
-smip = fou.lazy_import("sam2.sam2_image_predictor")
+sam2 = fou.lazy_import("sam2", callback=lambda: fou.ensure_import("sam2"))
```
brimoor, line 27: Unfortunately sam2 is not yet available via `pip`, so we can't add it to the requirements list in the models manifest.

However, adding `fou.ensure_import("sam2")` here and calling `dir(sam2)` in the constructor of each model will cause an informative error to be displayed at runtime if SAM2 isn't installed:

```
ImportError: The requested operation requires that 'sam2' is installed on your machine.

If you think this error is inaccurate, you can set `fiftyone.config.requirement_error_level` to 1 (warning) or 2 (ignore).
See https://docs.voxel51.com/user_guide/config.html for details.
```

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -215,7 +215,7 @@
                     "support": true
                 }
             },
-            "tags": ["segment-anything-2", "torch", "zero-shot"],
+            "tags": ["segment-anything", "torch", "zero-shot"],
```
brimoor, line 218: Okay with using `segment-anything` for the entire family of segment anything models? 

I believe the code snippets for SAM1 and SAM2-Image are identical, so the built docs should still be valid with this tag change.

docs/scripts/make_model_zoo_docs.py
prernadh: Awesome thanks!

fiftyone/utils/sam2.py
```diff
@@ -257,23 +261,24 @@ class SegmentAnything2VideoModel(fom.SamplesMixin, fom.Model):
     """
 
     def __init__(self, config):
-        self._fields = {}
```
prernadh, line 260: @brimoor Why did you get rid of this line? I assume it was by mistake?

prernadh (commented):
Changes look good except need to add the only line that initializes `model._field`

prernadh (approved):
Breaking line added back in, rest of it looks good to me

fiftyone/utils/sam2.py
brimoor: oh I didn't see it used anywhere but I guess it is!

PR #4689, opened 2024-08-16
Adding safe mode to drop index

minhtuev:

Adding a default safe mode to `dataset.drop_index` where the users are prompted to confirm before dropping the index

* Manual (✅ )
* Unit test (✅ )

        notes for FiftyOne users.

By default, `dataset.drop_index(...)` will include a confirmation prompt before proceeding.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced a `safe_mode` option for the `drop_index` method, allowing users to confirm before dropping indexes, enhancing safety against accidental deletions.
	- Added a `database_safe_mode` configuration option to improve safety during database interactions.

- **Tests**
	- Updated tests for the `drop_index` method to reflect the new `safe_mode` functionality, ensuring existing behaviors are maintained while allowing for more flexibility in index management.
	- Introduced new test cases to verify behavior when the `safe_mode` option is enabled.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -24,6 +24,7 @@ from pymongo import InsertOne, UpdateOne, UpdateMany
 import eta.core.serial as etas
 import eta.core.utils as etau
 
+import fiftyone as fo
 import fiftyone.core.aggregations as foa
 import fiftyone.core.annotation as foan
 import fiftyone.core.brain as fob
@@ -9261,14 +9262,18 @@ class SampleCollection(object):
                 indexes
             safe_mode (True): whether to use safe mode when dropping the index
         """
-        if safe_mode:
+        if fo.config.database_safe_mode and safe_mode:
             agreement = None
-            while agreement not in {"y", "n", "yes", "no"}:
+            acceptable = {"y", "n", "yes", "no"}
+            while agreement not in acceptable:
                 agreement = input("Are you sure you want to drop the index? (y/n)?").lower()
+                if agreement not in acceptable:
+                    print("Invalid response. Please enter 'y' or 'n'.")
 
             if agreement in {"n", "no"}:
                 return
 
+        logging.info("Dropping index '%s'...", field_or_name)
         name, is_frame_index = self._handle_frame_field(field_or_name)
 
         if is_frame_index:
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -9201,7 +9201,7 @@ class SampleCollection(object):
                     )
 
                 # We need to drop existing index and replace with a unique one
-                self.drop_index(field)
+                self.drop_index(field, safe_mode=False)
 
         is_frame_fields = []
         index_spec = []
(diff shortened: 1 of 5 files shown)


fiftyone/core/collections.py
```diff
@@ -9252,14 +9252,23 @@ def create_index(self, field_or_spec, unique=False, **kwargs):
...
         Args:
             field_or_name: a field name, ``embedded.field.name``, or compound
                 index name. Use :meth:`list_indexes` to see the available
                 indexes
+            safe_mode (True): whether to use safe mode when dropping the index
         """
+        if safe_mode:
+            agreement = None
+            while agreement not in {"y", "n", "yes", "no"}:
+                agreement = input("Are you sure you want to drop the index? (y/n)?").lower()
```
benjaminpkane, line 9267: We don't have a precedent of blocking input in the SDK, I believe. I like this, but I think it will break existing automation/scripts 🤔 

We could add a `database_safe_mode` as an opt-in setting to [FiftyOneConfig](https://github.com/voxel51/fiftyone/blob/991c1d0cead89eff78a1a11467293bd886b976a1/fiftyone/core/config.py#L59) that hard errors if unwanted index commands are attempted.

Thoughts? @swheaton @minhtuev @brimoor

swheaton: yeah agree. I don't think people are going to be dropping indexes willy nilly anyways, do you?

benjaminpkane: Hard to say. Also, since this is mostly Teams-specific, we could move this logic into a new API/Dataset permission

minhtuev: We can add config var to avoid backward compatible issue with the automation script but I think this is a necessary step to make our practice in sync with MongoDB best practices; in Compass they prompt the index name to confirm index deletion.

fiftyone/core/collections.py
```diff
@@ -9252,14 +9253,27 @@ def create_index(self, field_or_spec, unique=False, **kwargs):
 
         return name
 
-    def drop_index(self, field_or_name):
+    def drop_index(self, field_or_name, safe_mode=True):
```
minhtuev, line 9256: @coderabbitai generate unit testing code for safe_mode

tests/unittests/patches_tests.py
```diff
@@ -102,13 +102,13 @@ def test_to_patches(self):
         self.assertSetEqual(set(indexes), default_indexes)
 
         with self.assertRaises(ValueError):
-            view.drop_index("id")  # can't drop default index
+            view.drop_index("id", safe_mode=False)  # can't drop default index
```
minhtuev, line 105: When we set the env var `FIFTYONE_DATABASE_SAFE_MODE` to False, we do not need to set this flag any more :)

brimoor (changes requested):
If we were to do this, I definitely would insist on `database_safe_mode=False` being the default for backwards compatibility.

I have some additional thoughts that I'll share offline.

minhtuev:
As discussed offline, we will move the guardrail to the Index Management Panel.

PR #4693, opened 2024-08-16
Add `mask_path` to `fo.Detection` labels

### LINKED ISSUE #4486: [FR] Allow `mask_path` in `Detection` labels
### Proposal Summary

Currently `Detection` only accepts `mask` fields, which are serialized numpy arrays saved to the mongo database. It would instead be useful to be able to instruct fiftyone to use masks saved on disk, with a field `mask_path`, just like how `Segmentation` does: https://github.com/voxel51/fiftyone/blob/39201f4c84ddf4d1b08a306b897d16755f119c32/fiftyone/core/labels.py#L1006-L1021

### Motivation

This would make it possible to offload some disk usage from the mongo database to somewhere else.

-   [ ] App: FiftyOne application

### Details

I'm not quite familiar with the codebase yet, though from looking at `Segmentation` it doesn't look that hard. I would appreciate a maintainer's opinion on the feasibility of this feature.

### Willingness to contribute

-   [ ] Yes. I can contribute this feature independently
-   [x] Yes. I would be willing to contribute this feature with guidance from the FiftyOne community
-   [ ] No. I cannot contribute this feature at this time

Laurent2916:

Fix #4486

I genuinely couldn't figure out how to run the tests, I followed the [instructions](https://github.com/voxel51/fiftyone/blob/develop/CONTRIBUTING.md#developer-guide) but kept getting errors when using `pytest`, so I don't know if I have inadvertently broken something.
I wrote a quick snippet to test the feature instead:
```py
import tempfile
import numpy as np
from PIL import Image
import fiftyone as fo

MASK_SIZE = 256
BBOX = [0, 0, 1, 1]

# create a "mask"
mask = (np.random.rand(MASK_SIZE, MASK_SIZE) * 255).astype(np.uint8)
mask_image = Image.fromarray(mask)
mask_image.save(mask_path)
print(f"Saved mask to {mask_path}")

# check mask are the same, even though they are stored differently
detection1 = fo.Detection(label="detection1", mask=mask, bounding_box=BBOX)
detection2 = fo.Detection(label="detection2", mask_path=mask_path, bounding_box=BBOX)
assert np.allclose(detection1.get_mask(), detection2.get_mask()), "detection masks are different"  # type: ignore
print(detection1, detection2)

# check polyline conversion
polyline1 = detection1.to_polyline()
polyline2 = detection2.to_polyline()
assert np.allclose(polyline1.points, polyline2.points), "polyline points are different" # type: ignore

# check mask export
mask2_path = tempfile.mktemp(".png")
detection1.export_mask(mask2_path, update=True)
print(f"Exported mask to {mask2_path}")
mask2_image = Image.open(mask2_path)
mask2 = np.array(mask2_image)
assert np.allclose(mask, mask2), "exported mask is different"
print(detection1, detection2)

# check mask import
detection2.import_mask(update=True)
assert np.allclose(detection1.get_mask(), detection2.get_mask()), "imported mask is different"  # type: ignore
print(detection1, detection2)
```

-   [x] Yes. Give a description of this change to be included in the release notes for FiftyOne users.

Allows `Detection`s to store masks on disk, this allows to offload some disk usage from the mongo database to somewhere else. See also #4486

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced new capabilities for managing instance segmentation masks in the Detection class, including import/export functionality and transformations.
	- Enhanced mask retrieval processes across various modules for improved maintainability.

- **Bug Fixes**
	- Updated checks for mask presence to utilize the new mask retrieval methods, ensuring consistent access and functionality.

### DIFF
--- a/fiftyone/core/labels.py
+++ b/fiftyone/core/labels.py
@@ -401,17 +401,117 @@ class Detection(_HasAttributesDict, _HasID, Label):
         mask (None): an instance segmentation mask for the detection within
             its bounding box, which should be a 2D binary or 0/1 integer numpy
             array
+        mask_path (None):  the absolute path to the instance segmentation image
+            on disk
         confidence (None): a confidence in ``[0, 1]`` for the detection
         index (None): an index for the object
         attributes ({}): a dict mapping attribute names to :class:`Attribute`
             instances
     """
 
+    _MEDIA_FIELD = "mask_path"
+
     label = fof.StringField()
     bounding_box = fof.ListField(fof.FloatField())
     mask = fof.ArrayField()
+    mask_path = fof.StringField()
     confidence = fof.FloatField()
     index = fof.IntField()
+
+    @property
+    def has_mask(self):
+        """Whether this instance has a mask."""
+        return self.mask is not None or self.mask_path is not None
+
+    def get_mask(self):
+        """Returns the segmentation mask for this instance.
+
+        Returns:
+            a numpy array, or ``None``
+        """
+        if self.mask is not None:
+            return self.mask
+
+        if self.mask_path is not None:
+            return _read_mask(self.mask_path)
+
+        return None
+
+    def import_mask(self, update=False):
+        """Imports this instance's mask from disk to its :attr:`mask`
+        attribute.
+
+        Args:
+            outpath: the path to write the map
+            update (False): whether to clear this instance's :attr:`mask_path`
+                attribute after importing
+        """
+        if self.mask_path is not None:
+            self.mask = _read_mask(self.mask_path)
+
+            if update:
+                self.mask_path = None
+
+    def export_mask(self, outpath, update=False):
+        """Exports this instance's mask to the given path.
+
+        Args:
+            outpath: the path to write the mask
+            update (False): whether to clear this instance's :attr:`mask`
+                attribute and set its :attr:`mask_path` attribute when
+                exporting in-database segmentations
+        """
+        if self.mask_path is not None:
+            etau.copy_file(self.mask_path, outpath)
+        else:
+            _write_mask(self.mask, outpath)
+
+            if update:
+                self.mask = None
+                self.mask_path = outpath
+
+    def transform_mask(self, targets_map, outpath=None, update=False):
+        """Transforms this instance's mask according to the provided targets
+        map.
+
+        This method can be used to transform between grayscale and RGB masks,
+        or it can be used to edit the pixel values or colors of a mask without
+        changing the number of channels.
+
+        Note that any pixel values not in ``targets_map`` will be zero in the
+        transformed mask.
+
+        Args:
+            targets_map: a dict mapping existing pixel values (2D masks) or RGB
+                hex strings (3D masks) to new pixel values or RGB hex strings.
+                You may convert between grayscale and RGB using this argument
+            outpath (None): an optional path to write the transformed mask on
+                disk
+            update (False): whether to save the transformed mask on this
+                instance
+
+        Returns:
+            the transformed mask
+        """
+        mask = self.get_mask()
+        if mask is None:
+            return
+
+        mask = _transform_mask(mask, targets_map)
+
+        if outpath is not None:
+            _write_mask(mask, outpath)
+
+            if update:
+                self.mask = None
+                self.mask_path = outpath
+        elif update:
+            if self.mask_path is not None:
+                _write_mask(mask, self.mask_path)
+            else:
+                self.mask = mask
+
+        return mask
 
     def to_polyline(self, tolerance=2, filled=True):
         """Returns a :class:`Polyline` representation of this instance.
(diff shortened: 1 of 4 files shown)


fiftyone/utils/coco.py
```diff
@@ -1299,7 +1299,7 @@ def from_label(
             x, y, w, h = label.bounding_box
             bbox = [x * width, y * height, w * width, h * height]
 
-            if label.mask is not None:
+            if label.get_mask() is not None:
```
brimoor, line 1302: `label.has_mask` would be more performant here

fiftyone/utils/coco.py
```diff
@@ -2110,7 +2110,7 @@ def _coco_objects_to_detections(
         )
 
         if detection is not None and (
-            not load_segmentations or detection.mask is not None
+            not load_segmentations or detection.get_mask() is not None
```
brimoor, line 2113: `label.has_mask` would be more performant here

fiftyone/utils/cvat.py
```diff
@@ -6400,7 +6400,7 @@ def _create_detection_shapes(
                     }
                 )
             elif label_type in ("instance", "instances"):
-                if det.mask is None:
+                if det.get_mask() is None:
```
brimoor, line 6403: `label.has_mask` would be more performant here

fiftyone/core/labels.py
```diff
@@ -401,18 +401,118 @@ class Detection(_HasAttributesDict, _HasID, Label):
...
+        if self.mask_path is not None:
+            etau.copy_file(self.mask_path, outpath)
+        else:
+            _write_mask(self.mask, outpath)
+
+            if update:
+                self.mask = None
+                self.mask_path = outpath
+
+    def transform_mask(self, targets_map, outpath=None, update=False):
```
brimoor, line 473: This method only makes sense for `Segmentation` fields, where the masks are non-binary and thus the pixel values represent class semantics.

Here, the masks must be strictly binary so `targets_map` doesn't make sense.

fiftyone/core/labels.py
```diff
@@ -401,18 +401,118 @@ class Detection(_HasAttributesDict, _HasID, Label):
...
             array
+        mask_path (None):  the absolute path to the instance segmentation image
+            on disk
         confidence (None): a confidence in ``[0, 1]`` for the detection
         index (None): an index for the object
         attributes ({}): a dict mapping attribute names to :class:`Attribute`
             instances
     """
 
+    _MEDIA_FIELD = "mask_path"
```
brimoor, line 412: Note to Voxel51, this would be the first instance of a media field that could be embedded in a list (`Detections`), which means these methods will need to be updated:
- https://github.com/voxel51/fiftyone/blob/1824ddb5f579842664b9caf2f400ab2934191ff1/fiftyone/utils/data/exporters.py#L2345
- https://github.com/voxel51/fiftyone/blob/1824ddb5f579842664b9caf2f400ab2934191ff1/fiftyone/utils/data/exporters.py#L2041

There could be more such places too.

brimoor (changes requested):
Hi @Laurent2916! 👋 

Looks like you're interested in working with detection tasks where the object masks are stored as images on disk rather than as numpy arrays in the database.

Out of curiosity, how big are your instance masks in your use cases? In terms of resolution, number of objects per image, etc.

From a data modeling precedent standpoint, we have another closely related proposal in https://github.com/voxel51/fiftyone/pull/4500. In that case, the objective is also to work with detection datasets defined by on-disk images rather than in-database arrays, but there is an additional assumption that the objects are strictly non-overlapping and thus it is proposed to be achieved by adding a single full-resolution mask at the `Detections.mask|mask_path` level.

There's a quite detailed discussion there about how in the author's opinion, object detections represented in that way should be distinct from the existing `Segmentation` label type, which is identical from a data storage standpoint (single per-image mask) but has different semantics in terms of the pixel values representing instances rather than classes.

I'm pointing out https://github.com/voxel51/fiftyone/pull/4500 because, while we could in principle have all three of these:
- `Segmentation.mask_path`
- `Detections.mask_path`
- `Detection.mask_path`

the added representational flexibility incurs extra complexity in making the rest of `fiftyone` work (in-App visualization, import/export, additional complexities to support cloud-backed datasets in FiftyOne Teams, etc).

Laurent2916:
Hi @brimoor!

Thanks for the suggestions, I've applied them.

I'm working with ~1000x1000px masks on average. I have about 30k images for now, with each image containing 2-4 masks. This means that currently, masks are responsible for a good chunk of the disk usage of my mongo db.

The masks I have aren't "non-overlapping", they are inferred from various segmentation models and thus overlap most of the time (they are also non-binary/grayscale, even though the UI doesn't support displaying this). So using `Detections.mask_path` doesn't quite fit my use-case I'm afraid.

Laurent2916:
Hi @brimoor, when you get a chance, could you please take a look again at this PR and let me know your thoughts ?

sashankaryal:
Hi @Laurent2916, I cherry-picked your commits in #5120

sashankaryal:
Closing this PR in favor of #5120

PR #4698, opened 2024-08-18
Update indexes when renaming/deleting fields

brimoor:
When renaming and deleting fields, any indexes that touch the affected fields should also be updated, just like how `dataset.app_config` is automatically synchronized.

Note: unfortunately MongoDB does not support renaming indexes; they must be dropped and recreated

### DIFF
--- a/fiftyone/core/odm/mixins.py
+++ b/fiftyone/core/odm/mixins.py
@@ -6,6 +6,7 @@ Mixins and helpers for dataset backing documents.
 |
 """
 from collections import OrderedDict
+import itertools
 
 from bson import ObjectId
 from pymongo import UpdateOne
@@ -469,11 +470,6 @@ class DatasetMixin(object):
             if field is not None:
                 schema_paths.append((path, new_path))
 
-        # This fixes https://github.com/voxel51/fiftyone/issues/3185
-        # @todo improve list field updates in general so this isn't necessary
-        if schema_paths:
-            cls._reload_fields()
-
         if simple_paths:
             _paths, _new_paths = zip(*simple_paths)
             cls._rename_fields_simple(_paths, _new_paths)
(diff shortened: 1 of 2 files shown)


fiftyone/core/odm/mixins.py
```diff
@@ -484,6 +480,11 @@ def _rename_fields(cls, sample_collection, paths, new_paths):
                 sample_collection, _paths, _new_paths
             )
 
+        # This fixes https://github.com/voxel51/fiftyone/issues/3185
+        # @todo improve list field updates in general so this isn't necessary
+        if schema_paths:
```
brimoor, line 485: Previously `_reload_fields()` was called earlier, before issuing potentially long-running aggregations to rename the fields in the database. Now the reload happens later, immediately before the dataset doc's schema is updated, which should reduce potential errors due to concurrent dataset edits.

fiftyone/core/odm/mixins.py
```diff
@@ -497,6 +498,9 @@ def _rename_fields(cls, sample_collection, paths, new_paths):
         dataset_doc.app_config._rename_paths(paths, new_paths)
         dataset_doc.save()
 
+        if schema_paths:
+            cls._rename_indexes(paths, new_paths)
```
brimoor, line 502: Index updates are performed at the very end so that, if the dataset is big and the user ctrl + c's the operation sometime while the index operations are being performed, the data model changes will at least be finished.

fiftyone/core/odm/mixins.py
```diff
@@ -1118,6 +1125,21 @@ def _delete_field_schema(cls, path):
         doc._undeclare_field(field_name)
         _delete_field_doc(field_docs, field_name)
 
+    @classmethod
+    def _rename_indexes(cls, paths, new_paths):
+        updates = _get_index_updates(cls._dataset, paths, new_paths=new_paths)
+
+        for name, new_index_spec in updates.items():
+            cls._dataset.drop_index(name)
```
minhtuev, line 1133: A little bit sad that we have to do this instead of renaming the actual index datastructure, but this is another datapoint for @ritch ; in the future field changes (adding/renaming/deleting) can be treated as db migration that can be scheduled to run overnight or at a specified time to minimize disruption.

minhtuev (approved):
Looking good, thanks Brian!

PR #4709, opened 2024-08-20
Imavid/replay

CamronStaley:

Fix two bugs in imavid:
- When clicking play on a finished video it doesn't replay from the start
- When selecting a frame in one video, opening another will resume at that frame even if it's outside the range of that video. This would cause the UI to break

1. Use any image dataset (I used the demo dataset)
2. Select group by
3. Group by some fields ordered
4. Select display as video in settings
5. Test samples

        notes for FiftyOne users.

Fix bugs in images as video for grouping. Now you can press play to restart the video and if you pause on a specific frame and select a different sample you will be put back to the first frame.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced play/pause control to manage multiple video state types, improving playback flexibility.
	- Improved logic for pausing video playback to ensure the first frame displays without interruption.
	- Added control for looping behavior during playback, enhancing the user experience.

- **Bug Fixes**
	- Resolved playback behavior to ensure seamless restarts when reaching the end of the video.

- **Documentation**
	- Updated method descriptions to reflect changes in functionality and initialization logic for improved user clarity.

### DIFF
--- a/app/packages/looker/src/elements/common/actions.ts
+++ b/app/packages/looker/src/elements/common/actions.ts
@@ -609,7 +609,6 @@ const videoEscape: Control<VideoState> = {
         options: { showHelp, showJSON, selectedLabels },
         lockedToSupport,
       }) => {
-        console.log("closing window");
         if (showHelp) {
           dispatchEvent("panels", { showHelp: "close" });
           return { showHelp: "close" };
@@ -640,7 +639,7 @@ const videoEscape: Control<VideoState> = {
 
         if (frameNumber !== 1) {
           return {
-            frameNumber: 1, //THIS IS WHERE THE ISSUE IS WITH CLOSING FRAME
+            frameNumber: 1,
             playing: false,
           };
         }
--- a/app/packages/looker/src/elements/common/actions.ts
+++ b/app/packages/looker/src/elements/common/actions.ts
@@ -459,41 +459,55 @@ export const previousFrame: Control<VideoState | ImaVidState> = {
   },
 };
 
-export const playPause: Control<VideoState> = {
+export const playPause: Control<VideoState | ImaVidState> = {
   title: "Play / pause",
   shortcut: "Space",
   eventKeys: " ",
   detail: "Play or pause the video",
   action: (update, dispatchEvent) => {
-    update(
-      ({
-        frameNumber,
-        playing,
-        duration,
-        config: { frameRate, support, thumbnail },
-        lockedToSupport,
-      }) => {
-        if (thumbnail) {
-          return {};
-        }
-        const start = lockedToSupport ? support[0] : 1;
-        const end = lockedToSupport
-          ? support[1]
-          : getFrameNumber(duration, duration, frameRate);
+    update((state: ImaVidState | VideoState) => {
+      if (state.config.thumbnail) {
+        return {};
+      }
+      dispatchEvent("options", { showJSON: false });
 
-        dispatchEvent("options", { showJSON: false });
+      const isImaVid = (state as ImaVidState).currentFrameNumber;
+      if (isImaVid) {
+        const {
+          currentFrameNumber,
+          playing,
+          config: { frameStoreController },
+        } = state as ImaVidState;
+        const reachedEnd =
+          currentFrameNumber >= frameStoreController.totalFrameCount;
         return {
-          playing: !playing && start !== end,
-          frameNumber:
-            end === frameNumber
-              ? lockedToSupport
-                ? support[0]
-                : 1
-              : frameNumber,
+          currentFrameNumber: reachedEnd ? 1 : currentFrameNumber,
           options: { showJSON: false },
+          playing: !playing && !reachedEnd,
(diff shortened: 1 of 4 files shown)


app/packages/looker/src/elements/common/actions.ts
```diff
@@ -459,41 +459,56 @@ export const previousFrame: Control<VideoState | ImaVidState> = {
...
-                : 1
-              : frameNumber,
+          currentFrameNumber: reachedEnd ? 1 : currentFrameNumber,
           options: { showJSON: false },
+          playing: !playing || reachedEnd,
         };
       }
-    );
+
+      //seperate imavid from video state
```
sashankaryal, line 491: nit: typo: separate, also might want to hoist this comment before `if (isImaVid)`

sashankaryal (commented):
Looks good! 🚀 

I noticed one small bug though, curious if you can reproduce it. If loop video is on, frame number goes to 1 -> 2 -> 1 -> 2 -> 3 ....

I've attached a gif.

imavid-bug-restart-1-small

benjaminpkane:
Can you check behavior for the base cases of 1 frame and 2 frames? I believe play/pause should be disabled for 1 frame. Maybe an e2e spec for those two scenarios? That would give us confidence in this fix 🚀  😄

CamronStaley:
@sashankaryal seems to be an issue in demo environment as well so probably unrelated to my changes. Going to get a fix for it though since agreed it's not a great experience.

@benjaminpkane we talked offline but for documentation purposes the question you were asking is what happens for imavids that are just a single frame when pressing play. In that case the play button does nothing and the code behind why that happens is here:
- [playButton sets current frame back to 1 if we have reached the end which is always true for single frame imavids](https://github.com/voxel51/fiftyone/blob/864a51863f28ef060351146dbad84c9c40f6aff5/app/packages/looker/src/elements/imavid/play-button.ts#L24)
- [drawFrame attempts to retrieve the current sample again from the frame store but because the frame count is equal to the size of the totalFrame counts we just set pause back to true and do nothing](https://github.com/voxel51/fiftyone/blob/864a51863f28ef060351146dbad84c9c40f6aff5/app/packages/looker/src/elements/imavid/index.ts#L240)

benjaminpkane:
> @sashankaryal seems to be an issue in demo environment as well so probably unrelated to my changes. Going to get a fix for it though since agreed it's not a great experience.
> 
> @benjaminpkane we talked offline but for documentation purposes the question you were asking is what happens for imavids that are just a single frame when pressing play. In that case the play button does nothing and the code behind why that happens is here:
> 
> * [playButton sets current frame back to 1 if we have reached the end which is always true for single frame imavids](https://github.com/voxel51/fiftyone/blob/864a51863f28ef060351146dbad84c9c40f6aff5/app/packages/looker/src/elements/imavid/play-button.ts#L24)
> * [drawFrame attempts to retrieve the current sample again from the frame store but because the frame count is equal to the size of the totalFrame counts we just set pause back to true and do nothing](https://github.com/voxel51/fiftyone/blob/864a51863f28ef060351146dbad84c9c40f6aff5/app/packages/looker/src/elements/imavid/index.ts#L240)

Sounds good. Defer to @sashankaryal for review. Just nudging for consistency between the regular video player and imavid player. For the user :sparkles:

Here is code for a one frame video dataset

```python
from eta.core.video import VideoProcessor
import fiftyone as fo
import fiftyone.zoo as foz

OUT = "/Users/ben/code/scratch/one.mp4"

processor = VideoProcessor(
    dataset.first().filepath,
    out_video_path=OUT,
    frames=[1, 1],
)

with processor:
    for img in processor:
        processor.write(img)


dataset = fo.Dataset("one")
dataset.persistent = True
dataset.add_sample(fo.Sample(OUT))
``` 

Screenshot 2024-08-21 at 2 51 11 PM

app/packages/looker/src/elements/imavid/index.ts
```diff
@@ -227,7 +230,7 @@ export class ImaVidElement extends BaseElement<ImaVidState, HTMLImageElement> {
     // if abs(frameNumberToDraw, currentFrameNumber) > 1, then skip
     // this is to avoid drawing frames that are too far apart
     // this can happen when user is scrubbing through the video
-    if (Math.abs(frameNumberToDraw - this.frameNumber) > 1) {
+    if (Math.abs(frameNumberToDraw - this.frameNumber) > 1 && !this.isLoop) {
```
CamronStaley, line 233: this seems to fix the looping issue. For some reason when looping at lower speeds specifically this method would get triggered causing it to "buffer" and then try showing a frame again. This would cause that weird jumping between 2 and 3. Disabling this functionality while looping seems to fix the problem without any side-effects from what I can tell testing locally.

imanjra:
Works great! Super minor suggestion (may ignore): thoughts on replacing the play button icon with the replay icon?

sashankaryal (approved):
looks good! 🚀

PR #4710, opened 2024-08-20
fix looker3d bugs

sashankaryal:

This PR fixes bugs in fo3d navigation and fo3d status logs.

E2E.

## Summary by CodeRabbit

- **New Features**
	- Enhanced error logging functionality to utilize modal sample IDs.
	- Improved rendering logic in the UI to streamline loading states and conditionally display components based on the initialization of the scene.
	- Introduced a new selector for more granular access to asset loading logs associated with individual samples.

- **Bug Fixes**
	- Adjusted component dependencies to ensure correct sample context is used across various components.

- **Style**
	- Updated CSS for the Container to prevent horizontal overflow, enhancing layout responsiveness.

- **Tests**
	- Expanded test suite to include modal sidebar interactions and multi-sample rendering capabilities for comprehensive validation.


app/packages/looker-3d/src/ErrorBoundary.tsx
```diff
@@ -15,7 +15,7 @@ const AddFo3dErrorLogs = ({
   error: Error;
   boundaryName?: string;
 }) => {
-  const thisSampleId = useRecoilValue(currentSampleId);
+  const thisSampleId = useRecoilValue(modalSampleId);
```
benjaminpkane, line 18: In group datasets, I don't believe the `modalSampleId` necessarily represents a 3D sample. Does `looker-3d` depend on this id beyond key usage?

sashankaryal: This keying makes sense here for 3d status logs. We just need something "unique" about the current view. `currentSampleId` wouldn't work since it'd be null when something else is pinned.

benjaminpkane: Ok,  just noting `@fiftyone/looker-3d` remains a singleton library until we address this

benjaminpkane (approved):
LGTM 🚀

benjaminpkane:
I would check group datasets before merging. A `${sampleId}-${groupId}` might be required, e.g. with 

```python
import fiftyone as fo
import fiftyone.zoo as foz

session = fo.launch_app(dataset, group_id=dataset.first().group.id)
```

app/packages/looker-3d/src/ErrorBoundary.tsx
sashankaryal: Good point.

imanjra (approved):
LGTM

Probably out of scope and unrelated, but here is small things I found from local testing:
- The configuration position (intermittent)
image
- Would be nice to add an icon for sample panel
image
- would be nice to make this scrollable
image

benjaminpkane (approved):

benjaminpkane (approved):
🚀

PR #4712, opened 2024-08-22
fix: wrong argument description in `extract_patch` function

imyhxy:
## What changes are proposed in this pull request?

The description of the `alpha` argument in the `extract_patch` function is misleading. `alpha` greater than 0 will expand the box, less than 0 will contract the box. So, setting `alpha` to 1.1 will actually expand the box by 110%.

## How is this patch tested? If it is not, please explain why.

Just modify docs, no test required.

### Is this a user-facing change that should be mentioned in the release notes?

-   [x] No. You can skip the rest of this section.
-   [ ] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [x] Documentation: FiftyOne documentation changes
-   [ ] Other

## Summary by CodeRabbit

- **Documentation**
	- Updated the documentation for the `alpha` parameter in the `extract_patch` function to clarify the correct values for expanding and contracting the box size.

PR #4713, opened 2024-08-22
Fixing alpha docs

brimoor:
Applies the fix ID'd in https://github.com/voxel51/fiftyone/pull/4712 in more places.

## Summary by CodeRabbit

- **Documentation**
	- Updated examples for the `alpha` parameter across multiple components to clarify its usage for box expansion and contraction.


benjaminpkane (approved):
🧹

PR #4718, opened 2024-08-23
fix bug where obj doesn't load if custom mtl isn't provided

CamronStaley:

When loading a 3d sample with cloud backed media that doesn't have a custom mtl we are not able to load the asset because we aren't first checking if a preTrasformedObjPath is present. These changes fix that.

(Details)

        notes for FiftyOne users.

When using cloud backed media that don't contain a custom mtl path objects should be able to load properly.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced a new property, `preTransformedObjPath`, for enhanced flexibility in loading object paths.
	- Improved loading process by allowing the use of a pre-transformed object path when available. 

- **Bug Fixes**
	- Updated dependency tracking to ensure proper recalculations of memoized values based on the new property.

### DIFF
--- a/app/packages/looker-3d/src/fo3d/mesh/Obj.tsx
+++ b/app/packages/looker-3d/src/fo3d/mesh/Obj.tsx
@@ -24,13 +24,15 @@ const ObjMeshDefaultMaterial = ({
   obj: ObjAsset;
   onLoad?: () => void;
 }) => {
-  const { objPath } = obj;
+  const { objPath, preTransformedObjPath } = obj;
 
   const { fo3dRoot } = useFo3dContext();
 
   const objUrl = useMemo(
-    () => getSampleSrc(getResolvedUrlForFo3dAsset(objPath, fo3dRoot)),
-    [objPath, fo3dRoot]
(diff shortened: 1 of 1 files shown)


app/packages/looker-3d/src/fo3d/mesh/Obj.tsx
```diff
@@ -24,13 +24,15 @@ const ObjMeshDefaultMaterial = ({
...
 
   const { fo3dRoot } = useFo3dContext();
 
   const objUrl = useMemo(
-    () => getSampleSrc(getResolvedUrlForFo3dAsset(objPath, fo3dRoot)),
-    [objPath, fo3dRoot]
+    () =>
+      preTransformedObjPath ??
+      getSampleSrc(getResolvedUrlForFo3dAsset(objPath, fo3dRoot)),
+    [objPath, preTransformedObjPath, fo3dRoot]
```
sashankaryal, line 35: good catch

sashankaryal (approved):
LGTM!

PR #4569, opened 2024-07-12
Add track ids if present for ultralytics models 

Rusteam:
## What changes are proposed in this pull request?

If running `track` method from `ultralytics.YOLO` model, then track ids will be available video files.
If track ids are present, we add then to `fiftyone.Detection(index=track_id)` and other labels.

## How is this patch tested? If it is not, please explain why.

Works with detection and pose estimation models for the ultralytics integration.

### Is this a user-facing change that should be mentioned in the release notes?

-   [ ] No. You can skip the rest of this section.
-   [x] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

When running ultralytics models on video files, track ids will be internally added to label fields with a instance index.

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [x] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Enhanced tracking capabilities by including track IDs in detection, instance, polyline, and keypoint results.

PR #4720, opened 2024-08-23
Fix ultralytics extract track IDs

swheaton:

[This PR](https://github.com/voxel51/fiftyone/pull/4569) broke ultralytics models when track ID isn't present. Though I don't know what that even is TBH.

Notably, there are no automated tests. I'm not going to add them now.

```python
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.huggingface as fouh
dataset = fouh.load_from_hub(
    "Voxel51/VisDrone2019-DET",
    max_samples=25,
    overwrite=True
    )
model = foz.load_zoo_model(
    "yolov8l-coco-torch",
    device="cuda"
    )
```

        notes for FiftyOne users.

Fix ultralytics models without track ID

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of non-tracking cases, ensuring the system returns appropriate values based on the number of boxes detected.

### DIFF
--- a/fiftyone/utils/ultralytics.py
+++ b/fiftyone/utils/ultralytics.py
@@ -58,7 +58,7 @@ def _extract_track_ids(result):
     return (
         result.boxes.id.detach().cpu().numpy().astype(int)
         if result.boxes.is_track
-        else [None] * len(result.boxes.conf.size(0))
+        else [None] * result.boxes.conf.size(0)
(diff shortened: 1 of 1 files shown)


kaixi-wang (approved):
had the same fix but you got it first lol

PR #4729, opened 2024-08-26
fix copied doc updates not insert

swheaton:
## What changes are proposed in this pull request?

Another weird bug that doesn't cause issues really until to-come changes are introduced.

If you do this code (done 3 times with dataset clone)
```python
doc_copy = doc.copy()
_id = bson.ObjectId()
doc_copy.id = _id
```
Then `mongoengine` sets `_created` to `False` which means it thinks it's an update object not a new one.
When you call `save()` it emits an upsert call instead of an insert.
OK, not so bad, maybe a little weird but whatever ...

But in https://github.com/voxel51/fiftyone/pull/4597 @brimoor proposes an optimization where only changed fields are serialized to get the document. In combo, this causes all kinds of strange behavior. It just so happens to work in that PR because `doc._changed_fields` is uninitialized (a code smell in mongoengine, they have a [TODO](https://github.com/MongoEngine/mongoengine/blob/master/mongoengine/base/document.py#L40) to clean it up...) and so `doc._delta()` returns the whole doc.

But say we cleared changed fields because we didn't know about this strange requirement.
```python
import bson
import fiftyone.core.odm as foo

run_doc = foo.RunDocument(config={"foo": "bar"})
run_doc.save()
doc_copy = run_doc.copy()
doc_copy.id = bson.ObjectId()

doc_copy._clear_changed_fields()
doc_copy.version = "51.51"
print(doc_copy._get_changed_fields()  # ["version"]
doc_copy.save(upsert=True)
doc_copy.reload()

# Oops our config field is {} because it's not a changed field so update only wrote "version" field
assert doc_copy.config == {"foo": "bar"}

doc_copy.delete()
run_doc.delete()
```

Ok but we don't clear changed_field so we're fine? Nope, balancing on a thread due to another mongoengine weirdness where `_get_changed_fields()` can return something due to embedded documents being edited

```python
import bson
import fiftyone as fo

ds = fo.Dataset()

# Pretending to clone the dataset doc
doc_copy = ds._doc.copy()
doc_copy.id = bson.ObjectId()
doc_copy.sample_collection_name=f"samples.{str(doc_copy.id)}"
doc_copy.name="blah"
doc_copy.slug="blah"

# Making an embedded document update
doc_copy.sample_fields[0].description = "blah"

# Wha? changed fields is actually ["sample_fields.0.description"] because
#  simple fields aren't tracked but embedded docs are
assert doc_copy._get_changed_fields() == []

ds.delete()
doc_copy.delete()
```

[...]

PR #4730, opened 2024-08-26
Fields have created_at attribute

swheaton:

Adds `created_at` property to all `fo.Field`s. Default is `None` but it is populated on newly-created fields.
This is only stored in the dataset doc `sample_field` and `frame_field` embedded document list fields, so the footprint is small.

**Note**: `created_at` is not propagated through to embedded document fields. Could add this on later if desired but for now the feature achieved what I needed without doing so.

Depends on branch `last-modified-at` so must be merged there.
Also depends on some other bug fixes such as https://github.com/voxel51/fiftyone/pull/4729. But that was copied into the commit. Might just have merge conflicts.

Added automated tests checking field created_at is only updated when the field is new.

```python
from datetime import datetime
import time
import fiftyone as fo

def _assert_equalish(t1, t2):
  assert (t2-t1).total_seconds() < 0.1

now = datetime.utcnow()
ds = fo.Dataset()

for field in ds.get_field_schema().values():
  _assert_equalish(now, field.created_at)

time.sleep(1)
ds.add_sample_field("new_field", fo.StringField)
assert (ds.get_field("new_field").created_at - now).total_seconds() > 1
```

        notes for FiftyOne users.

Added `created_at` property to every field.

-   [ ] Other

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -487,6 +487,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
                 "metadata",
                 fof.EmbeddedDocumentField,
                 embedded_doc_type=doc_type,
+                created_at=datetime.utcnow(),
             )
             field_doc = foo.SampleFieldDocument.from_field(field)
             self._doc.sample_fields[idx] = field_doc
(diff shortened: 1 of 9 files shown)


brimoor (commented):
Looking good! I like it 💪 

I'll give it a pass to make sure all cases are covered. Cloning datasets/views, merging fields, renaming/cloning fields, etc.

brimoor:
Excellent, seems like you covered all bases 💪

PR #4733, opened 2024-08-27
Draft: Adding Med-SAM2 model and Med-SAM2 dataset to HF Collection

evatt-harvey-salinger:

This PR simply adds the 3D Med-SAM2 model to the model zoo. This model is a fine-tuning of SAM2 hiera-tiny, published on [Hugging Face](https://huggingface.co/datasets/jiayuanz3/REFUGE/tree/main) and the [Med-SAM2 Github repo](https://github.com/MedicineToken/Medical-SAM2) along with their paper: [Medical SAM 2: Segment Medical Images As Video Via Segment Anything Model 2](https://arxiv.org/abs/2408.00874).

Simultaneously, I have collected the BTCV imagery used to train Med-SAM2 and converted it into a Fiftyone video dataset. This dataset has been (is being) [uploaded](https://huggingface.co/datasets/Voxel51/BTCV-CT-as-video-MedSAM2-dataset/tree/main) to the Hugging Face Hub into the Voxel51 organization.

Soon, I will add a tutorial that guides users to load the BTCV dataset, load the Med-SAM2 model, and apply the model onto the test set of the dataset. The basic recipe is laid out below.

You must add the [SAM2](https://github.com/facebookresearch/segment-anything-2) package to your environment.
```python
import os
import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone.utils.huggingface import load_from_hub

dataset = load_from_hub("Voxel51/BTCV-CT-as-video-MedSAM2-dataset")

# Form the prompt dataset (recommended settings with a GPU)
prompt_dataset = form_prompt_dataset(
    dataset.limit(5), 
    prompt_frames_fraction=0.25, 
    # max_prompt_frames=None,
    # prompt_organs=None,
)

# Form the prompt dataset (recommended settings without a GPU)
prompt_dataset = form_prompt_dataset(
    dataset.limit(2), 
    # prompt_frames_fraction=0.25, 
    max_prompt_frames=10,
    prompt_organs=["liver"],
)

model = foz.load_zoo_model("MedSAM2-video-torch")

# Prompt with boxes
prompt_dataset.apply_model(
    model,
    label_field="pred_segmentations",
    prompt_field="frames.gt_detections",
)
session = fo.launch_app(prompt_dataset)
```

```python
def form_prompt_dataset(
    dataset, prompt_frames_fraction=0.25, max_prompt_frames=None, prompt_organs=None, seed=51
):
    """
    Create a prompt dataset by selecting a subset of frames and organs from the input dataset.

    To recreate the MedSAM2 paper use the default values:
    - prompt_frames_fraction=0.25
    - max_prompt_frames=None

    Args:
        dataset: The input FiftyOne dataset.
        prompt_frames_fraction (float, optional): The fraction of frames to retain in the prompt dataset (0-1).
        max_prompt_frames (int, optional): The maximum number of frames to include in the prompt dataset.
        prompt_organs (list, optional): A list of organs to filter the prompt dataset by.

    Returns:
        FiftyOne dataset: The prompt dataset with selected frames and filtered labels.
    """
    import random
    random.seed(seed)
    
    prompt_dataset = dataset.clone()
    
    # Get the frame IDs from the prompt dataset 
    # (a nested list of frame IDs per video sample)
    frame_ids = prompt_dataset.values("frames.id")
    
    frame_ids_to_delete = []

    for sample_frame_ids in frame_ids:
        random.shuffle(sample_frame_ids)
        
        num_prompt_frames = len(sample_frame_ids)

        if prompt_frames_fraction is not None:
            num_prompt_frames = int(len(sample_frame_ids) * prompt_frames_fraction)

        # Limit the number of prompt frames to the provided maximum value
        if max_prompt_frames is not None:
            num_prompt_frames = min(num_prompt_frames, max_prompt_frames)

        # Add the remaining frame IDs to the list of frame IDs to delete
        frame_ids_to_delete.extend(sample_frame_ids[num_prompt_frames:])

    # Delete the selected frames from the prompt dataset and remove their detections
    (
        prompt_dataset
        .select_frames(frame_ids_to_delete)
        .set_field("frames.gt_detections", None)
        .save()
    )
    # Filter the prompt dataset by the specified prompt organs, if provided
    if prompt_organs:
        (
            prompt_dataset
            .filter_labels(
                "frames.gt_detections", 
                filter=F("label").is_in(prompt_organs)
            )
            .save()
        )

    return prompt_dataset
```

Yes! This is a great example of the exciting capabilities of the new SAM2 model, and how it may be fine-tuned for improving performance in specific domains.

        notes for FiftyOne users.

Building upon the recent integration of SAM 2, the specialized CT-scan model Med-SAM2 is now available in the FiftyOne Model Zoo! The dataset used to fine-tune this model is also available on Voxel51's Hugging Face Hub. By treating CT-scan imagery as videos, this model shows the exciting opportunities for fine-tuning SAM2 to specific domains.

-   [ ] Other

### DIFF
--- a/fiftyone/zoo/models/manifest-torch.json
+++ b/fiftyone/zoo/models/manifest-torch.json
@@ -449,6 +449,44 @@
             "date_added": "2024-08-05 14:38:20"
         },
         {
+            "base_name": "MedSAM2-video-torch",
+            "base_filename": "MedSAM2_pretrain.pth",
+            "version": null,
+            "description": "Fine-tuned SAM2-hiera-tiny model from paper: Medical SAM 2 - Segment Medical Images as Video via Segment Anything Model 2 <https://arxiv.org/abs/2408.00874>`_",
+            "source": "https://github.com/MedicineToken/Medical-SAM2",
+            "size_bytes": 155906050,
+            "manager": {
+                "type": "fiftyone.core.models.ModelManager",
+                "config": {
+                    "url": "https://huggingface.co/jiayuanz3/MedSAM2_pretrain/resolve/main/MedSAM2_pretrain.pth?download=true"
+                }
+            },
+            "default_deployment_config_dict": {
+                "type": "fiftyone.utils.sam2.SegmentAnything2VideoModel",
+                "config": {
+                    "entrypoint_fcn": "sam2.build_sam.build_sam2_video_predictor",
+                    "entrypoint_args": { "model_cfg": "sam2_hiera_t.yaml" }
+                }
+            },
+            "requirements": {
+                "packages": ["torch", "torchvision"],
+                "cpu": {
+                    "support": true
+                },
+                "gpu": {
+                    "support": true
+                }
+            },
+            "tags": [
+                "segment-anything",
+                "torch",
+                "zero-shot",
+                "video",
+                "MedSAM-2"
+            ],
+            "date_added": "2024-08-17 14:48:00"
+        },
+        {
             "base_name": "deeplabv3-resnet50-coco-torch",
             "base_filename": "deeplabv3_resnet50_coco-cd0a2569.pth",
             "version": null,


prernadh:
Hey @evatt-harvey-salinger Thanks for adding Med Sam 2 to FiftyOne. Few high level comments- 
1. The detections in the dataset don't have the `index` field populated. Is that information you could add in so that the detections across frames can be linked to one another. 
2. In the current format that it is, it won't be possible to run the code snippet you have in the PR description on CPU since you are trying to run SAM2 by giving it prompts from every single detection that the dataset has. This won't work correctly since you will end up creating multiple object tracking labels for the SAME object since index is not specified. Could you only provide one detection per object (basically only one annotated frame for each object) in each video and then apply the model using these detections as prompts?

prernadh (changes requested):
I have added a PR comment with the changes you can make- let me know if you have any questions.

evatt-harvey-salinger:
Hey @prernadh,

1. Interesting, let me see about adding that in. The original dataset makes no distinction between instances of objects, simply a segmentation map for the entire image. I used `Segmentation.to_detections()` to populated the detections field. An assumption I could make would be that there are only 1 instance of each class in video sample (1 spleen, 1 aorta, 1 right kidney, etc). I'll do some digging to see if that assumption holds. Often, the same organ has to distinct segmentation regions. 
2. Yes, this is what I meant by the "WIP" commend `# Only retain the first detection in each video sample", which I'll update now for clarity. I'm going to filter the detections so that that there is only 1 detection for each class, on the first frame where that organ is present. I'm not sure how to account for the case I mentioned above, where there are two distinct regions arise for the same object...

evatt-harvey-salinger:
Here's an example from img0002.mp4. The "stomach" segmentation region on the right has been visible for a few frames, but then region left pops up out of nowhere. It's eventually absorbed the first "stomach" region, but would be missed by the model I only prompt the first frame
Screenshot 2024-08-27 at 2 18 30 PM

evatt-harvey-salinger:
I think what makes the most sense is to have the right and left region be their own detection with separate indices... then I can send a prompt anytime a new region pops up.

prernadh:
@evatt-harvey-salinger Usually you would address this by adding in a prompt for the frame when the new blob shows up for the "stomach" label. But you would have to add in the `index` field so that the SAM model knows that the two prompts for stomach are referring to the same instance of the object

evatt-harvey-salinger:
Copy, sounds good. I'll upload a new version tonight, and hopefully populate that test snippet above. Thanks @prernadh!

evatt-harvey-salinger:
Alrighty, I have updated the dataset so that each class maintains an index throughout the video. (This is mostly a safe assumption, except for the '"portal vein and splenic vein"', but it's the best that I could do.)

I've also added a code snippet for creating a `prompt_dataset` which only includes detections when there are more instances that in the previous frame. Checkout the video attached to see what that looks like.


https://github.com/user-attachments/assets/58ca4ef9-e50f-433c-b923-8f6240ddcee5

prernadh (approved):
LGTM!

PR #4739, opened 2024-08-27
Fix sample selection with map panel

benjaminpkane:

Sample ID to grid index map should be cleared when the grid is reset

Hook test ✨ 

## Release Notes

* Fixed sample selection when interacting with plugins that use an extended selection

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced a custom hook `useRecords` for enhanced record management.
  - Added unit tests for the `useRecords` hook to ensure reliability.
  - Introduced unit tests for `addRange` and `removeRange` functions in the `useSelectSample` module.

- **Improvements**
  - Streamlined import statements for better clarity and type safety.
  - Enhanced error handling in record retrieval logic.
  - Improved flexibility and efficiency in the `useSpotlightPager` hook by adjusting its parameter structure.


app/packages/core/src/components/Grid/useRecords.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type Records = MutableRefObject<Map<string, number>>;
+
+export default (clear: string) => {
+  const records = useRef(new Map<string, number>());
+
+  useMemo(() => {
+    clear;
+    records.current.clear();
+  }, [clear]);
```
sashankaryal, line 12: What's going on here? I don't get why it's a memo and not an effect.

sashankaryal (changes requested):

app/packages/core/src/components/Grid/useRecords.ts
benjaminpkane: The spotlight instance is a memo, but I suppose an unmount in an effect makes more sense here 👍

benjaminpkane: I think `useMemo` is the right call here. I removed the `useRef` weirdness. Now a new records `Map` is created with each new Spotlight instance. They have the same lifecycle, so I think this makes sense

benjaminpkane:
I think the real issue popped up during cleanup. Range selection needed to omit selections that the grid doesn't know about. Added unit tests to assert behavior

app/packages/core/src/components/Grid/useRecords.ts
sashankaryal: `return useMemo(() => new Map(), [clear])` makes sense to me. Just that previously it was being used like an effect and I couldn't wrap my head around it 😅

sashankaryal (approved):
LGTM! Tests look good, too.

PR #4753, opened 2024-08-29
fix set_data clobbering old values

ritch:

`set_data|state("my.path", 42)` was resulting in:

```
# original state|data
{
  "foo": "bar"
}
# after call
{
  "my": {"path": 42}
}
# when it should be
{
  "foo": "bar",
  "my": {"path": 42}
}
```

This PR fixes the issue.

You can easily test this with following panel:

```python
class CounterExample(foo.Panel):
    @property
    def config(self):
        return foo.PanelConfig(
            name="example_counter",
            label="Examples: Counter",
        )

    def on_load(self, ctx):
        ctx.panel.state.my_count = 0
        ctx.panel.data.info = {"foo": "bar"}

    def increment(self, ctx):
        ctx.panel.state.my_count += 1
        ctx.panel.set_data("info.bat", "baz")

    def decrement(self, ctx):
        ctx.panel.state.my_count -= 1

    def reset(self, ctx):
        ctx.panel.state.my_count = 0

    def say_hi(self, ctx):
        ctx.ops.notify("Hi!")

    def render(self, ctx):
        panel = types.Object()

        # Add menu to panel
        menu = panel.menu("menu", variant="square", color="51")
        menu.btn(
            "increment",
            icon="add",
            label="Increment",
            on_click=self.increment,
        )
        menu.btn(
            "decrement",
            icon="remove",
            label="Decrement",
            on_click=self.decrement,
        )
        menu.btn("reset", icon="360", label="Reset", on_click=self.reset)
        menu.btn(
            "say_hi",
            icon="emoji_people",
            label="Say hi!",
            on_click=self.say_hi,
            variant="contained",
            color="white",
        )

        # Add counter info to panel
        panel.message("my_count", f"Count: {ctx.panel.state.my_count}")
        panel.obj("info", view=types.JSONView())

        return types.Property(
            panel,
            view=types.GridView(
                align_y="center",
                align_x="center",
                orientation="vertical",
                height=100,
            ),
        )
```

## Summary by CodeRabbit

- **Improvements**
	- Streamlined panel state update process for enhanced efficiency and clarity.
- **New Features**
	- Introduced new tests for the `PanelRef` class, covering various scenarios for setting state and data, including deep path strings and write-only property access.

### DIFF
--- a/fiftyone/operators/panel.py
+++ b/fiftyone/operators/panel.py
@@ -212,10 +212,7 @@ class PanelRefState(PanelRefBase):
             value (None): the value, if key is a string
         """
         d = key if isinstance(key, dict) else {key: value}
-
-        for k, v in d.items():
-            super().set(k, v)
-
+        super().set(d)
         self._ctx.ops.patch_panel_state(d)
 
     def clear(self):
@@ -249,10 +246,7 @@ class PanelRefData(PanelRefBase):
             value (None): the value, if key is a string
         """
         d = key if isinstance(key, dict) else {key: value}
-
-        for k, v in d.items():
-            super().set(k, v)
-
+        super().set(d)
         self._ctx.ops.patch_panel_data(d)
 
     def get(self, key, default=None):
--- a/fiftyone/operators/panel.py
+++ b/fiftyone/operators/panel.py
@@ -213,12 +213,10 @@ class PanelRefState(PanelRefBase):
         """
         d = key if isinstance(key, dict) else {key: value}
 
-        args = {}
         for k, v in d.items():
             super().set(k, v)
-            pydash.set_(args, k, v)
 
-        self._ctx.ops.patch_panel_state(args)
+        self._ctx.ops.patch_panel_state(d)
 
     def clear(self):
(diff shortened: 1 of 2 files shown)


fiftyone/operators/panel.py
```diff
@@ -213,12 +213,10 @@ def set(self, key, value=None):
         """
         d = key if isinstance(key, dict) else {key: value}
 
-        args = {}
         for k, v in d.items():
             super().set(k, v)
```
brimoor, line 217: Looks like this can be further simplified as:

```suggestion
        super().set(d)
```

fiftyone/operators/panel.py
```diff
@@ -252,12 +250,10 @@ def set(self, key, value=None):
...
```
brimoor, line 254:
```suggestion
        super().set(d)
```

brimoor (changes requested):
Ah nice catch 😅

brimoor (dismissed):
LGTM

brimoor (approved):

PR #4763, opened 2024-09-03
Improve colorscheme ux

lanzhenw:

- Added validation to ColorScheme class, so users can get error message instantly when argument input is not valid. 
- Improve the colorscheme documentation
- Improve UX of setting colorscale and mask target colors in the UI by pruning invalid values
- Fixed a bug of creating continuous colors from unsorted colorscale list
- Segmentation: the intTarget is no longer limited between 1 and 255. It can be any non-negative integer

example:
Screenshot 2024-09-04 at 6 40 49 PM

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Added `color_by` parameter to dynamically adjust color schemes based on dataset values.
	- Introduced enhanced input validation and error handling for user inputs in various components.
	- Improved flexibility of color scale definitions in the `Colorscale` component.
	- Updated user interface to clarify requirements for custom colorscales.
	- Removed input limits in several components, allowing for a broader range of user inputs.

- **Bug Fixes**
	- Improved clarity and usability of color scheme configuration by specifying constraints and formatting requirements.

- **Documentation**
	- Enhanced user guide with detailed notes on configuring color schemes, including updates on the `default_colorscale` and removal of redundant notes.


lanzhenw:
@coderabbitai review

docs/source/user_guide/app.rst
```diff
@@ -1056,6 +1056,16 @@ You can also programmatically configure a session's color scheme by creating
     Refer to the |ColorScheme| class for documentation of the available
     customization options.
 
+.. note::
+
+    For maskTargetsColors, intTarget must be an integer between 1 and 255.
```
brimoor, line 1061: Can we instead update the frontend to allow targets > 255? The backend should support it and that was the intent as the user could have more than 255 classes in their dataset

docs/source/user_guide/app.rst
```diff
@@ -1056,6 +1056,16 @@ You can also programmatically configure a session's color scheme by creating
...
     customization options.
 
+.. note::
+
+    For maskTargetsColors, intTarget must be an integer between 1 and 255. 
+    Do not set color for intTarget 0, as it is reserved. 
+
+.. note::
+
+    For colorscales, the list of colors must be sorted by value in ascending
```
brimoor, line 1066: Can we instead update the frontend so it sorts by `value`? Should be a simple improvement right?

brimoor (changes requested):

docs/source/user_guide/app.rst
lanzhenw: Yeah, I would remove the FE constraints on > 255. Will only prune 0 and negative number.

docs/source/user_guide/app.rst
lanzhenw: The FE already does it, but I can put the sort in the get_colormap code. And will add more validation to the class to let users know if input is not valid.

brimoor: Okay great, definite +1 to adding the sorting logic to `get_colormap()`.

docs/source/user_guide/app.rst
brimoor: To clarify, are you proposing the following?
- Add validation when saving color schemes so that invalid things like negative/0 will raise an error?
- If, for some reason, bad data got into a color scheme, ignore it in the frontend? IE ignore negative/0 values?

lanzhenw: The confusion is mostly from SDK users, so I am proposing to: 
- add validation methods when we initiate the `ColorScheme` Class, so when a user enter bad input in the python SDK, they can see error messages immediately.
- in the FE, do not allow new bad inputs. For existing bad inputs, still display the value as is, but display a small warning text.

docs/source/user_guide/app.rst
lanzhenw: Just tested with all types of inputs. Plotly now already handle unsorted list and more than rgb colors. We don't need to change any code. I will remove this note from documentation.

fiftyone/core/odm/dataset.py
```diff
@@ -315,6 +316,137 @@ def _id(self):
     def _id(self, value):
         self.id = str(value)
 
+    def _validate(self):
```
brimoor, line 319: @lanzhenw I changed this to a private `_validate()` method because if we were to implement it as `clean()`, it would cause it to become impossible to load an existing dataset that happens to have an invalid color scheme.

The only backwards compatible thing I can think of to do is to tell users to call `dataset.app_config.color_scheme._validate()` if they want to know if their color scheme is valid or not.

fiftyone/core/odm/dataset.py
```diff
@@ -250,14 +251,18 @@ class ColorScheme(EmbeddedDocument):
...
         default_mask_targets_colors (None): a list of dicts with the following
             keys specifying index and color for 2D masks of the dataset. If a
             field does not have field specific mask targets colors, this list
             will be used:
 
-            -   ``intTarget``: integer target value
+            -   ``intTarget``: an integer target value
             -   ``color``: a color string
+
+            Note that the pixel value ``0`` is a reserved "background" class
```
brimoor, line 264: @lanzhenw if this is not already the case, can you update the frontend so that if maskTargets are provided that contain `target=0`, it will simply ignore the value rather than crashing?

Here's how this is documented for `Segmentation` fields:
Screenshot 2024-09-11 at 11 26 35 PM

Note that it doesn't say that `target=0` is not allowed. It just says that this is a special value that is always rendered as invisible in the App, which means that if the user provided `0: "background"` in their mask targets, then they won't see `background` in the tooltip when hovering over 0 pixels in the App.

The consistent behavior for color schemes would be to silently ignore a color if it is provided for pixel value 0.

brimoor (approved):
The backend aspects of this LGTM to me now, but I need to defer to @benjaminpkane for review of the frontend aspects.

Also seeking confirmation from @lanzhenw that the App will indeed gracefully ignore `target=0` mask targets if provided.

app/packages/core/src/components/ColorModal/colorPalette/Colorscale.tsx
```diff
@@ -284,5 +293,5 @@ export default Colorscale;
 
 const validateFloat = (n: number) => {
```
benjaminpkane, line 294: nit: this is more specific than float validation, I believe `validateUnitInterval` may be an apt name

app/packages/core/src/components/ColorModal/controls/ManualColorScaleList.tsx
```diff
@@ -148,6 +148,10 @@ const ManualColorScaleList: React.FC<ManualColorScaleListProp> = ({
     [input, values, onSyncUpdate]
   );
 
+  const onValidateNumber = useCallback((value: number | string) => {
```
benjaminpkane, line 151: Similar to above
```suggestion
  const onValidateUnitInterval = useCallback((value: number | string) => {
```

app/packages/core/src/components/Common/Input.tsx
```diff
@@ -150,58 +156,87 @@ export const NumberInput = React.memo(
       const theme = useTheme();
       color = color ?? theme.primary.plainColor;
       const display = [null, undefined].includes(value) ? "" : Number(value);
+      const [error, setError] = useState<string | null>(null);
+      let errorMsg: string[] | string = [];
+      if (min || min === 0) {
```
benjaminpkane, line 161:
```suggestion
      if (min >= 0) {
```

app/packages/core/src/components/Common/Input.tsx
```diff
@@ -150,58 +156,87 @@ export const NumberInput = React.memo(
...
+        errorMsg.push(`Min: ${min}.`);
+      }
+      if (max || max === 0) {
```
benjaminpkane, line 164:
```suggestion
      if (max >= 0) {
```

app/packages/core/src/components/Common/Input.tsx
lanzhenw: This is meant to be used for all number types (not just positive numbers). I will change to `typeof(min) == 'number` to be more concise.

fiftyone/core/odm/dataset.py
lanzhenw: It's currently not the case. App crashes when intTarget 0 color is set. Working on it.

app/packages/core/src/components/ColorModal/colorPalette/GlobalColorscale.tsx
```diff
@@ -77,6 +77,14 @@ const GlobalColorscale: React.FC = () => {
...
           (a, b) => (a.value as number) - (b.value as number)
         );
+        // a valid list must include 0 and 1
+        // if not, we will insert the missing value
+        if (list[0].value !== 0) {
+          list.unshift({ value: 0, color: list[0].color });
+        }
+        if (list[list.length - 1].value !== 1) {
+          list.push({ value: 1, color: list[list.length - 1].color });
+        }
```
benjaminpkane, line 87: Nice!

benjaminpkane (approved):

PR #4765, opened 2024-09-03
Add support for summary fields

minhtuev:

Adds support for creating and managing summary fields on datasets.

* Manual tests (✅)
* Unit tests (✅)

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.set_field("frames.detections.detections.confidence", F.rand()).save()

# Generate a summary field for object labels
dataset.create_summary_field("frames.detections.detections.label")

# Generate a summary field for [min, max] confidences
dataset.create_summary_field("frames.detections.detections.confidence")

# Generate a summary field for object labels and counts
dataset.create_summary_field(
    "frames.detections.detections.label",
    field_name="frames_detections_label2",
    include_counts=True,
)

# Generate a summary field for per-label [min, max] confidences
dataset.create_summary_field(
    "frames.detections.detections.confidence",
    field_name="frames_detections_confidence2",
    group_by="label",
)

# List the summary fields on the dataset
fo.pprint(dataset.list_summary_fields())

# Inspect contents of these summary fields
fo.pprint(dataset.values("frames_detections_label"))
fo.pprint(dataset.values("frames_detections_confidence"))
fo.pprint(dataset.values("frames_detections_label2"))
fo.pprint(dataset.values("frames_detections_confidence2"))

# Verify that newly created summary fields don't needed updating
fo.pprint(dataset.check_summary_fields())

# Modify the source field
label_upper = F("label").upper()
dataset.set_field("frames.detections.detections.label", label_upper).save()

# Verify that the summary fields now need updating
update_fields = dataset.check_summary_fields()
fo.pprint(update_fields)

# Update the summary fields
for field_name in update_fields:
    dataset.update_summary_field(field_name)

# Summary fields are now updated and no longer need updating
fo.pprint(dataset.check_summary_fields())

# Delete summary fields
dataset.delete_summary_fields(dataset.list_summary_fields())
fo.pprint(dataset.list_summary_fields())
```

## Summary by CodeRabbit

- **New Features**
	- Added methods for managing summary fields in datasets, including creation, listing, updating, and deletion.
	- Introduced optional parameters for enhanced filtering in schema retrieval functions.
	- Improved tools for summarizing and indexing data, particularly for video datasets.
	- Enhanced user interface for managing sidebar groups and filtering sample fields.

- **Bug Fixes**
	- Streamlined error handling in the `drop_index` method.

- **Tests**
	- Added comprehensive tests for frame summary functionalities and summary fields to ensure accurate data handling and reporting.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -9482,13 +9482,14 @@ class SampleCollection(object):
 
         return name
 
-    def drop_index(self, field_or_name):
+    def drop_index(self, field_or_name, quiet=False):
         """Drops the index for the given field or name.
 
         Args:
             field_or_name: a field name, ``embedded.field.name``, or compound
                 index name. Use :meth:`list_indexes` to see the available
                 indexes
+            quiet (booL): whether to raise an error if no index exists
         """
         name, is_frame_index = self._handle_frame_field(field_or_name)
 
@@ -9516,10 +9517,12 @@ class SampleCollection(object):
                 index_map[key] = key
 
         if name not in index_map:
-            itype = "frame index" if is_frame_index else "index"
-            raise ValueError(
-                "%s has no %s '%s'" % (self.__class__.__name__, itype, name)
-            )
+            if not quiet:
+                itype = "frame index" if is_frame_index else "index"
+                msg = "%s has no %s '%s'" % (self.__class__.__name__, itype, name)
+                raise ValueError(msg)
+            else:
+                return
 
         coll.drop_index(index_map[name])
 
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1423,11 +1423,9 @@ class SampleCollection(object):
             read_only (None): whether to restrict to (True) or exclude (False)
                 read-only fields. By default, all fields are included
             info_keys (None): an optional key or list of keys that must be in
-                a field's ``info`` dict in order for it to be included in the
-                returned schema. If ``None``, no filtering is performed.
-            created_after (None): an optional ``datetime`` to filter the
-                returned schema by, such that the field was `created_at` after
-                this time. If ``None``, no filtering is performed.
+                the field's ``info`` dict
+            created_after (None): an optional ``datetime`` specifying a minimum
+                creation date
             include_private (False): whether to include fields that start with
                 ``_`` in the returned schema
             flat (False): whether to return a flattened schema where all
@@ -1469,11 +1467,9 @@ class SampleCollection(object):
             read_only (None): whether to restrict to (True) or exclude (False)
                 read-only fields. By default, all fields are included
             info_keys (None): an optional key or list of keys that must be in
-                a field's ``info`` dict in order for it to be included in the
-                returned schema. If ``None``, no filtering is performed.
-            created_after (None): an optional ``datetime`` to filter the
-                returned schema by, such that the field was `created_at` after
-                this time. If ``None``, no filtering is performed.
+                the field's ``info`` dict
+            created_after (None): an optional ``datetime`` specifying a minimum
+                creation date
             include_private (False): whether to include fields that start with
                 ``_`` in the returned schema
             flat (False): whether to return a flattened schema where all
@@ -9482,14 +9478,13 @@ class SampleCollection(object):
 
         return name
 
-    def drop_index(self, field_or_name, quiet=False):
-        """Drops the index for the given field or name.
+    def drop_index(self, field_or_name):
+        """Drops the index for the given field or name, if necessary.
 
         Args:
             field_or_name: a field name, ``embedded.field.name``, or compound
                 index name. Use :meth:`list_indexes` to see the available
                 indexes
-            quiet (booL): whether to raise an error if no index exists
         """
         name, is_frame_index = self._handle_frame_field(field_or_name)
 
@@ -9516,15 +9511,8 @@ class SampleCollection(object):
             else:
                 index_map[key] = key
 
-        if name not in index_map:
-            if not quiet:
-                itype = "frame index" if is_frame_index else "index"
-                msg = "%s has no %s '%s'" % (self.__class__.__name__, itype, name)
-                raise ValueError(msg)
-            else:
-                return
-
-        coll.drop_index(index_map[name])
+        if name in index_map:
+            coll.drop_index(index_map[name])
 
     def _get_default_indexes(self, frames=False):
         if frames:
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1403,6 +1403,8 @@ class SampleCollection(object):
         ftype=None,
         embedded_doc_type=None,
         read_only=None,
+        info_keys=None,
+        created_after=None,
         include_private=False,
         flat=False,
         mode=None,
@@ -1418,10 +1420,16 @@ class SampleCollection(object):
                 iterable of types to which to restrict the returned schema.
                 Must be subclass(es) of
                 :class:`fiftyone.core.odm.BaseEmbeddedDocument`
-            include_private (False): whether to include fields that start with
-                ``_`` in the returned schema
             read_only (None): whether to restrict to (True) or exclude (False)
                 read-only fields. By default, all fields are included
+            info_keys (None): an optional key or list of keys that must be in
+                a field's ``info`` dict in order for it to be included in the
+                returned schema. If ``None``, no filtering is performed.
+            created_after (None): an optional ``datetime`` to filter the
+                returned schema by, such that the field was `created_at` after
+                this time. If ``None``, no filtering is performed.
+            include_private (False): whether to include fields that start with
+                ``_`` in the returned schema
             flat (False): whether to return a flattened schema where all
                 embedded document fields are included as top-level keys
             mode (None): whether to apply the above constraints before and/or
@@ -1440,6 +1448,8 @@ class SampleCollection(object):
         ftype=None,
         embedded_doc_type=None,
         read_only=None,
+        info_keys=None,
+        created_after=None,
         include_private=False,
         flat=False,
         mode=None,
@@ -1458,6 +1468,12 @@ class SampleCollection(object):
                 :class:`fiftyone.core.odm.BaseEmbeddedDocument`
             read_only (None): whether to restrict to (True) or exclude (False)
                 read-only fields. By default, all fields are included
+            info_keys (None): an optional key or list of keys that must be in
+                a field's ``info`` dict in order for it to be included in the
+                returned schema. If ``None``, no filtering is performed.
+            created_after (None): an optional ``datetime`` to filter the
+                returned schema by, such that the field was `created_at` after
+                this time. If ``None``, no filtering is performed.
             include_private (False): whether to include fields that start with
                 ``_`` in the returned schema
             flat (False): whether to return a flattened schema where all
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -59,7 +59,7 @@ fot = fou.lazy_import("fiftyone.core.stages")
 foud = fou.lazy_import("fiftyone.utils.data")
 
 
-_INVERTED_INDEX_KEY = "_inverted_index"
+_FRAME_SUMMARY_KEY = "_frame_summary"
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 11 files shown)


fiftyone/core/dataset.py
```diff
@@ -1630,6 +1650,267 @@ def add_frame_field(
...
+                rollup_path,
+                fo.EmbeddedDocumentField,
+                embedded_doc_type=fo.Classifications,
+                info=info,
+            )
+        else:
+            self.add_sample_field(rollup_path, fo.ListField, info=info)
+
+        # Optionally create mongodb index on rollup field
+        if create_index:
```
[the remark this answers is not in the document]
swheaton: Well the reason I added it here is so that the inserts would be done after the index was created, thinking maybe mongodb could optimize better and the additional delay would be amortized over the set_values() calls.

But now I'm thinking you're right it may be better to do it at the end and maybe we can use the non-blocking index creation param so that we don't have to wait but it's going in the background.

tests/unittests/import_export_tests.py
```diff
@@ -3294,6 +3294,11 @@ def test_fiftyone_dataset(self):
         created_at2 = dataset2.values("created_at")
         last_modified_at2 = dataset2.values("last_modified_at")
 
+        self.assertTrue(
```
minhtuev, line 3297: do i need to remove this? @swheaton

fiftyone/core/fields.py
```diff
@@ -23,7 +23,13 @@
 import fiftyone.core.utils as fou
 
 
-def validate_constraints(ftype=None, embedded_doc_type=None, read_only=None):
+def validate_constraints(
```
minhtuev, line 26: might need to move this out of this PR or wait for https://github.com/voxel51/fiftyone/pull/4597 to merge

fiftyone/core/dataset.py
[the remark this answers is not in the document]
brimoor: Has this been benchmarked? My intuition is that declaring the index first and then setting values will be faster than the other way around

fiftyone/core/collections.py
```diff
@@ -9516,15 +9511,8 @@ def drop_index(self, field_or_name, quiet=False):
...
-        if name not in index_map:
-            if not quiet:
-                itype = "frame index" if is_frame_index else "index"
-                msg = "%s has no %s '%s'" % (self.__class__.__name__, itype, name)
-                raise ValueError(msg)
-            else:
-                return
-
-        coll.drop_index(index_map[name])
+        if name in index_map:
```
minhtuev, line 9514: Nice!

fiftyone/core/odm/embedded_document.py
```diff
@@ -48,6 +48,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.validate()
 
+    def __hash__(self):
```
minhtuev, line 51: @brimoor : do we need to add unit tests for this?

fiftyone/core/dataset.py
```diff
@@ -3692,7 +4230,7 @@ def _save(self, view=None, fields=None):
             self._deleted = True
             raise ValueError("Dataset '%s' is deleted" % name)
 
-    def _save_field(self, field):
+    def _save_field(self, field, _enforce_read_only=True):
```
minhtuev, line 4233: I assume the reason for us to pass this as a private argument is that we don't want to publicly publish it?

brimoor: In general, private methods are already "undocumented, call at your own risk" methods. But when it comes to bypassing things like read-only constraints, just prefer to send an extra signal, even to internal callers, to not use this parameter unless you reallllly know what you're doing

fiftyone/core/fields.py
```diff
@@ -86,11 +96,26 @@ def validate_constraints(ftype=None, embedded_doc_type=None, read_only=None):
...
+        has_contraints = True
+        if not isinstance(info_keys, (list, str)):
+            raise ValueError("info_keys must be a single or a list of str's")
+
+    if created_after is not None:
+        has_contraints = True
+        if not isinstance(created_after, datetime):
+            raise ValueError("created_after must be a datetime")
+
     return has_contraints
```
minhtuev, line 109: How about: `return (ftype or embedded_doc_type or readonly or info_keys or created_after or created_after) is not None`?  We can get rid of these `has_constraints = True` statements

fiftyone/core/odm/embedded_document.py
brimoor: I had to add this in order for the new unit test in this PR to work - needed to convert `list(DynamicEmbeddedDocument)` to `set(DynamicEmbeddedDocument)` - so it's already implicitly tested

tests/unittests/dataset_tests.py
```diff
@@ -711,6 +711,233 @@ def test_index_in_progress(self):
...
+                "frames_gt_confidence",
+                "frames_gt_label_counts",
+                "frames_gt_confidence_by_label",
+            },
+        )
+
+        for field_name in update_fields:
+            dataset.update_summary_field(field_name)
+
+        update_fields = dataset.check_summary_fields()
```
minhtuev, line 901: nice!

tests/unittests/dataset_tests.py
```diff
@@ -711,6 +711,233 @@ def test_index_in_progress(self):
...
+        self.assertFalse("frames_gt_label_counts.count" in db_indexes)
+        self.assertFalse("frames_gt_confidence_by_label.label" in db_indexes)
+        self.assertFalse("frames_gt_confidence_by_label.min" in db_indexes)
+        self.assertFalse("frames_gt_confidence_by_label.max" in db_indexes)
+
+        update_fields = dataset.check_summary_fields()
+
+        self.assertListEqual(update_fields, [])
+
+        label_upper = F("label").upper()
```
minhtuev, line 878: noob question - does this just change all the field names to upper case?

fiftyone/core/dataset.py
```diff
@@ -1626,6 +1644,526 @@ def add_frame_field(
...
+
+        field = self.get_field(field_name)
+        if field is not None:
+            if overwrite and _SUMMARY_FIELD_KEY in field.info:
+                self.delete_summary_field(field_name)
+            else:
+                raise ValueError(f"Field '{field_name}' already exists")
+
+        index_fields = []
+        summary_info = {"path": path, "index_type": index_type}
```
minhtuev, line 1788: probably should rename this into `field_type`

fiftyone/core/dataset.py
```diff
@@ -1626,6 +1644,526 @@ def add_frame_field(
...
+            if include_counts:
+                label_field = field_name + ".label"
+                count_field = field_name + ".count"
+                index_fields.extend([label_field, count_field])
+
+                self.add_sample_field(
+                    field_name,
+                    fof.ListField,
+                    subfield=fof.EmbeddedDocumentField,
+                    embedded_doc_type=foo.DynamicEmbeddedDocument,
```
minhtuev, line 1800: Noob question - what is dynamic embedded document?

brimoor: https://docs.voxel51.com/user_guide/using_datasets.html#custom-embedded-documents

fiftyone/core/dataset.py
```diff
@@ -1626,6 +1644,526 @@ def add_frame_field(
...
+                self.add_sample_field(
+                    field_name,
+                    fof.EmbeddedDocumentField,
+                    embedded_doc_type=foo.DynamicEmbeddedDocument,
+                    info={_SUMMARY_FIELD_KEY: summary_info},
+                )
+                self.add_sample_field(min_field, type(_field))
+                self.add_sample_field(max_field, type(_field))
+
+        if sidebar_group is not False:
```
minhtuev, line 1854: Is there an advantage in not having a sidebar group for field summary? We can simplify by just having a `summaries` group by default

tests/unittests/dataset_tests.py
brimoor: yep: https://docs.voxel51.com/api/fiftyone.core.expressions.html#fiftyone.core.expressions.ViewExpression.upper

fiftyone/core/dataset.py
brimoor: IDK personal preference 🤷

fiftyone/core/dataset.py
brimoor: The default *is* to create a `summaries` group by default, Passing `sidebar_group=False` is just a bail out option in case the user doesn't want us to touch the sidebar groups at all.

For context, `dataset.app_config.sidebar_groups=None` by default and is only lazily initialized if the user actually drags fields around in the App (or modifies it via the SDK). This is exceptionally niche, but providing `sidebar_group=False` as an option allows for a case where a user prefer for `dataset.app_config.sidebar_groups` to stay as `None`

fiftyone/core/dataset.py
minhtuev: I think the main confusing point for me is that we are not creating an actual db index, but a summary field, so `index_type` is a little bit confusing if someone else has to read it again in a couple of months.

fiftyone/core/dataset.py
minhtuev: I see, makes sense. Just a hypothetical thought, if we don't allow `sidebar_group=False`, the user has to clear the side bar groups manually through SDK, or UI?

minhtuev:
Thanks @brimoor - this is awesome! I have a couple of thoughts:
1. `create_field_summary` is getting quite long, we should either consider refactoring it or add more comments to it. If we choose to refactor, we can do it in a separate PR (with TODO) so we can close this one.
2. If we delete all summary fields, does the `summary` sidebar group disappear automatically?

fiftyone/core/dataset.py
minhtuev: Haha nice, the Fiftyone nested document doll ~

tests/unittests/dataset_tests.py
```diff
@@ -711,6 +711,233 @@ def test_index_in_progress(self):
...
+        dataset.set_field("gt.detections.label", label_upper).save()
+        dataset.set_field("frames.gt.detections.label", label_upper).save()
+
+        update_fields = dataset.check_summary_fields()
+
+        self.assertSetEqual(
+            set(update_fields),
+            {
+                "gt_label",
```
minhtuev, line 887: Another noob question, if the field names are upper-cased in line 878 above, not sure why these field names are still in lower case

brimoor: the upper-casing is applied to the values in the field, not the names of the summary fields

fiftyone/core/dataset.py
brimoor: If sidebar groups exist and the user doesn't like them, then yes they either have to drag-to-reorder in the App or they could reset via the SDK if desired like so:

```py
dataset.app_config.sidebar_groups = None
dataset.save()
```

fiftyone/core/dataset.py
brimoor: oh I thought you wanted `summary_info -> field_type`. You want `index_type -> field_type`. Yeah I agree with that, will update

tests/unittests/dataset_tests.py
minhtuev: If that is the case, not sure why `gt_confidence`, `frames_gt_confidence`, `gt_label_counts` and `frames_gt_label_counts` should be in the update_fields. Should these fields be numeric?

brimoor: That's where the note in `check_summary_fields()` documentation comes in; unfortunately it's only a heuristic:

```
        Summary fields may need to be updated whenever there have been
        modifications to the dataset's samples since the summaries were last
        generated.

        Note that inclusion in this list is only a heuristic, as any sample
        modifications may not have affected the summary's source field.
```

swheaton:
Thank you for the detailed review @minhtuev 🙌🏼

brimoor:
> 2. If we delete all summary fields, does the `summary` sidebar group disappear automatically?

If a sidebar group becomes empty, it will remain in the App sidebar with an `X` button that the user can click to remove the group if they want, so its easy to get rid of.

fiftyone/core/dataset.py
brimoor: done ✅

brimoor (approved):

PR #4773, opened 2024-09-05
FOEPD-1665: Add support for materializing views

brimoor:
Adds a `materialize()` view stage that materializes the current view into a temporary database collection.

Apply this stage to an expensive view (eg an unindexed filtering operation on a large dataset) if you plan to perform multiple downstream operations on the view.

# Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.filter_labels("ground_truth", F("label") == "cat")
materialized_view = view.materialize()

# `materialized_view` contains the same contents as `view`
print(materialized_view.count("ground_truth.detections"))

# but it doesn't require aggregation to serve its contents!
fo.pprint(view._pipeline())
fo.pprint(materialized_view._pipeline())

# so downstream tasks like statistics are more efficient
print(view.count_values("ground_truth.detections.label"))
print(materialized_view.count_values("ground_truth.detections.label"))

# Materialized views are fully-compliant with the DatasetView interface, so
# you are free to chain additional view stages as usual
complex_view = materialized_view.shuffle().limit(10).to_patches("ground_truth")
print(complex_view.count_values("ground_truth.label"))
```

Like other generated views (`to_patches()`, `to_frames()`, and `to_clips()`), materialized views automatically sync changes back to their source dataset:

```py
sample = materialized_view.first()

sample["foo"] = "bar"
sample.save()

assert "foo" in dataset.get_field_schema()
assert dataset[sample.id].foo == "bar"

materialized_view.set_values("has_cats", ["True"] * len(materialized_view))

print(materialized_view.count_values("has_cats"))
print(dataset.count_values("has_cats"))

materialized_view.set_field("has_cats", "False (teehee)").save()

print(materialized_view.count_values("has_cats"))
print(dataset.count_values("has_cats"))

materialized_view.exclude_fields(["foo", "has_cats"]).keep_fields()

assert not dataset.has_field("foo")
assert not materialized_view.has_field("foo")

materialized_view.limit(0).keep()

assert len(materialized_view) == 0
assert len(dataset) == 0
```

They also respect the `include_indexes` parameter introduced in https://github.com/voxel51/fiftyone/pull/5955:

```py
dataset.compute_metadata()

dataset.create_index("metadata.size_bytes")
dataset.create_index("frames.detections.detections.label")

# No custom indexes are included by default
view = dataset.limit(5).materialize()
assert "metadata.size_bytes" not in view.list_indexes()
assert "frames.detections.detections.label" not in view.list_indexes()

view = dataset.limit(5).materialize(include_indexes=True)
assert "metadata.size_bytes" in view.list_indexes()
assert "frames.detections.detections.label" in view.list_indexes()

view = dataset.limit(5).materialize(include_indexes=["metadata.size_bytes"])
assert "metadata.size_bytes" in view.list_indexes()
assert "frames.detections.detections.label" not in view.list_indexes()
```


kaixi-wang:
> Apply this stage to an expensive view (eg an unindexed filtering operation on a large dataset) if you plan to perform multiple downstream operations on the view.

I’m a little confused about the use case. Why would creating a new collection + overhead of syncing be preferable to creating an index? Also if the permanent sample collection is indexed, materializing the view to an unindexed collection could lead to things being slower

docs/source/user_guide/using_views.rst
```diff
@@ -1006,6 +1006,161 @@ contains duplicate sample IDs:
...
+
+Materialized views are fully-compliant with the |DatasetView| interface, so you
+are free to chain additional view stages as usual:
+
+.. code-block:: python
+    :linenos:
+
+    complex_view = materialized_view.to_patches("ground_truth").limit(50)
+    print(complex_view.count_values("ground_truth.label"))
+    # {'person': 50}
```
swheaton, line 1073: I was seeing what I thought was unexpected behavior with `complex_view.keep()`. However the below is also unexpected to me so maybe I misunderstand things. So let's start here then explore `complex_view.keep()` and `materialized_view.keep()`, etc.

```python
dataset.count("ground_truth.detections")
# 1232

view = (
        dataset
        .filter_labels("ground_truth", F("label") == "person")
        .sort_by(F("ground_truth.detections.label").length(), reverse=True)
        .limit(10)
    )
patches=view.to_patches("ground_truth")
view.count_values("ground_truth.detections.label")                                                                                                                                                                             
# {'person': 140}
patches.count_values("ground_truth.label")
# {'person': 140}
patches.keep()
dataset.count("ground_truth.detections")
# 1232
```
Based on documentation
> Deletes all patches that are **not** in this view from the underlying dataset.

I would expect one of the following to be reasonable (documentation should be updated to make this clear at minimum):
1. Only patches in that view are kept. So the dataset has 200 samples still but only 140 detections, all "person".
2. Only patches in that view are kept, for only the samples represented in the view. The other 190 are left alone.

Based on above, neither of these seems to be the case? (Tested in develop also to rule out that this branch made some breaking change)

docs/source/user_guide/using_views.rst
```diff
@@ -1006,6 +1006,161 @@ contains duplicate sample IDs:
...
+    deleted, the collection(s) will be automatically regenerated from the
+    underlying dataset when the saved materialized view is loaded.
+
+Note that, since materialized views are served from a cache, their contents
+are not immediately affected if the source dataset is edited. To pull in
+changes, you can call
+:meth:`reload() <fiftyone.core.materialize.MaterializedView.reload>` to
+rehydrate an existing view, or you can use
+:meth:`materialize() <fiftyone.core.collections.SampleCollection.materialize>`
+to create a new materalized view:
```
swheaton, line 1102: We really need to fix nonpersistent dataset cleanup in Enterprise, this will only accelerate the litter :(

fiftyone/core/materialize.py
```diff
@@ -0,0 +1,621 @@
...
+                specified, only these fields are overwritten
+        """
+        if etau.is_str(fields):
+            fields = [fields]
+
+        super().save(fields=fields)
+
+        self._sync_source(fields=fields)
+
+    def keep(self):
```
swheaton, line 218: note to selves: add appropriate edit decorations in Enterprise

swheaton (commented):
Entirely possible I missed something due to sheer number of lines. But it looks right, works as documented, and automated tests are solid.

My only confusion is around behavior of edits to generated views. There may or may not be more to try/say for this PR depending on conclusion of that thread.

docs/source/user_guide/using_views.rst
```diff
@@ -1006,6 +1006,161 @@ contains duplicate sample IDs:
...
+    # {'person': 140}
+
+.. note::
+
+    Like :ref:`non-persistent datasets <dataset-persistence>`, the temporary
+    collections created by materializing views are automatically deleted when
+    all active FiftyOne sessions are closed.
+
+Materialized views are fully-compliant with the |DatasetView| interface, so you
+are free to chain additional view stages as usual:
```
swheaton, line 1066: Haha wow you keep calling materialize() on views and it still syncs, recursively I'm guessing.

```python
mat1 = ds.limit(150).materialize()
mat2 = mat1.limit(100).materialize()
mat3 = mat2.limit(10).materialize()
mat3.set_field("tags", ["from-mat3"]).save()
assert ds.limit(10).values('tags') == ["from-mat3"] * 10
```

docs/source/user_guide/using_views.rst
swheaton: Perhaps we should warn users about this, particularly given what we learned the other day about MongoDB performance and number of file descriptors (num_collections + num_indexes).

If they are going to use this feature, they might want to consider always creating a saved view so that all usages would reference the same backing dataset instead of creating a new one every time.

swheaton:
Returning here: The theory is interesting but I'm wondering what specific use cases we might be targeting as potential candidates and if we have performance data showing this additional complexity would indeed be worth it for the codebase and users to take on.

docs/source/user_guide/using_views.rst
```diff
@@ -1006,6 +1006,161 @@ contains duplicate sample IDs:
     The :ref:`FiftyOne App <fiftyone-app>` is not designed to display views
     with duplicate sample IDs.
 
+.. _materializing-views:
```
brimoor, line 1009: @swheaton you commented live on whether we should use the "materialize" name for this feature. The consensus was:
1. It is better to choose a tech-y name for this feature that alludes to how it is implemented rather than a name like "optimize" that is more business value-y but risks the user not thinking about *how it works*, which is an important piece of whether it is appropriate to use
2. Materialized view is a generic database term, so we'd still use it if we weren't built on MongoDB

Net-net, "materialize" it is ✅

docs/source/user_guide/using_views.rst
brimoor: **Should** you do it? no. **Can** you do it? yep!

docs/source/user_guide/using_views.rst
brimoor: Right this is a bit subtle: `to_patches().keep()` is a no-op. Deletion only occurs when you inject filters in between `to_patches()` and `keep()`. 

For example `view.to_patches().skip(10).keep()` will delete 10 patches from the dataset, but *will not* delete any patches that were never in `view` to begin with.

brimoor: I've updated the docstrings to more carefully explain this in https://github.com/voxel51/fiftyone/pull/4773/commits/4788438a8e6fb28502bc498e5e5fe20dc93b6689

docs/source/user_guide/using_views.rst
brimoor: > We really need to fix nonpersistent dataset cleanup in Enterprise, this will only accelerate the litter :(

Definitely agree that we need to get cleanup working. My argument live was that this is likely a small delta compared to the relatively larger problem that even just `fo.Dataset()` creates 2 collections and like 10 indexes that never go away.

> If they are going to use this feature, they might want to consider always creating a saved view so that all usages would reference the same backing dataset instead of creating a new one every time

yes I agree that, once https://github.com/voxel51/fiftyone/pull/6067 is merged, saved views should be heavily encouraged for repeat-use cases. That's something that CS can start preaching, and we can consider ways to highlight this in the docs in a follow-up PR in my opinion (it applies to all generated views, not just materialized ones)

fiftyone/core/materialize.py
brimoor: yep I have a commit already in place for that. I'll handle it ✅

brimoor:
> I’m a little confused about the use case. Why would creating a new collection + overhead of syncing be preferable to creating an index? Also if the permanent sample collection is indexed, materializing the view to an unindexed collection could lead to things being slower

@kaixi-wang for views that are index friendly, yes that pathway should be used instead! `materialize()` is still necessary in situations where you want to optimize downstream usage of complex views that require `COLLSCAN` to initially generate.

> thinking a little out of the box, but what if materialize view was not creating a new mongo collection but involved loading the specified subset into a performant in memory data frame on the client (eg polars https://pola.rs/). there would be work needed to support writing changes back, but a columnar storage will make filtering fast and local availability will enable subsequent work with the data to not be bottlenecked by constant network I/o or slow mongo collscans

@kaixi-wang yeah this makes sense. But I don't see any way to achieve that short of going all-in and building a generic `SampleCollection` interface that can be attached to any backing DB, not just MongoDB.

> Returning here: The theory is interesting but I'm wondering what specific use cases we might be targeting as potential candidates and if we have performance data showing this additional complexity would indeed be worth it for the codebase and users to take on.

@swheaton this was motivated by listening to user stories over many CS<>Product syncs where customers wanted us to optimize situations where they perform an expensive filter on a huge dataset and then want to work with the resulting smaller set in multiple ways without having to rerun the full query from scratch at every interaction.

Re benchmarking: at risk of sounding flippant, given the implementation, `materialize()` **must** yield arbitrarily large performance gains as the size of the dataset increases and the size of the matching view decreases. The key is that you must be willing to pay the one-time cost of materializing for the downstream payoff. But this makes sense; you can `materialize()` + save view offline (via a DO, say) and then you get fast interactive filters when you're using the view online in the App. I think this is a nice option to have in the FO toolbelt.

PR #4784, opened 2024-09-09
add product analytics rate limiting

ritch:
Adds rate limiting and debouncing to the analytics package.

## Summary by CodeRabbit

- **New Features**
	- Enhanced analytics functionality with configurable event rate limiting and debouncing options.
	- Improved event tracking efficiency by preventing excessive or rapid event logging.
- **Bug Fixes**
	- Added debugging messages for better visibility on event tracking behavior and limits.
	- Ensured analytics do not load without a write key and respect the `doNotTrack` flag.

### DIFF
--- a/app/packages/analytics/src/usingAnalytics.ts
+++ b/app/packages/analytics/src/usingAnalytics.ts
@@ -9,7 +9,6 @@ export type AnalyticsInfo = {
 };
 
 export type AnalyticsConfig = {
-  eventRateLimit: number;
   debounceInterval: number;
 };
 
@@ -17,7 +16,7 @@ let _analytics: Analytics = null;
 
 export default function usingAnalytics(info: AnalyticsInfo): Analytics {
   if (!_analytics) {
-    _analytics = new Analytics(info);
+    _analytics = new Analytics();
   }
   if (info) {
     _analytics.load(info);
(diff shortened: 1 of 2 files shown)


app/packages/analytics/src/usingAnalytics.ts
```diff
@@ -23,6 +25,20 @@ export default function usingAnalytics(info: AnalyticsInfo): Analytics {
 export class Analytics {
   private _segment?: AnalyticsBrowser;
   private _debug = false;
+  private _eventCount = 0; // Tracks number of events per session
+  private _lastEventTimestamps: Record<string, number> = {}; // Tracks last event times
+  private _rateLimit = 500; // Default max events per session
```
kaixi-wang, line 30: we should do a quick check against the existing events before deciding on these defaults as this could potentially skew the data significantly and lead to inaccurate analyses

app/packages/analytics/src/usingAnalytics.ts
```diff
@@ -6,13 +6,15 @@ export type AnalyticsInfo = {
   userGroup: string;
   doNotTrack?: boolean;
   debug: boolean;
+  eventRateLimit?: number; // Max events per session
```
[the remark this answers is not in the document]
ritch: Lifetime of the browser tab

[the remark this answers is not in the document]
ritch: refresh would clear this

app/packages/analytics/src/usingAnalytics.ts
jleven: I would vote pretty strongly against a max events per session limit.

jleven (commented):
I think the debounce is a reasonable stop-gap. I would need some convincing to get on board with a events-per-session limit, do we think the debounce won't solve the problem on its own?

I'd prefer if we could (eventually) fix the problem at its source and stop emitting extra events, but the debounce seems like a good immediate fix.

app/packages/analytics/src/usingAnalytics.ts
ritch: Sounds good to me

kaixi-wang (approved):
lgtm!

PR #4787, opened 2024-09-10
Support large deletions by sample/frame IDs

brimoor:
The `delete_samples()` and `clear_frames()` methods did not currently handle the case where the user wants to delete >100k samples/frames. This patches that.

## Example usage

Unit tests should already cover these methods, but here's a quick test I ran live:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.delete_samples(dataset[:2])
assert len(dataset) == 8

dataset[:2].clear_frames()
assert dataset[:2].values(F("frames").length()) == [0, 0]

frame_ids = dataset[2:4].values("frames.id", unwind=True)
dataset._clear_frames(frame_ids=frame_ids)
assert dataset[2:4].values(F("frames").length()) == [0, 0]

view = dataset[4:6].to_clips([[(1, 7),  (3, 10)], [(1, 10)]])
view.keep_frames()
assert dataset[4:6].values(F("frames").length()) == [10, 10]

view = dataset.match_frames(F("frame_number") <= 10)
view.keep_frames()
assert all(c <= 10 for c in dataset.values(F("frames").length()))

dataset.clear()
assert len(dataset) == 0
assert dataset.count("frames") == 0
```

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -4244,12 +4244,22 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         else:
             contains_videos = self._contains_videos(any_slice=True)
 
+        ops = []
         if sample_ids is not None:
-            d = {"_id": {"$in": [ObjectId(_id) for _id in sample_ids]}}
-        else:
-            d = {}
+            batch_size = fou.recommend_batch_size_for_value(
+                ObjectId(), max_size=100000
+            )
 
-        self._sample_collection.delete_many(d)
+            for _ids in fou.iter_batches(sample_ids, batch_size):
+                ops.append(
+                    DeleteMany(
+                        {"_id": {"$in": [ObjectId(_id) for _id in _ids]}}
+                    )
+                )
+        else:
+            ops.append(DeleteMany({}))
+
+        foo.bulk_write(ops, self._sample_collection)
         fos.Sample._reset_docs(
             self._sample_collection_name, sample_ids=sample_ids
         )
@@ -4336,9 +4346,19 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
                 frame_ids = view.values("frames.id", unwind=True)
 
         if frame_ids is not None:
-            self._frame_collection.delete_many(
-                {"_id": {"$in": [ObjectId(_id) for _id in frame_ids]}}
+            ops = []
+            batch_size = fou.recommend_batch_size_for_value(
+                ObjectId(), max_size=100000
             )
+
+            for _ids in fou.iter_batches(frame_ids, batch_size):
+                ops.append(
+                    DeleteMany(
+                        {"_id": {"$in": [ObjectId(_id) for _id in _ids]}}
+                    )
+                )
+
+            foo.bulk_write(ops, self._frame_collection)
             fofr.Frame._reset_docs_by_frame_id(
                 self._frame_collection_name, frame_ids
             )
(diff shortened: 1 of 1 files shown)


minhtuev (approved):
LGTM - added some comments :)

fiftyone/core/dataset.py
```diff
@@ -4244,12 +4244,22 @@ def _clear(self, view=None, sample_ids=None):
         else:
             contains_videos = self._contains_videos(any_slice=True)
 
+        ops = []
         if sample_ids is not None:
-            d = {"_id": {"$in": [ObjectId(_id) for _id in sample_ids]}}
+            batch_size = fou.recommend_batch_size_for_value(
+                ObjectId(), max_size=100000
```
minhtuev, line 4250: should we define this as a constant somewhere?

fiftyone/core/dataset.py
```diff
@@ -4336,9 +4346,19 @@ def _clear_frames(self, view=None, sample_ids=None, frame_ids=None):
                 frame_ids = view.values("frames.id", unwind=True)
 
         if frame_ids is not None:
-            self._frame_collection.delete_many(
-                {"_id": {"$in": [ObjectId(_id) for _id in frame_ids]}}
+            ops = []
+            batch_size = fou.recommend_batch_size_for_value(
+                ObjectId(), max_size=100000
```
minhtuev, line 4351: use the same constant as above

fiftyone/core/dataset.py
```diff
@@ -4244,12 +4244,22 @@ def _clear(self, view=None, sample_ids=None):
...
+
+            for _ids in fou.iter_batches(sample_ids, batch_size):
+                ops.append(
+                    DeleteMany(
+                        {"_id": {"$in": [ObjectId(_id) for _id in _ids]}}
+                    )
+                )
         else:
-            d = {}
+            ops.append(DeleteMany({}))
```
minhtuev, line 4260: Just curious, do we need to explicitly pass in an empty delete here? Does passing in an empty list for ops raise an error?

fiftyone/core/dataset.py
```diff
@@ -4350,12 +4370,26 @@ def _clear_frames(self, view=None, sample_ids=None, frame_ids=None):
 
             sample_ids = view.values("id")
 
+        ops = []
         if sample_ids is not None:
-            d = {"_sample_id": {"$in": [ObjectId(_id) for _id in sample_ids]}}
+            batch_size = fou.recommend_batch_size_for_value(
```
minhtuev, line 4375: Just a thought, what if we can do a refactor, something like `dataset._delete(...)` where we can decide which code path to use based on the number of documents? Is there any advantage to using `self._sample_collection.delete_many`?

fiftyone/core/dataset.py
brimoor: `DeleteMany({})` means delete *everything*, not delete *nothing* 😄

fiftyone/core/dataset.py
brimoor: It's actually not a constant, as `recommend_batch_size_for_value()` decides what batch size to use based on the current `fo.config` which could be changed dynamically by the user:
https://github.com/voxel51/fiftyone/blob/c699f8dfc16acd01d17acc012aa961e691e538e8/fiftyone/core/utils.py#L1678

fiftyone/core/dataset.py
brimoor: I admittedly haven't benchmarked thoroughly, but I'd expect that `delete_many(...)` and `bulk_write([DeleteMany(...)], ...)` are basically equivalent.

fiftyone/core/dataset.py
swheaton: I think he meant the `max_size=100000`

PR #4806, opened 2024-09-17
Protect against infinitely growing content size batcher

swheaton:

If content size batcher is not getting backpressure properly, it would just keep increasing batch size to infinity. Eventually causing problems.
If some versions were misaligned, this was possible to rear its ugly head in Teams.

Let's just cap the batcher (via `max_batch_size`) to `target_size`, which would mean 1 byte per object.

This is essentially what was happening (fixed backpressure) so let's just prevent it.
```python
import fiftyone as fo

batcher = fo.core.utils.ContentSizeDynamicBatcher(None, init_batch_size=100)

for _ in range(10):
  print(next(batcher))
  batcher.apply_backpressure(100 * 1024)
```

Before 👎🏼 
```
100
1024
10486
107377
1099540
11259290
115295130
1180622131
12089570621
123797203159
...
```
After 👍🏼 
```
100
1024
10486
107377
1048576
1048576
...
```

## Summary by CodeRabbit

- **New Features**
	- Enhanced batch processing by enforcing a maximum limit on the `max_batch_size` parameter, improving memory efficiency.
  
- **Bug Fixes**
	- Updated tests to reflect changes in batch processing logic, accommodating larger input sizes and ensuring expected outputs align with new parameters.

- **Documentation**
	- Improved readability of test function signatures for better clarity.

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -1462,6 +1462,9 @@ class ContentSizeDynamicBatcher(BaseDynamicBatcher):
         progress=False,
         total=None,
     ):
+        # If unset or larger, max batch size must be 1 byte per object
+        if max_batch_size is None or max_batch_size > target_size:
+            max_batch_size = target_size
         super().__init__(
             iterable,
             target_size,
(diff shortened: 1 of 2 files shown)


minhtuev (approved):
Nice!

PR #4810, opened 2024-09-17
add pending state for delegated_ops collection

CamronStaley:

Add pending state for delegated operations collection

tested in fiftyone celery poc

```
Total currently running or pending jobs: 1
Total currently running or pending jobs: 1
Executing operation: 66e8ccd98cecb3ff22558ad1
Celery task id: d45b78b2-f4ff-4f7f-b050-28275f09b0eb
Total currently running or pending jobs: 2
```

image

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced a 'PENDING' state for operations in the command-line interface, enhancing filtering and sorting capabilities.
	- Added new retrieval methods for pending and running operations, improving operation management.
	- Expanded sorting options to include 'PENDING_AT'.

- **Bug Fixes**
	- Updated help text for command-line arguments to reflect new operation states.

- **Documentation**
	- Enhanced CLI documentation to include new operation states and sorting options.


fiftyone/factory/repos/delegated_operation_doc.py
```diff
@@ -68,6 +69,7 @@ def from_pymongo(self, doc: dict):
             doc["completed_at"] if "completed_at" in doc else None
         )
         self.failed_at = doc["failed_at"] if "failed_at" in doc else None
+        self.pending_at = doc["pending_at"] if "pending_at" in doc else None
         self.pinned = doc["pinned"] if "pinned" in doc else None
         self.dataset_id = doc["dataset_id"] if "dataset_id" in doc else None
         self.run_link = doc["run_link"] if "run_link" in doc else None
```
swheaton, line 75: nit: I agree with a suggestion from coderabbit, can we just simply replace all of these `doc["thing"] if "thing" in doc else None` with `doc.get("thing")`

swheaton (changes requested):
Please add to `tests/unittests/delegated_operators_tests.py` for new state / functions.

swheaton:
>fiftyone/core/cli.py (1)
Line range hint 12-24: Consider adjusting the fee structure or discount percentages.
>
>The current implementation of adding a flat $20 fee to any discounted bill could negate the discount benefit, especially for smaller purchases or lower loyalty tiers. For example, a $100 purchase with a 10% discount for 3 years of loyalty would become $110 after the fee, which is more than the original price.
>
>This might frustrate customers who barely qualify for a discount tier, as the fee could cost them more than the discount saves them. It may also disincentivize customers from maintaining loyalty if the discounts are not worthwhile after the fee.
>
>Consider revising either the discount percentages or the flat fee structure to better align with the goal of rewarding customer loyalty. The fees and discounts should combine to benefit the customer.

lol @coderabbitai this is a wild hallucination, `cli.py` has nothing even close to this 😆

fiftyone/factory/repos/delegated_operation_doc.py
kaixi-wang: Agree that the 'if thing in doc' syntax should be simplified to just 'get' but is it possible doc is None and if so, should it throw? If not maybe what's needed here is just a 'if doc'?

swheaton: I don't think it's possible. It would throw before and it would throw now

swheaton:
Per offline discussion, `pending` will now become `scheduled`. Naming not behavior change for this PR.

fiftyone/factory/repos/delegated_operation_doc.py
```diff
@@ -46,32 +46,36 @@ def __init__(
...
     def from_pymongo(self, doc: dict):
         # required fields
-        self.operator = doc["operator"]
-        self.queued_at = doc["queued_at"]
-        self.run_state = doc["run_state"]
-        self.label = doc["label"] if "label" in doc else None
-        self.updated_at = doc["updated_at"] if "updated_at" in doc else None
+        self.operator = doc.get("operator")
+        self.queued_at = doc.get("queued_at")
+        self.run_state = doc.get("run_state")
```
swheaton, line 59: Don't really need to use `get()` for these as they're required to be set. It just pushes the errors to later

fiftyone/factory/repos/delegated_operation_doc.py
```diff
@@ -46,32 +46,36 @@ def __init__(
...
-        )
-        self.started_at = doc["started_at"] if "started_at" in doc else None
-        self.completed_at = (
-            doc["completed_at"] if "completed_at" in doc else None
-        )
-        self.failed_at = doc["failed_at"] if "failed_at" in doc else None
-        self.pinned = doc["pinned"] if "pinned" in doc else None
-        self.dataset_id = doc["dataset_id"] if "dataset_id" in doc else None
-        self.run_link = doc["run_link"] if "run_link" in doc else None
+        self.delegation_target = doc.get("delegation_target", None)
```
swheaton, line 62: technically `None` is already the default so not necessary, but it's fine to leave it

swheaton (approved):
couple minor things but overall lgtm

PR #4829, opened 2024-09-21
Add the option of using panoptic segmentation masks

kemaleren:
[Panoptic segmentations](https://arxiv.org/abs/1801.00868) contain both instance objects and semantic segmentation. Instead of [PR#4500](https://github.com/voxel51/fiftyone/pull/4500), which implemented disk-backed detections as a new type of label, this PR extends the `Segmentation` class. It supports all the same functionality and more, so this PR replaces that one.

This pull request modifies the `Segmentation` class to allow it to represent both semantic and panoptic segmentations. The `mask` attribute for panoptic segmentations must be a 2-channel array, where the first channel contains a semantic segmentation and the second channel contains an object index. It would have been cleaner to maintain two separate arrays -- one for semantic segmentation and one for objects -- but this implementation maintains backwards compatibility with the way semantic segmentations work, which is to keep everything in a single `mask` array or single image indicated by `mask_path`.

For instance, the following segmentation contains three objects, each with a different class (10, 20, and 30), as well as a region of "stuff" with another class (99):

```
class_mask = np.array([
    [10, 99, 20, 20],
    [10, 10, 99, 20],
    [30, 30, 99, 99],
    [30, 99, 99, 99],
])

instance_mask = np.array([
    [1, 0, 2, 2],
    [1, 1, 0, 2],
    [3, 3, 0, 0],
    [3, 0, 0, 0],
])

panoptic_mask = np.stack(
    [class_mask, instance_mask], axis=-1
)
segmentation = fo.Segmentation(mask=panoptic_mask, is_panoptic=True)
```

When saving to or loading  from an image, such as a PNG, the third channel is left empty for panoptic segmentations. Therefore the `is_panoptic` boolean is necessary to indicate whether the third channel is empty actually, or whether it represents all `0` values in the blue channel of an RGB segmentation mask.

This PR also adds the `tifffile` package to allow saving and loading two-channel TIFF images, which allows a higher bit depth for saving saving segmentation class indices or object labels.

Conversions to/from `Detection`, `Detections`, `Polyline`, and `Polylines` are implemented. Two more `mask_type` values are now recognized, in addition to the "stuff" and "thing" types, because there are a number of ways that panoptic segmentations can be converted to other labels:

* "stuff" - object instance labels are ignored. This option is identical to "stuff" on a normal semantic segmentation.
* "thing" - object instances are first converted to individual detections/polylines. Then any remaining connected regions are also converted to individual detections/polylines, just like converting "thing" semantic classes.
* "panoptic" - object instances are first converted to individual detections/polylines. Then any remaining regions are converted to one detection/polyline, just like converting "stuff" semantic classes. This option is called "panoptic" because it most closely aligns with the original intention of treating objects as "thing" and non-object regions as "stuff".
* "object" - i.e. objects only. Object instances are first converted to individual detections/polylines. Then any remaining regions are ignored.

I'm opening this PR so we can start a discussion about this approach and make sure it aligns with the FiftyOne team's expectations. Before final approval, the following should be implemented and tested:

* User documentation
* Evaluations need to be updated to work with panoptic segmentations. Panoptic Quality (PQ) should be implemented.
* Other code that imports or transforms segmentations needs to be updated to support the panoptic option.

There are unit tests for all possible conversions and saving/loading the semantic segmentations.

        notes for FiftyOne users.

The `Segmentation` class now supports panoptic segmentations, which indicate not only the segmentation class of a pixel but also which object to which it belongs.  Panoptic `Segmentation` objects can be converted to detections and polylines, as well as to semantic segmentations, and they can be saved as TIF files.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced panoptic segmentation functionality, allowing for the handling of instance masks alongside class masks.
	- Added methods to facilitate conversion between semantic and panoptic segmentations.

- **Bug Fixes**
	- Updated mask processing methods to correctly handle the new panoptic attribute.

- **Chores**
	- Added `tifffile` as a required dependency for handling TIFF files.

- **Tests**
	- Implemented unit tests for validating segmentation conversions, enhancing the testing framework for segmentation functionalities.

### DIFF
--- a/fiftyone/core/labels.py
+++ b/fiftyone/core/labels.py
@@ -8,6 +8,7 @@ Labels stored in dataset samples.
 from functools import partial
 import itertools
 import warnings
+import pathlib
 
 from bson import ObjectId
 import cv2
@@ -15,6 +16,7 @@ import numpy as np
 import scipy.ndimage as spn
 import skimage.measure as skm
 import skimage.segmentation as sks
+import tifffile
 
 import eta.core.frameutils as etaf
 import eta.core.image as etai
@@ -447,7 +449,9 @@ class Detection(_HasAttributesDict, _HasID, Label):
             **attributes,
         )
 
-    def to_segmentation(self, mask=None, frame_size=None, target=255):
+    def to_segmentation(
+        self, mask=None, frame_size=None, target=255, panoptic=False
+    ):
         """Returns a :class:`Segmentation` representation of this instance.
 
         The detection must have an instance mask, i.e., its :attr:`mask`
@@ -463,6 +467,7 @@ class Detection(_HasAttributesDict, _HasID, Label):
                 provided
             target (255): the pixel value or RGB hex string to use to render
                 the object
+            panoptic (False): whether to return a panoptic segmentation
 
         Returns:
             a :class:`Segmentation`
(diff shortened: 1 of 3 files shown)


benjaminpkane (approved):
Hi @kemaleren. Proposal looks great! I will merge this and work through the App details. We can follow up offline about the other work you outlined

PR #4830, opened 2024-09-22
Adding builtin operators for more of the FO interface

brimoor:
## Change log

Added the following new builtin operators:
- `edit_field_info`
- `clone_frame_field`
- `rename_frame_field`
- `clear_frame_field`
- `delete_frame_field`
- `create_index`
- `drop_index`
- `create_summary_field`
- `update_summary_field`
- `delete_summary_field`
- `add_group_slice`
- `rename_group_slice`
- `delete_group_slice`
- `load_saved_view`
- `save_view`
- `edit_saved_view_info`
- `delete_saved_view`
- `edit_workspace_info`
- `sync_last_modified_at`

Enhancements to existing builtin operators
- Intelligent handling of read-only fields
- Intelligent handling of default fields
- Enhanced many fields' descriptions

Other updates
- Improved the robustness of the v1.0.0 migration (eg handle case where an existing dataset has an incompatible `created_at` or `last_modified_at` field)
- Optimized `max(last_modified_at)` computations via a new `_get_last_modified_at()` method
- Ensured that `last_modified_at` times are exactly in-sync when creating/updating saved views

## Tested by

Manual tests with the `quickstart`, `quickstart-video`, and `quickstart-groups` datasets.

### DIFF
--- a/fiftyone/operators/builtin.py
+++ b/fiftyone/operators/builtin.py
@@ -2111,6 +2111,52 @@ class DeleteWorkspace(foo.Operator):
         ctx.dataset.delete_workspace(name)
 
 
+class SyncLastModifiedAt(foo.Operator):
+    @property
+    def config(self):
+        return foo.OperatorConfig(
+            name="sync_last_modified_at",
+            label="Sync last modified at",
+            dynamic=True,
+        )
+
+    def resolve_input(self, ctx):
+        inputs = types.Object()
+
+        instructions = """
+This operation updates the `last_modified_at` property of the dataset if
+necessary to incorporate any modification timestamps to its samples.
+        """
+
+        inputs.str(
+            "instructions",
+            default=instructions.strip(),
+            view=types.MarkdownView(read_only=True),
+        )
+
+        if ctx.dataset._has_frame_fields():
+            inputs.bool(
+                "include_frames",
+                default=True,
+                required=False,
+                label="Include frames",
+                description=(
+                    "Whether to sync the `last_modified_at` property of each "
+                    "video sample first if necessary to incorporate any "
+                    "modification timestamps to its frames"
+                ),
+            )
+
+        return types.Property(
+            inputs, view=types.View(label="Sync last modified at")
+        )
+
+    def execute(self, ctx):
+        include_frames = ctx.params.get("include_frames", True)
+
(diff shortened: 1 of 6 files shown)


fiftyone/operators/builtin.py
```diff
@@ -434,90 +1835,428 @@ class LoadWorkspace(foo.Operator):
...
+            "color",
+            default=None,
+            required=False,
+            label="Color",
+            description=(
+                "An optional RGB color string like `#FF6D04` for the workspace"
+            ),
+        )
+
+        # @todo infer this automatically from current App spaces
```
brimoor, line 1921: @imanjra @ritch can do we have a way to get the current `session.spaces` layout during operator execution?

fiftyone/operators/builtin.py
```diff
@@ -434,90 +1835,428 @@ class LoadWorkspace(foo.Operator):
...
+        return
+
+    workspace_selector = types.AutocompleteView()
+    for key in workspaces:
+        workspace_selector.add_choice(key, label=key)
+
+    # @todo default to current workspace name, if one is currently open
+    inputs.enum(
+        "name",
+        workspace_selector.values(),
```
brimoor, line 2021: @imanjra @ritch here I'm wanting to add a `default` parameter that specifies the name of the current workspace, if a saved workspace is currently loaded. IE to the equivalent of `default=ctx.view.name` for saved views

fiftyone/operators/builtin.py
imanjra: No, not at the moment. It should be fairly trivial to add it in.

fiftyone/operators/builtin.py
imanjra: I believe the current workspace is not available to operator context. This should also be fairly trivial to add.

fiftyone/operators/builtin.py
ritch: If this could also include the opened panel names that would be super useful

ritch:
This is awesome!

imanjra:
+1 this is awesome! 🚀

imanjra (approved):
Rest LGTM. Tested locally 🚀

PR #4845, opened 2024-09-25
Add timeline ops

ritch:
Adds the following operators:

 - SetFrameNumber
 - SetPlayheadState

## Summary by CodeRabbit

- **New Features**
	- Introduced `SetPlayheadState` class for setting the playhead state in timelines.
	- Added `SetFrameNumber` class to specify and set a frame number in timelines.
	- Both classes are now available as built-in operators within the application, enhancing timeline control capabilities.

### DIFF
--- a/app/packages/operators/src/built-in-operators.ts
+++ b/app/packages/operators/src/built-in-operators.ts
@@ -1215,7 +1215,11 @@ export class SetPanelTitle extends Operator {
 type SetPlayheadStateHooks = {
   setPlayheadState: (state: fop.PlayheadState, timeline_name?: string) => void;
 };
-type SetPlayheadStateParams = { state: any; timeline_name?: string };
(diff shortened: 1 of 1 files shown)


app/packages/operators/src/built-in-operators.ts
```diff
@@ -1211,6 +1212,56 @@ export class SetPanelTitle extends Operator {
   }
 }
 
+type SetPlayheadStateHooks = {
+  setPlayheadState: (state: fop.PlayheadState, timeline_name?: string) => void;
+};
+type SetPlayheadStateParams = { state: any; timeline_name?: string };
+export class SetPlayheadState extends Operator {
```
sashankaryal, line 1219: nit: whitespace

and in line 1218, can state not be typed `fop.PlayheadState`?

also we're taking in `timeline_name` but not doing anything with it...

PR #4857, opened 2024-09-27
timeline fix: loadrange in imavid

sashankaryal:
Implements `loadRange()` in imavid. Previously we were relying on the loading logic in the render loop which resulted in unintended effects

## Release Notes

- **New Features**
  - Introduced a callback for handling animation stutters, enhancing playback experience.
  - Added a method to check and manage buffer fetching, improving data handling.

- **Improvements**
  - Enhanced buffer management and fetching logic across multiple components.
  - Simplified rendering and playback control for better performance.
  - Updated default frame rate for playback functionality.

- **Bug Fixes**
  - Updated logic for fetching additional data, ensuring smoother playback.

- **Style**
  - Changed the loading state color in the Seekbar component for better visibility.

- **Tests**
  - Added tests for buffer management functionality to ensure reliability.

### DIFF
--- a/app/packages/core/src/components/Modal/ImaVidLooker.tsx
+++ b/app/packages/core/src/components/Modal/ImaVidLooker.tsx
@@ -1,6 +1,5 @@
 import { useTheme } from "@fiftyone/components";
 import { AbstractLooker, ImaVidLooker } from "@fiftyone/looker";
-import { BUFFERING_PAUSE_TIMEOUT } from "@fiftyone/looker/src/lookers/imavid/constants";
 import { BaseState } from "@fiftyone/looker/src/state";
 import { FoTimelineConfig, useCreateTimeline } from "@fiftyone/playback";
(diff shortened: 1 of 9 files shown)


app/packages/core/src/components/Modal/ImaVidLooker.tsx
```diff
@@ -210,6 +257,10 @@ export const ImaVidLookerReact = React.memo(
       name: timelineName,
       config: timelineCreationConfig,
       waitUntilInitialized: readyWhen,
+      // using this mechanism to resume fetch if it was paused
+      // ideally we have control of fetch in this component but can't do that yet
```
benjaminpkane, line 261: Is the grid moving to react? :)

benjaminpkane (dismissed):
LGTM

app/packages/core/src/components/Modal/ImaVidLooker.tsx
sashankaryal: I definitely haven't thought about that one way or the other. 😅

ritch (approved):

PR #4869, opened 2024-09-30
Add spinner variant

Br2850:
- [x] Add `LoadingView` spinner variant

Screenshot 2024-09-30 at 3 12 23 PM
Screenshot 2024-09-30 at 3 13 10 PM


Br2850:
Failing 1 test because of circular dependency, but will merge and address that test in `feat/fo-assistant` after Ritchie's fix

PR #4893, opened 2024-10-04
add support for atomic state transitions

tom-vx51:

This PR adds a new `set_queued` method to the `DelegatedOperationService` which provides a mechanism for transitioning operators to the `queued` run state. 

In addition, this updates the existing `update_run_state` to include an optional `required_state` field to provide a means to control state transitions. This new field allows actors to perform a state transition _only if the operator matches the `required_state`_. This enables concurrency-safe state transitions where multiple actors may attempt to transition the same DO. With this new parameter, only the first attempted transition will succeed; subsequent attempts will fail due to a `required_state` mismatch.

**Important note!!** `update_run_state` can now return `None` if no matching records are found; this is consistent with the behavior of the underlying `find_one_and_update` call and can be used to determine whether a state transition occurred.

Unit tests

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced a `set_queued` method for managing operation states.
	- Enhanced state transition control with a new `required_state` parameter across multiple methods.

- **Bug Fixes**
	- Improved robustness of state updates by ensuring transitions only occur if the current state matches the specified `required_state`.

- **Tests**
	- Expanded unit tests to include checks for the new `set_queued` method and refined assertions for operation state transitions, ensuring accurate tracking of operation statuses.

### DIFF
--- a/fiftyone/factory/repos/delegated_operation.py
+++ b/fiftyone/factory/repos/delegated_operation.py
@@ -44,6 +44,7 @@ class DelegatedOperationRepo(object):
         result: ExecutionResult = None,
         run_link: str = None,
         progress: ExecutionProgress = None,
+        required_state: ExecutionResult = None,
     ) -> DelegatedOperationDocument:
         """Update the run state of an operation."""
         raise NotImplementedError("subclass must implement update_run_state()")
@@ -243,6 +244,7 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
         result: ExecutionResult = None,
         run_link: str = None,
         progress: ExecutionProgress = None,
+        required_state: ExecutionRunState = None,
     ) -> DelegatedOperationDocument:
         update = None
 
@@ -310,13 +312,17 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
             update["$set"]["status"] = progress
             update["$set"]["status"]["updated_at"] = datetime.utcnow()
 
+        collection_filter = {"_id": _id}
+        if required_state is not None:
+            collection_filter["run_state"] = required_state
+
         doc = self._collection.find_one_and_update(
-            filter={"_id": _id},
+            filter=collection_filter,
             update=[update],
             return_document=pymongo.ReturnDocument.AFTER,
         )
 
-        return DelegatedOperationDocument().from_pymongo(doc)
+        return DelegatedOperationDocument().from_pymongo(doc) if doc is not None else None
 
     def update_progress(
         self,
--- a/fiftyone/factory/repos/delegated_operation.py
+++ b/fiftyone/factory/repos/delegated_operation.py
@@ -301,6 +301,14 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
                     "updated_at": datetime.utcnow(),
                 }
             }
+        elif run_state == ExecutionRunState.QUEUED:
+            update = {
+                "$set": {
+                    "run_state": run_state,
+                    "queued_at": datetime.utcnow(),
+                    "updated_at": datetime.utcnow(),
+                }
+            }
 
         if run_link is not None:
             update["$set"]["run_link"] = run_link
--- a/fiftyone/factory/repos/delegated_operation.py
+++ b/fiftyone/factory/repos/delegated_operation.py
@@ -44,7 +44,7 @@ class DelegatedOperationRepo(object):
         result: ExecutionResult = None,
         run_link: str = None,
         progress: ExecutionProgress = None,
-        required_state: ExecutionResult = None,
+        required_state: ExecutionRunState = None,
     ) -> DelegatedOperationDocument:
         """Update the run state of an operation."""
         raise NotImplementedError("subclass must implement update_run_state()")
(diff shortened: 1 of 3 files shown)


fiftyone/operators/delegated.py
```diff
@@ -80,7 +80,7 @@ def set_progress(self, doc_id, progress):
         """
         return self._repo.update_progress(_id=doc_id, progress=progress)
 
-    def set_running(self, doc_id, progress=None, run_link=None):
+    def set_running(self, doc_id, progress=None, run_link=None, required_state: ExecutionRunState = None):
```
swheaton, line 83: this line appears to be too long. did you run the precommit hooks which include black?

fiftyone/operators/delegated.py
```diff
@@ -90,6 +90,8 @@ def set_running(self, doc_id, progress=None, run_link=None):
                 operation
             run_link (None): an optional link to orchestrator-specific
                 information about the operation
+            required_state (ExecutionRunState): an optional required state of the operation. If provided, the
```
swheaton, line 93: black will not reformat docstring but we will manually insert line breaks to keep to below 79

fiftyone/operators/delegated.py
swheaton, line 83: There was a recent argument about type hints. The resulting doctrine was that within a submodule, we have to keep it consistent. All type hints or no type hints.
Ritchie mentioned wanting to start with the `operators` submodule, but clearly this has not happened yet in `develop`. So best to remove type hints for your new parameter unless you want to take on the burden of adding them for the whole `operators` module.

fiftyone/operators/delegated.py
```diff
@@ -90,6 +90,8 @@ def set_running(self, doc_id, progress=None, run_link=None):
...
+                update will only be performed if the referenced operation matches this state.
 
         Returns:
             a :class:`fiftyone.factory.repos.DelegatedOperationDocument`
```
swheaton, line 97: Recommend updating return section on all set_* methods docstrings to say something like could be `None` if no such run is found, particularly in the case that `required_state` is set.

fiftyone/operators/delegated.py
swheaton, line 93: Typically in our docstrings the default value is in parentheses and the type may be in text in the description.
```python
required_state (None): an optional :class:`ExecutionRunState` as a required current state of the operation.
```

tests/unittests/delegated_operators_tests.py
swheaton, line 1: thanks for updating tests 💪🏼

swheaton (changes requested):
formatting and docstrings

fiftyone/factory/repos/delegated_operation.py
```diff
@@ -310,13 +320,17 @@ def update_run_state(
             update["$set"]["status"] = progress
             update["$set"]["status"]["updated_at"] = datetime.utcnow()
 
+        collection_filter = {"_id": _id}
+        if required_state is not None:
+            collection_filter["run_state"] = required_state
+
         doc = self._collection.find_one_and_update(
-            filter={"_id": _id},
+            filter=collection_filter,
```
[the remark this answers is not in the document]
swheaton: I am not sure why we are passing it as a list. But I just tested it and both forms work.
I think I'd prefer if we not wrap in a list, because that feels more standard.

kaixi-wang, line 328: If you filter on more than Id then you potentially end up with multiple entries per id if upsert is true or you will not update the doc at all if not upserting... is this what should happen or should that be considered an error?

fiftyone/factory/repos/delegated_operation.py
```diff
@@ -310,13 +320,17 @@ def update_run_state(
...
             update=[update],
             return_document=pymongo.ReturnDocument.AFTER,
         )
 
-        return DelegatedOperationDocument().from_pymongo(doc)
+        return DelegatedOperationDocument().from_pymongo(doc) if doc is not None else None
```
kaixi-wang, line 333: If the method is update run state then if it fails to update shouldn't it error?

tom-vx51: This behavior is consistent with the underlying `find_one_and_update`, which returns `None` if no records match the filter. I generally avoid exceptions when there's a reasonable alternative but could be convinced otherwise if you feel strongly.

fiftyone/factory/repos/delegated_operation.py
tom-vx51: `upsert` is `False` here so I don't think that's a case we need to worry about, but definitely something to keep in mind. The intended behavior is for no update to occur if the filter doesn't match.

fiftyone/operators/delegated.py
tom-vx51: Had it set to 99, whoops!

swheaton: yeah we use default of 79 in this repo. In others we have chosen to go with 99

swheaton:
See my PR where `required_state` is put to use. https://github.com/voxel51/fiftyone/pull/4907
That can be merged into this PR, or merged afterwards.

swheaton (approved):
lgtm now. I fix/add a couple remaining requests in https://github.com/voxel51/fiftyone/pull/4907 and https://github.com/voxel51/fiftyone/pull/4908

swheaton:
let merge after tests pass

PR #4902, opened 2024-10-07
add spaces context to operators

imanjra:

- Add `spaces` to operator context
- Add panel events for spaces and workspace change

Using a test panel and operator

        notes for FiftyOne users.

See above

-   [ ] Other

## Release Notes

- **New Features**
	- Enhanced `CustomPanel` to handle changes in spaces and workspace contexts.
	- Added new properties to the `ExecutionContext` for improved context management during operator execution.
	- Introduced optional parameters in the `register_panel` method to respond to changes in spaces and workspace.
	- Improved workspace management features with context-aware defaults for user inputs.
	- Expanded `Panel` class to handle additional context change events during startup.

- **Documentation**
	- Updated documentation for developing plugins to include new methods for managing spaces and workspace state changes.

These updates improve user interaction and flexibility within the application.

### DIFF
--- a/app/packages/operators/src/operators.ts
+++ b/app/packages/operators/src/operators.ts
@@ -1,5 +1,6 @@
 import { AnalyticsInfo, usingAnalytics } from "@fiftyone/analytics";
-import { ServerError, getFetchFunction, isNullish } from "@fiftyone/utilities";
+import { SpaceNode, spaceNodeFromJSON, SpaceNodeJSON } from "@fiftyone/spaces";
+import { getFetchFunction, isNullish, ServerError } from "@fiftyone/utilities";
 import { CallbackInterface } from "recoil";
 import { QueueItemStatus } from "./constants";
 import * as types from "./types";
@@ -92,6 +93,8 @@ export type RawContext = {
   };
   groupSlice: string;
   queryPerformance?: boolean;
+  spaces: SpaceNodeJSON;
+  workspaceName: string;
 };
 
 export class ExecutionContext {
@@ -139,6 +142,12 @@ export class ExecutionContext {
   }
   public get queryPerformance(): boolean {
     return Boolean(this._currentContext.queryPerformance);
+  }
+  public get spaces(): SpaceNode {
+    return spaceNodeFromJSON(this._currentContext.spaces);
+  }
+  public get workspaceName(): string {
+    return this._currentContext.workspaceName;
   }
 
   getCurrentPanelId(): string | null {
@@ -548,6 +557,8 @@ async function executeOperatorAsGenerator(
       view: currentContext.view,
       view_name: currentContext.viewName,
       group_slice: currentContext.groupSlice,
+      spaces: currentContext.spaces,
+      workspace_name: currentContext.workspaceName,
     },
     "json-stream"
   );
@@ -712,6 +723,8 @@ export async function executeOperatorWithContext(
           view_name: currentContext.viewName,
           group_slice: currentContext.groupSlice,
           query_performance: currentContext.queryPerformance,
+          spaces: currentContext.spaces,
+          workspace_name: currentContext.workspaceName,
         }
       );
       result = serverResult.result;
@@ -815,6 +828,8 @@ export async function resolveRemoteType(
       view: currentContext.view,
       view_name: currentContext.viewName,
       group_slice: currentContext.groupSlice,
+      spaces: currentContext.spaces,
+      workspace_name: currentContext.workspaceName,
     }
   );
 
@@ -889,6 +904,8 @@ export async function resolveExecutionOptions(
       view: currentContext.view,
       view_name: currentContext.viewName,
       group_slice: currentContext.groupSlice,
+      spaces: currentContext.spaces,
+      workspace_name: currentContext.workspaceName,
     }
   );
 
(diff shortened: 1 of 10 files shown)


brimoor (changes requested):
@imanjra thanks for adding this! 🙌 

What do you think about deleting `ctx.workspace` from the Python interface in favor of just ensuring that, if the current spaces are a saved workspace, then `ctx.spaces.name` will be set to the current workspace name? This is how `session.spaces.name` works btw. It would also be consistent with how saved views work: we have `ctx.view` with `ctx.view.name` populated when the current view is saved.

Can you also add `ctx.ops.set_spaces(spaces=spaces, name=name)` to programmatically load a spaces configuration or a saved workspace by name? Again for consistency with saved views.

And one more thing: can you update the builtin operators to remove the `@todo`s:
- `@voxel51/save_workspace`: [remove this](https://github.com/voxel51/fiftyone/blob/c8e6de2f75311a57c54f432fe62241a623f15d4f/fiftyone/operators/builtin.py#L1942-L1962) and instead just use the current `ctx.spaces` in `execute()`
- `@voxel51/edit_workspace_info`: use `default=ctx.spaces.name` for [the `name` prop](https://github.com/voxel51/fiftyone/blob/c8e6de2f75311a57c54f432fe62241a623f15d4f/fiftyone/operators/builtin.py#L2038-L2045)
- `@voxel51/delete_workspace`: use `default=ctx.spaces.name` for [the `name` prop](https://github.com/voxel51/fiftyone/blob/c8e6de2f75311a57c54f432fe62241a623f15d4f/fiftyone/operators/builtin.py#L2106-L2114)

docs/source/plugins/developing_plugins.rst
```diff
@@ -2060,6 +2060,32 @@ subsequent sections.
             ctx.panel.set_state("event", "on_change_group_slice")
             ctx.panel.set_data("event_data", event)
 
+        def on_change_spaces(self, ctx):
```
brimoor, line 2063: Thanks for updating the docs! One more place to document `ctx.spaces` is the [Execution context](https://docs.voxel51.com/plugins/developing_plugins.html#execution-context) section 😄

imanjra:
@brimoor great suggestions! I'll make the changes to merge the two and update the docs. Thank you!

fiftyone/operators/operations.py
```diff
@@ -356,6 +356,8 @@ def register_panel(
                 current extended selection changes
             on_change_group_slice (None): an operator to invoke when the group
                 slice changes
+            on_change_spaces (None): an operator to invoke when the current spaces changes
```
brimoor, line 359: Is `on_change_workspace` needed? Comparing with view-related events for context, we only have `on_change_view`, not `on_change_view_name`.

`on_change_spaces` should fire every time that `on_change_workspace` fires, plus it would also fire when changing between two different unsaved spaces configurations.

app/packages/operators/src/operators.ts
```diff
@@ -889,6 +904,8 @@ export async function resolveExecutionOptions(
       view: currentContext.view,
       view_name: currentContext.viewName,
       group_slice: currentContext.groupSlice,
+      spaces: currentContext.spaces,
+      workspace_name: currentContext.workspaceName,
```
brimoor, line 908: I do see that we have both `view` and `view_name` here, so it is consistent to have both `spaces` and `workspace_name` defined here, and also to use both of them in the implementation of `ExecutionContext.spaces` 👍

docs/source/plugins/developing_plugins.rst
```diff
@@ -2060,6 +2061,32 @@ subsequent sections.
...
+            The current state of spaces will be available via ``ctx.spaces``.
+            """
+            event = {
+                "data": ctx.spaces,
+                "description": "the current state of spaces",
+            }
+            ctx.panel.set_state("event", "on_change_spaces")
+            ctx.panel.set_data("event_data", event)
+
+        def on_change_workspace(self, ctx):
```
brimoor, line 2077: Potentially this should be deleted if you agree with my inclination to remove the `on_change_workspace` event.

If it remains, then I believe `ctx.workspace` here should now be `ctx.spaces.name`.

fiftyone/operators/builtin.py
```diff
@@ -1939,18 +1939,18 @@ def resolve_input(self, ctx):
             ),
         )
 
-        # @todo infer this automatically from current App spaces
+        current_spaces = ctx.spaces.to_json(True) if ctx.spaces else None
```
brimoor, line 1942: Ah, to clarify: here I was thinking that, now that we have `ctx.spaces`, we can just delete the `spaces` parameter entirely from `resolve_input()` and just use `ctx.spaces` in `execute()`.

Pasting in a spaces configuration as JSON was not something I wanted to support, it was just a placeholder until `ctx.spaces` was available 😄 

To support programmatic execution, we could still allow the `spaces` parameter to be optionally provided like so:

```py
def execute(self, ctx):
    spaces = ctx.params.get("spaces", None)

    if spaces is not None:
        spaces = _parse_spaces(spaces)
    else:
        spaces = ctx.spaces
```

brimoor (changes requested):

fiftyone/operators/operations.py
imanjra: Agreed. Removing `on_change_workspace` as it's little redundant

app/packages/operators/src/operators.ts
imanjra: Yup! This is needed to load a workspace instead of building spaces from raw JSON

fiftyone/operators/builtin.py
imanjra: Ahh I see. Yah, that makes sense 😄

imanjra: Perhaps, for programmatic execution, I will leave the support. However, I'll remove from `resolve_input`

brimoor (approved):
LGTM 🏅

PR #4904, opened 2024-10-07
Adding confidence to SAM predictions from box prompts

prernadh:

Added confidence scores to SAM predictions that are generate from bounding box prompts

Tested manually by applying model on a dataset and confirming that confidence scores are saved on the predictions.

        notes for FiftyOne users.

-   [x] Other: SAM utils changes

## Summary by CodeRabbit

- **New Features**
	- Enhanced prediction outputs for the segmentation models to include confidence scores alongside existing data (boxes, labels, masks).
	- Improved consistency in output formats across different model methods.

- **Bug Fixes**
	- Updated method signatures to ensure accurate return values for both segmentation models.

### DIFF
--- a/fiftyone/utils/sam.py
+++ b/fiftyone/utils/sam.py
@@ -271,14 +271,19 @@ class SegmentAnythingModel(fout.TorchSamplesMixin, fout.TorchImageModel):
                 device=sam_predictor.device,
             )
 
-            masks, _, _ = sam_predictor.predict_torch(
+            masks, scores, _ = sam_predictor.predict_torch(
                 point_coords=None,
                 point_labels=None,
                 boxes=transformed_boxes,
                 multimask_output=False,
             )
             outputs.append(
-                {"boxes": input_boxes, "labels": labels, "masks": masks}
+                {
+                    "boxes": input_boxes,
+                    "labels": labels,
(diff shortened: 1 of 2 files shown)


mythrandire:
Not a review comment, just a tangential observation - @prernadh any reason why we shouldn't also include the final return value (`low_res_masks_np`) from the `predict` method? I don't see a reason to not expose it, seeing as the change mirrors the addition of scores.

mythrandire (approved):
No comments, fairly straightforward addition to expose [prediction IOU scores](https://github.com/facebookresearch/sam2/blob/29267c8e3965bb7744a436ab7db555718d20391a/sam2/sam2_image_predictor.py#L303).

prernadh:
> Not a review comment, just a tangential observation - @prernadh any reason why we shouldn't also include the final return value (low_res_masks_np) from the predict method? I don't see a reason to not expose it, seeing as the change mirrors the addition of scores.
I would like to avoid saving too many sample fields unless necessary. My understanding of the `low_res_masks_np` is that its useful if you are using this predict method as an intermediate step and need to pass these low resolution masks further downstream to a model etc. I don't think that is required currently but if a user asks for it, it should be a small change

PR #4436, opened 2024-05-28
store delegated operation IO in metadata

imanjra:
## What changes are proposed in this pull request?

Store delegated operation IO schema in metadata

## How is this patch tested? If it is not, please explain why.

Unit and manual testins

### Is this a user-facing change that should be mentioned in the release notes?

-   [x] No. You can skip the rest of this section.
-   [ ] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

(Details in 1-2 sentences. You can just refer to another PR with a description
if this PR is part of a larger change.)

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Enhanced metadata handling in delegated operations, including input and output schemas.

- **Bug Fixes**
  - Improved accuracy in setting and extracting metadata during operation execution.

- **Tests**
  - Added new unit tests to validate metadata handling and schema extraction in delegated operations.

- **Refactor**
  - Streamlined code to include metadata parameters in various methods for better operation context management.

- **Chores**
  - Removed unused `ExecutionContext` import to clean up the codebase.

PR #4908, opened 2024-10-10
delegated operator fixes

swheaton:

Addresses some weirdness introduced in https://github.com/voxel51/fiftyone/pull/4436.

1. I believe the `if` on line 272 should be indented to be under the `COMPLETED` case. Otherwise the if statement flow is goofy and should be reworked anyways.
2. Passing an array as the `update` param is unusual, though legal. I'm wondering what was intended with the addition of the `$ifNull`? As written, I don't think it does much because the `output_schema` is fixed.

test still passes

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced logic for updating operation states, ensuring correct handling of null metadata.
  
- **Bug Fixes**
	- Improved handling of `outputs_schema` to prevent unintended document creation.

- **Tests**
	- Added a test case to verify system behavior when metadata is null, ensuring robust updates to `outputs_schema`.

### DIFF
--- a/fiftyone/factory/repos/delegated_operation.py
+++ b/fiftyone/factory/repos/delegated_operation.py
@@ -262,6 +262,8 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
             else None
         )
 
+        needs_pipeline_update = False
+
         if run_state == ExecutionRunState.COMPLETED:
             update = {
                 "$set": {
@@ -272,10 +274,11 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
                 }
             }
 
-        if outputs_schema:
-            update["$set"]["metadata.outputs_schema"] = {
-                "$ifNull": [outputs_schema, {}]
-            }
+            if outputs_schema:
+                update["$set"]["metadata.outputs_schema"] = (
+                    outputs_schema or {}
+                )
+                needs_pipeline_update = True
 
         elif run_state == ExecutionRunState.FAILED:
             update = {
@@ -325,9 +328,15 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
         if required_state is not None:
             collection_filter["run_state"] = required_state
 
+        # Using pipeline update instead of a single update doc fixes a case
+        #   where `metadata` is null and so accessing the dotted field
+        #   `metadata.output_schema` creates the document instead of erroring.
+        if needs_pipeline_update:
+            update = [update]
+
         doc = self._collection.find_one_and_update(
             filter=collection_filter,
-            update=[update],
+            update=update,
             return_document=pymongo.ReturnDocument.AFTER,
         )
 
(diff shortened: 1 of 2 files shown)


fiftyone/factory/repos/delegated_operation.py
```diff
@@ -312,7 +312,7 @@ def update_run_state(
 
         doc = self._collection.find_one_and_update(
             filter={"_id": _id},
-            update=[update],
+            update=update,
```
kaixi-wang, line 315: I believe this should be an array (aggregation pipeline). Otherwise passing a dictionary will be interpreted as the new doc so you'd essentially be setting the doc to have a new field "$set" with value equal to the dictionary value

kaixi-wang (changes requested):

fiftyone/factory/repos/delegated_operation.py
swheaton: You may be thinking of `find_one_and_replace()`. This change works and tests pass.
It's more standard to pass a dict with the updates. It's possible, though uncommon, to use an agg pipeline, but I didn't see why that was necessary.

swheaton: Alternatively if you don't specify `$set` and pass a [replacement document](https://www.mongodb.com/docs/manual/reference/method/db.collection.update/#std-label-update-method-replacement-document) instead, then it would behave as you say.

kaixi-wang: https://www.mongodb.com/docs/manual/reference/method/db.collection.findOneAndUpdate/

I'm just going off of the mongo docs

kaixi-wang: But yeah the alternative is to remove set

swheaton: And yes the test `test_updates_progress` tests for this. After talking with Ibrahim I am also making an update and adding a test for a specific corner case that's broken.

fiftyone/factory/repos/delegated_operation.py
```diff
@@ -310,9 +313,15 @@ def update_run_state(
             update["$set"]["status"] = progress
             update["$set"]["status"]["updated_at"] = datetime.utcnow()
 
+        # Using pipeline update instead of a single update doc fixes a case
```
kaixi-wang, line 316: Since the update can either be a pipeline or a single update, why do we need to only make it a pipeline under certain conditions and not just always like the current code? Is there a different bug that can happen by consistently using pipeline for all cases?

swheaton: It's nonstandard. I'd prefer to use the standard approach and deviate only when necessary.

swheaton:
@imanjra @kaixi-wang please re-review

kaixi-wang (approved):
lgtm

PR #4917, opened 2024-10-10
Added two env vars for enabling query performance

minhtuev:

Added two env vars that control the default behaviors for query performance mode

* Manual testing

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Updated terminology in the Options component from "Lightning mode" to "Query Performance mode."
  - Introduced new configuration options related to query performance.

- **Bug Fixes**
  - Improved exception handling in the saved_views method for better reliability.

- **Documentation**
  - Updated GraphQL schema and fragments to include new fields for query performance settings.

- **Chores**
  - Marked the lightning_threshold attribute for deprecation in the configuration settings.


fiftyone/core/config.py
```diff
@@ -348,12 +348,21 @@ def __init__(self, d=None):
...
+        # deprecate lightning threshold in favor of a timer-based approach
         self.lightning_threshold = self.parse_int(
             d,
             "lightning_threshold",
             env_var="FIFTYONE_APP_LIGHTNING_THRESHOLD",
             default=None,
         )
+        self.enable_query_performance = self.parse_bool(
+            d, "enable_query_performance",
+            env_var="FIFTYONE_APP_ENABLE_QUERY_PERFORMANCE", default=True
```
kaixi-wang, line 360: Why are we defaulting to true for both enable and default?

Also the naming is super confusing. I have no idea what query performance true/false actually does.

minhtuev: @kaixi-wang : do you have a suggestion for name change?

benjaminpkane (approved):
LGTM

fiftyone/core/config.py
minhtuev: FIFTYONE_ENABLE_QUERY_PERFORMANCE=true/false: whether to disable the entire query performance feature, including the panel, toasts, etc; default: true

FIFTYONE_DEFAULT_QUERY_PERFORMANCE=true/false: whether to enable query performance by default for each dataset (if query performance is enabled systemwide per first env var); default: true

PR #4949, opened 2024-10-19
[WIP] refactor js operator runtime

ritch:
Refactors the JS operator runtime into a separate submodule that matches the existing interfaces. All the code is the same, but now factored into tiny testable bite sized files/functions.

This also includes a new unit (as in the runtime sub modules is the unit) test suite.


app/packages/operators/src/runtime/hooks/useExectuionContext.test.tsx
jleven, line 1: There's a typo in the name of this file:

```diff
-useExectuionContext.test.tsx
+useExecutionContext.test.tsx
```

app/packages/operators/src/runtime/hooks/useInvocationRequestQueue.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * useInvocationRequestQueue
+ *
+ * A hook to manage the queue of invocation requests, providing methods to handle success and error events.
+ *
+ * @returns {UseInvocationRequestQueueReturn} - An object containing requests, success, and error handling functions.
+ */
+export default function useInvocationRequestQueue(): UseInvocationRequestQueueReturn {
+  const ref = useRef<InvocationRequestQueue>();
```
kaixi-wang, line 18: What is the lifecycle of the queue?

ritch: I've added [a readme](https://github.com/voxel51/fiftyone/blob/5ac6ec44461b071fcfc81628912c8f3dff081f6c/app/packages/operators/src/runtime/README.md) to answer this question!

PR #4969, opened 2024-10-23
fix non-editable views state issue and conditional row actions for TableView

imanjra:

fix non-editable views state issue and conditional row actions for TableView

Using a python panel

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced a new utility function to enhance schema handling.
	- Added a constant to define non-editable view names.
	- New property added to customize the selected cell's background color in the table component.
- **Improvements**
	- Refactored row actions in the TableView component for dynamic retrieval based on the current row.
	- Enhanced logic for state initialization in the DynamicIO component, ensuring it only initializes for editable schemas.
- **Style**
	- Minor adjustments to the formatting and alignment of table cells.


imanjra:
Converting to draft to fix highlight colouring

app/packages/core/src/plugins/SchemaIO/utils/index.ts
```diff
@@ -92,6 +92,27 @@ const COMPOSITE_VIEWS = [
   "ButtonGroupView",
 ];
 
+const NON_EDITABLE_VIEWS = [
```
ritch, line 95: This seems hard to maintain. I would rather we define this in each view somehow or in a way that we won't forget to add things to this list. In other words I don't think we'll be able to maintain this list.

imanjra: Yah, good point. I will follow-up. I would say let's go with this for now to unblock Minh's panel for release

minhtuev (approved):
Bug fix works for TableView, verified here: 
https://www.loom.com/share/581654042f7e42b192beafe9ace62d09

Thanks Ibrahim! ✅

PR #4970, opened 2024-10-23
externalize all fo pacakges

sashankaryal:
This PR adds to the list of externalized FO packages. This is relevant in JS plugin build.

## Summary by CodeRabbit

- **New Features**
	- Enhanced global scope by adding new properties to the `Window` interface, providing access to functionalities from the `@fiftyone` library.
	- Introduced multiple new modules for improved data handling and visualization capabilities.

PR #4984, opened 2024-10-24
fix externalize bug

sashankaryal:
#4970 introduced changes that externalized more FO packages to be used in JS plugins. However, some of the packages externalized is causing the app to not build (specifically, `@fiftyone/plugin` module is not being resolved properly). The reason is unknown and will be investigated later.

This PR removes the culprit packages. This is a fine trade-off for now since these packages aren't expected to be used in JS plugins.

## Summary by CodeRabbit

- **New Features**
	- Streamlined import process for the `Looker3d` module, enhancing accessibility for users.
	- Updated global declarations to reflect new dependencies, improving module organization.

- **Bug Fixes**
	- Removed outdated imports and properties, ensuring a cleaner and more efficient codebase.

### DIFF
--- a/app/packages/looker-3d/package.json
+++ b/app/packages/looker-3d/package.json
@@ -32,10 +32,12 @@
         "@types/react": "^18.2.48",
         "@types/react-dom": "^18.2.18",
         "@types/three": "^0.169.0",
+        "@vitejs/plugin-react": "^1.3.0",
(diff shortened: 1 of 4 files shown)


minhtuev (approved):
👍

ritch (approved):

PR #4993, opened 2024-10-25
SliderView, DropDownView minor improvements

manivoxel51:

- don't show the unit dropdown after chat with design - use % for now and wait for feedback
- SliderView is now using defaultValue and is uncontrolled!
- DropdownView gets a new prop to allow re-clicking a dropdown option if enabled. This helps for example the cliking on "Reset Threshold" in Data Quality Panel.
- uncontrolled slider - use defaultValue instead of value prop
- Had to use useEffect for resetting data from external sources.

Screen Shot 2024-10-25 at 3 27 29 PM

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Standardized percentage formatting for slider inputs with updated labels ("Min %" and "Max %").
	- Removed unit selection option from slider input fields.
	- Enhanced dropdown interactivity with the addition of `addOnClickToMenuItems` for immediate updates upon menu item selection.
- **Bug Fixes**
	- Improved handling and validation of percentage values in input fields.

### DIFF
--- a/app/packages/core/src/plugins/SchemaIO/components/SliderView.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/SliderView.tsx
@@ -198,7 +198,7 @@ export default function SliderView(props) {
           }}
         >
           <MenuItem value="%">%</MenuItem>
-          {/* <MenuItem value="flt">flt</MenuItem> */}
+          <MenuItem value="flt">flt</MenuItem>
         </Select>
       </FormControl>
     ),
--- a/app/packages/core/src/plugins/SchemaIO/components/SliderView.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/SliderView.tsx
@@ -1,18 +1,10 @@
-import {
-  Box,
-  FormControl,
-  Grid,
-  MenuItem,
-  Select,
-  Slider,
-  TextField,
-  Typography,
-} from "@mui/material";
-import { isNumber } from "lodash";
-import React, { useEffect, useMemo, useRef, useState } from "react";
+import { Box, Grid, Slider, TextField, Typography } from "@mui/material";
+import { isNumber, isEqual } from "lodash";
+import React, { useEffect, useRef, useState } from "react";
 import { useKey } from "../hooks";
 import { autoFocus, getComponentProps } from "../utils";
 import FieldWrapper from "./FieldWrapper";
+import { ViewPropsType } from "../utils/types";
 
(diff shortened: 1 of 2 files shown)


lanzhenw (approved):

app/packages/core/src/plugins/SchemaIO/components/SliderView.tsx
```diff
@@ -79,17 +74,18 @@ export default function SliderView(props) {
     multipleOf: schemaMultipleOf,
     view,
   } = schema;
+
   const {
     min: viewMin,
     max: viewMax,
-    value_format: valueFormat = "%",
```
Br2850, line 85: I feel like we should keep the option to define the valueFormat from the Python context still. Because we don’t want to come back and change all this later, since we are now defining this outside of Python’s control as a const in React

app/packages/core/src/plugins/SchemaIO/components/SliderView.tsx
```diff
@@ -245,20 +241,20 @@ export default function SliderView(props) {
             <Grid container justifyContent="space-between" pl={1}>
               <Grid item pl={labelPosition === "left" ? "100px" : "0"}>
                 <SliderInputField
-                  label="Min"
+                  label="Min %"
```
Br2850, line 244: Instead of harcoding this, can we just make this a variable? So if someone in Python defines valueFormat = %, we can map this in as label[valueFormat]

Idk if we’ll end up adding icons as labels instead like talked about in Slack, or even just text like “units”, but this would make it more flexible

Br2850 (commented):
Left 2 comments about considering code for the future. I think it’s best we still give the Python Panel user the power when it comes to how the uniting looks. But if they choose %, we just render all the % options that are currently being hardcoded

app/packages/core/src/plugins/SchemaIO/components/SliderView.tsx
manivoxel51: makes sense

app/packages/core/src/plugins/SchemaIO/components/SliderView.tsx
manivoxel51: makes sense - I included the unit at minimum for now - we should ask design how she wants to handle this

Br2850:
Awesome LGTM

lanzhenw:
LGTM

lanzhenw (approved):

PR #5018, opened 2024-10-31
remove reload_on_navigation panel option

sashankaryal:
Panel component subscribing to `fos.modalUniqueId` has negative performance consequences on the app, most importantly the sample modal plugin. This PR removes that.

## Motivation

Keying the panel component by "modal unique ID", which is a concatenation of sample and group id, leads to a slippery slope where the contract is not well-defined. This behavior was added to support `shouldReloadOnNavigation` panel option for modal panels, but what is "navigation" anyway? This gets tricky in dynamic groups of groups, for instance. Also, `key=foo` is a bad react pattern.

Removing this panel option also incentivizes the plugin author to write better code where they control the scope and maintain strict ownership of the resources they create / subscribe to, and their lifecycles.

## Summary by CodeRabbit

- **New Features**
	- Introduced an optional `reloadOnNavigation` property for future plugin functionality (commented out).
  
- **Bug Fixes**
	- Removed the `reloadOnNavigation` option from various components and methods, streamlining the panel registration and rendering process.
  
- **Refactor**
	- Simplified the `Panel` component by eliminating unnecessary state dependencies and rendering logic.
  
- **Documentation**
	- Updated documentation to reflect the removal of the `reload_on_navigation` parameter in relevant classes and methods.

### DIFF
--- a/app/packages/operators/src/Panel/register.tsx
+++ b/app/packages/operators/src/Panel/register.tsx
@@ -23,7 +23,6 @@ export default function registerPanel(ctx: ExecutionContext) {
     panelOptions: {
       allowDuplicates: ctx.params.allow_duplicates,
       helpMarkdown: ctx.params.help_markdown,
-      reloadOnNavigation: ctx.params.reload_on_navigation,
       surfaces: ctx.params.surfaces,
     },
   });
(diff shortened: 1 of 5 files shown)


benjaminpkane (approved):
LGTM 🚀

PR #5029, opened 2024-11-01
Adding min() and max() aggregations

brimoor:
Adds `min()` and `max()` aggregations.

These can be computed via `bounds()[0]` and `bounds()[1]` today, but it's nice to have dedicated methods because:
- These are more efficient when you don't need both extrema
- Yields more readable code when you want only a particular extremum

## Summary by CodeRabbit

- **New Features**
	- Introduced new aggregation methods for computing minimum (`min()`) and maximum (`max()`) values in datasets.
	- Added new public entities, `Min` and `Max`, to enhance aggregation functionalities.
	- Updated documentation to include examples for the new aggregation methods.

- **Bug Fixes**
	- Improved handling of summary fields to ensure accurate updates based on dataset modifications.

- **Tests**
	- Added unit tests for the new `min` and `max` functionalities to validate their performance and correctness.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -659,6 +659,20 @@ class SampleCollection(object):
         return self._get_extremum(path, -1)
 
     def _get_extremum(self, path, order):
+        #
+        # This method exists in addition to `min()` and `max()` aggregations
+        # for two reasons:
+        #
+        # 1. `$sort + $limit 1` is more efficient than `$group _id: None` when
+        #    the field is indexed
+        #
+        # 2. When `path` is a frame-level field, these methods are optimized to
+        #    directly aggregate on the frames collection, which is something
+        #    that the Aggregation classes do not yet support. In other words,
+        #    `dataset._max("frames.last_modified_at")` is currently more
+        #    performant than `dataset.max("frames.last_modified_at")`
+        #
+
         path, is_frame_field = self._handle_frame_field(path)
         path = self._handle_db_field(path, frames=is_frame_field)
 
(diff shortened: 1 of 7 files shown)


fiftyone/core/collections.py
```diff
@@ -646,55 +646,58 @@ def _sync_samples_last_modified_at(self):
...
 
-    def _get_last_modified_at(self, frames=False):
-        if frames and not self._contains_videos(any_slice=True):
+    def _min(self, path):
+        return self._get_extremum(path, 1)
+
+    def _max(self, path):
+        return self._get_extremum(path, -1)
+
+    def _get_extremum(self, path, order):
```
brimoor, line 661: This exists in addition to `min()` and `max()` for two reasons:
1. `$sort + $limit 1` is more efficient than `$group _id: None` when the field is indexed
2. When `path` is a frame-level field, these methods are optimized to directly aggregated on the frames collection, which is something that the `Aggregation` classes do not yet support. In other words, `dataset._max("frames.last_modified_at")` is more performant than `dataset.max("frames.last_modified_at")`

swheaton (approved):
lgtm, just recommend your review comment be in code for future readers

tests/unittests/aggregation_tests.py
```diff
@@ -233,6 +233,92 @@ def test_sum(self):
...
+        s.save()
+        self.assertEqual(d.max("detections.detections.confidence"), 1)
+        self.assertEqual(d.max(1 + F("detections.detections.confidence")), 2)
+
+        d = fo.Dataset()
+        s = fo.Sample(filepath="video.mp4")
+        s[1]["detection"] = fo.Detection(label="label", confidence=1)
+        d.add_sample(s)
+        self.assertEqual(d.max("frames.detection.confidence"), 1)
+
```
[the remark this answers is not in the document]
brimoor: fancy

fiftyone/core/collections.py
brimoor: Good call, just added 👍

PR #5032, opened 2024-11-01
Adds readonly property to PillBadge and Modal button

manivoxel51:

- adds a readonly property to PillBadge and ModalView - needed for enforcing permissioning in teams
- adds an optional tooltipTitle property so we can show things like "you don't have sufficient permissions." in teams context. @imanjra lmk if there is a better way and I'll followup with improvements.

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced a `disabled` property to the `ModalButtonView` for enhanced button control.
	- Added `readOnly` and `tooltipTitle` properties to the `PillBadge` component, allowing it to render in a non-interactive state and provide additional context.
	- Updated the `PillBadgeView` to reflect the new `readOnly` prop for better schema integration.
	- Added a `tooltipTitle` property to the `DropdownView` for enhanced user guidance.

- **Improvements**
	- Enhanced button interactivity based on tagging functionality.
	- Improved usability of the `PillBadge` component by restricting user input when in read-only mode.
	- Adjusted the default cursor style in the `ImageView` component to provide clearer user interaction feedback.

### DIFF
--- a/app/packages/components/src/components/PillBadge/PillBadge.tsx
+++ b/app/packages/components/src/components/PillBadge/PillBadge.tsx
@@ -10,12 +10,14 @@ const PillBadge = ({
   variant = "filled",
   showIcon = true,
   operator,
+  readOnly = false,
 }: {
   text: string | string[] | [string, string][];
   color?: string;
   variant?: "filled" | "outlined";
   showIcon?: boolean;
   operator?: () => void;
+  readOnly?: boolean;
 }) => {
   const getInitialChipSelection = (
     text: string | string[] | [string, string][]
@@ -90,6 +92,7 @@ const PillBadge = ({
       ) : (
         <FormControl fullWidth>
           <Chip
+            disabled={readOnly}
             icon={
               showIcon ? (
                 <CircleIcon color={"inherit"} sx={{ fontSize: 10 }} />
--- a/app/packages/components/src/components/PillBadge/PillBadge.tsx
+++ b/app/packages/components/src/components/PillBadge/PillBadge.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from "react";
 import CircleIcon from "@mui/icons-material/Circle";
-import { Chip, FormControl, MenuItem, Select } from "@mui/material";
+import { Chip, FormControl, MenuItem, Select, Tooltip } from "@mui/material";
 import { usePanelEvent } from "@fiftyone/operators";
 import { usePanelId } from "@fiftyone/spaces";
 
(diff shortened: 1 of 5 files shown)


app/packages/components/src/components/PillBadge/PillBadge.tsx
```diff
@@ -90,6 +92,7 @@ const PillBadge = ({
       ) : (
         <FormControl fullWidth>
           <Chip
+            disabled={readOnly}
```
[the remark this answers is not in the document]
manivoxel51: fair point

lanzhenw (approved):

app/packages/components/src/components/PillBadge/PillBadge.tsx
```diff
@@ -10,12 +10,16 @@ const PillBadge = ({
   variant = "filled",
   showIcon = true,
   operator,
+  readOnly = false,
```
Br2850, line 13: Why is this needed for PillBadge?

To set a readOnly PillBadge view you simply pass down a singular string. If only a string is given to this component (with or without a color) it becomes unselectable and is no longer a writeable component

Br2850 (changes requested):
ModalButtonView doesn't have a disabled field at the parent level so that makes sense

But I'm a bit confused as to why the changes to PillBadge & PillBadgeView are needed? If you just pass in a string, the component is automatically read-only

Here's an example: 
```
schema = {"text": "Disabled", "color": "secondary"}
badge = types.PillBadgeView(**schema)
```

app/packages/components/src/components/PillBadge/PillBadge.tsx
manivoxel51: if we simply use 
```
schema = {"text": "Disabled", "color": "secondary"}
badge = types.PillBadgeView(**schema)
```

The badge doesn't reflect it is a dropdown select - the downward caret will be missing. disabled / readOnly are legit properties to use in common components. if you really don't like the readOnly prop, happy to huddle and resolve!

Br2850: But isn't that the point of readOnly? If I should only be reading the state of something, I don't understand why they would need to see the dropdown caret?

Having a dropdown caret implies that they can change the state of the pill, but we don't want that

So we should just show the pill and whatever the current status is without the ability to change it - which is just the string & a color choice

Br2850 (approved):
After demo, looks good, this gives us options if we need it

PR #5038, opened 2024-11-04
additional panel meta (is_new, beta, category)

ritch:
Here's an example using all of the new settings in python and JS:

```python
class CounterExample(foo.Panel):
    @property
    def config(self):
        return foo.PanelConfig(
            name="example_counter",
            label="Examples: Counter",
            category=foo.Categories.ANALYZE, # new
            is_new=foo.is_new("2024-11-01"), # new
            beta=True # new,
            priority=52000 # new
        )
```

```typescript
registerComponent({
  name: "Embeddings",
  label: "Embeddings",
  component: Embeddings,
  type: PluginComponentType.Panel,
  activator: () => true,
  Icon: WorkspacesIcon,
  panelOptions: {
    TabIndicator: EmbeddingsTabIndicator,
    priority: BUILT_IN_PANEL_PRIORITY_CONST,
    category: Categories.Curate, // new
    isNew: true, // new
    beta: true, // new
    priority: 51001 // existing
  },
});
```

New information is displayed only in the `AddPanel*` component.

image
image

## Release Notes

- **New Features**
	- Introduced a categorization system for panels, allowing users to easily identify and organize panels by their respective categories.
	- Added visual indicators for "NEW" and "BETA" features in the UI, enhancing user awareness of recent additions and experimental features.

- **Improvements**
	- Enhanced the `AddPanelButton` and `AddPanelItem` components for better usability and organization.
	- Improved panel registration capabilities with new metadata options, including priority and categorization.

- **Bug Fixes**
	- Minor adjustments to ensure consistent behavior across components.

### DIFF
--- a/app/packages/plugins/src/index.ts
+++ b/app/packages/plugins/src/index.ts
@@ -320,7 +320,6 @@ export function getCategoryLabel(category: CategoryID): string {
 }
 
 export function getCategoryForPanel(panel: PluginComponentRegistration) {
-  console.log(panel);
   return panel.panelOptions?.category || "custom";
 }
 
--- a/app/packages/plugins/src/index.ts
+++ b/app/packages/plugins/src/index.ts
@@ -319,6 +319,11 @@ export function getCategoryLabel(category: CategoryID): string {
   }
 }
 
+export function getCategoryForPanel(panel: PluginComponentRegistration) {
+  console.log(panel);
+  return panel.panelOptions?.category || "custom";
+}
+
 type Category = {
   id: CategoryID;
   label: string;
--- a/app/packages/plugins/src/index.ts
+++ b/app/packages/plugins/src/index.ts
@@ -297,6 +297,33 @@ export enum PluginComponentType {
    */
 }
 
+type CategoryID = "import" | "curate" | "analyze" | "custom";
+
+export enum Categories {
+  Import = "import",
+  Curate = "curate",
+  Analyze = "analyze",
+  Custom = "custom",
+}
+
+export function getCategoryLabel(category: CategoryID): string {
+  switch (category) {
+    case "import":
+      return "Import";
+    case "curate":
+      return "Curate";
+    case "analyze":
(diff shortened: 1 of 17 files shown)


minhtuev (commented):
Thanks Ritchie, meta info look good for me, not sure how to determine priority value thoug.

app/packages/plugins/src/index.ts
```diff
@@ -297,6 +297,37 @@ export enum PluginComponentType {
    */
 }
 
+type CategoryID = "import" | "curate" | "analyze" | "custom";
```
Br2850, line 300: Given that @brimoor this was your original vision for the categories of workflows with Assistant, do we want to simply just add all of these at right now?

Not sure if this list is still the "right" list in terms of categorization, but food for thought @ritch 

Screenshot 2024-11-04 at 3 13 35 PM

fiftyone/operators/utils.py
```diff
@@ -66,3 +67,40 @@ def is_method_overridden(base_class, sub_class_instance, method_name):
...
+            Defaults to 30.
+
+    Returns:
+        bool: True if the release date is within the specified number of days, False otherwise.
+
+    Examples:
+        >>> is_new("2024-11-09")  # Using a string date and default days
+        True  # if today's date is within 30 days after 2024-11-09
+
+        >>> is_new(datetime(2024, 11, 9), days=15)  # Using a datetime object with 15-day threshold
```
Br2850, line 95: That's cool, I like that we can manually define when `NEW` should disappear 🙌🏽

Br2850 (approved):
Looks great! 🎉 Similar sentiment to Minh' not sure how we would self define priority - maybe it should be a calculation of (new + category, etc.) but other than that this is super awesome!

app/packages/plugins/src/index.ts
ritch: We've been discussing the list in a slack thread

PR #5048, opened 2024-11-06
Bug fixes for Query Performance Toast

CamronStaley:

Done:
- Change dismiss to be saved in session storage
- Change button size to match field sizes
- Make toasts show up everytime there's an event not just the first time
- When toast is on a frame filter field we will point users to the [summary page](https://docs.voxel51.com/user_guide/using_datasets.html#summary-fields)
- Updated toast time shown to 10 seconds 
- Don't show toast if field is already indexed
- Fire queryperformance event whenever one of the two calls are slower than 5 seconds:
    -    [lightningModeQuery](https://github.com/voxel51/fiftyone/blob/d83ee2d21402a94c699be5c4d0fe6cee7c8cd2ed/app/packages/state/src/recoil/queryPerformance.ts#L21)
    - [paginateSamplesQuery](https://github.com/voxel51/fiftyone/blob/c940dd82c1e22b58a1d7142d7da356711f8074ec/app/packages/core/src/components/Grid/useSpotlightPager.ts#L82)

Changed TIMEOUT to 0 so the toast is always triggered and then added a console log statement to QPToast to show that the path is set accordingly.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced `NumericFieldFilter` component to conditionally calculate bounds based on query performance.
	- Extended display duration of the `QueryPerformanceToast` from 5 to 10 seconds.
	- Introduced a new link for summary fields in the user guide.

- **Bug Fixes**
	- Improved interaction logic in the `QueryPerformanceToast` for better user experience.

- **Refactor**
	- Removed deprecated `QueryPerformance` components and related event handling from the `FilterablePathEntries` component.

### DIFF
--- a/app/packages/core/src/components/Filters/NumericFieldFilter/state.ts
+++ b/app/packages/core/src/components/Filters/NumericFieldFilter/state.ts
@@ -7,9 +7,12 @@ export const FLOAT_NONFINITES: Nonfinite[] = ["inf", "ninf", "nan"];
 export const hasBounds = selectorFamily({
   key: "hasBounds",
   get:
-    (params: { path: string; modal: boolean }) =>
+    (params: { path: string; modal: boolean; shouldCalculate: boolean }) =>
     ({ get }) => {
-      return Boolean(get(boundsAtom(params))?.every((b) => b !== null));
+      const shouldCalculate = params.shouldCalculate ?? true;
+      return shouldCalculate
+        ? Boolean(get(boundsAtom(params))?.every((b) => b !== null))
+        : Boolean(false);
     },
 });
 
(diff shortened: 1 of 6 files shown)


app/packages/core/src/components/Filters/NumericFieldFilter/state.ts
```diff
@@ -7,9 +7,12 @@ export const FLOAT_NONFINITES: Nonfinite[] = ["inf", "ninf", "nan"];
 export const hasBounds = selectorFamily({
   key: "hasBounds",
   get:
-    (params: { path: string; modal: boolean }) =>
+    (params: { path: string; modal: boolean; shouldCalculate: boolean }) =>
     ({ get }) => {
-      return Boolean(get(boundsAtom(params))?.every((b) => b !== null));
+      const shouldCalculate = params.shouldCalculate ?? true;
```
CamronStaley, line 12: I added this because I was seeing that whenever a sidebar section was opened we were calling to lightningModeQuery to get the bounds which in QP mode we should never see that until a user actually clicks a button to view a slider or opens a dropdown. This was firing off because if we aren't in QP mode we want to check if something has bounds to determine if we want to show it at all instead of just saying "No Results". This way in QP we don't check bounds until the user selects the field and in non QP mode we always check it.

minhtuev (approved):
Nice great work thanks Cam!

PR #5059, opened 2024-11-07
panel meta fixes

ritch:
image

Improved padding / margin for panel metadata. Also fixed issue where NEW and BETA would show at the same time.


Br2850 (approved):
👍🏽 LGTM I just manually got rid of the usage of `is_new` instead of calling both, but this is a good failsafe if you feel like folks might not remember this in the future

imanjra (approved):
LGTM

PR #5086, opened 2024-11-11
Exection store interface cleanup

brimoor:
## Change log

- bug fix: must use `datetime.utcnow()` rather than `datetime.now()` to ensure TTL cleanup works as expected
- Updated the `ExecutionStoreRepo` interface so that all methods are restricted to the current dataset context (or the global datasetless context). Previously some methods were scoped to the current dataset while others were not, which could cause unexpected behavior. New `list_stores_global()` and `has_store_global()` methods are added that really do query across all contexts, if this is desired
- added a `foo.drop_orphan_stores()` utility
- added private `_has_store()`, `_list_stores()`, and `_get_store()` utils to the `SampleCollection` class for internal usage
- general linting of docstrings to be Sphinx-friendly

## Example usage

```py
import fiftyone as fo
import fiftyone.operators.store as foos

d = fo.Dataset()

svc = foos.ExecutionStoreService(dataset_id=d._doc.id)
svc.create_store("test")
svc.set_key("test", "foo", {"spam": "eggs"})

print(d._has_stores())
print(d._list_stores())
store = d._get_store("test")
print(store.get("foo"))

foo.delete_dataset(d.name, dry_run=True)
foo.delete_dataset(d.name, dry_run=False)

foo.drop_orphan_stores(dry_run=True)
foo.drop_orphan_stores(dry_run=False)
```

## Summary by CodeRabbit

- **New Features**
	- Added methods for managing execution stores, including checking for stores, listing stores, and fetching specific stores.
	- Introduced a function to remove orphaned execution stores associated with non-existent datasets.
	- Enhanced the `ExecutionStoreService` with new methods for store and key management.

- **Bug Fixes**
	- Updated internal logic for dataset deletion to improve cleanup processes.

- **Documentation**
	- Improved documentation clarity for several methods and functions throughout the codebase.

- **Tests**
	- Enhanced unit tests to ensure dataset ID handling in execution store operations.

### DIFF
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -580,6 +580,36 @@ def drop_orphan_runs(dry_run=False):
             _delete_run_results(conn, orphan_result_ids)
 
 
+def drop_orphan_stores(dry_run=False):
+    """Drops all orphan execution stores from the database.
+
+    Orphan stores are those that are associated with a dataset that no longer
+    exists in the database.
+
+    Args:
+        dry_run (False): whether to log the actions that would be taken but not
+            perform them
+    """
+    conn = get_db_conn()
+    _logger = _get_logger(dry_run=dry_run)
+
+    dataset_ids = set(conn.datasets.distinct("_id"))
+
+    store_ids = set(conn.execution_store.distinct("dataset_id"))
+    store_ids.discard(None)
+
+    orphan_store_ids = list(store_ids - dataset_ids)
+
+    if orphan_store_ids:
+        _logger.info(
+            "Deleting %d orphan store(s): %s",
+            len(orphan_store_ids),
+            orphan_store_ids,
+        )
+        if not dry_run:
+            _delete_stores(conn, orphan_store_ids)
+
+
 def stream_collection(collection_name):
     """Streams the contents of the collection to stdout.
 
@@ -1127,15 +1157,12 @@ def delete_dataset(name, dry_run=False):
         if not dry_run:
             _delete_run_results(conn, result_ids)
 
-    dataset_id = dataset_dict["_id"]
-    svc = foos.ExecutionStoreService(dataset_id=dataset_id)
-    num_docs = sum(
-        len(svc.list_keys(store_name)) for store_name in svc.list_stores()
-    )
-    if num_docs > 0:
-        _logger.info("Deleting %d store doc(s)", num_docs)
+    svc = foos.ExecutionStoreService(dataset_id=dataset_dict["_id"])
+    num_stores = svc.count_stores()
+    if num_stores > 0:
+        _logger.info("Deleting %d store(s)", num_stores)
         if not dry_run:
-            foos.cleanup_store_for_dataset(dataset_id)
+            svc.cleanup()
 
 
 def delete_saved_view(dataset_name, view_name, dry_run=False):
@@ -1613,6 +1640,10 @@ def _delete_run_docs(conn, run_ids):
 def _delete_run_results(conn, result_ids):
     conn.fs.files.delete_many({"_id": {"$in": result_ids}})
     conn.fs.chunks.delete_many({"files_id": {"$in": result_ids}})
+
+
+def _delete_stores(conn, dataset_ids):
+    conn.execution_store.delete_many({"dataset_id": {"$in": dataset_ids}})
 
 
 _RUNS_FIELDS = ["annotation_runs", "brain_methods", "evaluations", "runs"]
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -37,7 +37,6 @@ fob = fou.lazy_import("fiftyone.core.brain")
 fod = fou.lazy_import("fiftyone.core.dataset")
 foe = fou.lazy_import("fiftyone.core.evaluation")
 fors = fou.lazy_import("fiftyone.core.runs")
-foos = fou.lazy_import("fiftyone.operators.store")
 
 
 logger = logging.getLogger(__name__)
@@ -1157,12 +1156,14 @@ def delete_dataset(name, dry_run=False):
         if not dry_run:
             _delete_run_results(conn, result_ids)
 
-    svc = foos.ExecutionStoreService(dataset_id=dataset_dict["_id"])
-    num_stores = svc.count_stores()
+    _id = dataset_dict["_id"]
+    num_stores = conn.execution_store.count_documents(
+        {"dataset_id": _id, "key": "__store__"}
+    )
     if num_stores > 0:
         _logger.info("Deleting %d store(s)", num_stores)
         if not dry_run:
-            svc.cleanup()
+            conn.execution_store.delete_many({"dataset_id": _id})
 
 
 def delete_saved_view(dataset_name, view_name, dry_run=False):
--- a/fiftyone/factory/repos/execution_store.py
+++ b/fiftyone/factory/repos/execution_store.py
@@ -52,8 +52,7 @@ class ExecutionStoreRepo(object):
                 store_name=store_name,
                 key="__store__",
                 dataset_id=self._dataset_id,
-            ),
-            {},
+            )
         )
         return bool(result)
 
--- a/fiftyone/factory/repos/execution_store.py
+++ b/fiftyone/factory/repos/execution_store.py
@@ -1,35 +1,30 @@
 """
 Execution store repository.
+
+| Copyright 2017-2024, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
 """
 
-import datetime
-import bson
+from datetime import datetime
+
+from bson import ObjectId
 from pymongo.collection import Collection
+
 from fiftyone.operators.store.models import StoreDocument, KeyDocument
 
 
-def _where(store_name, key=None, dataset_id=None):
-    query = {"store_name": store_name}
-    if key is not None:
-        query["key"] = key
-    if dataset_id is not None:
-        query["dataset_id"] = dataset_id
-    return query
-
-
-class ExecutionStoreRepo:
+class ExecutionStoreRepo(object):
     """Base class for execution store repositories."""
 
     COLLECTION_NAME = "execution_store"
 
-    def __init__(
-        self, collection: Collection, dataset_id: bson.ObjectId = None
-    ):
+    def __init__(self, collection: Collection, dataset_id: ObjectId = None):
         self._collection = collection
         self._dataset_id = dataset_id
 
     def create_store(self, store_name, permissions=None) -> StoreDocument:
-        """Creates a store in the execution store."""
+        """Creates a store associated with the current context."""
         store_doc = StoreDocument(
             store_name=store_name,
             value=permissions,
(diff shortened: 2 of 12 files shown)


fiftyone/factory/repos/execution_store.py
```diff
@@ -38,15 +33,46 @@ def create_store(self, store_name, permissions=None) -> StoreDocument:
...
+    def count_stores(self) -> int:
+        """Counts the stores associated with the current context."""
+        return self._collection.count_documents(
+            dict(key="__store__", dataset_id=self._dataset_id),
+        )
+
+    def delete_store(self, store_name) -> int:
+        """Deletes the specified store."""
+        result = self._collection.delete_many(
+            dict(store_name=store_name, dataset_id=self._dataset_id)
```
brimoor, line 67: I think it is important to explicitly include `dataset_id: None` in this query (and all similar queries) when using an `ExecutionStoreService()` with no dataset context. As is, someone working with no dataset context could delete a store associated with a dataset.

In other words, I think there are `n + 1` possible contexts: one for each dataset, plus one additional "no dataset" context that is scoped to stores that are not associated with a dataset (which we are not currently using, but we could use soon to store app-wide state).

I added a  couple `XXX_global()` methods that really do operate across the entire execution store. But I think users should have to explicitly opt into those methods, rather than the current behavior where some methods are scoped to datasets but then other methods are "global" with `XXX_for_dataset()` methods that are scoped specifically to the dataset.

fiftyone/core/odm/database.py
```diff
@@ -1615,4 +1642,8 @@ def _delete_run_results(conn, result_ids):
     conn.fs.chunks.delete_many({"files_id": {"$in": result_ids}})
 
 
+def _delete_stores(conn, dataset_ids):
+    conn.execution_store.delete_many({"dataset_id": {"$in": dataset_ids}})
```
ritch, line 1646: Shouldn't this use `ExecutionStoreService` ? Or if not what is the reason to deviate from the pattern in delete()

fiftyone/factory/repos/execution_store.py
```diff
@@ -1,35 +1,30 @@
...
     COLLECTION_NAME = "execution_store"
 
-    def __init__(
-        self, collection: Collection, dataset_id: bson.ObjectId = None
-    ):
+    def __init__(self, collection: Collection, dataset_id: ObjectId = None):
         self._collection = collection
         self._dataset_id = dataset_id
 
     def create_store(self, store_name, permissions=None) -> StoreDocument:
```
ritch, line 26: While we're at it - should remove reference to permissions

fiftyone/factory/repos/execution_store.py
```diff
@@ -1,35 +1,30 @@
...
-        """Creates a store in the execution store."""
+        """Creates a store associated with the current context."""
```
ritch, line 27: context seems misleading here. Its simpler: if a `dataset_id` is provided, it will associate stores with that dataset, and they will be deleted when the dataset is deleted. Context here seems like an overloaded term.

fiftyone/factory/repos/execution_store.py
```diff
@@ -38,15 +33,46 @@ def create_store(self, store_name, permissions=None) -> StoreDocument:
         self._collection.insert_one(store_doc.to_mongo_dict())
         return store_doc
 
+    def has_store(self, store_name):
+        """Checks whether a store with the given name exists in the current
+        context.
```
ritch, line 38: Same comment as above.

ritch: Need to add test(s) for this

fiftyone/factory/repos/execution_store.py
ritch: > I think it is important to explicitly include dataset_id: None in this query

Yes... originally I was trying to avoid having a `dataset_id: null` in every document... however I agree with the explicitness, because `{key: 'blah'}` is an ambiguous query.

fiftyone/factory/repos/execution_store.py
```diff
@@ -38,15 +33,46 @@ def create_store(self, store_name, permissions=None) -> StoreDocument:
...
+        result = self._collection.delete_many(
+            dict(store_name=store_name, dataset_id=self._dataset_id)
+        )
+        return result.deleted_count
 
     def set_key(self, store_name, key, value, ttl=None) -> KeyDocument:
-        """Sets or updates a key in the specified store"""
-        now = datetime.datetime.now()
+        """Sets or updates a key in the specified store."""
+        now = datetime.utcnow()
```
ritch, line 73: Good catch. This points at missing true mongo tests which would have caught this when running in CI. Note that I did test these locally, however my mongo had the same timezone.

We should still do the same manual testing now that we've changed this logic. I'll do that before we merge this.

fiftyone/factory/repos/execution_store.py
```diff
@@ -102,53 +128,79 @@ def set_key(self, store_name, key, value, ttl=None) -> KeyDocument:
...
+            dict(store_name=store_name, dataset_id=self._dataset_id),
+            {"key": 1},
+        )
+        return [d["key"] for d in result]
+
+    def count_keys(self, store_name) -> int:
+        """Counts the keys in the specified store."""
+        return self._collection.count_documents(
+            dict(store_name=store_name, dataset_id=self._dataset_id)
+        )
```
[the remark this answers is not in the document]
ritch: This is a good suggestion. I also think this could be added to the `_where()` util instead of removing it. That util could also be a method like so:

```python
query = self._where_key(store_name) # handles $ne and self.dataset_id
```

fiftyone/core/odm/database.py
brimoor: This module is like a "break glass in case of emergency" module that can cleanup stuff when the builtin operations fail. For example, it has a `delete_dataset()` method that will delete all remnants of a dataset that can't be loaded via `fo.load_dataset()`.

So `drop_orphan_stores()` is implemented without relying on `ExecutionStoreService` so that if, hypothetically, `ExecutionStoreService` required the dataset to actually exist (which it currently does not, but could in the future), this method would still be able to do the garbage collection by going straight to the database.

fiftyone/factory/repos/execution_store.py
brimoor: Done in https://github.com/voxel51/fiftyone/pull/5086/commits/0ffb9838b0b8f457a1ddd9e16cb341053c4a29ec

fiftyone/factory/repos/execution_store.py
brimoor: I updated the class docstrings to explain what "context" means: https://github.com/voxel51/fiftyone/pull/5086/commits/07a515faa178c42c92542ca095f3affc129b9611

fiftyone/factory/repos/execution_store.py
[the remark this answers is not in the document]
brimoor: I implemented the `{"$ne": "__store__"}` suggestion in https://github.com/voxel51/fiftyone/pull/5086/commits/800cdad9963a8bd3cc15d525513456214e9e8ced.

It felt a bit magic to me to embed this logic in a `_where()` utility for minimal benefit. Just my opinion

fiftyone/factory/repos/execution_store.py
```diff
@@ -102,53 +128,79 @@
...
 
     def delete_key(self, store_name, key) -> bool:
-        """Deletes the document that matches the store name and key."""
+        """Deletes the document that matches the store name and key, if one
+        exists.
+        """
         result = self._collection.delete_one(
-            _where(store_name, key, self._dataset_id)
+            dict(store_name=store_name, key=key, dataset_id=self._dataset_id)
         )
```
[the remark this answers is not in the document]
brimoor: It's important to include `{"dataset_id": None}` in the query. This is not the same thing as `{}`.

fiftyone/factory/repos/execution_store.py
```diff
@@ -102,53 +128,79 @@
...
-
     def update_ttl(self, store_name, key, ttl) -> bool:
         """Updates the TTL for a key."""
         expiration = KeyDocument.get_expiration(ttl)
         result = self._collection.update_one(
-            _where(store_name, key, self._dataset_id),
+            dict(store_name=store_name, key=key, dataset_id=self._dataset_id),
             {"$set": {"expires_at": expiration}},
         )
         return result.modified_count > 0
```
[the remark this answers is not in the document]
brimoor: It's important to include `{"dataset_id": None}` in the query. This is not the same thing as `{}`.

fiftyone/factory/repos/execution_store.py
```diff
@@ -102,53 +128,79 @@
     def get_key(self, store_name, key) -> KeyDocument:
         """Gets a key from the specified store."""
         raw_key_doc = self._collection.find_one(
-            _where(store_name, key, self._dataset_id)
+            dict(store_name=store_name, key=key, dataset_id=self._dataset_id)
         )
         key_doc = KeyDocument(**raw_key_doc) if raw_key_doc else None
```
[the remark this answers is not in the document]
brimoor: It's important to include `{"dataset_id": None}` in the query. This is not the same thing as `{}`.

fiftyone/factory/repos/execution_store.py
```diff
@@ -102,53 +128,79 @@
...
 
-    def delete_store(self, store_name) -> int:
-        """Deletes the entire store."""
+    def list_keys(self, store_name) -> list[str]:
+        """Lists all keys in the specified store."""
+        result = self._collection.find(
+            dict(store_name=store_name, dataset_id=self._dataset_id),
+            {"key": 1},
+        )
+        return [d["key"] for d in result]
```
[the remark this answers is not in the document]
brimoor: done

fiftyone/factory/repos/execution_store.py
```diff
@@ -102,53 +128,79 @@
...
+            dict(store_name=store_name, dataset_id=self._dataset_id)
+        )
+
+    def cleanup(self) -> int:
+        """Deletes all stores and keys associated with the current context."""
         result = self._collection.delete_many(
-            _where(store_name, dataset_id=self._dataset_id)
+            dict(dataset_id=self._dataset_id)
         )
         return result.deleted_count
```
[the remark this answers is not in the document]
brimoor: It's important to include `{"dataset_id": None}` in the query. This is not the same thing as `{}`.

fiftyone/factory/repos/execution_store.py
brimoor: I updated the class docstrings to explain what "context" means: https://github.com/voxel51/fiftyone/pull/5086/commits/07a515faa178c42c92542ca095f3affc129b9611

fiftyone/operators/store/service.py
```diff
@@ -117,42 +139,58 @@
...
             store_name: the name of the store
 
         Returns:
-            a list of keys in the store
+            True/False
         """
-        return self._repo.list_keys(store_name)
+        return self._repo.has_store_global(store_name)
+
+    def list_stores_global(self) -> list[StoreDocument]:
```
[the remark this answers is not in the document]
brimoor: we require `Python>=3.9`.

fiftyone/operators/store/service.py
```diff
@@ -60,10 +45,49 @@
         Returns:
             a :class:`fiftyone.store.models.StoreDocument`
         """
-        return self._repo.create_store(store_name=store_name)
+        return self._repo.create_store(store_name)
+
+    def list_stores(self) -> list[str]:
```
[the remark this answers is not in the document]
brimoor: we require `Python>=3.9`.

fiftyone/operators/store/service.py
```diff
@@ -117,42 +139,58 @@ def update_ttl(
         Returns:
             a :class:`fiftyone.store.models.KeyDocument`
         """
-        return self._repo.update_ttl(
-            store_name=store_name, key=key, ttl=new_ttl
-        )
+        return self._repo.update_ttl(store_name, key, new_ttl)
+
+    def list_keys(self, store_name: str) -> list[str]:
```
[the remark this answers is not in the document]
brimoor: we require `Python>=3.9`.

fiftyone/factory/repos/execution_store.py
```diff
@@ -1,52 +1,88 @@
...
+        """Checks whether a store with the given name exists in the current
+        context.
+        """
+        result = self._collection.find_one(
+            dict(
+                store_name=store_name,
+                key="__store__",
+                dataset_id=self._dataset_id,
+            ),
+            {},
```
ritch, line 56: What is the purpose of this? Isn't this only for `find` args, like `limit`?

ritch: Removed

ritch (approved):

PR #5095, opened 2024-11-12
ExecutionStore + MongoDB integration tests and fixes

ritch:
## Summary by CodeRabbit

- **New Features**
	- Introduced dynamic collection name support in the execution store repository.
	- Added methods for checking key existence and deleting stores globally.
	- Enhanced store creation with optional metadata.
	- New property for accessing metadata in store documents.

- **Bug Fixes**
	- Improved error handling in TTL retrieval logic.

- **Tests**
	- Comprehensive unit tests added for `ExecutionStoreService` functionalities.
	- Updated existing tests for consistency and clarity.

### DIFF
--- a/tests/unittests/execution_store_service_tests.py
+++ b/tests/unittests/execution_store_service_tests.py
@@ -18,7 +18,7 @@ TEST_COLLECTION_NAME = "execution_store_test_collection"
 
 @pytest.fixture
 def dataset():
-    return fo.Dataset(name="test_dataset")
+    return fo.Dataset()
 
 
 @pytest.fixture


fiftyone/factory/repo_factory.py
```diff
@@ -57,15 +57,16 @@ def delegated_operation_repo() -> DelegatedOperationRepo:
...
         dataset_id: Optional[ObjectId] = None,
     ) -> ExecutionStoreRepo:
-        if (
-            MongoExecutionStoreRepo.COLLECTION_NAME
-            not in RepositoryFactory.repos
-        ):
-            RepositoryFactory.repos[
-                MongoExecutionStoreRepo.COLLECTION_NAME
-            ] = MongoExecutionStoreRepo(
+        repo_key = (
```
ritch, line 60: This is a bug fix where dataset_ids were being incorrectly used on "cached" repos.

fiftyone/factory/repos/execution_store.py
```diff
@@ -198,21 +198,42 @@ def has_store_global(self, store_name):
...
+            {
+                "$group": {
+                    "_id": {
+                        "store_name": "$store_name",
+                        "dataset_id": "$dataset_id",
+                    }
+                }
+            },
+            {
+                "$project": {
```
ritch, line 213: The `key=__store__` document is optional. This was returning no results in a few of the integration tests.

Another option would be to ensure a store document existed in set_key, which is an extra mongo operation.

fiftyone/factory/repos/execution_store.py
```diff
@@ -34,23 +35,42 @@ def __init__(self, collection: Collection, dataset_id: ObjectId = None):
         self._collection = collection
         self._dataset_id = dataset_id
 
-    def create_store(self, store_name) -> StoreDocument:
+    def create_store(
```
Br2850, line 38: I saw a PR where @imanjra plans to change this to `def store()` so just a heads up in case a conflict arises

fiftyone/operators/store/service.py
```diff
@@ -47,10 +47,12 @@ def __init__(
             repo = RepositoryFactory.execution_store_repo(
                 dataset_id=dataset_id
             )
-
+        self._dataset_id = dataset_id
         self._repo: ExecutionStoreRepo = repo
 
-    def create_store(self, store_name: str) -> StoreDocument:
+    def create_store(
```
Br2850, line 53: Same `creat_store` comment as before, I don't care which method name is used - I added a try/except for both just in case in DQ panel

fiftyone/factory/repos/execution_store.py
brimoor: Ah okay good catch 👍

tests/unittests/decorators.py
```diff
@@ -43,6 +45,27 @@ async def wrapper(*args, **kwargs):
     return wrapper
 
 
+def drop_stores(func, pattern="*"):
```
brimoor, line 48: Danger! ☠️☠️☠️

Looks like the default `@drop_stores` syntax will drop all of my persistent dataset's stores!

Note that the existing `@drop_datasets` decorator used in tests only deletes non-persistent datasets, which by definition should be safe to delete.

I definitely think we need to tweak this PR so that running unit tests doesn't affect any persistent datasets that the user may have in their DB.

brimoor (changes requested):
Thanks for taking this on 💪 

I think there's an important change needed to `@drop_stores` before this can be merged!

tests/unittests/decorators.py
ritch: What if we assert here that the dataset is non persistent?

ritch: Best idea I have so far is to only ever delete stores with a `__test__` prefix - via `@drop_stores(pattern="__test__*")`.

Then I'll rename all the tests store names to match. However this won't solve the issue where tests would fail...that would require some additional isolation. Maybe something like:

```python
original_collection_name = ExecutionStoreService.collection_name
ExecutionStoreService.collection_name = "test_stores"
func(*args, **kwargs*)
ExecutionStoreService.collection_name = original_collection_name
```

tests/unittests/execution_store_service_tests.py
```diff
@@ -0,0 +1,384 @@
...
+
+@pytest.fixture
+def svc():
+    return ExecutionStoreService()
+
+
+@pytest.fixture
+def svc_with_dataset():
+    dataset = fo.Dataset(name="test_dataset")
+    dataset.save()
```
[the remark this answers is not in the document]
ritch: Removed

tests/unittests/execution_store_service_tests.py
```diff
@@ -0,0 +1,396 @@
...
+from fiftyone.operators.store import ExecutionStoreService
+
+from decorators import drop_collection, drop_datasets
+
+TEST_COLLECTION_NAME = "execution_store_test_collection"
+
+
+@pytest.fixture
+def dataset():
+    return fo.Dataset(name="test_dataset")
```
brimoor, line 21: could just use `fo.Dataset()` here to ensure this test won't fail if someone happens to have a pre-existing `test_dataset`

brimoor (dismissed):
LGTM

tests/unittests/execution_store_service_tests.py
ritch: removed `name=`

brimoor (approved):

PR #5103, opened 2024-11-13
Discord Community Link

benjaminpkane:
Screenshot 2024-11-13 at 12 29 09 PM


brimoor (changes requested):
We'll need to check when Discord is officially in GA . `fiftyone==1.1.0` might be too early for that.

PR #5104, opened 2024-11-13
chore(ci): AS-326 Conditional GitHub Actions

afoley587:

This aims to create conditional workflow steps to say things like:

* only docs are changing, don't run python tests
* only an e2e playwright change is happening, no need for x tests
* a python or js code changed, we should re-run everything to be safe.

The goal is to speed up build times and save money on github actions

* [Docs Only Changes](https://github.com/voxel51/fiftyone/actions/runs/11822906578) - This didn't trigger anything. Which is expected.
* [`setup.py` Changes](https://github.com/voxel51/fiftyone/actions/runs/11822940087) - This triggered everything. Which is expected.

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -10,13 +10,43 @@ on:
       - release/v[0-9]+.[0-9]+.[0-9]+
 
 jobs:
+  modified-files:
+    runs-on: ubuntu-latest
+    outputs:
+      app-changes: ${{ steps.filter.outputs.app }}
+      e2e-changes: ${{ steps.filter.outputs.e2e-pw }}
+      fiftyone-changes: ${{ steps.filter.outputs.fiftyone }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            app:
+              - 'app/**'
+            e2e-pw:
+              - 'e2e-pw/**'
+            fiftyone:
+              - 'fiftyone/**'
+              - 'package/**'
+              - 'requirements/**'
+              - 'tests/**'
+              - 'requirements.txt'
+              - 'setup.py'
+
   build:
+    needs: modified-files
+    if: ${{ needs.modified-files.outputs.app-changes == 'true' || needs.modified-files.outputs.fiftyone-changes == 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -10,13 +10,43 @@ on:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            app:
+              - 'app/**'
+            e2e-pw:
+              - 'e2e-pw/**'
```
sashankaryal, line 28: We'll want to run e2e tests when `app/**` or `fiftyone/**` change as well.

sashankaryal (changes requested):
Love this!!!

.github/workflows/pr.yml
findtopher: that's handled down on like 43
> if: ${{ needs.modified-files.outputs.app-changes == 'true' || needs.modified-files.outputs.e2e-changes == 'true' || needs.modified-files.outputs.fiftyone-changes == 'true' }}

.github/workflows/pr.yml
```diff
@@ -51,6 +83,7 @@ jobs:
     if: always()
     steps:
       - run: sh -c ${{
-          needs.build.result == 'success' &&
-          needs.lint.result == 'success' &&
-          needs.test.result == 'success' }}
+          (needs.build.result == 'success' || needs.build.result == 'skipped') &&
+          (needs.lint.result == 'success' || needs.lint.result == 'skipped') &&
+          (needs.test.result == 'success' || needs.test.result == 'skipped') && 
+          (needs.e2e.result == 'success' || needs.e2e.result == 'skipped') }}
```
findtopher, line 89: nit: who doesn't put a newline at the end of the file?  What are we, barbarians? 😉

afoley587: newline added!

.github/workflows/pr.yml
afoley587: Yes, exactly. It's handled a few lines below!

sashankaryal (approved):
LGTM! 🥇

PR #5111, opened 2024-11-14
always disable orchestrator registration

swheaton:
Always disable orchestrator_registration in OSS

## Summary by CodeRabbit

- **Bug Fixes**
	- Simplified the orchestrator registration logic to always return `False`, ensuring consistent behavior regardless of environment settings.


CamronStaley (approved):
Approving since I believe these are OSS only (could maybe add a comment to say that it's OSS specific to avoid someone picking this if there's a merge conflict in the future)

PR #5128, opened 2024-11-15
No wait summary field index creation

minhtuev:

Added a param (`wait`) to pass to index creation for summary fields

* Local testing

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -1694,6 +1694,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         group_by=None,
         read_only=True,
         create_index=True,
+        wait=True,
     ):
         """Populates a sample-level field that records the unique values or
         numeric ranges that appear in the specified field on each sample in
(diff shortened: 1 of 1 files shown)


fiftyone/core/dataset.py
```diff
@@ -1898,7 +1901,7 @@ def create_summary_field(
 
         if create_index:
             for _field_name in index_fields:
-                self.create_index(_field_name)
+                self.create_index(_field_name, wait=wait)
```
brimoor, line 1904: Have you benchmarked the cost of `create_index()` vs `_populate_summary_field()`? Note that we create the index *before* populating the field here.

I bet running the aggregation to populate the field takes longer, which would mean that `create_summary_field(wait=False)` would not actually return fast.

brimoor (changes requested):

fiftyone/core/dataset.py
minhtuev: That's my fear too; I hope that at least setting `wait=False` will allow the execution of this function to terminate a little bit earlier...

minhtuev: Probably the right approach is allowing the users to schedule summary field operations, but it's currently out of scope for v2.2.0; one thing we can do from UI/UX side is to add a user warning to summary field operations so at least the users are aware of it.

PR #5155, opened 2024-11-19
Prevent video time from getting assigned NaN when frame rate is undefined

tom-vx51:

Prevent video time from getting assigned NaN when frame rate is undefined. This currently causes a crash during playback; with this change, playback will be disabled if the frame rate is undefined, but will not crash.

local app

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling in the time calculation function to prevent non-finite values from being returned, enhancing application stability.


benjaminpkane (approved):
🚀

PR #5156, opened 2024-11-20
improve overlay rendering performance

sashankaryal:

- Accumulate mask decoding promises and run `Promise.allSettled()` right before painting. This means we do not block in a loop for each mask to deserialize
- Separate mask decoding loop from mask painting loop
- Use a more resilient custom fetch function with linear backoff to fetch mask paths. I found `fetchRetry` from `getFetchFunction()` to not be reliable. We can safely not use `getFetchFunction()` in `imputeOverlayFromPath()` since overlay URL is guaranteed to be a well-formed URL here.
- When loading a dataset with lots of samples and overlays, we sometimes get into `ERR:INSUFFICIENT_RESOURCES` (both in image looker or worker). This PR adds resilience against that behavior.

Unit tests for new fetch function. Local performance test with the following script:

The dataset I worked with to get into `ERR:INSUFFICIENT_RESOURCES` was generated with the following script. This is a quickstart dataset with thousands of instance segmentations.

NOTE: Make sure to have both your grid spacing and zoom to the smallest possible values to accommodate the largest possible amount of samples

```
import fiftyone as fo
import fiftyone.zoo as foz
UINT8_MAX = 2**8 - 1
UINT16_MAX = 2**16 - 1

def get_mask(height, width, dtype=np.uint8):
    if dtype == np.uint8:
        return np.random.binomial(1, 0.7, size=(height, width)).astype(np.uint8) * 255
    elif dtype == np.uint16:
        return np.random.binomial(1, 0.7, size=(height, width)).astype(np.uint16) * 65535
    else:
        raise Exception("dtype not supported")

def get_mask_path_from_mask(mask):
    img = Image.fromarray(mask)
    img.save(mask_path)
    return mask_path

dataset = foz.load_zoo_dataset("quickstart", dataset_name="detection_mask_path", persistent=True)
samples = dataset.take(len(dataset))

# multiply dataset 3-fold
for _ in range(3):
    dataset.add_samples(samples.clone())

# add same mask (but different files) to all fields
mask = get_mask(256, 256)

for sample in dataset:
    mask_path = get_mask_path_from_mask(mask)
    gt_detections = sample.ground_truth.detections
    for detection in gt_detections:
        detection["mask"] = None
        detection["mask_path"] = mask_path
    pr_detections = sample.predictions.detections
    for detection in pr_detections:
        detection["mask"] = None
        detection["mask_path"] = mask_path
    sample.save()
```

        notes for FiftyOne users.

-   [ ] Other

## Release Notes

- **New Features**
  - Introduced a linear backoff retry mechanism for HTTP fetch requests, enhancing reliability during image loading and data fetching.

- **Improvements**
  - Enhanced error handling for image loading failures, allowing for multiple retry attempts before indicating an error.
  - Updated overlay image fetching logic to improve concurrency and error handling.
  - Simplified input for retry codes in the fetch function and adjusted retry delay to optimize response times.
  - Improved error handling in the error display logic, allowing for better management of the error state.

- **Tests**
  - Added comprehensive test suite for the new fetch functionality, covering success and error scenarios.

### DIFF
--- a/app/packages/looker/src/elements/common/error.ts
+++ b/app/packages/looker/src/elements/common/error.ts
@@ -103,12 +103,11 @@ export class ErrorElement<State extends BaseState> extends BaseElement<State> {
       }
     }
 
+    if (!error && this.errorElement) {
+      this.errorElement.remove();
+      this.errorElement = null;
+    }
+
     return this.errorElement;
   }
 }
-
-const onClick = (href) => {
-  let openExternal;
-
-  return null;
-};
--- a/app/packages/looker/src/elements/image.ts
+++ b/app/packages/looker/src/elements/image.ts
@@ -6,13 +6,24 @@ import type { ImageState } from "../state";
 import type { Events } from "./base";
 import { BaseElement } from "./base";
 
+const MAX_IMAGE_LOAD_RETRIES = 10;
+
 export class ImageElement extends BaseElement<ImageState, HTMLImageElement> {
   private src = "";
-  private imageSource: HTMLImageElement;
+  protected imageSource: HTMLImageElement;
+
+  private retryCount = 0;
+  private timeoutId: number | null = null;
 
   getEvents(): Events<ImageState> {
     return {
       load: ({ update }) => {
+        if (this.timeoutId !== null) {
+          window.clearTimeout(this.timeoutId);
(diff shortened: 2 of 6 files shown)


benjaminpkane (approved):
LGTM

PR #5171, opened 2024-11-21
fix qp null exception when setting state

CamronStaley:

If the toast triggers the QP operator for creating an index and it defaults to summary, switching to index and then creating the index would throw a null state exception although the panel would close and create the index. These changes stop that from happening.

Before:

After:

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved state management by adding checks to prevent updates when the state is null or undefined, enhancing overall app stability.
  
- **Refactor**
	- Updated code formatting for better clarity and consistency without affecting functionality.
  
- **Enhancements**
	- Refined resource management logic in routing to improve how resources are reused based on the `hard` flag.

### DIFF
--- a/app/packages/app/src/routing/RouterContext.ts
+++ b/app/packages/app/src/routing/RouterContext.ts
@@ -223,7 +223,7 @@ const makeGetEntryResource = <T extends OperationType>() => {
     hard: boolean;
     handleError?: (error: unknown) => void;
   }): Resource<Entry<T>> => {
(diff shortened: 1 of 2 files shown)


minhtuev (approved):
Reviewed with @imanjra , change looks good to unblock QP panel, but we still need to analyze root cause @benjaminpkane

benjaminpkane (approved):

PR #5180, opened 2024-11-22
Fix execute_operator() with delegation

swheaton:
`resolve_type_with_context()` had morphed since introduction and was now recreating the Context even though we already have it. The bug was that when scheduled from the SDK, `operator_uri` is not a part of `request_params`. When scheduled from the UI, it is, plus a whole bunch of other attributes. Introduced sometime pre-v1.0

Repurpose/refactor `resolve_type_with_context()` to be more what the name implies it's gonna do. Then make `resolve_type()` use that also.
If we reuse the `ExecutionContext` we already had (which was properly created), it is cleaner plus has a valid `operator_uri` which fixes this particular bug.

I tried to make some exposing tests but I was getting frustrated and I have limited hours before going out for a week. So ended up scrapping it.
We definitely need more testing in this area of the code.

Here's @brimoor's manual test that works now (no error on resolving output schema).

```python
import fiftyone as fo
import fiftyone.zoo as foz

compute_metadata = foo.get_operator("@voxel51/utils/compute_metadata")

# Schedule a delegated operation to (re)compute metadata
compute_metadata(dataset, overwrite=True, delegate=True)
```

```
$ fiftyone delegated launch
Running operation xyz (@voxel51/utils/compute_metadata)
Computing metadata...
 100% |██████████████████████████| 200/200 [50.6ms elapsed, 0s remaining, 4.0K samples/s] 

Operation xyz complete
```

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for output schema resolution, now logging warnings instead of raising exceptions.
	- Simplified type resolution process with improved error management.

- **Documentation**
	- Updated comments and docstrings for clarity on method functionality.

### DIFF
--- a/fiftyone/operators/delegated.py
+++ b/fiftyone/operators/delegated.py
@@ -514,11 +514,12 @@ class DelegatedOperationService(object):
         result = await do_execute_operator(operator, ctx, exhaust=True)
 
         outputs_schema = None
-        request_params = {**context.request_params, "results": result}
         try:
-            outputs = await resolve_type_with_context(
-                request_params, "outputs"
-            )
+            # Resolve output types now
+            ctx.request_params["target"] = "outputs"
+            ctx.request_params["results"] = result
+
+            outputs = resolve_type_with_context(operator, ctx)
             if outputs is not None:
                 outputs_schema = outputs.to_json()
         except (AttributeError, Exception):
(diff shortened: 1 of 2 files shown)


brimoor (approved):
LGTM

PR #5181, opened 2024-11-22
Don't require pipeline updates when setting output schema

brimoor:
On `develop`, running a delegated operation with an output schema like [this one](https://github.com/voxel51/fiftyone-plugins/blob/d589a19fd7f996517d8d84b20d5d5a23c5c33cc6/plugins/brain/__init__.py#L90-L93) would fail on MongoDB <5 with the following error:

```
pymongo.errors.OperationFailure: Invalid $set :: caused by :: an empty object is not a valid value. Found empty object at path metadata.outputs_schema.type.properties, full error: {'ok': 0.0, 'errmsg': 'Invalid $set :: caused by :: an empty object is not a valid value. Found empty object at path metadata.outputs_schema.type.properties', 'code': 40180, 'codeName': 'Location40180'}
```

The reason was because [line 337](https://github.com/voxel51/fiftyone/blob/6a5b17f74ac5ddff753eead60a9772aee9506afb/fiftyone/factory/repos/delegated_operation.py#L337) was attempting an update aggregation like this:

```
[
    {
        '$set': {
            'run_state': 'completed',
            'completed_at': datetime.datetime(2024, 11, 22, 21, 16, 17, 69660),
            'updated_at': datetime.datetime(2024, 11, 22, 21, 16, 17, 69663),
            'result': {
                'result': None,
                'executor': None,
                'error': None,
                'error_message': None,
                'delegated': False,
                'validation_ctx': None,
            },
            'metadata.outputs_schema': {
                'type': {'name': 'Object', 'properties': {}},
                'default': None,
                'required': False,
                'choices': None,
                'invalid': False,
                'error_message': '',
                'on_change': None,
                'view': {
                    'name': 'View',
                    'label': 'Request complete',
                    'description': None,
                    'caption': None,
                    'space': None,
                    'placeholder': None,
                    'read_only': None,
                    'component': None,
                    'componentsProps': None,
                    'container': None,
                },
            },
        },
    },
]
```

and sadly this part:

```
'type': {'name': 'Object', 'properties': {}},
```

is [not supported on MongoDB < 5](https://www.mongodb.com/community/forums/t/mongoservererror-invalid-set-caused-by-an-empty-object-is-not-a-valid-value/148344).

The solution is to ensure that the `metadata` key always exists. Then we don't need the `needs_pipeline_update` flag! 🎉 

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of metadata in queued operations, ensuring it defaults to an empty dictionary.
	- Streamlined logic for updating run states, simplifying the process without altering core functionality.

- **Bug Fixes**
	- Improved error handling in the queue operation method for better dataset ID resolution.

- **Tests**
	- Updated unit tests to reflect changes in metadata handling, ensuring consistency in expectations.
	- Removed outdated tests related to null metadata handling.

### DIFF
--- a/fiftyone/factory/repos/delegated_operation.py
+++ b/fiftyone/factory/repos/delegated_operation.py
@@ -177,15 +177,15 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
                 raise ValueError("Missing required property '%s'" % prop)
             setattr(op, prop, kwargs.get(prop))
 
-        # also set the delegation target (not required)
         delegation_target = kwargs.get("delegation_target", None)
         if delegation_target:
             setattr(op, "delegation_target", delegation_target)
 
-        # also set the metadata (not required)
         metadata = kwargs.get("metadata", None)
         if metadata:
             setattr(op, "metadata", metadata)
+        else:
+            setattr(op, "metadata", {})
 
         context = None
         if isinstance(op.context, dict):
@@ -262,8 +262,6 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
             else None
         )
 
-        needs_pipeline_update = False
-
         if run_state == ExecutionRunState.COMPLETED:
             update = {
                 "$set": {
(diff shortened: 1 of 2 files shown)


swheaton (approved):
🚢

PR #5187, opened 2024-11-25
Show "missing" counts in confusion matrices

brimoor:
## Change log

When evaluating object detections, confusion matrices need the ability to show an additional row/column to capture “missing” values:
- "missing": false negative ground truth (no matched prediction)
- "missing": false positive predictions (no matched ground truth)

## TODO

- [x] Update `set_view()` callbacks to correctly handle "missing" cases
- [x] The panel currently throws an exception when I click on a "missing" row/col (bottom row or rightmost col)
- [x] The backend now includes the "missing" class as the **LAST** row/col of the confusion matrices (see the example code below). When the user requests `k=10` classes, we need to display 11 total rows/cols: the 10 rows/cols representing the classes they requested PLUS the last row/col so that we're showing the "missing" data. This means the frontend logic for applying the `limit=k` logic will need to select the first `k` rows plus the last row of the appropriate confusion matrix, rather than just the first `k` rows 

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "predictions",
    gt_field="ground_truth",
    eval_key="eval",
    classwise=False,
)

```

```py

freq = Counter(results.ytrue)
if results.missing in freq:
    freq.pop(results.missing)

results.missing
# '(none)'

classes = sorted(freq, key=freq.get, reverse=True)[:5]
# ['person', 'kite', 'bird', 'car', 'carrot']

cmat, labels, _ = results._confusion_matrix(
    classes=classes,
    include_other=False,
    include_missing=True,
    tabulate_ids=False,
)

"""
cmat
array([[665,   2,   1,   0,   0,  43],
       [  0, 123,   1,   0,   0,  15],
       [  3,   2,  84,   0,   0,  22],
       [  0,   0,   0,  45,   0,  12],
       [  0,   0,   0,   0,  34,  12],
       [472,  59,  23, 154, 344,   0]])

labels
['person', 'kite', 'bird', 'car', 'carrot', '(none)']
"""
```

### DIFF
--- a/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
@@ -51,6 +51,7 @@ import { formatValue, getNumericDifference, useTriggerEvent } from "./utils";
 const KEY_COLOR = "#ff6d04";
 const COMPARE_KEY_COLOR = "#03a9f4";
 const DEFAULT_BAR_CONFIG = { sortBy: "default" };
+const NONE_CLASS = "(none)";
 
 export default function Evaluation(props: EvaluationProps) {
   const {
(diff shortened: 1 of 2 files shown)


imanjra (dismissed):
LGTM

brimoor:
@imanjra there's a couple frontend TODO items in the PR description that will need to be addressed before we can merge this. Could you take a look??

imanjra (approved):
LGTM

imanjra (approved):

PR #5192, opened 2024-11-25
Updated doc for `panel.img`

minhtuev:

Doc fix for `panel.img`

N/A

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
	- Clarified the description of the `href` argument in the `img` method to accurately reflect its purpose as the image URL.
	- Updated the default value of the `href` parameter to a more relevant image source URL.

### DIFF
--- a/fiftyone/operators/types.py
+++ b/fiftyone/operators/types.py
@@ -327,11 +327,12 @@ class Object(BaseType):
         Examples::
 
             import fiftyone.operators.types as types
+            ctx.panel.state.my_img = "/path/to/my/image.jpg"
 
             panel = types.Object()
             panel.img(
-                "my_img",
-                href="https://path/to/my/image",
+                "my_img", # uses the value in ctx.panel.state.my_img
+                href="https://path/to/navigate/to",
                 on_click=self.do_something,
                 prompt=False,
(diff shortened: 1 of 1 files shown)


fiftyone/operators/types.py
```diff
@@ -328,12 +328,10 @@ def img(
 
             import fiftyone.operators.types as types
 
-            ctx.panel.state.my_img = "https://i.imgur.com/TGEZj1Rl.jpg"
-
             panel = types.Object()
             panel.img(
                 "my_img",
-                href="https://voxel51.com",
+                href="https://path/to/my/image",
```
ritch, line 334: I believe the previous doc was correct. This is not the path to the image. Instead this property defines a string which is a path to an image.

fiftyone/operators/types.py
```diff
@@ -345,7 +343,7 @@ def img(
 
         Args:
             name: the name of the property from state
-            href (None): the URL to navigate to when the button is clicked
+            href (None): the URL of the image
```
ritch, line 346: How about:

> the url to navigate to when the image is clicked

ritch (approved):

PR #5226, opened 2024-12-05
List hf hub datasets

harpreetsahota204:

Added proper documentation and improved implementation of the `list_hub_datasets()` function in `fiftyone/utils/huggingface.py`. The changes include:
- Added comprehensive docstring explaining function purpose and usage
- Improved code readability with clear comments

The changes can be tested by:
1. Running the function directly:
```python
from fiftyone.utils.huggingface import list_hub_datasets
list_hub_datasets()
```

2. Checking the docstring help:
```python
help(list_hub_datasets)
```

-   [x] Yes. You can now easily list the datasets we have on our HF org

Added documentation and improved implementation of `list_hub_datasets()` function, making it easier for users to discover and load FiftyOne datasets from the Hugging Face Hub.

-   [] Core: Core `fiftyone` Python library changes
-   [x] Other: utils/huggingface.py

## Summary by CodeRabbit

- **New Features**
	- Introduced a new function to list all FiftyOne datasets available on the Hugging Face Hub, enhancing user access to discoverable datasets.

### DIFF
--- a/fiftyone/utils/huggingface.py
+++ b/fiftyone/utils/huggingface.py
@@ -67,22 +67,30 @@ logger = logging.getLogger(__name__)
 def list_hub_datasets():
     """Lists all FiftyOne datasets available on the Hugging Face Hub.
 
-    This function prints instructions for loading datasets and displays a list of
-    all datasets published by Voxel51 on the Hugging Face Hub.
+
+
+    Examples::
+
+        from fiftyone.utils.huggingface import load_from_hub
+        
+        # Return an iterable of all datasets in Voxel51 org
+        org_datasets = list_hub_datasets()
+        
+        # Print names of datasets
+        for dataset in org_datasets:
+            dataset_name = dataset.id
+            print(dataset_name)
+
+        # If a dataset looks interesting, you can load any of the following datasets as follows
+        dataset = load_from_hub("Voxel51/dataset-name")
+
+    Returns:
+        generator: A generator of dataset objects from the Hugging Face Hub API 
+             containing metadata about each dataset published by Voxel51
     """
-    from huggingface_hub import list_datasets
-    
-    print('You can load any of the following datasets as follows:\n')
-    print('from fiftyone.utils.huggingface import load_from_hub')
-    print('dataset = load_from_hub("Voxel51/dataset-name")\n')
-    print('Datasets on the Hugging Face Hub:\n')
-    
-    # List all datasets for Voxel51
-    org_datasets = list_datasets(author="Voxel51")
-    
-    for dataset in org_datasets:
-        dataset_name = dataset.id.split('Voxel51/')[-1]
-        print(dataset_name)
+
+
+    return hfh.list_datasets(author="Voxel51")
 
 
 def push_to_hub(
--- a/fiftyone/utils/huggingface.py
+++ b/fiftyone/utils/huggingface.py
@@ -31,7 +31,6 @@ from fiftyone.core.sample import Sample
 import fiftyone.core.utils as fou
 import fiftyone.types as fot
 
-from huggingface_hub import list_datasets
 
 hfh = fou.lazy_import(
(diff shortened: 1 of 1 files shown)


fiftyone/utils/huggingface.py
```diff
@@ -63,6 +64,26 @@
 
 logger = logging.getLogger(__name__)
 
+def list_hub_datasets():
+    """Lists all FiftyOne datasets available on the Hugging Face Hub.
+
+    This function prints instructions for loading datasets and displays a list of
+    all datasets published by Voxel51 on the Hugging Face Hub.
+    """
+    from huggingface_hub import list_datasets
```
brimoor, line 73: Use `hfh.list_datasets` here so that the user will get the informative error message that `lazy_import()` enables if they don't have the `huggingface_hub` package installed

fiftyone/utils/huggingface.py
```diff
@@ -63,6 +64,26 @@
 
 logger = logging.getLogger(__name__)
 
+def list_hub_datasets():
```
brimoor, line 67: I recommend this method work similarly `fo.list_datasets()` where the method returns the names and the user can print them themselves if that's what they want to do. Or they do something programmatically with the names

fiftyone/utils/huggingface.py
```diff
@@ -63,6 +64,26 @@
...
+    print('You can load any of the following datasets as follows:\n')
+    print('from fiftyone.utils.huggingface import load_from_hub')
+    print('dataset = load_from_hub("Voxel51/dataset-name")\n')
+    print('Datasets on the Hugging Face Hub:\n')
+    
+    # List all datasets for Voxel51
+    org_datasets = list_datasets(author="Voxel51")
+    
+    for dataset in org_datasets:
+        dataset_name = dataset.id.split('Voxel51/')[-1]
```
brimoor, line 84: I would argue that `Voxel51/` prefix should be kept. Then the example code will be much cleaner:

```py
import fiftyone.utils.huggingface as hf

# List all FiftyOne-compatible datasets in the HF Hub
names = hf.list_hub_datasets()
print(names)

# Load a dataset
dataset = hf.load_from_hub(names[0])
```

fiftyone/utils/huggingface.py
```diff
@@ -63,6 +64,26 @@
...
 
+def list_hub_datasets():
+    """Lists all FiftyOne datasets available on the Hugging Face Hub.
+
+    This function prints instructions for loading datasets and displays a list of
+    all datasets published by Voxel51 on the Hugging Face Hub.
+    """
+    from huggingface_hub import list_datasets
+    
+    print('You can load any of the following datasets as follows:\n')
```
brimoor, line 75: Rather than using `print()` here, please put the example code in the function's docstring, formatted [like this](https://github.com/voxel51/fiftyone/blob/1649f2f2e9a98f7499d08921dbc8e9ac050732c2/fiftyone/core/collections.py#L5528-L5564) so that it is rendered in the docs [like this](https://docs.voxel51.com/api/fiftyone.core.collections.html#fiftyone.core.collections.SampleCollection.limit).

brimoor (changes requested):

fiftyone/utils/huggingface.py
```diff
@@ -64,6 +59,33 @@
...
+        print(datasets)
+
+    Args:
+        info (False): whether to return dataset names (False) or
+            ``huggingface_hub.hf_api.DatasetInfo`` objects (True)
+
+    Returns:
+        a list of dataset names or objects
+    """
+    datasets = list(hfh.list_datasets(filter="library:fiftyone"))
```
brimoor, line 82: This used to be `author="Voxel51"` but I switched it to `filter="library:fiftyone"` so that we include all FiftyOne-compatible datasets

brimoor (approved):
LGTM

PR #5231, opened 2024-12-06
fix import timeouts on increasing datasets by precomputing batch size

CamronStaley:

When importing a dataset that changes greatly in size between samples with the content size dynamic batcher we were seeing timeouts from mongo because based on the previous batch of samples we would predict future batch sizes which wasn't scaling up fast enough. Since most situations we have the content sizes before doing any writes to the database we can precompute these values and create batch sizes upfront.

- Set default batcher to "size"
- Downloaded the zip in the jira ticket above
- Ran the following command:
```
import fiftyone as fo

dns = '/Users/camronstaley/Downloads/Dr_Wright_Drives_Beta.zip'
ds=fo.Dataset.from_archive(dns,dataset_type=fo.types.FiftyOneDataset)
```

Output:
```
(TEAMS) (3.10.14) ➜  scripts python import_archive.py
/Users/camronstaley/Workplace/envs/TEAMS/lib/python3.10/site-packages/paramiko/pkey.py:100: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  "cipher": algorithms.TripleDES,
/Users/camronstaley/Workplace/envs/TEAMS/lib/python3.10/site-packages/paramiko/transport.py:259: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  "class": algorithms.TripleDES,
Assuming '/Users/camronstaley/Downloads/Dr_Wright_Drives_Beta' contains the extracted contents of '/Users/camronstaley/Downloads/Dr_Wright_Drives_Beta.zip'
Importing samples...
 100% |███████████████████████████████████████████████████████████████████████████████████████████| 3/3 [154.1ms elapsed, 0s remaining, 19.5 samples/s] 
Importing frames...
 100% |█████████████████████████████████████████████
```

batch sizes created:
```
[3670, 3666, 3663, 3653, 3653, 3670, 3666, 3663, 3653, 3653, 1968, 50, 50, 50, 50, 50, 50, 52, 50, 50, 50, 50, 50, 50, 50, 69, 50, 50, 86, 56, 51, 51, 51, 51, 53, 51, 64, 94, 56, 51, 50, 50, 50, 51, 50, 51, 50, 50, 50, 50, 50, 50, 50, 113, 63, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 65, 51, 51, 51, 51, 96, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 53, 57, 51, 51, 2203, 3031, 3023, 3022, 3022, 1015]
```

batch sizes created before these changes:
```
batch content size: 1048992
length of batch: 3666                                                                                                                                  
batch content size: 1050321
length of batch: 3660                                                                                                                                  
batch content size: 1050664
length of batch: 3653                                                                                                                                  
batch content size: 1048607
length of batch: 3653                                                                                                                                  
batch content size: 45736747
length of batch: 457
```
Above is just a small chunk of the batch sizes but the idea is that it ramped from 0 - 3600ish which maintained the 1 mb target size until it reached the 3rd video which was way more densely populated with labels. This resulted in the batch size being 45736747 and often times it timed out.

-   [] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

-   [ ] Other

## Release Notes

- **New Features**
  - Introduced new methods for managing dataset samples, including adding images and videos, and saving/loading dataset views.
  - Added a new `ContentSizeBatcher` for improved batch processing based on content size.

- **Enhancements**
  - Enhanced sample management with improved validation and dynamic schema expansion.
  - Streamlined error handling for clearer messages and better edge case management.

- **Bug Fixes**
  - Updated tests to ensure coverage for the new batching functionality while maintaining existing test integrity. 

- **Documentation**
  - Updated method documentation to reflect new functionalities and parameters.

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -1552,6 +1552,117 @@ class StaticBatcher(Batcher):
         return self.batch_size
 
 
+class ContentSizeBatcher(Batcher):
+    """Class for iterating over the elements of an iterable with a dynamic
+    batch size to achieve a desired content size.
+
+    The batch sizes emitted when iterating over this object are dynamically
+    scaled such that the total content size of the batch is as close as
+    possible to a specified target size.
+
+    This batcher does not require backpressure feedback because it calculates
+    the total size of the iterable object before batching.
+
+    This class is often used in conjunction with a :class:`ProgressBar` to keep
+    the user appraised on the status of a long-running task.
+
+    Example usage::
+
+        import fiftyone.core.utils as fou
+
+        elements = range(int(1e7))
+
+        batcher = fou.ContentSizeBatcher(
+            elements,
+            target_size=2**20,
+            progress=True
+        )
+
+        with batcher:
+            for batch in batcher:
+                print("batch size: %d" % len(batch))
+
+    Args:
+        iterable: an iterable to batch over. If ``None``, the result of
+            ``next()`` will be a batch size instead of a batch, and is an
+            infinite iterator.
+        target_size (1048576): the target batch bson content size, in bytes
+        min_batch_size (1): the minimum allowed batch size
+        max_batch_size (None): an optional maximum allowed batch size
+        return_views (False): whether to return each batch as a
+            :class:`fiftyone.core.view.DatasetView`. Only applicable when the
+            iterable is a :class:`fiftyone.core.collections.SampleCollection`
+        progress (False): whether to render a progress bar tracking the
+            consumption of the batches (True/False), use the default value
+            ``fiftyone.config.show_progress_bars`` (None), or a progress
+            callback function to invoke instead
+        total (None): the length of ``iterable``. Only applicable when
+            ``progress=True``. If not provided, it is computed via
+            ``len(iterable)``, if possible
+    """
+
+    def __init__(
+        self,
+        iterable,
+        target_size=2**20,
+        min_batch_size=1,
+        max_batch_size=None,
+        return_views=False,
+        progress=False,
+        total=None,
+    ):
+        iterable, iterable_copy = itertools.tee(iterable)
+        super().__init__(
+            iterable, return_views=return_views, progress=progress, total=total
+        )
+        self.batch_sizes = self._compute_batch_sizes(
+            target_size, min_batch_size, max_batch_size, iterable_copy
+        )
+        self.curr_batch = 0
+
+    def _compute_batch_sizes(
+        self, target_size, min_batch_size, max_batch_size, iterable
+    ):
+        batch_sizes = []
+        curr_batch_size = 0
+        curr_batch_content_size = 0
+
+        for obj in iterable:
+            try:
+                curr_batch_content_size += len(
+                    json_util.dumps(self._make_dict(obj))
+                )
+            except Exception:
+                curr_batch_content_size += len(str(obj))
+            curr_batch_size += 1
+            if curr_batch_size >= min_batch_size and (
+                curr_batch_content_size >= target_size
+                or curr_batch_size == max_batch_size
+            ):
+                batch_sizes.append(curr_batch_size)
+                curr_batch_size = 0
+                curr_batch_content_size = 0
+
+        if curr_batch_size:
+            batch_sizes.append(curr_batch_size)
+
+        return batch_sizes
+
+    def _make_dict(self, obj):
+        return (
+            obj.to_mongo_dict(include_id=True)
+            if hasattr(obj, "to_mongo_dict")
+            else obj
+        )
+
+    def _compute_batch_size(self):
+        size = 1
+        if self.curr_batch < len(self.batch_sizes):
+            size = self.batch_sizes[self.curr_batch]
+        self.curr_batch += 1
+        return size
+
+
 def get_default_batcher(iterable, progress=False, total=None):
     """Returns a :class:`Batcher` over ``iterable`` using defaults from your
     FiftyOne config.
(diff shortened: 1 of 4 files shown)


fiftyone/core/utils.py
```diff
@@ -1552,6 +1552,117 @@ def _compute_batch_size(self):
...
+        self, target_size, min_batch_size, max_batch_size, iterable
+    ):
+        batch_sizes = []
+        curr_batch_size = 0
+        curr_batch_content_size = 0
+
+        for obj in iterable:
+            try:
+                curr_batch_content_size += len(
+                    json_util.dumps(self._make_dict(obj))
```
kaixi-wang, line 1633: Couldn't this extra serialization potentially be very expensive?

fiftyone/core/utils.py
```diff
@@ -1552,6 +1552,117 @@ def _compute_batch_size(self):
...
+            else obj
+        )
+
+    def _compute_batch_size(self):
+        size = 1
+        if self.curr_batch < len(self.batch_sizes):
+            size = self.batch_sizes[self.curr_batch]
+        self.curr_batch += 1
+        return size
+
```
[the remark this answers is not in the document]
swheaton: Provide a link to documentation that advises against using the original iterator after calling `tee`.

fiftyone/core/utils.py
CamronStaley: We currently do this already in the [_upsert_samples_batch](https://github.com/voxel51/fiftyone/blob/680656b85368482c24a69a545a25bb7b5387e0f3/fiftyone/core/dataset.py#L3433) so we are basically doubling how much we are doing it in that one method. In most other cases (such as [bulk_write](https://github.com/voxel51/fiftyone/blob/680656b85368482c24a69a545a25bb7b5387e0f3/fiftyone/core/odm/database.py#L820) and [insert_documents](https://github.com/voxel51/fiftyone/blob/680656b85368482c24a69a545a25bb7b5387e0f3/fiftyone/core/odm/database.py#L783)) its the same total computation since we are just moving it from post computing for each batch to precomputing for the entire collection.

CamronStaley: though I think it might be possible to refactor _upsert_sample_batch so that we just pass the dict as an iterable to begin with instead of calculating it twice

fiftyone/core/utils.py
```diff
@@ -1552,6 +1552,117 @@ def _compute_batch_size(self):
...
+        iterable, iterable_copy = itertools.tee(iterable)
+        super().__init__(
+            iterable, return_views=return_views, progress=progress, total=total
+        )
+        self.batch_sizes = self._compute_batch_sizes(
+            target_size, min_batch_size, max_batch_size, iterable_copy
+        )
+        self.curr_batch = 0
+
+    def _compute_batch_sizes(
```
brimoor, line 1623: This is not ideal because it requires iterating over the entire iterator before any writes start happening.

A large scale `add_samples()` call could take 1 hour or more to complete. With this implementation, it will take an hour just to compute safe batch sizes, during which time there will be no progress bar or indication to the user of what's happening, and *then* will it start adding samples to the dataset for the next hour.

We used dynamic batching with backpressure to try to find a compromise where writes will start immediately while also tuning the batch size to maximize throughput.

Perhaps there's a way to keep the greedy batching but add a safeguard to split unexpectedly large batches into multiple writes?

brimoor (changes requested):

fiftyone/core/utils.py
CamronStaley: I think an alternative is we can override the _next_ implementation for just this batch size and build it as we go so instead of doing it upfront we can just iterate and build a list up to the max size and then return that.

We probably want to refactor a bit as well because currently we have two content size bathers: one for backpressure where we don't know the size of the items before batching (think it's just a bunch of ops and not actually looking at the items) and another for if we have the items and can calculate their sizes.

brimoor: >  currently we have two content size bathers

What are you referring to? We currently have three batchers:
https://github.com/voxel51/fiftyone-teams/blob/318d3abb38b478f557180bb6aed4f25681cc7c64/fiftyone/core/utils.py#L1579-L1606

- `LatencyDynamicBatcher`: targets a specific latency between calls to generate the next batch
- `ContentSizeDynamicBatcher`: targets a specific content size in each batch
- `StaticBatcher`: simple fixed batch size

Latency and content size are correlated heuristics, and both would encounter problems if there are wildly different `Sample` sizes in a collection

brimoor: I guess you're referring to latency and content. 

The latency batcher is the default for OSS because it is effectively parameterized by the frame rate on the progress bars you want to see (and the assumption that the database is a big boi and will take as much content in 0.2 seconds as you can throw at it, for example)

The content batcher is the default for Teams when there's an API connection involved because the assumption is that the API gateway has a hard limit on request size, so we want to be certain to not violate that.

https://github.com/voxel51/fiftyone-teams/blob/318d3abb38b478f557180bb6aed4f25681cc7c64/fiftyone/core/config.py#L171

CamronStaley: The batcher in this PR "ContentSizeBatcher" is separate from ContentSizeDynamicBatcher because it doesn't use backpressure and has a different use case where we can view content size upfront which, keep me honest @swheaton, isn't always the case. Thats what I meant by specifically two "content size batchers" I know there's other batchers as well.

I am proposing a couple things:
1. rename both of the content size batchers so that it's clear one is used for computing before writing the batch and the other is for estimating based on the previous batch's results
2. make it so this new batcher we have added computes batch size everytime _next_ is called so that we don't have to do it all upfront (this would require overriding the inherited next method in its current state so maybe a refactor would be good)

PR #5236, opened 2024-12-06
do not use operator executor for disabled buttons

ritch:
## Summary by CodeRabbit

- **New Features**
	- Enhanced the `OperatorExecutionButton` to directly render a disabled state when the button is inactive, improving user experience by simplifying the button's behavior.

- **Bug Fixes**
	- Resolved issues with rendering logic when the button is disabled, ensuring consistent behavior across the application.

### DIFF
--- a/app/packages/operators/src/components/OperatorExecutionButton/index.tsx
+++ b/app/packages/operators/src/components/OperatorExecutionButton/index.tsx
@@ -93,6 +93,14 @@ export const OperatorExecutionButton = ({
 
   if (isLoading) return null;
 
+  if (disabled) {
+    return (
+      <Button disabled {...props} variant={props.variant}>
+        {children}
+      </Button>
+    );
+  }
+
   if (showWarning) {
     return (
       <TooltipProvider title={warningMessage}>


app/packages/operators/src/components/OperatorExecutionButton/index.tsx
```diff
@@ -37,6 +37,14 @@ export const OperatorExecutionButton = ({
   disabled?: boolean;
   children: React.ReactNode;
 }) => {
+  if (disabled) {
+    return (
+      <Button disabled {...props}>
```
manivoxel51, line 42: would this change show the tooltip when disabled as well?

could cause an issue in DQ panel where we use `disabled` and `title` to show permission messaging.
```
operator_button = types.OperatorExecutionButtonView(
                    icon="expand_more",
                    operator=operator_mapping[issue_type],
                    on_success=self.computation_handler,
                    on_error=self.cancel_compute,
                    params={
                        "issue_type": issue_type,
                    },
                    on_option_selected=self._on_compute_option_selected,
                    label=button_string,
                    disabled=no_access or is_computing,
                    title=(NOT_PERMITTED_TEXT if no_access else ""),
                    componentsProps={
                        "container": {
                            "sx": {
                                "display": "flex",
                                "justifyContent": "center",
                            }
                        }
                    },
                )
                card_content.obj(
                    "execution_button",
                    view=operator_button,
                )
```

we can wrap this Button with TitleProvider and pass the props to keep it backward compatible

manivoxel51 (commented):
🍨

PR #5242, opened 2024-12-09
Include more brain modules in docs builds

brimoor:
There are currently some broken links on [this page](https://docs.voxel51.com/brain.html#brain-leaky-splits) because they link to modules like `fiftyone.brain.internal.XXX` that were previously omitted from the docs build.

Seeing that the Brain is now open source, we can happily include more modules in the docs builds and fix these links.

### DIFF
--- a/docs/generate_docs.bash
+++ b/docs/generate_docs.bash
@@ -102,7 +102,7 @@ echo "Generating API docs"
 sphinx-apidoc --force --no-toc --separate --follow-links \
     --templatedir=docs/templates/apidoc \
     -o docs/source/api fiftyone \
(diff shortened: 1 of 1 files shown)


findtopher (approved):
I assume we're cloning the brain somewhere in here for those links to work

```
❯ ls -ald brain
ls: brain: No such file or directory

╭─ ~/workspace/fiftyone/develop/fiftyone develop
```

brimoor:
[yep](https://github.com/voxel51/fiftyone/blob/882055568b0fd6d83c4bcd01c5cc28eaceb16230/docs/generate_docs.bash#L53)

PR #5247, opened 2024-12-09
sample update / overlay recoloring performance optimization

sashankaryal:

This PR makes sample update / overlay recoloring faster, by:
1) cleaning up existing label bitmaps to prevent memory leaks during sample update, and 
2) transferring ownership of targets array buffer to the worker (realloc) during sample update to avoid expensive copying operation

Locally. I used the following code snippet:

```python
import fiftyone as fo
import fiftyone.zoo as foz
import numpy as np
from PIL import Image
import tempfile

UINT8_MAX = 2**8 - 1
UINT16_MAX = 2**16 - 1

def get_mask(height, width, dtype=np.uint8):
    if dtype == np.uint8:
        return np.random.binomial(1, 0.7, size=(height, width)).astype(np.uint8) * 255
    elif dtype == np.uint16:
        return np.random.binomial(1, 0.7, size=(height, width)).astype(np.uint16) * 65535
    else:
        raise Exception("dtype not supported")

def get_mask_path_from_mask(mask):
    img = Image.fromarray(mask)
    img.save(mask_path)
    return mask_path

try:
    fo.delete_dataset("detection_mask_path")
except:
    pass

dataset = foz.load_zoo_dataset("quickstart", dataset_name="detection_mask_path", persistent=True)

samples = dataset.take(len(dataset))
# multiply dataset n fold
for _ in range(3):
    dataset.add_samples(samples.clone())

mask = get_mask(256, 256)
for sample in dataset:
    mask_path = get_mask_path_from_mask(mask)
    gt_detections = sample.ground_truth.detections
    for detection in gt_detections:
        detection["mask"] = None
        detection["mask_path"] = mask_path
    pr_detections = sample.predictions.detections
    for detection in pr_detections:
        detection["mask"] = None
        detection["mask_path"] = mask_path
    sample.save()
```

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced new properties for overlays, enhancing flexibility and usability.
	- Added a method for cleaning up overlays to prevent memory leaks.

- **Improvements**
	- Enhanced bitmap resource management in overlay handling.
	- Simplified cleanup methods across multiple overlay classes for better efficiency.

- **Bug Fixes**
	- Improved error handling during bitmap operations to ensure smoother performance.

### DIFF
--- a/app/packages/looker/src/lookers/abstract.ts
+++ b/app/packages/looker/src/lookers/abstract.ts
@@ -23,7 +23,7 @@ import {
 import { Events } from "../elements/base";
 import { COMMON_SHORTCUTS, LookerElement } from "../elements/common";
 import { ClassificationsOverlay, loadOverlays } from "../overlays";
-import { CONTAINS, Overlay } from "../overlays/base";
+import { CONTAINS, LabelMask, Overlay } from "../overlays/base";
 import processOverlays from "../processOverlays";
 import {
   BaseState,
@@ -515,7 +515,29 @@ export abstract class AbstractLooker<
   abstract updateOptions(options: Partial<State["options"]>): void;
 
   updateSample(sample: Sample) {
-    this.loadSample(sample);
+    // collect any mask targets array buffer that overalys might have
+    // we'll transfer that to the worker instead of copying it
+    const arrayBuffers: ArrayBuffer[] = [];
+
+    for (const overlay of this.pluckedOverlays ?? []) {
+      // we paint overlays again, so cleanup the old ones
+      // this helps prevent memory leaks from, for instance, dangling ImageBitmaps
+      overlay.cleanup();
+
+      let overlayData: LabelMask = null;
+
+      if ("mask" in overlay.label) {
+        overlayData = overlay.label.mask as LabelMask;
+      } else if ("map" in overlay.label) {
+        overlayData = overlay.label.map as LabelMask;
+      }
+
+      if (overlayData?.data?.buffer) {
+        arrayBuffers.push(overlayData.data.buffer);
+      }
+    }
+
+    this.loadSample(sample, arrayBuffers);
   }
 
   getSample(): Promise<Sample> {
@@ -698,7 +720,7 @@ export abstract class AbstractLooker<
     );
   }
 
-  private loadSample(sample: Sample) {
+  private loadSample(sample: Sample, transfer: Transferable[] = []) {
     const messageUUID = uuid();
 
     const labelsWorker = getLabelsWorker();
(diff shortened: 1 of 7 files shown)


app/packages/looker/src/lookers/abstract.ts
```diff
@@ -515,7 +515,48 @@ export abstract class AbstractLooker<
   abstract updateOptions(options: Partial<State["options"]>): void;
 
   updateSample(sample: Sample) {
-    this.loadSample(sample);
+    // collect any mask targets array buffer that overlays might have
+    // we'll transfer that to the worker instead of copying it
+    const arrayBuffers: ArrayBuffer[] = [];
+
+    for (const overlay of this.pluckedOverlays ?? []) {
```
benjaminpkane, line 522: I think `sampleOverlays` should be used here. `pluckedOverlays` contains the current frame overlays, and that is handled separately in VideoLooker.updateSample

app/packages/looker/src/worker/index.ts
```diff
@@ -327,6 +327,10 @@ const processSample = async ({
   labelTagColors,
   schema,
 }: ProcessSample) => {
+  if (!sample) {
+    return;
+  }
```
benjaminpkane, line 332: nit: comment for this return

app/packages/looker/src/lookers/abstract.ts
sashankaryal: Yeah... although we'd have to _copy_ the bitmaps to the worker and I'd much rather avoid that. In my next PR, where I'll be working on async overlay loading, we'll be adding a "loading" state, maybe that'll make it better?

app/packages/looker/src/worker/index.ts
sashankaryal: removed it

sashankaryal:
@benjaminpkane  In the latest commits, I made it so that we retain the bitmaps and clean them in worker callback. However, mask targets will be transferred, and so super super rare edge case where user wants to know mask target value when sample is processing (hover on segmentation overlay), they might not see anything. I think it's an acceptable compromise. 

I might reconsider cleaning overalys in `updateSample()` and not in the worker callback after we introduce more async UI stuff, but later.

app/packages/looker/src/overlays/base.ts
```diff
@@ -67,6 +67,7 @@ export interface Overlay<State extends Partial<BaseState>> {
   draw(ctx: CanvasRenderingContext2D, state: State): void;
   isShown(state: Readonly<State>): boolean;
   field?: string;
+  label?: BaseLabel;
```
benjaminpkane, line 70: This is not true for `Classifications`. Is the `CoordinateOverlay` type is sufficient for the purpose of this PR?
```suggestion
```

app/packages/looker/src/lookers/abstract.ts
```diff
@@ -730,7 +776,20 @@ export abstract class AbstractLooker<
...
       uuid: messageUUID,
-    } as ProcessSample);
+    } as ProcessSample;
+
+    try {
+      labelsWorker.postMessage(workerArgs, transfer);
+    } catch (error) {
+      // rarely we'll get a DataCloneError
+      // if one of the buffers is detached and we didn't catch it
+      // try again without transferring the buffers (copying them)
```
benjaminpkane, line 786: Is guaranteeing  no `DataCloneError` errors possible? How does a bitmap on the main thread get detached before being transfered?

benjaminpkane: Oh, I see this is mentioned in `AbstractLooker.loadSample`

benjaminpkane (approved):

PR #5261, opened 2024-12-11
Refactor builtin plugins

ritch:
# Builtin Plugins Migration

## Background

There have been several attempts to include plugins along with FiftyOne. To ensure consistency in implementation among both built-in and external plugins, we need to normalize how plugins are included in FiftyOne and FiftyOne Teams.

## Current State

Currently, there is no concept of built-in plugins, only a list of `BUILT_IN_OPERATORS` and `BUILT_IN_PANELS`. These contradict the list of `PluginContexts` that are typically loaded for external plugins. This approach also does not support the concept of a built-in JS panel or plugin.

### 1. `plugin_context.register(MyClass, builtin=True)`
- This approach allows us to annotate an artifact during construction and registration.
- These parameters can be used in the future to easily reflect class metadata.

### 2. New `fiftyone/fiftyone/builtins` Directory
- This directory would be loaded as if it were another `FIFTYONE_PLUGINS_DIR`.
- **Note**: This will impact the behavior of the SDK and `list_plugins()`, as built-in plugins will be included. We can discuss how we want this to work.
- It would mimic the structure of `fiftyone-plugins/plugins`, enabling a typical plugin structure and removing the need for `BUILT_IN_OPERATORS` and `BUILT_IN_PANELS`.

### 3. Migrate Existing Built-In Panels and Operators
- Move existing built-in panels and operators from various areas of the codebase to the new directory.

#### Components to Migrate:
- Core Operators
- Query Performance Panel
- Data Quality Panel
- Model Evaluation Panel
- Model Evaluation Operators
- Data Lens Python Operators + Utils
- Data Lens JS Panel (+ Registration)

## Summary by CodeRabbit

- **New Features**
	- Introduced new parameters to enhance operator and plugin listing functionalities, allowing users to filter built-in operators and plugins.
	- Added new classes and functions for improved operator registration and execution.
	- New YAML configuration files for both operators and panels, detailing core functionalities and available features.
	- Added built-in panels for model evaluation.

- **Documentation**
	- Updated documentation for developing plugins, including enhanced examples and clearer instructions on the plugin structure and requirements.


fiftyone/plugins/core.py
```diff
@@ -490,7 +492,10 @@ def _parse_plugin_metadata(metadata_path):
 
 def _list_plugins(enabled=None):
     plugins = []
-    for metadata_path in _iter_plugin_metadata_files():
+    external_plugin_paths = _iter_plugin_metadata_files()
+    builtin_plugin_paths = _iter_plugin_metadata_files(BUILTIN_PLUGINS_DIR)
```
ritch, line 496: This now includes built ins in all listings. Lets discuss how we actually want to handle this. More thought needed here.

fiftyone/plugins/utils.py
```diff
@@ -245,3 +245,8 @@ def _do_get_plugin_info(task):
             spec = gh_repo
 
         logger.debug("Failed to retrieve plugin info for %s: %s", spec, e)
+
+
+def get_builtin_plugins_dir():
+    """Returns the path to the built-in plugins directory."""
```
ritch, line 251: Why built-in vs builtin you may ask... just following the norms from python itself:
image

ritch (commented):
Yep

fiftyone/plugins/core.py
ritch: I'm leaning towards something like `include_builtins=False`.

plugins/operators/__init__.py
```diff
@@ -16,8 +17,10 @@
 import fiftyone.operators.types as types
 from fiftyone.core.odm.workspace import default_workspace_factory
 
-# pylint: disable=no-name-in-module
-from fiftyone.operators.builtins.panels.model_evaluation import EvaluationPanel
+
+from .foo import bar
+
+print(bar())
```
imanjra, line 23: ✂️

plugins/operators/foo.py
```diff
@@ -0,0 +1,2 @@
+def bar():
```
imanjra, line 1: ✂️

imanjra (commented):
Going to extend this in a few to bring back the model eval panel

imanjra (approved):
Tested locally. Works great. LGTM

fiftyone/operators/registry.py
```diff
@@ -90,9 +96,11 @@ def list_operators(self, include_builtin=True, type=None):
         for pctx in self.plugin_contexts:
             operators.extend(pctx.instances)
 
-        if include_builtin:
-            operators.extend(BUILTIN_OPERATORS)
-            operators.extend(BUILTIN_PANELS)
+        if not include_builtin:
+            operators = [op for op in operators if op._builtin is False]
```
kaixi-wang, line 100: Instead of making it a property, can we just have a hardcoded list so that people can’t just mark their own as builtin?

plugins/operators/fiftyone.yml
```diff
@@ -0,0 +1,39 @@
...
+  - load_saved_view
+  - save_view
+  - edit_saved_view_info
+  - delete_saved_view
+  - list_workspaces
+  - load_workspace
+  - save_workspace
+  - edit_workspace_info
+  - delete_workspace
+  - sync_last_modified_at
```
kaixi-wang, line 38: We already have the list pretty much here so it wouldn’t be much of a change to reference the list instead of the operator itself but would prevent people from saving their own as builtin

kaixi-wang (changes requested):

fiftyone/operators/registry.py
ritch: I think its impossible to prevent hijacking in this way. We would need some other mechanisms to fully prevent that. Either way lets first get the new structure in place then we can improve on the builtin mechanism.

plugins/operators/fiftyone.yml
ritch: This is only part of this list this needs to support...in teams we have the other built-ins.

fiftyone/plugins/context.py
```diff
@@ -88,18 +88,19 @@ def can_register(self, instance):
...
             Any errors are logged rather than being raised.
 
         Args:
-            cls: an :class:`fiftyone.operators.operator.Operator` class
+            cls: an :class:`fiftyone.operators.operator.Operator` or :class:`fiftyone.operators.panel.Panel` class
+            builtin (False): For internal use. Whether the cls is a builtin operator or panel
         """
         try:
-            instance = cls()
+            instance = cls(_builtin=builtin)
```
kaixi-wang, line 103: this is unnecessary to expose since you have BUILTIN_PLUGINS_DIR defined and iterate through that.

fiftyone/plugins/core.py
```diff
@@ -488,9 +492,18 @@ def _parse_plugin_metadata(metadata_path):
...
 
 
-def _list_plugins(enabled=None):
+def _list_plugins(enabled=None, include_builtin=False):
     plugins = []
-    for metadata_path in _iter_plugin_metadata_files():
+    external_plugin_paths = list(_iter_plugin_metadata_files())
+    builtin_plugin_paths = list(
+        _iter_plugin_metadata_files(BUILTIN_PLUGINS_DIR)
+    )
```
kaixi-wang, line 500: if necessary, here you can assign the _builtin prop based off being listed from the builtin plugins dir

kaixi-wang (changes requested):

kaixi-wang (approved):
just note that this makes it easy to for non-builtins to be marked as builtin, which could lead to problems

kaixi-wang (approved):
lgtm thanks!

ritch:
> just note that this makes it easy to for non-builtins to be marked as builtin, which could lead to problems
Resolve this by only resolving builtin based on path. 

Now all plugins (including JS) are aware of being builtin, and its all based on one way of defining a plugin: the `PluginDefinition` / `fiftyone.yml`. Builtin is now determined based on where this file is. If its in the expected builtin plugin directory then it is considered a builtin plugin.

kaixi-wang (approved):

PR #5268, opened 2024-12-13
Optimize view callbacks for model evaluation panel

brimoor:
## Change log

Improves the views loaded by the model evaluation panel in the following callbacks:
- clicking the TP/FP/FN icons in the Summary card
- clicking classes in the Class Performance card
- clicking cells in the Confusion Matrix card

Specific improvements:
- Include both ground truth and predicted labels in the filters where appropriate
- use `match()` rather than `filter_labels()` for classification evaluations
- when comparing two models, filter both sets of predictions in the Class Performance callbacks

## TODO

- [x] Events are not firing when cells are clicked in the "compare to" confusion matrix?
    - [x] "compare to" callbacks *are working* for the TP/FP/FN buttons (in which case `options["key"] == compare_eval_key`) and I wired up the `type == "matrix"` code path to use the same key. So I believe that once the events start firing, `load_view()` should work as expected
- [x] I believe the frontend will need to be updated so that the selection state of the Class Performance chart updates as expected when you click on a bar

## Summary by CodeRabbit

- **New Features**
	- Enhanced evaluation logic for model performance, simplifying binary classification checks.
	- Improved handling of different view types for classification and detection scenarios.
	- Expanded filtering capabilities for evaluation stages in the confusion matrix.

- **Refactor**
	- Consolidated binary classification logic into the main evaluation method.
	- Updated view loading process to enhance flexibility and clarity.
	- Improved interaction with the confusion matrix for better user experience.

### DIFF
--- a/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
@@ -1217,6 +1217,14 @@ export default function Evaluation(props: EvaluationProps) {
                             ].join(" <br>") + "<extra></extra>",
                         },
                       ]}
+                      onClick={({ points }) => {
+                        const firstPoint = points[0];
+                        loadView("matrix", {
+                          x: firstPoint.x,
+                          y: firstPoint.y,
+                          key: compareKey,
+                        });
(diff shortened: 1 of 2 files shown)


imanjra (dismissed):
LGTM

imanjra (approved):
Tested locally. LGTM.

PR #5273, opened 2024-12-13
Similarity race condition patch

ehofesmann:

To reproduce (as of v2.1.3):

* Load the quickstart dataset and compute similarity with CLIP
* Create a view stage with Limit 200
* Sort by text similarity against dogs

This error gets raised by the App:

To catch this error, launch the path with:
```
python /path/to/your/install/of/fiftyone/server/main.py
```

```python
File "/home/eric/work/fiftyone/fiftyone-teams/fiftyone/utils/torch.py", line 580, in __exit__
    self._no_grad.__exit__(*args)
AttributeError: 'NoneType' object has no attribute '__exit__'
 ```

TBD

        notes for FiftyOne users.

Fixed a race condition when sorting by text similarity on a view in the App.

-   [X] App: FiftyOne application changes
-   [ ] Other

### DIFF
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -572,6 +572,7 @@ class TorchImageModel(
         if self.config.cudnn_benchmark is not None:
             torch.backends.cudnn.benchmark = self._benchmark_orig
             self._benchmark_orig = None
+
         if self._no_grad is not None:
             self._no_grad.__exit__(*args)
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM. It is not possible for the reported error to still happen with this change, so it has to solve it!

minhtuev:
Thanks Brian!

PR #5274, opened 2024-12-13
Add COCO-styled mAR to evaluate_detections 

manushreegangwar:

This PR adds COCO-styled mean average recall to `evaluate_detections()`.

```py
import fiftyone as fo
import fiftyone.zoo as foz

    "predictions",
    gt_field="ground_truth",
    eval_key="eval",
    compute_mAP=True,
)

print(results.mAR())

# Open Model Evaluation panel and see mAR in the table :)
```

## Summary by CodeRabbit

- **New Features**
    - Introduced Mean Average Recall (mAR) metric in evaluation summaries.
    - Added functionality to visualize and interact with evaluation results in the app.
    - Enhanced COCO evaluation capabilities to compute mAR alongside mAP.

- **Documentation**
    - Expanded documentation on evaluation methods, including mAR and its integration with COCO-style evaluations.
    - Updated examples and explanations for clarity and usability.

### DIFF
--- a/fiftyone/utils/eval/coco.py
+++ b/fiftyone/utils/eval/coco.py
@@ -230,9 +230,9 @@ class COCOEvaluation(DetectionEvaluation):
             matches,
             precision,
             recall,
-            recall_sweep,
             iou_threshs,
             classes,
+            recall_sweep=recall_sweep,
             thresholds=thresholds,
             missing=missing,
             backend=self,
@@ -253,9 +253,9 @@ class COCODetectionResults(DetectionResults):
         precision: an array of precision values of shape
             ``num_iou_threshs x num_classes x num_recall``
         recall: an array of recall values
-        recall_sweep: an array of recall values of shape ``num_iou x num_classes``
         iou_threshs: an array of IoU thresholds
         classes: the list of possible classes
+        recall_sweep (None): an array of recall values of shape ``num_iou x num_classes``
         thresholds (None): an optional array of decision thresholds of shape
             ``num_iou_threshs x num_classes x num_recall``
         missing (None): a missing label string. Any unmatched objects are
@@ -271,9 +271,9 @@ class COCODetectionResults(DetectionResults):
         matches,
         precision,
         recall,
-        recall_sweep,
         iou_threshs,
         classes,
+        recall_sweep=None,
         thresholds=None,
         missing=None,
         backend=None,
@@ -296,7 +296,9 @@ class COCODetectionResults(DetectionResults):
         )
 
         self._classwise_AP = np.mean(precision, axis=(0, 2))
-        self._classwise_AR = np.mean(recall_sweep, axis=0)
+        self._classwise_AR = (
+            np.mean(recall_sweep, axis=0) if recall_sweep is not None else None
+        )
 
     def plot_pr_curves(
         self, classes=None, iou_thresh=None, backend="plotly", **kwargs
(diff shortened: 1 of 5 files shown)


fiftyone/utils/eval/coco.py
```diff
@@ -292,6 +296,7 @@ def __init__(
         )
 
         self._classwise_AP = np.mean(precision, axis=(0, 2))
+        self._classwise_AR = np.mean(recall_sweep, axis=0)
```
manushreegangwar, line 299: @brimoor I am using `config.compute_mAP` for mAR since they are closely related. Perhaps the flag should be renamed.

fiftyone/utils/eval/coco.py
```diff
@@ -229,6 +230,7 @@ def generate_results(
             matches,
             precision,
             recall,
+            recall_sweep,
```
manushreegangwar, line 233: ```COCODetectionResults.__init__() missing 1 required positional argument: 'recall_sweep' (operation: @voxel51/operators/model_evaluation_panel_builtin#load_evaluation)```

TODO update evaluation panel code to support this additional param

fiftyone/utils/eval/coco.py
```diff
@@ -251,6 +253,7 @@ class COCODetectionResults(DetectionResults):
         precision: an array of precision values of shape
             ``num_iou_threshs x num_classes x num_recall``
         recall: an array of recall values
+        recall_sweep: an array of recall values of shape ``num_iou x num_classes``
```
brimoor, line 256: You can't add new positional arguments to `EvaluationResults` subclasses because existing evaluations won't have this value and thus `dataset.load_evaluation_results("<existing_eval_key>")` would start failing.

fiftyone/utils/eval/coco.py
brimoor: The naming is less than ideal, but for backwards compatibility we should probably not rename it.

fiftyone/utils/eval/coco.py
manushreegangwar: I made this optional and the panel doesn't throw any errors now.

fiftyone/utils/eval/coco.py
```diff
@@ -253,6 +255,7 @@ class COCODetectionResults(DetectionResults):
         recall: an array of recall values
         iou_threshs: an array of IoU thresholds
         classes: the list of possible classes
+        recall_sweep (None): an array of recall values of shape ``num_iou x num_classes``
```
brimoor, line 258: nit: our [style guide](https://github.com/voxel51/fiftyone/blob/develop/STYLE_GUIDE.md) says line length should be <80.

Unfortunately the code linter in pre-commit hooks doesn't automatically format comments, so we have to manually apply this convention for docstrings/comments.

Adding a vertical ruler to your IDE at 80 characters helps with this.

brimoor (changes requested):
@manushreegangwar can you please also document mAR in the relevant places so that users know we support COCO mAR now?
- https://docs.voxel51.com/user_guide/evaluation.html#map-and-pr-curves
- https://docs.voxel51.com/integrations/coco.html#map-and-pr-curves
- In the `COCOEvaluationConfig`:

```
        compute_mAP (False): whether to perform the necessary computations so
            that mAP, mAR, and PR curves can be generated
```

brimoor (commented):
This LGTM now, but let's wait until `fiftyone==1.2.0` is merged back into `develop` and then rebase this PR so that we can confirm that mAR behaves as expected in the Model Evaluation panel in the App:
- It is displayed when `compute_mAP=True` was passed
- Otherwise, it does not appear

brimoor (approved):
LGTM

PR #5279, opened 2024-12-16
Add support for custom evaluation metrics

brimoor:
See https://github.com/voxel51/fiftyone-plugins/pull/190

## Summary by CodeRabbit

- **New Features**
	- Added support for custom metrics across various evaluation methods in FiftyOne.
	- Enhanced evaluation configurations to include custom metric specifications.
	- Introduced new base classes for more flexible evaluation methods.
	- Integrated custom metrics into the evaluation panel for better display of results.

- **Improvements**
	- Updated evaluation result classes to handle custom metrics.
	- Expanded evaluation framework to support more dynamic metric computation.
	- Added utility methods for managing custom metrics in different evaluation contexts.

- **Technical Enhancements**
	- Refactored evaluation method classes to improve modularity.
	- Added type safety and error handling for custom metric processing.
	- Improved logging and reporting of evaluation results.

### DIFF
--- a/fiftyone/operators/evaluation_metric.py
+++ b/fiftyone/operators/evaluation_metric.py
@@ -40,7 +40,7 @@ class EvaluationMetric(Operator):
     """Base class for evaluation metric operators."""
 
     def get_parameters(self, ctx, inputs):
-        """Defines any necessary properties to collect this evaluaiton metric's
+        """Defines any necessary properties to collect this evaluation metric's
         parameters from a user during prompting.
 
         Args:
--- a/fiftyone/operators/evaluation_metric.py
+++ b/fiftyone/operators/evaluation_metric.py
@@ -1,0 +1,127 @@
+"""
+Evaluation metric operators.
+
+| Copyright 2017-2025, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
+from .operator import Operator, OperatorConfig
+
+
+class EvaluationMetricConfig(OperatorConfig):
+    """Configuration class for evaluation metrics.
+
+    Args:
+        name: the name of the evaluation metric
+        label (name): a label for the evaluation metric
+        description (None): a description of the evaluation metric
+        eval_types (None): an optional list of evaluation method types that
+            this metric supports
+        lower_is_better (True): whether lower values of the metric are better
+        **kwargs: other kwargs for :class:`fiftyone.operators.OperatorConfig`
+    """
+
+    def __init__(
+        self,
+        name,
+        label=None,
+        description=None,
+        eval_types=None,
+        lower_is_better=True,
+        **kwargs,
+    ):
+        super().__init__(name, label=label, description=description, **kwargs)
+        self.eval_types = eval_types
+        self.lower_is_better = lower_is_better
+
+
+class EvaluationMetric(Operator):
+    """Base class for evaluation metric operators."""
+
+    def get_parameters(self, ctx, inputs):
+        """Defines any necessary properties to collect this evaluaiton metric's
+        parameters from a user during prompting.
+
+        Args:
+            ctx: an :class:`fiftyone.operators.ExecutionContext`
+            inputs: a :class:`fiftyone.operators.types.Property`
+        """
+        pass
+
+    def parse_parameters(self, ctx, params):
+        """Performs any necessary execution-time formatting to this evaluation
+        metric's parameters.
+
+        Args:
+            ctx: an :class:`fiftyone.operators.ExecutionContext`
+            params: a params dict
+        """
+        pass
+
+    def compute(self, samples, results, **kwargs):
+        """Computes the evaluation metric for the given collection.
+
+        Args:
+            samples: a :class:`fiftyone.core.collections.SampleCollection`
+            results: an :class:`fiftyone.core.evaluation.EvaluationResults`
+            **kwargs: arbitrary metric-specific parameters
+
+        Returns:
+            an optional aggregate metric value to store on the results
+        """
+        raise NotImplementedError("Subclass must implement compute()")
+
+    def get_fields(self, samples, config, eval_key):
+        """Lists the fields that were populated by the evaluation metric with
+        the given key, if any.
+
+        Args:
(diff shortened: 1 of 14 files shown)


fiftyone/utils/eval/base.py
```diff
@@ -6,17 +6,124 @@
...
+        if not self.config.custom_metrics:
+            return {}
+
+        if isinstance(self.config.custom_metrics, list):
+            return {m: None for m in self.config.custom_metrics}
+
+        return self.config.custom_metrics
+
+    def compute_custom_metrics(self, samples, eval_key, results):
+        results.custom_metrics = {}
```
manushreegangwar, line 52:
```suggestion
        if self.config.custom_metrics:
            results.custom_metrics = {}
```

brimoor: Addressed like [this](https://github.com/voxel51/fiftyone/pull/5279/commits/49c33075212dd3092eac07a83031eac6b7baba8c) ✅

plugins/panels/model_evaluation/__init__.py
```diff
@@ -146,6 +146,12 @@ def get_mar(self, results):
         except Exception as e:
             return None
 
+    def get_custom_metrics(self, results):
+        try:
+            return results.custom_metrics
```
manushreegangwar, line 151: We are still missing the CustomMetric class that @ritch suggested: https://github.com/voxel51/fiftyone/pull/5389#issuecomment-2596844520


results.custom_metrics should return a list of CustomMetric objects. Working on adding those on top of these changes.

brimoor: I don't see value in further modifications to this PR. `results.custom_metrics` is currently JSON in roughly same shape as what @ritch suggested in https://github.com/voxel51/fiftyone/pull/5389#issuecomment-2596844520, so it achieves the same thing, just without the class interface. That seems fine to me as `results.custom_metrics` is not intended for direct usage by SDK users. It's really just an internal data structure to support the Model Evaluation panel.

Note that I refactored so that custom metrics are included in `results.metrics()` and `results.print_metrics()`. So those are the only things that SDK users should need to care about IMO.

Also note that introducing a class might be complex as `results` need to be serializable.

manushreegangwar: Deferring to @ritch for the app side changes.

Looks good to me on the SDK side. I am okay with merging this without the modifications proposed for the app.

fiftyone/operators/evaluation_metric.py
```diff
@@ -0,0 +1,127 @@
...
+        super().__init__(name, label=label, description=description, **kwargs)
+        self.eval_types = eval_types
+        self.lower_is_better = lower_is_better
+
+
+class EvaluationMetric(Operator):
+    """Base class for evaluation metric operators."""
+
+    def get_parameters(self, ctx, inputs):
+        """Defines any necessary properties to collect this evaluaiton metric's
```
manushreegangwar, line 43:
```suggestion
        """Defines any necessary properties to collect this evaluation metric's
```

manushreegangwar (approved):
LGTM.

Ran `evaluate_detections` (app + SDK) and `evaluate_regression` (SDK) with and without custom metrics.

plugins/panels/model_evaluation/__init__.py
ritch: > I don't see value in further modifications to this PR. results.custom_metrics is currently JSON in roughly same shape as what @ritch suggested in https://github.com/voxel51/fiftyone/pull/5389#issuecomment-2596844520, so it achieves the same thing, just without the class interface.

If devs end up using this feature we can add in the ability to define a `fiftyone.operator.types.Property` and use that to render the value. -1 for an adhoc way to define labels, tooltips, type info, min/max values, max number of digits to display etc. That should all be defined via an operator property and that type system that already supports a lot of these.

ritch (approved):

fiftyone/utils/eval/base.py
```diff
@@ -5,18 +5,192 @@
...
+        Also includes any custom metrics from :attr:`custom_metrics`.
+
+        Args:
+            *args: subclass-specific positional arguments
+            **kwargs: subclass-specific keyword arguments
+
+        Returns:
+            a dict
+        """
+        return self._get_custom_metrics()
```
manushreegangwar, line 159: Why add this method here? Wouldn't this return just the custom metrics since base results don't have any other metrics?

manushreegangwar: All derived classes override this so it's not directly used. I am guessing this is for future eval methods which may only use custom metrics.

manushreegangwar (approved):

fiftyone/utils/eval/base.py
brimoor: I added this for two reasons:
1. So I had something to link to in the `EvaluationMetricConfig` docstring
2. To establish "metrics" as a concept that all evaluation protocols should contemplate. And yes, the base implementation simply returns *only* custom metrics in case a future evaluation protocol doesn't implement any builtin metrics and relies solely on custom metrics

fiftyone/utils/eval/base.py
```diff
@@ -147,8 +177,7 @@ def _get_custom_metrics(self):
 
         try:
             for uri, metric in self.custom_metrics.items():
-                name = uri.rsplit("/", 1)[-1]
-                metrics[name] = metric["value"]
+                metrics[metric["key"]] = metric["value"]
```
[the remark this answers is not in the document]
brimoor: Yep. In order to support applying a metric multiple times, we'd need to do a few more things, including the ability for a metric to customize its `aggregate_key` and `label` (for the ME panel) based on its `kwargs`, not just statically defined by its `EvaluationMetricConfig`.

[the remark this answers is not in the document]
brimoor: I recommend waiting until we have signal that users want to do this before implementing.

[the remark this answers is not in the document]
manushreegangwar: Yes, I agree. We can wait on this.

PR #5283, opened 2024-12-16
Dense Mask IoU

mwoodson1:

Now when calling `fo.utils.iou.compute_ious(preds, gts, tolerance=None)` with pixel masks, the IoU will be computed using pixels instead of converting to contours. This is especially needed when trying to evaluate segmentation models which typically use dense IoU as a metric instead of one computed on contours.

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Release Notes

- **Documentation**
	- Enhanced documentation for model evaluation in FiftyOne framework
	- Clarified Intersection over Union (IoU) calculations for instance segmentations and polygons
	- Added detailed guidance on using `use_masks` and `use_boxes` parameters

- **Improvements**
	- Added new functions to compute IoU metrics for dense masks and bounding boxes
	- Improved flexibility in IoU computation methods
	- Enhanced support for different detection task evaluations

- **Technical Updates**
	- Refined docstrings in evaluation configuration classes
	- Introduced more precise IoU calculation techniques

### DIFF
--- a/fiftyone/utils/iou.py
+++ b/fiftyone/utils/iou.py
@@ -136,11 +136,6 @@ def compute_ious(
         )
 
     if use_masks:
-        # @todo when tolerance is None, consider using dense masks rather than
-        # polygonal approximations?
-        if tolerance is None:
-            tolerance = 2
-
         return _compute_mask_ious(
             preds,
             gts,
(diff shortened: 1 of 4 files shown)


mwoodson1:
@ehofesmann I know you mentioned there would be some edge cases regarding different pred and ground truth resolutions. Do you think my implementation covers this sufficiently? If not can you provide a small test case I can use?

ehofesmann:
> @ehofesmann I know you mentioned there would be some edge cases regarding different pred and ground truth resolutions. Do you think my implementation covers this sufficiently? If not can you provide a small test case I can use?

Very nice work! I haven't run this myself, but I do believe that this todo would need to be addressed in order to support preds and gts with different resolutions. You could try to test it with something like:

```python
gt_mask = np.zeros((90,90))
gt_mask[:60,:] = 1
pred_mask = np.zeros((50,50))
pred_mask[25:,:] = 1

gt = fo.Detection(bounding_box=[0.1,0.1,0.1,0.1], label="test", mask=gt_mask)
pred = fo.Detection(bounding_box=[0.1,0.1,0.1,0.1], label="test", mask=pred_mask)
fo.utils.iou._dense_iou(pred, gt) 
```

ehofesmann:
> I'm not sure I understand the use case here. When do we have predictions and ground truths of different resolutions? And for the example you shared what is the expected behaviour

I could definitely imagine a case where the ground truth labels were annotated on higher res images than what a segmentation model results. And also that 2 segmentation models might have resulting masks of different resolutions and you want to compare one with the other.

For the example I would expect there to be an overlap of 1/6 (one mask covers the bottom 2/3 of a square image and the other mask the top 1/2)

mwoodson1:
Got it, this should be handled correctly now.

brimoor (approved):
Code LGTM from a static review standpoint.

Note: I haven't run this myself; relying on you for implementation correctness.

Just to double confirm: as @ehofesmann pointed out, FO does not enforce nor assume that the masks for each object have the same resolution, so the IoU implementation should gracefully handle two `Detection` instances whose masks are different sizes.

I added some documentation for this change here: https://github.com/voxel51/fiftyone/pull/5283/commits/79080ffc01dfa69db523641eb3333c951c415584

mwoodson1:
@manushreegangwar @jacobsela could one of you validate the correctness of my implementation? Happy to provide guidance on how to test offline.

fiftyone/utils/iou.py
```diff
@@ -528,6 +525,65 @@ def compute_bbox_iou(gt, pred, gt_crowd=False):
...
+    y1 = round(pred_bb[1] * pred_img_h)
+    x2 = round(x1 + (pred_bb[2] * pred_img_w))
+    y2 = round(y1 + (pred_bb[3] * pred_img_h))
+    pred_mask_full[y1:y2, x1:x2] = pred_mask
+
+    if gt_img_w != pred_img_w or gt_img_h != pred_img_h:
+        gt_size = gt_img_w * gt_img_h
+        pred_size = pred_img_w * pred_img_h
+        if gt_size > pred_size:
+            pred_mask_full = etai.resize(
```
jacobsela, line 573: This resolves to `cv2.resize` which by default uses linear interpolation. I'm pretty sure values in (0, 1) are treated as boolean True in `np.logical_and` which may cause some strange edge case bugs in thin objects for example. I can check this more thoroughly later. Besides that looks good to me but I haven't run it yet.

jacobsela (commented):
Haven't had a chance to verify by running code yet. Generally looks fine.

fiftyone/utils/iou.py
```diff
@@ -528,6 +525,65 @@ def compute_bbox_iou(gt, pred, gt_crowd=False):
...
+    pred_img_w = round(pred_mask_w / pred_bb[2])
+    pred_img_h = round(pred_mask_h / pred_bb[3])
+
+    gt_mask_full = np.zeros((gt_img_h, gt_img_w))
+    pred_mask_full = np.zeros((pred_img_h, pred_img_w))
+
+    x1 = round(gt_bb[0] * gt_img_w)
+    y1 = round(gt_bb[1] * gt_img_h)
+    x2 = round(x1 + (gt_bb[2] * gt_img_w))
+    y2 = round(y1 + (gt_bb[3] * gt_img_h))
```
manushreegangwar, line 560: You can make a function out of this since it is repeated for `pred`

fiftyone/utils/iou.py
```diff
@@ -624,6 +684,61 @@ def _compute_bbox_ious(
...
+
+    if sparse:
+        ious = defaultdict(list)
+    else:
+        ious = np.zeros((len(preds), len(gts)))
+
+    if iscrowd is not None:
+        gt_crowds = [iscrowd(gt) for gt in gts]
+    else:
+        gt_crowds = [False] * len(gts)
```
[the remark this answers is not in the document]
mwoodson1: Good catch.

fiftyone/utils/iou.py
mwoodson1: This is true, unfortunately if detections and ground truths are of different aspect ratios or sizes then some sort of resizing needs to be done.

manushreegangwar (approved):

fiftyone/utils/iou.py
jacobsela: I think the best thing to do is to change the interpolation method or introduce some thresholding after the resize.

mwoodson1: I updated the interpolation method to handle this but it looks like the [PR](https://github.com/voxel51/fiftyone/pull/5369) Brian mentioned should add more general thresholding of mask values. At that time we might consider making the interpolation method a parameter instead of fixed.

PR #5289, opened 2024-12-17
add support for collection overlays

sashankaryal:
We don't have a precedent for `sources` in labels accounting for collection label types like `Detections`. This PR adds support for that.

## Summary by CodeRabbit

- **New Features**
  - Enhanced handling of overlay data processing with new parameters for improved flexibility.
  - Updated media fields related to detections for better metadata retrieval.

- **Bug Fixes**
  - Improved error handling during overlay image fetching and decoding processes.

- **Refactor**
  - Streamlined functions for additional media fields and detection handling.
  - Adjusted function signatures for better structure in media field handling.

### DIFF
--- a/app/packages/looker/src/worker/disk-overlay-decoder.ts
+++ b/app/packages/looker/src/worker/disk-overlay-decoder.ts
@@ -92,9 +92,7 @@ export const decodeOverlayOnDisk = async (
   }
 
   // convert absolute file path to a URL that we can "fetch" from
-  const overlayImageUrl = getSampleSrc(
-    sources[`${field}.${overlayPathField}`] || label[overlayPathField]
-  );
+  const overlayImageUrl = getSampleSrc(source || label[overlayPathField]);
   const urlTokens = overlayImageUrl.split("?");
 
   let baseUrl = overlayImageUrl;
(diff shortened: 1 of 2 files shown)


app/packages/looker/src/worker/disk-overlay-decoder.ts
```diff
@@ -74,16 +80,29 @@ export const decodeOverlayOnDisk = async (
...
 
+  // if we have an explicit source defined from sample.urls, use that
+  // otherwise, use the path field from the label
+  let source = sources[`${field}.${overlayPathField}`];
+
+  if (typeof overlayCollectionProcessingParams !== "undefined") {
+    // example: for detections, we need to access the source from the parent label
+    // like: if field is "prediction_masks", we're trying to get "predictiion_masks.detections[INDEX].mask"
+    source =
+      sources[
```
tom-vx51, line 91: any concern that the key we're constructing here doesn't exist on `sources`?

sashankaryal: in which case it'd return undefined. In fact, it won't exist in OSS

tom-vx51 (approved):

app/packages/looker/src/worker/disk-overlay-decoder.ts
tom-vx51: my eyes told me the `].` deref was outside the key so thought we might be derefing an `undefined`. my mistake 🙏

PR #5291, opened 2024-12-18
simplify delegated operation doc

ritch:
- removes redundant `params` from operations docs
 - removes panel_state from operations docs, since it is unused

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method to enhance the serialization of request parameters by excluding specific keys.

- **Improvements**
	- Updated the constructor to enforce stricter type requirements for the context parameter.
	- Streamlined serialization and deserialization processes for the context attribute.
	- Enhanced error handling and data integrity within repository methods related to operation queuing and state updates.
	- Improved validation for progress parameters in the update progress method.

### DIFF
--- a/fiftyone/factory/repos/delegated_operation.py
+++ b/fiftyone/factory/repos/delegated_operation.py
@@ -214,6 +214,7 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
             context.request_params["dataset_id"] = str(op.dataset_id)
             context.request_params["dataset_name"] = context.dataset.name
 
+        op.context = context
         doc = self._collection.insert_one(op.to_pymongo())
         op.id = doc.inserted_id
         return DelegatedOperationDocument().from_pymongo(op.__dict__)
--- a/fiftyone/factory/repos/delegated_operation_doc.py
+++ b/fiftyone/factory/repos/delegated_operation_doc.py
@@ -24,17 +24,14 @@ class DelegatedOperationDocument(object):
         self,
         operator: str = None,
         delegation_target: str = None,
-        context: dict = None,
+        context: ExecutionContext = None,
         is_remote: bool = False,
     ):
         self.operator = operator
         self.label = None
         self.delegation_target = delegation_target
-        self.context = (
-            context.to_dict()
-            if isinstance(context, ExecutionContext)
-            else context
-        )
+        assert isinstance(context, ExecutionContext)
+        self.context = context
         self.run_state = (
             ExecutionRunState.SCHEDULED
             if is_remote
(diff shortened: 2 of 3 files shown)


fiftyone/factory/repos/delegated_operation.py
```diff
@@ -214,6 +214,7 @@ def queue_operation(self, **kwargs: Any) -> DelegatedOperationDocument:
             context.request_params["dataset_id"] = str(op.dataset_id)
             context.request_params["dataset_name"] = context.dataset.name
 
+        op.context = context
```
ritch, line 217: This fixes an existing issue that I think might explain some of the weirdness/handling of `op.context` as `dict`!

Without this change [this](https://github.com/voxel51/fiftyone/pull/5291/files#diff-8c618b37055d71658b7666c9e677d916dfbb8b9f43876a5a2a983be19b63c100R192) is a no-op.

fiftyone/operators/executor.py
```diff
@@ -889,6 +889,11 @@ def store(self, store_name):
         dataset_id = self.dataset._doc.id
         return ExecutionStore.create(store_name, dataset_id)
 
+    def _get_serialized_request_params(self):
```
ritch, line 892: Not really liking this as the interface...but didn't feel that changing `.serialize()` was the right move. However there is no need for `panel_state` in a serialized `ctx`. Instead it should really be outside of `request_params` to begin with. I've started that change elsewhere, but its a much more complex change and regresses some functionality in the dashboard plugin.

fiftyone/factory/repos/delegated_operation_doc.py
```diff
@@ -24,17 +24,14 @@ def __init__(
...
     ):
         self.operator = operator
         self.label = None
         self.delegation_target = delegation_target
-        self.context = (
-            context.to_dict()
-            if isinstance(context, ExecutionContext)
-            else context
-        )
+        assert isinstance(context, ExecutionContext) or context is None
```
kaixi-wang, line 33: don't use assert

ritch: used this instead:

```python
        try:
            assert isinstance(context, ExecutionContext) or context is None
        except:
            raise AttributeError
```

fiftyone/factory/repos/delegated_operation_doc.py
```diff
@@ -24,17 +24,17 @@ def __init__(
...
         self.operator = operator
         self.label = None
         self.delegation_target = delegation_target
-        self.context = (
-            context.to_dict()
-            if isinstance(context, ExecutionContext)
-            else context
-        )
+        try:
+            assert isinstance(context, ExecutionContext) or context is None
```
kaixi-wang, line 34: You can just do if... else raise.

kaixi-wang (approved):
Lgtm

CamronStaley (approved):

PR #5305, opened 2024-12-20
Bump to ETA v0.13.1

brimoor:
This optimizes `dataset.first()` by ensuring that `len(dataset)` is not called unnecessarily.

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -973,19 +973,6 @@ class ProgressBar(etau.ProgressBar):
         self._progress = progress
         self._callback = callback
 
-    def __call__(self, iterable):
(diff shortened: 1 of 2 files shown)


fiftyone/core/utils.py
```diff
@@ -973,19 +973,6 @@ def __init__(self, total=None, progress=None, quiet=None, **kwargs):
         self._progress = progress
         self._callback = callback
 
-    def __call__(self, iterable):
```
brimoor, line 976: No longer needed; this is now handled by https://github.com/voxel51/eta/commit/9b4db83db034fd4f29323ca94bbac2963c3b3c16

findtopher (approved):
voxel51-eta v0.13.1 has been published

PR #4990, opened 2024-10-25
Segmentation annotations default flow modifications

geke-mir:
## What changes are proposed in this pull request?

Require users to specify a `mask_targets` argument for new segmentations annotations, otherwise fall back onto the dataset's defaults specified in `dataset.mask_targets` or `dataset.default_mask_targets`.

## How is this patch tested? If it is not, please explain why.

Manual testing of annotation flow with a test CVAT instance.

### Is this a user-facing change that should be mentioned in the release notes?

-   [ ] No. You can skip the rest of this section.
-   [x] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

When requesting semantic segmentations annotations, users must specify `mask_targets` for fields that do not yet exist.  If no targets are specified, the targets will be pulled from `dataset.mask_targets` or `dataset.default_mask_targets` if either exist.

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [x] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced annotation functionality requiring mask targets for new segmentation fields.
	- Improved error handling for empty samples during annotation.

- **Bug Fixes**
	- Refined logic for determining mask targets, ensuring proper retrieval from existing samples or defaults.

- **Documentation**
	- Added detailed docstrings for better clarity on function parameters and return values.

PR #5318, opened 2024-12-26
Fallback to mask targets when annotating existing segmentation fields

brimoor:
Merging https://github.com/voxel51/fiftyone/pull/4990 plus some documentation updates.

## Summary by CodeRabbit

- **Documentation**
	- Clarified handling of `mask_targets` and `classes` parameters in the annotation process across CVAT, Labelbox, and Label Studio integrations.
	- Enhanced explanations regarding the requirements for new and existing label fields in the FiftyOne annotation API.
	- Updated user guide to reflect changes in the `annotate()` method, improving clarity on parameter usage.


ehofesmann (approved):
LGTM! I manually tested the different approaches for providing mask targets and it is working well!

PR #5321, opened 2024-12-27
Add FiftyOneTorchDataset

### LINKED ISSUE #1302: [BUG] Failed to bind to port with PyTorch DistributedDataParallel
### System information

-   **Ubuntu 18.04**:
-   **FiftyOne installed from pip**:
-   **FiftyOne version v0.13.2**:
-   **Python 3.6.9 (virtual env)**:

### Commands to reproduce

```python -u -m torch.distributed.launch --nproc_per_node=4 test.py```

### Describe the problem

FiftyOne can't bind ports after calling the script with PyTorch `DistributedDataParallel`. There are no problems with a single GPU, but with 2 and more GPUs script terminating instantly and very rare works without issues. Probably it's directly related to the multi-process parallelism of `DistributedDataParallel`.  

### Other info / logs

```
*****************************************
Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed.
*****************************************
Process 16911 (service/main.py --51-service db --multi) did not respond
{"t":{"$date":"2021-09-27T06:21:42.600Z"},"s":"I",  "c":"CONTROL",  "id":20697,   "ctx":"main","msg":"Renamed existing log file","attr":{"oldLogPath":"/home/ubuntu/.fiftyone/var/lib/mongo/log/mongo.log","newLogPath":"/home/ubuntu/.fiftyone/var/lib/mongo/log/mongo.log.2021-09-27T06-21-42"}}
Subprocess ['/home/ubuntu/OID/lib/python3.6/site-packages/fiftyone/db/bin/mongod', '--dbpath', '/home/ubuntu/.fiftyone/var/lib/mongo', '--logpath', '/home/ubuntu/.fiftyone/var/lib/mongo/log/mongo.log', '--port', '0', '--nounixsocket'] exited with error 100:
{"t":{"$date":"2021-09-27T06:21:42.609Z"},"s":"I",  "c":"CONTROL",  "id":20697,   "ctx":"main","msg":"Renamed existing log file","attr":{"oldLogPath":"/home/ubuntu/.fiftyone/var/lib/mongo/log/mongo.log","newLogPath":"/home/ubuntu/.fiftyone/var/lib/mongo/log/mongo.log.2021-09-27T06-21-42"}}
Subprocess ['/home/ubuntu/OID/lib/python3.6/site-packages/fiftyone/db/bin/mongod', '--dbpath', '/home/ubuntu/.fiftyone/var/lib/mongo', '--logpath', '/home/ubuntu/.fiftyone/var/lib/mongo/log/mongo.log', '--port', '0', '--nounixsocket'] exited with error 100:
{"t":{"$date":"2021-09-27T06:21:42.627Z"},"s":"I",  "c":"CONTROL",  "id":20697,   "ctx":"main","msg":"Renamed existing log file","attr":{"oldLogPath":"/home/ubuntu/.fiftyone/var/lib/mongo/log/mongo.log","newLogPath":"/home/ubuntu/.fiftyone/var/lib/mongo/log/mongo.log.2021-09-27T06-21-42"}}
Subprocess ['/home/ubuntu/OID/lib/python3.6/site-packages/fiftyone/db/bin/mongod', '--dbpath', '/home/ubuntu/.fiftyone/var/lib/mongo', '--logpath', '/home/ubuntu/.fiftyone/var/lib/mongo/log/mongo.log', '--port', '0', '--nounixsocket'] exited with error 100:
Uncaught exception
  File "train_test.py", line 7, in <module>
    from data_loader_fletch import create_data_loaders
  File "/home/ubuntu/LaboroTrainingOpenImages/data_loader_fletch.py", line 1, in <module>
    import fiftyone as fo
  File "/home/ubuntu/OID/lib/python3.6/site-packages/fiftyone/__init__.py", line 25, in <module>
    from fiftyone.__public__ import *
  File "/home/ubuntu/OID/lib/python3.6/site-packages/fiftyone/__public__.py", line 14, in <module>
    foo.establish_db_conn(config)
  File "/home/ubuntu/OID/lib/python3.6/site-packages/fiftyone/core/odm/database.py", line 77, in establish_db_conn
    port = _db_service.port
  File "/home/ubuntu/OID/lib/python3.6/site-packages/fiftyone/core/service.py", line 295, in port
    return self._wait_for_child_port()
  File "/home/ubuntu/OID/lib/python3.6/site-packages/fiftyone/core/service.py", line 179, in _wait_for_child_port
    return find_port()
  File "/home/ubuntu/OID/lib/python3.6/site-packages/retrying.py", line 49, in wrapped_f
    return Retrying(*dargs, **dkw).call(f, *args, **kw)
  File "/home/ubuntu/OID/lib/python3.6/site-packages/retrying.py", line 212, in call
    raise attempt.get()
  File "/home/ubuntu/OID/lib/python3.6/site-packages/retrying.py", line 247, in get
    six.reraise(self.value[0], self.value[1], self.value[2])
  File "/home/ubuntu/OID/lib/python3.6/site-packages/six.py", line 719, in reraise
    raise value
  File "/home/ubuntu/OID/lib/python3.6/site-packages/retrying.py", line 200, in call
    attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
  File "/home/ubuntu/OID/lib/python3.6/site-packages/fiftyone/core/service.py", line 177, in find_port
    raise ServiceListenTimeout(etau.get_class_name(self), port)
fiftyone.core.service.ServiceListenTimeout: fiftyone.core.service.DatabaseService failed to bind to port
```

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community.

### PR DESCRIPTION
**Pull request by jacobsela**
## PR Needs

1. Does it work as advertised
2. Does it satisfy the needs of an ML practitioner trying to train with FiftyOne 
3. Does it work in a scalable and efficient manner 

## Example Snippet

```python
import torch
from torchvision.transforms import v2 as T
from torchvision.models import resnet18
from PIL import Image

import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone.utils.torch import FiftyOneTorchDataset

transforms = T.Compose([
    T.Resize((224, 224)),
    T.PILToTensor(),
    T.ToDtype(torch.float32),
    T.Normalize(mean=(0.5, 0.5, 0.5), std=(0.5, 0.5, 0.5)),
])

# the key to this method is defining a get_item.
# this function is the contact point between FiftyOne and PyTorch
# it accepts a sample or sample dict from FiftyOne
# and returns model inputs
# because of serializability concerns, this must be a top level function
def get_item(sample):
    image = Image.open(sample['filepath'])
    image = transforms(image)
    label = sample['ground_truth']['label']
    label = torch.tensor(classes.index(label), dtype=torch.long)
    return {'images': image, 'labels': label}

# required for multiprocessing
if __name__ == "__main__":

    if torch.multiprocessing.get_start_method(allow_none=True) is None:
        torch.multiprocessing.set_start_method("forkserver")
        torch.multiprocessing.set_forkserver_preload(['fiftyone', 'torch'])

    # only works with persistent datasets
    dataset.persistent = True

    # fully functional PyTorch dataset
    torch_dataset = dataset.to_torch(get_item)

    print(1, torch_dataset[0]['images'].shape)
    print(2, torch_dataset[0]['labels'])

    model = resnet18()
    model.fc = torch.nn.Linear(512, 10)

    res = model(torch_dataset[0]['images'].unsqueeze(0))
    print(3, res)

    # you can use a dataloader with the datset and use it for training
    # have to create a new torch dataset here because the previous one is not pickleable
    # this is because we access the samples in the dataset
    # rendering the dataset unpickleable
    torch_dataset = dataset.to_torch(get_item)
    dataloader = torch.utils.data.DataLoader(
        torch_dataset,
        batch_size=10,
        num_workers=2,
        worker_init_fn=FiftyOneTorchDataset.worker_init # remember to pass the init function!
    )

    loss_fn = torch.nn.CrossEntropyLoss()
    optimizer = torch.optim.Adam(model.parameters())

    # regular training loop
    for i, sample_batch in enumerate(dataloader):
        print(4, sample_batch['images'].shape)
        print(5, sample_batch['labels'])
        
        res = model(sample_batch['images'])
        print(6, res.shape)

        loss = loss_fn(res, sample_batch['labels'])
        print(7, loss)

        optimizer.zero_grad()
        loss.backward()
        optimizer.step()
        break
```

Expected output:
```
Split 'test' already downloaded
Loading existing dataset 'cifar10-test'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
Split 'test' already downloaded
Loading existing dataset 'cifar10-test'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
1 torch.Size([3, 224, 224])
2 tensor(3)
3 tensor([[-0.0039, -0.0782,  0.3495, -0.1928,  0.1146,  0.1937, -0.5634,  0.4737,
         -0.9304,  0.1114]], grad_fn=<AddmmBackward0>)
Split 'test' already downloaded
Loading existing dataset 'cifar10-test'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
Split 'test' already downloaded
Loading existing dataset 'cifar10-test'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
4 torch.Size([10, 3, 224, 224])
5 tensor([3, 8, 8, 0, 6, 6, 1, 6, 3, 1])
6 torch.Size([10, 10])
7 tensor(2.7122, grad_fn=<NllLossBackward0>)
```

1. New class `FiftyOneTorchDataset`.
2. added top-level `to_torch` method.
3. some DDP utils.
4. Recipes for use (still missing DDP example)

Resolves #1302 

1. Benchmarked memory footprint
6. Benchmarked dataset latency
7. Validated data loaders load correct samples for each trainer
8. Trained mint classifier w/ new class.
9. Currently training MAGE in a DDP setup - will report results.

        notes for FiftyOne users.

New `to_torch` method allows for easy transfer to a fully functional `torch.utils.data.Dataset` object.
Please see recipes for examples and tutorials.

-   [x] Other - `fiftyone.utils.torch`

## Summary by CodeRabbit

- **New Features**
	- Introduced a script for training and validating models using the MNIST dataset with PyTorch.
	- Added a new Jupyter notebook for a basic training example with the MNIST dataset.
	- Implemented a method to convert sample collections into a PyTorch-compatible format.
	- Added a new class for creating PyTorch datasets from FiftyOne datasets, enhancing integration with distributed training.
	- Introduced several utility functions and classes to optimize data handling in distributed training environments.
	- Added a comprehensive script for training models using Distributed Data Parallel (DDP) with PyTorch.

- **Bug Fixes**
	- Improved data handling and efficiency in multi-worker environments.

- **Documentation**
	- Enhanced clarity with markdown explanations in the new Jupyter notebook.


jacobsela:
MAGE training successful. Very coarse test, but nice to see still. 
image

jacobsela:
@mwoodson1 @manushreegangwar Can you please give this a review?

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1467,186 @@ def _is_string_array(targets):
...
+        - get_item: a `Callable[:class:`fo.core.sample.SampleView`, Any]`
+            Must be a serializable function.
+        - cache_fields (None): a list of strings. Fields to cache in memory. If this
+            argument is passed, get_item should be from from a dict with keys and values
+            corresponding to the sample's fields and values to the model input.
+            This argument is highly recommended, as it offers a significant performance
+            boost.
+            IMPORTANT:  For proper functionality, the field type must be serializable.
+                        DO NOT PASS OBJECT FIELDS. This will cause memory usage of this object
+                        to multiply by (num dataloaders * num gpus + num gpus).
```
mwoodson1, line 1484: Since this seems to be a rather important notice, you could do some checks in the code itself to ensure `cache_fields` is serializable and not an object field. Better to not rely too much on humans reading the documentation thoroughly :)

fiftyone/utils/torch.py
```diff
@@ -2030,3 +2217,198 @@ def _load_image(image_path, use_numpy, force_rgb):
...
+
+# taken from https://github.com/ppwwyyxx/RAM-multiprocess-dataloader/blob/795868a37446d61412b9a58dbb1b7c76e75d39c4/serialize.py#L19
+# as well as https://github.com/facebookresearch/detectron2/blob/main/detectron2/utils/comm.py
+class NumpySerializedList:
+    def __init__(self, lst: list):
+        def _serialize(data):
+            buffer = pickle.dumps(data, protocol=-1)
+            return np.frombuffer(buffer, dtype=np.uint8)
+
+        print(
```
mwoodson1, line 2230: In general I think method printing like this should be logged instead. Makes it easier to debug in production enviornments where logs get piped but print statements do not.

mwoodson1 (dismissed):
I left some specific notes, mostly regarding documentation but here is some high level feedback

- The documentation for the FiftyOneTorchDataset class is much too long imo and should be relocated to a user guide or tutorial documentation
- A lot of the documentation seems distributed training focused. Some users may use this feature without training on multiple machines, especially our open source users. Documentation for both is important but should be separated.
- There are a lot of gotchas with this feature. The biggest one I see is only mentioned in the basic_example.ipynb example stating you need to call to_torch again after sampling, otherwise it's not suitable for multiprocessing. Put as many of these into warnings in the code itself or make sure to call all of them out in one place to avoid confusion.

fiftyone/utils/torch.py
jacobsela: Just did some testing:
1. code fails during dataset creation if field values are not serializable.
2. If the field values are serializable memory usage remains fine.

This is because the main process serializes into bytes and puts bytes on shared memory. When workers read bytes from shared memory they:
1. do not copy-on-read the bytes themselves because they are a numpy array / torch tensor
2. make copies of just the sample field values they currently need (see `__getitem__` for `NumpySerializedList` and `TorchSerializedList` - `pickle.loads` creates an instance of whatever underlying object was serialized)

(2) is fine because regular GC will clean up the loaded instances when they are no longer needed. We are just trying to avoid having the entire field values loaded on memory all at once for all of the workers.

Updated documentation to reflect this.

fiftyone/utils/torch.py
```diff
@@ -26,12 +30,16 @@
 import fiftyone.core.odm as foo
 import fiftyone.core.utils as fou
 import fiftyone.utils.image as foui
+import fiftyone.core.collections as focol
```
manushreegangwar, line 33: `foc` is used as the alias elsewhere

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1468,189 @@ def _is_string_array(targets):
         return False
 
 
+class FiftyOneTorchDataset(Dataset):
+    """A class that accepts a FO dataset and creates a corresponding torch.utils.data.Dataset
+
+    Args:
+        - samples: a :class:`fo.core.collections.SampleCollection`
```
manushreegangwar, line 1475:
```suggestion
        samples: a :class:`fo.core.collections.SampleCollection`
```

manushreegangwar: as per https://github.com/voxel51/fiftyone/blob/develop/STYLE_GUIDE.md#docstrings

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1467,186 @@ def _is_string_array(targets):
...
+            This argument is highly recommended, as it offers a significant performance
+            boost.
+            IMPORTANT:  For proper functionality, the field type must be serializable.
+                        DO NOT PASS OBJECT FIELDS. This will cause memory usage of this object
+                        to multiply by (num dataloaders * num gpus + num gpus).
+        - local_process_group (None) - only pass if running DDP. The process group with
+            each of the processes running the main train script on the machine this
+            object is on.
+
+    Notes:
```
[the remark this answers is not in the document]
jacobsela: I think that the most common usage should be covered in the class docs, but I do see your point.  The recipes in this PR cover all of the notes mentioned.

However, I would like to have a place in the API documentation that lists all of these points, because the user shouldn't have to read a tutorial to understand the behavior of the object. See https://pytorch.org/docs/stable/data.html as an example (in fact, a lot of the points I mention are thoroughly discussed there). How would you suggest to do this?

fiftyone/utils/torch.py
jacobsela: Good catch, will fix.

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1468,189 @@ def _is_string_array(targets):
...
+            corresponding to the sample's fields and values to the model input.
+            This argument is highly recommended, as it offers a significant performance
+            boost.
+            Please note :   the field values must be pickle serializable i.e.
+                            `pickle.dumps(field_value)` should not raise an error.
+                            `pickle.loads(pickle.dumps(field_value))` should have all of the
+                            functionality of the original field value that you would need
+                            in your get_item function.
+
+        - local_process_group (None) - only pass if running DDP. The process group with
```
manushreegangwar, line 1489: nit: expand DDP for clarity

fiftyone/utils/torch.py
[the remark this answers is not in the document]
manushreegangwar: These are great notes but +1 to what Markus said.

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1468,189 @@ def _is_string_array(targets):
...
+             Reads into the FO object during training may slow your code down significantly.
+        - Writing
+            - Writing currently happens on the process from which it is called. This
+            shows moderate slowdown, and will be adressed.
+
+    """
+
+    def __init__(
+        self,
+        samples: focol.SampleCollection,
```
manushreegangwar, line 1549: Regarding type annotations, see [this](https://github.com/voxel51/fiftyone/blob/develop/STYLE_GUIDE.md#docstrings:~:text=Our%20docstring%20style%20is%20derived%20from%20Google%20Python%20style%2C%20with%20the%20exception%20that%20we%20do%20not%20use%20type%20annotations.) -- I personally prefer using them.

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1468,189 @@ def _is_string_array(targets):
...
+        get_item: Callable[fos.SampleView, Any],
+        cache_fields: list[str] = None,
+        local_process_group=None,
+    ):
+        super().__init__()
+
+        self.name = samples._dataset.name
+        # either a whole dataset or a view
```
manushreegangwar, line 1557: nit
```suggestion
        # Either a whole dataset or a view
```

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1468,189 @@ def _is_string_array(targets):
...
 
+class FiftyOneTorchDataset(Dataset):
+    """A class that accepts a FO dataset and creates a corresponding torch.utils.data.Dataset
+
+    Args:
+        - samples: a :class:`fo.core.collections.SampleCollection`
+        - get_item: a `Callable[:class:`fo.core.sample.SampleView`, Any]`
+            Must be a serializable function.
+        - cache_fields (None): a list of strings. Fields to cache in memory. If this
+            argument is passed, get_item should be from from a dict with keys and values
```
manushreegangwar, line 1479:
```suggestion
            argument is passed, get_item should be from a dict with keys and values
```

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1468,189 @@ def _is_string_array(targets):
...
+            else None
+        )
+        self.get_item = get_item
+        # make this a file we call for very large datasets
+        # TODO: remove duplication with cached_fields, just have
+        # ids always be one of the cached fields.
+        self.ids = _to_bytes_array(samples.values("id"))
+
+        self.cached_fields = None
+        self.cache_fields = cache_fields
```
manushreegangwar, line 1570: I find myself misreading `cached_fields` as `cache_fields` and vice-versa.

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1468,189 @@ def _is_string_array(targets):
...
+                    # don't need to pass actual data if not local rank 0
+                    # we read it from shared memory anyways
+                    self.cached_fields[field_name] = TorchShmSerializedList(
+                        [], local_process_group
+                    )
+
+    @property
+    def samples(self):
+        if self._samples == None:
+            self._load_samples()
```
manushreegangwar, line 1610: In general, properties should do trivial computations. Loading an FO dataset does not fit the bill IMO.

fiftyone/utils/torch.py
```diff
@@ -2030,3 +2221,198 @@ def _load_image(image_path, use_numpy, force_rgb):
...
+        return pickle.loads(bytes)
+
+
+class TorchSerializedList(NumpySerializedList):
+    def __init__(self, lst: list):
+        super().__init__(lst)
+        self._addr = torch.from_numpy(self._addr)
+        self._lst = torch.from_numpy(self._lst)
+
+    def __getitem__(self, idx):
```
manushreegangwar, line 2265: Wouldn't this throw an error if `idx` >= len? Add a check.

fiftyone/utils/torch.py
```diff
@@ -2030,3 +2221,198 @@ def _load_image(image_path, use_numpy, force_rgb):
...
+        end_addr = self._addr[idx].item()
+        bytes = memoryview(self._lst[start_addr:end_addr].numpy())
+        return pickle.loads(bytes)
+
+
+class TorchShmSerializedList(TorchSerializedList):
+    def __init__(self, lst: list, local_process_group):
+        if get_local_rank(local_process_group) == 0:
+            super().__init__(lst)
+        if get_local_rank(local_process_group) == 0:
```
manushreegangwar, line 2276: Why is this checking for the same condition in line 2274 and 2276?

manushreegangwar (changes requested):

fiftyone/utils/torch.py
jacobsela: Yes, builtin to PyTorch.

fiftyone/utils/torch.py
jacobsela: This is a standard way of lazy loading attributes (which is needed here). The alternative with `__getattr__` is far less clean. Do you know of other ways of doing this?

fiftyone/utils/torch.py
jacobsela: good catch

fiftyone/utils/torch.py
jacobsela: I'm happy to expand, but what's unclear?

jacobsela: oh you mean the actual word ok

fiftyone/utils/torch.py
jacobsela: unfortunately `foc` is laready taken by `import fiftyone.core.config as foc`

jacobsela:
@mwoodson1 Thank you for the feedback. I've addressed some of the points and wanted to discuss some of the others.

> The documentation for the FiftyOneTorchDataset class is much too long imo and should be relocated to a user guide or tutorial documentation

I agree that it's cumbersome. I wrote a lengthier explanation on the comment above. The recipes added to this PR cover all of the points mentioned. I still think these points should be in the API docs. If the class itself isn't the right place we need to find a different solution.

> A lot of the documentation seems distributed training focused. Some users may use this feature without training on multiple machines, especially our open source users. Documentation for both is important but should be separated.

DDP applies for both multi node and single-node, multi-gpu training. It's relevant almost universally for people that regularly train models. That said, I do agree that the added text is unappealing for new users. This goes back to the previous point, if not there, where should the DDP notes live?

> There are a lot of gotchas with this feature. The biggest one I see is only mentioned in the basic_example.ipynb example stating you need to call to_torch again after sampling, otherwise it's not suitable for multiprocessing. Put as many of these into warnings in the code itself or make sure to call all of them out in one place to avoid confusion.

I agree, it's a big issue. I've added warnings and early error throwing where possible. Unfortunately, a lot of the pitfalls occur outside of code we have purview over. In the example you gave (don't call `self.samples` or query the torch_dataset before instantiating the dataloader), the user's code will fail as soon as they try to pull samples out of a multi-worker dataloader built on an already-queried torch_dataset (even before the worker_init is called!). How can I check for this without having access to the user's code? I have yet to come up with a creative solution. This is the case for a lot of the common pitfalls mentioned.

fiftyone/utils/torch.py
jacobsela: already in the docstring

fiftyone/utils/torch.py
jacobsela: fixed

fiftyone/utils/torch.py
jacobsela: renamed `cache_fields` to `cache_field_names`

fiftyone/utils/torch.py
```diff
@@ -2030,3 +2264,197 @@
...
+    authkey = bytes(torch.multiprocessing.current_process().authkey)
+    all_keys = all_gather(authkey, local_process_group)
+    local_leader_key = all_keys[int(os.environ["RANK"]) - local_rank]
+    if authkey != local_leader_key:
+        logger.info(
+            "Process authkey is different from the key of local leader. This might happen when "
+            "workers are launched independently."
+        )
+        logger.info("Overwriting local authkey ...")
+        multiprocessing.current_process().authkey = local_leader_key
```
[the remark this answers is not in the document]
jacobsela: Torchrun sets these environment variables no need to check

fiftyone/utils/torch.py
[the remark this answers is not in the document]
mwoodson1: I think moving some of this into a user guide makes sense. You may want a product perspective on this though, they likely will have a good idea on what to do for this type of documentation.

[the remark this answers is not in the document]
jacobsela: @brimoor any thoughts?

fiftyone/utils/torch.py
mwoodson1: I agree with Manushree here. While less clean, using `_getattr__` is the more Pythonic way for this type of lazy loading scenario IMO. I don't think it's worth blocking this but if it can be switched in a reasonable amount of time I'd prefer it.

manushreegangwar: +1 not a blocker. You can decide @jacobsela

jacobsela: In that case, for the sake of getting this through in a timely manner I'll leave it as is for now.

daniel-bogdoll:
@jacobsela If you are looking for more inspiration, there were also 2 approaches to this discussed here: https://discord.com/channels/1266527359511564372/1301923324388970587

jacobsela:
Hi @daniel-bogdoll thank you for taking interest in this project! I skimmed over the discord thread and it seems to me that the features discussed there are covered. Is there anything specific you’d like to see?

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1467,186 @@ def _is_string_array(targets):
...
+        self.name = samples._dataset.name
+        # either a whole dataset or a view
+        self.stages = (
+            samples._serialize()
+            if isinstance(samples, fov.DatasetView)
+            else None
+        )
+        self.get_item = get_item
+        # make this a file we call for very large datasets
+        # TODO: remove duplication with cached_fields, just have
```
griffbr, line 1564: @jacobsela, if you are going to save this for a future PR, I recommend adding a link to a jira card for this TODO so it is documented and does not get lost.

griffbr:
@jacobsela, `ipython ./docs/source/recipes/torch-dataset-examples/basic_example.ipynb` leads to the following error:
```
Overwriting existing directory '/Users/b/fiftyone/quickstart'
Downloading dataset to '/Users/b/fiftyone/quickstart'
Downloading dataset...
 100% |███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████|  187.5Mb/187.5Mb [862.8ms elapsed, 0s remaining, 217.4Mb/s]      
Extracting dataset...
Parsing dataset metadata
Found 200 samples
Dataset info written to '/Users/b/fiftyone/quickstart/info.json'
Loading existing dataset 'quickstart'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
True
<class 'fiftyone.core.sample.Sample'>
67c09bebdf722eb263d147fc
/Users/b/fiftyone/quickstart/data/004534.jpg
67c09bebdf722eb263d147fc
File ~/Dropbox/git/fiftyone/fiftyone/core/odm/mixins.py:120, in DatasetMixin.get_field(self, field_name)
    119 try:
--> 120     return super().get_field(field_name)
    121 except AttributeError:

File ~/Dropbox/git/fiftyone/fiftyone/core/odm/document.py:323, in MongoEngineBaseDocument.get_field(self, field_name)
    321         return None
--> 323     raise AttributeError(
    324         "%s has no field '%s'" % (self.__class__.__name__, field_name)
    325     )
    327 return getattr(self, field_name)

AttributeError: samples.67c09beadf722eb263d147c3 has no field 'detections.detections'
```

Also, `ipython ./docs/source/recipes/torch-dataset-examples/simple_training_example.ipynb` almost works but then:
```
[TerminalIPythonApp] WARNING | File 'notebook' doesn't exist
(fo-src) b@ava fiftyone % ipython ./docs/source/recipes/torch-dataset-examples/simple_training_example.ipynb 
Split 'test' already downloaded
Loading 'mnist' split 'train'
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 60000/60000 [9.6s elapsed, 0s remaining, 6.4K samples/s]      
Loading 'mnist' split 'test'
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10000/10000 [1.6s elapsed, 0s remaining, 6.1K samples/s]         
Dataset 'mnist' created
App launched. Point your web browser to http://localhost:5151

Welcome to

███████╗██╗███████╗████████╗██╗   ██╗ ██████╗ ███╗   ██╗███████╗
██╔════╝██║██╔════╝╚══██╔══╝╚██╗ ██╔╝██╔═══██╗████╗  ██║██╔════╝
█████╗  ██║█████╗     ██║    ╚████╔╝ ██║   ██║██╔██╗ ██║█████╗
██╔══╝  ██║██╔══╝     ██║     ╚██╔╝  ██║   ██║██║╚██╗██║██╔══╝
██║     ██║██║        ██║      ██║   ╚██████╔╝██║ ╚████║███████╗
╚═╝     ╚═╝╚═╝        ╚═╝      ╚═╝    ╚═════╝ ╚═╝  ╚═══╝╚══════╝ v1.3.0

If you're finding FiftyOne helpful, here's how you can get involved:

|
|  ⭐⭐⭐ Give the project a star on GitHub ⭐⭐⭐
|  https://github.com/voxel51/fiftyone
|
|  🚀🚀🚀 Join the FiftyOne Discord community 🚀🚀🚀
|  https://community.voxel51.com/
|

{'test': 10000, 'validation': 6000, 'train': 54000}
AssertionError                            Traceback (most recent call last)
Cell In[1], line 1
----> 1 mnist_training.main(subset, 10, 10, 'cuda:1', '/home/jacobs/fiftyone/docs/source/recipes/torch-dataset-examples/mnist_weights')

File ~/Dropbox/git/fiftyone/docs/source/recipes/torch-dataset-examples/mnist_training.py:15, in main(dataset, num_classes, num_epochs, device, save_dir)
     13 global DEVICE
     14 DEVICE = torch.device(device)
---> 15 model = utils.setup_model(num_classes).to(DEVICE)
     16 loss_function = torch.nn.CrossEntropyLoss(reduction="none")
     17 dataloaders = utils.create_dataloaders(
     18     dataset,
     19     utils.mnist_get_item,
   (...)
     22     persistent_workers=True,
     23 )

File ~/fo-src/lib/python3.11/site-packages/torch/nn/modules/module.py:1343, in Module.to(self, *args, **kwargs)
   1340         else:
   1341             raise
-> 1343 return self._apply(convert)

File ~/fo-src/lib/python3.11/site-packages/torch/nn/modules/module.py:903, in Module._apply(self, fn, recurse)
    901 if recurse:
    902     for module in self.children():
--> 903         module._apply(fn)
    905 def compute_should_use_set_data(tensor, tensor_applied):
    906     if torch._has_compatible_shallow_copy_type(tensor, tensor_applied):
    907         # If the new tensor has compatible tensor type as the existing tensor,
    908         # the current behavior is to change the tensor in-place using `.data =`,
   (...)
    913         # global flag to let the user control whether they want the future
    914         # behavior of overwriting the existing tensor or not.

File ~/fo-src/lib/python3.11/site-packages/torch/nn/modules/module.py:930, in Module._apply(self, fn, recurse)
    926 # Tensors stored in modules are graph leaves, and we don't want to
    927 # track autograd history of `param_applied`, so we have to use
    928 # `with torch.no_grad():`
    929 with torch.no_grad():
--> 930     param_applied = fn(param)
    931 p_should_use_set_data = compute_should_use_set_data(param, param_applied)
    933 # subclasses may have multiple child tensors so we need to use swap_tensors

File ~/fo-src/lib/python3.11/site-packages/torch/nn/modules/module.py:1329, in Module.to.<locals>.convert(t)
   1322     if convert_to_format is not None and t.dim() in (4, 5):
   1323         return t.to(
   1324             device,
   1325             dtype if t.is_floating_point() or t.is_complex() else None,
   1326             non_blocking,
   1327             memory_format=convert_to_format,
   1328         )
-> 1329     return t.to(
   1330         device,
   1331         dtype if t.is_floating_point() or t.is_complex() else None,
   1332         non_blocking,
   1333     )
   1334 except NotImplementedError as e:
   1335     if str(e) == "Cannot copy out of meta tensor; no data!":

File ~/fo-src/lib/python3.11/site-packages/torch/cuda/__init__.py:310, in _lazy_init()
    305     raise RuntimeError(
    306         "Cannot re-initialize CUDA in forked subprocess. To use CUDA with "
    307         "multiprocessing, you must use the 'spawn' start method"
    308     )
    309 if not hasattr(torch._C, "_cuda_getDeviceCount"):
--> 310     raise AssertionError("Torch not compiled with CUDA enabled")
    311 if _cudart is None:
    312     raise AssertionError(
    313         "libcudart functions unavailable. It looks like you have a broken build?"
    314     )

AssertionError: Torch not compiled with CUDA enabled
```

Have you tested your examples work for `mps` or `cpu` devices?

fiftyone/utils/torch.py
jacobsela: good call, new backlog ticket: https://voxel51.atlassian.net/browse/ML-130?atlOrigin=eyJpIjoiMjZjMDIzOTM0Zjg2NDYwZWE2ODg1ZjBjMzNmMzkxN2UiLCJwIjoiaiJ9

manushreegangwar (dismissed):
Minor comments/ questions. Tested all the notebooks. LGTM.

fiftyone/utils/torch.py
```diff
@@ -2030,3 +2264,197 @@ def _load_image(image_path, use_numpy, force_rgb):
...
+    """
+    if dist.get_backend() == "nccl":
+        return dist.new_group(backend="gloo")
+    else:
+        return dist.group.WORLD
+
+
+# see https://github.com/ppwwyyxx/RAM-multiprocess-dataloader/issues/5
+def local_broadcast_process_authkey(local_process_group):
+    if int(os.environ["LOCAL_WORLD_SIZE"]) == 1:
```
kaixi-wang, line 2448: Typically this is added to/accessed through fo.config

jacobsela:
@kaixi-wang @brimoor leaving this here as well https://github.com/voxel51/fiftyone/issues/1302#issuecomment-2718692973

fiftyone/utils/torch.py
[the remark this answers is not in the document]
jacobsela: Until there is a dedicated place for these kinds of notes, I'm leaving this here.

fiftyone/utils/torch.py
```diff
@@ -1460,6 +1468,232 @@ def _is_string_array(targets):
...
+             Reads into the FO object during training may slow your code down significantly.
+        - Writing
+            - Writing currently happens on the process from which it is called. This
+            shows moderate slowdown, and will be adressed.
+
+    """
+
+    def __init__(
+        self,
+        samples: focol.SampleCollection,
```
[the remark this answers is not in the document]
jacobsela: It can be either a DatasetView or a Dataset, which is why I made the hint their common parents (SamplesCollection). I can change/remove the hint if it's misleading though.

fiftyone/utils/torch.py
jacobsela: Do you mean the authkey or the environment variable? Generally what this is here to address is that the way the user launches the training processes is through `torchrun`. By default, this sets these environment variables and this is how you can manage the behavior of the different processes. This happens entirely outside of FO and will NOT be based on the config variables like `max_process_pool_workers`.

minhtuev:
@jacobsela : we probably need a CUDA-enabled instance to test this PR right?

jacobsela:
@minhtuev Actually no! This PR just facilitates a way of loading model inputs from a fiftyone dataset. We do so by having the function `to_torch` that turns a fiftyone dataset into a `torch.utils.data.Dataset`. Reductively, all it does is create an object that when subscripted calls a function `get_item` passed in by the user on the sample at that index passed.

In fact the following code is equivalent to running `func` on the dataset in parallel with 4 workers each processing a batch of 16 samples at a time:

```python
from tqdm import tqdm
from torch.utils.data import DataLoader
import fiftyone as fo
from fiftyone.utils.torch import FiftyOneTorchDataset

def func(sample):
    sample["new_field"] = sample["id"]
    sample.save()
    return None
def dummy_collate(batch):
    return None

torch_dataset = dataset.to_torch(get_item=func)
dataloader = DataLoader(
    torch_dataset,
    batch_size=16,
    drop_last=False,
    num_workers=4,
    collate_fn=dummy_collate,
    worker_init_fn=FiftyOneTorchDataset.worker_init
)
for _ in tqdm(dataloader, total=len(dataloader)):
    continue
for sample in dataset:
    assert sample["new_field"] == sample["id"]
print("Success")
```

By passing a different function to `get_item`, e.g.

```python
def classification_get_item(sample):
    return sample["filepath"], sample["ground_truth.label"]
```

We can efficiently load model inputs for training and have high throughput.

minhtuev:
I am running into this error:
```
File /opt/anaconda3/envs/voxel311/lib/python3.11/site-packages/torch/utils/data/dataloader.py:701, in _BaseDataLoaderIter.__next__(self)
    698 if self._sampler_iter is None:
    699     # TODO(https://github.com/pytorch/pytorch/issues/76750)
    700     self._reset()  # type: ignore[call-arg]
--> 701 data = self._next_data()
    702 self._num_yielded += 1
    703 if (
    704     self._dataset_kind == _DatasetKind.Iterable
    705     and self._IterableDataset_len_called is not None
    706     and self._num_yielded > self._IterableDataset_len_called
    707 ):

File /opt/anaconda3/envs/voxel311/lib/python3.11/site-packages/torch/utils/data/dataloader.py:1448, in _MultiProcessingDataLoaderIter._next_data(self)
   1445     return self._process_data(data)
   1447 assert not self._shutdown and self._tasks_outstanding > 0
-> 1448 idx, data = self._get_data()
   1449 self._tasks_outstanding -= 1
   1450 if self._dataset_kind == _DatasetKind.Iterable:
   1451     # Check for _IterableDatasetStopIteration

File /opt/anaconda3/envs/voxel311/lib/python3.11/site-packages/torch/utils/data/dataloader.py:1412, in _MultiProcessingDataLoaderIter._get_data(self)
   1408     # In this case, `self._data_queue` is a `queue.Queue`,. But we don't
   1409     # need to call `.task_done()` because we don't use `.join()`.
   1410 else:
   1411     while True:
-> 1412         success, data = self._try_get_data()
   1413         if success:
   1414             return data

File /opt/anaconda3/envs/voxel311/lib/python3.11/site-packages/torch/utils/data/dataloader.py:1256, in _MultiProcessingDataLoaderIter._try_get_data(self, timeout)
   1254 if len(failed_workers) > 0:
   1255     pids_str = ", ".join(str(w.pid) for w in failed_workers)
-> 1256     raise RuntimeError(
   1257         f"DataLoader worker (pid(s) {pids_str}) exited unexpectedly"
   1258     ) from e
   1259 if isinstance(e, queue.Empty):
   1260     return (False, None)

RuntimeError: DataLoader worker (pid(s) 90889, 90890, 90891, 90892) exited unexpectedly
```

fiftyone/utils/torch.py
kaixi-wang: I was referring to the environment variable, but didn't know about the torchrun assumption... I need to take some time to read the doc strings, but without full context, it seems like an antipattern and introduces implicit dependencies

kaixi-wang (commented):
Perhaps this deserves its own file as an integration (eg cvat) as it sounds like this code needs to be run in a specific way and a lot of the functions are only applicable under those run conditions. Mixing it in with the other torch utils may lead to people attempting to use it in the wrong context or trying to do other fiftyone things with it

jacobsela:
@minhtuev 
> ```
> AttributeError: Can't get attribute 'func' on <module '__main__' (built-in)>
> ```

This is most likely a pickling issue. Make sure that `func` is top level (i.e. outside of '__main__' or some function). Can you send me your code so that I can confirm?

fiftyone/utils/torch.py
jacobsela: That's fair. I'm not sure it's explicitly documented but the idea is that this will be used within a standard torch (possibly distributed) train run, and that if it is distributed it has been launched with `torchrun`. Regardless now that I'm looking at this code it's inconsistent with the rest of the repo (which uses the `get_local_rank` and other utils), so I'll rewrite it.

minhtuev (dismissed):
Did some testing with @jacobsela, discovered that there are some incompatibility with running Pytorch parallel processing inside Jupyter notebook/iPython on MacOS.

We also tested and verified that regular pool processing works fine, so this is just a problem with this library.

jacobsela:
@minhtuev I've been doing some reading and something isn't adding up. Yesterday, we tested the following snippet and it fails within an interactive interpreter when using spawn, but works with fork (I confirmed this on linux after you ran it on mac):
```python
from tqdm import tqdm
from torch.utils.data import DataLoader
import fiftyone as fo
from fiftyone.utils.torch import FiftyOneTorchDataset

import torch.multiprocessing as mp
mp.set_start_method("spawn", force=True) # or mp.set_start_method("fork", force=True)

def func(sample):
    sample["new_field"] = sample["id"]
    sample.save()
    return None
def dummy_collate(batch):
    return None

torch_dataset = dataset.to_torch(get_item=func)
dataloader = DataLoader(
    torch_dataset,
    batch_size=16,
    drop_last=False,
    num_workers=4,
    collate_fn=dummy_collate,
    worker_init_fn=FiftyOneTorchDataset.worker_init
)
for _ in tqdm(dataloader, total=len(dataloader)):
    continue
for sample in dataset:
    assert sample["new_field"] == sample["id"]
print("Success")
```
With the error: `AttributeError: Can't get attribute 'func' on <module '__main__' (built-in)>`.

The thing is, this is *expected* behavior according to the [official docs](https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods). In fact, it is explicitly mentioned in the note [here](https://docs.python.org/3/library/multiprocessing.html#using-a-pool-of-workers).

This means that when this works when using `map_samples`, something inherently different is happening.

I'm not sure if this is a bug or a feature.

manushreegangwar:
Tested on fiftyone teams using https://github.com/voxel51/machine_learning/pull/2

manushreegangwar (approved):

jacobsela:
With confirmation that this works on Teams, I'm merging with the understanding that there may be some features / use cases that we don't fully support yet. We can address those piecemeal over time.

Current state + future work noted in [this ticket](https://voxel51.atlassian.net/browse/ML-130?atlOrigin=eyJpIjoiZTI2NWNmOWRkODk5NDgxNDk3YjZhYjQxZDk5NDQzMWEiLCJwIjoiaiJ9).

PR #5323, opened 2024-12-27
Adding label IDs in Segmentation Evaluation results

prernadh:

Adding Segmentation Ids in Segementation Evaluation results fixes the click through behavior when you attach plots to sessions in FiftyOne Notebooks.

Manually tested

        notes for FiftyOne users.

Adds in click-through capabilities for segmentation evaluation confusion matrices.

-   [ ] Other

Code to test- 
```import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset('quickstart',max_samples=3)

# Generate some toy semantic segmentations
for idx, sample in enumerate(dataset.iter_samples(autosave=True)):
    mask = idx*np.ones((10,10),dtype=np.uint8)
    seg = fo.Segmentation(mask=mask)
    sample['segmentation'] = seg

    mask = (idx+1)*np.ones((10,10),dtype=np.uint8)
    seg = fo.Segmentation(mask=mask)
    sample['segmentation2'] = seg

res = dataset.evaluate_segmentations(
    "segmentation2",
    gt_field="segmentation",
    eval_key="eval_seg"
)

plot = res.plot_confusion_matrix()
plot.show()

session.plots.attach(plot)

# Click on dark cells of confusion matrix. Expected clicking on eg (r,c) = (0,1) to select the first sample in the view.```

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

- **New Features**
  - Enhanced segmentation evaluation with more detailed performance insights.
  - Added capability to track pixel-level classification matches across samples.

- **Improvements**
  - Expanded evaluation results to provide more granular information about segmentation performance.
  - Updated evaluation process to include detailed pixel-level match data in results reporting.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/fiftyone/utils/eval/segmentation.py
+++ b/fiftyone/utils/eval/segmentation.py
@@ -391,9 +391,9 @@ class SimpleEvaluation(SegmentationEvaluation):
                     bandwidth=bandwidth,
                 )
                 sample_conf_mat += image_conf_mat
-                non_zero_indexes = np.nonzero(sample_conf_mat)
+                non_zero_indexes = np.nonzero(image_conf_mat)
                 for index in zip(*non_zero_indexes):
-                    matches.append((classes[index[0]], classes[index[1]], sample_conf_mat[index[0], index[1]], gt_seg.id, pred_seg.id))
+                    matches.append((classes[index[0]], classes[index[1]], image_conf_mat[index[0], index[1]], gt_seg.id, pred_seg.id))
 
                 if processing_frames and save:
                     facc, fpre, frec = _compute_accuracy_precision_recall(
--- a/fiftyone/utils/eval/segmentation.py
+++ b/fiftyone/utils/eval/segmentation.py
@@ -423,7 +423,8 @@ class SimpleEvaluation(SegmentationEvaluation):
             missing = classes[0] if values[0] in (0, "#000000") else None
         else:
             missing = None
-
+        if len(matches) == 0:
+            matches = None
         return SegmentationResults(
             samples,
             self.config,
--- a/fiftyone/utils/eval/segmentation.py
+++ b/fiftyone/utils/eval/segmentation.py
@@ -390,11 +390,13 @@ class SimpleEvaluation(SegmentationEvaluation):
                     values,
                     bandwidth=bandwidth,
                 )
-                sample_conf_mat += image_conf_mat
+                
                 non_zero_indexes = np.nonzero(image_conf_mat)
                 for index in zip(*non_zero_indexes):
                     matches.append((classes[index[0]], classes[index[1]], image_conf_mat[index[0], index[1]], gt_seg.id, pred_seg.id))
 
+                sample_conf_mat += image_conf_mat
+                
                 if processing_frames and save:
                     facc, fpre, frec = _compute_accuracy_precision_recall(
                         image_conf_mat, values, average
--- a/fiftyone/utils/eval/segmentation.py
+++ b/fiftyone/utils/eval/segmentation.py
@@ -509,7 +509,7 @@ class SegmentationResults(BaseEvaluationResults):
         )
 
     @staticmethod
-    def _parse_confusion_matrix(confusion_matrix, classes, ytrue_ids_dict, ypred_ids_dict):
+    def _parse_confusion_matrix(confusion_matrix, classes):
         ytrue = []
         ypred = []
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: fiftyone/utils/eval/segmentation.py [hunk 1]
```diff
@@ -495,21 +513,34 @@ def _from_dict(cls, d, samples, config, eval_key, **kwargs):
...
+                        ytrue_ids+= ytrue_ids_dict[index]
+                        ypred_ids+= ypred_ids_dict[index]
+                        ytrue_multiplier = len(ytrue_ids_dict[index])
+                    else:
+                        ytrue_multiplier = 1
+                    for p in range(ytrue_multiplier):
+                        ytrue.append(classes[i])
+                        ypred.append(classes[j])
+                        weights.append(cij/ytrue_multiplier)
+                    #Note: The weights aren't equally divided by the different ytrue_ids, but it works out for the confusion matrix calculations.
```
allenleetc, line 541: Agree, ideally the weights would be the actual weights/counts from each pair of (ytrue_ids, ypred_ids). It seems to work okay as-is but long-run there may be other ramifications. For instance, one idea, if I click on a misclassification (off-diagonal) cell, ideally I would see the resulting samples sorted in descending order of weight, with the first sample containing the most misclassified pixels of that type.

fiftyone/utils/eval/segmentation.py
```diff
@@ -495,21 +513,34 @@ def _from_dict(cls, d, samples, config, eval_key, **kwargs):
...
+                index = (i, j)
                 cij = confusion_matrix[i, j]
                 if cij > 0:
-                    ytrue.append(classes[i])
-                    ypred.append(classes[j])
-                    weights.append(cij)
-
-        return ytrue, ypred, weights
-
+                    if ytrue_ids_dict is not None and ypred_ids_dict is not None:
```
allenleetc, line 531: Minor, maybe create a bool flag for this as this is re-used above in line 522 and also in every cell in this doubly-nested loop

fiftyone/utils/eval/segmentation.py
```diff
@@ -495,7 +509,7 @@ def _from_dict(cls, d, samples, config, eval_key, **kwargs):
         )
 
     @staticmethod
-    def _parse_confusion_matrix(confusion_matrix, classes):
+    def _parse_confusion_matrix(confusion_matrix, classes, ytrue_ids_dict, ypred_ids_dict):
```
kaixi-wang, line 512: What's the purpose of this? You don't seem to use the new args in the method.

prernadh: Fixed it, those were remaining from my previous commit.

fiftyone/utils/eval/segmentation.py
prernadh: Fixed and updated in latests commit

fiftyone/utils/eval/segmentation.py
prernadh: Fixed in latest commit

fiftyone/utils/eval/segmentation.py
```diff
@@ -449,13 +458,20 @@ def __init__(
...
         backend=None,
     ):
         pixel_confusion_matrix = np.asarray(pixel_confusion_matrix)
-        ytrue, ypred, weights = self._parse_confusion_matrix(
-            pixel_confusion_matrix, classes
-        )
+        ytrue_ids = None
+        ypred_ids = None
+
+        if matches is not None:
```
brimoor, line 469: @prernadh mmm this an appealing way to store the data because it means `_parse_confusion_matrix` doesn't need to be called when loading results 🧠 

Prior to this refactor and before you created https://github.com/voxel51/fiftyone/pull/5330, I had forked your initial implementation with dicts and implemented Model Evaluation panel callbacks in https://github.com/voxel51/fiftyone/pull/5331.

As I mentioned on that PR, there's a few implementation details we probably aught to benchmark to make sure the panel is as performant as possible, but I could definitely see us using the data model here 👍

prernadh: Sounds good, what were you thinking in terms of benchmarking @brimoor ?

prernadh:
Subsumed by #5332

PR #4614, opened 2024-08-04
Bugfix: support branches that contain slashes when downloading plugins

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/4610

## Tested by

Verifying that we can download a plugin from a branch that contains a slash:

```shell
export FIFTYONE_PLUGINS_DIR=/tmp/plugins

# Download a plugin from branch `feature/update-index-plugin` that contains a slash
fiftyone plugins download \
    https://github.com/voxel51/fiftyone-plugins/tree/feature/update-index-plugin \
    --plugin-names '@voxel51/indexes'

rm -rf /tmp/plugins
unset FIFTYONE_PLUGINS_DIR
```

Also verified that all relevant syntaxes in `fiftyone.plugins.utils` work as expected:

```py
import fiftyone.plugins.utils as fopu

plugins = fopu.list_zoo_plugins()
plugins = fopu.list_zoo_plugins(info=True)

plugins = fopu.find_plugins("https://github.com/voxel51/fiftyone-plugins")
plugins = fopu.find_plugins("https://github.com/voxel51/fiftyone-plugins/tree/main", path="plugins/annotation")

info = fopu.get_plugin_info("https://github.com/voxel51/voxelgpt")
info = fopu.get_plugin_info("https://github.com/voxel51/fiftyone-plugins/tree/main/plugins/annotation")
info = fopu.get_plugin_info("https://github.com/voxel51/fiftyone-plugins/blob/main/plugins/annotation/fiftyone.yml")
info = fopu.get_plugin_info("voxel51/fiftyone-plugins", path="plugins/annotation")
info = fopu.get_plugin_info("voxel51/fiftyone-plugins", path="plugins/annotation/fiftyone.yml")
```

## Summary by CodeRabbit

- **New Features**
	- Enhanced plugin search capabilities with the ability to specify subdirectories in GitHub repositories.
	- Improved access to plugin information by allowing path specifications for YAML files or directories.

- **Bug Fixes**
	- Refined GitHub URL parsing logic to support a broader range of valid URLs.

- **Tests**
	- Added a new test for verifying GitHub repository URL parsing, improving test coverage.

- **Documentation**
	- Updated documentation for functions and classes to enhance clarity on expected inputs and usage.

PR #5324, opened 2024-12-27
Handle tree paths in GitHub identifiers

brimoor:
## Change log

The [@voxel51/plugins/install_plugin](https://github.com/voxel51/fiftyone-plugins/tree/main/plugins/plugins) operator stopped working as a side effect of https://github.com/voxel51/fiftyone/pull/4614 because the operator tried to issue a command like this:

```py
import fiftyone.plugins as fop

url = "https://github.com/voxel51/fiftyone-plugins/tree/main/plugins/dashboard"
fop.download_plugin(url, plugin_names=["@voxel51/dashboard"])
```

where the `url` is pulled [directly from here](https://github.com/voxel51/fiftyone-plugins/blob/d458e19bf9141bb8cd68c0d382cda98fbb2e1169/plugins/dashboard/fiftyone.yml#L6).

When the operator was initially written, the above `download_plugin()` syntax worked, but only because the `GitHubRepository` class would automatically strip `main/plugins/dashboard` down to `main`. However, that was causing our plugins API to not support branches/tags that contain `/` (what https://github.com/voxel51/fiftyone/pull/4614 solved).

This PR brings us back to the best of both worlds: `GitHubRepository(repo)` now supports `repo` strings of the form `<ref>[/<path>]`, where both `<ref>` and `<path>` may contain `/`.

With this patch, the [@voxel51/plugins/install_plugin](https://github.com/voxel51/fiftyone-plugins/tree/main/plugins/plugins) operator works again.

And, as an added bonus, the `download_plugin()` method now supports scoping downloads by subdirectory!

```py
fop.download_plugin("https://github.com/voxel51/fiftyone-plugins/tree/main/plugins/dashboard")
```

Previously the only way to install a subset of plugins from a repository was to specify them by name via the `plugin_names=` kwarg.

### Search for plugins in a specific tree root

```py
import fiftyone.plugins.utils as fopu

plugins = fopu.find_plugins("https://github.com/voxel51/fiftyone-plugins/tree/main/plugins/annotation")
assert len(plugins) == 1
```

### Download plugin from a specific tree root

```py
import fiftyone.plugins as fop

fop.download_plugin("https://github.com/voxel51/fiftyone-plugins/tree/main/plugins/dashboard")
```

### New `GitHubRepository(..., safe=True)` syntax

This new syntax is what enables the other syntaxes to work under the hood:

```py
import fiftyone.utils.github as foug

url = "https://github.com/voxel51/fiftyone-plugins/tree/main/plugins/dashboard"

repo = foug.GitHubRepository(url)
assert repo.ref == "main/plugins/dashboard"
assert repo.safe_path == None

repo = foug.GitHubRepository(url, safe=True)
assert repo.ref == "main"
assert repo.safe_path == "plugins/dashboard"

url = "https://github.com/voxel51/fiftyone-plugins/tree/f5e1d93233caace681b427d7e13677cbcfe58a06/plugins/dashboard"

repo = foug.GitHubRepository(url, safe=True)
assert repo.ref == "f5e1d93233caace681b427d7e13677cbcfe58a06"
assert repo.safe_path == "plugins/dashboard"
```

### DIFF
--- a/fiftyone/plugins/core.py
+++ b/fiftyone/plugins/core.py
@@ -22,6 +22,7 @@ import eta.core.web as etaw
 import fiftyone as fo
 import fiftyone.constants as foc
 from fiftyone.core.config import locate_app_config
+import fiftyone.core.storage as fos
 import fiftyone.core.utils as fou
 from fiftyone.plugins.definitions import PluginDefinition
 from fiftyone.utils.github import GitHubRepository
@@ -172,6 +173,8 @@ def download_plugin(url_or_gh_repo, plugin_names=None, overwrite=False):
                 ``https://github.com/<user>/<repo>/tree/<branch>`` or
                 ``https://github.com/<user>/<repo>/commit/<commit>``
             -   a GitHub ref string like ``<user>/<repo>[/<ref>]``
+            -   a GitHub tree path like
+                ``https://github.com/<user>/<repo>/tree/<branch>/<path>``
             -   a publicly accessible URL of an archive (eg zip or tar) file
 
         plugin_names (None): a plugin name or iterable of plugin names to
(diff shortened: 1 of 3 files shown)


kaixi-wang (approved):
nice! 
image

PR #5325, opened 2024-12-28
Add `create_index` kwarg to geo stages

brimoor:
Adds a `create_index` kwarg to `GeoNear` and `GeoWithin` stages, for consistency with the `SortBy` and `GroupBy` stages.

Note that, unlike sorting/grouping, the geo stages **require** the spherical index to exist in order to function. However, it feels appropriate to provide a `create_index=False` syntax to provide users a way to guarantee that nothing expensive like index creation will occur when creating/using a view.

### DIFF
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -3017,14 +3017,20 @@ def _extract_filter_field(val):
 
 
 class _GeoStage(ViewStage):
-    def __init__(self, location_field):
+    def __init__(self, location_field=None, create_index=True):
         self._location_field = location_field
         self._location_key = None
+        self._create_index = create_index
 
     @property
     def location_field(self):
         """The location field."""
         return self._location_field
+
+    @property
+    def create_index(self):
+        """Whether to create the required spherical index, if necessary."""
+        return self._create_index
 
     def validate(self, sample_collection):
         if self._location_field is None:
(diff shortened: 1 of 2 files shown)


fiftyone/core/stages.py
```diff
@@ -3039,8 +3045,9 @@ def validate(self, sample_collection):
             # Assume the user directly specified the subfield to use
             self._location_key = self._location_field
 
-        # These operations require a spherical index
-        sample_collection.create_index([(self._location_key, "2dsphere")])
+        if self._create_index:
+            # These operations require a spherical index
```
minhtuev, line 3049: Should we add a logging statement here for creating index?

fiftyone/core/stages.py
```diff
@@ -3225,6 +3239,12 @@ def _params(cls):
                 "placeholder": "",
                 "default": "None",
             },
+            {
```
minhtuev, line 3242: Just a curious question, why do we return this json info list for the arguments & params? Who are the endpoint consumers of these? :)

minhtuev (approved):
LGTM 👍 🚀 rocke

fiftyone/core/stages.py
brimoor: `_params()` is used by the App to render the view stages in the view bar

fiftyone/core/stages.py
brimoor: Definitely a reasonable suggestion. This module doesn't have a convention of logging things so I'll punt on it now for consistency

PR #5331, opened 2024-12-31
Add Model Evaluation panel callbacks for segmentation tasks

brimoor:
## Example usage

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=10, shuffle=True)

dataset.apply_model(model, "resnet101")

CLASSES = (
    "tvmonitor"
)
classes = CLASSES.split(",")
mask_targets = {idx: label for idx, label in enumerate(classes)}

rgb_mask_targets = {
    fof.int_to_hex(random.getrandbits(64)): label
    for label in CLASSES.split(",")
}

_mask_targets = {v: k for k, v in mask_targets.items()}
_rgb_mask_targets = {v: k for k, v in rgb_mask_targets.items()}
targets_map = {_mask_targets[c]: _rgb_mask_targets[c] for c in classes}

dataset.clone_sample_field("resnet101", "resnet101_rgb")
foul.transform_segmentations(
    dataset,
    "resnet101_rgb",
    targets_map,
)

dataset.clone_sample_field("resnet50", "resnet50_rgb")
foul.transform_segmentations(
    dataset,
    "resnet50_rgb",
    targets_map,
)

dataset.mask_targets["resnet101"] = mask_targets
dataset.mask_targets["resnet50"] = mask_targets
dataset.mask_targets["resnet101_rgb"] = rgb_mask_targets
dataset.mask_targets["resnet50_rgb"] = rgb_mask_targets
dataset.save()

# Evaluation with int mask targets
dataset.evaluate_segmentations(
    "resnet50",
    gt_field="resnet101",
    eval_key="eval",
)

# Evaluation with RGB mask targets
dataset.evaluate_segmentations(
    "resnet50_rgb",
    gt_field="resnet101_rgb",
    eval_key="eval_rgb",
    mask_targets=dataset.mask_targets["resnet50_rgb"],
)

sesssion = fo.launch_app(dataset)
```

### DIFF
--- a/fiftyone/utils/eval/segmentation.py
+++ b/fiftyone/utils/eval/segmentation.py
@@ -333,7 +333,7 @@ class SimpleEvaluation(SegmentationEvaluation):
         if mask_targets is not None:
             if fof.is_rgb_mask_targets(mask_targets):
                 mask_targets = {
-                    _hex_to_int(k): v for k, v in mask_targets.items()
+                    fof.hex_to_int(k): v for k, v in mask_targets.items()
                 }
 
             values, classes = zip(*sorted(mask_targets.items()))
@@ -349,8 +349,9 @@ class SimpleEvaluation(SegmentationEvaluation):
 
         nc = len(values)
         confusion_matrix = np.zeros((nc, nc), dtype=int)
-        ypred_ids = {}
-        ytrue_ids = {}
+        weights_dict = {}
+        ytrue_ids_dict = {}
+        ypred_ids_dict = {}
 
         bandwidth = self.config.bandwidth
         average = self.config.average
@@ -392,16 +393,19 @@ class SimpleEvaluation(SegmentationEvaluation):
                     bandwidth=bandwidth,
                 )
                 sample_conf_mat += image_conf_mat
-                non_zero_indexes = np.nonzero(sample_conf_mat)
-                for index in zip(*non_zero_indexes):
-                    if index not in ypred_ids:
-                        ypred_ids[index] = [pred_seg.id]
-                    else:
-                        ypred_ids[index].append(pred_seg.id)
-                    if index not in ytrue_ids:
-                        ytrue_ids[index] = [gt_seg.id]
-                    else:
-                        ytrue_ids[index].append(gt_seg.id)
+
+                for index in zip(*np.nonzero(image_conf_mat)):
+                    if index not in weights_dict:
+                        weights_dict[index] = []
+                    weights_dict[index].append(int(image_conf_mat[index]))
+
+                    if index not in ytrue_ids_dict:
+                        ytrue_ids_dict[index] = []
+                    ytrue_ids_dict[index].append(gt_seg.id)
+
+                    if index not in ypred_ids_dict:
+                        ypred_ids_dict[index] = []
+                    ypred_ids_dict[index].append(pred_seg.id)
 
                 if processing_frames and save:
                     facc, fpre, frec = _compute_accuracy_precision_recall(
@@ -436,8 +440,9 @@ class SimpleEvaluation(SegmentationEvaluation):
             eval_key,
             confusion_matrix,
             classes,
-            ypred_ids=ypred_ids,
-            ytrue_ids=ytrue_ids,
+            weights_dict=weights_dict,
+            ytrue_ids_dict=ytrue_ids_dict,
+            ypred_ids_dict=ypred_ids_dict,
             missing=missing,
             backend=self,
         )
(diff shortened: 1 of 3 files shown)


fiftyone/utils/eval/segmentation.py
```diff
@@ -490,25 +541,64 @@ def _from_dict(cls, d, samples, config, eval_key, **kwargs):
...
+                        ypred.append(classes[j])
+                        weights.append(cij)
+
+        return ytrue, ypred, weights, ytrue_ids, ypred_ids
+
+
+def _parse_index_dict(d):
+    import ast
+
+    return {ast.literal_eval(k): v for k, v in d.items()}
```
brimoor, line 601: This is probably slow; may want to consider other data models here

plugins/panels/model_evaluation.py
```diff
@@ -585,6 +610,89 @@ def load_view(self, ctx):
                     view = eval_view.filter_labels(
                         pred_field, F(eval_key) == field, only_matches=True
                     )
+        elif info.config.type == "segmentation":
+            results = ctx.dataset.load_evaluation_results(eval_key)
```
brimoor, line 614: `Dataset` objects do cache evaluation results, so if `load_evaluation_results()` is called multiple times on the same `dataset` singleton instance, the results only have to be loaded from the database once. However, I don't think we can count on multiple API requests re-using the same dataset object...

An alternative would be to store the necessary data as panel state. However that is also not desirable because the necessary data here is potentially huge lists of IDs, and panel state is serialized with every panel callback, not just `load_view()`.

So, the best pattern for this case needs some though.

plugins/panels/model_evaluation.py
```diff
@@ -585,6 +610,89 @@ def load_view(self, ctx):
...
+                _, gt_id2 = ctx.dataset._get_label_field_path(gt_field2, "_id")
+            if pred_field2 is not None:
+                _, pred_id2 = ctx.dataset._get_label_field_path(
+                    pred_field2, "_id"
+                )
+
+            if view_type == "class":
+                # All GT/predictions that contain class `x`
+                k = classes_map[x]
+                ytrue_ids, ypred_ids = _get_ids_slice(results, k)
```
brimoor, line 652: These are the callbacks that fire when a user clicks on a cell in the `Class performance` section. The current data model requires building up a list of IDs for entire row/column slices of the ID dicts. We could potentially optimize the data model for this case. Needs some more thought.

plugins/panels/model_evaluation.py
```diff
@@ -585,6 +610,89 @@ def load_view(self, ctx):
...
+                    expr |= F(gt_id2).is_in(ytrue_ids2)
+                    expr |= F(pred_id2).is_in(ypred_ids2)
+
+                view = eval_view.match(expr)
+            elif view_type == "matrix":
+                # Specific confusion matrix cell
+                i = classes_map[x]
+                j = classes_map[y]
+                ytrue_ids = _to_object_ids(
+                    results.ytrue_ids_dict.get((i, j), [])
```
brimoor, line 667: Confusion matrix callbacks require the IDs for a particular `(i, j)` entry; so the current data model is roughly optimal for this case.

plugins/panels/model_evaluation.py
```diff
@@ -585,6 +610,89 @@ def load_view(self, ctx):
...
+                ypred_ids = _to_object_ids(
+                    results.ypred_ids_dict.get((i, j), [])
+                )
+                expr = F(gt_id).is_in(ytrue_ids)
+                expr &= F(pred_id).is_in(ypred_ids)
+                view = eval_view.match(expr)
+            elif view_type == "field":
+                if field == "tp":
+                    # All true positives
+                    inds = results.ytrue == results.ypred
```
brimoor, line 678: Segmentation tasks do not currently show TP/FP/FN counts in the `Metrics performance section`. But if we were to add them, then this implements the callback for the "grid" icons that would appear.

fiftyone/utils/eval/segmentation.py
```diff
@@ -490,25 +541,64 @@ def _from_dict(cls, d, samples, config, eval_key, **kwargs):
...
     @staticmethod
-    def _parse_confusion_matrix(confusion_matrix, classes):
+    def _parse_confusion_matrix(
+        confusion_matrix,
+        classes,
+        weights_dict=None,
+        ytrue_ids_dict=None,
+        ypred_ids_dict=None,
+    ):
+        have_ids = ytrue_ids_dict is not None and ypred_ids_dict is not None
```
brimoor, line 559: This is a non-trivial amount of data transformation, which, as currently implemented, must be performed every time a `SegmentationResults` object is loaded. May want to consider optimizing the data model here for speed.

brimoor:
Closing in favor of https://github.com/voxel51/fiftyone/pull/5332.

PR #5332, opened 2025-01-01
Add Model Evaluation panel callbacks for segmentation tasks

brimoor:
## Change log

- Adds class/confusion matrix callbacks when evaluating segmentations

## TODO

- [x] Benchmark performance of using label IDs to implement `load_view()`
- [x] Investigate alternatives that don't require calling `load_evaluation_results()`? [cf this comment](https://github.com/voxel51/fiftyone/pull/5331#discussion_r1900245655)
- [ ] Figure out a way to show TP/FP/FN data in the ME panel?
- [ ] Figure out a way to support TP/FP/FN filtering in the App sidebar? 

## Example usage

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=10, shuffle=True)

dataset.apply_model(model, "resnet101")

CLASSES = (
    "tvmonitor"
)
classes = CLASSES.split(",")
mask_targets = {idx: label for idx, label in enumerate(classes)}

rgb_mask_targets = {
    fof.int_to_hex(random.getrandbits(64)): label
    for label in CLASSES.split(",")
}

_mask_targets = {v: k for k, v in mask_targets.items()}
_rgb_mask_targets = {v: k for k, v in rgb_mask_targets.items()}
targets_map = {_mask_targets[c]: _rgb_mask_targets[c] for c in classes}

dataset.clone_sample_field("resnet101", "resnet101_rgb")
foul.transform_segmentations(
    dataset,
    "resnet101_rgb",
    targets_map,
)

dataset.clone_sample_field("resnet50", "resnet50_rgb")
foul.transform_segmentations(
    dataset,
    "resnet50_rgb",
    targets_map,
)

dataset.mask_targets["resnet101"] = mask_targets
dataset.mask_targets["resnet50"] = mask_targets
dataset.mask_targets["resnet101_rgb"] = rgb_mask_targets
dataset.mask_targets["resnet50_rgb"] = rgb_mask_targets
dataset.save()

# Evaluation with int mask targets
dataset.evaluate_segmentations(
    "resnet50",
    gt_field="resnet101",
    eval_key="eval",
)

# Evaluation with RGB mask targets
dataset.evaluate_segmentations(
    "resnet50_rgb",
    gt_field="resnet101_rgb",
    eval_key="eval_rgb",
    mask_targets=dataset.mask_targets["resnet50_rgb"],
)

sesssion = fo.launch_app(dataset)
```

## Release Notes

- **New Features**
  - Added color conversion utilities for hex and integer representations
  - Enhanced segmentation evaluation with custom metrics support
  - Introduced dataset caching mechanism
  - Improved model evaluation panel functionality

- **Improvements**
  - Refined handling of segmentation results
  - Added more robust error handling for custom metrics
  - Streamlined segmentation evaluation processes

- **Technical Enhancements**
  - Updated utility functions for more flexible data processing
  - Improved compatibility between different evaluation components

### DIFF
--- a/fiftyone/utils/eval/segmentation.py
+++ b/fiftyone/utils/eval/segmentation.py
@@ -369,7 +369,7 @@ class SimpleEvaluation(SegmentationEvaluation):
         if mask_targets is not None:
             if fof.is_rgb_mask_targets(mask_targets):
                 mask_targets = {
-                    _hex_to_int(k): v for k, v in mask_targets.items()
+                    fof.hex_to_int(k): v for k, v in mask_targets.items()
                 }
 
             values, classes = zip(*sorted(mask_targets.items()))
@@ -385,8 +385,9 @@ class SimpleEvaluation(SegmentationEvaluation):
 
         nc = len(values)
         confusion_matrix = np.zeros((nc, nc), dtype=int)
-        ypred_ids = {}
-        ytrue_ids = {}
+        weights_dict = {}
+        ytrue_ids_dict = {}
+        ypred_ids_dict = {}
 
         bandwidth = self.config.bandwidth
         average = self.config.average
@@ -428,16 +429,19 @@ class SimpleEvaluation(SegmentationEvaluation):
                     bandwidth=bandwidth,
                 )
                 sample_conf_mat += image_conf_mat
-                non_zero_indexes = np.nonzero(sample_conf_mat)
-                for index in zip(*non_zero_indexes):
-                    if index not in ypred_ids:
-                        ypred_ids[index] = [pred_seg.id]
-                    else:
-                        ypred_ids[index].append(pred_seg.id)
-                    if index not in ytrue_ids:
-                        ytrue_ids[index] = [gt_seg.id]
-                    else:
-                        ytrue_ids[index].append(gt_seg.id)
+
+                for index in zip(*np.nonzero(image_conf_mat)):
+                    if index not in weights_dict:
+                        weights_dict[index] = []
+                    weights_dict[index].append(int(image_conf_mat[index]))
+
+                    if index not in ytrue_ids_dict:
+                        ytrue_ids_dict[index] = []
+                    ytrue_ids_dict[index].append(gt_seg.id)
+
+                    if index not in ypred_ids_dict:
+                        ypred_ids_dict[index] = []
+                    ypred_ids_dict[index].append(pred_seg.id)
 
                 if processing_frames and save:
                     facc, fpre, frec = _compute_accuracy_precision_recall(
@@ -472,8 +476,9 @@ class SimpleEvaluation(SegmentationEvaluation):
             eval_key,
             confusion_matrix,
             classes,
-            ypred_ids=ypred_ids,
-            ytrue_ids=ytrue_ids,
+            weights_dict=weights_dict,
+            ytrue_ids_dict=ytrue_ids_dict,
+            ypred_ids_dict=ypred_ids_dict,
             missing=missing,
             backend=self,
         )
@@ -488,6 +493,11 @@ class SegmentationResults(BaseClassificationResults):
         eval_key: the evaluation key
         pixel_confusion_matrix: a pixel value confusion matrix
         classes: a list of class labels corresponding to the confusion matrix
+        weights_dict (None): a dict mapping ``(i, j)`` tuples to pixel counts
+        ytrue_ids_dict (None): a dict mapping ``(i, j)`` tuples to lists of
+            ground truth IDs
+        ypred_ids_dict (None): a dict mapping ``(i, j)`` tuples to lists of
+            predicted label IDs
         missing (None): a missing (background) class
         custom_metrics (None): an optional dict of custom metrics
         backend (None): a :class:`SegmentationEvaluation` backend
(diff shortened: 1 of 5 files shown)


fiftyone/utils/eval/segmentation.py
```diff
@@ -438,6 +452,9 @@ class SegmentationResults(BaseEvaluationResults):
         eval_key: the evaluation key
         pixel_confusion_matrix: a pixel value confusion matrix
         classes: a list of class labels corresponding to the confusion matrix
+        matches (None): a list of
+            ``(gt_label, pred_label, pixel_count, gt_id, pred_id)``
+            matches
```
prernadh, line 457: I like this data model better - reads cleaner and like you said avoids calling _parse_confusion_matrix when present.

prernadh, line 457: I like this data model better - reads cleaner and like you said avoids calling _parse_confusion_matrix when present.

fiftyone/utils/eval/segmentation.py
```diff
@@ -484,12 +507,31 @@ def dice_score(self):
...
+        ytrue_ids = d.get("ytrue_ids", None)
+        ypred_ids = d.get("ypred_ids", None)
+
+        if ytrue is not None and ypred is not None and weights is not None:
+            if ytrue_ids is None:
+                ytrue_ids = itertools.repeat(None)
+
+            if ypred_ids is None:
+                ypred_ids = itertools.repeat(None)
+
```
prernadh, line 522:
```suggestion
```

prernadh: Nit: Won't need the `None` check here because of the previous if-statement

plugins/panels/model_evaluation.py
```diff
@@ -585,6 +589,76 @@ def load_view(self, ctx):
...
+                    expr |= F(pred_id2).is_in(ypred_ids2)
+
+                view = eval_view.match(expr)
+            elif view_type == "matrix":
+                # Specific confusion matrix cell
+                ytrue_ids, ypred_ids = _get_segmentation_conf_mat_ids(
+                    results, x, y
+                )
+                expr = F(gt_id).is_in(ytrue_ids)
+                expr &= F(pred_id).is_in(ypred_ids)
```
prernadh, line 637: Mmmm, I see this expression should evaluate much faster than `select_labels` since you are specifying which labels to look for in which fields. Nice

plugins/panels/model_evaluation.py
```diff
@@ -605,3 +679,102 @@ def render(self, ctx):
...
+    elif field == "fn":
+        # False negatives
+        inds = results.ypred == results.missing
+        ytrue_ids = _to_object_ids(results.ytrue_ids[inds])
+        return ytrue_ids, None
+    else:
+        # False positives
+        inds = results.ytrue == results.missing
+        ypred_ids = _to_object_ids(results.ypred_ids[inds])
+        return None, ypred_ids
```
prernadh, line 776: Would we want to move this tp/fp/fn calculation to utils/eval/segmentation.py and make it a sample level field so we can filter on it - similar to detections?

prernadh (changes requested):

prernadh (commented):
Just a question about high level TP/FP/FN design

plugins/panels/model_evaluation.py
```diff
@@ -585,6 +589,76 @@ def load_view(self, ctx):
...
+                    )
+                    expr = F(gt_id).is_in(ytrue_ids)
+                    view = eval_view.match(expr)
+                else:
+                    # All false positives
+                    _, ypred_ids = _get_segmentation_tp_fp_fn_ids(
+                        results, field
+                    )
+                    expr = F(pred_id).is_in(ypred_ids)
+                    view = eval_view.match(expr)
```
prernadh, line 661: We currently don't display FP/FN/TP in the summary table- `get_tp_fp_fn` function will have to be updated for segmentations if we ever want to reach this section of the code is my understanding

fiftyone/utils/eval/segmentation.py
```diff
@@ -424,6 +437,7 @@ def evaluate_samples(
             eval_key,
             confusion_matrix,
             classes,
+            matches=matches,
```
prernadh, line 440:
```suggestion
            matches=matches if matches!=[] else None,
```

prernadh: This line is causing tests to fail

plugins/panels/model_evaluation.py
brimoor: I'm not sure what we'd store as a sample-level field. The TP/FP/FN designation has to do with each **region** in the segmentation mask, so there would usually be multiple per sample (like object detection tasks for example), but the the mask is stored in a single `Segmentation` field (unlike object detection).

prernadh: Hmm.. I see what you are saying- but I guess my confusion is arising from the fact that if we are marking labels as TP/FP/FN, we should be able to filter by it at the sample level too.

fiftyone/utils/eval/segmentation.py
brimoor: Actually they are required:

In previous versions of this code, `ytrue`, `ypred`, and `weights` were not persisted as properties of `SegmentationResults`. If a user loads such a pre-existing segmentation evaluation and then calls `results.save()`, this will create a new version of the results that *does persist* `ytrue`, `ypred` and `weights` (as parsed from `_parse_confusion_matrix()`). However, there still won't be `ytrue_ids` and `ypred_ids` for these results, so these `if` statements are needed to ensure that the *next time* these results are loaded, we'll be able to construct the `matches` object.

prernadh: Makes sense!

fiftyone/utils/eval/segmentation.py
brimoor: Good catch! Just fixed this in a slightly different way, for consistency with how object detection handles this: https://github.com/voxel51/fiftyone/blob/46bd6bde2b01641bc2ae056ef3672c3556b316d8/fiftyone/utils/eval/detection.py#L543-L553

plugins/panels/model_evaluation.py
brimoor: Exactly!

plugins/panels/model_evaluation.py
brimoor: Yeah I went ahead an implemented the callback so it would be available *if* we found a reasonable way to show TP/FP/FN icons in the panel. Not quite sure what the best way to show this info would be though.

plugins/panels/model_evaluation.py
brimoor: I agree that it would be nice to support more sample-level filtering, but I don't know what to do!

plugins/panels/model_evaluation.py
prernadh: Oh I understand now! Realized you had also left comment on the other PR explaining this. Apologies

plugins/panels/model_evaluation.py
prernadh: Yep understood. Makes sense to leave as is for now then. Maybe something to discuss with the ML team

prernadh (approved):
LGTM! 😄

prernadh (dismissed):
LGTM!

PR #5333, opened 2025-01-01
Cleanup builtin plugins interface

brimoor:
## Change log

Code
- ensure that `enable_plugin()`, `disable_plugin()` and `delete_plugin()` can't be run on builtin plugins
- consolidate `include_builtin` and `builtins_only` arguments into a single `builtin` argument, for consistency with the existing `enabled` argument

CLI
- include `builtin` column in `fiftyone plugins list` output
- include `--builtins-only` and `--no-builtins` filtering options for `fiftyone plugins|operators list`

Docs
- include plugins API reference in docs build
- fix various broken links for builtin operators in the docs

## Example usage

```shell
# Now includes `builtin` column
fiftyone plugins list

# New filtering options
fiftyone plugins list --builtins-only
fiftyone plugins list --no-builtins
fiftyone operators list --builtins-only
fiftyone operators list --no-builtins

# Not allowed
fiftyone plugins enable '@voxel51/operators'
fiftyone plugins disable '@voxel51/operators'
fiftyone plugins delete '@voxel51/operators'
```

## Release Notes

- **Documentation**
	- Updated documentation for CLI commands related to operators and plugins
	- Added API reference section for plugins
	- Improved clarity of plugin and operator documentation
	- Updated user guide references for model evaluation

- **New Features**
	- Enhanced filtering for operators and plugins (builtin/non-builtin)
	- Added support for more flexible plugin management
	- Introduced example Databricks connector and Data Lens connector

- **Bug Fixes**
	- Improved path normalization for plugin directories
	- Refined error handling for built-in plugins
	- Fixed potential `StopIteration` errors during long-running operations

- **Chores**
	- Updated plugin and panel configuration files
	- Streamlined plugin listing and retrieval logic

### DIFF
--- a/docs/generate_docs.bash
+++ b/docs/generate_docs.bash
@@ -103,10 +103,16 @@ sphinx-apidoc --force --no-toc --separate --follow-links \
     --templatedir=docs/templates/apidoc \
     -o docs/source/api fiftyone \
         fiftyone/brain/internal/models \
+        fiftyone/constants \
+        fiftyone/internal \
         fiftyone/server \
         fiftyone/service \
         fiftyone/management \
         fiftyone/api
+
+sphinx-apidoc --force --no-toc --separate --follow-links \
+    --templatedir=docs/templates/apidoc \
+    -o docs/source/plugins/api plugins
 
 # Remove symlink
 unlink fiftyone/brain
--- a/fiftyone/core/cli.py
+++ b/fiftyone/core/cli.py
@@ -2905,7 +2905,7 @@ class OperatorsCommand(Command):
 
 
 class OperatorsListCommand(Command):
-    """List operators and panels that you've installed locally.
+    """List operators and panels that are installed locally.
 
     Examples::
 
@@ -2918,7 +2918,10 @@ class OperatorsListCommand(Command):
         # List disabled operators and panels
         fiftyone operators list --disabled
 
-        # Only list panels
+        # List non-builtin operators and panels
+        fiftyone operators list --no-builtins
+
+        # List panels
         fiftyone operators list --panels-only
     """
 
@@ -2939,11 +2942,25 @@ class OperatorsListCommand(Command):
             help="only show disabled operators and panels",
         )
         parser.add_argument(
+            "-b",
+            "--builtins-only",
+            action="store_true",
+            default=None,
+            help="only show builtin operators and panels",
+        )
+        parser.add_argument(
+            "-c",
+            "--no-builtins",
+            action="store_true",
+            default=None,
+            help="only show non-builtin operators and panels",
+        )
+        parser.add_argument(
             "-o",
             "--operators-only",
             action="store_true",
             default=None,
-            help="only show operators",
+            help="only show non-panel operators",
         )
         parser.add_argument(
             "-p",
@@ -2968,6 +2985,13 @@ class OperatorsListCommand(Command):
         else:
             enabled = "all"
 
+        if args.builtins_only:
+            builtin = True
+        elif args.no_builtins:
+            builtin = False
+        else:
+            builtin = "all"
+
         if args.operators_only:
             type = "operator"
         elif args.panels_only:
@@ -2975,11 +2999,11 @@ class OperatorsListCommand(Command):
         else:
             type = None
 
-        _print_operators_list(enabled, type, args.names_only)
+        _print_operators_list(enabled, builtin, type, args.names_only)
 
 
-def _print_operators_list(enabled, type, names_only):
-    operators = foo.list_operators(enabled=enabled, type=type)
+def _print_operators_list(enabled, builtin, type, names_only):
+    operators = foo.list_operators(enabled=enabled, builtin=builtin, type=type)
 
     if names_only:
         operators_map = defaultdict(list)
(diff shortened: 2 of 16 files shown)


fiftyone/plugins/definitions.py
```diff
@@ -13,7 +13,7 @@
 import eta.core.serial as etas
 
 import fiftyone as fo
-import fiftyone.plugins.constants as constants
+import fiftyone.plugins.constants as fpc
```
ritch, line 16: This style is different in the other files. Change the others to be consistent.

fiftyone/core/cli.py
```diff
@@ -3558,14 +3606,15 @@ def _print_plugins_list(enabled, names_only):
...
-    headers = ["plugin", "version", "enabled", "directory"]
+    headers = ["plugin", "version", "enabled", "builtin", "directory"]
     rows = []
     for pd in plugin_defintions:
         rows.append(
             {
                 "plugin": pd.name,
                 "version": pd.version or "",
-                "enabled": pd.name in enabled_plugins,
+                "enabled": pd.builtin or pd.name in enabled_plugins,
```
ritch, line 3616: Good catch 👍

fiftyone/operators/registry.py
```diff
@@ -32,25 +32,25 @@ def get_operator(operator_uri, enabled=True):
...
 
-def list_operators(enabled=True, type=None, builtins_only=False):
+def list_operators(enabled=True, builtin="all", type=None):
     """Returns all available operators.
 
     Args:
         enabled (True): whether to include only enabled operators (True) or
             only disabled operators (False) or all operators ("all")
+        builtin (None): whether to include only builtin operators (True) or
+            only non-builtin operators (False) or all operators ("all")
```
ritch, line 42: Purely from a docs pov this is confusing since it lists both None as the default and “all”. These appear functionally equivalent when reviewing the implementation. Changing the default to “all” seems right.

fiftyone/plugins/core.py
```diff
@@ -97,9 +127,12 @@ def delete_plugin(plugin_name):
     Args:
         plugin_name: the plugin name
     """
-    plugin = _get_plugin(plugin_name)
+    plugin = get_plugin(plugin_name)
+    if plugin.builtin:
+        raise ValueError(f"Cannot delete builtin plugin '{plugin_name}'")
```
kaixi-wang, line 132: Good catch

fiftyone/plugins/definitions.py
ritch: [fixed](https://github.com/voxel51/fiftyone/pull/5333/commits/75c481b9dbae6e96482a7d81a152c333b6775341)

fiftyone/operators/registry.py
brimoor: yes intended to write `builtin ("all")` in the docs. Will fix now 👍

docs/generate_docs.bash
```diff
@@ -103,11 +103,17 @@ sphinx-apidoc --force --no-toc --separate --follow-links \
...
     -o docs/source/api fiftyone \
         fiftyone/brain/internal/models \
+        fiftyone/constants \
+        fiftyone/internal \
         fiftyone/server \
         fiftyone/service \
         fiftyone/management \
         fiftyone/api
 
+sphinx-apidoc --force --no-toc --separate --follow-links \
```
ritch, line 113: What is this for?

ritch: I'm not sure it makes sense to add this as "API Reference" for plugins. I think the existing `fiftyone.operators.operations` is what I would expect in a section like that.

ritch: Also is there any way to adjust the header/breadcrumb - it confusingly says "Docs / Plugins Overview / plugins"

brimoor: I agree that it's not very usable in its current form. The only reason I added this is that https://github.com/voxel51/fiftyone/pull/5261 broke a few links in the plugins docs and release notes that used to link to (equally unhelpful) resources about the available builtin operators in the `fiftyone` API reference tree.

ritch (approved):
We should look at how the new builtin plugins are displayed in the docs. Otherwise LGTM.

docs/generate_docs.bash
brimoor: One possible solution is to add class docstrings to all operators/panels that describe how to use them, like I did for `ViewExpression` here: 
https://docs.voxel51.com/api/fiftyone.core.expressions.html?highlight=viewexpression#fiftyone.core.expressions.ViewExpression

That would be a bit awkward though for this use case.

It's not that useful to have the panels in the API reference, as they aren't meant to be used programmatically and really should already have dedicated documentation sections for them.

There is some value in having a list of all the builtin operators, in the sense that they are totally undocumented to-date. It would be reasonable for _some_ of them to have programmatic execution examples, but generally they are really intended to be invoked via the App's Operator browser. But a docstring is an awkward place to document in-App usage. What we really need is some kind of operator documentation page that shows GIFs of every single operator in use.

ritch: > What we really need is some kind of operator documentation page that shows GIFs of every single operator in use.

Sure - I've always thought that we could allow for self-documentation, since operators define their own input/output.

We should change the "API reference". Based on what you said above I think the only value of this is the listing of what is in `/plugins`. In that case we should call it "Built-in plugins" or something similar.

PR #5337, opened 2025-01-02
clean overlays in detach function and not during reconciliation

sashankaryal:

- Cleanup overlays in looker's `detach` as opposed to reconciliation function
- For single channel masks, modify buffer in-place instead of allocating another UInt8 array
- Set targets to null in overlay cleanup. They should already be garbage collected, but this is just a safety mechanism to reduce ref to the buffer by one in case overaly refs are maintained somewhere. 
- Reuse offscreen canvas. I did micro-benchmarking with creating new offscreen canvas vs reusing it. I noticed a difference of 2ms at most, so this is kind of a useless change. Resizing the canvas probably offsets the cost of allocating a new canvas and destroying it.

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Performance Improvements**
	- Enhanced memory management for overlays by implementing cleanup mechanisms
	- Optimized image processing in canvas decoding with off-screen canvas and efficient data handling

- **Bug Fixes**
	- Improved resource cleanup for looker and scroll reader components
	- Added safeguards to prevent potential memory leaks during object destruction

- **New Features**
	- Introduced more robust overlay processing for heatmap and segmentation overlays

### DIFF
--- a/app/packages/looker/src/processOverlays.ts
+++ b/app/packages/looker/src/processOverlays.ts
@@ -4,6 +4,8 @@
 
 import { CONTAINS, Overlay } from "./overlays/base";
 import { ClassificationsOverlay } from "./overlays/classifications";
+import HeatmapOverlay from "./overlays/heatmap";
+import SegmentationOverlay from "./overlays/segmentation";
 import { BaseState } from "./state";
 import { rotate } from "./util";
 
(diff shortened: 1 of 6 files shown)


app/packages/looker/src/processOverlays.ts
```diff
@@ -30,6 +32,25 @@ const processOverlays = <State extends BaseState>(
 
     if (!(overlay.field && overlay.field in bins)) continue;
 
+    // todo: find a better approach / place for this.
+    // for instance, this won't work in detection overlay, where
+    // we might want the bounding boxes but masks might not have been loaded
```
benjaminpkane, line 37: I think these instances can be filtered out during overlay creation. But this is a good place for now 👍

app/packages/spotlight/src/createScrollReader.ts
```diff
@@ -73,6 +74,14 @@ export default function createScrollReader(
...
       element.removeEventListener("scroll", scroll);
       element.removeEventListener("scrollend", scrollEnd);
+
+      if (timeout) {
+        clearTimeout(timeout);
+      }
+
+      if (animationFrameId) {
+        cancelAnimationFrame(animationFrameId);
+      }
```
benjaminpkane, line 84: It seems like there is redundancy here. I would be surprised if the closure here is causing a memory leak. But it doesn't hurt 👍

benjaminpkane (approved):
LGTM 🐎

PR #5338, opened 2025-01-03
POC: use GPU to process overlays

sashankaryal:

successful POC where we use webgl to process grayscale segmentation masks. Currently, this processing happens in the CPU and is extremely compute intensive.

(Details)

        notes for FiftyOne users.

-   [ ] Other


jleven (commented):
Quick question: What happens if the client doesn’t have a GPU?

sashankaryal:
> What happens if the client doesn’t have

Webgl should automatically fallback to software rendering.

PR #5342, opened 2025-01-05
Handle shadowed plugins

brimoor:
## Change log

- Updates the plugins API to gracefully ignore plugins that are "shadowed" by other plugins of the same name
- Ensures that `download_plugin(..., overwrite=True)` and `create_plugin(..., overwrite=True)` cannot overwrite builtin plugins

## Example usage

```shell
PLUGINS_DIR="$(fiftyone config plugins_dir)"

# (if necessary) remove source install of `fiftyone-plugins`
unlink "$PLUGINS_DIR/fiftyone-plugins"

# Install a plugin
fiftyone plugins download \
    --plugin-names @voxel51/dashboard

# Duplicate the plugin
ORIG_DIR="$PLUGINS_DIR/@voxel51/dashboard"
DUP_DIR="$PLUGINS_DIR/also_dashboard"
cp -R $ORIG_DIR $DUP_DIR

# See how duplicate plugins are handled
fiftyone plugins list
"""
plugin                            version    enabled    builtin    shadowed    directory
--------------------------------  ---------  ---------  ---------  ----------  ------------------------------------------------------
@voxel51/operators                1.0.0      ✓          ✓                      /Users/Brian/dev/fiftyone/plugins/operators
@voxel51/panels                   1.0.0      ✓          ✓                      /Users/Brian/dev/fiftyone/plugins/panels
@voxel51/dashboard                1.0.1      ✓                                 /Users/Brian/fiftyone/__plugins__/@voxel51/dashboard
@voxel51/dashboard                1.0.1                            ✓           /Users/Brian/fiftyone/__plugins__/also_dashboard
"""

# Verify that the App works
# EG launch the Dashboard panel
fiftyone app launch

# Only affects the "active" plugin
# "Shadowed" plugin remains disabled
fiftyone plugins disable @voxel51/dashboard
fiftyone plugins list

# Only affects the "active" plugin
# "Shadowed" plugin remains disabled
fiftyone plugins enable @voxel51/dashboard
fiftyone plugins list

# Shows the "active" plugin
# The `shadow_paths` row indicates that this plugin is shadowing `also_dashboard`
fiftyone plugins info @voxel51/dashboard

# Show info about a "shadowed" plugin
fiftyone plugins info $DUP_DIR

# Delete a shadowed plugin
rm -rf $DUP_DIR

# The `shadowed` column is only shown if there are actually shadowed plugins
fiftyone plugins list
"""
plugin                            version    enabled    builtin    directory
--------------------------------  ---------  ---------  ---------  ------------------------------------------------------
@voxel51/operators                1.0.0      ✓          ✓          /Users/Brian/dev/fiftyone/plugins/operators
@voxel51/panels                   1.0.0      ✓          ✓          /Users/Brian/dev/fiftyone/plugins/panels
@voxel51/dashboard                1.0.1      ✓                     /Users/Brian/fiftyone/__plugins__/@voxel51/dashboard
"""

# Cleanup
fiftyone plugins delete @voxel51/dashboard

# (if necessary)
cd /path/to/fiftyone-plugins
ln -s "$(pwd)" "$(fiftyone config plugins_dir)/fiftyone-plugins"
```

## Release Notes

- **New Features**
  - Enhanced plugin management with support for shadowed plugins.
  - Expanded `fiftyone plugins info` command to accept plugin name or directory path.

- **Documentation**
  - Updated CLI documentation to clarify plugin information retrieval.
  - Improved command usage examples for plugin-related commands.

- **Improvements**
  - Added more comprehensive plugin listing capabilities, including shadowed plugins.
  - Enhanced plugin detection and management logic.
  - Improved error handling and logging for plugin operations.

These updates provide more flexibility and clarity in managing FiftyOne plugins, making it easier to discover and interact with plugin resources.

### DIFF
--- a/fiftyone/plugins/core.py
+++ b/fiftyone/plugins/core.py
@@ -5,7 +5,6 @@ Core plugin methods.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-from collections import Counter
 from dataclasses import dataclass
 import json
 import logging
@@ -13,6 +12,7 @@ import os
 from packaging.requirements import Requirement
 import re
 import shutil
+from typing import Optional
 
 import yaml
 
@@ -44,19 +44,22 @@ class PluginPackage:
 
     name: str
     path: str
+    shadow_paths: Optional[list[str]] = None
 
     def __repr__(self):
         return f"Plugin(name={self.name}, path={self.path})"
 
 
-def list_plugins(enabled=True, builtin=False):
-    """Returns the definitions of available plugins.
+def list_plugins(enabled=True, builtin=False, shadowed=False):
+    """Lists available plugins.
 
     Args:
         enabled (True): whether to include only enabled plugins (True) or only
             disabled plugins (False) or all plugins ("all")
         builtin (False): whether to include only builtin plugins (True) or only
             non-builtin plugins (False) or all plugins ("all")
+        shadowed (False): whether to include only "shadowed" duplicate plugins
+            (True) or only usable plugins (False) or all plugins ("all")
 
     Returns:
         a list of :class:`PluginDefinition` instances
@@ -67,12 +70,17 @@ def list_plugins(enabled=True, builtin=False):
     if builtin == "all":
         builtin = None
 
+    if shadowed == "all":
+        shadowed = None
+
     plugins = []
-    for p in _list_plugins(enabled=enabled, builtin=builtin):
+    for p in _list_plugins(
+        enabled=enabled, builtin=builtin, shadowed=shadowed
+    ):
         try:
-            plugins.append(_load_plugin_definition(p))
+            plugins.append(_load_plugin_definition(plugin=p))
         except:
-            logger.debug(f"Failed to parse plugin at '{p.path}'")
+            logger.info(f"Failed to parse plugin at '{p.path}'")
 
     return plugins
 
@@ -141,7 +149,8 @@ def list_downloaded_plugins():
     Returns:
         a list of plugin names
     """
-    return _list_plugins_by_name(builtin=False)
+    plugins = _list_plugins(builtin=False, shadowed=False)
+    return [p.name for p in plugins]
 
 
 def list_enabled_plugins():
@@ -150,7 +159,8 @@ def list_enabled_plugins():
     Returns:
         a list of plugin names
     """
-    return _list_plugins_by_name(enabled=True, builtin=False)
+    plugins = _list_plugins(enabled=True, builtin=False, shadowed=False)
+    return [p.name for p in plugins]
 
 
 def list_disabled_plugins():
@@ -159,20 +169,25 @@ def list_disabled_plugins():
     Returns:
         a list of plugin names
     """
-    return _list_plugins_by_name(enabled=False, builtin=False)
+    plugins = _list_plugins(enabled=False, builtin=False, shadowed=False)
+    return [p.name for p in plugins]
 
 
-def get_plugin(name):
+def get_plugin(name=None, plugin_dir=None):
     """Gets the definition for the given plugin.
 
     Args:
-        name: the plugin name
+        name (None): the plugin name
+        plugin_dir (None): a directory containing the plugin
 
     Returns:
         a :class:`PluginDefinition`
     """
+    if plugin_dir is not None:
+        return _load_plugin_definition(plugin_dir=plugin_dir)
+
     plugin = _get_plugin(name)
-    return _load_plugin_definition(plugin)
+    return _load_plugin_definition(plugin=plugin)
 
 
 def find_plugin(name):
@@ -233,9 +248,12 @@ def download_plugin(url_or_gh_repo, plugin_names=None, overwrite=False):
     elif plugin_names is not None:
         plugin_names = set(plugin_names)
 
-    existing_plugin_dirs = {p.name: p.path for p in _list_plugins()}
-
+    existing_plugins = {
+        p.name: p for p in list_plugins(enabled="all", builtin="all")
+    }
     downloaded_plugins = {}
+    skipped_plugins = set()
+
     with etau.TempDir() as tmpdir:
         root_dir = tmpdir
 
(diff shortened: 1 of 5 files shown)


ritch:
```
(fiftyone-env2) ➜  plugins git:(prompt-on-cancel) ✗ fiftyone plugins disable @voxel51/panels
  File "/Users/ritchie/.pyenv/versions/fiftyone-env2/bin/fiftyone", line 8, in <module>
    sys.exit(main())
  File "/Users/ritchie/Projects/voxel51/fiftyone/fiftyone/core/cli.py", line 4715, in main
    args.execute(args)
  File "/Users/ritchie/Projects/voxel51/fiftyone/fiftyone/core/cli.py", line 4698, in <lambda>
    parser.set_defaults(execute=lambda args: command.execute(parser, args))
  File "/Users/ritchie/Projects/voxel51/fiftyone/fiftyone/core/cli.py", line 3986, in execute
    fop.disable_plugin(name)
  File "/Users/ritchie/Projects/voxel51/fiftyone/fiftyone/plugins/core.py", line 125, in disable_plugin
    raise ValueError(
ValueError: Cannot change enablement of builtin plugin '@voxel51/panels'
```

Lots of unrelated output, would have expected just the error message instead of the trace.

ritch:
Manually tested duplicate plugins, duplicate builtins, disabling builtins (possibly unrelated to this PR!) ✅

ritch:
Also tested duplicate builtins in the app ✅

fiftyone/plugins/core.py
```diff
@@ -44,19 +44,22 @@ class PluginPackage:
...
+def list_plugins(enabled=True, builtin=False, shadowed=False):
+    """Lists available plugins.
 
     Args:
         enabled (True): whether to include only enabled plugins (True) or only
             disabled plugins (False) or all plugins ("all")
         builtin (False): whether to include only builtin plugins (True) or only
             non-builtin plugins (False) or all plugins ("all")
+        shadowed (False): whether to include only "shadowed" duplicate plugins
+            (True) or only usable plugins (False) or all plugins ("all")
```
ritch, line 62: Can `builtin` and `shadowed` be used together? The wording here is confusing.

fiftyone/plugins/core.py
```diff
@@ -278,15 +296,21 @@ def download_plugin(url_or_gh_repo, plugin_names=None, overwrite=False):
...
-                etau.delete_dir(existing_dir)
-                plugin_dir = existing_dir
+                if existing_plugin.builtin:
+                    raise ValueError(
+                        f"Cannot overwrite builtin plugin '{plugin.name}'"
+                    )
+
+                logger.info(f"Overwriting existing plugin '{plugin.name}'")
+                plugin_dir = existing_plugin.directory
+                etau.delete_dir(plugin_dir)
```
ritch, line 313: When does this happen? I would hope that something like `fiftyone plugins download XXX --force` would be required to avoid accidentally deleting something. Possibly exists, just commenting here to make sure we double check this.

fiftyone/plugins/core.py
```diff
@@ -621,24 +668,25 @@ def _iter_plugin_metadata_files(root_dir=None, strict=False):
...
-    plugin = None
-    for _plugin in _list_plugins(enabled=enabled, builtin=builtin):
-        if _plugin.name == name:
-            if check_for_duplicates and plugin is not None:
-                raise ValueError(f"Multiple plugins found with name '{name}'")
+def _get_plugin(name, enabled=None, builtin=None):
+    for plugin in _list_plugins(
+        enabled=enabled, builtin=builtin, shadowed=False
+    ):
+        if plugin.name == name:
```
ritch, line 675: We use this style all over and its not that efficient. Instead of loading everything and then searching for a name we should load only until we get the named plugin. Mostly concerned about unnecessary fs calls.

Maybe a `_find_plugin(name=, plugin_dir=)` would be useful for this.

Can handle such improvements elsewhere.

fiftyone/plugins/definitions.py
```diff
@@ -194,6 +202,7 @@ def to_dict(self):
         """
         return {
             "name": self.name,
+            "builtin": self.builtin,
```
ritch, line 205: ?

ritch: I find that moving this arbitrarily adds to the OSS => teams conflicts.

ritch (approved):
LGTM

fiftyone/plugins/core.py
```diff
@@ -44,19 +44,22 @@ class PluginPackage:
 
     name: str
     path: str
+    shadow_paths: Optional[list[str]] = None
```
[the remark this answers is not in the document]
brimoor: My official mantra is follow the existing convention in the module you're contributing to, so I had to do it 😄 

And... also had to do it because it's required when defining data classes 😉

fiftyone/plugins/core.py
brimoor: Correct, user must specify `download_plugin(..., overwrite=True)` or `fiftyone plugins download ... --overwrite` in order to overwrite an existing plugin.

They'll get an error without forcing the overwrite.

fiftyone/plugins/core.py
brimoor: yes refactoring `_list_plugins()` to be a generator would 2x average performance here.

Importantly, however, I believe all of the App's plugin usage goes through `build_plugin_contexts()`, which is cached via `@plugins_cache`.

fiftyone/plugins/definitions.py
brimoor: I moved it so that `builtin` appears immediately below `name` when printing plugin info:

```
$ fiftyone plugins info @voxel51/operators
key                     value
----------------------  -----------------------------------------------------
name                    @voxel51/operators
builtin                 True
...
```

That's a bit hacky of course as it relies on dicts having a stable iteration order, which isn't necessarily guaranteed in general.

brimoor: +1 for not causing diffs that have no functional purpose though

ritch: Ah actually this makes sense, just didn't realize.

PR #5347, opened 2025-01-06
Removed registry._EXTRA_OPERATORS

minhtuev:

Used patching so we don't have to inject operators for testing

* Ran unit test locally

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

- **Refactor**
	- Removed `_EXTRA_OPERATORS` variable from the operator registry.

- **Tests**
	- Modified test function to use mocked `list_operators` method.
	- Introduced a new test suite for the `OperatorRegistry` class, validating various scenarios for operator listing.

### DIFF
--- a/tests/unittests/operators/executor_tests.py
+++ b/tests/unittests/operators/executor_tests.py
@@ -1,16 +1,13 @@
 import pytest
-from unittest.mock import MagicMock, patch
-from starlette.exceptions import HTTPException
+from unittest.mock import patch
 
 import fiftyone.operators.types as types
 from fiftyone.operators.operator import Operator
 from fiftyone.operators.executor import (
     execute_or_delegate_operator,
     ExecutionResult,
-    ExecutionContext,
 )
 from fiftyone.operators import OperatorConfig
-import fiftyone.operators.registry as registry
 
 
 ECHO_URI = "@voxel51/operators/echo"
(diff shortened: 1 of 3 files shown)


tests/unittests/operators/executor_tests.py
```diff
@@ -30,12 +27,11 @@ def execute(self, ctx):
         return {"message": ctx.params.get("message", None)}
 
 
-# Force registration of the operator for testing
-registry._EXTRA_OPERATORS.append(EchoOperator(_builtin=True))
-
-
 @pytest.mark.asyncio
-async def test_execute_or_delegate_operator():
+@patch("fiftyone.operators.registry.OperatorRegistry.list_operators")
```
ritch, line 31: Interesting! I just used this in teams but as `unittest.mock.patch()`. 

Is there a way to use the real `list_operators`? Thats a big part of whats being tested here. Although `_EXTRA_OPERATORS` isn't a better solution.

ritch (approved):

tests/unittests/operators/executor_tests.py
minhtuev: I thought about it! We can achieve the same effect by mocking out `fopc.build_plugin_contexts` but I think it's cleaner to mock out `list_operators`.

I can add a unit test for `list_operators` too so that it is also covered by tests.

PR #5348, opened 2025-01-06
Add on_cancel callback for operator prompt

ritch:

Adds a new `on_cancel` callback when prompting for an operator. This will be called if a user cancels the operator prompt. NOTE: this behavior is not intended or available for immediately executing operators, such as when `skip_prompt=True` or the operator does not have input.

The example below is available as [a panel example here](https://github.com/voxel51/fiftyone-plugins/pull/194).

```python
class PromptExample(foo.Panel):
    def render(self, ctx):
        panel = types.Object()
        msg = ctx.panel.state.message
        result = f"message: {msg}" if msg else "..."
        panel.md(
            f"""
            ### Prompt Example

            This is an example of a prompt that can be used to gather user input.
                 
            {result}
        """
        )
        panel.btn("prompt", label="Prompt", on_click=self.on_click_prompt)
        return types.Property(
            panel,
            view=types.GridView(margin=5),
        )

    def on_success(self, ctx):
        ctx.panel.state.message = ctx.params.get("result", {}).get("message")
        ctx.ops.notify("Prompt was successful!")

    def on_cancel(self, ctx):
        ctx.ops.notify("Prompt was canceled", variant="warning")

    def on_click_prompt(self, ctx):
        ctx.prompt(
            "@voxel51/panel-examples/example_message",
            on_success=self.on_success,
            # NEW
            on_cancel=self.on_cancel,
        )
```

(Details)

        notes for FiftyOne users.

Adds a new `on_cancel` event for `ctx.prompt()`.

-   [x] Plugins
-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Added a new `on_cancel` parameter to prompt operations across multiple components.
	- Introduced enhanced cancellation handling for operator prompts.
	- Implemented a flexible mechanism to execute custom logic when a prompt is canceled.
	- Added new properties `onCancel` and `prompt` to relevant components for improved interaction.

- **Improvements**
	- Extended prompt functionality to support more robust user interaction scenarios.
	- Improved error handling and user experience during operation cancellations.
	- Streamlined component logic for better management of operator execution and menu interactions.

### DIFF
--- a/app/packages/operators/src/components/OperatorExecutionTrigger/index.tsx
+++ b/app/packages/operators/src/components/OperatorExecutionTrigger/index.tsx
@@ -46,7 +46,6 @@ export const OperatorExecutionTrigger = ({
   onClick?: () => void;
   onSuccess?: ExecutionCallback;
   onError?: ExecutionErrorCallback;
-  onCancel?: () => void;
   prompt?: boolean;
   executionParams?: object;
   executorOptions?: OperatorExecutorOptions;
--- a/app/packages/operators/src/components/OperatorExecutionTrigger/index.tsx
+++ b/app/packages/operators/src/components/OperatorExecutionTrigger/index.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useMemo, useRef, useState } from "react";
-import { Box } from "@mui/material";
+import { Box, Button } from "@mui/material";
 import { OperatorExecutionMenu } from "../OperatorExecutionMenu";
 import {
   ExecutionCallback,
@@ -10,6 +10,7 @@ import {
   OperatorExecutionOption,
   useOperatorExecutionOptions,
   useOperatorExecutor,
+  usePromptOperatorInput,
 } from "../../state";
 
 /**
@@ -38,12 +39,8 @@ import {
  * @param disabled If true, context menu will never open
  */
 export const OperatorExecutionTrigger = ({
-  operatorUri,
   onClick,
-  onSuccess,
-  onError,
-  executionParams,
-  executorOptions,
+  executionOptions,
   onOptionSelected,
   disabled,
   children,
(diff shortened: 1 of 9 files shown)


app/packages/operators/src/state.ts
```diff
@@ -579,6 +579,11 @@ export const useOperatorPrompt = () => {
     },
     [operator, promptingOperator, cachedResolvedInput]
   );
+  const onCancel = promptingOperator.options?.onCancel;
+  const cancel = () => {
```
Br2850, line 583: @ritch I like this as a seperate function vs `close()` but the question now is - where does `cancel()` get invoked in a prompt? 

We need to invoke this whenever:

1. The "cancel" button is specifically clicked
2. The user clicks outside of the modal

This is likely the BrowserEvent or subscription you & Ibrahim were talking about

If that is the case, then this PR vs mine will be cleaner in the sense that the panel user doesn't have to worry about params & conditionals to make sure they are not just getting a generic close operation

ritch: This was alerady handled [here](https://github.com/voxel51/fiftyone/blob/e7f361b8618fd1dc2b801a5b9831652c1166d06a/app/packages/operators/src/utils.ts#L69)

app/packages/operators/src/components/OperatorExecutionTrigger/index.tsx
```diff
@@ -54,40 +46,18 @@ export const OperatorExecutionTrigger = ({
   onClick?: () => void;
   onSuccess?: ExecutionCallback;
   onError?: ExecutionErrorCallback;
+  onCancel?: () => void;
```
imanjra, line 49: Looks like `onCancel` is not being used in this component? Except for maybe `Box`?

imanjra (approved):
LGTM

PR #5355, opened 2025-01-07
Proactive vs. reactive content size batcher

swheaton:

TBD

TBD

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -3360,10 +3360,6 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
             if sample.media_type == fom.VIDEO:
                 sample.frames.save()
 
-        if batcher is not None and batcher.manual_backpressure:
-            # @todo can we infer content size from insert_many() above?
-            batcher.apply_backpressure(dicts)
-
         return [str(d["_id"]) for d in dicts]
 
     def _upsert_samples(
@@ -3426,10 +3422,6 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
 
             if sample.media_type == fom.VIDEO:
                 sample.frames.save()
-
-        if batcher is not None and batcher.manual_backpressure:
-            # @todo can we infer content size from bulk_write() above?
-            batcher.apply_backpressure(dicts)
 
     def _make_dict(
         self,
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -1069,16 +1069,8 @@ def _report_progress_dt(progress, dt):
 
 
 class Batcher(abc.ABC):
-    """Base class for iterating over the elements of an iterable in batches."""
-
-    manual_backpressure = False
-
     def __init__(
-        self,
-        iterable,
-        return_views=False,
-        progress=False,
-        total=None,
+        self, iterable, return_views=False, progress=False, total=None
     ):
         import fiftyone.core.collections as foc
 
@@ -1100,7 +1092,6 @@ class Batcher(abc.ABC):
         self._render_progress = bool(progress)  # callback function: True
         self._last_offset = None
         self._num_samples = None
-        self._manually_applied_backpressure = True
 
     def __enter__(self):
         self._in_context = True
@@ -1140,17 +1131,15 @@ class Batcher(abc.ABC):
 
         return self
 
+    @abc.abstractmethod
     def __next__(self):
-        if (
-            self.manual_backpressure
-            and not self._manually_applied_backpressure
-        ):
-            raise ValueError(
-                "Backpressure value not registered for this batcher"
-            )
+        pass
 
-        self._manually_applied_backpressure = False
 
+class BaseBatcher(Batcher):
+    """Base class for iterating over the elements of an iterable in batches."""
+
+    def __next__(self):
         if self._render_progress and self._last_batch_size is not None:
             self._pb.update(count=self._last_batch_size)
 
@@ -1185,21 +1174,12 @@ class Batcher(abc.ABC):
 
         return batch
 
-    def apply_backpressure(self, *args, **kwargs):
-        """Apply backpressure needed to rightsize the next batch.
-
-        Required to be implemented and called every iteration, if
-        ``self.manual_backpressure == True``.
-
-        Subclass defines arguments and behavior of this method.
-        """
-
     @abc.abstractmethod
     def _compute_batch_size(self):
         """Return next batch size. Concrete classes must implement."""
 
 
-class BaseDynamicBatcher(Batcher):
+class BaseDynamicBatcher(BaseBatcher):
     """Class for iterating over the elements of an iterable with a dynamic
     batch size to achieve a desired target measurement.
 
@@ -1210,6 +1190,8 @@ class BaseDynamicBatcher(Batcher):
     Concrete base classes define the target measurement and method of
     calculation.
     """
+
+    manual_backpressure = False
 
     def __init__(
         self,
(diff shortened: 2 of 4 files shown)


fiftyone/core/utils.py
```diff
@@ -1539,7 +1534,102 @@ def _compute_batch_size(self):
...
+            callback function to invoke instead
+        total (None): the length of ``iterable``. Only applicable when
+            ``progress=True``. If not provided, it is computed via
+            ``len(iterable)``, if possible
+    """
+
+    def __init__(
+        self,
+        iterable,
+        size_calculation_fn,
```
CamronStaley, line 1584: what's this size_calculation_fn? I don't see it in the comments or an example of it being passed in anywhere unless I am just missing it.

CamronStaley:
workin on fixing that one test not sure why it's having issues yet

fiftyone/core/dataset.py
```diff
@@ -3336,41 +3356,22 @@ def add_collection(
...
-            self._validate_samples(samples)
-
-        now = datetime.utcnow()
-        dicts = [
-            self._make_dict(sample, created_at=now, last_modified_at=now)
-            for sample in samples
-        ]
-
+    def _add_samples_batch(self, samples, batcher=None):
+        dicts = [sample[1] for sample in samples]
```
swheaton, line 3360: I was very confused until I realized samples is tuple(sample, transformed_dict).... this is where type hints would be 🤌

fiftyone/core/utils.py
```diff
@@ -1138,19 +1138,24 @@ def __iter__(self):
                 )
                 self._render_progress = False
 
+        if self._transform_fn is not None:
+            self._iter = self._apply_transform(self._iter)
```
swheaton, line 1142: could also just do `self._iter = map(self._transform_fn, self._iter)`

fiftyone/core/dataset.py
```diff
@@ -3336,41 +3356,22 @@ def add_collection(
...
+        for sample, d in samples:
             doc = self._sample_dict_to_doc(d)
             sample._set_backing_doc(doc, dataset=self)
             if sample.media_type == fom.VIDEO:
                 sample.frames.save()
 
-        if batcher is not None and batcher.manual_backpressure:
-            # @todo can we infer content size from insert_many() above?
+        if batcher and batcher.manual_backpressure:
             batcher.apply_backpressure(dicts)
```
swheaton, line 3375: we don't need this any longer do we? esp. since you (or was it me i cant remember lol) moved it to Dynamic content size batcher only.

fiftyone/core/dataset.py
```diff
@@ -3429,16 +3466,15 @@ def _upsert_samples_batch(
...
             sample._set_backing_doc(doc, dataset=self)
 
             if sample.media_type == fom.VIDEO:
                 sample.frames.save()
 
-        if batcher is not None and batcher.manual_backpressure:
-            # @todo can we infer content size from bulk_write() above?
-            batcher.apply_backpressure(dicts)
+        if batcher and batcher.manual_backpressure:
+            batcher.apply_backpressure([sample[1] for sample in samples])
```
swheaton, line 3477: same

fiftyone/core/utils.py
```diff
@@ -1539,7 +1569,140 @@ def _compute_batch_size(self):
...
+                ):
+                    break
+                curr_batch_size += 1
+                batch_content_size += next_element_size
+                curr_batch.append(self._prev_element)
+            except StopIteration:
+                self._prev_element = None
+                break
+        self._last_batch_size = curr_batch_size
+        return curr_batch
```
swheaton, line 1684: I think I'm the one that wrote this lol. But I'm wondering if there's a cleaner way, this is a little ugly 🤣

swheaton (commented):
done with first pass, comments in line; gonna do a second pass in IDE

fiftyone/core/utils.py
CamronStaley: yeah you wrote this lol I will try and refactor tmr

fiftyone/core/dataset.py
CamronStaley: is it not still needed for latency batcher? I will have to look on IDE in the morning but yeah can clean these up if it's no longer used outside of dynamic content size.

fiftyone/core/utils.py
```diff
@@ -1539,7 +1569,140 @@ def _compute_batch_size(self):
...
+        return len(json_util.dumps(obj))
+    except Exception:
+        return len(str(obj))
+
+
+def get_default_batcher(
+    iterable,
+    progress=False,
+    total=None,
+    size_calc_fn=default_calculate_size,
```
swheaton, line 1703: I think this can default None, since the default is handled in ContentSizeBatcher init, yes?

Also, its missing from docstring

fiftyone/core/utils.py
```diff
@@ -1539,7 +1569,140 @@ def _compute_batch_size(self):
...
+
+    def __next__(self):
+        if self._render_progress and self._last_batch_size:
+            self._pb.update(count=self._last_batch_size)
+
+        if self._prev_element is None:
+            raise StopIteration()
+
+        batch_content_size = self._size_calculation_fn(self._prev_element)
+        curr_batch_size = 1
```
swheaton, line 1661: unnecessary as we can just use len(curr_batch) where we need it

fiftyone/core/utils.py
```diff
@@ -1433,6 +1453,7 @@ class ContentSizeDynamicBatcher(BaseDynamicBatcher):
         total (None): the length of ``iterable``. Only applicable when
             ``progress=True``. If not provided, it is computed via
             ``len(iterable)``, if possible
+        transform_fn (None): a transform function to apply to each item of the batch
     """
 
     manual_backpressure = True
```
swheaton, line 1459: Yeah I think we can get rid of this variable entirely

fiftyone/core/utils.py
```diff
@@ -1069,7 +1069,6 @@ def progress_dt(pb):
 
 
 class Batcher(abc.ABC):
```
swheaton, line 1071: I don't like what I did with the class naming. How about this becomes `BaseBatcher` and `BaseBatcher` becomes something like `BaseChunkyBatcher`. Meaning the batch size is determined per chunk via `_compute_batch_size()`
Would have to explain this in docstring of course 🙃

swheaton:
Awesome 👏🏼 . I'll test it out next. And re-review when you say go

fiftyone/core/dataset.py
swheaton: It is not. ContentSizeDynamicBatcher is the only one where manual_backpressure is True.
I had defined that variable on the root batcher since we can get 1 of a few different batchers, and I didn't want to do `hasattr(batcher, "manual_backpressure")`.
But that batcher isn't in get_default_batcher anymore so we can get rid of it

fiftyone/core/utils.py
swheaton: ya i can help

fiftyone/core/dataset.py
kaixi-wang: Yeah can we add a comment at least?

fiftyone/core/utils.py
swheaton: failed to come up with something better than this so just renamed prev_elt to next_elt which makes more sense, and added some comments.

swheaton:
LGTM! Pushed some minor coderabbit suggestions.
Also tested and it works! No readtimeout or duplicate key error.

🚢 🇮🇹 
(This is technically my PR so I cannot review it)

CamronStaley (approved):

PR #5356, opened 2025-01-08
add default labels visibility in dataset app config + selective labels rendering optimization

sashankaryal:

This PR adds a new dataset app config for configuring default visibility of labels. If `app_config.default_visibility_labels` is not set, all labels are not rendered by default.

Also adds an optimization of not rendering labels if it's unchecked in the sidebar.

### Notes to the reviewer

The most relevant files on the app side are: 
    - `useSelect.ts`
    - `labelsVisibility.ts`
    - `useDetectNewActiveLabelFields.ts`
    - `lookers/abstract.ts`
    - `worker/index.ts`
    - `worker/async-labels-rendering-manager.test.ts`

Known Issues:
- Recoloring samples too fast has a race condition (data clone error in `postMessage`). This doesn't crash the app, but might lead to confusion. I think it's safe to fix it in upcoming PRs.

Unit tests where applicable. E2E tests.

For local smoke testing, use following script:
```
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("coco-2014", split="validation")

dataset.app_config.default_visibility_labels = {"include": ["detections"]}
dataset.save()
```

Place a debugger in `painter.ts` for `Segmentations` and observe it doesn't get called. Scroll through the grid and check/uncheck fields to see different behavior.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced customizable default visibility settings for datasets, allowing users to refine which labels are displayed by default.
	- Added new hooks for selective rendering of modals based on active label fields and current frame data for different types of lookers.
	- Implemented methods for refreshing overlays and samples in the `ImaVidLooker` class to enhance user interaction with video data.
	- Added a new function to synchronize label fields for lookers, improving the management of active labels.
	- Enhanced error messaging in the UI to provide more detailed feedback during issues.
- **Improvements**
	- Enhanced overlay rendering and sample refresh behavior for a smoother, more reliable viewing experience.
	- Optimized media viewer components (image and video) for improved responsiveness and performance.
- **Tests**
	- Expanded test coverage to ensure robust label processing and media interaction, contributing to overall system stability.


kaixi-wang:
If the default is not set, does that mean previously rendered things will suddenly not be visible unless someone explicitly sets it? 

Considering the multi-user case, Should this be set at the dataset level for all users or per dataset per user?

app/packages/core/src/components/Grid/Grid.tsx
```diff
@@ -46,6 +47,15 @@ function Grid() {
...
   const getFontSize = useFontSize(id);
 
+  const getCurrentActiveLabelFields = useRecoilCallback(
+    ({ snapshot }) =>
+      () => {
+        return snapshot
+          .getLoadable(fos.activeLabelFields({ modal: false }))
+          .getValue();
+      }
+  );
```
benjaminpkane, line 57: Note, this callback is created on every render. Intentional?

sashankaryal: good catch. nope! will fix

app/packages/utilities/src/schema.ts
```diff
@@ -1,3 +1,16 @@
...
+export const SEGMENTATION_EMBEDDED_DOC_TYPE =
+  "fiftyone.core.labels.Segmentation";
+export const HEATMAP_EMBEDDED_DOC_TYPE = "fiftyone.core.labels.Heatmap";
+
+export const DENSE_LABEL_EMBEDDED_DOC_TYPES = [
+  DETECTION_EMBEDDED_DOC_TYPE,
+  DETECTIONS_EMBEDDED_DOC_TYPE,
+  SEGMENTATION_EMBEDDED_DOC_TYPE,
+  HEATMAP_EMBEDDED_DOC_TYPE,
+];
```
benjaminpkane, line 12: Duplicates values in `./constants`. As noted offline, I will follow up

app/packages/looker/src/worker/shared.ts
```diff
@@ -1,5 +1,12 @@
 import { HEATMAP } from "@fiftyone/utilities";
 
+export type DenseLabelRenderStatus =
+  | null
+  | "pending"
+  | "painting"
+  | "painted"
+  | "decoded";
```
benjaminpkane, line 8: These statuses are used quite frequently. Something like the below could be useful
```suggestion
export const DECODED = "decoded";
export const PAINTED = "painted";
export const PAINTING = "painting";
export const PENDING = "pending";
export type DenseLabelRenderStatus =
  | null
  | DECODED
  | PENDING
  | PAINTED
  | PAINTING
```

app/packages/looker/src/lookers/abstract.ts
```diff
@@ -517,9 +530,73 @@ export abstract class AbstractLooker<
...
+    const updateTimeoutMs = 10000;
+
+    if (UPDATING_SAMPLES_IDS.has(id)) {
+      UPDATING_SAMPLES_IDS.delete(id);
+      this.cleanOverlays(true);
+
+      // to prevent deadlock, we'll remove the id from the set after a timeout
+      const timeoutId = setTimeout(() => {
+        UPDATING_SAMPLES_IDS.delete(id);
+      }, updateTimeoutMs);
```
benjaminpkane, line 549: I don't see how this has an effect, noting that id is deleted on 542, as well.

Can UPDATING_SAMPLES_IDS be replace with state internal to Looker? Like isBatching? Introducing a global store feels somewhat inflexible

app/packages/core/src/components/Grid/useSelect.ts
```diff
@@ -24,11 +30,62 @@ export default function useSelect(
         }
 
         retained.add(id.description);
+
+        const newFieldsIfAny = getNewFields(id.description);
+
+        const overlays =
+          instance instanceof VideoLooker
+            ? instance.pluckedOverlays
+            : instance.sampleOverlays;
```
benjaminpkane, line 39: I don't think exposing these is necessary. `refreshSample` could capture the overlay filtering logic happening below

app/packages/looker/src/worker/async-labels-rendering-manager.ts
```diff
@@ -0,0 +1,199 @@
...
+import { v4 as uuid } from "uuid";
+import { ProcessSample } from ".";
+import { Coloring, Sample } from "..";
+import { LookerUtils } from "../lookers/shared";
+import { createWorker } from "../util";
+
+export type AsyncLabelsRenderingJob = {
+  sample: Sample;
+  labels: string[];
+  lookerRef: Lookers;
```
benjaminpkane, line 11: I think this ref could be replaced a function that returns the required options. This would would exposing the internal `state`

benjaminpkane (changes requested):
A couple nits that are probably worth following up on, but once video is working this LGTM!

app/packages/utilities/src/schema.ts
sashankaryal: Sounds good.

benjaminpkane (approved):
LGTM

PR #5357, opened 2025-01-08
Better exception handling reserved attribute

minhtuev:

Currently, when initializing FiftyOne objects like Detection with reserved attribute names (e.g., has_mask), the code fails with a confusing AttributeError indicating that a property has no setter. This error message isn't user-friendly and doesn't clearly explain that these are reserved attributes that shouldn't be set during initialization.

1) Modified the `DynamicEmbeddedDocument.__init__` to proactively check for reserved attributes (implemented as properties) before calling the parent class initializer. If a reserved attribute is detected, it raises a clear FiftyoneDocumentException with a descriptive message.

2) Added comprehensive unit tests to verify this behavior using the Detection class as a test case.

* Local testing (✅ )
* Unit test (✅ )

Example:
```
In [1]: import fiftyone as fo

In [2]: fo.Detection(has_mask="none")
FiftyoneDocumentException                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 fo.Detection(has_mask="none")

File ~/workspace/fiftyone/fiftyone/core/odm/embedded_document.py:58, in DynamicEmbeddedDocument.__init__(self, *args, **kwargs)
     55 for key, value in kwargs.items():
     56     # Check if the key exists and is a property
     57     if hasattr(self.__class__, key) and isinstance(getattr(self.__class__, key), property):
---> 58        raise FiftyoneDocumentException(f"Attribute {key} already exists for {self.__class__.__name__}")
     60 super().__init__(*args, **kwargs)
     61 self.validate()

FiftyoneDocumentException: Attribute has_mask already exists for Detection
```

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced a new exception class `DynamicEmbeddedDocumentException` for improved error handling related to dynamic document operations.
	- Enhanced document initialization with stricter validation for reserved attributes.

- **Tests**
	- Added a new test class for validating the behavior of the `Detection` class.
	- Verified exception handling for reserved attributes during object initialization and attribute setting.
	- Ensured correct extraction of attribute names from exceptions.

### DIFF
--- a/fiftyone/core/odm/embedded_document.py
+++ b/fiftyone/core/odm/embedded_document.py
@@ -11,12 +11,6 @@ import mongoengine
 from .document import DynamicMixin, MongoEngineBaseDocument
 
 
-class FiftyOneDynamicDocumentException(Exception):
-    """Exception raised when an error occurs in a document operation."""
-
-    pass
-
-
 class BaseEmbeddedDocument(MongoEngineBaseDocument):
     """Base class for documents that are embedded within other documents and
     therefore are not stored in their own collection in the database.
@@ -37,6 +31,12 @@ class EmbeddedDocument(BaseEmbeddedDocument, mongoengine.EmbeddedDocument):
         self.validate()
 
 
+class DynamicEmbeddedDocumentException(Exception):
+    """Exception raised when an error occurs in a dynamic document operation."""
+
+    pass
+
+
 class DynamicEmbeddedDocument(
     DynamicMixin,
     BaseEmbeddedDocument,
@@ -47,8 +47,6 @@ class DynamicEmbeddedDocument(
     database.
 
     Dynamic documents can have arbitrary fields added to them.
-
-    Raises FiftyoneDocumentException if a kwarg attribute is reserved.
     """
 
     meta = {"abstract": True, "allow_inheritance": True}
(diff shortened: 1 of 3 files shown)


fiftyone/core/odm/embedded_document.py
```diff
@@ -40,11 +45,22 @@ class DynamicEmbeddedDocument(
...
 
     meta = {"abstract": True, "allow_inheritance": True}
 
     def __init__(self, *args, **kwargs):
+        for key, value in kwargs.items():
+            # Skip MongoDB internal fields
+            if key.startswith('_'):
+                continue
+
+            # Check if the key exists and is a property
```
brimoor, line 60: Looks like you don't have [pre-commit hooks](https://github.com/voxel51/fiftyone/blob/develop/CONTRIBUTING.md#pre-commit-hooks) installed

fiftyone/core/odm/embedded_document.py
```diff
@@ -40,11 +45,22 @@ class DynamicEmbeddedDocument(
...
 
     Dynamic documents can have arbitrary fields added to them.
+
+    Raises FiftyoneDocumentException if a kwarg attribute is reserved.
     """
 
     meta = {"abstract": True, "allow_inheritance": True}
 
     def __init__(self, *args, **kwargs):
+        for key, value in kwargs.items():
```
brimoor, line 55: Have you benchmarked this implementation? While validation is nice, in the context of something I/O intensive like iterating over a dataset or adding a bunch of samples to a dataset, this would introduce millions of extra iterations, string comparisons, etc.

Doing validation to prevent bad data from being added to the database is a necessary expense. But in this case, the damage is already done. We're raising a more informative error message, but the user still can't work with the dataset (in the case where we're iterating over an existing dataset).

A faster alternative if we really want this validation is to [use try-except](https://docs.python.org/2/faq/design.html#how-fast-are-exceptions).

brimoor (changes requested):

fiftyone/core/odm/embedded_document.py
minhtuev: Ah, the performance consideration makes sense, I haven't done the benchmarking but I think it's safer to use `try-except`. I did think about it but I wanted to do more precise validation by iterating through the key-value pairs.

fiftyone/core/odm/embedded_document.py
minhtuev: Haha oops this is a new conda env I just created :)

Mizokuiam:
Thank you for raising this issue! I'll look into it and try to help if I can.

fiftyone/core/odm/embedded_document.py
```diff
@@ -40,13 +47,37 @@ class DynamicEmbeddedDocument(
...
+            if match:
+                key = match.group("attribute1") or match.group("attribute2")
+                # Check if the key exists and is a property
+                if (
+                    key is not None
+                    and hasattr(self.__class__, key)
+                    and isinstance(getattr(self.__class__, key), property)
+                ):
+                    raise FiftyOneDynamicDocumentException(
+                        f"Attribute {key} already exists for {self.__class__.__name__}"
```
brimoor, line 77: How about something like this for the error message?

```
Invalid attribute name '{key}'. '{key}' is a reserved keyword for {self.__class__.__name__} objects
```

For consistency with our nomenclature [here](https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/odm/utils.py#L157-L175), for example.

minhtuev: Done ✅

brimoor (changes requested):
I just tried this like so:

```py
import fiftyone as fo

fo.Detection(has_mask="not allowed")
```

and didn't see the informative error message because the exception was this:

```
AttributeError: can't set attribute
```

which is the same thing I see if I do this:

```py
d = fo.Detection()
d.has_mask = "not allowed"
```

minhtuev:
@brimoor : that's a little odd, I checked locally and my error came out as expected

```
In [1]: import fiftyone as fo
   ...:
   ...: fo.Detection(has_mask="not allowed")
File ~/workspace/fiftyone/fiftyone/core/odm/embedded_document.py:58, in DynamicEmbeddedDocument.__init__(self, *args, **kwargs)
     57 try:
---> 58     super().__init__(*args, **kwargs)
     59     self.validate()

File /opt/anaconda3/envs/voxel311/lib/python3.11/site-packages/mongoengine/document.py:90, in EmbeddedDocument.__init__(self, *args, **kwargs)
     89 def __init__(self, *args, **kwargs):
---> 90     super().__init__(*args, **kwargs)
     91     self._instance = None

File /opt/anaconda3/envs/voxel311/lib/python3.11/site-packages/mongoengine/base/document.py:142, in BaseDocument.__init__(self, *args, **values)
    141     for key, value in dynamic_data.items():
--> 142         setattr(self, key, value)
    144 # Flag initialised

File /opt/anaconda3/envs/voxel311/lib/python3.11/site-packages/mongoengine/base/document.py:209, in BaseDocument.__setattr__(self, name, value)
    207     super().__setattr__("_created", False)
--> 209 super().__setattr__(name, value)

AttributeError: property 'has_mask' of 'Detection' object has no setter

During handling of the above exception, another exception occurred:

FiftyOneDynamicDocumentException          Traceback (most recent call last)
Cell In[1], line 3
      1 import fiftyone as fo
----> 3 fo.Detection(has_mask="not allowed")

File ~/workspace/fiftyone/fiftyone/core/odm/embedded_document.py:76, in DynamicEmbeddedDocument.__init__(self, *args, **kwargs)
     70     # Check if the key exists and is a property
     71     if (
     72         key is not None
     73         and hasattr(self.__class__, key)
     74         and isinstance(getattr(self.__class__, key), property)
     75     ):
---> 76         raise FiftyOneDynamicDocumentException(
     77             f"Invalid attribute name '{key}'. '{key}' is a reserved keyword for {self.__class__.__name__} objects"
     78         )
     79 # If the error is not due to a reserved attribute, raise the original error
     80 raise e

FiftyOneDynamicDocumentException: Invalid attribute name 'has_mask'. 'has_mask' is a reserved keyword for Detection objects

minhtuev:
@brimoor : I have added that message to exception handling and also updated attribute setter `__setattr__` with the same exception handling as well

brimoor (approved):
LGTM, thanks for diving in and improving the error handling here @minhtuev 🥇

Issue #4961, opened 2024-10-22
[FR] Support non-binary/grayscale masks in `fo.Detection`


Laurent2916:
### Proposal Summary

Currently [`fo.Detection`](https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/labels.py#L391-L414) supports:

> an instance segmentation mask for the detection within its bounding box, which should be a 2D binary or 0/1 integer numpy array

In practice it is possible to store a [0, 1] float numpy array (i.e. a continuous mask), since there doesn't seem to be checks in place to prevent that. This is very useful when using/training segmentation models that operate on such continuous masks (i.e. non binary/thresholded masks) like MVANet, RMBG-1.4, BiRefNet, etc.

I was wondering if supporting continuous mask would be feasible in Fiftyone, because continuous masks currently are thresholded (at 0) in the UI, which often makes visualizing them difficult. Having the ability to enable, disable and change the threshold value in the UI would be very useful. 

Note: Using continuous masks probably breaks some methods such as `to_polyline` and `to_segmentation`, though I don't use them often so I don't know.

### What areas of FiftyOne does this feature affect?

-   [x] App: FiftyOne application
-   [x] Core: Core `fiftyone` Python library
-   [ ] Server: FiftyOne server

### Willingness to contribute

The FiftyOne Community welcomes contributions! Would you or another member of your organization be willing to contribute an implementation of this feature?

-   [ ] Yes. I can contribute this feature independently
-   [x] Yes. I would be willing to contribute this feature with guidance from the FiftyOne community
-   [ ] No. I cannot contribute this feature at this time

deltheil:
> [...] continuous masks currently are thresholded (at 0) in the UI [...] Having the ability to enable, disable and change the threshold value in the UI would be very useful.

I had a look at the [Looker package](https://github.com/voxel51/fiftyone/tree/develop/app/packages/looker) which takes care of this. So roughly:

- The FiftyOne server graphql API returns masks as they are stored in Mongo, i.e. base64-encoded and zlib compressed numpy format (e.g. see `mainSampleQuery`)
- They are deserialized client-side part of [numpy.ts](https://github.com/voxel51/fiftyone/blob/4b1d9cae3f162aae1a6fb286da9b2f4ef8908226/app/packages/looker/src/numpy.ts#L172-L174)
[...]

PR #5369, opened 2025-01-10
add detection mask threshold to dataset app config

tom-vx51:

This PR adds a configurable detection mask threshold to the dataset app config. For masks containing non-binary values (e.g. a uint8 mask with values in [0,255]), FiftyOne currently treats all non-zero elements as masked; this effectively acts as a default mask threshold of 0. This change allows for setting a custom threshold, which causes masks to render only for mask values exceeding the threshold.

Tested via local app development. Modifying `dataset.app_config.mask_threshold` results in conditional masking based on the configured threshold (default = 0).

        notes for FiftyOne users.

Add support for configuring mask thresholds for non-binary detection masks.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Added a configurable mask threshold for detection masks
	- Users can now set a custom threshold for rendering non-binary masks

- **Documentation**
	- Updated user guide with instructions on configuring detection mask threshold
	- Added explanation of how mask threshold affects mask rendering

- **Technical Improvements**
	- Introduced `maskThreshold` across multiple components and interfaces
	- Enhanced dataset configuration options for mask visualization


sashankaryal (approved):
lgtm! 🚀

fiftyone/core/odm/dataset.py
```diff
@@ -553,6 +555,7 @@ class DatasetAppConfig(EmbeddedDocument):
     sidebar_groups = ListField(
         EmbeddedDocumentField(SidebarGroupDocument), default=None
     )
+    mask_threshold = IntField(default=0)
```
brimoor, line 558: I *think* that the App gets all its data via `dataset.app_config`, which makes this concern moot.

But... just be aware that since there's no database migration here, if there were some code path that directly pulled a dataset's app config from the database, then this would be missing from the dict.

docs/source/user_guide/using_datasets.rst
```diff
@@ -477,6 +477,32 @@ You can disable frame filtering for a video dataset as follows:
     Did you know? You can also globally disable frame filtering for all video
     datasets via your :ref:`App config <configuring-fiftyone-app>`.
 
+.. _dataset-app-config-detection-mask-threshold:
+
+Configure detection mask threshold
```
brimoor, line 482: I believe this only applies to instance segmentations, right? It would be nice to link to this section of the docs from the section that describes instance segmentations:
https://docs.voxel51.com/user_guide/using_datasets.html#instance-segmentations

Speaking of, note that this section currently provides the following guidance:

> Masks stored directly in the database must be 2D numpy arrays containing either booleans or 0/1 integers that encode the extent of the instance mask within the [bounding_box](https://docs.voxel51.com/api/fiftyone.core.labels.html#fiftyone.core.labels.Detection.bounding_box) of the object.
>
> For masks stored on disk, the [mask_path](https://docs.voxel51.com/api/fiftyone.core.labels.html#fiftyone.core.labels.Detection.mask_path) attribute should contain the file path to the mask image. We recommend storing masks as single-channel PNG images, where a pixel value of 0 indicates the background (rendered as transparent in the App), and any other value indicates the object.

Does that section need to be updated in light of this proposed change?

Also note that there is some connection between how masks are rendered in the App and the semantics of other operations, such as when evaluating instance segmentations:
https://github.com/voxel51/fiftyone/blob/60971f8b6ba80cf48629fb3b6f4a32a3a44dcd19/fiftyone/utils/eval/detection.py#L125-L127

If `evaluate_detections()` treats mask definitions differently than the App does, then it may be misleading to users why the quantitative evaluation metrics don't align with what they see in the App. This is tricky though, because `app_config.mask_threshold` is an App-centric configuration setting, while `evaluate_detections()` is an SDK concept.

brimoor (changes requested):

docs/source/user_guide/using_datasets.rst
brimoor: Update: after tracing down how `evaluate_detections()` handles instance masks, they are currently parsed specifically here:
https://github.com/voxel51/eta/blob/7280590da6b03a76d21d0c3d24aa2067b091985b/eta/core/image.py#L1089

@mwoodson1 is also actively working on an alternative implementation ub #5283 where instance mask arithmetic would instead be implemented like this:
https://github.com/voxel51/fiftyone/blob/72d933ba344fbd30dca8261b11155f88514eaf4a/fiftyone/utils/iou.py#L575-L576

So I believe the SDK is effectively hard-coded to `mask_threshold=0` for the purposes of evaluating instance segmentations.

brimoor: Another place where instance segmentation semantics matter on the backend is [`Detections.to_segmentation()`](https://github.com/voxel51/fiftyone/blob/60971f8b6ba80cf48629fb3b6f4a32a3a44dcd19/fiftyone/core/labels.py#L514), which ends up treating all instance masks with `mask_threshold=0` because of [this line](https://github.com/voxel51/eta/blob/7280590da6b03a76d21d0c3d24aa2067b091985b/eta/core/image.py#L976).

fiftyone/core/odm/dataset.py
tom-vx51: Yeah confirmed that this works for existing datasets. The app does a nullish check and defaults to 0 if it's not contained in the config.

benjaminpkane:
I do think considering the pattern established by `Dataset.default_mask_targets` and `Dataset.mask_targets` may be valuable. As Brian mentioned, this may not be strictly an App setting. And a global value may become restrictive

For example, `Dataset.default_instance_mask_threshold` and `Dataset.instance_mask_thresholds`

docs/source/user_guide/using_datasets.rst
tom-vx51: Good notes, will update the docs.
> If evaluate_detections() treats mask definitions differently than the App does, then it may be misleading to users why the quantitative evaluation metrics don't align with what they see in the App. This is tricky though, because app_config.mask_threshold is an App-centric configuration setting, while evaluate_detections() is an SDK concept.

This change was only meant to be a visual configuration, but I agree with your concern. I admittedly haven't tested this behavior in the SDK and was trusting the experience in [this comment](https://github.com/voxel51/fiftyone/issues/4961#issuecomment-2494939562). Would you like any SDK changes as part of this PR? Or additional documentation? Or something else?

brimoor: > Would you like any SDK changes as part of this PR? Or additional documentation? Or something else?
Depends on the motivation for the change. Can you share some additional context?

brimoor: If the user is only interested in visualizing but not evaluating data, then we could just ship this as-is after updating the documentation as mentioned so that its clear that `dataset.app_config.mask_threshold` is a **visual-only** change.

If the user is also likely interested in evaluating these instance segmentations, or if our primary motivation is internal development, then I'd want to understand a bit more.

The main use case I could imagine here would be that users are working with instance segmentation models that generate probabilistic masks that are represented initially as float values in `[0, 1]`, which would then be converted to `uint8(0, 255)` for storage reasons. In this case, a user might want to evaluate different candidate thresholds for binarizing the mask. In this context, what I'd want is:
1. Ability to visualize probabilistic instance masks as heatmaps (like our `Heatmap` label type)
2. Ability to visualize different candidate `mask_threshold` values in the App via a slider or similar
    - Point 1 could be a special case of this where the mask threshold is disabled
3. Ability to define an appropriate `mask_threshold` for instance segmentation field(s), which would then be used subsequently by the App and via the SDK
    - As [Ben mentions](https://github.com/voxel51/fiftyone/pull/5369#issuecomment-2583313425), the most natural way to support that would be `dataset.default_instance_mask_threshold` and `dataset.instance_mask_thresholds`, for consistency with how mask targets are stored today

We could choose to *not* build Point 3, instead recommending to the user to use 1 and 2 to explore different thresholds, and then once you've picked one, binarize your instance masks and store these on your FO dataset for downstream tasks like evaluation.

mdecarlo: @brimoor , the user [is only interested in visualizing data](https://github.com/voxel51/fiftyone/issues/4961#issuecomment-2628844098). However, I want to push back on this way forward:

> then we could just ship this as-is after updating the documentation as mentioned so that its clear that dataset.app_config.mask_threshold is a visual-only change

Three questions:
1. Do we have any precedence for "visual-only" segmentations (or detections, or classifications), such that we can feel confident users won't get confused or frustrated given only a fragment buried in documentation? 
2. Presumably non-binary masks could be passed through to in-App model evaluations, including via the Model Eval Panel. Let's say the user sets the threshold to 75 on the dataset. How would model evaluation behave? And assuming it works in some justifiable way, wouldn't the results be misleading when juxtaposed against masks visualized in the App?
3. If any piece of my concern in (2)☝️ is valid, then can you think of other in-App operators or panels with analogous concerns?

brimoor: There is no precedent for a setting like `dataset.app_config.mask_threshold` that would change rendering in the App but wouldn't affect how the data is treated in the SDK.

My intuition is that if we added this, then it's only a matter of time before a user requests 

The most use 

It sounds like the use case here is to 
If we add `dataset.app_config.mask_threshold` as an App-only parameter, then it's only a matter of time before

brimoor: There is no precedent for a setting like `dataset.app_config.mask_threshold` that would change rendering behavior in the App but **wouldn't** affect how the data is treated in the SDK. This would be a first.

If the user evaluates data that they've been visualizing with a `mask_threshold > 1`, then yes the results will likely be misleading. There are other situations as well. I pointed out [Detections.to_segmentation()](https://github.com/voxel51/fiftyone/blob/60971f8b6ba80cf48629fb3b6f4a32a3a44dcd19/fiftyone/core/labels.py#L514) above in this thread. Another case off the top of my head is [this method](https://github.com/voxel51/fiftyone/blob/ad05c8df9e6aee47116f40fd166b09738dcee5b5/fiftyone/utils/cvat.py#L6469), which is used to upload instance segmentations to CVAT; it would not do the right thing when it encounters pixel values >1.

Basically it is unsafe to store instance segmentations with pixel values >1 in FO today because that violates our documented convention.

The only protection here would be security through obscurity. Users would have to know what they're doing in order to get into this threshold exploration mode.

Since the use case here is to help users pick a threshold for their model, the most useful thing IMO would be to have an interactive slider in the App where I could try different thresholds to arrive at the best setting for my dataset. I would also be interested in seeing a heatmap-style rendering where transparency is used to encode values in `[0, 255]`. This would correspond to `mask_threshold =- 1`, say.

Once I pick a threshold, I would need to binarize my data by applying the threshold and converting the masks in-place into the expected 0-1 format in order for the rest of the App's workflows to be supported.

brimoor: The idiomatic thing to do in FO would be to add a utility like  `fiftyone.utils.labels.binarize_instances()` that would allow for programmatically converting continuous masks to binary.

When using the interactive threshold exploration UX in the App that I described above, I'd love to have a way to trigger the binarization.

brimoor: I'm now thinking the best path forward here to support this FR would be to **not** add a persistent `dataset.app_config.mask_threshold` setting. Instead, it would be most logical to add the interactive threshold exploration UX.

By default, everything >0 would be "object", for consistency with how the SDK would treat the data.

However, if you've stored masks with values in `[0, 255]`, then you could drag to explore different thresholds in the App and then finalize the binarization if you found a threshold that you prefer to `mask_threshold=1`.

PR #5376, opened 2025-01-11
Added support for batch size and multi processing to evaluate_detections

minhtuev:

This PR introduces parallel processing capabilities to the detection evaluation functionality in FiftyOne. Key changes include:

1) Implemented parallel processing of samples using ProcessPoolExecutor or ThreadPoolExecutor
2) Added configurable number of workers (defaults to CPU count - 1)
3) Maintained proper batch-based sample downloading through FiftyOne's iter_samples(), with batch_size exposed as an argument 

The parallelization significantly improves performance for large datasets while maintaining the same accuracy and output format as the original

* Local testing 

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/eval/detection.py
+++ b/fiftyone/utils/eval/detection.py
@@ -139,8 +139,8 @@ def evaluate_detections(
         progress (None): whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
-        batch_size (None): the batch size at which to process samples. By
-            default, all samples are processed in a single (1) batch
+        batch_size (None): the batch size at which to process samples. If None,
+            use the default batching strategy (usually latency)
         num_workers (None): number of parallel workers. Defaults to CPU count - 1
         **kwargs: optional keyword arguments for the constructor of the
             :class:`DetectionEvaluationConfig` being used
--- a/fiftyone/utils/eval/detection.py
+++ b/fiftyone/utils/eval/detection.py
@@ -5,7 +5,7 @@ Detection evaluation.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
+from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
 import inspect
 import itertools
@@ -45,7 +45,6 @@ def evaluate_detections(
     progress=None,
     batch_size=None,
     num_workers=None,
-    executor_type="process",
     **kwargs,
 ):
     """Evaluates the predicted detections in the given samples with respect to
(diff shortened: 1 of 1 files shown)


fiftyone/utils/eval/detection.py
```diff
@@ -176,45 +188,67 @@ def evaluate_detections(
...
-            if processing_frames and save:
-                doc[tp_field] = tp
-                doc[fp_field] = fp
-                doc[fn_field] = fn
-
-        if save:
-            sample[tp_field] = sample_tp
-            sample[fp_field] = sample_fp
-            sample[fn_field] = sample_fn
+    with contextlib.ExitStack() as stack:
```
minhtuev, line 205: Need to clean up this bit

kaixi-wang:
How did you test this/on what platform? Pymongo is not fork safe and since fiftyone uses a single global mongo client, iterating over samples could result in deadlock. https://www.mongodb.com/docs/languages/python/pymongo-driver/current/faq/#can-i-use-pymongo-with-multiprocessing-

swheaton:
Also we are pausing this for now as discussed the other day 😊

minhtuev:
@kaixi-wang : I am running this locally on my mac computer against local & remote DB. The samples are fetched on the main thread and the actual IOU computation is parallelized, so I don't think there is a danger to Pymongo.

fiftyone/utils/eval/detection.py
```diff
@@ -176,45 +183,50 @@ def evaluate_detections(
...
+            else:
+                docs = [sample]
+
+            future = executor.submit(
+                _process_sample, docs, eval_method, eval_key, processing_frames
+            )
+            futures.append((future, sample))
+
+        # Collect results
+        for future, sample in futures:
```
minhtuev, line 219: We probably should define a constant somewhere so thaat we can collect the results earlier without having to maintain a large list

minhtuev:
Related PR:
https://github.com/voxel51/fiftyone/pull/5362/files

fiftyone/utils/eval/detection.py
```diff
@@ -135,6 +139,9 @@ def evaluate_detections(
         progress (None): whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
+        batch_size (None): the batch size at which to process samples. By
+            default, all samples are processed in a single (1) batch
```
brimoor, line 143: It's not true that the default is a batch size of 1. `iter_samples(autosave=True)` uses a batching strategy by default:
https://github.com/voxel51/fiftyone/blob/67bdf7e60a752b2686c2050d4f7ec16141fef988/fiftyone/core/view.py#L500-L516 

The default batcher sends requests every 0.2 seconds:
https://github.com/voxel51/fiftyone/blob/67bdf7e60a752b2686c2050d4f7ec16141fef988/fiftyone/core/config.py#L159-L182

minhtuev: Ah, so if batch_size is not passed, we use latency strategy by default 👍

PR #5375, opened 2025-01-10
AutoComplete and DropDown operator View improvements

ritch:
## What changes are proposed in this pull request?

 - Changes the behavior of a dropdown to match the AutoCompleteView, but with dropdown semantics (eg. no user input)
 - Adds `values_only` param to `AutoCompleteView` for ensuring only the values are provided in the corresponding property (eg. `list()`).
 - Coerce empty string and arrays to `null` / `None` for `str` and `list` properties.

## How is this patch tested? If it is not, please explain why.

Using a purpose built test operator

## Summary by CodeRabbit

- **New Features**
	- Enhanced Autocomplete component with more flexible selection options
	- Added support for multiple dropdown selections
	- Introduced value coercion for input handling

- **Improvements**
	- Added options to allow user input in dropdowns
	- Implemented duplicate selection control
	- Improved input validation and processing

- **Technical Updates**
	- Updated component logic for more robust input management
	- Expanded schema-based input handling capabilities

PR #5379, opened 2025-01-13
Allow lists wherever possible in builtin operators

brimoor:
## Change log

- Updates the builtin operators to use `list()` to allow users to act on multiple fields/entities wherever it makes sense
- Adopts the cleaner syntax made possible by https://github.com/voxel51/fiftyone/pull/5375

### DIFF
--- a/plugins/operators/__init__.py
+++ b/plugins/operators/__init__.py
@@ -244,30 +244,28 @@ def _clone_sample_field_inputs(ctx, inputs):
     for key in field_keys:
         field_selector.add_choice(key, label=key)
 
-    inputs.enum(
+    field_prop = inputs.str(
         "field_name",
-        field_selector.values(),
         label="Sample field",
-        description=(
-            "The field to copy. You can use `embedded.field.name` to clone "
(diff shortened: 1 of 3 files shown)


plugins/operators/__init__.py
```diff
@@ -669,28 +641,33 @@ def _clear_sample_field_inputs(ctx, inputs):
...
-    field_keys = list(schema.keys())
-    field_selector = types.AutocompleteView()
-    for key in field_keys:
-        field_selector.add_choice(key, label=key)
+    field_names = ctx.params.get("field_names", None) or []
 
-    field_prop = inputs.enum(
-        "field_name",
-        field_selector.values(),
+    field_choices = types.AutocompleteView(multiple=True)
```
ritch, line 646: Suggestion: add `allow_duplicates=False` which prevents adding a given value multiple times. Also `multiple` is not needed since the setting is inferred from the schema being a `list`.

```
types.AutocompleteView(allow_duplicates=False)
```

ritch (approved):

PR #5389, opened 2025-01-15
Add custom metric to Summary in Model Evaluation panel

manushreegangwar:

Adds `Custom Metrics` to Summary in Model Evaluation panel

image

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
@@ -1,3 +1,4 @@
+import _ from "lodash";
 import { Dialog } from "@fiftyone/components";
 import { editingFieldAtom, view } from "@fiftyone/state";
 import {
@@ -196,7 +197,6 @@ export default function Evaluation(props: EvaluationProps) {
   const evaluationTimestamp = evaluationInfo.timestamp;
   const evaluationConfig = evaluationInfo.config;
   const evaluationMetrics = evaluation.metrics;
-  const evaluationCustomMetrics = evaluation.custom_metrics;
   const evaluationType = evaluationConfig.type;
   const evaluationMethod = evaluationConfig.method;
   const compareEvaluationInfo = compareEvaluation?.info || {};
@@ -204,8 +204,6 @@ export default function Evaluation(props: EvaluationProps) {
   const compareEvaluationTimestamp = compareEvaluationInfo?.timestamp;
   const compareEvaluationConfig = compareEvaluationInfo?.config || {};
   const compareEvaluationMetrics = compareEvaluation?.metrics || {};
-  const compareEvaluationCustomMetrics =
-    compareEvaluation?.custom_metrics || {};
   const compareEvaluationType = compareEvaluationConfig.type;
   const isObjectDetection = evaluationType === "detection";
   const isClassification = evaluationType === "classification";
@@ -468,21 +466,8 @@ export default function Evaluation(props: EvaluationProps) {
           : false,
       hide: !showTpFpFn,
     },
+    ...formatCustomMetricRows(evaluation, compareEvaluation),
   ];
-
-  for (const key in evaluationCustomMetrics) {
-    const cm_keys = Object.keys(evaluationCustomMetrics[key]);
-    summaryRows.push({
-      id: key,
-      property: evaluationCustomMetrics[key][cm_keys[0]].toString(),
-      value: evaluationCustomMetrics[key][cm_keys[1]],
-      compareValue: compareEvaluationCustomMetrics[key][cm_keys[1]],
-      lesserIsBetter: true,
-      filterable: false,
-      active: false,
-      hide: false,
-    });
-  }
 
   const perClassPerformance = {};
   for (const key in evaluation?.per_class_metrics) {
@@ -1774,3 +1759,52 @@ type PLOT_CONFIG_TYPE = {
 type PLOT_CONFIG_DIALOG_TYPE = PLOT_CONFIG_TYPE & {
   open?: boolean;
 };
+
+type CustomMetric = {
+  label: string;
+  value: any;
+};
+
+type CustomMetrics = {
+  [operatorName: string]: CustomMetric;
+};
+
+type SummaryRow = {
+  id: string;
+  property: string;
+  value: any;
+  compareValue: any;
+  lesserIsBetter: boolean;
+  filterable: boolean;
+  active: boolean;
+  hide: boolean;
+};
+
+function formatCustomMetricRows(evluationMetrics, comparisonMetrics) {
+  const results = [] as SummaryRow[];
+  const customMetrics = _.get(
+    evluationMetrics,
+    "custom_metrics",
+    {}
+  ) as CustomMetrics;
+  for (const [operatorName, customMetric] of Object.entries(customMetrics)) {
+    const compareValue = _.get(
+      comparisonMetrics,
+      `custom_metrics.${operatorName}.value`,
+      null
+    );
+    const hasComparisonValue = compareValue !== null;
+
+    results.push({
+      id: operatorName,
+      property: customMetric.label,
+      value: customMetric.value,
+      compareValue,
+      lesserIsBetter: true, // TODO - this needs to be provided from the customMetrics object
+      filterable: false,
+      active: false,
+      hide: !hasComparisonValue,
+    });
+  }
+  return results;
+}
--- a/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
@@ -465,6 +465,12 @@ export default function Evaluation(props: EvaluationProps) {
           : false,
       hide: !showTpFpFn,
     },
+    {
+      id: "custom_metrics",
+      property: "Custom Metrics",
+      value: evaluationMetrics.custom_metrics,
+      compareValue: compareEvaluationMetrics.custom_metrics,
+    },
   ];
 
   const perClassPerformance = {};
--- a/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
@@ -1793,7 +1793,7 @@ function formatCustomMetricRows(evluationMetrics, comparisonMetrics) {
       `custom_metrics.${operatorName}.value`,
       null
     );
-    const hasComparisonValue = compareValue !== null;
+    const hasOneValue = customMetric.value !== null || compareValue !== null;
 
     results.push({
       id: operatorName,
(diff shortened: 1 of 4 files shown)


fiftyone/operators/builtins/panels/model_evaluation/__init__.py
```diff
@@ -146,6 +146,9 @@ def get_mar(self, results):
         except Exception as e:
             return None
 
+    def get_custom_metrics(self, results):
+        return results.custom_metrics
```
brimoor, line 150: `results.custom_metrics` will be `None` if there are no custom metrics.

Do you want to return `results.custom_metrics or {}` here? Or is `json.dumps(None)` okay for the frontend?

fiftyone/operators/builtins/panels/model_evaluation/__init__.py
```diff
@@ -11,7 +11,7 @@
 import traceback
 
 import numpy as np
-
+import json
```
brimoor, line 14: nit: [style guide](https://github.com/voxel51/fiftyone/blob/develop/STYLE_GUIDE.md#imports) says that the new import should go here:

```py
from collections import defaultdict, Counter
import json
import os
import traceback
```

(alphabetized within in the builtins block)

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
```diff
@@ -465,6 +465,12 @@ export default function Evaluation(props: EvaluationProps) {
           : false,
       hide: !showTpFpFn,
     },
+    {
+      id: "custom_metrics",
```
brimoor, line 469: Defer to @ritch to review the frontend implementation here.

From a UX standpoint, I propose that:
- Each custom metric gets its own row in the Summary table
- Therefore, if there are no custom metrics, nothing is shown

brimoor (changes requested):

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
manushreegangwar: I also prefer this design but would need @ritch's help on the frontend part for this.

fiftyone/operators/builtins/panels/model_evaluation/__init__.py
manushreegangwar: I updated this to return `None` as `{}`. `json.dumps(None)` returns `'null'`.

fiftyone/operators/builtins/panels/model_evaluation/__init__.py
manushreegangwar: Done. Wondering why this is not handled by the pre-commit hooks?

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
ritch: You would need to build an object in this array **per custom metric**.

```js
{
  id: "unique_id_per_custom_metric",
  property: "Human readable label",
  value: evaluationMetrics.custom_metrics['unique_id_per_custom_metric'], // value to display
  compareValue: compareEvaluationMetrics.custom_metrics['unique_id_per_custom_metric'], // value to compare with
  lesserIsBetter: true, // needs to come from the evaluation results & custom metric
  filterable: true, // safe to say this is always false?
  active: false, // always false for custom metric
  hide: false, // this should always be false
}
```

ritch: I think this means we need to revisit the structure of `evaluation_results.custom_metrics`. The info above needs to be included (eg. "lesserIsBetter").

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
```diff
@@ -467,6 +470,19 @@ export default function Evaluation(props: EvaluationProps) {
...
+      id: key,
+      property: evaluationCustomMetrics[key].label.toString(),
+      value: evaluationCustomMetrics[key].value,
+      compareValue: compareEvaluationCustomMetrics[key].value,
+      lesserIsBetter: true,
+      filterable: true,
+      active: false,
+      hide: false,
+    });
+  }
```
manushreegangwar, line 484: @ritch could you take a look at this?

fiftyone/utils/eval/base.py
```diff
@@ -181,6 +185,16 @@ def __init__(
         self.missing = missing
         self.custom_metrics = custom_metrics
 
+    @property
+    def additional_metrics(self):
```
kaixi-wang, line 189: Why switch from custom_metrics to additional_metrics?

manushreegangwar: I need two types of output dicts for custom metrics: 

`self.custom_metrics` has more fields because they are needed by the app.

`additional_metrics` is added for SDK with minimal information.

I don't want to use `result._custom_metrics` and `result.custom_metrics`  because the former is updated outside of this class. I can change this from a property to a function -- something like `custom_metrics_report`.

brimoor: I think we should pick a format for `results.custom_metrics` such that both SDK and App use cases have what they need.

Operators have a few concepts:
- `name`: operator name, eg `mae`
- `uri`: the operator's URI, eg `@voxel51/example-metrics/mae`
- `label`: a human-readable name, eg `Mean Absolute Error`

The reason I chose `{label: value}` format for `results.custom_metrics` is that I felt that the human-readable `label` is the right thing to show both in the App, and also when when interesting with the results in the SDK via `print(results.custom_metrics)`.

Note that `results.config.custom_metrics` is also available, which contains the list of URIs `[uri, ...]` of the operators that were run.

However, note **also** that evaluations are persisted in the database, while plugins can be uninstalled, renamed, etc. at any time. So it is possible that the user is viewing an evaluation for which the operator no longer exists under the name/URI/label on which it was computed. That's why I put `try: except` around everything in https://github.com/voxel51/fiftyone/pull/5279.

fiftyone/utils/eval/base.py
```diff
@@ -58,7 +59,10 @@ def compute_custom_metrics(self, samples, eval_key, results):
                     samples, eval_key, results, **kwargs or {}
                 )
                 if value is not None:
-                    results.custom_metrics[operator.config.label] = value
+                    results.custom_metrics[operator.config.name] = {
```
brimoor, line 62: Why is operator `name` needed by the frontend? Can't this be avoided?

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
```diff
@@ -1757,3 +1759,52 @@ type PLOT_CONFIG_TYPE = {
...
+      null
+    );
+    const hasComparisonValue = compareValue !== null;
+
+    results.push({
+      id: operatorName,
+      property: customMetric.label,
+      value: customMetric.value,
+      compareValue,
+      lesserIsBetter: true, // TODO - this needs to be provided from the customMetrics object
```
brimoor, line 1803: What does `lesserIsBetter` do?

fiftyone/utils/eval/base.py
manushreegangwar: I can update the app side of things to use combination of `results.config.custom_metrics` and `results.custom_metrics` which I will restore to `{label: value}`

manushreegangwar: > However, note also that evaluations are persisted in the database, while plugins can be uninstalled, renamed, etc. at any time. So it is possible that the user is viewing an evaluation for which the operator no longer exists under the name/URI/label on which it was computed. That's one reason I put try: except around everything in https://github.com/voxel51/fiftyone/pull/5279. Breaking changes to "custom" things should not cause "builtin" things to fail.

I understand but in the app we don't try to reload the operator, so even if the operator is not present, the user should be able to view the computed value.

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
ritch:
```
const showTrophy = lesserIsBetter
                      ? difference < 0
                      : difference > 0;
```

Its used to show the trophy icon. It could also be used in the future to show red vs green when displaying numbers.

ritch:
Proposal for a CustomMetric class that would provide the right information for the SDK and App:

```python
CustomMetric(
    id="example_metric",
    operator_uri="@voxel51/metric-examples/example_metric",
    value=42,
    lower_is_better=True,
    metric_property=types.Property(
        types.Number(float=True),
        label="Example metric",
        view=types.NumberView(title="This will show as a tooltip")
    )
)
```

fiftyone/utils/eval/regression.py
```diff
@@ -425,6 +426,8 @@ def __init__(
         self.ids = ids
         self.missing = missing
         self.custom_metrics = custom_metrics
+        # Stores configs for custom metrics with a return value (used by the model evaluation panel).
+        self.custom_metrics_config = []
```
ritch, line 430: I think we should merge both `custom_metrics_config` and `custom_metrics` and use the style/shape that I mentioned [here](https://github.com/voxel51/fiftyone/pull/5389#issuecomment-2596844520).

manushreegangwar: Can you see [this](https://github.com/voxel51/fiftyone/pull/5389#discussion_r1918805950) comment by @brimoor?

If we want keep  `results.custom_metrics` in a form that easy for SDK users, changing this to the format you are suggesting won't work.

Option 1 is to remove `self.custom_metrics_config` and add the `CustomMetric` objects to a new attribute under `results`.

Option 2 (which I prefer) is to keep just one `self.custom_metrics` variable that stores `CustomMetric` objects and expect the SDK users to use a method that converts `self.custom_metrics` into a more readable form.

ritch: Option 2 is simple, it just requires a `.custom_metrics()` implementation similar to [other results throughout fiftyone](https://github.com/voxel51/fiftyone/blob/a687aeaae88b8f73895588f8228fbedf9214dd2a/fiftyone/utils/eval/base.py#L116).

Perhaps something like this:

```python
def custom_metrics(self):
  return {m['id']: m for m in self._custom_metrics}
```

manushreegangwar:
> Proposal for a CustomMetric class that would provide the right information for the SDK and App:
> 
> ```python
> CustomMetric(
>     id="example_metric", # need some identifier, unique to the set of metrics
>     operator_uri="@voxel51/metric-examples/example_metric", # this might not be needed
>     value=4.2, # can be any serializable value but most likely a number or string
>     lower_is_better=True, # allows for color (red vs green) and icon semantics (trophy icon)
>     metric_property=types.Property( # required, so the app knows how to render the value
>         types.Number(float=True),
>         label="Example metric",
>         view=types.NumberView(title="This will show as a tooltip") # optional
>     )
> )
> ```
> 
> Note that the `types` are from `fiftyone.operator.types` and can be used to render an arbitrary value in the app.

Making this change throws serialization error in evaluation run. `types.Property` is not serializable.

fiftyone/utils/eval/base.py
manushreegangwar: It is used in summaryRow as [id](https://github.com/voxel51/fiftyone/pull/5389/files#diff-dc63da9c0259cdf08ca27f8b7c5c932dcebaa187d426ddf93c9d6a36435855b8R1800)

brimoor (approved):

brimoor (approved):

PR #5398, opened 2025-01-17
Use estimated document count when possible for `count()` operations

brimoor:
## Change log

Use estimated document counts for `dataset.count()` and `dataset.count("frames")`.

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

# These all now use estimated document counts
_ = dataset.summary()
_ = len(dataset)
_ = dataset.count()
_ = dataset.count("frames")

# And so do these
view = dataset.select_group_slices(_allow_mixed=True)
_ = view.summary()
_ = len(view)
_ = view.count()
```

## Discussion

When are estimated document counts not exact? See
https://www.mongodb.com/docs/manual/reference/method/db.collection.estimatedDocumentCount/#behavior

For context, aside from cases where a user is directly calling `count()`, there are a couple key operations that call `count()` internally:

1. `len(dataset)` calls `dataset.count()`
2. `print(dataset)` calls `dataset.summary()` which calls `dataset.count()`

In my opinion, 2 is a clear case where it makes sense to use estimated document counts. It's more important for `print(dataset)` to be fast than it is for the count to be accurate in every possible case.

1 is a bit dubious though. On one hand, it's reasonable to assume that there are cases where you want to know *exactly* how many samples are in a dataset. On the other hand, there are times, especially in interactive shells, where you don't need precision, you just want to know, well, *roughly* how big the dataset is (ya know, an *estimate*).

My current opinion is that it is probably okay for `len(dataset)` to be an estimate. After all, datasets may have samples added/deleted at any time. So there's really no guarantee how long an "exact" document count will remain exact. And therefore it you're writing code that relies on `len(dataset)` being exact for a long period of time, then you're probably asking for trouble.

An alternative strategy would be to make this optimization *optional*, eg by introducing a syntax like this:

```py
# Allow estimated document counts, if possible
dataset_or_view.count(..., allow_estimates=True)
```

We could then make sure that `print(dataset)` uses the optimization, while making the default `dataset.count(..., allow_estimates=False)` so that `len(dataset)` is exact.

## Summary by CodeRabbit

- **Performance Optimization**
	- Enhanced document counting method with estimated count functionality.
	- Added performance-improving checks for sample and frame collections.
	- Introduced a new method to determine full dataset collection status.

- **New Features**
	- Added a method to quickly obtain an estimated count of samples or frames in the dataset.
	- Enhanced modal components to ensure correct re-rendering based on selected media fields.
	- Improved operator execution handling by simplifying the interaction model and removing unused props.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -7726,6 +7726,20 @@ class SampleCollection(object):
         Returns:
             the count
         """
+
+        # Optimization: use estimated document count when possible
+        if (
+            isinstance(self, fod.Dataset)
+            and expr is None
+            and (
+                field_or_expr is None
+                or (field_or_expr == "frames" and self._has_frame_fields())
+            )
+        ):
+            frames = field_or_expr == "frames"
+            # pylint: disable=no-member
+            return self._estimated_count(frames=frames)
+
         make = lambda field_or_expr: foa.Count(
             field_or_expr, expr=expr, safe=safe
         )
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -7728,17 +7728,19 @@ class SampleCollection(object):
         """
 
         # Optimization: use estimated document count when possible
-        if (
-            isinstance(self, fod.Dataset)
-            and expr is None
+        if self._is_full_collection() and (
+            expr is None
             and (
                 field_or_expr is None
-                or (field_or_expr == "frames" and self._has_frame_fields())
+                or (
+                    etau.is_str(field_or_expr)
+                    and field_or_expr == "frames"
+                    and self._has_frame_fields()
+                )
             )
         ):
             frames = field_or_expr == "frames"
-            # pylint: disable=no-member
-            return self._estimated_count(frames=frames)
+            return self._dataset._estimated_count(frames=frames)
 
         make = lambda field_or_expr: foa.Count(
             field_or_expr, expr=expr, safe=safe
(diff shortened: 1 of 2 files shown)


kaixi-wang (approved):
This will speed up a lot and iterating based on a count rather than mapping is very unreliable

kaixi-wang (changes requested):
actually looks like the tests are failing so it might not be this simple

brimoor:
> actually looks like the tests are failing so it might not be this simple

ah I forgot that this optimization isn't applicable to grouped datasets

fiftyone/core/collections.py
```diff
@@ -7726,6 +7726,20 @@ def count(self, field_or_expr=None, expr=None, safe=False):
         Returns:
             the count
         """
+
+        # Optimization: use estimated document count when possible
+        if (
+            isinstance(self, fod.Dataset)
```
swheaton, line 7732: should fix it. the way groups are currently implemented, we cannot use estimated count.
```suggestion
            isinstance(self, fod.Dataset)
            and self.media_type != fom.GROUP
```

brimoor: I generalized it slightly more in https://github.com/voxel51/fiftyone/pull/5398/commits/d11868cb1e37fa2f6a06264a3895cc728879f131 so that estimated counts can also be used for grouped datasets when applicable, namely:

```py
grouped_dataset.select_group_slices().count()
```

swheaton (approved):
makes sense to me! I believe this was proposed previously but rejected due to some of the stated reasons. Glad we are making it now.
Measured with remote DB and 1.9M samples ... a 4000x speedup for `len(dataset)`! Congrats to you and @kaixi-wang lol 🔥 

🚢 🇮🇹

PR #5406, opened 2025-01-17
fix jpg segmentations

sashankaryal:

fix jpg segmentations

(Details)

        notes for FiftyOne users.

-   [ ] Other


app/packages/looker/src/worker/canvas-decoder.ts
```diff
@@ -121,6 +128,31 @@ export const decodeWithCanvas = async (blob: Blob, cls: string) => {
...
+        height,
+        channels
+      );
+      channels = 1;
+    }
+
+    // note: for JPG segmentations with RGB mask targets, we don't need to recast
+    // although depending on the JPG compression, we might have some artifacts.
+    // even the slightest change in color can cause the mask to be rendered as
+    // background color (transparent) instead of the actual mask color
```
benjaminpkane, line 153: I wouldn't be opposed to considering this a hard error if it is easy to present/document

app/packages/looker/src/worker/canvas-decoder.ts
```diff
@@ -67,7 +69,12 @@ export const recastBufferToMonoChannel = (
   return uint8Array.slice(0, totalPixels).buffer;
 };
 
-export const decodeWithCanvas = async (blob: Blob, cls: string) => {
+export const decodeWithCanvas = async (
+  blob: Blob,
+  cls: string,
+  field: string,
+  coloring: Coloring
+) => {
```
benjaminpkane, line 77: This could help `decodeWithCanvas` responsibilities to a minimum
```suggestion
export const decodeWithCanvas = async (
  blob: Blob,
  cls: string,
  isRgb: boolean
) => {
```

benjaminpkane (approved):
LGTM 🥇

app/packages/looker/src/worker/canvas-decoder.ts
sashankaryal: I update some of this in https://github.com/voxel51/fiftyone/pull/5413

app/packages/looker/src/worker/canvas-decoder.ts
sashankaryal: Yeah... the thing is you can have JPGs with zero compression and that might not be a problem, so probably not the best idea to consider it a hard error.

PR #5407, opened 2025-01-17
Optimize dataset first/head and last/tail 

kaixi-wang:

Optimizes `first()`, `last()`, `head()`, and `tail()` calls for datasets.

- Tested locally and on a 11,460,588 sample dataset that first, last, head, and tail return instantaneously


fiftyone/core/collections.py
```diff
@@ -949,7 +961,7 @@ def head(self, num_samples=3):
         Returns:
             a list of :class:`fiftyone.core.sample.Sample` objects
         """
-        return [s for s in self[:num_samples]]
+        return [s for s in self._get_first(limit=num_samples)]
```
brimoor, line 964: I'm curious: do these changes optimize `first()` and `head()`? Or are you just standardizing on `_get_first()` usage?

For `first()`, we're replacing iterating one sample on `pipeline=[]` with `pipeline=[{$sort: {_id: 1}}, {$limit: 1}]`.

For `head()` we're going from `pipeline=[{$limit: n}]` to `pipeline=[{$sort: {_id: 1}}, {$limit: n}]`.

My intuition is that both should perform the same. Definitely would be weird if `{$sort: {_id: 1}}` made things faster. And I'd hope that `{$sort: {_id: 1}}` wouldn't make things any slower since `_id` is indexed.

fiftyone/core/collections.py
```diff
@@ -916,17 +916,28 @@ def _get_per_sample_frames_bytes(self):
         results = self._aggregate(pipeline=pipeline, attach_frames=True)
         return {str(r["_id"]): r["size_bytes"] for r in results}
 
+    def _get_first(self, limit=None, field=None, reverse=False):
+        pipeline = [{"$limit": limit if limit else 1}]
+        direction = -1 if reverse else 1
+        if field is None:
+            # Sort by _id if no field is specified to ensure first is deterministic
+            field = "_id"
+        pipeline.insert(0, {"$sort": {field: direction}})
```
brimoor, line 925: nit: can you just write it like this for simplicity? 

```py
pipeline = [
    {"$sort": {field: direction}},
    {"$limit": limit if limit else 1},
]
```

fiftyone/core/collections.py
```diff
@@ -916,17 +916,28 @@ def _get_per_sample_frames_bytes(self):
...
 
         Returns:
             a :class:`fiftyone.core.sample.Sample` or
             :class:`fiftyone.core.sample.SampleView`
         """
-        try:
-            return next(iter(self))
-        except StopIteration:
-            raise ValueError("%s is empty" % self.__class__.__name__)
+        return next(self._get_first())
```
brimoor, line 940: This is returning dicts, not `Sample` or `SampleView` objects as promised. Which means that `DatasetView.first()`  is currently broken.

The existing pattern for situations like this where `Dataset.fcn` and `DatasetView.fcn` need different implementations is to raise `NotImplementedError` in `SampleCollection.fcn` and then implement the logic in the other two places.

brimoor (changes requested):

fiftyone/core/collections.py
```diff
@@ -916,26 +916,45 @@ def _get_per_sample_frames_bytes(self):
         results = self._aggregate(pipeline=pipeline, attach_frames=True)
         return {str(r["_id"]): r["size_bytes"] for r in results}
 
+    def _get_first(self, limit=1, field="_id", reverse=False):
+        direction = -1 if reverse else 1
+        pipeline = [{"$sort": {field: direction}}, {"$limit": limit}]
+
+        return self._aggregate(
+            pipeline=pipeline, detach_frames=True, detach_groups=True
```
brimoor, line 924: `iter_samples()` includes `detach_frames=True, detach_groups=True` so we should use it here too

fiftyone/core/collections.py
```diff
@@ -947,9 +966,11 @@ def head(self, num_samples=3):
             num_samples (3): the number of samples
 
         Returns:
-            a list of :class:`fiftyone.core.sample.Sample` objects
+            a list of :class:`fiftyone.core.sample.Sample` or
+            :class:`fiftyone.core.sample.SampleView` objects
         """
-        return [s for s in self[:num_samples]]
+        cursor = self._get_first(limit=num_samples)
+        return [self._make_sample(d) for d in cursor]
```
brimoor, line 973: we want to return an empty list here, not raise an exception, for backwards compatibility

fiftyone/core/collections.py
```diff
@@ -961,9 +982,13 @@ def tail(self, num_samples=3):
...
 
         Returns:
-            a list of :class:`fiftyone.core.sample.Sample` objects
+            a list of :class:`fiftyone.core.sample.Sample` or
+            :class:`fiftyone.core.sample.SampleView` objects
         """
-        return [s for s in self[-num_samples:]]
+        cursor = self._get_first(reverse=True, limit=num_samples)
+        samples = [self._make_sample(d) for d in cursor]
+        samples.reverse()
```
brimoor, line 990: `reverse()` here so that sample order is maintained, for backwards compatibility

fiftyone/core/collections.py
```diff
@@ -1008,27 +1033,28 @@ def one(self, expr, exact=False):
...
 
         Returns:
-            a :class:`fiftyone.core.sample.SampleView`
+            a class:`fiftyone.core.sample.Sample` or
+            :class:`fiftyone.core.sample.SampleView`
         """
-        view = self.match(expr)
-        matches = iter(view)
+        limit = 2 if exact else 1
+        view = self.match(expr).limit(limit)
```
brimoor, line 1040: added `$limit` here as I believe your benchmarking of `dataset.first()` showed that explicitly including a limit is better than only calling `next()` once on a cursor with no limit

fiftyone/core/collections.py
```diff
@@ -916,26 +916,45 @@ def _get_per_sample_frames_bytes(self):
         results = self._aggregate(pipeline=pipeline, attach_frames=True)
         return {str(r["_id"]): r["size_bytes"] for r in results}
 
+    def _get_first(self, limit=1, field="_id", reverse=False):
+        direction = -1 if reverse else 1
+        pipeline = [{"$sort": {field: direction}}, {"$limit": limit}]
```
brimoor, line 921: It should be noted that the implementation of `view.last()` is now significantly different than it used to be.

Previously it would run two pipelines: `view_pipeline + [{"$count": n}]` to get the count followed by `view_pipeline + [{"$skip": n - 1}]` to get the last sample.

Now it runs a single pipeline: `view_pipeline + [{"$sort": {"_id": -1}}, {"$limit": 1}]` that directly grabs the last sample.

`$count` and `$skip` require collection scans, which was definitely not ideal and is the key thing that this PR is addressing in the first place. However, at least they probably(?) don't require too much DB memory.

I'm uncertain how performant `[{$sort: _id}, {$limit: 1}]` is when appended to a potentially complex pipeline. I worry that the sorting might require more memory?

brimoor (commented):
@kaixi-wang can you benchmark this updated version on your test datasets?

Per [this comment](https://github.com/voxel51/fiftyone/pull/5407/files#r1921238537), I'm particularly interested in the performance of `complex_view.last()` before/after this PR.

fiftyone/core/collections.py
```diff
@@ -1008,27 +1033,28 @@ def one(self, expr, exact=False):
...
 
         try:
-            sample = next(matches)
+            d = next(matches)
         except StopIteration:
             raise ValueError("No samples match the given expression")
 
         if exact:
             try:
                 next(matches)
```
minhtuev, line 1050: Nice! Is there a way to check if the iterator still has more elements instead of calling next()?

fiftyone/core/collections.py
```diff
@@ -916,26 +916,45 @@ def _get_per_sample_frames_bytes(self):
...
         """
+        cursor = self._get_first()
+
         try:
-            return next(iter(self))
+            d = next(cursor)
         except StopIteration:
             raise ValueError("%s is empty" % self.__class__.__name__)
 
+        return self._make_sample(d)
```
minhtuev, line 941: Do users use these operations like `first` and `last` often enough that we can implement some caching here too?

kaixi-wang: I don't know about other users, but personally I don't think we'd want to cache the results. I usually call first and last after some mutation as a sanity check

fiftyone/core/collections.py
kaixi-wang: Good catch 👍

fiftyone/core/view.py
```diff
@@ -439,6 +439,119 @@ def _make_view_stages_str(self):
             ]
         )
 
+    def first(self):
```
brimoor, line 442: Unfortunately we can't use `_get_first()` for views because they may contain sorting operations of their own.

For views, I can't think of anything we can easily do that's better than retaining the existing implementation.

brimoor (approved):

PR #5413, opened 2025-01-20
add 16 bit png support

### LINKED ISSUE #5363: [BUG] recent versions don't render uint16 semantic segmentation masks anymore?
### Describe the problem

We updated from 1.0.1 to 1.2.0 and somehow our uint16 semantic segmentation masks are often not rendered anymore.

Can you reproduce this on your side or at least confirm that relevant code has been changed between these two versions?

### Code to reproduce issue

When running the code below, then masks are rendered using `1.0.1`, but not with `1.2.0`.

There seem to be some exceptions: When using `1.2.0` and launching a dataset with `np.uint8` masks and then switching to a dataset with `uint16` masks in the UI, they are rendered again (sometimes)?  It's a bit tedious to reproduce, since a restart seems to be necessary when switching versions.

```python
import fiftyone as fo
from PIL import Image
import numpy as np
from pathlib import Path


def _save(x: np.ndarray, path: Path):
    Image.fromarray(x).save(path)
    return path


# write image
hw = 32
rng = np.random.default_rng(42)
img = rng.integers(80, 120, size=(hw, hw, 3), dtype=np.uint8)
mask = rng.integers(0, 255, size=(hw, hw), dtype=np.uint16)

# save sample
img_path = _save(img, "/tmp/img.png")
mask_path = _save(mask, "/tmp/mask.png")
sample = fo.Sample(filepath=img_path)
sample["semseg"] = fo.Segmentation(mask_path=mask_path)

# create dataset
ds = fo.Dataset("tmp_055", overwrite=True, persistent=False)
ds.add_sample(sample)
fo.launch_app(ds)
```

### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 22.04): Linux Ubuntu 20.04.6 LTS
-   **Python version** (`python --version`): 3.11
-   **FiftyOne version** (`fiftyone --version`): 1.01 or 1.2.0
-   **FiftyOne installed from** (pip or source): uv pip 

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community

sashankaryal:
addresses #5363

## Release Notes

- **New Features**
  - Enhanced PNG image decoding support with improved handling of 16-bit images
  - Added advanced mask decoding functionality for different image formats

- **Bug Fixes**
  - Improved error handling for image decoding processes
  - Simplified image channel determination logic

- **Performance**
  - Streamlined image processing methods
  - Optimized overlay mask decoding mechanisms

- **Technical Improvements**
  - Removed deprecated PNG color type detection
  - Updated decoding strategies for more robust image processing


benjaminpkane (changes requested):
I think a regression test is worthwhile for this. A vitest test case that asserts 4 x 4 16 bit png result maybe?

sashankaryal:
> I think a regression test is worthwhile for this. A vitest test case that asserts 4 x 4 16 bit png result maybe?

good suggestion. done

benjaminpkane (approved):
LGTM 🖼️

PR #5419, opened 2025-01-22
fix mime type error

sashankaryal:

Sometimes mime type is missing and we get just `binary/octet-stream` for masks. The app should be resilient against that.

Note: unit tests for masks decoding are added in #5413 

(Details)

-   [] No. You can skip the rest of this section.
        notes for FiftyOne users.

Fix masks loading bug

-   [ ] Other

## Release Notes

- **New Features**
  - Improved image processing flexibility in canvas decoding
  - Enhanced overlay processing with more robust filtering

- **Bug Fixes**
  - Refined image type handling in canvas decoding

- **Dependency Updates**
  - Upgraded boto3 to version 1.36.2
  - Updated ipywidgets version constraints
  - Added awscli 1.37.2 to test requirements

- **Chores**
  - Removed specific version constraint for itsdangerous package
  - Refactored overlay and image processing logic

### DIFF
--- a/app/packages/looker/src/worker/canvas-decoder.ts
+++ b/app/packages/looker/src/worker/canvas-decoder.ts
@@ -73,6 +73,7 @@ export const decodeWithCanvas = async (blob: Blob, cls: string) => {
   let channels: number = 4;
 
   if (blob.type !== "image/jpg" && blob.type !== "image/jpeg") {
+    // note that the following function doesn't rely on MIME type and instead reads the file header
     const colorType = await getMaybePngcolorType(blob);
     if (colorType !== undefined) {
       // according to PNG specs:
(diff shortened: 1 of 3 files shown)


app/packages/looker/src/worker/canvas-decoder.ts
```diff
@@ -70,8 +72,9 @@ export const recastBufferToMonoChannel = (
 export const decodeWithCanvas = async (blob: Blob, cls: string) => {
   let channels: number = 4;
 
-  if (blob.type === "image/png") {
-    const colorType = await getPngcolorType(blob);
+  if (blob.type !== "image/jpg" && blob.type !== "image/jpeg") {
```
benjaminpkane, line 75: This implies we still rely on MIME type, specifically for jpeg?

benjaminpkane (dismissed):

app/packages/looker/src/worker/canvas-decoder.ts
sashankaryal: Not really... more like a heuristic. We don't want to unnecessarily check if it's PNG if the mime type says it's a jpeg.

Note that all of this is used just to determine number of channels.

benjaminpkane (approved):

PR #5425, opened 2025-01-23
Fix multi config test

kaixi-wang:

- Test fails if database admin set to false, leading to extra configs and failures upon rerun due to duplicate key error

- Tests pass with FIFTYONE_DATABASE_ADMIN = false

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

- **Tests**
	- Enhanced configuration cleanup test with conditional logic for database admin privileges
	- Added clarifying comment to explain test behavior
	- Implemented cleanup step for test configurations

### DIFF
--- a/tests/unittests/utils_tests.py
+++ b/tests/unittests/utils_tests.py
@@ -461,14 +461,14 @@ class ConfigTests(unittest.TestCase):
             # Add some duplicate documents
             db.config.insert_one(
                 {
-                    "_id": ObjectId.from_datetime(datetime(2022, 1, 1)),
+                    "_id": new_config_ids[0],
                     "version": "0.14.4",
                     "type": "fiftyone",
                 }
             )
             db.config.insert_one(
                 {
-                    "_id": ObjectId.from_datetime(datetime(2023, 1, 1)),
+                    "_id": new_config_ids[1],
                     "version": "0.1.4",
                     "type": "fiftyone",
                 }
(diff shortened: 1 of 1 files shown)


tests/unittests/utils_tests.py
```diff
@@ -463,12 +473,19 @@ def test_multiple_config_cleanup(self):
...
+        if fo.config.database_admin:
+            # Ensure that duplicate documents are automatically cleaned up if run by database admin
+            self.assertEqual(len(list(db.config.aggregate([]))), 1)
+        else:
+            # Otherwise, the duplicates are not cleaned up
+            self.assertEqual(len(list(db.config.aggregate([]))), 3)
         self.assertEqual(config, orig_config)
 
+        # Clean up the fake configs
+        db.config.delete_many({"_id": {"$in": new_config_ids}})
```
swheaton, line 487: put this in a try-finally block so they'll still be deleted even if an assertion or other call fails

swheaton:
Fine with me, thanks for the cleanup. 1 thing then good

tests/unittests/utils_tests.py
kaixi-wang: good idea

minhtuev (approved):
Running unit test passses locally with database_admin=False

swheaton:
running into the `first() == "bar"` issue now. i'll start looking into this as well when i can

kaixi-wang:
merged this since it has nothing to do with the failing tests

PR #5436, opened 2025-01-25
Support adding custom metrics to existing evaluations

brimoor:
## Change log

- Custom metrics can now be added to existing evaluations via a new `results.add_custom_metrics()` method

## Example usage

```shell
# Install the example metrics plugin
fiftyone plugins download \
    --plugin-names @voxel51/metric-examples
```

```py
import random
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

dataset.delete_sample_field("ground_truth")

for idx, sample in enumerate(dataset.iter_samples(autosave=True, progress=True), 1):
    ytrue = random.random() * idx
    ypred = ytrue + np.random.randn() * np.sqrt(ytrue)
    confidence = random.random()
    sample["ground_truth"] = fo.Regression(value=ytrue)
    sample["predictions"] = fo.Regression(value=ypred, confidence=confidence)

results = dataset.evaluate_regressions(
    "predictions",
    gt_field="ground_truth",
    eval_key="eval",
)
```

Later in another session:

```py
import fiftyone as fo

dataset = fo.load_dataset("cifar10-test")
results = dataset.load_evaluation_results("eval")

results.add_custom_metrics(
    [
        "@voxel51/metric-examples/absolute_error",
        "@voxel51/metric-examples/squared_error",
    ]
)

print(dataset.bounds("eval_absolute_error"))
print(dataset.bounds("eval_squared_error"))

results.print_metrics()

dataset.delete_evaluation("eval")

assert not dataset.has_field("eval_absolute_error")
assert not dataset.has_field("eval_squared_error")
```

## Release Notes

- **New Features**
  - Enhanced model evaluation capabilities with support for custom metrics.
  - Added ability to dynamically add custom metrics to existing evaluations.

- **Documentation**
  - Updated user guide with comprehensive details on model evaluation techniques.
  - Expanded documentation on custom metrics implementation and usage.

- **Improvements**
  - Introduced more flexible evaluation methods for regressions, classifications, detections, and segmentations.
  - Improved metric filtering and management during model evaluations.

These updates provide users with greater control and customization options when evaluating machine learning models in FiftyOne.

### DIFF
--- a/fiftyone/utils/eval/base.py
+++ b/fiftyone/utils/eval/base.py
@@ -28,6 +28,8 @@ class BaseEvaluationMethodConfig(foe.EvaluationMethodConfig):
     """Base class for configuring evaluation methods.
 
     Args:
+        custom_metrics (None): an optional list of custom metrics to compute
+            or dict mapping metric names to kwargs dicts
         **kwargs: any leftover keyword arguments after subclasses have done
             their parsing
     """
@@ -42,17 +44,27 @@ class BaseEvaluationMethod(foe.EvaluationMethod):
         config: an :class:`BaseEvaluationMethodConfig`
     """
 
-    def _get_custom_metrics(self):
+    def _get_custom_metrics(self, metric_uris=None):
         if not self.config.custom_metrics:
             return {}
 
         if isinstance(self.config.custom_metrics, list):
             return {m: None for m in self.config.custom_metrics}
 
-        return self.config.custom_metrics
+        custom_metrics = self.config.custom_metrics
 
-    def compute_custom_metrics(self, samples, eval_key, results):
-        for metric, kwargs in self._get_custom_metrics().items():
+        if metric_uris is not None:
+            custom_metrics = {
+                k: v for k, v in custom_metrics.items() if k in metric_uris
+            }
+
+        return custom_metrics
+
+    def compute_custom_metrics(
+        self, samples, eval_key, results, metric_uris=None
+    ):
+        custom_metrics = self._get_custom_metrics(metric_uris=metric_uris)
+        for metric, kwargs in custom_metrics.items():
             try:
                 operator = foo.get_operator(metric)
                 value = operator.compute(samples, results, **kwargs or {})
@@ -79,10 +91,11 @@ class BaseEvaluationMethod(foe.EvaluationMethod):
                     e,
                 )
 
-    def get_custom_metric_fields(self, samples, eval_key):
+    def get_custom_metric_fields(self, samples, eval_key, metric_uris=None):
         fields = []
 
-        for metric in self._get_custom_metrics().keys():
+        custom_metrics = self._get_custom_metrics(metric_uris=metric_uris)
+        for metric in custom_metrics.keys():
             try:
                 operator = foo.get_operator(metric)
                 fields.extend(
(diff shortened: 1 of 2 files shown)


fiftyone/utils/eval/base.py
```diff
@@ -144,6 +161,42 @@ def __init__(
...
+        if not custom_metrics:
+            return
+
+        _custom_metrics = self.config.custom_metrics
+
+        if _custom_metrics is None:
+            _custom_metrics = {}
+
+        if isinstance(_custom_metrics, list):
+            _custom_metrics = {k: None for k in _custom_metrics}
```
manushreegangwar, line 180: `_custom_metrics = self._get_custom_metrics()` should do the same.

fiftyone/utils/eval/base.py
```diff
@@ -144,6 +161,42 @@ def __init__(
...
+        if all(v is None for v in _custom_metrics.values()):
+            _custom_metrics = list(_custom_metrics.keys())
+
+        self.config.custom_metrics = _custom_metrics
+        self.save_config()
+
+        self.backend.compute_custom_metrics(
+            self.samples, self.key, self, metric_uris=metric_uris
+        )
+        self.save()
```
manushreegangwar, line 198: Updating `config` with new custom metrics and then using `metric_uris` to filter seems a bit convoluted. Also, it would recompute any metric that was in the original config. Some suggestions to simplify this method:

1. Change method to function: `_get_custom_metrics(custom_metrics: Union(Dict, List))`
2. Update calls to `_get_custom_metrics(custom_metrics=metric_uris if metric_uris else self.config.custom_metrics)` in BaseEvaluationMethod.
3. Update `config.custom_metrics`. Why not change list to dict (if needed) in the first update to avoid checking that all the values are None? 

Also, when adding new metrics, we should check against `results.custom_metrics` and make sure that any of the new operator uris that are in `results.custom_metrics.keys()` should not be recomputed.

manushreegangwar (changes requested):

fiftyone/utils/eval/base.py
brimoor: It doesn't recompute existing metrics, it just:
1. Adds new metrics to the config (taking care to not double add metrics in case they were already applied)
2. Saves the config *before* metric computation occurs, so that even if an error occurs, the config will correctly have a record that the metric(s) were partially applied
3. Computes *only* the metrics that the user passed to `add_custom_metrics()`

manushreegangwar: Clarification: _Also, it would recompute any new metric added by the user that was also in the original config_

Is that not the case? I will go over the implementation again.

brimoor: Correct, it will compute any new metrics and recompute any existing metrics, but not recompute anything that wasn't directly passed to `add_custom_metrics()`

manushreegangwar: My comment is regarding the _recompute any existing metrics_ part. Do we want to always recompute existing metrics every time `add_custom_metric` gets an existing metric uri? 

I can see the value in recomputing but would like to be able to control when not to recompute/overwrite. For example, I have 5 evaluation runs on a large dataset and I want to make sure all the runs have 10 chosen custom metrics. Some of these runs already have n of these 10 metrics so I don't want to recompute them.

brimoor: That use case is already possible like so:

```py
# Custom metrics to apply
custom_metrics = [...]

results = dataset.load_evaluation_results(eval_key)

# Recomputing existing metrics is a valid use case
# But if you don't want to recompute, that's easy too
custom_metrics = [m for m in custom_metrics if m not in results.config.custom_metrics]

results.add_custom_metrics(...)
```

brimoor: done here: https://github.com/voxel51/fiftyone/pull/5436/commits/91e7618cae2334885903ee3927f17977b03f31ab

manushreegangwar (approved):
LGTM! :rocket:

PR #5442, opened 2025-01-27
imavid: return 0 instead of NaN if total frames is 1 or 0

sashankaryal:
minor bugfix to prevent cases where NaN might be returned as a result of division by zero

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of edge cases in timeline visualization calculations to prevent potential division by zero errors
	- Enhanced robustness of percentage conversion function when dealing with minimal frame counts


benjaminpkane (approved):
LGTM. `convertFrameNumberToPercentage` is an easy function to add unit tests for 😄

PR #5446, opened 2025-01-29
Use estimated count on full dataset view

kaixi-wang:

- use estimated count for full dataset view so dataset slices with negative starts can be optimized

- tested locally with 11M sample dataset

```
# With Changes
In [18]: %time v=ds[-2:]
CPU times: user 5.4 ms, sys: 886 µs, total: 6.29 ms
Wall time: 770 ms
```

```
# Without changes
In [19]: %time v=ds[-2:]
CPU times: user 3.63 s, sys: 496 ms, total: 4.12 s
Wall time: 16min 58s
```

```
In [20]: len(ds)
Out[20]: 11460588

```

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Refactor**
  - Simplified logic for determining full collection status.
  - Streamlined condition checks to improve clarity and maintainability.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -10619,7 +10619,6 @@ class SampleCollection(object):
         return _handle_id_fields(self, field_name)
 
     def _is_full_collection(self):
-        """Checks if the current instance represents a full collection."""
         if isinstance(self, fod.Dataset) and self.media_type != fom.GROUP:
             return True
 
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -10619,18 +10619,22 @@ class SampleCollection(object):
         return _handle_id_fields(self, field_name)
 
     def _is_full_collection(self):
+        """Checks if the current instance represents a full collection."""
         if isinstance(self, fod.Dataset) and self.media_type != fom.GROUP:
             return True
 
-        # pylint:disable=no-member
-        if (
-            isinstance(self, fov.DatasetView)
-            and self._dataset.media_type == fom.GROUP
-            and len(self._stages) == 1
-            and isinstance(self._stages[0], fos.SelectGroupSlices)
-            and self._pipeline() == []
-        ):
-            return True
+        if isinstance(self, fov.DatasetView):
+            if not self._pipeline() and self._dataset.media_type != fom.GROUP:
+                return True
+
+            # pylint:disable=no-member
+            if (
+                self._dataset.media_type == fom.GROUP
+                and len(self._stages) == 1
+                and isinstance(self._stages[0], fos.SelectGroupSlices)
+                and not self._pipeline()
+            ):
+                return True
 
         return False
 
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -10625,7 +10625,7 @@ class SampleCollection(object):
         if isinstance(self, fov.DatasetView):
             if not (
                 self._pipeline()
-                or isinstance(self, fov.ClipsView)
+                or self._is_clips
                 or self._dataset.media_type == fom.GROUP
             ):
                 return True
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -10623,10 +10623,14 @@ class SampleCollection(object):
             return True
 
         if isinstance(self, fov.DatasetView):
-            # pylint:disable=no-member
-            if not self._stages and self._dataset.media_type != fom.GROUP:
+            if not (
+                self._pipeline()
+                or isinstance(self, fov.ClipsView)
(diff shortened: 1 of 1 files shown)


fiftyone/core/collections.py
```diff
@@ -10622,15 +10622,22 @@ def _is_full_collection(self):
...
-            isinstance(self, fov.DatasetView)
-            and self._dataset.media_type == fom.GROUP
-            and len(self._stages) == 1
-            and isinstance(self._stages[0], fos.SelectGroupSlices)
-            and self._pipeline() == []
-        ):
-            return True
+        if isinstance(self, fov.DatasetView):
+            if not (
+                self._pipeline()
```
brimoor, line 10627: Note that `_pipeline()` can be expensive for some views. EG if a view contains `SortBySimilarity` it can even run model inference!!!

That's why I wrote the existing `if` statement in this function in a seemingly funny way: I first check things that are super cheap and only call `_pipeline()` last once I've determined that the view is simple and thus `_pipeline()` will be fast. Taking advantage of [short-circuiting](https://docs.python.org/2/library/stdtypes.html#boolean-operations-and-or-not).

I'd request that a similar treatment be applied to any additional logic added to this method.

Side note, I've been staring at this first `if` for a minute and can't figure out what it is saying(!)

kaixi-wang: ohh thanks for calling this out! I was wondering why we used _pipeline() and stages here

swheaton: Yeah I don't like the refactor either, it's much harder to understand imo.

swheaton:
Why would a view have a negative start?

swheaton:
I guess it doesn't matter, len(ds.view()) should be fast as well

kaixi-wang:
> Why would a view have a negative start?

Slices from the end to see the tail

swheaton:
> > Why would a view have a negative start?
> 
> Slices from the end to see the tail

Oh well if it's at the tail, then it's not the full collection?

kaixi-wang:
We overwrite the slice operator so that dataset[-N:] triggers a skip aggregation and to get the value of skip, there's a hidden len(dataset.view()) that is previously excluded in this condition

swheaton:
Ah ok so really the optimization is for `dataset.view()` like I mentioned above. Got it.

fiftyone/core/collections.py
kaixi-wang: Updated the syntax...

Side note, kinda funny how the bot summary described this syntax as simplified and streamlined 🤖

swheaton: the rabbit sees the glass half full 😃

fiftyone/core/collections.py
```diff
@@ -10622,15 +10622,22 @@ def _is_full_collection(self):
         if isinstance(self, fod.Dataset) and self.media_type != fom.GROUP:
             return True
```
swheaton, line 10623:
```suggestion
        # Case 1: full non-group dataset
        if isinstance(self, fod.Dataset) and self.media_type != fom.GROUP:
            return True
```

fiftyone/core/collections.py
```diff
@@ -10622,15 +10622,22 @@ def _is_full_collection(self):
...
+            ):
+                return True
+
+            if (
+                self._dataset.media_type == fom.GROUP
+                and len(self._stages) == 1
+                and isinstance(self._stages[0], fos.SelectGroupSlices)
+                and not self._pipeline()
+            ):
+                return True
```
swheaton, line 10640: lets add comments about the cases
```suggestion
        if isinstance(self, fov.DatasetView):
            # Case 2: Empty non-group dataset view
            # pylint:disable=no-member
            if (
                not len(self._stages)
                and not self._is_clips
                and self._dataset.media_type != fom.GROUP
            ):
                return True

            # Group dataset with all slices selected and no additional stages
            if (
                self._dataset.media_type == fom.GROUP
                and len(self._stages) == 1
                and isinstance(self._stages[0], fos.SelectGroupSlices)
                and not self._pipeline()
            ):
                return True
```

fiftyone/core/collections.py
```diff
@@ -10622,15 +10622,22 @@ def _is_full_collection(self):
...
-            and len(self._stages) == 1
-            and isinstance(self._stages[0], fos.SelectGroupSlices)
-            and self._pipeline() == []
-        ):
-            return True
+        if isinstance(self, fov.DatasetView):
+            # pylint:disable=no-member
+            if (
+                not len(self._stages)
+                and not self._is_clips
```
swheaton, line 10629: Is clips the only case? Or is it any generated dataset?

PR #5449, opened 2025-01-29
`beam.map()`

brimoor:
## Change log

Adds a `beam_map()` method that leverages Apache Beam to apply a `map_fcn` to each sample in a `sample_collection` in parallel, optionally saving any sample edits and/or applying a `reduce_fcn` or `aggregate_fcn` to the outputs.

## Example 1: map-save

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.beam as foub

def map_fcn(sample):

foub.beam_map(view, map_fcn, save=True, progress=True)

```

```
Batch 02/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 811.38it/s]
Batch 13/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 806.92it/s]
Batch 12/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 795.11it/s]
Batch 06/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 806.14it/s]
Batch 14/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 790.75it/s]
Batch 03/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 788.72it/s]
Batch 01/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 789.38it/s]
Batch 11/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 794.18it/s]
Batch 04/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 789.54it/s]
Batch 07/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 785.78it/s]
Batch 10/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 796.16it/s]
Batch 15/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 783.64it/s]
Batch 16/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 783.61it/s]
Batch 09/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 790.07it/s]
Batch 05/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 1008.09it/s]
Batch 08/16: 100%|█████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 1009.33it/s]
{'TRUCK': 5000, 'AIRPLANE': 5000, 'AUTOMOBILE': 5000, 'SHIP': 5000, 'FROG': 5000, 'BIRD': 5000, 'CAT': 5000, 'DEER': 5000, 'HORSE': 5000, 'DOG': 5000}
```

Same operation with `iter_samples(autosave=True)`:

```py
for sample in view.iter_samples(autosave=True, progress=True):
    map_fcn(sample)
```

```
 100% |██████████████████████████████████████████████████████████████████| 50000/50000 [27.6s elapsed, 0s remaining, 1.6K samples/s]      
```

## Example 2: map-reduce

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.beam as foub

def map_fcn(sample):
    return sample.ground_truth.label.lower()

class ReduceFcn(foub.ReduceFcn):
    def create_accumulator(self):
        return Counter()

    def add_input(self, accumulator, input):
        sample_id, value = input
        accumulator[value] += 1
        return accumulator

    def merge_accumulators(self, accumulators):
        accumulator = Counter()
        for a in accumulators:
            accumulator.update(a)
        return accumulator

    def extract_output(self, accumulator):
        counts = dict(accumulator)
        self._store_output(counts)

counts = foub.beam_map(view, map_fcn, reduce_fcn=ReduceFcn, progress=True)
print(counts)
```

## Example 3: map-aggregate

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.beam as foub

def map_fcn(sample):
    return sample.ground_truth.label.lower()

def aggregate_fcn(sample_collection, values):
    return dict(Counter(values.values()))

counts = foub.beam_map(view, map_fcn, aggregate_fcn=aggregate_fcn, progress=True)
print(counts)
```

### DIFF
--- a/fiftyone/utils/beam.py
+++ b/fiftyone/utils/beam.py
@@ -5,14 +5,19 @@
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+import itertools
 import logging
+import os
+
 import numpy as np
 
 import fiftyone.core.dataset as fod
 import fiftyone.core.utils as fou
 import fiftyone.core.view as fov
 
+os.environ["GRPC_VERBOSITY"] = "NONE"
 fou.ensure_import("apache_beam")
+tqdm = fou.lazy_import("tqdm")
 
 import apache_beam as beam
 from apache_beam.options.pipeline_options import PipelineOptions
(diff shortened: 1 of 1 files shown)


fiftyone/utils/beam.py
```diff
@@ -338,6 +343,130 @@ def beam_export(
...
+            configures how to run the pipeline. By default, the pipeline will
+            be run via Beam's direct runner using multiprocessing with
+            ``num_workers`` workers
+        progress (False): whether to render progress bar(s) for each worker
+        verbose (False): whether to log the Beam pipeline's messages
+    """
+    if shard_method == "slice":
+        n = len(sample_collection)
+    else:
+        ids = sample_collection.values("id")
```
kaixi-wang, line 404: I've noticed that this can be a bottleneck when the sample collection is large. Can we alternatively define bounds of _ids and avoid creating a huge list?

brimoor: Do you mean like how `shard_method="slice"` works?

fiftyone/utils/beam.py
```diff
@@ -508,6 +703,135 @@ def process(self, element, **kwargs):
...
+        for a in accumulators:
+            accumulator.update(a)
+        return accumulator
+
+    def extract_output(self, accumulator):
+        output = self.reduce_fcn(self._sample_collection, accumulator)
+        if output is None or self.result_key is None:
+            return
+
+        _set_key(self._sample_collection, self.result_key, output)
```
minhtuev, line 820: do we need to return `output` here so that the output from the return function is returned?

brimoor: No, the output is stored via `_set_key()` and loaded via `_get_key()` back in the main method.

Apache Beam supports distributed computing, so in general `ReduceFn.extract_output()` may be called on a different machine and thus any I/O must be done via disk, database, etc.

fiftyone/utils/beam.py
```diff
@@ -338,6 +345,194 @@ def beam_export(
...
+
+        dataset = foz.load_zoo_dataset("cifar10", split="train")
+        view = dataset.select_fields("ground_truth")
+
+        #
+        # Example 1: map
+        #
+
+        def map_fcn(sample):
+            sample.ground_truth.label = sample.ground_truth.label.upper()
```
kaixi-wang, line 397: Concerned that by importing fiftyone in the main process we are essentially passing the mongo client to the worker processes instead of independently creating the client inside the worker, which may result in deadlock in real workloads

brimoor: Database connections aren't serializable so it can't be happening.

minhtuev:
I have done a little bit of refactor and bug fixes so that we can run evaluate detections with `beam_map`, the initial results look good :)

https://github.com/voxel51/fiftyone/pull/5471/files

fiftyone/utils/beam.py
```diff
@@ -338,6 +345,194 @@ def beam_export(
...
+        edges = list(range(0, n + 1, shard_size))
+        if edges[-1] < n:
+            edges.append(n)
+    else:
+        # Split collection into exactly `num_workers` shards
+        edges = [int(round(b)) for b in np.linspace(0, n, num_workers + 1)]
+
+    if shard_method == "slice":
+        # Slice batches
+        slices = list(zip(edges[:-1], edges[1:]), 1)
```
minhtuev, line 465: Bug fix:

```suggestion
        slices = list(zip(edges[:-1], edges[1:]))
```

fiftyone/utils/beam.py
```diff
@@ -414,12 +425,46 @@ def reduce_fcn(sample_collection, values):
         counts = foub.beam_map(view, map_fcn, reduce_fcn=reduce_fcn, progress=True)
         print(counts)
 
+        #
+        # Example 3: map-reduce with `reduce_cls`
```
minhtuev, line 429: nice!

fiftyone/utils/beam.py
brimoor: Actually you're correct that mongo clients can be passed to worker processes. However I still don't believe that is happening with `beam_map()` because I **DID** initially see warnings about sharing mongo clients in https://github.com/voxel51/fiftyone/pull/5480 which I had to manually address, but I don't see those errors here.

So it appears that Beam's direct multiprocessing runner has a more localized way of spawning/forking new processes? 🤷

minhtuev: According to the documentation, each worker is in a sub process and communicates through gRPC, so they are probably initializing new connection 

https://beam.apache.org/documentation/runners/direct/#:~:text=multi_threading%3A%20Runner%20and%20workers%20communicate,worker%20runs%20in%20a%20thread

PR #5451, opened 2025-01-31
parallelize end-to-end specs

sashankaryal:
This PR aims to accomplish the following:

1. Make e2e specs embarrassingly parallel to execute. The more CPU we throw at our automated test suite, the faster it ought to finish executing.
2. Makes specs run faster by skipping toasts. (we save roughly 2-6 seconds each spec)
3. Fix all flaky tests
4. Fix previously skipped screenshot assertions
5. Upgrade all PW related dependencies
6. Cache pip so that we don't re-install Python deps in every run

The net effect of this work is that whereas it used to take a total of 60-90 minutes in the past to run the full suite of tests, with these changes, it's reduced to a total of 14-20 minutes. This is with two cores in the runner that we use for e2e, of course it'd decrease linearly with the number of added cores in the runner. 

## Release Notes

- **Testing Improvements**
  - Updated Playwright configuration for more parallel test execution.
  - Enhanced error handling in test fixtures and components.
  - Improved web server and database management during tests.
  - Modified test suite structure for better organization and sequential execution.
  - Introduced a new class for managing web server lifecycle.
  - Added a new asynchronous method for verifying multiple sidebar entries simultaneously.

- **Environment Configuration**
  - Added new environment variable for process output logging.
  - Updated Node.js version in CI/CD workflow.
  - Refined environment variable handling for case-insensitive checks.

- **Performance Optimization**
  - Disabled analytics and performance toasts during Playwright testing.
  - Improved port availability checking mechanism.

- **Maintenance**
  - Removed unused application configuration constants.
  - Refactored test suite organization and error handling.
  - Introduced a new function for deleting databases programmatically.
  - Updated Dockerfile to use a newer base image version.
  - Added a new shell script for managing MongoDB databases.


benjaminpkane (approved):
LGTM! Let's add more cores 🚀

PR #5467, opened 2025-02-04
add log path field to DO

CamronStaley:

Adding a log_path field for DOs for executors that don't use airflow or have a prebuilt UI a path to a cloud bucket can be provided here for allowing users to download logs in the UI.

n/a

        notes for FiftyOne users.

-   [ ] Other


minhtuev:
Ah just got around to read this PR now since we were in meetings this morning. Overall, LGTM, just a minor comment: we should consider adding a unit test so that the log_path field (and others) can be tested and validated automatically.

PR #5479, opened 2025-02-08
[WIP] Optimize `values()` calls

brimoor:
## Change log

- Optimizes `values("id")` and `values("filepath")` pipelines by using field inclusion syntax where possible
- Optimizes all `values()` calls by omitting `_id` from return results where possible

## Context

Currently, `values("id")` executes the following pipeline:

```py
pipeline = [
    {
        "$project": {
            "value": {
                "$cond": {
                    "if": {"$gt": ["$_id", None]},
                    "then": {"$toString": "$_id"},
                    "else": None,
                },
            },
        },
    },
]
```

However, the conditional logic in that pipeline is not actually needed for *required* fields like `id` and `filepath`.

I wanted to see how much faster `values("id")` would be if the following pipeline were instead used:

```py
pipeline = [{"$project": {"_id": 1}}]
```

with the `str()` conversion instead happening in Python rather than in the database.

## Findings 

The benchmarking section below shows the following findings on a local database with a 500k sample dataset:
- The `{"$project": {"_id": 1}}` pipeline runs in 2.1 seconds compared to the original pipeline, which runs in 3.4 seconds

## What about index usage?

`values("id")` does not leverage the `_id` field index in either of the above pipelines by default. It uses a `COLLSCAN`. So, I also wanted to see how much faster/slower both pipelines are when they use an `IXSCAN`, which can be forced via `hint={"_id": 1}`.

Interestingly, adding `hint={"_id": 1}` makes both pipelines 0.4-0.6 seconds *slower* on my local machine, so I am *not* proposing to add `hint={"_id": 1}` anywhere at this time.

## Create large dataset

```py
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz

def expand_dataset(dataset, target_num_samples):
    num_doubles = int(np.ceil(np.log2(target_num_samples / len(dataset))))
    with fo.ProgressBar(start_msg="Expanding dataset") as pb:
        for i in pb(list(range(1, num_doubles + 1))):
            # Cloning avoids the halloween problem, which I observed if directly using `dataset` in-place of `tmp` here
            # https://www.mongodb.com/docs/manual/reference/operator/aggregation/merge/#output-to-the-same-collection-that-is-being-aggregated
            if i == num_doubles:
                tmp = dataset[:(target_num_samples - len(dataset))].clone()
            else:
                tmp = dataset.clone()

            dataset.add_collection(tmp, new_ids=True)
            tmp.delete()

    "cifar10",
    split="train",
    dataset_name="zzz",
    persistent=True,
)

expand_dataset(dataset, 500000)
```

## Test real usage

```py
import fiftyone as fo
import eta.core.utils as etau

dataset.clone_sample_field("id", "id2")
# dataset.create_index("id2")

# Uses optimized values() call
# Time elapsed: 2.7 seconds
with etau.Timer():
    _ = dataset.values("id")

# Using existing values() call
# Time elapsed: 3.8 seconds
with etau.Timer():
    _ = dataset.values("id2")
```

## Benchmarking

```py
import fiftyone as fo
import eta.core.utils as etau

# Current behavior: in-database string conversion
# Time elapsed: 2.8 seconds
with etau.Timer():
    pipeline = [
        {
            "$project": {
                "value": {
                    "$cond": {
                        "if": {"$gt": ["$_id", None]},
                        "then": {"$toString": "$_id"},
                        "else": None,
                    },
                },
            },
        },
    ]
    results = dataset._sample_collection.aggregate(pipeline)
    ids = [d["value"] for d in results]

# With field inclusion
# Time elapsed: 2.1 seconds
with etau.Timer():
    pipeline = [{"$project": {"_id": 1}}]
    results = dataset._sample_collection.aggregate(pipeline)
    ids = [d["_id"] for d in results]
    ids = [str(_id) for _id in ids]

# With field inclusion + index hint
# Time elapsed: 2.5 seconds
with etau.Timer():
    pipeline = [{"$project": {"_id": 1}}]
    results = dataset._sample_collection.aggregate(pipeline, hint={"_id": 1})
    ids = [d["_id"] for d in results]
    ids = [str(_id) for _id in ids]

# With field setting + hint
# Time elapsed: 3.4 seconds
with etau.Timer():
    pipeline = [{"$project": {"value": "$_id"}}]
    results = dataset._sample_collection.aggregate(pipeline, hint={"_id": 1})
    ids = [d["value"] for d in results]
    ids = [str(_id) for _id in ids]
```

## Summary by CodeRabbit

- **New Features**
  - Improved performance for retrieving "id" and "filepath" fields with optimized aggregation pipelines.
  - Option to return raw MongoDB aggregation pipelines for advanced use cases.

- **Bug Fixes**
  - Ensured the MongoDB document `_id` field is excluded from aggregation results when not needed.
  - Allowed batch aggregation pipelines to customize field names and exclude `_id` explicitly.

- **Tests**
  - Added tests to verify optimized pipeline generation and correct value types for "id", "_id", and "filepath" fields.

### DIFF
--- a/fiftyone/core/aggregations.py
+++ b/fiftyone/core/aggregations.py
@@ -2898,12 +2898,17 @@ class Values(Aggregation):
             context=context,
         )
 
-        self._field = self._manual_field or field
-        self._big_field = big_field
-        self._num_list_fields = len(list_fields)
+        # Optimization: call str() in memory rather than $toString in database
+        if id_to_str:
+            field = fof.ObjectIdField()
+            id_to_str = False
 
-        pipeline.extend(
-            _make_extract_values_pipeline(
+        # Optimization: use field inclusion syntax when possible
+        if self._expr is None and path == "_id":
+            big_field = path
+            _pipeline = [{"$project": {"_id": 1}}]
+        else:
+            _pipeline = _make_extract_values_pipeline(
                 path,
                 list_fields,
                 id_to_str,
@@ -2911,7 +2916,12 @@ class Values(Aggregation):
                 self._big_result,
                 big_field,
             )
-        )
+
+        self._field = self._manual_field or field
+        self._big_field = big_field
+        self._num_list_fields = len(list_fields)
+
+        pipeline.extend(_pipeline)
 
         return pipeline
 
--- a/fiftyone/core/aggregations.py
+++ b/fiftyone/core/aggregations.py
@@ -2899,7 +2899,7 @@ class Values(Aggregation):
         )
 
         # Optimization: call str() in memory rather than $toString in database
-        if id_to_str:
+        if id_to_str and not self._raw:
             field = fof.ObjectIdField()
             id_to_str = False
 
--- a/fiftyone/core/aggregations.py
+++ b/fiftyone/core/aggregations.py
@@ -2903,10 +2903,15 @@ class Values(Aggregation):
             field = fof.ObjectIdField()
             id_to_str = False
 
+        #
         # Optimization: use field inclusion syntax when possible
-        if self._expr is None and path == "_id":
+        #
+        # Note that this is only allowed because these fields are *required*
+        # _make_extract_values_pipeline is required when fields may be missing
+        #
+        if self._expr is None and path in ("_id", "filepath"):
             big_field = path
-            _pipeline = [{"$project": {"_id": 1}}]
+            _pipeline = [{"$project": {big_field: 1}}]
         else:
             _pipeline = _make_extract_values_pipeline(
                 path,
@@ -2991,7 +2996,11 @@ def _make_extract_values_pipeline(
             expr = _extract_list_values(inner_list_field, expr)
 
     if big_result:
-        return [{"$project": {big_field: expr.to_mongo(prefix="$" + root)}}]
+        project = {big_field: expr.to_mongo(prefix="$" + root)}
+        if big_field != "_id":
+            project["_id"] = False
+
+        return [{"$project": project}]
 
     return [
         {"$project": {"value": expr.to_mongo(prefix="$" + root)}},
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -10630,6 +10630,9 @@ class SampleCollection(object):
             if _group_slices:
                 group_slices.update(_group_slices)
 
+            # Optimization: some aggregations are allowed to change field name
+            big_field = getattr(aggregation, "_big_field", big_field)
+
             try:
                 assert len(_pipeline) == 1
                 project[big_field] = _pipeline[0]["$project"][big_field]
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -10642,6 +10642,9 @@ class SampleCollection(object):
                     "$project stage; found %s" % _pipeline
                 )
 
+        if "_id" not in project:
+            project["_id"] = False
+
         return self._pipeline(
             pipeline=[{"$project": project}],
             attach_frames=attach_frames,
--- a/tests/unittests/aggregation_tests.py
+++ b/tests/unittests/aggregation_tests.py
@@ -745,6 +745,57 @@ class DatasetTests(unittest.TestCase):
         self.assertListEqual(values2, expected)
 
     @drop_datasets
+    def test_values_optimized(self):
+        dataset = fo.Dataset()
+        dataset.add_samples(
+            [
+                fo.Sample(filepath="image1.png"),
+                fo.Sample(filepath="image2.png"),
+                fo.Sample(filepath="image3.png"),
+                fo.Sample(filepath="image4.png"),
+                fo.Sample(filepath="image5.png"),
+                fo.Sample(filepath="image6.png"),
+            ]
+        )
+
+        pipeline = dataset.aggregate(fo.Values("id"), _mongo=True)
+        self.assertListEqual(pipeline, [{"$project": {"_id": 1}}])
+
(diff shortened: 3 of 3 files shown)


kaixi-wang:
Interesting... the project _id and hint at the db level I would expect to be doing the same thing...

When I run it, even on a small dataset, hint is always faster...
image

I'll check the explain plans/do a little more investigating

fiftyone/core/aggregations.py
```diff
@@ -2884,20 +2884,30 @@ def to_mongo(self, sample_collection, big_field="values", context=None):
             context=context,
         )
 
-        self._field = self._manual_field or field
-        self._big_field = big_field
-        self._num_list_fields = len(list_fields)
+        # Optimization: call str() in memory rather than $toString in database
```
kaixi-wang, line 2887: not really related to this PR, but would it be possible to not convert it to string? or would that break a lot of things/not be easy to change?

Just thinking about in-memory conversion to string (if done not while iterating over the cursor but done in a list comprehension) doesn't scale well for millions of samples

fiftyone/core/aggregations.py
```diff
@@ -2884,20 +2884,30 @@ def to_mongo(self, sample_collection, big_field="values", context=None):
...
+        if id_to_str:
+            field = fof.ObjectIdField()
+            id_to_str = False
 
-        pipeline.extend(
-            _make_extract_values_pipeline(
+        # Optimization: use field inclusion syntax when possible
+        if self._expr is None and path == "_id":
+            big_field = path
+            _pipeline = [{"$project": {"_id": 1}}]
```
kaixi-wang, line 2895: What is big_field? I think I looked into it once, and it looked like it was just used for frames... I may be wrong, but I don't think frame _ids are used much, just sample_id, frame_number, and in that case, we should project those fields

kaixi-wang:
Also the results on 11million samples: 
image

fiftyone/core/aggregations.py
brimoor: `values()` is a "big" aggregation because it returns IDs one at a time (in separate documents). All other aggregations are true aggregations: they do the combination in the database and return the results in a single document (ie `min()`, `max()`, etc).

Returning `values()` results as "big" style aggregations is necessary to avoid errors when the dataset contains >16MB of IDs.

fiftyone/core/aggregations.py
brimoor: `values("id")` returns strings, `values("_id")` returns ObjectIDs. So it is up to the caller on whether they want to potentially optimize things by directly working with ObjectIDs rather than strings.

You're right that it should be slightly faster to use ObjectIDs. It removes the `$toString` conversion in the database, and `ObjectId` is fewer bytes than `str(ObjectId)` so it would decrease network I/O somewhat.

fiftyone/core/collections.py
```diff
@@ -10244,7 +10244,7 @@ def _add_view_stage(self, stage):
         """
         raise NotImplementedError("Subclass must implement _add_view_stage()")
 
-    def aggregate(self, aggregations):
+    def aggregate(self, aggregations, _mongo=False):
```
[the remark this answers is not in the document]
kaixi-wang: What does _mongo do?

kaixi-wang:
There seems to be a failing test for video counts.

fiftyone/core/collections.py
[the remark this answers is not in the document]
swheaton: returns the pipeline(s) instead of running them.

brimoor:
@swheaton @kaixi-wang okay tests are passing now. Can you please review?

tests/unittests/aggregation_tests.py
```diff
@@ -744,6 +744,57 @@ def test_values_unwind(self):
...
+
+        self.assertEqual(len(ids1), 6)
+        self.assertEqual(len(_ids1), 6)
+        self.assertEqual(len(filepaths1), 6)
+        self.assertEqual(len(ids2), 6)
+        self.assertEqual(len(_ids2), 6)
+        self.assertEqual(len(filepaths2), 6)
+
+        self.assertIsInstance(ids1[0], str)
+        self.assertIsInstance(_ids1[0], ObjectId)
```
kaixi-wang, line 792: Can you include a check that the values line up? I worry removing _id might affect the consistency of the order of results, particularly if values is called on a view with a match stage involving an indexed field.

brimoor: Are you suggesting that changing the last stage of an aggregation pipeline from

```py
{"$project": {"value": ...}}
```

```py
{"$project": {"value": ..., "_id": False}}
```

could change the order of the returned documents? That would be very surprising to me.

brimoor: If you're worried about the interplay of `"_id": False` and previous view stages that leverage indexes, then there would need to be specific tests added to validate that. It wouldn't address your concern to validate the order of the `values()` calls in this new test.

Note that there are existing tests in this module that validate the order of `values()` calls.

swheaton (changes requested):
1. We should not be targeting release branch, dev cut has passed and this affects all users.
2. I don't think there is enough data to support this change yet. Particularly the `{_id: False}` change associated with https://github.com/voxel51/fiftyone/pull/5739, because there exists some cases where this actually makes things way worse (Kacey found this, I have confirmed it). But even the base filepath/id change, which should be more targeted and therefore safer, needs to have varied sample/dataset shapes, remote connection testing, etc. Willing to build and kick off this testing after release using our scripts.

brimoor:
@swheaton ok fair enough, retargeted to `develop` ✅

PR #5480, opened 2025-02-09
Add `map_samples()`

brimoor:
## Change log

Adds a `map_samples()` method to `SampleCollection` that leverages a multiprocessing pool to apply a `map_fcn` to each sample in the collection in parallel, optionally saving any sample edits and/or applying a `reduce_fcn` function to "reduce" the outputs.

### Map

```py
import fiftyone as fo
import fiftyone.zoo as foz

def map_fcn(sample):

view.map_samples(map_fcn, progress=True)

```

```
Batch 04/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 899.01it/s]
Batch 03/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 894.90it/s]
Batch 11/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 900.14it/s]
Batch 06/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 895.61it/s]
Batch 13/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 903.09it/s]
Batch 08/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 895.33it/s]
Batch 05/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 893.26it/s]
Batch 02/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 889.17it/s]
Batch 01/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 888.16it/s]
Batch 09/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 893.69it/s]
Batch 12/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 896.80it/s]
Batch 14/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 903.28it/s]
Batch 10/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 893.63it/s]
Batch 07/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 891.26it/s]
Batch 15/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 905.06it/s]
Batch 16/16: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 911.72it/s]
{'DEER': 5000, 'HORSE': 5000, 'AIRPLANE': 5000, 'FROG': 5000, 'BIRD': 5000, 'AUTOMOBILE': 5000, 'SHIP': 5000, 'CAT': 5000, 'TRUCK': 5000, 'DOG': 5000}102.01it/s]
```

### Map-"reduce"

```py
import fiftyone as fo
import fiftyone.zoo as foz

def map_fcn(sample):
    return sample.ground_truth.label.upper()

def reduce_fcn(sample_collection, values):
    return dict(Counter(values.values()))

counts = view.map_samples(map_fcn, reduce_fcn=reduce_fcn, progress=True)
print(counts)
```

```
Batch 03/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1420.54it/s]
Batch 02/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1410.60it/s]
Batch 01/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1404.99it/s]
Batch 05/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1404.51it/s]
Batch 11/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1420.08it/s]
Batch 06/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1404.55it/s]
Batch 04/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1398.35it/s]
Batch 10/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1413.71it/s]
Batch 08/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1405.86it/s]
Batch 09/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1407.91it/s]
Batch 12/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1418.37it/s]
Batch 13/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1424.65it/s]
Batch 07/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1396.38it/s]
Batch 14/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1457.15it/s]
Batch 15/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1458.80it/s]
Batch 16/16: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 3125/3125 [00:02<00:00, 1451.57it/s]
{'SHIP': 5000, 'DEER': 5000, 'FROG': 5000, 'DOG': 5000, 'HORSE': 5000, 'CAT': 5000, 'TRUCK': 5000, 'AIRPLANE': 5000, 'BIRD': 5000, 'AUTOMOBILE': 5000}
```


minhtuev:
This is awesome, I can check this out and run some benchmarking with evaluate_detections. Should we do another pass with Ray as well? We can do a comparison between Apache Beam vs. Pool Multiprocessing vs. Ray.

docs/source/user_guide/using_datasets.rst
```diff
@@ -5612,6 +5612,204 @@ The benefit of the above approach versus passing ``autosave=True`` to
...
+
+        .. code-block:: python
+            :linenos:
+
+            from collections import Counter
+
+            def map_fcn(sample):
+                return sample.ground_truth.label.lower()
+
+            def reduce_fcn(sample_collection, values):
```
jleven, line 5703: I’m pretty uncomfortable naming this “reduce”. We put ourselves in an awkward position moving forward if we name this “reduce” since it is not a “reduce” function. If we introduce a proper reduce function later, what would we name it? Might users familiar with reduce find this confusing? Let’s please name it “aggregate” (or something else) instead of misnaming it using a word that already has a different meaning. And let’s not refer to “map-reduce” if we aren’t performing a map reduce operation.

brimoor: Note that `reduce_fcn` supports two options:
- This version you've commented on, which I agree is more like an "aggregate function"
- A `ReduceFcn` subclass, which is modeled after Beam's `CombineFn` and is like a traditional reduce operator

For clarity, I could get onboard with adding a new `aggregate_fcn` argument that callers must use if they want to provide an "aggregate function" like this.

This would reserve `reduce_fcn` for the true "reduce" functionality. Thinking ahead, if we want to use `map_samples(map_fcn, reduce_fcn=)` internally in the SDK *and* support swapping out the multiprocessing backend for distributed computing backends (Beam, Ray, etc), we'd need to find some kind of universal interface for defining reduce functions, so that the caller's reduction could be performed by any backend. Comparing the `ReduceFcn` classes from this PR and #5449, this feels tricky to achieve.

brimoor: If we think it is important to support multiple backends in the future, it might make sense to ship `map_samples(map_fcn, aggregate_fcn=)` and *NOT* include a `reduce_fcn=` argument just yet.

After all, most reductions could be expressed idiomatically as [FiftyOne aggregations](https://docs.voxel51.com/user_guide/using_aggregations.html).

jacobsela:
@brimoor Can this launch non-daemon processes?

docs/source/user_guide/using_datasets.rst
kaixi-wang: Agree with Josh and your suggestion to not include reduce. Even the beam docs call it aggregate as it's not really the same thing as a traditional reduce most people are familiar with. Also embedding reduce inside of map() really doesn't make sense for other backends that have a true reducer like spark

brimoor:
@kaixi-wang this PR uses `multiprocessing.Pool`. Beam experiments are in https://github.com/voxel51/fiftyone/pull/5449.

What makes you think there will be memory issues? `iter_samples(autosave=True)` does not things in memory beyond what is sent in each save batch, which by default is around 1MB per worker.

For me, running the example code in the PR description achieves an ~8x speedup over the same operation using a single `iter_samples(autosave=True)` call when using `num_workers=multiprocessing.cpu_count()=16` on my machine. And I see no reason those gains would change for larger datasets.

brimoor:
> @brimoor Can this launch non-daemon processes?

@jacobsela can you elaborate on what you're asking?

kaixi-wang:
> @kaixi-wang this PR uses `multiprocessing.Pool`. Beam experiments are in #5449.
> 
> What makes you think there will be memory issues? `iter_samples(autosave=True)` does not store things in memory beyond what is sent in each save batch, which by default is around 1MB per worker.
> 
> For me, running the example code in the PR description achieves an ~8x speedup over the same operation using a single `iter_samples(autosave=True)` call when using `num_workers=multiprocessing.cpu_count()=16` on my machine. And I see no reason those gains would change for larger datasets.

Oh yeah sorry. Mixed up the PRs. But yes in your example, there “should” be no issue since you are only using values on the sample you are iterating over. However, based on my recent attempts/frustrations to make/do anything with large datasets, I’m cautious in recommending a solution for big data without verifying/testing it… although I do trust that you know this part of the code significantly better than me 

Additionally, the concern with memory specifically stems from thinking about the case where external data/side inputs are required. Since each process worker will require a copy of all the same inputs (except for samples) and there doesn’t appear to be direct support for selectively partitioning/streaming or sharing objects,  this version still might not be ideal for solving the common pain point of adding fields/setting values that are not computed directly from the existing fields.

kaixi-wang:
> > @brimoor Can this launch non-daemon processes?

@jacobsela are you thinking about potential issues relating to nested multiprocessing and that daemon processes can’t create child processes? For example, if  the map fn were to call some method like ‘compute_<metric>’ that used multiprocessing under the hood ?

jacobsela:
@brimoor @kaixi-wang Exactly, daemon processes aren't able to launch their own child processes. I wrote my own similar [util](https://github.com/voxel51/fiftyone_model_training/blob/09472bd51b636a3e6ff6e33a4e2d6377ee3ec6b4/fiftyone/train/utils/multiprocessing.py#L30) for this purpose and because it uses Pool it ran into the same issue.

If we want to parallelize model inference/training/anything that uses a data loader it is pretty useful to be able to launch more nested child processes.

minhtuev:
@jacobsela : that might be a little bit out of scope of what we want to do here, on one hand, I think it's reasonable to assume that `map_fnc` does not launch its child processes. On another hand, since `map_samples` is designed to be a replacement for `iter_samples`, do we use `iter_samples` anywhere in inference/training?

jacobsela:
@minhtuev  I understand, I don't mean for this to be an explicit request. I was just wondering for my own use cases. FWIW the way `compute_embeddings` and `apply_model` are currently implemented does use `iter_samples`.

minhtuev:
@jacobsela : thanks for the input, I have created a ticket to investigate all usage patterns of `iter_samples` in Fiftyone this week. So we can take stock of how much of these we can translate to `map_samples`.

fiftyone/utils/multiprocessing.py
```diff
@@ -0,0 +1,376 @@
...
+            to reduce the map outputs. See above for usage information
+        save (None): whether to save any sample edits applied by ``map_fcn``.
+            By default this is True when no ``reduce_fcn`` is provided and
+            False when a ``reduce_fcn`` is provided
+        num_workers (None): the number of workers to use. The default is
+            :meth:`fiftyone.core.utils.recommend_process_pool_workers`
+        shard_size (None): an optional number of samples to distribute to each
+            worker at a time. By default, samples are evenly distributed to
+            workers with one shard per worker
+        shard_method ("id"): whether to use IDs (``"id"``) or slices
```
minhtuev, line 137: Do we have any recommendations or guidance on when to use slices or ids?

fiftyone/utils/multiprocessing.py
```diff
@@ -0,0 +1,376 @@
...
+
+    if return_outputs:
+        values = {}
+
+    if progress:
+        desc = f"Batch {i + 1:0{len(str(num_batches))}}/{num_batches}"
+        with tqdm(total=total, desc=desc, position=i) as pb:
+            for sample in batch_view.iter_samples(autosave=save):
+                result = map_fcn(sample)
+                if return_outputs and result is not None:
```
minhtuev, line 365: Something we can explicitly call out in docstring, in the aggregate/reduce step we will filter None return value

fiftyone/utils/multiprocessing.py
```diff
@@ -0,0 +1,376 @@
...
+
+    import fiftyone as fo
+    import fiftyone.core.odm.database as food
+    import fiftyone.core.view as fov
+
+    global sample_collection, map_fcn, save, return_outputs, progress
+
+    # Ensure that each process creates its own MongoDB clients
+    # https://pymongo.readthedocs.io/en/stable/faq.html#using-pymongo-with-multiprocessing
+    food._disconnect()
```
minhtuev, line 324: Haha funny acronym!

minhtuev: Do we automatically reconnect the mongodb connection after `map_samples` finishes the computation?

docs/source/user_guide/using_datasets.rst
brimoor: Added `aggregate_fcn` as discussed.

fiftyone/utils/multiprocessing.py
brimoor: ID is robust to concurrent edits to the dataset and should thus be preferred (why it is the default).

fiftyone/utils/multiprocessing.py
brimoor: This code runs in the worker processes, which are destroyed when the operation completes.

fiftyone/core/collections.py
```diff
@@ -3212,6 +3213,243 @@ def _set_labels(self, field_name, sample_ids, label_docs, progress=False):
     def _delete_labels(self, ids, fields=None):
         self._dataset.delete_labels(ids=ids, fields=fields)
 
+    def update_samples(
```
brimoor, line 3216: Added an `update_samples()` method for the case where the user just wants to apply edits to a collection.

fiftyone/core/collections.py
```diff
@@ -3212,6 +3213,243 @@ def _set_labels(self, field_name, sample_ids, label_docs, progress=False):
...
+        unless this method is called in a daemon process (subprocess), in which
+        case no workers are used.
+
+        When only a ``map_fcn`` is provided, this function effectively performs
+        the following map operation with the outer loop in parallel::
+
+            for batch_view in fou.iter_slices(sample_collection, shard_size):
+                for sample in batch_view.iter_samples(autosave=save):
+                    sample_output = map_fcn(sample)
+                    yield sample.id, sample_output
```
brimoor, line 3305: @j053y `map_samples(map_fcn)` now returns a generator as suggested ✅

fiftyone/core/utils.py
```diff
@@ -2294,6 +2292,10 @@ def recommend_thread_pool_workers(num_workers=None):
 def recommend_process_pool_workers(num_workers=None):
     """Recommends a number of workers for a process pool.
 
+    By default, ``multiprocessing.cpu_count()`` workers will be recommended if
+    you are running on macOS or Linux, while a single worker will be
+    recommended on Windows.
```
brimoor, line 2297: TODO: is multiprocessing on Windows more stable now?

Can we use `num_workers=multiprocessing.cpu_count()` by default instead of `num_workers=1`?

fiftyone/utils/multiprocessing.py
```diff
@@ -0,0 +1,572 @@
...
+    Returns:
+        one of the following:
+
+        -   the output of ``reduce_fcn`` or ``aggregate_fcn``, if provided
+        -   a generator that emits ``(sample_id, map_output)`` tuples, if
+            ``return_outputs`` is True
+        -   None, otherwise
+    """
+    if num_workers is None:
+        if multiprocessing.current_process().daemon:
```
brimoor, line 172: Automatically defaulting to no multiprocessing when this code is invoked in a daemon process ✅

fiftyone/utils/multiprocessing.py
```diff
@@ -0,0 +1,572 @@
...
+        lock = ctx.RLock()
+        tqdm.set_lock(lock)
+    else:
+        worker_progress = False
+        lock = None
+
+    if return_outputs:
+        batch_count = multiprocessing.Value("i", 0)
+        sample_count = None
+        queue = multiprocessing.Queue()
```
brimoor, line 231: @j053y using a queue to implement process communication as suggested ✅

fiftyone/utils/multiprocessing.py
```diff
@@ -0,0 +1,572 @@
...
+        sample_count = None
+        queue = None
+
+    pool = ctx.Pool(
+        processes=num_workers,
+        initializer=_init_worker,
+        initargs=(
+            dataset_name,
+            view_stages,
+            pickle.dumps(map_fcn),
```
brimoor, line 247: Using `dill` to serialize the function allows us to support `map_fcn` that are lambda functions with `'spawn'` contexts  ✅ 

Without this, the example code in the docs/PR descriptions only worked for `'fork'` contexts.

fiftyone/utils/multiprocessing.py
```diff
@@ -0,0 +1,572 @@
...
+            batch_count,
+            sample_count,
+            queue,
+            save,
+            worker_progress,
+            lock,
+        ),
+    )
+
+    pb = fou.ProgressBar(total=num_samples, progress=progress)
```
brimoor, line 257: Using `multiprocessing.Queue/Value` allows for a single global progress bar in the main context ✅ 

Users can optionally pass `progress="workers"` if they want to see the cool per-worker progress bars 😎

minhtuev: Nice!

fiftyone/utils/multiprocessing.py
minhtuev: We should add a logging here, in case users are expecting multiprocessing output

fiftyone/utils/multiprocessing.py
minhtuev: I can run a test tomorrow to see if this works on Windows as well

fiftyone/utils/multiprocessing.py
```diff
@@ -0,0 +1,572 @@
...
+def _map_samples_single(
+    sample_collection,
+    map_fcn,
+    reduce_fcn=None,
+    aggregate_fcn=None,
+    return_outputs=True,
+    save=False,
+    progress=None,
+):
+    if reduce_fcn is not None:
```
minhtuev, line 343: We probably should raise an error/warning if both `reduce_fnc` and `aggregate_fcn` are passed in

fiftyone/core/utils.py
minhtuev: Crated a ticket to track this, will test tomorrow

fiftyone/utils/multiprocessing.py
minhtuev: Will run a benchmark today to test the performance of the queue implementation ✅

fiftyone/utils/multiprocessing.py
minhtuev: Is there any other reason (other than performance) that users should prefer slices over ids? We do see a slight performance improvement when using slices instead of ids in our benchmark:

```
=== Performance Summary ===
Sorted by execution time:
configuration shard_method  num_workers    duration
slice_16w    slice        16           119.791122
id_16w       id           16           129.170237
slice_8w     slice        8            212.950201
id_8w        id           8            220.344598
slice_4w     slice        4            335.403332
id_4w        id           4            370.033176
slice_2w     slice        2            604.814190
id_2w        id           2            659.206682
baseline     none         0            1051.305521
slice_1w     slice        1            1105.534975
id_1w        id           1            1227.948978

Average duration by shard method:
shard_method
slice      475.698764
id         521.340734
none       1051.305521

Average duration by number of workers:
num_workers
0  1051.305521
1  1166.741976
2  632.010436
4  352.718254
8  216.647400
16 124.480679

File: fiftyone/utils/multiprocessing.py [hunk 13]
```diff
@@ -0,0 +1,572 @@
...
+            the map outputs. See above for usage information
+        aggregate_fcn (None): an optional function to aggregate the map
+            outputs. See above for usage information
+        return_outputs (True): whether to return the map outputs. This has no
+            effect when a ``reduce_fcn`` or ``aggregate_fcn`` is provided
+        save (False): whether to save any sample edits applied by ``map_fcn``
+        num_workers (None): the number of workers to use. The default is
+            :meth:`fiftyone.core.utils.recommend_process_pool_workers`. If this
+            value is <= 1, all work is done in the main process
+        shard_size (None): an optional number of samples to distribute to each
```
minhtuev, line 153: I think we should consider removing shard_size since it is confusing to the users. If num_workers * shard_size < n, we will just end up using max_shard_size anyway :)  so might as well not need it

fiftyone/core/collections.py
```diff
@@ -3212,6 +3213,243 @@ def _set_labels(self, field_name, sample_ids, label_docs, progress=False):
...
+            progress (None): whether to render a progress bar (True/False), use
+                the default value ``fiftyone.config.show_progress_bars``
+                (None), or a progress callback function to invoke instead, or
+                "workers" to render per-worker progress bars
+
+        Returns:
+            the output of ``reduce_fcn`` or ``aggregate_fcn``, if provided,
+            otherwise a generator that emits ``(sample_id, map_output)`` tuples
+        """
+        return self._map_samples(
```
kaixi-wang, line 3417: Why not directly call foum.map_samples?

kaixi-wang, line 3417: Why not directly call foum.map_samples?

fiftyone/utils/multiprocessing.py
minhtuev: Verified that multiprocessing can work in windows :)

fiftyone/core/utils.py
minhtuev: Tested on windows, multiprocessing does work for me, didn't run into any issue except the initial fiftyone setup.

PR #5499, opened 2025-02-19
chore: update Dockerfile to handle multiple python versions

findtopher:

A long time ago, in a galaxy pretty close to this one, we made some hard-coded decisions that weren't great, but got us where we needed to be that day.  This change goes back and makes things a little cleaner and easier to automate.

Today's original goal was to change:
```
-# CMD [ \
-#     "python", \
-#     ".fiftyone-venv/lib/python3.11/site-packages/fiftyone/server/main.py", \
-#     "--port", \
-#     "5151" \
-#     ]
```
to 
```
+CMD [ \
+    "python", \
+    "-m", \
+    "fiftyone.server.main", \
+    "--port", \
+    "5151" \
+    ]
```
So the server would run regardless of the python version...  things have, sorta gotten out of hand...

### Test OSS Default Builds (Interactive, source)
```
❯ export RELEASE_VERSION=1.4.0.dev1
❯ make docker
❯ docker run -p 5151:5151 -it local/fiftyone fiftyone --version
FiftyOne v1.4.0.dev1, Voxel51, Inc.
❯ docker run -p 5151:5151 -it local/fiftyone
Python 3.11.11 (main, Feb  4 2025, 04:55:09) [GCC 12.2.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.32.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import fiftyone as fo

In [2]:
```

### Test server build from source artifact
```
❯ docker build -t local/fiftyone:${RELEASE_VERSION}-server --target server .
❯ docker run -p 5151:5151 -it local/fiftyone:${RELEASE_VERSION}-server fiftyone --version
FiftyOne v1.4.0.dev1, Voxel51, Inc.
❯ docker run -p 5151:5151 -it local/fiftyone:${RELEASE_VERSION}-server
[2025-02-20 00:47:14 +0000] [1] [INFO] Running on http://0.0.0.0:5151 (CTRL + C to quit)
```

### Test interactive build from released artifact
```
❯ docker build -t local/fiftyone:1.3.0-interactive --build-arg BUILD_TYPE=released --build-arg FO_VERSION=1.3.0 .
❯ docker run -it --rm local/fiftyone:1.3.0-interactive fiftyone --version
FiftyOne v1.3.0, Voxel51, Inc.
❯ docker run -it --rm -p 5151:5151 local/fiftyone:1.3.0-interactive
Python 3.11.11 (main, Feb  4 2025, 04:55:09) [GCC 12.2.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.32.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]:
```

### Test server build from released artifact
```
❯ docker build -t local/fiftyone:1.3.0-server --build-arg BUILD_TYPE=released --build-arg FO_VERSION=1.3.0 --target server .
❯ docker run -it --rm local/fiftyone:1.3.0-server fiftyone --version
FiftyOne v1.3.0, Voxel51, Inc.
❯ docker run -it --rm local/fiftyone:1.3.0-server
[2025-02-20 00:53:24 +0000] [1] [INFO] Running on http://0.0.0.0:5151 (CTRL + C to quit)
```

## Summary by CodeRabbit

- **New Features**
  - Introduced flexible container build configurations, allowing selection between development and production setups.
  - Enhanced startup commands to ensure smoother server and interactive session launches.

- **Chores**
  - Streamlined and clarified build instructions to improve overall image construction and maintenance.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,8 @@
 #
 # ARGs::
 #
+#   BUILD_TYPE (interactive): Allows interactive or server builds without
+#       editing the Dockerfile
 #   PIP_INDEX_URL (https://pypi.org/simple): Allow the use of caching proxies
 #   PYTHON_VERSION (3.11): The Python base image to use
 #   ROOT_DIR (/fiftyone): The name of the directory within the container that
@@ -24,13 +26,15 @@
 #       -it local/fiftyone
 #
 
+# The type of build to run
+ARG BUILD_TYPE=interactive
+
 # The base python image to build from
 ARG PYTHON_VERSION=3.11
 
 # Collect wheels for future installation
-FROM python:${PYTHON_VERSION} AS builder
+FROM python:${PYTHON_VERSION} AS interactive
 ARG PIP_INDEX_URL=https://pypi.org/simple
-
 # default: use local wheel
 #
 COPY dist dist
@@ -40,14 +44,16 @@ RUN pip --no-cache-dir install -q -U pip setuptools wheel \
         dist/*.whl \
         ipython
 
+FROM python:${PYTHON_VERSION} AS server
+ARG PIP_INDEX_URL=https://pypi.org/simple
 # server: use published pypi package
 #
-# ARG FO_VERSION
-# ENV FO_VERSION=${FO_VERSION}
-# RUN pip --no-cache-dir install -q -U pip setuptools wheel \
-#     && pip wheel --wheel-dir=/wheels \
-#         fiftyone==${FO_VERSION} \
-#         ipython
+ARG FO_VERSION
+ENV FO_VERSION=${FO_VERSION}
+RUN pip --no-cache-dir install -q -U pip setuptools wheel \
+    && pip wheel --wheel-dir=/wheels \
+        fiftyone==${FO_VERSION} \
+        ipython
 
 #
 # Other packages you might want to add to the list above:
@@ -63,9 +69,12 @@ RUN pip --no-cache-dir install -q -U pip setuptools wheel \
 #   pydicom: DICOM images
 #
 
+FROM ${BUILD_TYPE} AS builder
+# This is an empty target because you can't use a variable in `RUN --mount`
+
 
 # Create a smaller image with wheels installed
-FROM python:${PYTHON_VERSION}-slim AS final
+FROM python:${PYTHON_VERSION}-slim AS shared
 ARG PIP_INDEX_URL=https://pypi.org/simple
 
 # The name of the shared directory in the container that should be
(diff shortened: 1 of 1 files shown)


afoley587 (approved):
Tested both the `ipython` and `server` by commenting out the relevant portions of the Dockerfile and building. Seems OK to me! Should we think about adding different targets (one for ipython and one for server) so a user doesn't have to mdoify the dockerfile?

findtopher:
I thought about that, but then I wind up with four targets, because the local 'use the source image' build and the 'use the published image' build would also have to be unique targets... it felt like... it was a lot...

afoley587 (approved):

findtopher:
One more time, @afoley587 - this time actually with the targets you suggested

afoley587:
Is it expected to need to pass a FO_VERSION as a build arg as opposed to copying `dist` as we do for interactive mode? I would expect the two modes to act the same in that regard

findtopher:
Yes, folks for 'interactive mode' are generally doing source builds locally

Server mode are the Dockerfiles we publish to docker hub, and we only use the published packages for those

afoley587:
Additionally, to run the server mode, I had to run 

```shell
docker build -t local/fiftyone . --build-arg=BUILD_TYPE=server --build-arg=FO_VERSION=1.3.0 --target=server-final
```

which isn't too bad. But, we could likely remove the BUILD_TYPE and FO_VERSION and just use targets if we wanted to

findtopher:
you need the BUILD_TYPE because of:
```
FROM ${BUILD_TYPE} AS builder
# This is an empty target because you can't use a variable in `RUN --mount`
```

Otherwise I have to do a lot of duplication in the Dockerfile...

findtopher:
If we hard-code FO_VERSION in the Dockerfile then we have to update that for every release, which I guess is fine, but feels like I could just pass in an arg that I like?

afoley587:
So, in the past releases, we never had the fiftyone version build arg. So I guess my point is "I don't understand why we don't just copy the dist in". This would allow us to build the docker image and pypi package directly from GH actions (possibly parallel). But, if we want to pass the build arg in, then it'll be harder to automate those two together - a GH action would have to be updated (which seems like hard coding still). 

Additionally, now if a user wants to run the server, then can't test their changes locally without modifying the dockerfile, right? 

If we're just copying the dist in, then the shared portion stays the same and the targets just change their CMD args

findtopher:
> So, in the past releases, we never had the fiftyone version build arg.

For every DockerHub published version we've ever published I had to pass in the version, just like the build and deploy script in the Summary.  We extrapolate the version from the git tag that starts the publish process, so we can pass it to a future Docker Build workflow.

> Copying the dist in would allow us to build the docker image and pypi package directly from GH actions.

Since we do multi-architecture builds, using multiple servers, `dist` doesn't exist on half of the architectures.

> Additionally, now if a user wants to run the server, then can't test their changes locally without modifying the dockerfile, right?

Which is the way that it has always been; interactive, source builds typically launch the server from the interactive `ipython` command line.  The intent is to keep the 'default' build (`make docker`) exactly the same as it always has been, while adding ease-of-use for building and deploying the Docker Hub images that, so far, I have done in a very manual way.

afoley587 (approved):
ok, makes sense

findtopher:
OK @afoley587 - one more time - you should be able to run it in any format you feel inclined now

Source build or released build
Interactive CMD or server CMD

The summary has been heavily updated to reflect all the options I can think of

PR #5502, opened 2025-02-19
Pass custom label schema attribute keys to annotation backends

ehofesmann:

Previously, the keys of attribute properties in an annotation label schema were filtered to only include a preconfigured set of values. For example, if you were to pass in this label schema:
```
{
  "gt": {
    "type": "detections", 
    "classes": [
      {
        "classes": ["c1"], 
        "attributes": {"a1": {"type": "select", "values": ["v1", "v2"], "otherattrf": "yes"}
  }
}
```
It would only pass along the "type" and "values" properties and remove the "otherattrf". This means that if a new annotation backend has custom attribute properties, there is no way to define those in the label schema.

The cvat, labelbox, and labelstudio tests were rerun and passed.

Printing the label schema from inside of an annotation backend now shows the custom attributes.

        notes for FiftyOne users.

Annotation label schema attributes now support custom attributes for annotation backends.

-   [x] Other

## Summary by CodeRabbit

- **Refactor**
  - Improved the handling of attribute data to ensure original definitions remain unchanged, enhancing overall stability.

### DIFF
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -955,11 +955,12 @@ def _format_attributes(backend, attributes):
         if not attr:
             attr = backend.recommend_attr_tool(name, None)
 
-        attr_type = attr.get("type", None)
-        values = attr.get("values", None)
-        default = attr.get("default", None)
-        mutable = attr.get("mutable", None)
-        read_only = attr.get("read_only", None)
+        _attr = attr.copy()
+        attr_type = _attr.get("type", None)
+        values = _attr.get("values", None)
+        default = _attr.get("default", None)
+        mutable = _attr.get("mutable", None)
+        read_only = _attr.get("read_only", None)
 
         if attr_type is None:
             raise ValueError(
@@ -978,7 +979,7 @@ def _format_attributes(backend, attributes):
                 )
             )
 
-        _attr = {"type": attr_type}
+        _attr["type"] = attr_type
 
         # Parse `values` property
         if values is not None:
--- a/fiftyone/utils/annotations.py
+++ b/fiftyone/utils/annotations.py
@@ -955,12 +955,12 @@ def _format_attributes(backend, attributes):
         if not attr:
             attr = backend.recommend_attr_tool(name, None)
 
-        _attr = attr.copy()
-        attr_type = _attr.get("type", None)
-        values = _attr.get("values", None)
-        default = _attr.get("default", None)
-        mutable = _attr.get("mutable", None)
-        read_only = _attr.get("read_only", None)
+        _attr = deepcopy(attr)
+        attr_type = _attr.pop("type", None)
+        values = _attr.pop("values", None)
+        default = _attr.pop("default", None)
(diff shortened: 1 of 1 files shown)


fiftyone/utils/annotations.py
```diff
@@ -955,11 +955,12 @@ def _format_attributes(backend, attributes):
         if not attr:
             attr = backend.recommend_attr_tool(name, None)
 
-        attr_type = attr.get("type", None)
-        values = attr.get("values", None)
-        default = attr.get("default", None)
-        mutable = attr.get("mutable", None)
-        read_only = attr.get("read_only", None)
+        _attr = attr.copy()
```
minhtuev, line 958: Should we use `deepcopy` here instead?

fiftyone/utils/annotations.py
```diff
@@ -978,7 +979,7 @@ def _format_attributes(backend, attributes):
                 )
             )
 
-        _attr = {"type": attr_type}
+        _attr["type"] = attr_type
```
minhtuev, line 982: Not a 100% sure how this works, we already are getting `attr_type` from `_attr` in line 959 and `attrr_type` is not modified or updated in between?

fiftyone/utils/annotations.py
```diff
@@ -955,11 +955,12 @@ def _format_attributes(backend, attributes):
...
             attr = backend.recommend_attr_tool(name, None)
 
-        attr_type = attr.get("type", None)
-        values = attr.get("values", None)
-        default = attr.get("default", None)
-        mutable = attr.get("mutable", None)
-        read_only = attr.get("read_only", None)
+        _attr = attr.copy()
+        attr_type = _attr.get("type", None)
+        values = _attr.get("values", None)
```
brimoor, line 960: @ehofesmann FYI I believe this slightly changes how `_attr` is constructed.

For example:
- Previously if `attr["mutable"] == None`, then `_attr` would **not** contain a `"mutable"` key, since it is only added if it is non-`None`
- Now as implemented, `_attr` would contain a `"mutable"` key with value `None`

I'm not sure if that change in behavior breaks anything, but you could maintain the current behavior if you used `pop()` instead of `get()`:

```py
attr_type = _attr.pop("type", None)
values = _attr.pop("values", None)
default = _attr.pop("default", None)
mutable = _attr.pop("mutable", None)
read_only = _attr.pop("read_only", None)
```

fiftyone/utils/annotations.py
brimoor: If @ehofesmann implements my suggestion above, then this line would be required.

As currently implemented, I agree it is a no-op.

fiftyone/utils/annotations.py
brimoor: The current implementation doesn't modify nested values so it doesn't matter. However, agree that `deepcopy` would be more proper.

brimoor (changes requested):

minhtuev (approved):
LGTM, thanks Eric!

brimoor (approved):
LGTM

PR #5506, opened 2025-02-20
add support for arbitrary media types

tom-vx51:

Currently, all datasets must have media types that are natively supported by FiftyOne (image, video, point-cloud, 3d), or groups of such media types. This PR adds support to the SDK and app for arbitrary media types.

```python
ds = fo.Dataset()

# previously an error, now this is allowed
sample = fo.Sample(filepath="/path/to/file.csv", media_type="csv")

ds.add_sample(sample)
print(ds.media_type)
# csv
```

```python
ds = fo.Dataset()

# media type is still inferred for known types
sample = fo.Sample(filepath="/path/to/image.png")

ds.add_sample(sample)
print(ds.media_type)
# image
```

```python
ds = fo.Dataset()

# unspecified media type now defaults to "unknown" if the file path is not recognized as an image, video, etc.
sample = fo.Sample(filepath="/path/to/data.json")

ds.add_sample(sample)
print(ds.media_type)
# unknown
```

Samples with non-native media types are rendered with relevant metadata in the grid.

image

and are shown with metadata and the existing JSON viewer in the modal

image

unit tests, local app

        notes for FiftyOne users.

Add support for creating samples with arbitrary `media_type`.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced a new metadata view that cleanly displays detailed sample information and file properties.
  - Added support for an `UnknownSample` type in the GraphQL schema to handle unrecognized media types.
  - Added utility functions for media type validation, enhancing the application's capability to handle various media formats.
  - Enhanced the `formatDateTime` function for improved flexibility and introduced a new `formatLongDateTime` function.
  - Expanded documentation to clarify the option of specifying media types during sample creation.
  - Added new media types, `unknown` and `custom type`, to enhance dataset handling.

- **Refactor**
  - Streamlined the logic for handling 3D media types, enhancing rendering consistency.
  - Updated media schema to use string types for media types, simplifying the data structure.
  - Simplified media type handling in the server logic by using a default case for unknown types.
  - Refactored the logic for determining media types in various components to utilize utility functions.

- **Bug Fixes**
  - Removed validation for media types in dataset methods to simplify logic.

- **Tests**
  - Expanded test coverage for media type validation utilities to ensure reliability.
  - Modified sample instantiation in unit tests to include media type specification.
  - Added unit tests for new utility functions related to file metadata.


sashankaryal (approved):
👑

brimoor (changes requested):
Can we please not merge this into `develop` until we're sure the SDK is fully wired up to support custom media types in a way we're comfortable with? And we have documentation for it, etc.

brimoor (dismissed):
@tom-vx51 I was playing with this tonight and noticed that the 3D visualizer is not appearing as expected on this branch, both for "flat" 3D datasets and grouped datasets that contain 3D slices:

```py
import fiftyone as fo
import fiftyone.zoo as foz

session = fo.launch_app()

session.dataset = foz.load_zoo_dataset("quickstart-3d")

session.dataset = foz.load_zoo_dataset("quickstart-groups")
```

Can you take a look?

app/packages/looker/src/elements/file.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    return this.element;
+  }
+
+  #handleScaling(state: Readonly<BaseState>) {
+    this.#container.style.scale = `${this.#getScaleFactor(
+      state.windowBBox[2],
+      state.windowBBox[3]
+    )}`;
+  }
```
benjaminpkane, line 106: Love it

app/packages/state/src/hooks/useCreateLooker.ts
```diff
@@ -105,24 +107,39 @@ export default <T extends AbstractLooker<BaseState>>(
...
+        // media types for which we have a dedicated looker
+        const nativeMediaTypes = [
+          // undefined media_type is assumed to be an image
+          undefined,
+          "image",
+          "video",
+          "3d",
+          "point-cloud",
+          "group",
+        ];
```
benjaminpkane, line 119: Maybe this is an opportunity to break out at least the constructor selection logic to a new function? `useCreateLooker` is terribly bloated and nearly untestable at this point 😅

benjaminpkane (commented):
Pretty cool it only took this much to add a new Looker implementation 🫡 

My proposal is we create a `feat/other-media` feature branch for this PR to target. I sense all of the details that will follow will cause unnecessary interruptions on `develop`

benjaminpkane:
> @tom-vx51 I was playing with this tonight and noticed that the 3D visualizer is not appearing as expected on this branch, both for "flat" 3D datasets and grouped datasets that contain 3D slices:
> 
> ```python
> import fiftyone as fo
> import fiftyone.zoo as foz
> 
> session = fo.launch_app()
> 
> session.dataset = foz.load_zoo_dataset("quickstart-3d")
> 
> session.dataset = foz.load_zoo_dataset("quickstart-groups")
> ```
> 
> Can you take a look?

This may be related to removing the `MediaType` enum. Which is ok, although my understanding is we are only adding the `other` type value for this work

tom-vx51:
> my understanding is we are only adding the `other` type value for this work

This was one of the proposed options, but I don't think it really solves the problem. The issue currently is that non-native media types need to set an additional metadata field (i.e. not `media_type`) and have plugins/lookers/etc. check for that field. If we simply add support for a generic "other" `media_type`, then this issue is still true - all we know is that the sample _isn't_ an image/video/etc, but we don't know what it actually is. 

The approach in this PR is to remove the restriction on `media_type` values entirely, and allow/recommend for components to key off of `media_type` as the source of truth for selecting appropriate behavior

tom-vx51:
> the 3D visualizer is not appearing as expected on this branch, both for "flat" 3D datasets and grouped datasets that contain 3D slices

This should be resolved now. These samples were not being correctly identified as 3d due to some media type mismatches.

benjaminpkane:
A few tests failing in the SDK.

I personally don't see any major changes in the SDK that are needed in order to call this ready. Writing the documentation may confirm/deny. Evaluation, models, etc. already validate for an expected media type.

One note is the below yields an `image` dataset, which I do think is an issue. Adding a sample that is not an image should not create an image dataset
```python
ds = fo.Dataset()
sample = fo.Sample(filepath="/path/to/file.csv")
ds.add_sample(sample)
```

tom-vx51:
> One note is the below yields an `image` dataset, which I do think is an issue. Adding a sample that is not an image should not create an image dataset
> 
> ```python
> ds = fo.Dataset()
> sample = fo.Sample(filepath="/path/to/file.csv")
> ds.add_sample(sample)
> ```

Agreed this is not intuitive. This is due to the behavior in [get_media_type()](https://github.com/voxel51/fiftyone/blob/574a5776c232381a8c6c5375c41cdcda3bbbb93a/fiftyone/core/media.py#L23), where we check a few specific cases but default to image. I'm hesitant to change the default, but we could do something like

```python
# ...other checks

mime_type = etau.guess_mime_type(filepath)
if not mime_type.startswith("image/"):
    return OTHER

return IMAGE
```

Are there any specific mime types we'd need to treat differently?

brimoor:
> > One note is the below yields an `image` dataset, which I do think is an issue. Adding a sample that is not an image should not create an image dataset
> > ```python
> > ds = fo.Dataset()
> > sample = fo.Sample(filepath="/path/to/file.csv")
> > ds.add_sample(sample)
> > ```
> 
> Agreed this is not intuitive. This is due to the behavior in [get_media_type()](https://github.com/voxel51/fiftyone/blob/574a5776c232381a8c6c5375c41cdcda3bbbb93a/fiftyone/core/media.py#L23), where we check a few specific cases but default to image. I'm hesitant to change the default, but we could do something like
> 
> ```python
> # ...other checks
> 
> mime_type = etau.guess_mime_type(filepath)
> if not mime_type.startswith("image/"):
>     return OTHER
> 
> return IMAGE
> ```
> 
> Are there any specific mime types we'd need to treat differently?

@tom-vx51 exactly, I think we need to integrate the concepts from https://github.com/voxel51/fiftyone/pull/5440 into this PR.

I was calling it "other" in that PR. However, now that we've decided to allow users to define custom media types, how about adding a fallback "unknown" media type.

Example behavior:

```py
import fiftyone as fo

dataset = fo.Dataset()

# Unknown and undeclared media type, defaults to "unknown"
sample = fo.Sample(filepath="foo.ext")
assert sample.media_type == "unknown"

dataset.add_sample(sample)
assert dataset.media_type == "unknown"
```

Note that, as I called out in https://github.com/voxel51/fiftyone/pull/5440, there are a few places sprinkled throughout the codebase where we could use different/more informative error handling now that we've decided to support non-builtin media types.

Can you take a crack at updating those in the appropriate ways in this PR?

benjaminpkane (approved):
LGTM 🔥

swheaton:
@tom-vx51 this is a potentially breaking change, for example it caused voxelhub tests to [fail](https://github.com/voxel51/voxel-hub/actions/runs/14500439916/job/40678580922?pr=1713).

Now, it could be said that my tests were imprecise and no real use case would rely on the fallback media type being 'image' not 'unknown'. But I don't know if this is the case and in the meantime it's technically a change of behavior.

tom-vx51:
hub strikes again! I admittedly did not test in hub, sorry about that.

No objection to more thorough testing; are there any changes that need to be made here? FWIW it does attempt to determine an appropriate media type based on the resolved mime type, so I do expect things to work in a "real" scenario.

swheaton:
> hub strikes again! I admittedly did not test in hub, sorry about that.
> 
> No objection to more thorough testing; are there any changes that need to be made here? FWIW it does attempt to determine an appropriate media type based on the resolved mime type, so I do expect things to work in a "real" scenario.

I don't think so, just making note. As long as we're sure the mime type checker would catch any image type thrown at it.

tom-vx51:
> As long as we're sure the mime type checker would catch any image type thrown at it.

agreed! it's using the existing eta utils so as long as the mime type resolves appropriately then it should be good https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/media.py#L34

PR #5517, opened 2025-02-25
add reload sample operators

ritch:


app/packages/operators/src/built-in-operators.ts
```diff
@@ -1155,11 +1155,13 @@ export class SetActiveFields extends Operator {
   useHooks(): {
     setActiveFields: (fields: string[]) => void;
   } {
+    // NOTE: useRecoilValue(fos.modal) is always false here
```
ritch, line 1158: @benjaminpkane this seems like a regression in the behavior of `fos.modal` - what I have here is a hack (using currentSample). Any idea what is going on?

brimoor (commented):
@benjaminpkane can you give this a review for @ritch? Note his question [here](https://github.com/voxel51/fiftyone/pull/5517/files#r1973896278)

ritch:
This needs more design discussion before moving forward.

PR #5525, opened 2025-02-28
Added map_samples with multiprocessing

minhtuev:

Added map_samples and update_samples implementation with multiprocessing 

* Added and updated unit tests (✅ )
* Run benchmark test (TBD)

        notes for FiftyOne users.

-   [ ] Other


fiftyone/core/collections.py
```diff
@@ -3212,6 +3214,63 @@ def _set_labels(self, field_name, sample_ids, label_docs, progress=False):
...
+            num_workers (None): Number of workers (if applicable).
+            shard_method ("id"): Method for sharding ('id' or 'slice').
+            progress (None): Whether to show a progress bar.
+            backend (None): Backend execution strategy, defaults to `sequential` if None.
+
+        Returns:
+            Iterator[Any]: Processed sample results.
+        """
+        # Default to sequential if backend is not provided
+        if backend is None:
```
CamronStaley, line 3241: is there a reason we set this here instead of in the params?

fiftyone/core/collections.py
```diff
@@ -3212,6 +3214,63 @@ def _set_labels(self, field_name, sample_ids, label_docs, progress=False):
...
+
+        Args:
+            map_fcn: Function to apply to each sample.
+            num_workers (None): Number of workers (if applicable).
+            shard_method ("id"): Method for sharding ('id' or 'slice').
+            progress (None): Whether to show a progress bar.
+            backend (None): Backend execution strategy, defaults to `sequential` if None.
+        """
+        # Default to sequential if backend is not provided
+        if backend is None:
```
CamronStaley, line 3267: same as above

fiftyone/core/map/map.py
```diff
@@ -28,7 +28,6 @@ def update_samples(
         self,
         sample_collection,
         map_fcn,
-        save: Optional[bool] = None,
```
CamronStaley, line 31: would this break people's implementation who upgrade? I know we want them to use map_samples instead of this way of updating but shouldn't we just document it instead of potentially breaking their integrations?

fiftyone/core/map/process.py
```diff
@@ -0,0 +1,401 @@
...
+        Returns:
+            one of the following:
+            -   a generator that emits ``(sample_id, map_output)`` tuples, if
+                ``return_outputs`` is True
+            -   None, otherwise
+        """
+        if num_workers is None:
+            if multiprocessing.current_process().daemon:
+                num_workers = 1
+            elif fo.config.default_map_workers is not None:
```
CamronStaley, line 149: I see this is a new config value should we be defaulting to None or should we default to something like how it's done for pytorch:

https://github.com/voxel51/fiftyone/blob/72115d64bd4e4aab2e37ef8519ea07f48532607d/fiftyone/utils/torch.py#L1425

fiftyone/core/map/map.py
minhtuev: I'm not sure what you meant, this is equivalent to `autosave` param of `iter_samples`

CamronStaley: Oh my bad I hadn't looked at the rest of the file to see this is part of the new functionality being added in a previous PR / commit. I thought this was an existing method customers had been exposed to and we were modifying the definition potentially breaking their usages.

Plz ignore

fiftyone/core/map/process.py
CamronStaley: also can we unify this in some way across the app? seems confusing to have multiple "num_workers" unless there's a reason map_workers needs to be different

j053y (changes requested):
The changes seemed to have made it to the sample collection class in core again. 

There also seems to be some extra work done in `eval`. 

Let's limit these initial PRs to just the utility functions. The we'll expose them and use them later.

tests/unittests/map.py
j053y, line 1: Can you split this file up into a directory with tests for each component in a corresponding file? It makes the separation of concerns better, reduces file size, and makes targeting specific tests easier.

j053y (changes requested):

tests/unittests/map.py
minhtuev: Sure done

j053y (approved):
:nut_and_bolt:

PR #5528, opened 2025-02-28
add do log file size

CamronStaley:

Adding a log size field to DOs so that we can optionally provide a log size on upload to be returned by the backend if present. If the log size isn't present in the backend we will attempt to fetch it from the cloud storage location (less efficient but still works in cases where X customer has a custom log file but isn't storing that size in mongo like we are).

Ran executor locally and then tested that log file sizes were reflected / stored in mongodb:

Screenshot 2025-02-27 at 9 50 49 PM

        notes for FiftyOne users.

-   [ ] Other


swheaton (approved):
🚢 🇮🇹

PR #5537, opened 2025-03-04
Looker cache "frozen" fixes

benjaminpkane:

The "frozen" intermediary looker cache state is important when creating a new `Spotlight` (layout). Previously shown instances are frozen during this transition, which marks them as neither shown, hidden, nor pending. Pending is a not yet loaded yet hidden instance

Fixes state transitions for these frozen instances. See recording for an issue currently on `develop` related to these transitions

https://github.com/user-attachments/assets/f404dfcf-dc23-4d4f-8e52-e400db7b729b

Unit tests

-   [ ] Other

## Summary by CodeRabbit

- **Refactor**
	- Streamlined rendering and state management for a more consistent and reliable display.
- **Bug Fixes**
	- Enhanced cancellation handling to prevent unintended actions during aborted operations.

These improvements contribute to a smoother and more stable user experience without altering any public features.

### DIFF
--- a/app/packages/looker/src/cache.ts
+++ b/app/packages/looker/src/cache.ts
@@ -67,7 +67,10 @@ export const createCache = <T extends Instance | Lookers = Lookers>(
   const shown = new Map<string, T>();
 
   const get = (key: string) =>
-    shown.get(key) ?? hidden.get(key)?.instance ?? pending.get(key);
+    shown.get(key) ??
+    hidden.get(key)?.instance ??
+    pending.get(key) ??
+    frozen.get(key);
 
   const hide = (key?: string) => {
     if (!key) {
@@ -75,8 +78,9 @@ export const createCache = <T extends Instance | Lookers = Lookers>(
       return;
     }
 
-    const instance = shown.get(key);
+    const instance = shown.get(key) ?? frozen.get(key);
     shown.delete(key);
+    frozen.delete(key);
     if (!instance) {
       return;
     }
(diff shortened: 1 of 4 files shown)


app/packages/looker/src/cache.ts
```diff
@@ -67,16 +67,20 @@ export const createCache = <T extends Instance | Lookers = Lookers>(
   const shown = new Map<string, T>();
 
   const get = (key: string) =>
-    shown.get(key) ?? hidden.get(key)?.instance ?? pending.get(key);
+    shown.get(key) ??
+    hidden.get(key)?.instance ??
+    pending.get(key) ??
+    frozen.get(key);
```
tom-vx51, line 73: do we need `frozen` at all? from what I can tell, we can't just leave the instances in `shown` because they won't get reattached with a new spotlight, but couldn't we just `hide()` them all and skip `frozen`?

benjaminpkane: Hiding the instances ends up thrashing the the LRU cache because it also has a max size bytes threshold. `hidden` is the only state that has a max size / max size bytes setting

tom-vx51: played with it a bit locally and i think i understand what's happening - everything gets moved to `frozen`, but then some elements get moved to `shown` and the remainder go to `hidden`. I thought that everything went to `hidden` first so it makes sense that this is nicer to the cache.

I think we can probably find a way that we're not trying to maintain 4 caches internally, but for another day.

tom-vx51 (approved):

benjaminpkane:
> I think we can probably find a way that we're not trying to maintain 4 caches internally, but for another day.

Of course! We can iterate

PR #5538, opened 2025-03-04
Optimize Query Performance counts with count scans

benjaminpkane:

With Query Performance enabled, excessive aggregation were still being run from the original sidebar mode.

Removing all aggregations except `Count` when query performance is enabled, and ensuring count scans occur when possible, yields at least 4x faster db responses for sample level scalars. Tested with BDD100k (69,863 samples). 172ms db time to 39ms db time.

Server aggregation assertion

## Release Notes
* Optimized sidebar counts for :ref:`Query Performance <app-optimizing-query-performance>` mode

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced a performance toggle for aggregation queries that optimizes data processing.
  - Added a `queryPerformance` flag to aggregation forms to influence performance behavior.

- **Schema Updates**
  - Updated aggregation input forms to include a performance flag.
  - Adjusted dataset and aggregation result fields to provide more flexible, optional outputs.

- **Tests**
  - Enhanced test scenarios to validate the behavior of the performance toggle in aggregation queries.

### DIFF
--- a/app/packages/relay/src/queries/__generated__/datasetQuery.graphql.ts
+++ b/app/packages/relay/src/queries/__generated__/datasetQuery.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<919aae82cf7adfefba901db3a992a34e>>
+ * @generated SignedSource<<a8037981df94264ef3088abf98a8f0a1>>
  * @lightSyntaxTransform
  * @nogrep
  */
@@ -88,7 +88,7 @@ export type datasetQuery$data = {
       readonly slug: string | null;
     } | null;
     readonly " $fragmentSpreads": FragmentRefs<"datasetFragment">;
-  };
+  } | null;
   readonly " $fragmentSpreads": FragmentRefs<"configFragment" | "savedViewsFragment" | "stageDefinitionsFragment" | "viewSchemaFragment">;
 };
 export type datasetQuery = {
--- a/app/packages/relay/src/queries/__generated__/datasetQuery.graphql.ts
+++ b/app/packages/relay/src/queries/__generated__/datasetQuery.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<a8037981df94264ef3088abf98a8f0a1>>
+ * @generated SignedSource<<919aae82cf7adfefba901db3a992a34e>>
  * @lightSyntaxTransform
  * @nogrep
  */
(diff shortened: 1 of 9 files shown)


fiftyone/server/aggregations.py
```diff
@@ -42,6 +42,7 @@ class AggregationForm:
     slices: t.Optional[t.List[str]]
     view: BSONArray
     view_name: t.Optional[str] = None
+    query_performance: t.Optional[bool] = False
```
kaixi-wang, line 45: where does this value come from?/how is it set?

app/packages/relay/src/queries/__generated__/datasetQuery.graphql.ts
```diff
@@ -88,7 +88,7 @@ export type datasetQuery$data = {
       readonly slug: string | null;
     } | null;
     readonly " $fragmentSpreads": FragmentRefs<"datasetFragment">;
-  } | null;
+  };
```
kaixi-wang, line 91: is this safe? Could no dataset due to no permissions cause problems?

kaixi-wang (commented):
So this removes label counts when qp is enabled? Not sure if this should go through product first

fiftyone/core/aggregations.py
```diff
@@ -537,9 +538,15 @@ class Count(Aggregation):
     """
 
     def __init__(
-        self, field_or_expr=None, expr=None, safe=False, _unwind=True
+        self,
+        field_or_expr=None,
+        expr=None,
+        safe=False,
+        _optimize=False,
```
minhtuev, line 545: why don't we set it to true by default? :)

app/schema.graphql
```diff
@@ -238,7 +239,7 @@ type Dataset {
   appConfig: DatasetAppConfig
   info: JSON
   estimatedFrameCount: Int
-  estimatedSampleCount: Int!
+  estimatedSampleCount: Int
```
benjaminpkane, line 242: It's unclear to me why the required `!` was removed. I will follow up

minhtuev (dismissed):
Change looks good to me, thanks Ben 🚢 small comments otherwise

app/packages/relay/src/queries/__generated__/datasetQuery.graphql.ts
benjaminpkane: This shouldn't be changing. I need to figure out why generated output is different. Thanks!

fiftyone/server/aggregations.py
benjaminpkane: It comes from the App setting `@fiftyone/state/queryPerformance` atom

kaixi-wang (approved):
lgtm

kaixi-wang (approved):
👍

PR #5539, opened 2025-03-04
Lasso with point field (updates)

brimoor:
See https://github.com/voxel51/fiftyone-brain/pull/239 for usage.

### DIFF
--- a/fiftyone/server/routes/embeddings.py
+++ b/fiftyone/server/routes/embeddings.py
@@ -235,31 +235,13 @@ class EmbeddingsExtendedStage(HTTPEndpoint):
         return await run_sync_task(self._post_sync, data)
 
     def _post_sync(self, data):
-        point_field = data["pointField"]
         dataset_name = data["datasetName"]
         stages = data["view"]
         patches_field = data["patchesField"]  # patches field of plot, or None
         selected_ids = data["selection"]  # selected IDs in plot
         slices = data["slices"]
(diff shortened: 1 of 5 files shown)


fiftyone/server/routes/embeddings.py
```diff
@@ -235,42 +235,56 @@ async def post(self, request: Request, data: dict) -> dict:
...
                     {
                         "$match": {
-                            point_field: {
+                            points_field: {
                                 "$geoWithin": {
-                                    "$polygon": lasso_points_as_tuples
+                                    "$polygon": [
+                                        [x, y]
+                                        for x, y in zip(
+                                            lasso_points["x"],
```
ritch, line 277: This no longer guarantees that the lasso_points are valid...was that an intentional change?

fiftyone/server/routes/embeddings.py
```diff
@@ -235,42 +235,56 @@ async def post(self, request: Request, data: dict) -> dict:
...
+                )
 
-        if point_field:
-            lasso_points_as_tuples = _get_fiftyone_geowithin(lassoPoints)
             stage = fos.Mongo(
                 [
                     {
                         "$match": {
-                            point_field: {
+                            points_field: {
```
ritch, line 272: I used an `fos.Mongo` stage here just to get things working, however should we convert this to a `GeoWithin`?

PR #5540, opened 2025-03-04
treat detections as non-dense by default

sashankaryal:

Detections are unique in that they are considered "dense" only when they include masks, but "non-dense" when they consist solely of a bounding box (noting that a bounding box is always present). Previously, the application treated detection fields as dense labels by default. This pull request reverses that approach, so now only heatmaps and segmentation fields will be treated as dense-by-default. Users can still customize visibility settings as needed through the dataset app config.

Using two datasets:
- `quickstart`
- `masks-correctness` at https://github.com/voxel51/dataset-forge/tree/main

        notes for FiftyOne users.

Yet to be released feature, so doesn't apply.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced a new label type, "segmentation," in the schema.

- **Refactor**
  - Updated dense label processing by streamlining the supported document categories to include only segmentation and heatmap types.

- **Bug Fixes**
  - Adjusted test cases to reflect the new label type and ensure correct functionality for label visibility.
  - Modified test cases to focus on "predictions" instead of "ground_truth."

- **Chores**
  - Removed sidebar interactions from test setup to simplify the initial state of tests.


benjaminpkane (approved):
🎨

PR #5541, opened 2025-03-05
Disable comparing model by type in model evaluation panel

lanzhenw:

- Disable the comparing evaluation model options based on type and method, since it only makes sense to compare models of the same type and method. 
- Add tooltip to explain the disabled reason

Disabled option tooltips:

Eval on regression type is disabled as expected. 

Screenshot 2025-03-05 at 5 33 59 PM

Updated dropdown selections

Screenshot 2025-03-05 at 5 34 14 PM

How to test it?

Use the code snippet [here](https://docs.voxel51.com/user_guide/evaluation.html#simple-evaluation-default) to generate different embedding type fields and evaluation runs
Then open model eval model and select the comparison

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Enhanced the evaluation view with tooltips for disabled selection options, providing clearer user guidance.
	- Introduced a new property for customizable icon color in the evaluation icon component.
	- Added a new function to compute and sort evaluation keys based on specific criteria.
- **Refactor**
	- Streamlined the logic for filtering and comparing evaluation options to ensure consistent behavior.
- **Chores**
	- Optimized dependency usage by removing unnecessary imports, leading to improved performance.


app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
```diff
@@ -129,17 +137,20 @@ export default function Evaluation(props: EvaluationProps) {
...
+    const currentMethod = currentEval?.info?.config?.method || "";
     const evaluations = data?.evaluations || [];
-    for (const evaluation of evaluations) {
-      const { key, type, method } = evaluation;
-      if (key !== name) {
-        keys.push({ key, type, method });
-      }
-    }
-    return keys;
+    return computeSortedCompareKeys(
```
manivoxel51, line 146: this seems wrong to me. list vs object right?

manivoxel51 (commented):
🍨  chatted on slack 

Let's test that
- different types are disabled if comparing with tooltip
- same types are comparable - valid
```
results = dataset.evaluate_regressions(
    "predictions",
    gt_field="ground_truth",
    eval_key="eval_reg",
)
```
to create different type for testing

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
lanzhenw: Thanks for catching it.

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
```diff
@@ -577,7 +589,10 @@ export default function Evaluation(props: EvaluationProps) {
               borderColor: theme.palette.divider,
             }}
           >
-            <ColorSquare color={KEY_COLOR} />
+            <EvaluationIcon
```
manivoxel51, line 592: option 1 : I think we should keep both the color and eval icon? 🤔 
option 2: Or have EvaluationIcon accept a color and pass KEY_COLOR so the icon has eval color.

The color becomes relevant when comparing evaluation runs and possibly other cases. might be product q

manivoxel51: can follow up with product and separately address

manivoxel51 (approved):
🍨

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/Evaluation.tsx
lanzhenw: Option 2 sounds good. 
Having both did not look pleasant. In figma, one side used colorsquare, the other side used icon only. It feels off, too.

lanzhenw:
Updated the comparison key icon with the compare key color

Screenshot 2025-03-05 at 7 40 20 PM

PR #5551, opened 2025-03-07
fix: wrong check for has_mask

imyhxy:

I was importing the `CrowdHuman` Dataset and export it to the `COCODetectionDataset` which take me more than 2 hours. 😨 So confused! So I profiled the exporting progress and found out that more than 90% of the time is spent on the `_instance_to_coco_segmentation` call, but the `CrowdHuman` doesn't have mask. The reason is that the `has_mask` property of the `Label` returns `bool` result, but the `coco.py` and `cvat.py` still checking it with `None`, which always result in `True`.

This modification change the exporting time from 2hrs+ to 2min in the CrowdHuman dataset.

I have test it on the `CrowdHuman` dataset, which is imported and export to `COCODetectionDataset`

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/coco.py
+++ b/fiftyone/utils/coco.py
@@ -1304,7 +1304,7 @@ class COCOObject(object):
             x, y, w, h = label.bounding_box
             bbox = [x * width, y * height, w * width, h * height]
 
-            if label.has_mask is not None:
+            if label.has_mask:
                 segmentation = _instance_to_coco_segmentation(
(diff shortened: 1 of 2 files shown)


imyhxy:
> Wow. Apologies for the interruption(s) and thank you for investigating and fixing @imyhxy 🙇

No need for apologies, you guy are doing a wonderful library! 🤗

PR #5554, opened 2025-03-07
Fix recomputing frames on a video dataset

benjaminpkane:

Resolves frames issue encountered [by community](https://discord.com/channels/1266527359511564372/1344036099634233344). 

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset.compute_metadata()
frames = dataset.to_frames(sample_frames=True)
frames.compute_metadata()
dataset.to_frames(sample_frames=True)  # error
``` 

Out
```
  File "/Users/ben/code/scratch/recompute.py", line 9, in <module>
    dataset.to_frames(sample_frames=True)  # error
  File "/Users/ben/code/fiftyone/fiftyone/core/collections.py", line 7627, in to_frames
    return self._add_view_stage(fos.ToFrames(**kwargs))
  File "/Users/ben/code/fiftyone/fiftyone/core/dataset.py", line 7481, in _add_view_stage
    return self.view().add_stage(stage)
  File "/Users/ben/code/fiftyone/fiftyone/core/collections.py", line 4299, in add_stage
    return self._add_view_stage(stage)
  File "/Users/ben/code/fiftyone/fiftyone/core/view.py", line 1815, in _add_view_stage
    view = stage.load_view(self)
  File "/Users/ben/code/fiftyone/fiftyone/core/stages.py", line 8512, in load_view
    frames_dataset = fovi.make_frames_dataset(
  File "/Users/ben/code/fiftyone/fiftyone/core/video.py", line 651, in make_frames_dataset
    dataset._sample_doc_cls.merge_field_schema(frame_schema, overwrite=True)
  File "/Users/ben/code/fiftyone/fiftyone/core/odm/mixins.py", line 321, in merge_field_schema
    if field is not None and field.read_only:
AttributeError: 'StringField' object has no attribute 'read_only'
``` 

Account for frame metadata in video_tests.py

## Release Notes

* Fixed recomputing frames on a video dataset

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Video datasets now support frame-level metadata, enabling richer information capture for improved video data management.

- **Bug Fixes**
  - Enhanced data field handling to ensure that designated immutable fields remain protected, preventing unintended modifications and increasing overall application stability.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1425,7 +1425,7 @@ class SampleCollection(object):
                 return None, None, read_only
 
             resolved_keys.append(field.db_field or field.name)
-            read_only = field.read_only
+            read_only = getattr(field, "read_only", False)
             last_key = idx == len(keys) - 1
 
             if last_key and not leaf:
@@ -2499,7 +2499,7 @@ class SampleCollection(object):
         if field is None:
             field = self.get_field(field_name)
 
-        if field is not None and field.read_only:
+        if field is not None and getattr(field, "read_only", False):
             raise ValueError("Cannot edit read-only field '%s'" % field_name)
 
         _field_name, _, list_fields, _, id_to_str = self._parse_field_name(
@@ -2658,7 +2658,7 @@ class SampleCollection(object):
         if field is None:
             field = self.get_field(field_name)
 
-        if field is not None and field.read_only:
+        if field is not None and getattr(field, "read_only", False):
             raise ValueError("Cannot edit read-only field '%s'" % field_name)
 
         _field_name, is_frame_field, _, _, id_to_str = self._parse_field_name(
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1425,7 +1425,7 @@ class SampleCollection(object):
                 return None, None, read_only
 
             resolved_keys.append(field.db_field or field.name)
-            read_only = getattr(field, "read_only", False)
+            read_only = field.read_only
             last_key = idx == len(keys) - 1
 
             if last_key and not leaf:
@@ -2499,7 +2499,7 @@ class SampleCollection(object):
         if field is None:
             field = self.get_field(field_name)
 
-        if field is not None and getattr(field, "read_only", False):
+        if field is not None and field.read_only:
             raise ValueError("Cannot edit read-only field '%s'" % field_name)
 
         _field_name, _, list_fields, _, id_to_str = self._parse_field_name(
@@ -2658,7 +2658,7 @@ class SampleCollection(object):
         if field is None:
             field = self.get_field(field_name)
 
-        if field is not None and getattr(field, "read_only", False):
+        if field is not None and field.read_only:
             raise ValueError("Cannot edit read-only field '%s'" % field_name)
 
         _field_name, is_frame_field, _, _, id_to_str = self._parse_field_name(
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1425,7 +1425,7 @@ class SampleCollection(object):
                 return None, None, read_only
 
             resolved_keys.append(field.db_field or field.name)
-            read_only = field.read_only
+            read_only = getattr(field, "read_only", False)
             last_key = idx == len(keys) - 1
 
             if last_key and not leaf:
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -2176,7 +2176,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
                     error_level,
                 )
             else:
-                if getattr(field, "read_only", False):
+                if field.read_only:
                     field.read_only = False
                     field.save()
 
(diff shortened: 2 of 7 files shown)


swheaton:
Good find @benjaminpkane thank you!

brimoor (dismissed):
Ah this is my bad. It's code smell that there are still a few places where `mongoengine.Field` are encountered in the wild rather than `fiftyone.Field`.

I encountered this bug a few times when I originally added `field.read_only`, which is why there are already a few places that use `getattr(field, "read_only", False)` instead of `field.read_only`.

However, it appears I missed a few places 😢 

Note that there are still places in non `odm` modules that use `field.read_only` directly. I *think* they are safe, but to be honest with you I can't remember exactly when `mongoengine.Field` may slip through...

swheaton (dismissed):
🚢 🇮🇹

swheaton:
Looks like `_cls` in `Metadata` for one

benjaminpkane:
The new solution still yields an error for the code snippet and updated test

swheaton:
Added some more protections, just in case.

fiftyone/core/dataset.py
```diff
@@ -2176,7 +2176,7 @@ def _delete_summary_fields(self, field_names, error_level):
                     error_level,
                 )
             else:
-                if field.read_only:
```
brimoor, line 2179: This change is definitely not necessary. If `field` is a `mongoengine.Field` then the next line will fail too

fiftyone/core/fields.py
```diff
@@ -152,7 +152,9 @@ def matches_constraints(
         ):
             return False
 
-    if read_only is not None and read_only != field.read_only:
+    if read_only is not None and read_only != getattr(
```
brimoor, line 155: This method documents that it only supports `fo.Field` instances, and other stuff in here like `field.info` only supports `fo.Field`, so let's not make this change here.

We'll just confuse our future selves with seemingly unnecessary `getattr`.

fiftyone/core/dataset.py
```diff
@@ -2176,7 +2176,7 @@ def _delete_summary_fields(self, field_names, error_level):
...
+                if getattr(field, "read_only", False):
```
brimoor, line 2179: This change is definitely not necessary. If `field` is a `mongoengine.Field` then `field.save()` will fail too.

fiftyone/core/collections.py
```diff
@@ -1425,7 +1425,7 @@ def _parse_field(self, path, include_private=False, leaf=False):
                 return None, None, read_only
 
             resolved_keys.append(field.db_field or field.name)
-            read_only = field.read_only
+            read_only = getattr(field, "read_only", False)
```
brimoor, line 1428: This *is* unfortunately necessary, albeit only when accessing private stuff

```py
import fiftyone.zoo as foz

# throws an exception w/o this fix
dataset.get_field("ground_truth.detections._cls", include_private=True)
```

fiftyone/core/fields.py
swheaton: I don't think it's that confusing but sure I'll revert my commit and we can get this in

fiftyone/core/fields.py
```diff
@@ -227,10 +229,14 @@ def validate_field(
             )
 
     if read_only is not None:
-        if read_only != field.read_only:
+        if read_only != getattr(field, "read_only", False):
```
brimoor: This method also documents that it only supports `fo.Field`

fiftyone/core/odm/dataset.py
```diff
@@ -107,7 +107,7 @@ def to_field(self):
             db_field=self.db_field,
             description=self.description,
             info=self.info,
-            read_only=self.read_only,
+            read_only=getattr(self, "read_only", False),
```
brimoor: This isn't necessary. `SampleFieldDocument` always has `read_only`

fiftyone/core/odm/dataset.py
```diff
@@ -136,7 +136,7 @@ def from_field(cls, field):
             db_field=field.db_field,
             description=field.description,
             info=field.info,
-            read_only=field.read_only,
+            read_only=getattr(field, "read_only", False),
```
brimoor: This method documents that it only supports `fo.Field`

fiftyone/core/odm/utils.py
```diff
@@ -321,7 +321,7 @@ def get_field_kwargs(field):
         "db_field": field.db_field,
         "description": field.description,
         "info": field.info,
-        "read_only": field.read_only,
+        "read_only": getattr(field, "read_only", False),
```
brimoor: I don't believe the changes in this module are necessary

fiftyone/core/collections.py
swheaton: ok will put back just this one only

jleven: Has this one been put back?

swheaton (approved):
lgtm still 🚢 🇮🇹

jleven (approved):

fiftyone/core/collections.py
swheaton: commit https://github.com/voxel51/fiftyone/pull/5554/commits/a7bd66bf9409be4b5403647127712392551c97d5

PR #5529, opened 2025-02-28
fix infinite initialization of dynamic io fields

ritch:
## What changes are proposed in this pull request?

Fixes an issue that shows up when using the following `fiftyone.operators.type`:

```
inputs = types.Object()
inputs.list(
    "things",
    types.String(),
    label="List of things"
    view=types.AutocompleteView(multiple=True),
)
```

See this [issue](https://github.com/voxel51/fiftyone-plugins/issues/208) for details.

## How is this patch tested? If it is not, please explain why.

Tested using the `edit_dataset_info` operator.

## Release Notes

Fixed an issue with list fields causing a "Maximum update depth exceeded" react error

## Summary by CodeRabbit

- **New Features**
	- Enhanced the behavior of autocomplete inputs for multi-select, ensuring accurate default settings and improved handling when no selections are made.
  
- **Refactor**
	- Optimized the initialization flow in dynamic input components to ensure that setup occurs only once, enhancing overall stability and performance.

PR #5560, opened 2025-03-10
enhancement for empty values for list and str types

imanjra:

This reverts *some* of the null coercion changes from https://github.com/voxel51/fiftyone/pull/5375. Namely:
- for string fields: empty `""` values are *NOT* coerced to `None` if the default value of the field is `""`
- for list fields: empty `[]` values are *NOT* coerced to `None` if the default value of the field is `[]`

Using various operators with list/str type

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced logic for value coercion based on user interaction and schema definitions.
  - Added new functions to determine when values should be coerced, improving data handling.

- **Refactor**
  - Adjusted how input values are processed and stored, ensuring more predictable handling of data.

### DIFF
--- a/app/packages/core/src/plugins/SchemaIO/index.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/index.tsx
@@ -3,6 +3,7 @@ import { cloneDeep, get, set } from "lodash";
 import React, { useCallback, useEffect, useRef } from "react";
 import DynamicIO from "./components/DynamicIO";
 import { clearUseKeyStores, SchemaIOContext } from "./hooks";
+import { coerceValue } from "./utils";
 
 export function SchemaIOComponent(props) {
   const { onChange, onPathChange, id, shouldClearUseKeyStores } = props;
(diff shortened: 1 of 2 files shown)


brimoor (changes requested):
@imanjra based on your [comment here](https://github.com/voxel51/fiftyone/pull/5529#issuecomment-2709207840), it sounds like reverting this null coercion behavior fixes [your concern here](https://github.com/voxel51/fiftyone/pull/5529#discussion_r1977598248). Can we find another solution to your concern on https://github.com/voxel51/fiftyone/pull/5529 than this?

For context- the null coercion change we introduced as part of https://github.com/voxel51/fiftyone/pull/5375 was an intentional change. The motivation was that in this common syntax:

```py
inputs.str("str_field", required=False)
```

the property defaults to `None`. However, in a dynamic form, if the user types something into `str_field` and then deletes it, the prop now takes value `""`. This is a common gotcha for plugin devs, as they don't realize that they need to handle both `None` and `""` in their `execute()` method to handle the case of a user who decides they don't want to use an optional string field after all. In fact, many of the core plugins in `fiftyone-plugins` suffer from this bug today, which is why https://github.com/voxel51/fiftyone/pull/5375 is so convenient; it fixed the bugs without needing to update the plugins themselves.

brimoor:
> Only coerce value to None if it was modified by user and if the default value are not empty ("" for str and [] for list)

Clever! This address my comment 👍

brimoor (approved):
The [updated behavior](https://github.com/voxel51/fiftyone/pull/5560#issuecomment-2714194707) sounds good to me 👍 

Defer to @ritch for code owner approval

ritch (approved):

PR #5563, opened 2025-03-10
fix python panel state change from DynamicIO

imanjra:

fix python panel state change from DynamicIO

Using model evaluation python panel

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Refactor**
  - Improved state updates so that when updating panel data, array values are now entirely replaced rather than merged, ensuring more predictable behavior.
  - Enhanced type safety for panel state management to boost reliability.

### DIFF
--- a/app/packages/operators/src/useCustomPanelHooks.ts
+++ b/app/packages/operators/src/useCustomPanelHooks.ts
@@ -1,4 +1,4 @@
-import { debounce, merge } from "lodash";
+import { debounce, merge, mergeWith } from "lodash";
 import { useCallback, useEffect, useMemo } from "react";
 
 import { usePanelState, useSetCustomPanelState } from "@fiftyone/spaces";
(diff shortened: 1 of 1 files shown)


app/packages/operators/src/useCustomPanelHooks.ts
```diff
@@ -191,8 +191,12 @@ export function useCustomPanelHooks(props: CustomPanelProps): CustomPanelHooks {
   ]);
 
   const handlePanelStateChange = (newState) => {
-    setCustomPanelState((state: any) => {
-      return merge({}, state, newState);
+    setCustomPanelState((state: unknown) => {
+      return mergeWith({}, state, newState, (objValue, srcValue) => {
```
manivoxel51, line 195: looking good!

this would work in case of addition , deletion, and update to an state that is an array. I think we should be good but let's get Ritchie's thoughts on it too

manivoxel51 (commented):
🍨

app/packages/operators/src/useCustomPanelHooks.ts
```diff
@@ -191,8 +191,12 @@ export function useCustomPanelHooks(props: CustomPanelProps): CustomPanelHooks {
...
+        if (Array.isArray(objValue)) {
+          return srcValue; // Overwrite instead of merging arrays
```
lanzhenw, line 197: nice catch

lanzhenw (approved):
Nice catch. LGTM

PR #3824, opened 2023-11-20
Add PIL image rotation according to exif info (PR 3779 alt)

swheaton:
## What changes are proposed in this pull request?

Introduce exif orientation support when importing images into a dataset - see https://github.com/voxel51/fiftyone/issues/3778

## How is this patch tested? If it is not, please explain why.

Tested on the code posted in the Issue linked above.

Before the fix image dimensions are flipped and annotation is displayed in the wrong position:
image

After the fix dimensions are correct and annotation is displayed in the correct position:
image

### Is this a user-facing change that should be mentioned in the release notes?

-   [x] No. You can skip the rest of this section.
-   [ ] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [x] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

PR #5564, opened 2025-03-10
fix slow compute metadata on PNGs due to getexif on png

CamronStaley:

Running compute metadata on PNGs was being reported as unusually slow especially on larger images. These changes should give a decent speedup, but makes it so "flipped" PNGs will no longer be treated as such on the backend.

The reason for the slow down was that getexif will stream through the entire file if called on a PNG. We found that this method doesn't actually work for most other filetypes and only really efficiently works on JPEGS so we are restricting it to that use case.

Uploaded some large PNG's and JPG's to gcloud and then tested how long it took to get metadata:

```
import time
import fiftyone as fo


img_url = "some-url"
start_time = time.time()
metadata = fo.ImageMetadata.build_for(img_url)
end_time = time.time()

print(f"Time taken: {end_time - start_time} seconds")
```

Before changes:

```
Time taken with 812kb PNG: 0.4665961265563965 seconds
Time taken with 331kb JPG: 0.2698838710784912 seconds
Time taken with 3mb PNG: 0.6951353549957275 seconds
Time taken with 3mb JPG: 0.31340694427490234 seconds
```

After changes:

```
Time taken with 812kb PNG: 0.43336963653564453 seconds
Time taken with 331kb JPG: 0.2811110019683838 seconds
Time taken with 3mb PNG: 0.47894811630249023 seconds
Time taken with 3mb JPG: 0.3149549961090088 seconds
```

As you can see JPG sees very little difference in time take while PNG sees a pretty drastic change especially when the file size is larger or internet connection is worse due to it streaming the entire file instead of the first X bytes.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
	- Adjusted image processing so that dimension corrections are applied only when needed for specific image formats.
	- Enhanced error handling to provide clearer feedback during metadata processing, ensuring that exceptions are managed appropriately.

### DIFF
--- a/fiftyone/core/metadata.py
+++ b/fiftyone/core/metadata.py
@@ -459,8 +459,8 @@ def get_image_info(f):
     img = Image.open(f)
 
     # Flip the dimensions if image metadata requires us to. PIL.Image doesn't
-    #   handle by default.
-    if _image_has_flipped_dimensions(img):
+    #   handle by default. Image flipping only efficiently supported on JPEG.
+    if img.format == "JPEG" and _image_has_flipped_dimensions(img):
         width, height = img.height, img.width
     else:
         width, height = img.width, img.height
@@ -555,13 +555,12 @@ def _do_compute_metadata(args):
 def _compute_sample_metadata(
     filepath, media_type, skip_failures=False, cache=None
 ):
-    if not skip_failures:
-        return _get_metadata(filepath, media_type, cache=cache)
-
     try:
         return _get_metadata(filepath, media_type, cache=cache)
     except:
-        return None
+        if skip_failures:
+            return None
+        raise
 
 
 def _get_metadata(filepath, media_type, cache=None):


swheaton:
Here's the original PR introducing EXIF stuff: https://github.com/voxel51/fiftyone/pull/3824

Anddraca (approved):
lgtm! great catch

swheaton (approved):
I do not think EXIF is valid for any type but jpeg. It was [added](https://ftp-osl.osuosl.org/pub/libpng/documents/pngext-1.5.0.html#R.eXIf) to PNG as an extended chunk type. It seems to exist largely for conversion between jpeg or tiff, and PNG.

The original bug was fairly obscure anyways, so as the author (therefore introducer of this bug 😬 ), I am 100% on board with limiting this to JPEG where getting EXIF does not need to load any more data.

PR #5591, opened 2025-03-17
Fix points_field attr reference when using older fiftyone.brain versions

ritch:
This prevents the following error when `fiftyone` is using a version of `fiftyone.brain` that does not include the new `config.points_field` attribute.

> points_field = results.config.points_field\nAttributeError: 'UMAPVisualizationConfig' object has no attribute 'points_field'\n

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced system robustness by adding a conditional check to ensure certain configuration settings are present before use, thereby preventing potential runtime errors.

### DIFF
--- a/fiftyone/server/routes/embeddings.py
+++ b/fiftyone/server/routes/embeddings.py
@@ -76,7 +76,11 @@ class OnPlotLoad(HTTPEndpoint):
 
         patches_field = results.config.patches_field
         is_patches_plot = patches_field is not None
-        points_field = results.config.points_field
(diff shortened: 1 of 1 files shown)


imanjra (approved):
LGTM

brimoor (changes requested):
`fiftyone==1.4.0` will require `fiftyone-brain>=0.20`, so we don't need to make this change. cf https://github.com/voxel51/fiftyone/pull/5589

PR #5596, opened 2025-03-18
[Benchmark] Evaluate detections with map_samples

minhtuev:

(Details)

        notes for FiftyOne users.

-   [ ] Other


fiftyone/utils/map.py
```diff
@@ -46,6 +48,11 @@ def map_samples(
         workers,
         batch_method,
     )
+    print("parallelize_method:", parallelize_method)
+    print("workers:", workers)
+    print("batch_method:", batch_method)
+    print("mapper:", mapper)
+    print("map_fcn:", map_fcn)
```
j053y, line 55: Stray prints.

fiftyone/utils/map.py
```diff
@@ -13,13 +13,15 @@
...
-    skip_failures=True,
+    map_fcn: Callable[[Any], None],
+    workers: Optional[int] = None,
+    batch_method: Union[Literal["id"], Literal["slice"]] = "id",
+    progress: Optional[Union[bool, Literal["worker"]]] = None,
+    save: bool = False,
+    parallelize_method: Union[
+        Literal["process"], Literal["thread"]
+    ] = "process",
+    skip_failures: bool = False,
```
j053y, line 24: I took the types out here to be more in line with the rest of `fiftyone` for now. But if we're gonna keep them let's ensure they are correct and as simplified as possible.

```suggestion
    map_fcn: Callable[[Any], Any],
    workers: Optional[int] = None,
    batch_method: str = "id",    
    progress: Optional[Union[bool, Literal["worker"]]] = None,
    save: bool = False,
    parallelize_method: str = "process",
    skip_failures: bool = False,
```

`batch_method` and `parallelize_method` are both strings because they use a factories (so we can easily add more without updating this file).

fiftyone/utils/map.py
```diff
@@ -56,10 +63,12 @@ def update_samples(
     sample_collection,
```
j053y, line 63: Same as above method, either remove the types or make them correct and simplified.

j053y (changes requested):

fiftyone/utils/map.py
minhtuev: I see OK. I don't think we will be updating `batch_methods` and `parallelize_method` very often, so it might be OK to use literal with finite set; it makes better type hinting in my opinion :)

PR #5597, opened 2025-03-18
docs(deprecations): mongodb 5 deprecation docs

afoley587:

We are beginning to deprecate mongoDB 5. We will still support it but, we should make it be known on `deprecation.voxel51.com` that it is in the process of being deprecated.

This PR should not be in any release notes or anything of that sort as it's purely a docs PR. 

CS has been made aware and will discuss with enterprise customers.

See image below:

Screenshot 2025-03-18 at 8 44 00 AM

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Documentation**
  - Added deprecation details for MongoDB 5.0, including a support end notice on July 1, 2025, and an end-of-life transition effective October 2024.
  - Noted that releases after July 2025 may no longer support MongoDB 5.0 and older.
  - Improved the formatting of the MongoDB 4.4 deprecation information.
- **Bug Fixes**
  - Adjusted version comparison logic to check major versions for MongoDB compatibility, enhancing accuracy in version checks.
  - Updated warning message in tests to clarify that it refers to the oldest supported major version of MongoDB.


afoley587:
@benjaminpkane I believe it would only log for full matches (it's doing a full `==`). We could check major versions only

benjaminpkane:
The warning should probably emit for all versions `>=5` and `<6`

afoley587:
Would you like that as part of this docs PR or should I open a separate one?

benjaminpkane (changes requested):
I think it should be in this PR for completeness 🚀

afoley587:
@benjaminpkane - did a test running with mongo `5.0.4`:

```shell
Type "help", "copyright", "credits" or "license" for more information.
>>> import fiftyone as fo
>>> fo.list_datasets()
You are running the oldest supported version of mongo. Please refer to https://deprecation.voxel51.com for deprecation notices.
[]
>>> 
```

I think checking if the version is between >=5 and <6 is the same as checking major versions. 

Let me know your thoughts!

brimoor (changes requested):
@afoley587 I believe some of this FCV stuff is already sitting on `release/v1.4.0`, so can you please target this at the release branch too? Or alternatively pull the existing stuff out of the release and target `develop` instead? Goal is to ship all or nothing here.

Some requests on the functionality itself:
- We have an existing [`database_validation`](https://github.com/voxel51/fiftyone/blob/a6c5852df81084737ec930650d0c7b591e840a9f/fiftyone/core/odm/database.py#L314) config setting that promises to enable/disable database version checking. Can you update this new logic to respect that setting?
- When you do log stuff, can you include [this message](https://github.com/voxel51/fiftyone/blob/a6c5852df81084737ec930650d0c7b591e840a9f/fiftyone/core/odm/database.py#L317-L321) in your logs explaining that users can suppress this log message by disabling database version checking? I don't like polluting user's shell with log messages they can't get rid of 😄 
- Can you update [this documentation](https://github.com/voxel51/fiftyone/blob/a6c5852df81084737ec930650d0c7b591e840a9f/docs/source/user_guide/config.rst?plain=1#L373-L396) to reflect that we now *require* MongoDB 5 or later?
- Can you add `MONGODB` to the [`MAX_ALLOWABLE_FCV_DELTA`](https://github.com/voxel51/fiftyone/blob/a6c5852df81084737ec930650d0c7b591e840a9f/fiftyone/constants.py#L105C1-L105C24) constant name to clarify that it refers to MongoDB versioning, not FiftyOne? 
- Noting that it seems a little inconsistent that we only do this FCV validation on FiftyOne-managed databases (`_db_service != None`), while [`_validate_db_version()`](https://github.com/voxel51/fiftyone/blob/a6c5852df81084737ec930650d0c7b591e840a9f/fiftyone/core/odm/database.py#L217) is applied in all cases
    - I do agree with the choice to not issue an admin command against a database that the user manually configured. But a warning seems pertinent in all cases, right?

afoley587:
@brimoor closing this in favor of https://github.com/voxel51/fiftyone/pull/5613. That should have all of your notes addressed!

PR #5611, opened 2025-03-19
Bug fix for progress bar in map samples a

minhtuev:

Bug fix for progress bar in map samples and updated map_samples type hinting

* Tested locally

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/map/process.py
+++ b/fiftyone/core/map/process.py
@@ -136,8 +136,11 @@ class ProcessMapper(fomm.LocalMapper[T]):
                     if batch_count.value >= num_batches:
                         break
                 else:
+                    # Update progress bar
+                    pb.update()
+
                     if err is not None:
-                        # When skipping failures, simply yield the the
+                        # When skipping failures, simply yield the
                         # sample ID and the error.
                         if skip_failures:
                             yield sample_id, err, None
@@ -150,8 +153,6 @@ class ProcessMapper(fomm.LocalMapper[T]):
                     else:
                         # Yield successfully mapped sample
                         yield sample_id, None, result
-                finally:
-                    pb.update()
 
             queue.close()
             queue.join_thread()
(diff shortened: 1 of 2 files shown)


fiftyone/core/map/process.py
```diff
@@ -264,3 +266,14 @@ def _map_batch(args: Tuple[int, int, fomb.SampleBatch]):
...
+
+def _process_worker_error(
+    process_skip_failures, process_cancel_event, process_queue, sample_id, err
+):
+    if process_skip_failures:
+        # Cancel other workers as soon as possible.
+        process_cancel_event.set()
+
+    # Add sample ID and error to the queue.
+    process_queue.put((sample_id, err, None))
```
j053y, line 279: No need for a function here if it's not re-used.

fiftyone/utils/map.py
```diff
@@ -13,13 +13,13 @@
...
 def map_samples(
     sample_collection,
-    map_fcn,
-    workers=None,
-    batch_method="id",
-    progress=None,
-    save=False,
-    parallelize_method="process",
-    skip_failures=True,
+    map_fcn: Callable[[Any], None],
```
j053y, line 16:
```suggestion
    map_fcn: Callable[[Any], Any],
```

j053y (changes requested):

fiftyone/core/map/process.py
minhtuev: I keep it because it makes the code a little bit cleaner and easier to read :)

j053y: I would disagree, I think it makes it harder to read. Now I have to bounce around to another method to see what is happening. More methods != greater readability. Especially when the content is only like 3 lines and it isn't reused.

I like to imagine coming back after 6 months without touching this code. Then being able to understand what a function is doing without have bounce around to different functions is ideal. While that's not always achievable in this case it is.

j053y: This is bigger philosophical pattern I'm digging into. 

Is this really a big deal?
With this 1 specific method: No, not necessarily. I agree, it's still readable enough.
With 10 methods like this: Yes it makes everything far harder to understand what the single worker should being doing.

Readability should be targeted not for us, making this feature, but for developers that have never seen this code at all. My point being is let's apply good patterns globally and try not not make exceptions. Functions should only be created with good reason (re-useability, easier testing, etc.). Arbitrarily making functions because it's pretty just makes the code needlessly complicated.

minhtuev: I think these are all good points; personally, I do like functions with smaller bodies, but we can move forward with this change being reverted.

PR #5441, opened 2025-01-27
chore(fiftyone-odm): AS-331 Automated FCV Upgrades

afoley587:
## What changes are proposed in this pull request?

Various versions of mongoDB are being deprecated and moving to EOL. FiftyOne should be able to be updated with those changes. MongoDB has a few version numbers and, of importance, are the server version and the feature compatibility version (FCV). The server version is the version of the binary that is running while the FCV tells the database which server features to be compatible with.

With embedded databases (ones spawned by FO), we should check this FCV and, if required, increase it to the same major version as the server. For reference, see [the mongodb docs](https://www.mongodb.com/docs/manual/release-notes/8.0-upgrade-standalone/#std-label-8.0-upgrade-standalone).

## How is this patch tested? If it is not, please explain why.

Unit tests were added to `tests/unittests/fcv_tests.py`. The mongoDB instance is mocked in those tests so that the database is not corrupted during the other tests (and causes flakes).

Manual tests were also performed.

### Test 1: Nominal Upgrade Path On Debian and Ubuntu VM

```shell
$ sudo apt-get install -y python3-pip python3-venv git
$ python3 -m venv .venv
$ source .venv/bin/activate
$ pip install fiftyone
```

In a python interpreter:

```shell
>>> import fiftyone as fo
>>> import fiftyone.zoo as foz

>>> dataset = foz.load_zoo_dataset("quickstart")
>>> session = fo.launch_app(dataset)
>>> dataset.persistent = True
>>> exit()
```

Simulated a mongo standalone upgrade:
```shell
$ wget -O mongo-next.tgz \
    https://fastdl.mongodb.org/linux/mongodb-linux-aarch64-ubuntu2204-7.0.4.tgz
$ tar xvzf mongo-next.tgz
$ find mongodb-linux-* \
    -name mongod -exec cp {} \
    $HOME/.venv/lib/python3.11/site-packages/fiftyone/db/bin/mongod \;
$ $HOME/.venv/lib/python3.11/site-packages/fiftyone/db/bin/mongod --version
```

Then built fiftyone from source on my branch:

```shell
$ git clone https://github.com/voxel51/fiftyone.git \
    -b afoley/AS-331-pymongo-fcv-updates
$ cd fiftyone
$ bash install.bash
$ export PYTHONPATH=$PWD:$PYTHONPATH
```

Then ran the app to verify the logs:

```shell
>>> import fiftyone as fo
>>> fo.list_datasets()
[...]

PR #5613, opened 2025-03-19
chore: mongodb 5 deprecation and fcv enhancements

afoley587:

We are beginning to deprecate mongoDB 5. We will still support it but, we should make it be known on `deprecation.voxel51.com` that it is in the process of being deprecated.

This PR also builds upon https://github.com/voxel51/fiftyone/pull/5597 but, enhances FCV validation, adds the ability to disable logging, and enhances some tests.

CS has been made aware and will discuss with enterprise customers.

See image below:

Screenshot 2025-03-18 at 8 44 00 AM

        notes for FiftyOne users.

Release notes should be pulled from https://github.com/voxel51/fiftyone/pull/5441

-   [ ] Other

## Summary by CodeRabbit

- **Documentation**
  - Added deprecation details for MongoDB 5.0, including a support end notice on July 1, 2025, and an end-of-life transition effective October 2024.
  - Noted that releases after July 2025 may no longer support MongoDB 5.0 and older.
  - Improved the formatting of the MongoDB 4.4 deprecation information.
- **Bug Fixes**
  - Adjusted version comparison logic to check major versions for MongoDB compatibility, enhancing accuracy in version checks.
  - Updated warning message in tests to clarify that it refers to the oldest supported major version of MongoDB.


fiftyone/core/odm/database.py
```diff
@@ -222,17 +222,16 @@ def establish_db_conn(config):
     connect(config.database_name, **_connection_kwargs)
 
     db_config = get_db_config()
+
+    _update_fc_version(config.database_validation, db_config.type, _client)
```
benjaminpkane, line 226: Why does this always happen now? Should it at least happen if and only if a migration will also be attempted?

afoley587: In a prior discussion between us and topher, we noted this should always happen (not only during migrations). I can certainly change that to move it into migrations!

afoley587: Additionally, I can move the invocation of this to:

```
    if os.environ.get("FIFTYONE_DISABLE_SERVICES", "0") != "1":
        _update_fc_version(config.database_validation, db_config.type, _client) <---- here
        fom.migrate_database_if_necessary(config=db_config)
```

brimoor: Doing things to the database was indeed the intention behind the "admin migration" functionality that occurs when `migrate_database_if_necessary()` is called.

In fact, we used to have an [0.7.1 admin migration](https://github.com/voxel51/fiftyone/blob/v0.15.0/fiftyone/migrations/revisions/admin/v0_7_1.py) that set feature compatibility versions! However, at some point we [deleted that migration](https://github.com/voxel51/fiftyone/tree/develop/fiftyone/migrations/revisions/admin).

I am fuzzy on the details, but I think the reason we removed this migration is that all upward migrations are automatically run when a user imports FO for the first time on a fresh install, and we didn't necessarily want to blindly set feature compatibility versions if it was not necessary.

This PR no longer *always* sets feature compatibility versions, but I remain somewhat scared about taking actions like this automatically...

brimoor: @benjaminpkane @afoley587 I forget: how are MongoDB binaries distributed with FO? Can we assume that users that rely on FO's "managed" Mongo installs will automatically get new mongo versions when they upgrade their `fiftyone` install if we want them to?

I'm not fully groking why we need to mess with feature compatibility version flags now.

afoley587: I can strip out any actionable commands (i.e. any `client.admin.command(cmd)`) and leave the log messages in if that seems more reasonable. If you guys have walked this path before, you have WAY more context than me.

afoley587: @brimoor When a new `fiftyone.db` is installed, it will clobber the old binary. We will shortly need to start managing those binaries as mongo 5.0 is EOL and we should deprecate it soon.

So, if I do a `pip upgrade` of fiftyone, it's very likely a new binary will overwrite my old one and the data will be unusable. The FCV will allow the binary to vary for 1 major version. This assumes someone doesn't jump from a fiftyone version with a 4.4 db -> fiftyone version with a 6.0.

We can, alternatively, leave in the warning messages and direct them to a page which says "here's how to upgrade your FCV"

brimoor: If our goal is just to tell users that they're using an unsupported MongoDB version, isn't that what the `_validate_db_version()` method we already have (which relies on `MIN_MONGODB_VERSION`) provides for us?

afoley587: Also, the binaries can differ by 1 version seamlessly. The issue is if someone then upgrades another time and the binary version (server + fcv) vary by more than 1, mongo will just fail to start.

brimoor: Isn't FCV about preventing a newer MongoDB version from persisting data which is incompatible with earlier versions of the database?

If we actually want to use some feature of a newer MongoDB that makes a breaking change to persisting data, wouldn't we need to bump `MIN_MONGODB_VERSION` and include some kind of "admin migration"?

brimoor: Upgrading a MongoDB binary and then forcing it to behave like an older version seems counterproductive to a goal of ridding ourselves of supporting older MongoDB versions.

brimoor: Not trying to be difficult, I actually just don't understand what problem we're solving for here.

benjaminpkane: I will share my thoughts tomorrow AM when I have more headspace for this 😅

afoley587: You're correct in your understanding of FCV. I guess the question becomes:

what happens when we force a mongodb server of 5.0 instead of 4.0? After that, we would force a minimum version of 6.0, right? What if the user has never upgraded FCV to 5.0? Then fiftyone can't bring up their DB, have to roll back the binary in `site-packages`, etc.

As a note, this is ONLY for the databases that fiftyone manages itself. Excluded from this is any database that is external to our `DatabaseService` that we internally start.

afoley587: i think that, at a minimum, we should log something that says "hey, you're running a 6.0 db with an fcv of 5.0, you might want to do something about that. If you upgrade fiftyone again, you'll have to downgrade your mongodb binary" even if we don't do it automatically

brimoor: I thought FCV was only relevant if you want to GUARANTEE that a database can be downgraded to an older MongoDB version without manually removing any data. IE it is relevant when downgrading, not upgrading.

In our case, if we really want to start using a new feature, we bump `MIN_MONGODB_VERSION` and it is what it is. You can no longer use FO with older MongoDB versions than that.

brimoor: And if we decide to do a public service by forcing users to respect MongoDB EOL versions, we can again achieve that by bumping `MIN_MONGODB_VERSION`, thus forcing users to set `database_validation=False` if they want to ignore our advice.

afoley587: It's not really about us using a new feature, it's about someone automatically getting a new binary when they upgrade fiftyone. But, if we want to provide upgrade paths to users like:

* upgrade to fiftyone x.x.x
* upgrade your fcv to y.0
* profit

that's cool too. The issue is when we deprecate more verisons of mongo, their databases will just cease to work/start and fiftyone will report the `ServiceTimeoutError`. At that point, we'll want a runbook to respond to issues and point to people to the right place in the docs

afoley587: For posterity, here's the error if FCV isn't bumped:

```
{"t":{"$date":"2025-03-20T00:07:41.393+00:00"},"s":"I",  "c":"STORAGE",  "id":4795906, "ctx":"initandlisten","msg":"WiredTiger opened","attr":{"durationMillis":613}}
{"t":{"$date":"2025-03-20T00:07:41.393+00:00"},"s":"I",  "c":"RECOVERY", "id":23987,   "ctx":"initandlisten","msg":"WiredTiger recoveryTimestamp","attr":{"recoveryTimestamp":{"$timestamp":{"t":0,"i":0}}}}
{"t":{"$date":"2025-03-20T00:07:41.430+00:00"},"s":"W",  "c":"CONTROL",  "id":22178,   "ctx":"initandlisten","msg":"/sys/kernel/mm/transparent_hugepage/enabled is 'always'. We suggest setting it to 'never' in this binary version","tags":["startupWarnings"]}
{"t":{"$date":"2025-03-20T00:07:41.430+00:00"},"s":"W",  "c":"CONTROL",  "id":5123300, "ctx":"initandlisten","msg":"vm.max_map_count is too low","attr":{"currentValue":262144,"recommendedMinimum":1677720,"maxConns":838860},"tags":["startupWarnings"]}
{"t":{"$date":"2025-03-20T00:07:41.432+00:00"},"s":"F",  "c":"CONTROL",  "id":20573,   "ctx":"initandlisten","msg":"Wrong mongod version","attr":{"error":"UPGRADE PROBLEM: Found an invalid featureCompatibilityVersion document (ERROR: Location4926900: Invalid featureCompatibilityVersion document in admin.system.version: { _id: \"featureCompatibilityVersion\", version: \"5.0\" }. See https://docs.mongodb.com/master/release-notes/6.0-compatibility/#feature-compatibility. :: caused by :: Invalid feature compatibility version value '5.0'; expected '6.0' or '6.3' or '7.0'. See https://docs.mongodb.com/master/release-notes/6.0-compatibility/#feature-compatibility.). If the current featureCompatibilityVersion is below 6.0, see the documentation on upgrading at https://docs.mongodb.com/master/release-notes/6.0/#upgrade-procedures."}}
```

So our embedded mongo will fail to start. Again, I think that's fine if we tell people - but just want awareness here as we need to start upgrading mongodb versions (5.0 was EOL in october and 4.0 was EOL in feb 14)

afoley587: @brimoor @benjaminpkane let me know your thoughts on the above. If we want to remove the FCV stuff, I can revert f35a7408ab8596dd20a590eefa439937ee16a70d on the release branch (just want to do it today).

I think the only thing I really want in is the deprecation notices, so I'm definitely happy to follow whatever you both think is the right path.

If we want to have a more formal "How to deprecate mongodb 5" conversation later, that's also great with me. And, if that's really just bumping the binary versions in `package/db` - that's even easier - this will just be the first time we're doing this so we will want to make an easy process to roll out when 6 / 7 go EOL in a few months

findtopher: > I thought FCV was only relevant if you want to GUARANTEE that a database can be downgraded to an older MongoDB version without manually removing any data. IE it is relevant when downgrading, not upgrading.

No, [here](https://www.mongodb.com/community/forums/t/mongodb-4-0-to-5-0-migration/191711) are the upgrade instructions for 4-5 - you have to bump the FCV as part of the upgrade or you don't get the 5.0 features.

More importantly, if you don't bump the FCV, when we upgrade to 6.0 anybody who is still in 4.4. compatibility mode will fail the upgrade (and have to roll back to 5 to bump the FCV).

I'm confused by the suggestion that we might make OSS users, who aren't managing a MongoDB, log into their database and execute upgrade commands; isn't the entire idea of `pip install fiftyone` that the simplest form of this is managed by us?  Users who have external databases are not impacted by this change; Enterprise customers are not impacted by this change...

For OSS users who were on 4.4, then got bumped to 5.0 (when fiftyone-db bumped them) they have been running in 4.4 FCV ever since then.  When we bump `MIN_MONGODB_VERSION` to 6.0 in August those users' databases will simply fail to start, creating work and effort where none is required.

I don't understand the pushback to bumping the FCV, as instructed by MongoDB during upgrades.

benjaminpkane: Yes, agree Topher!

benjaminpkane (approved):
The essence of this PR is to address the MongoDB 4 -> 5 transitiion which prepares us for a 5 -> 6 transition.

I think this is PR necessary for the same reasons @findtopher has outlined for builtin DB contexts that we plan to automatically upgrade

fiftyone/core/odm/database.py
```diff
@@ -420,13 +434,21 @@ def _update_fc_version(client: pymongo.MongoClient):
...
+
     fc_version, server_version = _get_fcv_and_version(client)
     _logger = _get_logger()
 
-    if _is_fcv_upgradeable(fc_version, server_version):
+    if (
+        database_validation
+        and _is_fcv_upgradeable(fc_version, server_version)
+        and _db_service is not None
+        and client_type == "fiftyone"
```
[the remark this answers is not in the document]
afoley587: Noted - I removed the `client_type` check.

afoley587:
@benjaminpkane @brimoor @findtopher 

How do we feel about merging? Or do I revert all FCV managements and do a docs-only PR.

benjaminpkane:
It seems like this should be docs only until there is more offline discussion. The FCV upgrade won't have impact until 5.0 is dropped, and that isn't happening until July

afoley587:
okay, let me revert the original fcv pr

PR #5615, opened 2025-03-20
Selection id bottleneck

ritch:
Dynamically determines if the embeddings plot should resolve highlighted plot points based on:

 - `points_field` - the points_field must be defined, meaning this view has a corresponding spatial index
 - `filters` - filters must not be provided, since those selected IDs must be resolved by the API
 - appropriate combination of patches|sample plot and grid sources - since only a flat collection can be spatially queried without resolved selected IDs

This way, when server-side id resolution is not required, the ids are never sent.

## Summary by CodeRabbit

- **New Features**
	- Refined selection handling to ensure that content updates and data fetching now adapt to active filters and view settings, resulting in more accurate and consistent displays.

### DIFF
--- a/app/packages/embeddings/src/useExtendedStageEffect.tsx
+++ b/app/packages/embeddings/src/useExtendedStageEffect.tsx
@@ -5,6 +5,7 @@ import { usePanelStatePartial } from "@fiftyone/spaces";
 import { fetchExtendedStage } from "./fetch";
 import { atoms as selectionAtoms } from "./usePlotSelection";
 import { usePointsField } from "./useBrainResult";
+import { shouldResolveSelection } from "./utils";
 
 export default function useExtendedStageEffect() {
   const datasetName = useRecoilValue(fos.datasetName);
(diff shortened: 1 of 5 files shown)


app/packages/embeddings/src/useExtendedStageEffect.tsx
```diff
@@ -23,11 +24,17 @@ export default function useExtendedStageEffect() {
 
   useEffect(() => {
     if (loadedPlot && Array.isArray(selection)) {
+      const shouldIncludeSelection = shouldResolveSelection(
+        view,
+        null,
```
lanzhenw, line 29: why is filter is `null` here? Can users lasso select on filters?

sashankaryal (approved):

app/packages/embeddings/src/useExtendedStageEffect.tsx
ritch: The extended stage is either a

`fos.Mongo` - for using the lasso points to create a spatial query, which **does not** require `selection` (the sample or patch IDs)

or

`fos.Select`,  `fos.MatchLabels`, `fos.SelectBy` - which do require the selection to be present.

Note that we cannot use the presence of just the `pointsField` - since there are other requirements for allowing spatial querying such as when `SelectBy` is used to match the ids of a patch plot with each `patch.sample_id`.

PR #5616, opened 2025-03-20
Fix incorrect `$geoWithin` injection in embeddings backend

ritch:
Fixes an issue where the `$geoWithin` `fos.Mongo` stage was being added incorrectly. Now the embeddings backend will ensure that the plot source and the view source are the same before injecting the `$geoWithin` stage.

**You should never see this**

We should never see a `ToPatches` stage w/ a `Mongo` stage here (after clicking the bookmark).

image

We also ensure that this stage is never injected in the other combination: sample view and patches plot. The only way for the panel to handle this currently is by the old mechanism: `fos.MatchLabels`.

## Summary by CodeRabbit

- **Refactor**
	- Refined the internal logic for processing selections to improve consistency between view types and corresponding plot sources, ensuring smoother handling of selection actions.
	- Updated the constant used for identifying patches views to enhance clarity in logic comparisons.

### DIFF
--- a/app/packages/embeddings/src/utils.ts
+++ b/app/packages/embeddings/src/utils.ts
@@ -6,7 +6,7 @@ export type Filters = {
   [key: string]: {};
 } | null;
 
-const PATCHES_VIEW_STAGE = "fiftyone.core.stages.PatchesView";
+const PATCHES_VIEW_STAGE = "fiftyone.core.stages.ToPatches";
 
 export function shouldResolveSelection(
   view: ViewStage[],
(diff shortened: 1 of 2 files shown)


app/packages/embeddings/src/utils.ts
```diff
@@ -6,7 +6,7 @@ export type Filters = {
   [key: string]: {};
 } | null;
 
-const PATCHES_VIEW_STAGE = "fiftyone.core.stages.PatchesView";
+const PATCHES_VIEW_STAGE = "fiftyone.core.stages.ToPatches";
```
ritch, line 9: This fixes a slightly different issue, where the wrong view stage was used to determine whether samples view was using `ToPatches`.

imanjra (approved):
LGTM

PR #5623, opened 2025-03-21
Bump brain version

brimoor:
Would be nice to require `fiftyone-brain>=0.20.1` so that we can ensure `fiftyone==1.4.0` users won't encounter the small bug fixed in https://github.com/voxel51/fiftyone-brain/pull/246


findtopher (approved):
:shipit:

PR #5642, opened 2025-03-25
Project Map Samples: implementation of `map_samples` and `update_samples`

minhtuev:

This pull request introduces efficient methods for iterating over and modifying samples in FiftyOne datasets and views. These new methods, `map_samples` and `update_samples`, provide powerful, parallelized interfaces for sample processing, significantly improving performance and usability.

### Contributors
- Tech lead: @j053y 
- Contributors: @minhtuev, @Anddraca 

### Key Changes
New Methods in `SampleCollection`:

- `map_samples`: Applies a function to each sample in parallel and yields results as a generator.
- `update_samples`: Applies a function that updates each sample in-place and saves the results in parallel.
- `update_samples`: Updates samples in-place by applying a specified function.

### Enhancements to Parallelization:

- Added support for multithreading and multiprocessing through configurable parameters such as `batch_method`,  `batch_size`, `parallelize_method`, and `workers`.
- New `fo.config.default_map_workers` setting allows global control over parallel worker count.
- Automatically disables multiprocessing when invoked in daemon processes for compatibility.
- Supports both `spawn` and `fork` multiprocessing contexts.
- `shard_size` and `shard_method` provide control over batching strategies.

- Introduced options to show progress bars for each worker (progress="workers").

### Core Library Updates:

- Added a new module: `fiftyone.core.map`, containing the batching and parallelization logic.
- Introduced MapperFactory for creating mappers to handle various parallelization strategies (e.g., threading, processing).

### Unit Tests:

- New test cases added to `dataset_tests.py` and `view_tests.py` to ensure the correctness and reliability of `map_samples` and `update_samples`.
- Additional test files for the map module, including batching strategies (id_batch, slice_batch) and parallelization methods (process, threading).

### Configuration Updates:

- Added support for default parallelization settings in fiftyone.core.config.
- Introduced environment variables for controlling thread and process pool workers:
FIFTYONE_DEFAULT_PARALLELIZATION_METHOD
FIFTYONE_DEFAULT_THREAD_POOL_WORKERS
FIFTYONE_DEFAULT_PROCESS_POOL_WORKERS

### Bug Fixes and Refactoring:

- Fixed multiprocessing context issues on macOS and Linux by defaulting to fork for Python versions below 3.14.
- Refactored database connection logic to ensure proper cleanup and disconnection when necessary.

* Manual testing (✅ )
* Added unit tests (✅ )

        notes for FiftyOne users.

This release introduces map_samples and update_samples methods to FiftyOne, enabling efficient, parallelized sample iteration and modification. These methods provide significant performance improvements for large datasets and include flexible options for batching, parallelization, and progress monitoring.

-   [ ] Other

### Example 1: update_samples()

```python
import fiftyone as fo
import fiftyone.zoo as foz

def update_fcn(sample):

view.update_samples(update_fcn)

```

```
Loading existing dataset 'cifar10-train'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 50000/50000 [5.1s elapsed, 0s remaining, 34.3K samples/s]
{'AIRPLANE': 5000, 'CAT': 5000, 'AUTOMOBILE': 5000, 'DEER': 5000, 'SHIP': 5000, 'DOG': 5000, 'BIRD': 5000, 'TRUCK': 5000, 'FROG': 5000, 'HORSE': 5000}
```

You can optionally show per-worker progress bars via `progress="workers"`:

```python
view.update_samples(update_fcn, progress="workers")
```
```
Batch 04/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 899.01it/s]
Batch 03/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 894.90it/s]
Batch 11/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 900.14it/s]
Batch 06/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 895.61it/s]
Batch 13/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 903.09it/s]
Batch 08/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 895.33it/s]
Batch 05/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 893.26it/s]
Batch 02/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 889.17it/s]
Batch 01/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 888.16it/s]
Batch 09/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 893.69it/s]
Batch 12/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 896.80it/s]
Batch 14/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 903.28it/s]
Batch 10/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 893.63it/s]
Batch 07/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 891.26it/s]
Batch 15/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 905.06it/s]
Batch 16/16: 100%|██████████████████████████████████████████████████████████████████| 3125/3125 [00:03<00:00, 911.72it/s]
{'DEER': 5000, 'HORSE': 5000, 'AIRPLANE': 5000, 'FROG': 5000, 'BIRD': 5000, 'AUTOMOBILE': 5000, 'SHIP': 5000, 'CAT': 5000, 'TRUCK': 5000, 'DOG': 5000}
```

### Example 2: map_samples()

```python

import fiftyone as fo
import fiftyone.zoo as foz

def map_fcn(sample):
    return sample.ground_truth.label.upper()

counts = Counter()
for _, label in view.map_samples(map_fcn):
    counts[label] += 1

print(dict(counts))
```

```
Loading existing dataset 'cifar10-train'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 50000/50000 [4.2s elapsed, 0s remaining, 46.9K samples/s]
{'FROG': 5000, 'TRUCK': 5000, 'DEER': 5000, 'AUTOMOBILE': 5000, 'BIRD': 5000, 'SHIP': 5000, 'HORSE': 5000, 'CAT': 5000, 'AIRPLANE': 5000, 'DOG': 5000}
```

You can also show per-worker progress bars:

```python
counts = Counter()
for _, label in view.map_samples(map_fcn, progress="workers"):
    counts[label] += 1

print(dict(counts))
```

```
Loading existing dataset 'cifar10-train'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
Batch 07/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4167/4167 [00:01<00:00, 2921.37it/s]
Batch 12/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4167/4167 [00:01<00:00, 2890.01it/s]
Batch 08/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4166/4166 [00:01<00:00, 2880.94it/s]
Batch 10/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4167/4167 [00:01<00:00, 2707.27it/s]
Batch 04/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4167/4167 [00:01<00:00, 2689.78it/s]
Batch 02/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4166/4166 [00:01<00:00, 2676.12it/s]
Batch 03/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4167/4167 [00:01<00:00, 2467.76it/s]
Batch 05/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4166/4166 [00:01<00:00, 2301.27it/s]
Batch 06/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4167/4167 [00:01<00:00, 2224.42it/s]
Batch 11/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4166/4166 [00:01<00:00, 2111.78it/s]
Batch 01/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4167/4167 [00:02<00:00, 2021.93it/s]
Batch 09/12: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4167/4167 [00:02<00:00, 1934.60it/s]
{'HORSE': 5000, 'FROG': 5000, 'BIRD': 5000, 'TRUCK': 5000, 'AUTOMOBILE': 5000, 'DOG': 5000, 'DEER': 5000, 'SHIP': 5000, 'CAT': 5000, 'AIRPLANE': 5000}   | 3741/4166 [00:01<00:00, 3783.06it/s]
Batch 06/12:  83%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████▎                      | 3468/4167 [00:01<00:00, 3470.37it/s]
```

## Summary by CodeRabbit

- **New Features**
  - Introduced parallelized sample mapping and updating capabilities with flexible batch methods and worker configurations.
  - Added configuration options for default parallelization method and worker thread/process counts.
- **Bug Fixes**
  - Enhanced unique filename generation to ensure correctness and uniqueness across multiprocessing contexts.
- **Tests**
  - Added extensive unit tests covering sample mapping implementations, batching strategies, factory creation, and unique filename generation.
- **Documentation**
  - Minor formatting improvements to user guide documentation.


fiftyone/core/collections.py
```diff
@@ -3898,6 +3897,93 @@ def list_evaluations(self, type=None, method=None, **kwargs):
...
+        self,
+        map_fcn: Callable[[Any], Any],
+        *_,
+        workers: Optional[int] = None,
+        batch_method: str = "id",
+        batch_size: Optional[int] = None,
+        progress: Optional[Union[bool, Literal["worker"]]] = None,
+        parallelize_method: str = "process",
+        skip_failures: bool = False,
+        save: bool = False,
```
swheaton, line 3910: sorry no types in this module

fiftyone/core/collections.py
```diff
@@ -3898,6 +3897,93 @@ def list_evaluations(self, type=None, method=None, **kwargs):
...
+        Applies `map_fcn` to each sample using the specified backend strategy
+        and returns an iterator.
+
+        Args:
+            map_fcn: Function to apply to each sample.
+            workers (None): Number of workers.
+            batch_method: Explicit method for batching samples. Supported
+              methods are 'id' and 'slice'
+            batch_size (None): Number of samples per batch. If None, the batch size
+                is automatically calculated to be the number of samples per worker.
```
swheaton, line 3922: Wondering if this is a wise default. The smaller the batches, the more optimal the queueing will be. Splitting into exactly N batches for N workers is the least optimal. I do not know if this would have a material impact in real scenarios.
Smaller batches are nicer for checkpointing also - if a batch of 1M fails you may not know where it ended and so have to do the whole 1M again.

fiftyone/core/collections.py
```diff
@@ -3898,6 +3897,93 @@ def list_evaluations(self, type=None, method=None, **kwargs):
...
+    ):
+        """
+        Applies `map_fcn` to each sample using the specified backend strategy
+        and returns an iterator.
+
+        Args:
+            map_fcn: Function to apply to each sample.
+            workers (None): Number of workers.
+            batch_method: Explicit method for batching samples. Supported
+              methods are 'id' and 'slice'
```
swheaton, line 3920: The reader will have no idea what these methods are.

fiftyone/core/config.py
```diff
@@ -244,6 +244,27 @@ def __init__(self, d=None):
...
+            "default_thread_pool_workers",
+            env_var="FIFTYONE_DEFAULT_THREAD_POOL_WORKERS",
+            default=None,
+        )
+        self.default_process_pool_workers = self.parse_int(
+            d,
+            "default_process_pool_workers",
+            env_var="FIFTYONE_DEFAULT_PROCESS_POOL_WORKERS",
+            default=None,
+        )
```
swheaton, line 266: how is this semantically different from max_workers vars a few lines below?

swheaton (changes requested):
Couple initial comments. More to come.
Address the code rabbit comments though!

fiftyone/core/collections.py
```diff
@@ -3898,6 +3897,93 @@ def list_evaluations(self, type=None, method=None, **kwargs):
...
+        )
+
+        yield from mapper.map_samples(
+            self,
+            map_fcn,
+            progress=progress,
+            save=save,
+            skip_failures=skip_failures,
+        )
+
```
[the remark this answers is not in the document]
minhtuev: Thanks, fixed

fiftyone/core/collections.py
```diff
@@ -3898,6 +3897,93 @@
...
+        for _ in mapper.map_samples(
+            self,
+            update_fcn,
+            progress=progress,
+            save=True,
+            skip_failures=skip_failures,
+        ):
+            ...
+
```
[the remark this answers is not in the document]
minhtuev: Same

fiftyone/core/collections.py
minhtuev: Hmm true, I haven't done any benchmarking here. Thoughts? @j053y @Anddraca

tests/unittests/map/batcher/test_id_batch.py
```diff
@@ -0,0 +1,89 @@
...
+        sample_collection, workers := 8, batch_size
+    )
+    #####
+
+    expected_batch_size = min(
+        (batch_size if batch_size is not None else len(samples) // workers),
+        max_batch_size,
+    )
+    expected_batch_count = len(samples) // expected_batch_size
+
```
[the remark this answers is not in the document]
minhtuev: No need, this is just for unit testing

fiftyone/core/collections.py
minhtuev: Hmm we can add a couple of sentences to explain them here but it might be too long. Should we add a link to documentation instead?

fiftyone/core/config.py
minhtuev: We do use them slightly differently:
```
workers = (
                config.default_thread_pool_workers
                or fou.recommend_thread_pool_workers()
            )

if config.max_thread_pool_workers is not None:
    workers = min(workers, config.max_thread_pool_workers)
```

I guess max workers is the maximum of number of workers for threads/processes, so another way for admin to configure this behavior. @j053y

j053y: Default is the desired number of x" to use max is the limit. When both supplied in the config it's makes less sense. but the max is used when "x" is manually provided at runtime to limit the number if it exists.

swheaton: I see

fiftyone/core/collections.py
kaixi-wang: > Wondering if this is a wise default. The smaller the batches, the more optimal the queueing will be. Splitting into exactly N batches for N workers is the least optimal.

Do you have evidence of this? From my tests of multithreading, this is not true

fiftyone/core/collections.py
```diff
@@ -3898,6 +3907,97 @@ def list_evaluations(self, type=None, method=None, **kwargs):
...
+        )
+
+        yield from mapper.map_samples(
+            self,
+            map_fcn,
+            progress=progress,
+            save=save,
+            skip_failures=skip_failures,
+        )
+
```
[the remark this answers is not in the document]
swheaton: Comments in order
- Yeah I'm not sure what was intended by `*_`. If it's supposed to denote that the following parameters must be keyword-only arguments, then you want just `*,` without the `_`.
- Type hints are not used in `fiftyone.core` submodule for consistency. Other submodules may have them if it's applied consistently to the whole submodule.
- No, the factory appears to raise an error at the appropriate time

fiftyone/core/map/batcher/id_batch.py
```diff
@@ -0,0 +1,72 @@
...
+
+class SampleIdBatch(fomb.SampleBatch):
+    """Sample batch using ids"""
+
+    @classmethod
+    @functools.cache
+    def get_max_batch_size(cls):
+        """Get max batch size"""
+        return fou.recommend_batch_size_for_value(
+            bson.ObjectId(), max_size=100000
```
kaixi-wang, line 32: Why max at 100_000? (The mongo limit is 100Mb)

fiftyone/core/map/batcher/id_batch.py
```diff
@@ -0,0 +1,72 @@
...
+    @classmethod
+    def split(
+        cls,
+        sample_collection: SampleCollection[T],
+        num_workers: int,
+        batch_size: Optional[int] = None,
+    ) -> List["SampleIdBatch"]:
+        num_workers = max(num_workers, 1)
+
+        ids = sample_collection.values("id")
```
kaixi-wang, line 44: On large datasets, this alone could take a long time. Since id is default, end users will experience a hang with no indication of what's going on.

fiftyone/core/map/batcher/slice_batch.py
```diff
@@ -0,0 +1,56 @@
...
+T = TypeVar("T")
+
+
+class SampleSliceBatch(fomb.SampleBatch):
+    """Sample batch using slices"""
+
+    @classmethod
+    def split(
+        cls,
+        sample_collection: SampleCollection[T],
```
kaixi-wang, line 25: Is this different from a fiftyone e sample collection? You can slice a dataset and dataset view, but I don't know if you can actually slice a sample collection

swheaton (commented):
Checkpointing the review.

fiftyone/core/map/batcher/id_batch.py
swheaton: Yeah we should utilize the find-hint approach here, for sure.

Either
```python
if is full dataset:
    ids = list(sample_collection._sample_collection.find({}, {"_id": True}, hint={"_id": 1}))
```
Or if we don't wanna use protected stuff, we can make a `fast_ids()` method or something like this.

fiftyone/core/map/batcher/slice_batch.py
swheaton: sample collection usually denotes either-dataset-or-view.

fiftyone/core/map/mapper.py
```diff
@@ -0,0 +1,254 @@
...
+            progress (Union[bool, Literal[&quot;workers&quot;]]): Whether or
+              not and how to render progress.
+            save (bool, optional): Whether to save mutated samples mutated in
+              the map function. Defaults to False.
+            skip_failures (bool, optional): Whether to gracefully continue
+              without raising an error if the map function raises an exception
+              for a sample. Defaults to True.
+
+        Yields:
+            MapSampleBatchesReturnType: The sample ID, the exception raised
```
swheaton, line 216: This (`MapSampleBatchesReturnType `) is not actually a defined thing? So just say "a tuple of ... "

fiftyone/core/map/mapper.py
```diff
@@ -0,0 +1,254 @@
...
+        self,
+        sample_collection: SampleCollection[T],
+        map_fcn: Callable[[T], R],
+        *_,
+        progress: Optional[Union[bool, Literal["workers"]]] = None,
+        save: bool = False,
+        skip_failures: bool = True,
+    ) -> Iterator[Tuple[bson.ObjectId, R]]:
+        # If workers if 1 on a the same local machine, no need for the
+        # overhead of trying to parallelize.
```
swheaton, line 231: grammatically incorrect, also not sure what it's trying to say about same local machine.

fiftyone/core/map/process.py
```diff
@@ -0,0 +1,302 @@
...
+        ctx = fou.get_multiprocessing_context()
+
+        if progress is None:
+            progress = fo.config.show_progress_bars
+
+        if progress == "workers":
+            worker_progress = True
+            progress = False
+            lock = ctx.RLock()
+            tqdm.set_lock(lock)
```
swheaton, line 86: wait so we are using `tqdm` instead of eta ProgressBar now? I thought we figured out the issue but I may be wrong? I'm confused.

fiftyone/core/map/process.py
```diff
@@ -0,0 +1,302 @@
...
+                ),
+            )
+
+            sample_errors: List[Tuple[bson.ObjectId, Exception, None]] = []
+
+            # Initialize backoff parameters
+            initial_timeout = 0.1
+            max_timeout = 5.0
+            backoff_factor = 2
+            current_timeout = initial_timeout
```
swheaton, line 151: can we not utilize a back off library?

fiftyone/core/map/process.py
```diff
@@ -0,0 +1,302 @@
...
+                config.default_process_pool_workers
+                or fou.recommend_process_pool_workers()
+            )
+
+        if config.max_process_pool_workers is not None:
+            workers = min(workers, config.max_process_pool_workers)
+
+        return cls(batch_cls, workers, batch_size)
+
+    def _map_samples_multiple_workers(
```
swheaton, line 68: this function is getting a lil lengthy

fiftyone/core/map/process.py
```diff
@@ -0,0 +1,302 @@
...
+                except Empty:
+                    # Apply exponential backoff, but cap at max_timeout
+                    current_timeout = min(
+                        current_timeout * backoff_factor, max_timeout
+                    )
+
+                    # Reset backoff if we've hit too many consecutive timeouts
+                    # This prevents getting stuck with very long timeouts
+                    if current_timeout == max_timeout:
+                        current_timeout = initial_timeout
```
swheaton, line 167: why would it go back to the initial timeout instead of just staying at the max timeout? Seems a little weird to keep looping back around

fiftyone/core/map/process.py
```diff
@@ -0,0 +1,302 @@
...
+                    current_timeout = initial_timeout
+                except Empty:
+                    # Apply exponential backoff, but cap at max_timeout
+                    current_timeout = min(
+                        current_timeout * backoff_factor, max_timeout
+                    )
+
+                    # Reset backoff if we've hit too many consecutive timeouts
+                    # This prevents getting stuck with very long timeouts
+                    if current_timeout == max_timeout:
```
swheaton, line 166: I would use `>=` since it's possible that float precision messes it up

fiftyone/core/map/process.py
```diff
@@ -0,0 +1,302 @@
...
+    # pylint:disable=global-variable-undefined
+    global process_sample_collection
+    global process_map_fcn
+    global process_batch_count
+    global process_sample_count
+    global process_queue
+    global process_cancel_event
+    global process_save
+    global process_progress
+    global process_skip_failures
```
swheaton, line 232: yikes what's going on here?

fiftyone/core/map/process.py
```diff
@@ -0,0 +1,302 @@
...
+    i, num_batches, batch = args
+
+    try:
+        sample_collection = batch.create_subset(process_sample_collection)
+
+        sample_iter = sample_collection.iter_samples(autosave=process_save)
+
+        pb = None
+        if process_progress:
+            desc = f"Batch {i:0{len(str(num_batches))}}/{num_batches}"
```
swheaton, line 271: ah interesting. could use a util method or comment, this tripped me up for a second

fiftyone/core/map/threading.py
```diff
@@ -0,0 +1,225 @@
...
+
+T = TypeVar("T")
+R = TypeVar("R")
+
+ResultQueue = queue.Queue[
+    Tuple[bson.ObjectId, Union[Exception, None], Union[R, None]]
+]
+
+
+class ThreadMapper(fomm.LocalMapper):
```
swheaton, line 40: Any way we can DRY these 2 implementations? They've got a lot of similarities.

fiftyone/core/utils.py
```diff
@@ -2064,6 +2129,280 @@ def get_alt_path(self, output_path, alt_dir=None):
...
 
+def __rm_unique_filename_tmpdir():
+    with suppress(Exception):
+        shutil.rmtree(f"/tmp/fo-unq/{os.getpid()}")
+
+
+atexit.register(__rm_unique_filename_tmpdir)
+
+
+class MultiProcessUniqueFilenameMaker(object):
```
swheaton, line 2140: I cannot review this also in this PR ... 😮‍💨

tests/unittests/dataset_tests.py
```diff
@@ -1022,6 +1023,89 @@ def test_iter_samples(self):
             )
         )
 
+    @drop_datasets
+    def test_update_samples(self):
```
swheaton, line 1027: I don't know if we should be using the "process" backend in unit tests? Multiple processes is nonstandard. Threads too but that has already been broken long ago.

fiftyone/core/collections.py
jleven: Sounds like @swheaton is suggesting that there be a maximum allowed batch size as part of the default calculation to ensure we don't hit the 16MB limit. Though, I'm a little surprised that limit applies here... why would we store the list of IDs for a batch in mongodb? Can we not store it in an array / list?

fiftyone/core/config.py
jleven: If it's confusing, add a comment for the next person 😄

fiftyone/core/map/process.py
jleven: There was one case where @Anddraca was unable to use the FO progress bar, so instead of delay this work indefinitely, we decided to use what works and make a ticket to try again later.

jleven: I may be remembering wrong, though. I thought that was in the Thread implementation, not the Process one...

fiftyone/core/utils.py
jleven: https://github.com/voxel51/fiftyone/pull/5715

fiftyone/core/map/process.py
minhtuev: Sure - I thought going back to an initial time out is nice so that we don't have to wait at max time out

fiftyone/core/map/process.py
minhtuev: Yup we use TQDM for worker progress bar in process too

docs/source/user_guide/using_datasets.rst
```diff
@@ -5717,6 +5717,235 @@ The benefit of the above approach versus passing ``autosave=True`` to
...
+    :func:`recommend_process_pool_workers() <fiftyone.core.utils.recommend_process_pool_workers>`
+    to choose a number of worker processes, unless the method is called in a
+    daemon process (subprocess), in which case no workers are used
+
+.. note::
+
+    You can set ``workers<=1`` to disable the
+    use of multithreading and multiprocessing pools in
+    :meth:`update_samples() <fiftyone.core.collections.SampleCollection.update_samples>`.
+
```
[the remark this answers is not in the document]
brimoor: I think its important to be able to force `update_samples()` to fallback to an implementation that uses a single process/thread, in case these core methods are being used in a situation where multithreading/multiprocessing isn't allowed or isn't working

fiftyone/core/map/batcher/id_batch.py
brimoor: Exposing this as `values(..., unordered=True)` felt natural at first glance to me.

However, adding a kwarg to the public interface like that should really promise that it works for as many possible syntaxes of `values()` as possible. Not just one or two specific things.

Unfortunately I don't think the "fast values" approach that uses `find()` will be able to translate cases like `values(expr)`, because mongo aggregations and mongo find operations support slightly different sets of operations 😢

swheaton: Yeah and find() isn't always fast, just in a few particular cases.

fiftyone/core/collections.py
swheaton: No not quite, I'm talking about the input query document. This method gets a big ol' list of IDs and passes them into a view. It will look something like this, except with a million Object IDs ... not only is the query document huge but the match is probably going to be quite slow.
```python
{$match: {_id: {$in: [ObjectId('642f31df73aa78f43a202cbf'), ObjectId('642f31df73aa78f43a202ce4'), ... ]}}}
```

1 million IDs, if we assume it's using bson which needs 12 bytes per, will be 12MB. Once the 16MB limit is passed, the query will be rejected. We've seen it happen with customers who tried to select a ton of object IDs in a view.

swheaton: But actually I found out after reading more code that they do actually limit batch size via this default approach, to 100,000. This seems arbitrary but who knows.
If we use the unordered ID approach then we don't have to pass in a list of IDs at all, only the start and stop IDs.
Making most of my point moot (although I still think it's better to have medium sized batches than small or giant).

fiftyone/core/map/batcher/id_batch.py
swheaton: Input query limit is 16MB I think, just like a normal document. 100MB is something else - working set document size or something

docs/source/user_guide/using_datasets.rst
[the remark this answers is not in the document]
swheaton: Ok so just to be able to call a single method and have it work. Got it.

fiftyone/core/map/batcher/id_batch.py
kaixi-wang: Yeah I wasn't thinking about passing lists of ids when thinking about batch sizes ☠️

fiftyone/core/map/batcher/id_batch.py
brimoor: yep and `values("id")` and `values("filepath")` are by far the most common values operations, so we should definitely use this trick whenever we can tolerate unordered listing.

fiftyone/core/collections.py
brimoor: > But actually I found out after reading more code that they do actually limit batch size via this default approach, to 100,000. This seems arbitrary but who knows.

This logic exists in a variety of places in the codebase:

```py
batch_size = recommend_batch_size_for_value(value, max_size=100000)
```

and it does two things:
1. Ensures that a JSON request involving a list of `batch_size` elements of `type(value)` will not exceed the maximum batch size allowed by `fiftyone.config.batcher_target_size_bytes`. This exists to satisfy any constraints of API gateways that are intercepting our requests to the database. The default value for this is 1MB, for example
2. The `max_size=100000` part is there because, even if somebody set `batcher_target_size_bytes` to infinite, we *still* must respect MongoDB's 16MB size limit of query request documents. For context, 100k `ObjectIDs` is ~4MB, so it's comfortably under the 16MB limit.

Why express `max_size` as a maximum batch size rather than just *also* enforcing a 16MB size limit inside the function? Like @swheaton is saying, there's a performance tradeoff, `$select` from a very long list of IDs starts to drop off in performance.

I honestly can't remember why I settled on 100k, but I may have done some rough benchmarking and found that performance of batched operations starts to drop-off above that level. 

Note that, because the default `batcher_target_size_bytes` is 1MB, the limiting factor is point 1 above and so when a user calls `map_samples()` or `update_samples()` without specifying a batch size, they get roughly `batch_size=25,000` by default.

brimoor: So in reality, when `map_samples()` is called on a collection of 1M samples, the default behavior is that the task is split up into 40 shards, regardless of whether the user's machine has 4, 8, 16, or 32 cores.

Importantly, if the collection is small enough that splitting the IDs equally across all available cores would result in < 25,000 IDs per core, then that should definitely be done by default. The whole point is to always make full use of all cores, while avoiding creating batches that are too big.

I haven't reviewed this PR in detail, but what I described above is how the batching logic works in https://github.com/voxel51/fiftyone/pull/5480, which I think was copied over here. And it should stay that way IMO.

fiftyone/core/map/process.py
brimoor: We should always use eta `ProgressBar` where possible, because it supports some special syntaxes like `progress=callback_fn` which is useful for reporting the progress of a running DO. 

However, eta `ProgressBar` does not support rendering nested progress bars in subprocesses as cleanly as `tqdm` does, so we can fallback to `tqdm` specifically when the user wants to see per-worker progress bars. And this is fine, because if that's how the user wants to see progress bars, then the `progress=callback_fn` syntax doesn't make sense anyway, as the workers would be stepping on each other.

Since the per-worker progress case is the exception, not the rule, it needs to have the "new" syntax. I proposed `progress="workers"` in https://github.com/voxel51/fiftyone/pull/5480.

`progress=True` should be reserved for the typical case, which is a single eta `ProgressBar` that tracks the progress of the entire operation.

fiftyone/core/map/threading.py
j053y: Maybe a little bit. They use different synchronization primitives and the two methods do things slight differently. It might make it more confusing to read and work in the name of being DRY.

fiftyone/core/map/process.py
j053y: Multiprocessing setup

fiftyone/core/collections.py
swheaton: So unless I'm missing something, in `id` mode it would split into 40 "shards" because of the recommend batch size. Which is good, but not what is documented in the high level function.

In `slice` mode, it breaks it into exactly `num_workers` number of shards. Which is fine for the query document size, but I am arguing here it is not a sane default if it would batches larger than, say, 25k to be consistent.

fiftyone/core/map/factory.py
```diff
@@ -0,0 +1,97 @@
...
+        cls,
+        mapper_key: Optional[str],
+        workers: Optional[int],
+        batch_method: Optional[str] = None,
+        batch_size: Optional[int] = None,
+        **mapper_extra_kwargs,
+    ) -> fomm.Mapper:
+        """Create a mapper instance"""
+
+        config = focc.load_config()
```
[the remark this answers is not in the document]
j053y: Circular import causes the unit tests to fail

fiftyone/core/map/process.py
Anddraca: Yep, agree with Brian. tqdm is only used in threading for the per worker use case, and if you see in this processing use case, we're still using eta's ProgressBar further down. And in the threading use case, it's still used for the global (i.e. `progress=True` case) as Brian described

fiftyone/core/map/mapper.py
minhtuev: Addressed here: https://github.com/voxel51/fiftyone/pull/5653/files

docs/source/user_guide/using_datasets.rst
```diff
@@ -5717,6 +5717,235 @@ The benefit of the above approach versus passing ``autosave=True`` to
...
+You can also pass `progress="workers"` to
+:meth:`update_samples() <fiftyone.core.collections.SampleCollection.update_samples>`
+to render progress bar(s) for each worker:
+
+.. code-block:: python
+    :linenos:
+
+    view.update_samples(update_fcn, workers=16, progress="workers")
+
```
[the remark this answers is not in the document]
minhtuev: Thanks

docs/source/user_guide/using_datasets.rst
```diff
@@ -5717,6 +5717,235 @@ The benefit of the above approach versus passing ``autosave=True`` to
...
+    daemon process (subprocess), in which case no workers are used
+
+.. note::
+
+    You can set ``workers<=1`` to disable the
+    use of multithreading and multiprocessing pools in
+    :meth:`update_samples() <fiftyone.core.collections.SampleCollection.update_samples>`.
+
+-   The ``batch_method`` parameter controls how samples are grouped into batches for processing. When set to "slice",
+    samples are grouped sequentially, while "id" groups them by their unique IDs.
```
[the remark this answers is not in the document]
minhtuev: Updated

docs/source/user_guide/using_datasets.rst
```diff
@@ -5717,6 +5717,235 @@ The benefit of the above approach versus passing ``autosave=True`` to
...
+    Batch 11/16: 100%|█████████████████████████████████████████████████| 3125/3125 [896.80it/s]
+    Batch 12/16: 100%|█████████████████████████████████████████████████| 3125/3125 [903.28it/s]
+    Batch 13/16: 100%|█████████████████████████████████████████████████| 3125/3125 [893.63it/s]
+    Batch 14/16: 100%|█████████████████████████████████████████████████| 3125/3125 [891.26it/s]
+    Batch 15/16: 100%|█████████████████████████████████████████████████| 3125/3125 [905.06it/s]
+    Batch 16/16: 100%|█████████████████████████████████████████████████| 3125/3125 [911.72it/s]
+
+.. _map-reduce-operations:
+
+Mapping samples operations
```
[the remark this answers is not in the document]
minhtuev: I think the duplicated instructions are by design @brimoor. We can combine the two sections (map-samples, update-samples into one for both). Let me know if that is what we want to do.

docs/source/user_guide/using_datasets.rst
```diff
@@ -5717,6 +5717,235 @@ The benefit of the above approach versus passing ``autosave=True`` to
...
+-   The samples are not processed in any particular order
+
+-   Your ``map_fcn`` should not modify global state or variables defined
+    outside of the function
+
+-   If your ``map_fcn`` modifies samples in-place, you must pass ``save=True``
+    to save these edits
+
+-   Your ``map_fcn`` should not return any samples as the document objects are
+    not able to serialize and deserialize.
```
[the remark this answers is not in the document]
minhtuev: Updated

docs/source/user_guide/using_datasets.rst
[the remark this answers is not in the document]
minhtuev: Already addressed in a different PR

fiftyone/core/map/batcher/id_batch.py
```diff
@@ -0,0 +1,72 @@
...
+        self.sample_ids = sample_ids
+
+    @property
+    def total(self) -> int:
+        return len(self.sample_ids)
+
+    def create_subset(
+        self, sample_collection: SampleCollection[T]
+    ) -> SampleCollection[T]:
+        return fov.make_optimized_select_view(
```
kaixi-wang, line 70: Have we tested this with complex views instead of just full datasets? Since we've already identified the specific sample_ids, is retaining the view and then selecting the sample_id faster than just directly selecting the sample_id from the collection directly?

swheaton (approved):
removing change request and approving, on 2 conditions:
- Plan to address, log, or close concerns, discussed offline, will be enacted upon
- We are not claiming speed improvement victory and this will not be documented - therefore the test burden is smaller - only correctness and not performance.

fiftyone/core/map/process.py
minhtuev: I have opened a draft PR to utilize backoff library and will do more testing tomorrow; just checking, is this a blocking request?
https://github.com/voxel51/fiftyone/pull/5749

swheaton: no

PR #5643, opened 2025-03-25
Expose map_samples to dataset collection and dataset view

minhtuev:

Exposed map_samples and updated_samples to dataset collection and dataset view

* Tested locally
* Added unit tests

        notes for FiftyOne users.

-   [ ] Other


j053y (changes requested):
Let's also remove the version in `fiftyone.utils` and copy the implementation to these methods. It's unnecessary to real reason to have the middle man.

minhtuev:
Hey @j053y, I checked the code in this file (collections.py), and the pattern here is that each function calls a function in utils and passes the params. I don't have a strong opinion here, do you still think we should go ahead and remove the functions in utils/map?

```
    def list_evaluations(self, type=None, method=None, **kwargs):
        """Returns a list of evaluation keys on this collection.
        Args:
            type (None): a specific evaluation type to match, which can be:
                -   a string
                    :attr:`fiftyone.core.evaluations.EvaluationMethodConfig.type`
                -   a :class:`fiftyone.core.evaluations.EvaluationMethod` class
                    or its fully-qualified class name string
            method (None): a specific
                :attr:`fiftyone.core.evaluations.EvaluationMethodConfig.method`
                string to match
            **kwargs: optional config parameters to match
        Returns:
            a list of evaluation keys
        """
        return foev.EvaluationMethod.list_runs(
            self, type=type, method=method, **kwargs
        )
```

j053y (approved):
:nut_and_bolt:

PR #5646, opened 2025-03-25
decouple execution store logic from mongo

sashankaryal:

We have both `ExecutionStoreRepo` and `MongoExecutionStoreRepo` but `ExecutionStoreRepo` is very tightly coupled to Mongo. This PR makes `ExecutionStoreRepo` an abstract base class that `MongoExecutionStoreRepo` concretizes. Also add an implementation for `InMemoryExecutionStoreRepo` that is a lightweight implementation of execution store repo we can use for unit / integration testing.

Unit tests

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced a refined execution store interface with consistent operations, offering both persistent and in-memory management options.

- **Documentation**
  - Enhanced service documentation to clearly detail constructor parameters, ensuring proper configuration without ambiguity.

- **Tests**
  - Added comprehensive test suites covering core operations such as creation, retrieval, updating, deletion, and cleanup to ensure robust functionality.
  - Updated existing tests to utilize the new MongoDB execution store repository implementation.

### DIFF
--- a/fiftyone/factory/repos/execution_store.py
+++ b/fiftyone/factory/repos/execution_store.py
@@ -1,44 +1,283 @@
 """
-Execution store repository.
+Execution store repository interface and implementations.
 
 | Copyright 2017-2025, Voxel51, Inc.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
 
+from abc import ABC, abstractmethod
 from datetime import datetime
+from typing import Any, Dict, List, Optional
 
 from bson import ObjectId
-from pymongo.collection import Collection
-from typing import Any, Dict
 
-from fiftyone.operators.store.models import StoreDocument, KeyDocument
+from fiftyone.operators.store.models import KeyDocument, StoreDocument
 
 
-class ExecutionStoreRepo(object):
-    """Base class for execution store repositories.
+class ExecutionStoreRepo(ABC):
+    """Abstract base class for execution store repositories.
 
-    Each instance of this repository has a context:
-
-    -   If a ``dataset_id`` is provided, it operates on stores associated with
-        that dataset
-    -   If no ``dataset_id`` is provided, it operates on stores that are not
-        associated with a dataset
+    Each instance operates in a context:
+    - If a `dataset_id` is provided, it operates on stores associated with that dataset.
+    - If no `dataset_id` is provided, it operates on stores not associated with any dataset.
 
     To operate on all stores across all contexts, use the ``XXX_global()``
     methods that this class provides.
     """
 
-    COLLECTION_NAME = "execution_store"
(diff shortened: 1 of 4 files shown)


fiftyone/factory/repos/execution_store.py
```diff
@@ -261,42 +488,149 @@ def count_stores_global(self) -> int:
...
-        """
-        result = self._collection.delete_many(dict(store_name=store_name))
+    def delete_store_global(self, store_name: str) -> int:
+        result = self._collection.delete_many({"store_name": store_name})
         return result.deleted_count
 
 
-class MongoExecutionStoreRepo(ExecutionStoreRepo):
-    """MongoDB implementation of execution store repository."""
+class InMemoryExecutionStoreRepo(ExecutionStoreRepo):
```
tom-vx51, line 499: 😮‍💨

tom-vx51 (approved):

fiftyone/factory/repos/execution_store.py
```diff
@@ -141,92 +370,92 @@ def set_key(self, store_name, key, value, ttl=None) -> KeyDocument:
...
+                "store_name": store_name,
+                "key": key,
+                "dataset_id": self._dataset_id,
+            },
             {"$set": {"expires_at": expiration}},
         )
         return result.modified_count > 0
 
-    def delete_key(self, store_name, key) -> bool:
-        """Deletes the document that matches the store name and key, if one
```
ritch, line 184: What do the docs render like with this change?

fiftyone/factory/repos/execution_store.py
ritch, line 499: 🔥

fiftyone/factory/repos/execution_store.py
ritch: this change = moving to the abstract class

sashankaryal: So detached from that world. Is there something I should be concerned about? Are you worried because only the ABC has docstrings but not Mongo ES repo?

ritch: Ya, as long as the docs linked from `ctx` have the right info we should be alright.

ritch: https://docs.voxel51.com/api/fiftyone.operators.store.html#fiftyone.operators.store.ExecutionStore

ritch: This is where they link to ^

sashankaryal: I see! The repo link should link to ABC and it should have all the docs

Issue #5335, opened 2025-01-02
[BUG] evaluate_detections() is skipping detections


AntoninDvl:
### Describe the problem

Hi ! Thanks for this awesome project !

I've been using 51 for quite a while now. I recently tried the `evaluate_detections() `method  with `fiftyone.core.labels.Detections` and `COCO` evaluation method, which is very powerful.

When I upgraded Fiftyone to 1.1.0 (and also in 1.2.0), I've noticed a strange behavior that I can not explain.

It looks like some predictions are not evaluated. Their `eval_id` remains empty, and the detection is neither evaluated as a `fp` or `tp`. When investigated, I discovered it seems it was all the predictions that are False Positive and where there is also a GT on the same image ( a non-negative image). 

### Code to reproduce issue

This is how I created the evaluation :
dataset = fo.load_dataset('my_dataset')

```
dataset.evaluate_detections(
            pred_field='preds',
            gt_field="ground_truth",
            eval_key=f"eval_test",
            classwise=False,
            compute_mAP=False,
            iou=0.5,
        )
```

### System information

-   **OS Platform and Distribution** : Linux Ubuntu 22.04
-   **Python version** (`python --version`): Python 3.10.13
-   **FiftyOne version** (`fiftyone --version`): 1.2.0
-   **FiftyOne installed from** (pip or source): poetry

### Other info/logs

Include any logs or source code that would be helpful to diagnose the problem.
If including tracebacks, please include the full traceback. Large logs and
files should be attached. Please do not use screenshots for sharing text. Code
snippets should be used instead when providing tracebacks, logs, etc.

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another
member of your organization be willing to contribute a fix for this bug to the
FiftyOne codebase?

-   [X] Yes. I would be willing to contribute a fix for this bug with guidance
        from the FiftyOne community
-   [ ] No. I cannot contribute a bug fix at this time

mwoodson1:
Can you provide any more details on how to reproduce the issue, what kind of data, labels, etc? I tried reproducing this using `coco-2017` validation set and `faster-rcnn` predictions. With that data and predictions I can see many False Positives but no samples ended up with empty `eval_id`s.

AntoninDuvalPreligens:
Hi @mwoodson1. Unfortunatly, I can't share directly the data. I can share however a sample which had this issue.

[...]

PR #5649, opened 2025-03-26
Fix evaluate_detections for non-overlapping bboxes

kaixi-wang:

- https://github.com/voxel51/fiftyone/issues/5335
- ensures that predictions that don't overlap with any ground truth are assigned a fp for the closest ground truth

- add unit test
- test running compute_evaluations and verify results:
```
import fiftyone as fo

dataset = fo.Dataset()
gt = fo.Detections(detections=[fo.Detection(label="test", bounding_box=[0.1,0.1,0.1,0.1])])
preds = fo.Detections(detections=[
    fo.Detection(label="test", bounding_box=[0.1,0.1,0.5,0.5]),
    fo.Detection(label="test", bounding_box=[0.0,0.0,0.1,0.1]),
    fo.Detection(label="test", bounding_box=[0.9,0.9,0.1,0.1]),
])
sample = fo.Sample(filepath="fake.png", ground_truth=gt, predictions=preds)
dataset.add_sample(sample)
eval_key = "eval"
pred_field="predictions"
gt_field = "ground_truth"
    pred_field=pred_field,
    gt_field=gt_field,
    eval_key=eval_key,
    classwise=False,
    iou=0.5,
)
print(dataset.values(eval_key+"_fp"))
# v1.0.2: [3]
# v1.1: [2]
# now: [3]
print(dataset.values(pred_field+".detections."+eval_key))
# v1.0.2: [['fp', 'fp', 'fp']]
# v1.1: [['fp', 'fp', None]]
# now: [['fp', 'fp', 'fp']]

## Release Notes

<!--
-   [x] Selected option
-->

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/iou.py
+++ b/fiftyone/utils/iou.py
@@ -681,7 +681,16 @@ def _compute_bbox_ious(
                 ious[i, j] = iou
                 if is_symmetric:
                     ious[j, i] = iou
-
+        # If a prediction has no intersection with a ground_truth,
+        # find the nearest and assign iou=0
+        if sparse and not ious.get(pred.id):
+            # pylint: disable=not-an-iterable
+            for nearest in rtree_index.nearest(box, 1):
+                gt = gts[nearest]
+                iou = 0
+                ious[pred.id].append((gt.id, iou))
+                if is_symmetric:
+                    ious[gt.id].append((pred.id, iou))
     return ious
 
 
--- a/tests/unittests/utils/iou_tests.py
+++ b/tests/unittests/utils/iou_tests.py
@@ -1,0 +1,63 @@
+import unittest
+
+from bson import ObjectId
+
+import fiftyone as fo
+from fiftyone.utils.iou import compute_ious
+
+
+class ComputeIouTests(unittest.TestCase):
+    def test_compute_iou_bbox_sparse_nonoverlapping(self):
+        iscrowd = None
+        classwise = False
+        sparse = True
+        eval_results = {
+            "mask": None,
+            "mask_path": None,
+            "confidence": None,
+            "index": None,
+            "eval_iou": None,
+            "eval_id": "",
+        }
+        ids = [ObjectId() for _ in range(4)]
+        gt = [
+            fo.Detection(
+                id=ids[0],
+                label="test",
+                bounding_box=[0.1, 0.1, 0.1, 0.1],
+                **eval_results
+            )
+        ]
+        preds = [
+            fo.Detection(
+                id=ids[1],
+                label="test",
+                bounding_box=[0.1, 0.1, 0.5, 0.5],
+                **eval_results
+            ),
+            fo.Detection(
+                id=ids[2],
+                label="test",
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: fiftyone/utils/iou.py [hunk 1]
```diff
@@ -681,7 +681,16 @@ def _compute_bbox_ious(
                 ious[i, j] = iou
                 if is_symmetric:
                     ious[j, i] = iou
-
+        # If a prediction has no intersection with a ground_truth,
+        # find the nearest and assign iou=0
+        if sparse and not ious.get(pred.id):
```
brimoor, line 686: When `sparse=True`, this method specifically promises to only store matches with IoU>0. If we included non-overlapping objects then it wouldn't be "sparse" anymore.

So, I do think you're on the right track as far as what the root cause of https://github.com/voxel51/fiftyone/issues/5335 is. But, I think the fix needs to be in the code where `compute_ious(sparse=True)` is **called**: which is in `fiftyone.uils.eval.coco` and `fiftyone.utils.eval.openimages`.

brimoor (changes requested):

fiftyone/utils/iou.py
kaixi-wang: interesting point... so eval.coco calls compute_ious with sparse=True, yet the issue claims that non-overlapping predictions should be counted as false positives?

kaixi-wang: updated so that no matches are stored, but the id is preserved...
If it needs to be where it's called ( `_coco_evaluation_setup` and the corresponding openimages methods), I should probably pass this off to ML

brimoor: Non-overlapping ground truth objects:

```
eval="fn"
eval_id=""
eval_iou=None
```

Non-overlapping predictions:

```
eval="fp"
eval_id=""
eval_iou=None
```

swheaton: I agree, seems like more of ML territory

kaixi-wang: Can you give me a little more context? I'm not sure what you're pointing out

ehofesmann: I like Kacey's current proposal of having it return a dict with a key for every `pred.id` but it is just an empty list if there are no matches. I still need to test it, but that should resolve #5335. 

This would comply with the promise to only store matches with IoU>0 since there is no corresponding `(gt.id, iou)` that gets returned, and having an empty list for every `pred.id` is already a pattern for this method when `sparse=True` here: https://github.com/voxel51/fiftyone/blob/4e1fc87a2c7b6b823cfc76988503ae23a3d6c984/fiftyone/utils/iou.py#L98

brimoor:
Resolved by https://github.com/voxel51/fiftyone/pull/5697

fiftyone/utils/iou.py
brimoor: @ehofesmann I like the suggestion; implemented in https://github.com/voxel51/fiftyone/pull/5697 ✅

PR #5658, opened 2025-03-28
Guard against map function returning a sample

Anddraca:

With the implementation of `map_samples`, it is known that returning a `fo.Sample` object from the provided map function will result in errors that can be difficult to parse or understand the root cause of.

The Sample object cannot always be pickled, so it cannot be passed between processes. This results in an unclear error and a failure. We can instead block that possibility by not allowing users to return Sample objects directly from the `map_fcn` and avoid the unexpected behavior.

This PR introduces a check for this to make sure that it will not attempt to run and fail.

Updated unit tests, tested locally. 

To test:
use any dataset, i.e. quickstart, and supply a map function that returns a Sample. 

Alternatively, the logic can be tested outside of fiftyone with something like this script:

```import fiftyone as fo

dataset = fo.zoo.load_zoo_dataset("quickstart")

sample_col = dataset._get_sample_collection()

first_sample = dataset.first()


def map_fcn(sample):
    # make a copy of the sample
    sample_copy = sample.copy()
    
    # testing that we only update the copy and 
    # this test does not modify anything in the database
    sample_copy.tags.append("map_tag")


    return sample_copy


print("IS IT A SAMPLE? ", isinstance(map_fcn(first_sample), fo.Sample))```

### CODE REVIEW

**Comment by kaixi-wang** [comment]
> it is known that returning a fo.Sample object from the provided map function will result in errors that can be difficult to parse or understand the root cause of.

Out of curiosity, is this described or discussed more in detail somewhere? Understanding this could be important to understanding the limitations of map_samples better

**Comment by minhtuev** [comment]
Verified that it works

```
In [11]: for _, value in dataset.map_samples(map_fcn, workers=4):
    ...:     print(value)
    ...:
    
File ~/workspace/fiftyone/fiftyone/core/map/factory.py:92, in MapperFactory.create(cls, key, sample_collection, workers, map_fcn, batch_method, **mapper_extra_kwargs)
     90 # check here if the map fcn returns a sample and raise if it do
     91 if cls.__check_if_return_is_sample(sample_collection, map_fcn):
---> 92     raise ValueError(
     93         "The map function must not return a Sample object"
     94     )
     96 return cls.__MAPPER_CLASSES[key](
     97     sample_collection, workers, batch_method, **mapper_extra_kwargs
     98 )

ValueError: The map function must not return a Sample object
```

**Review by minhtuev — APPROVED** [review]
LGTM, thanks Tom!

Btw, is it really impossible to return a Sample object?

File: fiftyone/core/map/factory.py [hunk 1]
```diff
@@ -82,9 +81,6 @@ def create(
         if key is None:
             key = cls.default()
 
-        if key not in cls.__MAPPER_CLASSES:
-            raise ValueError(f"Could not create mapper for: '{key}'")
-
```
j053y, line 87: Why did this get removed?

j053y (changes requested):
Don't know why unrelated chunk of code got removed other than that looks good!

j053y (approved):

PR #5666, opened 2025-03-31
Add support for adding samples as a generator

brimoor:
Adds an optional `generator=True` parameter to all `Dataset` methods that add samples to datasets that causes the methods to yield control to the caller after each batch of samples is added. 

This is useful in situations where the caller would like to perform some custom logic *during* the operation, such as report progress to an external entity as is done in https://github.com/voxel51/fiftyone-plugins/pull/227.

As a toy example, we can implement our own progress bar outside of `add_samples()` as follows:

```py
import fiftyone as fo

n = 5000
make_sample = lambda i: fo.Sample(filepath=f"image{i}.jpg")
samples = map(make_sample, range(n))

dataset = fo.Dataset()
with fo.ProgressBar(total=n) as pb:
    for ids in dataset.add_samples(samples, generator=True, progress=False):
        pb.update(len(ids))
```

## Summary by CodeRabbit

- **New Features**
  - Added a generator option to dataset sample addition and import methods, allowing users to yield sample ID batches incrementally during data ingestion.
- **Documentation**
  - Updated docstrings to describe the new generator parameter and its behavior.
- **Tests**
  - Added tests to verify generator-based sample addition functionality.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -3251,6 +3251,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         expand_schema=True,
         dynamic=False,
         validate=True,
+        generator=False,
         progress=None,
         num_samples=None,
     ):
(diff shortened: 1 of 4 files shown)


swheaton:
@coderabbitai review

swheaton (approved):
Sure. Briefly tested and lgtm
🚢 🇮🇹

benjaminpkane (approved):

Issue #566, opened 2020-09-29
[FR] Add support for optionally loading segmentations in COCODetectionDataset format


brimoor:
Some resources on the topic:

- https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocotools/mask.py
- https://stackoverflow.com/q/49494337
- https://www.immersivelimit.com/tutorials/create-coco-annotations-from-scratch

There are some questions here, like should the output be stored as `Detections` with instance `mask`s, or should the output be stored as `Polylines`? Or both (leveraging a `labels_dict` return type)? Need to think about the label class contract of the importer/exporters...

One wrinkle is that single objects are stored as vertices, while crowds are stored as RLE'd dense masks. One would probably prefer all masks to be stored in one format or the other in FiftyOne, not mixed.

brimoor:
As of https://github.com/voxel51/fiftyone/pull/612, `COCODetectionDataset` format now supports loading instance segmentations.

In fact, it [provides parameters](https://voxel51.com/docs/fiftyone/api/fiftyone.utils.coco.html?highlight=cocodetection#fiftyone.utils.coco.COCODetectionDatasetImporter) for controlling whether segmentations are loaded, omitted, or loaded as polylines rather than detections with dense segmentation masks.

PR #5667, opened 2025-03-31
Add support for passing custom batchers to `add_samples()`

brimoor:
## Change log

- Adds support for passing your own batcher to `add_samples(..., batcher=)`.
- Adds a shorthand `add_samples(..., batcher=False)` syntax to add all samples in a single batch (ie disable batching for that particular method invocation)

## Commentary

I originally implemented this as a means to achieve the same outcome of https://github.com/voxel51/fiftyone-plugins/pull/227 of reintroducing the progress callbacks; specifically, I was using `add_samples(..., batcher=False)` to add samples in batches that we manually constructed by `@voxel51/io/import_samples`. However, I then came up with https://github.com/voxel51/fiftyone/pull/5666, which seemed a bit cleaner.

Nonetheless, it is still potentially interesting and useful to let users build + pass their own batchers, so I thought I'd throw up a PR for discussion anyway, even though this is new capability isn't in use anywhere just yet.

Some of my thoughts while implementing this:
- Allowing users to pass `batcher=False` is a bit scary as you can't send arbitrarily large batches over API in general
- The `partial()` syntax to configure some parameters of a custom `Batcher` class is cute, but perhaps too cute?
- Ideally, my OCD would want us to expose `batcher` in all the relevant places throughout the codebase, not *just* `add_samples()`. EG like I did in https://github.com/voxel51/fiftyone/pull/566

## Example ussage

```py
from functools import partial

import fiftyone as fo

samples = [fo.Sample(filepath=f"image{i}.jpg") for i in range(1000)]

#
# No batching
#

dataset = fo.Dataset()
dataset.add_samples(samples, batcher=False)

assert len(dataset) == 1000

#
# Custom batcher
#

batcher = partial(fou.StaticBatcher, batch_size=10)

dataset = fo.Dataset()
dataset.add_samples(samples, batcher=batcher)

assert len(dataset) == 1000
```

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -13,6 +13,7 @@ from collections import defaultdict
 from contextlib import contextmanager
 from copy import deepcopy
 from datetime import date, datetime
+from functools import partial
 import glob
 import hashlib
 import importlib
@@ -1068,14 +1069,31 @@ def _report_progress_dt(progress, dt):
     return progress_dt
 
 
-class BaseBatcher(abc.ABC):
+class Batcher(abc.ABC):
+    """Base class for iterating over the elements of an iterable in batches.
+
+    Args:
+        iterable: an iterable to batch over
+        transform_fn (None): a transform function to apply to each item
+        return_views (False): whether to return each batch as a
+            :class:`fiftyone.core.view.DatasetView`. Only applicable when the
+            iterable is a :class:`fiftyone.core.collections.SampleCollection`
+        progress (False): whether to render a progress bar tracking the
+            consumption of the batches (True/False), use the default value
+            ``fiftyone.config.show_progress_bars`` (None), or a progress
+            callback function to invoke instead
+        total (None): the length of ``iterable``. Only applicable when
+            ``progress=True``. If not provided, it is computed via
+            ``len(iterable)``, if possible
+    """
+
     def __init__(
         self,
         iterable,
+        transform_fn=None,
         return_views=False,
         progress=False,
         total=None,
-        transform_fn=None,
     ):
         import fiftyone.core.collections as foc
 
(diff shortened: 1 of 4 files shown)


fiftyone/core/utils.py
```diff
@@ -1068,14 +1069,31 @@ def progress_dt(pb):
     return progress_dt
 
 
-class BaseBatcher(abc.ABC):
+class Batcher(abc.ABC):
+    """Base class for iterating over the elements of an iterable in batches.
+
+    Args:
```
brimoor, line 1075: Always document kwargs of important classes, especially ones linked to from important public methods

fiftyone/core/utils.py
```diff
@@ -1697,36 +1726,51 @@ def default_calculate_size(obj):
 
 def get_default_batcher(
     iterable,
+    batcher=None,
```
brimoor, line 1729: This doesn't really return a "default" batcher if the user provides a `batcher`, but couldn't really think of an alternative that didn't require renaming this method

fiftyone/core/utils.py
```diff
@@ -1697,36 +1726,51 @@ def default_calculate_size(obj):
...
-            before adding them to the batch to ensure we are under max batch size.
-            This is applied after transform_fn if both are provided.
-        transform_fn (None): a transform function to apply to each item of the batch
 
     Returns:
-        a :class:`Batcher`
+        a :class:`Batcher` instance
     """
+    if batcher is False:
+        batcher = partial(StaticBatcher, batch_size=-1)
```
brimoor, line 1764: Supporting this is basically the whole point of this PR

CamronStaley (commented):
Changes seem solid just needs conflicts resolved lmk if you have issues with that since I think they're my changes that are conflicting

swheaton (approved):
🚢 🇮🇹

PR #5675, opened 2025-04-01
Fix label tags sidebar filter for Query Performance

benjaminpkane:

Fixes label tag filtering the QP sidebar

https://github.com/user-attachments/assets/17b84285-4175-4650-928f-bba8a503f13a

Selector tests

## Release Notes

* Fixed label tags filtering in the
  :ref:`Query Performance <app-optimizing-query-performance>` sidebar

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
	- Fixed label tags filtering in the Query Performance sidebar to ensure counts are displayed correctly in grid and modal views.
	- Enhanced count reliability by defaulting to zero when values are missing.
- **Documentation**
	- Updated release notes to include details for the latest releases (FiftyOne Enterprise 2.7.2 and FiftyOne 1.4.1) with pertinent bug fixes.

### DIFF
--- a/app/packages/state/src/recoil/pathData/counts.ts
+++ b/app/packages/state/src/recoil/pathData/counts.ts
@@ -67,9 +67,10 @@ export const count = selectorFamily({
         }
 
         if (first === "_label_tags" && value) {
-          return get(cumulativeCounts({ ...params, ...MATCH_LABEL_TAGS }))[
-            value
-          ];
+          return (
+            get(cumulativeCounts({ ...params, ...MATCH_LABEL_TAGS }))[value] ??
+            0
+          );
         }
 
         if (split.length < 2) {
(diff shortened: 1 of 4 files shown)


app/packages/state/src/recoil/pathData/counts.ts
```diff
@@ -67,9 +67,10 @@ export const count = selectorFamily({
         }
 
         if (first === "_label_tags" && value) {
-          return get(cumulativeCounts({ ...params, ...MATCH_LABEL_TAGS }))[
-            value
-          ];
+          return (
+            get(cumulativeCounts({ ...params, ...MATCH_LABEL_TAGS }))[value] ??
```
kaixi-wang, line 71: Is it possible for the get to return none and then throw an error when trying to access value from none?

findtopher:
could you bump the `VERSION` in `setup.py` while you're in here?

findtopher (approved):
Looks good from Aloha!

app/packages/state/src/recoil/pathData/counts.ts
benjaminpkane: This is a small detail I fixed when checking the legacy sidebar mode. We want zero to return here as it does on line 95. Currently "of X" is shown instead of "0 of X". See screenshot.
Screenshot 2025-04-01 at 10 40 17 AM

PR #5680, opened 2025-04-02
`execution_cache` module & decorator

ritch:
## Change log

Adds a new `fiftyone.operators.execution_cache` decorator for function caching

## Motivation

Enables operators and panels to more easily use the `ExecutionStore` as a caching mechanism by consolidating the implementation into a one-liner.

```python

@foo.execution_cache(ttl=90)
def get_values(ctx, field):
    return ctx.dataset.values(field)

class ValuesOperator(foo.Operator):
    @property
    def config(self):
        return foo.OperatorConfig(
            name="values_example",
            label="Get Values",
        )

    def execute(self, ctx):
        return {"values": get_values(ctx, "filepath")}
```

image
image

## Summary by CodeRabbit

- **Documentation**
  - Added a new section in the plugin development docs that explains how to cache function results, with detailed examples and clear instructions for managing cache entries.

- **New Features**
  - Introduced an execution caching mechanism to optimize performance by storing and reusing function outputs.
  - Enabled configurable options (e.g., time-to-live, scoping, custom key generation) along with controls for cache clearing and policy updates.
  - Added a configuration toggle for enabling/disabling caching via environment settings.
  - Enhanced the `ExecutionStore` with methods for cache management, including setting and clearing cache entries.
  - Added a new decorator for caching function results in the execution store, supporting various scoping options.
  - Implemented serialization and deserialization functions for cache data management.
  - Introduced utility functions for cache key management and context handling.

- **Tests**
  - Added comprehensive unit tests for the `execution_cache` decorator and cache utility functions, ensuring correct functionality and error handling.
  - Added a test to verify behavior of the `clear_cache` method on a non-existent store.


fiftyone/operators/cache/decorator.py
```diff
@@ -0,0 +1,179 @@
...
+        result = expensive_query.uncached(ctx, path)
+
+        # Clear the cache for the given arguments
+        expensive_query.clear_cache(ctx, path)
+    """
+
+    def decorator(func):
+        func.store_name = store_name
+        func.link_to_dataset = link_to_dataset
+        func.exec_cache_version = version
```
ritch, line 112: We should add a cache setter as well. Something like:

```python
@execution_cache(ttl=60)
def my_func(ctx, field):
    return ctx.dataset.values(field)

# in an operator
def execute(self, ctx):    
    my_func.set_cache(ctx, [1,2,3])
```

fiftyone/operators/cache/manager.py
```diff
@@ -0,0 +1,36 @@
...
+
+| Copyright 2017-2025, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+from fiftyone.operators.store import ExecutionStoreService
+from .utils import resolve_store_name, get_store_for_func
+
+
+def clear_all_caches():
```
ritch, line 12: This is missing the `collection_name` arg for consistency elsewhere.

fiftyone/operators/cache/manager.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Clears all named execution stores and their cache entries.
+    """
+    svc = ExecutionStoreService()
+    svc.clear_cache()
+
+
+def clear_function_cache(ctx, func, store_name=None, link_to_dataset=True):
```
ritch, line 20: This might not make sense as a separate utility. We could keep all utils on the func itself, since a reference is required one way or another.

fiftyone/operators/cache/utils.py
```diff
@@ -0,0 +1,204 @@
...
+import json
+import hashlib
+
+import eta.core.serial as etas
+
+import fiftyone as fo
+from fiftyone.operators.store import ExecutionStore
+
+
+def get_function_id(func):
```
ritch, line 19: Need to clean this up and:

 - decide which are public vs private
 - add doc strings to public methods

fiftyone/operators/cache/utils.py
```diff
@@ -0,0 +1,239 @@
...
+    key_string = json.dumps(structured_key, sort_keys=True)
+    return hashlib.sha256(key_string.encode("utf-8")).hexdigest()
+
+
+def _get_cache_key_list(ctx_index, args, kwargs, key_fn):
+    if key_fn:
+        try:
+            cache_key_list = key_fn(*args, **kwargs)
+            if not isinstance(cache_key_list, list):
+                raise ValueError("Custom key function must return a list.")
```
tom-vx51, line 189: we should probably document this requirement. Might also be a good idea to use tuples instead since they are immutable

fiftyone/operators/cache/utils.py
```diff
@@ -0,0 +1,239 @@
...
+            value = get_value()
+            if value:
+                scoped_keys.append(value)
+            elif allow_missing is False:
+                skip_cache = True
+
+    return scoped_keys, skip_cache
+
+
+def _convert_args_to_dict(args):
```
tom-vx51, line 226: `_convert_args_to_list`?

fiftyone/operators/store/store.py
```diff
@@ -21,18 +21,32 @@ class ExecutionStore(object):
         store_name: the name of the store
         store_service: an
             :class:`fiftyone.operators.store.service.ExecutionStoreService`
+        default_policy ("persist"): the default eviction policy for the store.
```
tom-vx51, line 24: [nit] we should probably include the information from the `set` doc string, or preferably make this an enum

fiftyone/operators/cache/manager.py
```diff
@@ -0,0 +1,36 @@
...
+from fiftyone.operators.store import ExecutionStoreService
+from fiftyone.operators.cache.utils import _get_store_for_func
+
+
+def clear_all_caches():
+    """
+    Clears all named execution stores and their cache entries.
+    """
+    svc = ExecutionStoreService()
+    svc.clear_cache()
```
tom-vx51, line 17: this feels dangerous; could we scope to an operator?

tom-vx51 (changes requested):

fiftyone/operators/cache/manager.py
ritch: Need to update the doc string. This function will only delete documents where the policy is `"evict"`. This is the explicit behavior and the reason for the key policy. It won't delete `policy="persisted"` keys

fiftyone/operators/store/store.py
ritch: there is an enum that backs this internally, I thought about this for a while and landed on sticking with the libraries patterns...such as [this](https://docs.voxel51.com/api/fiftyone.core.dataset.html?highlight=media_type#fiftyone.core.dataset.Dataset.media_type) or [this](https://docs.voxel51.com/api/fiftyone.utils.annotations.html?highlight=label_type#fiftyone.utils.annotations.AnnotationBackend.supported_label_types) or [this (method)](https://docs.voxel51.com/api/fiftyone.brain.html#fiftyone.brain.compute_visualization).

fiftyone/operators/cache/manager.py
ritch: Actually going to remove this module for now and only expose the `func.clear_cache()` utility. This will preserve scope settings like `operator_scoped=True`.

docs/source/plugins/developing_plugins.rst
```diff
@@ -2469,6 +2469,113 @@ that are available to all datasets via the
...
+
+.. code-block:: python
+    :linenos:
+
+    import fiftyone as fo
+
+    def serialize_sample(sample):
+        return sample.to_dict()
+
+    def deserialize_sample(data):
```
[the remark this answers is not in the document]
ritch: it is a bulleted list silly rabit

tom-vx51 (approved):

PR #5686, opened 2025-04-03
Handle missing label tags in looker

benjaminpkane:

Fixe rendering a lookers when `tags` is `None` on a label

```python
import fiftyone.zoo as foz

first = dataset.first()
first.ground_truth.detections[0].tags = None
first.save()
``` 
Screenshot 2025-04-03 at 9 49 12 AM

Unit test

## Release Notes

* Fixed rendering a samples in the App that are missing a label tags list

-   [ ] Other

### DIFF
--- a/app/packages/looker/src/overlays/utils.test.ts
+++ b/app/packages/looker/src/overlays/utils.test.ts
@@ -1,0 +1,12 @@
+import { describe, expect, it } from "vitest";
+
+import { shouldShowLabelTag } from "./util";
+
+describe("shouldShowLabelTag", () => {
+  it("handles missing tags", () => {
+    expect(shouldShowLabelTag(undefined, undefined)).toBe(false);
+    expect(shouldShowLabelTag(undefined, ["one"])).toBe(true);
+    expect(shouldShowLabelTag(["one"], ["one"])).toBe(true);
+    expect(shouldShowLabelTag(["one"], ["two"])).toBe(false);
+  });
+});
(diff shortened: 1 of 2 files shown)


app/packages/looker/src/overlays/utils.test.ts
```diff
@@ -0,0 +1,13 @@
...
+import { shouldShowLabelTag } from "./util";
+
+describe("shouldShowLabelTag", () => {
+  it("handles missing tags", () => {
+    expect(shouldShowLabelTag(undefined, undefined)).toBe(false);
+    expect(shouldShowLabelTag(undefined, ["one"])).toBe(true);
+    expect(shouldShowLabelTag(["one"], ["one"])).toBe(true);
+    expect(shouldShowLabelTag(["one"], ["two"])).toBe(false);
+    expect(shouldShowLabelTag(["one"], undefined)).toBe(false);
+  });
```
lanzhenw, line 12:
```Suggestion  
   it("handles missing tags and overlapping tags when filtering labels", () => {
    // when no label tag filter is applied
    expect(shouldShowLabelTag(null, null)).toBe(false);
    expect(shouldShowLabelTag(null, ["one"])).toBe(true);
    expect(shouldShowLabelTag(undefined, undefined)).toBe(false);
    expect(shouldShowLabelTag(undefined, ["one"])).toBe(true);

    // when filter tag is applied and overlaps
    expect(shouldShowLabelTag(["one"], ["one", "two", "three"])).toBe(true);

    // when filter tag is applied and does not overlap
    expect(shouldShowLabelTag(["one"], ["two"])).toBe(false);
    expect(shouldShowLabelTag(["one"], undefined)).toBe(false);
  });
```

lanzhenw (approved):
🚀  tested.

PR #5695, opened 2025-04-05
do not require class labels to match when `classwise=False`

kevinconka:

When passing `classwise=False` to `evaluate_detections`, the results are not what the user would expect.

As confirmed in this [discord message](https://discord.com/channels/1266527359511564372/1351668013912428668/1352438986886483990), when setting `classwise=False` in COCO evaluation, the code has an inconsistency:

- It allows predictions to match ground truths across different classes (based on IoU)
- BUT it still requires class labels to match exactly to be considered a true positive

This PR adds minimal changes to ignore class label matching when `classwise=False`.

Minimal setup to test class-agnostic results:

```python
import fiftyone as fo

n = 6  # number of samples
samples = [
    fo.Sample(
        filepath=f"img_{i}.jpg",
        ground_truth=fo.Detections(
            detections=[
                fo.Detection(label=f"class_{i}", bounding_box=[0.2, 0.2, 0.5, 0.5])
            ]
        ),
        predictions=fo.Detections(
            detections=[
                fo.Detection(
                    label=f"class_{n - i}",  # 1 match if classwise=True and n is even
                    bounding_box=[0.2, 0.2, 0.4, 0.4],  # 0.64 IoU
                    confidence=0.8,
                )
            ]
        ),
    )
    for i in range(n)
]

dataset = fo.Dataset("my_dataset", persistent=False, overwrite=True)
dataset.add_samples(samples)

# Create a list of thresholds from 0.5 to 0.95 with step 0.05
    pred_field="predictions",
    ground_truth="ground_truth",
    iou=0.5,
    classwise=False,
)
results.print_metrics(average="weighted")
```

### Before patch
```
100% |█████████████████████| 6/6 [381.7ms elapsed, 0s remaining, 15.7 samples/s]  
Evaluating detections...
 100% |█████████████████████| 6/6 [37.1ms elapsed, 0s remaining, 161.7 samples/s] 
accuracy   0.17
precision  0.17
recall     0.17
fscore     0.17
support    6
```

### After patch
```
100% |█████████████████████| 6/6 [403.3ms elapsed, 0s remaining, 14.9 samples/s]  
Evaluating detections...
 100% |█████████████████████| 6/6 [38.5ms elapsed, 0s remaining, 155.9 samples/s] 
accuracy   1
precision  1
recall     1
fscore     1
support    6
```

        notes for FiftyOne users.

Fix inconsistency when using `classwise=False` in COCO evaluation. Do not require class labels to match exactly for a prediction to be considered a true positive.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced a new option for classwise evaluation, allowing predictions to be matched to ground truth based on class labels. This update refines performance metrics, particularly in scenarios with crowd objects, by ensuring evaluation accuracy through proper label matching.

### DIFF
--- a/fiftyone/utils/eval/coco.py
+++ b/fiftyone/utils/eval/coco.py
@@ -500,6 +500,7 @@ def _coco_evaluation_single_iou(gts, preds, eval_key, config):
         eval_key=eval_key,
         id_key=id_key,
         iou_key=iou_key,
+        classwise=config.classwise,
     )
 
     # omit iscrowd
@@ -524,6 +525,7 @@ def _coco_evaluation_iou_sweep(gts, preds, config):
             eval_key="_eval",
             id_key=id_key,
             iou_key=iou_key,
+            classwise=config.classwise,
         )
         for iou_thresh, id_key in zip(iou_threshs, id_keys)
     ]
@@ -595,7 +597,7 @@ def _coco_evaluation_setup(
 
 
 def _compute_matches(
-    cats, pred_ious, iou_thresh, iscrowd, eval_key, id_key, iou_key
+    cats, pred_ious, iou_thresh, iscrowd, eval_key, id_key, iou_key, classwise
 ):
     matches = []
 
@@ -623,9 +625,7 @@ def _compute_matches(
                     if gt[id_key] != _NO_MATCH_ID and not gt_iscrowd:
                         continue
 
-                    # If matching classwise=False
-                    # Only objects with the same class can match a crowd
-                    if gt_iscrowd and gt.label != pred.label:
+                    if gt_iscrowd and (classwise and gt.label != pred.label):
                         continue
 
                     # Crowds are last in order of GTs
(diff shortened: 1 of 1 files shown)


brimoor (changes requested):
This would be a breaking change in existing behavior, which I don't think we should ship for backwards compatibility reasons. Users are likely relying on the existing semantics in practice today.

We could theoretically add a third option, say `classwise="any"`, with the proposed semantics. But if you want objects with `label="cat"` and `label="dog"` that are matched to be considered a true positive, I would argue that you should just set the label of every object to the same value, say `label="object"`, because you really aren't evaluating the classes at all.

A question that could be asked is whether `classwise=False` has useful semantics as currently implemented. It says:
- always always match objects with the largest IoU, irrespective of their class
- if this causes two objects with different labels to be matched, then mark them as FP/FN because, although the localization was good, the classification was not correct

This is slightly different than `classwise=True`, which says:
- never allow objects with different `label` to be matched
- within objects of the same `label`, match the pair with largest IoU

The distinction of whether objects with different labels can be "matched" is meaningful in the confusion matrix:
- When `classwise=True`, the confusion matrix is always strictly diagonal, except for the last row/column which denote unmatched objects
- When `classwise=False`, the confusion matrix can have an off diagonal entries in any location, eg when a `"cat"` and `"dog"` are matched. BUT, these off diagonal entries are still counted as FP/FN for evaluation metric purposes

```py
import fiftyone as fo

sample = fo.Sample(
    filepath="image.jpg",
    ground_truth=fo.Detections(
        detections=[
            fo.Detection(
                label="cat",
                bounding_box=[0, 0, 0.2, 0.2],
            ),
        ]
    ),
    predictions=fo.Detections(
        detections=[
            fo.Detection(
                label="dog",
                bounding_box=[0, 0, 0.2, 0.2],
                confidence=0.9,
            ),
        ]
    )
)

dataset = fo.Dataset()
dataset.add_sample(sample)

# The default: don't match "cat" and "dog"
results1 = dataset.evaluate_detections(
    "predictions",
    gt_field="ground_truth",
    classwise=True,
    eval_key="eval1",
)

# This encodes an unmatched GT and an unmatched prediction separately
print(results1.confusion_matrix(include_missing=True))
"""
[[0 0 1]
 [0 0 0]
 [0 1 0]]
"""

# Allow cat and dog to be "matched" for confusion matrix purposes, but still
# count them as TP/FN for metric purposes
results2 = dataset.evaluate_detections(
    "predictions",
    gt_field="ground_truth",
    classwise=False,
    eval_key="eval2",
)

# Off-diagonal entry encodes that a "cat" and "dog" were matched but incorrectly classified
print(results2.confusion_matrix(include_missing=True))
"""
[[0 1 0]
 [0 0 0]
 [0 0 0]]
"""
```

FYI- the default is `classwise=True` because that is how the official COCO evaluation protocol's semantics were defined. But we offer `classwise=False` for users that want "interesting" confusion matrices while enforcing that classification labels must match in order for something to be considered a true positive.

PR #5697, opened 2025-04-07
Fix #5335: missing false positives in detection evaluations when classwise=False 

### LINKED ISSUE #5335: [BUG] evaluate_detections() is skipping detections

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/5335

Unit tests are updated to protect against this regression reappearing.

As suspected by @ehofesmann in https://github.com/voxel51/fiftyone/issues/5335#issuecomment-2632025882, this bug was introduced  in `fiftyone==1.1` when object detection evaluation was optimized with r-trees.

FYI- after further investigation, the bug **only occurs** when `classwise=False` is passed to `evaluate_detections()`, which is not the default behavior.

## Summary by CodeRabbit

- **Refactor**
  - Simplified the internal logic for handling unmatched predictions to ensure a consistent evaluation process.
  - Improved the data initialization approach within evaluation routines for enhanced clarity.
  
- **Tests**
  - Updated test cases to include additional detection scenarios.
  - Adjusted expected evaluation outcomes to align with the refined metrics.

### DIFF
--- a/tests/unittests/evaluation_tests.py
+++ b/tests/unittests/evaluation_tests.py
@@ -1084,10 +1084,15 @@ class DetectionsTests(unittest.TestCase):
             predictions=fo.Detections(
                 detections=[
                     fo.Detection(
+                        label="cat",
+                        bounding_box=[0.6, 0.6, 0.4, 0.4],
+                        confidence=0.9,
+                    ),
+                    fo.Detection(
                         label="dog",
                         bounding_box=[0.1, 0.1, 0.4, 0.4],
                         confidence=0.9,
-                    )
+                    ),
                 ]
             ),
         )
@@ -1163,11 +1168,17 @@ class DetectionsTests(unittest.TestCase):
             predictions=fo.Detections(
                 detections=[
                     fo.Detection(
+                        label="cat",
+                        bounding_box=[0.6, 0.6, 0.4, 0.4],
+                        confidence=0.9,
+                        mask=np.full((8, 8), True),
+                    ),
+                    fo.Detection(
                         label="dog",
                         bounding_box=[0.1, 0.1, 0.4, 0.4],
                         confidence=0.9,
                         mask=np.full((8, 8), True),
-                    )
+                    ),
                 ]
             ),
         )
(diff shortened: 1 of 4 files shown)


tests/unittests/evaluation_tests.py
```diff
@@ -1083,11 +1083,16 @@ def _make_detections_dataset(self):
             ),
             predictions=fo.Detections(
                 detections=[
+                    fo.Detection(
```
brimoor, line 1086: These updated unit tests fail on `develop` due to the "missing false positive" scenario, but now they pass.

ehofesmann (approved):
LGTM, good catch on that `elif` case! Tested this a few times and confirmed that it results in the expected behavior now :raised_hands:

manushreegangwar (approved):
LGTM!

kaixi-wang (approved):
👍

Issue #1607, opened 2022-02-17
[BUG] Unable to filter videos with >16MB of frame labels in the App


brimoor:
Trying to use the App's `label`/`confidence` filters to filter frame labels of a video with >16MB of frame labels can cause an error like so:

Shouldn't we be able to detach frames when loading the grid below? Frame data is only needed when streaming data to an individual Looker instance
https://github.com/voxel51/fiftyone/blob/4c96e97a640524b3398e34a8e4c56429276500c0/fiftyone/server/main.py#L265

<img width="1194" alt="Screen Shot 2022-02-17 at 11 31 45 AM" src="https://user-images.githubusercontent.com/25985824/154540433-5a6cea7f-1340-43da-b4b2-650cbb64ba1e.png">

benjaminpkane:
I have brought this up before, but I don't think I was heard. This is a core library issue and does not necessarily require my attention.

See [here](https://stackoverflow.com/questions/45724785/aggregate-lookup-total-size-of-documents-in-matching-pipeline-exceeds-maximum-d).

brimoor:
If you find yourself with a video dataset with one or more samples whose frame labels exceed 16MB, here are some strategies to workaround the issue:

First, use [frames_sizes()](https://github.com/voxel51/fiftyone/issues/1822#issuecomment-1173151476) to retrieve the IDs of samples whose frame labels exceed 16MB.

**If only a few samples are too large:**

Consider either omitting (using [exclude()](https://voxel51.com/docs/fiftyone/api/fiftyone.core.collections.html#fiftyone.core.collections.SampleCollection.exclude)) or deleting (using [delete_samples()](https://voxel51.com/docs/fiftyone/api/fiftyone.core.dataset.html#fiftyone.core.dataset.Dataset.delete_samples)) the samples from the dataset.

```py
# Option 1: omit large samples when you need to use the App
view = dataset.exclude(ids_of_large_samples)
session = fo.launch_app(view)

# Option 2: permanently delete large samples
dataset.delete_samples(ids_of_large_samples)
session = fo.launch_app(dataset)
```

**If many samples are too large**

Option 1: If your frame labels are distributed across multiple frame fields, create separate datasets containing subsets of the frame labels that each total <16MB individually:

```py
print(dataset.get_frame_field_schema())
# field1, field2, etc

dataset1 = dataset.clone()
dataset1.delete_frame_field("field2")

dataset2 = dataset.clone()
dataset2.delete_frame_field("field1")
[...]

PR #5450, opened 2025-01-30
Limit frames lookup pipeline for grid

benjaminpkane:
## What changes are proposed in this pull request?

Limit the lookup pipeline for frames in grid samples request to one, as only the first frame is needed

## How is this patch tested? If it is not, please explain why.

Todo

## Release Notes

Todo

### What areas of FiftyOne does this PR affect?

-   [x] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

PR #5705, opened 2025-04-10
Optimize frame lookups to be as late as possible in aggregations

brimoor:
## Change log

- Optimizes frame lookups to be as late as possible in aggregation pipelines

cc @ehofesmann this seemingly small change results in big performance improvements in certain cases:
1. Significantly improves the performance of frame-level aggregations on views where `len(view) << len(dataset)`
2. Resolves many (but *not yet* all) instances of "16MB errors", aka [#1607](https://github.com/voxel51/fiftyone/issues/1607)

See below for examples of these cases.

## Example optimizations

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset.compute_metadata()
```

### Case 1: a view that doesn't filter on frame-level information

```py
view = dataset.match(F("metadata.size_bytes") > 2671165)
_ = view.count("frames")
```

#### Previous pipeline ❌ ❌ ❌ ❌ 

`$match` is between `$lookup` and `$unwind`, which means that this pipeline will fail with "16MB errors" on datasets with many frames.

```
[
    {
        '$lookup': {
            'from': 'frames.samples.67f7211b32b0fedf4532de6f',
            'let': {'sample_id': '$_id'},
            'pipeline': [
                {'$match': {'$expr': {'$eq': ['$$sample_id', '$_sample_id']}}},
                {'$sort': {'frame_number': 1}},
            ],
            'as': 'frames',
        },
    },
    {'$match': {'$expr': {'$gt': ['$metadata.size_bytes', 2671165]}}},
    {'$unwind': '$frames'},
    {'$count': 'count'},
]
```

#### New pipeline ✅ ✅ ✅ ✅   

`$match` comes first _and then_ `$lookup + $unwind` in immediate succession, which means no "16MB error" issues!

Additionally, `$lookup` is now only performed on samples that match the sample-level filter, which could result in *substantially faster* aggregations when `len(view) << len(dataset)`.

```
[
    {'$match': {'$expr': {'$gt': ['$metadata.size_bytes', 2671165]}}},
    {
        '$lookup': {
            'from': 'frames.samples.67f720ab45345727feab06bd',
            'let': {'sample_id': '$_id'},
            'pipeline': [
                {'$match': {'$expr': {'$eq': ['$$sample_id', '$_sample_id']}}},
                {'$sort': {'frame_number': 1}},
            ],
            'as': 'frames',
        },
    },
    {'$unwind': '$frames'},
    {'$count': 'count'},
]
```

### Case 2: an aggregation that involves explicit frame unwinding

```py
_ = dataset.values("frames[].frame_number")
```

#### Previous pipeline ❌ ❌ ❌ ❌ 

`$project` is between `$lookup` and `$unwind`, which means that this pipeline will fail with "16MB errors" on datasets with many frames.

```
[
    {
        '$lookup': {
            'from': 'frames.samples.67f7211b32b0fedf4532de6f',
            'let': {'sample_id': '$_id'},
            'pipeline': [
                {'$match': {'$expr': {'$eq': ['$$sample_id', '$_sample_id']}}},
                {'$sort': {'frame_number': 1}},
            ],
            'as': 'frames',
        },
    },
    {'$project': {'frames.frame_number': True}},
    {'$unwind': '$frames'},
    {
        '$project': {
            'values': {
                '$cond': {
                    'if': {'$gt': ['$frames.frame_number', None]},
                    'then': '$frames.frame_number',
                    'else': None,
                },
            },
        },
    },
]
```

#### New pipeline ✅ ✅ ✅ ✅   

`$lookup + $unwind` in immediate succession _and then_ `$project`, which means no "16MB error" issues!

```
[
    {
        '$lookup': {
            'from': 'frames.samples.67f720ab45345727feab06bd',
            'let': {'sample_id': '$_id'},
            'pipeline': [
                {'$match': {'$expr': {'$eq': ['$$sample_id', '$_sample_id']}}},
                {'$sort': {'frame_number': 1}},
            ],
            'as': 'frames',
        },
    },
    {'$unwind': '$frames'},
    {'$project': {'frames.frame_number': True}},
    {
        '$project': {
            'values': {
                '$cond': {
                    'if': {'$gt': ['$frames.frame_number', None]},
                    'then': '$frames.frame_number',
                    'else': None,
                },
            },
        },
    },
]
```

## Summary by CodeRabbit

- **New Features**
	- Introduced an advanced option in aggregations that allows direct retrieval of underlying processing pipelines.
  
- **Refactor**
	- Enhanced video sample processing to ensure consistent handling of frame data.
	- Streamlined data view construction for more efficient integration of frame and group information.

- **Bug Fixes**
	- Improved frame label selection logic to correctly include frame-level data when no specific fields are chosen.
	- Adjusted aggregation pipeline construction to properly unwind video frames when needed.

- **Tests**
	- Added tests verifying correct aggregation behavior and pipeline structure for video frames and nested detections.

### DIFF
--- a/fiftyone/core/view.py
+++ b/fiftyone/core/view.py
@@ -1655,66 +1655,39 @@ class DatasetView(foc.SampleCollection):
         if _attach_frames_idx is None and (attach_frames or frames_only):
             _attach_frames_idx = len(_pipelines)
 
-        #######################################################################
-        # Insert frame lookup pipeline(s) if needed
-        #######################################################################
-
         if _attach_frames_idx1 is not None and _attach_frames_idx is not None:
             _attach_frames_idx = _attach_frames_idx1
 
-        if _attach_frames_idx0 is not None and _attach_frames_idx is not None:
-            # Two lookups are required; manually do the **last** one and rely
-            # on dataset._pipeline() to do the first one
-            attach_frames = True
-            _pipeline = self._dataset._attach_frames_pipeline(
-                limit=limit_frames, support=support
-            )
-            _pipelines.insert(_attach_frames_idx, _pipeline)
-        elif _found_flattened_videos and _attach_frames_idx is not None:
-            # Must manually attach frames after the group $lookup
-            attach_frames = None  # special syntax: frames already attached
-            _pipeline = self._dataset._attach_frames_pipeline(
-                limit=limit_frames, support=support
-            )
-            _pipelines.insert(_attach_frames_idx, _pipeline)
-        elif _attach_frames_idx0 is not None or _attach_frames_idx is not None:
-            # Exactly one lookup is required; rely on dataset._pipeline() to
-            # do it
-            attach_frames = True
+        _inserts = []
 
-        # @todo use the optimization below instead, which injects frames as
-        # late as possible in the pipeline. We can't currently use it because
-        # there's some issue with poster frames in the App if the frames are
-        # not attached first...
+        # Handles incrementing idx values in case of multiple inserts
+        def _adjust(idx):
+            idx += sum(i < idx for i in _inserts)
+            _inserts.append(idx)
+            return idx
 
-        """
+        # Insert frame lookup pipeline(s) if needed
         if _attach_frames_idx0 is not None or _attach_frames_idx is not None:
             attach_frames = None  # special syntax: frames already attached
 
             if _attach_frames_idx0 is not None:
                 _pipeline = self._dataset._attach_frames_pipeline(
-                    support=support
+                    limit=limit_frames, support=support
                 )
-                _pipelines.insert(_attach_frames_idx0, _pipeline)
+                _pipelines.insert(_adjust(_attach_frames_idx0), _pipeline)
 
             if _attach_frames_idx is not None:
-                if _attach_frames_idx0 is not None:
-                    _attach_frames_idx += 1
-
                 _pipeline = self._dataset._attach_frames_pipeline(
-                    support=support
+                    limit=limit_frames, support=support
                 )
-                _pipelines.insert(_attach_frames_idx, _pipeline)
-        """
-
-        #######################################################################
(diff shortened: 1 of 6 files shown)


fiftyone/core/view.py
```diff
@@ -1637,66 +1637,39 @@ def _pipeline(
...
-                limit=limit_frames, support=support
-            )
-            _pipelines.insert(_attach_frames_idx, _pipeline)
-        elif _attach_frames_idx0 is not None or _attach_frames_idx is not None:
-            # Exactly one lookup is required; rely on dataset._pipeline() to
-            # do it
-            attach_frames = True
+        _inserts = []
 
-        # @todo use the optimization below instead, which injects frames as
```
brimoor, line 1667: @benjaminpkane I first tried to make this optimization awhile ago, but, as documented here, when I initially tried it, there was some App rendering issue where poster frames would not show labels.

However, I am not seeing any issues with first-frame label rendering now, so it seems like whatever the previous limitation was has now been lifted! 🤞🤞🤞

A couple questions for you:
- Do you know what might have changed to resolve the symptom reported in this `@todo`?
- Do you foresee any issues with `$lookup` potentially occurring later in pipelines now?

benjaminpkane: It is likely due to https://github.com/voxel51/fiftyone/pull/5450. I don't see any issue with this PR. Though it looks like some tests need to be updated

brimoor: Good catch, test has been updated!

kaixi-wang:
I remember looking into frame aggregations a long time ago and found that swapping the base collection from the sample to the frame collection overcomes nearly all 16mb issues and avoids the costly lookup+ unwind. Wondering why that approach wouldn't work instead?

brimoor:
I agree we should find a way to minimize unnecessary lookups, but its not quite as simple as just swapping the base collections; it will take more work than that.

swheaton:
Ben has more weight on this one but lgtm also 👍🏼

PR #5708, opened 2025-04-10
add databricks useragent

ritch:
Add databricks useragent info for all usage of fiftyone.


fiftyone/utils/databricks.py
```diff
@@ -0,0 +1,18 @@
+"""
+Databricks utilities.
+
+| Copyright 2017-2025, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+import fiftyone as fo
```
ritch, line 8: This might cause a circular dep...we only really need the __version__.

brimoor (changes requested):
I know this is just a POC, but as an implementation detail: semantically, I don't love the idea of a `databricks` module being imported for all FO users, most of which aren't using databricks at all.

It would be better to lazily do this when something databricks-y is actually done.

fiftyone/utils/databricks.py
brimoor: This indeed looks like a circular dep to me

ritch:
> It would be better to lazily do this when something databricks-y is actually done.

I can think of a few, but no single place would guarantee the coverage other than the existing impl: `import fiftyone`

Other places that might make sense:

 - `load_dataset`
 - `launch_app`
 - `sort_by_similarity`

benjaminpkane:
Note, this is how we currently [detect databricks notebooks](https://github.com/voxel51/fiftyone/blob/7710beb5a8aee0f228681ef5eb239f03284cd8a0/fiftyone/core/context.py#L73)

fiftyone/utils/useragent.py
```diff
@@ -0,0 +1,19 @@
+"""
+Databricks utilities.
+
+| Copyright 2017-2025, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+import fiftyone.constants as foc
```
ritch, line 8: @brimoor I've resolved the circular dep by using the constant directly.

fiftyone/utils/useragent.py
```diff
@@ -0,0 +1,19 @@
+"""
+User agent utilities.
+
+| Copyright 2017-2025, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+import fiftyone.constants as foc
+import fiftyone.core.utils as fou
```
findtopher, line 9: this is an unused import; is it necessary?

fiftyone/utils/useragent.py
```diff
@@ -0,0 +1,18 @@
...
+import fiftyone.constants as foc
+
+
+def with_fiftyone_useragent():
+    try:
+        from databricks.sdk import useragent
+
+        useragent.with_partner("voxel51")
+        useragent.with_product("fiftyone", foc.VERSION)
+    except:
```
brimoor, line 17: better to swallow *all* exceptions here IMO as databricks agent info is strictly optional

brimoor (approved):
LGTM

PR #5711, opened 2025-04-10
Use `FiftyOneTorchDataset` when applying `TorchImageModel`s

jacobsela:
## Change log

1. Formalizes and officially documents the `SampleCollection.to_torch()` interface for optimized model inference
2. Generalized `FiftyOneTorchDataset` by adding a new `GetItem` class for loading model inputs for samples
3. Adds a `GetItem` for `TorchImageModel`
4. Refactored methods like `apply_model()` and `compute_embeddings()` to use `FiftyOneTorchDataset` by default when possible
5. All models can expose themselves to the new implementation by implementing `SupportsGetItem`

## Example usage

The results of `apply_model` are the same before and after this code change:

```py
import fiftyone as fo
import fiftyone.zoo as foz

model = foz.load_zoo_model("clip-vit-base32-torch", classes=["truck", "dog"])

dataset.apply_model(model, label_field="predictions", batch_size=64)
```

### Benchmarked throughput on OSS to make sure there isn't slowdown

Using same code as above with `clip-vit-base32-torch`:
- Before: ~1.1k samples / second
- After: ~1.4k samples / second


fiftyone/core/models.py
```diff
@@ -713,56 +722,105 @@ def handle_errors(batch):
...
+        worker_init_fn=None
+        if not use_fotd
+        else fout.FiftyOneTorchDataset.worker_init,
     )
 
 
+def _make_fo_torch_dataset(
+    samples,
+    model,
+    skip_failures,  # have to implement this
```
manushreegangwar, line 816: nit: remove comment

fiftyone/utils/torch.py
```diff
@@ -253,6 +255,124 @@ def ensure_torch_hub_requirements(
...
+    Subclasses should interact with `required_fields` and `field_mapping` through the properties.
+    the method :method:`fiftyone.utils.torch.GetItem.add_required_fields`
+    can be used to add fields to required_fields.
+    Required fields should be set in the subclass's `__init__` method.
+    Required fields cannot repeat. If a field is added twice, it will be ignored.
+    Make sure to check if a superclass has already added a field before adding to it.
+    """
+
+    def __init__(self, field_mapping=None, **kwargs):
+        super().__init__(**kwargs)
```
manushreegangwar, line 278: I don't see the parent class.

fiftyone/utils/torch.py
```diff
@@ -253,6 +255,124 @@ def ensure_torch_hub_requirements(
...
+        if not hasattr(self, "_field_mapping_dict"):
+            self._field_mapping_dict = {}
+        if value is None:  # generally on init
+            value = {k: k for k in self.required_fields}
+            for k, v in self._field_mapping_dict.items():
+                # if mixins have already set the field mapping, we should keep it
+                value[k] = v
+        if not isinstance(value, dict):
+            raise ValueError("field_mapping must be a dictionary.")
+        for k, v in value.items():
```
manushreegangwar, line 341: nit: only keys are used.

fiftyone/utils/torch.py
```diff
@@ -253,6 +255,124 @@ def ensure_torch_hub_requirements(
...
+
+        self._field_mapping_dict = value
+
+    def update_field_mapping(self):
+        """
+        Update the field mapping for the get_item function.
+        """
+        for f in self.required_fields:
+            if f not in self.field_mapping:
+                self.field_mapping[f] = f
```
manushreegangwar, line 359: you can use `update_field_mapping` in `field_mapping` setter.

fiftyone/utils/torch.py
```diff
@@ -253,6 +255,124 @@ def ensure_torch_hub_requirements(
...
+
+    def validate_compatible_samples(self, samples):
+        """
+        Validate that the samples are compatible with the GetItem.
+        This should be called every time so we can fail quickly and loudly when needed.
+        TODO: Add type checking for the fields.
+        """
+        # is this the correct way to do this?
+        sample = samples.first()
+        foval.get_fields(sample, self.field_mapping.values(), allow_none=False)
```
manushreegangwar, line 369: This would only check that the fields exist with a value that is not None on the first sample. This doesn't gaurantee the fields are not None on all samples.

fiftyone/utils/torch.py
```diff
@@ -586,6 +744,15 @@ def __exit__(self, *args):
             self._no_grad.__exit__(*args)
             self._no_grad = None
 
+    def _build_get_item(self, **kwargs):
+        return ImageGetItem(
+            transform=self._transforms,
```
manushreegangwar, line 749: I have noticed some of the models (mostly non-torch) overriding the `transforms` property but not `self._transforms`. Best to use `self.transforms` here and elsewhere in this file but can be a separate refactor.

fiftyone/core/models.py
```diff
@@ -713,56 +722,105 @@ def handle_errors(batch):
...
+    field_mapping=None,
+):
+    get_item = model._build_get_item(field_mapping=field_mapping)
```
manushreegangwar, line 819: this is a private method. You can make a public wrapper.

fiftyone/core/models.py
```diff
@@ -685,23 +695,22 @@ def _iter_batches(video_reader, batch_size):
...
-        samples=samples,
-        transform=model.transforms,
-        use_numpy=use_numpy,
-        force_rgb=True,
-        skip_failures=skip_failures,
-    )
+    def __call__(self, batch):
+        return self._collate_fn(batch)
 
+    @staticmethod
```
manushreegangwar, line 713: this is fine but avoid the use of `staticmethod`. you can make a private function. better for testing.

manushreegangwar (changes requested):
Mostly looks good. Few comments. I will test the changes on my end this week.

fiftyone/utils/torch.py
jacobsela: There is no parent class (technically the base python class for all such objects is `object`). That line is there to avoid MRO bugs with mixins. This is also why I added the mixins test for this class.

fiftyone/utils/torch.py
jacobsela: the `field_mapping` setter is currently used in `update_field_mapping`. This function is there for clarity.

fiftyone/utils/torch.py
jacobsela: It doesn't just check that the values are not `None`, it also checks that the `get_item` is compatible with the sample (see the next two lines).

However, it is true that this doesn't guarantee anything for all of the samples, just the first one. This is fine though. There isn't really a generic way of checking that all field values are compatible for all samples without running the get item on all of them. Since this is very computationally costly, we rely on the heuristic of just checking the first sample.

In general, the goal of this method isn't to guarantee that the `GetItem` will work, it's just to catch simple errors (like the dataset not having all of the required fields) early on.

fiftyone/utils/torch.py
jacobsela: I agree this is something we should probably separately address.

We need to establish either `self._transforms` or the property `self.transforms` as the single source of truth. If different subclasses are implementing different parts of this code then this should be fixed. I tend to agree that `self.transforms` makes more sense.

fiftyone/core/models.py
jacobsela: python doesn't really have a notion of "private" methods, this is just a convention. added anyways.

fiftyone/core/models.py
jacobsela: Why is the use of `staticmethod` problematic?

manushreegangwar: In general, writing unit tests for `staticmethod` can be less than ideal. The same for overrides in subclasses. Adds a bit of complexity that can be avoided.

tests/unittests/get_item_tests.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def test_add_required_fields(self):
+        gi = DummyGetItem()
+        self.assertEqual(set(gi.required_fields), set(["foo", "bar"]))
+        gi.add_required_fields("baz")
+        self.assertEqual(set(gi.required_fields), set(["foo", "bar", "baz"]))
+        gi.add_required_fields("baz")
+
+        gi.add_required_fields("baz")
+        self.assertEqual(set(gi.required_fields), set(["foo", "bar", "baz"]))
```
manushreegangwar, line 45: Shouldn't you be checking against `gi.required_fields` since `add_required_fields` is expected to handle the `set` part?

manushreegangwar (commented):
Tested on teams

`19.5s for 10000 samples`

jacobsela:
@manushreegangwar This doesn't work out of the box in teams as shown in the demo. Please see the branch on teams that I DMed you. You should see substantially more speedup.

tests/unittests/get_item_tests.py
jacobsela: good catch, fixed.

manushreegangwar (approved):

manushreegangwar:
Performance on OSS with `cifar-10` dataset and clip:
on this branch:  1.5K samples/s
on develop: 981.9 samples/s

fiftyone/utils/torch.py
```diff
@@ -255,6 +256,124 @@ def ensure_torch_hub_requirements(
...
+        if not defined, we will assume the fields are the same
+        When this is updated, the behavior of __call__ should be updated accordingly.
+        This is what allows the user to specify the fields in the sample that are used as input to the model.
+        """
+        if not hasattr(self, "_field_mapping_dict"):
+            self._field_mapping_dict = {}
+        return self._field_mapping_dict
+
+    @field_mapping.setter
+    def field_mapping(self, value):
```
brimoor, line 332: We already have the existing concept of `SamplesMixin`, which defines a `needs_fields` property:

```py
class SamplesMixin(object):
    """Mixin for :class:`Model` classes that need samples for prediction.

    Models can implement this mixin to declare that they require one or more
    fields of the current sample when performing inference on its media.

    The fields are get/set via :meth:`needs_fields`, which is a dict that maps
    model-specific keys to sample field names::

        model.needs_fields = {"key1": "field1", "key2": "field2", ...}
    """
```

Importantly, `needs_fields` is used in `apply_model()` to select only the fields that the model actually needs during inference:

https://github.com/voxel51/fiftyone/blob/c677ff4fe24f2f2bbc0d765510f294bb4d0a8f7c/fiftyone/core/models.py#L198-L202

This optimization is very import when a dataset has many fields, as it ensures that only the required fields are loaded from the database.

`GetItem.field_mapping` is serving exactly the same purpose, so it should also be used by `apply_model()` to perform a `select_fields()` call to limit what fields are loaded into memory during inference. Is that happening? Or am I just not seeing it in the code?

brimoor (changes requested):
@jacobsela I know you've been working on this for awhile and you want to wrap it up. At the same time, as this is a core interface, I want to make sure we're happy with the interface precedent that we're adding here.

I'm happy to chat live if helpful to finalize any design details so we can get this merged and done ASAP.

fiftyone/utils/torch.py
```diff
@@ -255,6 +256,124 @@ def ensure_torch_hub_requirements(
...
+        return self.sample_dict_to_input(sample)
+
+    @property
+    def required_fields(self):
+        """a list of fields that are required in the sample for the GetItem to work."""
+        if not hasattr(self, "_required_fields_list"):
+            self._required_fields_list = ()
+        return self._required_fields_list
+
+    def add_required_fields(self, value):
```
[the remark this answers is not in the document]
jacobsela: Mainly so that mixins can add required fields without directly editing the underlying `self._required_fields_list`. This allows for easily building modular `GetItem`s. Though frankly I am starting to dislike mixins.

fiftyone/utils/torch.py
```diff
@@ -255,6 +256,124 @@ def ensure_torch_hub_requirements(
...
+    Required fields should be set in the subclass's `__init__` method.
+    Required fields cannot repeat. If a field is added twice, it will be ignored.
+    Make sure to check if a superclass has already added a field before adding to it.
+    """
+
+    def __init__(self, field_mapping=None, **kwargs):
+        super().__init__(**kwargs)
+        self.field_mapping = field_mapping
+
+    def sample_dict_to_input(self, sample_dict):
```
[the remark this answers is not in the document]
jacobsela: We did, haven't got a chance to do it, didn't realize it was blocking.

fiftyone/utils/torch.py
```diff
@@ -255,6 +256,124 @@ def ensure_torch_hub_requirements(
...
+        Returns:
+            model input
+        """
+        raise NotImplementedError("Subclass should implement this method.")
+
+    def __call__(self, sample):
+        return self.sample_dict_to_input(sample)
+
+    @property
+    def required_fields(self):
```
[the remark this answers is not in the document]
jacobsela: So `required_fields` is just a way for the model to advertise what fields it needs, it's equivalent to the keys of the `needs_fields` property in `SamplesMixin`. `field_mapping` is exactly `SamplesMixin.needs_fields`.

fiftyone/utils/torch.py
jacobsela: No, you're 100% right, this is on the optimizations backlog for this project. I can bump it up.

fiftyone/utils/torch.py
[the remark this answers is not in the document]
brimoor: If you want people to use `GetItem`, you can't change it once you ship it!!!

fiftyone/utils/torch.py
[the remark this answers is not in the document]
brimoor: I'm still not seeing why `add_required_fields` needs to exist. Why isn't this sufficient?

```py
get_item = GetItem(field_mapping=field_mapping, ...)
```

fiftyone/utils/torch.py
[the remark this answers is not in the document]
brimoor: If we don't need optional fields, then we don't need `required_fields` as a separately defined concept. It can just be inferred as `list(field_mapping.values())`.

fiftyone/utils/torch.py
brimoor: I'm willing to bet that any performance enhancement you're getting from supporting batch inference + data loaders will be lost if you don't include `select_fields()`.

fiftyone/utils/torch.py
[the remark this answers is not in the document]
jacobsela: That is sufficient from a functionality standpoint. I was hoping to make use of something like this pattern:

```
class GetItemEmbeddingsMixin:
    """
    Mixin for GetItem classes that are used for embedding tasks.
    """
    def __init__(self, dtype=torch.float32, **kwargs):
        super().__init__(**kwargs)
        self.add_required_fields("embeds")
        self.embeddings_dtype = dtype

    def load_embeddings(self, sample):
        embeds = sample[self.field_mapping[self._gi_embeds_field]]
        return torch.tensor(embeds, dtype=self.embeddings_dtype)
```

And then all `GetItems` that require embeddings can just subclass that mixin. Without `add_required_fields` the mixin would have to directly edit the internal list, and I wanted to use that interface to avoid conflicts.

fiftyone/utils/torch.py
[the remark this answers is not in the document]
jacobsela: That's fair enough. This will require some finesse. Let's talk about it offline.

fiftyone/utils/torch.py
```diff
@@ -255,6 +256,124 @@ def ensure_torch_hub_requirements(
...
+        Update the field mapping for the get_item function.
+        """
+        for f in self.required_fields:
+            if f not in self.field_mapping:
+                self.field_mapping[f] = f
+
+    def validate_compatible_samples(self, samples):
+        """
+        Validate that the samples are compatible with the GetItem.
+        This should be called every time so we can fail quickly and loudly when needed.
```
kaixi-wang, line 365: For performance reasons, I don’t think we should ever call validation on a per sample basis in the torch pipeline. If you must ensure fields, do it before assigning the samples by filtering only the ones with non null values when setting the sample ids

fiftyone/core/models.py
```diff
@@ -250,6 +250,7 @@ def apply_model(
                 skip_failures,
                 filename_maker,
                 progress,
+                field_mapping=kwargs.get("field_mapping", None),
```
brimoor, line 253: @jacobsela doesn't this code:

https://github.com/voxel51/fiftyone/blob/81c954b1292dd2e67d57b096cdd8287d67462eee/fiftyone/core/models.py#L198-L202

mean that, as currently written, `FiftyOneTorchDataset` will only ever get `samples.select_fields()` passed to it, and therefore that models that use `required_key` to specify non-default fields won't have access to those fields and thus inference won't work?

fiftyone/utils/torch.py
```diff
@@ -519,8 +587,53 @@ def __init__(self, d):
...
+    ):
+        super().__init__(**kwargs)
+        self.transform = transform
+        self.raw_inputs = raw_inputs
+        self.using_half_precision = using_half_precision
+        self.use_numpy = use_numpy
+
+    def __call__(self, d):
+        img = _load_image(
+            # hardcoded because that's how it was in fiftyone.core.models._make_data_loader
```
brimoor, line 609: What is hard-coded here? Feels like you're apologizing for something but IDK what

fiftyone/utils/torch.py
```diff
@@ -519,8 +587,53 @@ def __init__(self, d):
...
+        return img
+
+    @property
+    def required_keys(self):
+        return ["filepath"]
+
+
 class TorchImageModel(
-    TorchEmbeddingsMixin, fom.TorchModelMixin, fom.LogitsMixin, fom.Model
+    fom.SupportsGetItem,
```
brimoor, line 632: Just to clarify, this one line has **huge** consequences- the vast majority of zoo models are `TorchImageModel`, which means that they are *all* now going to use `to_torch()` to run inference when passed to `apply_model()`, `compute_embeddings()`, etc.

Are we really ready to make that commitment? 😄 

For example, I'm pretty sure upon static review that `required_keys` would not actually work as-is due to `select_fields()`.

fiftyone/utils/torch.py
```diff
@@ -1612,6 +1731,9 @@ def __init__(
                 f"Your start method is {start_method}. It is recommended to use 'spawn' or 'forkserver' with this class."
             )
 
+        self.field_mapping = get_item.field_mapping.copy()
+        samples = samples.select_fields(list(self.field_mapping.values()))
```
brimoor, line 1735: Either this field selection needs to happen in `apply_model()`, `compute_embeddings()`, etc boilerplate. Or it needs to *not* happen there so it can *only* happen here.

brimoor: btw I'm somewhat surprised that this code actually works in data loaders, which run in separate processes.

In the `map_samples()` project, we had to manually reestablish database connections in each worker process:
https://github.com/voxel51/fiftyone/blob/be5889872e0b820df117b574105f0ce498f88d8b/fiftyone/core/map/process.py#L234-L246

brimoor (changes requested):

fiftyone/utils/torch.py
jacobsela: I see the point on select fields, will fix.

And I do individually connect each worker to the database. see `_load_samples`

fiftyone/utils/torch.py
jacobsela: I believe I meant `force_rgb`

fiftyone/utils/torch.py
brimoor: It's not a matter of my comfort. We just need our zoo models to continue to work. So we need to test all models that now use `SupportsGetItem` to ensure they do.

We have these tests as a starting point for that:
https://github.com/voxel51/fiftyone/blob/develop/tests/intensive/model_zoo_tests.py

fiftyone/utils/torch.py
brimoor: Ah okay, `_load_samples()`. Good! I knew that needed to exist.

Note that in the `map_samples()` code I linked, there is a `food._disconnect()` command. That was *required* in order for that project to work correctly in all cases due to this:
https://pymongo.readthedocs.io/en/stable/faq.html#using-pymongo-with-multiprocessing 

> Instances of MongoClient copied from the parent process have a high probability of deadlock in the child process due to [inherent incompatibilities between fork(), threads, and locks](https://pymongo.readthedocs.io/en/stable/faq.html#pymongo-fork-safe-details). PyMongo will attempt to issue a warning if there is a chance of this deadlock occurring.

Perhaps that is why you added this:

```py
        start_method = multiprocessing.get_start_method(allow_none=True)
        if start_method not in ["spawn", "forkserver"]:
            warnings.warn(
                f"Your start method is {start_method}. It is recommended to use 'spawn' or 'forkserver' with this class."
            )
```

I believe if you add `food._disconnect()` then `FiftyOneTorchDataset` should work with `fork`ed processes too. Moreover, I think it is best practice to re-establish database connection in each worker process anyway. Less chance for deadlocks.

brimoor: As a side note- I believe that Torch data loaders will default to `spawn` multiprocessing contexts on unix systems (even though the default is `fork`), as there are CUDA reasons why `spawn` is preferred.

Having said all that, best to insert `food._disconnect()` so that *our code* can handle whatever multiprocessing context the user may want to throw at us.

jacobsela: added

brimoor (approved):
@jacobsela LGTM! I cleaned up the documentation for us 😄 

Also, please squash merge to make it easier to cherry-pick this into `release/v1.6.0`

PR #5723, opened 2025-04-12
Automatically update `last_modified_at` when deleting samples/frames

brimoor:
## Change log

- `Dataset.last_modified_at` is now automatically updated when samples are deleted
- `Sample.last_modified_at` is now automatically updated when frames are deleted

## Motivation

Currently the behavior of the `last_modified_at` properties of datasets/samples/frames is as follows:
1. Modifications to a dataset's schema/metadata automatically increment `Dataset.last_modified_at`
2. Modifications to samples automatically increment `Sample.last_modified_at`
3. Modifications to frames automatically increment `Frame.last_modified_at`
4. Users can call `Dataset.sync_last_modified_at()` to manually pull timestamps "up one level":
    - `Dataset.last_modified_at` is increased if samples have been updated more recently than the dataset's metadata
    - `Sample.last_modified_at` is increased if that sample's frames have been updated more recently than the sample itself

This current logic leaves a gap: if a user **deletes** a sample/frame, then (unless that sample/frame happens to be the *newest*), there is no way to detect that a change has been made via the available mechanisms:
- `dataset.last_modified_at`: currently only tracks last update to dataset metadata
- `dataset.max("last_modified_at")`: currently only tracks last update to *currently existing* samples
- `dataset.max("frames.last_modified_at")`: currently only tracks last update to *currently existing* frames

With this PR, there is now an available mechanism to check for sample/frame deletions:
- Check `dataset.last_modified_at` to see if samples have been deleted 
- Check `dataset.max("last_modified_at")` to see if frames have been deleted

## Use case 

Why might one care about detecting deletions? Detecting when saved views need to be refreshed for example! https://github.com/voxel51/fiftyone-plugins/pull/122

## Implementation discussion

It is noted that the chosen implementation (automatic syncing "one level up") is inconsistent with Point 4 above (manual syncing "one level up"). But something needed to be done.

So, why did we choose "manual syncing" in Point 4 rather than "automatic syncing" in the first place? The thinking was that edits are extremely common, and thus it would be very expensive to increment `last_modified_at` in another collection as a side effect of a sample/frame update. Instead, we chose to index `last_modified_at` by default so that operations like `dataset.max("last_modified_at")` are performant.

Deletions are far less common operations (and moreover they are usually *batched*; indeed we only have `delete_samples()` and `delete_frames()` and no singular counterparts) so the cost of the extra `$set` for deletions is typically negligible.

### Image datasets

```py
import fiftyone as fo
import fiftyone.zoo as foz

last_modified_at1 = dataset.last_modified_at

# DatasetView.keep()
dataset.take(150).keep()
last_modified_at2 = dataset.last_modified_at
assert last_modified_at2 > last_modified_at1

# DatasetView.clear()
dataset.take(50).clear()
last_modified_at3 = dataset.last_modified_at
assert last_modified_at3 > last_modified_at2

# Dataset.delete_samples()
dataset.delete_samples(dataset.take(50))
last_modified_at4 = dataset.last_modified_at
assert last_modified_at4 > last_modified_at3

# Dataset.clear()
dataset.clear()
last_modified_at5 = dataset.last_modified_at
assert last_modified_at5 > last_modified_at4
```

### Video datasets

```py
import fiftyone as fo
import fiftyone.zoo as foz

last_modified_at1 = dataset.last_modified_at

# Dataset.delete_samples()
dataset.delete_samples(dataset.take(1))
last_modified_at2 = dataset.last_modified_at
assert last_modified_at2 > last_modified_at1

# Dataset.delete_frames()
frame_ids = dataset.match_frames(F("frame_number") > 100).values("frames.id", unwind=True)
dataset.delete_frames(frame_ids)
last_modified_at3 = dataset.max("last_modified_at")
assert last_modified_at3 > last_modified_at2

# DatasetView.keep_frames()
dataset.match_frames(F("frame_number") <= 50).keep_frames()
last_modified_at4 = dataset.max("last_modified_at")
assert last_modified_at4 > last_modified_at3

# DatasetView.clear_frames()
dataset.take(4).clear_frames()
last_modified_at5 = dataset.max("last_modified_at")
assert last_modified_at5 > last_modified_at4

# Dataset.clear_frames()
dataset.clear_frames()
last_modified_at6 = dataset.max("last_modified_at")
assert last_modified_at6 > last_modified_at5

# Frame deletions increment Sample.last_modified_at but *not* Dataset.last_modified_at
last_modified_at7 = dataset.last_modified_at
assert last_modified_at7 == last_modified_at2

# Call Dataset.sync_last_modified_at() to update Dataset.last_modified_at
dataset.sync_last_modified_at()
last_modified_at8 = dataset.last_modified_at
assert last_modified_at8 > last_modified_at7
```

## Summary by CodeRabbit

- **New Features**
  - Enhanced dataset tracking now updates modification timestamps automatically after sample and video frame changes.
  - Improved video sample handling with optional reloading of frame data for more precise control.
  
- **Tests**
  - Added comprehensive tests ensuring modification timestamps accurately reflect changes to dataset and video content.

### DIFF
--- a/fiftyone/core/frame.py
+++ b/fiftyone/core/frame.py
@@ -18,6 +18,7 @@ import fiftyone.core.odm as foo
 from fiftyone.core.singletons import FrameSingleton
 import fiftyone.core.utils as fou
 
+fos = fou.lazy_import("fiftyone.core.sample")
 fov = fou.lazy_import("fiftyone.core.view")
 
 
@@ -142,6 +143,10 @@ class Frames(object):
     @property
     def _frame_collection(self):
         return self._dataset._frame_collection
+
+    @property
+    def _sample_collection_name(self):
+        return self._dataset._sample_collection_name
 
     @property
     def _frame_collection_name(self):
@@ -488,9 +493,15 @@ class Frames(object):
                 "a dataset"
             )
 
-        delete_ops = self._save_deletions(deferred=deferred)
+        sample_ops, delete_ops = self._save_deletions(deferred=deferred)
         replace_ops = self._save_replacements(deferred=deferred)
-        return delete_ops + replace_ops
+
+        if deferred:
+            frame_ops = delete_ops + replace_ops
+        else:
+            frame_ops = None
+
+        return sample_ops, frame_ops
 
     def reload(self, hard=False):
         """Reloads all frames for the sample from the database.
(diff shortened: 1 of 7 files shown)


fiftyone/core/frame.py
```diff
@@ -677,37 +688,57 @@ def _to_frames_dict(self, include_private=False):
...
-
-        if self._delete_frames:
-            ops = [
-                DeleteOne(
-                    {
-                        "_sample_id": self._sample_id,
-                        "frame_number": frame_number,
-                    }
+        elif self._delete_frames:
+            sample_ops.append(
```
kaixi-wang, line 719: While I agree it makes sense to have the sample last updated at reflect deletion in frames, I'm concerned about performance, particularly for large datasets through the api, when nesting the update inside another operation. Perhaps doing it in 2 separate stages so that it can be handled differently through the api?

brimoor: Not sure what you're pointing out. The two `bulk_write()` methods are separate calls, which means separate API calls in the proxy.

benjaminpkane (approved):
LGTM

swheaton (approved):
without testing, looks fine. I swear we had added this in the original branch, but maybe it got lost when we were combining / updating branches.

PR #5725, opened 2025-04-13
Add evaluation-related fields to App sidebar group by default

brimoor:
## Change log

When performing evaluations, all fields added by the evaluation (including custom evaluation metrics, if applicable) are now added to a sidebar group with name `eval_key`. This reduces clutter and promotes better organization of fields in the App sidebar, which can have many fields on complex datasets, especially those with multiple evaluations.

The sidebar group is automatically renamed/deleted if the evaluation is later renamed/deleted.

## Tested by

- Unit tests are added verifying intended functionality and backwards compatibility/robustness to missing/renamed/existing/otherwise modified sidebar groups


ehofesmann (commented):
This makes a ton of sense to me!

fiftyone/core/collections.py
```diff
@@ -11384,6 +11384,60 @@ def _get_label_attributes_schema(self, label_field):
...
+        dataset = self._root_dataset
+        if dataset.app_config.sidebar_groups is None:
+            return
+
+        existing_group = None
+        for group in dataset.app_config.sidebar_groups:
+            if group.name == new_group_name:
+                existing_group = group
+
+        for group in dataset.app_config.sidebar_groups[:]:
```
swheaton, line 11421: is this doing anything?
```suggestion
        for group in dataset.app_config.sidebar_groups:
```

fiftyone/core/collections.py
```diff
@@ -11384,6 +11384,60 @@ def _get_label_attributes_schema(self, label_field):
...
+                    group.name = new_group_name
+
+                dataset.save()
+
+    def _delete_empty_sidebar_group(self, group_name):
+        dataset = self._root_dataset
+        if dataset.app_config.sidebar_groups is None:
+            return
+
+        for group in dataset.app_config.sidebar_groups[:]:
```
swheaton, line 11436: same
```suggestion
        for group in dataset.app_config.sidebar_groups:
```

swheaton (commented):
did not test but impl lgtm. 2 nits about weird looking `for group in dataset.app_config.sidebar_groups[:]`

fiftyone/core/collections.py
brimoor: Yeah it's subtle. `[:]` creates a copy of the list object in-memory so that I can `remove()` things from the original list while iterating over the copy in the loop.

swheaton: might be more clear to use `sidebar_groups.copy()` which would also create shallow copy like `[:]`

brimoor: Deal, but in return I get to complain when people use the walrus operator, which I think is unreadable 😄

brimoor: Okay, `feature_request := updated`

PR #5729, opened 2025-04-15
Convert ultralytics models to TorchImageModel

manushreegangwar:

These changes are for converting all `Ultralytics.YOLO` models to `TorchImageModel`

1. Load and apply all the YOLO models:
```
import fiftyone as fo
import fiftyone.zoo as foz

model_name = <MODEL_ZOO_NAME>
zoo_model = foz.load_zoo_model(
    model_name,
)
dataset.apply_model(zoo_model, label_field="yolo-torch-test")

from ultralytics import YOLO
model = YOLO("<MODEL_NAME>.pt")
dataset.apply_model(model, label_field="yolo-direct-test")
```
Correctness tested by comparing results against direct Ultralytics YOLO inference as well as inference on `develop`. 

2. Run `test/intensive/model_zoo_tests.py`

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced a unified model interface for all Ultralytics model types, streamlining configuration and usage.
  - Added dedicated output processors for each Ultralytics task (detection, segmentation, pose, classification, OBB), improving label conversion and postprocessing.
  - Added support for classification outputs with confidence thresholding and logits storage.

- **Refactor**
  - Consolidated multiple specialized model classes into a single, flexible model class.
  - Simplified model initialization, preprocessing, prediction, and output handling for improved maintainability and consistency.

### DIFF
--- a/fiftyone/utils/ultralytics.py
+++ b/fiftyone/utils/ultralytics.py
@@ -7,7 +7,6 @@ Utilities for working with
 |
 """
 
-from typing import Callable
 import numpy as np
 
 import fiftyone.core.labels as fol
@@ -52,6 +51,15 @@ def convert_ultralytics_model(model):
             "Unsupported model type; cannot convert %s to a FiftyOne model"
             % model
         )
+
+
+def _extract_track_ids(result):
+    """Get ultralytics track ids if present, else use Nones"""
+    return (
+        result.boxes.id.detach().cpu().numpy().astype(int)
+        if result.boxes.is_track
+        else [None] * result.boxes.conf.size(0)
+    )
 
 
 def obb_to_polylines(results, confidence_thresh=None, filled=False):
@@ -105,6 +113,59 @@ def _obb_to_polylines(result, filled, confidence_thresh=None):
         )
         polylines.append(polyline)
     return fol.Polylines(polylines=polylines)
+
+
+def to_keypoints(results, confidence_thresh=None):
+    """Converts ``ultralytics.YOLO`` keypoints to FiftyOne format.
+    Args:
+        results: a single or list of ``ultralytics.engine.results.Results``
+        confidence_thresh (None): a confidence threshold to filter keypoints
+    Returns:
+        a single or list of :class:`fiftyone.core.labels.Keypoints`
+    """
+    single = not isinstance(results, list)
+    if single:
+        results = [results]
+
+    batch = [
+        _to_keypoints(r, confidence_thresh=confidence_thresh) for r in results
+    ]
+
+    if single:
+        return batch[0]
+
+    return batch
+
+
+def _to_keypoints(result, confidence_thresh=None):
+    if result.keypoints is None:
+        return None
+
+    classes = np.rint(result.boxes.cls.detach().cpu().numpy()).astype(int)
+    points = result.keypoints.xyn.detach().cpu().numpy().astype(float)
+    if result.keypoints.conf is not None:
+        confs = result.keypoints.conf.detach().cpu().numpy().astype(float)
+    else:
+        confs = itertools.repeat(None)
+    track_ids = _extract_track_ids(result)
+
+    keypoints = []
+    for cls, _points, _confs, idx in zip(classes, points, confs, track_ids):
+        if confidence_thresh is not None:
+            _points[_confs < confidence_thresh] = np.nan
+
+        label = result.names[cls]
+        _confidence = _confs.tolist() if _confs is not None else None
+
+        keypoint = fol.Keypoint(
+            label=label,
+            points=_points.tolist(),
+            confidence=_confidence,
+            index=idx,
+        )
+        keypoints.append(keypoint)
+
+    return fol.Keypoints(keypoints=keypoints)
 
 
 class FiftyOneYOLOModelConfig(fout.TorchImageModelConfig, fozm.HasZooModel):
@@ -200,14 +261,15 @@ class FiftyOneYOLOModel(fout.TorchImageModel):
         return transforms, ragged_batches
 
     def _preprocess_im(self, img):
-        orig_img = img
         if not isinstance(img, torch.Tensor):
             if isinstance(img, Image.Image):
                 img = img.convert("RGB")
-            img = [np.array(img)]
-            img = self._ultralytics_preprocess(img)
-        output = {"img": img, "orig_img": orig_img}
-        return output
+            orig_img = img
+            return {
+                "img": self._ultralytics_preprocess([np.asarray(img)]),
+                "orig_img": orig_img,
+            }
+        return {"img": img, "orig_img": img}
 
     def _ultralytics_preprocess(self, im):
         im = np.stack(self._model.predictor.pre_transform(im))
@@ -268,16 +330,13 @@ class FiftyOneYOLOModel(fout.TorchImageModel):
 
         output = self._forward_pass(images)
 
-        if isinstance(output, dict):
-            # This is required for Ultralytics post-processing.
-            output["orig_imgs"] = orig_images
-            height, width = _get_image_dims(orig_images[0])
-            output["imgs"] = images
+        # This is required for Ultralytics post-processing.
+        output["orig_imgs"] = orig_images
+        height, width = _get_image_dims(orig_images[0])
+        output["imgs"] = images
 
         if self._output_processor is None:
-            _output = (
-                output.get("preds") if isinstance(output, dict) else output
-            )
+            _output = output.get("preds")
             if isinstance(_output, torch.Tensor):
                 _output = _output.detach().cpu().numpy()
 
@@ -488,9 +547,42 @@ class UltralyticsDetectionOutputProcessor(
                     "boxes": result.boxes.xyxy,
                     "labels": result.boxes.cls.int(),
                     "scores": result.boxes.conf,
+                    "track_ids": _extract_track_ids(result),
                 }
                 batch.append(pred)
         return batch
+
+    def _parse_output(self, output, frame_size, confidence_thresh):
+        width, height = frame_size
+
+        boxes = output["boxes"].detach().cpu().numpy()
+        labels = output["labels"].detach().cpu().numpy()
+        scores = output["scores"].detach().cpu().numpy()
+        track_ids = output["track_ids"]
+
+        detections = []
+        for box, label, score, idx in zip(boxes, labels, scores, track_ids):
+            if confidence_thresh is not None and score < confidence_thresh:
+                continue
+
+            x1, y1, x2, y2 = box
+            bounding_box = [
+                x1 / width,
+                y1 / height,
+                (x2 - x1) / width,
+                (y2 - y1) / height,
+            ]
+
+            detections.append(
+                fol.Detection(
+                    label=self.classes[label],
+                    bounding_box=bounding_box,
+                    confidence=score,
+                    index=idx,
+                )
+            )
+
+        return fol.Detections(detections=detections)
 
 
 class UltralyticsSegmentationOutputProcessor(
@@ -518,19 +610,20 @@ class UltralyticsSegmentationOutputProcessor(
                 continue
             else:
                 pred = {
-                    "boxes": result.boxes.xyxy,
-                    "boxes_xywhn": result.boxes.xywhn,
+                    "boxes": result.boxes.xywhn,
                     "labels": result.boxes.cls.int(),
                     "scores": result.boxes.conf,
                     "masks": result.masks.data,
+                    "track_ids": _extract_track_ids(result),
                 }
                 batch.append(pred)
         return batch
 
-    def _parse_output(self, output, frame_size, confidence_thresh):
-        boxes = output["boxes_xywhn"].detach().cpu().numpy().astype(float)
+    def _parse_output(self, output, _, confidence_thresh):
+        boxes = output["boxes"].detach().cpu().numpy().astype(float)
         labels = output["labels"].detach().cpu().numpy()
         masks = output["masks"].detach().cpu().numpy() > self.mask_thresh
+        track_ids = output["track_ids"]
 
         boxes[:, 0] -= boxes[:, 2] / 2.0
         boxes[:, 1] -= boxes[:, 3] / 2.0
@@ -541,7 +634,9 @@ class UltralyticsSegmentationOutputProcessor(
             scores = itertools.repeat(None)
 
         detections = []
-        for box, label, mask, score in zip(boxes, labels, masks, scores):
+        for box, label, mask, score, idx in zip(
+            boxes, labels, masks, scores, track_ids
+        ):
             if (
                 confidence_thresh is not None
                 and score is not None
@@ -568,6 +663,7 @@ class UltralyticsSegmentationOutputProcessor(
                     bounding_box=list(box),
                     mask=sub_mask.astype(bool),
                     confidence=score,
+                    index=idx,
                 )
             )
 
@@ -575,7 +671,7 @@ class UltralyticsSegmentationOutputProcessor(
 
 
 class UltralyticsPoseOutputProcessor(
-    fout.KeypointDetectorOutputProcessor, UltralyticsPostProcessor
+    fout.OutputProcessor, UltralyticsPostProcessor
 ):
     """Converts Ultralytics Pose model outputs to FiftyOne format."""
 
@@ -587,26 +683,9 @@ class UltralyticsPoseOutputProcessor(
         super().__init__(classes)
         self.post_processor = post_processor
 
-    def __call__(self, output, frame_size, confidence_thresh=None):
-        results = self.post_process(output)
-        preds = self._to_dict(results)
-        return super().__call__(preds, frame_size, confidence_thresh)
-
-    def _to_dict(self, results):
-        batch = []
-        for result in results:
-            if not result.keypoints:
-                continue
-            else:
-                pred = {
-                    "boxes": result.boxes.xyxy,
-                    "labels": result.boxes.cls.int(),
-                    "scores": result.boxes.conf,
-                    "keypoints": result.keypoints.xyn,
-                    "keypoints_scores": result.keypoints.conf,
-                }
-                batch.append(pred)
-        return batch
+    def __call__(self, output, _, confidence_thresh=None):
+        preds = self.post_process(output)
+        return to_keypoints(preds, confidence_thresh=confidence_thresh)
 
 
 class UltralyticsOBBOutputProcessor(
--- a/fiftyone/utils/ultralytics.py
+++ b/fiftyone/utils/ultralytics.py
@@ -55,137 +55,6 @@ def convert_ultralytics_model(model):
         )
 
 
-def _extract_track_ids(result):
-    """Get ultralytics track ids if present, else use Nones"""
-    return (
-        result.boxes.id.detach().cpu().numpy().astype(int)
-        if result.boxes.is_track
-        else [None] * result.boxes.conf.size(0)
-    )
-
-
-def to_detections(results, confidence_thresh=None):
-    """Converts ``ultralytics.YOLO`` boxes to FiftyOne format.
-
-    Args:
-        results: a single or list of ``ultralytics.engine.results.Results``
-        confidence_thresh (None): a confidence threshold to filter boxes
-
-    Returns:
-        a single or list of :class:`fiftyone.core.labels.Detections`
-    """
-    single = not isinstance(results, list)
-    if single:
-        results = [results]
-
-    batch = [
-        _to_detections(r, confidence_thresh=confidence_thresh) for r in results
-    ]
-
-    if single:
-        return batch[0]
-
-    return batch
-
-
-def _to_detections(result, confidence_thresh=None):
-    if result.boxes is None:
-        return None
-
-    classes = np.rint(result.boxes.cls.detach().cpu().numpy()).astype(int)
-    boxes = result.boxes.xywhn.detach().cpu().numpy().astype(float)
-    confs = result.boxes.conf.detach().cpu().numpy().astype(float)
-    track_ids = _extract_track_ids(result)
-
-    detections = []
-    for cls, box, conf, idx in zip(classes, boxes, confs, track_ids):
-        if confidence_thresh is not None and conf < confidence_thresh:
-            continue
-
-        label = result.names[cls]
-        xc, yc, w, h = box
-
-        detection = fol.Detection(
-            label=label,
-            bounding_box=[xc - 0.5 * w, yc - 0.5 * h, w, h],
-            confidence=conf,
-            index=idx,
-        )
-        detections.append(detection)
-
-    return fol.Detections(detections=detections)
-
-
-def to_instances(results, confidence_thresh=None):
-    """Converts ``ultralytics.YOLO`` instance segmentations to FiftyOne format.
-
-    Args:
-        results: a single or list of ``ultralytics.engine.results.Results``
-        confidence_thresh (None): a confidence threshold to filter boxes
-
-    Returns:
-        a single or list of :class:`fiftyone.core.labels.Detections`
-    """
-    single = not isinstance(results, list)
-    if single:
-        results = [results]
-
-    batch = [
-        _to_instances(r, confidence_thresh=confidence_thresh) for r in results
-    ]
-
-    if single:
-        return batch[0]
-
-    return batch
-
-
-def _to_instances(result, confidence_thresh=None):
-    if result.masks is None:
-        return None
-
-    classes = np.rint(result.boxes.cls.detach().cpu().numpy()).astype(int)
-    boxes = result.boxes.xywhn.detach().cpu().numpy().astype(float)
-    masks = result.masks.data.detach().cpu().numpy() > 0.5
-    confs = result.boxes.conf.detach().cpu().numpy().astype(float)
-    track_ids = _extract_track_ids(result)
-
-    # convert from center coords to corner coords
-    boxes[:, 0] -= boxes[:, 2] / 2.0
-    boxes[:, 1] -= boxes[:, 3] / 2.0
-
-    detections = []
-    for cls, box, mask, conf, idx in zip(
-        classes, boxes, masks, confs, track_ids
-    ):
-        if confidence_thresh is not None and conf < confidence_thresh:
-            continue
-
-        label = result.names[cls]
-        w, h = mask.shape
-        tmp = np.copy(box)
-        tmp[2] += tmp[0]
-        tmp[3] += tmp[1]
-        tmp[0] *= h
-        tmp[2] *= h
-        tmp[1] *= w
-        tmp[3] *= w
-        tmp = [int(b) for b in tmp]
-        y0, x0, y1, x1 = tmp
-        sub_mask = mask[x0:x1, y0:y1]
-
-        detection = fol.Detection(
-            label=label,
-            bounding_box=list(box),
-            mask=sub_mask.astype(bool),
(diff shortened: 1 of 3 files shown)


fiftyone/utils/ultralytics.py
```diff
@@ -364,104 +364,75 @@ def _to_keypoints(result, confidence_thresh=None):
...
-        raise NotImplementedError(
-            "Subclass must implement _format_predictions()"
-        )
-
-    def predict(self, arg):
-        image = Image.fromarray(arg)
-        predictions = self.model(image, verbose=False)
-        return self._format_predictions(predictions[0])
+    def _forward_pass(self, imgs):
+        return self._model(imgs, verbose=False)
```
jacobsela, line 414: Please make sure that the `__call__` method of `YOLO` models is actually just the forward pass. I'm not sure that this is the case. This may include preprocessing which should go under `transforms` or post processing that should go under `postprocess`.

jacobsela (changes requested):
Good starting point.

Please make sure that the key parts of the `TorchImageModel` contract are maintained. I've written some details for how + why in the comments.

Please let me know if what I wrote makes sense to you or if you want to go over the contracts of the various moving parts, I'm more than happy to help (and how we feel about these contracts will motivate if/how we maintain them going forward).

jacobsela:
Please note that the tests described in the description are not informative of the correctness of the change.

> Load and apply each of the following models:
>  FiftyOneYOLOClassificationModel
>  FiftyOneYOLODetectionModel
>  FiftyOneYOLOSegmentationModel
>  FiftyOneYOLOOBBModel
>  FiftyOneYOLOPoseModel
> import fiftyone as fo
> import fiftyone.zoo as foz
> 
> dataset = fo.load_dataset("quickstart")
> 
> model_name = <MODEL_ZOO_NAME>
> zoo_model = foz.load_zoo_model(
>     model_name,
> )
> dataset.apply_model(zoo_model, label_field="yolo-torch-test")
> 
> from ultralytics import YOLO
> model = YOLO("<MODEL_NAME>.pt")
> dataset.apply_model(model, label_field="yolo-direct-test")
> Run test/intensive/model_zoo_tests.py

1. `test/intensive/model_zoo_tests.py` checks if the models run, but doesn't have a correctness check.
2. The "direct check" doesn't actually apply the model directly. `apply_model` will convert the model to the equivalent FO format before inference.

fiftyone/utils/ultralytics.py
```diff
@@ -364,104 +364,75 @@ def _to_keypoints(result, confidence_thresh=None):
...
-    @property
-    def media_type(self):
-        return "image"
-
-    @property
-    def ragged_batches(self):
-        return False
-
-    @property
-    def transforms(self):
```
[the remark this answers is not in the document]
manushreegangwar: The model needs `PIL -> tensor` conversion and also image resizing for it to work. How are you proposing to pass those if not through `transforms`?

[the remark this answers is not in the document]
manushreegangwar: Okay, I see what you are suggesting. I will update this.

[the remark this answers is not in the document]
brimoor: For context, when you pass a `TorchImageModel` to `apply_model()`, it will use a data loader and it will pass `model.transforms` to that data loader:
https://github.com/voxel51/fiftyone/blob/580800dbf20cae088145a3d81921c90f37df7211/fiftyone/core/models.py#L699

If you implement a subclass of `TorchImageModel` that doesn't implement `transforms` or any of the config parameters that define it, then the default value of `transforms` is built like this:
https://github.com/voxel51/fiftyone/blob/580800dbf20cae088145a3d81921c90f37df7211/fiftyone/utils/torch.py#L807-L860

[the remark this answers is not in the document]
manushreegangwar: When overriding with ultralytics `preprocess` transforms https://github.com/voxel51/fiftyone/blob/bbc0ec4657b6c22e13460fef2e8d097824c2ef1d/fiftyone/utils/ultralytics.py#L437-L447, [pre_transform](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/predictor.py#L173) expects a list but `TorchImageDataset.__getitem__` operates on an individual image. Is there a way to apply this transform using the dataloader?

fiftyone/utils/ultralytics.py
```diff
@@ -419,223 +414,123 @@ def _load_model(self, config):
             else:
                 model.set_classes(config.classes)
 
+        if not model.predictor:
```
manushreegangwar, line 417: This needs to be done to call ultralytics `preprocess` as a transform in the dataloader. 

Since setup, preprocess, inference and postprocess are in https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/predictor.py#L274, preprocess is still called in `stream_inference` in ultralytics but with tensor inputs, it only calls does [this](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/predictor.py#L158-L159).

In order to remove `preprocess` from `stream_inference`, our forward pass would end up looking like:

```
def _forward_pass(self, imgs):
        # Setup source every time predict is called
        print("imgs ", imgs.size())
        self._model.predictor.setup_source(imgs)

        if not self._model.predictor.done_warmup:
            self._model.predictor.model.warmup(
                imgsz=(
                    1
                    if self._model.predictor.model.pt
                    or self._model.predictor.model.triton
                    else self._model.predictor.dataset.bs,
                    3,
                    *self._model.predictor.imgsz,
                )
            )
            self.done_warmup = True

        (
            self._model.predictor.seen,
            self._model.predictor.windows,
            self._model.predictor.batch,
        ) = 0, [], None

        self._model.predictor.run_callbacks("on_predict_start")
        for self._model.predictor.batch in self._model.predictor.dataset:
            self._model.predictor.run_callbacks("on_predict_batch_start")
            _, im0s, _ = self._model.predictor.batch
            preds = self._model.predictor.inference(im0s)

            # Postprocess
            results = self._model.predictor.postprocess(preds, im0s, im0s)
            self._model.predictor.run_callbacks("on_predict_postprocess_end")

            self._model.predictor.run_callbacks("on_predict_batch_end")
            return list(results)
```

fiftyone/utils/ultralytics.py
```diff
@@ -364,50 +168,66 @@ def _to_keypoints(result, confidence_thresh=None):
...
-        self.model.to(self.device)
+        super().__init__(config)
+
+    def _download_model(self, config):
+        config.download_model_if_necessary()
+
+    def _set_predictor(self, config, model):
+        custom = {
+            "conf": config.confidence_thresh,
+            "batch": 1,
```
jacobsela, line 195: Does this force the internal batch size to be 1? Does this work with higher batch sizes?

fiftyone/utils/ultralytics.py
```diff
@@ -364,50 +168,66 @@ def _to_keypoints(result, confidence_thresh=None):
...
+            "verbose": False,
+            "device": self._device,
+        }
+        args = {**custom, **model.overrides}
+        model.predictor = model.task_map[model.task]["predictor"](
+            overrides=args,
+            _callbacks=model.callbacks,
+        )
+        model.predictor.imgsz = (
+            config.image_size if config.image_size else (640, 640)
```
jacobsela, line 208: Where is the size 640 x 640 coming from?

fiftyone/utils/ultralytics.py
```diff
@@ -364,50 +168,66 @@ def _to_keypoints(result, confidence_thresh=None):
...
+        )
+        model.predictor.setup_model(model=model.model, verbose=False)
+        return model
 
     def _load_model(self, config):
```
jacobsela, line 213: Wouldn't it be better to use `super()._load_model` and just overwrite `entry point_fcn`?

fiftyone/utils/ultralytics.py
```diff
@@ -419,223 +239,228 @@ def _load_model(self, config):
...
+            return config.classes
+        if isinstance(self._model.names, dict):
+            return list(self._model.names.values())
         return None
 
-    @property
-    def preprocess(self):
-        return False
+    def _forward_pass(self, imgs):
+        self._model.predictor.setup_source(imgs)
```
jacobsela, line 258: Are lines 258 and 259 necessary? `imgs` are passed in line 260. Do these two lines somehow change relevant things as a side effect? It seems to me that line 260 leads to [this call](https://github.com/ultralytics/ultralytics/blob/3f7b8f65ebe657002e73ae162de7d822f1f3c15c/ultralytics/nn/autobackend.py#L570), is that right? If so, it seems that imgs are just passed through to the model.

I haven't looked closely enough to know for sure. Please let me know if I'm missing something.

fiftyone/utils/ultralytics.py
```diff
@@ -419,223 +239,228 @@ def _load_model(self, config):
...
+        # Taken from ultralytics.engine.predictor.preprocess.
+        img = np.stack(self._model.predictor.pre_transform(img))
+        img = img.transpose((0, 3, 1, 2))
+        img = np.ascontiguousarray(img)
+        img = torch.from_numpy(img)
+        img = img.half() if self._model.predictor.model.fp16 else img.float()
+        img /= 255
+        return torch.squeeze(img, axis=0)
+
+    def _build_output_processor(self, config):
```
jacobsela, line 295: why not user super here?

fiftyone/utils/ultralytics.py
```diff
@@ -419,223 +239,228 @@ def _load_model(self, config):
...
-        super().__init__(config)
-
-    def _format_predictions(self, predictions):
-        return to_detections(predictions)
-
-    def predict_all(self, args):
-        images = [Image.fromarray(arg) for arg in args]
-        predictions = self.model(images, verbose=False)
-        return self._format_predictions(predictions)
+    def _predict_all(self, imgs):
```
jacobsela, line 314: Not a PR comment but something to discuss at some point:
I ended up doing something similar for HF. It's obviously not great that we separately wrote similar code in different places, but it may be inevitable given our current constraints. Making `SamplesMixin` be a standard part of models should help with this. We should discuss this for our next steps with model refactoring.

fiftyone/utils/ultralytics.py
```diff
@@ -419,223 +239,228 @@ def _load_model(self, config):
...
+    return height, width
 
-    def _format_predictions(self, predictions):
-        return to_instances(predictions)
 
+class FiftyOneRTDETRModelConfig(FiftyOneYOLOModelConfig):
+    """Configuration for a :class:`FiftyOneRTDETRModel`."""
 
-class FiftyOneYOLOPoseModelConfig(FiftyOneYOLOModelConfig):
     pass
```
jacobsela, line 383: perfect.

jacobsela: Jokes aside, the fact that the code all gets consolidated to once class and works across the board is great. This was kind of the point of it all (outside just needing it to be a `TorchImageModel`). Stuff like this makes it a whole lot easier to maintain the code longterm. Great job.

fiftyone/utils/ultralytics.py
```diff
@@ -419,223 +239,228 @@ def _load_model(self, config):
...
+        return {
+            "img": self._model.predictor.preprocess(img),
+            "orig_img": orig_img,
+        }
 
 
 def _convert_yolo_classification_model(model):
-    config = FiftyOneYOLOClassificationModelConfig({"model": model})
-    return FiftyOneYOLOClassificationModel(config)
+    config = FiftyOneYOLOModelConfig(
```
jacobsela, line 412: Again, super clean. Sign of a job well done.

fiftyone/utils/ultralytics.py
```diff
@@ -667,3 +492,248 @@ def _to_detections(self, df):
...
+                pred = {
+                    "boxes": result.boxes.xyxy,
+                    "labels": result.boxes.cls.int(),
+                    "scores": result.boxes.conf,
+                    "track_ids": _extract_track_ids(result),
+                }
+                batch.append(pred)
+        return batch
+
+    def _parse_output(self, output, frame_size, confidence_thresh):
```
jacobsela, line 588: can this use super?

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -3253,8 +3253,15 @@
...
-                "type": "fiftyone.utils.ultralytics.FiftyOneYOLODetectionModel",
-                "config": {}
+                "type": "fiftyone.utils.ultralytics.FiftyOneYOLOModel",
+                "config": {
+                    "entrypoint_fcn": "ultralytics.YOLO",
+                    "entrypoint_args": {
+                        "model": "yolov8n.pt"
+                    },
+                    "output_processor_cls": "fiftyone.utils.ultralytics.UltralyticsDetectionOutputProcessor",
+                    "image_size": [640, 640]
```
jacobsela, line 3263: Why hardcode the `image_size` parameter here? And again, where does 640 x 640 come from?

fiftyone/utils/ultralytics.py
```diff
@@ -667,3 +492,248 @@ def _to_detections(self, df):
...
+    """Converts Ultralytics Oriented Bounding Box model outputs to FiftyOne format."""
+
+    def __init__(
+        self,
+        classes=None,
+        post_processor=None,
+    ):
+        super().__init__(classes)
+        self.post_processor = post_processor
+
```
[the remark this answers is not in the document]
jacobsela: This is a good point, would be nice to share the test code so that we can be aligned on expectations.

jacobsela (commented):
From a code cleanliness standpoint this is good. There are some comments I left, but I don't want to block this because of them (though they should be done, just maybe not rn given the time crunch).

I have a few concerns (further notes for them in the code):
1. separation between `transforms` and `_forward_pass`. I am not familiar enough with the Ultralytics code to be able to say for sure if the way it's done is right or not. Please make sure it is the case. It looks to me like there may be some unnecessary stuff in `_forward_pass`, or potentially things that should go in `transforms`.
2. batched inference - the predictor batch size is set to 1. What does this affect?

I saw you mentioned some gaps between direct YOLO inference and using the integration (on the order of magnitude of 1e-5). Some gaps are fine, OOM 1e-5 is definitely fine IMO. I wouldn't worry about it.

If the above concerns are good this is good to go IMO, ping me and I'll approve.

fiftyone/utils/ultralytics.py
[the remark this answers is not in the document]
jacobsela: @manushreegangwar resolving this because it's a huge message from code rabbit. I understand you very thoroughly tested these models, this is great, please share code when you can.

fiftyone/utils/ultralytics.py
manushreegangwar: There may be a workaround. `self._model.predictor.batch` is used under the hood in ultralytics but I can potentially move this out of here if all the variables are set as needed.

fiftyone/utils/ultralytics.py
manushreegangwar: Yes, I agree. We can also refactor `_predict_all` to make it more configurable without having to copy everything over and modifying.

fiftyone/utils/ultralytics.py
manushreegangwar: It is possible but I will have to loop over all the detections to add track ids. I would prefer to refactor the super class later to make this more seamless.

fiftyone/utils/ultralytics.py
manushreegangwar: https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/model.py#L536 is how ultralytics handles this.

fiftyone/utils/ultralytics.py
manushreegangwar: I also need some special handling for `kwargs = config.entrypoint_args` to avoid downloading the yolo model in current working directory (line 227)

fiftyone/utils/ultralytics.py
jacobsela: wow that's awful. ok so nit, make that some constant somewhere, but really this is more on them than us.

fiftyone/utils/ultralytics.py
jacobsela: those can be passed in the manifest or handled in the init/config. I think it's important that we consolidate as much as possible, which is why I'm being so pedantic about this, but not blocking.

fiftyone/utils/ultralytics.py
jacobsela: Like the other case, this can be done in the init by adding a new argument to `config.output_processor_args`.

fiftyone/utils/ultralytics.py
jacobsela: Absolutely. I wish we had fully done the `TorchImageModel` --> `TorchModel` refactor before this, what you wrote is one of the main points of that refactor.

fiftyone/utils/ultralytics.py
jacobsela: Hmm, I lean towards clarity + maintainability over efficiency in this case (especially because longterm we want to offload post processing from the main compute pathway), but I suppose it's fine for now.

fiftyone/utils/ultralytics.py
manushreegangwar: Done. We should revisit the super method so it doesn't require us to modify the config.

fiftyone/utils/ultralytics.py
manushreegangwar: Updated this to use super

fiftyone/utils/ultralytics.py
```diff
@@ -237,85 +115,11 @@ def _obb_to_polylines(result, filled, confidence_thresh=None):
     return fol.Polylines(polylines=polylines)
 
 
-def to_polylines(results, confidence_thresh=None, tolerance=2, filled=True):
```
brimoor, line 240: Public utility methods need to stay! For example, here's a case in our docs where `to_detection()` usage is documented:
https://docs.voxel51.com/integrations/ultralytics.html#object-detection

It's still valid to use these utilities if a user is for some reason directly working with an `ultralytics.YOLO` model and wants to convert its output to the appropriate FiftyOne types.

fiftyone/utils/ultralytics.py
```diff
@@ -7,20 +7,20 @@
...
 from PIL import Image
 
-from fiftyone.core.config import Config
 import fiftyone.core.labels as fol
-from fiftyone.core.models import Model
 import fiftyone.utils.torch as fout
 import fiftyone.core.utils as fou
 import fiftyone.zoo.models as fozm
+import eta.core.utils as etau
+import itertools
```
brimoor, line 18: Our style guide has rules about how imports should be ordered:
https://github.com/voxel51/fiftyone/blob/develop/STYLE_GUIDE.md#imports

Please follow them 😄 

In this case, `itertools` belongs back in its own block where it previously was in this PR, and `etau` belongs in its own block in between `PIL` and `fol`.

fiftyone/utils/ultralytics.py
```diff
@@ -667,3 +475,226 @@ def _to_detections(self, df):
...
+                    "labels": result.boxes.cls.int(),
+                    "scores": result.boxes.conf,
+                    "masks": result.masks.data,
+                    "track_ids": _extract_track_ids(result),
+                }
+                batch.append(pred)
+        return batch
+
+    def _parse_output(self, output, _, confidence_thresh):
+        boxes = output["boxes"].detach().cpu().numpy().astype(float)
```
brimoor, line 617: I believe this logic should go back into a `to_detections()` utility

manushreegangwar: Moved into `to_instances`

fiftyone/utils/ultralytics.py
manushreegangwar: Done

fiftyone/utils/ultralytics.py
manushreegangwar: Restored these functions.

fiftyone/utils/ultralytics.py
manushreegangwar: Done

fiftyone/utils/ultralytics.py
```diff
@@ -12,15 +12,14 @@
...
-from fiftyone.core.config import Config
 import fiftyone.core.labels as fol
-from fiftyone.core.models import Model
 import fiftyone.utils.torch as fout
 import fiftyone.core.utils as fou
 import fiftyone.zoo.models as fozm
 
 ultralytics = fou.lazy_import("ultralytics")
 torch = fou.lazy_import("torch")
+torchvision = fou.lazy_import("torchvision")
```
jacobsela, line 22: Can you add `fou.ensure_torch()` and `fou.ensure_package("transformers")` before the imports?

fiftyone/utils/ultralytics.py
```diff
@@ -419,223 +429,226 @@ def _load_model(self, config):
...
+
+        if isinstance(imgs, list) and len(imgs) and isinstance(imgs[0], dict):
+            orig_images = [img.get("orig_img") for img in imgs]
+            images = [img.get("img") for img in imgs]
+        else:
+            orig_images = imgs
+            images = imgs
+
+        height, width = None, None
+        if self.config.raw_inputs:
```
jacobsela, line 500: Shouldn't this always be true? If so let's remove the if-else

fiftyone/utils/ultralytics.py
```diff
@@ -667,3 +680,162 @@ def _to_detections(self, df):
...
+
+
+class UltralyticsPostProcessor(object):
+    # pylint: disable=not-callable
+    post_processor = None
+
+    def post_process(self, output):
+        imgs = output.get("imgs", None)
+        orig_imgs = output.get("orig_imgs", None)
+        if isinstance(orig_imgs, torch.Tensor):
```
jacobsela, line 692: Not blocking: are all of these options possible outputs from YOLO models? if not let's remove the branches we can.

jacobsela (approved):
LGTM

fiftyone/utils/ultralytics.py
brimoor: This module already imports our torch utils on line 16:

```py
```

which calls `fou.ensure_torch()` internally. So in fact we won't get to line 21 unless `torch` + `torchvision` are installed. So technically these lazy imports aren't actually adding any value as written:

```py
torch = fou.lazy_import("torch")
torchvision = fou.lazy_import("torchvision")
```

brimoor: I'm not requesting any change to this module, just providing context for why things like `ensure_torch()` and `lazy_import()` exist, and when to use them.

fiftyone/utils/ultralytics.py
```diff
@@ -667,3 +680,162 @@ def _to_detections(self, df):
...
+):
+    """Converts Ultralytics Classification model outputs to FiftyOne format."""
+
+    def __init__(self, classes=None, store_logits=False, post_processor=None):
+        super().__init__(classes=classes, store_logits=store_logits)
+        self.post_processor = post_processor
+
+    def __call__(self, output, frame_size, confidence_thresh=None):
+        results = self.post_process(output)
+        classifications = []
```
brimoor, line 724: Exposing this as a `to_classification()` utility and then calling it here would be consistent with the design of the rest of the module. We have `to_detections()`, `to_instances()`, `to_polylines()`, etc.

fiftyone/utils/ultralytics.py
```diff
@@ -667,3 +680,162 @@ def _to_detections(self, df):
...
+        self.post_processor = post_processor
+
+    def __call__(self, output, frame_size, confidence_thresh=None):
+        results = self.post_process(output)
+        preds = self._to_dict(results)
+        return super().__call__(preds, frame_size, confidence_thresh)
+
+    def _to_dict(self, results):
+        batch = []
+        for result in results:
```
brimoor, line 763: Calling `to_detections()` here rather than reimplementing it would be consistent with the rest of the module.

fiftyone/utils/ultralytics.py
brimoor: > Can you add `fou.ensure_package("transformers")` ...

@jacobsela I believe you mean `fou.ensure_package("ultralytics")`. That would have the effect of making the `fiftyone.utils.ultralytics` module unimportable unless the user actually has the `ultralytics` package installed. Whereas currently the module will be importable but if the user actually tries to use an Ultralytics model without the package installed, *then* they will get the error.

fiftyone/utils/ultralytics.py
manushreegangwar: Do you mean that the images will always be a tensor at this point? I will prune this in the collate pr.

fiftyone/utils/ultralytics.py
manushreegangwar: When directly using `predict`, all of these are possible. I will see if some of this can be removed before we get to post-processing.

fiftyone/utils/ultralytics.py
jacobsela: Thanks Brian, good to know

fiftyone/utils/ultralytics.py
```diff
@@ -364,50 +409,76 @@ def _to_keypoints(result, confidence_thresh=None):
...
+
+            if etau.is_str(entrypoint_fcn):
+                entrypoint_fcn = etau.get_function(entrypoint_fcn)
+
+            kwargs = config.entrypoint_args or {}
+            model = entrypoint_fcn(**kwargs)
+
+        model = model.to(self._device)
+        if self.using_half_precision:
+            model = model.half()
```
manushreegangwar, line 481: Re-introduced this because when calling `model.eval()` with older version of ultralytics [this happens](https://stackoverflow.com/questions/79067001/model-eval-on-pretrained-yolov8-causes-dowloading-of-datasets)

fiftyone/utils/ultralytics.py
manushreegangwar: Leaving this as is so users get an error when using ultralytics integration without ultralytics installed.

fiftyone/utils/ultralytics.py
manushreegangwar: Done

manushreegangwar:
@brimoor I have addressed your comments. Please take a look. Thanks!

brimoor (approved):
I'm approving now because the specific things I commented on in my review have been addressed.

@manushreegangwar @jacobsela I'm trusting that you'll verify that all of the example code in our Ultralytics integration continues to work as written before merging 😄 
https://docs.voxel51.com/integrations/ultralytics.html

If that is the case --- and the model outputs look correct obviously! --- then this PR is good to go from my seat.

brimoor:
@manushreegangwar please note that a `release/v1.6.0` branch already exists, and I believe you are intending this functionality to be used in that release, so you'll need to do one of two things:
1. Update this PR by rebasing its commits onto the `release/v1.6.0` branch
2. Merge this PR into `develop` and then create a new PR where you cherry-pick these commits onto `release/v1.6.0`

manushreegangwar:
> @manushreegangwar please note that a `release/v1.6.0` branch already exists, and I believe you are intending this functionality to be used in that release, so you'll need to do one of two things:
> 
> 1. Update this PR by rebasing its commits onto the `release/v1.6.0` branch
> 2. Merge this PR into `develop` and then create a new PR where you cherry-pick these commits onto `release/v1.6.0`

Going with option 2, same as  https://github.com/voxel51/fiftyone/pull/5834. 

Verified that https://docs.voxel51.com/integrations/ultralytics.html works the same as develop.

PR #5730, opened 2025-04-15
Optimize `split_labels()` and `delete_labels()`

brimoor:
## Change log

- optimizes `split_labels()` and `delete_labels(view=view)` syntaxes by using optimized per-sample update operations that can leverage the `_id` index rather than requiring full collection scans
- fixed a "BSON too large" error that would previously occur when deleting a sufficiently long list of IDs via `delete_labels(ids=ids)`

## Tested by

- unit tests added

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of label merging and deletion, especially for frame fields and views, ensuring more accurate and robust label management.
- **Tests**
  - Added new tests to verify correct behavior of label splitting and deletion for both sample-level and frame-level label fields.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -3979,11 +3979,12 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         if labels is not None:
             self._delete_labels(labels, fields=fields)
 
-        if ids is None and tags is None and view is None:
-            return
+        if view is not None:
+            labels = view._get_selected_labels(fields=fields)
+            self._delete_labels(labels, fields=fields)
 
-        if view is not None and view._dataset is not self:
-            raise ValueError("`view` must be a view into the same dataset")
+        if ids is None and tags is None:
+            return
 
         if etau.is_str(ids):
             ids = [ids]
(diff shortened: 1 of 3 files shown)


fiftyone/core/dataset.py
```diff
@@ -4005,41 +4006,29 @@ def delete_labels(
 
         now = datetime.utcnow()
 
+        batch_size = fou.recommend_batch_size_for_value(
```
ritch, line 4009: nit: this exact usage is now duplicated in 6 places

ritch (approved):
LGTM

fiftyone/core/dataset.py
brimoor: Right, we can chat offline about it if you want. The semantics are a bit subtle.

The purpose of the function is to ensure that the `UpdateMany` ops that this function generates aren't too big on a couple dimensions:
- they don't exceed the allowed JSON payload size dictated by `fo.config.batcher_target_size_bytes`
- we don't generate an `{$in: ids}` for which `len(ids) > 100000`, because performance of `$in` degrades beyond that

Both of these conditions depend on the nature of the actual `UpdateMany` operations being constructed, so I prefer to be explicit in calling `recommend_batch_size_for_value()` rather than just having a magic constant somewhere (besides this value does depend on the current value of `fo.config.batcher_target_size_bytes`).

In practice, most `UpdateMany` operations we construct look roughly the same, so all `recommend_batch_size_for_value()` calls end up looking the same too. But if we refactored the `UpdateMany` operations in this function, it could change the recommended `batch_size` for this function without changing the recommended batch size for other operations.

PR #5734, opened 2025-04-16
update doc string for "none" compression

CamronStaley:

Update string to be "none" for no compressor used instead of an empty string which would be ignored and default back to lz4.

n/a

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Documentation**
  - Updated documentation to specify that the value to disable API compression is now `none` instead of an empty string. The list of accepted values has been revised accordingly.


swheaton (approved):
🚢 🇮🇹

PR #5739, opened 2025-04-17
Optimize values() mongodb call

swheaton:
This is the pipeline for a few cases of values() calls.
```
>> ds.values('filepath')
# pipeline
{'value0': {'$cond': {'if': {'$gt': ['$filepath', None]}, 'then': '$filepath', 'else': None}}}

# sample output
{'_id': ObjectId('67608096daa9bc771cbd3dcc'), 'value0': '/Users/stuart/fiftyone/quickstart/data/000880.jpg'}
...

>> ds.values(['filepath', 'id'])
# pipeline
{'value0': {'$cond': {'if': {'$gt': ['$filepath', None]}, 'then': '$filepath', 'else': None}}, 'value1': {'$cond': {'if': {'$gt': ['$_id', None]}, 'then': {'$toString': '$_id'}, 'else': None}}}

# sample output
{'_id': ObjectId('67608096daa9bc771cbd3dcc'), 'value0': '/Users/stuart/fiftyone/quickstart/data/000880.jpg', 'value1': '67608096daa9bc771cbd3dcc'}
...
```

Anything else (passing expr instead of field, unwinding a list with `[]`, is not considered "big batchable" and so is not the target of this limited PR.

### What/Why
If we add `"_id": False` to the projection, it will omit the object ID which we don't use. In the single scalar field values case, the amount of data transferred is cut almost in half. The bigger the field, that advantage shrinks. However this is a very prevalent use case.

It does not make a material difference when the database is local. But it won't hurt it either.
It does give a performance improvement (~10-20% speedup) on some datasets when the database is remote or proxied.

Note it does not improve for all datasets and fields, because if the samples are particularly dense, the mongodb load may completely overshadow the network transfer aspect. Or if the size of _id relative to the "real" data is small, it does not make a big difference.

### Change Safety
The change is on the generic aggregate() function however it's safe because:
1. It's only for big batchable pipelines
2. Only `Values` aggregation supports this (`_is_big_batchable`)
3. `Values.parse_result()` only uses its `_big_field` which is usually `value0`, `value1`, and so on.

Since the _id field is proven to be unused, there is no way this change can make things worse.

```python
ds = fo.load_dataset("open-images-v7-full") # 1.9M samples

with etau.Timer() as t:
  _=ds.values('filepath')
```
This is also machine and network dependent, but on my particular setup with direct remote database, I got an average of 58s, down to an average 50s.
With a proxied remote database, average of 62s goes down to average 55s

        notes for FiftyOne users.

Improved `values(scalar_field)` call performance 10-20% over a low-speed MongoDB connection.

-   [ ] Other

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -10415,6 +10415,7 @@ class SampleCollection(object):
                     "Batchable aggregations must have pipelines with a single "
                     "$project stage; found %s" % _pipeline
                 )
+        project["_id"] = False
 
         return self._pipeline(
             pipeline=[{"$project": project}],


CamronStaley (approved):
seems legit

kaixi-wang (changes requested):
i don't think the explanation provided explains the change in performance because passing True explicitly can sometimes improve performance as well, which makes me think something else is happening (perhaps index lookup related or something)

CamronStaley:
adding my results with compression + lz4 `ds.values('filepath')`:

With these changes
```
(TEAMS) (3.10.14) ➜  scripts python benchmark.py -d malaria-cell-images -n 1000000 -r 3 -a retrieve_filepaths
Repetition 1
------------------------- Retrieve values (filepaths) --------------------------
Running...
Done in 17.412230968475342 seconds.

ACTION                       DATASET                CONNECTION_TYPE      VERSION  SAMPLE COUNT  TOTAL ELAPSED  ELAPSED PER SAMPLE
=================================================================================================================================
Retrieve values (filepaths)  malaria-cell-images            api_uri  2.7.0.dev29       1000000        17.4122              0.0000
Repetition 2
------------------------- Retrieve values (filepaths) --------------------------
Running...
Done in 17.00377106666565 seconds.

ACTION                       DATASET                CONNECTION_TYPE      VERSION  SAMPLE COUNT  TOTAL ELAPSED  ELAPSED PER SAMPLE
=================================================================================================================================
Retrieve values (filepaths)  malaria-cell-images            api_uri  2.7.0.dev29       1000000        17.0038              0.0000
Repetition 3
------------------------- Retrieve values (filepaths) --------------------------
Running...
Done in 16.805859088897705 seconds.

ACTION                       DATASET                CONNECTION_TYPE      VERSION  SAMPLE COUNT  TOTAL ELAPSED  ELAPSED PER SAMPLE
=================================================================================================================================
Retrieve values (filepaths)  malaria-cell-images            api_uri  2.7.0.dev29       1000000        16.8059              0.0000
```

without these changes:
```
(TEAMS) (3.10.14) ➜  scripts python benchmark.py -d malaria-cell-images -n 1000000 -r 3 -a retrieve_filepaths
Repetition 1
------------------------- Retrieve values (filepaths) --------------------------
Running...
Done in 21.160054922103882 seconds.

ACTION                       DATASET                CONNECTION_TYPE      VERSION  SAMPLE COUNT  TOTAL ELAPSED  ELAPSED PER SAMPLE
=================================================================================================================================
Retrieve values (filepaths)  malaria-cell-images            api_uri  2.7.0.dev29       1000000        21.1601              0.0000
Repetition 2
------------------------- Retrieve values (filepaths) --------------------------
Running...
Done in 21.426618814468384 seconds.

ACTION                       DATASET                CONNECTION_TYPE      VERSION  SAMPLE COUNT  TOTAL ELAPSED  ELAPSED PER SAMPLE
=================================================================================================================================
Retrieve values (filepaths)  malaria-cell-images            api_uri  2.7.0.dev29       1000000        21.4266              0.0000
Repetition 3
------------------------- Retrieve values (filepaths) --------------------------
Running...
Done in 21.77501916885376 seconds.

ACTION                       DATASET                CONNECTION_TYPE      VERSION  SAMPLE COUNT  TOTAL ELAPSED  ELAPSED PER SAMPLE
=================================================================================================================================
Retrieve values (filepaths)  malaria-cell-images            api_uri  2.7.0.dev29       1000000        21.7750              0.0000
(TEAMS) (3.10.14) ➜  scripts
```

swheaton:
Ok I'll try it with non indexed scalar fields

fiftyone/core/collections.py
```diff
@@ -10415,6 +10415,7 @@ def _build_batch_pipeline(self, aggs_map):
                     "Batchable aggregations must have pipelines with a single "
                     "$project stage; found %s" % _pipeline
                 )
+        project["_id"] = False
```
brimoor, line 10418: I agree with this change!

For completeness, we should also add this idea to `Values.to_mongo()` class so that we also omit `_id` from any `Values()` aggregations that are "big" (`agg._has_big_result == True`) but not "batchable" (`agg._is_big_batchable == False`)

brimoor: Also, since we're engaging in optimizations like this, we should go ahead and optimize `values("id")` as proposed in https://github.com/voxel51/fiftyone/pull/5479, because that's the most common `values()` operation!

swheaton: Ok I was 100% sure about big batchable so I started there but I didn't look into the other cases. We can do that.

And yeah let's do it, I'll get to that one ASAP

brimoor: @swheaton I already had most of the desired things in https://github.com/voxel51/fiftyone/pull/5479, so I went ahead and added this idea to that PR. It's ready for review now

swheaton: Ok. Still testing stuff.

swheaton:
@kaixi-wang you're right, this is super weird. It's a particular behavior of MongoDB, with no clear explanation from the explain plan.

But what I've found boils down to the cardinality of the field.
Low cardinality or complex/embedded fields?: `_id: [False|True]` can make it MUCH slower
High cardinality: `_id: False` is 10-20% faster.

I'm thinking that we use this only for filepath and id/_id. That's really the only cases we can guarantee it will at least be no worse than before. Which will tie in with @brimoor's #5479

brimoor (changes requested):
This proposal has been generalized and incorporated into #5479

swheaton:
Close for https://github.com/voxel51/fiftyone/pull/5479

PR #5743, opened 2025-04-18
Optimize retrieving indexed field values

kaixi-wang:

- force mongo to use a covered index query when possible to improve speed, particularly on large datasets

- context/background: aggregation projections can never be a fully covered query and will always involve a collection scan. By using find + hint, we force the lookup on the index only, vastly improving performance

- We can build upon this iteratively and expand the supported fields, but this will reduce the bottleneck in a lot of existing workflows. 

- added unit tests
- tested by running locally:

### values() optimization
```
In [16]: %%time
    ...: values=ds.values("_id")
    ...: 
    ...: 
CPU times: user 17.5 s, sys: 4.88 s, total: 22.4 s
Wall time: 7min 3s

In [17]: %%time
    ...: values=ds.values("_id", _enforce_natural_order=False)
    ...: 
    ...: 
CPU times: user 8.07 s, sys: 1.68 s, total: 9.75 s
Wall time: 19.7 s

```

```
In [19]: ds=fo.load_dataset('big-image-ds')

In [20]: %%time
    ...: vals=ds.values("filepath")
    ...: 
    ...: 
CPU times: user 3min 24s, sys: 1min 20s, total: 4min 45s
Wall time: 2h 9min 14s

In [21]: len(vals)
Out[21]: 32877680


In [25]: %%time
    ...: vals=ds.values("filepath", _enforce_natural_order=False)
    ...: 
    ...: 
CPU times: user 1min 24s, sys: 10.3 s, total: 1min 34s
Wall time: 4min 32s

In [26]: len(vals)
Out[26]: 32877680

```

### Optimize getting values of a compound index
```
In [14]: %%time
    ...: mapping=get_indexed_values(ds, "_id_1_filepath_1")
    ...: 
    ...: 
CPU times: user 9.57 s, sys: 1.86 s, total: 11.4 s
Wall time: 45.4 s

In [15]: %%time
    ...: values=ds.values(["_id","filepath"])
    ...: 
    ...: 
CPU times: user 18.6 s, sys: 3.36 s, total: 22 s
Wall time: 10min 38s
```

        notes for FiftyOne users.

Add a new function to retrieve values from indexed fields fast. 
Enable using this optimization on full datasets for values() calls by passing `_enforce_natural_order=False`

-   [ ] Other

## Summary by CodeRabbit

- **Performance Improvements**
  - Optimized retrieval of indexed field values for full datasets, speeding up queries on fields like `id`, `_id`, and `filepath`.
- **New Features**
  - Introduced efficient methods to fetch values from indexed fields, improving query performance.
  - Exposed new indexed value retrieval functions in the public API.
- **Bug Fixes**
  - Enhanced error handling with clear errors for queries on non-indexed or nonexistent fields.
- **Tests**
  - Added tests validating indexed value retrieval behavior and error handling.


swheaton (changes requested):
Wow I'm impressed! Mongo DB can be so silly sometimes. How could it not decide that that find query should use _id index ... Note that you can also force it by putting something dumb in the query slot.

Unfortunately we cannot merge as is. This will totally break the contract (even if unofficial) we have maintained for `values()`. They need to be in the same order. For example in `set_values()`, using this form of `values('id')` will give us a potentially different ordering, causing values to be assigned in the wrong order.

This would have to be part of a different path: _id sorted samples which we have discussed before. It would have to be opt-in and only useable in certain situations.  What could use this initially though, map_samples()? Faster paging than $skip/$limit?

kaixi-wang:
> Wow I'm impressed! Mongo DB can be so silly sometimes. How could it not decide that that find query should use _id index ... Note that you can also force it by putting something dumb in the query slot.
> 
> Unfortunately we cannot merge as is. This will totally break the contract (even if unofficial) we have maintained for `values()`. They need to be in the same order. For example in `set_values()`, using this form of `values('id')` will give us a potentially different ordering, causing values to be assigned in the wrong order.
> 
> This would have to be part of a different path: _id sorted samples which we have discussed before. It would have to be opt-in and only useable in certain situations. What could use this initially though, map_samples()? Faster paging than $skip/$limit?

yeah valid points. They key thing here is that for this level of speedup, we just can't use aggregate() because it will always end in a doc lookup rather than relying entirely on the index

brimoor:
> What could use this initially though, map_samples()?

+100 to this suggestion!

benjaminpkane:
With a compound index, i.e. `(_id, filepath)`, order should be maintained. See https://github.com/voxel51/fiftyone/commit/61e4ed06cc2861f54a7b8d57c723e6e6f3f8d0b5

swheaton:
> With a compound index, i.e. `(_id, filepath)`, order should be maintained. See [61e4ed0](https://github.com/voxel51/fiftyone/commit/61e4ed06cc2861f54a7b8d57c723e6e6f3f8d0b5)

How's that possible? Regular values() without a sort returns in insertion order. If we force it to use the index, it will return it in sorted order.

swheaton:
In that case of the server you're attaching ID to the result so it's fine

swheaton:
I think you are right that values('filepath') should be in _id order if it's using that compound index.
But _id order is not necessarily default/insertion order

benjaminpkane:
Ok, I don't know the use case details. Yes, the compound index ensures the value pairs are a correct. And the results will follow the `_id` order due to the index

swheaton:
> Ok, I don't know the use case details. Yes, the compound index ensures the value pairs are a correct. And the results will follow the `_id` order due to the index

Thanks though. We can probably use it in the future if we are able to relax some constraints or create new use case pathways

fiftyone/core/collections.py
```diff
@@ -9044,6 +9045,32 @@ def values(
...
+                if isinstance(field_or_expr, list):
+                    field = field_or_expr[0]
+                else:
+                    field = field_or_expr
+                return foo.get_indexed_values(
+                    self._dataset, field, values_only=True
+                )
+            except Exception as e:
+                # Fallback to values, but log the error to surface the recommendation to create an index
+                logging.debug(e)
```
minhtuev, line 9072: How about?
```suggestion
                logging.error(e)
```

tests/unittests/odm_tests.py
```diff
@@ -57,3 +57,51 @@ def test_doc_copy_with_new_id(self):
 
         finally:
             dataset_doc.delete()
+
+
+class GetIndexedValuesTests(unittest.TestCase):
```
minhtuev, line 62: nice!

tests/unittests/odm_tests.py
```diff
@@ -57,3 +57,51 @@ def test_doc_copy_with_new_id(self):
...
+            dataset = fo.Dataset()
+
+            samples = [
+                fo.Sample(filepath="sample_%d.jpg" % i) for i in range(10)
+            ]
+            new_ids = dataset.add_samples(samples)
+
+            dataset.add_sample_field("new_field", fo.IntField)
+            dataset.set_values("new_field", [i for i in range(10)])
+            dataset.create_index("new_field")
```
minhtuev, line 74: `create_index` is a blocking call so `new_field` index should exist, but maybe we can add a check in unit test to make sure ~

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,98 @@ def delete_runs(name, dry_run=False):
...
+                )
+            ]
+        return [
+            list(doc.values())
+            for doc in _iter_indexed_values(dataset, field_or_index, pipeline)
+        ]
+    return _iter_indexed_values(dataset, field_or_index, pipeline).to_list()
+
+
+def _iter_indexed_values(dataset, field_or_index, pipeline=None):
```
minhtuev, line 1533: I can see this being potentially a useful public method by itself :)

minhtuev:
Thanks Kacey, I added some small comments, but otherwise LGTM! Defer to Stuart or Brian on this :)

fiftyone/core/odm/database.py
kaixi-wang: yep thats why I split it out, but I didn't think it matched the typical patterns to expose it as a public method.

tests/unittests/odm_tests.py
kaixi-wang: yeah it's blocking. Failing the check would would be the same as failing the test though? I think we can add the check if it starts flaking, but for now, since it doesn't seem to be problematic, no reason to overcomplicate it

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,98 @@ def delete_runs(name, dry_run=False):
...
+    collection = dataset._sample_collection
+    if (
+        (field_or_index != "_id")
+        and (
+            field_or_index
+            not in (default_indexes := dataset._get_default_indexes())
+        )
+        and (
+            field_or_index not in (indexes := dataset.get_index_information())
```
minhtuev, line 1542: We might want to run a couple of scenarios with index creation for large indexes; if an index is under construction and we try to query it, MongoDB might make us wait for index creation to complete. We can use `get_index_information(include_stats=True)` to check if an index is in progress.

kaixi-wang: The block/exclusive lock is only at the start of index creation and at the end (https://www.mongodb.com/docs/manual/core/index-creation/#index-build-process). Otherwise it's in the background and things are slow, but there's nothing we can do about that and it's no slower than calling values(). Also, during the long background index building, include_stats doesn't report it as in progress. To see that you have to query db.currentOp() and check for index creation.

fiftyone/core/collections.py
```diff
@@ -9044,6 +9045,32 @@ def values(
         Returns:
             the list of values
         """
+        # If we do not need to follow insertion order, we can potentially use a covered index query
+        # to get the values directly from the index and avoid a collscan.
+        # In the future, we might not want to specify full_collection,
```
jleven, line 9050:
```suggestion
        # In the future, we might not want to require _is_full_collection,
```

fiftyone/core/odm/database.py
swheaton, line 1542: I don't love getting index information so often. Especially since some have complained about seemingly mundane things needing elevated db permissions.
Since this is protected by a private variable on `values()`, it's not that big of a deal since `_get_default_indexes()` would hit first and that does not hit the db.
But if this is ever used more readily then it may be a problem.

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,98 @@ def delete_runs(name, dry_run=False):
...
+    dataset, field_or_index, pipeline=None, values_only=False
+):
+    """Returns the values of the field for all samples in the given collection that are covered by the index.
+    Raises an error if the field is not indexed.
+
+    Args:
+        dataset: the dataset to query
+        field_or_index: the name of the indexed field to query or the index key for a compound index
+        pipeline: an optional pipeline to apply to the collection. For performance,
+            this should not include any fields not covered by the index.
```
swheaton, line 1497: hm this should not actually be a pipeline, it is really wanting a filter query document.

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,98 @@ def delete_runs(name, dry_run=False):
...
+            hint = {field_or_index: 1}
+        else:
+            proj = {f: 1 for f in field_or_index.strip("_1").split("_1_") if f}
+            hint = {k: v for k, v in proj.items()}
+            proj["_id"] = 0
+
+    elif field_or_index in indexes:
+        idx = indexes[field_or_index]["key"]
+        proj = {field: int(direction) for field, direction in idx}
+        hint = {k: v for k, v in proj.items()}
```
swheaton, line 1564: direction doesn't have anything to do with the projection, it's just coincidental that `bool(-1) is True`.
This would probably be more semantically accurate
```suggestion
        hint = {field: int(direction) for field, direction in idx}
        proj = {k: True for k in hint.keys()}
```

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,98 @@ def delete_runs(name, dry_run=False):
...
+        proj = {"_id": 1}
+        hint = {"_id": 1}
+    elif field_or_index in default_indexes:
+        if "_1" not in field_or_index:
+            proj = {field_or_index: 1, "_id": 0}
+            hint = {field_or_index: 1}
+        else:
+            proj = {f: 1 for f in field_or_index.strip("_1").split("_1_") if f}
+            hint = {k: v for k, v in proj.items()}
+            proj["_id"] = 0
```
swheaton, line 1559: logic is covered by statement below so we should not exclude it here. _id might be a part of a default index in the future.

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,98 @@ def delete_runs(name, dry_run=False):
...
+    ):
+        raise ValueError(
+            "Field '%s' is not indexed. Please create an index first or call `values()` instead"
+            % field_or_index
+        )
+    if field_or_index == "id" or field_or_index == "_id":
+        proj = {"_id": 1}
+        hint = {"_id": 1}
+    elif field_or_index in default_indexes:
+        if "_1" not in field_or_index:
```
swheaton, line 1553: won't work if a default index with descending order is added.

swheaton (commented):
Inline comments.

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,98 @@ def delete_runs(name, dry_run=False):
...
+                    dataset, field_or_index, pipeline
+                )
+            ]
+        return [
+            {"id": str(doc["_id"])}
+            for doc in _iter_indexed_values(dataset, field_or_index, pipeline)
+        ]
+    if values_only:
+        # If values_only is True, we need to extract the values from the dict
+        if "_1_" not in field_or_index:
```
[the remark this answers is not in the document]
jleven: Are you suggesting this logic can be rewritten using "mongo's names"?

fiftyone/core/odm/database.py
jleven: Do you have a suggestion for what might be a better approach?

fiftyone/core/collections.py
```diff
@@ -9044,6 +9045,32 @@ def values(
...
+        # If we do not need to follow insertion order, we can potentially use a covered index query
+        # to get the values directly from the index and avoid a collscan.
+        # In the future, we might not want to require _is_full_collection,
+        # but that needs further investigation as a complex view might require an aggregation
+        if (
+            not _enforce_natural_order
+            and (
+                isinstance(field_or_expr, str)
+                or (
+                    isinstance(field_or_expr, list) and len(field_or_expr) == 1
```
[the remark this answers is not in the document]
brimoor: yep this must be addressed 👍

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,98 @@ def delete_runs(name, dry_run=False):
     )
 
 
+def get_indexed_values(
+    dataset, field_or_index, pipeline=None, values_only=False
+):
+    """Returns the values of the field for all samples in the given collection that are covered by the index.
```
brimoor, line 1490: Please respect the 79 character line limit from the [style guide](https://github.com/voxel51/fiftyone/blob/develop/STYLE_GUIDE.md)!

The pre-commit hooks enforce it automatically on code, but you must manually enforce it in comments/docstrings.

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,98 @@ def delete_runs(name, dry_run=False):
     )
 
 
+def get_indexed_values(
```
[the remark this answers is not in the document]
brimoor: Agree with Stuart. Things in `fiftyone.core.odm` should operate on pymongo collections, not FO datasets. That convention is carefully followed in `fiftyone.core.odm` today.

fiftyone/core/odm/database.py
kaixi-wang: default indexes are hard coded right now and do not use descending order. Plus the qp panel also doesn't allow selecting descending order

fiftyone/core/odm/database.py
kaixi-wang: yeah I should probably call it query to distinguish it from an aggregation pipeline.

fiftyone/core/odm/database.py
swheaton: i suggest narrowing scope to only the cases we said we'd address immediately: filepath and id , or both

fiftyone/core/odm/database.py
swheaton: i understand that, im saying IF one was added.

fiftyone/core/collections.py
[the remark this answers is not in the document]
kaixi-wang: limiting it to id, _id, filepath here for now. It'll give the most bang for the buck with least risk. The method to read from index is exposed as a public method so if a user needs, they can always get the values using that separate function.

fiftyone/core/odm/database.py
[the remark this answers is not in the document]
kaixi-wang: I got rid of it and just wrapped it in a try. That way it doesn't check for indexes and potentially trigger the mongo permission error if they are using limited permissioned mongo_uris for direct db access.

fiftyone/core/odm/database.py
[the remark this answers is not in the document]
kaixi-wang: good point... I started with a collection but realized the default indexes were only accessible from the FO dataset and changed it without moving the function... fixing it

fiftyone/core/collections.py
```diff
@@ -9044,6 +9045,36 @@ def values(
...
+            elif isinstance(field_or_expr, list):
+                field = field_or_expr[0]
+
+            # Checking _is_full_collection last as it could potentially be expensive.
+            # In the future, we might be less restrictive, but that needs
+            # further investigation.
+            if (
+                field in ["id", "_id", "filepath"]
+                and self._dataset.media_type != fom.GROUP
+                and self._is_full_collection()
```
swheaton, line 9065: I think you had it right the first time, `_is_full_collection()` should handle the groups cases.
I don't see why it would be an expensive op, it's just a few comparisons.
```suggestion
                and self._is_full_collection()
```

fiftyone/core/collections.py
```diff
@@ -9044,6 +9045,36 @@ def values(
...
+
+        # If we do not need to follow insertion order, we can potentially use a
+        # covered index query to get the values directly from the index and
+        # avoid a COLLSCAN.
+        if not _enforce_natural_order:
+            field = None
+            if isinstance(field_or_expr, str):
+                field = field_or_expr
+            elif isinstance(field_or_expr, list):
+                field = field_or_expr[0]
```
swheaton, line 9057: this will return the wrong values if you called `ds.values(["id", "filepath"])` - it would return the same as `ds.values('id')`.

I suggest removing this and we will note it for follow on work.

```suggestion
```

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,90 @@ def delete_runs(name, dry_run=False):
...
+    """Returns the values of the field(s) for all samples in the given collection
+    that are covered by the index. Raises an error if the field is not indexed.
+
+    Args:
+        collection: a ``pymongo.collection.Collection`` or
+            ``motor.motor_asyncio.AsyncIOMotorCollection``
+        field_or_fields (str or list): the field name(s) to retrieve.
+        index_key (None): the name of the index to use. If None, the default
+            index name will be constructed from the field name(s).
+        query (dict, optional): a selection filter to apply when querying.
```
swheaton, line 1504: docstring style nit
```suggestion
        query (None): a dict selection filter to apply when querying.
```

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,90 @@ def delete_runs(name, dry_run=False):
...
+            return [{"id": str(doc["_id"])} for doc in cursor]
+
+        if values_only:
+            # If `values_only` is True, we need to extract the values from the dict
+            if isinstance(field_or_fields, str):
+                # Flatten single field values for consistency with `values()`
+                return [doc[field_or_fields] for doc in cursor]
+            return [list(doc.values()) for doc in cursor]
+
+        return cursor.to_list()
```
swheaton, line 1532: I know `to_list()` is faster. But in Teams I think it's currently going to send the entire list in one big response without batching, which could be bad.
Can we use the `list()` form, as it's a known entity - and circle back on this once we make Teams better handle `to_list()` ?

```suggestion
        return list(cursor)
```

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,90 @@ def delete_runs(name, dry_run=False):
...
+    Returns:
+        a list of values for the specified field or index keys for each sample
+        sorted in the same order as the index
+    """
+    try:
+        cursor = _iter_indexed_values(
+            collection, field_or_fields, index_key=index_key, query=query
+        )
+
+        if field_or_fields == "id":
```
swheaton, line 1520: Noting a minor case like `ds.values(["id", "filepath"])` where we still want to cast "_id" to string.
Not worth doing anything about it here though.

fiftyone/core/odm/database.py
```diff
@@ -1484,6 +1484,90 @@ def delete_runs(name, dry_run=False):
...
+        hint = "_id_"  # special case for _id
+    else:
+        if isinstance(field_or_fields, str):
+            proj[field_or_fields] = 1
+            if not hint:
+                hint = field_or_fields + "_1"
+        else:
+            proj.update({f: 1 for f in field_or_fields})
+            if not hint:
+                hint = "_".join([f + "_1" for f in field_or_fields])
```
swheaton, line 1566: Same minor case of `["filepath", "id"]` - this time the problem is the hint will be `filepath_1_id_1` instead of `filepath_1__id_1`.
Same suggestion to not worry about it right now.

swheaton (commented):
- Couple things I think we should change.
- Couple nits.
- Same minor case of `["filepath", "id"]` that has various issues. Just noting for posterity, as I don't think we need to fix or support that right now.

fiftyone/core/collections.py
kaixi-wang: Is full collection returns true for grouped datasets. However the current implementation with values for groups only returns a slice. If we were to try to use this for id, the counts wouldn't match. Plus, we don't index by group name + id by default so there isn't any builtin guarantee optimization to preserve existing behavior

fiftyone/core/collections.py
swheaton: Ok!

fiftyone/core/odm/database.py
kaixi-wang: sure

fiftyone/core/collections.py
kaixi-wang: oh interesting! glad to remove an unnecessary restriction here

Side note, I had checked:
```
ds.select_group_slices(ds.group_slices, _allow_mixed=True)._is_full_collection() # False
```
maybe an issue with _is_full_collection(), but not something to address here

swheaton: > Regarding _is_full_collection() as being potentially expensive, I recall Brian saying that _pipeline() could trigger off being frames/patches/etc collections or something and that's why it's also last in the method.

Yeah OK it should be safe though cuz of the other checks.

swheaton (dismissed):
thanks for addressing things. everything looks good now. Nice one!
🚢 🇮🇹

swheaton:
This will have to retarget release branch if we'd like to get it in though

fiftyone/core/collections.py
```diff
@@ -9044,6 +9045,35 @@ def values(
...
+                field in ["id", "_id", "filepath"]
+                and self._is_full_collection()
+            ):
+                try:
+                    return foo.get_indexed_values(
+                        self._dataset._sample_collection,
+                        field,
+                        values_only=True,
+                    )
+                except Exception as e:
```
brimoor, line 9072: Can we not use the `try`-`except` pattern here please? 

Here's my thinking: core methods need to be as fast as possible, and they need to work when used correctly or not work when used incorrectly. Catching an exception and gracefully continuing implies that we first tried something (potentially expensive), but it failed, so we're falling back to some other potentially expensive thing. It's in the best interest of the caller to just see the exception so that they know they shouldn't be using the method differently.

In other words, we shouldn't be shipping some feature that might log warnings here, because by definition that thing is doing more work than it could be.

brimoor (changes requested):

fiftyone/core/collections.py
swheaton: Doesn't matter too much because it's a private usage, but yeah that should be fine I'd think.

kaixi-wang: >  core methods need to be as fast as possible, and they need to work when used correctly or not work when used incorrectly. 
the `try: except:` allows us to attempt to run it without checking the indexes manually first. Since checking indexes requires elevated mongo permissions, I didn't want to potentially trigger that error in a values() call. 
I know that the sidebar frequently checks the indexes, but an exclusive sdk user potentially may have never hit those codpaths and needed those permissions .

If it's not in the index, the error will throw immediately (the find will not attempt to execute)

brimoor: I realize that in this particular case, the idea is "check to see if there is a covered index, and if there is, then use a more optimal code path to get the values". So in fact the pathway through the function is not deterministic, it may go one way or another. 

My point is that I'd like to see the code implemented more like this:

```py
if not _enforce_natural_order and _has_covered_indexes(...):
    return _get_indexed_values(...)
```

Then it is clear what is going on.

brimoor: In my opinion, the semantics of passing `_enforce_natural_order=False` are "get me the values in any order you want". So if we can get the values in natural order, why are we logging a warning?

If the goal is to provide a hint to the user that the method could be optimized by adding a certain index, then that seems more like a `debug()` use case to me.

kaixi-wang: I am fine changing it back to debug.

Is `_has_covered_indexes(...)` a blocking requirement?

brimoor: Also, suppose I accidentally contribute a bug to `get_indexed_values()` where it no longer worked. The way it is invoked by `values()` now with the naked `try`-`except` would suppress the real bug and turn it into a warning that isn't the user's fault!

swheaton (dismissed):
reapproval

fiftyone/core/collections.py
brimoor: > the try: except: allows us to attempt to run it without checking the indexes manually first. Since checking indexes requires elevated mongo permissions, I didn't want to potentially trigger that error in a values() call.
I know that the sidebar frequently checks the indexes, but an exclusive sdk user potentially may have never hit those codpaths and needed those permissions .

I appreciate the careful thinking here, but given the following:
- creating datasets requires the ability to create and index collections
- as you mention, the App's sidebar does lots of index checks

the reality is that all user personas need to be able to CRUD indexes

fiftyone/core/collections.py
```diff
@@ -9155,6 +9156,35 @@ def values(
...
+        # covered index query to get the values directly from the index and
+        # avoid a COLLSCAN.
+        if not _enforce_natural_order:
+            field = None
+            if isinstance(field_or_expr, str):
+                field = field_or_expr
+            elif isinstance(field_or_expr, list) and len(field_or_expr) == 1:
+                field = field_or_expr[0]
+
+            # Checking _is_full_collection last as it could potentially be expensive.
```
brimoor, line 9170: `_is_full_collection()` will never be expensive! It exists precisely so that there is a guaranteed fast check to see if we're operating on a full (mongo) collection. Please remove this expense claim 😄

fiftyone/core/collections.py
```diff
@@ -9155,6 +9156,35 @@ def values(
         Returns:
             the list of values
         """
+
+        # If we do not need to follow insertion order, we can potentially use a
+        # covered index query to get the values directly from the index and
+        # avoid a COLLSCAN.
+        if not _enforce_natural_order:
+            field = None
+            if isinstance(field_or_expr, str):
```
brimoor, line 9165: Note that this is also a valid syntax:

```py
dataset.values("filepath", expr=F().upper())
```

so you must also ensure that `expr is None` in order to use the index pathway.

fiftyone/core/collections.py
kaixi-wang: oh then i was misremembering 
removed

fiftyone/core/collections.py
```diff
@@ -9155,6 +9156,39 @@ def values(
...
+                field = field_or_expr
+            elif isinstance(field_or_expr, list) and len(field_or_expr) == 1:
```
brimoor, line 9167: The method signature promises that `field_or_expr` can be a tuple, so that needs to be supported here too.

There is an `etau.is_container()` utility that exists specifically to check for any container-like thing (list, tuple, even iterable)

fiftyone/core/collections.py
kaixi-wang: ahhh so much packed into every function! good catch thanks.

fiftyone/core/collections.py
```diff
@@ -9155,6 +9156,39 @@ def values(
...
+                        self._dataset._sample_collection,
+                        field,
+                        values_only=True,
+                    )
+                except ValueError as e:
+                    # Fallback to values, but log the error to surface the
+                    # recommendation to create an index
+                    logger.debug(e)
+                except Exception as e:
+                    # Unexpected error, so we should surface it in the logs
```
brimoor, line 9188: I disagree with this. Why should unexpected errors be tolerated in `get_indexed_values()` specifically? I think they should be raised like they are everywhere else.

fiftyone/core/odm/database.py
```diff
@@ -1542,6 +1542,90 @@ def delete_runs(name, dry_run=False):
...
+        index_key (None): the name of the index to use. If None, the default
+            index name will be constructed from the field name(s).
+        query (None): a dict selection filter to apply when querying.
+            For performance, this should only include fields that are in
+            the specified index.
+        values_only (False): whether to remove field names from the resulting list.
+            If True, the field names are removed and only the values will be
+            returned as a list for each sample. If False, the field names are
+            preserved and the values will be returned as a dict for each sample.
+    Returns:
```
brimoor, line 1569: I believe the docstring won't render correctly unless there is an empty line before `Returns:` here.

fiftyone/core/odm/database.py
```diff
@@ -1542,6 +1542,90 @@ def delete_runs(name, dry_run=False):
...
+    query=None,
+    values_only=False,
+):
+    """Returns the values of the field(s) for all samples in the given collection
+    that are covered by the index. Raises an error if the field is not indexed.
+
+    Args:
+        collection: a ``pymongo.collection.Collection`` or
+            ``motor.motor_asyncio.AsyncIOMotorCollection``
+        field_or_fields (str or list): the field name(s) to retrieve.
```
brimoor, line 1559: I know its annoying, but our docstring style is to use `()` to denote default values, not types. So I'd remove `(str or list)` here.

Note that `(None)` and `(False)` below are default values, not types 😄

fiftyone/core/collections.py
brimoor: Unless I am missing something about `get_indexed_values()` where it might not work deterministically?

kaixi-wang: `get_indexed_values()` itself will raise all errors, but I think it should be tolerated inside values() and only raise if values() also errors

brimoor: What's an example of a non-`ValueError` that `get_indexed_values()` could raise where it makes sense to gracefully continue here? 

My claim is that anything other than the specific "you should add an index" error that we're logging as a debug message represents a logical error in the implementation of `get_indexed_values()` that needs to be fixed, not ignored.

fiftyone/core/collections.py
kaixi-wang: good to keep in mind in the future, but since we are only currently supporting single fields, I believe the current conditional sufficies.

```
import eta.core.utils as etau

field_or_expr = ("id")
etau.is_container(field_or_expr) # False
isinstance(field_or_expr, str) # True
```

fiftyone/core/collections.py
kaixi-wang: ok I'm fine with raising unanticipated errors since they must specifically add  `_enforce_natural_order=False`, so anyone using normal values() cannot be impacted

fiftyone/core/collections.py
swheaton: Technically you could do this: `ds.values(("filepath",))`

brimoor (approved):
Thanks for addressing my review comments 🙇

PR #5703, opened 2025-04-09
FOTD Optimizations: Unify IDs with other fields, implement `__getitems__`, add tests.

jacobsela:
# Changes
1. refactored `_load_cached_fields` to `_load_fields`
2. the field `id` (that's always loaded to memory) now goes through this path as well
3. unittests
4. implement __getitems__ rather than __getitem__

## Unifying IDs with other fields

startup time test:
- 32 workers
- 20000 samples

### snippet to test

```python
from time import time_ns
import torch
import fiftyone as fo

def dummy_get_item(sample):
    return "foo"

if __name__ == "__main__":
    import torch.multiprocessing as mp
    mp.set_start_method('forkserver')
    mp.set_forkserver_preload(['torch', 'fiftyone'])

    dataset = fo.load_dataset("quickstart")
    torch_dataset = dataset.to_torch(dummy_get_item)
    dataloader = torch.utils.data.DataLoader(
        torch_dataset,
        batch_size=1,
        num_workers=32,
    )

    time_start = time_ns()
    for _ in dataloader:
        break
    time_end = time_ns()

    print(f"Time taken to load first batch: {(time_end - time_start) / 1_000_000} ms")
```

### before:
```
Time taken to load first batch: 39445.790492 ms
```

### after:
```
Time taken to load first batch: 5597.779764 ms
```

### Snippet to test:

```python
from time import time_ns
import torch
import fiftyone as fo

def id_get_item(sample):
    return sample['id']

if __name__ == "__main__":
    import torch.multiprocessing as mp
    mp.set_start_method('forkserver')
    mp.set_forkserver_preload(['torch', 'fiftyone'])

    BATCH_SIZES = [2 ** i for i in range(8)]

    dataset = fo.load_dataset("stanford-dogs")
    torch_dataset = dataset.to_torch(id_get_item)

    for batch_size in BATCH_SIZES:

        dataloader = torch.utils.data.DataLoader(
            torch_dataset,
            batch_size=batch_size,
            num_workers=32,
        )

        time_start = time_ns()
        for _ in dataloader:
            pass
        time_end = time_ns()

        print(f"==========================")
        print(f"Batch size: {batch_size}")
        print(f"Time to iterate dataset: {(time_end - time_start) / 1_000_000_000} s")
        print(f"Average time per sample: {(time_end - time_start) / len(torch_dataset) / 1_000_000_000} s")
        print(f"==========================")
```

### before:
```
==========================
Batch size: 1
Time to iterate dataset: 23.707311036 s
[...]

PR #5745, opened 2025-04-19
Add torch packages to test requirements

brimoor:
I pulled the latest `develop` and got unit test failures due to the new `tests/unittests/torch_dataset_tests.py` tests that were added by @jacobsela in https://github.com/voxel51/fiftyone/pull/5703.

Upon inspection, the solution was to upgrade my `torch` and `torchvision` packages, so I've pinned sufficiently new versions to `requirements/test.txt` to ensure that others don't encounter this bug, assuming they follow the dev install instructions and run `bash install.bash -d`.

### DIFF
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -22,7 +22,6 @@ from PIL import Image
 import eta.core.geometry as etag
 import eta.core.learning as etal
 import eta.core.utils as etau
-from torchvision.models.feature_extraction import create_feature_extractor
 
 import fiftyone.core.config as foc
 import fiftyone.core.labels as fol
@@ -36,11 +35,14 @@ import fiftyone.core.sample as fos
 import fiftyone.core.view as fov
 
 fou.ensure_torch()
+
 import torch
-import torchvision
-from torchvision.transforms import functional as F
 from torch.utils.data import Dataset
 import torch.distributed as dist
+
+import torchvision
+from torchvision.models.feature_extraction import create_feature_extractor
+from torchvision.transforms import functional as F
 
 
 logger = logging.getLogger(__name__)
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -8,5 +8,7 @@ pytest-cov==4.0.0
 pytest-mock==3.10.0
 pytest-asyncio
 shapely
-tensorflow==2.17.0
+tensorflow>=2.17.0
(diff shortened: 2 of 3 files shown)


fiftyone/utils/torch.py
```diff
@@ -22,7 +22,6 @@
 import eta.core.geometry as etag
 import eta.core.learning as etal
 import eta.core.utils as etau
-from torchvision.models.feature_extraction import create_feature_extractor
```
brimoor, line 25: @jacobsela for future reference, it's important that all `torch`/`torchvision` imports come **after** `fou.ensure_torch()` in FO modules.

By design, `torch` is not a mandatory requirement of `fiftyone`. It needs to be possible to use FiftyOne's core functionality without having `torch` installed. 

It's okay for *optional* functionality to require `torch`. But, by convention, to be as helpful to the user as possible, we added the `fou.ensure_torch()` method, which raises an informative error message that explains to the user that they've tried to use something that requires torch to be installed:

```
In [1]: fou.ensure_torch()

ImportError: The requested operation requires that 'torchvision' is installed on your machine.

If you think this error is inaccurate, you can set `fiftyone.config.requirement_error_level` to 1 (warning) or 2 (ignore).
See https://docs.voxel51.com/user_guide/config.html for details.
```

jacobsela: That makes sense, thanks for letting me know.

jacobsela (approved):
@brimoor LGTM, tests run fine on my end. Do you know what the exact failure was when you ran the tests? These tests use pretty generic torch functionality that's been around for many versions. Maybe I can account for them and have a broader torch requirement.

brimoor:
@jacobsela hmm, unfortunately I don't remember what version of `torchvision` I was running before I upgraded, but I was getting some kind of error with the new import you added:

```
from torchvision.models.feature_extraction import create_feature_extractor
```

However, I just tried downgrading to `torch==2.0.0` and `torchvision==0.15.0` and the above import still works, so IDK what my issue was... 🤷‍♂️ 

In any case, I'm glad to hear that `create_feature_extractor` has been around for many versions because, indeed, it is ideal for our core methods to have the broadest possible compatibility range for `torch`/`torchvision`, to avoid situations where people are precluded from using `fiftyone` because they need to work with older `torch`/`torchvision` packages for whatever non-FO reason.

PR #5747, opened 2025-04-22
[WIP] Fix double batching in save context

CamronStaley:

TODO:
- Is it possible to split the curr_batch_size_bytes by samples and frames so that we are accurately handling the batch size instead of potentially doing two half batches (though this might not happen often I am not sure)
- Testing + benchmarking

tbd

        notes for FiftyOne users.

-   [ ] Other


fiftyone/core/collections.py
```diff
@@ -184,7 +185,10 @@ def save(self, sample):
                     len(str(op)) for op in frame_ops
                 )
 
-            if self._curr_batch_size_bytes >= self.batch_size:
```
CamronStaley, line 187: this is what I am referring to in the TODO section. We have a single curr_batch_size_bytes which results in samples or frames or both being written once we reach curr_batch_size_bytes. Ideally we'd have something like curr_batch_size_bytes_samples and curr_batch_size_bytes_frames which are handled independently. I was a bit confused by the batch_ids stuff though so I need to look into that more before trying to refactor this.

brimoor: I would not recommend messing around with separating out sample and frame size tracking.

It is true that there are separate `bulk_write()` calls for samples and frames. So if the target size is 1MB, then the then the current guarantee is a bit conservative: the *sum* of the `bulk_write()` calls will be no more than 1MB. But the goal is not too do updates that are *exactly* 1MB, just to be within the ballpark, which this achieves. One of the bulk writes will be ~0.5MB or larger.

The `batch_ids` are relevant for cases like frame views and patch views, where the edits need to be synced to the view's parent collection.

There's also `_reload_parents`, which is relevant in cases where the `SaveContext` is being applied to a `DatasetView`, as we promise that any edits will be automatically synced to any in-memory `Sample` and `Frame` objects in the parent `Dataset` (this is the "samples are singletons" behavior that is somewhat of a thorn in our side because it is expensive to implement this).

It would not be straightforward to separate out the syncing and reloading of samples from their frames. Better to leave it as-is, where the all changes associated with a given sample are guaranteed to be performed in one call to `_save_batch()`.

fiftyone/core/collections.py
```diff
@@ -125,6 +125,7 @@ def __init__(
         self._batching_strategy = batching_strategy
         self._curr_batch_size = None
         self._curr_batch_size_bytes = None
+        self._encoding_ratio = 1.0
```
brimoor, line 128: What's this for? Looks unused and not directly related to the goal of removing double-batching, which you can achieve just via a `bulk_write(..., batch=False)` syntax.

fiftyone/core/odm/database.py
```diff
@@ -844,9 +844,20 @@ def bulk_write(ops, coll, ordered=False, progress=False):
...
+            batching strategies do not work for your use case.
 
+    Returns:
+        a list of :class:`pymongo.results.BulkWriteResult`
+    """
+    # use static batcher so we still have a progress bar when requested
+    batcher = (
+        fou.get_default_batcher(ops, progress=progress)
+        if batch
+        else fou.StaticBatcher(ops, progress=progress, batch_size=len(ops))
```
brimoor, line 858: Following up on a conversation we had offline: I am onboard with *not* exposing a `batcher` argument to the public FO interface as I initially have proposed in https://github.com/voxel51/fiftyone/pull/5667.

However, after I remove that (will do so shortly), I propose that you review that PR, which includes the addition of a `bulk_write(..., batcher=)` syntax that allows for providing a custom batcher, or `batcher=False` to disable batching.

Note that the `fiftyone.core.odm` module is strictly for internal usage (as is anything that is not directly imported into the root `fiftyone` namespace and is not a public method of something in the public namespace). And for internal usage, I think its reasonable to support providing custom batchers to `bulk_write()` for specific use cases. Indeed, with `SaveContext` we have a case where we want to disable batching 😄

brimoor: So concretely, I propose we merge https://github.com/voxel51/fiftyone/pull/5667 first, then you can use `bulk_write(..., batcher=False)` in this PR

brimoor (changes requested):

fiftyone/core/collections.py
CamronStaley: This is a new concept we introduced with compression across the API. Can talk more offline but the general use case is if you're encoding the data (compressing, base64, etc) to some size smaller than the actual object size, then you'd want to factor that value into your batching when calling through the API otherwise compression doesn't actually reduce the total number of batches and adds pointless overhead.

fiftyone/core/collections.py
CamronStaley: sounds good. Noting for myself I noticed that the content size batcher here doesn't account for the max number of operations we typically allow (100k). This probably never happened previously because operations will almost always hit 1M bytes before reaching 100k total ops, but now that we are support compression it could be possible that we have a batch of 100k ops at 1M bytes (though unlikely I have only ever seen this when doing insert_many with set_values calls which are very small). Regardless we should check both content size and ensure we are doing <100k ops.

CamronStaley: Also noting that this is my main concern with having custom batchers it becomes hard to consider all these random factors like compression that didn't exist previously. I also exposed a config variable for compression with direct to mongo calls but we don't currently have a way of getting back the compressed size ideally we'd eventually factor that into batching as well.

fiftyone/core/collections.py
```diff
@@ -170,21 +176,32 @@ def save(self, sample):
 
         if self._batching_strategy == "static":
             self._curr_batch_size += 1
```
CamronStaley, line 178: this doesn't seem right. It's weird to me in the default batchers we are typically looking at num operations but here it's num samples even though it could be multiple operations? @brimoor can you confirm what the expected behavior is here? I'd like to limit total number of ops based on what mongo can support. Also, do we think this logic should be included in the latency batcher as well? I believe it is for our default batcher.

fiftyone/core/odm/database.py
```diff
@@ -881,6 +881,7 @@ def bulk_write(ops, coll, ordered=False, batcher=None, progress=False):
             a progress callback function to invoke instead
     """
     batcher = fou.get_default_batcher(ops, batcher=batcher, progress=progress)
+    results = []
```
brimoor, line 884: please add a `Returns:` section to the docstring documenting what is returned 😄

fiftyone/core/collections.py
brimoor: > Also, do we think this logic should be included in the latency batcher as well? I believe it is for our default batcher

I'm not sure what you're asking/proposing to do. Can you explain?

brimoor: > It's weird to me in the default batchers we are typically looking at num operations but here it's num samples even though it could be multiple operations? @brimoor can you confirm what the expected behavior is here?

Batchers are generic things; they could be used for splitting any iterable into batches, not just `bulk_write()` calls.

So if a user performs this global configuration:

```py
fo.config.default_batcher = "static"
fo.config.batcher_static_size = 100
```

then all they're saying is that this code should emit batches that correspond to 100 elements of `iterable`:

```py
batcher = fou.get_default_batcher(iterable, ...)
for batch in batcher:
    ...
```

brimoor: The docstring for `save_context()` implies that `batching_strategy="static"` will save `batch_size` samples at a time, which would mean exactly `batch_size` ops for non-video datasets but could mean >> `batch_size` frame ops for video datasets. 

This is precisely why the `default_batcher` is not `"static"`, because in practice samples can be small or big so it is not useful to express things in terms of number of samples processed.

fiftyone/core/collections.py
```diff
@@ -170,37 +176,65 @@ def save(self, sample):
...
                 self._curr_batch_size = 0
         elif self._batching_strategy == "size":
             if sample_ops:
+                self._curr_batch_size += len(sample_ops)
                 self._curr_batch_size_bytes += sum(
                     len(str(op)) for op in sample_ops
                 )
 
             if frame_ops:
+                self._curr_batch_size += len(sample_ops)
```
brimoor, line 193: typo: `sample_ops -> frame_ops`

fiftyone/core/collections.py
CamronStaley: Idk if it comes from mongo or we arbitrarily defined it but we have a max limit of 100k mongo operations per batch when using the batchers. It's unlikely that we hit 100k ops before we reach 1M bytes, but with compression it's possible. For save context should we have parody with this functionality and if so should this apply to all 3 batching types or just size? For more context it applies to all 3 of our default batchers.

brimoor: In any case, to be precise you shouldn't use `_curr_batch_size` to enforce a maximum op limit, because it tracks samples, not ops, for the static batching case (as promised by the docstring).

fiftyone/core/collections.py
brimoor, line 193: So theoretically it could happen that `frame_ops` contains 99k ops, and then after pulling in the frame ops for the next sample, it could *exceed* 100k ops (eg the video could have more than 1k frames and the user edited all of them).

Despite the promise made by the new `max_batch_size` parameter, the proposed implementation wouldn't actually prevent the "100k limit" error.

Here's an alternative strategy: *don't* add `max_batch_size=` to `SaveContext` at all. Instead, just update `bulk_write()` to this:

```py
    if batcher is False:
        batcher = partial(StaticBatcher, batch_size=-1, max_batch_size=100000)
```

Then it would gracefully split into multiple batches inside `bulk_write()` *only if necessary*.

brimoor: Go ahead and hard-code `max_batch_size=100000` here too:
https://github.com/voxel51/fiftyone/blob/7117db9e2e00688adc5b7804fe9aad84e7e95b85/fiftyone/core/utils.py#L1812

The other default batchers already encode the 100k limit.

(which means that, while batchers *could* be arbitrary things, we are still designing for the typical case that they will be used to generate bulk writes).

brimoor (approved):
LGTM

PR #5757, opened 2025-04-23
Optimize built-in App sort by similarity

benjaminpkane:

In the `/sort` route which resolves sorting by similarity, the serialized stage must be returned to include the `_state` kwarg which ensures the sort pipeline is not recomputed on grid paging requests

-   [ ] Other


jleven (commented):
@benjaminpkane, can you please add context about the testing you've done? Thank you!

benjaminpkane:
The issue the PR resolved is excessive pipeline creation for the sort by similarity stage, which is expensive. This is why `_state` caching exists. We want to /sort route to include that state in its response so other calls, e.g. in the grid, do not create the pipeline again when constructing the view on the server
* https://github.com/voxel51/fiftyone/blob/5c16b3b076d8fda45c6e87888e33fd7271e8dc8c/fiftyone/core/stages.py#L7750

PR #5758, opened 2025-04-23
Remove `batcher=` argument from public interface

brimoor:
Per offline discussion, lets not expose custom `batcher=` support added in https://github.com/voxel51/fiftyone/pull/5667 in the public interface just yet.

The stance is that we'll take responsibility for making batching as optimal as we can. And if users need to tune the strategy they can do so via the `fiftyone.config.*batcher*` config settings.

## Summary by CodeRabbit

- **Refactor**
  - Removed support for the `batcher` parameter when adding samples to datasets.
  - Updated related documentation and improved docstring formatting.

- **Tests**
  - Removed tests related to the `batcher` parameter when adding samples.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -3253,7 +3253,6 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         expand_schema=True,
         dynamic=False,
         validate=True,
-        batcher=None,
         progress=None,
         num_samples=None,
     ):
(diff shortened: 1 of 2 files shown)


swheaton (approved):
good idea to remove this surface area.
🚢 🇮🇹

CamronStaley (approved):
lgtm. mb I think I suggested not having a batcher param in the public methods, but wasn't sure that we had agreed on that. This seems safer since the batching gets complicated and I am hesitant to believe a user would make a more optimized batcher that's reliable.

Issue #5583, opened 2025-03-14
[BUG] Label mapping error when using Grounding-DINO for object detection


yumixx:
### Describe the problem

I was using `Grounding-DINO` model for an object detection task. The model works well when only one input class is provided. However, when I specify multiple input classes, the labels in outputs seem to be mapped incorrectly.

### Code to reproduce issue

```python
import fiftyone.zoo as foz

classes = ["cat", "face"]

model = foz.load_zoo_model(
    "zero-shot-detection-transformer-torch",
    "IDEA-Research/grounding-dino-tiny",
    classes=classes,
)

dataset.apply_model(model, label_field="grounding_dino")
```
# commands and/or screenshots here

Here is the label output:

![Image](https://github.com/user-attachments/assets/c4bcfbfb-ec07-4e83-9ea1-6f92379ae04b)

### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 22.04): ubuntu 22.04
-   **Python version** (`python --version`): 3.11.11
-   **FiftyOne version** (`fiftyone --version`): FiftyOne v1.3.2, Voxel51, Inc.
-   **FiftyOne installed from** (pip or source): pip

### Other info/logs

In fiftyone, the function used to predict is `image_process.post_process_object_detection` in file `transformers.py`, which defers from the function in huggingface example: `post_process_grounded_object_detection()`, where the integer label is already mapped to text labels.

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another
member of your organization be willing to contribute a fix for this bug to the
FiftyOne codebase?

-   [ ] Yes. I can contribute a fix for this bug independently
-   [ ] Yes. I would be willing to contribute a fix for this bug with guidance
        from the FiftyOne community
-   [ ] No. I cannot contribute a bug fix at this time

PR #5770, opened 2025-04-25
Fix label mapping in Grounding DINO by using post_process_grounded_object_detection

AdonaiVera:
This PR addresses the incorrect class mapping issue related to [Issue #5583](https://github.com/voxel51/fiftyone/issues/5583), initially detected by @yumixx.

### Problem
Previously, the code used `post_process_object_detection`, which returns label indices corresponding to token positions in the text prompt—not actual class indices. This caused confusion when multiple classes like `"a cat. a dog."` were passed as prompt, since the returned labels could not be reliably mapped back to class names.

### Solution
The fix involved switching from `post_process_object_detection` to `post_process_grounded_object_detection`, a more appropriate method provided by Hugging Face's [GroundingDinoProcessor](https://huggingface.co/docs/transformers/v4.51.3/en/model_doc/grounding-dino#transformers.GroundingDinoProcessor).

This new method internally handles token-to-class mapping correctly and is better suited for zero-shot multi-class detection scenarios.

### Additional Changes
- Removed the custom `id2label` mapping logic since `post_process_grounded_object_detection` handles it internally.
- Added validation through before/after examples.
- Confirmed the mapping now aligns correctly with the original class list.

> Note: Grounding DINO still produces a number of false positives, which is a known limitation from the original research, but the class mapping is now functioning as expected.

Before:
Screenshot 2025-04-24 at 3 13 59 PM

After:
Screenshot 2025-04-24 at 9 50 04 PM

Please let me know what you think or if you suggest any changes. 🙌


brimoor (commented):
@AdonaiVera thanks for the contribution!

@jacobsela can you please review this PR as code owner, since you are currently working on our Transformers integration?

AdonaiVera:
Perfect, thank you @brimoor!  🤗 

Hi @jacobsela, I checked out your PR: [https://github.com/voxel51/fiftyone/pull/5761](https://github.com/voxel51/fiftyone/pull/5761). I saw that you refactored the whole function, including changes like using `self.transform` directly instead of `self.image_processor`, along with many other updates.

If you think is better. I can wait for your PR to be merged and then update mine to match. The rest, like using `id2label`, stays the same. It’s a small change to get Grounding DINO working with multiclass.

To test multiclass, you can run this:

```python
import fiftyone as fo
import fiftyone.zoo as foz

classes = ["cat", "face"]

model = foz.load_zoo_model(
    "zero-shot-detection-transformer-torch",
    name_or_path="IDEA-Research/grounding-dino-tiny",
    classes=classes,
)

dataset.apply_model(model, label_field="grounding_dino")
```

Let me know what you think! Thank you Jacob! And nice update with the TorchImageModel! 🚀

jacobsela:
Hi @AdonaiVera,

First, thank you for the contribution!

Like you said, I am working on a larger refactor for this code, so I think that it does make sense to wait a little before we integrate this change.

In the meantime, if you are interested, I would think about the following things:
1. How can we make the code work for both grounded and regular post processing, what about different models that have slightly different interfaces or prompt requirements?
2. How can we guarantee that the user always receives boxes with one of the classes passed in? It seems to me that Grounding DINO is very sensitive to choice of prompt.

AdonaiVera:
Hi @jacobsela,
I pulled your latest PR and it solved most of the previous issues. I updated the PR with a few changes to handle some remaining edge cases.

One fix was related to `post_process_grounded_object_detection`, which returns regular Python lists instead of torch tensors. That was causing a `'list' object has no attribute 'detach'` error only for labels. I added a simple check to handle that properly.

Also, I noticed a warning from Hugging Face:

```
FutureWarning: The key `labels` will return integer ids in `post_process_grounded_object_detection` since v4.51.0. Use `text_labels` instead to retrieve string object names.
```

So I made sure to switch to `text_labels` when available.

Now, about your questions: 

**1. Making it work for grounded and regular post-processing:**

You're currently checking for the method name inside the `@processor.setter`. That works fine now, especially since we’re wrapping Hugging Face processors directly. As we add more models, we’ll likely run into differences in output formats, prompt handling, and class naming. Managing all of that with if-checks or attribute lookups can quickly become hard to maintain. Maybe we can introduce a common post_process() method across models, kind of like how Hugging Face’s PreTrainedModel works. Each model can handle its own logic inside that, so we don’t need to keep adding if-checks. This option also gives us more flexibility for optimization later on.

**2. Ensuring boxes for user-passed classes:**
This is actually a model-level limitation. There are related issues here:

[#63 – Prompt-sensitive false positives](https://github.com/IDEA-Research/GroundingDINO/issues/63)
[#85 – Model predicts boxes for irrelevant prompts](https://github.com/IDEA-Research/GroundingDINO/issues/85)

Grounding DINO tends to fuse features even when the text isn’t a good match, which causes false positives or missed classes. What if we add a post-processing step that doesn’t rely on exact string matching, but uses lightweight mapping instead? For example, we could match the first token of each detected phrase to the user-provided class list. It’s simple, fast, and helps align predictions even when the model outputs slightly noisy or merged tokens. It’s not perfect, but it could make results more reliable for open-set models like Grounding DINO, other option could be if we could compute similarity scores between each detected phrase and the user’s class list, selecting the best match above a threshold. If no match is strong enough, the result could be discarded or mapped to a fallback label. This would be slower but more robust, and could serve as a shared post-processing utility for all open-set models.

Let me know what you think — happy to help implement this if useful.

AdonaiVera:
Hey @jacobsela, I was thinking about your point 2, I totally agree it’s critical if we want people to really take advantage of zero-shot detection.

A simple solution could be to add an optional parameter like `semantic_similarity_filtering` when initializing the model. Then, we can re-map predicted labels to the closest user-passed class using existing `embed_prompts` and cosine similarity.

Here’s a sketch of how that could work:

```python
def filter_by_similarity(detected_label, target_classes, model, threshold=0.75):
    """Maps detected label to the most similar target class."""
    # Get embeddings
    label_emb = model.embed_prompts([detected_label])[0]
    target_embs = model.embed_prompts(target_classes)
    
    # Compute cosine similarity
    label_norm = np.linalg.norm(label_emb)
    target_norms = np.linalg.norm(target_embs, axis=1)
    similarities = np.dot(label_emb, target_embs.T) / (label_norm * target_norms)
    
    # Find best match
    best_idx = np.argmax(similarities)
    best_score = similarities[best_idx]
    
    # Return match if above threshold
    return target_classes[best_idx] if best_score >= threshold else None
```
This could work with any zero-shot detector, and since more of these models are being developed, adding this kind of filtering would be really useful for users

AdonaiVera:
Closing this PR because it was already addressed by @CharlesCNorton 🚀

PR #5779, opened 2025-04-26
fix win tests

CamronStaley:

fix win tests

na

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Tests**
  - Certain unit tests are now skipped on Windows platforms to improve test reliability across different operating systems.


tests/unittests/utils/decorators.py
```diff
@@ -0,0 +1,25 @@
...
+| Copyright 2017-2025, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+from functools import wraps
+import platform
+import unittest
+
+
+def skip_windows(func):
```
kaixi-wang, line 13: I think there is already a decorator https://github.com/voxel51/fiftyone/blob/fix/win-tests-map/tests/unittests/decorators.py#L65

CamronStaley: yeah I tried importing it directly but couldn't get the imports working in the test directories. I tried relative import, added __init__.py and import through that but just kept running into all sorts of problems. I saw Ritchie duplicated it here so I assume I am not the first person to have this issue: https://github.com/voxel51/fiftyone/blob/68b99a4886c45b4697aa6850909e60c9b1d165e7/tests/unittests/execution_cache/decorators.py#L16

swheaton: this doesn't work for you?
```python
from decorators import skip_windows
```

swheaton (approved):
think its good now
🚢 🇮🇹

PR #5784, opened 2025-04-26
Documenting `map_samples()` and `update_samples()`

brimoor:
## Change log

- Add user guide documentation for `map_samples()` and `update_samples()`
- Add example usage to `map_samples()` and `update_samples()` docstrings
- Document new config variables introduced in [#5642](https://github.com/voxel51/fiftyone/pull/5642)
- Fix import order regressions in `fiftyone.core.collections` module introduced by [#5642](https://github.com/voxel51/fiftyone/pull/5642)
- Remove unnecessary `multiprocessing.set_start_method()` on Windows introduced by [#5642](https://github.com/voxel51/fiftyone/pull/5642)

## Summary by CodeRabbit

- **New Features**
  - Added new documentation and user guide sections for efficient batch updating and mapping of samples, including parallelization options and progress tracking.
  - Introduced new configuration options for controlling parallelization methods and worker counts.

- **Documentation**
  - Updated terminology in release notes and user guide for clarity, replacing "efficient batch edits" with "save contexts."
  - Reformatted configuration documentation for improved readability.
  - Expanded documentation for new and existing batch processing methods.

- **Bug Fixes**
  - Improved clarity and grammar in code comments.

- **Chores**
  - Added "dill" to installation requirements and adjusted the order of dependencies.
  - Removed unnecessary blank lines and updated internal logic for multiprocessing context handling.


fiftyone/core/collections.py
```diff
@@ -3956,38 +3958,76 @@ def map_samples(
         self,
         map_fcn,
         save=False,
+        skip_failures=False,
+        parallelize_method=None,
```
brimoor, line 3962: We need to set `parallelize_method=None` here so that `fo.config.default_parallelization_method` will be used

fiftyone/core/utils.py
```diff
@@ -2828,9 +2828,6 @@ def get_multiprocessing_context():
         # subsequent usage of things like `multiprocessing.Queue()` will not
         # cause the default start method to switch to 'spawn'
         multiprocessing.set_start_method("fork", force=True)
-    elif sys.platform == "win32":
```
brimoor, line 2831: I'm not sure who added this `win32` block in #5642, but I don't think it is needed and thus we should remove it because we shouldn't be force-ably overwriting the user's `multiprocessing` start method if they have manually configured one.

For context, the `darwin/linux` block immediately above this does use `force=True`, but **ONLY** if the user hasn't manually set a start method. And even then, it is setting the start method for a specific reason as explained in the comment, which doesn't apply to Windows.

j053y (approved):
:nut_and_bolt:

PR #5785, opened 2025-04-27
Add support for setting new embedded document fields

brimoor:
Resolves https://github.com/voxel51/fiftyone-plugins/pull/225#issuecomment-2825901286.

## Change log

- Adds support for populating new embedded document fields via `sample["new.field"] = value` and related syntaxes

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("cifar10", split="test", max_samples=10)

dataset.add_sample_field(
    "data",
    fo.EmbeddedDocumentField,
)

# previously would not work, now does!
dataset.apply_model(model, label_field="data.predictions")

print(dataset.values("data"))
```

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced nested field handling to support dynamic creation and proper validation of embedded documents, improving schema declaration and attribute setting.

- **Tests**
	- Added unit tests confirming dynamic addition and persistence of new embedded document fields, including nested fields, in datasets and views.

### DIFF
--- a/fiftyone/core/odm/mixins.py
+++ b/fiftyone/core/odm/mixins.py
@@ -20,6 +20,7 @@ import fiftyone.core.utils as fou
 
 from .database import get_db_conn
 from .dataset import SampleFieldDocument
+from .embedded_document import DynamicEmbeddedDocument
 from .utils import (
     deserialize_value,
     serialize_value,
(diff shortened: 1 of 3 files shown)


allenleetc:
LGTM but defer to @swheaton @benjaminpkane

fiftyone/core/odm/mixins.py
```diff
@@ -141,16 +142,26 @@ def set_field(
...
 
         if len(chunks) > 1:
-            doc = self.get_field(chunks[0])
+            root = chunks[0]
+            doc = self.get_field(root)
+
+            if doc is None:
+                doc = DynamicEmbeddedDocument()
+                self[root] = doc
+                self._changed_fields.remove(root)
```
swheaton, line 151: should we respect the 1create` parameter here?

swheaton (commented):
1 comment otherwise lgtm

benjaminpkane (dismissed):

fiftyone/core/odm/mixins.py
brimoor: good call, updated 👍

PR #5786, opened 2025-04-27
[WIP] Adding support for private fields

brimoor:
## Change log

- Adds support for storing/declaring private fields on samples, frames, and embedded documents
- Creating private fields is not allowed by default; users must opt into allowing this by passing a (mostly) undocumented `allow_private=True` parameter to the relevant methods
- Private fields **are not** exposed by default by user-facing schema constructs like `get_field_schema()` and `print(sample)`, but they **are** available to system-facing constructs like `__getattr__`, `__getitem__`, `has_field()`, `values()`, etc. in such a way that it should "just work" to write methods that can optionally read/write from private fields

## TODO

- Use this feature internally to validate that the design results in clean code
- Should we document `allow_private=True` in more docstrings?

### Private fields

```py
import fiftyone as fo

dataset = fo.Dataset()

# Private fields can be manually declared...
dataset.add_sample_field("_private", fo.IntField, allow_private=True)

# ...in which case new samples can have private fields
sample = fo.Sample(filepath="image.jpg", _private=1)
dataset.add_sample(sample)

# The API allows checking for the existence of private fields
assert dataset.has_field("_private")
assert "_private" in dataset.get_field_schema(include_private=True)

# ...but they are not exposed to the user by default
assert "_private" not in dataset.get_field_schema()

# Private fields are not shown to the user...
print(sample)
print(dataset)

# ...but the data is in fact there
print(sample._private)
print(dataset.values("_private"))

# Private fields can be edited as you would expect
sample._private = 2
sample.save()

print(dataset.values("_private"))

# Private fields can be renamed and cloned
# `allow_private=True` is required when the *new* fields are private
dataset.rename_sample_field("_private", "_still_private", allow_private=True)
dataset.clone_sample_field("_still_private", "_also_private", allow_private=True)

print(sample._still_private)
print(sample._also_private)
print(dataset.values("_still_private"))
print(dataset.values("_also_private"))

# Private fields can be deleted
dataset.delete_sample_field("_also_private")

assert not dataset.has_field("_also_private")

# Private fields can be set in bulk
dataset.set_values("_still_private", [3])

# New private fields can be declared in bulk
dataset.set_values("_yep_private", [4], allow_private=True)

print(sample._yep_private)

# Private fields cannot be dynamically added to samples by default...
try:
    sample._not_allowed = 5
except:
    print("ERROR")
    sample.reload()

# ...but they can be added like this if necessary
sample.set_field("_is_allowed", 6, allow_private=True)
sample.save()

assert dataset.has_field("_is_allowed")

print(dataset.values("_is_allowed"))

# Private fields cannot be dynamically added by default...
sample = fo.Sample(filepath="image.jpg", _also_not_allowed=7)
try:
    dataset.add_sample(sample)
except:
    print("ERROR")

# ...but they can be added like this
sample = fo.Sample(filepath="image.jpg", _also_is_allowed=8)
dataset.add_sample(sample, allow_private=True)

assert dataset.has_field("_also_is_allowed")
print(dataset.values("_also_is_allowed"))
```

### Private attributes

```py
import fiftyone as fo

dataset = fo.Dataset()

# Private attributes can be added to embedded documents
sample = fo.Sample(
    filepath="image1.jpg",
    ground_truth=fo.Detections(
        detections=[fo.Detection(label="cat", _private=1)],
    ),
)

# You must pass `allow_private=True` in order to dynamically declare them
dataset.add_sample(sample, dynamic=True, allow_private=True)

assert dataset.has_field("ground_truth.detections._private")
assert "ground_truth.detections._private" in dataset.get_field_schema(flat=True, include_private=True)

# Private attributes aren't presented to the user...
detection = sample.ground_truth.detections[0]
print(detection)

# ...but the data is in fact there
print(detection._private)

# Private attributes can be edited at any time
detection._private = 2
sample.save()

print(dataset.values("ground_truth.detections._private"))

# Private attributes can also be added later
detection._also_private = 3
sample.save()

print(dataset.get_dynamic_field_schema(include_private=True))

# Private attributes can be manually declared at any time
dataset.add_sample_field(
    "ground_truth.detections._also_private",
    fo.IntField,
    allow_private=True,
)

assert dataset.has_field("ground_truth.detections._also_private")
print(dataset.values("ground_truth.detections._also_private"))

# Private attributes can be renamed and cloned
# `allow_private=True` is required when the *new* fields are private
dataset.rename_sample_field(
    "ground_truth.detections._private",
    "ground_truth.detections._still_private",
    allow_private=True,
)
dataset.clone_sample_field(
    "ground_truth.detections._still_private",
    "ground_truth.detections._also_still_private",
    allow_private=True,
)

detection = sample.ground_truth.detections[0]

print(detection._still_private)
print(detection._also_still_private)
print(dataset.values("ground_truth.detections._still_private"))
print(dataset.values("ground_truth.detections._also_still_private"))

# Private attributes can be deleted
dataset.delete_sample_field("ground_truth.detections._also_still_private")

assert not dataset.has_field("ground_truth.detections._also_still_private")

# Existing private attributes can be edited in bulk...
dataset.set_values("ground_truth.detections._still_private", [[4]])

print(dataset.values("ground_truth.detections._still_private"))

# ...and new private attributes can be declared in bulk
dataset.set_values(
    "ground_truth.detections._yep_private",
    [[5]],
    dynamic=True,
    allow_private=True,
)

assert dataset.has_field("ground_truth.detections._yep_private")

detection = sample.ground_truth.detections[0]
print(detection._yep_private)
```

### DIFF
--- a/fiftyone/core/clips.py
+++ b/fiftyone/core/clips.py
@@ -42,6 +42,8 @@ class ClipView(fos.SampleView):
             filtered in this view
     """
 
+    __slots__ = ()
+
     @property
     def _sample_id(self):
         return ObjectId(self._doc.sample_id)
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -1333,7 +1333,7 @@ class SampleCollection(object):
         return path in default_fields
 
     def _is_read_only_field(self, path):
-        _, _, read_only = self._parse_field(path, include_private=True)
+        _, _, read_only = self._parse_field(path)
         return read_only
 
     def _get_default_field(self, path):
@@ -1362,7 +1362,7 @@ class SampleCollection(object):
         embedded_doc_type=None,
         subfield=None,
         read_only=None,
-        include_private=False,
+        include_private=True,
         leaf=False,
     ):
         """Returns the field instance of the provided path, or ``None`` if one
@@ -1381,7 +1381,7 @@ class SampleCollection(object):
                 :class:`fiftyone.core.fields.Field`
             read_only (None): whether to optionally enforce that the field is
                 read-only (True) or not read-only (False)
-            include_private (False): whether to include fields that start with
+            include_private (True): whether to include fields that start with
                 ``_`` in the returned schema
             leaf (False): whether to return the subfield of list fields
 
@@ -1413,7 +1413,7 @@ class SampleCollection(object):
 
         return field
 
-    def _parse_field(self, path, include_private=False, leaf=False):
+    def _parse_field(self, path, include_private=True, leaf=False):
         keys = path.split(".")
 
         if not keys:
@@ -1573,7 +1573,12 @@ class SampleCollection(object):
             "Subclass must implement get_frame_field_schema()"
         )
 
-    def get_dynamic_field_schema(self, fields=None, recursive=True):
+    def get_dynamic_field_schema(
+        self,
+        fields=None,
+        recursive=True,
+        include_private=False,
+    ):
         """Returns a schema dictionary describing the dynamic fields of the
         samples in the collection.
 
@@ -1585,16 +1590,24 @@ class SampleCollection(object):
                 return dynamic fields. By default, all fields are considered
             recursive (True): whether to recursively inspect nested lists and
                 embedded documents
+            include_private (False): whether to include private fields
 
         Returns:
             a dict mapping field paths to :class:`fiftyone.core.fields.Field`
             instances or lists of them
         """
         return self._get_dynamic_field_schema(
-            fields=fields, recursive=recursive
+            fields=fields,
+            recursive=recursive,
+            include_private=include_private,
         )
 
-    def get_dynamic_frame_field_schema(self, fields=None, recursive=True):
+    def get_dynamic_frame_field_schema(
+        self,
+        fields=None,
+        recursive=True,
+        include_private=False,
+    ):
         """Returns a schema dictionary describing the dynamic fields of the
         frames in the collection.
 
@@ -1606,6 +1619,7 @@ class SampleCollection(object):
                 return dynamic fields. By default, all fields are considered
             recursive (True): whether to recursively inspect nested lists and
                 embedded documents
+            include_private (False): whether to include private fields
 
         Returns:
             a dict mapping field paths to :class:`fiftyone.core.fields.Field`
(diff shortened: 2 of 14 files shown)


swheaton (approved):
I do not know if it meets the usage needs, but code lgtm. Mostly a ton of plumbing and 1 tricky bit to get around mongoengine.

brimoor:
@swheaton thanks for the review!

Given that this is intended as an internal-only feature, I do want to wait for feedback from @ritch and @jacobsela on whether this makes sense to adopt for their use cases or if any changes would be required to effectively adopt this in practice.

There is some tension between hiding private fields from users while also ensuring that we are able to write clean code that uses private fields to do stuff (once they exist).

jacobsela:
@brimoor What are you looking for in this review?

I did a mockup for one-shot classification, and used private fields to store intermediate values (embeddings, logits). I can send you the notebook offline if you are interested. The code behaves as advertised.

My notes in terms of writing code with this:
1. If we use this we have to be very careful about writing cleanup methods that handle these private fields.
2. We are also going to potentially be putting a lot of new data on the user's DB. When we do use these, I think we should lean towards using them as media fields rather than to store raw values.

Actually now that I am thinking about it, due to point (2), I think it could potentially be interesting if we had these at the dataset rather than the sample level. This would allow for storing 1 path to whatever intermediate values a method/run needs for example, rather than at the sample level. This would also allow for entirely hiding runs/methods via a new private field in the dataset document. Is this possible with this code?

brimoor:
> If we use this we have to be very careful about writing cleanup methods that handle these private fields.

yep for sure, although that also applies to any public fields that we add to datasets. Both public and private fields can be deleted via `delete_sample_fields()` and related methods.

> We are also going to potentially be putting a lot of new data on the user's DB. When we do use these, I think we should lean towards using them as media fields rather than to store raw values.

yes we shouldn't store arbitrary amounts of data as sample fields. The specific use case for storing data on samples fields is when you need to be able to filter by these fields via dataset view.

> Actually now that I am thinking about it, due to point (2), I think it could potentially be interesting if we had these at the dataset rather than the sample level. This would allow for storing 1 path to whatever intermediate values a method/run needs for example, rather than at the sample level. This would also allow for entirely hiding runs/methods via a new private field in the dataset document. Is this possible with this code?

[ExecutionStore](https://docs.voxel51.com/plugins/developing_plugins.html#execution-store) can be used to store arbitrary data at the dataset-level.

ritch:
The name "private" is a bit confusing here since that usually refers to access constraints. This seems like a visibility feature to ensure that internal fields do not show up in user facing schemas.

In any case clarifying what these are in doc strings makes sense to me.

> Should we document allow_private=True in more docstrings?

Yes...I would name it `allow_internal` though.

brimoor:
There's already precedent in the SDK for calling fields that start with `_` "private". And there are existing `include_private` parameters throughout the SDK, such as `get_field_schema(include_private=True)`. 

So I think private is the natural naming here, or else we'd need breaking changes to existing parameters to maintain consistent nomenclature.

ritch (approved):

PR #5797, opened 2025-04-29
use optimized values in map_samples id batcher

kaixi-wang:

- use optimized values call in the id batcher to significantly reduce the time splitting datasets into batches
- this is particularly useful when map_samples is used on a full dataset consisting of many dense samples

- for context, on large datasets, batching ids can take >5min. To the user, this looks like nothing happens (no console logs/output/no progress bars) after they run map_samples() for an extended period of time, which is a bad experience and may make them think that it's not going to work.

tested with `ds.map_samples(map_fn, progress="workers", batch_method="id")`, for 2.1M (places) dataset:

```
Retrieved ids in  305.1424562931061 seconds with _enforce_natural_order=True
Retrieved ids in  17.984689235687256 seconds with _enforce_natural_order=False
```

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved consistency when retrieving sample IDs by updating the ordering behavior in certain operations. No changes to visible features or workflows.


minhtuev:
I added some break ppoints and verified that `id_batcher` is using the new values call:
```
In [1]: import fiftyone as fo

In [2]: def map_fcn(sample):
   ...:     labels = [detection.label for detection in sample.ground_truth.detections]
   ...:     return labels
   ...:

In [3]: dataset = fo.load_dataset("quickstart")

In [4]: results = []
   ...: for _, labels in dataset.map_samples(map_fcn, progress="workers", batch_method="id", num_workers=4):
   ...:     results.append(labels)
   ...:
Batch 1/4: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 50/50 [00:00<00:00, 425.97it/s]
Batch 3/4: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 50/50 [00:00<00:00, 401.12it/s]
Batch 2/4: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 50/50 [00:00<00:00, 296.29it/s]
Batch 4/4: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 50/50 [00:00<00:00, 233.73it/s]
Batch 4/4:  42%|███████████████████████████████████████████████████▏                                                                      | 21/50 [00:00<00:00, 196.50it/s]
In [5]: len(results)
Out[5]: 200
```

minhtuev (commented):
LGTM Kacey, but we will need to update unit tests too

```
    def assert_called_with(self, /, *args, **kwargs):
        """assert that the last call was made with the specified arguments.
    
        Raises an AssertionError if the args and keyword args passed in are
        different to the last call to the mock."""
        if self.call_args is None:
            expected = self._format_mock_call_signature(args, kwargs)
            actual = 'not called.'
            error_message = ('expected call not found.\nExpected: %s\n  Actual: %s'
                    % (expected, actual))
            raise AssertionError(error_message)
    
        def _error_message():
            msg = self._format_mock_failure_message(args, kwargs)
            return msg
        expected = self._call_matcher(_Call((args, kwargs), two=True))
        actual = self._call_matcher(self.call_args)
        if actual != expected:
            cause = expected if isinstance(expected, Exception) else None
>           raise AssertionError(_error_message()) from cause
E           AssertionError: expected call not found.
E           Expected: values('id')
E             Actual: values('id', _enforce_natural_order=False)
```

minhtuev (approved):

PR #5810, opened 2025-04-30
[Bug fix] Bug fix for map samples for empty collection

minhtuev:

Currently, for `map_samples`, we do `_sample_collection.first` to get a sample and test if `map_fcn` can operate successfully on the sample, and inspect the returned object and validate that it is not a Fiftyone document. 

When the dataset or view is empty, `first(...)` will raise an exception.

* Manual testing (✅ )
* Unit test (✅ )

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/map/mapper.py
+++ b/fiftyone/core/map/mapper.py
@@ -168,6 +168,8 @@ class Mapper(abc.ABC):
             Iterator[Tuple[bson.ObjectId, R]]: The sample ID and the result of
               the map function for the sample.
         """
+        if len(sample_collection) == 0:
+            return
 
         if check_if_return_is_sample(sample_collection, map_fcn):
             raise ValueError("`map_fcn` should not return Samples objects.")
--- a/fiftyone/core/map/mapper.py
+++ b/fiftyone/core/map/mapper.py
@@ -168,10 +168,15 @@ class Mapper(abc.ABC):
             Iterator[Tuple[bson.ObjectId, R]]: The sample ID and the result of
               the map function for the sample.
         """
-        if len(sample_collection) == 0:
-            return
+        try:
+            result = check_if_return_is_sample(sample_collection, map_fcn)
+        except ValueError as e:
+            if "is empty" in str(e):
+                return
+            else:
+                raise e
 
-        if check_if_return_is_sample(sample_collection, map_fcn):
+        if result:
             raise ValueError("`map_fcn` should not return Samples objects.")
 
         yield from self._map_samples(
--- a/tests/unittests/map/test_mapper.py
+++ b/tests/unittests/map/test_mapper.py
@@ -13,6 +13,8 @@ from unittest import mock
 import bson
 import pytest
 
+import fiftyone as fo
+from fiftyone import ViewField as F
 import fiftyone.core.sample as focs
 import fiftyone.core.map.batcher as fomb
 import fiftyone.core.map.mapper as fomm
@@ -77,9 +79,11 @@ class Mapper(fomm.LocalMapper):
     """Test implementation for abstract class"""
 
     @classmethod
-    def create(cls, *_, **__): ...
+    def create(cls, *_, **__):
+        ...
 
-    def _map_samples_multiple_workers(self, *_, **__): ...
+    def _map_samples_multiple_workers(self, *_, **__):
+        ...
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


minhtuev:
Currently, `map_samples` will throw an error for empty collection:
```
In [1]: def map_fcn(sample):
   ...:     return sample.id
   ...:

In [2]: from fiftyone import ViewField as F

In [3]: import fiftyone as fo

In [4]: dataset = fo.load_dataset("quickstart")

In [5]: view = dataset.filter_labels("predictions", F("confidence") > 1)

In [6]: for _, result in view.map_samples(map_fcn):
   ...:     print(result)
   ...:
StopIteration                             Traceback (most recent call last)
~/workspace/fiftyone/fiftyone/core/collections.py in ?(self)
    947             return next(iter(self))
    948         except StopIteration:
--> 949             raise ValueError("%s is empty" % self.__class__.__name__)

StopIteration:

...

~/workspace/fiftyone/fiftyone/core/collections.py in ?(self)
    945         """
    946         try:
    947             return next(iter(self))
    948         except StopIteration:
--> 949             raise ValueError("%s is empty" % self.__class__.__name__)

ValueError: DatasetView is empty
```

With bug fix:
```
In [1]: import fiftyone as fo

In [2]: from fiftyone import ViewField as F

In [3]: dataset = fo.load_dataset("quickstart")

In [4]: view = dataset.filter_labels("predictions", F("confidence") > 1)

In [5]: def map_fcn(sample):
   ...:     return sample.id
   ...:

In [6]: for _, result in view.map_samples(map_fcn):
   ...:     print(result)
   ...:

In [7]: exit
```

fiftyone/core/map/mapper.py
```diff
@@ -168,8 +168,15 @@ def map_samples(
             Iterator[Tuple[bson.ObjectId, R]]: The sample ID and the result of
               the map function for the sample.
         """
+        try:
+            result = check_if_return_is_sample(sample_collection, map_fcn)
```
minhtuev, line 172: We can do len(sample_collection) to check if a collection is empty, but the call might be expensive for very large datasets.

swheaton: Len for full dataset is fast now. The concern would be with a complex view.
In fact, even getting first() may take a while if the pipeline is complex.
We might want to do something about that, however for this release the fix for empty collection should be fine.
I think a warning but not error seems appropriate, but product may have an opinion @brimoor

swheaton: A validate_output param for example.
Or remove validation and put it lazily on the first sample of the workers' batches

fiftyone/core/map/mapper.py
```diff
@@ -168,8 +168,18 @@ def map_samples(
...
+        except ValueError as e:
+            if "is empty" in str(e):
```
Anddraca, line 174: Thank you for catching this, but we really shouldn't be looking for specific strings in errors. I understand we might not want to raise for any generic `ValueError`...since we're close to a release, can we at least make a followup ticket to address this more specifically?

fiftyone/core/map/mapper.py
```diff
@@ -168,8 +168,18 @@ def map_samples(
...
+                logger.warning(
+                    "Sample collection is empty. No samples to map."
```
brimoor, line 176: I definitely don't want to log a warning about the collection being empty. None of our existing methods log warnings in this case. They are just no-ops. We should follow that pattern here for consistency.

fiftyone/core/map/mapper.py
brimoor: > remove validation and put it lazily on the first sample of the workers' batches

This would be okay with me too. We use `warnings.warn()` elsewhere in cases like this to hopefully ensure that a single version of the log message makes it to the user's shell. But it also wouldn't be the worst thing to see `n` versions of this warning because you are definitely not doing the right thing if your `map_fcn` returns `Sample` objects.

Anddraca: > I would rather just remove this validation tbh. Yes it is an anti-pattern to return Sample objects. 

This was added because not having the validation still leads to an error, but it is unclear why the error is being raised and happens after work starts happening. Nothing indicates to the user that their `map_fcn` can't return a Sample...the assumption was that the lack of guidance on why your map_samples-empowered workflow wasn't running would have been undesirable. 

However, your call on what makes sense here. We can definitely remove it, especially if this is called out in documentation

swheaton: sgtm ... let's remove the validation for this release? And consider adding it lazily (or some other less invasive check) in develop?

brimoor: I'm onboard with @swheaton's proposal 👍

swheaton: In the interest of time for the release, I have opened this PR: https://github.com/voxel51/fiftyone/pull/5813

PR #5814, opened 2025-04-30
Map Samples: Dynamic iteration on samples within worker

j053y:
**This is NOT for the upcoming release**

. This is an idea for how to dynamic control what values get provided to the map function for map samples. 

By default (and in the original implementation) the individual workers just use `iter_samples` to pipe into the map function. This changes allows an `iter_fcn` to be provided to change how the samples get populated. An example is using `.values` to get only the parts of a sample necessary instead of the whole Sample object.

**Note**: The `iter_fcn` would be applied **after** the original sample collection is split into any necessary batches. It is just used in place of `iter_samples` in the various implementations. 

#### TODO
- [ ] Write units tests (other unit tests were modified to ensure original behavior but no new ones were added for the new behavior)

I used the following script to test the changes using a modified call to `values`  to illustrate `iter_fcn` usage.

```py
"""Test mapping different ways"""

import fiftyone as fo
import fiftyone.core.map as focm


DATASET_NAME = "mapper-test"
SAMPLE_COUNT = 1000
NUM_WORKERS = 4

try:
    dataset = fo.Dataset(DATASET_NAME, persistent=True)
    dataset.add_samples(
        [fo.Sample(filepath=f"/img/{i}.png") for i in range(SAMPLE_COUNT)],
        progress=True,
    )
except ValueError:
    dataset = fo.load_dataset(DATASET_NAME)


mappers = [
    focm.MapperFactory.create(method, NUM_WORKERS)
    for method in ("thread", "process")
]

for mapper in mappers:
    for map_fcn, iter_fcn in (
        # Using iter_samples (the default iter_fcn)
        (
            lambda sample: f"I am a sample: {repr(sample)}",
            None,
        ),
        # Needs to return a tuple with the first value being the sample id and
        # the other value being anything. In this case we'll call values on the
        # sample collection to get the id and the filepath.
        (
            lambda values: f"I am the values from a sample: {repr(values)}",
            lambda sample_collection: (
                (value[0], value)
                for value in zip(*sample_collection.values(["id", "filepath"]))
            ),
        ),
    ):
        kwargs = {}

        if iter_fcn is not None:
            kwargs["iter_fcn"] = iter_fcn

        mapped = list(mapper.map_samples(dataset, map_fcn, **kwargs))

        for sample_id, result in mapped:
            print(sample_id, result)
            break


```

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Added support for customizable iteration functions when mapping over samples, allowing greater flexibility in processing collections.

- **Refactor**
  - Improved error handling by reporting errors alongside results during mapping operations.
  - Enhanced progress bar behavior and configuration defaults for mapping tasks.
  - Decoupled iteration logic from mapping logic for more modular and extensible processing.
  - Updated multiprocessing and threading mappers to explicitly handle sample IDs and separate iteration from mapping.

- **Tests**
  - Expanded test coverage to validate both default and custom iteration functions in sample mapping.
  - Updated tests to use the public mapping API for improved reliability.


fiftyone/core/map/process.py
```diff
@@ -246,11 +253,11 @@ def _init_worker(
         process_sample_collection = dataset
 
     process_map_fcn = pickle.loads(map_fcn)
+    process_iter_fcn = pickle.loads(iter_fcn)
```
swheaton, line 256: I was going to comment on pickle and lambdas, but somehow passing a lambda to map_samples() works?? Not really sure how since lambdas should not be pickleable.

swheaton (commented):
idea is sound i think

fiftyone/core/map/mapper.py
```diff
@@ -168,18 +147,45 @@ def map_samples(
             Iterator[Tuple[bson.ObjectId, R]]: The sample ID and the result of
               the map function for the sample.
```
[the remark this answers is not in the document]
j053y: Yes that's exactly right. One of the requirements would be the iterator needs to be formatted to return a tuple of the id and thing you want to map.

fiftyone/core/map/mapper.py
```diff
@@ -168,18 +147,45 @@ def map_samples(
...
 
-        if check_if_return_is_sample(sample_collection, map_fcn):
-            raise ValueError("`map_fcn` should not return Samples objects.")
+            if check_if_return_is_sample(sample_collection, map_fcn):
+                raise ValueError(
+                    "`map_fcn` should not return Samples objects."
+                )
+        else:
+            if save is True:
+                logger.warning("Unable to save when `iter_fcn` is provided")
```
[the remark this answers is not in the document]
j053y: I had a similar thought and I think it's worth investigating in the future. For now It thought it would be to complex to tackle and was just focused on the fetching of data. 

Luckily the map function returns an iterator so those calls could happen outside of this (in batches or not)

kaixi-wang (commented):
I like the idea/direction for optimizing map_samples as resolving the entire fo.Sample was something I see as a problem/limit with the current implementation. 


> Note: The iter_fcn would be applied after the original sample collection is split into any necessary batches. It is just used in place of iter_samples in the various implementations.

Is the main reason not to just batch the values directly and iterate over the results due to values potentially being large? Otherwise, this will essentially still require 2 coll scans and if we could limit it to 1, that would be ideal...

kaixi-wang:
From the example you have, it sounds equivalent to replacing iter samples with select_fields (essentially adding a project stage to the aggregation). ATM, I'm having a hard time coming up with an example of a use case where iter_fn is not just a projection applied over the sample doc in the db. Do you have any specific examples for why we should expose iter_fn? Is there something other than limiting the fields that this unlocks? What if the iter fn returned data that currently doesn't exist on the sample by reading data from another source?

swheaton:
> From the example you have, it sounds equivalent to replacing iter samples with select_fields (essentially adding a project stage to the aggregation). ATM, I'm having a hard time coming up with an example of a use case where iter_fn is not just a projection applied over the sample doc in the db. Do you have any specific examples for why we should expose iter_fn? Is there something other than limiting the fields that this unlocks? What if the iter fn returned data that currently doesn't exist on the sample by reading data from another source?

Select fields is better than getting full sample, but it still brings in all the default fields, that you may or may not need. Plus the overhead of creating the FO and mongoengine objects is still there. If the iter_fn were to call `values()` instead of `iter_samples()`, then maybe it is a bit faster on the read-sample side.

kaixi-wang:
> Select fields is better than getting full sample, but it still brings in all the default fields, that you may or may not need. Plus the overhead of creating the FO and mongoengine objects is still there. If the iter_fn were to call values() instead of iter_samples(), then maybe it is a bit faster on the read-sample side.
yeah I wasn't suggesting we use select_fields() exactly (just the idea of projecting specific fields). The key difference with values() is that values() exhausts the cursor immediately and compiles all values into a list while select_fields I think only actually returns a cursor.

swheaton:
> > Select fields is better than getting full sample, but it still brings in all the default fields, that you may or may not need. Plus the overhead of creating the FO and mongoengine objects is still there. If the iter_fn were to call values() instead of iter_samples(), then maybe it is a bit faster on the read-sample side.
> 
> yeah I wasn't suggesting we use select_fields() exactly (just the idea of projecting specific fields). The key difference with values() is that values() exhausts the cursor immediately and compiles all values into a list while select_fields I think only actually returns a cursor.

It doesn't really matter because in this context we're already inside of a batch so we know we want to exhaust the cursor in any case.

fiftyone/core/map/process.py
brimoor: `pickle` cannot serialize lambdas, which was a total deal-breaker for the entire `map_samples()` project.

Fortunately `dill` **can** serialize lambdas. Back in business!

brimoor:
Do we have some concrete use cases in mind for this? I'm not quite able to interpret the examples I see in the PR description.

Note that this is already possible:

```py
def map_fcn(a, b, c):
    return ...

results = map(map_fcn, *dataset.values(["fieldA", "fieldB", "fieldC"]))
```

and that this retrieves all three field's values in a _single aggregation pipeline_.

(yes `values()` does annoyingly call `list()` internally instead of returning generators; we could separately fix that by adding ` values(..., generator=True)` syntax)

brimoor:
As @kaixi-wang points out, this feels close to using `dataset.select_fields().map_samples()` rather than `dataset.map_samples()`.

The idea of using `select_fields()` judiciously to limit what is loaded into memory is indeed a general FO best practice. We even demonstrate it in the main documentation for `map_samples()` itself:
https://docs.voxel51.com/user_guide/using_datasets.html#map-operations

> As the above snippet shows, you should optimize your iteration by [selecting only](https://docs.voxel51.com/user_guide/using_views.html#efficient-iteration-views) the required fields.

Note that `update_samples()` and `map_samples(..., save=True)` rely on `SaveContext`, which in turn requires us to be iterating over a `SampleCollection` so we can leverage `SaveContext`. So we can't part ways with `Sample` objects without rolling a whole new method for generating bulk writes.

minhtuev:
@brimoor: one of the potential candidate functions discussed for leveraging this new interface is `compute_metadata` since it uses fast values call to retrieve and set the fields.

> Do we have some concrete use cases in mind for this? I'm not quite able to interpret the examples I see in the PR description.
> 
> Note that this is already possible:
> 
> ```python
> def map_fcn(a, b, c):
>     return ...
> 
> results = map(map_fcn, *dataset.values(["fieldA", "fieldB", "fieldC"]))
> ```
> 
> and that this retrieves all three field's values in a _single aggregation pipeline_.
> 
> (yes `values()` does annoyingly call `list()` internally instead of returning generators; we could separately fix that by adding a ` values(..., generator=True)` syntax)

swheaton:
>Note that update_samples() and map_samples(..., save=True) rely on SaveContext, which in turn requires us to be iterating over a SampleCollection so we can leverage SaveContext. So we can't part ways with Sample objects without rolling a whole new method for generating bulk writes.

See above where I propose exactly this ;)

brimoor:
> See above where I propose exactly this ;)

Lol indeed, all credit to @swheaton for the idea. What I meant to imply was that it seems like such extra work would be needed to produce something that has non-trivial overlap with the existing alternate patterns mentioned above.

swheaton:
Oh yeah I'm not saying I think it would be better enough to be worth it, merely ideating since this was a draft PR

j053y:
The internal worker implementation of `map_samples` uses `iter_samples` to fetch the samples to map for each batch. This is replaces that iterator with a custom one.

Anything you do to a dataset/sample collection prior would still end up calling `iter_samples` in each worker. 

I also want to add a version `.values` that is an iterator to help with this.

j053y:
Also it seems like `select_fields` still has more overhead then `values`. I guess I'd be curious why `values` exists/is used internally in many methods instead of the FO best practice of `select_fields`.

swheaton:
> Also it seems like `select_fields` still has more overhead then `values`. I guess I'd be curious why `values` exists/is used internally in many methods instead of the FO best practice of `select_fields`.

If you need samples, `select_fields()` is your best bet. If you only need values, then `values()` should be better because it's a pure Mongo->Python operation as opposed to Mongo->mongoengine->fo.Sample

fiftyone/core/map/mapper.py
```diff
@@ -25,18 +26,19 @@
 import fiftyone.core.config as focc
 import fiftyone.core.map.batcher as fomb
 import fiftyone.core.sample as fos
+import fiftyone.core.utils as fou
 from fiftyone.core.map.typing import SampleCollection
 
 T = TypeVar("T")
 R = TypeVar("R")
+U = TypeVar("U")
```
swheaton, line 34: same for declarations in process.py and threading.py also. This will help us in the future
```suggestion
T = TypeVar("T")  # Sample type
R = TypeVar("R")  # Return value type of map_fcn
U = TypeVar("U")  # Return value type of iter_fcn if set - must be same as input type to map_fcn.
```

brimoor (changes requested):
Requesting changes so we can discuss use cases this afternoon, and, if we decide to move forward, discuss how to document this functionality so end-users know about it too.

brimoor (approved):
Discussed offline and LGTM 👍

PR #5821, opened 2025-05-01
Improve map loading performance 

kaixi-wang:

- improve aggregation to get coordinates by cleaning up the async/sync, removing unnecessary dataset_load if possible, and making it more likely to use an index on the points if it exists.

- tested locally and execution time goes from 1.5 min -> 30 sec for 1 million points when the cache is not used (verified by calling _fetch_geo_points.cache_info() and confirming miss)
image

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of geo point retrieval, ensuring only relevant fields are returned and optimizing data fetching for better performance.
  - Enhanced validation of field paths before processing geo point data requests.
- **Refactor**
  - Streamlined the process for fetching and projecting geo point data, resulting in more efficient and reliable operations.
  - Made view loading more flexible by allowing optional parameters for improved usability.

### DIFF
--- a/fiftyone/server/routes/geo.py
+++ b/fiftyone/server/routes/geo.py
@@ -15,7 +15,6 @@ from starlette.endpoints import HTTPEndpoint
 from starlette.requests import Request
 from starlette.responses import JSONResponse
 
-import fiftyone as fo
 import fiftyone.core.odm as foo
 import fiftyone.server.view as fosv
 from fiftyone.server.decorators import route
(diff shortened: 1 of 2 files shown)


fiftyone/server/routes/geo.py
```diff
@@ -28,49 +28,53 @@
...
-
-    collection = foo.get_async_db_conn()[view._dataset._sample_collection_name]
-
-    agg_results = foo.aggregate(collection, pipeline)
+        pipeline = view._pipeline()
+    else:
+        pipeline = []
+
+    # only return the minimum amount of data needed to minimize network overhead
+    # if there is a regular index on the coordinates, querying this will be faster
```
benjaminpkane, line 62: I don't think this is true. It may read from the index, but it will also read the documents
```suggestion
```

benjaminpkane (dismissed):
LGTM 🚀

benjaminpkane (approved):

PR #5822, opened 2025-05-01
move str casting from mongo agg to python

sashankaryal:

This gave about 20% speed boost in total aggregation time when tested locally with 1m points.

(Details)

        notes for FiftyOne users.

-   [ ] Other


fiftyone/server/routes/geo.py
```diff
@@ -69,8 +68,8 @@ async def _fetch_geo_points(
     results = {}
 
     async for result in agg_results:
-        if result and "sampleId" in result and "coordinates" in result:
-            results[result["sampleId"]] = result["coordinates"]
+        if result and "_id" in result and "coordinates" in result:
```
kaixi-wang, line 71: you don't need this check. Since you're projecting the fields, they will always be there, even if null (but that doesn't seem to matter since you aren't checking the values here either)

sashankaryal: sure. will remove

fiftyone/server/routes/geo.py
```diff
@@ -69,8 +68,8 @@ async def _fetch_geo_points(
     results = {}
 
     async for result in agg_results:
-        if result and "sampleId" in result and "coordinates" in result:
-            results[result["sampleId"]] = result["coordinates"]
+        if result:
+            results[str(result["_id"])] = result["coordinates"]
```
swheaton, line 72: Perhaps though, you mean to ignore results that do not have coordinates?

```python
        if result and result.get("coordinates"):
            results[result["sampleId"]] = result["coordinates"]
```

benjaminpkane:
Makes sense! Included in #5821. Closing this

PR #5828, opened 2025-05-01
Fix/ec ephem clear cache

ritch:

Fixes an issue where `fn.clear_all_caches()` was not clearing ephemeral caches.

Automated tests and in the dashboard cache implementation via the app and dasbhoard plugin.

### DIFF
--- a/fiftyone/operators/cache/decorator.py
+++ b/fiftyone/operators/cache/decorator.py
@@ -265,6 +265,8 @@ def execution_cache(
                 jwt_scoped=jwt_scoped,
                 collection_name=collection_name,
             )
+            print("store", store)
+            print("memory_cache", memory_cache)
             if store:
                 store.delete(cache_key)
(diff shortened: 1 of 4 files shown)


fiftyone/operators/cache/decorator.py
```diff
@@ -300,11 +304,17 @@ def set_cache(*args, **kwargs):
...
 
-        def clear_all_caches(dataset_id=None):
+        def clear_all_caches(dataset_id=None, ctx=None):
+            dataset_id = dataset_id or (ctx.dataset._doc.id if ctx else None)
             if dataset_id is None and func.link_to_dataset:
                 raise ValueError(
-                    "dataset_id must be provided when link_to_dataset=True"
+                    "dataset_id or ctx must be provided when link_to_dataset=True"
                 )
+            if residency == "ephemeral" or residency == "hybrid":
```
minhtuev, line 313: should we define this as enum?

PR #5834, opened 2025-05-01
Added optional `collate_fn` to `TorchImageModel`

jacobsela:

Added and optional `collate_fn` method to `TorchImageModel`. This is useful for handling interesting transforms. For example, we may receive a batch with various image shapes, and may want to dynamically pad them.

Ran `test_apply_model` and `test_torch_hub_feature_extractor` from `fiftyone.tests.intensive.model_tests`
Commented out parts that are not relevant for torch model (i.e. the TF checks), as well as the `AssertRaises` for loading models, as those parts are not relevant for the correctness of the change.

## Release Notes

You can use custom collate functions now with `TorchImageModels`.

        notes for FiftyOne users.

You can use custom collate functions now with `TorchImageModels`. Simply overwrite the `collate_fn` method in `TorchImageModel` and set the property `has_collate_fn` to `True`

-   [x] Other: `fiftyone.utils.torch`

## Summary by CodeRabbit

- **New Features**
  - Added support for custom batch collation functions in Torch-based image models, allowing users to override default data loading behavior.
- **Bug Fixes**
  - Enforced compatibility rules preventing simultaneous use of custom collation functions and ragged batches.

### DIFF
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -745,12 +745,6 @@ def _make_data_loader(samples, model, batch_size, num_workers, skip_failures):
                 return error
 
             try:
-                # collate_fn is just for TorchImageModels
-                # this means that it's fine to only assign
-                # collate_fn here, because other models
-                # use the other branches e.g. use_numpy==True
-                # or if a model has ragged_batches==True
-                # then it shouldn't have a collate_fn
                 if model.has_collate_fn:
                     return model.collate_fn(batch)
                 else:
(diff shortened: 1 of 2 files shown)


fiftyone/utils/torch.py
```diff
@@ -705,6 +743,11 @@ def predict_all(self, imgs):
     def _predict_all(self, imgs):
         if self._preprocess and self._transforms is not None:
             imgs = [self._transforms(img) for img in imgs]
+            if hasattr(self, "collate_fn"):
```
brimoor, line 746: Don't you want `self.has_collate_fn` here?

fiftyone/core/models.py
```diff
@@ -745,7 +745,16 @@ def collate_fn(batch):
                 return error
 
             try:
-                return tud.dataloader.default_collate(batch)
+                # collate_fn is just for TorchImageModels
```
brimoor, line 748: Alternatively, for clarity you could just define `has_collate_fn` and `collate_fn` on the base `Model` class. Then you wouldn't need to explain yourself here.

brimoor: In truth, the accurate statement is that models must implement `TorchModelMixin` to get here in the code. So to be proper, if you don't want to declare `has_collate_fn` and `collate_fn` on `Model`, then you should declare them on `TorchModelMixin`, because that's the class that controls whether this code may get called.

brimoor (changes requested):

fiftyone/utils/torch.py
jacobsela: oh right good catch

fiftyone/core/models.py
jacobsela: Right... I'm adding (not moving) the collate code to `TorchModelMixin` and removing the comment.

jacobsela: Not relevant to this PR but I'd want to rename or remove `TorchModelMixin` once we have a `TorchModel`. Can be renamed to `DataLoaderCompatibleMixin` or something of the sort.

brimoor (approved):
LGTM!

Please note that a `release/v1.6.0` branch already exists, and I believe you are intending this functionality to be used in that release, so you'll need to do one of two things:
1. Update this PR by rebasing its 7 commits onto the `release/v1.6.0` branch
2. Merge this PR into `develop` and then create a new PR where you cherry-pick these 7 commits onto `release/v1.6.0`

jacobsela:
Going with option 2

PR #5846, opened 2025-05-02
Exclude core.models from backend ownership

jleven:
This change removes one file `fiftyone/core/models.py` from backend ownership.

## Summary by CodeRabbit

- **Chores**
  - Updated code ownership rules to explicitly include a specific backend file.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,6 +13,7 @@
 # Backend
 /fiftyone/api/         @voxel51/backend-devs
 /fiftyone/core/        @voxel51/backend-devs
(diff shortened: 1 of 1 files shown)


jacobsela (approved):
Thank you Josh

minhtuev (approved):

swheaton:
Hm I disagree with this @jleven , Models.py is no less core than other files here.

If there's minimal changes, one of us will just approve it. It's not that big of a deal?

If there's ML specific stuff in there, let's take it out and put it elsewhere. But I'm not sure there is

swheaton:
Nvm taking offline

PR #5853, opened 2025-05-05
Add `last_deletion_at` property to datasets

brimoor:
## Change log

- Adds a `Dataset.last_deletion_at` property that is automatically updated when samples are deleted
- `Dataset.last_modified_at` is no longer automatically updated when samples are deleted
- Adds documentation for the behavior of FiftyOne's builtin datetime fields to the user guide
- Updates `_is_full_collection()` to include full views, eg `dataset.view()` and full generated views like `to_patches()`

## Motivation

https://github.com/voxel51/fiftyone/pull/5723 added automatic incrementing of `Dataset.last_modified_at` when samples are deleted. This was released in `fiftyone==1.5.0`.

After reflecting on this implementation in the context of use cases like https://github.com/voxel51/fiftyone-plugins/pull/122, I strongly believe that deletion times should be tracked separately from other dataset metadata changes.

For example, changing a dataset's description and creating a new saved view both increment `Dataset.last_modified_at`, but these have no effect on whether a saved `to_patches()` view needs to be reloaded.

As penance for my decision to change course from how deletion is tracked in `fiftyone==1.5.0`, I've added previously non-existent documentation on automatic datetime fields to the user guide 🙇 

## Summary by CodeRabbit

- **New Features**
  - Added a new property to datasets that records the datetime when a sample was last deleted.
  - Enhanced documentation with a comprehensive guide on builtin datetime fields for datasets and samples, including new examples and usage notes.

- **Bug Fixes**
  - Improved accuracy of dataset modification and deletion timestamps for better tracking.

- **Tests**
  - Updated and expanded tests to verify correct behavior of the new deletion timestamp and collection status features.

- **Documentation**
  - Expanded user guide with detailed explanations and examples for all builtin datetime fields and their update semantics.

### DIFF
--- a/docs/source/user_guide/using_datasets.rst
+++ b/docs/source/user_guide/using_datasets.rst
@@ -990,6 +990,16 @@ a |Sample| that has not been added to a |Dataset|:
     print(sample.id)
     # None
 
+The :meth:`last_deletion_at <fiftyone.core.dataset.Dataset.last_deletion_at>`
+property of a |Dataset| tracks the datetime that a sample was last deleted
+from the dataset:
+
+.. code-block:: python
+    :linenos:
+
+    print(dataset.last_deletion_at)
+    # datetime.datetime(2025, 5, 4, 21, 0, 52, 942511)
+
 .. _using-fields:
 
 Fields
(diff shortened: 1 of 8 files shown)


swheaton:
> For example, changing a dataset's description and creating a new saved view both increment Dataset.last_modified_at, but these have no effect on whether a saved to_patches() view needs to be reloaded.

Couldn't the same be said for insertion? @brimoor

docs/source/user_guide/using_datasets.rst
```diff
@@ -2110,6 +2120,187 @@ some workflows when it is available.
...
+-   when properties such as
+    :attr:`name <fiftyone.core.dataset.Dataset.name>`,
+    :attr:`persistent <fiftyone.core.dataset.Dataset.persistent>`,
+    :attr:`tags <fiftyone.core.dataset.Dataset.tags>`,
+    :attr:`description <fiftyone.core.dataset.Dataset.description>`,
+    :attr:`info <fiftyone.core.dataset.Dataset.info>`, and
+    :attr:`app_config <fiftyone.core.dataset.Dataset.app_config>` are edited
+-   when fields are added or deleted from the dataset's schema
+-   when group slices are added or deleted from the dataset's schema
+-   when saved views or workspaces are added, edited, or deleted
```
swheaton, line 2160: does it also include runs?

swheaton:
Looks alright to me but wondering if we should also have `last_insertion_at` or `last_addition_at`

benjaminpkane (dismissed):

swheaton (changes requested):
requesting changes merely to get an answer to insertion question

brimoor:
> Looks alright to me but wondering if we should also have last_insertion_at or last_addition_at

@swheaton good question. Here's my thinking:
- Insertion sets both `created_at` and `last_modified_at` and both are indexed by default. Therefore:
- `max("created_at")` is a performant way to check for last addition specifically
- `max("last_modified_at")` is a performant way to check for last edit OR addition

Deletion is special because it does not reliably change the min/max statistic of `created_at`/`last_modified_at`.

swheaton (dismissed):
got it thanks. looks good to me then

docs/source/user_guide/using_datasets.rst
brimoor: good call, I initially intended to provide an incomplete list but at this point its close enough to being a full list I should list more things 🙃

brimoor: updated 👍

PR #5861, opened 2025-05-05
Ignore fields not defined in dataset schema

kaixi-wang:

- currently samples with fields not in the dataset schema will throw an error during iter_samples() when the sample instances are created. This can be very confusing and unnecessarily limiting. 
```
FieldDoesNotExist: The fields "{'not_in_dataset_schema'}" do not exist on the document "samples.67d77c98dcc49de74086ebf8"
```
It would be better just to ignore undefined fields. This would also allow "hidden" fields to exist for internal features without needing to declare them on the dataset schema and resolve them during iter_samples().

- add automated tests

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of extra fields not defined in the dataset schema, ensuring they are ignored when loading samples, but still stored in the database.
- **Tests**
  - Added a unit test to verify that undeclared fields are ignored by the sample loading mechanism while remaining in the database.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -8533,6 +8533,7 @@ def _create_sample_document_cls(
 ):
     cls = type(sample_collection_name, (foo.DatasetSampleDocument,), {})
     cls._dataset = dataset
+    cls.STRICT = True
 
     _declare_fields(dataset, cls, field_docs=field_docs)
     return cls
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -8533,7 +8533,6 @@ def _create_sample_document_cls(
 ):
     cls = type(sample_collection_name, (foo.DatasetSampleDocument,), {})
     cls._dataset = dataset
-    cls.STRICT = True
 
     _declare_fields(dataset, cls, field_docs=field_docs)
     return cls
--- a/fiftyone/core/odm/sample.py
+++ b/fiftyone/core/odm/sample.py
@@ -77,7 +77,7 @@ class DatasetSampleDocument(DatasetMixin, Document):
     this class.
     """
 
-    meta = {"abstract": True}
+    meta = {"abstract": True, "strict": False}
 
     _is_frames_doc = False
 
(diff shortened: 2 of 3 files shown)


minhtuev:
Does this cause an issue for map_samples too? Usually I thought we have `skip_failures` param to ignore errors.

brimoor (changes requested):
There is a valid situation where undefined fields could in principle be encountered when iterating over samples:
1. I call `load_dataset()`
2. Now another user adds a new field to that dataset's schema and populates some values on its samples
3. I iterate over my copy of the `dataset` without first calling `dataset.reload()`

However, since that _can indeed_ happen, we already handle it here:
https://github.com/voxel51/fiftyone/blob/524e29c3710381dc853031e9a7ce40f75dadc17c/fiftyone/core/dataset.py#L8113-L8121

Therefore, I claim that if a `FieldDoesNotExist` error is raised, this represents a situation where something has gone wrong that needs immediate action to resolve. We shouldn't just ignore the undefined fields and hope for the best.

Suppose we did this and then some operation uses a `ReplaceOne` operation. I can't say for certain what would happen to any undeclared fields that happen to exist on the `Document` class, which is bad. In my opinion, in order to make a change like this, we would need to do a careful audit of the behavior of non-strict documents.

More importantly though, we'd need a motivating use case to make this change. Because the scenario I described above is already accounted for.

The idea of leveraging this to implement a hidden field functionality is noted. But that would need to be fully prototyped before making this proposed change.

(I personally don't think we should be skirting field declaration and validation; yes it adds computational overhead, but it also adds value. If we are specifically interested in optimizing performance, we should consider other things like ripping out `mongoengine` entirely)

kaixi-wang:
It's not about performance, but rather just annoying when everything else seems to work like evaluate_detections, values(), etc but as soon as I hit a codepath that resolves the full sample instance, I get that error. If it should be treated as an error state, then I would prefer if all reads and writes error out and not just those that call iter_samples.

kaixi-wang:
Ok will convert to draft for now

PR #5672, opened 2025-03-31
optimize fiftyone migrate --info

kaixi-wang:
## What changes are proposed in this pull request?

- improve `fiftyone migrate --info` so that it no longer loops through all datasets to get names and version separately

## How is this patch tested? If it is not, please explain why.

- run `fiftyone migrate --info` and returns almost instantly now (used to hang forever for 10k)

### Is this a user-facing change that should be mentioned in the release notes?

-   [ ] No. You can skip the rest of this section.
-   [ ] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

(Details in 1-2 sentences. You can just refer to another PR with a description
if this PR is part of a larger change.)

### What areas of FiftyOne does this PR affect?

-   [ ] App: FiftyOne application changes
-   [ ] Build: Build and test infrastructure changes
-   [ ] Core: Core `fiftyone` Python library changes
-   [ ] Documentation: FiftyOne documentation changes
-   [ ] Other

## Summary by CodeRabbit

- **New Features**
	- Introduced an enhancement that retrieves all dataset version information in a single, efficient operation.
- **Refactor**
	- Streamlined the process for gathering dataset revisions, reducing overhead and potentially enhancing migration performance.

PR #5868, opened 2025-05-06
Fix datasets revision 

kaixi-wang:

Filter out datasets that have no name when listing to match previous behavior. 
https://voxel51.atlassian.net/browse/FOEPD-1221

- Added dataset with no name and ran get_datasets_revisions()
- added unit and e2e test

        notes for FiftyOne users.

Fix fiftyone migrate command to handle datasets with missing names.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dataset listing and migration processes to exclude database entries without a dataset name, preventing empty or invalid datasets from appearing.
- **Tests**
  - Added comprehensive tests to validate dataset revision retrieval, including handling of missing fields and ensuring correct database query filters.

### DIFF
--- a/tests/unittests/migrations/runner_tests.py
+++ b/tests/unittests/migrations/runner_tests.py
@@ -1,0 +1,65 @@
+"""
+| Copyright 2017-2025, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
+import datetime
+import unittest
+from unittest.mock import patch, MagicMock
+
+from fiftyone.migrations import get_datasets_revisions
+
+
+class TestGetDatasetRevisions(unittest.TestCase):
+    @patch("fiftyone.core.odm.get_db_conn")
+    def test_all_fields_present(self, mock_get_db_conn):
+        mock_conn = MagicMock()
+        mock_conn.datasets.find.return_value = [
+            {"name": "dataset1", "version": "v1"},
+            {"name": "dataset2", "version": "v2"},
+        ]
+        mock_get_db_conn.return_value = mock_conn
+
+        result = get_datasets_revisions()
+        expected = {"dataset1": "v1", "dataset2": "v2"}
+        self.assertEqual(result, expected)
+
+    @patch("fiftyone.core.odm.get_db_conn")
+    def test_missing_version_field(self, mock_get_db_conn):
+        mock_conn = MagicMock()
+        mock_conn.datasets.find.return_value = [
+            {"name": "dataset1"},  # version is missing
+            {"name": "dataset2", "version": "v2"},
+        ]
+        mock_get_db_conn.return_value = mock_conn
+
+        result = get_datasets_revisions()
+        expected = {"dataset1": None, "dataset2": "v2"}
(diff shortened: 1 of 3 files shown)


brimoor (changes requested):
Actually I think we may want to revert or at least discuss the full change in https://github.com/voxel51/fiftyone/pull/5672.

`conn.datasets.find({}, ...)` is **NOT** a drop-in replacement for `fod.list_datasets()`. The latter filters out things like private datasets that back generated views:
https://github.com/voxel51/fiftyone/blob/524e29c3710381dc853031e9a7ce40f75dadc17c/fiftyone/core/dataset.py#L8317-L8333

Moreover, in Teams, `fod.list_datasets()` further scopes the available datasets according to the users permissions.

Directly querying `conn.datasets.find({}, ...)` here bypasses all of that stuff that was intentionally in place.

cc @swheaton to weigh in here.

kaixi-wang:
the change was implemented because the old version took way too long to complete

brimoor:
I understand that neither this PR nor https://github.com/voxel51/fiftyone/pull/5672 changed the behavior of `fod.list_datasets()`. 

My point is that https://github.com/voxel51/fiftyone/pull/5672 **DID** change the behavior of `fiftyone migrate`, which needs to be considered. It wasn't "just" an optimization.

tests/unittests/migrations/runner_tests.py
```diff
@@ -0,0 +1,65 @@
...
+    @patch("fiftyone.core.odm.get_db_conn")
+    def test_empty_collection(self, mock_get_db_conn):
+        mock_conn = MagicMock()
+        mock_conn.datasets.find.return_value = []
+        mock_get_db_conn.return_value = mock_conn
+
+        result = get_datasets_revisions()
+        self.assertEqual(result, {})
+
+    def test_missing_name(self):
```
brimoor, line 50: This test scares me. There's no automatic mechanism to cleanup bad `datasets` docs so if this `finally` block ever fails, eg due to a ctrl + c, then there is forever a bad doc in the user's `datasets` collection.

brimoor (commented):
Okay I slept on my prior review comments and here's where I got:
1. I'm onboard with `fiftyone migrate --info` including private datasets, specifically because I think it is fine and good for `fiftyone migrate --all` to migrate private datasets too, which it was not previously doing
2. Upon further thought, switching from `fo.list_datasets()` to `conn.datasets.find()` will not prevent permissions from being enforced in Teams, so that is okay

I rebased this to `release/v1.5.1` because empty docs in `datasets` collection **does** happen (for reasons I don't fully understand), so we need to regain robustness to that in core methods like `fiftyone migrate` ASAP.

Only reason I didn't approve yet is that `test_missing_name` scares me 👻

tests/unittests/migrations/runner_tests.py
kaixi-wang: tbh I also didn't like having to write an invalid dataset to a real db. I'm happy to remove it.

brimoor (approved):
LGTM

PR #5872, opened 2025-05-07
Do not require read access to admin database

brimoor:
We don't want to **require** read access to the `admin` database in order to use FO with `database_admin=True` and `database_validation=True`.

In order to satisfy that, we must skip automatic FCV validation when the user does not have access to the `admin` database.

See https://voxel51.atlassian.net/browse/FOEPD-1221 for details.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of MongoDB permission errors during feature compatibility version checks to prevent unnecessary warnings.
  - Reduced warning messages related to database access issues, now logging them as debug messages instead.

### DIFF
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -1935,8 +1935,6 @@ def _update_fc_version(client: pymongo.MongoClient):
     if fc_version is None:
         return
 
-    _logger = _get_logger()
-
     if (
         _is_fcv_upgradeable(fc_version, server_version)
         and _db_service is not None
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -298,7 +298,7 @@ def _delete_non_persistent_datasets_if_allowed():
             )
         )
     except:
-        logger.warning(
+        logger.debug(
             "Skipping automatic non-persistent dataset cleanup. This action "
             "requires read access of the 'admin' database"
         )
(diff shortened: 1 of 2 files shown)


fiftyone/core/odm/database.py
```diff
@@ -298,7 +298,7 @@ def _delete_non_persistent_datasets_if_allowed():
             )
         )
     except:
-        logger.warning(
+        logger.debug(
```
brimoor, line 301: Just noting out loud: automatic cleanup of non-persistent datasets that happens on-exit also requires read access to the `admin` database in order to function. And we've set the precedent that if we don't have that, we gracefully continue.

afoley587:
@brimoor - tests can be updated via https://github.com/voxel51/fiftyone/pull/5873

afoley587 (approved):
Thank you!

fiftyone/core/odm/database.py
```diff
@@ -1839,6 +1840,12 @@ def _get_fcv_and_version(
         return current_fcv, server_version
     except ServerSelectionTimeoutError as e:
         raise ConnectionError("Could not connect to `mongod`") from e
+    except:
```
jleven, line 1843: Are we sure we want to catch every error to respond with this message?

brimoor: It's definitely heavy-handed. But the actual error raised here could either be from `pymongo` itself (OSS) or from the pymongo proxy (Teams). IDK how the latter works.

kaixi-wang: Currently the proxy will return a no access error for any database that isn’t the fiftyone database

jleven: I say, let's fix in a follow-up, but let's get this patch release out today.

brimoor: Net-net, we're saying that FCV version checking is an optional service we provide. So I like the generic except here. If some other exception occurred, I wouldn't want to disrupt users either.

PR #5878, opened 2025-05-07
Create iterator version of values

kaixi-wang:
- add iter_values() which allows for processing values during the iteration rather than require exhausting the cursor into a list before starting any subsequent processing 

- https://voxel51.atlassian.net/browse/FOEPD-1224

- Added automated tests for parody with values
- Manually tested iter_values(fields) with different cases to verify iteration and compare final list with output of values(fields) 
 Example:
```
In [12]: for sid, fsid, frame_numbers, filepath, frame_paths in ds.iter_values(["id", "frames._sample_id", "frames.frame_number","filepath", "frames.filepath"]):
    ...:     print(f"sample id ={sid}, distinct sample ids in frames={set(fsid)},\n total, max, last frames={len(frame_numbers)}, {max(frame_numbers)}, {frame_numbers[-1]},\n sample {filepath=},last frame filepath={frame_paths[-1]}, total frame filepaths={len(frame
    ...: _paths)}")
    ...:     break
    ...: 
sample id =67a17dfbc3f95f4b26f4181c, distinct sample ids in frames={ObjectId('67a17dfbc3f95f4b26f4181c')},
 total, max, last frames=361, 361, 361,
 sample filepath='gs://kacey-test/WebUOT-238-Test/data/WebUOT-1M_Test_000003.mp4',last frame filepath=gs://kacey-test/WebUOT-238-Test/data/WebUOT-1M_Test_000003/000361.jpg, total frame filepaths=361


%%time
    ...: vid,vf,vd, vv=[],[],[],[]
    ...: import time
    ...: start=time.time()
    ...: printed=False
    ...: for x, y, z,zz in ds.iter_values(["id", "filepath", "deta_swin_large_o365",'emb_clip-vit-base32-torch'], ):
    ...:     if not printed:
    ...:         print("Time to first result can be processed in iter_values loop:", time.time()-start)
    ...:         printed=True
    ...:     vid.append(x)
    ...:     vf.append(y)
    ...:     vd.append(z)
    ...:     vv.append(zz)
    ...: print(f'Finish iterating in {time.time()}s')
    ...: 
    ...: 
3.0994415283203125e-06 seconds to create make function
aggregate took 0.3728311061859131
_iter_batched_results ['value0', 'value1', 'value2', 'value3']
has_extra_parsing True
Time to first result can be processed in iter_values loop: 0.3759453296661377
Finish iterating in 1746846486.78352s
CPU times: user 4min 17s, sys: 13.5 s, total: 4min 31s
Wall time: 15min 15s

%%time
   ...: vals=ds.values(["id", "filepath", "deta_swin_large_o365",'emb_clip-vit-base32-torch'])
   ...: 
   ...: 
aggregate took 0.3766591548919678
time to list batch result 675.8010149002075
total time to parse 940.7909009456635
CPU times: user 4min 21s, sys: 40.7 s, total: 5min 2s
Wall time: 15min 45s

```

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Introduced a memory-efficient method to lazily iterate over sample values in large collections.

- **Enhancements**
  - Added support for improved aggregation output transformation and lazy evaluation.
  - Enhanced error handling and debugging information during value transformations in aggregations.
  - Optimized vector field data conversion to reduce unnecessary copying.


j053y (approved):
Tested locally and looks good to me. :nut_and_bolt:

fiftyone/core/collections.py
```diff
@@ -9269,9 +9256,94 @@ def values(
             _big_result=_big_result,
             _raw=_raw,
             _field=_field,
+            _lazy=False,
         )
         return self._make_and_aggregate(make, field_or_expr)
 
+    def iter_values(
```
brimoor, line 9263: Can we do `values(..., generator=True)` instead?

For consistency with `add_samples(...,  generator=True)` and other such methods?

fiftyone/core/collections.py
```diff
@@ -9269,9 +9256,94 @@ def values(
...
+        unwind=False,
+        _allow_missing=False,
+        _big_result=True,
+        _raw=False,
+        _field=None,
+        _enforce_natural_order=True,
+    ):
+        """Wrapper for values() that returns an iterator over the values to enable
+        processing data iteratively without loading all values into memory first.
+
```
brimoor, line 9277: When adding a public method to a core class, we must always fully document all `Args`.

As you can see from the other `@aggregation` methods, we also commit to adding example usages to the docstrings.

Additionally, all methods that uses `Aggregation` under the hood need to be decorated with `@aggregation`.

However, in fact this is not a new aggregation, this is just calling `Values` with a new `_lazy=True` parameter. So I think it would be more appropriate to just add a `generator=True` option to `values()` so that we don't need to produce all this duplicate stuff.

fiftyone/core/collections.py
```diff
@@ -10476,8 +10548,9 @@ def aggregate(self, aggregations, _mongo=False):
                 instances
 
         Returns:
-            an aggregation result or list of aggregation results corresponding
-            to the input aggregation(s)
+            The aggregation result(s) corresponding to the input aggregation(s).
+            Returns a single result for a single aggregation, a list of results
+            for multiple aggregations, or a generator if all aggregations are lazy
```
brimoor, line 10553: nit: [style guide](https://github.com/voxel51/fiftyone/blob/develop/STYLE_GUIDE.md#python-style-guide) stays line limit is 79 characters. This last line is too long

I will never give up on pointing this out until/unless the rule is formally changed.

fiftyone/core/expressions.py
```diff
@@ -1859,6 +1859,19 @@ def if_else(self, true_expr, false_expr):
             {"$cond": {"if": self, "then": true_expr, "else": false_expr}}
         )
 
+    def if_null(self, false_expr):
+        """Returns either this expression or ``false_expr`` if this expression is null.
+        This is a shortcut for ``if_else(self.is_null(), false_expr, self)`` and is useful
```
brimoor, line 1864: The code here is not correct, it should read:

```
self.is_null().if_else(false_expr, self)
```

fiftyone/core/expressions.py
```diff
@@ -1859,6 +1859,19 @@ def if_else(self, true_expr, false_expr):
...
+        for replacing null values in a field with a default value.
+
+        Args:
```
brimoor, line 1867: By convention (look at all other methods in this module), all `ViewExpression` methods should contain an example usage in their docstring under an `Examples::` block. Can you please add this?

fiftyone/core/expressions.py
```diff
@@ -1859,6 +1859,19 @@ def if_else(self, true_expr, false_expr):
...
+            false_expr: a :class:`ViewExpression` or MongoDB expression dict
```
brimoor, line 1868: I believe there must be a newline between `Args` and `Returns` block in order for it to render correctly in the docs.

brimoor (changes requested):

fiftyone/core/collections.py
brimoor: In fact if this PR doesn't edit `Values.parse_result()` then I'm not convinced that this can be working correctly. That method does important stuff to correctly parse the values from the db, and it definitely converts to a list right now.

So if you're not calling the method then we haven't reached parity with how `values()` works. And if you aren't modifying the method, then the cursor is getting converted to a list.

I think there's a way to achieve this where you don't need to edit the `SampleCollection.aggregate()` method at all. It doesn't actually touch the mongo cursors for "big" aggregations like this; it just passes them through to `Aggregation.parse_result()`.

kaixi-wang: I couldn’t use parse result as is so I added _iter_batched_results to handle the values extraction efficiently

brimoor: There should be a way to use `map()` inside `parse_result()` so you don't have to reimplement the parsing that `Values.parse_result()` does.

Reimplementing that logic outside of the `Values` class would be a big anti-pattern. All logic about how `Values` aggregations function should be inside that class.

kaixi-wang: Currently parse results expects a list of docs and iterates over the list of all results, extracting big_field. When iterating, this means each input is a single doc. It would be very inefficient to individually extract each big result from each doc using parse results for every iteration of iter values. It's much preferable to extract all big fields from the single doc

brimoor: Right but we need to find a way to apply the logic that `parse_field()` **does** have. 

Your `_iter_batched_results()` implementation bypasses all of that, which means that [`Values._field`](https://github.com/voxel51/fiftyone/blob/deb8b349cfca487289dfe8030733c49aa2d10cc2/fiftyone/core/aggregations.py#L2877-L2881) isn't respected, for example. Which means that data is always returned from the db in "raw" form, it wouldn't do things like convert compressed bytes to `numpy` arrays in `VectorField`s, for example.

If you wrote more thorough unit tests for `iter_values()`, some of the possible usages just wouldn't be working yet.

I do see your point that we need a way to parse multiple "big" aggregations with a *single* iteration over the cursor. 

One solution would be to add an `Aggregation.parse_big_value()` method to the `Aggregation` interface that must be implemented by "big aggregations", ie those for which `Aggregation._has_big_result == True`. This method would be responsible for parsing an **individual value** from a doc emitted by mongo cursor. Your `_iter_batched_results()` method could then call this `parse_big_value()` method for each big aggregation rather than just getting the field value (which is equivalent to what `_raw == True` does).

With this design, the logic for value parsing is nicely encapsulated within the `Aggregation` classes, but it is exposed to `SampleCollection.aggregate()` in a way that it can be invoked when iterating over "big" cursors.

fiftyone/core/collections.py
j053y: We should add this to the GitHub Actions. This type of style check can be easily automated.

brimoor: Ideally to pre-commit hooks, yes. That would make me so happy!

Unfortunately our `black` linter doesn't do comments, only code. I presume there are some subtleties that make auto-formatting comments difficult.

j053y:
> Can we do values(..., generator=True) instead?

If you really want this we can do but it has a bit of a code smell to it in my opinion. Using parameters to change return types is definitely possible but I don't think it's a good pattern for maintainable and readable code. 

Also in a more general sense, functions with a lot of parameters can become difficult to understand and test. Values already has 9, it's not a great pattern to just keep adding parameters which in turn making the functionality more complex.

j053y:
@brimoor Upon further thought, you're right this is product facing decision so I'll remove my reservations about `.values(generator=True)` vs `.iter_values()` and follow whatever you want.

I still stand behind the points I was raising in general sense (for internal methods and such). I got a little caught up in the philosophy of it all and this might not be the right place to have this particular discussion.

fiftyone/core/collections.py
j053y: Yeah `pylint` can do the line-length where `black` falls down

kaixi-wang:
I had tried folding in iter_values() into values(_generator=True), but it breaks everything (even when _generator=False). My earlier version, around[ commit  ](https://github.com/voxel51/fiftyone/pull/5878/commits/e298af7f46106b6891102e3231353a32954d9a77 )show iter_values() just calling values(_stream). I suspect it has something to do with the @aggregation decorator... 
Also, to double check it wasn't just a bug in my earlier implementation, I added _generator param to values in this test pr and you can see that all the tests fail: https://github.com/voxel51/fiftyone/pull/5891

fiftyone/core/collections.py
brimoor: That would be great, please add that!

fiftyone/core/collections.py
kaixi-wang: Thanks for point out _field. I'm working on figuring this out

brimoor:
> @brimoor Upon further thought, you're right this is product facing decision so I'll remove my reservations about `.values(generator=True)` vs `.iter_values()` and follow whatever you want.
> 
> I still stand behind the points I was raising in general sense (for internal methods and such). I got a little caught up in the philosophy of it all and this might not be the right place to have this particular discussion.

@j053y I appreciate the discussion!

I agree that methods with many kwargs can make them complicated. From a design perspective, I see the public methods on `SampleCollection` more like database commands such as `SQL SELECT` than generic methods on a class. In other words, I think it makes sense for users to have methods that correspond 1-1 to semantic operations (like extracting slice(s) of values from the dataset) and thus, if a semantic operation has a variety of possible configuration options, then that will indeed result in a method with a variety of kwargs.

I definitely agree that having a method whose return data type changes based on a parameter value is not ideal. With the benefit of hindsight, I wish we shipped `values()` as a generator from the start, as that would give users the flexibility to deal with the stream of data however they wish. Unfortunately we didn't, so now we have to deal with backwards compatibility. The `generator=True` parameter seems like the least bad of the available options.

From a consistency standpoint, `iter_values()` would match `iter_samples()`, so that does make some sense. But on the other hand, we have `add_samples(..., generator=True)`, so there's prior art in that direction too.

My current thinking is that keeping the concept of values extraction in a single `values()` method wins out among all considerations.

fiftyone/core/collections.py
brimoor: Consider my `parse_big_value()` proposal above 😄 

We should do this in a way where any logic related to the specific parsing that a particular `Aggregation` does is fully encapsulated in that `Aggregation` class and not duplicated in a generic `SampleCollection` method.

kaixi-wang: I realized that list(ds.iter_values()) != to values() because the iteration means the result will be by sample not field. I added a warning to the docstring, but because of this, I feel more strongly to not fold it into values()

kaixi-wang: ok I have added a call to parse_results() in the iter_values() callstack to ensure that if parsing ever changes for values, iter_values() will also be handled. I think what I'm doing is doing something similar to your parse_big_value() suggestion,  however instead of introducing something new, I modified parse_results() itself to return a function  function to transform each of the results when self._lazy. I was hesitant introduce a new public parse_big_value() just for this...

Also added test with a VectorField to ensure that the parsed result of iter_values() matches that of values()

brimoor: > I realized that list(ds.iter_values()) != to values() because the iteration means the result will be by sample not field

I believe you're referring to how `values()` returns per-field slices:

```py
gen1, gen2 = dataset.values(["id", "filepath"], generator=True)

for id, filepath in zip(gen1, gen2):
    pass
```

I think you can use `itertools.tee()` to make that work.

In fact, that would probably be a clean solution to avoid the `parse_big_value()` concept altogether.

kaixi-wang: 🥴 eek I hadn't intended for/considered that the iterator form to be called like that...
In order to have the optimization one would expect (to immediately be able to process items rather than wait for the entire result to be returned first), the immediate return is single object and can't be unpacked into multiple values until you begin iterating. So your snippet would actually throw a value error:
```
 In [1]: import fiftyone as fo

In [2]: ds=fo.load_dataset('quickstart')

In [3]: gen1,gen2=ds.iter_values(['id','filepath'])
ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 gen1,gen2=ds.iter_values(['id','filepath'])

ValueError: too many values to unpack (expected 2)

```

However, this syntax works: 
```
In [4]: ids, filepaths=[],[]
   ...: for i,f in ds.iter_values(['id','filepath']):
   ...:     ids.append(i.upper())
   ...:     filepaths.append(f.upper())
```

kaixi-wang:
Tested with some more complex datasets:
```

def verify_values():
    video = fo.load_dataset('webuot-test')
    fields = ["id", "frames.yolov8s_coco_torch", "frames.gt", "frames.frame_number"]

    start = time.time()
    ids, embeds, frame_gts, frame_numbers = video.values(fields)
    print(f'Total time to get values={time.time() - start}s')

    start = time.time()
    ids_iter, embeds_iter, frame_gts_iter, frame_numbers_iter = [[] for _ in fields]
    for i, (sid, embed, gt, fn) in enumerate(video.iter_values(fields)):
        if len(ids_iter) == 0:
            print(f'first set of values returned in {time.time() - start}s')
        ids_iter.append(sid)
        embeds_iter.append(embed)
        frame_gts_iter.append(gt)
        frame_numbers_iter.append(fn)
    print(f'Iter values done after {time.time() - start}s')

    for i in range(len(video)):
        assert ids_iter[i] == ids[i]
        assert embeds_iter[i] == embeds[i]
        assert frame_gts_iter[i] == frame_gts[i]
        assert frame_numbers_iter[i] == frame_numbers[i]
    print('All values match')
```

```
In [105]: verify_values()
Total time to get values=70.36945509910583s
first set of values returned in 3.590708017349243s
Iter values done after 70.90396094322205s
All values match
```

kaixi-wang:
Also ran with the places dataset (2 million samples with embeddings) and included memory and time stats to highlight the significantly less amount of memory needed to process large datasets and the ability to immediately start processing the data:

```
def get_values(iter=False):
    fields = ["id", "filepath", "deta_swin_large_o365", 'emb_clip-vit-base32-torch']
    i = 0
    start = time.time()
    if not iter:
        for v in ds.values(fields):
            i += 1
            if i % 100_000 == 0:
                print(f'Elapsed time={time.time() - start}, iterations={i}')
    else:
        for v in ds.iter_values(fields):
            i += 1
            if i % 100_000 == 0:
                print(f'Elapsed time={time.time() - start}, iterations={i}')

    print(f'done after {time.time() - start}s and {i} iterations')
```

```

In [21]: track_memory(get_values)
done after 1692.7094440460205s and 4 iterations
Current memory usage: 2228.9 KiB; Peak: 15782711.3 KiB
Total time: 1696.4811840057373s
```
```
In [22]: track_memory(get_values, iter=True)
Elapsed time=55.43256092071533, iterations=100000
Elapsed time=110.87826704978943, iterations=200000
Elapsed time=174.63331604003906, iterations=300000
Elapsed time=304.7547540664673, iterations=400000
Elapsed time=394.4654870033264, iterations=500000
Elapsed time=465.7292308807373, iterations=600000
Elapsed time=513.239795923233, iterations=700000
Elapsed time=541.3452291488647, iterations=800000
Elapsed time=571.4457149505615, iterations=900000
Elapsed time=592.4214651584625, iterations=1000000
Elapsed time=609.1073710918427, iterations=1100000
Elapsed time=633.3143661022186, iterations=1200000
Elapsed time=658.8204748630524, iterations=1300000
Elapsed time=685.8621389865875, iterations=1400000
Elapsed time=712.7943592071533, iterations=1500000
Elapsed time=734.8036069869995, iterations=1600000
Elapsed time=755.9011399745941, iterations=1700000
Elapsed time=776.1638059616089, iterations=1800000
Elapsed time=796.290027141571, iterations=1900000
Elapsed time=821.3187799453735, iterations=2000000
Elapsed time=846.4069650173187, iterations=2100000
done after 861.6529550552368s and 2168460 iterations
Current memory usage: 328.0 KiB; Peak: 51194.1 KiB
Total time: 861.6535680294037s
```

fiftyone/core/aggregations.py
```diff
@@ -167,6 +170,20 @@ def default_result(self):
...
 
+    def _must_transform_output_field_value(self):
+        """Whether the output value of this aggregation must be transformed
+        with parse_result(). Note that this must be called after to_mongo() for
+        the field to be set.
+
+        Returns:
+            True/False
+        """
+        if hasattr(self, "_field"):
```
kaixi-wang, line 181: realized that I can only check this after to_mongo() has been called... so thinking of safer alternative to pass this info without leaking into collections.py

fiftyone/core/odm/database.py
```diff
@@ -394,7 +394,6 @@ def _aggregate(args):
         )
 
     if _stream:
-        # When `unwind` is used, each aggregation runs in its own independent pipeline.
```
kaixi-wang, line 397: address https://github.com/voxel51/fiftyone/pull/5911#discussion_r2089640892

fiftyone/core/collections.py
```diff
@@ -9306,14 +9401,24 @@ def _iter_values(
...
             _field_or_expr = []
             for field in field_or_expr:
-                if "[]" in field:
-                    field = field.replace("[]", "")
-                    logging.warning(
-                        'Single field unwinding "[]" is not '
-                        "supported when using iter_values "
-                        "and will be ignored."
+                if isinstance(field, str) and "[]" in field:
+                    raise ValueError(
```
kaixi-wang, line 9405: address https://github.com/voxel51/fiftyone/pull/5911#discussion_r2089729511

fiftyone/core/collections.py
```diff
@@ -10685,25 +10797,23 @@ def _iter_and_parse_agg_results(self, parsed_aggs, cursor):
         # Non-batchable big aggregations will result a cursor per aggregation
         handle_multiple_cursors = isinstance(cursor, list)
 
-        # Determine if extra parsing is needed for the Aggregation result
-        has_extra_parsing = any(
-            agg._field is not None and not agg._raw
-            for agg in parsed_aggs.values()
-        )
+        transformers = [agg.parse_result(None) for agg in parsed_aggs.values()]
```
kaixi-wang, line 10800: address https://github.com/voxel51/fiftyone/pull/5911#discussion_r2089744967

fiftyone/core/collections.py
```diff
@@ -10793,12 +10903,15 @@ def _parse_aggregations(self, aggregations, allow_big=True):
         stream = None
 
         for idx, aggregation in enumerate(aggregations):
-            # stream is True if all aggregations are lazy
-            if lazy := getattr(aggregation, "_lazy", False):
+            # all aggregations must be lazy to stream
+            if lazy := aggregation._is_lazy:
```
kaixi-wang, line 10907: address https://github.com/voxel51/fiftyone/pull/5911#discussion_r2089751636

fiftyone/core/collections.py
```diff
@@ -10793,12 +10903,15 @@ def _parse_aggregations(self, aggregations, allow_big=True):
...
                 if stream is None:
                     stream = lazy
                 elif stream != lazy:
-                    stream = False
+                    raise ValueError(
+                        "Cannot mix lazy and non-lazy aggregations"
```
kaixi-wang, line 10912: address https://github.com/voxel51/fiftyone/pull/5911#discussion_r2089762574

fiftyone/core/collections.py
```diff
@@ -10651,6 +10759,10 @@ def aggregate(self, aggregations, _mongo=False):
         # Parse batch results
         if batch_aggs:
             if stream:
+                if big_aggs:
+                    raise ValueError(
+                        "Cannot stream with both batch and big aggregations"
```
kaixi-wang, line 10764: address  https://github.com/voxel51/fiftyone/pull/5911#discussion_r2089759540

PR #5893, opened 2025-05-09
Manage ultralytics confidence thresholding

manushreegangwar:

- Enable users to set confidence threshold via `model.config.confidence_thresh` when loading a model 
- Update Ultralytics confidence based on `confidence_thresh` set during `apply_model` (this will override the confidence threshold set at model initialization)

```py
import fiftyone as fo
import fiftyone.zoo as foz

batch_size = 32
model_name_fo = "yolo11l-coco-torch"

#
# Use default confidence
#

model = foz.load_zoo_model(model_name_fo)
dataset.apply_model(model, label_field="predictions_yolo_25")

assert(len(dataset.filter_labels("predictions_yolo_25", F("confidence") < 0.25))==0)
count_025 = dataset.count("predictions_yolo_25.detections")

#
# Use confidence=0.15 at loading
#

model = foz.load_zoo_model(model_name_fo, confidence_thresh=0.15)
dataset.apply_model(model, label_field="predictions_yolo_15")

assert(len(dataset.filter_labels("predictions_yolo_15", F("confidence") < 0.15)) ==0)
count_015 = dataset.count("predictions_yolo_15.detections")

#
# Use confidence=0.1 during apply_model
#

dataset.apply_model(model, label_field="predictions_yolo_1", confidence_thresh=0.1)

assert(len(dataset.filter_labels("predictions_yolo_1", F("confidence") < 0.1)) ==0)
count_010 = dataset.count("predictions_yolo_1.detections")

assert(count_010 > count_015 > count_025)

print(count_010, count_015, count_025) # 51223 42407 33236
```

### DIFF
--- a/fiftyone/utils/ultralytics.py
+++ b/fiftyone/utils/ultralytics.py
@@ -23,8 +23,6 @@ ultralytics = fou.lazy_import("ultralytics")
 torch = fou.lazy_import("torch")
 torchvision = fou.lazy_import("torchvision")
 
-DEFAULT_ULTRALYTICS_CONF = 0.25
-
 
 def convert_ultralytics_model(model):
     """Converts the given Ultralytics model into a FiftyOne model.
@@ -458,13 +456,8 @@ class FiftyOneYOLOModel(fout.TorchImageModel):
             for k, v in config.overrides.items():
                 model.overrides[k] = v
 
-        conf = (
-            config.confidence_thresh
-            if config.confidence_thresh
-            else DEFAULT_ULTRALYTICS_CONF
-        )
         custom = {
-            "conf": conf,
+            "conf": config.confidence_thresh,
             "save": False,
             "mode": "predict",
             "rect": False,
@@ -601,6 +594,9 @@ class FiftyOneYOLOModel(fout.TorchImageModel):
         if self._using_half_precision:
             images = images.half()
 
+        if self.config.confidence_thresh:
+            self._model.predictor.args.conf = self.config.confidence_thresh
+
         output = self._forward_pass(images)
 
         # This is required for Ultralytics post-processing.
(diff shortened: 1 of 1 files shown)


fiftyone/utils/ultralytics.py
```diff
@@ -457,8 +458,13 @@ def _set_predictor(self, config, model):
...
                 model.overrides[k] = v
 
+        conf = (
+            config.confidence_thresh
+            if config.confidence_thresh
+            else DEFAULT_ULTRALYTICS_CONF
+        )
         custom = {
-            "conf": 0.0,
+            "conf": conf,
```
jacobsela, line 467: Can this accept `None`? that way we can just always use the default value in ultralytics without the constant.

fiftyone/utils/ultralytics.py
```diff
@@ -611,6 +617,9 @@ def _predict_all(self, imgs):
         if self.has_logits:
             self._output_processor.store_logits = self.has_logits
 
+        if self.config.confidence_thresh:
+            self._model.predictor.args.conf = self.config.confidence_thresh
```
jacobsela, line 621: does this value only have an effect for preprocessing or also within model inference itself? If the latter is true, it might be worthwhile to set this before `_forward_pass` is called.

jacobsela (commented):
Statically looks fine, left some questions that could potentially be important, let's merge if you are confident about them.

fiftyone/utils/ultralytics.py
manushreegangwar: No, just on postprocessing but I moved it anyway.

brimoor (commented):
@manushreegangwar this should also solve our problem with low confidence keypoints for pose models, right???

brimoor:
@manushreegangwar can you please retarget this at `release/v1.6.0` or promise to merge and then cherry-pick into `release/v1.6.0`?

brimoor (approved):
LGTM

manushreegangwar:
> @manushreegangwar can you please retarget this at `release/v1.6.0` or promise to merge and then cherry-pick into `release/v1.6.0`?

Merging into develop and cherry-picking into release/v1.6.0

PR #5895, opened 2025-05-10
Tweak ultralytics batch inference docs

brimoor:
Clarifies in Ultralytics documentation how to achieve **full** batch inference.

## Summary by CodeRabbit

- **Documentation**
  - Improved and clarified instructions for batch inference with Ultralytics models, including explicit examples for setting batch size and using rectangular resizing.
  - Consolidated and updated code examples for better clarity, removing redundant notes and examples.

### DIFF
--- a/docs/source/integrations/ultralytics.rst
+++ b/docs/source/integrations/ultralytics.rst
@@ -525,9 +525,9 @@ them as kwargs to
 .. code-block:: python
     :linenos:
 
+    # Use rectangular resizing with a batch size of 1
     model = foz.load_zoo_model(model_name, overrides={"rect": True})
-
-    dataset.apply_model(model, label_field="predictions", batch_size=16)
+    dataset.apply_model(model, label_field="predictions", batch_size=1)
 
 .. note::
 
--- a/docs/source/integrations/ultralytics.rst
+++ b/docs/source/integrations/ultralytics.rst
@@ -513,24 +513,6 @@ you can request batch inference via the pattern below:
         batch_size=batch_size,
     )
 
-The manual inference loops can be also executed using batch inference via the
-pattern below:
-
-.. code-block:: python
-    :linenos:
-
-    from fiftyone.core.utils import iter_batches
-
-    filepaths = dataset.values("filepath")
-    batch_size = 16
-
-    predictions = []
-    for paths in iter_batches(filepaths, batch_size):
-        results = model(paths)
-        predictions.extend(fou.to_detections(results))
-
-    dataset.set_values("predictions", predictions)
-
 Alternatively, to run inference on a batch size of 1 with rectangular resizing,
 use overrides as shown below:
 
@@ -549,6 +531,24 @@ use overrides as shown below:
         label_field="predictions",
         batch_size=batch_size,
     )
+
+The manual inference loops can be also executed using batch inference via the
+pattern below:
+
+.. code-block:: python
+    :linenos:
+
+    from fiftyone.core.utils import iter_batches
+
+    filepaths = dataset.values("filepath")
+    batch_size = 16
+
+    predictions = []
+    for paths in iter_batches(filepaths, batch_size):
+        results = model(paths)
+        predictions.extend(fou.to_detections(results))
+
+    dataset.set_values("predictions", predictions)
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


docs/source/integrations/ultralytics.rst
```diff
@@ -494,12 +493,25 @@ Batch inference
 
 When using
 :meth:`apply_model() <fiftyone.core.collections.SampleCollection.apply_model>`,
-you can request batch inference by passing the optional `batch_size` parameter:
+you can request batch inference via the pattern below:
 
 .. code-block:: python
     :linenos:
 
-    dataset.apply_model(model, label_field="predictions", batch_size=16)
```
brimoor, line 502: Better to just document how to **fully** run batch inference.

The old example code was flat out wrong it seems. Doing batch data loading but sequential forward passes doesn't count in my book 🤓

manushreegangwar: It does default to batch inference when passing batched data but it is better to be explicit.

manushreegangwar (approved):
LGTM :+1:

docs/source/integrations/ultralytics.rst
brimoor: The reason I made this PR is that we definitely want users to know how to achieve **full** batch inference, both in data loaders and in the forward pass. There's no value in documenting how to do batch data loading if the model ultimately does forward passes in a loop. It should be all or nothing.

I assumed that was difficult to implement for some reason and thus that `overrides={"batch": batch_size}` would be required for now. Which is suboptimal, but if that's what's required for full inference then that is what the documentation should read (hence this PR).

manushreegangwar: @brimoor In ultralytics, the default batch size is set to `16` ([not set to 1 as it mentions in the ultralytics docs](https://docs.ultralytics.com/modes/predict/#inference-arguments)).

When doing batch inference, we don't want the model `batch` = 1 in Ultralytics model. Anything > 1 is fine because of how the setup/initialization happens for Ultralytics model.

```
# Tell model to use batch inference
model = foz.load_zoo_model( model_name, overrides={"batch": batch_size}, )
```
So the above is not required to run batch inference because the default `batch` > 1 in ultralytics (not optimal if ultralytics default is changed to 1 -- this is when override would be needed).

The ideal solution would be to make sure `batch_size` in `apply_model` is communicated to `model._model` and that in turns updates the `model._model.predictor`. This is currently not in place. As an interim solution, I can set `batch` the same way as `conf` is set in https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/ultralytics.py#L460. If `batch_size` is not set when the model is loaded, Ultralytics default `batch` will be used.

docs/source/integrations/ultralytics.rst
```diff
@@ -494,12 +493,44 @@ Batch inference
...
+    batch_size = 16
+
+    # Tell model to use batch inference
+    model = foz.load_zoo_model(
+        model_name,
+        overrides={"batch": batch_size},
```
manushreegangwar, line 506: Not required for batch inference

docs/source/integrations/ultralytics.rst
brimoor: I see. So you're saying that with this syntax:

```py
dataset.apply_model(
    model,
    label_field="predictions",
    batch_size=batch_size,
)
```

the data loaders will use `batch_size` but the model will use a fixed batch size of 16 internally.

Indeed, **ideally** the two batch strategies would be in-sync, but for documentation purposes I am okay with removing `overrides={"batch": batch_size}` as that will never be something we want users to have to do. For now they won't have control over the model's batch size and once we do this:

> The ideal solution would be to make sure batch_size in apply_model is communicated to model._model and that in turns updates the model._model.predictor.

then batching will be "in-sync" according to the `batch_size` passed to `apply_model()`.

manushreegangwar: The way ultralytics inference is integrated, it bypasses `batch` parameter of `model._model` for **inference** -- `batch_size` tensor is directly passed to ultralytics model `inference` so `batch` has no effect.

`batch_size` and `batch` not being equal is okay but not ideal. This is because there are other places where `batch` is used in `ultralytics` but it is beyond the scope of batch inference -- a non-issue for our integration afaik.

> The ideal solution would be to make sure batch_size in apply_model is communicated to model._model and that in turns updates the model._model.predictor.

The above is still an ideal solution because it avoids any unforeseen issues from arising due to `batch` and `batch_size` not being equal. It is more future-proof.

brimoor: Okay now I get it. What a roller coaster! 

Now I am **definitely** removing `overrides` because it provides no value to `apply_model()` users.

PR #5900, opened 2025-05-12
Do not create sort indexes for grid sorting

benjaminpkane:

The `SortBy` stage creates an index by default. This interferes with a sort that is already valid via a compound index

Server tests

## Release Notes

* Fixed unwanted index creation for grid sorting when a compound index already applies

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sorting behavior to consistently default to ascending order unless specified otherwise.
  - Sorting now explicitly disables index creation for better performance and predictability.

- **Documentation**
  - Updated descriptions to clarify sorting options and behaviors.

- **Tests**
  - Added tests to verify sorting order and parameter handling.

### DIFF
--- a/fiftyone/server/view.py
+++ b/fiftyone/server/view.py
@@ -86,7 +86,7 @@ def get_view(
     reload=True,
     awaitable=False,
     sort_by=None,
-    desc=None,
+    desc=False,
 ):
     """Gets the view defined by the given request parameters.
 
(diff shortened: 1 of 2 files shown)


kaixi-wang (approved):
What! Did not know this, but definitely feel right not to do it automatically by just applying a view...

PR #5901, opened 2025-05-12
Fix ultralytics pickle error when using spawn and forkserver in torch.multiprocessing

manushreegangwar:

UIltralytics `predictor` uses a thread lock variable which causes issues when pickling. This PR fixes the issue by setting the thread lock variable to `None`.

```
import fiftyone as fo
import fiftyone.zoo as foz

model = foz.load_zoo_model("yolo11x-coco-torch")

dataset.apply_model(model, label_field="test_fork", num_workers=8, batch_size=32)
```

```
import torch.multiprocessing as mp
mp.set_start_method("spawn")

import fiftyone as fo
import fiftyone.zoo as foz

model = foz.load_zoo_model("yolo11x-coco-torch")

dataset.apply_model(model, label_field="test_spawn", num_workers=8, batch_size=32)
```

```
import torch.multiprocessing as mp
mp.set_start_method("forkserver")

import fiftyone as fo
import fiftyone.zoo as foz

model = foz.load_zoo_model("yolo11x-coco-torch")

dataset.apply_model(model, label_field="test_forkserver", num_workers=8, batch_size=32)
```

Compared that detection outputs are the same in "test_fork", "test_spawn" and "test_forkserver"

```
# For all three methods (fork, spawn, forkserver), checked model conversion outputs

model = YOLO("yolov8s-seg.pt")
dataset.apply_model(model, label_field="instances", num_workers=8, batch_size=32)
```

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved compatibility with multiprocessing to prevent errors during prediction tasks.
	- Standardized prediction output handling for more consistent results.

### DIFF
--- a/fiftyone/utils/ultralytics.py
+++ b/fiftyone/utils/ultralytics.py
@@ -474,6 +474,9 @@ class FiftyOneYOLOModel(fout.TorchImageModel):
         model.predictor.setup_model(model=model.model, verbose=False)
         model.predictor.setup_source([np.zeros((10, 10))])
         model.predictor.batch = next(iter(model.predictor.dataset))
+        # Remove thread lock to avoid issues with pickle in multiprocessing.
+        model.predictor._lock = None
+
         return model
 
     def _load_model(self, config):
@@ -602,13 +605,6 @@ class FiftyOneYOLOModel(fout.TorchImageModel):
         # This is required for Ultralytics post-processing.
         output["orig_imgs"] = orig_images
         output["imgs"] = images
-
-        if self._output_processor is None:
-            _output = output.get("preds")
-            if isinstance(_output, torch.Tensor):
-                _output = _output.detach().cpu().numpy()
-
-            return _output
 
         if self.has_logits:
             self._output_processor.store_logits = self.has_logits


fiftyone/utils/ultralytics.py
```diff
@@ -474,6 +474,9 @@ def _set_predictor(self, config, model):
         model.predictor.setup_model(model=model.model, verbose=False)
         model.predictor.setup_source([np.zeros((10, 10))])
         model.predictor.batch = next(iter(model.predictor.dataset))
+        # Remove thread lock to avoid issues with pickle in multiprocessing.
```
kaixi-wang, line 477: Is this safe? Usually locks are implemented for a reason...

manushreegangwar: Good question. In our integration, we are relying on process based parallelism. This lock variable is not being used during inference so this PR is simply removing that from the `predictor` object.

jacobsela: I skimmed through the ultralytics code and docs. As far as I can tell the issue is that if multiple threads share their model object they can get race conditions because their inference code modifies the model object in-place. I have no idea why they do that. I believe our usage of it avoids this (@manushreegangwar is this true?).

Since `apply_model`'s actual inference code is single threaded (afaik), I don't believe we would have issues here.

jacobsela (commented):
Assuming the output of using `apply_model` matches direct Ultralytics output for `fork`, `spawn`, and `forkserver` we can merge

fiftyone/utils/ultralytics.py
manushreegangwar: Yes, our integration avoids using multiple threads and we don't modify the model post initialization like Ultralytics does (the batch/dataloader is tied to the predictor in ultralytics, ours is independent of the model). I haven't noticed any issues in the experiments I have run.

manushreegangwar:
> Assuming the output of using `apply_model` matches direct Ultralytics output for `fork`, `spawn`, and `forkserver` we can merge

The outputs for all three methods are the same

jacobsela (approved):

PR #5911, opened 2025-05-14
Add internal helper for processing field values

kaixi-wang:

- Add internal method for getting just the field values

- manual testing + automated tests

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Added support for lazy (streaming) aggregation and value retrieval, enabling efficient iteration over large result sets without loading all data into memory.
  - Introduced new methods for generator-based iteration of field values and streaming aggregation results.
  - Added an expression method to handle null values with a default fallback.

- **Bug Fixes**
  - Improved handling of missing values in aggregation pipelines.

- **Tests**
  - Enhanced test coverage to verify consistency and correctness of lazy value iteration and streaming aggregation features.


j053y (approved):
:nut_and_bolt:

fiftyone/core/collections.py
```diff
@@ -10601,7 +10750,15 @@ def _parse_aggregations(self, aggregations, allow_big=True):
         big_aggs = {}
         batch_aggs = {}
         facet_aggs = {}
+        stream = None
+
         for idx, aggregation in enumerate(aggregations):
+            # stream is True if all aggregations are lazy
+            if lazy := getattr(aggregation, "_lazy", False):
```
brimoor, line 10757: More code smell. `_lazy` is not defined on the `Aggregation` interface.

This usage pattern would ideally be documented on the `Aggregation` class via attributes, like `_has_big_result`, `_is_big_batchable`, etc.

fiftyone/core/collections.py
```diff
@@ -10521,21 +10604,28 @@ def aggregate(self, aggregations, _mongo=False):
...
         # Run all aggregations
-        _results = foo.aggregate(self._dataset._sample_collection, pipelines)
+        _results = foo.aggregate(
+            self._dataset._sample_collection, pipelines, _stream=stream
+        )
 
         # Parse batch results
         if batch_aggs:
+            if stream:
+                return self._iter_and_parse_agg_results(batch_aggs, _results)
```
brimoor, line 10614: Comment on code readability: the early return here is messy. The flow of the `aggregate(aggs)` method is to break up the aggregations into categories, get their results, and insert them into `results` in the correct positions.

What would be *ideal* is to make `aggregate()` support a mix of lazy and non-lazy aggregations. Just like how the current implementation supports a mix of "big", "batchable", and "other" aggregations. It treats them appropriately behind the scenes but the user can still pass an arbitrary list of things and we'll get them done.

However, I guess we cannot do that as "stream" aggregations are lumped together in a single generator that emits `(result1, result2, ...)` tuples, which is a fundamentally different return format than what `aggregate()` provides for non-lazy results.

So, now I'm thinking we should raise an error if a mix of lazy and non-lazy aggregations are provided.

fiftyone/core/collections.py
```diff
@@ -10601,7 +10750,15 @@ def _parse_aggregations(self, aggregations, allow_big=True):
         big_aggs = {}
         batch_aggs = {}
         facet_aggs = {}
+        stream = None
+
         for idx, aggregation in enumerate(aggregations):
+            # stream is True if all aggregations are lazy
```
brimoor, line 10756: Suppose I do this:

```py
results = dataset.aggregate([fo.Values(..., _lazy=True), fo.Values(...)])
```

Since only one aggregation is "lazy", I'll get `stream=False` in this method, which means that the results of `fo.Values(..., _lazy=True)` will get parsed by the "non-lazy" implementation. But the lazy aggregation's `parse_result()` method will return a lambda (because it still thinks it will be used in "lazy" mode), not the values.

So the result will be non-sensical.

brimoor (commented):
Oh, so titling things "internal helper" gets them merged without a full review? 🙃

fiftyone/core/collections.py
```diff
@@ -9269,9 +9256,104 @@ def values(
...
+            _field=_field,
+            _lazy=True,
+        )
+
+        if isinstance(field_or_expr, (list, tuple)):
+            _field_or_expr = []
+            for field in field_or_expr:
+                if "[]" in field:
+                    field = field.replace("[]", "")
+                    logging.warning(
```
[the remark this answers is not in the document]
kaixi-wang: I can remove it. I thought your request was to make it work

kaixi-wang:
Note that we will continue to iterate on the public facing function and tweak implementation as necessary on https://github.com/voxel51/fiftyone/pull/5878

PR #5923, opened 2025-05-16
Optimize compute metadata

kaixi-wang:
- [FOEPD-1197] 
- refactor compute metadata for performance
  - use map samples with values iterator to avoid large values() calls
  - write saves in a separate thread to not block/stall map function execution
- Saw at least 2x performance improvement for the same default config (16 processes, 1000 batch size)
- Progress bars are now smooth and don't stall for every update

- Tested with num workers =1 and default (multiprocessing) on dataset with ~ 200K samples.
- smooth 
```
import fiftyone as fo
ds=fo.load_dataset('webuot-frames-local')

# optionally delete existing metadata to verify result
# ds._sample_collection.update_many({},[{"$set":{"metadata":None}}])

ds.compute_metadata(overwrite=True)

# spot check sample metadata is saved
# ds._sample_collection.aggregate([]).to_list(1)

 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 192644/192644 [26.5s elapsed, 0s remaining, 8.8K samples/s]       

# Optimized, multiprocess result:
CPU times: user 21.9 s, sys: 5.24 s, total: 27.2 s
Wall time: 29.9 s

# Develop branch multiprocessing:
192644/192644 [58.4s elapsed, 0s remaining, 3.6K samples/s]
CPU times: user 57.3 s, sys: 14.6 s, total: 1min 11s
Wall time: 1min 5s

```

- Tested in teams over an api connection with cloud paths, and for the places dataset (2M) , compute_metadata went from 1hr 20min (avg 483/sec) to just 20min (1.9K/sec).

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Refactor**
  - Improved metadata computation with asynchronous bulk writes and multi-threaded processing for better scalability and reliability.
  - Unified metadata update workflows with enhanced progress reporting and robust failure handling.
  - Removed unused internal checks and simplified related logic.

[FOEPD-1197]: https://voxel51.atlassian.net/browse/FOEPD-1197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fiftyone/core/map/mapper.py
+++ b/fiftyone/core/map/mapper.py
@@ -25,7 +25,6 @@ import bson
 
 import fiftyone.core.config as focc
 import fiftyone.core.map.batcher as fomb
-
 import fiftyone.core.utils as fou
 from fiftyone.core.map.typing import SampleCollection
 
--- a/fiftyone/core/map/mapper.py
+++ b/fiftyone/core/map/mapper.py
@@ -25,38 +25,41 @@ import bson
 
 import fiftyone.core.config as focc
 import fiftyone.core.map.batcher as fomb
-import fiftyone.core.sample as fos
+
+# import fiftyone.core.sample as fos
 import fiftyone.core.utils as fou
 from fiftyone.core.map.typing import SampleCollection
 
 T = TypeVar("T")  # Sample type
 R = TypeVar("R")  # Return value type of map_fcn
-U = TypeVar("U")  # Return value type of iter_fcn if set - must be same as input type to map_fcn.
+U = TypeVar(
+    "U"
+)  # Return value type of iter_fcn if set - must be same as input type to map_fcn.
 
 
 logger = logging.getLogger(__name__)
 
 
-def check_if_return_is_sample(
-    sample_collection: SampleCollection[T], map_fcn: Callable[[T], R]
-) -> bool:
-    """
-    Check if the map function returns a sample
-    """
+# def check_if_return_is_sample(
+#     sample_collection: SampleCollection[T], map_fcn: Callable[[T], R]
+# ) -> bool:
+#     """
+#     Check if the map function returns a sample
+#     """
 
-    first_sample = sample_collection.first()
-    if first_sample is None:
-        raise ValueError("Sample collection is empty")
+#     first_sample = sample_collection.first()
+#     if first_sample is None:
+#         raise ValueError("Sample collection is empty")
 
-    # make a copy outside of the db
-    sample_copy = first_sample.copy()
+#     # make a copy outside of the db
+#     sample_copy = first_sample.copy()
 
-    # run the map function on just the copy
-    # if it returns a Sample object
-    if isinstance(map_fcn(sample_copy), fos.Sample):
-        return True
+#     # run the map function on just the copy
+#     # if it returns a Sample object
+#     if isinstance(map_fcn(sample_copy), fos.Sample):
+#         return True
 
-    return False
+#     return False
 
 
 def _get_default_sample_iter(*, save: bool):
@@ -160,9 +163,6 @@ class Mapper(abc.ABC):
         else:
             if save is True:
                 logger.warning("Unable to save when `iter_fcn` is provided")
-                
-        
-
 
         yield from self._map_samples(
             sample_collection,
--- a/fiftyone/core/map/mapper.py
+++ b/fiftyone/core/map/mapper.py
@@ -26,7 +26,6 @@ import bson
 import fiftyone.core.config as focc
 import fiftyone.core.map.batcher as fomb
 
-# import fiftyone.core.sample as fos
 import fiftyone.core.utils as fou
 from fiftyone.core.map.typing import SampleCollection
 
--- a/fiftyone/core/metadata.py
+++ b/fiftyone/core/metadata.py
@@ -12,7 +12,6 @@ import logging
 import multiprocessing.dummy
 import os
 
-import fiftyone.core.odm as foo
 import requests
 from PIL import Image
 
--- a/fiftyone/core/metadata.py
+++ b/fiftyone/core/metadata.py
@@ -7,14 +7,12 @@ Metadata stored in dataset samples.
 """
 import queue
 import threading
-import time
-from collections import defaultdict, deque
+from collections import defaultdict
 import itertools
 import logging
 import multiprocessing.dummy
 import os
 
-from pymongo.errors import BulkWriteError
 from tqdm import tqdm
 
 import fiftyone.core.odm as foo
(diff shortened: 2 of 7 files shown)


fiftyone/core/metadata.py
```diff
@@ -382,28 +389,79 @@ def compute_metadata(
...
-    if num_workers <= 1:
-        _compute_metadata(
-            sample_collection, overwrite=overwrite, progress=progress
-        )
-    else:
-        _compute_metadata_multi(
+    if not overwrite:
+        sample_collection = sample_collection.exists("metadata", False)
+
+    if (sample_count := len(sample_collection)) == 0:
```
brimoor, line 402: Don't compute `len(sample_collection)`! This may cause a full collection scan on views which can take significant time for complex ones on big datasets.

fiftyone/core/metadata.py
```diff
@@ -382,28 +389,79 @@ def compute_metadata(
...
+        logger.info("No samples to compute metadata for")
```
brimoor, line 403: I request that this `if` block be removed. By convention, we don't log "nothing to do" messages in the FO SDK. We just silently return because there was nothing to do, which is a perfectly valid and successful outcome of the method.

fiftyone/core/metadata.py
```diff
@@ -382,28 +389,79 @@ def compute_metadata(
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
     """
-    num_workers = fou.recommend_thread_pool_workers(num_workers)
+    logger.info("Computing metadata...")
```
brimoor, line 392: The previous implementation intentionally only logged this message if there were actually >0 samples that needed metadata computed. I'd like to maintain that behavior. The underlying convention I'd like to follow is to only generate logs in core methods when things actually happen, not when they don't.

fiftyone/core/metadata.py
```diff
@@ -417,6 +475,92 @@ def compute_metadata(
...
+def _compute_metadata_iter_fcn(sample_collection):
+    cache = {}
+
+    for sid, filepath, media_type in sample_collection._iter_values(
+        ["_id", "filepath", "_media_type"], _allow_missing=True
+    ):
+        yield sid, tuple([sid, filepath, media_type, cache])
+
+
+def _compute_metadata_single(
```
brimoor, line 516: If we use `SampleCollection.map_samples()` then none of this code would be necessary because passing `num_workers<=1` to it automatically uses a non-thread/process pool implementation of mapping.

More reasons to use the core abstractions that we build!

fiftyone/core/metadata.py
```diff
@@ -382,28 +389,79 @@ def compute_metadata(
...
+            num_samples=sample_count,
+            skip_failures=skip_failures,
+            warn_failures=warn_failures,
         )
 
+    if not num_workers:
+        num_workers = fou.recommend_thread_pool_workers(num_workers)
+
+    mapper = focm.MapperFactory.create(
+        "process",
```
brimoor, line 420: We added a `default_parallelization_method` config setting specifically so that users could globally configure whether FO should processes or threads to parallelize things in the SDK.

So directly using `"process"` here is a big no-no! We should leverage the core abstractions we provide like `SampleCollection.map_samples()` to get work done and not directly invoke specific backends.

fiftyone/core/metadata.py
```diff
@@ -382,28 +389,79 @@ def compute_metadata(
...
+        sample_collection,
+        _compute_metadata_map_fcn,
+        iter_fcn=_compute_metadata_iter_fcn,
+        skip_failures=skip_failures,
+        progress=progress,
+    )
+
+    collection = sample_collection._root_dataset._sample_collection
+
+    # Collect update operations
```
brimoor, line 433: The next few lines are basically a partial implementation of a `set_values()` function that accepts a generator as input. We should just build that method and then use it here!

If we kept following this pattern, we'd have dozens of implementations of this exact same code throughout the codebase, which would be 😢  Imagine we find a bug or improvement to this code; we want to make that fix in one place, not dozens (or worse, only fix it in one place).

brimoor (changes requested):

fiftyone/core/metadata.py
kaixi-wang: Okay. I only added it because on large datasets it would previously hang with no user feedback of what was happening.

fiftyone/core/metadata.py
kaixi-wang: ok I kept it from josey's original implementation and it seemed that process was always faster, but if there is already a config, I can just use that.

fiftyone/core/metadata.py
kaixi-wang: The goal was to not get into the weeds with set_values() so I went with this. Unlike other places we are computing something and then setting a known field name, so we don't need the flexibility of set_values() or the overhead of all the checks.

fiftyone/core/metadata.py
kaixi-wang: you're right. I missed it in the mapper.py

fiftyone/core/metadata.py
kaixi-wang: Actually, I don't see anywhere else I can log it since all the rest is checked internal to the mapper so I just changed it to debug.

fiftyone/core/metadata.py
kaixi-wang: Note that we had tested it initially with `sample_collection.set_values("metadata", values, key_field="id")` (without modifying set_values) and saw no improvement with just map_samples + values iterator

fiftyone/core/map/mapper.py
```diff
@@ -25,38 +25,41 @@
...
 logger = logging.getLogger(__name__)
 
 
-def check_if_return_is_sample(
-    sample_collection: SampleCollection[T], map_fcn: Callable[[T], R]
-) -> bool:
-    """
-    Check if the map function returns a sample
-    """
+# def check_if_return_is_sample(
```
kaixi-wang, line 43: @j053y idk if we just want to delete this or leave it? If we do leave it, prob should have a comment

fiftyone/core/metadata.py
swheaton: There doesn't seem to be that much about metadata specifically. Perhaps this could be more generic and placed somewhere else. If this is an efficient pattern for setting values, it would be great for any subsequent `map_samples()` user to be able to use also.

fiftyone/core/map/mapper.py
j053y: The comment is where it is used below. I just commented out the method itself as it was causing circular dependency errors with the imports and types it uses.

fiftyone/core/metadata.py
kaixi-wang: To not increase the scope of this PR and because there's a lot more that needs to be considered if handling other fields/user inputs, I've created a separate task to track that work: https://voxel51.atlassian.net/browse/FOEPD-1336

PR #5924, opened 2025-05-16
Add GroupViT to model zoo via transformers integration

manushreegangwar:

This PR adds [GroupViT](https://huggingface.co/docs/transformers/en/model_doc/groupvit) to the model zoo via `transformers` integration

```
import fiftyone as fo
import fiftyone.zoo as foz

# Add semantic segmentation via zoo model using apply_model
model = foz.load_zoo_model("group-vit-segmentation-transformer-torch",
                           text_prompt="a photo of a",
                           classes=["cat", "dog", "other"])
dataset.apply_model(model, label_field="grpvit_seg", batch_size=8)

# Add semantic segmentation via zoo model using model.predict_all
segmenataions_out = model.predict_all(images)
    sample.add_labels(
        segmenataions_out[idx],
        label_field="grpvit_predict_seg",
    )
    sample.save()

# Add semantic segmentation via HF model
from PIL import Image
filepaths = dataset.values('filepath')
images = [Image.open(f) for f in filepaths]

from transformers import AutoProcessor, GroupViTModel
tr_model = GroupViTModel.from_pretrained("nvidia/groupvit-gcc-yfcc", output_segmentation=True)
processor = AutoProcessor.from_pretrained("nvidia/groupvit-gcc-yfcc")

inputs = processor(
    text=["a photo of a cat", "a photo of a dog", "a photo of a other"], images=images, return_tensors="pt",
    padding=True
)
outputs = tr_model(**inputs)

probs = (outputs.segmentation_logits).detach().cpu().numpy()
masks = probs.argmax(axis=1)

segmentation = [fol.Segmentation(mask=mask) for mask in masks]
    sample.add_labels(
        segmentation[idx],
        label_field="hf_grpvit_seg",
    )
    sample.save()
```
Compared outputs of all three methods -- same

-   [x ] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

Adds `"group-vit-segmentation-transformer-torch"` to FiftyOne model zoo

-   [x ] Other

### DIFF
--- a/fiftyone/utils/transformers.py
+++ b/fiftyone/utils/transformers.py
@@ -35,7 +35,6 @@ DEFAULT_SEGMENTATION_PATH = "nvidia/segformer-b0-finetuned-ade-512-512"
 DEFAULT_DEPTH_ESTIMATION_PATH = "Intel/dpt-hybrid-midas"
 DEFAULT_ZERO_SHOT_CLASSIFICATION_PATH = "openai/clip-vit-large-patch14"
 DEFAULT_ZERO_SHOT_DETECTION_PATH = "google/owlvit-base-patch32"
-DEFAULT_ZERO_SHOT_SEGMENTATION_PATH = "nvidia/groupvit-gcc-yfcc"
 
 
 def convert_transformers_model(model, task=None, **kwargs):
@@ -995,13 +994,7 @@ class FiftyOneTransformerForObjectDetection(FiftyOneTransformer):
 class FiftyOneZeroShotTransformerForSemanticSegmentationConfig(
     FiftyOneZeroShotTransformerConfig
 ):
-    def __init__(self, d):
-        if (
-            d.get("name_or_path", None) is None
-            and d.get("model", None) is None
-        ):
-            d["name_or_path"] = DEFAULT_ZERO_SHOT_SEGMENTATION_PATH
-        super().__init__(d)
+    pass
 
 
 class FiftyOneZeroShotTransformerForSemanticSegmentation(
--- a/fiftyone/utils/transformers.py
+++ b/fiftyone/utils/transformers.py
@@ -35,6 +35,7 @@ DEFAULT_SEGMENTATION_PATH = "nvidia/segformer-b0-finetuned-ade-512-512"
 DEFAULT_DEPTH_ESTIMATION_PATH = "Intel/dpt-hybrid-midas"
 DEFAULT_ZERO_SHOT_CLASSIFICATION_PATH = "openai/clip-vit-large-patch14"
 DEFAULT_ZERO_SHOT_DETECTION_PATH = "google/owlvit-base-patch32"
+DEFAULT_ZERO_SHOT_SEGMENTATION_PATH = "nvidia/groupvit-gcc-yfcc"
 
 
 def convert_transformers_model(model, task=None, **kwargs):
@@ -991,6 +992,36 @@ class FiftyOneTransformerForObjectDetection(FiftyOneTransformer):
         self.transforms.return_image_sizes = True
 
 
+class FiftyOneZeroShotTransformerForSemanticSegmentationConfig(
+    FiftyOneZeroShotTransformerConfig
+):
+    def __init__(self, d):
+        if (
+            d.get("name_or_path", None) is None
+            and d.get("model", None) is None
+        ):
+            d["name_or_path"] = DEFAULT_ZERO_SHOT_SEGMENTATION_PATH
+        super().__init__(d)
+
+
+class FiftyOneZeroShotTransformerForSemanticSegmentation(
+    FiftyOneZeroShotTransformer
+):
+    """FiftyOne wrapper around a ``transformers`` model for zero-shot image
+    classification.
+
+    Args:
+        config: a `FiftyOneZeroShotTransformerForSemanticSegmentationConfig`
+    """
+
+    def __init__(self, config):
+        # override output processor
+        if config.output_processor_cls is None:
+            config.output_processor_cls = "fiftyone.utils.transformers.TransformersSemanticSegmentatorOutputProcessor"
+
+        super().__init__(config)
+
+
 class FiftyOneTransformerForSemanticSegmentationConfig(
     FiftyOneTransformerConfig
 ):
@@ -1255,11 +1286,14 @@ class TransformersSemanticSegmentatorOutputProcessor(
     fout.SemanticSegmenterOutputProcessor
 ):
     def __init__(self, *args, **kwargs):
+        self.logits_key = kwargs.pop("logits_key", "logits")
         super().__init__(*args, **kwargs)
 
     def __call__(self, output, image_sizes, confidence_thresh=None):
         return super().__call__(
-            {"out": output.logits},  # to be compatible with the base class
+            {
+                "out": output[self.logits_key]
+            },  # to be compatible with the base class
             image_sizes,
             confidence_thresh=confidence_thresh,
         )
@@ -1293,7 +1327,6 @@ class TransformersDepthEstimatorOutputProcessor(fout.OutputProcessor):
                 )
 
     def __call__(self, output, image_sizes, confidence_thresh=None):
-
         output = self._depth_estimation_post_processor(output)
         output = np.array(
             [o["predicted_depth"].detach().cpu().numpy() for o in output]
@@ -1321,6 +1354,7 @@ MODEL_TYPE_TO_CONFIG_CLASS = {
     "depth-estimation": FiftyOneTransformerForDepthEstimationConfig,
     "zero-shot-image-classification": FiftyOneZeroShotTransformerForImageClassificationConfig,
     "zero-shot-object-detection": FiftyOneZeroShotTransformerForObjectDetectionConfig,
+    "zero-shot-semantic-segmentation": FiftyOneZeroShotTransformerForSemanticSegmentationConfig,
 }
 
 MODEL_TYPE_TO_MODEL_CLASS = {
@@ -1331,4 +1365,5 @@ MODEL_TYPE_TO_MODEL_CLASS = {
     "depth-estimation": FiftyOneTransformerForDepthEstimation,
     "zero-shot-image-classification": FiftyOneZeroShotTransformerForImageClassification,
     "zero-shot-object-detection": FiftyOneZeroShotTransformerForObjectDetection,
+    "zero-shot-semantic-segmentation": FiftyOneZeroShotTransformerForSemanticSegmentation,
 }
--- a/fiftyone/utils/transformers.py
+++ b/fiftyone/utils/transformers.py
@@ -1290,13 +1290,18 @@ class TransformersSemanticSegmentatorOutputProcessor(
         super().__init__(*args, **kwargs)
 
     def __call__(self, output, image_sizes, confidence_thresh=None):
-        return super().__call__(
-            {
-                "out": output[self.logits_key]
-            },  # to be compatible with the base class
-            image_sizes,
-            confidence_thresh=confidence_thresh,
-        )
+        logits = output[self.logits_key].detach().cpu()
+        probs = logits.softmax(dim=1).numpy()
+        masks = probs.argmax(axis=1)
+
+        confidence = probs.max(axis=1)
+        confidence_thresh = confidence_thresh if confidence_thresh else 0
+        conf_mask = confidence > confidence_thresh
+        masks[~conf_mask] = -1
+
+        # Increment class index by 1 since 0 is reserved for background in the app.
+        masks += 1
+        return [fol.Segmentation(mask=mask) for mask in masks]
 
 
 class TransformersDepthEstimatorOutputProcessor(fout.OutputProcessor):
--- a/fiftyone/zoo/models/manifest-torch.json
+++ b/fiftyone/zoo/models/manifest-torch.json
@@ -2891,7 +2891,8 @@
             "default_deployment_config_dict": {
                 "type": "fiftyone.utils.transformers.FiftyOneZeroShotTransformerForSemanticSegmentation",
                 "config": {
-                    "entrypoint_fcn": "transformers.GroupViTModel.from_pretrained",
+                    "name_or_path": "nvidia/groupvit-gcc-yfcc",
+                    "entrypoint_fcn": "transformers.GroupViTMod1el.from_pretrained",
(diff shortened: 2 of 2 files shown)


fiftyone/utils/transformers.py
```diff
@@ -563,9 +563,11 @@ def __init__(self, config):
         if config.entrypoint_fcn is None:
             config.entrypoint_fcn = "transformers.AutoModel.from_pretrained"
         if config.entrypoint_args is None:
-            config.entrypoint_args = {
-                "pretrained_model_name_or_path": config.name_or_path,
-            }
+            config.entrypoint_args = {}
```
jacobsela, line 566: remove this and put it in a different pr

fiftyone/utils/transformers.py
```diff
@@ -991,6 +993,36 @@ def __init__(self, config):
...
+    def __init__(self, config):
+        # override output processor
+        if config.output_processor_cls is None:
+            config.output_processor_cls = "fiftyone.utils.transformers.TransformersSemanticSegmentatorOutputProcessor"
+
+        # Do not default to use_fast = True since AutoProcessor
+        # (CLIPImageProcessorFast) is failing.
+        if not config.transforms_args:
+            config.transforms_args = {}
+        if not config.transforms_args.get("use_fast"):
```
jacobsela, line 1021: Why make this false if it doesn't work for just one model?

fiftyone/utils/transformers.py
```diff
@@ -1255,14 +1287,22 @@ class TransformersSemanticSegmentatorOutputProcessor(
...
-        return super().__call__(
-            {"out": output.logits},  # to be compatible with the base class
-            image_sizes,
-            confidence_thresh=confidence_thresh,
-        )
+        logits = output[self.logits_key].detach().cpu()
+        probs = logits.softmax(dim=1).numpy()
+        masks = probs.argmax(axis=1)
+
+        confidence = probs.max(axis=1)
```
jacobsela, line 1298: I see a couple of options for this:

1. Is the base implementation in `fiftyone.utils.torch.SemanticSegmenterOutputProcessor` is wrong? If so, let's move this there so that all the models get to enjoy this.
2. Is just the previous usage of it here wrong? In that case let's just fix the usage and keep the use of `super().__cal__`.
3. If neither of the above are true let's remove this change

fiftyone/utils/transformers.py
```diff
@@ -991,6 +993,36 @@ def __init__(self, config):
         self.transforms.return_image_sizes = True
 
 
+class FiftyOneZeroShotTransformerForSemanticSegmentationConfig(
+    FiftyOneZeroShotTransformerConfig
+):
+    def __init__(self, d):
```
jacobsela, line 999: We need a default zero shot segmentation model and we need to inject it here in the name_or_path like we do for the other classes.

fiftyone/utils/transformers.py
```diff
@@ -991,6 +993,36 @@ def __init__(self, config):
...
+        super().__init__(d)
+
+
+class FiftyOneZeroShotTransformerForSemanticSegmentation(
```
jacobsela, line 1003: Let's register this class in the model conversion at the top of this file.

jacobsela (changes requested):
Haven't tested yet, just looking over things statically for now:

1. If we are adding a new model class we should add all of the things around it to make it in line with the other types. So add a new default model, expose a top level "zero-shot-image-segmentation-torch" model that uses the default, and register the new class in the model conversion function.
2. I don't like changing common functionality (in this case the `__init__` of `FiftyOneTransformer` and the `__call__` of `TransformersSemanticSegmentatorOutputProcessor`) because of one new model. This kind of change means that either the previous implementation was wrong (in which case we should separately address it), or the new model deviates from the "common" behavior (in which case it doesn't make much sense to change the common code for it). Can you give some clarification as to which case we are in for this change?

fiftyone/utils/transformers.py
manushreegangwar: `transformers` doesn't have a standard `AutoModelForZeroShotSemanticSegmentation` like it does for object detection and image classification. That's why I didn't add it.

fiftyone/utils/transformers.py
manushreegangwar: See my comment above. There is no standard transformer class for this to operate.

fiftyone/utils/transformers.py
manushreegangwar: The base implementation is used by many non `transformer` models in the zoo. I will remove support for confidence thresholding and use `super().__cal__`. If confidence thresholding is needed, that can be done to the base implementation in a separate PR.

fiftyone/utils/transformers.py
manushreegangwar: Why? For easy cherry-picking?

jacobsela: Among other things. We should get in the habit of doing one PR for one concern. As you said, it will make our lives easier when cherry-picking or merging for releases/teams, and makes it so that we can get though changes faster (no need to hold some changes up because of discussions around another).

fiftyone/utils/transformers.py
jacobsela: There are two points here:
1. default model + injection in config: we can add this part regardless
2. How to automatically convert models: I see the issue you are talking about. However, we need to find a solution for this so that `convert_transformers_model` (which is used in `apply_model` when a transformers model is passed in) works as intended for zero shot semantic segmentation models. Currently, the code correctly recognizes the model as zero shot but then turns it into a `FiftyOneZeroShotTransformerForImageClassification` because that's the default for non-detection models.

fiftyone/utils/transformers.py
jacobsela: responded above

fiftyone/utils/transformers.py
manushreegangwar: Here's the PR https://github.com/voxel51/fiftyone/pull/5934

I am all for creating smaller, focused PRs. However, github doesn't handle related PRs as some of the tools so need to be mindful of the merge order: https://github.com/voxel51/fiftyone/pull/5934 should be merged before this.

fiftyone/utils/transformers.py
manushreegangwar: I have handled this a bit differently than what you are suggesting:

1. I exposed "zero-shot-semantic-segmentation-transformer-torch" in the manifest but there is no `DEFAULT` model and users are expected to provide a name_or_path and config kwargs. This is to avoid adding non-standard config to model manifeste. An error is raised when users don't provide a `name_or_path` or `model`
2. We need to update `convert_transformers_model` for models that need specific config settings. Also, there's no support for passing zero-shot classes.
3. Re: _Currently, the code correctly recognizes the model as zero shot but then turns it into a FiftyOneZeroShotTransformerForImageClassification because that's the default for non-detection models._ -- This is because groupvit can also work as a classification model. If you didn't specify the `task`, it will defaults to classification mode.

manushreegangwar: Let's discuss 2 and 3. We should address them in a seprate PR since this extends beyond semantic segmentation.

fiftyone/utils/transformers.py
manushreegangwar: Removed.

manushreegangwar:
@jacobsela can you take a look at this again?

fiftyone/utils/transformers.py
jacobsela: Please remove (1) or set a default model. There definitely shouldn't be top level models that by default give errors.

For 2 and 3, sure.

jacobsela (commented):
Code itself LGTM besides point mentioned on removing the top level `zero-shot-semantic-segmentation-transformer-torch`.

Can you please share the testing script that was used either here or offline?

fiftyone/utils/transformers.py
manushreegangwar: _There definitely shouldn't be top level models that by default give errors._ As long as the usage is properly documented, I don't see this as a problem (some zero-shot model do throw errors by default if no classes are provided). For now, I will remove 1 and revert to the original state where only groupvit is added to the zoo.

There's no default semantic segmentation model that would work with standard config parameters. If we find a semantic segmentation model that can work as a default model, this can be re-introduced.

manushreegangwar:
> Can you please share the testing script that was used either here or offline?

There's code in the PR description. Comparison was done qualitatively in the app and quantitatively using `evaluate_segmentations`.

docs/scripts/make_model_zoo_docs.py
```diff
@@ -221,7 +222,26 @@
     dataset.apply_model(model, label_field="predictions")
 
     session = fo.launch_app(dataset)
-    
+
+{% elif 'zero-shot-semantic-segmentation' in name and 'transformer' in name %}
+
+    # There is no default zero-shot semantic segmentation transformer model.
+    # Model and necessary config parameters should be specificed while loading.
+    model = foz.load_zoo_model(
```
brimoor: Is the code in the PR description still accurate? 

```py
model = foz.load_zoo_model("group-vit-segmentation-transformer-torch",
                           text_prompt="a photo of a",
                           classes=["cat", "dog", "other"])
dataset.apply_model(model, label_field="grpvit_seg", batch_size=8)
```

What you've documented here is more complex

brimoor: Oh I see, PR description code is for `group-vit-segmentation-transformer-torch`. This is for `zero-shot-semantic-segmentation-transformer-torch`.

Can't you add more of these kwargs to the manifest entry for `zero-shot-semantic-segmentation-transformer-torch` so that the usage snippet in the model zoo is concise like `group-vit-segmentation-transformer-torch` is?

manushreegangwar: There are only a handful of semantic segmentation torch models available in tranformers. Unlike detection and classification, `transformers` doesn't have `AutoModelForZeroShotSemanticSegmentation` class or default config parameters for semantic segmentation.

Instead of adding groupvit params as default, I was going to remove `zero-shot-semantic-segmentation-transformer-torch` to avoid non-standard config params. For example, `transforms_args={"use_fast": False}` is needed for groupvit only. Also, `output_processor_args={"logits_key": "segmentation_logits"}, entrypoint_args={"output_segmentation": True},` are not supported by other models.

The other option is to keep this as it is now.

brimoor: My north star is that the end user syntax for `load_zoo_model()` should be as clean as possible. Ideally user provides model names and classes and never any internal configuration details that don't directly impact what the model does to their data.

So in this case, sounds like removing `zero-shot-semantic-segmentation-transformer-torch` is the way to go.

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -2882,6 +2882,84 @@
...
+                "segmentation",
+                "embeddings",
+                "torch",
+                "transformers",
+                "zero-shot"
+            ],
+            "date_added": "2025-05-15 15:20:35"
+        },
+        {
+            "base_name": "group-vit-segmentation-transformer-torch",
```
brimoor, line 2886: If there are multiple GroupViT zero shot segmentation models and they all require the same kwargs encoded here, then perhaps you could document the `name_or_path` parameter in the model zoo entry so that users know they can use it to pick different models:

```py
# Use the default GroupViT model
model = foz.load_zoo_model("group-vit-segmentation-transformer-torch")
dataset.apply_model(...)

# Use a specific model
model = foz.load_zoo_model(
    "group-vit-segmentation-transformer-torch",
    name_or_path="...",
)
dataset.apply_model(...)
```

brimoor: This strategy would be a replacement for `zero-shot-semantic-segmentation-transformer-torch` which is *too generic* to actually be useful with a clean syntax.

If this doesn't cover all the semantic segmentation models in `transformers`, then we could just have new model zoo entries that cover different architectures and (I assume) the corresponding parameter configurations they require.

manushreegangwar: GroupViT only has `"nvidia/groupvit-gcc-yfcc"` documented in https://huggingface.co/docs/transformers/en/model_doc/groupvit#transformers.GroupViTModel.

docs/scripts/make_model_zoo_docs.py
```diff
@@ -221,7 +221,6 @@
     dataset.apply_model(model, label_field="predictions")
 
     session = fo.launch_app(dataset)
-
```
brimoor, line 224: nit: the spaces here are literally injected into the docs, so I think this change should be reverted

brimoor: @manushreegangwar I haven't built the docs locally myself, so I want to confirm: have you built the docs locally and confirmed that the entry for `group-vit-segmentation-transformer-torch` looks as expected?

manushreegangwar: image

Reverted. Groupvit shows up as this but I am going to add a "groupvit" tag and implement special handling for this tag so `text_prompt` is shown in the docs (same a clip). It works without the "text_prompt" but the masks are not as expected for the classes.

```
model = foz.load_zoo_model("group-vit-segmentation-transformer-torch",
                           text_prompt="a photo of a",
                           classes=["cat", "dog", "other"])
```

docs/scripts/make_model_zoo_docs.py
```diff
@@ -180,7 +181,6 @@
 
     embeddings = dataset.compute_embeddings(model)
 {% elif 'zero-shot-classification' in name and 'transformer' in name %}
-
```
manushreegangwar, line 183: This is for consistency. newline is not required here and below.

docs/scripts/make_model_zoo_docs.py
manushreegangwar: Instead of adding a new tag, I used "group-vit" in name since tags are reserved for a group of models.

brimoor (commented):
My comments have been addressed. If the built model zoo docs look good then I'm onboard 👍 

Defer to @jacobsela for final approval

manushreegangwar:
> My comments have been addressed. If the built model zoo docs look good then I'm onboard 👍
> 
> Defer to @jacobsela for final approval

image

@jacobsela, can you review this PR again?

brimoor (approved):

Issue #5917, opened 2025-05-16
[FR] Allow callable frames_patt argument


zanocrate:
### Proposal Summary

When extracting `FrameView`s from a video dataset, we can pass an optional `frames_patt` argument which determines the filename of the extracted image. I would like a more flexible way to determine the filename of each frame by allowing an arbitrary callable input which takes the Frame object of the relative frame (in my use case, I populated these with custom frame-level fields I wish to use to build the filename).

### Motivation

The main reason I would suggest implement doing this is the integration with CVAT annotations; here is the specific problem I encountered:

1. I had a video dataset I filtered using `.match_frames()` with some frame-level field filter.
2. Next, I turned this into a `FrameView` using `.to_frames(sparse=True)`; so far so good, and all the extracted frames spawn at the provided output directory, with names like `/path/to/video/000001.jpg`.
3. Now, the next step in my workflow is to send these frames to CVAT for annotation, by doing `frame_view.annotate()`. When I first tried this, my script died without any informative output.
4. By inspecting the CVAT logs, I found out that it couldn't accept the request as it found identical filenames, due to possibly different videos sampling the same frame number, thus resulting in the same exact filename which CVAT does not like. 
5. Thus the feature request, which is the first thing that came to my mind to solve this issue.

### What areas of FiftyOne does this feature affect?

-   [ ] App: FiftyOne application
-   [x ] Core: Core `fiftyone` Python library
-   [ ] Server: FiftyOne server

### Willingness to contribute

The FiftyOne Community welcomes contributions! Would you or another member of your organization be willing to contribute an implementation of this feature?

-   [ ] Yes. I can contribute this feature independently
-   [x ] Yes. I would be willing to contribute this feature with guidance from the FiftyOne community
-   [ ] No. I cannot contribute this feature at this time

brimoor:
Hi @zanocrate 👋 

Your use case of annotating frames views makes sense and should definitely be supported!

I've fixed this in https://github.com/voxel51/fiftyone/pull/5927 by tweaking our CVAT integration to support creating tasks that contain duplicate filenames. This will be included as part of `fiftyone==1.6.0` which is scheduled for release in 1-2 weeks.

For example, this will now work:

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video")

view = dataset.limit(2).to_frames(sample_frames=True)

# now works!
results = view.annotate("test", label_field="detections", launch_editor=True)
```

[...]

PR #5927, opened 2025-05-17
Avoid duplicate filenames when uploading to CVAT

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/5917

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.limit(2).to_frames(sample_frames=True)

# now works!
results1 = view.annotate("test1", label_field="detections", launch_editor=True)

# this also works
results2 = view.annotate(
    "test2",
    label_field="detections",
    launch_editor=True,
    task_size=120,
)
```

```py
results1.cleanup()
results2.cleanup()
```

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -4356,22 +4356,24 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
         files = {}
         open_files = []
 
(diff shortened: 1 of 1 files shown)


swheaton (approved):
lgtm

PR #5939, opened 2025-05-20
Update default num workers for torch

jacobsela:

Removed num workers being set to 0 on Mac.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved worker recommendation logic on macOS by removing the restriction that set the number of workers to zero, allowing better parallelism on macOS systems without CUDA.

### DIFF
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -1634,13 +1634,6 @@ def recommend_num_workers():
         # https://stackoverflow.com/q/20222534
         return 0
 
-    if sys.platform == "darwin" and not torch.cuda.is_available():
-        # There is a parallelism bug on macOS with CPU that prevents us from
-        # using `num_workers > 0`
-        # https://stackoverflow.com/q/64772335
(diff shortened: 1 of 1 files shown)


kaixi-wang (approved):
lgtm

also checked the pytorch github issue, and the problem seems to be macs with python <3.8 and we don't support 3.8 or lower anyways

brimoor (approved):
In order to ship this, we need to be sure that `apply_model()` with data loaders works on `darwin` with both `spawn` and `fork` multiprocessing contexts (because some FO methods like `map_samples()` will implicitly set the default context to `fork`).

Confirmed with the code below that this is true, so LGTM 👍 

```py
import multiprocessing

import fiftyone as fo
import fiftyone.zoo as foz

# optional: force the default multiprocessing context to 'fork' on darwin
_ = fou.get_multiprocessing_context()
print(multiprocessing.get_start_method())

    "cifar10",
    split="test",
    max_samples=100,
    shuffle=True,
)

dataset.apply_model(model, label_field="predictions", batch_size=8, num_workers=8)
```

PR #5942, opened 2025-05-21
fix(secrets.py): resolve attribute collisions

CharlesCNorton:
* **PluginSecretsResolver:** rename backing field `client` → `_client`; expose via read-only `@property client` to eliminate name clash  
* **SecretsDictionary:** rename custom `__dict__()` → `to_dict()` to restore normal reflection  

Call-site updates: use `resolver.client` (not `resolver.client()`) and `secrets.to_dict()`.

---

This PR removes two silent attribute collisions that could lead to runtime errors and broken introspection:

* **PluginSecretsResolver**
  * Stores the secrets client in `_client` instead of `client`.
  * Provides a property `client` that returns `_client`, ensuring `resolver.client` is always an `ISecretProvider` and never a method.

* **SecretsDictionary**
  * Replaces a custom `__dict__()` method with `to_dict()`, restoring Python’s built-in `obj.__dict__` so debuggers, serializers, and `vars()` work normally.

No functional behaviour changes beyond these naming fixes.

---

No new automated tests were added. The change is purely structural: it removes name collisions without altering logic paths. Correctness was verified by code inspection (attribute resolution order) and by confirming existing unit tests continue to pass locally.

---

- [ ] No. You can skip the rest of this section.  
- [x] Yes. Give a description of this change to be included in the release
        notes for FiftyOne users.

**Release-note blurb:**  
Fixed internal attribute collisions in plugin secret handling. `PluginSecretsResolver.client` is now a property (access without parentheses), and `SecretsDictionary` exposes `to_dict()` instead of overriding `__dict__()`.

- [ ] App
- [ ] Build
- [x] Core
- [ ] Documentation
- [ ] Other

### DIFF
--- a/fiftyone/plugins/secrets.py
+++ b/fiftyone/plugins/secrets.py
@@ -23,7 +23,7 @@ class PluginSecretsResolver:
     def __new__(cls):
         if cls._instance is None:
             cls._instance = super(PluginSecretsResolver, cls).__new__(cls)
-            cls._instance.client = _get_secrets_client()
+            cls._instance._client = _get_secrets_client()
         return cls._instance
 
     def register_operator(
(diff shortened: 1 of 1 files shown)


kaixi-wang (changes requested):
We specifically want to override the python default dict to not print values as in teams, this contains database secrets

jacobsela:
@CharlesCNorton For future reference, this kind of backend code is outside of the concerns of the ML team. I've added @kaixi-wang as a reviewer for this PR as a code owner.

PR #5944, opened 2025-05-21
labels-correspondence app follow-up work

sashankaryal:

Compliments @brimoor's https://github.com/voxel51/fiftyone/pull/5918

### Changelog

- Instance IDs now render distinct overlay indices if `.index` is missing
- Update tooltip from “Select similar” to “Select all”
- Stable "Color by instance" mode (was flickering previously because of missing index and no support for `instance._id`

Locally

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Improved modal management with a unique identifier for enhanced cross-state synchronization.
- **Enhancements**
  - Detection overlays now display more precise label indices, including support for instance-based indexing.
  - Tooltip shortcut text updated from "Select similar" to "Select all" for improved clarity.
  - Label hash generation now includes instance IDs for greater uniqueness.
  - Reduced port collision risk in end-to-end testing by adding random offset to server port assignment.
- **Refactor**
  - Internal state handling streamlined for modal and overlay components.

### DIFF
--- a/e2e-pw/src/oss/fixtures/index.ts
+++ b/e2e-pw/src/oss/fixtures/index.ts
@@ -28,7 +28,12 @@ const customFixtures = base.extend<object, CustomFixturesWithoutPage>({
         return;
       }
 
-      await use(3050 + workerInfo.workerIndex + workerInfo.parallelIndex);
+      // random number [0, 99] to avoid port collisions (rare edge case)
+      const rand = Math.floor(Math.random() * 100);
+
+      await use(
+        3050 + workerInfo.workerIndex + workerInfo.parallelIndex + rand
+      );
     },
     { scope: "worker" },
   ],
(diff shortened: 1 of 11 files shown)


brimoor (changes requested):
@sashankaryal everything looks great except one (hilarious) thing: we now have Mario Kart in FiftyOne! Watch out for the vehicle with the star powerup 🤣🤣🤣 

https://github.com/user-attachments/assets/a04aaceb-9d0e-4fb5-9344-91c17ea1feb9

^this is the `quickstart-video` dataset **before** running `foul.index_to_instance()`.

Looks like `vehicle 0` is getting rendered as `vehicle`, which causes color by instance mode to bug out.

Here's how I think it should work:

When the dataset has `instance` values populated:
- Use **1-based indexing** to generate overlay numbers, both for video and group datasets
- Any object with `instance=None` is treated as a 1-of-1 object and thus gets the Mario Kart treatment in "Color by instance" mode

When the dataset has `index` values populated:
- The original `index` values are preserved, including when `index=0`, which is treated as a valid "trajectory"
- Any objects with `index=None` are given the Mario Kart as with `instance=None`

e2e-pw/src/oss/fixtures/index.ts
```diff
@@ -28,7 +28,12 @@ const customFixtures = base.extend<object, CustomFixturesWithoutPage>({
         return;
       }
 
-      await use(3050 + workerInfo.workerIndex + workerInfo.parallelIndex);
+      // random number [0, 99] to avoid port collisions (rare edge case)
```
tom-vx51, line 31: 🤢 🤷

tom-vx51 (approved):

PR #5957, opened 2025-05-26
Map samples: use `deque`, `reload()`, and `fo.config`

brimoor:
## Change log

A small collection of enhancements to the `map_samples()` + `update_samples()` backends:

- Use `collections.deque` to exhaust generators in `update_samples()` for efficiency
- Add `reload()` calls when using multiprocessing so that schema changes are reflected in the main process, for consistency with the user experience when `num_workers=0` or `parallelize_method="thread"` are used
- Use global `fo.config` rather than loading a separate copy from disk, for consistency with the rest of the library 

## Tested by

- Unit tests added

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

# update_samples()

def update_fcn(sample):
    sample["foo"] = "bar"

# Empty view
dataset.limit(0).update_samples(update_fcn)
assert not dataset.has_field("foo")

# Full dataset
dataset.update_samples(update_fcn)
assert dataset.has_field("foo")

# map_samples()

def map_fcn(sample):
    sample["spam"] = "eggs"
    return "eggs"

# Empty view
for _ in dataset.limit(0).map_samples(map_fcn, save=True):
    pass

assert not dataset.has_field("spam")

# Full dataset
for _ in dataset.map_samples(map_fcn, save=True):
    pass

assert dataset.has_field("spam")
```

## Summary by CodeRabbit

- **New Features**
  - Improved synchronization of schema changes when processing samples in parallel, ensuring updates made during mapping or updating are accurately reflected in the main dataset.

- **Bug Fixes**
  - Addressed issues where schema edits in worker processes were not visible in the main process after parallel operations.

- **Tests**
  - Enhanced and expanded test coverage for parallel sample processing, including new tests for public interface methods and improved dataset setup and cleanup.

- **Chores**
  - Updated internal imports and configuration handling for improved maintainability and consistency.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -6,7 +6,7 @@ Interface for sample collections.
 |
 """
 
-from collections import defaultdict
+from collections import defaultdict, deque
 from copy import copy
 from datetime import datetime
 import fnmatch
@@ -4067,6 +4067,10 @@ class SampleCollection(object):
             skip_failures=skip_failures,
         )
 
+        # Sync any schema edits from workers to main process
+        if save and isinstance(mapper, focm.ProcessMapper):
+            self.reload()
+
     def update_samples(
         self,
         update_fcn,
@@ -4135,14 +4139,19 @@ class SampleCollection(object):
             parallelize_method, num_workers, batch_method, batch_size
         )
 
-        for _ in mapper.map_samples(
+        generator = mapper.map_samples(
             self,
             update_fcn,
             progress=progress,
             save=True,
             skip_failures=skip_failures,
-        ):
-            ...
+        )
+
+        deque(generator, maxlen=0)
+
+        # Sync any schema edits from workers to main process
+        if isinstance(mapper, focm.ProcessMapper):
+            self.reload()
 
     def rename_evaluation(self, eval_key, new_eval_key):
         """Replaces the key for the given evaluation with a new key.
--- a/fiftyone/core/map/factory.py
+++ b/fiftyone/core/map/factory.py
@@ -1,5 +1,6 @@
 """
-Factory for mapping backends
+Factory for mapping backends.
+
 | Copyright 2017-2025, Voxel51, Inc.
 | `voxel51.com <https://voxel51.com/>`_
 |
@@ -8,8 +9,7 @@ Factory for mapping backends
 import multiprocessing
 from typing import Dict, List, Optional, Type
 
-
-import fiftyone.core.config as focc
+import fiftyone as fo
 import fiftyone.core.map.batcher as fomb
 import fiftyone.core.map.mapper as fomm
 import fiftyone.core.map.process as fomp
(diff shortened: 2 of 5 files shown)


fiftyone/core/map/factory.py
```diff
@@ -70,7 +66,7 @@ def create(
         if mapper_key is None:
             # If the default parallelization method is set in the config, use
             # it no matter what.
-            mapper_key = config.default_parallelization_method
+            mapper_key = fo.config.default_parallelization_method
```
brimoor, line 69: This is for consistency with the rest of the library, because the following works in other cases:

```py
import fiftyone as fo

# stuff here...

fo.config.parallelize_method = "thread"

# Should respect runtime edit to `parallelize_method`

dataset.map_samples(...)
```

tom-vx51 (approved):

fiftyone/core/collections.py
```diff
@@ -4135,14 +4139,19 @@ def update_fcn(sample):
...
             self,
             update_fcn,
             progress=progress,
             save=True,
             skip_failures=skip_failures,
-        ):
-            ...
+        )
+
+        deque(generator, maxlen=0)
```
j053y, line 4150: This is a fun trick, but is it actually making things better? 

It seems like just another import or something a developer would have to look up if not familiar?

j053y: Just ran some tests locally and looping and deque (with a generator that's feeding 1,000,000 values at a steady rate) seems to perform almost the exactly the same. Sometimes loop is faster, and sometimes deque is faster depending on the run, but neither seems significantly better than the other.

brimoor: The reason I made this change is that @tom-vx51 specifically chose to use `deque(map_samples(fcn, save=True)` over `update_samples(fcn)`, and so, as Michael Jordan would say, I took it personal and had to tweak `update_samples()` immediately so it would suit his fancy next time 😄 

Trinity also used it in a few places in the codebase in the past. So I figured with these two as evidence, better make the change!

tom-vx51: You're all correct to question the net effect; if the whole point of this code is to provide a performance benefit, then it makes sense to me to use the optimal patterns to achieve that. In tight loops, the difference can be 10-20%. Admittedly, this will rarely be a tight loop, and so the iteration itself is unlikely to have a noticeable impact. We can say for sure that it won't ever be slower, and we already have a precedent of using it elsewhere https://github.com/voxel51/fiftyone-teams/blob/develop/fiftyone/operators/executor.py#L425-L427

swheaton: Yep exactly and note the comment which was my suggestion above
```python
                # Fastest way to exhaust sync generator, re: itertools consume()
                #   https://docs.python.org/3/library/itertools.html
                collections.deque(result, maxlen=0)
```

PR #5963, opened 2025-05-28
More docs for linking/selecting labels

brimoor:
Follow-up to https://github.com/voxel51/fiftyone/pull/5959 that adds some additional documentation around selecting labels/instances in the App.


sashankaryal (approved):
gracias

PR #5978, opened 2025-06-03
add support for completely custom plotly tooltip

imanjra:

(Details)

        notes for FiftyOne users.

-   [ ] Other


app/packages/components/src/components/Plot/Plotly.tsx
```diff
@@ -1,9 +1,50 @@
-import { useTheme } from "@fiftyone/components";
+import { Box, useTheme } from "@mui/material";
 import { merge } from "lodash";
 import React, { useMemo } from "react";
 import Plot, { PlotParams } from "react-plotly.js";
+import PlotlyTooltip, { TooltipValue } from "./PlotlyTooltip";
 
 export default function EvaluationPlot(props: EvaluationPlotProps) {
```
ritch, line 7: nit:

Rename this to match the file

app/packages/utilities/src/type-check.ts
```diff
@@ -18,6 +18,23 @@ export function isObjectIdString(value: string, strict = true) {
   return isHex(value) && strict ? value.length === 24 : value.length <= 24;
 }
 
+export function isFunctionalComponent(
```
ritch, line 21: This seems out of the ordinary...

1. it is named in a way that makes me think only things like `const MyComp = () => <h1>Hello</h1>` would return `true` ...I wouldn't expect `useEffect` to return `true`. We should clarify this somehow.
2. Its odd that we need this... it might be self explanatory but adding an explanation for why it is useful / needed would help clarify this for our future selves.

ritch (approved):

PR #5981, opened 2025-06-04
Fail children when parent is manually failed

kaixi-wang:

- set run_state of children to failed when parent is set as failed
- https://voxel51.atlassian.net/browse/FOEPD-1286

- added automated test

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured that when a partitioned delegated operation fails, all related child operations are also marked as failed and include an appropriate error message.

- **Tests**
  - Added a unit test to verify that failing a partitioned delegated operation correctly updates the status of all its child operations.

### DIFF
--- a/fiftyone/factory/repos/delegated_operation.py
+++ b/fiftyone/factory/repos/delegated_operation.py
@@ -392,7 +392,16 @@ class MongoDelegatedOperationRepo(DelegatedOperationRepo):
         ):
             # If a parent operation is failed, also mark the children as failed
             self._collection.update_many(
-                {"group_id": doc["_id"]}, {"$set": {"run_state": run_state}}
+                {
+                    "group_id": doc["_id"],
+                    "run_state": {"$nin": ["failed", "completed"]},
+                },
+                {
+                    "$set": {
+                        **update["$set"],
+                        "result": {"error": "parent operation failed"},
+                    }
+                },
             )
 
         return (


fiftyone/factory/repos/delegated_operation.py
```diff
@@ -390,6 +383,12 @@ def update_run_state(
             return_document=pymongo.ReturnDocument.AFTER,
         )
 
+        if doc.get("num_partitions") and run_state is ExecutionRunState.FAILED:
```
CamronStaley: couldn't this be moved into the run_state == ExecutionRunState.FAILED above?

kaixi-wang: I only have the doc id so I can't check if it is a parent op. Also, I was thinking that I only want to fail the children if the parent op was failed successfully...

fiftyone/factory/repos/delegated_operation.py
```diff
@@ -390,6 +385,16 @@ def update_run_state(
...
         )
 
+        if (
+            doc
+            and doc.get("num_partitions")
+            and run_state is ExecutionRunState.FAILED
+        ):
+            # If a parent operation is failed, also mark the children as failed
+            self._collection.update_many(
+                {"group_id": doc["_id"]}, {"$set": {"run_state": run_state}}
```
CamronStaley: should you set the other fields that get set for the parent as well:
                    "run_state": run_state,
                    "failed_at": datetime.utcnow(),
                    "updated_at": datetime.utcnow(),
                    "result": execution_result_json,
maybe result should be something like "parent was manually failed" or whatever some sort of meaningful message so users aren't confused what happened.

kaixi-wang: I actually had set it at first but then changed it to run_state only in case other children have actual run related data/results that correspond to actual execution events. Also, I figured this would be a good way to know that the child didn't fail, but was forced into failure without needing to parse the execution error in the doc

swheaton: I think we can do what Cam suggests if we filter for uncompleted children. Completed/failed children should remain as they were.

```python
self._collection.update_many({
        "group_id": doc["_id"],
        "run_state": {"$nin": ["failed", "completed"]}
    },
    {"$set": update}
)
```

kaixi-wang: Why make it more complicated with redundant data? Is there something that would break if not all the fields are filled out?

swheaton:
```
"run_state": run_state,
"failed_at": datetime.utcnow(),
"updated_at": datetime.utcnow(),
"result": {"error": "parent manually failed"},
```
these fields are redundant? how?

kaixi-wang: Run state is updated to failed. The other fields are on the parent. The result reason is not always manually failed (It could be an error), so I can't just use the same update as is. And if I filter out completed, then it doesn't satisfy the req that if a parent fails, all children are marked failed. We aren't using a relational db

swheaton: I see. I guess I'm saying I'm changing the requirement :D 

> the req that if a parent fails, all children are marked failed

If a parent fails, all _uncompleted_ children are marked failed with an appropriate error message

fiftyone/constants.py
```diff
@@ -14,7 +14,7 @@
 from importlib.metadata import metadata
 
 
-CLIENT_TYPE = "fiftyone"
+CLIENT_TYPE = "fiftyone-teams"
```
swheaton: oops

fiftyone/factory/repos/delegated_operation.py
```diff
@@ -390,6 +385,25 @@ def update_run_state(
...
+            # If a parent operation is failed, also mark the children as failed
+            self._collection.update_many(
+                {
+                    "group_id": doc["_id"],
+                    "run_state": {"$nin": ["failed", "completed"]},
+                },
+                {
+                    "$set": {
+                        **update["$set"],
+                        "result": {"error": "parent operation failed"},
```
swheaton, line 402: Hm I didn't think this would work since update already has "result" key defined, but it does in this syntax!
If it were `dict(**update["$set"], result={"error": "parent operation failed"})`, then it fails with `TypeError: dict() got multiple values for keyword argument 'result'`

fiftyone/constants.py
kaixi-wang: 🤫

PR #6001, opened 2025-06-12
fix frame field support type rendering bug

sashankaryal:

Fix a bug where support values where not being passed as an array, which the renderer expects

Before:

After:

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of tag rendering for certain field types to ensure values are correctly displayed, especially for fields that require single-element arrays.


benjaminpkane (approved):
LGTM 🐛

PR #6007, opened 2025-06-13
Replace docs search bar with Algolia DocSearch

dgarcia360:

Replaces docs search bar with Algolia DocSearch:

image

image

- [x] Searchbar renders locally
- [x] Searchbar works locally

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -223,9 +223,9 @@ pushfeedback_project = "1nx7ekqhts"
 pushfeedback_feedback_button_text = "Feedback"
 
 # -- Options for sphinx-docsearch --------------------------------------------
-docsearch_app_id = "7CWWSPCBVR"
-docsearch_api_key = "2824253b47efccdc7610e4b424dd85d3"
-docsearch_index_name = "voxel51"
+docsearch_app_id = os.environ.get("DOCSEARCH_APP_ID", "8ZYQ0G7IMC")
+docsearch_api_key = os.environ.get("DOCSEARCH_API_KEY", "")
+docsearch_index_name = os.environ.get("DOCSEARCH_API_KEY", "voxel51")
 docsearch_container = "#searchbox"
 
 # -- Custom app setup --------------------------------------------------------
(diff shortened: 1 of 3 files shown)


findtopher (approved):
👍

docs/source/conf.py
```diff
@@ -221,6 +222,12 @@
 pushfeedback_project = "1nx7ekqhts"
 pushfeedback_feedback_button_text = "Feedback"
 
+# -- Options for sphinx-docsearch --------------------------------------------
+docsearch_app_id = "7CWWSPCBVR"
+docsearch_api_key = "2824253b47efccdc7610e4b424dd85d3"
```
kevin-dimichel: Should we be hard coding this API Key in clear text into a public repo?  If not, this credential should be rotated and we should inject it via an environment variable.

findtopher: https://sphinx-docsearch.readthedocs.io/configuration.html

This looks like a key that is exposed when the site is published, so having it in the repo didn't seem like a problem (in my opinion)

kevin-dimichel: https://sphinx-docsearch.readthedocs.io/credentials.html suggests otherwise?

findtopher: Seems legit; looks like we need to rotate those credentials and update the deploy workflow to set up a .env file!

dgarcia360: Even though the suggestion to use a `.env` file comes from `sphinx-docsearch` docs, it's safe to use the search public key in frontend code. From Algolia DocSearch UI:

image

Note that the key is also visible whenever a user makes a request via the browser console:

image

In any case, I’ve moved it to a secret and rotated the key. The downside of this approach is that local search will no longer work.

Next, someone with the appropriate permissions must configure the key as a GitHub Secret ([instructions](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-an-environment)). The name of the env variable is `DOCSEARCH_API_KEY`.

@findtopher @kevin-dimichel — could you take care of this? Please contact me at david(@)techdocs(.)studio and I'll share the key.

dgarcia360:
We’ve received access to Algolia DocSearch program and completed the crawler configuration. Ready for review!

cc/ @jimmyguerrero

PR #5956, opened 2025-05-26
Add support for deleting all visible labels in selected samples

brimoor:
## Change log

- Upgrades the builtin `delete_selected_labels` to support deleting some/all labels in selected samples in the App
- This feature uses the new `ctx.active_fields` property to default to deleting only **visible** labels in the selected samples 🎉 
- Optimizes label deletion on generated views

### Regular dataset

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")
session = fo.launch_app(dataset)
```

https://github.com/user-attachments/assets/1ab0a449-1392-43e9-86b3-2a4cf229a234

### Patches view

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")
patches = dataset.to_patches("ground_truth")

session = fo.launch_app(patches)
```

https://github.com/user-attachments/assets/bbf051b1-30a0-4cc1-822e-97a6444624ce

### Frames view

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video")
frames = dataset.to_frames(sample_frames=True)

session = fo.launch_app(frames)
```

https://github.com/user-attachments/assets/a3d9faf5-4a3b-4f21-bb9e-a28926b31dff

## Summary by CodeRabbit

- **New Features**
  - Enhanced label deletion capabilities, allowing deletion of labels from selected samples and fields, with improved handling for video, patch, and clip views.
  - Added support for displaying and using active fields in plugin operators, providing more context-aware operations.
- **Bug Fixes**
  - Improved synchronization of label deletions between views (clips, frames, patches) and their source datasets for more accurate updates.
- **Documentation**
  - Updated plugin development documentation to describe the new active fields property in the operator and panel execution context.
- **Tests**
  - Added comprehensive tests to verify correct label deletion behavior for patches, clips, and frames in video datasets.

PR #6008, opened 2025-06-13
Revert "Add support for deleting all visible labels in selected samples"

jleven:
Reverts voxel51/fiftyone#5956

## Summary by CodeRabbit

- **Bug Fixes**
	- Simplified and streamlined the process for deleting selected labels, reducing complexity and improving reliability when removing labels from datasets.
- **Documentation**
	- Updated plugin development and operator documentation for improved clarity regarding active fields and label deletion.
- **Tests**
	- Removed outdated or redundant tests related to label deletion in patches, clips, and frames views to reflect the updated deletion logic.


kaixi-wang (approved):

fiftyone/core/collections.py
```diff
@@ -3291,19 +3291,8 @@ def _set_labels(self, field_name, sample_ids, label_docs, progress=False):
                 ops, ids=ids, frames=is_frame_field, progress=progress
             )
 
-    def _delete_labels(self, labels, fields=None):
-        self._dataset._delete_labels(labels, fields=fields)
-
-    def _map_values(self, in_values, in_field, *out_fields):
-        view = self.select_by(in_field, in_values)
```
brimoor, line 3298: This `select_by()` statement is critical: in the next line, we are not calling `values()` on the entire dataset, but rather only the specific samples whose `in_field` contains values in `in_values`. In the context of this PR, this means that the performance of `_delete_labels()` is now proportional to the number of labels actually being deleted (previously it always required a full collection scan, as I will describe below). **Moreover**,  the `_delete_labels()` methods in this PR invoke `_map_values()` with an `in_field` that is indexed (`"id"`, `"sample_id"`, etc), so `select_by()` can leverage indexes here.

In fact, as stated in the [PR description of #5956](https://github.com/voxel51/fiftyone/pull/5956), this PR **optimized** the performance of label deletion on generated views. The reason is that the previous implementation of these `_delete_labels()` methods would [cause an entire collection scan](https://github.com/voxel51/fiftyone/blob/2219d475de207e31aa3469d667103b2eb050eb15/fiftyone/core/dataset.py#L4073-L4083) looking to delete the specified labels by ID. Now with this PR, all deletions include the sample ID that contains the label ID to delete, which allows for the deletions to [operate directly on the sample document](https://github.com/voxel51/fiftyone/blob/2219d475de207e31aa3469d667103b2eb050eb15/fiftyone/core/dataset.py#L4223-L4232), avoiding a full collection scan.

kaixi-wang: Is it possible for this method to be triggered for >800k labels? If so, this would trigger the 16MB doc limit error.

brimoor: This PR's usage of `_map_values()` only applies to deleting labels that users have selected in the App, so no.

If we need `_map_values()` to support 800k+ values in the future, we can update it to support that. There are likely many other methods in this same boat.

PR #6029, opened 2025-06-18
add FIFTYONE_DEFAULT_MEDIA_TYPE env var

Laurent2916:

Added a small environment variable (`FIFTYONE_DEFAULT_MEDIA_TYPE`) to be able to set the default value returned by `get_media_type`.

Not sure about the `from fiftyone.__public__ import config` though, config should probably be fetched some other way.

Tested by patching my fiftyone package in an other project.

        notes for FiftyOne users

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Added a configurable default media type setting, allowing users to specify the default media type via configuration or environment variable.
- **Improvements**
  - The system now uses the configurable default media type when the media type cannot be determined, instead of a fixed value.

### DIFF
--- a/fiftyone/core/config.py
+++ b/fiftyone/core/config.py
@@ -280,6 +280,12 @@ class FiftyOneConfig(EnvConfig):
             env_var="FIFTYONE_EXECUTION_CACHE_ENABLED",
             default=True,
         )
+        self.default_media_type = self.parse_string(
+            d,
+            "default_media_type",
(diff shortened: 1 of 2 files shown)


brimoor (changes requested):
Hi @Laurent2916 👋 

Can you elaborate on your use case here?

Note that you can manually declare the media type when adding samples like this:

```py
sample = fo.Sample(filepath="foo.bar", media_type="spam")

dataset.add_samples([..., sample, ...])
```

PR #6043, opened 2025-06-23
Scenario analysis bugfix: we don't yet support None values

brimoor:
Scenario analysis doesn't yet support defining scenarios of the form `F(field) == None`. So, for now we update the Scenario analysis creation flow to **omit** `None` values from the UI.

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

ANIMALS = ["cat", "dog", None]
BOOLS = [True, False, None]

animals = [random.choice(ANIMALS) for _ in range(len(dataset))]
dataset.set_values("animals", animals)

bools = [random.choice(BOOLS) for _ in range(len(dataset))]
dataset.set_values("bools", bools)

dataset.take(66).clear_sample_field("uniqueness")

dataset.evaluate_detections(
    "predictions",
    gt_field="ground_truth",
    eval_key="eval",
)

# Try to create scenarios based on `bools`, `animals`, and `uniqueness` fields
```

### DIFF
--- a/plugins/operators/model_evaluation/__init__.py
+++ b/plugins/operators/model_evaluation/__init__.py
@@ -795,7 +795,8 @@ class ConfigureScenario(foo.Operator):
             return ShowOptionsMethod.CHECKBOX, {
                 "true": counts.get(True, 0),
                 "false": counts.get(False, 0),
-                "none": counts.get(None, 0),
+                # @todo consider supporting None
+                # "none": counts.get(None, 0),
             }
 
         # Retrieve distinct values (may be slow for large datasets)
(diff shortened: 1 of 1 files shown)


plugins/operators/model_evaluation/__init__.py
```diff
@@ -821,10 +822,14 @@ def get_scenarios_picker_type(self, ctx, field_name):
...
         values = dataset_or_view.count_values(field_name)
+
+        # @todo consider supporting None
+        values.pop(None, None)
+
         return (
             (ShowOptionsMethod.EMPTY, None)
             if not values
-            else (ShowOptionsMethod.CHECKBOX, dict(sorted(values.items())))
+            else (ShowOptionsMethod.CHECKBOX, values)
```
jleven: Why remove the sorting? Is the sorting happening somewhere else?

brimoor: @imanjra I meant to comment on this in the code.

I believe you included `dict(sorted(values.items()))` so that the checkboxes/lists we produce in the UI will be sorted alphabetically. However, dicts in Python are not **guaranteed** to have a stable ordering in all situations when you iterate over them. So if we want stable sorting, the interface should use lists of tuples instead.

For now I just removed the sorting since `dict()` could ruin it.

brimoor: update: I added back the sorting because its not harmful and *could* work. But points above still stand.

imanjra (approved):
LGTM

PR #6045, opened 2025-06-24
fix(execution-store): enforce ObjectId type for dataset_id and add test

ritch:
This PR adds type checking for dataset_id in the execution store and includes tests.

```python
import fiftyone.operators.store as fos
import fiftyone as fo

ds = fo.load_dataset('quickstart')
dataset_id = ds._doc.id
str_dataset_id = str(dataset_id)

try:
    # this will now throw to prevent incorrect / confusing behavior
    svc = fos.ExecutionStoreService(dataset_id=str_dataset_id)
except Exception as e:
    # prints "dataset_id must be an ObjectId, got str"
    print(e)

svc = fos.ExecutionStoreService(dataset_id=dataset_id)
print(svc.count_stores_global()) # => 3
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for dataset IDs to ensure only valid ObjectId types are accepted, preventing errors with incorrect input.
  
- **Tests**
  - Added and corrected tests to verify proper dataset ID handling and error reporting for invalid types.

- **Documentation**
  - Clarified that dataset IDs must be of ObjectId type in service class documentation.

### DIFF
--- a/fiftyone/factory/repos/execution_store.py
+++ b/fiftyone/factory/repos/execution_store.py
@@ -41,6 +41,10 @@ class ExecutionStoreRepo(ABC):
             dataset_id (Optional[ObjectId]): the dataset ID to operate on
             is_cache (False): whether the store is a cache store
         """
+        if dataset_id is not None and not isinstance(dataset_id, ObjectId):
+            raise ValueError(
+                f"dataset_id must be an ObjectId, got {type(dataset_id).__name__}"
+            )
         self._dataset_id = dataset_id
 
     @abstractmethod
@@ -301,6 +305,10 @@ class MongoExecutionStoreRepo(ExecutionStoreRepo):
     def __init__(
         self, collection, dataset_id: Optional[ObjectId] = None, is_cache=False
     ):
+        if dataset_id is not None and not isinstance(dataset_id, ObjectId):
+            raise ValueError(
+                f"dataset_id must be an ObjectId, got {type(dataset_id).__name__}"
+            )
         super().__init__(dataset_id, is_cache)
         self._collection = collection
         self._create_indexes()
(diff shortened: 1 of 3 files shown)


imanjra (dismissed):
LGTM

fiftyone/factory/repos/execution_store.py
```diff
@@ -41,6 +41,10 @@ def __init__(self, dataset_id: Optional[ObjectId] = None, is_cache=False):
             dataset_id (Optional[ObjectId]): the dataset ID to operate on
             is_cache (False): whether the store is a cache store
         """
+        if dataset_id is not None and not isinstance(dataset_id, ObjectId):
```
kaixi-wang, line 44: you can just do
```suggestion
        if dataset_id is not None:
            self._dataset_id = ObjectId(dataset_id)
```

This will auto convert the string to an ObjectId or raise an error otherwise

fiftyone/factory/repos/execution_store.py
```diff
@@ -301,6 +305,10 @@ class MongoExecutionStoreRepo(ExecutionStoreRepo):
     def __init__(
         self, collection, dataset_id: Optional[ObjectId] = None, is_cache=False
     ):
+        if dataset_id is not None and not isinstance(dataset_id, ObjectId):
```
kaixi-wang, line 308: same here

fiftyone/factory/repos/execution_store.py
ritch: Right, but for sanity I want the caller to handle that. I want to avoid having to support a string value at all.

jleven (approved):

PR #6064, opened 2025-06-25
UniqueFilenameMaker: cross-platform tempdir creation + test key behaviors

brimoor:
## Change log

- Make `MultiProcessUniqueFilenameMaker`'s temp directory creation platform-agnostic, so we can support Windows
- Revert to previous behavior of `UniqueFilenameMaker` when `idempotent=True` and `ignore_exts=True`

### Setup

```shell
mkdir -p /tmp/ufm
touch /tmp/ufm/000001.jpg
```

### UniqueFilenameMaker

```py

filename_maker = fou.UniqueFilenameMaker(
    output_dir="/tmp/ufm",
    ignore_exts=True,
    ignore_existing=False,
    idempotent=True,
)

print(filename_maker.get_output_path("/datasets/cifar10/test/000001.jpg"))
print(filename_maker.get_output_path("/datasets/cifar10/test/000001.jpg"))
print(filename_maker.get_output_path("/datasets/cifar10/test/000001.png"))
print(filename_maker.get_output_path("/datasets/cifar10/test/000001.png"))
print(filename_maker.get_output_path("/datasets/cifar10/validation/000001.jpg"))
print(filename_maker.get_output_path("/datasets/cifar10/validation/000001.jpg"))
```

```
/tmp/ufm/000001-2.jpg
/tmp/ufm/000001-2.jpg
/tmp/ufm/000001-3.png
/tmp/ufm/000001-3.png
/tmp/ufm/000001-4.jpg
/tmp/ufm/000001-4.jpg
```

Explanation of correctness of output paths:
1. Output filenames start at `000001-2` because of `ignore_exts=True` + `ignore_existing=False` and the existing `000001` filename in the output directory
2. Output filenames come in pairs because of `idempotent=True`
3. Second unique output filename is `000001-3` because `ignore_exts=True` causes `000001.jpg` and `000001.png` to be seen as duplicates. This is intentional and specifically relied upon by various exporters such as VOC, KITTI, and YOLO that need to generate `.txt` files with the same root name as each media file. In such cases, we cannot allow output filenames to have the same root but different extensions because the corresponding `.txt` files would have clashing names

### MultiProcessUniqueFilenameMaker

```py
import os

filename_maker = fou.MultiProcessUniqueFilenameMaker(
    os.getpid(),
    output_dir="/tmp/ufm",
    ignore_exts=True,
    ignore_existing=False,
    idempotent=True,
)

print(filename_maker.get_output_path("/datasets/cifar10/test/000001.jpg"))
print(filename_maker.get_output_path("/datasets/cifar10/test/000001.jpg"))
print(filename_maker.get_output_path("/datasets/cifar10/test/000001.png"))
print(filename_maker.get_output_path("/datasets/cifar10/test/000001.png"))
print(filename_maker.get_output_path("/datasets/cifar10/validation/000001.jpg"))
print(filename_maker.get_output_path("/datasets/cifar10/validation/000001.jpg"))
```

```
/tmp/ufm/000001-2.jpg
/tmp/ufm/000001-2.jpg

# incorrect: with ignore_exts=True, these should cause an increment in the "-#" counter
/tmp/ufm/000001-2.png
/tmp/ufm/000001-2.png

# incorrect: these need to *not* be the same as the first two paths
/tmp/ufm/000001-2.jpg
/tmp/ufm/000001-2.jpg
```

Issues with `MultiProcessUniqueFilenameMaker` that need solutions:
1. Does not correctly handle the case where multiple paths with the same root but different extensions are provided (when `ignore_exts=True`)
2. Does not correctly handle the case where multiple paths with the same filenames but different directories are provided

### DIFF
--- a/tests/unittests/utils/test_unique_filename_maker.py
+++ b/tests/unittests/utils/test_unique_filename_maker.py
@@ -46,7 +46,8 @@ def cleanup():
     # it does work when manually running
     # cleaning up touched files after the session finishes
     with contextlib.suppress(Exception):
-        shutil.rmtree(f"/tmp/fo-unq/{os.getpid()}")
+        tmp_dir = focu.MultiProcessUniqueFilenameMaker.gettempdir(os.getpid())
+        shutil.rmtree(tmp_dir)
 
 
 # =========================================================================
@@ -110,14 +111,14 @@ def cleanup():
             {"ignore_existing": True},
             id="{ignore_existing=True, idempotent=True}",
         ),
-        pytest.param(
-            {"ignore_exts": True, "idempotent": False},
-            id="{ignore_exts=True, idempotent=False}",
-        ),
-        pytest.param(
-            {"ignore_exts": True, "idempotent": True},
-            id="{ignore_exts=True, idempotent=True}",
-        ),
+        # pytest.param(
+        #    {"ignore_exts": True, "idempotent": False},
+        #     id="{ignore_exts=True, idempotent=False}",
+        # ),
+        # pytest.param(
+        #     {"ignore_exts": True, "idempotent": True},
+        #     id="{ignore_exts=True, idempotent=True}",
+        # ),
     ),
 )
 @skip_windows  # TODO: don't skip on Windows
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -35,6 +35,7 @@ import struct
 import subprocess
 import sys
 import shutil
+import tempfile
 import timeit
 import types
 import uuid
@@ -2079,11 +2080,10 @@ class UniqueFilenameMaker(object):
 
         self._idx = len(filenames)
         for filename in filenames:
-            key = filename
-
-            # Adding ignore extension to filename counts
             if self.ignore_exts:
                 key, _ = os.path.splitext(filename)
+            else:
+                key = filename
 
             self._filename_counts[key] += 1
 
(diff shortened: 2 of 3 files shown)


jleven:
> Second unique output filename is 000001-3 because ignore_exts=True causes 000001.jpg and 000001.png to be seen as duplicates.

If they are seen as duplicates, then doesn't `idempotent` mean "same input => same output" and they should produce the same filename?

tests/unittests/utils/test_unique_filename_maker.py
```diff
@@ -110,14 +111,14 @@ def cleanup():
...
-            id="{ignore_exts=True, idempotent=True}",
-        ),
+        # pytest.param(
+        #    {"ignore_exts": True, "idempotent": False},
+        #     id="{ignore_exts=True, idempotent=False}",
+        # ),
+        # pytest.param(
+        #     {"ignore_exts": True, "idempotent": True},
+        #     id="{ignore_exts=True, idempotent=True}",
+        # ),
```
j053y:
```suggestion
        pytest.param(
           {"ignore_exts": True, "idempotent": False},
            id="{ignore_exts=True, idempotent=False}",
            marks=pytest.mark.skip(reason="Skipping this specific case until FOEPD-920 is complete")
        ),
        pytest.param(
            {"ignore_exts": True, "idempotent": True},
            id="{ignore_exts=True, idempotent=True}",
            marks=pytest.mark.skip(reason="Skipping this specific case until FOEPD-920 is complete")
        ),
```

I don't think we should ever be commenting out tests. This suggested method allows skipping of tests in a more formal way with information about why. They'll still show up in the test report (potentially incentivizing making the fix).

swheaton: nice, didn't know you can put skip on a param like this

PR #6067, opened 2025-06-26
Add support for reloading saved generated views

brimoor:
## Change log

- Adds a builtin `reload_saved_view` operator that allows for checking + reloading saved generated views
- Calling `reload()` on a saved generated view will now automatically update the saved view's metadata
- Backing datasets for saved generated views are now marked as persistent so that they are not affected by non-persistent dataset cleanup
    - Deleting a dataset or specific generated saved views will automatically mark the now-deprecated backing dataset as non-persistent
- Switch to `_state["dataset_id"]` in generated views to prevent unnecessary reloads when the parent dataset is renamed

## TODO before merging

1. ~~Why doesn't the icon asset of the builtin `reload_saved_view` operator render?~~
2. Fix caching issue where App erroneously still refers to the old backing dataset (which no longer exists and was replaced with a new dataset of the same name) when `patches.reload()` is called in SDK

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

limit = dataset.limit(100)

dataset.save_view("limit", limit)
dataset.save_view("patches", patches)

lma1 = dataset._get_saved_view_doc("patches").last_modified_at

dataset.delete_samples(dataset[:50])

session = fo.launch_app(patches)

# Use `reload_saved_view` to reload `patches` view
# This works as expected

lma2 = dataset._get_saved_view_doc("patches").last_modified_at
assert lma2 > lma1

dataset.delete_samples(dataset[:50])

patches.reload()

lma3 = dataset._get_saved_view_doc("patches").last_modified_at
assert lma3 > lma2

# Load `patches` view in App
# This causes Bug 2 above in App
```


fiftyone/core/stages.py
```diff
@@ -501,12 +503,15 @@ def _parse_params(self, samples):
 
     def _serialize_view(self, view):
         return {
-            "dataset": view._root_dataset.name,
+            "dataset_id": str(view._root_dataset._doc.id),
```
kaixi-wang, line 506: it just occurred to me, but will this break backwards/forwards compatibility?

fiftyone/core/dataset.py
```diff
@@ -9022,11 +9060,28 @@ def _do_load_dataset(obj, name):
...
-def _delete_dataset_doc(dataset_doc):
+def _handle_delete_generated_saved_view(dataset, view_doc):
+    if not any("_state" in s for s in view_doc.view_stages):
+        return
+
+    try:
+        # When deleting *generated* saved views, we mark the backing
+        # dataset as non-persistent so it can be garbage collected
+        view = dataset._load_saved_view_from_doc(view_doc)
+        view._dataset.persistent = False
```
kaixi-wang: why would we not want to just delete the generated dataset instead

brimoor: My thinking was that someone could still be working with the same view in another process, and we wouldn't want the view to suddenly start throwing exceptions over there. Marking as non-persistent guarantees (if cleanup actually occurs!) that the dataset will only be deleted when it is not in-use

fiftyone/core/dataset.py
```diff
@@ -4525,6 +4525,10 @@ def save_view(
         self._doc.last_modified_at = now
         self._doc.save(virtual=True)
 
+        view._set_name(name)
+        if view._is_generated:
+            view._dataset.persistent = True
```
kaixi-wang: this is going to mean some api permissions work is necessary for teams

fiftyone/core/dataset.py
kaixi-wang: Would like to hear other opinions, but I don't think users should be allowed to continue working with a view that's deleted

brimoor: My perspective is that there's no such thing as deleting a "view". Deleting a "saved view" removes the record of the view from the dataset. But its still a valid "view" in the sense that it expresses a filter that can still be applied to the dataset to retrieve samples and do stuff with them.

fiftyone/core/dataset.py
brimoor: Yep I have a plan for that. Can chat offline

tom-vx51 (approved):

PR #6078, opened 2025-06-27
Fix partial range filters in sidebar

benjaminpkane:

With manual input for numeric fields added, backend support for only a min _or_ a max was overlooked. Specifically for list fields

Updated server view test

## Release Notes

N/A

-   [ ] Other


jleven (approved):
Looks good

PR #6080, opened 2025-06-28
fix(useBrowserStorage): improve handling of undefined values

ritch:

Fixes an issue where the `useBrowserStorage` utility would persist an `"undefined"` value in localStorage, which is not valid JSON:

image

[FOEPD-1648](https://voxel51.atlassian.net/browse/FOEPD-1648)

(Details)

        notes for FiftyOne users.

-   [ ] Other

[FOEPD-1648]: https://voxel51.atlassian.net/browse/FOEPD-1648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of "undefined" values in browser storage to prevent errors and unintended behavior.
  * Ensured that undefined values are removed from storage instead of storing the string "undefined".
  * Enhanced reliability when using custom parsing functions.
* **Tests**
  * Added comprehensive tests for browser storage behavior, including handling of undefined, null, and custom parsing scenarios.

### DIFF
--- a/app/packages/state/src/hooks/useBrowserStorage.ts
+++ b/app/packages/state/src/hooks/useBrowserStorage.ts
@@ -16,11 +16,15 @@ export const useBrowserStorage = <T = string>(
 
   // Pass initial state function to useState so logic is only executed once
   const [storedValue, setStoredValue] = useState<T>(() => {
-    const item = storage.getItem(key);
+    let item = storage.getItem(key);
 
     if (item) {
       if (parseFn) {
         return parseFn.parse(item);
+      }
+      // Workaround for existing "undefined" values in storage
+      if (item === "undefined") {
+        return;
       }
 
       return JSON.parse(item);
@@ -44,12 +48,17 @@ export const useBrowserStorage = <T = string>(
         setStoredValue(value);
       }
 
+      // Convert undefined to null to avoid "undefined" values in storage
+      if (valueToStore === undefined) {  
+        valueToStore = null;
+      }
+
       storage.setItem(
         key,
         parseFn ? parseFn.stringify(valueToStore) : JSON.stringify(valueToStore)
       );
     },
-    [key, storage]
+    [key, storage, parseFn]
   );
 
   return [storedValue, setValue] as const;
--- a/app/packages/state/src/hooks/useBrowserStorage.ts
+++ b/app/packages/state/src/hooks/useBrowserStorage.ts
@@ -48,9 +48,17 @@ export const useBrowserStorage = <T = string>(
         setStoredValue(value);
       }
 
-      // Convert undefined to null to avoid "undefined" values in storage
-      if (valueToStore === undefined) {  
-        valueToStore = null;
+      // Only apply undefined handling when no parseFn is provided
+      if (parseFn) {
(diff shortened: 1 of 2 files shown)


app/packages/state/src/hooks/useBrowserStorage.ts
```diff
@@ -44,12 +48,17 @@ export const useBrowserStorage = <T = string>(
         setStoredValue(value);
       }
 
+      // Convert undefined to null to avoid "undefined" values in storage
+      if (valueToStore === undefined) {  
+        valueToStore = null;
```
jleven: Do we want to store null? Or just not store anything at all?

ritch: Local storage only stores strings. undefined isn’t valid, and empty string might mean something to the user of this util.

Since JSON drops undefined eg when a property is set to undefined , another option is to just remove the item from storage when undefined is set.

Storing "null" is the least invasive default since it’s valid JSON and clearly intentional.

Note parseFn can handle this encoding however it wants, so deleting the item would complicate that.

app/packages/state/src/hooks/useBrowserStorage.ts
```diff
@@ -44,12 +48,17 @@ export const useBrowserStorage = <T = string>(
...
+      }
```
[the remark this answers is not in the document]
ritch: What if parseFn handles undefined? This breaks that contract and just removes the item.

app/packages/state/src/hooks/useBrowserStorage.ts
```diff
@@ -16,12 +16,16 @@ export const useBrowserStorage = <T = string>(
 
   // Pass initial state function to useState so logic is only executed once
   const [storedValue, setStoredValue] = useState<T>(() => {
-    const item = storage.getItem(key);
+    let item = storage.getItem(key);
```
ritch: Can revert this…

app/packages/state/src/hooks/useBrowserStorage.ts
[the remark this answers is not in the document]
ritch: Good rabbit have a carrot 🥕

app/packages/state/src/hooks/useBrowserStorage.ts
ritch: Remove item looks good actually. Going to try that. Also this needs a test!

jleven: Yay! Test!

benjaminpkane (approved):

Br2850 (approved):
Great addition with the tests as well 👍🏽

Br2850:
@ritch Looks like your latest commit might have the Python tests no longer - make sure to check that before merging. I took a look at the `develop` history and they were passing + your commit before `add test suite` was passing with commit [e27c1d8](https://github.com/voxel51/fiftyone/pull/6080/commits/e27c1d81451b5ce857d577ad90e385623c8e1d7a)

PR #6087, opened 2025-06-30
Fix Management SDK references in release notes

brimoor:
## Change log

- add new `fiftyone.management` modules introduced in v1.7 to the Management SDK docs
- fix broken references to `fiftyone.management` modules in the release notes
- update release dates to July 1

### DIFF
--- a/docs/source/enterprise/management_sdk.rst
+++ b/docs/source/enterprise/management_sdk.rst
@@ -8,14 +8,14 @@ Enterprise Management SDK
 One of FiftyOne's core design principles is that you should be able to do
 everything *programmatically* if you want.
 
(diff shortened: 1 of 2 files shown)


kaixi-wang (dismissed):
thanks for fixing!

swheaton:
Thanks 🙏🏼 

I would be OK with leaving fom.orchestrators out entirely though, as it references stuff that doesn't fully exist yet until next release.

brimoor:
> Thanks 🙏🏼
> 
> I would be OK with leaving fom.orchestrators out entirely though, as it references stuff that doesn't fully exist yet until next release.

It's in the release so 🤷‍♂️

PR #6097, opened 2025-07-02
Remove tests using quantiles with inf and nan

swheaton:
`quantiles()` uses server-side JS function to sort. Sorting `inf` and `nan` is undefined, and we see that in Mongo 8 the order is somewhat different. The other assertions in this test are reasonable, but the quantile ones are not really.

Besides, this (server side JS) is deprecated functionality in mongo 8 anyways.

Remove offending lines

Tests pass in MongoDB 8

## Summary by CodeRabbit

* **Tests**
  * Updated unit tests by removing specific assertions related to quantile calculations with NaN and infinite values.


tom-vx51 (approved):

kaixi-wang:
I think there is some ordering that should be tested for because the values aren't all NaN or Inf

PR #6099, opened 2025-07-02
remove dead code in queue_operation thats impossible to get to

swheaton:
Removing dead code - the "elif" case is impossible to get to. The operation is newly created a few lines above and then dataset_id never is set nor is it a computed property, so there's never a chance for `op.dataset_id` to be True.

Also getting rid of `num_partitions` as that is not a thing anymore. Removing child/parent stuff from set-failed as it is not in this repo.

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal handling of dataset identification during operation queuing and removed propagation of failure states to child operations. No changes to user-facing features or functionality.


CamronStaley (approved):

swheaton:
> If we remove it from here, won't setting failed with the sdk potentially cause issues?

planning on Removing it from oss only

PR #6111, opened 2025-07-07
fix correct/incorrect trophy and ratio color

imanjra:

fix correct/incorrect trophy and ratio color

Using ME panel

        notes for FiftyOne users.

- For Incorrect row, show trophy is primary model incorrect count is lesser
- When lesser is better, show negative ratio in colour green

-   [ ] Other


brimoor (approved):
Confirmed that this is the behavior that we want and verified functionality via offline screenshots 🏆 ✅

findtopher (approved):
Aloha got called in for `setup.py` which, apparently, went away when the base branch got changed.

Seems fine to us!

PR #5183, opened 2024-11-23
Fix deleted datasets on App server

benjaminpkane:
```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart")
session = fo.launch_app(dataset)

#.wait for App, then delete dataset and recreate
dataset.delete()
dataset = foz.load_zoo_dataset("quickstart")

# refresh browser page
``` 
Screenshot 2024-11-22 at 8 10 20 PM

## Summary by CodeRabbit

- **New Features**
	- Enhanced dataset loading mechanism to ensure datasets are loaded correctly and efficiently.
	- Improved error handling for loading saved views, allowing for more robust dataset view creation.

- **Bug Fixes**
	- Adjusted error handling logic to better manage exceptions during dataset and view loading.

- **Refactor**
	- Updated method signatures for better control over dataset loading behavior.

PR #6114, opened 2025-07-07
use load_dataset to load dataset

kaixi-wang:

use load_dataset instead of directly initializing Dataset 

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/server/query.py
+++ b/fiftyone/server/query.py
@@ -610,7 +610,7 @@ async def serialize_dataset(
         if not fod.dataset_exists(dataset_name):
             return None
 
-        dataset = fo.Dataset(dataset_name, _create=False, _force_load=True)
+        dataset = fo.load_dataset(dataset_name, create_if_necessary=False)
         dataset.reload()
         view_name = None
         try:


fiftyone/server/query.py
```diff
@@ -610,7 +610,7 @@ def run():
         if not fod.dataset_exists(dataset_name):
             return None
 
-        dataset = fo.Dataset(dataset_name, _create=False, _force_load=True)
+        dataset = fo.load_dataset(dataset_name, create_if_necessary=False)
```
brimoor: nit: can we leave off `create_if_necessary=False` for clarity? We would never change the default value of that parameter. That would be far too disruptive.

fiftyone/server/query.py
```diff
@@ -610,7 +610,7 @@ def run():
         if not fod.dataset_exists(dataset_name):
             return None
 
-        dataset = fo.Dataset(dataset_name, _create=False, _force_load=True)
```
brimoor, line 613: Definitely need @benjaminpkane's sign-off on this change as he specifically wanted to use `_force_load=True` here in https://github.com/voxel51/fiftyone/commit/cdd4a488f8df1865e2a98c59f456429e34e763cf.

It does make sense that we *should* be using `load_dataset()` here; just noting that there might be *additional* considerations to complete the migration to `load_dataset()`.

brimoor: Specifically, after looking at the code more closely, `_force_load=True` is necessary to ensure that a deleted dataset is *immediately* recognized as such when `load_dataset()` is called, rather than returning a stale in-memory singleton of the `Dataset` before it was deleted. So this PR as-is would re-introduce that issue, which we don't want.

Also, just noting that the additional `dataset.reload()` immediately after the `load_dataset()` call is 🤦 

We reallllllly need "no singleton mode" in the App so that a simple `load_dataset()` call will do everything we want...
- always load the latest dataset doc from the DB
- avoid *double loads* that occur with the `load_dataset()` + `reload()` pattern when the `Dataset` has not been previously loaded
- immediately raise an error if the dataset is deleted
- allow for additional logic to occur in `load_dataset()` that doesn't exist in `Dataset`

kaixi-wang: Wouldn't the code above this handle the delete issue?
```        
if not fod.dataset_exists(dataset_name):
            return None
```
This queries from the db and thus should avoid the singleton issue? 

Best to wait for Ben's input though because the _force_load seems to be linked to updating last loaded at and I can see a potential issue with triggering that too often

swheaton: Alternatively I could make the Dataset constructor work with snapshots

benjaminpkane: > Alternatively I could make the Dataset constructor work with snapshots

If this is low hanging fruit, that would be ideal. Returning to `load_dataset` would regress https://github.com/voxel51/fiftyone/pull/5183

swheaton: Ok. Until we can confidently offer a no-singleton mode, let me work on this stop gap. The fruit is higher than this PR but still pretty low 😝

swheaton: @benjaminpkane do you have a response to Kacey's comment about `fod.dataset_exists()` ?

benjaminpkane: A dataset with `dataset_name` exists after deleting and recreating so I don't think that is sufficient

brimoor: Thinking from first principles, these are the things we need in `fiftyone/server/query.py`:
1. always load the latest dataset doc from the DB (eg so that schema changes are always recognized) 
2. if a dataset is deleted, immediately raise an error
3. don't return a stale `Dataset` singleton in a situation where a dataset was deleted and a new one was created with the same name
4. allow for additional logic (eg FOE-specific) to occur in a common place so that all publicly-documented instances of loading datasets will use it
5. avoid double loads that occur with the `load_dataset()` + `reload()` pattern when the Dataset has not been previously loaded

The only optional one is 5.

PR #6130, opened 2025-07-10
Refactor panel error handling

ritch:
Refactor panel error handling.


app/packages/operators/src/usePanelEvent.ts
```diff
@@ -33,19 +54,47 @@ export default function usePanelEvent() {
...
+      panel_state: currentPanelState ?? ((panelState as any)?.state || {}),
     };
 
     const eventCallback = (result, opts) => {
       decrement(panelId);
-      const msg =
-        result.errorMessage || result.error || "Failed to execute operation";
-      const computedMsg = `${msg} (operation: ${operator})`;
-      if (result?.error) {
-        notify({ msg: computedMsg, variant: "error" });
```
imanjra, line 45: Should we suppress the notify from `useOperatorExecutor` as well?

imanjra (approved):
LGTM

app/packages/operators/src/usePanelEvent.ts
ritch: Yep I've added a new option "skipErrorNotification" which is set to true when calling executeOperator via panel events.

PR #6131, opened 2025-07-10
Ultralytics: return empty labels, not None, when no objects are found

brimoor:
## Change log

When performing inference with `ultralytics` models, if no suitable objects were found, return empty labels (eg `Detections()`) rather than `None`.

This is important because the `Detections()` vs `None` distinction is semantically meaningful for downstream applications:
- `Detections()`: sample was processed but no suitable objects were found
- `None`: sample has not been processed

## Tested by

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=3)
model = foz.load_zoo_model("yolov8s-coco-torch")

dataset.apply_model(model, label_field="boxes", confidence_thresh=0.999999)

print(dataset.values("boxes"))
```

```
[<Detections: {'detections': []}>,
 <Detections: {'detections': []}>,
 <Detections: {'detections': []}>]
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency by returning empty label containers instead of `None` when no detection, instance, classification, OBB, polyline, or keypoint data is present. This ensures more predictable behavior when handling empty results.

### DIFF
--- a/fiftyone/utils/ultralytics.py
+++ b/fiftyone/utils/ultralytics.py
@@ -90,7 +90,7 @@ def to_detections(results, confidence_thresh=None):
 
 def _to_detections(result, confidence_thresh=None):
     if result.boxes is None:
-        return None
+        return fol.Detections()
 
     classes = np.rint(result.boxes.cls.detach().cpu().numpy()).astype(int)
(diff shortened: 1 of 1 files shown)


manushreegangwar (approved):
LGTM

PR #6136, opened 2025-07-14
fix(PluginDefinition): fix server path resolution for built in plugins

ritch:
Fixes an issue where builtin plugin definitions server paths were resovled incorrectly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of plugin server paths to correctly distinguish between built-in and user plugins, ensuring consistent URL formatting across operating systems.

* **Tests**
  * Added unit tests to verify correct server path generation for both built-in and user plugins.


kaixi-wang (approved):
thanks! Seems to solve the `Path ... is not in ...` logs in the api

PR #6138, opened 2025-07-15
Fix #6137: use named function to avoid pickle error

### LINKED ISSUE #6137: [BUG] AttributeError: Can't pickle local object '_make_patch_data_loader.<locals>.<lambda>'
On `fiftyone==1.7`, the following code raises this error on macOS:

```
AttributeError: Can't pickle local object '_make_patch_data_loader.<locals>.<lambda>'
```

**unless** I uncomment either of the following fixes:
- set `num_workers=0`
- force `fork` multiprocessing to be used

in which case it works.

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.brain as fob

# this makes it work
# import multiprocessing; multiprocessing.set_start_method("fork", force=True)

index = fob.compute_similarity(
    dataset,
    brain_key="clip_sim_patch",
    patches_field="ground_truth",
    embeddings="clip",
    model="clip-vit-base32-torch",
    # so does this
    # num_workers=0,
)
```

We started allowing multiple workers by default on macOS in [#5939](https://github.com/voxel51/fiftyone/pull/5939), which seems related.

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/6137

The code in https://github.com/voxel51/fiftyone/issues/6137 now runs successfully on macOS **without** forcing `num_workers=0` or forcing `fork` multiprocessing to be used ✅

### DIFF
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -1955,8 +1955,12 @@ def _make_patch_data_loader(
         dataset,
         batch_size=1,
         num_workers=num_workers,
-        collate_fn=lambda batch: batch[0],  # return patches directly
+        collate_fn=_patch_collate_fn,
     )
+
+
+def _patch_collate_fn(batch):
+    return batch[0]  # return patches directly
 
 
 def _parse_batch_size(batch_size, model, use_data_loader):


kaixi-wang (approved):
Tested compute_similarities with patches and no pickling error 👍  Thanks!

PR #6149, opened 2025-07-16
FOEPD-1759: Add optimized syntaxes for `set_values()`, `set_label_values()` and `map_values()` 

brimoor:
The original goal of this PR was to update the builtin `edit_field_values` operator so that it could be applied to views that filter label fields (cf [FOEPD-1474](https://voxel51.atlassian.net/browse/FOEPD-1474)). This is achieved ✅ 

The underlying technical need to achieve that goal was to have an alternative to the `view.map_values(path, map).save(root)` pattern that can be safely applied when `view` filters the `root` field. In that case, we need to use either `set_label_values()` or `set_values()` as appropriate to perform the field edits as individual DB ops.

Achieving this technical need led me down a deeply satisfying rabbit hole of optimizing some of our core batch edit methods:

1. Added a new syntax to `set_label_values()` that allows users to provide sample IDs in addition to label IDs for each update, for efficiency 

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.limit(5).filter_labels("predictions", F("confidence") > 0.5)

# New list syntax: provide sample IDs and label IDs
values = []
for sid, lids in zip(*view.values(["id", "predictions.detections.id"])):
    for lid in lids:
        values.append({"sample_id": sid, "label_id": lid, "value": True})

dataset.set_label_values("predictions.detections.random", values)

print(dataset.count_values("predictions.detections.random"))
# {True: 25, None: 5595}
```

2. Added a new syntax to `set_values()` where users can update frame fields by directly providing a mapping from frame IDs to values

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

# New dict syntax: provide a dict mapping frame IDs to values
frame_ids = dataset.values("frames.id", unwind=True)
values = {id: random.random() for id in frame_ids}

dataset.set_values("frames.random", values, key_field="frames.id")
print(dataset.bounds("frames.random"))
# (0.00013, 0.9993)
```

3. Added a new `fiftyone.utils.data.map_values()` utility that performs the same operation as `SampleCollection.map_values()` but it immediately saves the mapped values to the database rather than creating a view

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.data as foud

ANIMALS = [
    "bear", "bird", "cat", "cow", "dog", "elephant", "giraffe",
    "horse", "sheep", "zebra"
]

# Map all animals to string "animal"
mapping = {a: "animal" for a in ANIMALS}

foud.map_values(dataset, "ground_truth.detections.label", mapping)

print(dataset.count_values("ground_truth.detections.label"))
# {"animal": 183, ...}
```

[FOEPD-1474]: https://voxel51.atlassian.net/browse/FOEPD-1474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -13,6 +13,7 @@ from operator import itemgetter
 import fnmatch
 import itertools
 import logging
+import operator
 import os
 import random
 import string
@@ -2440,6 +2441,13 @@ class SampleCollection(object):
 
         .. note::
 
+            If you are setting attributes of a nested list of labels, such as
+            attributes of the objects in a
+            :class:`fiftyone.core.labels.Detections` field, then consider using
+            :meth:`set_label_values` instead for greater efficiency.
+
+        .. note::
+
             If the values you are setting can be described by a
             :class:`fiftyone.core.expressions.ViewExpression` applied to the
             existing dataset contents, then consider using :meth:`set_field` +
@@ -2686,6 +2694,12 @@ class SampleCollection(object):
             wish to modify. See :meth:`set_values` and :meth:`set_field` if
             your updates are not keyed by label ID.
 
+        .. note::
+
+           This method is *most efficient* when you use the list syntax for
+            ``values``, which includes the sample/frame ID of each label that
+            you are modifying.
+
         Examples::
 
             import fiftyone as fo
@@ -2700,6 +2714,22 @@ class SampleCollection(object):
 
             view = dataset.filter_labels("predictions", F("confidence") > 0.99)
 
+            # Option 1: provide label IDs *and* sample IDs
+
+            values = []
+            sample_ids, label_ids = view.values(["id", "predictions.detections.id"])
+            for sid, lids in zip(sample_ids, label_ids):
+                for lid in lids:
+                    values.append({"sample_id": sid, "label_id": lid, "value": True})
+
+            dataset.set_label_values("predictions.detections.high_conf", values)
+
+            print(dataset.count("predictions.detections"))
+            print(len(values))
+            print(dataset.count_values("predictions.detections.high_conf"))
+
+            # Option 2: provide only label IDs
+
             label_ids = view.values("predictions.detections.id", unwind=True)
             values = {_id: True for _id in label_ids}
 
@@ -2711,7 +2741,15 @@ class SampleCollection(object):
 
         Args:
             field_name: a field or ``embedded.field.name``
-            values: a dict mapping label IDs to values
+            values: the label values to set, in one of the following formats:
+
+                -   a list of dicts of the form
+                    ``{"sample_id": sample_id, "label_id": label_id, "value": value}
+                    when setting sample-level labels
+                -   a list of dicts of the form
+                    ``{"frame_id": frame_id, "label_id": label_id, "value": value}
+                    when setting frame-level labels
+                -   a dict mapping label IDs to values
             skip_none (False): whether to treat None data in ``values`` as
                 missing data that should not be set
             dynamic (False): whether to declare dynamic attributes of embedded
@@ -2722,8 +2760,13 @@ class SampleCollection(object):
                 use the default value ``fiftyone.config.show_progress_bars``
                 (None), or a progress callback function to invoke instead
         """
+        if isinstance(values, list):
+            _values = map(operator.itemgetter("value"), values)
+        else:
+            _values = values.values()
+
         field, _ = self._expand_schema_from_values(
-            field_name, values.values(), dynamic=dynamic, flat=True
+            field_name, _values, dynamic=dynamic, flat=True
         )
 
         if field is None:
@@ -2732,6 +2775,11 @@ class SampleCollection(object):
         if field is not None and field.read_only:
             raise ValueError("Cannot edit read-only field '%s'" % field_name)
 
+        label_field, root, is_list_field, _ = self._parse_label_attribute(
+            field_name
+        )
+        _root, _ = self._handle_frame_field(root)
+
         _field_name, is_frame_field, _, _, id_to_str = self._parse_field_name(
             field_name, omit_terminal_lists=True
         )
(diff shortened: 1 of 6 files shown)


fiftyone/utils/data/base.py
```diff
@@ -17,14 +18,197 @@
...
+        if is_list_leaf:
+            expr = F(leaf).exists() & (
+                F(leaf).filter(F().is_in(inputs)).length() > 0
+            )
+        else:
+            expr = F(leaf).is_in(inputs)
+
+        view = sample_collection.filter_labels(label_field, expr)
+
+        doc_ids, label_ids, curr_values = view.values(
```
kaixi-wang, line 143: Just want to mention that values() really doesn't scale well and is often the cause of things appearing to hang for significant periods of time

brimoor: `view` filters down to only the values that need edited, so the `values()` call will scale proportionally to the number of labels being edited. I don't know of any better way to achieve what we need here.

fiftyone/core/collections.py
```diff
@@ -11834,6 +12029,19 @@ def _make_set_field_pipeline(
             context=context,
         )
 
+    def _edits_field(self, path):
+        if not isinstance(self, fov.DatasetView):
+            return False
+
+        # pylint:disable=no-member
```
exupero, line 12036:
```suggestion
        # pylint:disable-next=no-member
```

fiftyone/utils/data/base.py
```diff
@@ -17,14 +18,201 @@
 import eta.core.utils as etau
 import eta.core.video as etav
 
+from fiftyone.core.expressions import ViewField as F
```
exupero, line 21: What's the motivation for aliasing this as a single-letter variable name?

swheaton: longstanding convention

exupero: Digging in some more, it looks like it's used as a DSL for building filter queries, so seems reasonable to have a shorthand.

fiftyone/core/collections.py
```diff
@@ -2722,8 +2865,16 @@ def set_label_values(
                 use the default value ``fiftyone.config.show_progress_bars``
                 (None), or a progress callback function to invoke instead
         """
+        if isinstance(values, list):
+            _values = [d["value"] for d in values]
+        else:
+            _values = values.values()
```
exupero, line 2871: What are your feelings on using a ternary in situations like this? e.g.
```suggestion
        _values = [d["value"] for d in values] if isinstance(values, list) else values.values()
```

fiftyone/core/collections.py
```diff
@@ -2809,6 +2928,48 @@ def set_label_values(
...
+            _sample_ids, _label_ids = view.values(["id", label_id_path])
+
+            if is_list_field:
+                for _sample_id, _lids in zip(_sample_ids, _label_ids):
+                    if _lids:
+                        for _label_id in _lids:
+                            id_map[_label_id] = _sample_id
+            else:
+                for _sample_id, _label_id in zip(_sample_ids, _label_ids):
+                    id_map[_label_id] = _sample_id
```
exupero, line 2969: How does this compare to composing dict and zip?
```suggestion
                id_map = dict(zip(_label_ids, _sample_ids))
```

fiftyone/core/collections.py
```diff
@@ -2809,6 +2928,48 @@ def set_label_values(
...
+            if is_list_field:
+                for _fids, _flids in zip(_frame_ids, _label_ids):
+                    for _frame_id, _lids in zip(_fids, _flids):
+                        if _lids:
+                            for _label_id in _lids:
+                                id_map[_label_id] = _frame_id
+            else:
+                for _fids, _flids in zip(_frame_ids, _label_ids):
+                    for _frame_id, _label_id in zip(_fids, _flids):
+                        id_map[_label_id] = _frame_id
```
exupero, line 2958: What's your opinion on dict comprehensions?
```suggestion
                id_map = {_label_id: _frame_id
                          for _fids, _flids in zip(_frame_ids, _label_ids)
                          for _frame_id, _label_id in zip(_fids, _flids)}
```

fiftyone/utils/data/base.py
kaixi-wang: Replacing values with _iter_values() would allow processing while streaming data rather than materializing everything at once and avoids the multiple iterations over the lists.

But ideally, this should be expressed as an aggregation or update_many pipeline to apply the change directly in MongoDB. That avoids client-side iteration entirely, improves performance, and eliminates concurrency concerns from working with stale data.

kaixi-wang: For example, working off your example in the docstrings, to add a category field on each detection object based on the label, it could look something like this: 
```
ANIMALS = [
    "bear", "bird", "cat", "cow", "dog", "elephant", "giraffe",
    "horse", "sheep", "zebra"
]
PLANTS = [
    "flower", "grass", "tree"
]
default_value = None
mapping = {a: "animal" for a in ANIMALS}
mapping.update({p: "plant" for p in PLANTS})
branches = [
    {
        'case': {
            '$eq': [
                '$$det.label',
                key]},
        'then': value} for key, value in mapping.items()
]
new_detection_field = "category"
ds._sample_collection.update_many({}, {'$set': {'predictions.detections': {'$map': {'as': 'det',
                                                                                    'in': {'$mergeObjects': ['$$det',
                                                                                                             {
                                                                                                                 new_detection_field: {
                                                                                                                     '$switch': {
                                                                                                                         'branches': branches,
                                                                                                                         'default': default_value}}}]},
                                                                                    'input': 
                                                                                        '$predictions.detections'}}}})
```
Something like this requires minimal i/o and works for any scale ( as long as the mapping < 16MB), is straightforward, and doesn't require executing the same filter/query more than once over the collection.

brimoor: Your proposal is exactly what `view.map_values(path, map).save(root)` does. And indeed, it is used when it can be: https://github.com/voxel51/fiftyone/blob/6616dabcba5bd55634880da08e4266a4f1eafd64/fiftyone/utils/data/base.py#L102-L107 

However, as I explained in the PR description, we cannot use that approach in situations where `view` filters the field whose attributes we are editing, as `$set` would overwrite the **entire field**.

fiftyone/core/collections.py
brimoor: oh nice, I didn't know `disable-next=no-member` existed 🥇

fiftyone/core/collections.py
brimoor: I personally like to use ternaries when they fit on one line

fiftyone/core/collections.py
brimoor: I wrote it with a `for` loop so that it visually looks similar to the other branches in the methods that require more complex nested `for` loops

fiftyone/core/collections.py
brimoor: I personally like to use them when they fit on one line, but I find nested multiline comprehensions can sometimes be harder to read.

fiftyone/core/collections.py
brimoor: Actually in this case there are *two* `view` property accesses in this method that need to be ignored, so its a bit cleaner to just ignore once for the rest of the method

exupero (approved):
Looks good to me; didn't test.

fiftyone/utils/data/base.py
kaixi-wang: iter_values() still would be better

Also, the problem is not that it's not possible at all with the database. Taking a look at the aggregation involved in the bug ticket, the problem is that the view redefines the original field, so just appending to that view will cause overwriting the entire field. For example:
``` 
{'$addFields': {'predictions.detections': {'$filter': {'input': '$predictions.detections', 'cond': {'$gt': [{'$size': {'$ifNull': [{'$filter': {'input': '$$this.tags', 'as': 'this', 'cond': {'$in': ['$$this', ['turkey']]}}}, []]}}, 0]}}}}},
```

kaixi-wang (commented):
The issue from the ticket is the addFields in the aggregation that overwrites the unfiltered array. I think that can be addressed without requiring reading all values locally

fiftyone/utils/data/base.py
brimoor: `iter_values()` can't be used (or at least it wouldn't do anything different in most cases) because we're calling `view.values()` not `dataset.values()`

brimoor: > Also, the problem is not that it's not possible at all with the database. Taking a look at the aggregation involved in the bug ticket, the problem is that the view redefines the original field, so just appending to that view will cause overwriting the entire field

Right, this is exactly the problem that needs a solution. When the input `view` modifies a field, it results in an aggregation pipeline that uses `$set` or its alias `$addFields`, and in those cases we cannot implement this operation by appending a `$merge` to the aggregation pipeline, nor can we implement it like this:

```
dataset._sample_collection.update_many({}, {'$set': {...}})
```

because the user is not requesting that every sample/label in the dataset have its values mapped but rather a particular subset of the samples and/or labels as described by the current `view` on which they are invoking `map_values()`.

brimoor: Repeating once more for clarity:

The usage of `values()` + `set_values()` introduced in this PR is **ONLY** used in the specific case where `foud.map_values(view, ...)` is called on a `view` that edits the field whose values are being mapped:
https://github.com/voxel51/fiftyone/blob/6616dabcba5bd55634880da08e4266a4f1eafd64/fiftyone/utils/data/base.py#L102-L107 

In all other cases, `view.map_values(path, map).save(root)` is used under the hood, which performs the operation fully in the database.

PR #6150, opened 2025-07-16
Use GeoWithin for Map Queries

ritch:
Refactor the map panel selection to use GeoWithin instead of Select (by id).

[FOEPD-1690](https://voxel51.atlassian.net/browse/FOEPD-1690)

[FOEPD-1690]: https://voxel51.atlassian.net/browse/FOEPD-1690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added support for spatial selection using polygons, allowing users to select data based on geographic areas.
* **Improvements**
  * Enhanced consistency and reliability when handling spatial selections and active fields in map interactions.
* **Bug Fixes**
  * Prevented override updates when a spatial selection is active, ensuring correct selection behavior.

### DIFF
--- a/app/packages/state/src/recoil/selectors.ts
+++ b/app/packages/state/src/recoil/selectors.ts
@@ -472,13 +472,26 @@ export const similarityMethods = selector<{
 export const extendedStagesUnsorted = selector({
   key: "extendedStagesUnsorted",
   get: ({ get }) => {
-    const sampleIds = get(atoms.extendedSelection)?.selection;
+    const extendedSelection = get(atoms.extendedSelection);
+    const sampleIds = extendedSelection?.selection;
+    const spatialSelection = extendedSelection?.spatialSelection;
     const extendedSelectionOverrideStage = get(
       atoms.extendedSelectionOverrideStage
     );
 
     if (extendedSelectionOverrideStage) {
       return extendedSelectionOverrideStage;
+    }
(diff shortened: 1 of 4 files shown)


app/packages/state/src/recoil/selectors.ts
```diff
@@ -472,14 +472,27 @@ export const similarityMethods = selector<{
...
       return extendedSelectionOverrideStage;
     }
+
+    if (spatialSelection) {
+      return {
+        "fiftyone.core.stages.GeoWithin": {
+          boundary: spatialSelection.polygon,
+          location_field: spatialSelection.field,
+          strict: true,
+          create_index: true,
```
brimoor: We should use `create_index=False` here and instead require users to create their own spatial index offline if they require it.

Note that I'm going to update this to `create_index=False` as well in a separate PR for the same reason:
https://github.com/voxel51/fiftyone/blob/5b055b70fa1cb5c5a25a0e5b8ea721f20ca4ee20/fiftyone/core/stages.py#L3370-L3371

Creating view stages shouldn't have expensive side effects like creating indexes.

PR #6151, opened 2025-07-17
Cherry-pick: fix(plugin-core-tests): use str instead of PosixPath

ritch:


kaixi-wang (approved):
lgtm
just double checking, this is just because of how the fixture is defned and not a potential issue in the sdk itself?

PR #6161, opened 2025-07-21
FOEPD-306 No singleton cache mode

swheaton:
- Make singleton cache optional (`fo.config.singleton_cache` defaults `True`)
- Default to _disabling_ singleton cache for the app server
- Consolidate `refresh()` calls into `load_dataset(name, refresh=True)`. So that that function can take care of whether to reload or not (if no singleton cache we don't need to reload even if refresh is True).

Utilizes https://github.com/voxel51/fiftyone/pull/3956 and https://github.com/voxel51/fiftyone/pull/3945
It was easier to redo the changes than to merge/rebase

Testing document (mostly about app)
https://docs.google.com/document/d/1lrkmwq6ec2mtEFJ3JTdm9IuvVWv2lWgjHWRVzBe4Tgg/edit?tab=t.0

### Usage in SDK
But also note that it can help performance tremendously in certain cases of SDK usage.

```python
import time
import fiftyone as fo
ds = fo.load_dataset("bdd100k-validation")

def test_it(ds):
  # We have to hold a ref to the sample otherwise the singleton will be kicked out of the weakref dict
  samps=[]
  for s in ds:
    samps.append(s)

  st = time.time()
  ds.set_values("blah", [51] * len(ds))
  et = time.time()

  del samps
  return et-st

fo.config.singleton_cache=False
print("No Singleton:", test_it(ds))
fo.config.singleton_cache=True
print("Singleton:", test_it(ds))
```
```
No Singleton: 0.3039100170135498
Singleton: 48.95887994766235
```

## Question
Should we disable singleton cache across the board? Currently unable to find a case where having the singleton makes anything significantly faster in the first place.

        notes for FiftyOne users.

Disables singleton cache by default for the app server.
Re-enable by setting `fo.config.singleton_cache=True`

-   [ ] Other

### DIFF
--- a/fiftyone/core/config.py
+++ b/fiftyone/core/config.py
@@ -280,6 +280,12 @@ class FiftyOneConfig(EnvConfig):
             env_var="FIFTYONE_EXECUTION_CACHE_ENABLED",
             default=True,
         )
+        self.singleton_cache = self.parse_bool(
+            d,
+            "singleton_cache",
+            env_var="FIFTYONE_SINGLETON_CACHE",
+            default=True,
+        )
         self._init()
 
     @property
--- a/fiftyone/core/config.py
+++ b/fiftyone/core/config.py
@@ -347,6 +347,10 @@ class FiftyOneConfig(EnvConfig):
                         e,
                     )
 
+        # Default no singleton cache for fiftyone app server
+        if os.environ.get("FIFTYONE_SERVER", False):
+            self.singleton_cache = True
+
         if self.timezone and self.timezone.lower() not in {"local", "utc"}:
             try:
                 pytz.timezone(self.timezone)
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -142,7 +142,7 @@ def _validate_dataset_name(name, skip=None):
     return slug
 
 
-def load_dataset(name, create_if_necessary=False):
+def load_dataset(name, create_if_necessary=False, reload=False):
     """Loads the FiftyOne dataset with the given name.
 
     To create a new dataset, use the :class:`Dataset` constructor.
@@ -156,6 +156,7 @@ def load_dataset(name, create_if_necessary=False):
     Args:
         name: the name of the dataset
         create_if_necessary (False): if no dataset exists, create an empty one
+        reload (False): whether to reload the dataset if necessary
 
     Raises:
         DatasetNotFoundError: if the dataset does not exist and
@@ -165,7 +166,7 @@ def load_dataset(name, create_if_necessary=False):
         a :class:`Dataset`
     """
     try:
-        return Dataset(name, _create=False)
+        return Dataset(name, _create=False, _reload=reload)
     except DatasetNotFoundError as ex:
         if create_if_necessary:
             return Dataset(name, _create=True)
@@ -314,6 +315,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         overwrite=False,
         _create=True,
         _virtual=False,
+        _reload=False,
         **kwargs,
     ):
         if name is None and _create:
--- a/fiftyone/core/odm/utils.py
+++ b/fiftyone/core/odm/utils.py
@@ -698,13 +698,14 @@ class DocumentRegistryError(Exception):
 _document_registry = DocumentRegistry()
 
 
-def load_dataset(id=None, name=None):
+def load_dataset(id=None, name=None, reload=False):
     """Loads the dataset from the database by its unique id or name. Throws
     an error if neither id nor name is provided.
 
     Args:
         id (None): the unique id of the dataset
         name (None): the name of the dataset
+        reload (False): whether to reload the dataset if necessary
 
     Returns:
         a :class:`fiftyone.core.dataset.Dataset`
@@ -713,20 +714,20 @@ def load_dataset(id=None, name=None):
     import fiftyone.core.dataset as fod
 
     if name:
-        return fod.load_dataset(name)
+        return fod.load_dataset(name, reload=reload)
 
     if not id:
         raise ValueError("Must provide either id or name")
 
     db = foo.get_db_conn()
     try:
-        uid = ObjectId(id)
+        _id = ObjectId(id)
     except:
         # Although _id is an ObjectId by default, it's possible to set it to
         # something else
-        uid = id
+        _id = id
 
-    res = db.datasets.find_one({"_id": uid}, {"name": True})
+    res = db.datasets.find_one({"_id": _id}, {"name": True})
     if not res:
-        raise ValueError(f"Dataset with _id={uid} does not exist")
-    return fod.load_dataset(res.get("name"))
+        raise ValueError(f"Dataset with _id={_id} does not exist")
+    return fod.load_dataset(res.get("name"), reload=reload)
--- a/fiftyone/core/singletons.py
+++ b/fiftyone/core/singletons.py
@@ -9,6 +9,11 @@ FiftyOne singleton implementations.
 from collections import defaultdict
 import weakref
 
+import fiftyone.core.utils as fou
+
+# TODO fix this
+fo = fou.lazy_import("fiftyone")
+
 
 class DatasetSingleton(type):
     """Singleton metaclass for :class:`fiftyone.core.dataset.Dataset`.
@@ -24,7 +29,7 @@ class DatasetSingleton(type):
         cls._instances = weakref.WeakValueDictionary()
         return cls
 
-    def __call__(cls, name=None, _create=True, *args, **kwargs):
+    def __call__(cls, name=None, _create=True, _reload=False, *args, **kwargs):
         instance = cls._instances.pop(name, None)
 
         if (
@@ -46,8 +51,13 @@ class DatasetSingleton(type):
                 return cls.__call__(
                     name=name, _create=_create, *args, **kwargs
                 )
+            if _reload:
+                instance.reload()
 
-        cls._instances[name] = instance
+        if fo.config.singleton_cache:
+            # If the singleton cache is enabled, we store the instance
+            #  so that it can be retrieved later
+            cls._instances[name] = instance
 
         return instance
 
@@ -110,7 +120,8 @@ class SampleSingleton(DocumentSingleton):
         return cls
 
     def _register_instance(cls, obj):
-        cls._instances[obj._doc.collection_name][obj.id] = obj
+        if fo.config.singleton_cache:
+            cls._instances[obj._doc.collection_name][obj.id] = obj
 
     def _get_instance(cls, doc):
         try:
@@ -261,9 +272,10 @@ class FrameSingleton(DocumentSingleton):
         return cls
 
     def _register_instance(cls, obj):
-        cls._instances[obj._doc.collection_name][obj.sample_id][
-            obj.frame_number
-        ] = obj
+        if fo.config.singleton_cache:
+            cls._instances[obj._doc.collection_name][obj.sample_id][
+                obj.frame_number
+            ] = obj
 
     def _get_instance(cls, doc):
         try:
--- a/fiftyone/core/singletons.py
+++ b/fiftyone/core/singletons.py
@@ -11,7 +11,8 @@ import weakref
 
 import fiftyone.core.utils as fou
 
-# TODO fix this
+# This is required to avoid circular imports, in order to get `fo.config` at
+#   runtime.
 fo = fou.lazy_import("fiftyone")
 
 
@@ -29,7 +30,16 @@ class DatasetSingleton(type):
         cls._instances = weakref.WeakValueDictionary()
         return cls
 
+    def _make_instance(cls, name, _create, *args, **kwargs):
+        instance = cls.__new__(cls)
+        instance.__init__(name=name, _create=_create, *args, **kwargs)
+        return instance
+
     def __call__(cls, name=None, _create=True, _reload=False, *args, **kwargs):
+        skip_cache = not fo.config.singleton_cache
+        if skip_cache:
+            return cls._make_instance(name, _create, *args, **kwargs)
+
         instance = cls._instances.pop(name, None)
 
         if (
@@ -38,8 +48,9 @@ class DatasetSingleton(type):
             or instance.deleted
             or instance.name is None
         ):
-            instance = cls.__new__(cls)
-            instance.__init__(name=name, _create=_create, *args, **kwargs)
+            instance = cls._make_instance(
+                name=name, _create=_create, *args, **kwargs
+            )
             name = instance.name  # `__init__` may have changed `name`
         else:
             try:
@@ -54,10 +65,9 @@ class DatasetSingleton(type):
             if _reload:
                 instance.reload()
 
-        if fo.config.singleton_cache:
-            # If the singleton cache is enabled, we store the instance
-            #  so that it can be retrieved later
-            cls._instances[name] = instance
+        # If the singleton cache is enabled, we store the instance
+        #  so that it can be retrieved later
+        cls._instances[name] = instance
 
         return instance
 
--- a/fiftyone/core/stages.py
+++ b/fiftyone/core/stages.py
@@ -8168,7 +8168,7 @@ class ToPatches(ViewStage):
             name = None
 
         try:
-            last_dataset = fod.load_dataset(name)
+            last_dataset = fod.load_dataset(name, reload=reload)
         except:
             last_dataset = None
 
@@ -8328,7 +8328,7 @@ class ToEvaluationPatches(ViewStage):
             name = None
 
         try:
-            last_dataset = fod.load_dataset(name)
+            last_dataset = fod.load_dataset(name, reload=reload)
         except:
             last_dataset = None
 
@@ -8501,7 +8501,7 @@ class ToClips(ViewStage):
             name = None
 
         try:
-            last_dataset = fod.load_dataset(name)
+            last_dataset = fod.load_dataset(name, reload=reload)
         except:
             last_dataset = None
 
@@ -8651,7 +8651,7 @@ class ToTrajectories(ViewStage):
             name = None
 
         try:
-            last_dataset = fod.load_dataset(name)
+            last_dataset = fod.load_dataset(name, reload=reload)
         except:
             last_dataset = None
 
@@ -8857,7 +8857,7 @@ class ToFrames(ViewStage):
             name = None
 
         try:
-            last_dataset = fod.load_dataset(name)
+            last_dataset = fod.load_dataset(name, reload=reload)
         except:
             last_dataset = None
 
--- a/fiftyone/core/state.py
+++ b/fiftyone/core/state.py
@@ -162,25 +162,16 @@ class StateDescription(etas.Serializable):
         """
         dataset = d.get("dataset", None)
         if dataset is not None:
-            dataset = fod.load_dataset(dataset)
+            dataset = fod.load_dataset(dataset, reload=True)
 
         stages = d.get("view", None)
         view = None
         view_name = d.get("view_name", None)
         if dataset is not None:
             if view_name:
-                try:
-                    view = dataset.load_saved_view(view_name)
-                except Exception as e:
-                    dataset.reload()
-                    view = dataset.load_saved_view(view_name)
-
+                view = dataset.load_saved_view(view_name)
             elif stages:
-                try:
-                    view = fov.DatasetView._build(dataset, stages)
-                except:
-                    dataset.reload()
-                    view = fov.DatasetView._build(dataset, stages)
+                view = fov.DatasetView._build(dataset, stages)
 
         config = (
             fo.AppConfig.from_dict(d["config"])
--- a/fiftyone/server/mutation.py
+++ b/fiftyone/server/mutation.py
@@ -89,7 +89,9 @@ class Mutation(SetColorScheme):
         info: Info,
     ) -> bool:
         state = get_state()
-        state.dataset = fod.load_dataset(name) if name is not None else None
+        state.dataset = (
+            fod.load_dataset(name, reload=True) if name is not None else None
+        )
 
         state.color_scheme = build_color_scheme(
             None, state.dataset, state.config
@@ -218,7 +220,7 @@ class Mutation(SetColorScheme):
             return None
 
         result_view = None
-        ds = fod.load_dataset(dataset_name)
+        ds = fod.load_dataset(dataset_name, reload=True)
         state.dataset = ds
 
         # Create the view using the saved view doc if loading a saved view
@@ -232,7 +234,7 @@ class Mutation(SetColorScheme):
         # Otherwise, build the view using the params
         if result_view is None:
             result_view = await get_view(
-                dataset_name,
+                ds,
                 stages=view if view else None,
                 filters=form.filters if form else None,
                 extended_stages=form.extended if form else None,
@@ -290,7 +292,7 @@ class Mutation(SetColorScheme):
         dataset = state.dataset
         use_state = dataset is not None
         if dataset is None:
-            dataset = fod.load_dataset(dataset_name)
+            dataset = fod.load_dataset(dataset_name, reload=True)
 
         if dataset is None:
             raise ValueError(
@@ -313,7 +315,6 @@ class Mutation(SetColorScheme):
             view_name, result_view, description=description, color=color
         )
         if use_state:
-            dataset.reload()
             state.view = dataset.load_saved_view(view_name)
             await dispatch_event(subscription, fose.StateUpdate(state=state))
 
@@ -341,7 +342,7 @@ class Mutation(SetColorScheme):
                 view_name,
             )
 
-        dataset = fod.load_dataset(dataset_name)
+        dataset = fod.load_dataset(dataset_name, reload=True)
         if not dataset:
             raise ValueError(f"No dataset found with name {dataset_name}")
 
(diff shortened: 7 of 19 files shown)


fiftyone/core/state.py
```diff
@@ -162,25 +162,16 @@ def from_dict(cls, d):
...
-                    view = dataset.load_saved_view(view_name)
-
+                view = dataset.load_saved_view(view_name)
             elif stages:
-                try:
-                    view = fov.DatasetView._build(dataset, stages)
-                except:
-                    dataset.reload()
-                    view = fov.DatasetView._build(dataset, stages)
+                view = fov.DatasetView._build(dataset, stages)
```
swheaton, line 174: @kaixi-wang in the branch you created, you had added this code back to `fix server view load`. Can you explain why that is necessary?

fiftyone/server/mutation.py
```diff
@@ -232,7 +234,7 @@ async def set_view(
         # Otherwise, build the view using the params
         if result_view is None:
             result_view = await get_view(
-                dataset_name,
+                ds,
```
swheaton, line 237: this seemed fine to me to avoid double loading, but lmk

fiftyone/server/query.py
```diff
@@ -607,11 +605,11 @@ async def serialize_dataset(
...
 ) -> Dataset:
     def run():
-        if not fod.dataset_exists(dataset_name):
+        try:
+            dataset = fo.load_dataset(dataset_name, reload=True)
+        except fod.DatasetNotFoundError:
             return None
 
-        dataset = fo.Dataset(dataset_name, _create=False, _force_load=True)
-        dataset.reload()
```
swheaton, line 614: @benjaminpkane inspect this to see if the original bug this code fixed would still show up or not (either singleton_cache mode or not)

fiftyone/server/routes/embeddings.py
```diff
@@ -79,7 +85,7 @@ def _post_sync(self, data):
             return {"error": msg}
 
         view = fosv.get_view(
-            dataset_name,
+            dataset,
```
swheaton, line 88: this seemed fine to me to avoid double loading, but lmk

fiftyone/server/routes/embeddings.py
```diff
@@ -193,7 +199,7 @@ def _post_sync(self, data):
         results = dataset.load_brain_results(brain_key)
 
         view = fosv.get_view(
-            dataset_name,
+            dataset,
```
swheaton, line 202: this seemed fine to me to avoid double loading, but lmk

fiftyone/server/samples.py
```diff
@@ -109,20 +109,11 @@ async def paginate_samples(
...
-    if after is None:
-        # first page, force dataset reload
-        view = await run_sync_task(run, True)
-    else:
-        # not first page, optimistically skip dataset reload
-        try:
-            view = await run_sync_task(run, False)
-        except:
-            view = await run_sync_task(run, True)
+    view = await run_sync_task(run)
```
swheaton, line 116: @kaixi-wang in the branch you created, you had added this code back to fix server view load. Can you explain why that is necessary?

fiftyone/server/routes/embeddings.py
```diff
@@ -64,12 +64,18 @@ def _post_sync(self, data):
             dataset = fosu.load_and_cache_dataset(dataset_name)
```
swheaton, line 64: @ritch take a look at `load_and_cache_dataset()` and see if it matches the expected behavior for both singleton and non singleton modes.

plugins/operators/model_evaluation/__init__.py
```diff
@@ -1069,7 +1070,10 @@ def get_scenario_names(self, ctx):
         return [scenario.get("name") for _, scenario in scenarios.items()]
 
     def resolve_input(self, ctx):
-        cache_dataset(self.get_dataset(ctx))
+        # Force ctx.dataset to load and store dataset so it remains consistent.
+        d = self.get_dataset(ctx)
+        if fo.config.singleton_cache:
+            cache_dataset(d)
```
swheaton: I'm not really sure what model eval is trying to do. Someone double check me, this was my best guess.

plugins/panels/model_evaluation/__init__.py
```diff
@@ -7,27 +7,28 @@
 """
 
 import os
-import fiftyone.utils.eval as foue
```
swheaton, line 10: headers were improperly ordered

plugins/panels/model_evaluation/__init__.py
```diff
@@ -1207,14 +1208,14 @@ def _init_segmentation_results(dataset, results, gt_field):
...
-    import fiftyone.server.utils as fosu
+    if fo.config.singleton_cache:
+        #
+        # Ensure the dataset singleton is cached so that subsequent callbacks on
+        # this panel will use the same `dataset` and hence `results`
+        #
+        import fiftyone.server.utils as fosu
 
-    fosu.cache_dataset(dataset)
+        fosu.cache_dataset(dataset)
```
swheaton: I'm not really sure what model eval is trying to do. Someone double check me, this was my best guess.

fiftyone/server/routes/embeddings.py
swheaton: Remembered ritchie is out for 2 weeks. @benjaminpkane ?

fiftyone/core/config.py
```diff
@@ -341,6 +347,10 @@ def _init(self):
                         e,
                     )
 
+        # Default no singleton cache for fiftyone app server
+        if os.environ.get("FIFTYONE_SERVER", False):
+            self.singleton_cache = True
```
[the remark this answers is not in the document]
swheaton: right. fixed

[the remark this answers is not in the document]
benjaminpkane: @swheaton The 🐰 is correct, right? The server should always have the cache disabled?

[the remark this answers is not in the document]
swheaton: yep bunny to the rescue!

plugins/operators/model_evaluation/__init__.py
kaixi-wang: I think it should be removed. `cache_dataset` already has a deprecation notice and testing model eval without it has so far not caused any issue

fiftyone/server/query.py
benjaminpkane: LGTM

fiftyone/server/samples.py
kaixi-wang: bad merge. I don't think it is necessary. I'll do some more testing to verify though

plugins/operators/model_evaluation/__init__.py
brimoor: `fosu.load_and_cache_dataset()` and `fosu.cache_dataset()` are used in situations where we are **explicitly wanting** to cache something in-memory to avoid having to reload it from the DB on every route call.

I believe the only specific use cases so far are:
- Embeddings panel: cache `load_brain_results()`
- Evaluation panel: cache `load_evaluation_results()`

```py
dataset = fo.load_dataset(...)

# Loading brain/eval results from the DB can take significant time
brain_results = dataset.load_brain_results(...)
eval_results = dataset.load_evaluation_results(...)

# However, once loaded, brain/eval results are cached on the `Dataset` object
also_brain_results = dataset.load_brain_results(...)  # nothing is loaded
also_eval_results = dataset.load_evaluation_results(...)  # nothing is loaded
```

brimoor: We need to maintain this caching behavior or else the panels will get very slow on datasets with big brain/eval results.

(yes, in a multiple App server context without sticky sessions, this style of caching is not a perfect solution; we'd need something like Redis to fully implement caching)

brimoor: Now that `Dataset` is not a singleton in App server context, we'll need to tweak the implementations of `fosu.load_and_cache_dataset()` and `fosu.cache_dataset()` and the panels that call them to ensure that the desired thing still happens:
- the `Dataset` object on which `load_XXX_results()` gets called is stored in a `_cache` and subsequent backend requests within the same feature that rely on accessing the same  `load_XXX_results()` will use this cached `Dataset` object rather than loading a fresh copy from the DB

fiftyone/server/routes/embeddings.py
brimoor: See my response here: https://github.com/voxel51/fiftyone/pull/6161#discussion_r2219471812

plugins/panels/model_evaluation/__init__.py
kaixi-wang: Also don't know this is doing what is intended... It looks like it's only used for _init_segmentation_results(), however I haven't been able to get scenario evaluations with segmentation to work on release/branches without singleton changes

plugins/operators/model_evaluation/__init__.py
swheaton: > cache_dataset already has a deprecation notice and testing model eval without it has so far not caused any issue

I added that in my branch. I now have removed it

fiftyone/server/samples.py
swheaton: ok. might be an area (second page load) where it's slower than with singleton

fiftyone/core/singletons.py
```diff
@@ -24,7 +30,16 @@ def __new__(metacls, *args, **kwargs):
         cls._instances = weakref.WeakValueDictionary()
         return cls
 
-    def __call__(cls, name=None, _create=True, *args, **kwargs):
+    def _make_instance(cls, name, _create, *args, **kwargs):
+        instance = cls.__new__(cls)
+        instance.__init__(name=name, _create=_create, *args, **kwargs)
+        return instance
+
```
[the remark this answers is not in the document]
swheaton: yeah but that's how it was in the `__new__()` function and I don't really want to change it in this PR.

plugins/operators/model_evaluation/__init__.py
swheaton: Based on my reading of the code, it appears to me that the embeddings stuff was for results caching purposes.

And the model eval stuff was so that it would utilize the same dataset throughout the panel life. But `ctx.dataset` property holds the dataset object in `ctx._dataset` so that will keep it around without the need for artificially "caching" it.

benjaminpkane (approved):
Server code LGTM

brimoor (changes requested):
Requesting changes until this thread is fully resolved and we've verified that `load_XXX_results()` is not getting called multiple times when using the Embeddings/Evaluation panels (at least in OSS): https://github.com/voxel51/fiftyone/pull/6161#discussion_r221947181

plugins/operators/model_evaluation/__init__.py
kaixi-wang: Also, it's possible subsequent requests aren't going to the same plugin pod, so I don't think it's possible to guarentee the same dataset throughout the lifetime of the panel anyways

(regardless I'm currently testing with and without caching)

kaixi-wang: > We need to maintain this caching behavior or else the panels will get very slow on datasets with big brain/eval results.

I've been testing with and without cache enabled in teams with log/print statements throughout, and for the model eval panel, even with the dataset caching disabled and singleton cache disabled, I do not see evals being reloaded when switching comparisons (after the first load).  In fact, I don't see any loading from the evaluations cache aside from saving to it when it's enabled. 

For embeddings, I do see it loading from cache, however, I have not observed any change in performance in the datasets I have tested with (<~20k)

fiftyone/core/odm/utils.py
```diff
@@ -698,13 +698,14 @@ class DocumentRegistryError(Exception):
 _document_registry = DocumentRegistry()
 
 
-def load_dataset(id=None, name=None):
+def load_dataset(id=None, name=None, reload=False):
```
brimoor, line 701: FYI I'm proposing to move this function in https://github.com/voxel51/fiftyone/pull/6067, so there will be a merge conflict

fiftyone/core/stages.py
```diff
@@ -8168,7 +8168,7 @@ def load_view(self, sample_collection, reload=False):
             name = None
 
         try:
-            last_dataset = fod.load_dataset(name)
+            last_dataset = fod.load_dataset(name, reload=reload)
```
brimoor: Note that https://github.com/voxel51/fiftyone/pull/6067 tweaks the syntax of how `last_dataset` is loaded in generated views:

```py
        try:
            # `_force_load=True` forces an exception to be raised if the
            # dataset has been deleted but this process' singleton doesn't
            # know about it
            last_dataset = fod.Dataset(name, _create=False, _force_load=True)
        except:
```

So now that the new `reload` parameter exists in `load_dataset()`, we basically **ALWAYS** want this:

```py
last_dataset = fod.load_dataset(name, reload=True)
```

The `reload` parameter of `load_view()` means something slightly different.

plugins/panels/model_evaluation/__init__.py
brimoor: The key objective is to ensure that `load_evaluation_results()` is not called multiple times per Model Evaluation panel session (at least in OSS where there's a single App server process) **specifically for segmentation evaluations**, because the rest of this method does some fairly expensive data transformation + storage on the `results` object that we don't want to regenerate every time a callback is triggered in the panel.

brimoor: So to achieve that in a world where `fo.load_dataset()` doesn't return singletons, we will likely have to make tweaks to methods in this panel's methods.

Using existing methods, I guess it would look something like this:

```py
# swap `ctx.dataset` for a `dataset` that supports cached `dataset.results`
dataset = fosu.load_and_cache_dataset(ctx.dataset.name)

# this is a no-op if results were already loaded
results = dataset.load_evaluation_results(eval_key)
```

fiftyone/server/samples.py
kaixi-wang: Scrolling doesn't feel different

fiftyone/core/stages.py
swheaton: got it, changed here  - 224a5b17ecccd3e2d5d003bf77eddfbc069ae619

kaixi-wang: I'm concerned setting this true will cause notable performance issues/slowdowns

fiftyone/core/stages.py
```diff
@@ -8168,7 +8168,9 @@ def load_view(self, sample_collection, reload=False):
             name = None
 
         try:
-            last_dataset = fod.load_dataset(name)
+            # Always reload the main dataset, the `reload` parameter is for
+            #   reloading patches dataset
+            last_dataset = fod.load_dataset(name, reload=True)
```
kaixi-wang, line 8173: I'm concerned setting this true will cause notable performance issues/slowdowns

swheaton: Can you explain please, I didn't fully understand what you were saying earlier

plugins/panels/model_evaluation/__init__.py
brimoor: @swheaton just checking- are there more changes coming?

Or are there objections to my stance that we need to preserve caching behavior that `fosu.load_and_cache_dataset()` and `fosu.cache_dataset()` offer for Embeddings/Evaluation panels before shipping this?

swheaton: Oh. I guess I was thinking that what I did would preserve the behavior already. Lmk where you think this is not correct.

1. `load_and_cache_dataset()` / `embeddings.py`
   - singleton: loads (singleton-cached) dataset and stores a reference into the pseudo-cache to keep the reference alive.
   - no singleton: Uses the actual cache
```python
    if not fo.config.singleton_cache:
        # If singleton cache is disabled, return the cached dataset if it exists
        dataset = _cache.get(name)
        if dataset:
            dataset.reload()
            return dataset
```
2. Model Eval panel
   - singleton: same behavior
   - no singleton: the below comment indicated using the same dataset was the only concern. Since `ctx.dataset` is basically a local cache of the one dataset (stores in `ctx._dataset`), I figured the behavior would be the same? Unless the panel doesn't retain the same `ExecutionContext`, then that would not be the case.

```
        #
        # Ensure the dataset singleton is cached so that subsequent callbacks on
        # this panel will use the same `dataset` and hence `results`
        #
```

swheaton: Will go ahead and make that change real quick

swheaton: Although, singleton cache is _already_ turned off for operators/panels via `contextvars` right? So the existing code is not doing much anyway?

brimoor: @swheaton okay here's a proposal that implements the logic I think we need to allow Embeddings/Model Evaluation panels to use cached run results: https://github.com/voxel51/fiftyone/pull/6176

brimoor: ^happy to chat live about this as needed

swheaton: k accepted the proposal

plugins/operators/model_evaluation/__init__.py
```diff
@@ -1069,7 +1056,8 @@ def get_scenario_names(self, ctx):
         return [scenario.get("name") for _, scenario in scenarios.items()]
 
     def resolve_input(self, ctx):
-        cache_dataset(self.get_dataset(ctx))
+        # force `ctx.dataset` to be cached
+        _ = self.get_dataset(ctx)
```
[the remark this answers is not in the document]
swheaton: the point is to cache it separately regardless of singleton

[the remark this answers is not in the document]
brimoor: ignoring `singleton_cache` is intentional in this case

plugins/operators/model_evaluation/__init__.py
```diff
@@ -53,18 +50,13 @@ def config(self):
...
-        ttl=PROMPT_SCOPED_CACHE_TTL,
-    )
+    # we use `fosu.cache_dataset()` rather than `@execution_cache` here so that
+    # the cached dataset can be reused outside of the current prompt session
     def get_dataset(self, ctx):
         """
         Returns the dataset for the current context.
         """
-        return ctx.dataset
+        return fosu.cache_dataset(ctx.dataset)
```
imanjra, line 59: From the testing I did a few weeks ago, caching data with the execution cache was needed for faster access after first time.

plugins/operators/model_evaluation/__init__.py
```diff
@@ -196,13 +188,8 @@ def get_subset_def_data_for_eval_key(
...
         return key
 
-    @execution_cache(
-        prompt_scoped=True,
-        residency="ephemeral",
-        serialize=dataset_serialize_deserialize,
-        deserialize=dataset_serialize_deserialize,
-        ttl=PROMPT_SCOPED_CACHE_TTL,
-    )
+    # there is no need to use `@execution_cache` here because evaluation
```
imanjra, line 191: same as above. It was needed from my testing

plugins/operators/model_evaluation/__init__.py
brimoor: Thanks for joining the party @imanjra! We should definitely validate that caching is indeed happening as expected before merging this PR 💯 

Here's what I think is going on:
1. I did not realize that `Dataset` singleton behavior was in fact **disabled** for plugin execution, which meant that the usage of `fosu.cache_dataset()` that I had originally added on `develop` was not actually sufficient to ensure that the `Dataset` objects on which evaluation results were cached were actually reused by subsequent API requests that use them
2. The proposed changes in this PR take into account that `Dataset` are not singletons. Specifically, I believe that now `ctx.dataset` has been replaced with `fosu.cache_dataset(ctx.dataset)` in all methods that access evaluation results, meaning that they should all now benefit from a shared cached `Dataset` object.

imanjra: Nicee!

imanjra (approved):

plugins/operators/model_evaluation/__init__.py
swheaton: @imanjra can you give steps for repeating this testing on this PR? Thanks

> From the testing I did a few weeks ago, caching data with the execution cache was needed for faster access after first time.

imanjra: @swheaton it was basically bunch of print statement with elapsed time calculated with `time` module. Then I deployed to ephem env and used create scenarios flow in ME panel on bdd10k with 100,000 samples (10k samples cloned 10 times)

Let me gather all the lines of `fiftyone/plugins/operators/model_evaluation/__init__.py` I benchmarked

imanjra: If you'd like, I can deploy this to my ephem and give it a go.

imanjra: Lines benchmarked:
- https://github.com/voxel51/fiftyone/blob/6ffb7d3604000d4caf1c91fc749fa54662208cb2/plugins/operators/model_evaluation/__init__.py#L199
- https://github.com/voxel51/fiftyone/blob/6ffb7d3604000d4caf1c91fc749fa54662208cb2/plugins/operators/model_evaluation/__init__.py#L202
- https://github.com/voxel51/fiftyone/blob/6ffb7d3604000d4caf1c91fc749fa54662208cb2/plugins/operators/model_evaluation/__init__.py#L231

- Entire function at https://github.com/voxel51/fiftyone/blob/6ffb7d3604000d4caf1c91fc749fa54662208cb2/plugins/operators/model_evaluation/__init__.py#L412
- https://github.com/voxel51/fiftyone/blob/6ffb7d3604000d4caf1c91fc749fa54662208cb2/plugins/operators/model_evaluation/__init__.py#L1071
- https://github.com/voxel51/fiftyone/blob/6ffb7d3604000d4caf1c91fc749fa54662208cb2/plugins/operators/model_evaluation/__init__.py#L1095 to Line 1113
- The entire resolve_input

swheaton: @imanjra if you can, that would be much appreciated. You'll be much faster at recreating previous testing than me trying to do it fresh

fiftyone/core/dataset.py
```diff
@@ -314,6 +315,7 @@ def __init__(
         overwrite=False,
         _create=True,
         _virtual=False,
+        _reload=False,
```
imanjra, line 318: noticed this when resolving conflict... is this used/needed?

swheaton: Yes. It's a little tricky but while it is not used in this class' `__init__` function, it is passed to `fiftyone.core.DatasetSingleton.__call__()` where it is used if the dataset comes from the singleton cache, and unused if the dataset is new or if singleton cache is disabled.

plugins/operators/model_evaluation/__init__.py
kaixi-wang: Yes same. Execution cache as opposed to the singleton cache was the only thing that actually seemed to effectively use cached results.

fiftyone/core/stages.py
kaixi-wang: Reloading the patches dataset sounds like potentially regenerating the patches collection or otherwise syncing the fields, which could make loading take unexpectedly long and in many cases is an unnecessary bottleneck... but I haven't gotten a chance to test/dig into the code so I guess just ignore for now...

fiftyone/server/utils.py
```diff
@@ -25,43 +25,58 @@ def load_and_cache_dataset(name):
...
         dataset: a :class:`fiftyone.core.dataset.Dataset`
+
+    Returns:
+        a :class:`fiftyone.core.dataset.Dataset`
     """
-    _cache[dataset.name] = dataset
+    return _cache_dataset(dataset.name, dataset=dataset)
+
+
+def _cache_dataset(name, dataset=None):
```
kaixi-wang, line 61: is there a reason to not just rely on the dataset.name? What if a dataset is renamed? Would it be possible to cache by id?

plugins/operators/model_evaluation/__init__.py
```diff
@@ -36,10 +37,6 @@
 PROMPT_SCOPED_CACHE_TTL = 60 * 60  # 1 hour
 
 
-def dataset_serialize_deserialize(dataset):
```
kaixi-wang, line 39: why did this function even exist  :(

plugins/operators/model_evaluation/__init__.py
kaixi-wang: @imanjra did you confirm that removing the execution was fine or that the cached evaluation results are ever read when interacting with the app?

imanjra: Yah, I was able to verify it working with these changes. EC can be removed.

plugins/operators/model_evaluation/__init__.py
imanjra: Verified. Works great! 🔥

plugins/operators/model_evaluation/__init__.py
imanjra: The default ExecutionCache serializer and deserializer cannot serialize/deserialize dataset. For `ephemeral` residency (in-memory) cache, we don't need to serialize or deserialize.

_Serializing is mainly for storing cached values in MongoDB or similar_

kaixi-wang (approved):
Didn't test the latest myself but I'm good with Ibrahim having tested the execution cache changes and my testing before looked fine

fiftyone/core/stages.py
swheaton: yeah its just going to reload the base dataset if necessary. The logic below determines whether it should regenerate patches collection I think.

PR #6169, opened 2025-07-23
Add builtin operator to convert existing indexes to unique

kaixi-wang:

- Add operator to convert existing indexes to unique or check for key violations to correct first
- Can be used to make the filepath field unique without the need to drop and recreate the index

- Tested as a plugin 
- 
image
Pasted Graphic 3
image

- Added unit tests

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

* **New Features**
  * Added an operator to convert non-unique dataset indexes to unique indexes, including a dry run option to detect uniqueness violations before applying changes.

* **Tests**
  * Added comprehensive unit tests covering input validation, dry run functionality, error handling, and successful uniqueness enforcement.

### DIFF
--- a/plugins/operators/__init__.py
+++ b/plugins/operators/__init__.py
@@ -1500,10 +1500,7 @@ class ConvertIndexToUnique(foo.Operator):
             collection_name = ctx.dataset._sample_collection_name
 
         index_info = ctx.dataset.get_index_information()[field]
-        key_pattern = {
-            field_name: direction
-            for field_name, direction in index_info["key"]
-        }
+        key_pattern = dict(index_info["key"])
 
         # Special sequence for converting to unique index
         # https://www.mongodb.com/docs/manual/core/index-unique/convert-to-unique/
--- a/plugins/operators/__init__.py
+++ b/plugins/operators/__init__.py
@@ -1510,80 +1510,78 @@ class ConvertIndexToUnique(foo.Operator):
         # https://www.mongodb.com/docs/manual/core/index-unique/convert-to-unique/
 
         # 1. Prepare the collection for unique index conversion
-        result = self._safe_collmod(
+        err_result = self._safe_collmod(
             db,
             collection_name,
-            dry_run=dry_run_only,
             index_spec={"keyPattern": key_pattern, "prepareUnique": True},
-            error_result={
-                **result,
-                "message": "Unable to prepare collection for unique index conversion.",
-            },
+            error_msg="Unable to prepare collection for unique index conversion.",
         )
-        if result.get("status") == "error":
-            return result
+        if err_result:
+            return {**result, **err_result}
 
         # 2. Check for unique key violations
-        result = self._safe_collmod(
+        err_result = self._safe_collmod(
             db,
             collection_name,
             dry_run=dry_run_only,
             index_spec={"keyPattern": key_pattern, "unique": True},
-            error_result={
-                **result,
-                "message": "Error while checking for unique key violations.",
-            },
+            error_msg="Error while checking for unique key violations.",
         )
 
-        if dry_run_only or result["status"] == "error":
+        if dry_run_only or err_result:
             # If dry run or error, restore the collection to its original state and return
-            return self._safe_collmod(
+            self._safe_collmod(
                 db,
                 collection_name,
-                dry_run=dry_run_only,
                 index_spec={"keyPattern": key_pattern, "prepareUnique": False},
-                success_result={
-                    **result,
-                    "message": "Dry run completed successfully. No unique key violations exist currently.",
-                },
             )
+            if err_result:
+                return {**result, **err_result}
+            return {
+                **result,
+                "status": "success",
+                "message": "No unique key violations found.",
+            }
 
         # 3. If no unique key violations, proceed to make the index unique
-        return self._safe_collmod(
+        err_result = self._safe_collmod(
             db,
             collection_name,
             dry_run=False,
             index_spec={"keyPattern": key_pattern, "unique": True},
-            success_result={
-                **result,
-                "message": f"Unique index conversion completed successfully for {field=}.",
-            },
-            error_result={
-                **result,
-                "message": f"Error while making index '{field}' unique.",
-            },
+            error_msg=f"Error while making index '{field}' unique.",
         )
+        if err_result:
+            # If error, restore the collection to its original state and return
+            self._safe_collmod(
+                db,
+                collection_name,
+                index_spec={"keyPattern": key_pattern, "prepareUnique": False},
+            )
+            return {**result, **err_result}
+        return {
+            **result,
+            "status": "success",
+            "message": "Unique index conversion successful.",
+        }
 
     def _safe_collmod(
         self,
         db,
         collection_name,
-        dry_run=False,
+        dry_run=None,
         index_spec=None,
-        success_result=None,
-        error_result=None,
+        error_msg=None,
     ):
         try:
             db.command(
                 "collMod", collection_name, index=index_spec, dryRun=dry_run
             )
-            if success_result:
-                return {**success_result, "status": "success"}
         except Exception as e:
             # Raise exception for unexpected errors
-            if not error_result:
+            if not error_msg:
                 raise e
-            return {**error_result, "status": "error", "error": str(e)}
+            return {"message": error_msg, "status": "error", "error": str(e)}
 
     def resolve_output(self, ctx):
         outputs = types.Object()
--- a/plugins/operators/__init__.py
+++ b/plugins/operators/__init__.py
@@ -1416,6 +1416,176 @@ class CreateIndex(foo.Operator):
         ctx.dataset.create_index(field_name, unique=unique, wait=False)
 
 
+class ConvertIndexToUnique(foo.Operator):
+    @property
+    def config(self):
+        return foo.OperatorConfig(
+            name="convert_index_to_unique",  # Must match what's in fiftyone.yml
+            label="Convert an existing index to unique",
+            description="Convert an existing non-unique index to a unique index in the dataset if possible or identify any unique key violations that would occur.",
+            allow_immediate_execution=True,
+            allow_delegated_execution=True,
+            default_choice_to_delegated=False,
+        )
+
+    def resolve_input(self, ctx):
+        import re
+
+        inputs = types.Object()
+        indexes = ctx.dataset.get_index_information()
+
+        if not indexes:
+            prop = inputs.str(
+                "index_name",
+                label="This dataset has no non-unique indexes to update",
+                view=types.Warning(),
+            )
+            prop.invalid = True
+            return
+
+        indexes = [
+            i
+            for i, v in indexes.items()
+            if not re.match(r"\.?id$", i) and "unique" not in v
+        ]
+
+        index_selector = types.DropdownView()
+        for key in indexes:
+            index_selector.add_choice(key, label=key)
+
+        inputs.str(
+            "index_name",
+            label="Index",
+            description="The existing index to make unique",
+            view=index_selector,
+            required=True,
+        )
+        inputs.bool(
+            "dry_run",
+            required=False,
+            default=False,
+            label="Check for unique key violations only (dry run)",
+        )
+
+        return types.Property(
+            inputs,
+            view=types.View(label="Convert existing index to unique"),
+        )
+
+    def execute(self, ctx):
+        field = ctx.params.get("index_name")
+        dry_run_only = ctx.params.get("dry_run")
+        result = {
+            "field": field,
+            "dry_run": "True" if dry_run_only else "False",
+            "status": None,
+            "error": None,
+        }
+        if not field:
+            result["status"] = "error"
+            result["error"] = "No index name provided."
+            return result
+
+        if field not in ctx.dataset.get_index_information():
+            result["status"] = "error"
+            result["error"] = f"Index '{field}' does not exist in the dataset."
+            return result
+
+        db = ctx.dataset._sample_collection.database
+        if field.startswith(ctx.dataset._FRAMES_PREFIX):
+            collection_name = ctx.dataset._frames_collection_name
+        else:
+            collection_name = ctx.dataset._sample_collection_name
+
+        index_info = ctx.dataset.get_index_information()[field]
+        key_pattern = {
+            field_name: direction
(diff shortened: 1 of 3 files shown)


tests/unittests/operators/builtin_tests.py
```diff
@@ -0,0 +1,124 @@
...
+            None,
+            Exception("violation"),
+            None,
+        ]
+
+        op = ConvertIndexToUnique()
+        result = op.execute(ctx)
+
+        assert result["status"] == "error"
+        assert "violations" in result["message"].lower()
```
exupero, line 99: Is this "violations" text derived from the exception that's thrown (which uses singular "violation")? Or are the two strings unrelated?

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,190 @@ def execute(self, ctx):
...
+        if field.startswith(ctx.dataset._FRAMES_PREFIX):
+            collection_name = ctx.dataset._frames_collection_name
+        else:
+            collection_name = ctx.dataset._sample_collection_name
+
+        index_info = ctx.dataset.get_index_information()[field]
+        key_pattern = {
+            field_name: direction
+            for field_name, direction in index_info["key"]
+        }
```
exupero:
```suggestion
        key_pattern = dict(index_info["key"])
```

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,190 @@ def execute(self, ctx):
...
+
+        # Special sequence for converting to unique index
+        # https://www.mongodb.com/docs/manual/core/index-unique/convert-to-unique/
+        try:
+            db.command(
+                "collMod",
+                collection_name,
+                index={"keyPattern": key_pattern, "prepareUnique": True},
+            )
```
exupero: Should these collMods be done if it's a dry run? Are they temporary, rolled back, or inconsequential in some way?

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,190 @@ def execute(self, ctx):
...
+                "message"
+            ] = "Error while checking for unique key violations."
+            result["error"] = str(e)
+
+        if dry_run_only or result["status"] == "error":
+            db.command(
+                "collMod",
+                collection_name,
+                index={"keyPattern": key_pattern, "prepareUnique": False},
+            )
```
exupero: All the other collMods are wrapped in a try/except Exception; should this one be too?

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,190 @@ def execute(self, ctx):
...
+                },
+            )
+            result["status"] = "success"
+            result[
+                "message"
+            ] = f"Index '{field}' has been successfully made unique."
+        except Exception as e:
+            result["status"] = "error"
+            result["message"] = f"Error while making index '{field}' unique."
+            result["error"] = str(e)
```
exupero: These try/excepts wrapping calls to db.command makes me wonder if we should have some kind of safe_command wrapper function, e.g.
```python
result = db.safe_command(
    "collMod",
    collection_name,
    index={...},
    success_result={**result, "status": "success", "message": "..."},
    error_result={**result, "status": "error", "message": "..."}
)

File: plugins/operators/__init__.py (hunk 3 above)
Comment (kaixi-wang) [reply plugins/operators/__init__.py]: This has to be run to run even the dryRun. If this initial prepare fails, there isn't anything to roll back.

**Comment by swheaton** [comment]
what is the motivation for doing this? I feel we should run by product before adding any builtin operators.

File: plugins/operators/__init__.py (hunk 4 above)
Comment (kaixi-wang) [reply plugins/operators/__init__.py]: The first one is a guard against a mongo version that may not support this (I couldn't find in the docs what the fcv was for this). The others are just to capture the expected error and return it in a normal output. 
Any failure with this one should probably be surfaced as a real error

**Review by swheaton — CHANGES_REQUESTED** [review]
have not reviewed content but I do think product must weigh in on a builtin operator. Otherwise you can always make your own plugin

File: plugins/operators/__init__.py [hunk 6]
```diff
@@ -1416,6 +1416,202 @@ def execute(self, ctx):
...
+            name="convert_index_to_unique",  # Must match what's in fiftyone.yml
+            label="Convert an existing index to unique",
+            description="Convert an existing non-unique index to a unique index in the dataset if possible or identify any unique key violations that would occur.",
+            allow_immediate_execution=True,
+            allow_delegated_execution=True,
+            default_choice_to_delegated=False,
+        )
+
+    def resolve_input(self, ctx):
+        import re
```
brimoor, line 1432: nit: prefer builtin module imports to be added at the top-level. Local imports should be reserved for special cases where a top-level import isn't possible

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,202 @@ def execute(self, ctx):
...
+
+        inputs = types.Object()
+        indexes = ctx.dataset.get_index_information()
+
+        if not indexes:
```
brimoor, line 1437: Shouldn't this block be placed *after* you filter out existing unique indexes? Datasets will always have default indexes so this would never get triggered as-is.

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,202 @@ def execute(self, ctx):
         ctx.dataset.create_index(field_name, unique=unique, wait=False)
 
 
+class ConvertIndexToUnique(foo.Operator):
+    @property
+    def config(self):
+        return foo.OperatorConfig(
+            name="convert_index_to_unique",  # Must match what's in fiftyone.yml
+            label="Convert an existing index to unique",
```
brimoor, line 1424: We have an existing convention of the `label` exactly matching the operator `name` when converted to sentence case. So here we'd have:

```py
label="Convert index to unique"
```

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,202 @@ def execute(self, ctx):
...
+                view=types.Warning(),
+            )
+            return types.Property(
+                inputs,
+                invalid=True,
+            )
+
+        indexes = [
+            i
+            for i, v in indexes.items()
```
brimoor, line 1450: We should have a discussion about whether we should or should not allow **default indexes** to be modified. 

For reference, here are the default indexes for a dataset that contains frame-level fields:

```py
import fiftyone as fo
import fiftyone.zoo as foz

fo.pprint(dataset.get_index_information())
```

```
{
    'id': {'v': 2, 'key': [('_id', 1)]},  # this is `unique` but mongo annoyingly doesn't report it here
    'filepath': {'v': 2, 'key': [('filepath', 1)]},
    'created_at': {'v': 2, 'key': [('created_at', 1)]},
    'last_modified_at': {'v': 2, 'key': [('last_modified_at', 1)]},
    'frames.id': {'v': 2, 'key': [('_id', 1)]},  # this is `unique` but mongo annoyingly doesn't report it here
    'frames._sample_id_1_frame_number_1': {
        'v': 2,
        'key': [('_sample_id', 1), ('frame_number', 1)],
        'unique': True,
    },
    'frames.created_at': {'v': 2, 'key': [('created_at', 1)]},
    'frames.last_modified_at': {'v': 2, 'key': [('last_modified_at', 1)]},
}
```

It would definitely be bad to allow `created_at`/`last_modified_at` indexes to be made unique, as there's no reason to expect their values to be unique. So that rules out Choice 3 below.

Other data points
- `drop_index()` [prevents dropping default indexes](https://github.com/voxel51/fiftyone/blob/21c77ec954b144e24b739e8a3d43c0bcebaf292f/fiftyone/core/collections.py#L10586-L10595)
- In order to perform an operation like `merge_samples(..., key_field="filepath")`, the `filepath` index is [temporarily upgraded](https://github.com/voxel51/fiftyone/blob/21c77ec954b144e24b739e8a3d43c0bcebaf292f/fiftyone/core/dataset.py#L10246-L10252) to being unique. Thus if the `filepath` index were *already* unique, then merge operations with `key_field="filepath"` would be optimized, in exchange for not being able to add samples with duplicate filepaths to the dataset
- Upgrading a non-unique index to unique would generally not break any existing functionality that required a covered index, with the exception of preventing new samples with duplicate values from being added
- For context, the `filepath` is **not** unique by default because patch views, for example, do contain samples with duplicate filepaths (whenever an image contains multiple object patches)

So, our choices:
1. Don't allow any default indexes to be modified
2. Don't allow default indexes to be modified, with a specific exception for `filepath` which may be converted to unique if the user desires
3. Allow all default indexes to be upgraded to unique

I think I am okay with Option 2, as I believe there are plenty of use cases where users would be okay with *not* being allowed to have duplicate filepaths (in fact this protection could be a feature)

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,202 @@ def execute(self, ctx):
...
+            label="Check for unique key violations only (dry run)",
+        )
+
+        return types.Property(
+            inputs,
+            view=types.View(label="Convert existing index to unique"),
+        )
+
+    def execute(self, ctx):
+        field = ctx.params.get("index_name")
```
brimoor, line 1478: `field` is a potentially confusing variable name as index names are only equal to field names in certain cases. Compound indexes and frame indexes are cases where `index_name != field_name`. I'd recommend calling this variable `name` or `index_name` for clarity

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,202 @@ def execute(self, ctx):
...
+            result["status"] = "error"
+            result["error"] = f"Index '{field}' does not exist in the dataset."
+            return result
+
+        index_info = all_index_info[field]
+
+        db = ctx.dataset._sample_collection.database
+        collection_name = (
+            ctx.dataset._frames_collection_name
+            if field.startswith(ctx.dataset._FRAMES_PREFIX)
```
brimoor, line 1503: It's possible for an image dataset to have a `frames.foo` field, so you need slightly different logic here to *only* switch to frame collection processing if in fact the index refers to a frame-level field.

Here's how that is done in `drop_index()`:
https://github.com/voxel51/fiftyone/blob/21c77ec954b144e24b739e8a3d43c0bcebaf292f/fiftyone/core/collections.py#L10586

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,202 @@ def execute(self, ctx):
...
+        inputs.bool(
+            "dry_run",
+            required=False,
+            default=False,
+            label="Check for unique key violations only (dry run)",
+        )
+
+        return types.Property(
+            inputs,
+            view=types.View(label="Convert existing index to unique"),
```
brimoor, line 1474: nit: by convention, this `label` should match the operator's name:

```py
label="Convert index to unique"
```

plugins/operators/__init__.py
```diff
@@ -1416,6 +1416,202 @@ def execute(self, ctx):
...
+            label="Index",
+            description="The existing index to make unique",
+            view=index_selector,
+            required=True,
+        )
+        inputs.bool(
+            "dry_run",
+            required=False,
+            default=False,
+            label="Check for unique key violations only (dry run)",
```
brimoor, line 1469: I would recommend making the description a bit more verbose so that users fully understand what happens when this is checked vs not checked.

For example (note that `label` may contain markdown):

```py
label="**(Dry run)**: check whether the field contains unique values without modifying the index
```

plugins/operators/__init__.py
brimoor: One consequence of Option 2 is that if a user converts `filepath` to unique, then I'm pretty sure there is no combination of public methods that will allow them to demote it back to a non-unique index, as `drop_index()` won't allow default indexes to be dropped (as it should).

I'm tempted to request that we add a `convert_index_to_non_unique` operator so that users can *undo* this operation from the App if needed. Unfortunately I believe that operator would have to do this though:

```py
dataset._sample_collection.drop_index(...)
dataset._sample_collection.create_index(..., unique=True)
```

brimoor (commented):
Just flagging that there are at least two places in the codebase that would benefit from this faster method of upgrading indexes to unique:
- https://github.com/voxel51/fiftyone/blob/21c77ec954b144e24b739e8a3d43c0bcebaf292f/fiftyone/core/collections.py#L10518-L10519
- https://github.com/voxel51/fiftyone/blob/b02f98c310c04ec23d52672c4f02faeb58de2e6e/fiftyone/core/dataset.py#L9675-L9680

Which may suggest that we should put this logic in a `SampleCollection.convert_index_to_unique()` method so these methods can utilize it.

kaixi-wang:
> @kaixi-wang do you have benchmarks on how much faster `prepareUnique: true` is than dropping + recreating the index?

For the places dataset, it took 20s using this way
Pasted Graphic 6

Not sure how long dropping and recreating the index takes because just creating an index to drop took several hours.

kaixi-wang:
> Just flagging that there are at least two places in the codebase that would benefit from this faster method of upgrading indexes to unique:
> 
> 1. in `create_index()`:

> 2. in `merge_samples()`:

> Moreover, the primary use case that I see for upgrading an index to unique is:
> 
> ```python
> dataset.merge_samples(..., key_field="filepath")
> ```
> so from a user experience standpoint, inserting this code into 2 would be the best way to make sure users benefit from it.


 Yes merging on filepath was what first motivated me to find a solution. I did not realize we were automatically dropping and recreating indexes (or I forgot), but this could definitely explain some of the inconsistent issues I've seen reported with merge_samples... 

> Along those lines, we should perhaps consider if we want to **stop** downgrading the `filepath` index to non-unique after `merge_samples()` completes. Options for that:
> 
> 1. **Stop** downgrading existing indexes to non-unique in `merge_samples()`, instead assuming that if the user did something that requires that field to contain unique values, then they are likely to want it to **remain** containing unique values
> 2. Don't change the default behavior of `merge_samples()`, but **do** add a new opt-in syntax to allow it, something like say:
> 
> ```python
> dataset.merge_samples(..., key_field="filepath", allow_index_updates=True)
> ```
> 
Ideally we wouldn't be dropping/creating/updating any indexes without requiring opt-in from users as these can have significant hidden performance implications. However, I lean towards option 1 in that if we make it unique, just keep it unique because I can't imagine why it would be desirable/expected that a merge key only be unique some of the time.

Agree that the converse of adding an option to make non-unique is a good idea regardless though.

kaixi-wang:
closing in favor of https://github.com/voxel51/fiftyone/pull/6365/

PR #6198, opened 2025-07-28
FOEPD-1848, FOEPD-1851, FOEPD-1854: Log errors to stderr, optionally log JSON

exupero:

- Log errors and higher to stderr
- Log everything below errors to stdout
- Add support for env var `FIFTYONE_LOGGING_FORMAT` set to `json` to log JSON

Tested manually by adding this logging code to `fiftyone/__init__.py`:

```python
logger.info('THIS IS INFO')
logger.debug('THIS IS DEBUG')
logger.warning('THIS IS WARNING')
logger.error('THIS IS ERROR')
try:
    raise ValueError("This is an error")
except ValueError as e:
    logger.error("An exception occurred", exc_info=True)
```

Let me know if you have recommendations for adding unit tests; I didn't find any right off.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable logging output format, allowing users to choose between text and JSON formats for log messages.
  * Introduced options to direct logging output to standard output, standard error, or both.
* **Refactor**
  * Improved logging system by introducing separate handlers for standard output and error streams, and enhanced formatting options for better log readability and structure.
* **Documentation**
  * Updated user guide to include new logging configuration options for output destination and format.


fiftyone/core/decorators.py
```diff
@@ -0,0 +1,32 @@
...
+        nonlocal result
+
+        if has_run:
+            return result
+
+        has_run = True
+        result = f(*args, **kwargs)
+        return result
+
+    return wrapper
```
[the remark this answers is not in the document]
exupero: The original code had the same behavior, this is just a decorator that abstracts the state into non-global vars, and it's okay if it runs multiple times. Feel free to suggest a more accurate name.

fiftyone/core/decorators.py
```diff
@@ -0,0 +1,32 @@
...
+
+    The function is run the first time it is called, and subsequent calls
+    return the result of the first call.
+    """
+    has_run = False
+    result = None
+
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        nonlocal has_run
```
kaixi-wang: interesting. haven't come across nonlocal before

fiftyone/core/logging.py
```diff
@@ -5,32 +5,68 @@
...
+
+    converter = time.gmtime
+
+    def format(self, record):
+        log_entry = {
+            "timestamp": f'{self.formatTime(record, "%Y-%m-%dT%H:%M:%S")}{record.msecs/1000:.3f}Z',
+            "level": record.levelname,
+            "message": record.getMessage(),
+            "filename": record.filename,
+            "lineno": record.lineno,
```
kaixi-wang, line 31:
```suggestion
            "lineno": record.lineno,
            "severity": record.levelname,
```

kaixi-wang: can we specifically add severity? It's what google logs seems to want

exupero: Changed `level` to `severity`. Let me know if we still need `level`.

swheaton (changes requested):
If you add env vars to open source, can you add to the configuration [documentation](https://docs.voxel51.com/user_guide/config.html)?
(This may invalidate the need to add these vars to the deployment documentation as discussed elsewhere, or we could do it again to be redundant).
The file to edit for the above linked documentation site would be: https://github.com/voxel51/fiftyone/blob/develop/docs/source/user_guide/config.rst

Thanks for making stderr optional and also alphabetizing though!

fiftyone/core/logging.py
```diff
@@ -5,32 +5,74 @@
...
+            _logger.addHandler(stdout_handler)
+            _logger.addHandler(stderr_handler)
 
 
 def init_logging():
     """Initializes FiftyOne's package-wide logging.
 
     The logging level is set to ``fo.config.logging_level``.
     """
-    global handler
```
kaixi-wang, line 23: no sure if there was a reason for the global handler before

swheaton (approved):
🚢 🇮🇹

fiftyone/core/logging.py
exupero: I didn't find any usages of `handler` outside this module

PR #6200, opened 2025-07-28
fix: prevent scroll from overriding sidebar highlight after click

AdonaiVera:

@prernadh  found the issue, and I tested it to confirm the problem. In the documentation site, when a user clicks a sidebar link, it briefly highlights the correct section, but a scroll event immediately fires and overrides it with the wrong one.
This PR introduces a hotfix by adding a flag to suppress the next scroll event after a click, ensuring the correct section stays highlighted.

https://github.com/user-attachments/assets/35022e96-1b59-4e98-8872-50fc3a94e365

I'm not sure if this is already fixed in the new documentation version, but this makes it work correctly in the current one.

Manually tested by clicking sidebar links and confirming that the correct section remains highlighted. Verified that the scroll suppression only applies once and resets correctly.

* [x] No. You can skip the rest of this section.

* [ ] Build: Build and test infrastructure changes
* [x] Documentation: FiftyOne documentation changes
* [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sidebar behavior to prevent unwanted updates when clicking sidebar links, resulting in a smoother navigation experience.

### DIFF
--- a/docs/source/_static/js/custom.js
+++ b/docs/source/_static/js/custom.js
@@ -9,6 +9,7 @@ $(function () {
     10;
 
   let lastSection = undefined;
+  let suppressNextScroll = false;
 
   function updateSidebar(e) {
     let currentSection = undefined;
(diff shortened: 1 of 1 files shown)


brimoor:
@AdonaiVera nice fix! 💪

If you retarget to `release/v1.7.2` we can deploy this as part of the upcoming release!

AdonaiVera:
Amazing @brimoor I just did :muscle:

brimoor (approved):
LGTM based on the video you shared 👍

PR #6202, opened 2025-07-30
fix ply points rendering bug

sashankaryal:

This PR fixes PLY point cloud rendering bug and adds an e2e spec that asserts correct rendering behavior of both PLY meshes and point clouds. Additionally, this PR also introduces tooltip behavior similar to PCD files.

Locally. E2E spec.

        notes for FiftyOne users.

-   [ ] Other


e2e-pw/src/oss/poms/modal/looker-3d-controls.ts
```diff
@@ -21,4 +23,33 @@ export class Looker3DControlsPom {
...
+        );
+        return logs?.textContent === SUCCESS_MSG_INJECTED;
+      },
+      SUCCESS_MSG,
+      { timeout: 10000 }
+    );
+    // takes a bit of time for 3d assets to mount after load
+    // todo: figure out if we can emit event on canvas paint
+    // eslint-disable-next-line playwright/no-wait-for-timeout
+    await this.page.waitForTimeout(150);
```
jleven, line 41: How confident are you that this won't result in a flaky test?

jleven:
@sashankaryal -- where in this code change is the bug fix? Calling that out, when so much of the code is the new hover functionality and the test, would be very helpful for the reviewer.

e2e-pw/src/oss/poms/modal/looker-3d-controls.ts
sashankaryal: I wrote that line with an aching heart :( 
Our loading manager emits "load" events but not "paint" events... The GPU is a black box and AFAIK can't set up event listeners for its operations like draw calls.

----
Only way to know is see if it causes any flakiness. Seems good so far.

app/packages/looker-3d/src/fo3d/mesh/Ply.tsx
```diff
@@ -49,7 +57,7 @@ const PlyWithPointsMaterial = ({
 
   const pointsContainerRef = useRef();
 
-  const pointsMaterialElement = usePcdMaterial(
+  const { pointsMaterial, shadingMode } = usePcdMaterial(
     name,
```
sashankaryal, line 61: This was the source of the regression. `usePcdMaterial` in the past returned a JSX element that represented points material. With dynamic PCD work, it started returning an object, hence the `Objects are not valid as a React child` error.

The actual JSX for material is now in the dereferenced `pointsMaterial`.

jleven: Ahh, so this is another example of where Typescript should have caught something but didn't because we're not seeing type errors?

e2e-pw/src/oss/poms/modal/looker-3d-controls.ts
jleven: Yup yup! Let's keep an eye on it, then.

jleven (approved):
LGTM

app/packages/looker-3d/src/fo3d/mesh/Ply.tsx
sashankaryal: Yes

PR #6204, opened 2025-07-30
FOEPD-1814: Annotation schema prototype

benjaminpkane:

Adds a proposed `SampleCollection.compute_annotation_schema` method

This is a starting point to drill down on initial boundaries for human annotation.

Questions to consider
* Should we persist annotation schemas? Where should they live? e.g. on a field document?
* Which annotation types should inspect the sample collection for possible values?
* The dataset schema (which defines field types) already exists. An annotation schema should supplement (and not contradict) this information

This work takes inspiration from the already documented [label schemas](https://docs.voxel51.com/user_guide/annotation.html#label-schema) used for third-party integrations

Python tests

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added the ability to compute and retrieve a UI-facing annotation schema for a specific dataset field via the public API and collection-level method. Supports categorical, boolean, date/number, and nested annotation types, returning type, attributes, and classes for UI consumption.

* **Tests**
  * Added unit tests validating schema generation across multiple primitive and annotation field types.

### DIFF
--- a/fiftyone/core/annotation.py
+++ b/fiftyone/core/annotation.py
@@ -99,7 +99,8 @@ def compute_annotation_schema(collection, field_name):
         - tags
 
     Args:
-        field_name: the field to process
+        collection: a :class:`SampleCollection`
+        field_name: the field name to process
 
     Raises:
         ValueError: if the field does not exists or annotation for its
--- a/fiftyone/core/annotation.py
+++ b/fiftyone/core/annotation.py
@@ -99,8 +99,8 @@ def compute_annotation_schema(collection, field_name):
         - tags
 
     Args:
-        collection: a :class:`SampleCollection`
-        field_name: the field name to process
+        collection: a :class:`fiftyone.core.collections.SampleCollection`
+        field_name: a field name or ``embedded.field.name`` to process
 
     Raises:
         ValueError: if the field does not exists or annotation for its
@@ -108,7 +108,6 @@ def compute_annotation_schema(collection, field_name):
 
     Returns:
         an annotation schema dictionary
-
     """
     if collection.media_type != fom.IMAGE:
         raise ValueError("only image datasets are supported")
--- a/fiftyone/core/annotation.py
+++ b/fiftyone/core/annotation.py
@@ -8,7 +8,6 @@ Annotation runs framework.
 
 import fiftyone.core.fields as fof
 import fiftyone.core.labels as fol
-import fiftyone.core.media as fom
 from fiftyone.core.runs import (
     BaseRun,
     BaseRunConfig,
@@ -109,9 +108,6 @@ def compute_annotation_schema(collection, field_name):
     Returns:
         an annotation schema dictionary
     """
-    if collection.media_type != fom.IMAGE:
-        raise ValueError("only image datasets are supported")
-
     if field_name is None:
         raise ValueError("field_name is required")
 
@@ -134,7 +130,9 @@ def compute_annotation_schema(collection, field_name):
             return {"default": None, "type": "input"}
 
     if is_list:
-        raise ValueError(f"unsupported annotation field {field}")
+        raise ValueError(
+            f"unsupported annotation field {field}; only StringField lists are supported"
+        )
 
     if isinstance(field, fof.BooleanField):
         return {
@@ -157,16 +155,6 @@ def compute_annotation_schema(collection, field_name):
 
     if not isinstance(field, fof.EmbeddedDocumentField):
         raise ValueError(f"unsupported annotation field {field}")
-
-    if not field.document_type in (
-        fol.Classification,
-        fol.Classifications,
-        fol.Detection,
-        fol.Detections,
-    ):
-        raise ValueError(
-            f"unsupported annotation document type {field.document_type}"
-        )
 
     _type = str(field.document_type.__name__).lower()
     if issubclass(field.document_type, fol._HasLabelList):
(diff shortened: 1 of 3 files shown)


fiftyone/core/annotation.py
```diff
@@ -78,3 +82,119 @@ class AnnotationResults(BaseRunResults):
...
+
+    Raises:
+        ValueError: if the field does not exists or annotation for its
+        field type is not supported
+
+    Returns:
+        an annotation schema dictionary
+
+    """
+    if collection.media_type != fom.IMAGE:
```
brimoor: why only image datasets?

If this is because the UX will initially only support classification/detections on image datasets, then same comment from below applies: perhaps we can go ahead and support all types in the SDK and then enforce any supported type limitations in the HA UX?

fiftyone/core/annotation.py
```diff
@@ -78,3 +82,119 @@ class AnnotationResults(BaseRunResults):
...
+    Annotation types are:
+        - checkbox
+        - input
+        - select
+        - radio
+        - text
+        - tags
+
+    Args:
+        collection: a :class:`SampleCollection`
```
brimoor: I believe you must use `fiftyone.core.collections.SampleCollection` for this to work in rendered docs. Relative names only work in class methods when referring to the class itself

fiftyone/core/annotation.py
```diff
@@ -78,3 +82,119 @@ class AnnotationResults(BaseRunResults):
...
+                "default": None,
+                "type": "checkbox" if is_list else "select",
+                "values": collection.distinct(field_name),
+            }
+        except:
+            # too many distinct values
+            return {"default": None, "type": "input"}
+
+    if is_list:
+        raise ValueError(f"unsupported annotation field {field}")
```
brimoor: nit: would be nice to explain in the error that only string lists are supported

fiftyone/core/annotation.py
```diff
@@ -78,3 +82,119 @@ class AnnotationResults(BaseRunResults):
...
+            fof.FloatField,
+            fof.IntField,
+        ),
+    ):
+        return {"type": "input", "default": None}
+
+    if not isinstance(field, fof.EmbeddedDocumentField):
+        raise ValueError(f"unsupported annotation field {field}")
+
+    if not field.document_type in (
```
brimoor: Do we need to raise an error in this method? Or can we go ahead and support all label types in this method and only ignore unsupported fields in the HA UX?

fiftyone/core/collections.py
```diff
@@ -3339,6 +3339,21 @@ def compute_metadata(
             progress=progress,
         )
 
+    def compute_annotation_schema(self, field_name):
+        """Compute the annotation schema for a field
+
+        Args:
+            field_name: the field to process
```
brimoor: nit: for consistency with other methods, I think we should explicitly document when a field name can be embedded:

```py
"""
            field_name: a field name or ``embedded.field.name`` to process
"""
```

fiftyone/core/collections.py
```diff
@@ -3339,6 +3339,21 @@ def compute_metadata(
             progress=progress,
         )
 
+    def compute_annotation_schema(self, field_name):
+        """Compute the annotation schema for a field
```
brimoor: For consistency with existing `SampleCollection` methods and to maximize usability for SDK users, I'd like us to include a full copy of the docstring from `foa.compute_annotation_schema()` here.

fiftyone/core/annotation.py
```diff
@@ -78,3 +82,119 @@ class AnnotationResults(BaseRunResults):
     """Base class for annotation run results."""
 
     pass
+
+
+def compute_annotation_schema(collection, field_name):
```
brimoor, line 86: What do you think about calling this "label schema" instead of "annotation schema" for consistency with the existing nomenclature?
https://docs.voxel51.com/user_guide/annotation.html#label-schema

A link to these docs would be nice in this docstring as well 😄

brimoor: How about allowing the user to provide a list of field names as well, in which case the returned schema is a dict in exactly the format described in the link above?

fiftyone/core/annotation.py
```diff
@@ -5,6 +5,10 @@
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+
+import fiftyone.core.fields as fof
```
brimoor, line 9: The existing utils like the `annotate()` method that use `label_schema`s is in the  `fiftyone.utils.annotations` module. Perhaps it would be good to move this method over there too for consistency, since the concepts are related?

I fully admit that the distinction between a `core` method and a `utils` method is a bit vague/under-specified.

Historically, I've used `core` for core data model and types that permeate all of FO, while `utils` was used for use-case dependent things that only apply to specific optional workflows or specific optional data types/external integrations that aren't required in order to use core FO.

The `annotate()` method initially wound up in `fiftyone.utils.annotations` because annotation was a purely external integration.

For context, stuff related to model evaluation is in `fiftyone.utils.eval`.

The only thing that *might* warrant putting label schema stuff in `core` is that this is *starting* to border on a core data model concept...

fiftyone/core/annotation.py
benjaminpkane: Thanks. Will adjust

fiftyone/core/annotation.py
benjaminpkane: Ok, will adjust

PR #6216, opened 2025-08-01
Upgrade pillow to 11.3.0

kaixi-wang:

pillow upgraded to [11.3.0](https://pillow.readthedocs.io/en/stable/releasenotes/11.3.0.html#) from version [6.2.0](https://pillow.readthedocs.io/en/stable/releasenotes/6.2.0.html) due to the vulnerability [CVE-2025-48379](https://www.cve.org/CVERecord?id=CVE-2025-48379)

(Details)

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ INSTALL_REQUIRES = [
     "numpy",
     "packaging",
     "pandas",
-    "Pillow>=6.2",
+    "Pillow>=11.3.0",
     "plotly>=4.14",
     "pprintpp",
     "psutil",


swheaton (changes requested):
Don't think this is a good idea to do so far upstream, it may break sdk users.
We should pin it in the services affected.

kaixi-wang:
> Don't think this is a good idea to do so far upstream, it may break sdk users. We should pin it in the services affected.

actually looking at the releasenotes, the issue was only introduced in 11.2.0 so can we just disallow that version? (https://pillow.readthedocs.io/en/stable/releasenotes/11.3.0.html#security)

afoley587 (approved):

PR #6219, opened 2025-08-04
Fix instance segmentation model tags in zoo manifest

CharlesCNorton:
Update 18 YOLO segmentation models entries in the model zoo (manifest-torch.json) to use "instances" tag instead of  "segmentation" tag. These models output Detection labels with masks  (instance segmentation) rather than Segmentation labels (semantic  segmentation), so they should be tagged as "instances" for consistency.

Affected models:
- All yolov8*-seg models
- All yolov9*-seg models  
- All yolo11*-seg models
- All yoloe* segmentation models

This PR updates the tags for 18 YOLO segmentation models in `manifest-torch.json` from `"segmentation"` to `"instances"`. These models perform instance segmentation (detecting individual objects with masks) rather than semantic segmentation (pixel-wise classification), so their tags should reflect this distinction. The change ensures consistency across the model zoo where instance segmentation models use the "instances" tag.

This is a metadata-only change to the model manifest JSON file. The change can be and was verified by:
- Checking that the JSON remains valid after the tag updates
- Confirming that model filtering/searching by "instances" tag now includes these YOLO segmentation models
- Verifying that the models still load correctly through the model zoo
- Pulling the branch, building, and loading the affected models.

        notes for FiftyOne users.

Fixed model zoo tagging: YOLO segmentation models are now correctly tagged as "instances" instead of "segmentation" to reflect that they perform instance segmentation.

-   [ ] Other

### DIFF
--- a/fiftyone/zoo/models/manifest-torch.json
+++ b/fiftyone/zoo/models/manifest-torch.json
@@ -3803,7 +3803,7 @@
                     "support": true
                 }
             },
-            "tags": ["segmentation", "coco", "torch", "yolo", "official"],
+            "tags": ["instances", "coco", "torch", "yolo", "official"],
             "date_added": "2024-04-02 19:22:51"
         },
         {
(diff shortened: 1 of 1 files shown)


manushreegangwar (approved):
Thanks for fixing this, @CharlesCNorton. LGTM!

PR #6231, opened 2025-08-06
Fix visualization for semantic segmentation and add confidence thresholding

manushreegangwar:

- Enforce class indices to start at 1 for models where class index 0 doesn't correspond to background class. This is to fix semantic segmentation mask visualization (app doesn't visualize mask pixels set to 0 class index).
- Enable confidence thresholding for semantic segmentation masks so that any mask pixel with confidence below threshold is assigned to class index 0.

```
# Semantic segmentation models in the zoo
'deeplabv3-resnet101-coco-torch',
 'deeplabv3-resnet50-coco-torch',
 'fcn-resnet101-coco-torch',
 'fcn-resnet50-coco-torch',
 'group-vit-segmentation-transformer-torch',
 'segformer-b0-ade20k-torch',
 'segformer-b1-ade20k-torch',
 'segformer-b2-ade20k-torch',
 'segformer-b3-ade20k-torch',
 'segformer-b4-ade20k-torch',
 'segformer-b5-ade20k-torch',
 'segmentation-transformer-torch'
```

1. Run semantic segmentation for all models with multiple confidence thresolds
```
import fiftyone as fo
import fiftyone.zoo as foz

for model_name in seg_models:
    if model_name == 'group-vit-segmentation-transformer-torch':
        model = foz.load_zoo_model("group-vit-segmentation-transformer-torch",
                    text_prompt="a photo of a",
                    classes=["cat", "dog", "bear", "person", "vehicle", "other"])
    else:
        model = foz.load_zoo_model(model_name)

    for conf in [0.0, 0.25, 0.8]:
        label_field = (f"{model_name}_conf{conf}").replace(".","")
        print(label_field)
        dataset.mask_targets[label_field] = model.mask_targets
        dataset.apply_model(model, label_field=label_field, batch_size=8, confidence_thresh=conf)
```

2. Run semantic segmentation on `develop` and compare output against this branch for confidence threshold set to 0.0
```
for model_name in seg_models[:-1]:
    if model_name == 'group-vit-segmentation-transformer-torch':
        model = foz.load_zoo_model("group-vit-segmentation-transformer-torch",
                    text_prompt="a photo of a",
                    classes=["cat", "dog", "bear", "person", "vehicle", "other"])
    else:
        model = foz.load_zoo_model(model_name)

    label_field = (f"{model_name}_dev")
    print(label_field)
    if model.mask_targets:
        dataset.mask_targets[label_field] = model.mask_targets
    dataset.apply_model(model, label_field=label_field, batch_size=8, skip_failures=False)

# For models with no 0 in mask target keys, increment masks by 1 for evaluation (groupvit, segmentation-transformer-torch)

for model_name in seg_models:
    label_field = (f"{model_name}_dev")
    br_label_field = (f"{model_name}_conf00")
    results = dataset.evaluate_segmentations(
        br_label_field,
        gt_field=label_field,
        eval_key=model_name.replace("-","_"),
    )
    print(results.print_metrics())
```

3. Qualitative evaluation in the app to check whether mask visualization looks good.

        notes for FiftyOne users.

All semantic segmentation torch models support confidence thresholding and the class indices start at 1 for segmentation masks.
 
-   [x] Other

## Summary by CodeRabbit

* **New Features**
  * Option to reserve a dedicated background index in semantic segmentation outputs.
  * Per-pixel confidence thresholding to mark low-confidence pixels as background.
  * Accepts either raw logits or softmaxed outputs and supports more flexible input arguments.

* **Improvements**
  * Mask-to-class mappings now derive from model classes and respect background reservation.

### DIFF
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -1655,7 +1655,7 @@ class SemanticSegmenterOutputProcessor(OutputProcessor):
         confidence_thresh = kwargs.pop("confidence_thresh", None)
         if confidence_thresh:
             confidence = probs.max(axis=1)
-            conf_mask = confidence > confidence_thresh
+            conf_mask = confidence >= confidence_thresh
             masks[~conf_mask] = 0
         return [fol.Segmentation(mask=mask) for mask in masks]
 
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -1623,8 +1623,19 @@ class SemanticSegmenterOutputProcessor(OutputProcessor):
         Returns:
             a list of :class:`fiftyone.core.labels.Segmentation` instances
         """
-        probs = output["out"].detach().cpu().numpy()
+        out = output["out"].detach().cpu()
+        if not (torch.all(out >= 0.0) and torch.all(out <= 1.0)):
+            probs = out.softmax(dim=1).numpy()
+        else:
+            probs = out.numpy()
         masks = probs.argmax(axis=1)
+        confidence_thresh = kwargs.pop("confidence_thresh", None)
+        confidence_thresh = confidence_thresh if confidence_thresh else 0
+        conf_mask = masks > confidence_thresh
+        masks[~conf_mask] = -1
+
+        # Increment class index by 1 since 0 is reserved for background in the app.
+        masks += 1
         return [fol.Segmentation(mask=mask) for mask in masks]
 
 
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -1642,10 +1642,10 @@ class SemanticSegmenterOutputProcessor(OutputProcessor):
         Returns:
             a list of :class:`fiftyone.core.labels.Segmentation` instances
         """
-        out = output["out"]
+        out = output["out"].detach().cpu()
         if not self.has_softmax_out:
             out = out.softmax(dim=1)
-        probs = out.detach().cpu().numpy()
+        probs = out.numpy()
 
         masks = probs.argmax(axis=1)
         if self.no_background_cls:
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -963,10 +963,17 @@ class TorchImageModel(
             mask_targets_path = fou.fill_patterns(config.mask_targets_path)
             return etal.load_labels_map(mask_targets_path)
 
-        if config.classes:
-            mask_targets = {
-                idx + 1: val for idx, val in enumerate(config.classes)
-            }
+        if self.classes and config.output_processor_args:
+            if config.output_processor_args.get("no_background_cls", False):
+                mask_targets = {
+                    idx + 1: val for idx, val in enumerate(self.classes)
+                }
+            else:
+                # Class at index 0 is treated as background class.
+                # This class is neither visualized in the app nor used in evaluate_segmentations.
+                mask_targets = {
+                    idx: val for idx, val in enumerate(self.classes)
+                }
             return mask_targets
 
         return None
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -1617,11 +1617,15 @@ class SemanticSegmenterOutputProcessor(OutputProcessor):
         classes (None): the list of class labels for the model. This parameter
             is not used
         no_background_cls (False): if true, class indices are incremented by 1 in the mask
+        has_softmax_out (True): if false, softmax is applied to output predictions.
     """
 
-    def __init__(self, classes=None, no_background_cls=False):
+    def __init__(
+        self, classes=None, no_background_cls=False, has_softmax_out=True
+    ):
         self.classes = classes
         self.no_background_cls = no_background_cls
+        self.has_softmax_out = has_softmax_out
 
     def __call__(self, output, *args, **kwargs):
         """Parses the model output.
@@ -1638,11 +1642,11 @@ class SemanticSegmenterOutputProcessor(OutputProcessor):
         Returns:
             a list of :class:`fiftyone.core.labels.Segmentation` instances
         """
-        out = output["out"].detach().cpu()
-        if not (torch.all(out >= 0.0) and torch.all(out <= 1.0)):
-            probs = out.softmax(dim=1).numpy()
-        else:
-            probs = out.numpy()
+        out = output["out"]
+        if not self.has_softmax_out:
+            out = out.softmax(dim=1)
+        probs = out.detach().cpu().numpy()
+
         masks = probs.argmax(axis=1)
         if self.no_background_cls:
             # Increment class index by 1 since 0 is reserved for background in the app.
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -963,7 +963,11 @@ class TorchImageModel(
             mask_targets_path = fou.fill_patterns(config.mask_targets_path)
             return etal.load_labels_map(mask_targets_path)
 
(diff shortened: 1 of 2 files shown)


fiftyone/utils/torch.py
```diff
@@ -1623,8 +1638,21 @@ def __call__(self, output, *args, **kwargs):
         Returns:
             a list of :class:`fiftyone.core.labels.Segmentation` instances
         """
-        probs = output["out"].detach().cpu().numpy()
+        out = output["out"].detach().cpu()
+        if not (torch.all(out >= 0.0) and torch.all(out <= 1.0)):
```
PeterOttoMoeller: Is this because some models already apply softmax? Would it be less brittle/more straightforward to add some kind of configuration parameter for these two cases (softmax already applied or not) instead of this numeric check?

fiftyone/utils/torch.py
```diff
@@ -1623,8 +1638,21 @@ def __call__(self, output, *args, **kwargs):
         Returns:
             a list of :class:`fiftyone.core.labels.Segmentation` instances
         """
-        probs = output["out"].detach().cpu().numpy()
+        out = output["out"].detach().cpu()
```
PeterOttoMoeller, line 1649: Could we keep all the following tensor operations that were added with this PR on the GPU? I am not sure if that would actually have a runtime advantage, but they all seem to be quite parallelizable

fiftyone/utils/torch.py
```diff
@@ -963,6 +963,19 @@ def _parse_mask_targets(self, config):
             mask_targets_path = fou.fill_patterns(config.mask_targets_path)
             return etal.load_labels_map(mask_targets_path)
 
+        if self.classes and config.output_processor_args:
+            if config.output_processor_args.get("no_background_cls", False):
```
PeterOttoMoeller, line 971: If the model DOES output a background class, we set `mask_targets` to start at 1 for the first class in `classes`. Do we know for certain that `classes` does not include "backgorund" as a class? 

In general, does the order of classes in `classes` always exactly correspond to the model's output? (Assuming this is indeed enforced somewhere else or simply has to be assumed to be true through correct configuration -- just wanting to make sure)

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -2769,7 +2769,11 @@
             "size_bytes": null,
             "default_deployment_config_dict": {
                 "type": "fiftyone.utils.transformers.FiftyOneTransformerForSemanticSegmentation",
-                "config": {}
+                "config": {
+                    "output_processor_args": {
+                        "no_background_cls": true
```
PeterOttoMoeller: Potentially dumb/obvious question: how did you find out that only these 2 specific models `segmentation-transformer-torch` and `group-vit-segmentation-transformer-torch` do not output 0 as background?

PeterOttoMoeller:
I am running this 

`semseg_models = foz.list_zoo_models(tags="segmentation")
# Remove tensorflow models
semseg_models = [model for model in semseg_models if not model.endswith("-tf")]

print(semseg_models)`

on current develop and am getting more than your 6 models in the PR description:

['deeplabv3-resnet101-coco-torch', 'deeplabv3-resnet50-coco-torch', 'fcn-resnet101-coco-torch', 'fcn-resnet50-coco-torch', 'group-vit-segmentation-transformer-torch', 'segformer-b0-ade20k-torch', 'segformer-b1-ade20k-torch', 'segformer-b2-ade20k-torch', 'segformer-b3-ade20k-torch', 'segformer-b4-ade20k-torch', 'segformer-b5-ade20k-torch', 'segmentation-transformer-torch']

fiftyone/utils/torch.py
manushreegangwar: Good question. This change doesn't take into consideration whether there is a "background" class in the model config or not. If the class at index "0" is background (in 'deeplabv3-resnet101-coco-torch' as an example), we will skip incrementing if `no_background_cls = False`. We assume that background class will be class index 0 or assumed to be not present. If there is "background"/ "other" class at any other index, it will be visualized. 

If there's a need to configure classes that are to be treated as class index 0, we can introduce a new parameter to configure such classes. I would do it as a separae PR. 

>  does the order of classes in classes always exactly correspond to the model's output?

https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/torch.py#L952-L966
This is not strictly enforced but is assumed that the model configuration is done correctly. Also, `mask_targets` doesn't affect model inference. It is only used for [visualizing per-pixel labels in the app.](https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/torch.py#L952-L966). Note that line 966 - 977 are not  required for visualizations to work correctly.

fiftyone/utils/torch.py
manushreegangwar: Good point. I will move softmax to gpu.

fiftyone/utils/torch.py
manushreegangwar: Do you mean add it as a model configuration parameter? Yes, that works.

fiftyone/zoo/models/manifest-torch.json
manushreegangwar: If you look at the visualization in the app, you can see that class at index 0 (let's say "cat") is not visualized correctly. More details in jira about this.

fiftyone/utils/torch.py
PeterOttoMoeller: Great, sounds good to me, feel free to resolve conversation

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -89,7 +92,14 @@
...
             },
-            "tags": ["segmentation", "coco", "torch", "resnet", "deeplabv3", "official"],
+            "tags": [
+                "segmentation",
+                "coco",
+                "torch",
+                "resnet",
+                "deeplabv3",
+                "official"
+            ],
```
manushreegangwar, line 102: I can undo the pylint changes if needed but the pre-commit hook is applying these.

fiftyone/utils/torch.py
manushreegangwar: Done

fiftyone/utils/torch.py
manushreegangwar: Added `has_softmax_out` to the output processor. All segmentation models currently in the zoo require this to be true.

fiftyone/utils/torch.py
```diff
@@ -1623,8 +1642,21 @@ def __call__(self, output, *args, **kwargs):
...
             a list of :class:`fiftyone.core.labels.Segmentation` instances
         """
-        probs = output["out"].detach().cpu().numpy()
+        out = output["out"]
+        if not self.has_softmax_out:
+            out = out.softmax(dim=1)
+        probs = out.detach().cpu().numpy()
+
         masks = probs.argmax(axis=1)
+        if self.no_background_cls:
```
PeterOttoMoeller, line 1655: Could all the following tensor operations not also stay on the GPU? Essentially till the return

fiftyone/zoo/models/manifest-torch.json
PeterOttoMoeller: Seems more readable to me this way, would keep

fiftyone/utils/torch.py
manushreegangwar: They may but since these are not very heavy operations it is probably better to move them to CPU

manushreegangwar: I have moved all post-processing to cpu for this PR but let's continue the discussion on GPU post-processing and tackle that in a separate PR.

PR #6233, opened 2025-08-07
use foc.VERSION in logs

kaixi-wang:

- use foc.VERSION because fo.__version__ in teams will sometimes throw an error that fo has no attribute __version__

- run app locally

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Refactor**
  * Updated internal handling of version information in log entries. No visible changes to end-user functionality.

### DIFF
--- a/fiftyone/core/logging.py
+++ b/fiftyone/core/logging.py
@@ -13,6 +13,7 @@ import sys
 import time
 
 import fiftyone as fo
+import fiftyone.constants as foc
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


exupero (commented):
Do we know why the error occurs? Looking at the [current code](https://github.com/voxel51/fiftyone-teams/blob/fa8bd037da501654fa899c35cb896ecbad011e28/fiftyone/__init__.py#L29), `fo.__version__` is just `foc.TEAMS_VERSION`, which is [essentially](https://github.com/voxel51/fiftyone-teams/blob/fa8bd037da501654fa899c35cb896ecbad011e28/fiftyone/constants.py#L54) `importlib.metadata.metadata('fiftyone')['version']`, and if I change `version` to a string that doesn't exist, it just returns None. Where's the error coming from? (Is it only produced by an older version of the code?)

PR #6235, opened 2025-08-07
FOEPD-1787 Add target-view utility for operators

swheaton:

Target-view means giving the operator user the choice of what the operator operates _on_: full dataset, current view, selections, etc.

This is an extremely common pattern in [existing plugins](https://github.com/voxel51/fiftyone-plugins) but is often very repetitious code. This PR adds a standardized target view utility for the operator author to utilize.
- Add `input_utils.resolve_target_view_inputs()` which creates and returns the input selection widget (if multiple choices exist) that the operator can use in its `resolve_inputs()` function to offer the user the target view choice.
- Expand `ExecutionContext.target_view()` to cover all the cases

See prior draft proposal: https://github.com/voxel51/fiftyone-teams/pull/1778

Added unit tests for `ExecutionContext.target_view` and the input utility.
Made compute metadata (upcoming PR) use this utility and then tested that it worked the same as it did before.

        notes for FiftyOne users.

Added a standardized target-view utility to the operator development toolkit.

-   [ ] Other


fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,244 @@
...
+    )
+    selected_samples_description = selected_samples_description or (
+        f"{action_description} only the selected samples"
+    )
+    selected_labels_description = selected_labels_description or (
+        f"{action_description} only the selected labels"
+    )
+
+    # Determine which target views are available
+    has_base_view = isinstance(ctx.view, fop.PatchesView)
```
kaixi-wang: Since this is a general util function, do frames and clips views also need to be handled?

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,244 @@
...
+    selected_labels_label="Selected labels",
+    selected_labels_description=None,
+):
+    """Creates and resolves a target view input for an operator.
+
+    This function adds an enum input to the provided ``inputs`` object that
+    allows the user to select which view to process. The available choices
+    depend on the current context and the provided flags.
+
+    The choices include:
```
kaixi-wang: What about saved view?

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,244 @@
...
+    - Entire dataset (if the operator supports patches views, this is the
+        base view)
```
kaixi-wang: I don't think base view is common knowledge/self explanatory

fiftyone/operators/input_utils.py
swheaton: like on any saved view? if you have it loaded already it would work with CURRENT_VIEW. allowing a selection of saved view would be more advanced than this is intended to be

fiftyone/operators/input_utils.py
swheaton: Honestly I got the base view aspect from BMoore's suggestion on my previous draft PR. I did not know enough about what a user would expect, to make decisions about frames/clips/etc., so I copied the patches only.
https://github.com/voxel51/fiftyone-teams/pull/1778#pullrequestreview-3082219041

fiftyone/operators/input_utils.py
swheaton: Ok will add a bit of explainer

fiftyone/operators/input_utils.py
swheaton: Ok yes I did not look too much into it but you are right, I will add these cases.

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,244 @@
...
+    base_view_description=None,
+    current_view_label="Current view",
+    current_view_description=None,
+    dataset_view_label="Dataset",
+    dataset_view_description=None,
+    selected_samples_label="Selected samples",
+    selected_samples_description=None,
+    selected_labels_label="Selected labels",
+    selected_labels_description=None,
+):
```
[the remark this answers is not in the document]
swheaton: I like that. Working on this change

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,251 @@
...
+            def config(self):
+                return foo.OperatorConfig(
+                    name="target_view_operator",
+                    label="Testing Target View Operator",
+                    dynamic=True,
+                )
+
+            def resolve_inputs(self, ctx):
+                inputs = types.Object()
+                target_view = foo.input_utils.resolve_target_view_inputs(
```
brimoor: Can you remove `input_utils` from the docstring here?

My goal is that everything user-facing should be in the main `fiftyone` namespace, or in a `fiftyone.XXX` namespace if it pertains to a specific feature (like operators here), but never deeper than one-level.

So I'm glad that you added `resolve_target_view_inputs()` to the `fiftyone.operators` namespace, because we indeed intend users to use it 😄

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,251 @@
...
+
+    The choices include:
+    - Entire dataset (if the current view is not a generated view)
+    - Base view (if the current view is a generated view such as
+        :class:`fiftyone.core.clips.ClipsView`, :class:`fiftyone.core.video.FramesView`,
+        or :class:`fiftyone.core.patches.PatchesView`), which is the semantic
+        equivalent of "entire dataset" for these views. The base view is the
+        view from which the generated view was created. For example,
+        ``dataset.limit(51).to_frames("ground_truth").limit(10)`` has a base
+        view of ``dataset.limit(51)``.
```
brimoor: The base view is actually `dataset.limit(51).to_frames("ground_truth")` in this example. It contains all the stuff *after* the generated view stage is applied but before any additional stages.

EG: if an operator is being applied to patches, it doesn't make sense for the user to apply it to something that's *not* a patches view.

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,251 @@
...
+):
+    """Creates and resolves a target view input for an operator.
+
+    This function adds an enum input to the provided ``inputs`` object that
+    allows the user to select which view to process. The available choices
+    depend on the current context and the provided flags.
+
+    The choices include:
+    - Entire dataset (if the current view is not a generated view)
+    - Base view (if the current view is a generated view such as
```
brimoor: Can you build the docs locally and ensure that this renders correctly as a bulleted list? I think you _might_ need to add an empty line above the first bullet and potentially adjust the indentation of multiline bullets.

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,251 @@
...
+        or has_selected_samples
+        or has_selected_labels
+        or allow_dataset_view
+    ):
+        target_choices = types.RadioGroup(orientation="horizontal")
+
+        if has_base_view:
+            # If the view is generated (clips, frames, patches, etc.), then the
+            # equivalent semantics of "entire dataset" is to process all
+            # patches in the base view
```
brimoor: nit: "samples" not "patches"

fiftyone/operators/executor.py
```diff
@@ -615,11 +617,20 @@ def target_view(self, param_name="view_target"):
...
+            return self.view
+        if target == constants.ViewTarget.DATASET:
             return self.dataset
+        if target == constants.ViewTarget.BASE_VIEW:
+            return self.view._base_view
+        if target == constants.ViewTarget.SELECTED_SAMPLES:
+            return self.view.select(self.selected)
+        if target == constants.ViewTarget.SELECTED_LABELS:
+            return self.view.select_labels(self.selected_labels)
+        if target == constants.ViewTarget.DATASET_VIEW:
```
brimoor, line 640: I can't think of a use case for `DATASET_VIEW`, did you have one in mind? I tend to think that `CURRENT_VIEW` and `DATASET` should suffice. 

FWIW, when I read this, I found "current view" vs "dataset view" confusing, so if we don't need it I'd recommend deleting it.

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,251 @@
...
+    via ``ctx.target_view()``.
+
+    The target view descriptions are generated based on the provided
+    ``action_description`` and the various description parameters. If a
+    description parameter is not ``None``, it will be used as the description
+    for the corresponding target view choice. Otherwise, a default description
+    will be generated such as f"{action_description} the entire dataset".
+
+
+    Examples::
```
brimoor: Can you add a short section to the docs about target views, please? 😄 

What you've written here is very nice. Copy+paste should suffice.

I think a good location is after the [Operator inputs](https://github.com/voxel51/fiftyone/blob/7a5c5ab9f4aec7f83b07bf84463c88b7df389112/docs/source/plugins/developing_plugins.rst?plain=1#L1017) section.

fiftyone/operators/input_utils.py
swheaton: yes you're right, typo.

fiftyone/operators/input_utils.py
swheaton: yeah sure. it is accessible via `foo. resolve_target_view_inputs()` so should be fine.

fiftyone/operators/executor.py
swheaton: Someone (I think you?) used it for clone dataset - to implement clone-all vs. clone-no-extras. So I figured I'd add it to try and fit as many of our operators as possible eventually.
It is default disallowed though, since it is rare

fiftyone/operators/input_utils.py
swheaton: Yeah sure I was going to add it along with the scheduler documentation but I can do this section now instead.

fiftyone/operators/input_utils.py
swheaton: yeah will fix

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,252 @@
...
+            ),
+        ]
+        for target_view, is_available, label, description in choice_order:
+            if is_available:
+                target_choices.add_choice(
+                    target_view, label=label, description=description
+                )
+            if default_target is None:
+                default_target = target_view
+
```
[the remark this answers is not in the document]
swheaton: The first choice encountered **is** the documented preference order. Your suggestion looks to have the same logic  in any case, but is significantly worse because it adds another loop

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,252 @@
...
+                constants.ViewTarget.BASE_VIEW,
+                label=base_view_label,
+                description=base_view_description,
+            )
+        else:
+            target_choices.add_choice(
+                constants.ViewTarget.DATASET,
+                label=dataset_label,
+                description=dataset_description,
+            )
```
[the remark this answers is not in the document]
swheaton: you're right, accepting the change

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,252 @@
...
+                selected_labels_description,
+            ),
+        ]
+        for target_view, is_available, label, description in choice_order:
+            if is_available:
+                target_choices.add_choice(
+                    target_view, label=label, description=description
+                )
+            if default_target is None:
+                default_target = target_view
```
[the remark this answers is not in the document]
swheaton: accepted

fiftyone/operators/executor.py
kaixi-wang: I think clone dataset could still work with CURRENT_VIEW and DATASET alone? Like full dataset without extras could still be achieved by selecting CURRENT_VIEW without any filters

swheaton: I think the thought was that if you do have view filters on, then you are still presented with the 2 options for full dataset clone (extras or no)

fiftyone/operators/input_utils.py
swheaton: done

fiftyone/operators/input_utils.py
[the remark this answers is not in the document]
swheaton: I decided not to do this after all. While it does seem cleaner to have not so many function parameters, it is not a common pattern in the codebase and so may be jarring to the user. Additionally the generated documentation looked a lot better / was a lot more useful, with direct parameters rather than fields on a `dataclass`.

fiftyone/operators/constants.py
```diff
@@ -0,0 +1,16 @@
+"""
+FiftyOne operator constants.
+
+| Copyright 2017-2025, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
+
+class ViewTarget:
```
tom-vx51, line 10: can we add documentation for what each of these means?

fiftyone/operators/input_utils.py
```diff
@@ -0,0 +1,263 @@
...
+            target_choices.values(),
+            default=default_target,
+            required=True,
+            label="Target view",
+            view=target_choices,
+        )
+        target = ctx.params.get(param_name, default_target)
+    else:
+        # Only one choice is available (dataset), so we don't need to prompt
+        target = constants.ViewTarget.DATASET
```
tom-vx51: any concern that the user won't realize other options are possible? I would vote for having disabled radio buttons

swheaton: I made it match the existing pattern in all of our V51 plugins. I don't really want to change the UX in this PR.
At least now once all plugins are ported to using this approach, if this change was desired we'd only have to change in one place.

fiftyone/operators/constants.py
swheaton: sure will copy over from other documentation locations

swheaton:
I have just realized there is already a `ViewTargetProperty` in `types.py` 😭 
I will have to consolidate these.

fiftyone/operators/executor.py
```diff
@@ -615,25 +617,67 @@ def target_view(self, param_name="view_target"):
...
+
+    # Alias for common reversal
+    view_target = target_view
+
+    @property
+    def has_generated_view(self):
+        """Whether the context's view is a generated view
+
+        Returns: a tuple of ``(bool, bool)`` indicating whether the context's
+            view is a generated view, and whether it has additional filters
```
tom-vx51: [🤓 nit] should probably be a separate property, returning raw tuples always feels scary

swheaton: haha yeah. I was trying to avoid doing the same computation twice but yeah I'll fix it

fiftyone/operators/types.py
```diff
@@ -2411,99 +2540,355 @@ class ViewTargetOptions(object):
...
 
         _type = Enum(options.values())
 
+        if not default_target or default_target not in options.values():
+            default_target = options.values()[-1]  # last option
+
         # Only 1 option so no need for a radio group, just hide it.
         if len(options.values()) == 1:
-            choice_view = HiddenView(read_only=True)
+            choice_view = HiddenView(read_only=False)
```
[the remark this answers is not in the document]
swheaton: provide a quote and link to the docs you think are contradictory

fiftyone/operators/executor.py
```diff
@@ -615,25 +617,67 @@ def target_view(self, param_name="view_target"):
...
+            )
+
+        if target == constants.ViewTarget.CURRENT_VIEW:
+            return self.view
+        if target == constants.ViewTarget.DATASET:
             return self.dataset
-        return self.view
+        if target == constants.ViewTarget.BASE_VIEW:
+            return self.view._base_view
+        if target == constants.ViewTarget.SELECTED_SAMPLES:
```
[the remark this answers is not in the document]
swheaton: Incorrect, `_base_view` is a property

fiftyone/operators/types.py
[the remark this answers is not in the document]
swheaton: No you were right. I will update the `default_target` parameter documentation to match the code logic. Selecting the most targeted choice as the default is the intended behavior.

fiftyone/operators/executor.py
brimoor: Ah okay. I'm onboard with keeping `DATASET_VIEW` as an option since it is disabled by default ✅

fiftyone/operators/executor.py
```diff
@@ -615,25 +619,94 @@ def target_view(self, param_name="view_target"):
...
+        return self.dataset
+
+    # Alias for common word reversal
+    view_target = target_view
+
+    def _init_has_generated_view(self):
+        stages = self.request_params.get("view", None)
+        generated_view_stages = {
+            f"{stage_class.__module__}.{stage_class.__name__}"
+            for stage_class in (
```
brimoor, line 652: Note to self: this will need to be updated when https://github.com/voxel51/fiftyone/pull/4773 is merged

fiftyone/operators/executor.py
```diff
@@ -615,25 +619,94 @@ def target_view(self, param_name="view_target"):
...
         """
-        target = self.params.get(param_name, None)
-        if target == "SELECTED_SAMPLES":
-            return self.view.select(self.selected)
-        if target == "DATASET":
+        target = self.params.get(param_name)
+        if not target:
+            target = (
+                constants.ViewTarget.BASE_VIEW
+                if self.has_generated_view
```
brimoor, line 626: @swheaton this usage of `self.has_generated_view` occurs in a method that already accesses `self.view`, so we could just call the existing `self.view._is_generated` property here rather than adding a new implementation of it.

(I will expand on this thought below)

fiftyone/operators/types.py
```diff
@@ -2411,99 +2539,355 @@ class ViewTargetOptions(object):
...
+            selected_labels_description (None): the description for the
+                "selected labels" target view. If ``None``, a default
+                description is generated
+        """
 
-        # Entire dataset is always an option
-        default_target = options.entire_dataset.value
-        options.entire_dataset.include = True
+        # Determine which target views are available
+        has_base_view = ctx.has_generated_view
```
brimoor, line 2840: @swheaton finishing my thought from https://github.com/voxel51/fiftyone/pull/6235/files#r2278096426, this looks like the only other usage of `has_generated_view` that is introduced.

I might have missed a more in depth discussion, but I think it would be fine to use `ctx.view._is_generated` here too. Loading a view is supposed to be fast in general. Yes, loading a generated view is the one case where "just" loading the view could trigger the expensive regeneration of a backing dataset, but by definition the only way to get a `ctx.view` that is generated is if you were just doing something with a generated view, so its non-persistent dataset should really already exist and thus `ctx.view` should be a no-op. Also I would say it is highly likely that something else downstream of this will access `ctx.view` anyway.

One motivation for pointing this out is that I'd prefer that whether a view is "generated" is kept as an implementation detail and not exposed as part of the public interface of `ExecutionContext` that users need to worry about.

Also, it would be nice for `SampleCollection._is_generated` to be the single source of truth for this determination. Multiple implementations can get out of sync. For example, as you point out in the docstring, `has_generated_view` doesn't work for saved views, so `ctx.view._is_generated` is more accurate.

fiftyone/operators/executor.py
swheaton: That's true. The main reason I did this was not for here but for where the view target property type is created. Initially I had it using `self.view` but I don't want to load dataset/view inside `resolve_input()`

I could change it for this method.

fiftyone/operators/types.py
swheaton: Agree with all of those points, I would greatly prefer to do it that way.
Except for that I'm worried about loading dataset/view since this function would be called on every `resolve_input()` call. On a large dataset, that could add noticeable lag - and system load.

But now I'm wondering if there is a way to only compute this on the first `resolve_input()` and not every time ... after all, I don't think it would be possible to change the view/selection while the operator modal is up

swheaton: (accessing a typical view should be very fast yes, but loading dataset is less so. And I don't think it would reuse the `fo.Dataset`)

brimoor: > But now I'm wondering if there is a way to only compute this on the first resolve_input() and not every time ... after all, I don't think it would be possible to change the view/selection while the operator modal is up

Sounds like you want `@execution_cache(prompt_scoped=True)` that caches `ctx.dataset` for the life of the input modal.

> Except for that I'm worried about loading dataset/view since this function would be called on every resolve_input() call. On a large dataset, that could add noticeable lag - and system load.

Yeah I get it. I too want `resolve_input()` to be as fast as possible. It's very important for dynamic operators usability (and most operators _are_ dynamic).

FWIW, most dynamic operators that use `ctx.target_view` are very likely also accessing `ctx.dataset` for something in `resolve_input()`. So basically I think the intentions of `ctx.has_generated_view` are sound (<-- make `resolve_input()` fast), but it seems like what we really need is a way to reuse `Dataset` objects for the life of a prompt scope, as if dataset loading is the bottleneck then in practice it will *still* probably be a bottleneck even if the target view utility doesn't call it.

brimoor: cc @ritch

brimoor: @swheaton can we start by making these generated view props private to declare our intention that end users shouldn't need to think about it?

swheaton: > @swheaton can we start by making these generated view props private to declare our intention that end users shouldn't need to think about it?

at minimum, yeah, can do. I will test on a large dataset and see how bad the resolve_input actually is. maybe its tolerable

brimoor: Data will be helpful, thanks!

You could very well be right that `ctx.dataset` is a bottleneck. But if it is, then in practice I think it will *remain* a bottleneck even if the target view utility isn't implicitly invoking it, because something else in `resolve_input()` probably will be.

swheaton: EDIT: going to do more testing on release now that I can upload plugins correctly. Input is getting stuck a lot on compute metadata

K so I tested and it did not seem to have an impact on the form loading time on large datasets.
Premature optimization something something .. 🙃 

I will revert to the other impl except still keep it in `ViewTargetProperty`.
We can always change it if we see problems in the wild.

PR #6248, opened 2025-08-11
Add LLMDet zero-shot OD family

CharlesCNorton:
Add 3 LLMDet zero-shot object detection models from iSEE-Laboratory:
llmdet-tiny-torch  (692MB)
llmdet-base-torch  (936MB)
llmdet-large-torch (1376MB)

All models:

* Use existing `FiftyOneZeroShotTransformerForObjectDetection` wrapper
* Support both CPU and GPU inference
* Open‑vocabulary zero‑shot detection from natural‑language prompts
* Apache 2.0 licensed

Add llmdet zero shot od family. (three models)

Validated Transformers implementation on the `quickstart` dataset: loaded each checkpoint, ran zero‑shot detection with COCO prompts, and confirmed reasonable detections and byte sizes. Spot‑checked zoo load.

* [ ] No. You can skip the rest of this section.
* [x] Yes. Give a description of this change to be included in the release
  notes for FiftyOne users.

Added three LLMDet open‑vocabulary zero‑shot object detectors (tiny/base/large) to the Torch model zoo.

* [ ] Build: Build and test infrastructure changes
* [ ] Documentation: FiftyOne documentation changes
* [x] Other

## Release Notes

* **New Features**
  * Extended object detection capabilities to support grounded detection models. The detector now handles both grounded and non-grounded detection outputs with appropriate processing pipelines.

### DIFF
--- a/fiftyone/utils/transformers.py
+++ b/fiftyone/utils/transformers.py
@@ -1511,6 +1511,11 @@ class TransformersDetectorOutputProcessor(fout.DetectorOutputProcessor):
         return res
 
     def _parse_output(self, output, frame_size, confidence_thresh, classes=None):
+        """Converts raw detector output to :class:`fiftyone.core.labels.Detections`.
+
+        Unlike the base class, this handles grounded detection models that
+        return text labels directly rather than class indices.
+        """
         detections = []
 
         scores = output["scores"].cpu().numpy()


fiftyone/utils/transformers.py
```diff
@@ -1456,6 +1456,7 @@ def processor(self):
     @processor.setter
     def processor(self, processor):
         self._processor = processor
+        self._is_grounded = False
```
lorenzoriano: This being an object variable should be in __init__ with relative doc

fiftyone/utils/transformers.py
```diff
@@ -1497,6 +1505,39 @@ def __call__(
 
         return res
 
+    def _parse_output(self, output, frame_size, confidence_thresh, classes=None):
```
lorenzoriano, line 1513: Can you add a docstring? Also the code is very similar to the original in DetectorOutputProcessor. Any chance it can be refactor to share the common code? For example by introducing support private functions with the shared code

fiftyone/utils/transformers.py
CharlesCNorton: Done in commit e2ab1d99.

fiftyone/utils/transformers.py
CharlesCNorton: docstring added\! re: refactoring, afaik these methods diverge quite a bit (grounded vs non-grounded, text labels vs class indices, etc). sharing code would probably add complexity without real benefit, but happy to reconsider if I'm missing something.

PR #6272, opened 2025-08-19
FOEPD-1948 target view utility followup

swheaton:

Followup from PR comment: https://github.com/voxel51/fiftyone/pull/6235#discussion_r2278116003

Originally I was trying to minimize work in `resolve_input()` since that can get called a lot. Premature optimization it turns out... Tested on 2M sample dataset and the different between using `ctx.view` and not, has no impact.

So instead of reinventing the wheel of determining if the thing has a generated view, current_view, etc., we will use what we have on `ctx.view`.

Same as that PR - compute metadata plugin with all different views and selections etc.

### DIFF
--- a/fiftyone/operators/executor.py
+++ b/fiftyone/operators/executor.py
@@ -620,10 +620,13 @@ class ExecutionContext(object):
             a :class:`fiftyone.core.collections.SampleCollection`
         """
         target = self.params.get(param_name)
+
+        # If no target is specified, default to the base view if the
+        #   current view is generated, otherwise default to the entire dataset
         if not target:
             target = (
                 constants.ViewTarget.BASE_VIEW
-                if self.has_generated_view
+                if self.view._is_generated  # pylint: disable=protected-access
                 else constants.ViewTarget.DATASET
             )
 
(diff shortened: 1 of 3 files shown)


fiftyone/operators/executor.py
```diff
@@ -620,10 +620,13 @@ def target_view(
...
         """
         target = self.params.get(param_name)
+
+        # If no target is specified, default to the base view if the
+        #   current view is generated, otherwise default to the entire dataset
         if not target:
             target = (
                 constants.ViewTarget.BASE_VIEW
-                if self.has_generated_view
+                if self.view._is_generated  # pylint: disable=protected-access
```
tom-vx51: a conversation for a different day, but imo this defeats the purpose of pylint

tom-vx51 (dismissed):

fiftyone/operators/executor.py
swheaton: indeed

swheaton: we have some stuff that is like semi-private. We need to use it in internal code but we don't want to expose to end users. Would need code restructuring to fix

brimoor (commented):
thanks for doing this! 🙌

PR #6277, opened 2025-08-19
Adding utils for 3d to 2d projections

prernadh:

Adding utilities for converting 3d detection labels of the lidar slice in a grouped dataset into 2d cuboid/polyline labels of the camera slice in the grouped dataset. 

Tested manually with nuScenes dataset

```python
import fiftyone as fo
import numpy as np
from pyquaternion import Quaternion
from fiftyone.utils.labels import detections_3d_to_cuboids_2d

dataset = fo.load_dataset("nuScenes")

cs_record_lidar = {'token': 'a183049901c24361a6b0b11b8013137c', 'sensor_token': 'dc8b396651c05aedbb9cdaae573bb567', 'translation': [0.943713, 0.0, 1.84023], 'rotation': [0.7077955119163518, -0.006492242056004365, 0.010646214713995808, -0.7063073142877817], 'camera_intrinsic': []}

pose_record_lidar = {'token': '9d9bf11fb0e144c8b446d54a8a00184f', 'timestamp': 1532402927647951, 'rotation': [0.5720320396729045, -0.0016977771610471074, 0.011798001930183783, -0.8201446642457809], 'translation': [411.3039349319818, 1180.8903791765097, 0.0]}

pose_record_cam = {'token': 'e3d495d4ac534d54b321f50006683844', 'timestamp': 1532402927612460, 'rotation': [0.5720063498929273, -0.0021434844534272707, 0.011564094980151613, -0.8201648693182716], 'translation': [411.4199861830012, 1181.197175631848, 0.0]}

cs_record_cam = {'token': '1d31c729b073425e8e0202c5c6e66ee1', 'sensor_token': '725903f5b62f56118f4094b46a4470d8', 'translation': [1.70079118954, 0.0159456324149, 1.51095763913], 'rotation': [0.4998015430569128, -0.5030316162024876, 0.4997798114386805, -0.49737083824542755], 'camera_intrinsic': [[1266.417203046554, 0.0, 816.2670197447984], [0.0, 1266.417203046554, 491.50706579294757], [0.0, 0.0, 1.0]]}

transforms = [(np.array(cs_record_lidar["translation"]), Quaternion(cs_record_lidar["rotation"]).rotation_matrix),
            (np.array(pose_record_lidar["translation"]), Quaternion(pose_record_lidar["rotation"]).rotation_matrix),
            (np.array(pose_record_cam["translation"]), Quaternion(pose_record_cam["rotation"]).rotation_matrix),
            (np.array(cs_record_cam["translation"]), Quaternion(cs_record_cam["rotation"]).rotation_matrix)]

to_target = [True, True, False, False]

cam_intrinsic = np.array([[1.26641720e+03, 0.00000000e+00, 8.16267020e+02],
       [0.00000000e+00, 1.26641720e+03, 4.91507066e+02],
       [0.00000000e+00, 0.00000000e+00, 1.00000000e+00]])

transformations = {}
camera_intrinsics = {}
forward_transformation_flags = {}
for group in dataset.iter_groups():
    spatial_sample = group["LIDAR_TOP"]
    camera_sample = group["CAM_FRONT"]
    transformations[spatial_sample.filepath] = transforms
    camera_intrinsics[camera_sample.filepath] = {"intrinsics":cam_intrinsic}
    forward_transformation_flags[spatial_sample.filepath] = to_target


detections_3d_to_cuboids_2d(dataset, "LIDAR_TOP", "CAM_FRONT", "ground_truth", 
                           "2d_projections_full_test", transformations, camera_intrinsics, forward_transformation_flags, 
                            transformation_key_field = "filepath", camera_key_field = "filepath", batch_size = 1000)
```

        notes for FiftyOne users.

Adds in 3d -> 2d projection functionality for multi-modal grouped datasets in FiftyOne.

-   [x] Other

For PR reviewers: 
- I tried to optimize the main function `detections_3d_to_cuboids_2d` using `map_samples` but in benchmarking I found that the save context is faster than using map_samples so will leave as is for now. 

TODO before its ready for review: 
- [x] Turn the single transformation pipeline into dicts mapping sample_ids to transformations
- [x] Turn single camera intrinsic into a dict mapping sample_id to camera_intrinsic
- [x] Turn source_to_target into a dict mapping sample_id to source_to_target
- [ ] Add distortion functions for post pinhole camera projection
       -  Will likely punt to a different PR

## Summary by CodeRabbit

- **New Features**
  - Convert 3D detections into 2D cuboid overlays for grouped, multi‑camera datasets.
  - New validation to surface clear errors for unsupported dynamic group views.

- **Documentation**
  - Added a guide on projecting 3D detections to 2D with usage notes and visualization guidance.
  - Note: the new doc section was duplicated in the patch.

- **Utilities**
  - Added 3D transformation, projection, and visibility utilities (two copies present).

- **Tests**
  - Added tests for transforms, projection, visibility, and box geometry (some test blocks duplicated).

### DIFF
--- a/fiftyone/utils/utils3d.py
+++ b/fiftyone/utils/utils3d.py
@@ -12,7 +12,6 @@ import os
 import warnings
 
 import numpy as np
-import scipy
 import scipy.spatial as sp
 from scipy.spatial.transform import Rotation as R
 from typing import Any, Dict, List, Tuple, Union
@@ -428,7 +427,7 @@ def _compute_intersection_points(box1, box2):
 
 def rpy_to_rotation(euler_rpy: List[float]):
     """Converts Euler angles in roll-pitch-yaw order to a
-    scipy.spatial.transforms Rotation.
+    scipy.spatial.transform Rotation.
 
     Args:
         euler_rpy: a list of Euler angles in roll-pitch-yaw order
@@ -506,8 +505,8 @@ def single_coordinate_transform(
     Args:
         points: A 3-element np.ndarray representing the (x, y, z) coordinates of
             the point
-        rot_quaternion: A scipy.spatial.transform.Rotation quaternion
-            representing the current orientation
+        rot: A scipy.spatial.transform.Rotation representing the current
+            orientation
         transformation: A tuple containing:
             - translation: 3-element array (tx, ty, tz)
             - rotation_matrix: (3, 3) rotation matrix


docs/source/user_guide/using_datasets.rst
```diff
@@ -4274,6 +4274,115 @@ new field on your dataset:
         mask_targets={1: "cat", 2: "dog"},
     )
 
+.. _projecting-3d-detections:
```
brimoor, line 4277: nit: I believe you need to install pre-commit hooks? They should automatically enforce our[ 79 character limit](https://github.com/voxel51/fiftyone-teams/blob/develop/STYLE_GUIDE.md#documentation-style-guide) on `rst` files like they do for `py` files.

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1205,208 @@ def _get_filepath(sample_or_frame, sample):
...
+    camera_slice_name: str,
+    lidar_slice_name: str,
+    in_field: str,
+    out_field: str,
+    transformations: Dict[str, fou3d.TransformationType],
+    camera_params: Dict[str, Dict[str, Any]],
+    forward_transform_flags: Dict[str, List[bool]] = None,
+    batch_size: int = 1000,
+    camera_model: Callable = fou3d.pinhole_projector,
+    progress: bool = True,
```
brimoor: nit: by convention, all builtin methods generally default to `progress=None` so that FO can respect the user's default value of `fo.config.show_progress_bars`. See the other methods in this module for examples of this. 

Can you update to `progress=None`?

fiftyone/utils/utils3d.py
```diff
@@ -417,6 +430,195 @@ def _compute_intersection_points(box1, box2):
     return intersection_points
 
 
+def rpy_to_quaternion(euler_rpy: List[float]):
+    """Converts Euler angles in roll-pitch-yaw order to a quaternion.
+
+    Args:
+        euler_rpy: a list of Euler angles in roll-pitch-yaw order
```
brimoor, line 433: nit: I *think* you need an empty line between the `Args:` and `Returns:` blocks in order for the docs to render correctly at fiftyone.ai

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1205,208 @@ def _get_filepath(sample_or_frame, sample):
         return sample_or_frame.filepath
     except:
         return sample.filepath
+
+
+def detections_3d_to_cuboids_2d(
+    dataset: fo.Dataset,
```
brimoor: Can you update this to `sample_collection: fo.SampleCollection`? It looks like this method will support any sample collection, not just entire datasets.

By convention, we make all builtin methods support `SampleCollection` whenever possible, as it is a core principle of FO that you can generally invoke most things on `DatasetView` too.

fiftyone/utils/utils3d.py
```diff
@@ -417,6 +430,195 @@ def _compute_intersection_points(box1, box2):
...
+
+
+def multiple_coordinate_transform(
+    points: List[float],
+    euler_rpy: List[float],
+    transformation_sequence: TransformationType,
+    forward_transform_flags: List[bool] = None,
+) -> Tuple[List[float], List[float]]:
+    """
+    Applies a sequence of 3D coordinate frame transformations to a point and its orientation.
```
brimoor: nit: the 79 character limit from the style guide also applies to docstrings, although unfortunately our pre-commit hooks don't automatically enforce this.

Can you manually enforce the 79 character limit on these docstrings? 🤓

brimoor (commented):
I defer to @danielgural for functional review, just leaving a couple procedural comments 🤓

fiftyone/utils/utils3d.py
AdonaiVera: Yes Brian, we have an exclude rule in the pre-commit hooks to avoid auto-fixing .rst files since that can break tables and content. I’m looking into a docs-specific hook that only fixes length/spacing without affecting tables. I’ll check and update you on what I find.

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,212 @@ def _get_filepath(sample_or_frame, sample):
         return sample_or_frame.filepath
     except:
         return sample.filepath
+
+
+def detections_3d_to_cuboids_2d(
+    dataset: SampleCollection,
```
brimoor: nit: can you rename this parameter to `sample_collection` for clarity? By convention, that's the variable name we use when the type is `SampleCollection`

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,212 @@ def _get_filepath(sample_or_frame, sample):
...
+            contains the keys required by the camera_model for each sample in
+            the dataset.
+        forward_transform_flags: a dict mapping sample IDs to lists of booleans
+            indicating whether to apply forward or inverse transformations for
+            each transform in the list of transformations for each sample
+        batch_size (1000): number of samples to process in each batch
+        camera_model (fou3d.pinhole_projector): a callable that takes 3D points
+            in the camera coordinate system and the camera parameters dict and
+            returns the projected 2D points in pixel coordinates
+        progress (True): whether to show a progress bar during processing
```
brimoor: nit: the docstring for `progress` needs updating to reflect that the default is now `None`.

You can copy + paste the same docstring from other methods in this module:

```
        progress (None): whether to render a progress bar (True/False), use the
```

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,212 @@ def _get_filepath(sample_or_frame, sample):
...
+    """
+    fov.validate_grouped_non_dynamic_collection(dataset)
+    fov.validate_collection_label_fields(dataset, in_field, fol.Detections)
+
+    lidar_slice = dataset.select_group_slices(lidar_slice_name)
+    camera_slice = dataset.select_group_slices(camera_slice_name)
+    camera_slice.compute_metadata()
```
brimoor: nit: it'd be nice to pass `progress` to `compute_metadata()` as well since it also provides control over progress printing:

```py
camera_slice.compute_metadata(progress=progress)
```

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,212 @@ def _get_filepath(sample_or_frame, sample):
...
+
+    values_2d_label = {}
+    count = 0
+    view_3d_label = lidar_slice.select_fields([in_field, "group"])
```
brimoor: `"group"` is a magic string and would need to be replaced with `lidar_slice.group_field` for generality, if you actually need to refer to the name of the collection's group field.

However, note that FO considers the `group_field` of a group collection a **default field**, which means that `select_fields()` **always** includes it like it does other default fields like `id`, `filepath`, etc.

So TLDR in this case you can just do this:

```py
view_3d_label = lidar_slice.select_fields(in_field)
```

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,212 @@ def _get_filepath(sample_or_frame, sample):
...
+
+    with view_3d_label.save_context() as context:
```
brimoor: Can you update this loop to use [`iter_groups()`](https://docs.voxel51.com/api/fiftyone.core.view.html#fiftyone.core.view.DatasetView.iter_groups) instead? It was created exactly for this purpose, and it would be nice for our builtin utils to demonstrate its use as a best practice 🤗 

```py
for group in sample_collection.select_fields(in_field).iter_groups(
    group_slices=[lidar_slice_name, camera_slice_name],
    progress=progress,
    autosave=True,
):
    lidar_sample = group[lidar_slice_name]
    camera_sample = group[camera_slice_name]

    # computation...

    camera_sample[out_field] = ...
```

It is true that using `save_context()` is a small optimization over using `autosave=True`, as the latter calls `sample.save()` on every sample that is loaded, even if some samples weren't actually edited (which is fine to do, it results in a no-op for samples that have not actually been edited). However, `iter_groups()` retrieves all slices in a single aggregation cursor, which is more efficient than this implementation, which requires an `get_group()` call that requires extra DB connections.

fiftyone/utils/utils3d.py
```diff
@@ -417,6 +432,218 @@ def _compute_intersection_points(box1, box2):
...
+    forward_transform_flags: List[bool] = None,
+) -> Tuple[List[float], List[float]]:
+    """
+    Applies a sequence of 3D coordinate frame transformations to a point and its
+    orientation. Each transformation consists of a translation vector and a
+    rotation matrix, applied in the order provided. The orientation is updated
+    at each step using quaternion multiplication.
+
+    Args:
+        points: A 3-element list representing the (x, y, z) coordinates of the
```
brimoor: nit: for multiline parameter descriptions, I believe you must indent subsequent lines by 4 spaces in order for the docstring to render correctly.

See other methods in this module for examples of correct formatting.

fiftyone/utils/utils3d.py
```diff
@@ -428,8 +655,8 @@ class OrthographicProjectionMetadata(DynamicEmbeddedDocument, fol._HasMedia):
             plane
         normal (None): the normal vector of the plane onto which the projection
             was performed. If not specified, ``[0, 0, 1]`` is assumed
-        width: the width of the image, in pixels
-        height: the height of the image, in pixels
+        width: the width of the image, in pixels height: the height of the
```
brimoor: Looks like there are some unwanted edits to `OrthographicProjectionMetadata` that will mess up the docstring. Can you revert them please?

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,212 @@ def _get_filepath(sample_or_frame, sample):
...
+    Args:
+        dataset: a
+            :class:`fiftyone.core.collections.SampleCollection`
+        camera_slice_name: the name of the camera slice in the dataset
+        lidar_slice_name: the name of the LiDAR slice in the dataset
+        in_field: the name of the :class:`fiftyone.core.labels.Detection` field
+            containing 3D labels to convert
+        out_field: the name of the :class:`fiftyone.core.labels.Polylines`
+            field to populate with 2D cuboids
+        transformations: a dict mapping sample IDs to transformation tuples
```
prernadh: Maps to a list of transformation tuples

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,212 @@ def _get_filepath(sample_or_frame, sample):
...
         return sample.filepath
+
+
+def detections_3d_to_cuboids_2d(
+    dataset: SampleCollection,
+    camera_slice_name: str,
+    lidar_slice_name: str,
+    in_field: str,
+    out_field: str,
+    transformations: Dict[str, fou3d.TransformationType],
```
prernadh, line 1215: allow users to choose key_field instead of forcing id

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,212 @@ def _get_filepath(sample_or_frame, sample):
         return sample_or_frame.filepath
     except:
         return sample.filepath
+
+
+def detections_3d_to_cuboids_2d(
+    dataset: SampleCollection,
+    camera_slice_name: str,
+    lidar_slice_name: str,
```
prernadh: nit: Fix naming -> replace `lidar` with something else

fiftyone/utils/labels.py
prernadh: Oh yes good point. Sure

docs/source/user_guide/using_datasets.rst
```diff
@@ -4274,6 +4274,134 @@ new field on your dataset:
...
+   dataset = fo.load_dataset("your_grouped_dataset")
+
+   # Example transformation: Assume that for all samples, the same
+   # transformation is applied. A 90 degree rotation around the x axis to align
+   # the point cloud and camera coordinate systems with 0 translation
+   rotation = np.array([[1, 0, 0],
+                        [0, 0, -1],
+                        [0, 1, 0]])
+   translation = np.array([0, 0, 0])
+   transformations = {sample.id: [(translation, rotation)] for sample in dataset}
```
allenleetc: Minor nit, is having `sample` do a vanilla iteration over `dataset` slightly discouraged? Instead eg could do (?)

`{_id: [etcetc] for _id in dataset.values('id')}`

docs/source/user_guide/using_datasets.rst
```diff
@@ -4274,6 +4274,134 @@ new field on your dataset:
...
+
+   # Example camera parameters: Assume that for all samples, the same camera
+   # intrinsics are used
+   camera_params = {
+       sample.id: {
+           "intrinsics": np.array([[1000, 0, 512],
+                                   [0, 1000, 384],
+                                   [0, 0, 1]])
+       }
+       for sample in dataset
```
allenleetc: Same comment as above

fiftyone/utils/labels.py
allenleetc: I also expected the "lidar"/3d slice name to come before the `camera_slice_name` (eg to match the src,dest ordering of in_field, out_field)

Maybe instead of in/out could be `slice_3d`, `slice_2d`, `field_3d`, `field_2d`. Just brainstorming

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,212 @@ def _get_filepath(sample_or_frame, sample):
...
+    transformations: Dict[str, fou3d.TransformationType],
+    camera_params: Dict[str, Dict[str, Any]],
+    forward_transform_flags: Dict[str, List[bool]] = None,
+    batch_size: int = 1000,
+    camera_model: Callable = fou3d.pinhole_projector,
+    progress: bool = None,
+):
+    """High-level orchestration of 3D → 2D label conversion. Processes the `in_field`
+    on the `lidar_slice_name` slice of a grouped dataset, uses the transformations
+    and camera intrinsics to converts the labels to 2D polylines and saves them in
```
allenleetc: "to convert the labels"

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,212 @@ def _get_filepath(sample_or_frame, sample):
...
+            the dataset.
+        forward_transform_flags: a dict mapping sample IDs to lists of booleans
+            indicating whether to apply forward or inverse transformations for
+            each transform in the list of transformations for each sample
+        batch_size (1000): number of samples to process in each batch
+        camera_model (fou3d.pinhole_projector): a callable that takes 3D points
+            in the camera coordinate system and the camera parameters dict and
+            returns the projected 2D points in pixel coordinates
+        progress (True): whether to show a progress bar during processing
+    """
```
allenleetc, line 1262: Brainstorm thought, in the future I wonder if it may be worth supporting scalar / non-dict values for `transformations`, `camera_params`, `forward_transform_flags` in cases where one or more of these is constant over the SampleCollection. (`forward_transform_flags` is perhaps the first most likely candidate)

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1205,206 @@
...
+                    f"Skipping sample {sample.id} because transformations or camera intrinsics are missing"
+                )
+                continue
+
+            group = dataset.get_group(sample.group.id)
+            cam_sample_id = group[camera_slice_name].id
+            width, height = (
+                camera_slice[cam_sample_id].metadata.width,
+                camera_slice[cam_sample_id].metadata.height,
+            )
```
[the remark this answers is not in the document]
allenleetc: @prernadh Hey this seems like a good point. If you had one 3d slice and multiple image slices, the transformations, camera parameters, and forwardflags should be keyed off the 2d/image slice sample IDs yes? Conceptually, I would store those on the 2d/image samples (if storing on dataset)

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1205,206 @@
...
+        return None, added_instance
+
+    points_cam, rotation_cam = fou3d.multiple_coordinate_transform(
+        det.location, det.rotation, transforms, forward_flags
+    )
+    if getattr(det, "instance", None) is None:
+        added_instance = True
+        det.instance = fo.Instance()
+        sample[in_field].detections[det_idx]["instance"] = det.instance
+
```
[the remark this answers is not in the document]
allenleetc: @prernadh I was also surprised to see `sample`, `in_field`, and `det_idx` in the input args list in addition to `det` as it seemed mildly redundant. I feel as I have been uncertain on the precise semantics of assignment in these cases however

fiftyone/utils/labels.py
prernadh: Done 👍

fiftyone/utils/labels.py
prernadh: Yes! Fixed :)

fiftyone/utils/utils3d.py
```diff
@@ -417,6 +432,215 @@ def _compute_intersection_points(box1, box2):
...
+    Args:
+        points: a 3xN np.ndarray containing the 3D coordinates of the points
+        cam_params: a dict containing the key 'intrinsics' that maps to a 3x3 or
+            4x4 np.ndarray representing the camera intrinsics matrix
+        normalize (True): whether to normalize the projected points by their
+            z-coordinate
+
+    Returns:
+        A 3xN np.ndarray containing the projected 2D coordinates of the points.
+        If `normalize` is True, the points are normalized by their z-coordinate.
```
allenleetc, line 588: I wonder when someone would want `normalize=False`

fiftyone/utils/utils3d.py
```diff
@@ -417,6 +432,215 @@ def _compute_intersection_points(box1, box2):
...
+    Returns:
+        True if all corners are visible in the image and all corners are in
+        front of the camera, False otherwise.
+    """
+    visible = np.logical_and(
+        corners_img[0, :] > 0, corners_img[0, :] < imsize[0]
+    )
+    visible = np.logical_and(visible, corners_img[1, :] < imsize[1])
+    visible = np.logical_and(visible, corners_img[1, :] > 0)
+    visible = np.logical_and(visible, corners_3d[2, :] > 1)
```
allenleetc: I am not sure I understand the criteria here. In world coords we may have arbitrary units, so what is the meaning of z>1?

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,223 @@ def _get_filepath(sample_or_frame, sample):
...
+    for group in sample_collection.select_fields(in_field).iter_groups(
+        group_slices=[spatial_slice_name, camera_slice_name],
+        progress=progress,
+        autosave=True,
+    ):
+        spatial_sample = group[spatial_slice_name]
+        camera_sample = group[camera_slice_name]
+        spatial_key = spatial_sample[transformation_key_field]
+        camera_key = camera_sample[camera_key_field]
+        transforms = transformations.get(spatial_key, None)
```
allenleetc: @prernadh are you sure you want `transformations` and `forward_transform_flags` to be keyed by the spatial sample ID? What if you have multiple camera slices, wouldn't those transformations and flags be keyed by the camera slice? (Conceptually, would those transformations be stored on the camera slices?)

docs/source/user_guide/using_datasets.rst
kaixi-wang: would the [translation, rotation] ever be different for individual samples within a group in a workflow? Seems like you just want to apply the transformation to an entire group_slice in this example, in which case you don't need to iterate so much, but idk if that's just for simplicity in the example

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,223 @@ def _get_filepath(sample_or_frame, sample):
...
+            continue
+
+        width, height = (
+            camera_slice[camera_sample.id].metadata.width,
+            camera_slice[camera_sample.id].metadata.height,
+        )
+
+        if width is None or height is None:
+            logger.warning(
+                f"Skipping camera sample {camera_sample.id} because camera metadata is missing. Please compute metadata for this sample first."
```
kaixi-wang: you already called compute metadata https://github.com/voxel51/fiftyone/pull/6277/files#diff-41b9587fdc5ac2683dcc9b89656cf2664c7a38aa9ec55b35f02cc032e87e961dR1270 so if it doesn't have metadata, there is probably an error

docs/source/user_guide/using_datasets.rst
kaixi-wang: yeah it feels like this iteration should just happen once in the body of iter_groups. Not sure why the need to pre-retrieve these values as you're querying for them again when you iterate...

fiftyone/utils/labels.py
prernadh: `forward_transform_flags` is an optional arg - so it does not need to be provided if its all forward transforms

fiftyone/utils/utils3d.py
prernadh: Nuscenes has similar functionality so just kept it in since its a utils function that should be reusable

fiftyone/utils/utils3d.py
prernadh: This z is not height, this would be the forward direction pointing into the camera plane. It's making sure the point is a tleast a certain distance away from the camera. Nuscenes utils does a very similar thing so I have kept it in here for parity

fiftyone/utils/labels.py
prernadh: Transformations relate two slices - spatial slice to camera slice in this example so it could be stored with either. I don't think users would want to store transformations on our samples at all. There might be multiple transformations involved and it will be a repetition of a lot of information. 

So I don't really think it matters whether we use spatial_sample.id or camera_sample.id since the user would have to construct this transformation matrix to get this information and it relates to both the slices.

docs/source/user_guide/using_datasets.rst
prernadh: Sure I can update this in our example docs. 

@kaixi-wang , yes this is a simplified example but users will likely have multiple transformations and different transformations for different subsets of their dataset. You might have multiple vehicles' data in the same dataset and the transformations for one vehicle is the same but is different for different vehicles.

docs/source/user_guide/using_datasets.rst
prernadh: Yep good points, will update.

fiftyone/utils/labels.py
allenleetc: OK. In my mind I had anchored the spatial slice more to worldcoords but I see your point

fiftyone/utils/utils3d.py
allenleetc: Yeah, understood. This functionality has general utility beyond A/V so the NuScenes convention may be less applicable in other use cases (eg where z<1 is relevant, or "1" in world coordinates represents a large distance). 

I think it's OK to leave here as a convention for A/V, just noting.

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1206,223 @@ def _get_filepath(sample_or_frame, sample):
...
         return sample.filepath
+
+
+def detections_3d_to_cuboids_2d(
+    sample_collection: SampleCollection,
+    camera_slice_name: str,
+    spatial_slice_name: str,
+    in_field: str,
+    out_field: str,
+    transformations: Dict[str, fou3d.TransformationType],
```
allenleetc, line 1215: V minor, the slice_name args and *_field args are not in consistent ordering (eg src,dst,src,dst)

fiftyone/utils/utils3d.py
prernadh: Yeah definitely fair. Let's see what feedback we get from users I guess

fiftyone/utils/labels.py
prernadh: Yeah but if compute_metadata skips a sample or something I wanted to explicitly log that projection is skipped for this sample.id

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1204,226 @@ def _get_filepath(sample_or_frame, sample):
...
+        if count % batch_size == 0:
+            sample_collection.set_values(
+                out_field, values_2d_label, key_field="id"
+            )
+            values_2d_label.clear()
+
+    if values_2d_label:
+        sample_collection.set_values(
+            out_field, values_2d_label, key_field="id"
+        )
```
swheaton: Why set `autosave=True` if you're going to write with `set_values()`? autosave does some batching on its own so you could try just using that. Or get rid of `autosave=True`

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1204,226 @@ def _get_filepath(sample_or_frame, sample):
...
+    detections_3d = sample[in_field]
+
+    if detections_3d is None:
+        return []
+    if len(detections_3d.detections) == 0:
+        return []
+
+    polylines = []
+
+    for _, det in enumerate(detections_3d.detections):
```
swheaton: `enumerate` is not necessary if you're not even using the index variable.

fiftyone/utils/utils3d.py
```diff
@@ -417,6 +428,225 @@ def _compute_intersection_points(box1, box2):
     return intersection_points
 
 
+def rpy_to_quaternion(euler_rpy: List[float]):
```
swheaton: just checking that the below utils are not available in a library we already use like scipy, open3d, etc. ?

fiftyone/utils/labels.py
prernadh: Autosaving is for the spatial_samples whose instances have been updated - specifically this line: https://github.com/voxel51/fiftyone/blob/411b47677bb50bf7f8d5e5a2c9bddc59665786d5/fiftyone/utils/labels.py#L1407

Yeah its a little complex but that was the best way I could think of updating it. Do you have recommendations on how to improve this? Sounds like your suggestion is allow `autosave` to do the saving on camera_slice_sample as well and drop `set_values`? Is that correct?

fiftyone/utils/labels.py
prernadh: Yes, remanent of a prev code logic. Can remove

fiftyone/utils/labels.py
swheaton: I mean it would be worth a quick try to see if you can leave everything to the autosave. But if it's significantly slower then I wouldn't spend much more time and just stick with what you've got.

fiftyone/utils/utils3d.py
prernadh: Yeah I do try to use other libraries as much as possible, but a lot of the utils are specific to how 3d detections are stored in FO. So there aren't libraries who do what we need as succinctly as these utils

fiftyone/utils/labels.py
brimoor: @prernadh +1 to @swheaton's suggestion to remove `set_values()` and rely on the autosave for all updates. That's what I had in mind in my original suggestion to use `autosave=True`:

```py
for group in sample_collection.select_fields(in_field).iter_groups(
    group_slices=[lidar_slice_name, camera_slice_name],
    progress=progress,
    autosave=True,
):
    lidar_sample = group[lidar_slice_name]
    camera_sample = group[camera_slice_name]

    # computation...

    camera_sample[out_field] = ...
```

Note that `iter_groups(..., autosave=True)` saves changes to *all samples in the emitted groups*, and thus in your current implementation it is already doing the work of checking to see if `camera_sample` has had any updates (which it won't currently); you might as well set `out_field` as shown above and let the autosave apply the updates.

kaixi-wang: > Autosaving is for the spatial_samples whose instances have been updated

Are you saying in the method you linked, it is expected to auto save an update to the existing field in addition to the conversion? 

I agree that you should go one way or the other (all auto or all explicit). IMO, it would be much easier long term to optimize/maintain if you were explicit with a set_values and no autosave, but if you go with autosave, highly recommend you keep the computation inside the body of the iteration like Brian laid out rather than breaking it out into a private helper as someone else could try to use it in a set_values() context without autosave and then it not work

prernadh: Sure - makes sense. Will update with using autosave for all samples. Would definitely like to optimize the code as much as possible though - so if autosave is not the recommended way for it @kaixi-wang , I am happy to try other methods to compare performance!

fiftyone/utils/utils3d.py
kaixi-wang: `fo.core.thread.transformation` seems to define a `Quarternion`. Is that different?/can it not be augmented instead of using pyq.Quaternion?

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1204,212 @@ def _get_filepath(sample_or_frame, sample):
...
+
+    if detections_3d is None:
+        return []
+    if len(detections_3d.detections) == 0:
+        return []
+
+    polylines = []
+
+    for det in detections_3d.detections:
+        ply = _process_detection_to_ply(
```
sashankaryal: nit: do we call polylines `ply` elsewhere? if yes, it's okay. if not, ply means something in 3D and i'd rather us use something else

fiftyone/utils/utils3d.py
prernadh: It doesn't have the same methods as pyq.Quaternion like q.rotation_matrix. I can add the method to convert the Quaternion to a rotation matrix though - so will take a look.

prernadh: Umm, actually the math to convert from rotation_matrix to quaternion is not the most stragihtforward. @sashankaryal What do you think? Worth adding to threed.transformations.Quaternion? 

This is the math - https://d3cw3dd2w32x2b.cloudfront.net/wp-content/uploads/2015/01/matrix-to-quat.pdf

sashankaryal: We've been using `scipy.spatial.transform` for 3D transformations - and pretty sure it offers a way to go from R to Q. Feel free to add it there!

fiftyone/utils/labels.py
```diff
@@ -1198,3 +1204,212 @@ def _get_filepath(sample_or_frame, sample):
...
+def detections_3d_to_cuboids_2d(
+    sample_collection: SampleCollection,
+    spatial_slice_name: str,
+    camera_slice_name: str,
+    in_field: str,
+    out_field: str,
+    transformations: Dict[str, fou3d.TransformationType],
+    camera_params: Dict[str, Dict[str, Any]],
+    forward_transform_flags: Optional[Dict[str, List[bool]]] = None,
+    camera_model: Callable = fou3d.pinhole_projector,
```
sashankaryal, line 1218: love that you can bring your own camera model!

fiftyone/utils/utils3d.py
```diff
@@ -417,6 +428,226 @@ def _compute_intersection_points(box1, box2):
...
+    assuming a pinhole camera.
+
+    The following orientation is assumed- x axis points to the right in the
+    image plane, y axis points down in the image plane and z axis points forward
+    from the camera.
+
+    Args:
+        points: a 3xN np.ndarray containing the 3D coordinates of the points
+        cam_params: a dict containing the key 'intrinsics' that maps to a 3x3 or
+            4x4 np.ndarray representing the camera intrinsics matrix
```
sashankaryal, line 582: what would 4x4 intrinsics mean for a pinhole camera? i understand what it'd mean for a non-pinhole camera (distortion, etc).

fiftyone/utils/utils3d.py
```diff
@@ -417,6 +428,226 @@ def _compute_intersection_points(box1, box2):
...
+        : camera_intrinsics.shape[0], : camera_intrinsics.shape[1]
+    ] = camera_intrinsics
+    nbr_points = points.shape[1]
+
+    points = np.concatenate((points, np.ones((1, nbr_points))))
+    points = np.dot(cam_int_pad, points)
+    points = points[:3, :]
+    if normalize:
+        points = points / points[2:3, :].repeat(3, 0).reshape(3, nbr_points)
+    return points[:2, :]
```
sashankaryal, line 610: should we consider culling points z<0 since they can't be projected to the image plane?

sashankaryal (approved):
Left some comments! Tested with an FO3D dataset, works flawlessly! Beautiful work 👏 

Can't wait to also make it available in the app JIT

fiftyone/utils/utils3d.py
prernadh: Yeah we perform those checks in the `points_in_front_of_camera` function in utils3d.py

fiftyone/utils/utils3d.py
prernadh: Yeah this is the superset case where we account for all acceptable intrinsic shapes

fiftyone/utils/utils3d.py
```diff
@@ -417,6 +428,226 @@ def _compute_intersection_points(box1, box2):
...
+    forward_transform_flags: List[bool] = None,
+) -> Tuple[List[float], List[float]]:
+    """
+    Applies a sequence of 3D coordinate frame transformations to a point and its
+    orientation. Each transformation consists of a translation vector and a
+    rotation matrix, applied in the order provided. The orientation is updated
+    at each step using quaternion multiplication.
+
+    Args:
+        points (list[float] or np.ndarray): A 3-element list/array representing
```
brimoor: nit: by convention we use the `()` to document default values, **not** types. 

(I know this is somewhat atypical; sorry!)

prernadh: Thanks - will fix!

fiftyone/utils/utils3d.py
prernadh: Will just use `scipy.spatial.transform` :D

PR #6279, opened 2025-08-20
Properly handle skip failures in compute metadata

CamronStaley:

Passing skip and warn failures all the way through compute metadata call stack. This will make it so there are no more silent failures (experienced this when we didn't have access to the media).

Copied the metadata.py over to FOT, then ran this script:

```
import fiftyone as fo

dataset_name = "quickstart-gcs-test-1"
fo.delete_dataset(dataset_name)
ds = fo.load_dataset("quickstart").clone(dataset_name)

filepaths = ds.values("filepath")

new_filepaths = []
for i in range(len(filepaths)):
    new_filepaths.append(f"gs://fiftyone-nonexistent-bucket/{i}")

ds.set_values("filepath", new_filepaths)

print(ds)
ds.compute_metadata(overwrite=True, warn_failures=True)
```

We are expecting to see a bunch of errors:
```
Failed to compute sample metadata for sample 'gs://fiftyone-nonexistent-bucket/196' due to: [Errno 2] No such file or directory: 'gs://fiftyone-nonexistent-bucket/196'
Failed to compute sample metadata for sample 'gs://fiftyone-nonexistent-bucket/197' due to: [Errno 2] No such file or directory: 'gs://fiftyone-nonexistent-bucket/197'
Failed to compute sample metadata for sample 'gs://fiftyone-nonexistent-bucket/198' due to: [Errno 2] No such file or directory: 'gs://fiftyone-nonexistent-bucket/198'
Failed to compute sample metadata for sample 'gs://fiftyone-nonexistent-bucket/199' due to: [Errno 2] No such file or directory: 'gs://fiftyone-nonexistent-bucket/199'
Failed to compute sample metadata for sample 'gs://fiftyone-nonexistent-bucket/1' due to: [Errno 2] No such file or directory: 'gs://fiftyone-nonexistent-bucket/1'
Failed to compute sample metadata for sample 'gs://fiftyone-nonexistent-bucket/165' due to: [Errno 2] No such file or directory: 'gs://fiftyone-nonexistent-bucket/165'
 100% |█████████████████████████████████████████████████████████████████████████████████████████| 200/200 [3.5ms elapsed, 0s remaining, 57.6K samples/s] 
```

        notes for FiftyOne users.

If you have compute metadata skip failures set to False it will now end execution when a sample fails to compute instead of silently swallowing them and then outputting "X samples couldn't populate metadata".

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Per-sample failure-handling for metadata computation: options to skip failing samples and to emit warnings when a sample fails.
  * Applies to both single- and multi-worker processing for more resilient metadata runs.

* **Behavior Change**
  * Metadata APIs now default to emitting warnings for per-sample failures, so failures are reported by default but do not stop processing.

### DIFF
--- a/fiftyone/core/metadata.py
+++ b/fiftyone/core/metadata.py
@@ -391,7 +391,11 @@ def compute_metadata(
 
     if num_workers <= 1:
         _compute_metadata(
-            sample_collection, overwrite=overwrite, progress=progress
+            sample_collection,
+            overwrite=overwrite,
+            progress=progress,
+            skip_failures=skip_failures,
+            warn_failures=warn_failures,
         )
     else:
         _compute_metadata_multi(
@@ -399,6 +403,8 @@ def compute_metadata(
             num_workers,
             overwrite=overwrite,
             progress=progress,
+            skip_failures=skip_failures,
+            warn_failures=warn_failures,
         )
 
     if skip_failures and not warn_failures:
@@ -469,7 +475,12 @@ def get_image_info(f):
 
 
 def _compute_metadata(
-    sample_collection, overwrite=False, batch_size=1000, progress=None
+    sample_collection,
+    overwrite=False,
+    batch_size=1000,
+    progress=None,
+    skip_failures=True,
+    warn_failures=False,
 ):
     if not overwrite:
         sample_collection = sample_collection.exists("metadata", False)
@@ -487,7 +498,14 @@ def _compute_metadata(
 
     cache = {}
     values = {}
-    inputs = zip(ids, filepaths, media_types, itertools.repeat(cache))
+    inputs = zip(
+        ids,
+        filepaths,
+        media_types,
+        itertools.repeat(cache),
+        itertools.repeat(skip_failures),
+        itertools.repeat(warn_failures),
+    )
 
     try:
         with fou.ProgressBar(total=num_samples, progress=progress) as pb:
@@ -509,6 +527,8 @@ def _compute_metadata_multi(
     overwrite=False,
     batch_size=1000,
     progress=None,
+    skip_failures=True,
+    warn_failures=False,
 ):
     if not overwrite:
         sample_collection = sample_collection.exists("metadata", False)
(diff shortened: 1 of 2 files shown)


fiftyone/core/metadata.py
```diff
@@ -296,7 +296,7 @@ def _parse_assets(scene, scene_path, cache=None):
                 asset_size += metadata.size_bytes
                 continue
 
-        tasks.append((None, asset_path, fom.MIXED, cache))
+        tasks.append((None, asset_path, fom.MIXED, cache, True, True))
```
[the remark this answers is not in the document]
swheaton: This is valid I think

fiftyone/core/metadata.py
```diff
@@ -545,19 +572,29 @@ def _compute_metadata_multi(
...
     try:
         return _get_metadata(filepath, media_type, cache=cache)
-    except:
+    except Exception as e:
+        if warn_failures:
+            logger.warning(
+                "Failed to compute sample metadata for sample '%s' due to: %s",
+                filepath,
+                e,
+            )
```
swheaton, line 660: This will excessively print warnings if there are many cases of the same failure (likely). Imagine getting 1M warnings spit out in the logs 😬 

It is very nice to have the actual error though instead of just "X samples failed, good luck figuring out which ones and why ;)"

Some options:
1. Add a [logging filter](https://stackoverflow.com/a/44692178/13228047) that filters this warning log after the first one. Then remove the filter when it's done.
2. Keep a count as a param to the private functions and only log if count < X or whatever.
3. Somehow keep track of one of the errors and bubble it back up so that the `logger.warning` occurs only once after all samples have been processed.

CamronStaley:
@swheaton updated the description with the lil script I am using

fiftyone/core/metadata.py
CamronStaley: This feels outside the scope of these changes imo. I think most places we don't dedupe logs and from a user perspective if I set "warn_failures" I'd expect to see all the failures. Interested to hear @jleven or @ehofesmann perspective tho

swheaton: Previously `warn_failures` would just print a warning summarizing how many samples failed. Now it will print the full error for every sample. If the failures are due to a fundamental issue like permissions, `ffmpeg`, etc., then you're going to get a million logs saying the same thing. This feels obnoxious to me.

We do not use `warn_failures` anywhere else really or log something within a tight sample loop, so I don't think the precedence argument is all that relevant

But understandably the implementation is awkward as I alluded to above so if we're OK with it then yeah lets not hold up the PR for it.

swheaton: Offline decision was to leave PR as is and circle back if it gets complaints.

PR #6280, opened 2025-08-20
Fix for sam/sam2 overwriting bounding boxes from prompt_label field

manushreegangwar:

SAM / SAM2 models with box prompting should preserve the original prompt boxes instead of overwriting them in generated `Detection` labels with instance `mask`

WIP

        notes for FiftyOne users.

SAM/SAM2 preserve box dimensions from box prompts in the generate `Detection` output labels.

-   [x] Other


fiftyone/utils/sam.py
```diff
@@ -279,7 +281,7 @@ def _forward_pass_boxes(self, imgs):
             )
             outputs.append(
                 {
-                    "boxes": input_boxes,
+                    "boxes": torch.tensor(boxes_xyxy),
```
manushreegangwar, line 284: @jacobsela these are on cpu now. good catch.

jacobsela: 👍

fiftyone/utils/sam.py
```diff
@@ -257,7 +258,8 @@ def _forward_pass_boxes(self, imgs):
             h, w = img.size(1), img.size(2)
 
             boxes = [d.bounding_box for d in detections.detections]
-            sam_boxes = np.array([_to_sam_box(box, w, h) for box in boxes])
+            boxes_xyxy = np.array([_to_abs_box(box, w, h) for box in boxes])
+            sam_boxes = np.round(boxes_xyxy).astype(int)
```
PeterOttoMoeller, line 262: This calls rounding and converting to int on all boxes instead of on each one individually, right? The reason is runtime optimization, presumably...? Why don't we go a step further and remove the loop here and convert all boxes simultaneously?

fiftyone/utils/sam2.py
```diff
@@ -400,11 +400,12 @@ def _forward_pass_boxes(self, video_reader, sample):
                     ann_obj_id = current_obj_idx
                     current_obj_idx += 1
                 classes_obj_id_map[ann_obj_id] = detection.label
-                box = fosam._to_sam_box(
+                box_xyxy = fosam._to_abs_box(
                     detection.bounding_box,
                     self._curr_frame_width,
                     self._curr_frame_height,
                 )
+                box = np.round(box_xyxy).astype(int)
```
PeterOttoMoeller: I haven't really looked deeply enough into this to have a strong opnion, here. I am assuming we call these operations here on individual boxes again because somehow the interfaces differ to the above models, correct? Still, makes me wonder if this shouldn't/couldn't be consolidated.

manushreegangwar: There's room for improvement and optimization here but since this is a bug fix PR, I am keeping it focused on the fix.  We should revisit the `SegmentAnything` classes to optimize them or perhaps integrate these models via `transformers` to avoid having to maintain separate classes. I haven't looked at the implementation in detail to say why it was originally implemented as a per-box format instead of a vectorized format.

fiftyone/utils/sam.py
manushreegangwar: That's a good point. I can create a vectorized version of `_to_abs_box` function but still need to preserve the per box version for the video model unless we do all the vectorization related changes. However, that is beyond the scope of this PR.

The reason for doing this change is to avoid rounded boxes to be outputed but it also has the added benefit of rounding being vectorized in line 262.

manushreegangwar: Updated to numpy vectorized form.

fiftyone/utils/sam2.py
PeterOttoMoeller: Great, looks good to me, do you mind creating a backlog ticket reflecting this? Feel free to close conversation.

manushreegangwar: there's a ticket is in the backlog for sam2 video models

PR #6283, opened 2025-08-21
Update average default to binary for binary classification eval

manushreegangwar:

1. Test https://docs.voxel51.com/user_guide/evaluation.html#binary-evaluation

```
dataset = fo.load_dataset("cifar10-binary")
classes = ["other", "cat"]

    "predictions",
    gt_field="ground_truth",
    eval_key="eval_binary", # eval_binary_dev on develop for the same dataset
    method="binary",
    classes=classes,
)

results.print_report()
```
Screenshot from 2025-08-21 13-52-59

2. Compare that other evaluations remain unchanged on `develop` vs this branch
- done for simple classification and detection type evaluations

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/eval/base.py
+++ b/fiftyone/utils/eval/base.py
@@ -739,14 +739,10 @@ class BaseClassificationResults(BaseEvaluationResults):
             self.ytrue, self.ypred, labels=labels, weights=self.weights
         )
 
-        precision, recall, fscore, _ = skm.precision_recall_fscore_support(
-            self.ytrue,
-            self.ypred,
+        precision, recall, fscore = self._precision_recall_fscore(
             average=average,
             labels=labels,
             beta=beta,
-            sample_weight=self.weights,
-            zero_division=0,
         )
 
         support = _compute_support(
@@ -891,6 +887,18 @@ class BaseClassificationResults(BaseEvaluationResults):
             return np.asarray(classes)
 
         return np.array([c for c in self.classes if c != self.missing])
+
+    def _precision_recall_fscore(self, labels, average, beta=1.0):
+        precision, recall, fscore, _ = skm.precision_recall_fscore_support(
+            self.ytrue,
+            self.ypred,
+            average=average,
+            labels=labels,
+            beta=beta,
+            sample_weight=self.weights,
+            zero_division=0,
+        )
+        return precision, recall, fscore
 
     def _confusion_matrix(
         self,
--- a/fiftyone/utils/eval/classification.py
+++ b/fiftyone/utils/eval/classification.py
@@ -5,6 +5,7 @@ Classification evaluation.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+
 from copy import deepcopy
 import inspect
 import itertools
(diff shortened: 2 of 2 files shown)


fiftyone/utils/eval/base.py
```diff
@@ -892,6 +888,18 @@ def _parse_classes(self, classes):
 
         return np.array([c for c in self.classes if c != self.missing])
 
+    def _precision_recall_fscore(self, labels, average, beta=1.0):
```
PeterOttoMoeller: Why do we introduce this function? 

Also, we re-introduce a default value for beta here, the same as is already defined in the scope where this function is called.

fiftyone/utils/eval/classification.py
```diff
@@ -901,6 +902,26 @@ def plot_roc_curve(self, backend="plotly", **kwargs):
             **kwargs,
         )
 
+    def metrics(self, classes=None, average="binary", beta=1.0):
```
PeterOttoMoeller, line 905: Is this redefinition necessary if we just call the parent class' version of this function? Also, this again reintroduces default values for the call arguments.

fiftyone/utils/eval/classification.py
```diff
@@ -901,6 +902,26 @@ def plot_roc_curve(self, backend="plotly", **kwargs):
...
+        return super().metrics(classes, average, beta)
+
+    def _precision_recall_fscore(self, labels, average, beta=1.0):
+        if average == "binary":
```
PeterOttoMoeller: Is this if/else the fix to the issue described in the PR description?

fiftyone/utils/eval/base.py
PeterOttoMoeller: I get why it was reintroduced, now. You want to overwrite this function in the grandchild class. That makes sense! Still, the comment about default values for function arguments remains valid, I think.

fiftyone/utils/eval/classification.py
```diff
@@ -901,6 +902,26 @@ def plot_roc_curve(self, backend="plotly", **kwargs):
...
+            pos_label = self._pos_label
+        else:
+            pos_label = 1  # set to sklearn default to suppress warning
+        precision, recall, fscore, _ = skm.precision_recall_fscore_support(
```
PeterOttoMoeller: instead of putting the call to `skm.precision_recall_fscore_support()` here again, couldn't we just make `pos_labels` part of the grandparent class' `_precision_recall_fscore` signature and then call `super()._precision_recall_fscore()` after setting `pos_labels` here?

fiftyone/utils/eval/classification.py
manushreegangwar: No, this is to suppress sklearn warning which is raised when `pos_label = self._pos_label` but average is not set to `binary`.

fiftyone/utils/eval/classification.py
manushreegangwar: Initially I was using `getattr` in the base class `_precision_recall_fscore` to set `_pos_label` but decided to override the method in this class to avoid possible errors related to renaming of `_pos_label` since it is not used elsewhere. I can revert the override. It is not necessary.

fiftyone/utils/eval/classification.py
manushreegangwar: I need to update the default value for `average` parameter here.

fiftyone/utils/eval/base.py
manushreegangwar: Good point. I will remove the default value of `beta`

fiftyone/utils/eval/base.py
```diff
@@ -892,6 +888,20 @@ def _parse_classes(self, classes):
...
+        pos_label = getattr(self, "_pos_label", None)
+        precision, recall, fscore, _ = skm.precision_recall_fscore_support(
+            self.ytrue,
+            self.ypred,
+            average=average,
+            labels=labels,
+            beta=beta,
+            sample_weight=self.weights,
+            zero_division=0,
+            pos_label=pos_label if average == "binary" else None,
```
manushreegangwar, line 901: https://github.com/scikit-learn/scikit-learn/blob/c5497b7f7/sklearn/metrics/_classification.py#L1784 
to suppress this warning

PR #6285, opened 2025-08-22
FOEPD-1964 Torch num worker recommender respects configured max

swheaton:

Consolidate a couple different impls of recommending num workers / process pool workers.

In doing so, makes sure tests and checks are always run. As well as respecting the configured max in `fo.config.max_process_pool_workers`.

TBD

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Worker recommendation utilities now accept optional hints to influence default worker counts across loaders and process pools.

* **Bug Fixes**
  * Unified worker-count resolution with robust fallbacks, clamping to configured maxima and enforcing non-negative values; daemon processes return 0 (no multiprocessing).

* **Documentation**
  * Clarified daemon behavior, 0 semantics, and resolution order for default worker counts.

* **Tests**
  * Added comprehensive tests for worker recommendation logic; removed obsolete single-worker enforcement test.

### DIFF
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -835,8 +835,7 @@ def _make_data_loader(
     # functionality
     use_numpy = not isinstance(model, TorchModelMixin)
 
-    if num_workers is None:
-        num_workers = fout.recommend_num_workers()
+    num_workers = fout.recommend_num_workers(num_workers)
 
     if model.has_collate_fn:
         user_collate_fn = model.collate_fn
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -1934,8 +1934,7 @@ def _make_patch_data_loader(
     # functionality
     use_numpy = not isinstance(model, TorchModelMixin)
 
-    if num_workers is None:
-        num_workers = fout.recommend_num_workers()
+    num_workers = fout.recommend_num_workers()
 
     dataset = fout.TorchImagePatchesDataset(
         samples=samples,
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -2873,19 +2873,43 @@ def recommend_thread_pool_workers(num_workers=None):
     return num_workers
 
 
-def recommend_process_pool_workers(num_workers=None):
+def recommend_process_pool_workers(num_workers=None, default_num_workers=None):
     """Recommends a number of workers for a process pool.
 
-    If a ``fo.config.max_process_pool_workers`` is set, this limit is applied.
+    If this process is a daemon, the number of workers is 0.
+
+    If ``num_workers`` is None, the following order is used to determine the
+    number of workers:
+
+    - The configured (``fo.config``) default number of workers
+    - The passed-in default number of workers (``default_num_workers``)
+    - The system CPU count
+
+    If ``fo.config.max_process_pool_workers`` is set, this limit is applied.
 
     Args:
         num_workers (None): a suggested number of workers
+        default_num_workers (None): a default number of workers to use if
+            ``num_workers`` is None and no configured default is set
 
     Returns:
         a number of workers
     """
-    if num_workers is None:
-        num_workers = multiprocessing.cpu_count()
+    try:
+        if multiprocessing.current_process().daemon:
+            num_workers = 0
+        elif num_workers is None:
+            # Order:
+            # 1. Configured default
+            # 2. Passed-in default
+            # 3. System CPU count
+            num_workers = (
+                fo.config.default_process_pool_workers
+                or default_num_workers
+                or multiprocessing.cpu_count()
+            )
+    except:
+        num_workers = 4
 
     if fo.config.max_process_pool_workers is not None:
         num_workers = min(num_workers, fo.config.max_process_pool_workers)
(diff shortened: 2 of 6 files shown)


fiftyone/core/utils.py
```diff
@@ -2873,19 +2873,43 @@ def recommend_thread_pool_workers(num_workers=None):
...
+            # Order:
+            # 1. Configured default
+            # 2. Passed-in default
+            # 3. System CPU count
+            num_workers = (
+                fo.config.default_process_pool_workers
+                or default_num_workers
+                or multiprocessing.cpu_count()
+            )
+    except:
```
CamronStaley: when would this throw an exception? Also should it default to 4 here or should we just have fo.config.default_process_pool_workers be 4 and set num workers equal to it here?

fiftyone/core/utils.py
```diff
@@ -2873,19 +2873,43 @@ def recommend_thread_pool_workers(num_workers=None):
...
+            ``num_workers`` is None and no configured default is set
 
     Returns:
         a number of workers
     """
-    if num_workers is None:
-        num_workers = multiprocessing.cpu_count()
+    try:
+        if multiprocessing.current_process().daemon:
+            num_workers = 0
```
kaixi-wang, line 2904: This is confusing. If it needs to be 1 then why not just return 1

kaixi-wang (changes requested):

fiftyone/core/utils.py
swheaton: No idea, just copied from the torch utils and wanted to ensure the same functionality.

No the idea is that it should probably never be 4, that's like a last resort. We will take the configured default if present, then the passed-in default, then cpu_count. And if we can't get cpu count for some reason then it goes to 4 :shrug:

fiftyone/core/utils.py
swheaton: Not entirely sure tbh I got this from the map samples stuff, I would have to look and see if `num_workers==1` makes sense or if 0 is required for some reason

swheaton: will fix to make more sense

fiftyone/core/utils.py
```diff
@@ -2873,22 +2873,58 @@ def recommend_thread_pool_workers(num_workers=None):
...
     Returns:
-        a number of workers
+        a number of workers. 0 means no multiprocessing
     """
-    if num_workers is None:
-        num_workers = multiprocessing.cpu_count()
+    try:
+        # "daemonic processes are not allowed to have children"
+        if multiprocessing.current_process().daemon:
+            num_workers = 0
```
kaixi-wang, line 2904: shouldn't this be 1 to match the previous? https://github.com/voxel51/fiftyone/pull/6285/files#diff-776c65a8e888d7a2cacab280fa5915af19d3f7eae54641bc2f30f4d44014bc9eL57

kaixi-wang (dismissed):

fiftyone/core/utils.py
swheaton: I changed it to make more sense and be consistent.
The code later on will treat 0 the same as 1 anyway (`num_workers <= 1`)
https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/map/mapper.py#L256-L270

tests/unittests/utils_tests.py
```diff
@@ -953,6 +954,65 @@ def test_quiet_overwrites_all(self):
...
+        self.assertEqual(fou.recommend_process_pool_workers(8, 4), 8)
+
+        # Negative number coerced to 0
+        self.assertEqual(fou.recommend_process_pool_workers(-1), 0)
+
+        # Number is capped by config
+        with patch.object(fo.config, "max_process_pool_workers", 2):
+            self.assertEqual(fou.recommend_process_pool_workers(8, 4), 2)
+
+    def test_default_from_config(self):
```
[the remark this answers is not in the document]
swheaton: Yes I will do that for the first bullet.
For the second, I'm passing the default arg to test that it is indeed ignored in favor of the explicitly passed num_workers.

CamronStaley (approved):

fiftyone/core/utils.py
kaixi-wang: yeah but the comment says 1 and anyone not digging through the code could make the mistake of using this in a daemon process and getting 0, and that being treated as not set and so it would use some other default

fiftyone/core/models.py
```diff
@@ -835,8 +835,7 @@ def _make_data_loader(
     # functionality
     use_numpy = not isinstance(model, TorchModelMixin)
 
-    if num_workers is None:
-        num_workers = fout.recommend_num_workers()
+    num_workers = fout.recommend_num_workers(num_workers)
```
jacobsela, line 838: why is the if removed?

swheaton: Because the `fo.core.utils.recommend_process_pool_workers()` now takes care of the condition. As well as enforces some other things, namely, a configurable cap on number of processes.

jacobsela: ah cool ok

fiftyone/core/utils.py
swheaton: I don't see a comment that says 1.

> digging through the code

In the docstring?

PR #6288, opened 2025-08-22
FOEPD-1976: Save predicted sample labels in background thread

exupero:

This PR updates `_apply_image_model_data_loader` to execute DB writes on a background thread so calls to `predict_all` aren't blocked. It only moves the top-level I/O-bound bottleneck off the main thread, it does not address any CPU-bound bottlenecks within `predict_all`.

Ran following script on voxelcompute instance and profiled with both built-in cProfile module and [line-profiler](https://pypi.org/project/line-profiler/).
```python
import sys

import fiftyone as fo
import fiftyone.zoo as foz


if __name__ == "__main__":
    suffix = sys.argv[1] if len(sys.argv) > 1 else "baseline"
    if not dataset.persistent:
        dataset.persistent = True
    dataset.apply_model(
	    model,
        num_workers=32,
        batch_size=16,
        confidence_thresh=0.1,
        label_field=f"profile_{suffix}"
	)
```
Code on develop ran in 5m47s, with 327s spent in `apply_model` and 198s in `predict_all` (60%); 77s was spent saving collections.
Screenshot 2025-08-20 at 12 30 28 PM
I do see some screwy behavior in the Snakeviz diagram, specifically that `apply_model` takes 327s, which besides the 275s in `predict_all` and saving collections also includes calling `check_perm_wrapper` for 327s (wat). Snakeviz says that calls the same `apply_model` model function. From looking at the code I'm not sure there's really any recursion happening and it's possibly a data collision in the cProfile data, but I don't know for sure, so I also profiled `_apply_image_model_data_loader` with line_profiler, which indicated that the 40% of non-`predict_all` time is spent in `sample.add_labels(...)` and `ctx.save(sample)`.

Code in this PR ran in 4m15s, with 236s spent in `apply_model` and 227s in `predict_all` (96%). Line-profile shows similar results.

Note that these changes make an overall improvement but `predict_all` is now slower:
visualization (2)
I tracked that difference down to a line in a library that only turns a tensor into a tuple, so the change is likely due to thrashing the GIL between threads. Preliminary investigation shows that it would be faster to use a background process instead of a thread, but some of the objects being saved cannot be pickled, so it's not possible to hand them off to another process.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Batch labeling and embedding saves now run asynchronously to improve throughput and reduce blocking during disk writes.
  * Added a configurable background executor to schedule and run batch save tasks.

* **Bug Fixes / Reliability**
  * Centralized, configurable batch-level error handling: optionally skip failed batches while continuing processing.
  * Clear warnings for skipped batches include contextual sample identifiers to aid diagnosis.

### DIFF
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -1219,6 +1219,18 @@ def _compute_image_embeddings_data_loader(
         pb = context.enter_context(fou.ProgressBar(samples, progress=progress))
         if embeddings_field is not None:
             ctx = context.enter_context(foc.SaveContext(samples))
+        submit = context.enter_context(
+            fou.async_executor(
+                max_workers=1,
+                skip_failures=skip_failures,
+                warning="Async failure saving embeddings",
+            )
+        )
+
+        def save_batch(sample_batch, embeddings_batch):
+            for sample, embedding in zip(sample_batch, embeddings_batch):
+                sample[embeddings_field] = embedding
+                ctx.save(sample)
 
         for sample_batch, imgs in zip(
             fou.iter_batches(samples, batch_size),
@@ -1244,9 +1256,7 @@ def _compute_image_embeddings_data_loader(
                 )
 
             if embeddings_field is not None:
-                for sample, embedding in zip(sample_batch, embeddings_batch):
-                    sample[embeddings_field] = embedding
-                    ctx.save(sample)
+                submit(save_batch, sample_batch, embeddings_batch)
             else:
                 embeddings.extend(embeddings_batch)
 
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -85,6 +85,7 @@ def handle_batch_error(skip_failures, sample_batch):
             sample_batch[0].id,
             sample_batch[-1].id,
             e,
+            exc_info=True,
         )
 
 
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -52,29 +52,6 @@ _ALLOWED_PATCH_TYPES = (
 
 
 @contextlib.contextmanager
-def _async_executor(
-    *, max_workers, skip_failures=False, warning="Async failure"
-):
-    with ThreadPoolExecutor(max_workers=max_workers) as executor:
-        _futures = []
-
-        def submit(*args, **kwargs):
-            future = executor.submit(*args, **kwargs)
-            _futures.append(future)
-            return future
-
-        yield submit
-
-        for future in _futures:
-            try:
-                future.result()
-            except Exception as e:
-                if not skip_failures:
-                    raise e
-                logger.warning(warning, exc_info=True)
-
-
-@contextlib.contextmanager
 def _handle_batch_error(skip_failures, sample_batch):
     try:
         yield
@@ -503,7 +480,7 @@ def _apply_image_model_data_loader(
         pb = context.enter_context(fou.ProgressBar(samples, progress=progress))
         ctx = context.enter_context(foc.SaveContext(samples))
         submit = context.enter_context(
-            _async_executor(
+            fou.async_executor(
                 max_workers=1,
                 skip_failures=skip_failures,
                 warning="Async failure labeling batches",
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -480,21 +480,21 @@ def _apply_image_model_data_loader(
     else:
         samples = samples.select_fields()
 
-    def save_batch(sample_batch, labels_batch):
-        for sample, labels in zip(sample_batch, labels_batch):
-            if filename_maker is not None:
-                _export_arrays(labels, sample.filepath, filename_maker)
-
-            sample.add_labels(
-                labels,
-                label_field=label_field,
-                confidence_thresh=confidence_thresh,
-            )
-            ctx.save(sample)
-
     with contextlib.ExitStack() as context:
         pb = context.enter_context(fou.ProgressBar(samples, progress=progress))
         ctx = context.enter_context(foc.SaveContext(samples))
+
+        def save_batch(sample_batch, labels_batch):
+            for sample, labels in zip(sample_batch, labels_batch):
+                if filename_maker is not None:
+                    _export_arrays(labels, sample.filepath, filename_maker)
+
+                sample.add_labels(
+                    labels,
+                    label_field=label_field,
+                    confidence_thresh=confidence_thresh,
+                )
+                ctx.save(sample)
 
         with futures(
             max_workers=1,
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -62,7 +62,7 @@ def futures(*, max_workers, skip_failures=False, warning="Async failure"):
             except Exception as e:
                 if not skip_failures:
                     raise e
-                logger.warning(warning, e)
+                logger.warning(warning, exc_info=True)
 
 
 def apply_model(
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -73,7 +73,7 @@ def futures(*, max_workers, skip_failures=False, warning="Async failure"):
 
 
 @contextlib.contextmanager
-def handle_batch_error(skip_failures, sample_batch):
+def _handle_batch_error(skip_failures, sample_batch):
     try:
         yield
     except Exception as e:
@@ -502,7 +502,7 @@ def _apply_image_model_data_loader(
         ctx = context.enter_context(foc.SaveContext(samples))
 
         def save_batch(sample_batch, labels_batch):
-            with handle_batch_error(skip_failures, sample_batch):
+            with _handle_batch_error(skip_failures, sample_batch):
                 for sample, labels in zip(sample_batch, labels_batch):
                     if filename_maker is not None:
                         _export_arrays(labels, sample.filepath, filename_maker)
@@ -523,7 +523,7 @@ def _apply_image_model_data_loader(
                 fou.iter_batches(samples, batch_size),
                 data_loader,
             ):
-                with handle_batch_error(skip_failures, sample_batch):
+                with _handle_batch_error(skip_failures, sample_batch):
                     if isinstance(imgs, Exception):
                         raise imgs
 
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -6,7 +6,6 @@ FiftyOne models.
 |
 """
 
-from concurrent.futures import ThreadPoolExecutor
 import contextlib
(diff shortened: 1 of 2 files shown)


fiftyone/core/models.py
```diff
@@ -487,18 +516,29 @@ def _apply_image_model_data_loader(
                     )
                     ctx.save(sample)
 
-            except Exception as e:
-                if not skip_failures:
-                    raise e
+        with _async_executor(
```
brimoor: You could avoid one level of indentation here by continuing the `context.enter_context()` pattern used above:

```py
submit = context.enter_context(_async_executor(...))

for sample_batch, imgs in zip(...):
    ...
```

Note that there's a concrete reason that `context.enter_context()` is used here: the [corresponding code in `fiftyone-teams`](https://github.com/voxel51/fiftyone-teams/blob/de11d3951876e86d00544a9c41d324478ff64473/fiftyone/core/models.py#L479-L498) contains an extra context that is unique to processing cloud data, and writing the code in that way minimizes merge conflicts when pulling changes in this code area over to `fiftyone-teams`.

fiftyone/core/models.py
```diff
@@ -49,6 +51,46 @@
 )
 
 
+@contextlib.contextmanager
```
brimoor, line 53: Seems like this utility may be useful in multiple contexts besides just model inference. Should we move it to `fiftyone.core.utils`?

fiftyone/core/models.py
```diff
@@ -461,21 +503,8 @@ def _apply_image_model_data_loader(
...
-
-                if needs_samples:
-                    labels_batch = model.predict_all(
-                        imgs, samples=sample_batch
-                    )
-                else:
-                    labels_batch = model.predict_all(imgs)
-
+        def save_batch(sample_batch, labels_batch):
+            with _handle_batch_error(skip_failures, sample_batch):
```
brimoor, line 490: @exupero @jacobsela @PeterOttoMoeller following up on a conversation I saw on another PR- I'd like to second the idea of incorporating async saving directly into `SaveContext`, as this would make async-saving available to many use cases without custom code.

Here's roughly what's happening in the code in question:

```py
# 0. Torch model inference
labels_batch = model.predict_all(imgs)

# 1. Validate that provided value is compliant with dataset schema
# 2. Store label(s) on Sample in-memory
sample.add_labels(...)

# 3. Compute sample delta represented as mongo operations
# 4. If enough ops have been accumulated, flush them to the DB
ctx.save(sample)
```

The only part that involves network communication is 4, so I'm surprised that putting 1-3 in a thread pool of size 1 (or larger for that matter) would actually result in more performance, due to Python GIL. Threading simple Python statements should not help anything. I guess that since 0 is leveraging low-level stuff like GPU it can happen in parallel with another thread?

brimoor: It is regrettable that 1 and 3 aren't negligible, given that they don't require any network communication (I saw another conversation stating that `add_labels()` takes some time). This is due to the underlying `mongoengine` library that we rely on being fairly wasteful in how it manages changes to documents. Removal of `mongoengine` from FO to improve performance has long been tech debt that we've wanted to address.

brimoor: I'm mostly interested in identifying the most reusable pattern for optimizing this code we can, as there are many examples of FO code that need to save updates to the DB, and my *intuition* is that the network overhead is the right/common bottleneck that we need to relieve.

exupero: > The only part that involves network communication is 4, so I'm surprised that putting 1-3 in a thread pool of size 1 (or larger for that matter) would actually result in more performance, due to Python GIL. Threading simple Python statements should not help anything. I guess that since 0 is leveraging low-level stuff like GPU it can happen in parallel with another thread?

Yes, my understanding is that GPU operations don't use Python bytecode and therefore don't lock the interpreter, allowing GIL-based operations to be executed in parallel.

> network overhead is the right/common bottleneck that we need to relieve.

I have found I/O to be the lowest hanging fruit. @jacobsela also identified costs of transforming tensors after inference, and moving that transformation logic to the background thread made additional improvements (those changes to come in a follow-on PR).

kaixi-wang: > have found I/O to be the lowest hanging fruit.

I have found download in addition to writes be a significant reason for underutilization of gpu and a huge bottleneck

> @jacobsela also identified costs of transforming tensors after inference, and moving that transformation logic to the background thread made additional improvements (those changes to come in a follow-on PR).

Do you mean for parsing the results?

exupero: Do you have a script that downloads media incrementally? The ones I was trying seemed to download all the images before applying a model.

> Do you mean for parsing the results?

I'm talking about the output of `_forward_pass` (appears to be tensor-like) which gets passed to `_output_processor` (which returns FO objects). Since `_output_processor` is CPU-bound, moving it to the background thread unblocks the next batch to run on the GPU.

brimoor: > Do you have a script that downloads media incrementally? The ones I was trying seemed to download all the images before applying a model.

@jacobsela's recent work to introduce `to_torch()` --- which is used internally by `apply_model(model, ...)` whenever `model` is a Torch model --- has made it so that media is downloaded in the data loader workers. So *not* all up front and *not* in the main process where `predict()` is called.

We should continue any discussion about media download offline, though, as it is not pertinent to the open source project.

fiftyone/core/utils.py
```diff
@@ -3143,3 +3143,26 @@ def validate_hex_color(value):
 
 
 fos = lazy_import("fiftyone.core.storage")
+
+
+@contextmanager
+def async_executor(
```
brimoor, line 3149: All public methods in core modules should get full docstrings (see other methods in module for reference). Can you add that please? 😄

fiftyone/core/models.py
brimoor: Thinking ahead to other workflows that may need optimizing: I expect the vast majority of code will be subject to Python GIL, and thus using `async_executor()` to offload anything besides `sample.save()` is not likely to help.

Therefore, the most generally useful abstractions will be:
1. Incorporating async saving directly into `SaveContext`, so that I/O is automatically offloaded to a background thread with zero code change to users
2. Horizontally scaling the whole operation onto multiple processes. IE we need to use the recently introduced `map_samples()` in more places in the SDK

brimoor: To be clear, I am onboard with making the bespoke optimizations proposed in this PR, as optimizing model inference is of particular value.

brimoor (commented):
@exupero can you apply this same change to `_compute_image_embeddings_data_loader()`?

Then we'll have optimized the two most common use cases:
- `apply_model(<torch model>, ...)`
- `compute_embeddings(<torch model>, ...)`

exupero:
I updated `_compute_image_embeddings_data_loader` to use the same approach (see patch below), but I didn't get any improvement in overall run time. Let me know if there's a different implementation to try. `apply_model` spends a lot of time validating samples and getting their attributes, which doesn't seem to be happening here, so saving context asynchronously may not have the same impact it had for `apply_model`.

Here's my test script. Let me know if there's a better scenario to benchmark:
```python
import sys

import fiftyone as fo
import fiftyone.zoo as foz

if __name__ == "__main__":
    suffix = sys.argv[1] if len(sys.argv) > 1 else "baseline"
    if not dataset.persistent:
        dataset.persistent = True
    model = foz.load_zoo_model("resnet101-imagenet-torch")
    dataset.compute_embeddings(model, num_workers=32, skip_failures=False)
```

And here's the code changes I tried:
```diff
diff --git i/fiftyone/core/models.py w/fiftyone/core/models.py
index 0bf2ca6ddd..51a506f496 100644
--- i/fiftyone/core/models.py
+++ w/fiftyone/core/models.py
@@ -1219,6 +1219,18 @@ def _compute_image_embeddings_data_loader(
         pb = context.enter_context(fou.ProgressBar(samples, progress=progress))
         if embeddings_field is not None:
             ctx = context.enter_context(foc.SaveContext(samples))
+        submit = context.enter_context(
+            fou.async_executor(
+                max_workers=1,
+                skip_failures=skip_failures,
+                warning="Async failure saving embeddings",
+            )
+        )
+
+        def save_batch(sample_batch, embeddings_batch):
+            for sample, embedding in zip(sample_batch, embeddings_batch):
+                sample[embeddings_field] = embedding
+                ctx.save(sample)
 
         for sample_batch, imgs in zip(
             fou.iter_batches(samples, batch_size),
@@ -1244,9 +1256,7 @@ def _compute_image_embeddings_data_loader(
                 )
 
             if embeddings_field is not None:
-                for sample, embedding in zip(sample_batch, embeddings_batch):
-                    sample[embeddings_field] = embedding
-                    ctx.save(sample)
+                submit(save_batch, sample_batch, embeddings_batch)
             else:
                 embeddings.extend(embeddings_batch)
 ```

With and without these changes, the script took about 55s to run on a GPU.

**Comment by brimoor** [comment]
@exupero this syntax:

```py
embeddings = dataset.compute_embeddings(model, ...)
```

computes embeddings and returns them as an array in-memory. So there's no writes to the database, and there's also no conversion from Torch tensors to FiftyOne `Label` instances. So there's no post-processing to offload.

However, there's this syntax:

```py
dataset.compute_embeddings(model, embeddings_field="<field-name>", ...)
```

where the embeddings are stored in an `embeddings_field` for each sample. This is analogous to `apply_model(..., label_field="<label-field>")`: we have database writes and sample field validation that will be optimized when offloaded to a thread.

kaixi-wang (approved):
Code looks good and makes sense, but didn't test.

fiftyone/core/utils.py
```diff
@@ -3143,3 +3143,45 @@ def validate_hex_color(value):
...
+    *, max_workers, skip_failures=False, warning="Async failure"
+):
+    """
+    Context manager that provides a function for submitting tasks to a thread
+    pool executor. All tasks are joined when the context is exited.
+
+    Example:
```
brimoor, line 3156: @exupero FYI in order for your example code to be rendered as a code block in the method's docstring at https://docs.voxel51.com, you need to use a double colon `Example::` (**unlike** `Args:` and `Returns:`, which are special Sphinx keywords).

For example: 
- source: https://github.com/voxel51/fiftyone/blame/2853b72cb393c1b08d1f14bcfd06da5ee7d3ae50/fiftyone/core/utils.py#L736-L744
- rendered: https://docs.voxel51.com/api/fiftyone.core.utils.html#fiftyone.core.utils.lazy_import

exupero: Thanks, fixed in #6309.

PR #6293, opened 2025-08-25
[NO MERGE] Ray postprocessing and writes. kinda goes zoom.

jacobsela:
fully offload post processing and writes.

Loading and offloading tensors from GPU is still on main process. Not 100% how (and if we should) get them off. At this point it may be better to just scale horizontally with DOs.

```python
from time import time_ns

import fiftyone as fo
import fiftyone.zoo as foz

dataset = fo.load_dataset("coco-val")

start = time_ns()
dataset.apply_model(
    model,
    label_field="predictions_ray",
    batch_size=16,
    num_workers=32,
    confidence_thresh=0.1,
    skip_failures=False,
)
end = time_ns()

print(f"Apply Model Time: {(end - start) / 1e9} seconds") # Apply Model Time: 147.059405114 seconds
```

Screenshot 2025-08-25 at 7 10 53 PM

all remaining time in `predict_all` that isn't forward pass is moving about tensors from cpu to gpu or back.
Screenshot 2025-08-25 at 7 20 01 PM

I guess if anything this is where multithreading would shine. Not 100% sure how to integrate it here though...

## Summary by CodeRabbit

- New Features
  - Parallelized image model inference and label writing for faster processing on large datasets
  - Automatic initialization of the distributed runtime when using parallel processing
- Refactor
  - Improved propagation of sample IDs through data pipelines to correctly associate predictions
  - Enhanced handling of model outputs without an output processor, including CPU-safe tensors and image size returns
  - More flexible batching in transformer pipelines, allowing non-tensor fields to pass through without padding

### DIFF
--- a/fiftyone/core/ray/base.py
+++ b/fiftyone/core/ray/base.py
@@ -1,0 +1,71 @@
+import ray
+
+import fiftyone.core.view as fov
+
+
+def serialize_samples(samples):
+    dataset_name = samples._root_dataset.name
+    stages = (
+        samples._serialize() if isinstance(samples, fov.DatasetView) else None
+    )
+    return dataset_name, stages
+
+
+def deserialize_samples(serialized_samples):
+    import fiftyone as fo
+
+    dataset_name, stages = serialized_samples
+
+    dataset = fo.load_dataset(dataset_name)
+    if stages is not None:
+        return fov.DatasetView._build(dataset, stages)
+    return dataset
+
+
+@ray.remote()
+class FiftyOneActor:
+    """Class for FiftyOne Ray actors.
+
+    Args:
+        serialized_samples: a serialized representation of a
+            :class:`fiftyone.core.collections.SampleCollection`
+    """
+
+    def __init__(self, serialized_samples, **kwargs):
+        super().__init__(**kwargs)
+        self.samples = deserialize_samples(serialized_samples)
+
+
+class ActorPoolContext:
+    """Context manager for a pool of Ray actors.
+
+    Args:
+        samples: a :class:`fiftyone.core.collections.SampleCollection`
+        actor_type: the :class:`FiftyOneActor` subclass to instantiate
+            for each worker
+        num_workers (int): the number of workers in the pool
+    """
+
+    def __init__(self, samples, actor_type, num_workers=4):
+        super().__init__()
+        self.serialized_samples_ref = ray.put(serialize_samples(samples))
+        self.num_workers = num_workers
+        self.actor_type = actor_type
+
+    def __enter__(self):
+        self.actors = [
+            self.actor_type.remote(self.serialized_samples_ref)
+            for _ in range(self.num_workers)
+        ]
+        self.pool = ray.util.ActorPool(self.actors)
+        return self
+
+    def __exit__(self, *args):
+        # Clean up refs
+        for actor in self.actors:
+            del actor
+
+        del self.serialized_samples_ref
+
+    def submit(self, ids, payloads):
+        self.pool.submit(lambda a, v: a.run.remote(*v), (ids, payloads))
--- a/fiftyone/core/ray/base.py
+++ b/fiftyone/core/ray/base.py
@@ -22,7 +22,6 @@ def deserialize_samples(serialized_samples):
     return dataset
 
 
-@ray.remote()
 class FiftyOneActor:
     """Class for FiftyOne Ray actors.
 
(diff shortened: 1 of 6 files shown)


fiftyone/core/ray/base.py
```diff
@@ -0,0 +1,72 @@
+import ray
+
+import fiftyone.core.view as fov
+
+
+def serialize_samples(samples):
```
kaixi-wang, line 6: This is not a good pattern because you're guaranteeing that the process will require a database connection just to resolve the file path. It would be much better to just resolve the file path directly

fiftyone/core/ray/writers.py
```diff
@@ -0,0 +1,40 @@
...
+
+        if self.post_processor is not None:
+            payloads = self.post_processor(
+                *payloads, confidence_thresh=self.confidence_thresh
+            )
+
+        with self.ctx:
+            for sample, payload in zip(samples_batch, payloads):
+                sample.add_labels(
+                    payload,
```
kaixi-wang, line 36: This is what actually needs to be refactored.

kaixi-wang (changes requested):
This adds Ray as a dependency and introduces the complexity of relying on another opinionated library but doesn't solve the root of the issue.

exupero:
Not being familiar with Ray, I have a couple high-level questions:
1. Is this using a background thread or process to offload work? (Or does Ray decide?)
2. If using a background process, does this get around the problem of SampleView and SaveContext not being picklable?

fiftyone/core/ray/base.py
jacobsela: What file paths are we talking about?

Generally speaking, we want these workers to have a database connection. One of the goals is for them to interact with FO datasets in parallel to the main process.

fiftyone/core/ray/writers.py
jacobsela: Any optimizations to the base code are welcome. That said, notice that we are offloading multiple things here from the main process, namely:
1. model output post processing
2. `add_labels`
3. writing to mongo (with `ctx.save()`)

all of these parts have some impact on final `apply_model` time.

jacobsela:
@kaixi-wang 

Re:

> This adds Ray as a dependency and introduces the complexity of relying on another opinionated library but doesn't solve the root of the issue.

There are two parts here:
*(1) the underlying issue*
The core problem this PR tries to address is offloading cpu-bound and I/O-bound operations from the main process, which we want to be only gpu-bound. It seems to do so, as indicated by the profiling attached. I am assuming that by underlying issue you mean the inherent performance of things like `add_labels`. Even if `add_labels` was optimized to take 0 time, there are other parts of the code that are a GPU utilization bottlenecks. An approach like this allows us to more easily offload _whatever_ we view as the bottleneck from any process. More generally, local optimizations can be really impactful for frequently called, especially slow operations. However, if we are interested in longer term scaling, our main concern should be dividing the code to pieces that we can individually scale according to a workload's needs.

This leads well to the second point, the use of ray (which is built to do exactly this).

*(2) adding ray as a dependency*
Yes, this is a decision we will have to make. Ray is just a means to an end here. Multiprocessing is a useful tool that can be helpful throughout our codebase. By introducing something like Ray, I hope to cut out a lot of development time that we would otherwise spend on making multiprocessing use in FiftyOne consistent, trusted, clean, and efficient. What are strong arguments for *not* using ray from your point of view? Let's discuss them and come to a decision.

fiftyone/utils/torch.py
```diff
@@ -921,9 +922,9 @@ def _predict_all(self, imgs):
 
         if self._output_processor is None:
             if isinstance(output, torch.Tensor):
-                output = output.detach().cpu().numpy()
+                output = output.detach().cpu()
```
jacobsela, line 925: notice this

fiftyone/core/ray/base.py
kaixi-wang: No you don't want to give workers database connections. There is zero benefit because the data (media) is not even in the database and cannot be retrieved using the database connection

fiftyone/utils/transformers.py
```diff
@@ -681,7 +681,17 @@ def _predict_all(self, args):
             )
 
         else:
-            return output
+            for k, v in output.items():
```
jacobsela, line 684: this

fiftyone/core/ray/base.py
jacobsela: These are writer workers. They must each hold some connection or share access to a pool of connections for us to write multiple things in parallel. Frankly we don't even need a ton of them because most of what they do is sit around waiting for I/O, so a single one that's multithreaded would probably be just fine.

Unrelated grievances with our multi-worker read system can be discussed elsewhere.

PR #6307, opened 2025-08-28
[DON'T MERGE] FOEPD-1977: Move label post-processing to background thread

exupero:

This PR updates `_apply_image_model_data_loader` to perform label post-processing on the same background thread as DB writes. To do so, it splits `_predict_all`'s logic into two sub-functions and makes them accessible via the functions `preprocess_and_forward_pass` and `postprocess` (better names welcome), then calls the latter on the background thread.

Ran following script on voxelcompute instance and profiled with the built-in cProfile module.
```python
import sys

import fiftyone as fo
import fiftyone.zoo as foz

if __name__ == "__main__":
    suffix = sys.argv[1] if len(sys.argv) > 1 else "baseline"
    if not dataset.persistent:
        dataset.persistent = True
    dataset.apply_model(model, num_workers=32, batch_size=16, confidence_thresh=0.1, label_field=f"profile_{suffix}", skip_failures=False)
```

With these changes, total running time of the script improved from 250s to 200s, almost 90% of which is spent in `_forward_pass`:
postprocessing

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- New Features
  - Added support for post-processing model outputs into final labels, enabling more accurate and consistent results.
  - Post-processing now leverages image size metadata when available.
  - Confidence thresholds are applied during post-processing for cleaner predictions.

- Refactor
  - Shifted to a two-stage prediction pipeline (raw output + post-processing) for both batch and asynchronous paths, improving reliability and consistency without changing public APIs.

### DIFF
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -486,8 +486,9 @@ def _apply_image_model_data_loader(
             )
         )
 
-        def save_batch(sample_batch, labels_batch):
+        def save_batch(sample_batch, model, output, image_sizes):
             with _handle_batch_error(skip_failures, sample_batch):
+                labels_batch = model.postprocess(output, image_sizes)
                 for sample, labels in zip(sample_batch, labels_batch):
                     if filename_maker is not None:
                         _export_arrays(labels, sample.filepath, filename_maker)
@@ -507,14 +508,15 @@ def _apply_image_model_data_loader(
                 if isinstance(imgs, Exception):
                     raise imgs
 
+                image_sizes = imgs.pop("fo_image_size", [(None, None)])
                 if needs_samples:
-                    labels_batch = model.predict_all(
+                    output = model.preprocess_and_forward_pass(
                         imgs, samples=sample_batch
                     )
                 else:
-                    labels_batch = model.predict_all(imgs)
+                    output = model.preprocess_and_forward_pass(imgs)
 
-                submit(save_batch, sample_batch, labels_batch)
+                submit(save_batch, sample_batch, model, output, image_sizes)
 
             pb.update(len(sample_batch))
 
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -485,8 +485,15 @@ def _apply_image_model_data_loader(
                 warning="Async failure labeling batches",
             )
         )
+        output_processor = model._output_processor
+        context.enter_context(fou.SetAttributes(model, _output_processor=None))
 
-        def save_batch(sample_batch, labels_batch):
+        def save_batch(sample_batch, output, image_sizes):
+            labels_batch = output_processor(
+                output,
+                image_sizes,
+                confidence_thresh=model.config.confidence_thresh,
+            )
             with _handle_batch_error(skip_failures, sample_batch):
                 for sample, labels in zip(sample_batch, labels_batch):
                     if filename_maker is not None:
@@ -507,14 +514,15 @@ def _apply_image_model_data_loader(
                 if isinstance(imgs, Exception):
                     raise imgs
 
+                image_sizes = imgs.pop("fo_image_size", [(None, None)])
                 if needs_samples:
-                    labels_batch = model.predict_all(
-                        imgs, samples=sample_batch
-                    )
+                    output = model.predict_all(imgs, samples=sample_batch)
                 else:
-                    labels_batch = model.predict_all(imgs)
+                    output = model.predict_all(imgs)
 
-                submit(save_batch, sample_batch, labels_batch)
+                submit(
+                    save_batch, sample_batch, output, image_sizes=image_sizes
+                )
 
             pb.update(len(sample_batch))
 
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -485,7 +485,9 @@ def _apply_image_model_data_loader(
                 warning="Async failure labeling batches",
             )
         )
-        output_processor = model._output_processor
+        output_processor = getattr(
+            model, "_output_processor", lambda output, *_, **__: output
+        )
         context.enter_context(fou.SetAttributes(model, _output_processor=None))
 
         def save_batch(sample_batch, output, image_sizes):
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -486,9 +486,8 @@ def _apply_image_model_data_loader(
             )
         )
 
-        def save_batch(sample_batch, model, output, image_sizes):
+        def save_batch(sample_batch, labels_batch):
             with _handle_batch_error(skip_failures, sample_batch):
-                labels_batch = model.postprocess(output, image_sizes)
                 for sample, labels in zip(sample_batch, labels_batch):
                     if filename_maker is not None:
                         _export_arrays(labels, sample.filepath, filename_maker)
@@ -508,15 +507,14 @@ def _apply_image_model_data_loader(
                 if isinstance(imgs, Exception):
                     raise imgs
 
-                image_sizes = imgs.pop("fo_image_size", [(None, None)])
                 if needs_samples:
-                    output = model.preprocess_and_forward_pass(
+                    labels_batch = model.predict_all(
                         imgs, samples=sample_batch
                     )
                 else:
-                    output = model.preprocess_and_forward_pass(imgs)
+                    labels_batch = model.predict_all(imgs)
 
-                submit(save_batch, sample_batch, model, output, image_sizes)
+                submit(save_batch, sample_batch, labels_batch)
 
             pb.update(len(sample_batch))
 
--- a/fiftyone/utils/transformers.py
+++ b/fiftyone/utils/transformers.py
@@ -654,31 +654,34 @@ class FiftyOneTransformer(TransformerEmbeddingsMixin, fout.TorchImageModel):
                     for _k, _v in self._last_output[k].items()
                 }
 
-    def _preprocess_and_forward_pass(self, args):
+    def _predict_all(self, args):
         if self.preprocess:
             args = self.collate_fn(self.transforms(args))
+
+        # this line is the only difference between this and the base class
+        # we should consolidate this function once post processing is properly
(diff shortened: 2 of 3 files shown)


jacobsela:
I was thinking about the numbers for a bit and I think most of the time save in this PR is coming from saving on I/O time between the CPU and GPU.

When comparing with the [mock PR](https://github.com/voxel51/fiftyone/pull/6293) I put up a few days ago, the diff in computation is as follows:
1. post processing is threaded in this PR rather than on a different process.
2. however the GPU-CPU I/O is on the main process in the other PR, while it's on a different thread in this one.

Those two points combined with the fact that the GPU/CPU I/O in the mock PR was ~50s, roughly the amount of time saved in this PR, lead me to believe that the above assertion is correct.

Now, I didn't offload the I/O in the other PR because I was getting a GPU memory leak (that I didn't want to deal with for the mockup) when doing the GPU-CPU I/O call in the other process, see the change in [this line](https://github.com/voxel51/fiftyone/pull/6293/files#r2308942856).

However what this leads me to believe is that we could get the combined effects of both PRs by having the "actual" post processing be on another process while having the I/O be threaded. Assuming we can stack the effects, we would be looking at ~100s for this code (going off the mock PR profiling).

fiftyone/core/models.py
```diff
@@ -486,8 +486,9 @@ def _apply_image_model_data_loader(
             )
         )
 
-        def save_batch(sample_batch, labels_batch):
+        def save_batch(sample_batch, model, output, image_sizes):
             with _handle_batch_error(skip_failures, sample_batch):
+                labels_batch = model.postprocess(output, image_sizes)
```
brimoor: Unless I'm missing something, this will cause `apply_model()` to start failing for all existing `TorchImageModel` instances besides `FiftyOneTransformer` which you've specifically implemented `postprocess()` for. We can't do that.

For context, the model zoo contains lots of instances of `TorchvisionImageModel(TorchImageModel)`. [EG here](https://github.com/voxel51/fiftyone/blob/2853b72cb393c1b08d1f14bcfd06da5ee7d3ae50/fiftyone/zoo/models/manifest-torch.json#L19)

fiftyone/core/models.py
```diff
@@ -507,14 +508,15 @@ def save_batch(sample_batch, labels_batch):
                 if isinstance(imgs, Exception):
                     raise imgs
 
+                image_sizes = imgs.pop("fo_image_size", [(None, None)])
```
brimoor, line 519: `fo_image_size` is unique to `FiftyOneTransformer` so it can't be referenced in `fiftyone.core.models`.

fiftyone/core/models.py
```diff
@@ -507,14 +508,15 @@ def save_batch(sample_batch, labels_batch):
...
                 if needs_samples:
-                    labels_batch = model.predict_all(
+                    output = model.preprocess_and_forward_pass(
```
brimoor: As @jacobsela pointed out, this is a sensitive change. The `Model` interface is a public-facing thing that was created so that users can write their own `Model` instances that implement our interface so that `apply_model()` can support them:
https://docs.voxel51.com/model_zoo/design.html

brimoor (changes requested):
If we wanted to provide a new syntax where postprocessing is not performed in the main process, then we'd need to support it in a backwards-compatible way in the `Model` class that would allow `apply_model()` to continue working for all existing `Model` instances.

I believe what you want is a similar idea to the existing `Model.preprocess == False`, which tells `Model.transforms` to be applied by the data loader and **not** in `predict()`/`predict_all()`.

For example, we could have `Model.postprocess == None` by default, but if a `Model` provides a non-None `Model.postprocess`, then `predict()`/`predict_all()` is assumed to provide a "raw" output that must be passed to `Model.postprocess()`. And then that can be called in a separate worker inside `apply_model()`.

exupero:
I should have noted that this is a proof of concept needing feedback for the proper way to apply it, so thanks for providing that feedback without my asking 💯.

@jacobsela can you provide benchmarks for reloading samples on a separate process? I spiked a branch with that mimicked the serialization and deserialization logic in your #6293, but having to reload samples from the DB appeared to be significantly slower than sharing memory between threads (slower as in my test script appeared to hang but was actually just going through background tasks at 15-20s per task). I might have missed something in your proof of concept, so I can share my spike code if you're interested.

My expectation is that the speedup in this branch is due to unblocking the GPU. Currently `predict_all` blocks on GPU-bound execution, then blocks on CPU-bound execution, but those can be done concurrently, so moving CPU operations to a background thread allows the next batch of GPU operations to begin while the CPU is still transforming tensors into FO detection objects.

Does Python cProfile data capture GPU-CPU I/O? If so, what should I look for? Or are you using a different tool to benchmark GPU-CPU I/O?

fiftyone/utils/transformers.py
```diff
@@ -654,35 +654,32 @@ def last_output(self, value):
...
         if self.preprocess:
             args = self.collate_fn(self.transforms(args))
 
-        # this line is the only difference between this and the base class
-        # we should consolidate this function once post processing is properly
-        # removed out of torchimagemodel
-        image_sizes = args.pop("fo_image_size", [(None, None)])
-
         for k, v in args.items():
             args[k] = v.to(self.device)
```
jacobsela: CPU to GPU copy

kaixi-wang:
> bothers @kaixi-wang about this approach. As far as I understand the concern stems from the repeated COLLSCANs that can be potentially slow, especially at the bottom of the collection. I'm not sure that in aggregate this is a problem just yet, but I have some ideas for things we can do to alleviate some of the problem if it turns out that it is

The issues are not just the COLLSCAN alone. it's the unnecessary database lookups to resolve samples when what you actually only need is filepath and storage client to resolve media.

jacobsela:
> The issues are not just the COLLSCAN alone. it's the unnecessary database lookups to resolve samples when what you actually only need is filepath and storage client to resolve media. It's overcomplicating everything in a way that impacts the ability to parallelize work

Resolving filepaths and media is for loading and preprocessing. It's unrelated to this PR. Please limit discussion to the topic at hand.

jacobsela:
> Here's a less intrusive take that mirrors the technique in #6293 of manipulating the model's `_output_processor` attribute to prevent post-prediction processing.

Ok this is a way better starting point. To address the points:

> It works, but I don't see that attribute as part of the model interface, so I'm assuming model classes aren't guaranteed to have it and this approach is a non-starter.

Yeah, this isn't guaranteed across the board currently. It is a part of the `TorchImageModel` interface, see code [here](https://github.com/voxel51/fiftyone/blob/30a246f929a651cea053b9941205f55128a21099/fiftyone/utils/torch.py#L427). Notice that the `apply_model` "sub function" you are working with (`_apply_image_model_data_loader`) only gets called for torch models or for `SupportsGetItem` (see the check [here](https://github.com/voxel51/fiftyone/blob/30a246f929a651cea053b9941205f55128a21099/fiftyone/core/models.py#L196)).

> Is there a more universal way to exclude post-processing from predict_all, or would doing that necessitate changing the model contract?

First, note that any change we would do (even if limited to `TorchImageModel`s and `_apply_image_model_data_loader` would require some interface change (even if only to those classes), see @brimoor's [point](https://github.com/voxel51/fiftyone/pull/6307/files#r2309121655).

I strongly recommend implementing the generic `TorchModel` interface for this change (again see the "arbitrary model input" doc shared with you). This is because the alternative (written below) is not significantly less work.

At the very least, `TorchImageModel`'s `_predict_all` needs to be changed so that if post processing happens outside of it, it delivers the correct payload. This would have cascading effects `TorchImageModel` subclasses, and the change would likely also have to be reflected there.

fiftyone/core/models.py
```diff
@@ -485,8 +485,17 @@ def _apply_image_model_data_loader(
...
             )
         )
+        output_processor = getattr(
+            model, "_output_processor", lambda output, *_, **__: output
+        )
+        context.enter_context(fou.SetAttributes(model, _output_processor=None))
 
-        def save_batch(sample_batch, labels_batch):
+        def save_batch(sample_batch, output, image_sizes):
+            labels_batch = output_processor(
```
kaixi-wang, line 494: is it actually more efficient to do this in a thread? How do the size of the output vectors compare to the final predicted objects?

exupero: It's not more efficient, it's more concurrent. Moving this to a background thread allows the output to be processed while the next batch's inputs are being processed and run. I.e.

(not to scale)
```mermaid
gantt
	title Async postprocessing
	dateFormat X
	axisFormat %s
	
	section Main thread
	Preprocess :active, preprocess, 0, 1
	Predict :active, predict, 1, 2
	Preprocess :active, preprocess, 2, 3
	Predict :active, predict, 3, 4
	
	section Background thread
	Postprocess :active, postprocess, 2, 3
	Write :active, write, 3, 4
	Postprocess :active, postprocess, 4, 5
	Write :active, write, 5, 6
```

With the test script, `output` is 4-5x larger (in bytes) than `labels_batch`, but since this is a thread, memory is shared, so it it shouldn't incur any greater memory usage than the previous version.

exupero:
Using the test script in the PR description, this flamegraph against develop shows more than half of `_predict_all`'s time is spent in the output processor (121s out of 219s):
Screenshot 2025-09-03 at 10 29 20 AM
However, the overall improvement of moving post-processing to the background thread only seems to be about 20% (note the green line, where run time reduced from 250s to 200s):
postprocessing
The flamegraph against this branch shows `_forward_pass` takes ~40% longer:
Screenshot 2025-09-03 at 10 39 15 AM
I expect the longer execution time is due to the GIL contention, so while moving post-processing to a background thread does make an overall improvement, further gains should be possible by moving to a background process (though that work is left for a separate PR).

exupero:
I've created [FOEPD-2017 Add post-processing hook to Model interface](https://voxel51.atlassian.net/browse/FOEPD-2017). Closing this until that work is complete.

PR #6331, opened 2025-09-09
Bug fix: `to_torch()` on empty collections

brimoor:
## Change log

- fixes a bug where `apply_model()` and `compute_embeddings()` would raise an error when applied to empty collections. now they will successfully complete as no-ops

## Tested by

```py
import fiftyone as fo
import fiftyone.zoo as foz

view = dataset.limit(0)

# these previously raised errors; now they succeed as no-ops
view.apply_model(model, label_field="clip_predictions")
view.compute_embeddings(model, embeddings_field="clip_embeddings")
```

### DIFF
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -2524,7 +2524,11 @@ class NumpySerializedList:
         self._lst = [_serialize(x) for x in lst]
         self._addr = np.asarray([len(x) for x in self._lst], dtype=np.int64)
         self._addr = np.cumsum(self._addr)
-        self._lst = np.concatenate(self._lst)
+        self._lst = (
+            np.concatenate(self._lst)
+            if self._lst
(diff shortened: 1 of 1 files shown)


brimoor:
For example, I discovered this bug because I built a workflow where I wanted to fill in missing predictions on a dataset, ie run inference on **only** samples that aren't already labeled.

From a clean code standpoint, I want to be able to express that like this:

````py
dataset.exists(label_field, bool=False).apply_model(..., label_field)
```

where, if the full dataset is already labeled, then this command will be a no-op, but I don't have to write any custom logic to handle that. The core library handles it more me.

As a general rule, for public methods on the core interface, I feel they should complete successfully in as many situations as possible, for maximum flexibility. If "the view is empty" is not allowed in a particular use case, then any error handling should be localized to that workflow's implementation.

PR #6334, opened 2025-09-10
Remove serverside JS usages

swheaton:
[$function](https://www.mongodb.com/docs/manual/reference/operator/aggregation/function/) is deprecated. It's also slow and bad security (which is why it's deprecated).

1. Translate `ViewExpression.sort` to mongo native `$sortArray`. This was added in 5.2 which wouldn't have been OK originally but is fine now since we require >=6.0.
2. In server count list items, replace the JS function with a fairly complicated mongo expression that also makes a histogram of values. Unfortunately it is not nearly as succinct as the JS function, but it will probably be faster since it's mongo native

NOTE: `ViewExpression.sort` docstring examples do not work in any configuration I've tried, with code going back 4+ years. I've tried mongodb versions back to 4.2. Both local (Mac) and some deployed (Linux). Either the docstrings have been totally wrong forever, the function has been broken forever, or I'm doing something silly. Would love to know in any case.

App runs if you disable serverside JS 🤷🏼
Gemini says they're logically equivalent 🤷🏼 🤖 
Letting unit tests run to see results of that.
I'm open to more tests to run? I don't really know what the server function populates (2)

        notes for FiftyOne users.

Remove usage of deprecated $function MongoDB stage

-   [ ] Other

## Summary by CodeRabbit

* **Refactor**
  * Array sorting now uses database-side sorting for deterministic results; the previous client-side function-based sorter was removed. Reversing arrays remains supported.
  * List-item counting now uses a database aggregation to produce item-count maps for lists, deriving counts from centralized label data for improved consistency.

* **Chores**
  * Updated internal dependencies to support the new aggregation-based approaches.

* **Documentation**
  * Updated docs to reflect the sorting removal and new counting behavior.

### DIFF
--- a/fiftyone/core/expressions.py
+++ b/fiftyone/core/expressions.py
@@ -2663,6 +2663,82 @@ class ViewExpression(object):
         """
         return ViewExpression({"$reverseArray": self})
 
+    def sort(self, key=None, numeric=False, reverse=False):
+        """Sorts this expression, which must resolve to an array.
+
+        If no ``key`` is provided, this array must contain elements whose
+        BSON representation can be sorted by JavaScript's ``.sort()`` method.
+
+        If a ``key`` is provided, the array must contain documents, which are
+        sorted by ``key``, which must be a field or embedded field.
+
+        Examples::
+
+            #
+            # Sort the tags of each sample in a dataset
+            #
+
+            import fiftyone as fo
+            from fiftyone import ViewField as F
+
+            dataset = fo.Dataset()
+            dataset.add_samples(
+                [
+                    fo.Sample(filepath="im1.jpg", tags=["z", "f", "p", "a"]),
+                    fo.Sample(filepath="im2.jpg", tags=["y", "q", "h", "d"]),
+                    fo.Sample(filepath="im3.jpg", tags=["w", "c", "v", "l"]),
+                ]
+            )
+
+            # Sort the `tags` of each sample
+            view = dataset.set_field("tags", F("tags").sort())
+
+            print(view.first().tags)
+
+            #
+            # Sort the predictions in each sample of a dataset by `confidence`
+            #
+
+            import fiftyone as fo
+            import fiftyone.zoo as foz
+            from fiftyone import ViewField as F
+
+            dataset = foz.load_zoo_dataset("quickstart")
+
+            view = dataset.set_field(
+                "predictions.detections",
+                F("detections").sort(key="confidence", numeric=True, reverse=True)
+            )
+
+            sample = view.first()
+            print(sample.predictions.detections[0].confidence)
+            print(sample.predictions.detections[-1].confidence)
+
+        Args:
+            key (None): an optional field or ``embedded.field.name`` to sort by
+            numeric (False): whether the array contains numeric values. By
+                default, the values will be sorted alphabetically by their
+                string representations
+            reverse (False): whether to sort in descending order
+
+        Returns:
+            a :class:`ViewExpression`
+        """
+        sort_order = pymongo.DESCENDING if reverse else pymongo.ASCENDING
+
+        if key is not None:
+            sort_by = {key: sort_order}
+        else:
+            sort_by = sort_order
+
+        sort_stage = {
+            "$sortArray": {
+                "input": self,
+                "sortBy": sort_by,
+            }
+        }
+        return ViewExpression(sort_stage)
+
     def filter(self, expr):
         """Applies the given filter to the elements of this expression, which
         must resolve to an array.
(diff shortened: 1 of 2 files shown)


kaixi-wang (commented):
doesn't seem like it gives the correct results. 
added 8 label tags 'test' to sample then ran
```

view= _add_labels_tags_counts(ds)
pipeline2=view._pipeline()
pipeline2.extend([{
    "$match": {
        "_label_tags": {"$ne": {}}
    }
},  {"$project":{"_label_tags":1}}])

# proposed
In [37]: list(ds._sample_collection.aggregate(pipeline))
Out[37]: [{'_id': ObjectId('6829068fb59076c98c88bdcc'), '_label_tags': {'test': 1}}]

# existing
list(ds._sample_collection.aggregate(pipeline2))
Out[34]: [{'_id': ObjectId('6829068fb59076c98c88bdcc'), '_label_tags': {'test': 8.0}}]

**Comment by swheaton** [comment]
Fixed. The only difference now is that the count is an integer instead of a float, idk if that matters

File: fiftyone/core/expressions.py [hunk 1]
```diff
@@ -2723,33 +2724,20 @@ def sort(self, key=None, numeric=False, reverse=False):
...
+            sort_by = sort_order
 
-        return self._function(sort_fcn)
+        sort_stage = {
+            "$sortArray": {
+                "input": self,
+                "sortBy": sort_by,
+            }
+        }
+        return ViewExpression(sort_stage)
```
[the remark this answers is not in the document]
swheaton: FiftyOne supports >=6.0
https://docs.voxel51.com/user_guide/config.html#using-a-different-mongodb-version

swheaton:
idk why tests would pass in ubuntu but fail in windows. those tests shouldn't have anything to do with the OS? Unless their MongoDB installations differ in behavior which would be wild?

swheaton:
Ok so confusingly this is because windows downloads v6 while ubuntu downloads v8 of mongod. So it's a version thing not a platform thing

fiftyone/core/expressions.py
```diff
@@ -2723,33 +2724,20 @@ def sort(self, key=None, numeric=False, reverse=False):
...
-                comp = "(a, b) => a.{key} - b.{key}"
-            else:
-                comp = "(a, b) => ('' + a.{key}).localeCompare(b.{key})"
-
-            comp = comp.format(key=key)
-        elif numeric:
-            comp = "(a, b) => a - b"
-        else:
-            comp = ""
+        sort_order = pymongo.DESCENDING if reverse else pymongo.ASCENDING
```
brimoor, line 2727: A couple comments on the implementation:

- The `numeric` parameter is no longer used. Assuming that the new implementation sorts strings alphabetically and numbers numerically by default, can you delete the `numeric` parameter?
- The docstring currently mentions "JavaScript's `.sort()` method". Can you update to say something like this?

```
If no ``key`` is provided, this array must contain sortable non-document values.
```

fiftyone/server/view.py
```diff
@@ -947,19 +947,43 @@ def _add_label_tags(path, field, view):
...
-        "counts[i] = 1 + (counts[i] || 0);"
-        "});"
-        "return counts;"
-        "}"
-    )
-
-    return view.set_field(
-        path, F(path)._function(function), _allow_missing=True
-    )
+    expr = {
```
brimoor, line 950: Defer to @benjaminpkane to evaluate the suitability of this new implementation

brimoor (commented):
FWIW- the example code for `sort()` still runs successfully for me on `develop`. I'm running MongoDB 5.0.2 on an x86 mac.

I'm onboard with swapping out `$function` for `$sortArray` though. Although looks like I'll have to upgrade to at least Mongo 5.2 if I want to use this function going forward 😝

fiftyone/server/view.py
swheaton: yeah it's not working in mongo6. Solution is getting really ugly ...

fiftyone/core/expressions.py
swheaton: Alright will clean up if we actually want to go forward with this. Sort of a prototype at the moment

fiftyone/server/view.py
benjaminpkane: I think changing this to existence instead of counts could work. These counts are for the `tag:count` grid bubble values but showing the distinct tags is sufficient imo if we are technically blocked (also a performance improvement for the grid, less computation)

swheaton: I got a working version now I think. Check it out and lmk.
Of course if you want to change the grid behavior that would work too but that's up to you + your squad

PR #6341, opened 2025-09-12
FOEPD-2039 OSS changes to support DO pipelines

swheaton:

Add new pipeline option to queuing operations.
Add `PipelineOperator` which has a `resolve_pipeline()` function. When it's called, it returns a list of `PipelineStage` objects.

Note this is likely not its final form. There is an idea to use `types` and a similar pattern to inputs

Added some automated tests.
But mostly it's tested over in enterprise until I can implement the immediate execution of pipelines

## Summary by CodeRabbit

- New Features
  - Pipeline abstraction added (PipelineOperator, PipelineStage) and support for attaching pipelines to delegated operations; queued operations now can include pipeline details.

- Improvements
  - Safer serialization/deserialization and more robust ID handling for delegated operations; pipeline types exported for consumers.

- Behavior Changes
  - Runtime now raises explicit errors for immediate or distributed pipeline execution in non-enterprise environments.

- Tests
  - Unit tests added for pipeline serialization and delegation handling.

### DIFF
--- a/fiftyone/factory/repos/delegated_operation_doc.py
+++ b/fiftyone/factory/repos/delegated_operation_doc.py
@@ -5,7 +5,8 @@ FiftyOne delegated operation repository document.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
-
+import copy
+import dataclasses
 import logging
 from datetime import datetime
 
@@ -15,6 +16,7 @@ from fiftyone.operators.executor import (
     ExecutionRunState,
     ExecutionProgress,
 )
+from fiftyone.operators.pipeline import PipelineStage
 
 logger = logging.getLogger(__name__)
 
@@ -62,6 +64,8 @@ class DelegatedOperationDocument(object):
 
         # distributed task fields
         self.parent_id = None  # Only on children
+        self.pipeline_index = 0
+        self.pipeline = None  # Only on pipeline parent
 
     @property
     def num_distributed_tasks(self):
@@ -95,7 +99,7 @@ class DelegatedOperationDocument(object):
         self.parent_id = doc.get("parent_id", None)
 
         # internal fields
-        self.id = doc["_id"]
+        self.id = doc.get("_id", doc.get("id"))
         self._doc = doc
 
         # nested fields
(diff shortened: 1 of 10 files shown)


fiftyone/factory/repos/delegated_operation_doc.py
```diff
@@ -62,6 +64,8 @@ def __init__(
 
         # distributed task fields
         self.parent_id = None  # Only on children
+        self.pipeline_index = 0
```
kaixi-wang, line 67: do you really need 2 variables? Can't self.pipeline equal the index if its a pipeline or just none if not a pipeline

fiftyone/factory/repos/delegated_operation_doc.py
```diff
@@ -95,7 +99,7 @@ def from_pymongo(self, doc: dict):
         self.parent_id = doc.get("parent_id", None)
 
         # internal fields
-        self.id = doc["_id"]
+        self.id = doc.get("_id", doc.get("id"))
```
kaixi-wang, line 102: since _id is usually ObjectId and id is a string, are we saying self.id is either?

fiftyone/factory/repos/delegated_operation_doc.py
```diff
@@ -127,14 +131,24 @@ def from_pymongo(self, doc: dict):
             if "updated_at" in doc["status"]:
                 self.status.updated_at = doc["status"]["updated_at"]
 
+        if "pipeline" in doc and doc["pipeline"] is not None:
```
kaixi-wang, line 134: since you just raise an error later, should you even be setting it?

kaixi-wang (commented):
is there communication between the stages/passing of the same context? I'm somewhat confused, but I don't know if that's just because most of the implementation is not in oss?

fiftyone/factory/repos/delegated_operation_doc.py
swheaton: pipeline index is the current stage of operation that it's on.
pipeline is the definition of the stages

fiftyone/factory/repos/delegated_operation_doc.py
swheaton: yeah uhh idk. We should cast to `str` I guess

fiftyone/factory/repos/delegated_operation_doc.py
swheaton: Attempt to minimize oss-teams delta

CamronStaley (approved):
Seems okay will probably make more sense once I see how pipeline and pipeline_index are used.

PR #6343, opened 2025-09-12
make num writer threads configurable

jacobsela:

Make num writer threads used in `apply_model` and `compute_embeddings` configurable. This has a pretty big effect on runtime (>2.5x in reasonable cases).

Also something weird may be happening with memory usage.

Code and data will be shared privately, I don't want to dump it here.

### Performance impact of num threads on I/O heavy jobs

**Task**: Computing dense embeddings with DINO v3
**Variables**:
1. num threads
2. model size
3. embedding size (controlled by image size during inference)

**`compute_embeddings` benchmark time vs num threads**
image

**analysis**
Without a large enough number of threads, the writing jobs just continue getting queued and pile up. As we exit the async execution, we have to do a blocking wait for them to all finish to complete the `compute_embeddings` operation. This effectively means that rather than saving time in `compute_embeddings`, we are just deferring the wait for I/O till later.

By increasing the number of writer threads, we can avoid this issue. Interestingly, this effect seems to be more pronounced for larger image sizes  and model sizes.

I'm not sure why the speed flatlines so quickly with num threads. The writing still continues after the GPU workload ends even for 16 threads. We may be hitting the limits of mongo throughput. Something to consider for an embeddings rework @brimoor .

### Memory consumption vs num samples

**Task**: Computing dense embeddings with DINO v3
**Variables**:
1. num threads
2. num samples

**`compute_embeddings` memory usage over time**
image
image

**analysis**
As the writing jobs pile up, we use more and more memory. It is critical that tasks like this don't have memory requirements that scale this quickly with the number of samples. In this case some back pressure and garbage collection will likely do.

Adding more threads should in theory help, because we would clear the job queue faster, thus not allowing it to pile up in the first place. Given that the write throughput scaling we get with threads isn't good, it doesn't resolve the problem. Regardless, we should have the guardrails mentioned above.

### action items
We need some limit on the number of embeddings that can be floating in memory at once.
1. add configurable back pressure to `async_executor`
2. figure out if we should manually garbage collect here or if this will happen naturally.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added an optional num_writer_workers setting to control writer-thread concurrency for asynchronous labeling/saving and embedding computation, improving throughput for heavy workloads.
  * Async executor now accepts an explicit max_workers default derived from user config or a CPU-based recommendation when unset.

* **Bug Fixes / Behavior**
  * Thread recommendation now clamps non-positive inputs to at least 1 worker to avoid zero/negative worker counts.

* **Documentation**
  * Docstrings updated to describe the new parameter and defaulting behavior.

### DIFF
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -75,7 +75,7 @@ def apply_model(
     store_logits=False,
     batch_size=None,
     num_workers=None,
-    num_writer_thread_workers=None,
+    num_writer_workers=None,
     skip_failures=True,
     output_dir=None,
     rel_dir=None,
@@ -106,6 +106,8 @@ def apply_model(
             batching
         num_workers (None): the number of workers to use when loading images.
             Only applicable for Torch-based models
+        num_writer_workers (None): the number of thread workers to use when
+            writing predictions. Only applicable for Torch-based models.
         skip_failures (True): whether to gracefully continue without raising an
             error if predictions cannot be generated for a sample. Only
             applicable to :class:`Model` instances
@@ -285,7 +287,7 @@ def apply_model(
                 filename_maker,
                 progress,
                 field_mapping,
-                num_writer_thread_workers,
+                num_writer_workers,
             )
 
         if batch_size is not None:
@@ -459,7 +461,7 @@ def _apply_image_model_data_loader(
     filename_maker,
     progress,
     field_mapping,
-    num_writer_thread_workers,
+    num_writer_workers,
 ):
     needs_samples = isinstance(model, SamplesMixin)
 
@@ -483,7 +485,7 @@ def _apply_image_model_data_loader(
         ctx = context.enter_context(foc.SaveContext(samples))
         submit = context.enter_context(
             fou.async_executor(
-                max_workers=num_writer_thread_workers,
+                max_workers=num_writer_workers,
                 skip_failures=skip_failures,
                 warning="Async failure labeling batches",
             )
@@ -907,7 +909,7 @@ def compute_embeddings(
     embeddings_field=None,
     batch_size=None,
     num_workers=None,
-    num_writer_thread_workers=None,
+    num_writer_workers=None,
     skip_failures=True,
     progress=None,
     **kwargs,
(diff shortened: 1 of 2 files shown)


fiftyone/core/models.py
```diff
@@ -75,6 +75,7 @@ def apply_model(
     store_logits=False,
     batch_size=None,
     num_workers=None,
+    num_writer_thread_workers=None,
```
brimoor: Can we use `num_writer_workers` for generality? In case we were to support configuring to use processes instead of threads in the future, for example.

Also, all kwargs of public methods need to be added to the docstring 😄

fiftyone/core/models.py
```diff
@@ -904,6 +907,7 @@ def compute_embeddings(
     embeddings_field=None,
     batch_size=None,
     num_workers=None,
+    num_writer_thread_workers=None,
```
brimoor: same as above

brimoor (commented):
Thinking out loud: should we document this parameter in the `apply_model()` and `compute_embeddings()` implementations of the main `SampleCollection` interface? Or do we consider this a strictly internal thing that we don't want end users to think about?

Observation 1: this parameter is only applicable when `model` is a Torch model that supports data loaders. Ideally all parameters in the public interface support *all models*, which is one piece of evidence to suggest that this is perhaps not worthy of inclusion in the `SampleCollection` interface.

Observation 2: the `SampleCollection` methods have `**kwargs` in their signatures, so it would still be possible to utilize this functionality, albeit undocumented.

What say you? 🤔

fiftyone/core/utils.py
```diff
@@ -3168,6 +3168,12 @@ def async_executor(
     Raises:
         Exception: if a task raises an exception and ``skip_failures == False``
     """
+    if max_workers is None:
+        max_workers = (
+            fo.config.default_thread_pool_workers
```
kaixi-wang, line 3179: from the data, it looks like there is no benefit after 8 (most the benefits are seen with 4) so it seems like we should cap it rather than use the default (cpu_count()). For high cpu machines, this could have deleterious affects.

fiftyone/core/models.py
jacobsela: changed an added

fiftyone/core/utils.py
jacobsela: The data attached is for computing dense embeddings with a (relatively) large batch size. It's of interest because we will use this workload in det RERs. It's also likely reflective of other very heavy mongo I/O workloads. Conversely, something like a regular `apply_model` on detections with a low batch size will have many small payloads and may benefit from more threads.

Given that the user can already configure the number of threads via any of:
- the argument `num_writer_workers`
- `fo.config.default_thread_pool_workers`
- `fo.config.max_thread_pool_workers`

I would rather not put a hard cap here.

What concerns do you have for very many threads being used? Maybe they can be separately adressed.

PR #6344, opened 2025-09-13
Don't create indexes by default in view stages

brimoor:
## Change log

Updates the default behavior of the `sort_by()`, `group_by()`, `geo_within()`, and `geo_near()` view stages to **NOT** automatically create an index when used.

Index creation can be quite expensive on large datasets, so it was bad manners of us to be silently performing expensive computation without the user's explicit direction, as instantiating view stages is supposed to be fast (excluding *generated views*).

**Note:** this is technically a breaking change as `geo_near()` **requires** an index in order to work. However, I believe a breaking change is worth it to encourage best practices of separating index creation from view stage instantiation. Also, `geo_near()` has limited utility (it can only be the *first* stage in a view).

## Tested by

```py
import random

import fiftyone as fo
import fiftyone.zoo as foz

```

### `sort_by()`

```py
classes = ["cat", "dog", "rabbit"]
predictions = [
    fo.Classification(label=random.choice(classes))
    for _ in range(len(dataset))
]

view = dataset.sort_by("predictions.label")
_ = len(view)

assert "predictions.label" not in dataset.list_indexes()

view = dataset.sort_by("predictions.label", create_index=True)
_ = len(view)

assert "predictions.label" in dataset.list_indexes()
dataset.drop_index("predictions.label")
```

### `group_by()`

```py
view = dataset.group_by("predictions.label")
_ = len(view)

assert "predictions.label" not in dataset.list_indexes()

view = dataset.group_by("predictions.label", create_index=True)
_ = len(view)

assert "predictions.label" in dataset.list_indexes()
dataset.drop_index("predictions.label")
```

### `geo_within()`

```py
MANHATTAN = [
    [
        [-73.949701, 40.834487],
        [-73.896611, 40.815076],
        [-73.998083, 40.696534],
        [-74.031751, 40.715273],
        [-73.949701, 40.834487],
    ]
]

view = dataset.geo_within(MANHATTAN)
_ = len(view)

assert "location.point" not in dataset.list_indexes()

view = dataset.geo_within(MANHATTAN, create_index=True)
_ = len(view)

assert "location.point" in dataset.list_indexes()
dataset.drop_index("location.point")
```

### `geo_near()`

```py
try:
    view = dataset.geo_near([-73.9855, 40.7580])
    _ = len(view)
    assert False
except:
    assert "location.point" not in dataset.list_indexes()

view = dataset.geo_near([-73.9855, 40.7580], create_index=True)
_ = len(view)

assert "location.point" in dataset.list_indexes()
dataset.drop_index("location.point")
```

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -2200,6 +2200,8 @@ class SampleCollection(object):
 
         if has_frame_fields:
             frame_numbers = results.pop(0)
+        else:
+            frame_numbers = None
 
         sample_ids = results[0]
         all_label_ids = results[1:]
@@ -5791,7 +5793,7 @@ class SampleCollection(object):
         min_distance=None,
         max_distance=None,
         query=None,
-        create_index=True,
+        create_index=False,
     ):
         """Sorts the samples in the collection by their proximity to a
         specified geolocation.
@@ -5799,7 +5801,8 @@ class SampleCollection(object):
         .. note::
 
             This stage must be the **first stage** in any
-            :class:`fiftyone.core.view.DatasetView` in which it appears.
+            :class:`fiftyone.core.view.DatasetView` in which it appears, and it
+            **requires** a spherical index on the specified location field.
 
         Examples::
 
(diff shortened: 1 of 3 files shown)


brimoor:
> Lgtm but looks like something tests need to be fixed?

fixed thanks 👍

fiftyone/core/collections.py
```diff
@@ -5814,14 +5817,18 @@ def geo_near(
             # Sort the samples by their proximity to Times Square
             #
 
-            view = dataset.geo_near(TIMES_SQUARE)
+            view = dataset.geo_near(TIMES_SQUARE, create_index=True)
```
kaixi-wang, line 5820: could be nice to have a debug log that a spherical index is being created since it could cause system slowdowns during the build process

kaixi-wang (approved):

PR #6348, opened 2025-09-19
Update installation script

mcdoh:

- convert from Bash script to posix-compliant sh script
- check that Python is within 3.9 - 3.12
  - open3d does not yet support 3.13
- check if nvm is not already installed
- disable confirmation of yarn installation
  - with output piped to /dev/null the user never sees request

- uninstall nvm
- uninstall yarn
- run `./install` on:
  - Zsh on Mac
  - Bash on Mac
  - Zsh on Linux
  - Bash on Linux

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - New POSIX-compatible installer with command-line options, Python 3.9–3.12 gating, optional front-end build, Node/NVM and Yarn/Corepack setup, and choices for installing MongoDB, FiftyOne core, Brain, and ETA (source or packaged).
- **Bug Fixes**
  - Improved option parsing, clearer errors, safer version detection and path handling.
- **Documentation**
  - Mac/Linux install instruction updated to ./install.sh; guidance to export PYTHONPATH after install.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -57,10 +57,10 @@ OS=$(uname -s)
 MIN_MINOR=9
 MAX_MINOR=12
 
-if command -v python3 >/dev/null 2>&1; then
-    PYTHON=python3
-elif command -v python >/dev/null 2>&1; then
+if command -v python >/dev/null 2>&1; then
     PYTHON=python
+elif command -v python3 >/dev/null 2>&1; then
+    PYTHON=python3
 else
     echo "ERROR: Neither python3 nor python found in PATH."
     exit 1
@@ -79,7 +79,7 @@ fi
 echo "Python $PY_VER is supported."
 
 # Ensure pip targets this Python interpreter
-PIP="$PYTHON_BIN -m pip"
+PIP="$PYTHON -m pip"
 
 # Do this first so pip installs with a built app
 if [ "$BUILD_APP" = true ]; then


install.sh
```diff
@@ -42,80 +42,106 @@ while getopts "hbdempo" FLAG; do
...
+
+# Extract major.minor version info
+PY_VER=$(python3 -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
+MAJOR=$(echo "$PY_VER" | cut -d. -f1)
+MINOR=$(echo "$PY_VER" | cut -d. -f2)
+
+if [ "$MAJOR" -ne 3 ] || [ "$MINOR" -lt "$MIN_MINOR" ] || [ "$MINOR" -gt "$MAX_MINOR" ]; then
+    echo "Python $PY_VER is NOT supported. Please use Python 3.$MIN_MINOR - 3.$MAX_MINOR."
+    exit 1
+fi
```
benjaminpkane, line 77: Awesome

install.sh
```diff
@@ -42,80 +42,105 @@ while getopts "hbdempo" FLAG; do
...
+MIN_MINOR=9
+MAX_MINOR=12
+
+if ! command -v python3 >/dev/null 2>&1; then
+    echo "Python 3 is not installed or not in PATH"
+    exit 1
+fi
+
+# Extract major.minor version info
+PY_VER=$(python3 -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
```
exupero: I'm using a pyenv virtualenv on Python 3.10 but this command says I'm using Python 3.13, I think because the `python3` command isn't shimmed as part of the virtual env. Is there a way to detect that situation and handle it appropriately?

exupero (changes requested):
Exploratory testing notes:
- ❌ Ran in pyenv virtualenv but complained about using Python 3.13 even though virtualenv uses 3.10
- ✅ Commented out Python 3.13 check, ran `./install.sh -p && ./install.sh -d` instead of `bash install.bash -p && bash install.bash -d`, scripted completed
- ✅ Ran app, worked

install.sh
mcdoh: Fixed, thanks!

install.sh
```diff
@@ -0,0 +1,227 @@
...
+NODE_VERSION=22.14.0
+OS=$(uname -s)
+
+MIN_MINOR=9
+MAX_MINOR=12
+
+if command -v python3 >/dev/null 2>&1; then
+    PYTHON=python3
+elif command -v python >/dev/null 2>&1; then
+    PYTHON=python
```
exupero: To pick up my virtualenv's Python, I had to swap the order of these clauses so the shimmed `python` is used if it's found, regardless of whether `python3` is available.

install.sh
```diff
@@ -0,0 +1,227 @@
...
+
+if [ "$MAJOR" -ne 3 ] || [ "$MINOR" -lt "$MIN_MINOR" ] || [ "$MINOR" -gt "$MAX_MINOR" ]; then
+    echo "Python $PY_VER is NOT supported. Please use Python 3.$MIN_MINOR - 3.$MAX_MINOR."
+    exit 1
+fi
+
+echo "Python $PY_VER is supported."
+
+# Ensure pip targets this Python interpreter
+PIP="$PYTHON_BIN -m pip"
```
exupero: Is this supposed to be `$PYTHON`?
```suggestion
PIP="$PYTHON -m pip"
```
Without this change I'm getting a failure on line 159 below.

PR #6354, opened 2025-09-24
Fix pymongo connect after close

kaixi-wang:

- enable re-establishing a connection if the client has been closed 

- added automated test and tested locally
```


## Release Notes

<!--
-   [x] Selected option
-->

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -240,6 +240,8 @@ def establish_db_conn(config):
 
 def _connect():
     global _client
+    if _client and getattr(_client, "_closed", False):
+        _client = None
     if _client is None:
         global _connection_kwargs
 
--- a/fiftyone/core/odm/database.py
+++ b/fiftyone/core/odm/database.py
@@ -275,6 +275,8 @@ def _async_connect(use_global=False):
     _connect()
 
     global _async_client
+    if _async_client and getattr(_async_client, "_closed", False):
+        _async_client = None
     if not use_global or _async_client is None:
         global _connection_kwargs
         client = mtr.AsyncIOMotorClient(
--- a/tests/unittests/odm_tests.py
+++ b/tests/unittests/odm_tests.py
@@ -109,3 +109,14 @@ class GetIndexedValuesTests(unittest.TestCase):
 
         finally:
             dataset.delete()
+
+
+class GetDbConnTests(unittest.TestCase):
+    def test_get_db_conn(self):
+        conn = foo.get_db_conn()
+        self.assertIsNotNone(conn)
+        conn.client.close()
+        assert conn.client._closed == True
+        # should be able to reconnect
+        conn = foo.get_db_conn()
+        assert conn.client._closed == False

### CODE REVIEW

File: fiftyone/core/odm/database.py [hunk 1]
```diff
@@ -240,6 +240,8 @@ def establish_db_conn(config):
 
 def _connect():
     global _client
+    if _client and getattr(_client, "_closed", False):
+        _client = None
```
exupero: Might be nice to encapsulate this in a helper function, less because it's used twice, more to move an implementation detail out of view:
```python
def get_client(client):
    return client if client and not getattr(client, "_closed", False) else None

File: fiftyone/core/odm/database.py [hunk 2]
```diff
@@ -273,6 +275,8 @@ def _async_connect(use_global=False):
     _connect()
 
     global _async_client
+    if _async_client and getattr(_async_client, "_closed", False):
+        _async_client = None
```
exupero: Are there only these two clients, or could this logic be needed elsewhere that we haven't run into yet?

Also, is it possible to test this usage, or would that a disproportionate effort?

tests/no_wrapper/multiprocess_tests.py
```diff
@@ -30,6 +30,19 @@ def _check_process(port):
     assert int(port) == food._connection_kwargs["port"]
 
 
+class GetDbConnTests(unittest.TestCase):
```
exupero: Is this test multiprocessing related? Or should it be moved to a different/new test file?

kaixi-wang: it is semi-related as it is most commonly used with multiprocessing, however, I prob should have created a new file

fiftyone/core/odm/database.py
kaixi-wang: yeah there are only these 2 global clients that are relevant. I've updated the sync test to at least query datasets to verify connection

tests/no_wrapper/db_conn_tests.py
```diff
@@ -0,0 +1,54 @@
...
+
+import fiftyone.core.odm as foo
+
+
+class GetDbConnTests(unittest.TestCase):
+    def test_get_db_conn(self):
+        # initial connection
+        conn = foo.get_db_conn()
+        doc = conn.datasets.find_one({}, {"_id": 1})
+        self.assertIsNotNone(doc)
```
exupero: Is there some setup I need to do for this unit test, e.g. run the database? I'm getting a failure on this line that `doc` is `None`.

kaixi-wang: 🤦‍♀️  my bad. didn't think anyone running the tests would have an empty datasets collection. rabbit was right. fixed

exupero: Is mine empty because I didn't run any other tests before it that would have created a doc?

exupero (approved):
- ✅ Ran tests, passed
- ✅ Reverted changes in fiftyone/, ran tests, both failed 👍

PR #6359, opened 2025-09-25
Add heartbeat for DOs

CamronStaley:

Making it so that monitor thread will update the "update_at" field on the DO at a set interval. This will tell us that the worker is still alive and executing so that we don't automatically mark it as "failed".

Unit tests and locally

command:
```
fiftyone delegated launch -m 
```

results:
```
Running operation 68cdc2e07da656d6f4722cdb (@voxel51/camstale_test/test_wait_multi)

Running operation 68cdc2e07da656d6f4722cdb (@voxel51/camstale_test/test_wait_multi) in child process
[test_wait_multi] Main operator process started.
[test_wait_multi] Spawning 5 worker processes...
[test_wait_multi] Started worker process 1 with PID 48763.
[test_wait_multi] Started worker process 2 with PID 48764.
[test_wait_multi] Started worker process 3 with PID 48765.
[test_wait_multi] Started worker process 4 with PID 48766.
[test_wait_multi] Started worker process 5 with PID 48767.
[test_wait_multi] All worker processes have been spawned.
[test_wait_multi] Now waiting for them to complete...
Pinging operation 68cdc2e07da656d6f4722cdb
Terminating process tree (PID: 48733) for operation 68cdc2e07da656d6f4722cdb. Reason: Operation marked as FAILED externally
[test_wait_multi] All worker processes have completed. Operator finished.
Terminating process tree (PID: 48733) for operation 68cdc2e07da656d6f4722cdb. Reason: Executor shutting down
Process 48733 for op 68cdc2e07da656d6f4722cdb already terminated.
```

I saw that the update_at field was updated everytime it logged a ping

        notes for FiftyOne users.

If you use the fiftyone delegated launch command to run a local executor of delegated operations with the "-m" monitoring flag we will now automatically update the DO update_at field at set intervals to confirm the process is still running.

-   [ ] Other

## Summary by CodeRabbit

- New Features
  - Added CLI option to set monitoring interval for delegated operations (--monitor-interval).
  - Added configuration option and env var to define the default monitoring interval.

- Documentation
  - Refined formatting of the “Configuration options” table for improved readability.

- Bug Fixes
  - Improved delegated operation monitoring with periodic backend pings and stronger handling of ping/get failures and external process termination.

- Tests
  - Updated tests to cover extended monitoring, ping errors, and external/internal failure scenarios.

### DIFF
--- a/fiftyone/operators/delegated.py
+++ b/fiftyone/operators/delegated.py
@@ -734,7 +734,7 @@ class DelegatedOperationService(object):
                     )
                     return ExecutionResult(error=reason)
                 else:
-                    logger.info("Pinging operation %s", operation_id)
+                    logger.debug("Pinging operation %s", operation_id)
                     self._repo.ping(operation_id)
             except Exception as e:
                 reason = f"Error in monitoring loop: {e}"


fiftyone/operators/delegated.py
```diff
@@ -733,6 +733,9 @@ def _monitor_operation(
                         child_process, operation_id, reason
                     )
                     return ExecutionResult(error=reason)
+                else:
+                    logger.info("Pinging operation %s", operation_id)
```
kaixi-wang: should this at least be debug? feels like it could get pretty noisy

fiftyone/core/cli.py
```diff
@@ -3193,8 +3193,17 @@ def setup(parser):
...
+        parser.add_argument(
+            "--monitor-interval",
+            type=int,
+            default=fo.config.delegated_operation_monitor_interval,
+            help="the interval in seconds at which to monitor the operation status (default: %(default)s)",
+        )
+
     @staticmethod
     def execute(parser, args):
+        if args.monitor_interval < 1:
```
kaixi-wang, line 3205: should we also check that if monitor_interval is passed that monitor is also passed?

CamronStaley: Currently we have a default value so we don't really know if the user provided something. I could change it to have no default, then determine based on that but then that kinda makes the env config value pointless. I think it's probably fine

fiftyone/core/cli.py
```diff
@@ -3193,8 +3193,17 @@ def setup(parser):
...
+            raise ValueError("monitor interval must be at least 1 second")
```
tom-vx51, line 3206: not a big deal, but we probably want the lower bound to be higher. 1s polling is pretty aggressive 😆

tom-vx51 (approved):
lgtm, will defer to scheduler folks for DO details

fiftyone/core/cli.py
kaixi-wang: yeah I was thinking having both was kinda redundant and maybe remove the bool...

kaixi-wang (approved):
lgtm

PR #6361, opened 2025-09-25
FOEPD-2119: Limit number of async tasks writing to DB

exupero:

#6288 allows asynchronous background tasks to accumulate without limit, on the assumption that predictions on a batch would take longer than saving that batch to the DB. In cases where saving to the DB takes longer, the number of async tasks can grow without limit and could cause the process to run out of memory.

This PR wraps the batch iterator in a new utility function that limits the number background tasks. After each iteration of the source iterator, if there are more background tasks than the specified limit, tasks are waited for until the backlog is reduced to the limit, at which point the next item is taken from the iterator.

This approach is open to discussion. If writing to the DB is the bottleneck and creates backpressure on the prediction pipeline, then as soon as enough async tasks accumulate, performance degenerates to the same throughput as never using async tasks. In that case, we might prefer simplicity and revert #6288 to go back a single thread.

Ran the following script on a machine with a GPU, with some added logging that recorded how many tasks had accumulated. Logs showed the expected limits were observed.

I also tried using [memory-profiler](https://pypi.org/project/memory-profiler/), but the profiles showed the same memory characteristics as the original code. I suspect memory usage is dominated by the downloaded samples, not by the data needed for the async tasks, so the effect of limiting the queue doesn't show:
Figure_1

@jacobsela has some more sophisticated testing he might be able to apply.

        notes for FiftyOne users.

-   [ ] Other


exupero:
@jacobsela are you able to run your analysis on this branch to see if it makes any difference?

fiftyone/core/utils.py
```diff
@@ -3206,19 +3206,76 @@ def async_executor(
...
         def submit(*args, **kwargs):
             future = executor.submit(*args, **kwargs)
-            _futures.append(future)
+            futures.append(future)
             return future
 
         yield submit
 
-        for future in _futures:
+        for future in futures:
```
exupero: These lines are an opportunistic cleanup in the original code.

With this branch this function is no longer used in fiftyone, so it could be removed, but the code has been released so there could be third-party uses of it. Suggestions on whether to keep it or get rid of it?

fiftyone/core/models.py
```diff
@@ -499,9 +492,15 @@ def save_batch(sample_batch, labels_batch):
...
-        for sample_batch, imgs in zip(
-            fou.iter_batches(samples, batch_size),
-            data_loader,
+        for submit, (sample_batch, imgs) in fou.async_iterator(
+            zip(
+                fou.iter_batches(samples, batch_size),
+                data_loader,
+            ),
+            limit=10,
+            max_workers=1,
```
exupero: Some defaults. Ideally these would be configurable by the user.

kaixi-wang: what is the reason for limit to 10?

fiftyone/core/utils.py
```diff
@@ -3206,19 +3206,76 @@ def async_executor(
...
+    with ThreadPoolExecutor(max_workers=max_workers) as executor:
+        futures = []
+
+        def submit(*args, **kwargs):
+            future = executor.submit(*args, **kwargs)
+            futures.append(future)
+            return future
+
+        for item in iterator:
+            yield submit, item
```
kaixi-wang, line 3310: is this supposed to be submit(item)?

fiftyone/core/models.py
exupero: Arbitrary. Didn't have any reason to choose something different. @jacobsela might be able to weigh in with an informed option.

fiftyone/core/utils.py
exupero: No. This is yielding the submit function and the current item to the consumer loop, so the consumer can call `submit` where it wants and with the function it wants to run async.

exupero:
Profiled memory on this branch with @jacobsela's testing script and found that limiting the number of background tasks doesn't seem to help. Even when using a limit of 0, which should be behaviorally equivalent to develop before #6288 (i.e. each batch's prediction is blocked on the previous batch being saved), memory usage climbs almost as high as without a limit. Increasing the number of workers speeds up execution but doesn't use that much less memory.
visualization(14)
(Note: the light blue run on 2000 samples, "queue limit 2, 2 workers", was aborted before finishing.)

PR #6362, opened 2025-09-25
Adding Docs for Getting Started with FOE 

sherpan:

Adding new documentation for "Getting Started with FiftyOne Enterprise"

(Details)

        notes for FiftyOne users.

-   [ ] Other


docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,198 @@
+.. _getting-started-with-foe:
```
brimoor: nit can you use `enterprise-getting-started` for the anchor link for consistency with the existing anchor links for most other FOE pages?

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,198 @@
...
+      Set the following environment variables:
+
+      .. code-block:: bash
+
+         export MINIO_ACCESS_KEY_ID=...
+         export MINIO_SECRET_ACCESS_KEY=...
+         export MINIO_DEFAULT_REGION=...
+
+Install the FOE Python SDK
+------------------------
```
brimoor: nit: the docs build throws warnings if the underlines don't exactly match the length of the titles above them. can you fix this throughout? 🤓

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,198 @@
...
+
+      Set the following environment variables:
+
+      .. code-block:: bash
+
+         export MINIO_ACCESS_KEY_ID=...
+         export MINIO_SECRET_ACCESS_KEY=...
+         export MINIO_DEFAULT_REGION=...
+
+Install the FOE Python SDK
```
brimoor: we use "FOE" internally, but in the docs we always say "FiftyOne Enterprise". Can you update throughout?

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,198 @@
...
+-------------------------------------------
+
+.. default-role:: code
+Configure cloud credentials to enable rendering of cloud-backed media within FiftyOne Enterprise.
+
+.. image:: /images/enterprise/getting_started_cloud_creds.gif
+   :alt: getting-started-cloud-creds
+   :align: center
+
+Set Cloud Credentials Locally
```
brimoor: Everything below here describes how to create your first dataset from the SDK, but it is possible to do the same thing directly in the FOE App.

For example:

1. Install the `@voxel51/io`, `@voxel51/utils`, `@voxel51/zoo`, and `@voxel51/brain` plugins
2. Create dataset via UI
3. Use `import_samples` to add samples for media in a cloud bucket 
4. Use `compute_metadata` to populate metadata
5. Use `apply_zoo_model` to add some model predictions
6. Use the Embeddings panel to generate an embeddings visualization

Moreover:
- 3,4,5,6 are perfect opportunities to introduce **delegated execution**
- 4,5 now support "fan out", which is a golden opportunity to introduce users to the fact that delegated execution supports parallel execution

In the spirit of driving awareness to in-App workflows, can you add this flow to the guide and put it **above** the SDK version?

sherpan:
@brimoor addressed your feedback and added the UI alternative

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+Getting Started with FiftyOne Enterprise
+========================================
+.. default-role:: code
+
+This guide provides comprehensive instructions to successful upload your first dataset to FiftyOne Enterprise (FOE). 
+
+Upload Cloud Credentials to your Deployment
+-------------------------------------------
+
+.. default-role:: code
```
brimoor: I believe this second `.. default-role:: code` is unnecessary

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+   :alt: getting-started-cloud-creds
+   :align: center
+
+Create a dataset via the SDK 
+-----------------------------
+
+Set Cloud Credentials Locally
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Next, configure the appropriate environment variables for your cloud provider within your local environment
```
brimoor: Can you change "cloud provider" in this sentence to this?

```
:ref:`cloud credentials <enterprise-cloud-credentials>`
```

As a best practice, always want to link to more detailed documentation for a concept whenever relevant.

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
+.. _enterprise-getting-started:
+
+Getting Started with FiftyOne Enterprise
+========================================
+.. default-role:: code
+
+This guide provides comprehensive instructions to successful upload your first dataset to FiftyOne Enterprise (FOE).
```
brimoor: nit: our [docs style guide](https://github.com/voxel51/fiftyone/blob/develop/STYLE_GUIDE.md#documentation-style-guide) requests line length of 79 characters or less. Can you apply this?

Adding a ruler in your code editor is a helpful tool 🤓

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+3. Copy and execute the provided bash command to install the SDK in your virtual environment
+
+.. image:: /images/enterprise/getting_started_install_sdk.gif
+   :alt: getting-started-install-sdk
+   :align: center
+
+Connect to Your Deployment
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To establish a connection between the FiftyOne Enterprise Python SDK and your deployment, configure the ``FIFTYONE_API_KEY`` and ``FIFTYONE_API_URI`` environment variables. For detailed setup instructions, refer to the `API Connection Documentation <https://docs.voxel51.com/api/>`_.
```
brimoor: Please replace this:

```
`API Connection Documentation <https://docs.voxel51.com/api/>`_
```

with this:

```
`:ref:`API connection documentation <enterprise-api-connection>`
```

The former style technically works, but it is meant for external links to other websites.

The `:ref:` style is meant for links to other pages within the docs, and it is preferred because no changes to this line of the docs would be required if we decided to rename our docs URL or even edit the section heading that you're linking to, because `enterprise-api-connection` is referring to the "anchor link":

https://github.com/voxel51/fiftyone/blob/c9dcfa42bec7626aabb6413117909af791c82da0/docs/source/enterprise/api_connection.rst?plain=1#L1

By convention, we do not edit anchor links after we've created them so that we don't cause broken links within our docs if we rename section headings.

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+Verify that your API connection is working correctly using the following method:
+
+.. code-block:: python
+
+   import fiftyone.management as fom  # if this fails, you may have the open-source SDK installed
+
+   fom.test_api_connection()  # API connection succeeded
+
+ 
+You can use ``fiftyone.config`` to debug and verify your configuration:
```
brimoor: perhaps explain what the user may should be looking for here?

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+
+      Set the following environment variables:
+
+      .. code-block:: bash
+
+         export MINIO_ACCESS_KEY_ID=...
+         export MINIO_SECRET_ACCESS_KEY=...
+         export MINIO_DEFAULT_REGION=...
+
+Install the FiftyOne Enterprise Python SDK
```
brimoor: From a semantic standpoint, I think the "install" step should come before configuring your cloud credentials.

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+
+   .. tab-item:: MinIO
+
+      Set the following environment variables:
+
+      .. code-block:: bash
+
+         export MINIO_ACCESS_KEY_ID=...
+         export MINIO_SECRET_ACCESS_KEY=...
+         export MINIO_DEFAULT_REGION=...
```
brimoor, line 145: How about adding something that helps users test whether their cloud creds are correctly registered?

```py
import fiftyone.core.storage as fos

fos.list_buckets("s3|gcs|azuer|minio")
```

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+ 
+You can use ``fiftyone.config`` to debug and verify your configuration:
+
+.. code-block:: python
+
+   import fiftyone as fo
+
+   print(fo.config)
+
+Create a Dataset and Add Samples
```
brimoor: How about linking to https://docs.voxel51.com/enterprise/cloud_media.html here so interested users can learn more about how to interact with cloud-backed media in the FOE SDK?

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+
+         dataset.add_samples(samples)
+         dataset.persistent = True
+
+   
+
+Compute Metadata
+~~~~~~~~~~~~~~~~
+
+Improve app performance by computing sample metadata:
```
brimoor: I would recommend expanding on this statement in two ways:

1. Explain that `compute_metadata()` is a builtin method that you can use to efficiently populate basic metadata such as file size, image height/width, etc for all of the samples in your dataset
2. Then you can mention that keeping the `metadata` field populated for all samples of your datasets is recommended because it enables the sample grid's tiling algorithm to run more efficiently 

I'd also recommend including a code output block that shows some example metadata, for completeness:

```py
sample = dataset.first()
print(sample.metadata)
```

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+
+   dataset.compute_metadata()
+
+Verify all samples have metadata by running the following:
+
+.. code-block:: python
+
+   len(dataset.exists("metadata", False))  # Should be 0
+.. note::
+   For video datasets, ensure ffmpeg is installed to enable metadata computation.
```
brimoor: I would recommend moving this note about `ffmpeg` installation to a subsection of the "install" section above. IE: if you're working with video datasets, you'll also want to install `ffmpeg`.

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+
+Create a dataset via the UI 
+-----------------------------
```
brimoor: nit: underline should match header length above to avoid warnings during docs build

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+Install the IO Plugin 
+~~~~~~~~~~~~~~~~~~~~~
+The IO plugin allows you to import data from your cloud storage buckets directly. The IO plugin can be downloaded from the following `repository <https://github.com/voxel51/fiftyone-plugins/tree/main/plugins/io>`_. Zip the folder and upload it to the Plugin section of the **Settings** page.
+
+.. note::
+   To install plugins to your deployment, you must have admin privileges and also set your deployment to be in Dedicated Plugins mode. It is also highly recommended to add Delegated Operators to your deployment.
```
brimoor: Isn't dedicated plugins mode an artifact of the past? I definitely hope that dedicated plugins is the default, and therefore that we shouldn't need to mention it here?

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+
+   len(dataset.exists("metadata", False))  # Should be 0
+.. note::
+   For video datasets, ensure ffmpeg is installed to enable metadata computation.
+
+Create a dataset via the UI 
+-----------------------------
+Install the IO Plugin 
+~~~~~~~~~~~~~~~~~~~~~
+The IO plugin allows you to import data from your cloud storage buckets directly. The IO plugin can be downloaded from the following `repository <https://github.com/voxel51/fiftyone-plugins/tree/main/plugins/io>`_. Zip the folder and upload it to the Plugin section of the **Settings** page.
```
brimoor: Since this getting started guide is oriented at end users who are likely not admins, how about just jumping directly to `Import your Dataset` section first, and then just add a small `.. note` at the bottom mentioning that an admin must install the IO plugin as a one-time setup step:

```
.. note::

    An admin must follow :ref:`these instructions <enterprise-plugins-install>`
    to install the
    `@voxel51/io <https://github.com/voxel51/fiftyone-plugins/blob/main/plugins/io/README.md>`_
    plugin in order to perform imports via the Enterprise UI. 
```

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,231 @@
...
+~~~~~~~~~~~~~~~~~~~
+Schedule the **import_samples**  operator to import your dataset from your cloud storage bucket.
+
+.. image:: /images/enterprise/getting_started_import_samples.gif
+   :alt: getting-started-import-samples
+   :align: center
+
+Compute Metadata
+~~~~~~~~~~~~~~~~
+To improve app performance, compute sample metadata by scheduling the **compute_metadata** operator as a delegated operation.
```
brimoor: Recommend adding the same content I suggested in the SDK section for explaining why metadata is useful.

docs/source/enterprise/getting_started.rst
sherpan: @brimoor given that we are switching the order around. the next step will test this via 

`fos.list_files(dirpath="")`

docs/source/enterprise/getting_started.rst
sherpan: unfortunately it looks like [built-in mode](https://github.com/voxel51/fiftyone-teams-app-deploy/tree/main/docker#jigsaw-step-6-configuring-fiftyone-enterprise-plugins) is default still

docs/source/enterprise/getting_started.rst
sherpan: oh i like this. then we re just remove the note when we eventually make it built in

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,267 @@
+.. _enterprise-getting-started:
+
+Getting Started with FiftyOne Enterprise
+========================================
+.. default-role:: code
+
+This guide provides comprehensive instructions to successfully upload your first dataset to FiftyOne Enterprise (FOE).
+
+Upload Cloud Credentials to your Deployment
```
brimoor: Like installing the IO plugin, uploading cloud credentials is something that admins do only once.

I can think of two ways to address that here:
1. Tweak this text to explain that this is a one time setup that only an admin needs to do
2. Convert this to a `.. note:` in the `Create a dataset via the UI ` section (since this step isn't strictly required to create a dataset via the SDK)

docs/source/enterprise/getting_started.rst
```diff
@@ -0,0 +1,267 @@
...
+~~~~~~~~~~~~~~~~~~~
+Schedule the **import_samples**  operator to import your dataset from your cloud storage bucket.
+
+.. image:: /images/enterprise/getting_started_import_samples.gif
+   :alt: getting-started-install-sdk
+   :align: center
+
+Compute Metadata
+~~~~~~~~~~~~~~~~
+**compute_metadata** is a builtin operator that effeciently populates basic
```
brimoor: It's not yet builtin sadly; must install the `@voxel51/utils` operator

docs/source/enterprise/getting_started.rst
sherpan: @brimoor i think option 1 works better. while  its technically not required to create a dataset via the SDK i still think the whole point is see and visualize their data so i rather not hide it there.

docs/source/enterprise/getting_started.rst
sherpan: @brimoor ah nice added it into the note to install the io and util plugins together

Issue #6346, opened 2025-09-18
[BUG] Mask filename creation process is incorrect when using the function objects_to_segmentations()


sanjeev-sudarsan:
### Describe the problem

When using the function `objects_to_segmentations()` to generate segmentation masks:

First run:

Mask files are created with the same filename as the source `filepath`.

Second run (and beyond):

New mask files are generated with the suffix -2, regardless of the overwrite setting.

Any further runs continue to overwrite only the -2 versions of the files, without creating -3, -4, etc.

Expected Behavior

If overwrite=True:
Existing mask files should be overwritten in place. The function should not delete the entire `out_dir` (its mentioned in the documentation that this is the expected behavior, but I would expect that only the relevant files are overwritten), since it may contain unrelated files outside of the current sample_collection.

If overwrite=False:
Each subsequent run should generate new mask files with incrementally increasing suffixes (-2, -3, -4, etc.), so no previous outputs are lost.

### Code to reproduce issue

```python
import fiftyone as fo
from fiftyone.utils.labels import objects_to_segmentations

dataset_with_segmentation_polylines = fo.load_dataset("segmentation_dataset")

objects_to_segmentations(
    sample_collection=dataset_with_segmentation_polylines,
    in_field="segmentation_polylines",
    out_field="segmentation_masks",
    mask_targets={100: "car", 200: "person"},
    output_dir="/path/to/segmentation_masks",
    rel_dir="/path/to/media_files",
    progress=True,
)
```

<img width="682" height="297" alt="Image" src="https://github.com/user-attachments/assets/c17af875-f6c6-4927-821f-b1e68032c621" />

### System information

-   **OS Platform and Distribution** (e.g., Linux Ubuntu 22.04): Linux Ubuntu 22.04)
-   **Python version** (`python --version`):  Python 3.10.12
-   **FiftyOne version** (`fiftyone --version`): FiftyOne v1.4.1
-   **FiftyOne installed from** (pip or source): pip

### Other info/logs

Include any logs or source code that would be helpful to diagnose the problem.
If including tracebacks, please include the full traceback. Large logs and
files should be attached. Please do not use screenshots for sharing text. Code
snippets should be used instead when providing tracebacks, logs, etc.

### Willingness to contribute

The FiftyOne Community encourages bug fix contributions. Would you or another
member of your organization be willing to contribute a fix for this bug to the
FiftyOne codebase?

-   [ ] Yes. I can contribute a fix for this bug independently
[...]

PR #6364, opened 2025-09-29
FOEPD-2166: Increment generated filenames to avoid overwriting existing files

brimoor:
Resolves https://github.com/voxel51/fiftyone/issues/6346

## Change log

- Fixes a bug where `fiftyone.utils.labels` utilities with `overwrite=False` options would erroneously start overwriting existing files if repeatedly executed more than 2 times
- Updates the `fiftyone.utils.labels` utilities with `overwrite=True` options so that they now only overwrite the specific files being exported (rather than deleting the entire directory)

## Tested by

Exporting to the same directory multiple times with `overwrite=False` now continually increments the `-%d` suffix:

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.core.storage as fos

    "coco-2017",
    split="validation",
    label_types="segmentations",
    label_field="instances",
    max_samples=10,
    only_matching=True,
)

mask_targets = {"#ff6d04": "person", "#499cef": "cat", "#6d04ff": "dog"}
output_dir = "/tmp/segmentations"

fos.delete_dir(output_dir)

for _ in range(3):
        dataset,
        "instances",
        "segmentations",
        mask_targets=mask_targets,
        output_dir=output_dir,
        overwrite=False,
    )

    fo.pprint(dataset.values("segmentations.mask_path"))
```

```
[
    '/tmp/segmentations/000000169076.png',
    '/tmp/segmentations/000000375278.png',
    '/tmp/segmentations/000000000139.png',
    '/tmp/segmentations/000000000785.png',
    '/tmp/segmentations/000000000872.png',
    '/tmp/segmentations/000000000885.png',
    '/tmp/segmentations/000000001000.png',
    '/tmp/segmentations/000000001268.png',
    '/tmp/segmentations/000000001296.png',
    '/tmp/segmentations/000000001353.png',
]
[
    '/tmp/segmentations/000000169076-2.png',
    '/tmp/segmentations/000000375278-2.png',
    '/tmp/segmentations/000000000139-2.png',
    '/tmp/segmentations/000000000785-2.png',
    '/tmp/segmentations/000000000872-2.png',
    '/tmp/segmentations/000000000885-2.png',
    '/tmp/segmentations/000000001000-2.png',
    '/tmp/segmentations/000000001268-2.png',
    '/tmp/segmentations/000000001296-2.png',
    '/tmp/segmentations/000000001353-2.png',
]
[
    '/tmp/segmentations/000000169076-3.png',
    '/tmp/segmentations/000000375278-3.png',
    '/tmp/segmentations/000000000139-3.png',
    '/tmp/segmentations/000000000785-3.png',
    '/tmp/segmentations/000000000872-3.png',
    '/tmp/segmentations/000000000885-3.png',
    '/tmp/segmentations/000000001000-3.png',
    '/tmp/segmentations/000000001268-3.png',
    '/tmp/segmentations/000000001296-3.png',
    '/tmp/segmentations/000000001353-3.png',
]
```

Exporting to an existing directory with `overwrite=True` now only overwrites the specific files being exported (rather than deleting the entire directory):

```py
    dataset,
    "instances",
    "segmentations",
    mask_targets=mask_targets,
    output_dir=output_dir,  # already contains files from previous exports
    overwrite=True,
)

fo.pprint(dataset.values("segmentations.mask_path"))

assert len(fos.list_files(output_dir)) == 30
```

```
[
    '/tmp/segmentations/000000169076.png',
    '/tmp/segmentations/000000375278.png',
    '/tmp/segmentations/000000000139.png',
    '/tmp/segmentations/000000000785.png',
    '/tmp/segmentations/000000000872.png',
    '/tmp/segmentations/000000000885.png',
    '/tmp/segmentations/000000001000.png',
    '/tmp/segmentations/000000001268.png',
    '/tmp/segmentations/000000001296.png',
    '/tmp/segmentations/000000001353.png',
]
```

## Summary by CodeRabbit

- New Features
  - Automatic filename de-duplication: exports and transforms now append “-N” to resolve name conflicts across 2D/3D segmentation workflows.

- Bug Fixes
  - Overwrite no longer deletes the entire output directory; existing files are preserved and new outputs get unique names.
  - Improved handling of filename collisions to avoid accidental overwrites.

- Documentation
  - Clarified overwrite behavior: the output directory is not removed; conflicts result in uniquely suffixed filenames.

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -2142,6 +2142,19 @@ class UniqueFilenameMaker(object):
 
         count = self._filename_counts[key]
         if count > 1:
+            # Handle existing filenames that use `-%d` suffix
+            if not self.ignore_existing:
+                while True:
+                    _key = name + ("-%d" % count)
+                    if not self.ignore_exts:
+                        _key += ext
+
+                    if _key in self._filename_counts:
+                        self._filename_counts[key] += 1
+                        count += 1
+                    else:
+                        break
+
             filename = name + ("-%d" % count) + ext
 
         if self.chunk_size is not None:
--- a/fiftyone/utils/labels.py
+++ b/fiftyone/utils/labels.py
@@ -67,8 +67,8 @@ def objects_to_segmentations(
             ``output_dir`` that match the shape of the input paths. The path is
             converted to an absolute path (if necessary) via
             :func:`fiftyone.core.storage.normalize_path`
-        overwrite (False): whether to delete ``output_dir`` prior to exporting
-            if it exists
+        overwrite (False): whether to overwrite any existing files in
+            ``output_dir``
         save_mask_targets (False): whether to store the ``mask_targets`` on the
             dataset
         progress (None): whether to render a progress bar (True/False), use the
@@ -99,12 +99,12 @@ def objects_to_segmentations(
 
     samples = sample_collection.select_fields(in_field)
 
-    if overwrite and output_dir is not None:
-        etau.delete_dir(output_dir)
-
     if output_dir is not None:
         filename_maker = fou.UniqueFilenameMaker(
-            output_dir=output_dir, rel_dir=rel_dir, idempotent=False
+            output_dir=output_dir,
+            rel_dir=rel_dir,
+            ignore_existing=overwrite,
+            idempotent=False,
         )
 
     for sample in samples.iter_samples(autosave=True, progress=progress):
@@ -195,8 +195,8 @@ def export_segmentations(
             converted to an absolute path (if necessary) via
             :func:`fiftyone.core.storage.normalize_path`
         update (True): whether to delete the arrays from the database
-        overwrite (False): whether to delete ``output_dir`` prior to exporting
-            if it exists
+        overwrite (False): whether to overwrite any existing files in
+            ``output_dir``
         progress (None): whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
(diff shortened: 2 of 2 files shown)


brimoor:
It's only a slight change in behavior for an *optional syntax* (passing `overwrite=True`).

I'm proposing it because I agree with the user who [requested](https://github.com/voxel51/fiftyone/issues/6346) this change that it is more likely that I may have other files in `output_dir` that I actually want to keep than it is that I want the entire directory to be deleted:

> Existing mask files should be overwritten in place. The function should not delete the entire out_dir (its mentioned in the documentation that this is the expected behavior, but I would expect that only the relevant files are overwritten), since it may contain unrelated files outside of the current sample_collection.

To your questions:

> a script which once kept a directory a constant size will now use increasing amounts of storage

The new `overwrite=True` syntax will overwrite any files generated by previous calls to this method if performed `n>1` times in a loop, so it won't cause runaway disk space, it just leaves any files unrelated to these utilities that previously existed in-place.

> downstream scripts which expect new files to have fixed names will now find old versions of the file

`overwrite=True` does produce fixed names; if you run it `n>1` times the files will have the same names. Either way though, users will use `dataset.values("<field>.mask_path")` as the source of truth for their mask paths, not `os.listdir()`. So the presence of additional files on disk is highly unlikely to change the behavior of an existing script that uses these methods.

Basically I think this change will allow more people to use `overwrite=True` that weren't able to use it, and anyone that was previously using `overwrite=True` will still be satisfied with what it does.

exupero:
> It's only a slight change in behavior for an optional syntax (passing `overwrite=True`).

I agree that when `overwrite=True` directory size should be flat. I'm asking about non-flat directory size when `overwrite=False`, per the PR description:

> Exporting to the same directory multiple times with `overwrite=False` now continually increments the `-%d` suffix:

That's the change in default behavior I'm concerned could surprise users. Per #6346, the current behavior is that "Any further runs continue to overwrite only the -2 versions of the files", so while that behavior isn't what we intend, it seems plausible that users could have scripts that depend on that behavior and won't look for the `-3`, `-4`, etc. filenames that will now be created.

Personally I'm fine with the breaking change. @jleven might have thoughts on what breaking changes are small enough to slip in, and which need to be handled with more consistency for users.

brimoor:
Oh I see. I definitely want to ship the `-2, -3, -4` part of the patch because the current behavior was just wrong/bad. It would overwrite files when the user passes `overwrite=False` and moreover it would do it in a bizarre and undocumented way.

PR #6365, opened 2025-09-29
Add support for updating existing indexes in-place

brimoor:
See [FOEPD-1846](https://voxel51.atlassian.net/browse/FOEPD-1846) for context

## Change log

1. Updates `create_index()` to support converting an existing non-unique index to a unique index in-place (using efficient `collMod` operations when possible)
2. Adds a `create_index(..., force=True)` syntax that supports other modifications to existing indexes in-place, such as downgrading an existing unique index to a non-unique index or changing the ordering of an existing index
3. Updates `merge_samples()` to **NOT** delete any unique indexes that it creates to support the merge operation (for efficiency)
4. Updates the builtin `create_index` operator to support the `force=True` syntax introduced in 2 above

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz

is_unique_index = lambda d, i: d.get_index_information()[i].get("unique", False)

dataset1 = dataset[:100].clone()
dataset2 = dataset[100:].clone()

#
# Change in behavior: the 'filepath' index is *not* automatically downgraded
# to non-unique after the merge
#

assert not is_unique_index(dataset1, "filepath")

dataset1.merge_samples(dataset2)

assert is_unique_index(dataset1, "filepath")

#
# New syntax: force 'filepath' to be downgraded back to non-unique
#

dataset1.create_index("filepath", unique=False, force=True)

assert not is_unique_index(dataset1, "filepath")

#
# New capability: change the ordering of an existing index
#

assert dataset1.get_index_information()["filepath"]["key"][0][1] == 1

dataset1.create_index([("filepath", -1)], force=True)

assert dataset1.get_index_information()["filepath"]["key"][0][1] == -1
```


fiftyone/core/dataset.py
```diff
@@ -10802,6 +10733,10 @@ def _index_frames(dataset, key_field, frame_key_field):
         _allow_missing=True,
     )
 
+    return dataset._frame_collection.create_index(
+        [(frame_key_field, 1), ("frame_number", 1)], unique=True
+    )
+
```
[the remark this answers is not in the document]
brimoor: OMG you're a genius rabbit! Partial indexes are exactly what we need here 🤯

tests/unittests/video_tests.py
```diff
@@ -164,13 +181,61 @@ def test_video_indexes(self):
         dataset.drop_index("frames.cls.label")
         self.assertNotIn("frames.cls.label", dataset.list_indexes())
 
-        compound_index_name = dataset.create_index(
-            [("frames.id", 1), ("frames.field", 1)]
+        dataset.create_index([("frames.field", -1)])
+        indexes = dataset.get_index_information()
+        indexes["frames.field"]["key"] == [("field", -1)]
+
```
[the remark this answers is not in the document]
brimoor: thanks for catching this

tests/unittests/dataset_tests.py
```diff
@@ -704,11 +731,47 @@ def test_indexes(self):
...
-        compound_index_name = dataset.create_index([("id", 1), ("field", 1)])
-        self.assertIn(compound_index_name, dataset.list_indexes())
+        dataset.create_index([("field", -1)])
+        indexes = dataset.get_index_information()
+        indexes["field"]["key"] == [("field", -1)]
+
+        dataset.create_index([("field", 1)])
+        indexes = dataset.get_index_information()
+        self.assertEqual(indexes["field"]["key"], [("field", -1)])  # still -1
+
```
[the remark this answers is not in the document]
brimoor: thanks for catching this

fiftyone/core/dataset.py
```diff
@@ -10133,6 +10081,10 @@ def _merge_samples_pipeline(
     #
     # Prepare frames merge pipeline
     #
+    # @todo this implementation creates a unique index on a temporary
+    # `frame_key_field`, which means that users will get errors if they try to
+    # create new frames while a merge is in progress. We need to fix this
+    #
```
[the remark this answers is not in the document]
brimoor: Unfortunately this doesn't work because I am using this index for a `$merge` operation, which requires a unique index on the destination collection that fully covers all documents. I tried your suggestion but I got an error when calling `$merge` when I include the `partialFilterExpression` argument. 

Do you have any ideas to simplify the merge operation for video datasets that avoids the need for a temporary `frame_key_field`?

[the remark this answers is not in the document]
brimoor: I like your suggestion. Please implement the change as a focused commit so I can review it!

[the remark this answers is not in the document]
brimoor: I like your suggestion. Can you implement it please?

fiftyone/core/collections.py
```diff
@@ -10548,42 +10523,132 @@ def create_index(self, field_or_spec, unique=False, wait=True, **kwargs):
...
+            else:
+                coll = self._dataset._get_sample_collection(
+                    write_concern=WriteConcern(w=0)
+                )
 
-        if single_field_index:
-            name = input_spec[0][0]
-        elif is_frame_index:
-            name = self._FRAMES_PREFIX + name
+        coll.create_index(index_spec, unique=unique, **kwargs)
```
kaixi-wang, line 10653: can we add a log to indicate that an index is being created? It could help debug slowdowns.

fiftyone/core/dataset.py
```diff
@@ -10133,6 +10081,10 @@ def _merge_samples_pipeline(
     #
     # Prepare frames merge pipeline
     #
+    # @todo this implementation creates a unique index on a temporary
```
kaixi-wang, line 10084: Do we have a ticket for this so it's not just lost in code?

fiftyone/core/collections.py
```diff
@@ -10548,42 +10523,132 @@ def create_index(self, field_or_spec, unique=False, wait=True, **kwargs):
...
+            if not convert_to_unique:
+                logger.info(
+                    f"Replacing existing index '{index_name}' "
+                    f"on dataset '{self._dataset.name}'"
+                )
+
+            coll.drop_index(_existing_db_name)
+
+        # Create new index, if necessary
+        if not wait:
```
kaixi-wang, line 10638: I know this isn't really changed by this pr, but setting w=0 makes the index operation return immediately, but subsequent operations in the same python session seem to hang. For example 

```
import fiftyone as fo
ds=fo.load_dataset('big-image-ds')
ds.create_index('db_pos') # returns immediately
ds = fo.load_dataset('quickstart') # hangs forever
```

```
import fiftyone as fo
ds=fo.load_dataset('big-image-ds')
ds.create_index('index'')  #returns immediately
ds._sample_collection.find_one({}) #hangs forever
```

brimoor: Hangs forever? Or hangs until the index is built and then succeeds?

Can you create a ticket for this and diagnose what's happening? Potentially we need to remove the `wait=False` option from `create_index()` eh?

fiftyone/core/dataset.py
brimoor: It's been resolved in https://github.com/voxel51/fiftyone/pull/6370!

fiftyone/core/collections.py
kaixi-wang: Not sure as the index build would have taken awhile as well for that dataset. 

Created ticket in the backlog to investigate https://voxel51.atlassian.net/browse/FOEPD-2223

fiftyone/core/odm/database.py
```diff
@@ -501,6 +501,16 @@ def get_db_conn():
     return _apply_options(db)
 
 
+def get_db_version():
+    """Returns the database version.
```
kaixi-wang, line 505: This is checking the same version as https://github.com/voxel51/fiftyone/blob/5872c4f3ec80a0f17b4aa78b16ec025d0035e8ba/fiftyone/core/odm/database.py#L343. Should that function be updated?

fiftyone/core/collections.py
```diff
@@ -10548,42 +10523,132 @@ def create_index(self, field_or_spec, unique=False, wait=True, **kwargs):
                 "or all frame-level fields"
             )
 
+        to_db_name = lambda spec: "_".join("%s_%s" % (f, o) for f, o in spec)
```
kaixi-wang, line 10526: Unlikely that people are using custom names, but just want to call out that this assumes the default naming scheme and wouldn’t work for named indexes

brimoor: Yeah, our public interface doesn't document support for custom names, so it's fine to not support it here

fiftyone/core/collections.py
brimoor: Adding a log for new index creation would cause some weird logging patterns, eg creating datasets/generated views would start issuing logs related to index creation, which would feel weirdly incomplete because creating a dataset does a number of other things that wouldn't get logs.

In general, I think our core SDK methods should be very light on logs; ie using `fiftyone` should feel similar to using scientific computing libraries like `numpy` or `pandas` (very light on logs). Any logging is best placed in the "application layer"; ie whatever is calling the core SDK methods.

fiftyone/core/odm/database.py
brimoor: Like to reduce code duplication? For me the shared code:

```py
Version(client.server_info()["version"])
```

doesn't quite reach the bar of needing a util method, as its just one line. If it were multiple lines of logic, I could see it.

fiftyone/core/collections.py
kaixi-wang: Without deep knowledge of the code, it would be really difficult to know from the "application layer" or callers that indexes are being built. Can we at least have a debug log? This would really help with triaging performance issues.

fiftyone/core/odm/database.py
kaixi-wang: Yeah like just update that code to call this one in case it changes for whoever reason.

fiftyone/core/collections.py
brimoor: okay added the debug log: https://github.com/voxel51/fiftyone/pull/6365/commits/bcc96b85b245916fff10fd29f9382cc6864c3493 ✅

kaixi-wang (approved):
lgtm 👍 hopefully this will make mitigate some of the perfomance issues/side effects of merge samples

PR #6369, opened 2025-09-29
feat(operators): new download file operators

ritch:
New download_file operators

[Example Operators PR](https://github.com/voxel51/fiftyone-plugins/pull/261)

## Summary by CodeRabbit

* **New Features**
  * Added a “Download file” operator to download files in-browser from a URL, with optional custom filename.
  * Automatically determines filename from provided input, response headers, or URL; falls back to direct link download on fetch failure or CORS issues.
  * Operator available in the operators palette and can be triggered programmatically via a new Python method.

### DIFF
--- a/plugins/operators/__init__.py
+++ b/plugins/operators/__init__.py
@@ -2701,6 +2701,29 @@ def _get_non_default_frame_fields(dataset):
     return schema
 
 
+class DownloadFileOperator(foo.Operator):
+    @property
+    def config(self):
+        return foo.OperatorConfig(
+            name="download_file",
+            label="Download file",
+        )
+
+    def resolve_input(self, ctx):
+        inputs = types.Object()
+        inputs.str(
+            "url",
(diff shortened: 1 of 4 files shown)


plugins/operators/__init__.py
```diff
@@ -2701,6 +2701,29 @@ def _get_non_default_frame_fields(dataset):
     return schema
 
 
+class DownloadFileOperator(foo.Operator):
+    @property
+    def config(self):
+        return foo.OperatorConfig(
```
ritch, line 2707: Needs to be unlisted...!

imanjra: Seems like proxy to js op, can we just remove it?

ritch: Nope, since this will be a real op in teams

ritch: Ill add a comment

PR #6371, opened 2025-10-01
Add support for pose estimation models in HF transformers integration

manushreegangwar:

This PR adds `FiftyOneTransformerForPoseEstimation` to support transformer-based keypoint prediction models ([ViTPose](https://huggingface.co/docs/transformers/model_doc/vitpose#vitpose)) to the fiftyone model zoo.

```
import fiftyone as fo
import fiftyone.zoo as foz

dataset_person = dataset.filter_labels("ground_truth", F("label") == "person")
```

- [x] Run inference via `apply_model`
```
for m_name in zoo_models:
    model = foz.load_zoo_model(m_name)
    dataset.apply_model(model, prompt_field="ground_truth", label_field=f"kpts_{m_name}")
    dataset_person.apply_model(model, prompt_field="ground_truth", label_field=f"kpts_person_{m_name}")
```
- [x] Run inference via `model.predict`

```
from fiftyone.utils.torch import _load_image
for m_name in zoo_models:
    model = foz.load_zoo_model(m_name)
    for sample in dataset_person:
        kpts = model.predict(img=_load_image(sample.filepath, True, True), sample=sample)
        sample[f"kpts_predict_{m_name}"] = kpts
        sample.save()
```

- [x] Run inference via HF interface and add results to FO dataset
```
from transformers import AutoImageProcessor, VitPoseForPoseEstimation
import torch

image_processor = AutoImageProcessor.from_pretrained("usyd-community/vitpose-base-simple")
model = VitPoseForPoseEstimation.from_pretrained("usyd-community/vitpose-base-simple").to(device)

for sample in dataset_person:
    image = _load_image(sample.filepath, True, True)
    detections = sample.get_field("ground_truth")
    height, width = image.shape[:2]
    boxes = _get_image_boxes((height, width), detections.detections)

    inputs = image_processor(image, boxes=[boxes], return_tensors="pt").to(model.device)
    with torch.no_grad():
        outputs = modell(**inputs)

    # Post-process outputs to generate keypoints

    sample["kpts_hf_vitpose-base-simple-torch"] = fol.Keypoints(keypoints=keypoints)
    sample.save()
```

- [x] Compare outputs from the above runs against one another (should be the same)

```
kpts_apply_model = dataset_person.values(f"kpts_person_{model_name}")
kpts_predict_inf = dataset_person.values(f"kpts_predict_{model_name}")
kpts_hf_inf = dataset_person.values(f"kpts_hf_{model_name}")

# Compare keypoints using np.allclose
```

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

* **New Features**
  * Added pose-estimation support in the Transformers integration with automatic model-type detection.
  * Outputs now include normalized keypoints with per-keypoint confidence scores.
  * Inference supports optional bounding-box prompts and dataset-index routing for mixture-of-experts models.
  * Pose-estimation results are surfaced in standard output formats alongside existing tasks.

* **Documentation**
  * Model zoo docs generation updated to handle pose-estimation models when combined with Transformers.

### DIFF
--- a/fiftyone/utils/transformers.py
+++ b/fiftyone/utils/transformers.py
@@ -1151,7 +1151,6 @@ class PoseEstimationGetItem(fout.GetItem):
             {
                 "images": img,
                 "boxes": [img_boxes],
-                "fo_image_size": [(height, width)],
             }
         )
 
@@ -1232,7 +1231,6 @@ class FiftyOneTransformerForPoseEstimation(
             args = {
                 "images": args,
                 "boxes": images_boxes,
-                "fo_image_size": images_hw,
             }
             args = self.collate_fn(self.transforms(args))
 
@@ -1400,8 +1398,6 @@ class _HFTransformsHandler:
         if isinstance(args, dict):
             # multiple inputs
             if self.return_image_sizes:
-                if args.get("fo_image_size"):
-                    image_size = args.pop("fo_image_size")
                 if not image_size and args.get("images", None) is not None:
                     image_size = (
                         [_get_image_size(img) for img in args["images"]]


CharlesCNorton (approved):

fiftyone/utils/torch.py
```diff
@@ -1513,6 +1513,62 @@ def _parse_output(self, output, frame_size, confidence_thresh):
         return fol.Detections(detections=detections)
 
 
+class KeypointOutputProcessor(OutputProcessor):
```
jacobsela, line 1516: This seems to be HF model specific, maybe it should go in the integration's file?

fiftyone/utils/transformers.py
```diff
@@ -1070,6 +1077,166 @@ def __init__(self, config):
...
+
+class PoseEstimationGetItem(fout.ImageGetItem):
+    """A :class:`GetItem` that loads images and bounding boxes to feed to
+    :class:`FiftyOneTransformerForPoseEstimation` instances."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def __call__(self, d):
+        img = fout._load_image(
```
jacobsela, line 1131: maybe use `super` here instead of rewriting image loading code?

fiftyone/utils/transformers.py
```diff
@@ -1070,6 +1077,166 @@ def __init__(self, config):
...
+            d["filepath"],
+            use_numpy=self.use_numpy,
+            force_rgb=True,
+        )
+        boxes = d["box_prompts"]
```
jacobsela: 💪

fiftyone/utils/transformers.py
```diff
@@ -1070,6 +1077,166 @@ def __init__(self, config):
...
+
+        return self.transform({"images": img, "boxes": boxes})
+
+    @property
+    def required_keys(self):
+        return ["filepath", "box_prompts"]
+
+
+class FiftyOneTransformerForPoseEstimation(
+    fout.TorchSamplesMixin, FiftyOneTransformer
```
jacobsela, line 1163: We generally want to be _either_ (i.e. XOR) a `SamplesMixin` or a `GetItem` based model. Since we already get the box prompts we want in `fiftyone.utils.transormers.PoseEstimationGetItem`, we don't need to also get the samples to have the values we need right?

fiftyone/utils/transformers.py
```diff
@@ -1070,6 +1077,166 @@ def __init__(self, config):
...
+                prompts.append(value)
+            else:
+                raise ValueError(
+                    "Sample %s is missing a prompt in field '%s'"
+                    % (sample.id, field_name)
+                )
+
+        return prompts
+
+    def _get_field(self):
```
jacobsela, line 1215: what is the purpose of this?

fiftyone/utils/transformers.py
```diff
@@ -1070,6 +1077,166 @@ def __init__(self, config):
...
+                image_sizes,
+                confidence_thresh=self.config.confidence_thresh,
+                box_prompts=boxes,
+            )
+        return output
+
+    def build_get_item(self, field_mapping=None):
+        if field_mapping is None:
+            field_mapping = {}
+        if "box_prompts" not in field_mapping:
```
jacobsela: We shouldn't inject values for the field mapping for the user, this is what the `field_mapping` is for.

fiftyone/utils/transformers.py
```diff
@@ -1070,6 +1077,166 @@ def __init__(self, config):
...
+            return self._output_processor(
+                output,
+                image_sizes,
+                confidence_thresh=self.config.confidence_thresh,
+                box_prompts=boxes,
+            )
+        return output
+
+    def build_get_item(self, field_mapping=None):
+        if field_mapping is None:
```
jacobsela: FYI this is already handled internally

fiftyone/utils/transformers.py
```diff
@@ -1070,6 +1077,166 @@ def __init__(self, config):
...
+        return PoseEstimationGetItem(
+            transform=self._transforms,
+            raw_inputs=self.config.raw_inputs,
+            using_half_precision=self._using_half_precision,
+            use_numpy=False,
+            field_mapping=field_mapping,
+        )
+
+    @staticmethod
+    def collate_fn(batch):
```
jacobsela: why do we need a custom `collate_fn` here?

fiftyone/utils/transformers.py
```diff
@@ -1200,7 +1372,29 @@ def __call__(self, args):
                         if isinstance(args["images"], list)
                         else [_get_image_size(args["images"])]
                     )
+            if args.get("boxes") is not None:
```
jacobsela: Can we put this logic in the `GetItem` instead? Truthfully all of the image size handling (code right before this new code) should be in the GetItem, it was my mistake to put it here.

fiftyone/utils/transformers.py
manushreegangwar: `transform` needs [boxes](https://huggingface.co/docs/transformers/model_doc/vitpose#transformers.VitPoseImageProcessor.preprocess). Can't use `super` without applying transform on images unless I unset `self.transform`

fiftyone/utils/transformers.py
manushreegangwar: May be I am missing something but how would the users apply `model.predict` on a single image if we don't have `SamplesMixin`?

fiftyone/utils/transformers.py
manushreegangwar: `dataset.apply_model(model, prompt_field="ground_truth", label_field="predictions")`

This is to get the "prompt_field" input. Following similar pattern as `sam` integration here. Not needed for getitem but used when calling `model.predict`

manushreegangwar: I can change the `else` to raise an error if we only want to support "prompt_field" as the key.

fiftyone/utils/transformers.py
manushreegangwar: I can remove this but that would mean we run inference as:
`dataset.apply_model(model, field_mapping={"box_prompts" : "ground_truth"} , label_field="predictions")`

I personally prefer:
`dataset.apply_model(model, prompt_field="ground_truth", label_field="predictions")`
and internally create the field_mapping  but I understand how that doesn't fit with the intended design for `field_mapping`.

fiftyone/utils/transformers.py
manushreegangwar: Sure, I will move this. I was divided between putting it here vs getitem.

fiftyone/utils/torch.py
manushreegangwar: Seems generic enough to be here since it doesn't use any HF data structures. Very similar to `KeypointDetectorOutputProcessor`. I thought about modifying `KeypointDetectorOutputProcessor` but seemed cleaner not to. I can merge the two if you feel strongly about not adding a new output processor here.

fiftyone/utils/transformers.py
manushreegangwar: Using boxes as a `list` was throwing an error in the `super` collate_fn

fiftyone/utils/torch.py
jacobsela: No strong feelings just bringing it up for due diligence.

fiftyone/utils/transformers.py
jacobsela: I see, in that case let's remove the inheritance from `ImageGetItem` just to avoid confusion. Am I missing some shared code between these two `GetItem`s?

fiftyone/utils/transformers.py
jacobsela:
```
input = {k : sample[v] for k, v in field_mapping.items()}
output = model.predict(input)
```

It may feel pretty ugly, but it is what it is for now. I'll note two things:
1. we can move the logic of `{k : sample[v] for k, v in field_mapping.items()}` from `FiftyOneTorchDataset` to `GetItem` and it would already feel a lot cleaner. I wish I had done this originally.
2. consolidating all models to have inference that looks like `output = model.predict(sample)` is the essence of the "Arbitrary Model Input" work package. I'm dying to do it already but haven't been able to get the traction yet. Please promote this!

fiftyone/utils/transformers.py
jacobsela: no all good, so this is effectively `field_mapping` : `_get_field`  <----> `GetItem` : `SamplesMixin`

fiftyone/utils/transformers.py
jacobsela: actually the syntax you wrote `dataset.apply_model(model, prompt_field="ground_truth", label_field="predictions")` does work! `apply_model` takes unused kwargs and constructs a `field_mapping` from them.

jacobsela: see [here](https://github.com/voxel51/fiftyone/blob/01cd8905228a9f026bf8e8bb9de3be729119b8c3/fiftyone/core/models.py#L182)

jacobsela: btw those few lines really exemplify the overlap between `SamplesMixin` and `GetItem`

fiftyone/utils/transformers.py
jacobsela: Ok I see a two options:
1. in `FiftyOneTransformerForPoseEstimation.collate_fn` pop 'boxes' and whatever else, handle them, and call `super` on the rest. Less pretty but faster.
2. Modify `FiftyOneTransformer.collate_fn` to also handle non-tensor values. Prettier but more work.

`FiftyOneTransformer.collate_fn` pads ragged batches dynamically so it's helpful for enabling batching where we otherwise wouldn't be able to use it. That said I don't love the implementation but I think less branches is better assuming we want to rewrite this stuff anyways.

fiftyone/utils/transformers.py
jacobsela: If you want to take it upon yourself to do the same for the other image size handling stuff this may be a good opportunity (no pressure obviously)

jacobsela: btw really great work. You're the first person that's not me to implement a `GetItem`! I feel like we should have a celebration of some sort.

docs/scripts/make_model_zoo_docs.py
```diff
@@ -227,6 +227,14 @@
     dataset.apply_model(model, label_field="predictions")
 
     session = fo.launch_app(dataset)
+
```
brimoor: @manushreegangwar be careful about adding empty lines in this file: they are literally added to the example code in the published docs 💀 

In this case, can you please remove the empty line after the two `launch_app()` lines affected by this diff? For consistency with the other `if` blocks you see in this file

fiftyone/utils/transformers.py
manushreegangwar: Changed to inherit from `GetItem`.

docs/scripts/make_model_zoo_docs.py
manushreegangwar: Good catch. I will remove the empty lines.

I wasn't noticing them in the built docs though. Did something change?
image

fiftyone/utils/torch.py
manushreegangwar: I will leave this here then

fiftyone/utils/transformers.py
```diff
@@ -721,14 +732,20 @@ def collate_fn(batch):
             for bdict in batch:
                 t = bdict[k]
                 pad_amounts = []
-                for d in reversed(range(len(t.shape))):
+                for d in reversed(range(1, len(t.shape))):
```
manushreegangwar, line 735: Added due to https://github.com/huggingface/transformers/blob/v4.57.0/src/transformers/models/vitpose/image_processing_vitpose.py#L532

fiftyone/utils/transformers.py
manushreegangwar: I haven't removed inheritance from SamplesMixin. It sounds like we need to do some work to make the interface better first. I can add a ticket for this refactoring if one doesn't exist yet.

fiftyone/utils/transformers.py
manushreegangwar: I added https://github.com/voxel51/fiftyone/pull/6371/files#diff-13100db40bf4740e3fc3ea44ae6c6d9498102a7da42b58a9a539173965e1dc24R179 because these models support get item and also fall under samples mixin for now. Lmk if there's a reason not to add that line.

fiftyone/utils/transformers.py
manushreegangwar: Updated the existing collate fn

fiftyone/utils/transformers.py
manushreegangwar: I haven't moved the other image size stuff out of here in this PR. The change would spill into `ImageGetItem` and don't want to conflate this PR with that.

fiftyone/utils/transformers.py
jacobsela: I think this consolidation is quite important. I don’t think the interface needs to be updated for this to happen. What changes are you interested in?

manushreegangwar: I agree that the consolidation is important and would improve the quality of model inference implementation. Here are my comments: 

1. If we do consolidation inside `FiftyOneTransformerForPoseEstimation` and use:
```
input = {k : sample[v] for k, v in field_mapping.items()}
# convert input to {"images": img, "boxes": [img_boxes]}
output = model.predict(input)
```
We expect the users to understand how to create `input` for each model, i.e, how to convert boxes from relative to absolute coordinates, load images from filepath etc. The `input` to `model.predict` should look like `{"images": img, "boxes": [img_boxes]}` for it to transform correctly during preprocessing. Or we take care of processing sample fields inside `def _predict_all` as it is currently for samplesmixin. I don't see the benefit of removing `SamplesMixin` in this case. Lmk if I misunderstood your suggestion here.

2. We can expect the users to use `model.build_get_item` instead of the above. However, this isn't documented in the FO docs yet. If we use getitem, then we get preprocessed inputs and therefore, preprocessing is not required during `model.predict`. Users need to know about this.

3. `output = model.predict(sample)` would be awesome! Let's discuss this.

jacobsela: 2 is what I was going for, sorry I just realized I sent you a very incomplete snippet (the one you reference in 1).

Not going to block on this for now if you are interested in doing 3, could be a nice first application to refactor this model with the code from 3.

Let's schedule some time to talk?

fiftyone/utils/transformers.py
jacobsela: makes sense, can we get a test for a few key models before merge?

fiftyone/utils/transformers.py
```diff
@@ -721,14 +732,20 @@ def collate_fn(batch):
...
                     diff = max_dims[d] - t.shape[d]
                     pad_amounts.extend([0, diff])  # (left_pad, right_pad)
                 padded_tensors.append(torch.nn.functional.pad(t, pad_amounts))
 
             # Concatenate along the first dimension
             res[k] = torch.cat(padded_tensors, dim=0)
 
+        for k in non_tensor_keys:
```
jacobsela, line 743: 👍

fiftyone/core/models.py
```diff
@@ -176,6 +176,7 @@ def apply_model(
     if supports_get_item and needs_samples:
         field_mapping = kwargs.pop("field_mapping", {})
         needs_fields = kwargs.pop("needs_fields", {})
+        field_mapping.update(kwargs)
```
jacobsela: This may potentially cause errors with other combo `SupportsGetItem` and `SamplesMixin`. AFAIK The only other one we have right now is SAM. Since that model is pretty important right now can we test it before a merge?

fiftyone/utils/transformers.py
```diff
@@ -1070,6 +1088,178 @@ def __init__(self, config):
...
+        if "prompt_field" in self.needs_fields:
+            prompt_field = self.needs_fields["prompt_field"]
+        else:
+            prompt_field = next(iter(self.needs_fields.values()), None)
+
+        return prompt_field
+
+    def _predict_all(self, args):
+        if self.preprocess:
+            images_hw = [_get_image_size(img) for img in args]
```
jacobsela, line 1225: Noting here that this makes the same mistake I made when first implementing this integration. This kind of preprocessing should be in the `GetItem`. Not blocking on this, mainly writing this to convey where we want to go. This is important for `model.predict(sample)`

fiftyone/utils/transformers.py
```diff
@@ -1191,16 +1385,21 @@ def __init__(self, processor, return_image_sizes=False, **kwargs):
         self.text_per_image = False  # passed in by model after init
 
     def __call__(self, args):
+        image_size = None
         if isinstance(args, dict):
             # multiple inputs
             if self.return_image_sizes:
-                if args.get("images", None) is not None:
+                if args.get("fo_image_size"):
```
jacobsela: Can you explain why this is needed?

fiftyone/core/models.py
manushreegangwar: Right. I reverted this since SAM models don't use getitem for box prompts and we cannot have fields in field_mapping that are not required by getitem.

fiftyone/utils/transformers.py
manushreegangwar: For sure. This will all go away once we default to using getitem.

fiftyone/utils/transformers.py
manushreegangwar: I missed adding this to poseestimatinogetitem. It should make sense now. Once we move the image sizes to `ImageGetItem` or a derived class for transformers, we can remove the `if` in line 1406

jacobsela: I'm saying that the image sizes are already calculated here, why do we need another if? We should either do this computation here, as was done previously, or in the GetItem (for all models).

manushreegangwar: Once this PR is merged, I will create a new one that does the refactoring related to `Once we move the image sizes to ImageGetItem or a derived class for transformers, we can remove the if in line 1406`. I am going to create a derived class for `ImageGetItem` in HF integration and add image sizes to that. 

Since this is not directly related to this PR, I want to do it in a separate PR.

jacobsela: I'm saying it doesn't make sense to do a half solution. Either you use the image size loading here OR in the GetItem, for all cases. Having another branch here that checks if `fo_image_size` is actually already computed, and calculating it if it's not is making existing code messier without fixing anything.

manushreegangwar: Removed but now we are computing image dimensions twice for PoseEstimation models. Not ideal. Will reintroduce this change with the rest of the GetItem change

manushreegangwar:
@jacobsela can you review this again?

PR #6368, opened 2025-09-29
Add patch sample endpoint in embedded server

CamronStaley:
## Reason

Allow deleting and upserting fields on sample via api call

## Testing

Started my server with:
```
python fiftyone/server/main.py
```

Then I ran this test script:
```
import requests
import fiftyone as fo

DATASET_NAME = "api-delete-test"

if DATASET_NAME in fo.list_datasets():
    fo.delete_dataset(DATASET_NAME)

dataset = fo.Dataset(DATASET_NAME)
dataset.persistent = False

sample = fo.Sample(filepath="/some/random/sample.jpg", tags=["initial"])

dataset.add_sample(sample)
dataset.save()

print(f"Created dataset '{DATASET_NAME}' for testing.")
print(f"Targeting sample ID: {sample.id}\n")

url = f"http://localhost:5151/dataset/{dataset._doc.id}/sample/{sample.id}"

patch_payload = {
    "ground_truth_2": {
        "_cls": "Detections",
        "detections": [
            {
                "_cls": "Detection",
                "label": "cat",
                "bounding_box": [
                    0.15,
                    0.15,
                    0.25,
                    0.25,
                ],
                "confidence": 0.99,
            },
            {
                "_cls": "Detection",
                "label": "cat",
                "bounding_box": [
                    0.15,
                    0.15,
                    0.25,
                    0.25,
                ],
                "confidence": 0.99,
            }
        ],
    },
    "ground_truth": {
        "_cls": "Detections",
        "detections": [
            {
                "_cls": "Detection",
                "label": "cat",
                "bounding_box": [
                    0.15,
                    0.15,
                    0.25,
                    0.25,
                ],
                "confidence": 0.99,
            }
        ],
    },
    "reviewer": "John Doe",
    "tags": None,
    "weather": {
        "_cls": "Classification",
        "label": "sunny",
        "confidence": 0.99,
    },
}

print("Sending PATCH request to delete labels...")
response = requests.patch(url, json=patch_payload)
print("Response from server:")
print(response.json())
response.raise_for_status()
print("-" * 30)
```

Here is the corresponding sample:

```
/** 
* Paste one or more documents here
*/
{
  "filepath": "/some/random/sample.jpg",
  "_media_type": "image",
  "_rand": 0.9990001799678636,
  "_dataset_id": {
    "$oid": "68dd9cf0d6c6359ceb1da8ba"
  },
  "created_at": {
    "$date": "2025-10-01T21:28:16.812Z"
  },
  "last_modified_at": {
[...]

PR #6372, opened 2025-10-01
Implement basic annotation client

tom-vx51:

This PR implements a basic frontend client for interacting with the endpoint added in #6368.

local

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Client-side sample patching with structured per-field updates and a dedicated patch request/response flow.
  * Configurable fetch behavior with richer result types and optional per-request error handlers.

* **Bug Fixes / Error Handling**
  * Improved mapping of HTTP error responses to specific error types for clearer diagnostics.

* **Chores**
  * Utility to build encoded API paths.
  * Broadened public exports to make client and link utilities more accessible.

### DIFF
--- a/app/packages/core/src/client/annotationClient.ts
+++ b/app/packages/core/src/client/annotationClient.ts
@@ -1,4 +1,10 @@
 import { encodeURIPath } from "./util";
+import { Sample } from "@fiftyone/looker";
+import {
+  MalformedRequestError,
+  NetworkError,
+  NotFoundError,
+} from "@fiftyone/utilities";
 
 /**
  * Label types which currently support mutation via {@link patchSample}.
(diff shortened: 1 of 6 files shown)


app/packages/core/src/client/annotationClient.ts
```diff
@@ -0,0 +1,60 @@
...
+  | "Classifications"
+  | "Detection"
+  | "Detections"
+  | "Polyline"
+  | "Polylines";
+
+/**
+ * A field specification can either be
+ *  - `null` indicating the field should be deleted, OR
+ *  - an object containing at least the `_cls` field to specify the label type
```
CamronStaley: Technically _cls wouldn't always be required if it's a primitive label on the sample itself. You could leave this, but below you'd need to make the Record not require a value of FieldSpecification

tom-vx51: good point. I'll see if I can make the type make sense without being so broad it's useless

app/packages/core/src/client/annotationClient.ts
```diff
@@ -0,0 +1,108 @@
...
+ *  - an object containing at least the `_cls` field to specify the label type, OR
+ *  - an arbitrary mapping to primitives or objects
+ */
+export type FieldSpecification = NullField | LabelField | AttributeField;
+
+export type PatchSampleRequest = {
+  datasetId: string;
+  sampleId: string;
+  delta: Record<string, FieldSpecification>;
+};
```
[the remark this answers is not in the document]
tom-vx51: `delta` entries indeed must not be simple scalars

app/packages/core/src/client/annotationClient.ts
```diff
@@ -0,0 +1,108 @@
...
+  const httpResponse = await fetch(
+    encodeURIPath(["dataset", request.datasetId, "sample", request.sampleId]),
+    {
+      method: "PATCH",
+      body: JSON.stringify(request.delta),
+      headers: {
+        "content-type": "application/json",
+      },
+    }
+  );
```
benjaminpkane: You'll need to use `getFetchFunction` so a properly constructed request is made (through a proxy, etc.)

PR #6374, opened 2025-10-02
Allow `compute_embeddings()` to write to embedded fields

brimoor:
## Change log

- Allow `compute_embeddings()` to write to embedded fields

## Tested by

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart", max_samples=10)

dataset.add_sample_field(
    "data",
    fo.EmbeddedDocumentField,
)

# this already works
dataset.apply_model(model, label_field="data.predictions")

# this didn't, but now it does!
dataset.compute_embeddings(model, embeddings_field="data.embeddings")
```

## Summary by CodeRabbit

- New Features
  - Support nested (dot-qualified) embedding fields across images, frames, videos, and patches, writing results to the correct location.

- Performance
  - Faster inference and embedding workflows by loading only the fields required for each operation, reducing I/O and memory use.

- Bug Fixes
  - Resolved errors when using nested embedding fields and ensured valid field constraints during processing.

- Refactor
  - Consolidated field-loading logic across inference and embedding paths for more consistent behavior without changing the public API.

### DIFF
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -1025,13 +1025,6 @@ def compute_embeddings(
             embeddings_field
         )
 
-        if "." in embeddings_field:
-            ftype = "frame" if _is_frame_field else "sample"
-            raise ValueError(
-                "Invalid `embeddings_field=%s`. Expected a top-level %s field "
-                "name that contains no '.'" % (embeddings_field, ftype)
-            )
-
         if dataset.media_type == fom.VIDEO and model.media_type == "image":
             if not dataset.has_frame_field(embeddings_field):
                 dataset.add_frame_field(embeddings_field, fof.VectorField)
@@ -1097,7 +1090,11 @@ def compute_embeddings(
 def _compute_image_embeddings_single(
     samples, model, embeddings_field, skip_failures, progress
 ):
-    samples = samples.select_fields()
+    if embeddings_field is not None and "." in embeddings_field:
+        root_field = embeddings_field.split(".", 1)[0]
+        samples = samples.select_fields(root_field)
+    else:
+        samples = samples.select_fields()
 
     embeddings = []
     errors = False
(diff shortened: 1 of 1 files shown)


fiftyone/core/models.py
```diff
@@ -1097,7 +1090,11 @@ def compute_embeddings(
 def _compute_image_embeddings_single(
     samples, model, embeddings_field, skip_failures, progress
 ):
-    samples = samples.select_fields()
+    if embeddings_field is not None and "." in embeddings_field:
+        root_field = embeddings_field.split(".", 1)[0]
+        samples = samples.select_fields(root_field)
+    else:
+        samples = samples.select_fields()
```
exupero: These 6 added blocks are all the same. Can we move them to a function?

brimoor: I was following the existing pattern of duplicating the logic in each of the `_compute_XXX_for_YYY()` methods, but, sure, just pushed https://github.com/voxel51/fiftyone/pull/6374/commits/7a9dba3bb5e63065ad755981353d2bfe78db68e7 that refactors all of these into shared utils.

jacobsela: @exupero FYI consolidating all of these functions into 1 is one of the big payoffs of the "arbitrary model input" work package.

brimoor: ^which I am at once excited by in theory and also not convinced that complexity will be removed, merely moved (eg into a class hierarchy rather than functional branches)

At the end of the day, there must be different code to handle:
- are we running batch inference (`predict_all()`) or individual inference (`predict()`)?
- are we processing images, videos, or frames?
- are data loaders used or not?

You need not respond here. 

> *"in god we trust, all others bring [pseudocode demonstrating the proposal]"*

exupero (approved):
Code looks good to me; didn't test.

PR #6384, opened 2025-10-06
FOEPD-2219: Disable async DB writes in `apply_model` and `compute_embeddings`

exupero:

Reverts DB writes in `apply_model` and `compute_embeddings` to happening on the main thread, rather than running on a background thread (introduced in #6288) because it can result in unbounded memory usage as tasks back up in the queue under DB pressure.

Ran a benchmarking script and compared memory usage from before #6288, the latest develop, and with these changes. Memory usage with these changes returns to being flat:
visualization(12)

        notes for FiftyOne users.

-   [ ] Other


kaixi-wang (approved):
lgtm. 
fwiw, looking at the code, there might be something in the save context that is causing memory to increase significantly when we queue up writes because the context itself appears to also queue save operations by batch size and then reload the docs in memory

PR #6387, opened 2025-10-07
Stop propagation of the Escape event when looker is a Modal [FOEPD-862]

jleven:
## TODO

- [x] Confirm that the way I'm checking `isModal` is the right way to do it.

Two changes:
1. Stops propagation of the Escape keyDown event when looker is a Modal
2. Using the Escape key to clear selections in the grid now first opens a confirmation modal

image

Clicked around in the app:
1. Select samples in the grid
2. Open modal
3. Close modal with escape key
4. Verify samples are still selected

Selecting and Clearing:
1. Select samples in the grid
2. Press escape key
3. Verify confirmation dialog pops up
4. Verify "cancel" does not clear selections / Verify "ok" does clear selections

When nothing is selected:
1. Make sure no samples are selected in the grid
2. Press escape key
3. Verify no confirmation dialog pops up

        notes for FiftyOne users.

Fixed bug where grid selections would be cleared when closing a modal using the escape key. Also added a confirmation dialog when clearing grid selections using the escape key.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Adds a confirmation dialog before clearing selected items to prevent accidental loss.

* **Bug Fixes**
  * Pressing Escape while a modal is open no longer triggers actions in the underlying grid.
  * Escape-based clearing of selections now requires explicit confirmation.

* **Chores**
  * Adds an optional modal flag in component configuration to enable modal-aware keyboard handling.

* **Tests**
  * End-to-end test updated to auto-accept the confirmation dialog during Escape-driven selection clearing.

### DIFF
--- a/e2e-pw/src/oss/specs/selection.spec.ts
+++ b/e2e-pw/src/oss/specs/selection.spec.ts
@@ -71,7 +71,7 @@ test.describe.serial("selection", () => {
       // verify selection clears on escape
       await grid.toggleSelectFirstSample();
       await grid.assert.isSelectionCountEqualTo(1);
-      page.on("dialog", (dialog) => dialog.accept());
+      page.once("dialog", (dialog) => dialog.accept());
       await page.press("body", "Escape");
       await grid.assert.isSelectionCountEqualTo(0);
 


app/packages/core/src/components/Grid/useEscape.ts
```diff
@@ -18,7 +18,16 @@ const useEscape = () => {
           }
 
           const modal = await snapshot.getPromise(fos.modalSelector);
-          modal === null && reset(fos.selectedSamples);
+          // TODO: modal is always `null` here right after a modal closes, so this isn't the condition we want
+          if (modal === null) {
+            if (
+              confirm(
+                "You are about to clear all selections. This cannot be undone. Are you sure?"
```
brimoor: nit: can we use this copy?

```
Are you sure you want to clear your current selection?
```

brimoor (commented):
Just gave this a test locally and have a couple comments:
1. Can we please update to *not* show the dialog box if there is no current selection?
2. (for technical reviewers) may need to confirm that the dialog box strategy works in FOE too

app/packages/core/src/components/Grid/useEscape.ts
```diff
@@ -18,7 +18,17 @@ const useEscape = () => {
           }
 
           const modal = await snapshot.getPromise(fos.modalSelector);
-          modal === null && reset(fos.selectedSamples);
+          const selectedSampleIds = await snapshot.getPromise(
+            fos.selectedSamples
+          );
+          // TODO: modal is always `null` here right after a modal closes, so this isn't the condition we want
+          if (modal === null && selectedSampleIds.size > 0) {
```
benjaminpkane, line 25: I'm not sure I understand. This seems to work. I assume only if propagation is stopped in looker?

jleven: I think `modal === null` was put here to try to accomplish what this PR is _actually_ accomplishing. Unfortunately, `modal` is always `null` here, so we do need this PR.

PR #6389, opened 2025-10-07
FOEPD-2119: Use SaveContext to write to DB in background thread

exupero:

#6288 performed DB writes on a background thread but showed increasing memory usage as batches were accumulated, so DB writes were moved back to the main thread in #6384. Testing on #6361 showed that limiting the number of tasks in the async queue did not reduce memory usage when computing embeddings with a large model.

This PR takes a different approach to putting DB writes on an async thread by using a subclass of `SaveContext` that submits saves to an async executor. Profiling shows memory usage remains flat while execution also finishes sooner, even when using only 1 async worker thread.

Ran a memory profiling script that computes embeddings with a large model, using 500, 1000, and 2000 samples. Using `async_executor` to write to the DB resulted in increasing memory usage, while using `AsyncSaveContext` did not and is faster, and using more async workers neither changed memory usage nor total execution time:
visualization(15)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Optional asynchronous background batching for save operations to improve throughput and responsiveness; image-model processing paths now enable async writes.

- **Refactor**
  - Per-instance concurrency controls and executor support added to make batched queuing thread-safe under async batching.

- **Bug Fixes**
  - Background save tasks are drained and awaited on context exit; errors from background writes are surfaced and failed batches retried to reduce lost writes.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -88,6 +88,9 @@ class _DummyFuture:
     def result(self):
         return self.value
 
+    def done(self):
+        return True
+
 
 class _DummyExecutor:
     def __enter__(self):
@@ -308,6 +311,17 @@ class SaveContext(object):
                 sample._reload_parents()
 
     def _save_batch(self):
+        pending = []
+        for future in self.futures:
+            if not future.done():
+                pending.append(future)
+            else:
+                try:
+                    future.result()
+                except Exception:
+                    pending.append(future)  # re-raise in __exit__
+        self.futures = pending
+
         future = self.executor.submit(self._do_save_batch)
         self.futures.append(future)
 


fiftyone/core/collections.py
```diff
@@ -234,6 +234,26 @@ def _save_batch(self):
             self._reload_parents.clear()
 
 
+class AsyncSaveContext(SaveContext):
```
jacobsela: 💪

fiftyone/core/collections.py
```diff
@@ -234,6 +234,26 @@ def _save_batch(self):
...
+        self.executor.__enter__()
+        return self
+
+    def __exit__(self, *args):
+        super().__exit__(*args)
+        self.executor.__exit__(*args)
+
+    def _save_batch(self):
+        self.executor.submit(super()._save_batch)
+
```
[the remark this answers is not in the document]
brimoor: @exupero I haven't thought deeply about this, but I think coderabbit makes a valid point here

brimoor (commented):
Makes intuitive sense to me that offloading DB writes to a worker thread would be faster! Yay!

I'd like to get this behavior into the public interface:

```py
for sample in dataset.iter_samples(..., autosave=True, **kwargs):
    ...

for sample in dataset.iter_groups(..., autosave=True, **kwargs):
    ...

for sample in dataset.save_context(**kwargs):
    ...

# note that this is imported into the public `fiftyone` namespace
with fo.SaveContext(dataset, **kwargs) as ctx:
    for sample in dataset:
        ..
```

not just used internally in the `fiftyone.core.models` module.

Moreover (assuming this is fully tested and stable), given that it is more performant, I think we should consider making it the *default behavior*.

Today, the `**kwargs` in the public interface let users configure save batching if desired via `batch_size` and `batching_strategy` on a per-call basis. And there are environment variables that allow users to configure the default autosave batching behavior across an entire session:

```
FIFTYONE_DEFAULT_BATCHER
FIFTYONE_BATCHER_STATIC_SIZE
FIFTYONE_BATCHER_TARGET_SIZE_BYTES
FIFTYONE_BATCHER_TARGET_LATENCY
```

Some options:

1. *Always* use async writes, because it is strictly better
2. Make async writes configurable via the above patterns by adding:

```
kwarg: async_writes=True/False
config: FIFTYONE_SAVE_CONTEXT_ASYNC_WRITES=True/False
```

In 2, assuming async is strictly better, I'd advocate for the default being `async_writes=True`.

fiftyone/core/collections.py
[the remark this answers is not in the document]
brimoor: Note that I am making some small tweaks to `SaveContext` in https://github.com/voxel51/fiftyone/pull/4773, so it would be great to get that merged first 😄

fiftyone/core/models.py
```diff
@@ -477,7 +478,11 @@ def _apply_image_model_data_loader(
 
     with contextlib.ExitStack() as context:
         pb = context.enter_context(fou.ProgressBar(samples, progress=progress))
-        ctx = context.enter_context(foc.SaveContext(samples))
+        ctx = context.enter_context(
+            foc.AsyncSaveContext(
+                samples, executor=ThreadPoolExecutor(max_workers=1)
+            )
+        )
```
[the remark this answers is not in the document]
exupero: Done.

exupero:
See #6391

fiftyone/core/collections.py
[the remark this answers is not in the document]
exupero: CodeRabbit's suggestion caused increasing memory usage, so I added a thread lock in `save` and `_save_batch` and confirmed samples aren't dropped.

fiftyone/core/collections.py
```diff
@@ -234,6 +235,39 @@ def _save_batch(self):
...
+        super().__enter__()
+        self.executor.__enter__()
+        return self
+
+    def __exit__(self, *args):
+        for future in self.futures:
+            future.result()
+        super().__exit__(*args)
+        self.executor.__exit__(*args)
+
```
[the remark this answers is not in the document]
exupero: Done.

fiftyone/core/collections.py
```diff
@@ -234,6 +235,137 @@ def _save_batch(self):
...
+        if self._batch_ids and self._is_generated:
+            with self.batch_ids_lock:
+                self.sample_collection._sync_source(ids=self._batch_ids)
+                self._batch_ids.clear()
+
+        if self._reload_parents:
+            with self.reloading_lock:
+                for sample in self._reload_parents:
+                    sample._reload_parents()
+
```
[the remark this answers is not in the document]
exupero: Copying the data causes memory usage to increase, so lock contention creates backpressure that slows the producer loop down to avoid consuming more memory. Even with the high likelihood of lock contention, my testing shows there is an improvement in overall run time.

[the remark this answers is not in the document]
exupero: No, you were right, I can copy the data in the locks and not lock the `bulk_write()` call. Profiling shows it doesn't use additional memory. My previous experiments with copying the data to make it thread-safe did result in more memory usage, but doing it directly within a method and not using a function closure probably helped the garbage collector know when it could clean up those objects, rather than keeping them around indefinitely. And/or introducing locks creates sufficient backpressure to avoid accumulating futures.

fiftyone/core/collections.py
```diff
@@ -234,6 +235,140 @@ def _save_batch(self):
...
+        self.executor.__enter__()
+        return self
+
+    def __exit__(self, *args):
+        super().__exit__(*args)
+        for future in self.futures:
+            future.result()
+        self.futures.clear()
+        self.executor.__exit__(*args)
+
```
exupero, line 205: Done.

fiftyone/core/collections.py
```diff
@@ -234,6 +235,148 @@ def _save_batch(self):
...
+            # super().__exit__() are awaited.
+            while self.futures:
+                futures = self.futures
+                self.futures = []
+                for future in futures:
+                    future.result()
+            self.futures.clear()
+        finally:
+            self.executor.__exit__(*args)
+
```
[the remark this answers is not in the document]
kaixi-wang: this seems reasonable to make draining more robust

fiftyone/core/collections.py
```diff
@@ -234,6 +235,148 @@ def _save_batch(self):
...
+    def save(self, sample):
+        """Registers the sample for saving in the next batch.
+
+        Args:
+            sample: a :class:`fiftyone.core.sample.Sample` or
+                :class:`fiftyone.core.sample.SampleView`
+        """
+        if sample._in_db and sample._dataset is not self._dataset:
```
kaixi-wang, line 216: is all of this just copied from SaveContext but with the locks? Feels like a lot of shared logic/similar code

fiftyone/core/collections.py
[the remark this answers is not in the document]
exupero: Done

fiftyone/core/collections.py
exupero: Yes. What are your thoughts on resolving that? Put the locks in `SaveContext`?

brimoor: > Put the locks in SaveContext?

Yes that seems like like the best path. Both to reduce code duplication and also because per [my comments here](https://github.com/voxel51/fiftyone/pull/6389#pullrequestreview-3310732058), I see this either as (1) the way that `SaveContext` always works, or (2) a behavior I can get via `fo.SaveContext(..., async_writes=True)` and similar.

> Note that I am making some small tweaks to SaveContext in https://github.com/voxel51/fiftyone/pull/4773, so it would be great to get that merged first 😄

^reminder about this as well. More reason not to duplicate implementation of `SaveContext` and to get that PR merged first

fiftyone/core/collections.py
```diff
@@ -77,6 +79,32 @@ def registrar(func):
 aggregation = _make_registrar()
 
 
+class DummyFuture:
```
brimoor: @exupero nit: can we name these `_DummyFuture` and `_DummyExecutor` since they aren't intended for direct public usage?

From a style perspective, we have two options:
1. If the function/class is intended for public use: give it a public name and it (and all public methods of it) are fully documented
2. If the function/class is not intended for public use: give it a `_` private name and document as little as you wish!

fiftyone/core/collections.py
```diff
@@ -106,6 +134,7 @@ def __init__(
         sample_collection,
         batch_size=None,
         batching_strategy=None,
+        async_writes=False,
```
brimoor, line 137: @exupero: can you add this to the docstring? style guide says all public kwargs of public methods should be documented

fiftyone/core/collections.py
```diff
@@ -130,6 +161,20 @@ def __init__(
...
         self._last_time = None
 
+        self.samples_lock = threading.Lock()
+        self.frames_lock = threading.Lock()
+        self.batch_ids_lock = threading.Lock()
+        self.reloading_lock = threading.Lock()
+
+        self.executor = (
+            # Using more than one worker will introduce race conditions in the state preserved between DB writes, but
+            # since prediction is the bottleneck, not writing to the DB, only one worker is necessary.
```
kaixi-wang: I think save context is used in more than just prediction-related tasks. Iirc, it's in iter_samples as well

fiftyone/core/collections.py
```diff
@@ -138,11 +183,32 @@ def __enter__(self):
...
+            # Loop-drain self.futures so any submissions triggered by
+            # self._save_batch() are awaited.
+            while self.futures:
+                futures = self.futures
+                self.futures = []
+                for future in futures:
+                    try:
+                        future.result()
+                    except Exception as e:
+                        if error is None:
```
kaixi-wang, line 200: Why not raise immediately?

fiftyone/core/collections.py
```diff
@@ -198,24 +268,43 @@ def save(self, sample):
...
+            try:
+                res = foo.bulk_write(
+                    sample_ops,
+                    self._sample_coll,
+                    ordered=False,
+                    batcher=False,
+                )[0]
+                encoded_size += res.bulk_api_result.get("nBytes", 0)
+            except Exception:
+                # requeue to avoid data loss
```
kaixi-wang: Is this safe? What if some operations completed?

fiftyone/core/collections.py
exupero: Generalized comment

fiftyone/core/collections.py
exupero: Draining all futures will hopefully keep them from holding onto any values, allowing those values to be GC'd. That's my interpolation; this was originally a CodeRabbit [suggestion](https://github.com/voxel51/fiftyone/pull/6389#discussion_r2414907984) without a concrete rationale.

fiftyone/core/collections.py
exupero: Good question. I take it `foo.bulk_write` isn't transactional? Is there a way to check what samples have been written and only requeue the ones that haven't been written?

fiftyone/core/collections.py
```diff
@@ -138,11 +176,32 @@ def __enter__(self):
...
+                    except Exception as e:
+                        if error is None:
+                            error = e
+            self.futures.clear()
+        finally:
+            self.executor.__exit__(*args)
+
+        if error and (not args or args[0] is None):
+            raise error
+
```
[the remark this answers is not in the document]
exupero: I don't think we use Ruff in this project.

fiftyone/core/collections.py
exupero: Discussed with @kaixi-wang. Re-queueing leaves `self._sample_ops` and `self._frame_ops` in the same state that develop does, but because another async save might have been queued on the executor, re-queueing could cause ops to be written twice. Unfortunately, those ops are not idempotent, so the DB could end up with duplicate samples or frames. It's only incidental that develop does not clear the lists of ops when `bulk_write` fails, and we're not aware of anything that is done with data that failed to be written (if so, that would also have non-idempotency problems). Thus, it should be okay to drop failed samples. (We could write them to a different instance attribute, but we don't have anything to do with that data.)

fiftyone/core/collections.py
```diff
@@ -130,6 +154,20 @@ def __init__(
...
         self._last_time = None
 
+        self.samples_lock = threading.Lock()
+        self.frames_lock = threading.Lock()
+        self.batch_ids_lock = threading.Lock()
+        self.reloading_lock = threading.Lock()
+
+        self.executor = (
+            # Using more than one worker will introduce race conditions in the state preserved between DB writes, but
+            # since DB writes shouldn't be the bottleneck, only one worker should be necessary.
```
kaixi-wang: Isn't this comment slightly confusing as I thought the original motivation behind using async writes was that db writes were a bottleneck?

exupero: Good catch. Clarified the comment to this:

> Using more than one worker will introduce race conditions in the state preserved between DB writes, but with DB writes done on even just one background thread, the GPU becomes the bottleneck and DB writes have ample time to complete before the next batch is ready, so multiple workers are unnecessary.

fiftyone/core/collections.py
```diff
@@ -130,6 +154,21 @@ def __init__(
...
         self._last_time = None
 
+        self.samples_lock = threading.Lock()
+        self.frames_lock = threading.Lock()
+        self.batch_ids_lock = threading.Lock()
+        self.reloading_lock = threading.Lock()
+
+        self.executor = (
+            # Using more than one worker will introduce race conditions in the state preserved between DB writes, but
+            # with DB writes done on even just one background thread, the GPU becomes the bottleneck and DB writes have
```
brimoor: Now this comment assumes that `apply_model()` is the user, but in general anyone can use this via `SaveContext(..., async_writes=True)`, and hopefully we soon have enough confidence to make `async_writes=True` the **default behavior**.

I would recommend just sticking with asserting that the choice of 1 worker here is intentional as >1 worker can introduce race conditions.

exupero: Good point. Done.

PR #6402, opened 2025-10-10
add support for subscribing to execution store updates

tom-vx51:

This PR adds support for subscribing to execution store updates, primarily for the purpose of real-time communication between the server and the application.

This implementation leverages server-sent events (SSE) to set up client-server subscriptions, and uses Mongo change streams if available (with a fallback to polling if change streams are not available).

<video src="https://github.com/user-attachments/assets/5a5c770d-a401-4443-a055-475c672d3dd2"/>

unit tests, local app

        notes for FiftyOne users.

Add support for subscribing to execution store updates from the front-end application

-   [ ] Other

## Summary by CodeRabbit

- **New Features**
  - Real-time execution-store streaming via SSE with per-dataset filtering, initial state sync, and subscription health indicators.
  - New SSE-capable operators and an endpoint to stream store changes.
  - Client-side subscription hook and public subscribe/unsubscribe APIs with connection lifecycle controls.
  - Background notification service started/stopped with the app lifecycle and robust change-stream/polling fallback.

- **Tests**
  - Unit tests covering notification service behavior, dataset filtering, change handling, polling fallback, and thread-safe subscription registry.

### DIFF
--- a/fiftyone/operators/store/store.py
+++ b/fiftyone/operators/store/store.py
@@ -42,6 +42,7 @@ class ExecutionStore(object):
         store_name: str,
         dataset_id: Optional[ObjectId] = None,
         default_policy: str = "persist",
+        collection_name: Optional[str] = None,
     ) -> "ExecutionStore":
         """Creates a new execution store.
 
@@ -59,7 +60,11 @@ class ExecutionStore(object):
         # Create store service with notification service
         store_service = ExecutionStoreService(
             dataset_id=dataset_id,
-            collection_name=MongoExecutionStoreRepo.COLLECTION_NAME,
+            collection_name=(
+                collection_name
+                if collection_name
+                else MongoExecutionStoreRepo.COLLECTION_NAME
+            ),
             notification_service=default_notification_service,
         )
 


fiftyone/operators/sse.py
```diff
@@ -0,0 +1,62 @@
...
+            dark_icon=dark_icon,
+        )
+        self.store_name = store_name
+
+
+class SseOperator(foo.Operator):
+    IS_SSE_OPERATOR = True
+
+    @property
+    def subscription_config(self):
```
ritch, line 44: Why do SseOperators have to implement subscriptions? Theres other usecases for an sse operator...this is similar to executoe_as_generator.

fiftyone/operators/server.py
```diff
@@ -229,11 +230,47 @@ async def post(self, request: Request, data: dict) -> dict:
         return result.to_dict() if result else {}
 
 
+class SubscribeToExecutionStoreAsOperator(HTTPEndpoint):
```
ritch, line 242: Shouldn't this just be a way to call an SseOperator... if one just happens to implement subscriptions great. Why are we tying the two concepts together?

PR #6411, opened 2025-10-13
Patch Sample and Field endpoints

CamronStaley:

Adding endpoints to patch at the sample or field level following these guidelines: https://datatracker.ietf.org/doc/html/rfc6902

Tested locally with unit tests and calling directly to the endpoint, then verifying to state of the sample in Mongo after edits are made.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Support JSON Patch (RFC 6902) for editing samples via application/json-patch+json.
  * Endpoint to patch individual items inside list fields by item ID.

* **Improvements**
  * Unified JSON-to-model transformations with clearer "No transform registered for class ..." errors.
  * Stricter content-type handling (415) and consistent 404s for missing datasets/samples.
  * Aggregated validation error reporting for PATCH requests.

* **Tests**
  * Expanded unit tests covering JSON Patch parsing, patch methods, route behavior, and edge cases.

### DIFF
--- a/fiftyone/server/utils/jsonpatch/methods.py
+++ b/fiftyone/server/utils/jsonpatch/methods.py
@@ -29,9 +29,6 @@ def to_json_pointer(
         raise ValueError(f"Invalid JSON pointer path: {path}") from err
 
 
-def do_it(): ...
-
-
 def get(src: T, path: Union[str, jsonpointer.JsonPointer]) -> V:
     """Gets a value from an object.
 
@@ -232,7 +229,7 @@ def remove(src: T, path: Union[str, jsonpointer.JsonPointer]) -> T:
     name = pointer.parts[-1]
 
     # ensure value exists
-    get(target, f"/{name}")
+    get(target, jsonpointer.JsonPointer.from_parts([name]))
 
     try:
         if hasattr(target, "__delitem__"):


fiftyone/server/routes/sample.py
```diff
@@ -53,49 +104,30 @@ async def patch(self, request: Request, data: dict) -> dict:
...
+            result = self._handle_patch(sample, data)
+        elif content_type == "application/json-patch+json":
+            result = handle_json_patch(sample, data)
+        else:
             raise HTTPException(
-                status_code=404,
-                detail=f"Sample '{sample_id}' not found in dataset '{dataset_id}'",
+                status_code=415,
+                detail=f"Unsupported Content-Type '{content_type}'",
             )
```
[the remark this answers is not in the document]
tom-vx51: +1 to provide better compatibility with third-party clients

fiftyone/server/utils/jsonpatch/methods.py
```diff
@@ -0,0 +1,305 @@
...
+    src: T, path: Union[str, jsonpointer.JsonPointer], value: V
+) -> Union[T, V]:
+    """The "add" operation performs one of the following functions,
+    depending upon what the target location references:
+
+    o  If the target location specifies an array index, a new value is
+        inserted into the array at the specified index.
+    o  If the target location specifies an object member that does not
+        already exist, a new member is added to the object.
+    o  If the target location specifies an object member that does exist,
```
tom-vx51, line 90: just curious, is this allowed under json patch spec? or just convenient to allow

fiftyone/server/utils/jsonpatch/methods.py
```diff
@@ -0,0 +1,305 @@
...
+
+    pointer = to_json_pointer(path)
+    if not pointer.parts:
+        return value
+
+    target = get(src, jsonpointer.JsonPointer.from_parts(pointer.parts[:-1]))
+    name = pointer.parts[-1]
+
+    try:
+        if hasattr(target, "__getitem__"):
```
tom-vx51: [nit] shared logic with `get`

j053y: Yep you're right, that particular value should actually be `__setitem__`. 

If you're talking more about the logic: get, set, and del all follow a similar pattern (of checking what the source object can do), but are just different enough that I thought abstracting it would make it more confusing and harder to work on.

fiftyone/server/utils/jsonpatch/methods.py
j053y: The docstrings for each operation are directly copied from the JSONPatch spec. So anything there is valid.

PR #6416, opened 2025-10-14
Add `If-Match` header to sample patch endpoints

j053y:

Add `If-Match` header to sample patch endpoints to avoid race conditions and other multi-user update problems.

Successful sample PATCH endpoints now return the `ETag` header.

Sample PATCH endpoints now accept the `If-Match` header which can be one of the following:
  - The `ETag` header from a previous request (this is preferred as it is more flexible than the following options)
  - The sample's `last_modified_at` value as an ISO 8601 date
  - The sample's `last_modified_at` value as a UNIX timestamp

The first example will compare against the current calculated ETag, while the latter two will compare directly against the sample's `last_modified_at` and raise a `412` if the values do not match.

- [x] Add `If-Match` header support to the sample routes (with tests!)
- [x] Refactored `fiftyone.server.utils` into sub-modules to group similar areas.
- [x] Refactored `fiftyone.server.decorators` to use refactored `fiftyone.server.utils`

Run the automated tests

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

* **New Features**
  * ETag utilities and If-Match enforcement for sample and sample-field PATCH operations (responses include ETag, 412 on mismatches).
* **Refactor**
  * Centralized JSON (de)serialization and standardized JSON responses; removed legacy auto-registration transform behavior.
* **Bug Fixes**
  * More robust patch parsing, value transformation, conditional save behavior, and clearer deserialization errors.
* **Tests**
  * Migrated tests to pytest with fixtures; expanded coverage for header-driven and error scenarios.
* **Documentation**
  * Endpoint docs updated to reflect ETag/If-Match semantics.

### DIFF
--- a/fiftyone/server/routes/sample.py
+++ b/fiftyone/server/routes/sample.py
@@ -131,7 +131,8 @@ def generate_sample_etag(sample: fo.Sample) -> str:
     value = base64.b64encode(
         sample.last_modified_at.isoformat().encode("utf-8")
     ).decode("utf-8")
-    return f'"{value}"'
+
+    return utils.http.ETag.create(value)
 
 
 def save_sample(
--- a/fiftyone/server/utils/http.py
+++ b/fiftyone/server/utils/http.py
@@ -13,9 +13,17 @@ class ETag:
     """Utility class for creating and parsing ETag strings."""
 
     @staticmethod
-    def create(value: Any) -> str:
+    def create(value: Any, is_weak: bool = False) -> str:
         """Creates an ETag string from the given value."""
-        return f'"{value}"'
+        # Wrap in quotes if not already quoted
+        if not (value.startswith('"') and value.endswith('"')):
+            value = f'"{value}"'
+
+        # Add weak prefix if necessary
+        if is_weak:
+            return f'W/"{value}"'
+
+        return value
 
     @staticmethod
     def parse(etag: str) -> tuple[str, bool]:


fiftyone/server/utils/json/serialization.py
```diff
@@ -0,0 +1,70 @@
...
+            )
+
+            if cls is None:
+                raise ValueError(
+                    f"No deserializer registered for class '{cls_name}'"
+                )
+
+            return cls.from_dict(value)
+
+    return value
```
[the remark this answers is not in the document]
j053y: Nope that's what we want

fiftyone/server/routes/sample.py
```diff
@@ -117,8 +144,13 @@ async def patch(self, request: Request, data: dict) -> dict:
...
             )
-        sample.save()
-        return result.to_dict(include_private=True)
+
+        result.save()
+
+        return utils.json.JSONResponse(
+            utils.json.serialize(result),
+            headers={"ETag": f'"{generate_sample_etag(result)}"'},
+        )
```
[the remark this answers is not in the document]
j053y: Great point, attempting to move it closer to actual save operation...

fiftyone/server/routes/sample.py
```diff
@@ -52,13 +56,29 @@ def get_sample(dataset_id: str, sample_id: str) -> fo.Sample:
...
+            return sample
+
+        if etag == sample.last_modified_at.isoformat():
+            return sample
+
+        if etag == str(sample.last_modified_at.timestamp()):
+            return sample
+
+        raise HTTPException(status_code=412, detail="ETag does not match")
+
```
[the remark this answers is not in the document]
j053y: Great point, attempting to move it closer to actual save operation...

fiftyone/server/utils/__init__.py
```diff
@@ -15,9 +15,8 @@
 
 import fiftyone.core.dataset as fod
 import fiftyone.core.fields as fof
-from fiftyone.server.utils.json_transform import (
-    transform as transform_json,
-)  # auto-register resource types
+from fiftyone.server.utils import json
+from fiftyone.server.utils import http
```
CamronStaley: do these need to be two separate lines?

fiftyone/server/utils/http.py
```diff
@@ -0,0 +1,33 @@
...
+        """Parses an ETag string into its value and whether it is weak."""
+
+        is_weak = False
+        if etag.startswith("W/"):
+            is_weak = True
+            etag = etag[2:]  # Remove "W/" prefix
+
+        # Remove surrounding quotes (ETags are typically quoted)
+        if etag.startswith('"') and etag.endswith('"'):
+            etag = etag[1:-1]
```
CamronStaley, line 39: does this still work as expected after the removal of W/ above? Not super familiar with etags based on the code I assume it's something like:

`W/"some-content"` or `"some-content"` or `some-content` or `W/some-content`?

is this the case?

fiftyone/server/utils/__init__.py
j053y: Probably not. Good catch.

fiftyone/server/utils/http.py
j053y: Yep that's exactly it. The first two are preferred, the last two are just being a little permissive on the format.

fiftyone/server/routes/sample.py
```diff
@@ -6,64 +6,190 @@
...
-from typing import Any
+
+from fiftyone.server import decorators, utils
 
 logger = logging.getLogger(__name__)
 
 
-def get_sample(dataset_id: str, sample_id: str) -> fo.Sample:
-    """Retrieves a sample from a dataset.
+def get_if_last_modified_at(
```
CamronStaley, line 26: is there a reason to have this here over having it in the utils file?

fiftyone/server/routes/sample.py
```diff
@@ -6,64 +6,190 @@
...
+            ...
+
+        if if_last_modified_at is None:
+            raise HTTPException(
+                status_code=400, detail="Invalid If-Match header"
+            )
+    return if_last_modified_at
+
+
+def get_sample(
```
CamronStaley, line 77: in FOT I changed this so that it takes in the request and returns the sample this way the locking logic could be added into this method. If you'd prefer both methods call some: touch_lock, then call get_sample I could see that argument. Lmk your thoughts

fiftyone/server/routes/sample.py
```diff
@@ -6,64 +6,190 @@
...
+    if if_last_modified_at is not None:
+        if sample.last_modified_at != if_last_modified_at:
+            raise HTTPException(
+                status_code=412, detail="If-Match condition failed"
+            )
 
     return sample
 
 
+def generate_sample_etag(sample: fo.Sample) -> str:
```
CamronStaley, line 123: again maybe could be moved to util method? It's only used here currently, but doesn't the lock stuff also have the concept of an etag? Where does that exist / how does it differ from this implementation?

fiftyone/server/utils/http.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+
+from typing import Any
+
+
+class ETag:
+    """Utility class for creating and parsing ETag strings."""
+
+    @staticmethod
+    def create(value: Any) -> str:
```
CamronStaley: is this used anywhere? I see in the sample method "generate_sample_etage" you're doing this same thing

fiftyone/server/routes/sample.py
j053y: Talked in person

fiftyone/server/routes/sample.py
j053y: Talked in person

fiftyone/server/utils/http.py
j053y: Nope. It should be, I'll add it in.

fiftyone/server/routes/sample.py
j053y: Talked in person

PR #6425, opened 2025-10-15
improve error details for when an operator returns non-serializable value

imanjra:

improve error details for when an operator returns non-serializable value

Using a test operator and python panel event

        notes for FiftyOne users.

improve error details for when an operator returns non-serializable value

#### Todo
Update docs to note the rule that operator, python-panel lifecycle methods, and python-panel events must return JSON-serializable values. If a complex value to be returned (i.e. DatasetView), it must be serialized or parsed into serialiazble object

-   [ ] Other

## Summary by CodeRabbit

* Bug Fixes
  * Error panel now shows a distinct "Trace" section and includes explicit message entries for operator errors.
  * Operator output displays more informative, context-aware error reasons.
  * Prevented empty output views from being rendered when there are no results.

* Refactor
  * Unified JSON response creation for more consistent API responses.
  * Improved numeric-type serialization for reliability.
  * Enhanced server error responses to include structured messages and stack traces.

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -15,6 +15,8 @@ from contextlib import contextmanager, suppress
 from copy import deepcopy
 from datetime import date, datetime
 from functools import partial
+from starlette.responses import Response
+from json import JSONEncoder
 import glob
 import hashlib
 import importlib
@@ -3190,6 +3192,27 @@ def validate_hex_color(value):
         raise ValueError(
             "%s is not a valid hex color string (eg: '#FF6D04')" % value
         )
+
+
+class Encoder(JSONEncoder):
+    """Custom JSON encoder that handles numpy types."""
+
+    def default(self, o):
+        if isinstance(o, np.floating):
+            return float(o)
+
+        if isinstance(o, np.integer):
+            return int(o)
+
+        return JSONEncoder.default(self, o)
+
+
+async def create_response(response: dict):
+    """Creates a JSON response from the given dictionary."""
+    return Response(
+        await run_sync_task(lambda: json_util.dumps(response, cls=Encoder)),
+        headers={"Content-Type": "application/json"},
+    )
 
 
 fos = lazy_import("fiftyone.core.storage")
--- a/fiftyone/operators/utils.py
+++ b/fiftyone/operators/utils.py
@@ -6,10 +6,17 @@ FiftyOne operator utilities.
 |
 """
 
-from datetime import datetime, timedelta
 import logging
+import traceback
 
-from .operator import Operator
+from datetime import datetime, timedelta
+from typing import Union
+
+from fiftyone.core.utils import create_response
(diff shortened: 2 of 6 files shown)


jleven (dismissed):
Looks good to me!!

You may want a second opinion, though.

ritch (dismissed):

fiftyone/core/utils.py
```diff
@@ -3192,4 +3194,25 @@ def validate_hex_color(value):
         )
 
 
+class Encoder(JSONEncoder):
```
kaixi-wang, line 3197: does it make sense to move this out of `server/`? seems very server specific

fiftyone/operators/utils.py
```diff
@@ -107,3 +114,43 @@ def is_new(release_date, days=30):
...
+        # Immediately re-raise starlette HTTP exceptions
+        if isinstance(e, HTTPException):
+            raise e
+
+        # Cast non-starlette HTTP exceptions as JSON with 500 status code
+        logging.exception(e)
+
+        msg = (
+            f"Failed to serialize operator result. {e}."
+            + " Make sure that the return value of the operation is JSON-serializable."
```
kaixi-wang, line 146: Would users know that dataset and view are not serializable?

fiftyone/core/utils.py
```diff
@@ -3192,4 +3194,25 @@ def validate_hex_color(value):
...
+    """Custom JSON encoder that handles numpy types."""
+
+    def default(self, o):
+        if isinstance(o, np.floating):
+            return float(o)
+
+        if isinstance(o, np.integer):
+            return int(o)
+
```
kaixi-wang, line 3206: haven't thought it through completely, but why wouldn't we just handle the serialization here? eg if isinstance(o, DatasetView): return o._serialize()?

fiftyone/core/utils.py
imanjra: I moved it out to core.utils as it is now also used by `create_operator_response`

fiftyone/operators/utils.py
imanjra: Not necessarily, but with stack trace and this message, it should give them pointers

fiftyone/core/utils.py
imanjra: Yah, I consider customizing but I think we should leave serialization of complex values to be case-by-case. For example, what if they have view in a nested object or what if another complex class is returned

fiftyone/operators/utils.py
jleven: I think this is enough for the release. Always room to improve moving forward.

fiftyone/core/utils.py
jleven: Given that this work is release blocking, it's helpful to keep the change as minimal and safe as possible

PR #6438, opened 2025-10-17
FOEPD-2109 PipelineRunInfo type to support always-run pipeline stage

swheaton:

Model changes to DelegatedOperationDocument to support always-run stages and other future functionality.
It separates the static definition of the pipeline `types.Pipeline` from run state info that is mutable - `types.PipelineRunInfo`.
This allows pipeline related information to be contained within one sub document instead of spread around - more organized.

The two fields we need added are:
- active: if True, the pipeline has not failed and is ongoing. If False, the pipeline has failed in the past but we are proceeding through the pipeline to always-run stages only.
- expected_children: either None if `active`, otherwise it is a list matching length `len(pipeline.stages)`. This list contains the number of children we should expect to see for each stage. Basically, stages that have been skipped are 0.

Added/edited unit tests
Real testing was with the whole system which will be outlined in the voxel hub PR.

        notes for FiftyOne users.

Added a new PipelineRunInfo type within DelegatedOperationDocument that keeps track of pipeline run state as it goes. Currently useful for internal purposes only.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Pipeline stages can be marked to always run.
  * Pipeline run state tracking added (active flag, expected children, current stage index).

* **Improvements**
  * Serialization/deserialization preserves stage always-run flags and pipeline run state.
  * Tests updated to validate always-run propagation and pipeline run state round-trip.

### DIFF
--- a/fiftyone/operators/_types/pipeline.py
+++ b/fiftyone/operators/_types/pipeline.py
@@ -39,7 +39,7 @@ class PipelineStage:
         name: Optional[str] = None,
         num_distributed_tasks: Optional[int] = None,
         params: Optional[Mapping[str, Any]] = None,
-        **kwargs,  # Accepts and ignores unused kwargs
+        **_,
     ):
         # Call the default dataclass initialization for the defined fields
         self.operator_uri = operator_uri
@@ -53,6 +53,11 @@ class PipelineStage:
         if not self.operator_uri:
             raise ValueError("operator_uri must be a non-empty string")
 
+        self.num_distributed_tasks = (
+            int(self.num_distributed_tasks)
+            if self.num_distributed_tasks is not None
+            else None
+        )
         if (
             self.num_distributed_tasks is not None
             and self.num_distributed_tasks < 1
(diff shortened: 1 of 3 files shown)


fiftyone/operators/_types/pipeline.py
```diff
@@ -128,3 +167,24 @@ def to_json(self):
...
+
+@dataclasses.dataclass
+class PipelineRunInfo:
+    """Information about a pipeline run.
+
+    Unlike the pipeline definition, this class is considered mutable.
+    """
+
+    active: bool = True
+    expected_children: Optional[List[int]] = None
```
CamronStaley, line 180: seems fine to me just curious when / how is this expected children going to be used?

CamronStaley: Nvm I shoulda read the description better: 

```
expected_children: either None if active, otherwise it is a list matching length len(pipeline.stages). This list contains the number of children we should expect to see for each stage. Basically, stages that have been skipped are 0.
```

Still curious where this will be set / read

swheaton: 🤫 it's a secret

no you will see soon but it is used by broker/pipeline manager for 2 things:
- getting total children count to use as denominator for a progress calculation
- getting currently expected number of children given the current stage. This is used to check against how many actual children we found, so that we don't try to complete or progress the parent when all children haven't been created yet.

It is written inside the update parent state machine at the time of first stage failure (and only done if there are always-run stages to go, otherwise it's unnecessary

PR #6467, opened 2025-10-24
implement basic feature flags

tom-vx51:

This PR implements a basic feature flagging system in FiftyOne. Brief description of features:
* Set server-side flags to control availability of features
* Client-side hooks to enable conditional rendering and business logic
* Client-side caching to reduce network traffic
* Automatic metrics tracking for feature-flagged code paths

local

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Feature flags system with environment-based configuration, a public API endpoint to list enabled flags, a React hook and component to conditionally render UI, and an in-memory feature cache.

* **Bug Fixes**
  * Improved analytics resilience by preventing tracking errors from disrupting the app.

* **Tests**
  * Added unit tests covering feature cache behavior.

### DIFF
--- a/app/packages/feature-flags/src/hooks/useFeature.ts
+++ b/app/packages/feature-flags/src/hooks/useFeature.ts
@@ -12,10 +12,10 @@ import { useEffect } from "react";
  */
 export const useFeature = ({
   feature,
-  enableTracking = true,
+  enableTracking,
 }: {
   feature: FeatureFlag;
-  enableTracking: boolean;
+  enableTracking?: boolean;
 }): boolean => {
   const cache = useFeatureCache();
   const trackEvent = useTrackEvent();


app/packages/feature-flags/src/components/FeatureFlagged.tsx
```diff
@@ -0,0 +1,25 @@
...
+  children,
+  fallback,
+}: {
+  feature: FeatureFlag;
+  children: React.ReactNode;
+  fallback?: React.ReactNode;
+}) => {
+  const isFeatureEnabled = useFeature({ feature });
+
+  const content = isFeatureEnabled ? children : fallback ?? <Fragment />;
```
jleven, line 23: Just curious, why `<Fragment />` instead of `null`?

tom-vx51: No good reason, just the way I learned

app/packages/feature-flags/src/hooks/useFeatureCache.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Hook which provides access to a shared cache of enabled features.
+ */
+export const useFeatureCache = (): FeatureCache => {
+  const cache = useAtomValue(featureCacheAtom);
+  const [isInitialized, setIsInitialized] = useAtom(isInitializedAtom);
+  const [isFetching, setIsFetching] = useAtom(isFetchingAtom);
+
+  useEffect(() => {
+    if (!isInitialized && !isFetching) {
+      listEnabledFeatures()
```
jleven, line 25: I think you need to set isFetching before you fetch...



```suggestion
    if (!isInitialized && !isFetching) {
      setIsFetching(true);
      listEnabledFeatures()
```

jleven: Ha, looks like you caught it already!

app/packages/feature-flags/src/hooks/useFeature.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Hook which provides the status of a given feature.
+ *
+ * @param feature Feature identifier
+ * @param enableTracking If enabled, will emit tracking events
+ * @returns true if the feature is enabled, else false
+ */
+export const useFeature = ({
+  feature,
+  enableTracking = true,
```
jleven: Not a blocker, but my instinct is we will want the default to be `false`. Tracking is really only interesting when we are doing a deliberate test, and most of the time, my bet is that we will not be testing, but deliberately hiding.

tom-vx51: ha fair point, updated

app/packages/feature-flags/vite.config.ts
```diff
@@ -0,0 +1,37 @@
+import { defineConfig } from "vite";
+import react from "@vitejs/plugin-react";
+import path from "path";
+import { viteExternalsPlugin } from "vite-plugin-externals";
+
+const isPluginBuild = process.env.STANDALONE !== "true";
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  mode: "development",
```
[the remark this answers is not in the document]
jleven: @tom-vx51 -- Looks like we make this mistake in a number of our packages. Not blocking, but maybe we should make a ticket and look into this?

app/packages/feature-flags/src/hooks/useFeature.ts
```diff
@@ -0,0 +1,35 @@
...
+  const isEnabled = cache.isFeatureEnabled(feature.toString());
+
+  useEffect(() => {
+    if (enableTracking) {
+      trackEvent("VFF:CHECK", {
+        feature,
+        isEnabled,
+      });
+    }
+  }, [enableTracking, feature, isEnabled]);
```
jleven, line 32: **non-blocking(suggestion)**: If I'm understanding this correctly, the cache will always start empty while it fetches from the backend. So, this will always fire with "feature disabled", and then fire again with "feature enabled" after the cache is populated. Could be cleaner if you didn't fire the tracking event until after the cache is initialized.

Of course, doesn't look like it's possible to know if the cache is initialized yet, so this would be a larger change, and I frankly don't think it's worth it.

tom-vx51: yeah good point, it should wait until it resolves. That info is available in `useFeatureCache` but isn't exposed. Easy enough to add, can do it in another PR :)

app/packages/feature-flags/src/hooks/useFeature.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @param feature Feature identifier
+ * @param enableTracking If enabled, will emit tracking events
+ * @returns true if the feature is enabled, else false
+ */
+export const useFeature = ({
+  feature,
+  enableTracking,
+}: {
+  feature: FeatureFlag;
+  enableTracking?: boolean;
```
moses-v51, line 18: Do we want this false by default?

tom-vx51: it acts falsey with the latest change 👍

moses-v51: Right, that's what I was questioning - whether that's the behavior we want.

tom-vx51: oh sorry 🙃 easy enough to change once someone has a strong opinion

PR #6474, opened 2025-10-24
Human Annotation Saved Context Fixes

benjaminpkane:

Fixes/verifies the support for save confirmations during the below actions
* Leaving focus via "<-" back button
* Modal navigation
* Modal closing
* Leaving focus by selecting a new label
* Canceling

https://github.com/user-attachments/assets/c9f0a72a-5ee1-45cf-8c7e-4dabfe35621e

## Release Notes

N/A

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Improved focus/confirmation flow when switching or deselecting annotations, prompting when there are unsaved changes.

* **Bug Fixes**
  * Modal overlay now appears above other UI layers for more reliable visibility.

* **Improvements**
  * Deletion now exits the edit flow immediately after confirmation for clearer behavior.
  * Removed an automatic overlay auto-selection behavior and refined event handling to prevent unintended updates.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/AnnotationSchema.test.tsx
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/AnnotationSchema.test.tsx
@@ -1,0 +1,125 @@
+import { LIGHTER_EVENTS, TransformOverlayCommand } from "@fiftyone/lighter";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+describe("AnnotationSchema", () => {
+  let mockScene: any;
+  let mockOverlay: any;
+  let commandExecutedHandler: any;
+
+  beforeEach(() => {
+    commandExecutedHandler = null;
+    mockScene = {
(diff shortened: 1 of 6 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useDelete.ts
```diff
@@ -1,33 +1,34 @@
...
+        type: LIGHTER_EVENTS.DO_REMOVE_OVERLAY,
+        detail: {
+          label,
+          schema: getFieldSchema(schema, label?.path)!,
         },
-      },
-    });
-  }, [exit, label, scene, schema, setter]);
+      });
+    removeOverlay(label?.data._id);
```
tom-vx51, line 32: I think this needs to be in the success callback like it was previously. If the event handler fails to delete, then the overlay should remain right?

benjaminpkane: Yes, except when the overlay is new/unsaved. I will follow up after a live discussion

PR #6494, opened 2025-10-28
Add support for filtering model predictions by class

brimoor:
## Release Notes

Adds support for filtering model predictions by class via a new `SampleCollection.apply_model(..., classes=)` syntax.

If necessary, `classes` is enforced *after* model inference, so this syntax is supported for *all models*.

However, models can optionally declare that they *internally* support class filtering by providing a `model.config.filter_classes` config setting, in which case `apply_model()` delegates filtering to the `Model` instance, for efficiency.

```py
import fiftyone as fo
import fiftyone.zoo as foz

model = foz.load_zoo_model("yolo11s-coco-torch")

# Tell supported models what classes they are allowed to return
if hasattr(model.config, "filter_classes"):
    model.config.filter_classes = ["cat", "dog"]

# Anything that's not a 'cat' or 'dog' is filtered out!
predictions = model.predict(...)
```

## Tested by

```py
import fiftyone as fo
import fiftyone.zoo as foz

```

### Test a raw `TorchModel`

```py

# All classes
dataset.apply_model(model, label_field="clip")
print(dataset.count_values("clip.label"))

# Specific classes via apply_model()
dataset.apply_model(model, label_field="clip", classes=["person"])
print(dataset.count_values("clip.label"))

# Specific classes via model.config.filter_classes
model.config.filter_classes = ["cat", "dog"]
dataset.apply_model(model, label_field="clip")
model.config.filter_classes = None  # reset manual filtering
print(dataset.count_values("clip.label"))
```

### Test an `ultralytics` model

```py
model = foz.load_zoo_model("yolo11s-coco-torch")

# All classes
dataset.apply_model(model, label_field="yolo11")
print(dataset.count_values("yolo11.detections.label"))

# Specific classes via apply_model()
dataset.apply_model(model, label_field="yolo11", classes=["person"])
print(dataset.count_values("yolo11.detections.label"))

# Specific classes via model.config.filter_classes
model.config.filter_classes = ["cat", "dog"]
dataset.apply_model(model, label_field="yolo11")
model.config.filter_classes = None  # reset manual filtering
print(dataset.count_values("yolo11.detections.label"))
```

### Test a `transformers` model

```py
model = foz.load_zoo_model(
    "zero-shot-classification-transformer-torch",
    classes=["cat", "dog", "person", "other"],
)

# All classes
dataset.apply_model(model, label_field="transformer")
print(dataset.count_values("transformer.label"))

# Specific classes via apply_model()
dataset.apply_model(model, label_field="transformer", classes=["person"])
print(dataset.count_values("transformer.label"))

# Specific classes via model.config.filter_classes
model.config.filter_classes = ["cat", "dog"]
dataset.apply_model(model, label_field="transformer")
model.config.filter_classes = None  # reset manual filtering
print(dataset.count_values("transformer.label"))
```

## Summary by CodeRabbit

* **New Features**
  * Added an optional "classes" filter to model application to restrict predictions to a chosen subset of labels.
  * Class filtering is supported end-to-end across inference pipelines (Torch, Transformers, Ultralytics, and CLIP flows).
  * Prediction post-processing and output processors now honor class restrictions so only requested classes appear in results.
  * Sample-level label operations gained label-filtering to save only allowed classes.


fiftyone/utils/open_clip.py
```diff
@@ -162,5 +162,8 @@ def _predict_all(self, imgs):
                 self._output_processor.store_logits = self.store_logits
 
         return self._output_processor(
-            output, frame_size, confidence_thresh=self.config.confidence_thresh
+            output,
+            frame_size,
+            confidence_thresh=self.config.confidence_thresh,
+            self._filter_classes,
```
manushreegangwar:
```suggestion
            classes=self._filter_classes,
```

brimoor:
@jacobsela if we don't want to add scope to the `Model` interface; then now is the time to make that decision! Otherwise we'll have a backwards-compatibility commitment to make.

The idea of offloading all post-processing (ie going from torch -> FO -> database) to a separate process makes sense in theory; the less work done in the GPU process, the more throughput it can have. But do you have a timeline for achieving it? As you know, we had to [revert our baby step in that direction due to memory leaks](https://github.com/voxel51/fiftyone/pull/6394), which doesn't exactly instill optimism in a larger scoped effort.

kaixi-wang:
Instead of adding attributes to the model, would it be possible to just pass in a post processor to the predict method

jleven:
If this comment is outdated based on the current discussion, please ignore. 

What happens when model.filter_classes and apply_model(classes=) disagree? Does apply_model overwrite the other?

brimoor:
> If this comment is outdated based on the current discussion, please ignore.
> 
> What happens when model.filter_classes and apply_model(classes=) disagree? Does apply_model overwrite the other?

`apply_model(classes=)` will set `model.filter_classes` (if the model supports filtering) for the lifetime of the function call and then return `model.filter_classes` to whatever it was before `apply_model()` was invoked.

brimoor:
@jacobsela okay thanks, I'm onboard with the general direction you've laid out here, and I'll take a crack at updating this PR to see how far in that direction I can get.

At minimum, I'll replace this:

```py
if model.can_filter_classes:
    model.filter_classes = [...]
```

with something where the filters are applied directly to the `output_processor`.

I have some implementation questions that I'll follow-up offline about.

manushreegangwar (approved):
LGTM :shipit:

PR #6514, opened 2025-10-30
coderabbit feedback and other minor changes

manushreegangwar:

Address coderabbit feedback and other minor changes

Same tests as https://github.com/voxel51/fiftyone/pull/6494

Additional tests for confidence thresholding
```
model_names = ["clip-vit-base32-torch", "resnet18-imagenet-torch", "classification-transformer-torch", "yolo11s-coco-torch"]
for model_name in model_names:
    model = foz.load_zoo_model(model_name)
    for conf in [None, 0.5]:
            label_field = f"{model_name}_{conf}_dev".replace(".","")
            dataset.apply_model(model, label_field=label_field, confidence_thresh=conf)

            # Compare dev against PR using dataset.count_values and dataset.evaluate_xyz
```

        notes for FiftyOne users.

-   [x] Other

## New Features
- Added optional class filtering when applying models, allowing predictions to be restricted to a specific subset of classes.
- Models now expose whether they support class filtering capabilities.
- Class filtering is supported across classifiers, detectors, segmenters, and keypoint models.


fiftyone/core/models.py
```diff
@@ -157,7 +157,7 @@ def apply_model(
         )
 
     if classes is not None:
-        classes = set(classes)
+        classes = list(set(classes))
```
brimoor, line 160: I had intentionally converted to `set()` here so that `label in classes` lookups will be `O(1)` instead of `O(len(classes))`.

fiftyone/core/sample.py
```diff
@@ -768,55 +765,53 @@ def _save(self, deferred=False):
...
             )
             for frame_number, frame_dict in label.items()
         }
     elif isinstance(label, dict):
         label = {
-            k: _apply_confidence_thresh(v, confidence_thresh)
+            k: _apply_label_filters(v, confidence_thresh, classes)
             for k, v in label.items()
         }
+    elif isinstance(label, fol._HasLabelList):
```
brimoor, line 784: Note that `confidence_thresh` has different semantics for `Keypoint(s)` instances than other `_HasLabelList` types: the `confidence` attribute is an array of per-point confidences rather than a single value. That's why there was nearly duplicate code in the previous implementation,.

I'm sure there's a way to have slightly less duplicate code, but we will need to bring back the correct behavior for `Keypoint(s)` as this implementation will not work for them.

fiftyone/core/sample.py
```diff
@@ -762,34 +765,54 @@ def _save(self, deferred=False):
...
+                or label.confidence < confidence_thresh
+            ):
+                remove_label = True
+
+        if classes is not None and hasattr(label, "label"):
+            if label.label not in classes:
+                remove_label = True
+
+        if remove_label:
+            label = label.__class__()
```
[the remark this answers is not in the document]
brimoor: No, `label.__class__()` is intentional. We want `None` to mean "unprocessed" while `label.__class__()` means "processed but no predictions matching the provided filters were found".

fiftyone/utils/torch.py
```diff
@@ -850,8 +850,9 @@ def filter_classes(self):
 
     @filter_classes.setter
     def filter_classes(self, value):
-        if value == self.classes:
-            value = None
+        if value and self.can_filter_classes:
```
brimoor, line 853: The `can_filter_classes` check doesn't make sense here. We should either decide to raise an exception if the user tries to set `filter_classes` when `can_filter_classes == False`, or we should just let the setter work in all cases. With this proposal, it can be called but its behavior changes in a confusing way depending on `can_filter_classes=True/False`.

fiftyone/core/models.py
brimoor: We could instead remove the cast here and force the `Model.filter_classes` setter to convert to `set()` internally, for efficiency.

The reason I didn't do that is that it puts the responsibility on the `Model` class to realize that set membership is more efficient, which would mean that logic needs to be duplicated for every `Model` instance that requires its own `filter_classes` implementation.

With that said, the proposed behavior in my PR would require that we update the docstring of the `Model.filter_classes` property to indicate that it can be any iterable, not strictly a list, so that it is contractually kosher for us to pass `classes` as a `set()` in `apply_model()`. But that *also* puts some onus on the `Model` class implementer on how their class-filtering implementation should work. So, there's a bit of work to be done by the `Model` author in either case.

fiftyone/utils/torch.py
brimoor: Also upon further thought, I think we should just delete this coercion entirely:

```py
if value == self.classes:
    value = None
```

It's cleaner to just require users to set:

```py
model.filter_classes = None
```

if they want filtering to be disabled completely.

I initially added the coercion for a nuanced reason (I was trying to piggy-back off of `Model.classes`; details are irrelevant now) but didn't end up going in that direction.

PR #6515, opened 2025-10-30
More 3D fixes

sashankaryal:
-

## Release Notes

* **New Features**
  * Added a second annotation plane view accessible via ⌘ + 4 shortcut
  * Added preview lines that display during polyline vertex transformations

* **Improvements**
  * Enhanced delete functionality to focus on point-level deletion for polylines
  * Increased scene bounds computation timeout for improved stability
  * Streamlined tooltip event handling for better performance

### DIFF
--- a/app/packages/looker-3d/src/annotation/SegmentPolylineRenderer.tsx
+++ b/app/packages/looker-3d/src/annotation/SegmentPolylineRenderer.tsx
@@ -1,7 +1,7 @@
 import * as fos from "@fiftyone/state";
 import { objectId } from "@fiftyone/utilities";
 import { Line as LineDrei } from "@react-three/drei";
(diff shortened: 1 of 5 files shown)


jleven:
I'm seeing a consistent timeout in the e2e test for `e2e-pw/src/oss/specs/3d/fo3d-pcd-stl.spec.ts` > "scene is rendered correctly". Specifically, `await modal.navigateNextSample()` is timing out at:

```
e2e-pw/src/oss/specs/3d/fo3d-pcd-stl.spec.ts:151 await modal.navigateNextSample()
e2e-pw/src/oss/poms/modal/index.ts:133 this.waitForSampleLoadDomAttribute(allowErrorInfo)
```

Rerunning, in case it was just a flake. But I wonder if the new 15 sec timeout is affecting this test, as that particular call has a 10sec timeout.

PR #6544, opened 2025-11-07
[Proposal] add ResolvableProperty type for operators

imanjra:

Add a ResolvavleProperty types to operator along with ResolvablePropertyView to support asynchronously resolving properly independent of resolve_input. This is particularly useful when a property is determined by computationally expensive logic.

```py
class PlotOperator(foo.Operator):
    @property
    def config(self):
        return foo.OperatorConfig(
            name="plot",
            label="Plot",
            dynamic=True,
        )

    def resolve_input(self, ctx):
        inputs = types.Object()
        inputs.str("name", label="Name")
        inputs.str("limit", label="Plot Limit")
        inputs.define_property(
            "plot",
            types.ResolvableProperty(
                resolver="@voxel51/plot/plot_resolver",
                dependencies=["limit"], # Optionally provide properties to watch
                auto_update=False, # Update only when the refresh button is clicked
                validate=True, # property must be resolved and validated with latest params
                params={} # params to pass to resolver
            ),
            label="Plot",
        )

        return types.Property(inputs)

    def execute(self, ctx):
        return {}


class PlotResolverOperator(foo.Operator):
    @property
    def config(self):
        return foo.OperatorConfig(
            name="plot_resolver",
            label="Plot Resolver",
            unlisted=True,
        )

    def execute(self, ctx):
        # Computationally-expensinve logic below
        time.sleep(5)
        x_values = [0, 1, 2, 3, 5, 6, 7, 8, 9, 10]
        y_values = [0] + [random.randint(5, 25) for _ in range(9)]

        return types.Property(
            types.Object(),
            view=types.PlotlyView(
                data=[
                    {
                        "x": x_values,
                        "y": y_values,
                        "type": "scatter",
                    }
                ],
                label="Resolved Plot",
                height="250px",
            ),
        ).to_json()
```

## Todos
- [x] Add support for update only on click
- [x] Add support for awaiting resolvable property
- [x] Add support for debounce/throttle configuration
- [x] Add support for binding refresh to specific params by path
- [x] Add loading state indicator
- [x] Add stale state indicator
- [x] Add error handling
- [x] Add validation

        notes for FiftyOne users.

Add a new operator input type `ResolvableProperty` and a corresponding new view `ResolvablePropertyView` to enable asynchronously resolving a property in operator's input.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Resolvable properties: runtime resolution with debounce/throttle, params, dependencies, auto/manual refresh, loading/error UI, and a new resolvable property view/type.
  * New plot resolver operator to generate scenario/plot previews.

* **Improvements**
  * Responsive plots: sizing now derives from view dimensions.
  * Validation plumbing: new validation error type, callbacks, propagation, and top-level validate export; validation can be forwarded to prompts/IO.
  * State helpers: exported unbound-state ref hook.

* **Other**
  * Plugin entrypoint loading now validated to prevent out-of-bounds paths.

### DIFF
--- a/plugins/operators/model_evaluation/__init__.py
+++ b/plugins/operators/model_evaluation/__init__.py
@@ -6,7 +6,6 @@ Scenario plugin.
 |
 """
 
-import fiftyone as fo
 import fiftyone.operators as foo
 import fiftyone.operators.types as types
 import fiftyone.core.fields as fof
@@ -14,7 +13,6 @@ import fiftyone.server.utils as fosu
 from fiftyone.operators.cache import execution_cache
 
 from bson import ObjectId
-from fiftyone.core.expressions import ViewField as F
 from .utils import (
     get_scenario_example,
     SCENARIO_BUILDING_CHOICES,
@@ -50,14 +48,6 @@ class ConfigureScenario(foo.Operator):
             unlisted=True,
         )
 
-    # we use `fosu.cache_dataset()` rather than `@execution_cache` here so that
-    # the cached dataset can be reused outside of the current prompt session
-    def get_dataset(self, ctx):
-        """
-        Returns the dataset for the current context.
-        """
-        return fosu.cache_dataset(ctx.dataset)
-
     @execution_cache(
         prompt_scoped=True, residency="ephemeral", ttl=PROMPT_SCOPED_CACHE_TTL
     )
@@ -65,7 +55,7 @@ class ConfigureScenario(foo.Operator):
         """
         Returns the number of samples in the dataset for the current context.
         """
-        dataset = self.get_dataset(ctx)
+        dataset = get_dataset(ctx)
         return dataset.count()
 
     def get_default_for_distribution_preview(self, ctx):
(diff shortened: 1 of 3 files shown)


app/packages/core/src/plugins/SchemaIO/components/ResolvablePropertyView.tsx
```diff
@@ -0,0 +1,156 @@
...
+            setError(error);
+          } else {
+            setResolvedSchema(
+              operatorToIOSchema(types.Property.fromJSON(schema))
+            );
+            setResolvedParams(serializedObservedData);
+          }
+        },
+      });
+    },
```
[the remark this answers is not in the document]
imanjra: Not quite. The `error` is nullish at line below `setError(error);` if `executeOperator` succeeds

app/packages/core/src/plugins/SchemaIO/components/ResolvablePropertyView.tsx
```diff
@@ -0,0 +1,246 @@
...
+      setResolveCount((count) => count + 1);
+      executeOperator(resolver, computedParams, {
+        callback: (result) => {
+          const { result: schema, error } = result;
+          setError(error);
+          if (schema) {
+            try {
+              const property = types.Property.fromJSON(schema);
+              setResolvedSchema(operatorToIOSchema(property));
+              if (validate && onValidationErrors && resolveCount <= 1) {
```
lanzhenw: Is resolveCount <= 1 safe? The callback captures it at creation time, not execution time. It's a stale value. 

Maybe we can switch this an explicit "first resolution isnce last dependency change" flag?
```
const firstResolutionRef = useRef(true);

useEffect(() => {
  firstResolutionRef.current = true;
}, [serializedObservedData, serializedProvidedParams]);

// inside callback:
if (validate && onValidationErrors && firstResolutionRef.current) {
  firstResolutionRef.current = false;
  …
}
```

imanjra: Good catch! This is an edge case which can definitely be possibly. Let me try to replicate it to see if I can re-create it

app/packages/core/src/plugins/SchemaIO/components/ResolvablePropertyView.tsx
```diff
@@ -0,0 +1,246 @@
...
+      : fullData;
+    return JSON.stringify(observedData);
+  }, [fullData, dependencies]);
+  const serializedProvidedParams = useMemo(() => {
+    return JSON.stringify(params || {});
+  }, [params]);
+
+  const schemaResolver = useCallback(
+    (
+      params: Record<string, unknown>,
```
erik-nieh, line 72: Do we always prefer `unknown` over `any`?

plugins/operators/model_evaluation/__init__.py
```diff
@@ -1287,3 +1052,271 @@ def execute(self, ctx):
...
+        ScenarioType.CUSTOM_CODE,
+        ScenarioType.LABEL_ATTRIBUTE,
+        ScenarioType.SAMPLE_FIELD,
+    ]:
+        return ScenarioType.CUSTOM_CODE
+
+    return scenario_type
+
+
+def resolve_subsets(scenario_type, params, values):
```
erik-nieh, line 1301: I noticed this while working on my PR, but do we run `mypy` anywhere? I think most of these views are untyped but I have seen some other python in the project that is.

app/packages/core/src/plugins/SchemaIO/components/ResolvablePropertyView.tsx
imanjra: I was able to replicate the race condition. Fixed in [this amend](https://github.com/voxel51/fiftyone/compare/8e14a08565192a6dd054f174cd8f390a2b61f12b..c3124f9c78c0530f8550a7bd430253b82c30008b)

app/packages/core/src/plugins/SchemaIO/components/ResolvablePropertyView.tsx
imanjra: I prefer `unknown` as it forces you to cast the final value into the expected type, whereas `any` can be used without casting it

For example

```js
function x(val: string) {
  return val
}

const myUnknownVal: unknown = "10"
const myAnyVal: any = "10"

x(myUnknownVal) // ts warning which can be resolved by doing x(myUnknownVal as string)
x(myAnyVal) // no ts warnign
```

plugins/operators/model_evaluation/__init__.py
imanjra: Good point. I don't think we do. Yah, there are some type inconsistencies. 

Will add types as a follow-up to this plugin

app/packages/core/src/plugins/SchemaIO/components/ResolvablePropertyView.tsx
lanzhenw: Checking the resolveId and return the early execution is neat.

app/packages/core/src/plugins/SchemaIO/components/ResolvablePropertyView.tsx
```diff
@@ -0,0 +1,256 @@
...
+              }
+            } catch (error) {
+              setError((error as Error).toString());
+            }
+          }
+        },
+      });
+    },
+    [resolver, onValidationErrors]
+  );
```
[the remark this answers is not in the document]
imanjra: This is not quite correct. This validation is  for validating nested value with respective nested schema.

app/packages/operators/src/state.ts
```diff
@@ -462,9 +462,12 @@ export const useOperatorPrompt = () => {
   const hooks = operator.useHooks(ctx);
   const executor = useOperatorExecutor(promptingOperator.operatorName);
   const [inputFields, setInputFields] = useState();
+  const [outputFields, setOutputFields] = useState();
   const [preparing, setPreparing] = useState(false);
   const [resolvedCtx, setResolvedCtx] = useState(null);
   const [resolvedIO, setResolvedIO] = useState({ input: null, output: null });
+  const [validationErrors, setValidationErrors] = useState([]);
+  const [customValidationErrors, setCustomValidationErrors] = useState([]);
   const notify = fos.useNotification();
```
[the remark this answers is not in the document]
imanjra: This is addressed in different PR https://github.com/voxel51/fiftyone/pull/6572

app/packages/core/src/plugins/SchemaIO/components/ResolvablePropertyView.tsx
```diff
@@ -0,0 +1,256 @@
...
+                onValidationErrors(basePath, errors);
+              }
+            } catch (error) {
+              setError((error as Error).toString());
+            }
+          }
+        },
+      });
+    },
+    [resolver, onValidationErrors]
```
lanzhenw, line 109: do we want `path` in the dependency?

imanjra: No, it's passed in to the callback

app/packages/core/src/plugins/SchemaIO/utils/types.ts
```diff
@@ -63,15 +63,20 @@ export type ViewPropsType<Schema extends SchemaType = SchemaType> = {
...
   /**
    * Custom. Available only when explicitly passed to the view.
    */
   onClick?: (
     e: React.MouseEvent,
-    params: { [key: string]: any },
+    params: Record<string, unknown>,
     props: ViewPropsType
   ) => void;
+  fullBody?: Record<string, unknown>;
```
lanzhenw: it should be `fullData`

jleven:
@ritch / @imanjra -- do we need to add documentation for this somewhere?

PR #6553, opened 2025-11-08
docs + more type safety for universal event bus

sashankaryal:
Add support for payload-less events, refactors hooks to a factory pattern for improved type safety, and docs

### Changelog

1. Events can have undefined payloads; dispatch makes payload optional.
2. Replaces `useEventHandler<T, E>()` with `createUseEventHandler<T>()()` to avoid repeating the event group type.
3. Use static registry and not `atomFamily`
4. Docs

### Breaking Changes
- `useEventHandler` replaced with `createUseEventHandler` factory pattern

### DIFF
--- a/app/packages/events/src/demo.ts
+++ b/app/packages/events/src/demo.ts
@@ -1,4 +1,4 @@
-import { EventDispatcher } from "./dispatch";
+import { getEventBus } from "./hooks";
 import { EventHandler } from "./types";
 
 export type DemoEventGroup = {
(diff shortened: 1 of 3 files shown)


app/packages/events/src/demo.tsx
```diff
@@ -13,8 +13,12 @@ export type DemoEventGroup = {
   "demo:eventC": {
     foo: string[];
   };
+  // Event with no payload
+  "demo:eventD": undefined;
 };
 
+const useDemoEventHandler = createUseEventHandler<DemoEventGroup>();
```
tom-vx51, line 20: nice

PR #6572, opened 2025-11-13
fix operator prompt sometimes getting stuck loading [FOEPD-2470]

imanjra:

fix operator prompt sometimes getting stuck loading

Using several built-in operators with prompt

        notes for FiftyOne users.

fix operator prompt sometimes getting stuck loading

-   [ ] Other

## Summary by CodeRabbit

* **Refactor**
  * Improved internal state management for operator parameter resolution, optimizing how resolved parameters are tracked and stored.
  * Enhanced debounce behavior for parameter resolution operations.

### DIFF
--- a/app/packages/operators/src/state.ts
+++ b/app/packages/operators/src/state.ts
@@ -463,7 +463,7 @@ export const useOperatorPrompt = () => {
   const executor = useOperatorExecutor(promptingOperator.operatorName);
   const [inputFields, setInputFields] = useState();
   const [preparing, setPreparing] = useState(false);
-  const [resolvedCtx, setResolvedCtx] = useState(null);
+  const [resolvedParams, setResolvedParams] = useState(null);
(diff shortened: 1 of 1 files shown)


app/packages/operators/src/state.ts
```diff
@@ -501,12 +505,12 @@ export const useOperatorPrompt = () => {
           resolveTypeError.current = e;
           setInputFields(null);
         }
-        setResolvedCtx(ctx);
+        setResolvedParams(ctx.params);
       },
       operator.isRemote ? RESOLVE_TYPE_TTL : 0,
-      { leading: true }
+      { leading: true, trailing: true }
```
tom-vx51, line 511: not sure if there's a reason we need to do this on both edges; it may be sufficient to just do trailing

imanjra: Yah, mostly agree. However, because it's a debounce, it'll add a slightly larger delay than it is now before anything in prompt appears.

PR #6579, opened 2025-11-14
annotation event bus

sashankaryal:
## Summary by CodeRabbit

* **New Features**
  * Typed annotation event system for reliable hover/selection and sidebar↔canvas syncing.
  * Command-driven save/delete with user-facing success/error notifications.
  * Staged 3D polyline editing with preview and a lightweight 3D annotation initializer.

* **Bug Fixes**
  * Safer overlay update handling to prevent runtime errors.
  * More robust parsing of boolean-like label fields.

* **Improvements**
  * Increased label visibility and refined zoom behavior in 3D view.
  * Render-loop activity tracking for improved stability.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/AnnotationSchema.tsx
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/AnnotationSchema.tsx
@@ -150,13 +150,13 @@ const useHandleChanges = () => {
           if (schema?.ftype === FLOAT_FIELD) {
             if (!data.length) return null;
             const parsed = Number.parseFloat(data);
-            return Number.isFinite(parsed) ? null : parsed;
+            return Number.isFinite(parsed) ? parsed : null;
           }
 
           if (schema?.ftype === INT_FIELD) {
             if (!data.length) return null;
             const parsed = Number.parseInt(data);
-            return Number.isFinite(parsed) ? null : parsed;
+            return Number.isFinite(parsed) ? parsed : null;
           }
         }
 
--- a/app/packages/looker-3d/src/state.ts
+++ b/app/packages/looker-3d/src/state.ts
@@ -435,10 +435,9 @@ export const hoveredVertexAtom = atom<{
  * Changes accumulate here as users manipulate polylines in the 3D canvas,
  * and is cleared once user commits changes or exits edit mode.
  */
-export const stagedPolylineTransformsAtom = atom<Record<
-  string,
-  PolylinePointTransformData
-> | null>({
+export const stagedPolylineTransformsAtom = atom<
+  Record<string, PolylinePointTransformData>
+>({
   key: "fo3d-stagedPolylineTransforms",
   default: {},
 });
--- a/app/packages/commands/src/hooks/useRegisterCommandHandler.test.ts
+++ b/app/packages/commands/src/hooks/useRegisterCommandHandler.test.ts
@@ -208,7 +208,7 @@ describe("useRegisterCommandHandler", () => {
     expect(result).toBe("hello");
   });
 
-  it("should handle registration errors gracefully", () => {
+  it("should handle registration errors gracefully", async () => {
     const consoleErrorSpy = vi
       .spyOn(console, "error")
       .mockImplementation(() => {});
@@ -234,7 +234,9 @@ describe("useRegisterCommandHandler", () => {
       expect.any(Error)
     );
 
-    expect(bus.execute(new TestCommand(5))).resolves.toEqual({ value: 5 });
+    await expect(bus.execute(new TestCommand(5))).resolves.toEqual({
+      value: 5,
+    });
 
     consoleErrorSpy.mockRestore();
   });
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useDelete.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useDelete.ts
@@ -2,7 +2,7 @@ import { DeleteAnnotationCommand, getFieldSchema } from "@fiftyone/annotation";
 import { useCommandBus } from "@fiftyone/commands";
 import { useLighter } from "@fiftyone/lighter";
 import * as fos from "@fiftyone/state";
-import { useAtomValue, useSetAtom } from "jotai";
+import { useAtom, useAtomValue, useSetAtom } from "jotai";
 import { useCallback } from "react";
 import { useRecoilValue } from "recoil";
 import { current, deleteValue } from "./state";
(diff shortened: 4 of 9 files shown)


app/packages/core/src/components/Modal/Lighter/useBridge.ts
```diff
@@ -17,7 +20,100 @@ import { useLighterTooltipEventHandler } from "./useLighterTooltipEventHandler";
...
+
+  useAnnotationEventHandler(
+    "annotation:notification:sidebarLabelHover",
+    useCallback(
+      (payload) => {
+        if (!scene) {
+          return;
+        }
+
+        scene.dispatchSafely({
```
tom-vx51, line 57: i assume it'll make sense as i continue reading, but do we need to maintain the two separate event systems, or can we have lighter migrate to the new events?

edit - it did not become clear. perhaps a later refactor

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/AnnotationSchema.tsx
```diff
@@ -143,22 +141,27 @@ const useSchema = () => {
...
+    ({ snapshot }) =>
+      async (currentField: string, path: string, data) => {
+        const expanded = await snapshot.getPromise(expandPath(currentField));
+        const schema = await snapshot.getPromise(field(`${expanded}.${path}`));
+
+        if (typeof data === "string") {
+          if (schema?.ftype === FLOAT_FIELD) {
+            if (!data.length) return null;
+            const parsed = Number.parseFloat(data);
+            return Number.isNaN(parsed) ? null : parsed;
```
tom-vx51: `Number.isFinite` might be more appropriate since it checks both `NaN` and `inf`/`-inf`

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useSave.ts
```diff
@@ -22,47 +24,53 @@ export default function useSave() {
...
-          });
-        },
+          addOverlay(label.overlay as BaseOverlay);
+        }
+
+        saved(label.data);
+        setSaving(false);
+        exit();
+        setNotification({
+          msg: `Label "${label.data.label ?? "Label"}" saved successfully.`,
```
tom-vx51: [nit] `Label "Label" saved successfully` in the fallback case; maybe just omit

app/packages/core/src/components/Modal/Sidebar/Annotate/utils.ts
```diff
@@ -8,16 +8,25 @@
  * @param records - The records to sanitize
  * @returns The sanitized records
  */
-export function coerceStringBooleans(
-  records: Record<string, unknown>
-): Record<string, unknown> {
+export function coerceStringBooleans<T extends Record<string, unknown>>(
```
tom-vx51, line 11: [note for myself] Conceptually similar to the logic in `transformSampleData` utility; will see if it makes sense to consolidate/unify.

app/packages/events/src/hooks/createUseEventHandler.ts
```diff
@@ -18,11 +18,28 @@ import { useEventBus } from "./useEventBus";
...
  * }
  * ```
+ *
+ * @remarks
+ * **⚠️ IMPORTANT**: Always wrap your handler functions in `useCallback` to ensure
+ * referential stability. Without `useCallback`, the handler will be a new function reference
+ * on every render, causing the event handler to be unregistered and re-registered unnecessarily,
+ * which can lead to performance issues and missed events.
+ *
+ * You can use `useCallback` directly as the second argument to get automatic type inference
```
tom-vx51, line 41: 🙌

app/packages/looker-3d/src/hooks/use-3d-annotation.ts
```diff
@@ -0,0 +1,128 @@
...
+        // Note: we don't support non-polyline annotations in 3D yet
+        return;
+      }
+
+      setSelectedLabelForAnnotation({
+        _id: payload.id,
+        ...payload.data,
+      });
+
+      const polylineData = payload.data as PolylineLabel;
```
tom-vx51, line 43: i see why we need the cast currently. Would be nice if this could be inferred somehow, maybe some TS magic as we add more label types

app/packages/looker-3d/src/hooks/use-3d-annotation.ts
```diff
@@ -0,0 +1,128 @@
...
+      });
+    }, [])
+  );
+
+  useAnnotationEventHandler(
+    "annotation:notification:sidebarValueUpdated",
+    useCallback(
+      (payload) => {
+        if (
+          payload.value?.["type"] !== "Polyline" ||
```
tom-vx51: similar, noting for future we should apply as much type safety as possible. I'd assume this `value` should always be defined

app/packages/looker-3d/src/hooks/use-3d-annotation.ts
```diff
@@ -0,0 +1,128 @@
...
+          !Array.isArray(payload.value?.["points3d"])
+        ) {
+          return;
+        }
+
+        const coerced = coerceStringBooleans(payload.value as PolylineLabel);
```
tom-vx51, line 73: cast needed here? this method assumes generic record right? I presume because `value` can be undefined :(

app/packages/looker-3d/src/hooks/use-3d-annotation.ts
```diff
@@ -0,0 +1,128 @@
...
+  );
+
+  useAnnotationEventHandler(
+    "annotation:notification:sidebarLabelHover",
+    useCallback((payload) => {
+      setHoveredLabel({
+        _id: payload.id,
+        path: "",
+        selected: false,
+        _cls: "",
```
tom-vx51: these strings intentionally empty?

app/packages/looker-3d/src/state.ts
```diff
@@ -295,6 +286,16 @@ export const hoveredLabelAtom = atom<OverlayLabel | null>({
 // ANNOTATION RELATED
 // =============================================================================
 
+export const annotationToolbarPositionAtom = atom<number>({
+  key: "fo3d-annotationToolbarPosition",
+  default: 50,
```
tom-vx51, line 296: units/description would be helpful here 🙏

app/packages/core/src/components/Modal/Lighter/useBridge.ts
sashankaryal: Yeah, later refactor. We definitely want to have only ONE event bus.

app/packages/looker-3d/src/hooks/use-3d-annotation.ts
sashankaryal: no because schema IO gives us string booleans in `value` like "false", or "true"

tom-vx51: i meant the `as PolylineLabel` :)

app/packages/looker-3d/src/hooks/use-3d-annotation.ts
sashankaryal: will fix. we only need to stored hovered label id and nothing else

PR #6582, opened 2025-11-17
add remove and find_and_execute APIs to object_3d

sashankaryal:
## Summary by CodeRabbit
* **New Features**
  * Added hashing for 3D objects for set/dict use.
  * Improved equality to require matching type, name, and transform.
  * New scene-graph tools to search/traverse and remove objects by reference, name, or UUID, including recursive removals.

* **Tests**
  * Added comprehensive unit tests covering removal and find-and-execute behaviors across direct and nested scene scenarios.

### DIFF
--- a/fiftyone/core/threed/object_3d.py
+++ b/fiftyone/core/threed/object_3d.py
@@ -81,11 +81,8 @@ class Object3D(object):
         )
 
     def __hash__(self):
-        """Return the hash of this object based on name, visible,
-        and transform matrix."""
-        # convert the matrix to a tuple for hashing
-        matrix_tuple = tuple(self.local_transform_matrix.flatten())
-        return hash((self.name, matrix_tuple))
+        """Return the hash of this object based on its immutable uuid."""
+        return hash(self._uuid)
 
     def __iter__(self):
         return iter(self.children)
@@ -249,6 +246,10 @@ class Object3D(object):
         It can be used for finding and removing nodes, collecting nodes,
         updating nodes, etc.
 
+        The traversal continues into the subtrees of both matching and
+        non-matching nodes. For matching nodes, the subtree is traversed
+        when on_match returns True and stop_on_first_match is False.
+
         Args:
             node: the node to start searching from
             predicate: a function that takes a child Object3D and returns
@@ -276,6 +277,12 @@ class Object3D(object):
             if predicate(child):
                 should_continue = on_match(node, child)
                 if stop_on_first_match or not should_continue:
+                    return True
+                # Recurse into matching child when on_match returns True
+                # and stop_on_first_match is False
+                if self.find_and_execute(
+                    child, predicate, on_match, stop_on_first_match
+                ):
                     return True
             else:
                 if self.find_and_execute(
@@ -376,7 +383,7 @@ class Object3D(object):
                 f"Objects with name '{name}' not found in scene graph"
             )
 
-    def remove_by_uuid(self, uuid: str) -> None:
+    def remove_by_uuid(self, target_uuid: str) -> None:
         """Remove the object with the given UUID from the scene graph recursively.
 
         This method searches recursively through the entire scene graph starting
@@ -384,19 +391,19 @@ class Object3D(object):
         parent's children list. UUIDs should be unique, so only one match is expected.
 
         Args:
-            uuid: the UUID of the object to remove
+            target_uuid: the UUID of the object to remove
 
         Raises:
             ValueError: if attempting to remove this object itself by UUID
             ValueError: if no object with the given UUID is found
         """
-        if self.uuid == uuid:
+        if self.uuid == target_uuid:
             raise ValueError("Cannot remove self from the scene graph")
 
         found = False
 
         def predicate(child: "Object3D") -> bool:
-            return child.uuid == uuid
+            return child.uuid == target_uuid
 
         def on_match(parent: "Object3D", child: "Object3D") -> bool:
             nonlocal found
@@ -411,7 +418,7 @@ class Object3D(object):
 
         if not found:
             raise ValueError(
-                f"Object with UUID '{uuid}' not found in scene graph"
+                f"Object with UUID '{target_uuid}' not found in scene graph"
             )
 
     def _get_asset_paths(self) -> List[str]:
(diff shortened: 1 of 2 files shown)


CamronStaley (approved):

swheaton (changes requested):
👀

fiftyone/core/threed/object_3d.py
```diff
@@ -223,6 +231,196 @@ def clear(self) -> None:
...
+            on_match: a function called when a match is found, takes
+            (parent, child) and returns True to continue searching, False to
+            stop
+            stop_on_first_match: if True, stop searching after first match
+            is processed
+
+        Returns:
+            True if a match was found and we should stop, False otherwise
+
+        Example:
```
swheaton, line 266: You will need a empty line after this in order for the code to render correctly in docs
```suggestion
        Examples::
        
```

fiftyone/core/threed/object_3d.py
```diff
@@ -223,6 +231,196 @@ def clear(self) -> None:
...
+        updating nodes, etc.
+
+        The traversal continues into the subtrees of both matching and
+        non-matching nodes. For matching nodes, the subtree is traversed
+        when on_match returns True and stop_on_first_match is False.
+
+        Args:
+            node: the node to start searching from
+            predicate: a function that takes a child Object3D and returns
+            True if it matches the search criteria
```
swheaton, line 256:
```suggestion
            predicate: a function that takes a child :class:`Object3D` and returns
                ``True`` if it matches the search criteria
```

fiftyone/core/threed/object_3d.py
```diff
@@ -223,6 +231,196 @@ def clear(self) -> None:
...
+            on_match: a function called when a match is found, takes
+            (parent, child) and returns True to continue searching, False to
+            stop
```
swheaton, line 259:
```suggestion
            on_match: a function called when a match is found, takes
                (parent, child) and returns True to continue searching, False to
                stop
```

fiftyone/core/threed/object_3d.py
```diff
@@ -223,6 +231,196 @@ def clear(self) -> None:
...
+            stop_on_first_match: if True, stop searching after first match
+            is processed
```
swheaton, line 261:
```suggestion
            stop_on_first_match: if ``True``, stop searching after first match
                is processed
```

fiftyone/core/threed/object_3d.py
```diff
@@ -223,6 +231,196 @@ def clear(self) -> None:
...
+
+        Returns:
+            True if a match was found and we should stop, False otherwise
```
swheaton, line 264:
```suggestion
            ``True`` if a match was found and we should stop, ``False`` otherwise
```

fiftyone/core/threed/object_3d.py
```diff
@@ -223,6 +231,196 @@ def clear(self) -> None:
...
+        when on_match returns True and stop_on_first_match is False.
+
+        Args:
+            node: the node to start searching from
+            predicate: a function that takes a child Object3D and returns
+            True if it matches the search criteria
+            on_match: a function called when a match is found, takes
+            (parent, child) and returns True to continue searching, False to
+            stop
+            stop_on_first_match: if True, stop searching after first match
```
swheaton, line 260: Why have this distinct from `on_match` ? It seems like if `on_match` just always returns ``False`` then it would be the same effect.

fiftyone/core/threed/object_3d.py
```diff
@@ -223,6 +231,196 @@ def clear(self) -> None:
...
+    ) -> bool:
+        """Recursively search the scene graph and execute an action on
+        matching nodes.
+
+        This is a generic method for traversing the scene graph and
+        performing operations on nodes that match a given predicate.
+        It can be used for finding and removing nodes, collecting nodes,
+        updating nodes, etc.
+
+        The traversal continues into the subtrees of both matching and
```
swheaton, line 249: Maybe want to note that it does a depth first search (as opposed to BFS) so that could affect expected outcome if `stop_on_first_match==True`

fiftyone/core/threed/object_3d.py
```diff
@@ -223,6 +231,196 @@ def clear(self) -> None:
...
+            # Find all nodes with a specific name and collect them
+            matches = []
+            def predicate(child):
+                return child.name == "target"
+            def on_match(parent, child):
+                matches.append(child)
+                return True  # continue searching
+            scene.find_and_execute(scene, predicate, on_match)
+        """
+        for child in list(node.children):
```
swheaton, line 276: nit: could use `node.traverse(include_self=False)`

fiftyone/core/threed/object_3d.py
```diff
@@ -223,6 +231,196 @@ def clear(self) -> None:
...
+        Returns:
+            True if the target was found and removed, False otherwise
+        """
+        for i, child in enumerate(parent.children):
+            if child is target:
+                del parent.children[i]
+                return True
+        return False
+
+    def remove(self, *objs: "Object3D") -> None:
```
swheaton, line 312: These are all very similar but not exact. I would like to consolidate but I'm not gonna force it

fiftyone/core/threed/object_3d.py
```diff
@@ -223,6 +231,196 @@ def clear(self) -> None:
...
+            return
+
+        if any(o is self for o in objs):
+            raise ValueError("Cannot remove self from the scene graph")
+
+        objs_to_remove = set(objs)
+        removed = set()
+
+        def predicate(child: "Object3D") -> bool:
+            return any(child is obj for obj in objs_to_remove)
```
swheaton, line 336: since you ensured the hash function just uses UUID, you can probably just do this
```suggestion
            return child in objs_to_remove
```

fiftyone/core/threed/object_3d.py
swheaton: although on second thought, the `list()` is critical because it's creating a copy of the children. Otherwise you would not be able to remove in place while iterating.
This might be wasteful if the objects were big, but we know they are not

tests/unittests/threed/object_3d_tests.py
```diff
@@ -314,3 +314,491 @@ def test_traverse(self):
         self.assertListEqual(
             list(sub.traverse(include_self=False)), objects[2:4]
         )
+
+
+class TestObject3DRemove(unittest.TestCase):
```
swheaton, line 319: lot of tests, thanks! Looks like AI written tests lol, but if you did it all manually, good on you sir

swheaton (approved):
a lot of nits and docstring formatting. doesn't have to block

PR #6585, opened 2025-11-17
Storage as submodule instead of single file

j053y:

Make storage a sub-module to allow for better extension.

n/a

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/core/storage/__init__.py
+++ b/fiftyone/core/storage/__init__.py
@@ -1,0 +1,9 @@
+"""
+File storage utilities.
+
+| Copyright 2017-2025, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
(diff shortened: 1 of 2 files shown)


fiftyone/core/storage/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""
+File storage utilities.
+
+| Copyright 2017-2025, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
+from fiftyone.core.storage.storage import *
```
brimoor, line 9: In other subpackages we've used this pattern to ensure that Sphinx refs continue to work. Can you add that here too?

https://github.com/voxel51/fiftyone/blob/dcb61dfa6b26a9d4299337e65aa66320a44c3ad6/fiftyone/core/map/__init__.py#L16-L21

PR #6592, opened 2025-11-19
add support for custom view target in ctx.target_view

tom-vx51:

This PR adds support for specifying a custom view target which can be resolved via `ctx.target_view()`.

This option adds flexibility to operators in cases where the desired view is known, but distinct from the session's _current_ view.

local, unit

        notes for FiftyOne users.

Added option to provide a custom view via context parameters when invoking an operator programmatically.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Operators now support custom view targets, enabling flexible view configuration control. Users can define custom view behaviors by passing JSON-serialized view stage definitions through the custom_view_target parameter when invoking operators. This enables caller-specified custom viewing options and dynamic view modification at runtime without requiring changes to operator code.

### DIFF
--- a/fiftyone/operators/executor.py
+++ b/fiftyone/operators/executor.py
@@ -717,6 +717,12 @@ class ExecutionContext(contextlib.AbstractContextManager):
             return self.view.select_labels(self.selected_labels)
         if target == constants.ViewTarget.DATASET_VIEW:
             return self.dataset.view()
+        if target == constants.ViewTarget.CUSTOM_VIEW_TARGET:
+            if (
+                view_stages := self.params.get("custom_view_target")
+            ) is not None:
+                # pylint: disable-next-line=protected-access
+                return fov.DatasetView._build(self.dataset, view_stages)
 
         return self.view if self.has_custom_view else self.dataset
 
--- a/fiftyone/operators/constants.py
+++ b/fiftyone/operators/constants.py
@@ -43,3 +43,10 @@ class ViewTarget:
 
     SELECTED_SAMPLES = "SELECTED_SAMPLES"
     """Selected samples in the app view, if any."""
+
+    CUSTOM_VIEW_TARGET = "CUSTOM_VIEW_TARGET"
+    """Custom view target specified by the caller.
+
+    When using this option, specify 'custom_view_target' in the
+    operator parameters with a list of JSON-serialized view stages to apply.
+    """
--- a/tests/unittests/operators/executor_tests.py
+++ b/tests/unittests/operators/executor_tests.py
@@ -103,6 +103,7 @@ class TestOperatorExecutionContext(unittest.TestCase):
                     constants.ViewTarget.SELECTED_SAMPLES,
                     view.select([selected]),
                 ),
+                (constants.ViewTarget.CUSTOM_VIEW_TARGET, ds.limit(5)),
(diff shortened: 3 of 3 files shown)


swheaton:
How do you plan to use this?

fiftyone/operators/executor.py
```diff
@@ -717,6 +717,12 @@ def target_view(self, param_name="view_target"):
             return self.view.select_labels(self.selected_labels)
         if target == constants.ViewTarget.DATASET_VIEW:
             return self.dataset.view()
+        if target == constants.ViewTarget.CUSTOM_VIEW_TARGET:
+            if (
+                view_stages := self.params.get("custom_view_target")
```
imanjra, line 722: This would make this param name a reserved keyword. My suggestion would be to go with one of two approaches below:

1. Use another request parameter (below [operators.ts#L606](https://github.com/voxel51/fiftyone/blob/feat/custom-view-target/app/packages/operators/src/operators.ts#L606) which are explicit and not controlled by the user or operator's inputs/schema

2. [Recommended/TypicalOperatorPattern] Introduce a new lifecycle method `resolve_view` which lets the user provide a custom view arbitrarily like below:

```py
    def resolve_target_view(self, ctx):
        my_custom_view_serialized = ctx.params.get("my_custom_view", None)
        
        view = DatasetView.build(...)

        return view

    def resolve_input(self, ctx):
        inputs = types.Object()

        inputs.list("my_custom_view", view=types.HiddenView()) # might be inputs.obj

        return types.Property(inputs)
```

_Edit: fixed example_

tom-vx51: I considered the reserved keyword concern, but IMO it's not an issue since it is only the case when you explicitly use `CUSTOM_VIEW_TARGET`, and so it would never "accidentally" conflict with user behavior.

Alternative (1) won't work afaik because the intent is specifically for a user to be able to control this view; it's distinct from any value inferred from the current app session.

I agree that (2) is more idiomatic w.r.t the current operator patterns, but I'm not sure how it would be used in practice.

`target_view` is a method of `ExecutionContext`, while the proposed `resolve_target_view` is within the operator. We'd currently do something like this:

```python
def execute(self, ctx):
    view = ctx.target_view()
    # use the view
``` 

We *could* move to a new paradigm of `resolve_target_view`, in which case we'd do

```python
def execute(self, ctx):
    view = self.resolve_target_view(ctx)
```

but this would be the same behavior as we have today, where we could directly just do

```python
def execute(self, ctx):
    view = self._get_view(ctx)  # some custom logic
```

It's not clear to me how the operator framework would leverage `ExecutionContext::target_view` and `Operator::resolve_target_view` without them being in conflict with one another. The impetus for this change is to allow for custom views in cases where the operator itself does not have explicit control (e.g. within a `PipelinOperator`'s child).

swheaton: Currently the API loads the `ExecutionContext` and calls `target_view()` to get the view to split into N pieces (if distributed). I'm not that keen on adding `resolve_target_view` because then we are adding the running of arbitrary custom user code where we otherwise did not have it.

imanjra: Sounds good. With a combination of `CUSTOM_VIEW_TARGET` as `target` and that param, it would be fairly explicit and should avoid unintended view override

PR #6594, opened 2025-11-19
Add HRM2.0 (4D-Humans) integration for 3D human mesh reconstruction

lorenzoriano:

This PR integrates the HRM2.0 (4D-Humans) model into FiftyOne, enabling state-of-the-art 3D human mesh reconstruction from single images.

Key changes include:
-   **New Utils Module**: Added `fiftyone/utils/hrm2.py` containing:
    -   `HRM2Model`: Wrapper for the 4D-Humans inference pipeline.
    -   `HRM2OutputProcessor`: Converts raw model outputs into FiftyOne labels.
    -   `apply_hrm2_to_dataset_as_groups`: Utility to process datasets and automatically structure them as grouped samples (linking 2D images with 3D reconstructions).
-   **Model Zoo**: Registered `hrm2-torch` (version 2.0b) in `fiftyone/zoo/models/manifest-torch.json`.
-   **Documentation**: Added integration documentation in `docs/source/integrations/hrm2.rst`.

-   Added unit tests in `tests/unittests/hrm2_tests.py` covering model loading, config parsing, and output processing.
-   Added intensive tests in `tests/intensive/hrm2_tests.py` for end-to-end inference checks.

        This PR integrates the HRM2.0 (4D-Humans) model into FiftyOne, enabling state-of-the-art 3D human mesh reconstruction from single images.

Added integration for the HRM2.0 (4D-Humans) model, enabling state-of-the-art 3D human mesh reconstruction from images directly within FiftyOne.

## Summary by CodeRabbit

* **New Features**
  * HRM2.0 human-pose integration: model loading, single/multi-person inference, SMPL mesh generation, BODY-25-like skeleton, 2D/3D keypoints, and 3D scene export; dataset grouping for linked 2D+3D samples
  * Image/tensor utilities: conversions, resizing, normalization, and detection-to-box helpers
  * Structured-input inference and runtime-output configuration merging
  * Utility to convert NumPy types to native Python types

* **Tests**
  * Added intensive and unit test suites covering HRM2 workflows, mesh export, utilities, and skeleton validation


fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,1947 @@
...
+
+    Args:
+        smpl_model_path: path to SMPL_NEUTRAL.pkl file
+        device: device to load model on
+        **kwargs: additional arguments
+
+    Returns:
+        loaded SMPL model
+    """
+    try:
```
brimoor: nit: we have a `lazy_import()` util that we tend to use so that all imports can be placed at the top of the module, for readability: 

```py
smplx = fou.lazy_import("smplx", callback=lambda: fou.ensure_import("smplx"))
```

A couple example usages:

https://github.com/voxel51/fiftyone/blob/df3508580ce7c6c37d540a4998e06e3698745a29/fiftyone/utils/sam2.py#L26

https://github.com/voxel51/fiftyone/blob/df3508580ce7c6c37d540a4998e06e3698745a29/fiftyone/utils/coco.py#L36-L38

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,1947 @@
...
+    dataset.apply_model(
+        model,
+        label_field=label_field,
+        batch_size=batch_size,
+        num_workers=num_workers,
+        output_dir=output_dir,
+    )
+
+    # Collect predictions with scene paths
+    predictions = []
```
brimoor: this can be expressed idiomatically as:

```py
predictions = dataset.values(label_field)
```

^much more efficient than iterating over the entire dataset(!)

docs/source/integrations/hrm2.rst
```diff
@@ -0,0 +1,263 @@
...
+
+    import fiftyone as fo
+    import fiftyone.zoo as foz
+    import fiftyone.utils.hrm2 as fouh
+
+    dataset = fo.load_dataset("your-dataset")
+
+    # 1. Generate person detections
+    detector = foz.load_zoo_model("yolov8n-coco-torch")
+    dataset.apply_model(detector, label_field="detections")
```
brimoor: We [just added](https://github.com/voxel51/fiftyone/pull/6494) a builtin syntax for filtering model outputs by class!

```py
dataset.apply_model(detector, label_field="detections", classes=["person"])
```

Can you use this to be idiomatic? 😄

fiftyone/core/odm/utils.py
```diff
@@ -492,50 +493,141 @@ def _parse_embedded_doc_fields(doc, dynamic):
...
         fields = _parse_embedded_doc_fields(value, dynamic)
-        _merge_embedded_doc_fields(fields_dict, fields)
+        fields_dict = _merge_embedded_doc_fields(fields_dict, fields)
 
     return _finalize_embedded_doc_fields(fields_dict)
 
 
-def _merge_embedded_doc_fields(fields_dict, fields):
-    for field in fields:
+def _merge_embedded_doc_fields(
```
brimoor: I think you ran into the same bug that I did when testing this PR; I put up a fix for it here: https://github.com/voxel51/fiftyone/pull/6609

fiftyone/core/labels.py
```diff
@@ -440,11 +449,22 @@ class Detection(_HasAttributesDict, _HasID, _HasMedia, _HasInstance, Label):
 
     Args:
         label (None): the label string
-        bounding_box (None): a list of relative bounding box coordinates in
-            ``[0, 1]`` in the following format::
+        bounding_box (None): bounding box coordinates. The format depends on
```
brimoor: ## TLDR

I do not want us to introduce the `relative_coordinate` option here.

## Context

By convention, we only introduce new `Label` classes and/or make changes to existing `Label` classes when the rest of the project **fully supports** these options.

In this case, the `Detection` class must support:
- visualization object detections in the App
- exporting object detections via `export()`
- running things like model evaluation via `evaluate_detections()`
- etc

It would be quite a lot of work to fully update all of these things to support `Detection` whose bounding boxes are stored in absolute coordinates, as each of these code paths would need to introduce conditionals that would process the detection object different depending on whether its bounding box is stored as relative or absolute coordinates.

^for this reason, we have taken the stance that FiftyOne should provide a single, opinionated `Label` format for each task type that provides a consistent data format from which the entire library can build against.

fiftyone/core/labels.py
```diff
@@ -1128,6 +1149,36 @@ class Keypoints(_HasLabelList, Label):
     keypoints = fof.ListField(fof.EmbeddedDocumentField(Keypoint))
 
 
+class Keypoint3D(EmbeddedDocument):
```
brimoor: Here's how 3D cuboids and 3D polylines are implemented:
- https://docs.voxel51.com/user_guide/using_datasets.html#d-detections
- https://docs.voxel51.com/user_guide/using_datasets.html#d-polylines

For consistency, I think we should adopt the convention of representing 3D keypoints using the `Keypoint` class where (for consistency with 3D polylines), we'd ask users to store the points in a `points3d` attribute.

I know its a bit strange that we didn't introduce `Detection3D` and `Polyline3D` classes. The reason we did not is that when working [group datasets](https://docs.voxel51.com/user_guide/groups.html), each slice (image, 3D, etc) must share a single field schema, and we wanted to allow users working with multimodal datasets to reuse a single field name (eg `ground_truth_boxes`) to store the boxes for their image slices and cuboids for their 3D slices. The way we made that possible was to adopt a convention where the `Detection` and `Polyline` classes can support either 2D or 3D.

Note that, per my previous comment about `Label` classes, ideally we would not use/document a `Keypiont(..., points3d=)` syntax until the App supports visualizing 3D keypoints stored in this format.

fiftyone/core/labels.py
```diff
@@ -1563,13 +1614,46 @@ def from_geo_json(cls, d):
...
 
 
-_LABEL_LIST_FIELDS = (
-    Classifications,
-    Detections,
-    Keypoints,
-    Polylines,
-    TemporalDetections,
-)
+class Camera(_HasID, Label):
```
brimoor: By convention, we only use the `Label` class for things that are visualized in the App.

When you just want to define a "struct" of data, use `DynamicEmbeddedDocument`:

```
class Camera(DynamicEmbeddedDocument):
```

Prior art: the [`Metadata` class](https://github.com/voxel51/fiftyone/blob/adf5e6f2db0fb1ccaad8ddfb4dd1d320c67069b8/fiftyone/core/metadata.py#L31).

fiftyone/core/labels.py
```diff
@@ -2103,3 +2188,547 @@ def _from_geo_json(d):
         polygons = None
 
     return points, lines, polygons
+
+
+class SMPLParams(EmbeddedDocument):
```
brimoor: Using `DynamicEmbeddedDocument` instead of `EmbeddedDocument` is usually preferred because it allows users to store additional attributes on these documents when using this class to store data on their samples.

Another reason is backwards/forwards compatibility. Like protocol buffers, we want the flexibility of attributes being added or removed in the future without causing validation errors.

fiftyone/core/labels.py
```diff
@@ -2103,3 +2188,547 @@ def _from_geo_json(d):
...
+            associated with this instance
+        instance (None): optional :class:`Instance` linking this 3D instance to
+            a 2D detection, segmentation, or other label in a grouped dataset
+        attributes (None): optional dictionary of free-form attributes such as
+            model scores, model names, or other metadata. Model-specific
+            parameters (e.g., SMPL body pose) should be stored here rather
+            than in dedicated fields.
+    """
+
+    instance_id = fof.IntField()
```
brimoor: Initially we started off using an int `index` attribute to represent instances for `Label` classes:
https://github.com/voxel51/fiftyone/blob/adf5e6f2db0fb1ccaad8ddfb4dd1d320c67069b8/fiftyone/core/labels.py#L469

But then we pivoted to having a UUID `instance` attribute of type `fo.Instance()`:
https://docs.voxel51.com/user_guide/groups.html#linking-labels-across-slices

`Label` classes that have instances implement the `_HasInstance` mixin:
https://github.com/voxel51/fiftyone/blob/adf5e6f2db0fb1ccaad8ddfb4dd1d320c67069b8/fiftyone/core/labels.py#L368-L384

Side note: you may be wondering why `_HasInstance` does not literally declare an `instance = fof.EmbeddedDocumentField(Instance)` field, or why we did not delete the `index` attribute once we introduced `instance`. It was for backwards-compatibility reasons which are probably best discussed IRL.

fiftyone/core/labels.py
```diff
@@ -2103,3 +2188,547 @@ def _from_geo_json(d):
...
+    ):
+        super().__init__(**kwargs)
+        self.pose = pose
+        self.detection = detection
+
+
+class HumanPoses2D(_HasLabelList, Label):
+    """A list of 2D human poses in an image.
+
+    This wrapper class provides proper schema registration for lists of
```
brimoor: Woah does the App actually support visualizing boxes and keypoints in `HumanPoses2D` fields in this format today?

I didn't think it was smart enough to automatically traverse lists of embedded documents whose *attributes* are `Label` classes!

fiftyone/core/labels.py
```diff
@@ -2103,3 +2188,547 @@ def _from_geo_json(d):
...
+        if update:
+            self.scene_path = scene_path
+
+        return scene_path
+
+
+class HumanPose2D(Label):
+    """2D human pose keypoints and detection for image samples.
+
+    This class wraps native FiftyOne label types (Keypoint and Detection) to
```
brimoor: FYI- I put this gist together showing the different data model options we have for representing pairs of boxes + keypoints and the pros/cons of each:
https://gist.github.com/brimoor/a33871d0b6130e6714bbf5ceb352c23d

Note that it has some other goodies in it:
- how to store [keypoint skeletons](https://docs.voxel51.com/user_guide/using_datasets.html#storing-keypoint-skeletons)
-  suggests a pattern of populating the `instance` attribute of the boxes/keypoints so that they are "linked" in the App, which provides some nice behaviors (consistent coloring, shift+click to select, etc) 

However, my gist might be slightly moot if the App is already smart enough to render data in your `HumanPoses2D` class!

fiftyone/core/labels.py
```diff
@@ -2103,3 +2188,547 @@ def _from_geo_json(d):
...
+class MeshInstances3D(_HasLabelList, Label):
+    """A list of 3D instances for a reconstructed scene.
+
+    This label type is a container for :class:`MeshInstance3D` instances associated with a single sample
+    or grouped sample. It also stores scene-level metadata such as a `.fo3d`
+    scene path and an optional scene-level camera.
+
+    Args:
+        instances (None): a list of :class:`MeshInstance3D` instances
+        scene_path (None): optional path to a `.fo3d` scene file that contains
```
brimoor: Declaring a `scene_path` attribute here is a bit of an anti-pattern.

In FiftyOne, the `filepath` field of 3D samples must be the `.fo3d` file for the scene, which would mean that any `Sample` that has a `MeshInstances3D` field would have duplicate FO3D files on it.

fiftyone/core/models.py
```diff
@@ -727,6 +697,12 @@ def _do_export_array(label, input_path, filename_maker):
             input_path, output_ext=".png"
         )
         label.export_map(map_path, update=True)
+    elif isinstance(label, fol.SMPLHumanPoses):
```
brimoor: Note to self: chat live to understand what's going on here

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2325 @@
...
+        gender="neutral",
+        batch_size=1,
+        create_transl=False,
+    )
+    logger.info("Loaded SMPL model from %s", smpl_model_path)
+
+    return model
+
+
+def apply_hrm2_to_dataset_as_groups(
```
brimoor, line 1626: By convention, whenever a method can be applied to a `Dataset` or `DatasetView`, we document this by declaring that the method accepts any `SampleCollection` as input:

```py
def apply_hrm2_to_dataset_as_groups(
    sample_collection: "fo.SampleCollection",
    ...
)
```

Also for consistency with other methods, `sample_collection` is always the *first argument*.

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2325 @@
...
+    if dataset.media_type == "group":
+        logger.warning(
+            "Dataset is already grouped. This will add new slices to "
+            "existing groups."
+        )
+
+    # Store original sample info before clearing
+    logger.info("Processing %d samples...", len(dataset))
+    original_data = []
+    for sample in dataset.iter_samples():
```
brimoor: Reading all samples into memory is an anti-pattern in general, as dataset contents could be way too large to fit in RAM.

A more idiomatic solution would be to create your new grouped dataset and then merge the original sample fields over via `merge_samples()`:

```py
# Merges samples with the same `filepath`
grouped_dataset.merge_samples(dataset, insert_new=False)
```

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -8406,6 +8409,49 @@
...
             "date_added": "2025-07-09 12:00:00"
+        },
+        {
+            "base_name": "hrm2-torch",
+            "base_filename": null,
+            "version": "2.0b",
+            "description": "Reconstructs 3D human body meshes from images using the SMPL body model with state-of-the-art accuracy",
+            "source": "https://github.com/shubham-goel/4D-Humans",
+            "author": "Shubham Goel, et al.",
+            "license": "BSD 3-Clause",
```
brimoor: I see MIT license on the GH repo: https://github.com/shubham-goel/4D-Humans

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -8406,6 +8409,49 @@
...
+            "size_bytes": null,
```
brimoor: Could you populate the `size_bytes` field with the size of the model on disk? We are pretty diligent about providing this information as it is shown to users in the docs. EG:
https://docs.voxel51.com/model_zoo/models/yolo11m_coco_torch.html

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -8406,6 +8409,49 @@
                 { "name": "MS COCO 2017", "url": "https://cocodataset.org" }
             ],
             "date_added": "2025-07-09 12:00:00"
+        },
+        {
+            "base_name": "hrm2-torch",
+            "base_filename": null,
+            "version": "2.0b",
```
brimoor: I would recommend setting model's `version=null`. The `version` key does exist in the framework, but we haven't been using it and I'm honestly not sure if it fully works...

For context, the `version` key was indeed intended for situations like this where you have multiple checkpoints for the same model architecture and you want to register them in the zoo under a common `base_name` in such a way where the highest versioned checkpoint is always used by default. However, in practice the `checkpoint_version` config parameter you added would already support this:

```py
model = foz.load_zoo_model("hrm2-torch", checkpoint_version="2.0a")
```

brimoor (changes requested):
@lorenzoriano this is incredibly thorough work! 💪💪💪

I'd like to sync live with you + some other folks on the team to identify what data model we should use to store the human pose, camera, meshes, etc data that you're introducing in this PR.

Introducing new `Label` and `EmbeddedDocument` classes is not something that we do lightly as we need to be sure that we're willing to live with them forever in a backwards/forwards compatible way 😄

fiftyone/core/odm/utils.py
lorenzoriano: I rebased on the latest `develop` but https://github.com/voxel51/fiftyone/pull/6609 doesn't quite solve the issue I was facing. I also have added unit tests to cover that specific case

lorenzoriano: Since I reverted the changes in labels.py this is not necessary anymore, but I think it'd be good to leave it if in the future we will want to have composite labels

fiftyone/zoo/models/manifest-torch.json
lorenzoriano: done

fiftyone/zoo/models/manifest-torch.json
lorenzoriano: done

fiftyone/zoo/models/manifest-torch.json
lorenzoriano: done

docs/source/integrations/hrm2.rst
lorenzoriano: done

fiftyone/core/odm/utils.py
brimoor: @lorenzoriano hmm, can you provide a minimal reproducible example here that shows the bug you're fixing? Like I did in the description of https://github.com/voxel51/fiftyone/pull/6609?

lorenzoriano: TL;DR I added test_deeply_nested_embedded_doc_merging which shows the bug.

The issue was with 3+ level nested embedded documents. When you have something like:

person.pose.joint.angle  # 3 levels deep

The old code would fail on the second merge because:
1. First sample adds nested fields as a list (that's what _parse_embedded_doc_fields
returns)
2. Second sample tries to merge, recursive call hits fields_dict[name] = field
3. But fields_dict is still a list at that point → TypeError: 'list' object does not support
item assignment

The fix normalizes everything to dicts before merging so the recursive calls work correctly.

Note that this fix is not related to this PR anymore, as I'm not adding new labels. I will probably separate this code and make a different PR

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+            If True, HRM2Person documents will contain vertex data for export.
+            Actual files are written when output_dir is provided to apply_model()
+        detections_field (None): optional field name containing person detections
+            to use for multi-person processing. If provided, HRM2 will process
+            each detected person separately. If None, processes the full image
+            as single-person mode
+        **kwargs: additional parameters for :class:`TorchImageModelConfig`
+    """
+
+    def __init__(self, d: Dict[str, Any]) -> None:
```
jacobsela, line 1822: The arguments here that are shared across every model should be in the manifest.

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+                ),  # Allow explicit override
+                "model_config_path": d.get(
+                    "model_config_path"
+                ),  # Allow explicit override
+                "init_renderer": False,
+            }
+        )
+
+        # Enable ragged batches for heterogeneous data from GetItem
+        d["ragged_batches"] = True
```
jacobsela: This disables batching. Is this required?

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+        if self.smpl_model_path is not None:
+            if not os.path.exists(self.smpl_model_path):
+                raise ValueError(
+                    f"SMPL model not found at {self.smpl_model_path}. "
+                    f"Please register at https://smpl.is.tue.mpg.de/ to "
+                    f"obtain the SMPL_NEUTRAL.pkl file."
+                )
+
+
+class HRM2ModelConfig(HRM2Config):
```
jacobsela: just consolidate into one class

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+        self.smpl_model_path = self.parse_string(
+            d, "smpl_model_path", default=None
+        )
+        self.checkpoint_version = self.parse_string(
+            d, "checkpoint_version", default="2.0b"
+        )
+        self.export_meshes = self.parse_bool(d, "export_meshes", default=True)
+
+        # Detections field for multi-person processing
+        self.detections_field = self.parse_string(
```
jacobsela: Why is this an attribute of the class? Shouldn't this be a runtime argument?

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+        # Set instance attributes BEFORE parent init (which calls _load_model)
+        self._hmr2: Optional[Any] = None
+        self._smpl: Optional[Any] = None
+        self._preprocessor: Optional[_HRM2CropHelper] = None
+
+        # Now initialize parent class
+        fout.TorchImageModel.__init__(self, config)
+
+    def _download_model(self, config: HRM2Config) -> None:
+        """Download model if it's a zoo model."""
```
jacobsela: This should just call the download script wrapper above right?

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+        **Loading Order (Critical for Output Processor Pattern):**
+        1. Load HRM2 model via parent's entrypoint
+        2. Load SMPL model (runtime resource needed by output processor)
+        3. Build output processor and inject SMPL model + device
+
+        This order allows the output processor to receive runtime resources
+        (SMPL model, device) that can't be serialized in the config.
+        """
+        # Load HRM2 model using parent's entrypoint mechanism
+        self._hmr2 = super()._load_model(config)
```
jacobsela, line 1978: why is this a new variable rather than just being `self._model`?

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+                img_shape,
+                smpl_faces,
+                bundle_frame_size,
+            )
+            results.append(bundle)
+
+        return results
+
+    @staticmethod
+    def _process_person(person_raw: Dict[str, Any]) -> Dict[str, Any]:
```
jacobsela, line 1065: Why not do this on the GPU if it has to be done? This doesn't matter if runtime is similar across gpu/cpu but it feels strange to me to have such a large operation in the output processing.

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+
+        This order allows the output processor to receive runtime resources
+        (SMPL model, device) that can't be serialized in the config.
+        """
+        # Load HRM2 model using parent's entrypoint mechanism
+        self._hmr2 = super()._load_model(config)
+
+        # Load SMPL model separately if configured
+        # This is a runtime resource that will be injected into the output processor
+        if config.smpl_model_path:
```
jacobsela, line 1982: if this is only used by the output processor why build this here? Shouldn't the out processor own this?

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+
+        Args:
+            field_mapping: optional dict mapping required_keys to dataset
+                field names. If not provided, will attempt to auto-configure from
+                the model's config.
+
+        Returns:
+            an :class:`HRM2GetItem` instance
+        """
+        # Copy field_mapping to avoid mutating caller's dict
```
jacobsela: All of this logic should be left to the user IMO.

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+        Args:
+            img: input image as PIL.Image, numpy array (HWC), or torch.Tensor
+
+        Returns:
+            Dict containing Keypoints, Detections, HRM2Person documents,
+            and export data (smpl_faces, frame_size)
+        """
+        return self.predict_all([img])[0]
+
+    def predict_all(
```
jacobsela: Is this override necessary? I think usage of `GetItem` and `collate_fn` should be enough here.

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+            batch_data: List of dicts from GetItem with 'image', 'detections', 'filepath'
+
+        Returns:
+            List of raw output dicts, one per image, each containing:
+                - "people": List[person_raw_dict] with raw tensors
+                - "img_shape": (height, width) of source image
+        """
+        raw_outputs = []
+
+        for data in batch_data:
```
jacobsela: Please no for loops over the batch dimension in the forward. That logic should be in `predict_all` if the model doesn't support batching, or this code should be batched if the model supports it.

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+            img = data["image"]
+            detections = data.get("detections")
+
+            # Process based on detection mode
+            if detections is not None and len(detections.detections) > 0:
```
jacobsela: This feels like it should be two different models.

jacobsela: Or alternatively to have an explicit `mode` argument, but deciding on behavior implicitly like this feels strange.

fiftyone/utils/hrm2.py
```diff
@@ -0,0 +1,2601 @@
...
+                output = self._inference_single_person(img)
+
+            # Carry filepath through for UID generation in output processor
+            output["filepath"] = data.get("filepath")
+
+            raw_outputs.append(output)
+
+        return raw_outputs
+
+    def _run_inference(self, img_t: torch.Tensor) -> Dict[str, Any]:
```
jacobsela: this should be the `_forward_pass`

jacobsela (changes requested):
Stopping here even though I've only gone through about half of the code.

The TL:DR is like this:
1. It's really important that we avoid overriding base class methods with our model integrations when possible. In the past this has been a huge pain point in maintaining and using the integrations in production. If the code is intended to be a "production" model that we can use in our tools, special care must be taken to have clean code that respects our OOP as much as possible.
2. The preprocess --> predict --> forward pass --> post process chain of events is especially important to cleanly maintain because strictly it allows us to (a) write clean, generic inference code and (b) easily distribute said code to good target hardware.

What I would recommend is that we go over this code together and figure out which parts can be consolidated (so that we can inherit from `super`) and which parts are actually unique. It may very well be possible that certain things that you are trying to do are not currently easy, or even possible at all to do with our current base classes. This is expected. In these cases we should either do some work on our base classes, or figure out creative ways of working around these issues.

One final point: it seems that a lot of code here is trying to resolve things for the user and work in many cases. I think a better approach is to have lean, simple code that is highly resusable. By leaving room a lot of room for modularity and configurability, the user can have a far easier time matching the generic code to their use case. A specific example is having the model automatically toggle inference modes depending on if `detections` are passed in, or creating the groups for the user. That code should be presented as examples, not as part of the integration code.

Let's meet to discuss sync when you have a chance.

fiftyone/utils/hrm2.py
lorenzoriano: done

fiftyone/utils/hrm2.py
lorenzoriano: fixed

fiftyone/utils/hrm2.py
lorenzoriano: Not sure what logic you're referring to, but I think having layers of input validation and augmentation is always useful

fiftyone/utils/hrm2.py
lorenzoriano: The base TorchImageModel._predict_all only accepts image-like inputs and strips everything else. It stacks or forwards them and passes only width, height to the output processor. I need to carry per-sample detection metadata into _forward_pass or _output_processor. Ideally we would change  TorchImageModel so that it can accept a richer input (like a dict, I believe that's what pytorch lightning does for example). I've modified torch.py to support the new interface

fiftyone/utils/hrm2.py
lorenzoriano: done

fiftyone/utils/hrm2.py
lorenzoriano: removed the single person mode entirely

fiftyone/utils/hrm2.py
lorenzoriano: done

fiftyone/utils/hrm2.py
lorenzoriano: it has to handle hrm2 and sbpl

fiftyone/utils/hrm2.py
lorenzoriano: it can be overriden via field_mapping. This is just a convenience, and follows patterns in the codebase

fiftyone/utils/hrm2.py
lorenzoriano: done

PR #6595, opened 2025-11-19
Perception Encoder Integration

jacobsela:

Starting an integration for [Perception Encoder](https://github.com/facebookresearch/perception_models/tree/main/apps/pe).

Only doing the ViT encoder of it now, can follow up with the rest in future PRs.

Install from the fork: https://github.com/voxel51/perception-models

I recommend running a visualization with this model on `foz.load_zoo_dataset("imagenet-sample")`

```python
import torch
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.pe as fope

dataset = foz.load_zoo_dataset("quickstart", persistent=True)

model = foz.load_zoo_model(
    "PE-Core-L14-336-Vision-Encoder", device="cuda:7", project=True,
)

dataset.compute_embeddings(
    model,
    "pe_integration_test",
    batch_size=32,
    num_workers=8,
)

samples = dataset[:10]

embeddings = np.array(samples.values("pe_integration_test"))

import perception_models.core.vision_encoder.pe as pe
import perception_models.core.vision_encoder.transforms as pet
from PIL import Image

model = pe.VisionTransformer.from_config(
    name="PE-Core-L14-336", pretrained=True
).to("cuda:7")
transform = pet.get_image_transform(image_size=532)

filepaths = samples.values("filepath")
images = [transform(Image.open(fp)).unsqueeze(0).to("cuda:7") for fp in filepaths]
images = torch.cat(images, dim=0)
with torch.no_grad():
    outputs = model(images).cpu().numpy()

passed = np.allclose(embeddings, outputs, rtol=1e-5, atol=1e-5)
print("PE integration test passed!" if passed else "PE integration test failed!")

diff = np.abs(embeddings - outputs)
print(f"diff max = {diff.max()}")
print(f"diff mean = {np.mean(diff)}")
print(f"diff std = {np.std(diff)}")
```

prints:
```
PE integration test passed!
diff max = 1.9073486328125e-06
diff mean = 1.7749498226748983e-07
diff std = 1.479596249964743e-07
```

        notes for FiftyOne users.

Added an integration with Perception Encoder (only ViT encoder for now).

-   [x] Other

## Summary by CodeRabbit

* **New Features**
  * Adds Perception Encoder support to load and run Meta vision models within FiftyOne.
  * Enables automatic checkpoint download from HuggingFace and sensible defaults for checkpoint paths and image transforms.
  * Provides configurable inference options — pooling, projection, normalization, token handling — and flexible layer selection for generating embeddings.

### DIFF
--- a/fiftyone/utils/pe.py
+++ b/fiftyone/utils/pe.py
@@ -9,8 +9,8 @@ fou.ensure_package("huggingface_hub>=0.20.0")
 from huggingface_hub import hf_hub_download
 
 fou.ensure_package("perception_models")
-import core.vision_encoder.pe as pe
-import core.vision_encoder.transforms as pe_transforms
+import perception_models.core.vision_encoder.pe as pe
+import perception_models.core.vision_encoder.transforms as pe_transforms
 
 DEFAULT_MODEL = "PE-Core-T16-384"
 


fiftyone/utils/pe.py
```diff
@@ -0,0 +1,151 @@
...
+def download_pe_checkpoint(name: str, checkpoint_path: str):
+    path = f"hf://facebook/{name}:{name}.pt"
+
+    # Load from huggingface
+    path = path[len("hf://") :]
+    repo, file = path.split(":")
+
+    local_dir = checkpoint_path.rsplit("/", 1)[0]
+
+    return hf_hub_download(repo_id=repo, filename=file, local_dir=local_dir)
```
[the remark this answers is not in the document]
jacobsela: resolved

fiftyone/utils/pe.py
```diff
@@ -0,0 +1,151 @@
...
+            if self.transforms_fcn == pe_transforms.get_image_transform:
+                # if the user has not specified otherwise, batches should not be ragged
+                self.ragged_batches = False
+
+        self.pool = self.parse_bool(d, "pool", True)
+        self.project = self.parse_bool(d, "project", True)
+        self.norm = self.parse_bool(d, "norm", True)
+        self.layer_idx = self.parse_int(d, "layer_idx", -1)
+        self.strip_cls_token = self.parse_bool(d, "strip_cls_token", False)
+
```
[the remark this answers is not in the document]
jacobsela: outdated

fiftyone/utils/pe.py
```diff
@@ -0,0 +1,149 @@
...
+        self.as_feature_extractor = True
+        self.layer_name = None
+
+        # defaults
+        # TODO: move to manifest
+        if self.entrypoint_fcn is None:
+            self.entrypoint_fcn = pe.VisionTransformer.from_config
+        if self.entrypoint_args is None:
+            self.entrypoint_args = {}
+        default_entrypoint_args = {
```
kaixi-wang, line 88: why are there "hardcoded" default args here and in the manifest https://github.com/voxel51/fiftyone/pull/6595/files#diff-d8412f633c4d9cc26284a55cf1ed0885f26efef3e3eca7997b3d6ad5b8d985adR8422

fiftyone/utils/pe.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def _forward_pass(self, imgs):
+        outputs = self._model.forward_features(
+            imgs,
+            norm=self.config.norm or self.config.project or self.config.pool,
+            layer_idx=self.config.layer_idx,
+            strip_cls_token=self.config.strip_cls_token,
+        )
+
+        if self.config.project or self.config.pool:
```
kaixi-wang, line 129: is this intentional? (if not self.config.pool, but self.config.project, that it still pools?)

fiftyone/utils/pe.py
```diff
@@ -0,0 +1,149 @@
...
+        self.layer_name = None
+
+        # defaults
+        # TODO: move to manifest
+        if self.entrypoint_fcn is None:
+            self.entrypoint_fcn = pe.VisionTransformer.from_config
+        if self.entrypoint_args is None:
+            self.entrypoint_args = {}
+        default_entrypoint_args = {
+            "name": "PE-Core-T16-384",
```
manushreegangwar: Could you define `DEFAULT_PE_MODEL = "PE-Core-T16-384"` and set this using that?

fiftyone/utils/pe.py
```diff
@@ -0,0 +1,149 @@
...
+        }
+        self.entrypoint_args = {
+            **default_entrypoint_args,
+            **self.entrypoint_args,
+        }
+        if "checkpoint_path" not in self.entrypoint_args:
+            self.entrypoint_args["checkpoint_path"] = (
+                fo.config.model_zoo_dir
+                + "/"
+                + self.entrypoint_args["name"]
```
manushreegangwar: If "name" is not in entrypoint args, this is throw an error

fiftyone/utils/pe.py
```diff
@@ -0,0 +1,149 @@
...
+fou.ensure_package("perception_models")
+import perception_models.core.vision_encoder.pe as pe
+import perception_models.core.vision_encoder.transforms as pe_transforms
+
+
+def download_pe_checkpoint(name: str, checkpoint_path: str):
+    path = f"hf://facebook/{name}:{name}.pt"
+
+    # Load from huggingface
+    path = path[len("hf://") :]
```
manushreegangwar: "hf://" in `path` seems unneccesary

manushreegangwar (commented):
Minor comments. Looks good otherwise 🚀 

@jacobsela Have you checked that documentation looks as expected for these models? Might need special handling.

fiftyone/utils/pe.py
jacobsela: yes

jacobsela: need to pool to project. Technically we can project each token individually but not sure if that will give intended results, leaving it for future work.

fiftyone/utils/pe.py
jacobsela: so that in applications users (like you) can just do `foz.load_zoo_model(PE-Core-L14-336-Vision-Encoder)` and get a model that behaves as intended.

kaixi-wang: Oh I wasn't questioning hardcoded defaults in general. I mean why are they in two places.

jacobsela: oh so that in the manifest we can just have minimal diffs from the defaults set here and so that if we want to change behavior across the board in the integration we can just do so here rather than for each entry in the manifest.

jacobsela:
@manushreegangwar docs look fine

manushreegangwar:
> @manushreegangwar docs look fine

@jacobsela for consistency, these should follow dino's [pattern](https://github.com/voxel51/fiftyone/blob/develop/docs/scripts/make_model_zoo_docs.py#L163C10-L163C14)

PR #6610, opened 2025-11-21
[HA] SDK Management + Primitives + JSON GUI Validation

benjaminpkane:

Adds label schema generation and validation

### JSON GUI

https://github.com/user-attachments/assets/bbe88ce4-c903-4eb1-8bf3-7dfd57bf6791

### Label schemas management

```python
import fiftyone as fo
import fiftyone.zoo as foz

# Set the label schemas
dataset.set_label_schemas(dataset.generate_label_schemas())

# Update a single label schema
dataset.update_label_schema(
    "uniqueness",
    {"component": "slider", "type": "float", "range": [0.0, 1.0]},
)

# Activate
dataset.activate_label_schemas("uniqueness")

# Rename (and delete) is handled
dataset.rename_sample_field("uniqueness", "uniqueness_renamed")
fo.pprint(dataset.active_label_schemas)
fo.pprint(dataset.label_schemas)

# Deactivate and clear label schemas
dataset.deactivate_label_schemas()
dataset.delete_label_schemas()
``` 

#### In
```python
import fiftyone as fo
import fiftyone.zoo as foz

fo.pprint(dataset.generate_label_schema())
```

#### Out
```
{
    'created_at': {
        'type': 'datetime',
        'component': 'datepicker',
        'read_only': True,
    },
    'filepath': {'type': 'str', 'component': 'text'},
    'ground_truth': {
        'attributes': {
            'attributes': {'type': 'dict', 'component': 'json'},
            'confidence': {'type': 'float', 'component': 'text'},
            'id': {'type': 'id', 'component': 'text', 'read_only': True},
            'index': {'type': 'int', 'component': 'text'},
            'mask_path': {'type': 'str', 'component': 'text'},
            'tags': {'type': 'list<str>', 'component': 'text'},
        },
        'classes': [
            'airplane',
            'apple',
            'backpack',
            'banana',
            'baseball glove',
            'bear',
            'bed',
            'bench',
            'bird',
            'boat',
            'book',
            'bottle',
            'bowl',
            'broccoli',
            'bus',
            'cake',
            'car',
            'carrot',
            'cat',
            'cell phone',
            'chair',
            'clock',
            'couch',
            'cow',
            'cup',
            'dining table',
            'dog',
            'donut',
            'elephant',
            'fire hydrant',
            'fork',
            'frisbee',
            'giraffe',
            'handbag',
            'horse',
            'hot dog',
            'kite',
            'knife',
            'laptop',
            'microwave',
            'motorcycle',
            'orange',
            'oven',
            'parking meter',
            'person',
            'pizza',
            'refrigerator',
            'sandwich',
            'scissors',
            'sheep',
            'sink',
            'skateboard',
            'skis',
            'snowboard',
            'spoon',
            'sports ball',
            'stop sign',
            'suitcase',
            'surfboard',
            'teddy bear',
            'tennis racket',
            'tie',
            'toothbrush',
            'traffic light',
            'train',
            'truck',
            'tv',
            'umbrella',
            'vase',
            'wine glass',
            'zebra',
        ],
        'component': 'dropdown',
        'type': 'detections',
    },
    'id': {'type': 'id', 'component': 'text', 'read_only': True},
    'last_modified_at': {
        'type': 'datetime',
        'component': 'datepicker',
        'read_only': True,
    },
    'metadata.height': {'type': 'int', 'component': 'text'},
    'metadata.mime_type': {'type': 'str', 'component': 'text'},
    'metadata.num_channels': {'type': 'int', 'component': 'text'},
    'metadata.size_bytes': {'type': 'int', 'component': 'text'},
    'metadata.width': {'type': 'int', 'component': 'text'},
    'predictions': {
        'attributes': {
            'attributes': {'type': 'dict', 'component': 'json'},
            'confidence': {
                'type': 'float',
                'component': 'slider',
                'range': [0.05003104358911514, 0.9999035596847534],
                'default': 0.05003104358911514,
            },
            'id': {'type': 'id', 'component': 'text', 'read_only': True},
            'index': {'type': 'int', 'component': 'text'},
            'mask_path': {'type': 'str', 'component': 'text'},
            'tags': {'type': 'list<str>', 'component': 'text'},
        },
        'classes': [
            'airplane',
            'apple',
            'backpack',
            'banana',
            'baseball glove',
            'bear',
            'bed',
            'bench',
            'bicycle',
            'bird',
            'boat',
            'book',
            'bottle',
            'bowl',
            'broccoli',
            'bus',
            'cake',
            'car',
            'carrot',
            'cat',
            'cell phone',
            'chair',
            'clock',
            'couch',
            'cow',
            'cup',
            'dining table',
            'dog',
            'donut',
            'elephant',
            'fire hydrant',
            'fork',
            'frisbee',
            'giraffe',
            'hair drier',
            'handbag',
            'horse',
            'hot dog',
            'keyboard',
            'kite',
            'knife',
            'laptop',
            'microwave',
            'motorcycle',
            'mouse',
            'orange',
            'oven',
            'parking meter',
            'person',
            'pizza',
            'potted plant',
            'refrigerator',
            'remote',
            'sandwich',
            'scissors',
            'sheep',
            'sink',
            'skateboard',
            'skis',
            'snowboard',
            'spoon',
            'sports ball',
            'stop sign',
            'suitcase',
            'surfboard',
            'teddy bear',
            'tennis racket',
            'tie',
            'toaster',
            'toilet',
            'toothbrush',
            'traffic light',
            'train',
            'truck',
            'tv',
            'umbrella',
            'vase',
            'wine glass',
            'zebra',
        ],
        'component': 'dropdown',
        'type': 'detections',
    },
    'tags': {
        'type': 'list<str>',
        'component': 'checkboxes',
        'values': ['validation'],
    },
    'uniqueness': {
        'type': 'float',
        'component': 'slider',
        'range': [0.15001302256126986, 1.0],
        'default': 0.15001302256126986,
    },
}
```

### Validating a field's label schema

```python
from fiftyone.core.annotation import validate_field_label_schema
import fiftyone.zoo as foz

# correct
validate_field_label_schema(
    dataset,
    "uniqueness",
    {
        "component": "slider",
        "default": 0.0,
        "range": [0.0, 1.0],
        "type": "float",
    },
)

# incorrect
validate_field_label_schema(
    dataset,
    "uniqueness",
    {
        "component": "slider",
        "default": 0.0,
        "range": [0.0, 1.0],
        "type": "int",
    },
)
``` 

New unit test modules

## Release Notes

N/A

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Introduces a label-schema system: generate, validate, edit, activate/deactivate, update, and delete per-field label schemas; UI supports scanning and editing label schemas.

* **Bug Fixes / Refactor**
  * Migrated from legacy annotation APIs to field-centric label-schema APIs and streamlined client/state flows.

* **Chores**
  * Added aggregated-exception handling dependency.

* **Tests**
  * Added comprehensive unit tests covering generation, validation, dataset operations, and UI interactions.

### DIFF
--- a/fiftyone/core/annotation/validate_label_schemas.py
+++ b/fiftyone/core/annotation/validate_label_schemas.py
@@ -244,7 +244,8 @@ def _validate_float_int_field_label_schema(
     field = collection.get_field(field_name)
     is_float = isinstance(field, fof.FloatField)
     _str_type = foac.FLOAT if is_float else foac.INT
-    _type = float if is_float else int
+    # a float field accepts float and int values
+    _type = (float, int) if is_float else int
 
     settings = foac.FLOAT_INT_SETTINGS
     component = label_schema.get(foac.COMPONENT, None)
--- a/tests/unittests/annotation/validate_label_schemas_tests.py
+++ b/tests/unittests/annotation/validate_label_schemas_tests.py
@@ -323,7 +323,7 @@ class LabelSchemaValidationTests(unittest.TestCase):
             {
                 "type": "float",
                 "component": "text",
-                "default": 1.0,
+                "default": 1.1,
                 "precision": 1,
             },
             fields="float_field",
@@ -335,8 +335,8 @@ class LabelSchemaValidationTests(unittest.TestCase):
             {
                 "type": "float",
                 "component": "slider",
-                "range": [0.0, 1.0],
-                "default": 0.0,
+                "range": [0.5, 1],
+                "default": 0.5,
             },
             fields="float_field",
             _allow_default=True,
@@ -351,15 +351,6 @@ class LabelSchemaValidationTests(unittest.TestCase):
                     "component": "slider",
                 },
                 fields="float_field",
-            )
-
-        # 'default' is not a float
-        with self.assertRaises(ExceptionGroup):
-            validate_label_schemas(
-                dataset,
-                {"type": "float", "component": "text", "default": 1},
-                fields="float_field",
-                _allow_default=True,
             )
 
         # 'precision' is negative
@@ -390,8 +381,8 @@ class LabelSchemaValidationTests(unittest.TestCase):
                 {
                     "type": "float",
                     "component": "slider",
-                    "range": [0.0, 1.0],
-                    "default": 2.0,
+                    "range": [0, 1],
+                    "default": 2,
                 },
                 fields="float_field",
                 _allow_default=True,
@@ -404,7 +395,7 @@ class LabelSchemaValidationTests(unittest.TestCase):
                 {
                     "type": "float",
                     "component": "slider",
-                    "range": [1.0, 0.0],
+                    "range": [1, 0],
                 },
                 fields="float_field",
             )


fiftyone/core/collections.py
```diff
@@ -3523,33 +3523,229 @@ def compute_metadata(
...
+
+        The ``type`` value for a field is inferred from the collection's field
+        schema. See
+        :meth:`fiftyone.core.collections.SampleCollection.get_field_schema`
+
+        Currently supported  media types for the collection are ``image`` and
+        ``3d``. See
+        :attr:`fiftyone.core.collections.SampleCollection.media_type`
+
+        Primitives and components::
```
brimoor: Can you build the docs locally and confirm that these docstrings render as expected?

Some things to check:
- I believe `::` is meant to create monospace code blocks, and in that case the text that follows must be indented in order for that to render as expected. But in this case I think you want something more like section headings, for which you may just want to bold via `**`
- bulleted lists may need an empty line above the first `-` in order to render as expected

fiftyone/core/collections.py
```diff
@@ -3523,33 +3523,229 @@ def compute_metadata(
...
-            - tags
+            import fiftyone as fo
+            import fiftyone.zoo as foz
 
-        Args:
-            collection: a :class:`fiftyone.core.collections.SampleCollection`
-            field_name: a field name or ``embedded.field.name`` to process
+            dataset = foz.load_zoo_dataset("quickstart")
+
+            fo.pprint(fo.generate_field_schema(dataset, scan_samples=True))
```
brimoor: In the version of this docstring on `SampleCollection`, I would use `dataset.generate_field_schema()` in the example code 🤓

fiftyone/core/collections.py
```diff
@@ -3523,33 +3523,229 @@ def compute_metadata(
...
+
+            #{
```
brimoor: nit: just to avoid all the `#`, perhaps you can format this as follows so that the example dict is copy+pastable? 

```
Example::
    
    <code here>

Example schema::

    <JSON here>
```

fiftyone/core/collections.py
```diff
@@ -3523,33 +3523,229 @@ def compute_metadata(
...
+            #        'component': 'slider',
+            #        'range': [0.15001302256126986, 1.0],
+            #        'default': 0.15001302256126986,
+            #    },
+            #}
+
+        Args:
+            fields (None): a field name, ``embedded.field.name`` or iterable of
+                such values
+            scan_samples (False): whether to scan the collection to populate
```
brimoor: suggestion: explain what happens when `scan_samples=False`:

```
            scan_samples (False): whether to scan the collection to populate
                component settings based on actual field values (ranges,
                values, defaults, etc). By default, the label schema is
                generated from *only* the statically available information in
                the dataset's field schema
```

fiftyone/core/annotation/generate_label_schema.py
```diff
@@ -0,0 +1,569 @@
...
+    Returns:
+        a list of supported fields
+    """
+    _ensure_collection_is_supported(sample_collection)
+    fields = _get_all_supported_fields(sample_collection)
+    return _flatten_fields(sample_collection, fields)
+
+
+def generate_label_schema(sample_collection, fields=None, scan_samples=True):
+    """Generates a label schema given a
```
brimoor: Same comments from `SampleCollection.generate_label_schema()` apply to this function's docstring

fiftyone/core/annotation/validate_label_schema.py
```diff
@@ -0,0 +1,590 @@
...
+from datetime import date, datetime
+from exceptiongroup import ExceptionGroup
+import json
+
+import fiftyone.core.annotation.constants as foac
+import fiftyone.core.fields as fof
+import fiftyone.core.labels as fol
+
+
+def validate_field_label_schema(sample_collection, field_name, label_schema):
```
brimoor: Can we make the public method signature like this?

```py
# fields=None means all fields are validated, but particular field(s) may be requested if desired
def validate_label_schema(sample_collection, label_schema, fields=None):
```

This would be for consistency with the other methods in this interface.

fiftyone/core/annotation/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+from fiftyone.core.runs import (
+    BaseRun,
+    BaseRunConfig,
+    BaseRunInfo,
+    BaseRunResults,
+)
+from fiftyone.core.odm import patch_annotation_runs
+
+from .generate_label_schema import generate_label_schema
```
brimoor: I think we should import `generate_label_schema()` and `validate_label_schema()` in the top-level `fiftyone` namespace.

In general, my goal is that end users only ever have to `import fiftyone`.

fiftyone/core/annotation/constants.py
```diff
@@ -0,0 +1,193 @@
...
+import fiftyone.core.labels as fol
+import fiftyone.core.media as fom
+import fiftyone.core.metadata as fomm
+from fiftyone.core.odm import DynamicEmbeddedDocument
+
+
+### Components
+
+
+CHECKBOX = "checkbox"
```
kaixi-wang, line 19: Why not use classes and enums?

benjaminpkane: I didn't see the benefit. Trying to keep things simple

fiftyone/core/annotation/validate_label_schemas.py
```diff
@@ -0,0 +1,654 @@
...
+
+        if key == foac.COMPONENT and value not in foac.BOOL_COMPONENTS:
+            _raise_component_error(field_name, value)
+        elif key == foac.DEFAULT:
+            _validate_default(field_name, value, bool, allow_default)
+        elif key == foac.READ_ONLY:
+            _validate_read_only(field_name, value)
+        elif key == foac.TYPE and value != foac.BOOL:
+            _raise_type_error(field, field_name, value)
+
```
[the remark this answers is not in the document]
benjaminpkane: incorrect

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -0,0 +1,183 @@
...
+    scan: () => {
+      setIsScanning(true);
+      generate.execute(
+        { field },
+        {
+          callback: (result) => {
+            if (result.result) {
+              setCurrent(result.result.label_schema);
+              setIsScanning(false);
+            }
```
lanzhenw, line 108: Suggest: 
```
callback: (result) => {
    if (result.result) {
      setCurrent(result.result.label_schema);
    }
    setIsScanning(false);  // we want to always reset `isScanning` to false regardless of `result.result`
```

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -0,0 +1,183 @@
...
+            },
+          }
+        );
+      } catch (e) {
+        if (e instanceof SyntaxError) {
+          setErrors([e.message]);
+        }
+
+        setIsValid(false);
+        return;
```
lanzhenw, line 158: Suggest:
```
setIsValid(false);
setIsValidating(false); 
return
```

fiftyone/core/annotation/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+from fiftyone.core.runs import (
+    BaseRun,
+    BaseRunConfig,
+    BaseRunInfo,
+    BaseRunResults,
+)
+from fiftyone.core.odm import patch_annotation_runs
+
+from .generate_label_schemas import generate_label_schemas
+from .validate_label_schemas import validate_label_schemas
```
exupero, line 19: These imports aren't used. Are they imported here so users can import them from the `fiftyone.core.annotation` module? If so, we should add a comment so no one deletes them.

fiftyone/core/dataset.py
```diff
@@ -2769,11 +2996,42 @@ def _delete_sample_fields(self, field_names, error_level):
...
+
+            keys = path.split(".")
+            parent = ".".join(keys[:-1])
+
+            if self._is_label_field(parent):
+                if parent not in label_schemas:
+                    # this is a label list field
+                    # e.g. remove "detections" from
+                    # "ground_truth.detections"
+                    parent = ".".join(parent.split(".")[:-1])
```
exupero: Same code with same explanatory comment is [here](https://github.com/voxel51/fiftyone/blob/9cc5c7a27ecd5dcaddc3d1241c1e954bf426218b/fiftyone/core/dataset.py#L2665-L2667), probably worth putting in a self-describing function.

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -0,0 +1,431 @@
...
+
+            elif len(values) <= foac.VALUES_THRESHOLD:
+                settings[foac.COMPONENT] = foac.DROPDOWN
+
+            if settings[foac.COMPONENT] in foac.VALUES_COMPONENTS:
+                settings[foac.VALUES] = values
+
+    except Exception:
+        # too many distinct values
+        pass
```
exupero: This shouldn't silently swallow all exceptions. Is there a more specific error or message we can check to make sure we're only catching the "too many distinct values" errors?

fiftyone/core/dataset.py
```diff
@@ -2450,13 +2641,49 @@ def _rename_sample_fields(self, field_mapping, view=None):
             sample_collection, paths, new_paths
         )
 
+        label_schemas = self.label_schemas
```
brimoor: @benjaminpkane in [`DatasetMixin._rename_fields()`](https://github.com/voxel51/fiftyone/blob/83261111a00e430e58413692175aca8d4cb34c31/fiftyone/core/odm/mixins.py#L623) there is precedent for handling side effects of field renaming such as updating the dataset's app config. So can you move this logic into that method as well, for consistency?

fiftyone/core/dataset.py
```diff
@@ -2450,13 +2641,49 @@ def _rename_sample_fields(self, field_mapping, view=None):
...
+                        attributes[new_name] = attributes.pop(name)
+
         fields, _, _, _ = _parse_field_mapping(field_mapping)
 
         if fields:
             fos.Sample._purge_fields(self._sample_collection_name, fields)
 
         fos.Sample._reload_docs(self._sample_collection_name)
         self._reload()
+        self.set_label_schemas(new_label_schemas)
```
brimoor: Move this logic into [`DatasetMixin._rename_fields()`](https://github.com/voxel51/fiftyone/blob/83261111a00e430e58413692175aca8d4cb34c31/fiftyone/core/odm/mixins.py#L623) as well

fiftyone/core/dataset.py
```diff
@@ -2769,11 +2996,42 @@ def _delete_sample_fields(self, field_names, error_level):
 
         fields, _ = _parse_fields(field_names)
 
+        label_schemas = self.label_schemas
```
brimoor: Same comment: move this logic into [`DatasetMixin._delete_fields()`](https://github.com/voxel51/fiftyone/blob/83261111a00e430e58413692175aca8d4cb34c31/fiftyone/core/odm/mixins.py#L748) for consistency

fiftyone/core/annotation/utils.py
```diff
@@ -0,0 +1,206 @@
...
+        for subfield in field.fields:
+            if _is_supported_primitive(subfield) or _is_supported_label(
+                subfield, collection.media_type, require_app_support
+            ):
+                flattened_fields.append(f"{field_name}.{subfield.name}")
+
+    return sorted(set(flattened_fields))
+
+
+def get_type(field):
```
kaixi-wang, line 151: are these types unique to annotations or should this function be in a more generic place?

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,194 @@ def deleted(self):
         """Whether the dataset is deleted."""
         return self._deleted
 
+    @property
+    def active_label_schemas(self):
```
kaixi-wang, line 1112: does active label_schema really need to be defined on the backend/persisted in the db? What if multiple annotators are using different label schemas for the same dataset?

fiftyone/core/annotation/__init__.py
```diff
@@ -0,0 +1,84 @@
+"""
+Annotation framework
```
kaixi-wang, line 2: should all of this be in core? Would it make sense to have most of this in fiftyone/annotation?

fiftyone/core/annotation/__init__.py
benjaminpkane: yes, thanks

fiftyone/core/dataset.py
benjaminpkane: There is only one label schema per field. In the future we can add support for many label fields, but we don't need it right now. 

It does need to be persisted so the manager can configure what other users see.

fiftyone/core/annotation/__init__.py
benjaminpkane: It's a good point, but this is where I ended up 😅

benjaminpkane:
@exupero thanks for the review, can you take a look again? 🙏

fiftyone/core/annotation/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+
+from fiftyone.core.runs import (
+    BaseRun,
+    BaseRunConfig,
+    BaseRunInfo,
+    BaseRunResults,
+)
+from fiftyone.core.odm import patch_annotation_runs
+
+# core methods are accesssible to fiftyone.core.annotation
```
exupero: [nit] typo
```suggestion
# core methods are accessible to fiftyone.core.annotation
```

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -0,0 +1,433 @@
...
+
+def _handle_float_or_int(
+    collection, field_name, is_list, settings, scan_samples
+):
+    settings[foac.COMPONENT] = foac.TEXT
+    if is_list or not scan_samples:
+        return settings
+
+    mn, mx = collection.bounds(field_name, safe=True)
+    if mn != mx and mn is not None:
```
exupero: Do we also need to check when `mx is not None`?

fiftyone/core/annotation/validate_label_schemas.py
```diff
@@ -0,0 +1,658 @@
...
+    elif exceptions:
+        raise exceptions[0]
+
+
+def _validate_default(
+    field_name, value, _type, allow_default, _range=None, values=None
+):
+    if not allow_default:
+        raise ValueError(
+            f"'default' setting is not allowed for field '{field_name}"
```
exupero: [nit] add close quote:
```suggestion
            f"'default' setting is not allowed for field '{field_name}'"
```

fiftyone/core/annotation/validate_label_schemas.py
```diff
@@ -0,0 +1,658 @@
...
+
+    raise exception
+
+
+def _validate_default_list(
+    field_name, value, _type, allow_default, values=None
+):
+    if not allow_default:
+        raise ValueError(
+            f"'default' setting is not allowed for field '{field_name}"
```
exupero: [nit] add close quote:
```suggestion
            f"'default' setting is not allowed for field '{field_name}'"
```

fiftyone/core/annotation/validate_label_schemas.py
```diff
@@ -0,0 +1,658 @@
...
+def _validate_values_setting(field_name, value, _type, key=foac.VALUES):
+    if not isinstance(value, list):
+        raise ValueError(
+            f"'{key}' setting for field '{field_name}' must be a list"
+        )
+
+    if not value:
+        raise ValueError(
+            f"'{key}' setting for field '{field_name}' must have at least "
+            f"one value "
```
exupero: [nit] remove extra space:
```suggestion
            f"one value"
```

fiftyone/core/collections.py
```diff
@@ -3518,33 +3518,248 @@ def compute_metadata(
...
+                    'range': [0.15001302256126986, 1.0],
+                },
+            }
 
         Args:
-            collection: a :class:`fiftyone.core.collections.SampleCollection`
-            field_name: a field name or ``embedded.field.name`` to process
+            sample_collection: the
+                :class:`fiftyone.core.collections.SampleCollection` to generate
+                the schema with
```
exupero: `sample_collection` isn't a parameter of this function:
```suggestion
```

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,201 @@ def deleted(self):
...
+
+        Args:
+            label_schemas: a label schemas ``dict``
+        """
+        if label_schemas is None:
+            label_schemas = {}
+
+        foa.validate_label_schemas(self, label_schemas)
+        self._doc.label_schemas = label_schemas
+        active = list(self.active_label_schemas or [])
```
exupero: `self.active_label_schemas` above always returns a list.
```suggestion
        active = self.active_label_schemas
```

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,201 @@ def deleted(self):
...
+
+        for field in self._doc.active_label_schemas:
```
exupero: Can you use `self.active_label_schemas` here too? Otherwise, do you need to deal with `self._doc.active_label_schemas` being `None` (`self.active_label_schemas` implies it might be something falsey)?

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,201 @@ def deleted(self):
...
+
+        Raises:
+            ValueError: if any fields are not in the label schema
+        """
+        if fields is None:
+            fields = sorted(self._doc.label_schemas)
+        elif etau.is_str(fields):
+            fields = [fields]
+
+        result = list(self._doc.active_label_schemas)
```
exupero: Can `self._doc.active_label_schemas` be `None` here?
```suggestion
        result = list(self._doc.active_label_schemas or [])
```

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,201 @@ def deleted(self):
...
+        Raises:
+            ValueError: if any fields are not in the label schema, or any field
+                is not currently active
+        """
+        if fields is None:
+            fields = self._doc.active_label_schemas
+        elif etau.is_str(fields):
+            fields = [fields]
+
+        result = list(self._doc.active_label_schemas)
```
exupero:
```suggestion
        result = list(self._doc.active_label_schemas or [])
```

fiftyone/core/odm/mixins.py
```diff
@@ -1997,3 +2039,34 @@ def _get_index_updates(dataset, paths, new_paths=None):
...
+    name = keys[-1]
+
+    if name not in attributes:
+        return
+
+    if new_path is None:
+        del attributes[name]
+    else:
+        new_name = new_path.split(".")[-1]
+        attributes[new_name] = attributes.pop(name)
```
exupero, line 2072: [learning question] Does this mutate instance state on something that's used by the caller?

plugins/operators/annotation.py
```diff
@@ -6,185 +6,162 @@
...
+                ctx.dataset, ctx.params.get("label_schemas", {})
+            )
+        except ExceptionGroup as exceptions:
+            for exception in exceptions.exceptions:
+                if isinstance(exception, ExceptionGroup):
+                    for subexception in exception.exceptions:
+                        errors.append(str(subexception))
+                    continue
+
+                errors.append(str(exception))
```
exupero: What about other types of errors? Should they be captured and returned, or allowed to raise normally?

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,201 @@ def deleted(self):
...
+        See
+        :meth:`generate_label_schemas` for ways in which to generate label
+        schemas.
+
+        Returns:
+            the dataset's label schemas ``dict``
+        """
+        return copy.deepcopy(self._doc.label_schemas) or {}
+
+    def set_label_schemas(self, label_schemas):
```
kaixi-wang, line 1145: note for teams, this is likely the method we would want to decorate with permissions to require dataset.manage

fiftyone/core/odm/dataset.py
```diff
@@ -70,7 +70,6 @@ class SampleFieldDocument(EmbeddedDocument):
     description = StringField(null=True)
     info = DictField(null=True)
     read_only = BooleanField(default=False)
-    schema = DictField(null=True)
```
kaixi-wang, line 73: is it safe to remove this? (any risk of backwards compatibility issues)

fiftyone/core/odm/mixins.py
```diff
@@ -595,6 +600,23 @@ def _rename_fields(cls, sample_collection, paths, new_paths):
...
+                if path in active_label_schemas:
+                    active_label_schemas[
+                        active_label_schemas.index(path)
+                    ] = new_path
+                continue
+
+            for field in label_schemas:
+                if field.startswith(f"{path}."):
+                    new_field = f"{new_path}.{field.split('.')[1]}"
+                    new_label_schemas[new_field] = new_label_schemas.pop(field)
```
kaixi-wang, line 614: label_schemas and new_label_schemas are pointing to the same dict, so mutating while iterating isn't safe?

fiftyone/core/annotation/utils.py
benjaminpkane: unique to annotation

fiftyone/core/annotation/generate_label_schemas.py
benjaminpkane: That check is unnecessary. Removing, thanks. 

`mx` must be `None` iff `mn` is `None`

fiftyone/core/dataset.py
benjaminpkane: I can use `self.active_label_schemas`

fiftyone/core/odm/dataset.py
benjaminpkane: I'm working on a migration

fiftyone/core/dataset.py
benjaminpkane: I don't believe so, but will change to `self.activate_label_schemas` for a guarantee

fiftyone/core/odm/mixins.py
benjaminpkane: Yes, it mutates `new_label_schemas`. Either deleting a label's field or renaming it

fiftyone/core/odm/mixins.py
benjaminpkane: label_schemas and new_label_schemas are not the same dict, but they are copies that start out identical

plugins/operators/annotation.py
benjaminpkane: If it is not an `ExceptionGroup` that comes from validation it should raise normally. I will tweak it a little. Any other exceptions outside the scope of "validation" will raise normally

fiftyone/core/odm/mixins.py
exupero: Confirmed that [`dataset.label_schemas` is a property](https://github.com/voxel51/fiftyone/blob/ba0da5a97266c7f0da3addc1161dc6cd4654b72c/fiftyone/core/dataset.py#L1136-L1148) that copies the underlying data, so `label_schemas` and `new_label_schemas` should not reference the same object.

fiftyone/core/annotation/__init__.py
exupero: @kaixi-wang What's the difference between fiftyone/core/annotation and fiftyone/annotation?

fiftyone/core/dataset.py
exupero: What about `update_label_schema` and `delete_label_schemas`?

fiftyone/core/annotation/constants.py
```diff
@@ -0,0 +1,192 @@
...
+    fof.DateTimeField,
+    fof.DictField,
+    fof.FloatField,
+    fof.FrameNumberField,
+    fof.IntField,
+    fof.ObjectIdField,
+    fof.StringField,
+    fof.UUIDField,
+)
+UNSUPPORTED_LABEL_TYPES = {
```
exupero, line 190: Should we only allow certain label types, rather than disallowing certain types? The six constants above this specify what's supported; this is the only one that specifies what's not supported. Can users define their own label types?

fiftyone/core/odm/mixins.py
```diff
@@ -595,6 +600,23 @@ def _rename_fields(cls, sample_collection, paths, new_paths):
...
+
+                _handle_nested_label_schema_change(
+                    dataset, path, new_path, label_schemas, new_label_schemas
+                )
```
exupero, line 618: How is this tested? I didn't see any tests of a `_rename_fields` function? Is it tested end-to-end somewhere?

fiftyone/core/odm/mixins.py
```diff
@@ -1997,3 +2039,34 @@ def _get_index_updates(dataset, paths, new_paths=None):
                 updates[name] = None
 
     return updates
+
+
+def _handle_nested_label_schema_change(
```
exupero: [nit] Slightly more direct name
```suggestion
def _change_nested_label_schema(
```

fiftyone/core/annotation/utils.py
```diff
@@ -0,0 +1,208 @@
...
+        raise ValueError(
+            f"{sample_collection.media_type} media is not supported yet"
+        )
+
+
+def get_supported_app_annotation_fields(sample_collection):
+    """Gets the supported App annotation fields for a
+    :class:`fiftyone.core.collections.SampleCollection`.
+
+    Currently supported  media types for the collection are ``image`` and
```
exupero: [nit] Extra space can be removed:
```suggestion
    Currently supported media types for the collection are ``image`` and
```

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,201 @@ def deleted(self):
...
+    def active_label_schemas(self):
+        """The list of active fields in the dataset's
+        :ref:`label schemas <annotation-label-schema>`.
+        """
+        return list(self._doc.active_label_schemas or [])
+
+    @active_label_schemas.setter
+    def active_label_schemas(self, fields):
+        if etau.is_str(fields):
+            fields = [fields]
```
exupero: [idea] This logic shows up a few times. Maybe have a `as_str_list(fields)` utility function?

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,201 @@ def deleted(self):
...
+        elif fields is None:
+            fields = []
+        else:
+            fields = list(fields)
+
+        for field in fields:
+            if field not in self._doc.label_schemas:
```
exupero, line 1123: Can `self._doc.label_schemas` be `None` here, causing an error? (`active_label_schemas` above only suggests it can be falsey, so not sure if it's guaranteed to be present.) Same question for a couple other inclusion checks on `self._doc.label_schemas` below.

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,201 @@ def deleted(self):
...
+                raise ValueError(
+                    f"field '{field}' does not have a label schema"
+                )
+
+        self._doc.active_label_schemas = fields
```
exupero, line 1128: Do I understand correctly that `self._doc.active_label_schemas` is a list of strings naming keys in `self._doc.label_schemas`? (If so, the nomenclature is a bit confusing, since one is schema definitions and the other is names of schema definitions, but I wouldn't say it's a deal-breaker.)

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,201 @@ def deleted(self):
...
+        self._doc.save()
```
exupero: Should these `self._doc.save()` calls use `self.save()`?

fiftyone/core/dataset.py
```diff
@@ -1105,6 +1108,201 @@ def deleted(self):
...
+
+        foa.validate_label_schemas(self, label_schemas)
+        self._doc.label_schemas = label_schemas
+        active = self.active_label_schemas
+
+        for field in self.active_label_schemas:
+            if field not in label_schemas:
+                active.remove(field)
+
+        self._doc.active_label_schemas = active
```
exupero: Does this accomplish something different than a list comprehension?
```suggestion
        self._doc.active_label_schemas = [field for field in self.active_label_schemas
                                          if field in label_schemas]
```
If not, a list comprehension has the advantages of 1) being more concise and transparent, 2) not mutating a collection, 3) not potentially leaving duplicate fields in `active` (since `.remove()` only removes the first occurrence), and 4) avoiding negative logic (i.e. no "not in" or "remove").

fiftyone/core/collections.py
```diff
@@ -3518,33 +3518,245 @@ def compute_metadata(
...
+
+        A label schema is defined by a ``type`` and ``component`` with respect
+        to a field. Further settings depend on the ``type`` and ``component``
+        combination as outlined below.
+
+        The ``type`` value for a field is inferred from the collection's field
+        schema. See
+        :meth:`fiftyone.core.collections.SampleCollection.get_field_schema`
+
+        Currently supported  media types for the collection are ``image`` and
```
exupero: [nit] Extra space can be removed:
```suggestion
        Currently supported media types for the collection are ``image`` and
```

fiftyone/core/collections.py
```diff
@@ -3518,33 +3518,245 @@ def compute_metadata(
...
+        **Primitives and components**
+
+        Supported primitive types are:
+
+            -   ``bool``: :class:`fiftyone.core.fields.BooleanField`
+            -   ``date``: :class:`fiftyone.core.fields.DateField`
+            -   ``datetime``: :class:`fiftyone.core.fields.DateTimeField`
+            -   ``dict``: :class:`fiftyone.core.fields.DictField`
+            -   ``float``: :class:`fiftyone.core.fields.FloatField`
+            -   ``id``: : :class:`fiftyone.core.fields.ObjectIdField` or
```
exupero: Is this supposed to have the extra colon before `:class:`?
```suggestion
            -   ``id``: :class:`fiftyone.core.fields.ObjectIdField` or
```

fiftyone/core/collections.py
```diff
@@ -3518,33 +3518,245 @@ def compute_metadata(
...
+                :class:`fiftyone.core.fields.UUIDField`
+            -   ``int``: :class:`fiftyone.core.fields.IntField` or
+                :class:`fiftyone.core.fields.FrameNumberField`
+            -   ``list<bool>``: :class:`fiftyone.core.fields.ListField` of
+                :class:`fiftyone.core.fields.BooleanField`
+            -   ``list<int>``: :class:`fiftyone.core.fields.ListField` of
+                :class:`fiftyone.core.fields.IntField`
+            -   ``list<float>``: :class:`fiftyone.core.fields.ListField` of
+                :class:`fiftyone.core.fields.FloatField`
+            -   ``list<str>``: : :class:`fiftyone.core.fields.ListField` of
```
exupero:
```suggestion
            -   ``list<str>``: :class:`fiftyone.core.fields.ListField` of
```

fiftyone/core/collections.py
```diff
@@ -3518,33 +3518,245 @@ def compute_metadata(
...
+        Supported ``str`` type components are:
+
+            -   ``dropdown``: the default if ``>5`` and ``<=1000`` values are
+                scanned
+            -   ``radio``: the default if ``<=5`` values are scanned
+            -   ``text``: the default if ``0`` values or ``>1000`` values are
+                scanned, or ``scan_samples`` is ``False``
+
+        ``float`` types support a ``precision`` setting when a ``text``
+        component is configured  for the number of digits to allow after the
```
exupero: [nit] Extra space can be removed:
```suggestion
        component is configured for the number of digits to allow after the
```

fiftyone/core/collections.py
```diff
@@ -3518,33 +3518,245 @@ def compute_metadata(
...
+
+        The ``label`` subfield of all label types are configured via ``classes``
+        and support the same settings as a ``str`` type. See the example output
+        below for ``detections`` fields in the quickstart dataset. If the label
+        type has a visual representation, that field is handled by the App's
+        builtin annotation UI, e.g. ``bounding_box`` for a ``detection``.
+        Primitive attributes of label types are configured via the
+        ``attributes`` setting.
+
+        When a label is marked is ``read_only``, all its attributes inherit the
```
exupero: [nit] Small grammar fix, `is` -> `as`:
```suggestion
        When a label is marked as ``read_only``, all its attributes inherit the
```

fiftyone/core/collections.py
```diff
@@ -3518,33 +3518,245 @@ def compute_metadata(
...
         Args:
-            collection: a :class:`fiftyone.core.collections.SampleCollection`
-            field_name: a field name or ``embedded.field.name`` to process
+            fields (None): a field name, ``embedded.field.name`` or iterable of
+                such values
+            scan_samples (True): whether to scan the collection to populate
+                component settings based on actual field values (ranges,
+                values, etc). By default, the label schema is generated from
+                *only* the statically available information in the dataset's
+                field schema
```
exupero, line 3749: Is there a contradiction here regarding the default behavior? `scan_samples=True` is the default and scans for more data, but then the doc says "By default, the label schema is generated from *only* the statically available information".

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -0,0 +1,433 @@
...
+
+    A label schema is defined by a ``type`` and ``component`` with respect
+    to a field. Further settings depend on the ``type`` and ``component``
+    combination as outlined below.
+
+    The ``type`` value for a field is inferred from the collection's field
+    schema. See
+    :meth:`fiftyone.core.collections.SampleCollection.get_field_schema`
+
+    Currently supported  media types for the collection are ``image`` and
```
exupero: [nit] Extra space can be removed:
```suggestion
    Currently supported media types for the collection are ``image`` and
```

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -0,0 +1,433 @@
...
+    **Primitives and components**
+
+    Supported primitive types are:
+
+        -   ``bool``: :class:`fiftyone.core.fields.BooleanField`
+        -   ``date``: :class:`fiftyone.core.fields.DateField`
+        -   ``datetime``: :class:`fiftyone.core.fields.DateTimeField`
+        -   ``dict``: :class:`fiftyone.core.fields.DictField`
+        -   ``float``: :class:`fiftyone.core.fields.FloatField`
+        -   ``id``: : :class:`fiftyone.core.fields.ObjectIdField` or
```
exupero:
```suggestion
        -   ``id``: :class:`fiftyone.core.fields.ObjectIdField` or
```

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -0,0 +1,433 @@
...
+            :class:`fiftyone.core.fields.UUIDField`
+        -   ``int``: :class:`fiftyone.core.fields.IntField` or
+            :class:`fiftyone.core.fields.FrameNumberField`
+        -   ``list<bool>``: :class:`fiftyone.core.fields.ListField` of
+            :class:`fiftyone.core.fields.BooleanField`
+        -   ``list<int>``: :class:`fiftyone.core.fields.ListField` of
+            :class:`fiftyone.core.fields.IntField`
+        -   ``list<float>``: :class:`fiftyone.core.fields.ListField` of
+            :class:`fiftyone.core.fields.FloatField`
+        -   ``list<str>``: : :class:`fiftyone.core.fields.ListField` of
```
exupero:
```suggestion
        -   ``list<str>``: :class:`fiftyone.core.fields.ListField` of
```

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -0,0 +1,433 @@
...
+    Supported ``str`` type components are:
+
+        -   ``dropdown``: the default if ``>5`` and ``<=1000`` values are
+            scanned
+        -   ``radio``: the default if ``<=5`` values are scanned
+        -   ``text``: the default if ``0`` values or ``>1000`` values are
+            scanned, or ``scan_samples`` is ``False``
+
+    ``float`` types support a ``precision`` setting when a ``text`` component
+    is configured  for the number of digits to allow after the decimal.
```
exupero: [nit] Extra space can be removed:
```suggestion
    is configured for the number of digits to allow after the decimal.
```

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -0,0 +1,433 @@
...
+
+        validate_label_schemas(
+            sample_collection,
+            label_schema,
+            fields=field_name,
+        )
+        schema[field_name] = label_schema
+
+    if is_scalar:
+        return next(iter(schema.values()))
```
exupero: Is it possible to put this in the loop above and return early for scalars? Then you don't have to build up a bunch of values we're eventually going to throw out (and we can avoid the awkward `next(iter(...))`).
```suggestion
        if is_scalar:
            return label_schema

        schema[field_name] = label_schema
```

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -0,0 +1,433 @@
...
+
+    return schema
+
+
+def _generate_field_label_schema(collection, field_name, scan_samples):
```
exupero, line 292: This function is somewhat complex. Are there any tests of it?

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -0,0 +1,433 @@
...
+    result = dict(
+        attributes={k: attributes[k] for k in sorted(attributes)},
+        **label,
+        type=_type,
+    )
+
+    if classes:
+        result[foac.CLASSES] = classes
+
+    return {k: result[k] for k in sorted(result)}
```
exupero, line 384: [learning question] What's the purpose of sorting the dict by its keys? It looks like the keys will be `"attributes"`, `"type"`, `foat.CLASSES`, and whatever keys are in `label`.

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -0,0 +1,433 @@
...
+
+            elif len(values) <= foac.VALUES_THRESHOLD:
+                settings[foac.COMPONENT] = foac.DROPDOWN
+
+            if settings[foac.COMPONENT] in foac.VALUES_COMPONENTS:
+                settings[foac.VALUES] = values
+
+    except OperationFailure:
+        # too many distinct values
+        pass
```
exupero, line 419: Does this whole block need to be in the `try`? Based on the comment, I would think the only thing that would throw the error would be `collection.distinct(field_name)` and the rest should be safe. Or is there some laziness and the collection has to be realized by e.g. `len(values)` before an error is raised?
```suggestion
    try:
        if scan_samples and field_name != foac.FILEPATH:
            values = collection.distinct(field_name)
        else:
            values = None
    except OperationFailure:
        # too many distinct values
        pass

    if values:
        if len(values) <= foac.CHECKBOXES_OR_RADIO_THRESHOLD:
            settings[foac.COMPONENT] = (
                foac.CHECKBOXES if is_list else foac.RADIO
            )

        elif len(values) <= foac.VALUES_THRESHOLD:
            settings[foac.COMPONENT] = foac.DROPDOWN

        if settings[foac.COMPONENT] in foac.VALUES_COMPONENTS:
            settings[foac.VALUES] = values
```

fiftyone/core/annotation/validate_label_schemas.py
```diff
@@ -0,0 +1,662 @@
...
+
+
+def _validate_field_label_schema(
+    sample_collection, field_name, label_schema, allow_default=False
+):
+    field = sample_collection.get_field(field_name)
+    if field is None:
+        raise ValueError(f"field '{field_name}' does not exist")
+
+    field_is_read_only = field.read_only
```
exupero, line 84: Any reason this needs to be a separate variable instead of just using `field.read_only` below? I don't see anything that would change the value between here and there.

fiftyone/core/annotation/validate_label_schemas.py
```diff
@@ -0,0 +1,662 @@
...
+            )
+        except Exception as exc:
+            exceptions.append(exc)
+
+    if len(exceptions) > 1:
+        raise ValidationErrors(
+            f"invalid attributes for field '{field_name}'", exceptions
+        )
+    elif exceptions:
+        raise exceptions[0]
```
exupero: Should this not still raise the exceptions as `ValidationErrors`? (I'm thinking of [this code](https://github.com/voxel51/fiftyone/blob/ba0da5a97266c7f0da3addc1161dc6cd4654b72c/plugins/operators/annotation.py#L153-L164) that expects a `ValidationError` from a different validation function. Seems like a good convention?)

exupero (commented):
I like the defensive programming: lots of validation and specific errors raised when things go wrong. 💯

I notice quite a few `isinstance` checks on `Field` subclasses. Should any logic move to methods on those classes, allowing us to use polymorphism instead of `if` branches? Or was it a deliberate choice not to add new methods to `Field` subclasses (e.g. because those are closed interfaces)?

fiftyone/core/annotation/constants.py
benjaminpkane: I'm adding a comment. Geo and temporal label types have subfields that cannot be represented by the HA annotation UX, e.g. a `support` tuple on the `TemporalDetection` label cannot be given a `type` and `component` in HA UX.

There are label types that are not supported by HA yet too, e.g. a `Segmentation`, but that is a different category. The SDK can still validate a schema for it

No, users are not (yet) able to add their own label types.

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/state.ts
```diff
@@ -158,7 +157,7 @@ export const isNew = atom((get) => {
 const fieldsOfType = atomFamily((type: LabelType) =>
   atom((get) => {
     const fields = new Array<string>();
-    for (const field in get(activeSchemas)) {
+    for (const field of get(activeLabelSchemas) ?? []) {
       if (IS[type].has(get(fieldType(field)))) {
```
erik-nieh, line 161: @benjaminpkane FYI the bug discussed earlier - when clicking "create new classification" or "create new detection" this line runs into an issue `IS[type]` returns undefined.

Seems that the "type" is `Classification` but the const `CLASSIFICATION` is `classifcation` (lowercase)

fiftyone/core/odm/mixins.py
benjaminpkane: Renaming is tested [here](https://github.com/voxel51/fiftyone/blob/7e1047299d70088ea878e88d7c4920108544d0d5/tests/unittests/annotation/dataset_label_schemas_tests.py#L151)

fiftyone/core/dataset.py
benjaminpkane: it accomplishes the same thing, sounds good

fiftyone/core/dataset.py
benjaminpkane: yes, all label schema crud methods on the dataset should have permissions

fiftyone/core/dataset.py
benjaminpkane: It will always be a `dict`, will add test assertion

fiftyone/core/dataset.py
benjaminpkane: A label schema is tied to a field. The name `label schema` comes from our [existing integration solution(s)](https://docs.voxel51.com/user_guide/annotation.html#label-schema).

I chose to pluralize it for more clarity around activating/deactivating individual fields through `active_label_schemas`, etc.

fiftyone/core/dataset.py
benjaminpkane: sure

fiftyone/core/annotation/generate_label_schemas.py
benjaminpkane: `schema` should only ever have one key when `is_scalar` is `True`, but returning early works too

fiftyone/core/annotation/generate_label_schemas.py
benjaminpkane: Label schema generation tests are [here](https://github.com/voxel51/fiftyone/blob/ha/generate-label-schema/tests/unittests/annotation/generate_label_schemas_tests.py)

fiftyone/core/annotation/generate_label_schemas.py
benjaminpkane: Hmm, I think I just did that for presentational reasons. It has no effect on behavior

fiftyone/core/annotation/validate_label_schemas.py
benjaminpkane: Below this we check if it is a list field, and then reassign `field = field.field` if so

So a list of strings being read only should be acknowledged. It's subfield being read only (the strings) is not supported/acknowledged

fiftyone/core/annotation/validate_label_schemas.py
benjaminpkane: sure, that's reasonable

PR #6616, opened 2025-11-24
move confusion matrix sorting to js in ME panel

imanjra:

- move confusion matrix sorting to js in ME panel
- Compress confusion_matrix cached in mongo using ES 

Using ME panel

        notes for FiftyOne users.

N/A

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Plotly-ready confusion matrix data and plotting helpers, default matrix config, class masking, and improved comparison handling with logarithmic colorscales.

* **Chores**
  * Unified confusion-matrix data flow, added compression/serialization helpers and caching hooks, safer handling when comparison data is missing.

* **UI**
  * Inline status text, right-aligned controls, and refined single-vs-comparison layout and interactions.

### DIFF
--- a/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/evaluation/overview/ConfusionMatrices.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/evaluation/overview/ConfusionMatrices.tsx
@@ -31,32 +31,38 @@ export default function ConfusionMatrices(props) {
     const {
       matrix,
       classes,
-      oranges_colorscale,
+      primary_colorscale,
       oranges_logarithmic_colorscale,
     } = evaluation?.confusion_matrix || {};
+    const { gt_field, pred_field } = evaluation?.info?.config || {};
     return getConfusionMatrixPlotlyData(
       {
         classes,
         matrix,
         colorscales: {
-          default: oranges_colorscale,
+          default: primary_colorscale,
           logarithmic: oranges_logarithmic_colorscale,
         },
         maskTargets: { primary: evaluationMaskTargets },
       },
-      confusionMatrixConfig
+      { ...confusionMatrixConfig, gtField: gt_field, predField: pred_field }
     );
   }, [evaluation, confusionMatrixConfig, evaluationMaskTargets]);
   const compareConfusionMatrixPlotlyData = useMemo(() => {
-    const { matrix, classes, blues_colorscale, blues_logarithmic_colorscale } =
-      compareEvaluation?.confusion_matrix || {};
+    const {
+      matrix,
+      classes,
+      secondary_colorscale,
+      blues_logarithmic_colorscale,
+    } = compareEvaluation?.confusion_matrix || {};
+    const { gt_field, pred_field } = compareEvaluation?.info?.config || {};
     if (matrix && classes) {
       return getConfusionMatrixPlotlyData(
         {
           classes,
           matrix,
           colorscales: {
-            default: blues_colorscale,
+            default: secondary_colorscale,
             logarithmic: blues_logarithmic_colorscale,
           },
           maskTargets: {
@@ -64,7 +70,7 @@ export default function ConfusionMatrices(props) {
             secondary: evaluationMaskTargets,
           },
         },
-        confusionMatrixConfig
+        { ...confusionMatrixConfig, gtField: gt_field, predField: pred_field }
       );
     }
     return [];
--- a/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/evaluation/scenario/Scenarios.tsx
+++ b/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/evaluation/scenario/Scenarios.tsx
@@ -1382,7 +1382,7 @@ function ConfusionMatrixChart(props) {
       classes: matrixData?.classes,
       matrix: matrixData?.matrix,
       colorscales: {
-        default: matrixData?.oranges_colorscale,
+        default: matrixData?.primary_colorscale,
         logarithmic: matrixData?.oranges_logarithmic_colorscale,
       },
       maskTargets: { primary: evaluationMaskTargets },
@@ -1396,7 +1396,7 @@ function ConfusionMatrixChart(props) {
           classes: compareMatrixData?.classes,
           matrix: compareMatrixData?.matrix,
           colorscales: {
-            default: compareMatrixData?.oranges_colorscale,
+            default: compareMatrixData?.primary_colorscale,
             logarithmic: compareMatrixData?.oranges_logarithmic_colorscale,
           },
           maskTargets: {
--- a/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/utils.ts
+++ b/app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/utils.ts
@@ -26,8 +26,7 @@ export function getClasses(
   maskTargets?: MaskTargets["primary"]
 ) {
   if (!matrixData) return [];
-  const classes = matrixData[`classes`];
-  return classes.map((c) => {
+  return matrixData.classes.map((c) => {
     return maskTargets?.[c] || c;
   });
 }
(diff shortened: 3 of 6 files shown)


app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/utils.ts
```diff
@@ -214,3 +138,148 @@ export function getInapplicableMetrics(evaluation) {
...
+
+type ClassWithCount = {
+  class: string;
+  count: number;
+};
+
+type MaskTargets = {
+  primary: Record<string, string>;
+  secondary?: Record<string, string>;
+};
```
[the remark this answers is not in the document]
imanjra: 1. Ok
2. Ok
3. They are not optional
4. sortBy is not optional anymore

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/evaluation/overview/ConfusionMatrices.tsx
```diff
@@ -4,56 +4,84 @@ import { Box, Stack, Typography } from "@mui/material";
...
-  const confusionMatrix = useMemo(() => {
-    return getMatrix(
-      evaluation?.confusion_matrices,
-      confusionMatrixConfig,
-      evaluationMaskTargets
+  const confusionMatrixPlotlyData = useMemo(() => {
+    const {
+      matrix,
+      classes,
+      oranges_colorscale,
```
tom-vx51: [nit] we should avoid encoding colors in variable names; no guarantee that this is consistent with the colors actually being used. Something like `primary`/`secondary` might be more appropriate.

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/evaluation/scenario/Scenarios.tsx
```diff
@@ -1368,35 +1369,45 @@ function ConfusionMatrixChart(props) {
...
     props.compareEvaluation?.mask_targets || {};
-  const matrices = subsetData?.confusion_matrices;
-  const matrixData = getMatrix(
-    matrices,
-    config,
-    evaluationMaskTargets,
-    undefined,
-    true
+  const matrixData = subsetData?.confusion_matrix;
+  const matrixPlotlyData = getConfusionMatrixPlotlyData(
```
tom-vx51, line 1380: [nit] can reduce code duplication by making these a function of the confusion matrix metadata

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/utils.ts
```diff
@@ -21,88 +21,12 @@ export function getNumericDifference(
...
-
-  return baseMatrix;
-}
-
-export function getClasses(matrices, maskTargets?) {
-  if (!matrices) return [];
-  const sortBy = "az";
-  const classes = matrices[`${sortBy}_classes`];
+  if (!matrixData) return [];
+  const classes = matrixData[`classes`];
```
tom-vx51: [nit] no need for backticks here

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/utils.ts
```diff
@@ -214,3 +138,154 @@ export function getInapplicableMetrics(evaluation) {
...
+
+  const sortedClassesWithCount = classesWithCount.sort(
+    (a: ClassWithCount, b: ClassWithCount) => {
+      if (a.class === NONE_CLASS) {
+        return 1;
+      }
+      if (b.class === NONE_CLASS) {
+        return -1;
+      }
+      if (sortBy === "az") return a.class.localeCompare(b.class);
```
tom-vx51, line 179: [nit] comments and/or constants would be helpful here to understand what's happening

app/packages/core/src/plugins/SchemaIO/components/NativeModelEvaluationView/utils.ts
```diff
@@ -214,3 +138,154 @@ export function getInapplicableMetrics(evaluation) {
...
+type MatrixOptions = {
+  sortBy: "az" | "za" | "mc" | "lc";
+  limit?: number;
+  classes?: string[];
+  skipZeroCount?: boolean;
+};
+
+type MatrixPlotDataConfig = MatrixOptions & {
+  log?: boolean;
+  gt_field?: string;
```
tom-vx51: [nit] camel case

plugins/panels/model_evaluation/__init__.py
```diff
@@ -15,14 +15,13 @@
...
-from collections import Counter, defaultdict
-import fiftyone as fo
+from collections import defaultdict
 import fiftyone.core.view as fov
 import fiftyone.operators.types as types
 import fiftyone.server.utils as fosu
 import fiftyone.utils.eval as foue
 from fiftyone import ViewField as F
-from fiftyone.core.plots.plotly import _to_log_colorscale
+from fiftyone.core.plots.plotly import _get_colorscale, _to_log_colorscale
```
tom-vx51, line 23: [nit] we should make these non-private if they're being imported

plugins/panels/model_evaluation/__init__.py
```diff
@@ -47,7 +52,7 @@
     os.environ.get("FIFTYONE_DISABLE_EVALUATION_CACHING") not in TRUTHY_VALUES
 )
 CACHE_TTL = 30 * 24 * 60 * 60  # 30 days in seconds
-CACHE_VERSION = "v2.0.0"
+CACHE_VERSION = "v2.13.0"
```
tom-vx51: does this need to be bumped on every release? this is an odd pattern

fiftyone/operators/cache/serialization.py
```diff
@@ -55,6 +55,8 @@ def auto_serialize(value):
         return value.tolist()
     elif isinstance(value, np.generic):
         return value.item()
+    elif isinstance(value, bytes):
+        return value
```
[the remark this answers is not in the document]
tom-vx51: @imanjra i'm not familiar with this code but please check this and verify the behavior

plugins/panels/model_evaluation/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    ]
+    compressed_matrix = {
+        "shape": matrix.shape,
+        "non_zero_matrix": non_zero_matrix,
+    }
+    compressed_matrix_binary = json.dumps(compressed_matrix).encode("utf-8")
+    return (
+        zlib.compress(compressed_matrix_binary)
+        if use_zlib
+        else compressed_matrix_binary
```
tom-vx51, line 24: [nit] a bit misleading that `compress_matrix` doesn't necessarily compress it. Would be better to separate into `serialize_matrix` + `compress_bytes` or similar.

plugins/panels/model_evaluation/__init__.py
imanjra: Not necessarily every release, but every time the structure/content of cached data is updated. Bumping version here forces the cache to be dropped and updated

plugins/panels/model_evaluation/utils.py
imanjra: Might've been missed, but it does compress. The compression is two parts:

Part 1 (always happens) - reconstruct excluding all `0` values:
```
    non_zero_indices = np.nonzero(matrix)
    non_zero_matrix = [
        {"i": int(i), "j": int(j), "v": int(matrix[i, j])}
        for i, j in zip(*non_zero_indices)
    ]
    compressed_matrix = {
        "shape": matrix.shape,
        "non_zero_matrix": non_zero_matrix,
    }
```

Part 2 (optional based on `use_zlib`) - binary compression:

```
    compressed_matrix_binary = json.dumps(compressed_matrix).encode("utf-8")
    return (
        zlib.compress(compressed_matrix_binary)
        if use_zlib
        else compressed_matrix_binary
    )
```

plugins/panels/model_evaluation/__init__.py
imanjra: _Also, the version here is independent of release version_

plugins/panels/model_evaluation/utils.py
tom-vx51: if it's sparse then yes, but we don't check the number of non-zeros so in the average case converting to json will actually inflate its size. It might be fine in this specific case, just noting that the method is misleading.

imanjra: Yah, good point. I agree. In general, this does not necessarily compress for a non-sparse matrix. However, for this case, the matrix is fairly sparse. I will rename it to be more accurate than now

PR #6627, opened 2025-12-01
FIx possible recursion during reload

kaixi-wang:

Fix potential recursive reload when the sample.id is not None, but does not exist in the database:

```
  File "/Users/kacey/.pyenv/versions/3.11.9/lib/python3.11/asyncio/base_events.py", line 641, in run_until_complete
    self.run_forever()
  File "/Users/kacey/.pyenv/versions/3.11.9/lib/python3.11/asyncio/base_events.py", line 608, in run_forever
    self._run_once()
  File "/Users/kacey/.pyenv/versions/3.11.9/lib/python3.11/asyncio/base_events.py", line 1928, in _run_once
    handle._run()
  File "/Users/kacey/.pyenv/versions/3.11.9/lib/python3.11/asyncio/events.py", line 84, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/kacey/Dev/voxel-hub/api/tests/e2e/mutation/test_e2e_dataset_snapshots_mutations.py", line 424, in test_it
    dataset.reload()
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/dataset.py", line 8624, in reload
    self._reload_docs(hard=True)
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/dataset.py", line 8667, in _reload_docs
    fos.Sample._reload_docs(self._sample_collection_name, hard=hard)
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/singletons.py", line 215, in _reload_docs
    sample.reload(hard=hard, include_frames=include_frames)
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/sample.py", line 623, in reload
    super().reload(hard=hard)
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/document.py", line 560, in reload
    self._reload_backing_doc()
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/sample.py", line 608, in _reload_backing_doc
    self._doc = self._dataset._sample_dict_to_doc(d)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/dataset.py", line 8526, in _sample_dict_to_doc
    self.reload()
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/dataset.py", line 8624, in reload
    self._reload_docs(hard=True)
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/dataset.py", line 8667, in _reload_docs
    fos.Sample._reload_docs(self._sample_collection_name, hard=hard)
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/singletons.py", line 215, in _reload_docs
    sample.reload(hard=hard, include_frames=include_frames)
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/sample.py", line 623, in reload
    super().reload(hard=hard)
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/document.py", line 560, in reload
    self._reload_backing_doc()
  File "/Users/kacey/Dev/voxel-hub/fiftyone-teams/fiftyone/core/sample.py", line 608, in _reload_backing_doc
    self._doc = self._dataset._sample_dict_to_doc(d)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

api tests pass

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness when reloading sample and frame data: missing backing documents now log a warning and are skipped instead of causing invalid updates.
  * On load failures the system now performs a hard in-memory reload and retries reconstruction of data.
  * Added debug logging for load errors and a safeguard to prevent infinite reload/retry loops.

### DIFF
--- a/fiftyone/core/dataset.py
+++ b/fiftyone/core/dataset.py
@@ -8261,7 +8261,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         doc = self._sample_dict_to_doc(d)
         return fos.Sample.from_doc(doc, dataset=self)
 
-    def _sample_dict_to_doc(self, d, _reload_backing_docs=True):
+    def _sample_dict_to_doc(self, d, *, _reload_backing_docs=True):
         try:
             return self._sample_doc_cls.from_dict(d)
         except Exception as e:
@@ -8283,7 +8283,7 @@ class Dataset(foc.SampleCollection, metaclass=DatasetSingleton):
         doc = self._frame_dict_to_doc(d)
         return fofr.Frame.from_doc(doc, dataset=self)
 
-    def _frame_dict_to_doc(self, d, _reload_backing_docs=True):
+    def _frame_dict_to_doc(self, d, *, _reload_backing_docs=True):
         try:
             return self._frame_doc_cls.from_dict(d)
         except Exception as e:


fiftyone/core/sample.py
```diff
@@ -536,7 +539,15 @@ def _reload_backing_doc(self):
             return
 
         d = self._dataset._sample_collection.find_one({"_id": self._id})
-        self._doc = self._dataset._sample_dict_to_doc(d)
+        if d:
+            self._doc = self._dataset._sample_dict_to_doc(d)
```
CamronStaley: do we wanna fix this here or in sample_dict_to_doc? That method feels weird to me why are we blindly reloading for any exception are we sure that `d == None` is the only time that this can happen? I think I have seen this before with temporary schema changes where the sample itself is in the dataset but the schema doesn't match.

```
    def _sample_dict_to_doc(self, d):
        try:
            return self._sample_doc_cls.from_dict(d)
        except:
            # The dataset's schema may have been changed in another process;
            # let's try reloading to see if that fixes things
            self.reload()

            return self._sample_doc_cls.from_dict(d)
```

kaixi-wang: >  temporary schema changes where the sample itself is in the dataset but the schema doesn't match.

Yeah a change in the in memory schema often triggers the exception block in your comment, but that usually just leads to the error message along the lines of `field does not exist`. 

I went with the current because it was the most straightforward/targeted to handle (return early if not in db). 

If we are to handle it in _sample_dict_to_doc(), would we want to throw an error if d is None? Or return early? I feel like there is uncertainty there.

swheaton: I agree. I think we could do both?
Though neither solution explains how we get into that situation in the particular test...

In particular, in this instance it is safe to do `self._reload(hard=True)` but NOT `self._reload_docs(hard=True)`. So we could just do the former instead of `self.reload()`. This code hasn't changed in years so it's interesting.

If we are just adding samples or something instead of within reloading backing docs, then it should be safe to do `self.reload()` including reload_docs; we don't want samples to be out of sync with the schema.

```python
    def _sample_dict_to_doc(self, d, _backing_doc=False):
        try:
            return self._sample_doc_cls.from_dict(d)
        except:
            # The dataset's schema may have been changed in another process;
            # let's try reloading to see if that fixes things
            self._reload(hard=True)
            should_reload_backing_docs = not _backing_doc
            if should_reload_backing_docs
                self._reload_docs(hard=True)

            return self._sample_doc_cls.from_dict(d)
```

kaixi-wang: taking a closer look at the code, it makes sense to check the field schema with `self._reload`,  but why would we ever reload all in memory sample docs if there's an exception loading a single sample from the dictionary? In other words, where would should_reload_backing_docs be true?

In the error path, self._reload_docs(hard=True) at best does nothing when there are no in memory samples, but if there are any problematic in memory samples, it just leads to sample.reload -> Document.reload -> _reload_backing_doc -> self._dataset._sample_dict_to_doc

swheaton: > where would should_reload_backing_docs be true?

when adding / updating a sample. This will typically be the case except for in the reload backing doc function specifically, where we would call this to prevent recursion:

```python
        if d:
            self._doc = self._dataset._sample_dict_to_doc(d, _backing_doc=True)
```

brimoor: Can you provide a minimum reproducible example of the error that this change is targeting?

Based on the code comment that is added here, I'm inferring that it has something to do with reloading a sample singleton when it has been deleted in another process?

I tried to reproduce as follows on `develop` and see an acceptable behavior as-is:

```py
# In process A

import fiftyone as fo

sample = fo.Sample(filepath="image.jpg")

dataset = fo.Dataset("test")
dataset.add_sample(sample)
```

```py
# In process B

import fiftyone as fo

dataset = fo.load_dataset("test")
dataset.delete_samples(dataset.first())
```

```py
# back in process A

sample.reload()
# DoesNotExist: Document does not exist
```

brimoor: Don't change the behavior of the `try-except` in `Dataset._sample_dict_to_doc()`. Per the code comment, the behavior is intentional to gracefully pull in changes to a dataset's schema that have happened in other processes (in-memory `dataset._doc` is a singleton and thus can become stale) without breaking whatever action the current process is taking (eg iterating over the dataset).

Reflecting a schema change requires reconstituting the `dataset._sample_doc_cls` instance, which I'm pretty sure (it's been years since I've looked at this) requires reconstituting any in-memory `Sample` and `Frame` objects in order for their validation logic to respect the now-different schema (hence the sample/frame doc reloading).

kaixi-wang: I was able to repro the recursion and confirm it was fixed by the changes in this pr.
To repro:

```
# In process A
import fiftyone as fo
import fiftyone.core.sample as fos

# setup to get samples into the singleton cache
ds=fo.load_dataset('reg-add-samples')
samples = []
for i in range(5):
    samples.append(fo.Sample(filepath=f"img{i}.jpg"))
ds.add_samples(samples)

# get the id of a sample in the cache
fos.Sample._instances
# defaultdict(weakref.WeakValueDictionary,{'samples.692e14b41c869f6acd8820bf': <WeakValueDictionary at 0x37ad2f8c0>})
vals=[val for val in fos.Sample._instances['samples.692e14b41c869f6acd8820bf'].values()]
vals[-1]['id']
# '692e614cb739f1bbd5712d0c'

```

```
# In process B
import fiftyone as fo
ds = fo.load_dataset('reg-add-samples')
ds._sample_collection.delete_one({"_id": ObjectId('692e614cb739f1bbd5712d0c')})
```


``` 
# In process A
ds.reload()
```

brimoor: Okay thanks, so it specifically happens when one calls `dataset.reload()` rather than `sample.reload()`. I can reproduce that as well 👍 

I am onboard with logging a warning if `d` is empty, but can we please be more opinionated about the log message? 

```py
logger.warning(f"Sample with ID {self._id} has been deleted")
```

swheaton: > Don't change the behavior of the try-except in Dataset._sample_dict_to_doc().

Why? This is leaving in a potential for infinite recursion if some other sort of error happens. I was suggesting no change in behavior except when coming from `_reload_backing_docs()`. If we try to reload backing docs again, we'll end up in the infinite loop

brimoor: Sorry let me clarify, my point was: don't delete the `try-except: reload()` from `Dataset._sample_dict_to_doc()` because it serves a purpose. 

If more tweaks are needed to prevent infinite recursion, that is okay.

swheaton: @kaixi-wang It might be a bit contrived however we should ensure infinite recursion is never possible

```python
# Fake a "session 1"
import fiftyone as fo
ds = fo.Dataset()
s = fo.Sample("blah.jpg")
ds.add_sample(s)

# Fake a "session 2" or some other sort of update.
conn = fo.core.odm.get_db_conn()
conn[ds._sample_collection_name].update_one({"_id": s._id}, {"$set": {"blah": "blahh"}})

# Back to "session 1"
ds.reload()  # infinite recursion here

ds.delete()
```

CamronStaley (approved):
Verified that the previously failing test no longer gets stuck in a loop in voxel-hub with these changes checked out locally. Would like to see approval from Stu or someone who can verify this doesn't negatively impact any weird cases that aren't covered by our automated tests.

fiftyone/core/dataset.py
```diff
@@ -8263,27 +8261,43 @@ def _make_sample(self, d):
         doc = self._sample_dict_to_doc(d)
         return fos.Sample.from_doc(doc, dataset=self)
 
-    def _sample_dict_to_doc(self, d):
+    def _sample_dict_to_doc(self, d, *, _reload_backing_docs=True):
```
swheaton, line 8264: nit: it's more like `_reload_backing_docs_on_error`. which is long. which is why in my example I just used `_backing_doc` and `should_reload_backing_doc = not _backing_doc`

kaixi-wang: Yeah `_reload_backing_docs_on_error` is more accurate, but `_backing_doc` alone was kinda without looking into the usage within the private function.

PR #6634, opened 2025-12-03
FiftyOne Labs CLI + plugin search upgrades

manushreegangwar:
## Release Notes

- added a `fiftyone plugins search` CLI command to search for plugins available in a github repo
- added a `fiftyone plugins list --glob-patt` option to list plugins matching a glob pattern
- added a `fiftyone labs` CLI command that supports installing, uninstalling, listing, and searching for FiftyOne Labs features

### Plugin search

```
fiftyone plugins search https://github.com/voxel51/fiftyone-plugins
fiftyone plugins search https://github.com/voxel51/fiftyone-plugins --path plugins/brain
```

### Plugin glob

```
fiftyone plugins list --glob-patt '@voxel51/*'
```

### FiftyOne Labs

```
fiftyone labs -h

fiftyone labs search
fiftyone labs search --path plugins/labs_panel

fiftyone labs install --all
fiftyone labs install @51labs/labs_panel --branch main --overwrite

fiftyone labs list

fiftyone labs uninstall @51labs/labs_panel
fiftyone labs uninstall --all

fiftyone labs list
```

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -6,6 +6,7 @@ jobs:
   test-e2e:
     timeout-minutes: 90
     runs-on: ubuntu-latest-m
+    name: Test e2e (Mongo ${{ matrix.mongodb-version }})
     env:
       FIFTYONE_DO_NOT_TRACK: true
       FIFTYONE_APP_DEFAULT_QUERY_PERFORMANCE: false
@@ -17,6 +18,14 @@ jobs:
     defaults:
       run:
         shell: bash
+    strategy:
+      fail-fast: false
+      matrix:
+        mongodb-version:
+          - "6"
+          - "7"
+          - "8.0"
+          - "latest"
     steps:
       - name: Checkout
         uses: actions/checkout@v6
@@ -36,7 +45,7 @@ jobs:
       - name: Start MongoDB
         uses: supercharge/mongodb-github-action@1.12.1
         with:
-          mongodb-version: latest
+          mongodb-version: ${{ matrix.mongodb-version }}
 
       - name: Setup
         uses: actions/setup-python@v6
@@ -114,7 +123,6 @@ jobs:
         with:
           path: ${{ env.PLAYWRIGHT_BROWSERS_PATH }}
           key: ${{ runner.os }}-playwright-${{ steps.playwright-version.outputs.PLAYWRIGHT_VERSION }}
-          working-directory: e2e-pw
 
       - name: Install Playwright browser if not cached
         run: yarn playwright install
@@ -133,6 +141,6 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v5
         with:
-          name: playwright-report
+          name: playwright-report-mongodb-${{ matrix.mongodb-version }}
           path: e2e-pw/playwright-report/
           retention-days: 30
--- a/.github/workflows/push-release.yml
+++ b/.github/workflows/push-release.yml
@@ -26,7 +26,7 @@ jobs:
           git checkout develop
           git pull origin merge/${{ inputs.ref_name || github.ref_name }} --no-rebase
           git pull origin ${{ inputs.ref_name || github.ref_name }} --no-rebase
-      - uses: peter-evans/create-pull-request@v7
+      - uses: peter-evans/create-pull-request@v8
         with:
           author: voxel51-bot <bot@voxel51.com>
           token: ${{ secrets.FIFTYONE_GITHUB_TOKEN }}
--- a/docs/source/tutorials/cosmos-transfer-integration.ipynb
+++ b/docs/source/tutorials/cosmos-transfer-integration.ipynb
@@ -42,7 +42,7 @@
     "\n",
     "### 2. Install Cosmos-Transfer 2.5 and Dependencies\n",
     "\n",
-    "Clone the [official repository](https://github.com/nvidia-cosmos/cosmos-transfer2.5) and set up the environment. You can also follow the instructions on this [Cosmos Transfer 2.5 Recipe](https://github.com/nvidia-cosmos/cosmos-cookbook/blob/main/docs/recipes/inference/transfer2_5/inference-biotrove-augmentation_w_FiftyOne/inference.md) and the [Setup Process](https://github.com/nvidia-cosmos/cosmos-cookbook/blob/main/docs/recipes/inference/transfer2_5/inference-biotrove-augmentation_w_FiftyOne/setup.md)\n",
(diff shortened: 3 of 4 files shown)


fiftyone/core/cli.py
```diff
@@ -4194,6 +4195,57 @@ def execute(parser, args):
...
+    """Install fiftyone labs features from the web.
+
+    You can install all labs features (plugins) from GitHub with the following::
+        fiftyone labs install
+    """
+
+    @staticmethod
+    def setup(parser):
+        parser.add_argument(
+            "--labs_repo",
```
brimoor: curious what the use case for a custom labs repo is here? 

Implementation-wise, this is just syntactic sugar around:

```
fiftyone plugins download --plugin-names XXX
```

so if someone is working with anything other than the official Labs repository, then should they just use that command?

fiftyone/core/cli.py
```diff
@@ -4194,6 +4195,57 @@ def execute(parser, args):
...
+
+    @staticmethod
+    def execute(parser, args):
+        parser.print_help()
+
+
+class LabsInstallCommand(Command):
+    """Install fiftyone labs features from the web.
+
+    You can install all labs features (plugins) from GitHub with the following::
```
brimoor: nit: can you use the same `Examples::` syntax that you see in other `Command` classes in this module for consistency?

fiftyone/core/cli.py
```diff
@@ -4194,6 +4195,57 @@ def execute(parser, args):
...
+        )
+
+    @staticmethod
+    def execute(parser, args):
+        fop.download_plugin(
+            args.labs_repo,
+            plugin_names=args.feature_names,
+            overwrite=args.overwrite,
+        )
+
```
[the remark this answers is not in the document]
brimoor: +1 to rabbit's suggestion to use hyphenated long options for consistency with the rest of the CLI

fiftyone/core/cli.py
```diff
@@ -4194,6 +4195,57 @@ def execute(parser, args):
...
+        subparsers = parser.add_subparsers(title="available commands")
+        _register_command(subparsers, "install", LabsInstallCommand)
+
+    @staticmethod
+    def execute(parser, args):
+        parser.print_help()
+
+
+class LabsInstallCommand(Command):
+    """Install fiftyone labs features from the web.
```
brimoor: `fiftyone labs` -> `FiftyOne Labs`

fiftyone/core/cli.py
```diff
@@ -4194,6 +4195,100 @@ def execute(parser, args):
...
+    def execute(parser, args):
+        if args.all:
+            names = fop.list_downloaded_plugins()
+        else:
+            names = args.name
+
+        for name in names:
+            if name.startswith("@51labs"):
+                fop.delete_plugin(name)
+
```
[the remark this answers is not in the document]
manushreegangwar: Good idea. Added

fiftyone/core/cli.py
manushreegangwar: Updated

fiftyone/core/cli.py
manushreegangwar: Done

fiftyone/core/cli.py
manushreegangwar: If I want to install features from a branch, I can use `labs-repo`:
```fiftyone labs install --labs-repo https://github.com/voxel51/labs/tree/<branch>```

Comment (manushreegangwar) [reply fiftyone/core/cli.py]: I don't mind removing this since there are other ways to do the above

**Review by swheaton — APPROVED** [review]
lgtm from static review. Thank you for addressing brian and code rabbit suggestions.

One thing I did notice was, how would I install specific feature names if I don't know what the features are?

**Comment by brimoor** [comment]
> One thing I did notice was, how would I install specific feature names if I don't know what the features are?

Yeah I was thinking about requesting that as well. In general I'd like to have a util that can scan a GH repo and tell me what plugins exist:

```
fiftyone plugins search $URL_OR_GH_REPO
```

(bonus points if this scan occurs without actually cloning the repo locally)

And then since `fiftyone labs install` just wraps `fiftyone plugins download`, I guess there would be a `fiftyone labs XXX` method that "rebrands" this functionality specifically for the Labs repo?

One concern I have about the `fiftyone labs` command in general is that `fiftyone plugins` contains lots of utils:

```
$ fiftyone plugins -h
usage: fiftyone plugins [-h] [--all-help] {list,info,download,requirements,create,enable,disable,delete} ...

Tools for working with FiftyOne plugins.

options:
  -h, --help            show this help message and exit
  --all-help            show help recursively and exit

available commands:
  {list,info,download,requirements,create,enable,disable,delete}
    list                List plugins that are installed locally.
    info                Prints info about plugins that are installed locally.
    download            Download plugins from the web.
    requirements        Handles package requirements for plugins.
    create              Creates or initializes a plugin.
    enable              Enables the given plugin(s).
    disable             Disables the given plugin(s).
    delete              Delete plugins from your local machine.
```

all of which are equally applicable to Labs features as well, as they are just plugins themselves...

So where does the rabbit hole end for `fiftyone labs`? Should we expect users to know that they are "just" plugins and thus they can use `fiftyone plugins YYY` commands to enable/disable/list/manage requirements/etc? Or should we wrap *all* of these commands as `fiftyone labs YYY` commands? The latter feels too duplicative to me, but if that's the case, I'm having a bit of a hard time convincing myself that it makes sense to wrap only some of this functionality in a `fiftyone labs` interface.

File: fiftyone/core/cli.py [hunk 6]
```diff
@@ -104,6 +105,7 @@ def setup(parser):
         _register_command(subparsers, "plugins", PluginsCommand)
         _register_command(subparsers, "utils", UtilsCommand)
         _register_command(subparsers, "zoo", ZooCommand)
+        _register_command(subparsers, "labs", LabsCommand)
```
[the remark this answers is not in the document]
manushreegangwar: Good call. Fixed

brimoor:
@manushreegangwar `fiftyone plugins search` works great! I forgot we already had search implemented in the SDK so it was easy to add to the CLI 🎉 

I tested this locally and felt like I wanted a `fiftyone labs list` CLI command to round out the functionality so I could have a fast way to see what Labs features I have locally, so I went ahead and added it to this PR.

^this functionality is also achievable via a new `--glob-patt` syntax I added:

```
fiftyone plugins list --glob-patt '@51labs/*
```

but that feels too difficult to construct from memory, so I lean towards keeping `fiftyone labs list` even though it is duplicitous. LMK what you think!

I was tempted to combine `fiftyone labs search` and `fiftyone labs list` into a single method that would simply print a table of all Labs features with column that would indicate which ones are currently installed. I do think that would be a helpful view. But, since we have options like `fiftyone labs search --branch --path` (whose value do make sense), it felt weird to combine the concepts because those flags aren't applicable to already installed Labs features locally. So ultimately I left the separate `fiftyone labs search|list` commands, which also gives us consistency with `fiftyone plugins list|search`. LMK what you think about this too.

btw I do agree that we don't need to add *yet more* duplications of plugin-related methods like `fiftyone labs enable`, etc. From the output of `fiftyone labs list` it is apparent that Labs features are actually plugins, which means that `fiftyone plugins enable` and related operations are fair game.

brimoor (approved):
Tested the commands in the PR description and everything works as expected

PR #6639, opened 2025-12-04
Update `get_implied_field_kwargs()` to support `EmbeddedDocumentField` subclasses

lorenzoriano:
## Summary

- updates `foo.get_implied_field_kwargs()` to correctly handle values that contain `EmbeddedDocumentField` subclasses

## Tested by

```py
import fiftyone as fo

class _GetImpliedInner(foo.EmbeddedDocument):
    x = fo.FloatField()
    score = fo.FloatField()

class _GetImpliedSubclassedEmbeddedDocField(fo.EmbeddedDocumentField):
    pass

class _GetImpliedOuter(foo.EmbeddedDocument):
    pose = _GetImpliedSubclassedEmbeddedDocField(_GetImpliedInner)

values = [
    _GetImpliedOuter(pose=_GetImpliedInner(x=1.0)),
    _GetImpliedOuter(pose=_GetImpliedInner(score=0.5)),
]

kwargs = foo.get_implied_field_kwargs(values)
fo.pprint(kwargs)

# Fails on `develop` (only the 'x' attribute is recognized)
# Succeeds on this branch (both 'x' and 'score' attributes are recognized)
assert len(kwargs["fields"][0]["fields"]) == 2
```

### DIFF
--- a/fiftyone/core/odm/utils.py
+++ b/fiftyone/core/odm/utils.py
@@ -10,6 +10,7 @@ import inspect
 import json
 import numbers
 import sys
+
 from bson import Binary, json_util, ObjectId, SON
 import numpy as np
 import pytz
@@ -413,7 +414,9 @@ def get_implied_field_kwargs(value, dynamic=False):
             value_type = next(iter(value_types))
             kwargs["subfield"] = value_type
 
-            if value_type == fof.EmbeddedDocumentField:
+            if inspect.isclass(value_type) and issubclass(
+                value_type, fof.EmbeddedDocumentField
+            ):
                 document_type = _get_list_subfield_type(value)
                 fields = _parse_embedded_doc_list_fields(value, dynamic)
                 kwargs["embedded_doc_type"] = document_type


fiftyone/core/odm/utils.py
```diff
@@ -547,7 +547,13 @@ def _finalize_embedded_doc_fields(fields_dict):
     for field in fields_dict.values():
         if field is not None:
             fields.append(field)
-            if field["ftype"] == fof.EmbeddedDocumentField:
+            ftype = field["ftype"]
+            is_embedded_doc = inspect.isclass(ftype) and issubclass(
```
brimoor: Can you provide the example call to `foo.get_implied_field_kwargs()` that resulted in a case where there was a `field` whose type was a strict subclass of `EmbeddedDocumentField` in this method?

brimoor: `ftype == fof.EmbeddedDocumentField` comparisons are used in a couple other methods in this module as well, so I would think that if we need a patch here, then we'd need a patch in those methods too

brimoor: Generally speaking, we do add more `EmbeddedDocument` subclasses to represent different data models in FO over time, but we do not define `EmbeddedDocumentField` subclasses, as the latter is just the generic concept of a field that contains a document.

lorenzoriano: Yup you're right, there isn’t an in-repo call path today that supplies a strict subclass of EmbeddedDocumentField.  The issubclass guards are defensive to avoid silently skipping nested schemas if a downstream plugin or future extension ever introduces a subclass (which is what I ended up doing at one point and discovered this issue). I updated the remaining equality check in _finalize_embedded_doc_fields to use the same issubclass pattern we already use in _merge_embedded_doc_fields, so merge/finalize behavior stays consistent and safe for hypothetical subclasses while remaining identical for current inputs.

fiftyone/core/odm/utils.py
```diff
@@ -492,62 +493,143 @@ def _parse_embedded_doc_fields(doc, dynamic):
...
+    that repeated traversals can be merged efficiently. This helper normalizes
+    the structure back into the list-of-field-specs format expected by
+    `create_field`, recursively applying the same conversion to any nested
+    embedded-document entries.
+    """
+    if not fields_dict:
+        return []
+
+    if isinstance(fields_dict, list):
+        # Older callers may still provide a list that already filtered out
```
brimoor: nit: we don't need to support backwards-compatibility in these private methods. We can assume they are strictly invoked by `get_implied_field_kwargs()`

brimoor (commented):
I'm still trying to wrap my head around the actual error case here. The methods you're tweaking are strictly invoked by calling `get_implied_field_kwargs()` so can you provide an example that raises an exception on `develop`?

```py

# what fails here?
value = ...
kwargs = foo.get_implied_field_kwargs(value, ...)
```

The unit tests are contrived ways of invoking the private methods directly, which users would never do. I'd like to have "integration" tests that validate the behavior of `get_implied_field_kwargs()` instead.

lorenzoriano:
> I'm still trying to wrap my head around the actual error case here. The methods you're tweaking are strictly invoked by calling `get_implied_field_kwargs()` so can you provide an example that raises an exception on `develop`?
> 
> ```python
> import fiftyone.core.odm as foo
> 
> # what fails here?
> value = ...
> kwargs = foo.get_implied_field_kwargs(value, ...)
> ```
> 
> The unit tests are contrived ways of invoking the private methods directly, which users would never do. I'd like to have "integration" tests that validate the behavior of `get_implied_field_kwargs()` instead.

The bug shows up when get_implied_field_kwargs() is asked to infer a schema for nested embedded-document structures, specifically when:
* The value is a list of embedded documents, and
* Those embedded documents themselves contain embedded-document fields.

I had this issue when working on the new (and now removed) labels in HRM2.  The old implementation of _merge_embedded_doc_fields() assumed its fields_dict argument was always a dict, but in practice it was sometimes a list of field-specs coming. When merging them, it would end up doing something equivalent to:
`fields_dict[name] = field  # `fields_dict` is actually a list here`
which on develop raises:
`TypeError: list indices must be integers or slices, not str`
from inside _merge_embedded_doc_fields(). 

I’ve added a test that uses get_implied_field_kwargs(). The new GetImpliedFieldKwargsTests.test_list_of_embedded_docs_merges_nested_schema builds a list of embedded documents where the nested field is a subclass of EmbeddedDocumentField, and each element’s inner document exposes different fields.

fiftyone/core/odm/utils.py
lorenzoriano: done

PR #6651, opened 2025-12-08
FOEPD-2576: fix camera position resolution bug

sashankaryal:
There's an expected precedence order which has a bug. See "Files changed" tab for reference + fix.

Expected behavior = user defined camera position in fo3d file should take precedence over browser saved camera position


app/packages/looker-3d/src/fo3d/MediaTypeFo3d.tsx
sashankaryal, line 344: Note: this is the expected precedence

tom-vx51 (approved):

PR #6657, opened 2025-12-11
Fix behavior change with sklearn>=1.8

swheaton:

In sklearn 1.8 and later, passing empty arrays raises an error instead of returning zeros.
https://github.com/scikit-learn/scikit-learn/pull/32549
Handle this case here now to maintain consistent functionality with sklearn < 1.8

Unit tests now pass with both scikit-learn 1.7.x and 1.8.0

        notes for FiftyOne users.

Fixes error raised with classification evaluation results on an empty view, with scikit-learn>=1.8

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Reporting now detects when there are no samples and displays "No samples to analyze".
  * Precision, recall, and F1-score computations return zeros for empty inputs, ensuring stable, consistent evaluation metrics.

### DIFF
--- a/fiftyone/utils/eval/base.py
+++ b/fiftyone/utils/eval/base.py
@@ -900,6 +900,17 @@ class BaseClassificationResults(BaseEvaluationResults):
         #   Handle this case here now to maintain consistent functionality with
         #   sklearn < 1.8
         if self.ytrue.size == 0:
+            # If average is None, return arrays of zeros. From function
+            # docstring:
+            # precision : float (if average is not None) or array of float,
+            #         shape =\
+            #         [n_unique_labels]
+            #         Precision score.
+            if average is None:
+                num_labels = 0 if labels is None else len(labels)
+                zeros = np.zeros(num_labels, dtype=float)
+                return zeros, zeros, zeros
+
             return 0.0, 0.0, 0.0
 
         precision, recall, fscore, _ = skm.precision_recall_fscore_support(


fiftyone/utils/eval/base.py
```diff
@@ -890,6 +890,17 @@ def _parse_classes(self, classes):
...
         pos_label = getattr(self, "_pos_label", None)
+
+        # In sklearn 1.8 and later, passing empty arrays raises an error
+        #   instead of returning zeros.
+        #   https://github.com/scikit-learn/scikit-learn/pull/32549
+        #   Handle this case here now to maintain consistent functionality with
+        #   sklearn < 1.8
+        if any(
+            arr is None or len(arr) < 1 for arr in (self.ytrue, self.ypred)
+        ):
```
exupero: Are `self.ytrue` and `self.ypred` not falsey when empty? Or are there cases where they are falsey (but not None or empty) that we don't want to catch?
```suggestion
        if not self.ytrue or not self.ypred:
```

brimoor: Annoyingly, numpy does not like truthy or falsey checks:

```py
import numpy as np
x = np.array([])
bool(x)
```

```
# DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
```

brimoor: However, this would be sufficient:

```py
if self.ytrue.size == 0:
    return 0.0, 0.0, 0.0
```

because `ytrue` must be an array of same length as `ypred`.

Note that there are other checks in this module that explicitly check this (and only this), so that would be my preference for consistency.

swheaton: I was matching it closer to what they do in sklearn where the error is raised but sure yeah that's fine

fiftyone/utils/eval/base.py
```diff
@@ -890,6 +890,17 @@ def _parse_classes(self, classes):
...
+            return 0.0, 0.0, 0.0
+
```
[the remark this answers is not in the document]
swheaton: [525d23b](https://github.com/voxel51/fiftyone/pull/6657/commits/525d23b8e3b565eb292fde67075d1cfe2eeff48f)

PR #680, opened 2020-11-10
Adding support for exporting/importing labels to/from Scale AI

brimoor:
Adds support for importing labels in [Scale AI format](https://docs.scale.com/reference#general-image-annotation), as well as exporting labels in a format that can be uploaded as pre-annotations via Scale AI tasks' `hypothesis` parameter.

All of the following label types are supported:

Images
- [General image annotation](https://docs.scale.com/reference#general-image-annotation), including boxes, polylines, polygons, and keypoints
- [Semantic segmentation](https://docs.scale.com/reference#semantic-segmentation-annotation)

Videos
- [General video annotation](https://docs.scale.com/reference#general-video-annotation)
- [Video playback](https://docs.scale.com/reference#video-playback)

Here's an example of exporting BDD labels into Scale AI format and then importing them back into the same dataset using the API introduced in this PR:

See `tests/misc/scale_tests.py` for additional tests.

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.utils.scale as fous

SCALE_EXPORT_PATH = "/tmp/scale-image-export.json"
SCALE_IMPORT_PATH = "/tmp/scale-image-import.json"
SCALE_ID_FIELD = "scale_id"

_dataset = foz.load_zoo_dataset("bdd100k", split="validation", shuffle=True, max_samples=10)
dataset = fo.Dataset()
dataset.add_samples(_dataset.take(10))

# Export labels in Scale format
fous.export_to_scale(
    dataset,
    SCALE_EXPORT_PATH,
    label_prefix="",  # all fields
)

# Convert to Scale import format
id_map = fous.convert_scale_export_to_import(
    SCALE_EXPORT_PATH, SCALE_IMPORT_PATH
)

for sample_id, task_id in id_map.items():
    sample = dataset[sample_id]
    sample[SCALE_ID_FIELD] = task_id
    sample.save()

# Import labels from Scale
fous.import_from_scale(
    dataset,
    SCALE_IMPORT_PATH,
    label_prefix="scale",
    scale_id_field=SCALE_ID_FIELD,
)

# Verify that we have two copies of the same labels
session = fo.launch_app(dataset)
```

Screen Shot 2020-11-09 at 9 07 32 PM

PR #6658, opened 2025-12-11
Fix TF2 SavedModel inference for non-callable models

CharlesCNorton:
## Summary
- Fixes `centernet-mobilenet-v2-fpn-512-coco-tf2` and other TF2 Model Zoo models failing with `TypeError: '_UserObject' object is not callable`
- In TensorFlow 2.20+, `tf.saved_model.load()` returns a `_UserObject` that is not directly callable
- Adds a patch to use `loaded.signatures['serving_default']` instead

## Testing
Verified locally with `centernet-mobilenet-v2-fpn-512-coco-tf2` on the quickstart dataset - inference produces detections successfully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Apply a lazy TensorFlow SavedModel compatibility patch when entering ETA model contexts to support both signature-based and direct-call TF2 detectors, avoid repeated patching, and gracefully handle missing TensorFlow (logs a warning if patching fails).

* **Tests**
  * Added unit tests for loader creation, callable vs signature-based SavedModels, generic output handling, idempotent patching behavior, and behavior when TensorFlow is unavailable.

### DIFF
--- a/fiftyone/utils/eta.py
+++ b/fiftyone/utils/eta.py
@@ -27,6 +27,7 @@ import eta.core.video as etav
 
 import fiftyone.core.labels as fol
 import fiftyone.core.models as fom
+import fiftyone.core.utils as fou
 
 
 _IMAGE_MODELS = (
@@ -44,20 +45,15 @@ _VIDEO_MODELS = (
     etal.VideoSemanticSegmenter,
 )
 
-_tf2_patched = False
+def _make_tf2_detection_model_loader():
+    """Creates a fixed TF2 detection model loader.
 
+    Some models export with signatures rather than as directly callable
+    objects. This loader handles both cases.
 
-def _patch_tf2_detection_model():
-    """Patches TF2 SavedModel loading for TF2 Model Zoo models.
-
-    TF2 Model Zoo models are exported with signatures rather than as directly
-    callable objects. This patches eta.detectors.tfmodels_detectors to use
-    loaded.signatures['serving_default'] when the loaded model is not callable.
+    Returns:
+        The fixed loader function, or None if TensorFlow is unavailable.
     """
-    global _tf2_patched
-    if _tf2_patched:
-        return
-
     try:
         import os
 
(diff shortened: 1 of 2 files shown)


fiftyone/utils/eta.py
```diff
@@ -100,6 +172,7 @@ def __init__(self, config, _model=None):
         fom.LogitsMixin.__init__(self)
 
     def __enter__(self):
+        _patch_tf2_detection_model()
```
lorenzoriano: Does this need to be unconditional? Can it not be triggered only if needed?

fiftyone/utils/eta.py
```diff
@@ -43,6 +44,77 @@
     etal.VideoSemanticSegmenter,
 )
 
+_tf2_patched = False
+
+
+def _patch_tf2_detection_model():
```
lorenzoriano: Would love to see a unit test for this

lorenzoriano (commented):
Nothing blocking on my side, but It'd be great to see testing and make the change more "surgical"

fiftyone/utils/eta.py
```diff
@@ -43,6 +44,77 @@
     etal.VideoSemanticSegmenter,
 )
 
+_tf2_patched = False
```
manushreegangwar: Have you looked at https://github.com/voxel51/fiftyone/blob/2666fe0688cae658b8f9fb9e69ca297e9adf5a8a/fiftyone/zoo/models/torch.py#L74? I would suggest doing a version check

fiftyone/utils/eta.py
```diff
@@ -43,6 +44,77 @@
...
+
+                if "detection_boxes" in detections:
+                    return (
+                        detections["detection_boxes"],
+                        detections["detection_scores"],
+                        detections["detection_classes"],
+                    )
+                else:
+                    return (
+                        detections["output_3"],
```
manushreegangwar: What is the format of `detections`? Why do you need to return this?

tests/unittests/eta_tests.py
```diff
@@ -0,0 +1,130 @@
...
+            "_load_tf2_detection_model_fixed"
+        )
+
+    @_requires_tensorflow
+    def test_patch_not_reapplied(self):
+        import fiftyone.utils.eta as eta_module
+        from fiftyone.utils.eta import _patch_tf2_detection_model
+
+        _patch_tf2_detection_model()
+        first_fn = None
```
manushreegangwar: nit: remove

fiftyone/utils/eta.py
```diff
@@ -43,6 +44,77 @@
...
+                        detections["output_1"],
+                        detections["output_2"],
+                    )
+
+            return predict
+
+        tfmodels._load_tf2_detection_model = _load_tf2_detection_model_fixed
```
manushreegangwar: This should be temporary, right? You can use https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/utils.py#L2676

fiftyone/utils/eta.py
CharlesCNorton: It seems that the signature returns generic keys instead of detection_boxes/scores/classes, so I mapped them to match. Let me know if you want this handled in a different way.

tests/unittests/eta_tests.py
CharlesCNorton: done!

fiftyone/utils/eta.py
CharlesCNorton: Much cleaner thank you!

CharlesCNorton:
addressed and merged in [#680](https://github.com/voxel51/eta/pull/680)

PR #6663, opened 2025-12-12
[FOTD] Support throwing an error for missing values

kaixi-wang:

- Enable returning an exception for None values in FiftyOneTorchDataset. 
- Include sample id in the message to help with debugging/resolving errors
- Only applies when vectorized = False

- added unit test

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added a validation option for PyTorch dataset conversion that enforces non-null values in required fields. When enabled, the system raises descriptive errors if None values are encountered during data loading operations.

* **Tests**
  * Added comprehensive test coverage for the new validation feature, including scenarios with partial data and combined error-handling behaviors.

### DIFF
--- a/tests/unittests/torch_dataset_tests.py
+++ b/tests/unittests/torch_dataset_tests.py
@@ -179,6 +179,8 @@ class FiftyOneTorchDatasetTests(unittest.TestCase):
                 assert_not_none=True,
                 skip_failures=True,
             )
+            # any sample with none for partial_field should
+            # be returned as a ValueError
             for i, val in enumerate(torch_dataset.__getitems__(indices)):
                 if i % 2 == 0:
                     self.assertFalse(isinstance(val, Exception))


brimoor (changes requested):
@jacobsela don't you think `GetItem` should be responsible for raising the error if it requires a non-None value?

jacobsela:
@brimoor 
For the first point would it alleviate the concern if we threw a new specific error type? If we want None values to be our N/A then we shouldn’t we have some recognition of that?

I agree with the second point. We should fail outright after the values call if vectorize is True and the new arg is set. This would also give us a clean path forward in regards to the decision of skipping missing values in training.

@kaixi-wang 
Alternatively we scrap this PR and then in the fo-ml PR we only do the following things:
1. Expose the FOTD skip failures
2. Add specific error *raising* just to the embeddings get item
3. Add the error handling collate we discussed.

The key thing is that in the second option the PR should be very lean.

I’m fine with either option.

PR #6683, opened 2025-12-17
FOEPD-2481 - Command and keybinding API

marcus-kestler:

The intent is to introduce a global command and keybinding API.  The design is a "context" based scoping.  This allows developers to implement scope local keybinding/commands as well as an inheritable stack based scope system.

The basic unit here is an Action.  An action is a unit of execution.  It can optionally be undoable by implementing the Undoable interface.  Actions are run in a context which maintains the undo/redo stack.  The ActionManager of each CommandCommand context maintains this, and executes actions.

On top of actions are Commands.  Commands are the component level representation of an Action.  The have an enabled/disabled state/function as well as optional labels and descriptions designed to be shared between different invocation points (keyboard, button, menu item , etc).  Commands can return actions that are automatically push on the undo stack.

The final layer is a keybinding.  This simply maps a command to a key sequence like "ctrl+s", or a multi-sequence like "ctrl+d, alt+f". 

I have replaced the keybindings in the Modal (s, x, f, Escape), ModalNavigation (ArrowLeft, Arrow Right), and undo redo globally (CTRL+x, META+x) (CTRL+SHIFT+x, Meta+y, Meta+SHIFT+x).

There are numerous new unit test to cover the core API.  I'll add hook test is a follow up, as this is already too big of a PR..
I manually tested that all replaced keybindings work (there are existing unit test covering some of them).

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Central command/command-bus with scoped contexts, hierarchical undo/redo, key‑binding manager, and new hooks for commands and key bindings.

* **Improvements**
  * Modal, navigation, and editor UIs use the centralized command system for consistent keyboard shortcuts and actions.
  * Undo/redo behavior unified and surfaced via hooks and a context manager.

* **Breaking Change**
  * useLighter no longer exposes undo/redo — migrate to the central undo/redo hooks.

* **Tests**
  * Added comprehensive tests for commands, contexts, key parsing/management, and undo/redo.

* **Documentation**
  * README updated to reflect the useLighter API change.

### DIFF
--- a/app/packages/commands/src/actions/ActionManager.ts
+++ b/app/packages/commands/src/actions/ActionManager.ts
@@ -90,7 +90,7 @@ export class ActionManager {
 
   /**
    * Redoes the last undone undoable action.
-   * @returns True is the redo was successful, false otherwise.
+   * @returns True if the redo was successful, false otherwise.
    */
   redo(): boolean {
     const undoable = this.redoStack.pop();
--- a/app/packages/commands/src/context/CommandContextManager.ts
+++ b/app/packages/commands/src/context/CommandContextManager.ts
@@ -222,7 +222,7 @@ export class CommandContextManager {
    * @param event the key event
    * @returns Nothing
    */
-  public async handleKeyDown(event: KeyboardEvent): Promise<void> {
+  public handleKeyDown(event: KeyboardEvent): void {
     const active = document.activeElement;
 
     // Prevent shortcuts when interacting with any form field
@@ -236,7 +236,7 @@ export class CommandContextManager {
     }
     const match = this.getActiveContext().handleKeyDown(event);
     if (match.full) {
-      await this.getActiveContext().executeCommand(match.full);
+      this.getActiveContext().executeCommand(match.full);
       event.stopPropagation();
       event.preventDefault();
     }
--- a/app/packages/commands/src/actions/ActionManager.test.ts
+++ b/app/packages/commands/src/actions/ActionManager.test.ts
@@ -4,15 +4,15 @@ import { DelegatingUndoable } from "./Undoable";
 
 describe("ActionManager", () => {
   let manager: ActionManager;
-  let execFn: () => Promise<void>;
-  let undoFn: () => Promise<void>;
+  let execFn: () => void;
+  let undoFn: () => void;
   let action: DelegatingUndoable;
   beforeEach(() => {
     manager = new ActionManager();
-    execFn = vi.fn(async () => {
+    execFn = vi.fn(() => {
       return;
     });
-    undoFn = vi.fn(async () => {
+    undoFn = vi.fn(() => {
       return;
     });
     action = new DelegatingUndoable("fo.test", execFn, undoFn);
--- a/app/packages/commands/src/context/context.manager.test.ts
+++ b/app/packages/commands/src/context/context.manager.test.ts
@@ -80,7 +80,7 @@ describe("CommandContextManager", () => {
     CommandContextManager.instance()
       .getActiveContext()
       .bindKey("ctrl+x", cmd.id);
-    await CommandContextManager.instance().handleKeyDown(
+    CommandContextManager.instance().handleKeyDown(
       new KeyboardEvent("keydown", { ctrlKey: true, key: "x" })
     );
     expect(execFn).toBeCalledTimes(1);
@@ -103,7 +103,7 @@ describe("CommandContextManager", () => {
       true
     );
     CommandContextManager.instance().pushContext(context);
-    await CommandContextManager.instance().handleKeyDown(
+    CommandContextManager.instance().handleKeyDown(
       new KeyboardEvent("keydown", { ctrlKey: true, key: "x" })
     );
     expect(execFn).toBeCalledTimes(1);
@@ -126,7 +126,7 @@ describe("CommandContextManager", () => {
       false
     );
     CommandContextManager.instance().pushContext(context);
-    await CommandContextManager.instance().handleKeyDown(
+    CommandContextManager.instance().handleKeyDown(
       new KeyboardEvent("keydown", { ctrlKey: true, key: "x" })
     );
     expect(execFn).toBeCalledTimes(0);


app/packages/commands/src/hooks/useUndoRedo.ts
```diff
@@ -0,0 +1,38 @@
...
+        await getActionManager().redo();
+    }, []);
+    useEffect(() => {
+        getActionManager().subscribe((undoEnabled, redoEnabled, undoCount, redoCount) => {
+            setUndoEnabled(undoEnabled);
+            setRedoEnabled(redoEnabled);
+            setUndoCount(undoCount);
+            setRedoCount(redoCount);
+        });
+    });
```
sashankaryal: This effect ought to return an unsub func for cleanup or else memleak

app/packages/commands/src/actions/ActionManager.ts
```diff
@@ -0,0 +1,159 @@
...
+  public subscribe(listener: ActionManagerListener): () => void {
+    this.listeners.add(listener);
+    return () => {
+      this.listeners.delete(listener);
+    }
+  }
+
+  private fireListeners() {
+    this.listeners.forEach((listener) => {
+      listener(this.canUndo(), this.canRedo(), this.getUndoStackSize(), this.getRedoStackSize()); ß
```
sashankaryal: nit: typo

app/packages/commands/src/keys/manager.ts
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * Manages a set of key bindings in scopes.
+ * Each scope must contain unique bindings.  
+ * Scopes may shadow bindings in other scopes with the 
+ * precedence being User, Plugin, Core.
+ * This allows plugins and user re-bindings.
+ * @see KeyParser for documentation on the binding format.
+ */
+export class KeyManager {
```
sashankaryal: nit: I see that we have a global instance of this and therefore not relevant _except for tests_, but maybe we should add a `cleanup()` equivalent function that removes the event listener.

app/packages/commands/src/hooks/useUndoRedo.ts
sashankaryal: also don't forgot to include deps list or else we'll have a subscription on every render

app/packages/commands/src/keys/KeyManager.ts
```diff
@@ -0,0 +1,139 @@
...
+              break;
+            }
+            //Matches a subset of the sequences. Mark the match so we can add the event to prior matches
+            hasMatch = true;
+          } else {
+            break;
+          }
+        }
+      }
+    }
```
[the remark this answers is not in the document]
marcus-kestler: Intentional so that the map can use a string key to find bindings.

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+    /**
+     * removes all contexts,listeners and creates a new default context.
+     * THIS IS FOR TESTING
+     */
+    public reset(): void {
+        this.defaultContext = new CommandContext(KnownContexts.Default);
+        this.contextStack = [this.defaultContext];
+        this.listeners.clear();
+        this.contexts.clear();
+    }
```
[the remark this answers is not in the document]
marcus-kestler: This is just for testing with the singleton.

app/packages/commands/src/hooks/useCommand.ts
```diff
@@ -0,0 +1,39 @@
+import { useCallback, useEffect, useRef } from "react";
+import { CommandContext } from "../context";
+import { CommandFunction } from "../types";
+
+/**
+ * Hook to create and register a command in a given context.
+ * The command is unregistered on unmount.
+ * @param context An acquired context @see useCommandContext
+ * @param id The id of the command
+ * @param execFn The. function to call when the command is executed
```
marcus-kestler: Fixed

app/packages/commands/src/hooks/useCommand.ts
```diff
@@ -0,0 +1,39 @@
...
+            label, description);
+        return () => {
+                context.unregisterCommand(cmd.id);
+        }
+
+    }, [context, id, exec, enable, label,description]);
+    
+    return useCallback(async ()=>{
+        await context.executeCommand(id);
+    }, [id, context]);
```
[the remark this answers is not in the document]
marcus-kestler: Fixed

app/packages/commands/src/actions/ActionManager.ts
```diff
@@ -0,0 +1,168 @@
...
+ * 2. The action's execute() method is called
+ * 3. The action is pushed onto undoStack
+ */
+
+export type UndoStateListener = (undoEnabled: boolean, redoEnabled: boolean) => void;
+export type ActionListener = (actionId: string, isUndo: boolean) => void;
+
+export class ActionManager {
+  private undoStack: Undoable[] = [];
+  private redoStack: Undoable[] = [];
```
tom-vx51, line 48: [nit] this should be `Action[]` yeah? Since you'd never call `undo()` from the redo stack

app/packages/commands/src/actions/ActionManager.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /**
+   * Undoes the last undoable.
+   * @returns The undone action, or undefined if no actions to undo.
+   */
+  async undo(): Promise<Undoable | undefined> {
+    const undoable = this.undoStack.pop();
+    if (undoable) {
+      await undoable.undo();
```
tom-vx51: what's our strategy for error handling? say this fails; who's responsible?

With this scheme, we could run into some undesirable states:
* `manger.undo()` called
* `undoable` popped from stack
* `undoable.undo()` fails
* => action is now permanently applied - no reference exists in either stack

app/packages/commands/src/context/CommandContext.ts
```diff
@@ -0,0 +1,274 @@
...
+import { Command, CommandFunction } from "../types";
+
+/**
+ * Represents a scoped execution environment consisting of
+ * an undo/redo stack, registered commands, and key bindings.
+ * Scope can be inherited to allow chaining of parent commands,
+ * key binding, etc.
+ */
+export class CommandContext {
+    private readonly actions = new ActionManager();
```
tom-vx51: [nit] would be nice if these were optional constructor params. More flexibility and easier testing via DI

app/packages/commands/src/context/CommandContext.ts
```diff
@@ -0,0 +1,274 @@
...
+        this.unsubscribes.forEach((unsubFn) => {
+            unsubFn();
+        });
+        this.unsubscribes = [];
+    }
+    /**
+     * Adds a listeners to the local ActionManager and all parent
+     * context's ActionManager to be notified of all undo state changes
+     * in the heirarchy.
+     * @param l the listener
```
tom-vx51: can we use more descriptive variable names?

app/packages/commands/src/context/CommandContext.ts
```diff
@@ -0,0 +1,274 @@
...
+     * Updates the local undo state (lastUndo/lastRedo) with the total state
+     * of all available scopes.  If there is a change, listeners are invoked.
+     */
+    private updateUndoState() {
+        const newUndo = this.canUndo();
+        const newRedo = this.canRedo();
+        if (newUndo !== this.lastCanUndo || newRedo !== this.lastCanRedo) {
+            this.lastCanUndo = newUndo;
+            this.lastCanRedo = newRedo;
+            this.fireUndoListeners();
```
tom-vx51: for my understanding, is this only meant to be called when the state changes? i.e. not necessarily when the stack changes?

app/packages/commands/src/context/CommandContext.ts
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    public deactivate() {
+        this.unsubscribes.forEach((unsubFn) => {
+            unsubFn();
+        });
+        this.unsubscribes = [];
+    }
+    /**
+     * Adds a listeners to the local ActionManager and all parent
```
tom-vx51: [typo] `s/listeners/listener`

app/packages/commands/src/context/CommandContext.ts
```diff
@@ -0,0 +1,274 @@
...
+        const newUndo = this.canUndo();
+        const newRedo = this.canRedo();
+        if (newUndo !== this.lastCanUndo || newRedo !== this.lastCanRedo) {
+            this.lastCanUndo = newUndo;
+            this.lastCanRedo = newRedo;
+            this.fireUndoListeners();
+        }
+    }
+    /**
+     * Executes and action and handles pushing it on the
```
tom-vx51: [typo] `s/and/an`

app/packages/commands/src/context/CommandContext.ts
```diff
@@ -0,0 +1,274 @@
...
+        const newRedo = this.canRedo();
+        if (newUndo !== this.lastCanUndo || newRedo !== this.lastCanRedo) {
+            this.lastCanUndo = newUndo;
+            this.lastCanRedo = newRedo;
+            this.fireUndoListeners();
+        }
+    }
+    /**
+     * Executes and action and handles pushing it on the 
+     * undo stack it if is Undoable.
```
tom-vx51: [typo] `s/it if/if it`

app/packages/commands/src/context/CommandContext.ts
```diff
@@ -0,0 +1,274 @@
...
+     */
+    public async undo(): Promise<boolean> {
+        if (this.actions.canUndo()) {
+            await this.actions.undo();
+            return true;
+        }
+        //if we have a parent, and we had no
+        //undo, propogate it up
+        if (this.parent) {
+            return await this.parent.undo();
```
tom-vx51: do we have examples of where this is used? This feels a little dangerous to me; having a scoped command context implies a separation of intent, so running off the end of the stack may be problematic

app/packages/commands/src/actions/action.manager.test.ts
tom-vx51, line 1: [nit] naming convention would be `ActionManager.test.ts` or `ActionManager.spec.ts`

app/packages/commands/src/registry/CommandRegistry.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Executes a registered command
+   * @param id The command id
+   * @returns true if the command was registered and enabled and executed.
+   */
+  public async executeCommand(id: string): Promise<boolean> {
+    const command = this.getCommand(id);
+    if (command && command.isEnabled()) {
+      const result = await command.execute();
+      if (result && "undo" in result && typeof result.undo === "function") {
```
tom-vx51: [nit] using this logic in a couple places, would be good to make it a utility function

app/packages/commands/src/types/command.ts
```diff
@@ -2,70 +2,81 @@
...
- * Constructor type for command classes.
- *
- * @template C - Command type
- */
-export type CommandCtor<C extends Command<any>> = new (...args: any[]) => C;
+  /**
+   * Replaces the current enablement function and update the enablement
+   * @param func 
+   */
+  public setEnablement(func: () => boolean) {
```
tom-vx51, line 60: any specific use for this? Feels odd to change the configuration after the command is registered

app/packages/commands/tsconfig.json
```diff
@@ -8,15 +8,15 @@
...
+        "strict": true,
         "forceConsistentCasingInFileNames": true,
         "module": "ESNext",
         "moduleResolution": "Node",
         "resolveJsonModule": true,
         "isolatedModules": true,
         "noEmit": true,
         "jsx": "react-jsx",
-        "types": ["vite/client", "node"]
+        "types": ["vite/client", "node"],
```
tom-vx51: trailing comma is not valid json

app/packages/commands/vitest.config.ts
```diff
@@ -1,4 +1,4 @@
-import { defineConfig } from "vitest/config";
+import { defineConfig } from "vitest/config.js";
```
tom-vx51, line 1: [nit] shouldn't be necessary

app/packages/core/src/components/Modal/Lighter/LighterSampleRenderer.tsx
tom-vx51, line 1: unrelated change, can omit

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useExit.ts
```diff
@@ -58,10 +59,8 @@ export default function useExit(revertLabel = true) {
     /**
      * 3D SPECIFIC LOGIC ENDS HERE.
      */
-
-    // We are leaving editing mode, clear the stack
-    scene?.clearUndoRedoStack();
-
+    
+    CommandContextManager.instance().clearUndoRedoStack();
```
tom-vx51, line 78: this is global yeah? previous logic was local to a specific Lighter scene

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+    ModalFullScreenToggle = "fo.modal.fullscreen.toggle",
+    ModalClose = "fo.modal.close",
+    ModalNextSample = "fo.modal.next.sample",
+    ModalPreviousSample = "fo.modal.previous.sample"
+}
+//callback for context changes
+export type CommandContextListener = (newId: string) => void;
+
+/**
+ * Manages a stack of contexts as was all created context.
```
tom-vx51, line 31: [typo] `s/as was all/as well as`?

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+        if (document) {
+            document.addEventListener("keydown", this.handleKeyDown.bind(this));
+        }
+        this.defaultContext.registerCommand(KnownCommands.Undo, async ()=>{
+            await this.getActiveContext().undo();
+        }, ()=>{
+            return this.getActiveContext().canUndo();
+        }, 
+        "Undo", "Undoes the previous command.")
+        this.defaultContext.bindKey("ctrl+z", KnownCommands.Undo);
```
tom-vx51: any way to standardize these keys so we can avoid typos? Might make more sense to put these in a dedicated mapping (outside of this class) for easier discoverability/maintenance

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+export class CommandContextManager {
+
+    private defaultContext = new CommandContext(KnownContexts.Default);
+    private contextStack = new Array<CommandContext>();
+    private static _instance: CommandContextManager | undefined;
+    private listeners = new Set<CommandContextListener>();
+    private contexts = new Map<string, CommandContext>();
+
+    constructor() {
+        this.contextStack.push(this.defaultContext);
```
tom-vx51: do we always want a default context?

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+            return this.getActiveContext().canRedo();
+        }, "Redo", "Redoes a previously undone command.");
+        this.defaultContext.bindKey("ctrl+shift+z", KnownCommands.Redo);
+        this.defaultContext.bindKey("meta+y", KnownCommands.Redo);
+        this.defaultContext.bindKey("meta+shift+z", KnownCommands.Redo);
+    }
+    /**
+     * @returns the single instance of this manager
+     */
+    public static instance(): CommandContextManager {
```
tom-vx51: we've been bitten a number of times by singletons. Might make more sense to have a `useCommandContextManager()` hook which provides a shared instance by default

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+    /**
+     * Factory method for CommandContexts for easy inheriting of the current context.
+     * @param id The id of the new context
+     * @param inheritCurrent If true, the current context will serve as a parent, so
+     * any commands not handled in the current context will propagate up to the parent,
+     * it's parent if there is one, and so on.
+     * @returns The new context.  This context is not active until it is pushed @see this.pushExecutionContext
+     */
+    public createCommandContext(id: string, inheritCurrent: boolean): CommandContext {
```
tom-vx51: confirming my understanding, this allows for 
* a single chain of inherited contexts
* an arbitrary number of "solo" contexts

are there times when we'd want to support multiple distinct chains? i.e. a tree instead of a list

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+    /**
+     * Gets a previously created context by id.  @see this.createCommandContext
+     * @param id the context id
+     * @returns The context or undefined if it doesn't exist
+     */
+    public getCommandContext(id: string): CommandContext | undefined {
+        return this.contexts.get(id);
+    }
+
+    public destroyContext(id: string){
```
tom-vx51: [nit] `destroy` often comes with side-effects; clearer if we just call it delete

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+        this.contexts.delete(id);
+    }
+    /**
+     * Get the current command context
+     * @returns the current command context
+     */
+    public getActiveContext(): CommandContext {
+        return this.contextStack[this.contextStack.length - 1];
```
tom-vx51: how do we know that this is the active context? say I've called `createCommandContext` several times with no inheritance

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+            listener(this.contextStack[this.contextStack.length - 1].id);
+        });
+    }
+
+    public async handleKeyDown(event: KeyboardEvent) {
+        const active = document.activeElement;
+
+        // Prevent shortcuts when interacting with any form field
+        if (
+            active?.tagName === "INPUT" ||
```
tom-vx51: is there a better way we can detect this?

app/packages/commands/src/context/CommandContextManager.ts
tom-vx51, line 1: a bit confused about how this class is meant to function. Can you help clarify the intended behavior with `contextStack`, `contexts`, inheritance, and the active context?

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+            const popped = this.contextStack.pop();
+            popped?.deactivate();
+            this.fireListeners();
+        }
+    }
+
+    /**
+     * clears all contexts except the default context
+     */
+    public toDefault(): void {
```
tom-vx51: [nit] maybe rename to reset? and move reset to a clear testing-only name

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+            event.repeat
+        ) {
+            return;
+        }
+        const match = this.getActiveContext().handleKeyDown(event);
+        if (match.full) {
+            await this.getActiveContext().executeCommand(match.full);
+            event.stopPropagation();
+            event.preventDefault();
+        }
```
tom-vx51: should this propagate to the next context if not handled?

app/packages/spotlight/src/events.ts
tom-vx51, line 1: unrelated, can omit

app/packages/lighter/src/interaction/InteractionManager.ts
```diff
@@ -468,7 +465,7 @@ export class InteractionManager {
    * Handles keyboard events for undo/redo shortcuts and shift modifier to maintain aspect ratio.
    * @param event - The keyboard event.
    */
-  private handleKeyDown = (event: KeyboardEvent): void => {
+  private handleKeyDown = async (event: KeyboardEvent): Promise<void> => {
```
tom-vx51, line 468: doesn't return a promise?

app/packages/lighter/src/commands/UpdateLabelCommand.ts
```diff
@@ -25,11 +25,11 @@ export class UpdateLabelCommand implements Command {
     this.nextLabel = nextLabel;
   }
 
-  execute(): void {
+  async execute(): Promise<void> {
```
tom-vx51: these don't return promises

app/packages/lighter/src/commands/TransformOverlayCommand.ts
```diff
@@ -33,11 +33,11 @@ export class TransformOverlayCommand implements Command {
     this.description = `Transform overlay ${overlayId}`;
   }
 
-  execute(): void {
+  async execute(): Promise<void> {
```
tom-vx51: these don't return promises

app/packages/lighter/src/commands/RemoveOverlayCommand.ts
```diff
@@ -18,11 +18,11 @@ export class RemoveOverlayCommand implements Command {
     this.description = `Remove overlay ${overlay.id}`;
   }
 
-  execute(): void {
+  async execute(): Promise<void> {
```
tom-vx51: these don't return promises :) `Undoable` should allow for `() => void | Promise<void>`

app/packages/commands/src/context/CommandContextManager.ts
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * Known contexts.  Add to these if you need
+ * to share a context between components.
+ */
+export enum KnownContexts {
+    Default = "fo.default",
+    Modal = "fo.modal"
+}
+
+export enum KnownCommands {
```
tom-vx51, line 17: i assume we know of more than these, not sure this is the best place to enumerate every possible command

app/packages/commands/src/hooks/useCommand.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param context An acquired context @see useCommandContext
+ * @param id The id of the command
+ * @param execFn The. function to call when the command is executed
+ * @param enablement A function to determine if the command is enabled
+ * @param label The short name of the command, ie Edit, Save, etc
+ * @param description A longer description fit for a tooltip
+ * @returns A function to invoke the command
+ */
+export const useCommand = (context: CommandContext, id: string, execFn: CommandFunction, enablement: () => boolean, label?: string, description?: string) => {
+    const exec = useRef(execFn);
```
tom-vx51: why refs?

app/packages/commands/src/hooks/useCommandContext.ts
```diff
@@ -0,0 +1,53 @@
...
+    deactivate: () => void
+} => {
+    const existed = useRef(false);
+    const boundContext = useMemo(() => {
+        if (typeof context === "string") {
+            const existing = CommandContextManager.instance().getCommandContext(context);
+            if (existing) {
+                existed.current = true;
+                return existing;
+            }
```
tom-vx51: should reset `existed.current`

app/packages/commands/src/hooks/useCommandContext.ts
```diff
@@ -0,0 +1,53 @@
...
+            return CommandContextManager.instance().createCommandContext(context, inheritCurrent ? inheritCurrent : true);
```
tom-vx51: `inheritCurrent ? inheritCurrent : true` === `true`?

app/packages/commands/src/hooks/useCommandContext.ts
```diff
@@ -0,0 +1,53 @@
...
+            }
+        }
+    }, [boundContext]);
+
+    const activate = useCallback(() => {
+        CommandContextManager.instance().pushContext(boundContext);
+    }, [boundContext]);
+
+    const deactivate = useCallback(() => {
+        CommandContextManager.instance().popContext();
```
tom-vx51: is this context guaranteed to be on the top of the stack?

app/packages/commands/src/hooks/useKeyBinding.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Registers a keybinding to a specific command in a context.
+ * @param command The command or command id to bind to the key sequence
+ * @param binding A key sequence specificed like: "ctrl+x", "s", "meta+x, ctrl+d".  @see KeyParser
+ * @param context The context of the binding.  @see CommandContext
+ */
+export const useKeyBinding = (command: string | Command, binding: string, context?: CommandContext) => {
+    const resolvedCtx = useMemo(() => {
+        return context || CommandContextManager.instance().getActiveContext();
```
tom-vx51: [nit] `??` preferred over `||` in most cases

app/packages/commands/src/keys/KeyParser.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+    /**
+     * Returns a string representation that is
+     * normalized to lower case and without 
+     * white space.
+     */
+    public toString(): string {
+        let ret = "";
+        if (this.isCtrl) {
+            ret += "ctrl";
```
tom-vx51: can we make these constants

app/packages/commands/src/keys/KeyParser.ts
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        if (this.isAlt) {
+            if (ret.length > 0) {
```
tom-vx51: [nit] lots of duplication

tom-vx51 (commented):
general comments:
* mostly there, but missing docstrings on some public methods in the classes
* run prettier

app/packages/commands/src/actions/ActionManager.ts
marcus-kestler: Only Undoables can be in either stack.  It goes in to the undo stack first.  Nothing can go directly to the redo stack.  If the redo stack were Actions, we'd have to coerce the type when it goes from redo->undo.

app/packages/commands/src/context/CommandContext.ts
marcus-kestler: Correct, it should only fire when there as an actual change of undo/redo state, not the stack depth.

app/packages/commands/src/hooks/useCommandContext.ts
marcus-kestler: If used in the component hierarchy correctly, yes.  You could shoot yourself in the foot if you do some custom push/pops outside of the hook system.  I will make this safer and only pop if the Id matches and warn if not.

app/packages/commands/src/registry/CommandRegistry.ts
```diff
@@ -0,0 +1,104 @@
...
+   * @param id The command id.  Use the "fo." prefix for our commands.  Plugins may register them as well.
+   * @param execute The lambda to call on execution
+   * @param undo If provided the command will automatically support undo.  This function will be called to 
+   * perform the undo.
+   * @param label A short name like Save, Delete, etc
+   * @param description A longer description such as "Deletes the selected annotation"
+   * @param enablement A function to determine if the command is enabled.  It is invoked when 
+   * registering a command and can be refresh with the @link updateEnabled function.
+   * @returns A command object that can be used locally to execute, enable etc.
+   */
```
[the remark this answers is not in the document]
marcus-kestler: Fixed

app/packages/commands/src/context/CommandContext.ts
marcus-kestler: It is used right now to make undo/redo a global op.  The modal context inherits from the default.

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useExit.ts
marcus-kestler: It is a global call, but the local context lighter registers makes it local to lighter, like it was.

app/packages/commands/src/context/CommandContextManager.ts
marcus-kestler: The parser is very forgiving.  It will allow "CtRL +      s", etc.  If you do put a bad binding in there it will throw.  The KeySequence class normalizes them for display and matching purposes already.

app/packages/commands/src/context/CommandContextManager.ts
marcus-kestler: This is for global keys right now.  Only lighter interact with undo/redo.  I do think we want it.  I'd love to move Escape handling to the default context as well to normalize that behavior.

app/packages/commands/src/context/CommandContextManager.ts
marcus-kestler: Yeah, I know there is some hate around singletons, but in this case I really don't want anyone instantiating this.  Since we have plugins that can access this, there was no other way to stop someone from creating a new one and doing something weird.

app/packages/commands/src/context/CommandContextManager.ts
marcus-kestler: It supports composable chains of contexts.  You can effectively do tree like behavior.  One of the use cases I considered was 2 lighter instances.  So in that case, lighter1 would have a context that inherited from modal and default.  Ligther2 would have the same.  You could then "swap" them on focus change for instance, and the behavior is analogous to a tree.  Even though there is a push/pop semantic for contexts, pushing to the the top does not mean that when you pop, the new active context is the parent of the popped context.  This is one point I considered, and may be confusing, but it seems to lend the most flexibility.

app/packages/commands/src/context/CommandContextManager.ts
marcus-kestler: The activate/deactivate callbacks must be used to control this.

app/packages/commands/src/context/CommandContextManager.ts
marcus-kestler: I just lifted this from many of the existing keydown handlers.  I'm not sure if there is a better way.

app/packages/commands/src/context/CommandContextManager.ts
marcus-kestler: The contextStack is a stack that is meant to handle the component hierarchy view of what context are available.  A child should now need to know what contexts are above in it some cases, and just wants to push a new context on, and pop when it is done.  Think of the Modal vs the global app context (default).  Now consider Lighter as another child (not yet implemented).  Lighter wants it's own set of undo/redo and commands specific to it.  So for each annotation edit, we could push an undo context and when the edit is done, pop it for instance, thus isolating the actions on a particular annotation cleanly from other edits.
"contexts" is used as a registry of contexts.  This is so that you can reference a created context that is not yours, or active by name.  This is needed for child components to be able to register commands in to a parent context that may not yet be active.
Inheritance is needed to allow child components to extend the commands, isolate the undo stack, introduce new key bindings in a scoped manner that allows for adding to but still allowing parent commands/bindings/undos.
The active context is really just there to allow for global execution and undo.  That's why there are the command bound in to the default for undo/redo.  As I mention I want Escape handling to go there as well.

app/packages/commands/vitest.config.ts
marcus-kestler: I believe eslint yelled about it.  Some automation wanted it.

app/packages/commands/src/context/CommandContextManager.ts
marcus-kestler: CommandContext::executeCommand handles the inheritance.  The event propagation doesn't effect that.

app/packages/commands/src/hooks/useCommand.ts
marcus-kestler: It is the "latest ref" pattern.  Avoids re-render when the function changes:
https://www.epicreact.dev/the-latest-ref-pattern-in-react

app/packages/commands/src/context/CommandContext.ts
marcus-kestler: id is the only thing required.

app/packages/commands/src/actions/ActionManager.ts
marcus-kestler: I don't think you want to keep it on the stack.  That would mean you have a bad action lodged in there that will continue to fail (presumably).
Perhaps we should discuss a strategy.

app/packages/commands/src/context/CommandContextManager.ts
marcus-kestler: Fixed.

app/packages/lighter/src/interaction/InteractionManager.ts
marcus-kestler: Fixed

app/packages/core/src/components/Modal/Modal.tsx
```diff
@@ -121,71 +135,103 @@ const Modal = () => {
...
-          if (hoveringSampleId) {
-            select(hoveringSampleId);
-          } else {
-            const modalSampleId = await snapshot.getPromise(fos.modalSampleId);
-            if (modalSampleId) {
-              select(modalSampleId);
+      async () => {
+        const current = await snapshot.getPromise(fos.modalSelector);
+        set(fos.selectedSamples, (selected) => {
+          const newSelected = new Set([...Array.from(selected)]);
```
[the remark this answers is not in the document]
marcus-kestler: Not my change.

app/packages/commands/src/utils.ts
```diff
@@ -0,0 +1,3 @@
+export function isUndoable(object: object) {
+  return "undo" in object && typeof object.undo === "function";
+}
```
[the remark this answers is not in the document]
marcus-kestler: No.

app/packages/commands/src/context/CommandContext.ts
marcus-kestler: Fixed

app/packages/commands/src/context/CommandContext.ts
marcus-kestler: Fixed

app/packages/commands/src/context/CommandContext.ts
marcus-kestler: Fixed

app/packages/commands/src/actions/action.manager.test.ts
marcus-kestler: Fixed

app/packages/commands/src/registry/CommandRegistry.ts
marcus-kestler: Added a utils.ts with a helper.

app/packages/commands/tsconfig.json
marcus-kestler: Fixed.

app/packages/core/src/components/Modal/Lighter/LighterSampleRenderer.tsx
marcus-kestler: Fixed

app/packages/commands/src/context/CommandContextManager.ts
marcus-kestler: Fixed

app/packages/spotlight/src/events.ts
marcus-kestler: Fixed.

app/packages/lighter/src/commands/TransformOverlayCommand.ts
marcus-kestler: Fixed

app/packages/commands/src/hooks/useCommandContext.ts
marcus-kestler: I'm not sure I understand.

app/packages/commands/src/hooks/useCommandContext.ts
marcus-kestler: Doh!  Fixed.

app/packages/commands/src/hooks/useKeyBinding.ts
marcus-kestler: Fixed.

app/packages/lighter/src/commands/MoveOverlayCommand.ts
```diff
@@ -42,25 +42,25 @@ export interface Movable {
...
-  execute(): void {
+  async execute(): Promise<void> {
     this.overlay.setBounds(this.newBounds);
   }
 
-  undo(): void {
+  async undo(): Promise<void> {
     this.overlay.setBounds(this.oldBounds);
   }
 }
```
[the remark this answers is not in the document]
marcus-kestler: This was existing code.

app/packages/commands/src/hooks/useCommandContext.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+    return CommandContextManager.instance().getActiveContext();
+  }, [context, inheritCurrent]);
+
+  useEffect(() => {
+    return () => {
+      if (!existed.current) {
+        CommandContextManager.instance().deleteContext(boundContext.id);
+      }
+    };
```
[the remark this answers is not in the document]
sashankaryal: yeah hope we don't destroy other contexts during cleanup

sashankaryal (commented):
I can't help but think it's a bit verbose and we can do some work to simplify the API. There's too much ceremony around context management...

Looking at the usage in Modal.tsx, the user has to:
1. Get context via `useCommandContext`
2. Call `activate()` in a useEffect and return `deactivate` for cleanup
3. Use two hooks for every binding (`useCommand + useKeyBinding`) and pass context to both

Can we add a thin layer that handles this ceremony? Something like:

```javascript
useKeybindings(KnownContexts.Modal, [
  { key: "x", id: KnownCommands.ModalSelect, handler: selectCallback, label: "Select", description: "Select Sample" },
  { key: "s", id: KnownCommands.ModalSidebarToggle, handler: sidebarFn, label: "Sidebar", description: "Show/Hide the sidebar" },
  { key: "f", id: KnownCommands.ModalFullScreenToggle, handler: fullscreenFn, label: "Fullscreen", description: "Enter/Exit full screen" },
  { key: "Escape", id: KnownCommands.ModalClose, handler: closeFn, label: "Close", description: "Close the window" },
]);
```

This would internally handle context acquisition, activation/deactivation, and both command + keybinding registration.

Separately - if not for scoped undo / redo, I would have strongly advocated for removing the context abstraction altogether because the "context stack" naturally emerges from mount order. Would you agree?

--
🎖️ (had fun reviewing this)

app/packages/commands/src/keys/KeyParser.ts
```diff
@@ -0,0 +1,218 @@
...
+            );
+          }
+          switch (keyLow) {
+            case "space":
+              keySequence.key = " ";
+              break;
+            //unescape \+ (= is the lower case)
+            case "\\+":
+              keySequence.key = "=";
+              break;
```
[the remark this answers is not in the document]
marcus-kestler: No, it is the lowercase representation.

app/packages/commands/src/hooks/useCommandContext.ts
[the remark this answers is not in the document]
marcus-kestler: I have fixed this.

app/packages/commands/src/actions/ActionManager.ts
```diff
@@ -0,0 +1,202 @@
...
+        this.fireActionListeners(undoable.id, true);
+      } catch (error) {
+        console.error(`An exception ocurred during undo execution for undoable ${undoable.id}`);
+        console.error(error);
+      }
+      this.fireUndoListeners();
+      return undoable;
+    }
+    return undefined;
+  }
```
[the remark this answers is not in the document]
marcus-kestler: Fixed

app/packages/commands/src/actions/ActionManager.ts
```diff
@@ -0,0 +1,202 @@
...
+        this.fireActionListeners(undoable.id, false);
+      } catch (error) {
+        console.error(`An exception occurred during redo execution for undoable ${undoable.id}`);
+        console.error(error);
+      }
+      this.fireUndoListeners();
+      return undoable;
+    }
+    return undefined;
+  }
```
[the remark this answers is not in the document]
marcus-kestler: Fixed

app/packages/core/src/components/Modal/ModalNavigation.tsx
```diff
@@ -161,6 +141,23 @@ const ModalNavigation = ({ closePanels }: { closePanels: () => void }) => {
...
+      description: "Previous Sample"
+    },
+    {
+      commandId: KnownCommands.ModalNextSample,
+      sequence: "ArrowRight",
+      handler: next,
+      label: "Next",
+      description: "Next Sample"
+    },
+  ]);
```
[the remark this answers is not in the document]
marcus-kestler: Fixed

app/packages/commands/src/actions/ActionManager.ts
```diff
@@ -0,0 +1,205 @@
...
+        this.fireUndoListeners();
+        return false;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Redoes the last undone undoable action.
+   * @returns True is the redo was successful, false otherwise.
```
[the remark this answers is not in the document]
marcus-kestler: Fixed

PR #6693, opened 2025-12-19
Validate frame numbers passed to `set_values()`

brimoor:
## Release notes

- all frame numbers passed to `set_values()` are now validated for type/value safety

## Tested by

- unit tests added
- example below

```py
import fiftyone as fo
import fiftyone.zoo as foz

# Frame numbers must be ints, not strings!
naughty_values = {
    dataset.first().filepath: {
        "1": fo.Detections(),
        "2": fo.Detections(),
        "63": fo.Detections()
    },
    dataset[1:].first().filepath: {
        "2": fo.Detections(),
        "63": fo.Detections(),
        "10000": fo.Detections(),
    }
}

# Previously this would cause corruption, but now correctly raises an error
dataset.set_values("frames.new_field", naughty_values, key_field="filepath")
# FrameError: Frame numbers must be integers; found <class 'str'>

# No corruption anyone!
assert all(isinstance(fn, int) for fn in dataset.values("frames[].frame_number"))
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for frame numbers to prevent invalid entries from being inserted during frame value parsing.
  * Updated exception handling for frame-related errors to provide more appropriate error classification.

* **Tests**
  * Added unit test to verify frame number validation and error handling behavior.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -34,6 +34,7 @@ import fiftyone.core.evaluation as foev
 import fiftyone.core.expressions as foe
 from fiftyone.core.expressions import ViewField as F
 import fiftyone.core.fields as fof
+import fiftyone.core.frame_utils as fofu
 import fiftyone.core.groups as fog
 import fiftyone.core.labels as fol
 import fiftyone.core.map as focm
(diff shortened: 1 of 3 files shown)


exupero (approved):
Testing notes:
- ✅ Test passes
- ✅ Without added validation test fails

I don't see any direct tests of `validate_frame_number`, so this PR adds an implicit test of the integral check in that function 👍

PR #6290, opened 2025-08-23
FOEPD-1965 function to better estimate number of useable CPUs

swheaton:
`os.cpu_count()` gives the number of CPUs available to the whole node.
In kubernetes e.g., this might be much different from the CPU limit that the container is actually operating under.

## What changes are proposed in this pull request?

Try in order:
    1. cgroup v2 CPU quota (``/sys/fs/cgroup/cpu.max``)
    2. cgroup v1 CPU quota (``/sys/fs/cgroup/cpu/cpu.cfs_quota_us``)
    3. cgroup v2 cpuset (``/sys/fs/cgroup/cpuset.cpus.effective``)
    4. cgroup v1 cpuset (``/sys/fs/cgroup/cpuset/cpuset.cpus``)
    5. ``os.process_cpu_count()`` (Python 3.13+)
    6. ``os.sched_getaffinity(0)`` (Linux)
    7. ``multiprocessing.cpu_count()``

### Docker manual tests
```bash
IMAGE="us-central1-docker.pkg.dev/computer-vision-team/dev-docker/fiftyone-app-internal:v2.14.1"

echo "Testing get_cpu_count() with real fiftyone code..."
echo "Using image: $IMAGE"
echo "=========================================="
echo ""

echo "Test 1: --cpus=2 (should return 2)"
docker run --rm --cpus=2 -v $(pwd):/code -e PYTHONPATH=/code $IMAGE python -c "
from fiftyone.core.utils import get_cpu_count
result = get_cpu_count()
print(f'  Result: {result}')
assert result == 2, f'FAIL: Expected 2, got {result}'
print('  ✓ PASS')
"

echo ""
echo "Test 2: --cpus=0.5 (fractional, should return 1)"
docker run --rm --cpus=0.5 -v $(pwd):/code -e PYTHONPATH=/code $IMAGE python -c "
from fiftyone.core.utils import get_cpu_count
result = get_cpu_count()
print(f'  Result: {result}')
assert result == 1, f'FAIL: Expected 1, got {result}'
print('  ✓ PASS')
"

echo ""
echo "Test 3: --cpuset-cpus=0,1 (should return 2)"
docker run --rm --cpuset-cpus=0,1 -v $(pwd):/code -e PYTHONPATH=/code $IMAGE python -c "
from fiftyone.core.utils import get_cpu_count
result = get_cpu_count()
print(f'  Result: {result}')
assert result == 2, f'FAIL: Expected 2, got {result}'
print('  ✓ PASS')
"

echo ""
[...]

PR #6697, opened 2025-12-24
Better defaults for model inference

kaixi-wang:

Improve inference performance for torch models by improving defaults. 
- Add/expose a new env var for model inference related batch_size `FIFTYONE_DEFAULT_MODEL_BATCH_SIZE`
- to differentiate from default_batch_size used by the sdk for database operations
- Expose `prefetch_factor` via env var `FIFTYONE_DEFAULT_DATALOADER_PREFETCH_FACTOR` and as an arg to pass to torch dataloaders
- cap number of recommended default workers to 4

Ran benchmarks for several models with apply_model, setting batch_size and num_workers on GPU and high CPU compute:

### With A100 GPU:
image
image
image

### With compute-optimized 48 cpu:

image
image

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Two new config options: default model batch size and default dataloader prefetch factor.
  * Prefetch control parameter added to model inference and embedding workflows.

* **Improvements**
  * More robust CPU detection across environments, used for worker defaults.
  * Smarter default worker recommendation with a capped maximum for stability.

* **Tests**
  * Unit tests updated to use the new CPU-count helper and expanded CPU-count coverage.

### DIFF
--- a/fiftyone/core/config.py
+++ b/fiftyone/core/config.py
@@ -160,6 +160,12 @@ class FiftyOneConfig(EnvConfig):
             d,
             "default_model_batch_size",
             env_var="FIFTYONE_DEFAULT_MODEL_BATCH_SIZE",
+            default=1,
+        )
+        self.default_dataloader_prefetch_factor = self.parse_int(
+            d,
+            "default_dataloader_prefetch_factor",
+            env_var="FIFTYONE_DEFAULT_DATALOADER_PREFETCH_FACTOR",
             default=16,
         )
         self.default_batch_size = self.parse_int(
--- a/fiftyone/core/config.py
+++ b/fiftyone/core/config.py
@@ -156,6 +156,12 @@ class FiftyOneConfig(EnvConfig):
             env_var="FIFTYONE_DEFAULT_ML_BACKEND",
             default=None,
         )
+        self.default_model_batch_size = self.parse_int(
+            d,
+            "default_model_batch_size",
+            env_var="FIFTYONE_DEFAULT_MODEL_BATCH_SIZE",
+            default=8,  # conservative minimum
+        )
         self.default_batch_size = self.parse_int(
             d,
             "default_batch_size",
--- a/fiftyone/core/config.py
+++ b/fiftyone/core/config.py
@@ -162,17 +162,11 @@ class FiftyOneConfig(EnvConfig):
             env_var="FIFTYONE_DEFAULT_MODEL_BATCH_SIZE",
             default=1,
         )
-        self.default_dataloader_prefetch_factor = self.parse_int(
-            d,
-            "default_dataloader_prefetch_factor",
-            env_var="FIFTYONE_DEFAULT_DATALOADER_PREFETCH_FACTOR",
-            default=16,
-        )
         self.default_batch_size = self.parse_int(
             d,
             "default_batch_size",
             env_var="FIFTYONE_DEFAULT_BATCH_SIZE",
-            default=None,
+            default=16,
         )
         self.default_batcher = self.parse_string(
             d,
--- a/fiftyone/core/config.py
+++ b/fiftyone/core/config.py
@@ -160,7 +160,7 @@ class FiftyOneConfig(EnvConfig):
             d,
             "default_model_batch_size",
             env_var="FIFTYONE_DEFAULT_MODEL_BATCH_SIZE",
-            default=8,  # conservative minimum
+            default=16,
         )
         self.default_batch_size = self.parse_int(
             d,
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -9,7 +9,6 @@ FiftyOne models.
 import contextlib
 import inspect
 import logging
-import os
 
 import numpy as np
 
@@ -848,6 +847,8 @@ def _make_data_loader(
     skip_failures,
     field_mapping,
 ):
+    import torch
+
     if batch_size is None:
         raise ValueError("batch_size cannot be None")
     # This function supports DataLoaders that emit numpy arrays that can
@@ -882,6 +883,10 @@ def _make_data_loader(
         worker_init_fn = None
 
     pin_memory = isinstance(model, fout.TorchImageModel) and model._using_gpu
+
+    # Avoid CPU oversubscription:
+    # https://docs.pytorch.org/docs/stable/notes/multiprocessing.html#avoid-cpu-oversubscription
+    torch.set_num_threads(fou.get_cpu_count() / num_workers)
 
     return tud.DataLoader(
         dataset,
--- a/fiftyone/core/models.py
+++ b/fiftyone/core/models.py
@@ -9,6 +9,7 @@ FiftyOne models.
 import contextlib
 import inspect
 import logging
+import os
 
 import numpy as np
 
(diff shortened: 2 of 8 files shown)


fiftyone/core/utils.py
```diff
@@ -2870,6 +2870,39 @@ def get_multiprocessing_context():
     return multiprocessing.get_context()
 
 
+def get_cpu_count():
```
brimoor, line 2873: @swheaton is proposing to add a `get_cpu_count()` utility in https://github.com/voxel51/fiftyone/pull/6290, so these two implementations should be reconciled and we should ship a single thing that does whatever combo of things we consider best practice.

fiftyone/utils/torch.py
```diff
@@ -27,12 +25,11 @@
 import fiftyone.core.config as foc
 import fiftyone.core.labels as fol
 import fiftyone.core.models as fom
-import fiftyone.core.media as fomd
 import fiftyone.core.odm as foo
 import fiftyone.core.utils as fou
 import fiftyone.utils.image as foui
-import fiftyone.core.collections as focol
 import fiftyone.core.view as fov
+from fiftyone.core.utils import get_cpu_count
```
brimoor: nit: this is already available as `fou.get_cpu_count()` due to `fou` import above

fiftyone/utils/torch.py
```diff
@@ -1811,9 +1808,9 @@ def recommend_num_workers(num_workers=None):
         # https://github.com/voxel51/fiftyone/issues/1531
         # https://stackoverflow.com/q/20222534
         return 0
-
     try:
-        default = multiprocessing.cpu_count() // 2
+        # >4 workers adds significant memory overhead without meaningful speedup
```
brimoor, line 1906: cc @jacobsela; agree with this assessment?

fiftyone/utils/torch.py
```diff
@@ -2034,7 +2031,6 @@ def _prepare_batch_db(self, indices):
                     break
 
             batch.append(d)
-
```
brimoor, line 2037: nit: undo trivial whitespace change

fiftyone/core/models.py
```diff
@@ -1994,18 +2003,20 @@ def _patch_collate_fn(batch):
     return batch[0]  # return patches directly
 
 
-def _parse_batch_size(batch_size, model, use_data_loader):
+def _parse_batch_size_and_num_workers(*, batch_size, model, num_workers):
```
brimoor: The reason the existing implementation handles `batch_size` default separately from `num_workers` is that `apply_model()` supports non-Torch models and `batch_size` is meaningful for all models, whereas `num_workers` is only relevant for Torch models that use data loaders.

With this change, running inference with a Tensorflow model will raise a "this method requires 'torch' to be installed" error:

```py
import fiftyone as fo
import fiftyone.zoo as foz

model = foz.load_zoo_model("efficientdet-d6-1280-coco-tf2")

# Raises "torch" not installed error
dataset.apply_model(model, label_field="preds")
```

because `_parse_batch_size_and_num_workers()` is executed and now imports `fiftyone.utils.torch`.

Unless/until `torch` is a mandatory requirement of `fiftyone`, I don't think that should happen.

fiftyone/core/models.py
```diff
@@ -1994,18 +2003,20 @@ def _patch_collate_fn(batch):
...
+    gpu_available = getattr(model, "_using_gpu", False)
+    num_workers = fout.recommend_num_workers(
+        num_workers, gpu_available=gpu_available
```
brimoor: I don't see `gpu_available` implemented on `fout.recommend_num_workers()`

fiftyone/core/config.py
```diff
@@ -156,6 +156,12 @@ def __init__(self, d=None):
             env_var="FIFTYONE_DEFAULT_ML_BACKEND",
             default=None,
         )
+        self.default_model_batch_size = self.parse_int(
+            d,
+            "default_model_batch_size",
+            env_var="FIFTYONE_DEFAULT_MODEL_BATCH_SIZE",
+            default=16,
```
brimoor: Is it really safe to assume that batches of 16 will work for all models/machines?

The reason I used `batch_size=1` as the default for this *global* setting is that `apply_model()` is a generic interface that supports non-image models, while `batch_size` is typically only relevant for image models.

One consequence of this change is that all models with `model.ragged_batches=True` would start to issue "this model does not support batching" warnings when invoked via the default `apply_model()` syntax (because such models, including all video models, don't support batching), but, as a general rule, I don't think users should get warnings about behaviors they didn't explicitly request.

I do think we should find a way to use larger batch sizes by default for applicable models. But my thinking is that the optimal batch size is heavily model-dependent, and thus any intelligent defaulting needs to be in the **model's config**. That's why [in this doc](https://docs.google.com/document/u/2/d/19VRbePJPHg_YICUwvAH1C1l_aC94Pznrxivjc_w4UAg/edit?tab=t.0) I proposed that models that support batching should specify a `recommended_batch_size` in their zoo manifest entry. `apply_model()` would then be tweaked to use that setting as the default batch size.

Meanwhile, the only sensible "global" default is `batch_size=1`.

brimoor (changes requested):

fiftyone/core/utils.py
kaixi-wang: Oh thanks for calling that out. Didn't know Stuart had another implementation.

fiftyone/core/config.py
kaixi-wang: Yeah I agree. Ideally models or families of models should have some way of influencing batch size as the data does show optimal setting is model dependent. 

From my testing, batch size 16 had marginal increase in memory.  The more detrimental factor was the uncapped num workers, especially with batch size 1.

fiftyone/utils/torch.py
jacobsela: The statement is not generally correct but it's fine as a default value when the user passes no input in.

jacobsela (changes requested):
This PR is doing a lot at once, some good some I don't fully understand. If we limit the PR to the good we can merge quickly and handle the rest in other PRs IMO.

Good:
- introducing `prefects_factor` arg. This is great. Note that the prefetch factor increases memory usage linearly. (It should still be technically better than more workers because we avoid the overhead of new processes).

Can go either way:
- cap default worker usage to 4 - generally fine, if people want more they can chuck in whatever value they want. I don't agree with the reasoning given though, there are envs / models / circumstances that benefit from higher num_workers.

Needs explanation:
- global default batch size > 1 - I agree with Brian's comments.
- changing default thread count - presumably threaded operations are I/IO heavy. We don't necessarily want to cap the number of threads to be the number of CPU cores because we may want to have > num cpu cores pending operations at once.

kaixi-wang:
> global default batch size > 1 - I agree with Brian's comments.

In my latest, I updated it to 1 as the default. The goal is just to have a separate env var for it because if someone sets the DEFAULT_BATCH_SIZE to 1, then that affects mongo operations and we should definitely not do that

> changing default thread count - presumably threaded operations are I/IO heavy. We don't necessarily want to cap the number of threads to be the number of CPU cores because we may want to have > num cpu cores pending operations at once.

Removed it. I was just going by the docs since they sound reasonable (https://docs.pytorch.org/docs/stable/notes/multiprocessing.html#avoid-cpu-oversubscription). However,  I didn't really observe any difference with setting it, and I believe the OOM issues we've seen with docker deployments is the inaccurate cpu_count() that is addressed with get_cpu_counts()

PR #6700, opened 2025-12-29
FOEPD-2765: Fiftyone Camera Data Model

sashankaryal:

- Add a first-class camera module with intrinsics/extrinsics data models, dataset-level refs, and projection utilities for 3D <-> 2D workflows (`fiftyone/core/camera.py`).
- Persist camera intrinsics and sensor extrinsics on datasets, including validation, resolution helpers for grouped datasets, and transform chaining (`Dataset.camera_intrinsics`, `Dataset.sensor_extrinsics`, `resolve_intrinsics`, `resolve_extrinsics`, `get_transform_chain`).
- Add extensive unit coverage.

### Example (`quickstart-groups`)
```python
  import fiftyone as fo
  import fiftyone.zoo as foz
  import numpy as np
  from scipy.spatial.transform import Rotation

  # =============================================================================
  # COORDINATE FRAME CONVENTIONS
  # =============================================================================
  # Ego frame (automotive standard): x-forward, y-left, z-up
  # Camera frame (OpenCV standard):  z-forward, x-right, y-down
  #
  # The rotation from camera frame to ego frame maps:
  #   camera z (forward) -> ego x (forward)
  #   camera x (right)   -> ego -y (left is negative y)
  #   camera y (down)    -> ego -z (down is negative z)
  # =============================================================================

  # Compute the camera-to-ego rotation quaternion
  R_cam_to_ego = np.array([
      [0, 0, 1],    # ego_x = camera_z
      [-1, 0, 0],   # ego_y = -camera_x  
      [0, -1, 0]    # ego_z = -camera_y
  ])
  quat_cam_to_ego = Rotation.from_matrix(R_cam_to_ego).as_quat().tolist()
  # Result: [0.5, -0.5, 0.5, -0.5]


  # 1. Load dataset from the zoo

  # 2. Define Static Intrinsics (Dataset Level)
  #    Standard 1080p camera parameters
  dataset.camera_intrinsics = {
      "left": fo.PinholeCameraIntrinsics(fx=1100, fy=1100, cx=960, cy=540),
      "right": fo.PinholeCameraIntrinsics(fx=1100, fy=1100, cx=960, cy=540),
  }

  # 3. Define Static Calibration (Dataset Level)
  #    Camera mounting positions relative to the vehicle center ("ego")
  dataset.sensor_extrinsics = {
      # Left camera: 1.5m forward, 0.5m left, 1.2m up
      "left::ego": fo.SensorExtrinsics(
          translation=[1.5, 0.5, 1.2],
          quaternion=quat_cam_to_ego,  # Rotation from camera frame to ego frame
          source_frame="left",
          target_frame="ego",
      ),
      # Right camera: 1.5m forward, 0.5m right, 1.2m up
      "right::ego": fo.SensorExtrinsics(
          translation=[1.5, -0.5, 1.2],
          quaternion=quat_cam_to_ego,
          source_frame="right",
          target_frame="ego",
      ),
      # Lidar: centered, 2.0m up (roof-mounted, z-up aligned with ego)
      "lidar::ego": fo.SensorExtrinsics(
          translation=[0.0, 0.0, 2.0],
          quaternion=[0, 0, 0, 1],  # Lidar z-up = ego z-up, no rotation needed
          source_frame="lidar",
          target_frame="ego",
      )
  }
  dataset.save()


  # Focus on the left camera
  dataset.group_slice = "left"

  # 4. Define Dynamic Ego Pose (Sample Level)
  #    Where the vehicle is in the world for this specific frame
  sample = dataset.first()

  # Vehicle is at world position (100, 50, 0), facing +x direction
  ego_pose = fo.SensorExtrinsics(
      translation=[100.0, 50.0, 0.0],
      quaternion=[0, 0, 0, 1],  # Ego x-axis aligned with world x-axis
      source_frame="ego",
      target_frame="world",
  )
  sample["ego_pose"] = ego_pose
  sample.save()

  # 5. Resolve the full transform chain: left camera -> ego -> world
  #    This composes: left::ego with ego::world
  extrinsics = dataset.resolve_extrinsics(
      sample,
      source_frame="left",
      target_frame="world",
      chain_via=["ego"]
  )

  # 6. Resolve intrinsics (infers "left" from dataset.group_slice)
  intrinsics = dataset.resolve_intrinsics(sample)

  # 7. Create projector and project a world point
  #    Stop sign at world coordinates (111.5, 50.5, 2.5)
  #    - 10m in front of camera (x direction)
  #    - centered laterally (same y as camera)  
  #    - 2.5m high (above camera which is at z=1.2)
  projector = fo.CameraProjector(intrinsics, extrinsics)
  world_point = [111.5, 50.5, 2.5]
  pixel = projector.project([world_point], in_camera_frame=False)

  print(f"World Point: {world_point}")
  print(f"Projects to pixel: {pixel[0].astype(int)}")  # [960, 397]
```

Design doc in ticket (INTERNAL ONLY): https://voxel51.atlassian.net/browse/FOEPD-2765

## How is this patch tested?
- `tests/unittests/camera_tests.py`
- `tests/unittests/dataset_tests.py`

## Summary by CodeRabbit

* **New Features**
  * Public camera geometry API: intrinsics, extrinsics, reference types, and a projector for projection/unprojection.
  * Dataset/collection/view properties to store, validate, persist and serialize per-sensor intrinsics/extrinsics; import/export now support them.

* **Behavior**
  * Projection/unprojection supports OpenCV and fisheye distortion, axis-convention handling, depth unprojection, and transform chaining/resolution.

* **Tests**
  * Extensive unit tests covering models, projector, dataset integration, resolution, persistence and edge cases.

### DIFF
--- a/fiftyone/core/camera.py
+++ b/fiftyone/core/camera.py
@@ -11,6 +11,7 @@ multi-sensor fusion workflows.
 """
 
 import warnings
+from abc import ABC, abstractmethod
 from typing import Optional, Union
 
 import cv2
@@ -30,6 +31,144 @@ SUPPORTED_CAMERA_CONVENTIONS = (
     CAMERA_CONVENTION_OPENCV,
     CAMERA_CONVENTION_OPENGL,
 )
+
+
+class ProjectionModel(ABC):
+    """Abstract base class for camera projection models.
+
+    Encapsulates projection and undistortion operations for different camera
+    models.
+    """
+
+    @abstractmethod
+    def project(
+        self,
+        points: np.ndarray,
+        K: np.ndarray,
+        distortion: Optional[np.ndarray],
+    ) -> np.ndarray:
+        """Project 3D points to 2D image coordinates.
+
+        Args:
+            points: (N, 3) array of 3D points in camera frame
+            K: (3, 3) intrinsic matrix
+            distortion: distortion coefficients, or None
+
+        Returns:
+            (N, 2) array of 2D pixel coordinates
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def undistort(
+        self,
+        points: np.ndarray,
+        K: np.ndarray,
+        distortion: Optional[np.ndarray],
+    ) -> np.ndarray:
+        """Undistort 2D image points, returning normalized coordinates.
+
+        Args:
+            points: (N, 2) array of 2D pixel coordinates
+            K: (3, 3) intrinsic matrix
+            distortion: distortion coefficients, or None
+
+        Returns:
+            (N, 2) array of normalized coordinates (z=1 plane in camera frame)
+        """
+        raise NotImplementedError
+
+
+class OpenCVProjectionModel(ProjectionModel):
+    """Standard OpenCV camera model with radial and tangential distortion."""
+
+    def project(
+        self,
+        points: np.ndarray,
+        K: np.ndarray,
+        distortion: Optional[np.ndarray],
+    ) -> np.ndarray:
+        """Project points using OpenCV's standard camera model."""
+        rvec = np.zeros(3, dtype=np.float64)
+        tvec = np.zeros(3, dtype=np.float64)
+
+        if distortion is None:
+            distortion = np.zeros(5, dtype=np.float64)
+
+        points_2d, _ = cv2.projectPoints(
+            points.reshape(-1, 1, 3),
+            rvec,
+            tvec,
+            K,
+            distortion,
+        )
+        return points_2d.reshape(-1, 2)
+
+    def undistort(
+        self,
+        points: np.ndarray,
+        K: np.ndarray,
+        distortion: Optional[np.ndarray],
+    ) -> np.ndarray:
+        """Undistort points using OpenCV's standard camera model.
+
+        Returns normalized (x, y) coordinates (z=1 plane in camera frame).
+        """
+        if distortion is None:
+            distortion = np.zeros(5, dtype=np.float64)
+
+        undistorted = cv2.undistortPoints(
+            points.reshape(-1, 1, 2),
+            K,
+            distortion,
+        )
+        return undistorted.reshape(-1, 2)
+
+
+class FisheyeProjectionModel(ProjectionModel):
+    """OpenCV fisheye camera model with equidistant projection."""
+
+    def project(
+        self,
+        points: np.ndarray,
+        K: np.ndarray,
+        distortion: Optional[np.ndarray],
+    ) -> np.ndarray:
+        """Project points using OpenCV's fisheye camera model."""
+        rvec = np.zeros(3, dtype=np.float64)
+        tvec = np.zeros(3, dtype=np.float64)
+
+        if distortion is None:
+            distortion = np.zeros(4, dtype=np.float64)
+
+        points_2d, _ = cv2.fisheye.projectPoints(
+            points.reshape(1, -1, 3),
+            rvec,
+            tvec,
+            K,
+            distortion,
+        )
+        return points_2d.reshape(-1, 2)
+
+    def undistort(
+        self,
+        points: np.ndarray,
+        K: np.ndarray,
+        distortion: Optional[np.ndarray],
+    ) -> np.ndarray:
+        """Undistort points using OpenCV's fisheye camera model.
+
+        Returns normalized (x, y) coordinates (z=1 plane in camera frame).
+        """
+        if distortion is None:
+            distortion = np.zeros(4, dtype=np.float64)
+
+        undistorted = cv2.fisheye.undistortPoints(
+            points.reshape(-1, 1, 2),
+            K,
+            distortion,
+        )
+        return undistorted.reshape(-1, 2)
 
 
 class CameraIntrinsics(DynamicEmbeddedDocument):
@@ -122,6 +261,14 @@ class CameraIntrinsics(DynamicEmbeddedDocument):
             a numpy array of distortion coefficients, or None if no distortion
         """
         return None
+
+    def get_projection_model(self) -> ProjectionModel:
+        """Returns the ProjectionModel instance for this camera model.
+
+        Returns:
+            a :class:`ProjectionModel` instance
+        """
+        return OpenCVProjectionModel()
 
     def camera_matrix_3x4(
         self, extrinsics: Optional["SensorExtrinsics"] = None
@@ -305,6 +452,14 @@ class OpenCVFisheyeCameraIntrinsics(CameraIntrinsics):
             ],
             dtype=np.float64,
         )
+
+    def get_projection_model(self) -> ProjectionModel:
+        """Returns the ProjectionModel instance for fisheye distortion.
+
+        Returns:
+            a :class:`FisheyeProjectionModel` instance
+        """
+        return FisheyeProjectionModel()
 
 
 class SensorExtrinsics(DynamicEmbeddedDocument):
@@ -687,6 +842,7 @@ class CameraProjector:
         self.intrinsics = intrinsics
         self.camera_to_reference = camera_to_reference
         self.camera_convention = camera_convention
+        self._projection_model = intrinsics.get_projection_model()
 
     @classmethod
     def from_reference_to_camera(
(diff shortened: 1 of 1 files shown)


fiftyone/core/odm/dataset.py
```diff
@@ -937,6 +937,8 @@ class DatasetDocument(Document):
     default_mask_targets = MaskTargetsField()
     skeletons = DictField(EmbeddedDocumentField(KeypointSkeleton))
     default_skeleton = EmbeddedDocumentField(KeypointSkeleton)
+    camera_intrinsics = DictField(EmbeddedDocumentField(CameraIntrinsics))
```
kaixi-wang: does this also work for grouped datasets?/will there ever need to be multiple sets for different groups?

fiftyone/core/dataset.py
```diff
@@ -1100,6 +1093,402 @@ def default_skeleton(self, skeleton):
         self._doc.default_skeleton = skeleton
         self.save()
 
+    @property
+    def camera_intrinsics(self):
+        """A dict mapping sensor/camera names to
+        :class:`fiftyone.core.camera.CameraIntrinsics` instances that define
+        the intrinsic parameters for each camera in the dataset.
+
+        Intrinsics are keyed by sensor name (typically matching group slice
```
kaixi-wang, line 1103: If the keys aren't constrained (eg must be a group slice name), how will the FE resolve the right parameters to the right media?

fiftyone/core/dataset.py
```diff
@@ -1100,6 +1093,402 @@ def default_skeleton(self, skeleton):
...
+                    )
+
+            if value.target_frame is not None:
+                if value.target_frame != expected_target:
+                    raise ValueError(
+                        f"Key '{key}' expects target_frame='{expected_target}' "
+                        f"but got target_frame='{value.target_frame}'"
+                    )
+
+    def resolve_intrinsics(self, sample):
```
kaixi-wang, line 1332: based on your subsequent REST usage, does this actually need to be in dataset.py? What about a function that takes in dataset_id and sample_id to resolve the exact data needed so every request doesn't require loading the full dataset and full sample?

fiftyone/core/dataset.py
sashankaryal: Good question. When FE needs to resolve intrinsics / extrinsics, it's working backwards from the sample. If there's no way to reach the params from the sample, it's considered, for the sake of argument, "orphan". These "orphan" parameters are less for the FE and more for other BE workflows involving projection / unprojection that users might develop.

That said - in the future, we might run into use cases where we want to allow users to pick other intrinsics during runtime from the app

fiftyone/core/dataset.py
```diff
@@ -1100,6 +1093,402 @@ def default_skeleton(self, skeleton):
...
+            import fiftyone as fo
+
+            dataset = fo.Dataset()
+
+            # Set extrinsics for sensors
+            dataset.sensor_extrinsics = {
+                "camera_front::ego": fo.SensorExtrinsics(
+                    translation=[1.5, 0.0, 1.2],
+                    quaternion=[0.0, 0.0, 0.0, 1.0],
+                    source_frame="camera_front",
```
kaixi-wang, line 1208: What do source_frame and target_frame reference? is source_frame suppose to be the name of a group slice? In other words, how does this map to the associated sample level media/data

fiftyone/core/dataset.py
```diff
@@ -1100,6 +1093,402 @@ def default_skeleton(self, skeleton):
...
+
+        if extrinsics is None:
+            return
+
+        for key, value in extrinsics.items():
+            if not isinstance(value, foc.SensorExtrinsics):
+                continue
+
+            # Parse key to get expected frames
+            parts = key.split("::", 1)
```
kaixi-wang, line 1309: if the keys are expected to always be "source_frame::target_frame", why even allow setting keys manually at all?

sashankaryal: added `add_extrinsics()` API

fiftyone/core/dataset.py
sashankaryal: (We chatted about this offline)

fiftyone/core/camera.py
```diff
@@ -0,0 +1,841 @@
...
+            fy=1000.0,
+            cx=960.0,
+            cy=540.0,
+        )
+
+        # Access the 3x3 intrinsic matrix
+        K = intrinsics.intrinsic_matrix
+    """
+
+    fx = fof.FloatField(required=True)
```
kaixi-wang, line 204: do focal lengths have to be positive?

fiftyone/core/camera.py
```diff
@@ -0,0 +1,841 @@
...
+            target_frame="ego",
+        )
+
+        # Access the 4x4 transformation matrix
+        T = extrinsics.extrinsic_matrix
+    """
+
+    translation = fof.ListField(fof.FloatField(), default=None)
+    quaternion = fof.ListField(fof.FloatField(), default=None)
+    source_frame = fof.StringField(default=None)
```
kaixi-wang, line 517: in add_extrinsics, source_frame looks to be required. Would it be better to just do validation in a constructor?

fiftyone/core/camera.py
```diff
@@ -0,0 +1,841 @@
...
+                f"camera_convention must be 'opencv' or 'opengl', "
+                f"got '{camera_convention}'"
+            )
+
+        self.intrinsics = intrinsics
+        self.extrinsics = extrinsics
+        self.camera_convention = camera_convention
+
+        # Precompute matrices
+        self._K = intrinsics.intrinsic_matrix
```
kaixi-wang: I think storing both instrinsics and a precomputed _K could lead to bugs/inconsistencies.
  - Changes to fx, fy, cx, cy, skew → Not reflected (uses self._K)
  - Changes to k1, k2, p1, p2 → ARE reflected (reads from self.intrinsics)

For example:
```
intrinsics = PinholeCameraIntrinsics(
        fx=1000.0, fy=1000.0, cx=960.0, cy=540.0
    )
projector = CameraProjector(intrinsics)
print(projector.intrinsics)
print(projector._K)

projector.intrinsics.fx=2000
print(projector.intrinsics)
print(projector._K)

# projector.intrinsics updates but not _K
```

fiftyone/core/camera.py
sashankaryal: "are usually" positive, IDK about "have to be" positive

fiftyone/core/dataset.py
exupero: I take it this was one of the topics discussed offline. Out of curiosity, what was the verdict?

fiftyone/core/dataset.py
exupero: I wasn't in the meeting, but I'm curious about the answer.

fiftyone/core/camera.py
sashankaryal: good flag, thanks... i'll add a ctor which inits translation / quat to origin/identity + validate source_frame. assignment time validation can stay as is.

fiftyone/core/camera.py
sashankaryal: nice catch again... i'll compute _K on-the-fly... there's a slight but insignificant performance cost

fiftyone/core/dataset.py
```diff
@@ -6,54 +6,48 @@
...
-    UpdateOne,
-)
-from pymongo.collection import Collection
-from pymongo.errors import CursorNotFound, BulkWriteError
-
 import eta.core.serial as etas
 import eta.core.utils as etau
+import mongoengine.errors as moe
+from bson import DBRef, ObjectId, json_util
+from deprecated import deprecated
```
exupero: Not used
```suggestion
```

fiftyone/core/dataset.py
```diff
@@ -6,54 +6,48 @@
...
+from pymongo import DeleteMany, InsertOne, ReplaceOne, UpdateMany, UpdateOne
+from pymongo.collection import Collection
```
exupero: Not used
```suggestion
```

fiftyone/core/dataset.py
```diff
@@ -1100,6 +1094,509 @@ def default_skeleton(self, skeleton):
...
+            # Edit existing intrinsics
+            dataset.camera_intrinsics["camera_front"].k1 = -0.12
+            dataset.save()  # must save after edits
+        """
+        return self._doc.camera_intrinsics
+
+    @camera_intrinsics.setter
+    def camera_intrinsics(self, intrinsics):
+        self._doc.camera_intrinsics = intrinsics
+        self.save()
```
exupero, line 1134: Any validation needed, like is done in the `sensor_extrinsics` setter?

fiftyone/core/camera.py
```diff
@@ -0,0 +1,869 @@
...
+        points_3d_back = projector.unproject(points_2d, depth)
+    """
+
+    def __init__(
+        self,
+        intrinsics: CameraIntrinsics,
+        extrinsics: Optional[SensorExtrinsics] = None,
+        camera_convention: str = "opencv",
+    ):
+        if camera_convention not in ("opencv", "opengl"):
```
exupero: Supported conventions should be a constant somewhere.

fiftyone/core/camera.py
```diff
@@ -0,0 +1,869 @@
...
+        """
+        if distortion is None:
+            distortion = np.zeros(4, dtype=np.float64)
+
+        undistorted = cv2.fisheye.undistortPoints(
+            points.reshape(-1, 1, 2),
+            self._K,
+            distortion,
+        )
+        return undistorted.reshape(-1, 2)
```
exupero: Not sure how extensible this needs to be, but looks like it should use object composition on a class that defines both projection and undistort methods.

kaixi-wang (commented):
I just added the new fields to a dataset and samples, and now the dataset doesn't load in the app that hasn't been updated, so it breaks backwards compatibility. 

image

fiftyone/core/dataset.py
sashankaryal: Less essential than extrinsics (which needs mathematical guarantees). We could be defensive and require `fx, fy > 0` and `cx, cy > 0` but they could technically be negative...I cannot think of any real-world use cases but the math maths.

fiftyone/core/camera.py
sashankaryal: Good point. Will add `ProjectionModel` ABC that includes projection and undistort methods

PR #6703, opened 2025-12-30
FOEPD-2855: Add REST endpoint for querying camera params

sashankaryal:

This PR is a follow up on https://github.com/voxel51/fiftyone/pull/6700 and provides a REST endpoint to query intrinsics and extrinsics params stored in a dataset. These endpoints act as a thin facade over existing dataset APIs, namely `dataset.resolve_intrinsics()` and `dataset.resolve_extrinsics()`.

Unit Tests.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added camera calibration endpoints to retrieve per-sample intrinsics and extrinsics, including batch endpoints and support for optional source/target frames and chained transforms.

* **Bug Fixes / Behavior Changes**
  * Standardized the default extrinsics target frame and tightened validation to ensure consistent handling of missing or mismatched target frames.

* **Refactor**
  * Centralized dataset/sample retrieval and updated route composition for camera endpoints.

* **Tests**
  * Added comprehensive unit tests for single/batch endpoints, parameter parsing, and error cases.

### DIFF
--- a/tests/unittests/server/routes/test_sample.py
+++ b/tests/unittests/server/routes/test_sample.py
@@ -350,7 +350,7 @@ class TestSampleRoutes:
         )
 
     @pytest.mark.asyncio
-    async def test_sample_not_found(self, mutator, mock_request, dataset_id):
+    async def test_sample_not_found(self, mutator, mock_request, dataset):
         """Tests that a 404 HTTPException is raised for a non-existent
         sample."""
         bad_id = str(ObjectId())
@@ -365,7 +365,7 @@ class TestSampleRoutes:
         assert exc_info.value.status_code == 404
         assert (
             exc_info.value.detail
-            == f"Sample '{bad_id}' not found in dataset '{dataset_id}'"
+            == f"Sample '{bad_id}' not found in dataset '{dataset.name}'"
         )
 
     @pytest.mark.asyncio
@@ -888,7 +888,7 @@ class TestSampleFieldRoute:
         )
 
     @pytest.mark.asyncio
-    async def test_sample_not_found(self, mutator, mock_request, dataset_id):
+    async def test_sample_not_found(self, mutator, mock_request, dataset):
         """Tests that a 404 is raised for a non-existent sample."""
         bad_id = str(ObjectId())
         mock_request.path_params["sample_id"] = bad_id
@@ -901,7 +901,7 @@ class TestSampleFieldRoute:
         assert exc_info.value.status_code == 404
         assert (
             exc_info.value.detail
-            == f"Sample '{bad_id}' not found in dataset '{dataset_id}'"
+            == f"Sample '{bad_id}' not found in dataset '{dataset.name}'"
         )
 
     @pytest.mark.asyncio


fiftyone/server/routes/camera.py
```diff
@@ -0,0 +1,173 @@
...
+            dataset_id,
+        )
+
+        dataset = _get_dataset(dataset_id)
+        sample = _get_sample(dataset, sample_id)
+
+        try:
+            intrinsics = dataset.resolve_intrinsics(sample)
+        except ValueError as err:
+            raise HTTPException(status_code=400, detail=str(err)) from err
```
CamronStaley: nit: Isn't 400 for a bad request? This looks like an error internally wouldn't that be a 500? Unless I guess this error only happens if dataset_id or sample_id are wrong? But in that case you'd probably wanna catch that when you get the dataset or the sample.

fiftyone/server/routes/camera.py
```diff
@@ -0,0 +1,173 @@
...
+            "Received GET request for camera extrinsics for sample %s "
+            "in dataset %s (source_frame=%s, target_frame=%s, chain_via=%s)",
+            sample_id,
+            dataset_id,
+            source_frame,
+            target_frame,
+            chain_via,
+        )
+
+        dataset = _get_dataset(dataset_id)
```
kaixi-wang: this seems like a pretty roundabout way to resolve this. Are you open to resolving this more straightforwardly with just a util method that takes in sample_id and dataset_id?

fiftyone/server/routes/camera.py
sashankaryal: Yeah it should be removed in `resolve_intrinsics`, I'll do so in a subsequent commit, thanks for the flag.

(However, in `resolve_extrinsics`, we raise `ValueError` for invalid `chain_via`, so I'll keep that. )

fiftyone/core/dataset.py
```diff
@@ -1322,12 +1322,13 @@ def _validate_sensor_extrinsics(self, extrinsics):
...
                 )
 
-            if value.target_frame is not None:
-                if value.target_frame != expected_target:
-                    raise ValueError(
-                        f"Key '{key}' expects target_frame='{expected_target}' "
-                        f"but got target_frame='{value.target_frame}'"
-                    )
+            # Treat target_frame=None as "world" for comparison
+            actual_target = value.target_frame or "world"
```
kaixi-wang: nit: Might be good to define the default as a top level constant

fiftyone/server/routes/camera.py
```diff
@@ -0,0 +1,327 @@
...
+from starlette.requests import Request
+from starlette.responses import JSONResponse
+
+import fiftyone.core.odm as foo
+from fiftyone.server import utils
+
+logger = logging.getLogger(__name__)
+
+
+def _get_dataset(dataset_id: str):
```
kaixi-wang: since the sample route has essentially the same functions, could you reuse/share that one?

kaixi-wang (approved):
didn't test but lgtm

PR #6718, opened 2026-01-08
Add Depth Anything V3 to model zoo

CharlesCNorton:
## Summary

- Add Depth Anything V3 models to the model zoo
- Create `fiftyone/utils/depth_anything.py` wrapper

## Models Added

| Model | Size | Description |
|-------|------|-------------|
| `depth-anything-v3-small-torch` | 99 MB | Lightweight, fast inference |
| `depth-anything-v3-base-torch` | 541 MB | Balanced performance |
| `depth-anything-v3-large-torch` | 1.64 GB | High accuracy |

## Usage

```python
import fiftyone as fo
import fiftyone.zoo as foz

model = foz.load_zoo_model("depth-anything-v3-base-torch")
dataset.apply_model(model, label_field="depth")

```

## Test plan

- [x] Model loading and inference
- [x] Output normalized heatmaps
- [x] Multiple input types (file path, PIL, NumPy, Tensor)

## Summary by CodeRabbit

* **New Features**
  * Added Depth Anything V3 integration with on-demand installation, image/batch support, per-image normalization and resizing, optional confidence/sky outputs, multi-view depth computation, and 3D export helpers (GLB/PLY).
* **Tests**
  * Added unit tests covering configuration, output processing, normalization, resizing, dtype handling, batching, zero-depth and other edge cases.

### DIFF
--- a/fiftyone/utils/depth_anything.py
+++ b/fiftyone/utils/depth_anything.py
@@ -26,7 +26,7 @@ import torch
 logger = logging.getLogger(__name__)
 
 
-def _ensure_depth_anything_3():
+def _ensure_depth_anything_3() -> None:
     if not fou.ensure_package("depth-anything-3", error_level=2):
         fou.install_package(
             "git+https://github.com/ByteDance-Seed/depth-anything-3.git"
@@ -226,7 +226,7 @@ class DepthAnythingV3ModelConfig(fout.TorchImageModelConfig, fozm.HasZooModel):
             to extract intermediate features from (e.g. ``[0, 5, 10]``)
     """
 
-    def __init__(self, d):
+    def __init__(self, d: dict) -> None:
         d = self.init(d)
         super().__init__(d)
 
@@ -283,12 +283,12 @@ class DepthAnythingV3Model(fout.TorchImageModel):
         config: a :class:`DepthAnythingV3ModelConfig`
     """
 
-    def _download_model(self, config):
+    def _download_model(self, config: DepthAnythingV3ModelConfig) -> None:
         from huggingface_hub import snapshot_download
 
         snapshot_download(config.name_or_path)
 
-    def _load_model(self, config):
+    def _load_model(self, config: DepthAnythingV3ModelConfig) -> torch.nn.Module:
         model = da3_api.DepthAnything3.from_pretrained(config.name_or_path)
         model = model.to(self._device)
         if self.using_half_precision:
@@ -297,7 +297,7 @@ class DepthAnythingV3Model(fout.TorchImageModel):
         return model
 
     @property
-    def media_type(self):
+    def media_type(self) -> str:
         return "image"
 
     def _forward_pass(self, imgs: List[np.ndarray]) -> Dict[str, Any]:


fiftyone/utils/depth_anything.py
```diff
@@ -0,0 +1,444 @@
...
+import torch
+
+logger = logging.getLogger(__name__)
+
+
+def _ensure_depth_anything_3():
+    if not fou.ensure_package("depth-anything-3", error_level=2):
+        fou.install_package(
+            "git+https://github.com/ByteDance-Seed/depth-anything-3.git"
+            "@2c21ea849ceec7b469a3e62ea0c0e270afc3281a"
```
lorenzoriano, line 33: Is this right? Pinning the zoo model to a specific commit?

fiftyone/utils/depth_anything.py
lorenzoriano, line 1: can you add type hints to the functions?

fiftyone/utils/depth_anything.py
```diff
@@ -0,0 +1,444 @@
...
+                conf_maps = conf_maps[np.newaxis, ...]
+
+        sky_masks = output.get("sky")
+        if sky_masks is not None:
+            if isinstance(sky_masks, torch.Tensor):
+                sky_masks = sky_masks.detach().cpu().numpy()
+            if len(sky_masks.shape) == 2:
+                sky_masks = sky_masks[np.newaxis, ...]
+
+        from PIL import Image
```
lorenzoriano: why the import here instead of top of the file?

fiftyone/utils/depth_anything.py
```diff
@@ -0,0 +1,444 @@
...
+
+        width, height = frame_size
+        results = []
+
+        for i, depth in enumerate(depth_maps):
```
lorenzoriano, line 143: somewhere you should have a check for None or empty depth_maps

fiftyone/utils/depth_anything.py
```diff
@@ -0,0 +1,444 @@
...
+        dataset.apply_model(model, label_field="depth")
+
+        session = fo.launch_app(dataset)
+
+    Args:
+        config: a :class:`DepthAnythingV3ModelConfig`
+    """
+
+    def _download_model(self, config):
+        pass
```
lorenzoriano: I think the model downloading part should be here, and `_load_model` will just load it

fiftyone/utils/depth_anything.py
```diff
@@ -0,0 +1,444 @@
...
+                logger.warning(
+                    "infer_gs=True but model returned no gaussians. "
+                    "Only giant and nested models support Gaussian Splatting"
+                )
+        if prediction.aux:
+            output["aux"] = prediction.aux
+        return output
+
+
+def compute_multiview_depth(
```
lorenzoriano: I'm not sure it's good to have a standalone function. Is that common in the repo? If I call it several times it'll create the model over and over. What about having it as a method of `DepthAnythingV3Model`?

```
    def compute_multiview_depth(self, filepaths, **kwargs):
        prediction = self._model.inference(
            filepaths,
            process_res=self.config.process_res,
            use_ray_pose=kwargs.get('use_ray_pose', False),
            ...
        )
```

fiftyone/utils/depth_anything.py
CharlesCNorton: no pypi package or release tags so commit pin is their latest. Will unpin when they publish.

fiftyone/utils/depth_anything.py
lorenzoriano: There are still functions that don't have type hints, can you (auto-) add them?

PR #6721, opened 2026-01-09
FOEPD-2894: Use DynamicEmbeddedDocument as fallback when new loading newer datasets with legacy clients

sashankaryal:

Use `DynamicEmbeddedDocument` as fallback when new cls is not found. This makes legacy clients of FO more resilient when loading datasets created by newer SDKs.

Tested loading example dataset from https://github.com/voxel51/fiftyone/pull/6700 in older version of FO.

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when loading datasets containing unknown embedded field types.
  * Unknown types now generate warnings and fall back to safe dynamic representations instead of preventing dataset loading.
  * Nested and list-based embedded documents are handled more reliably during deserialization.

* **Tests**
  * Added coverage for unknown embedded documents, including nested documents, lists, and preservation of recognized document types.

### DIFF
--- a/tests/unittests/odm_tests.py
+++ b/tests/unittests/odm_tests.py
@@ -188,3 +188,86 @@ class GetIndexedValuesTests(unittest.TestCase):
 
         finally:
             dataset.delete()
+
+
+class SanitizeUnknownEmbeddedDocsTests(unittest.TestCase):
+    """Tests for graceful handling of unknown embedded document types."""
+
+    def test_sanitize_simple_unknown_document(self):
+        """Test sanitizing a simple unknown embedded document."""
+        import fiftyone.core.odm.utils as foou
+
+        data = {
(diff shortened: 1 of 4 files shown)


tests/unittests/odm_tests.py
```diff
@@ -109,3 +109,86 @@ def test_get_indexed_values(self):
 
         finally:
             dataset.delete()
+
+
+class SanitizeUnknownEmbeddedDocsTests(unittest.TestCase):
```
kaixi-wang, line 193: for this type of change, it's prob best to do full e2e tests with fo.Dataset + samples
for example, testing to_dict(), from_dict(), clone(), add_sample_field(), etc

PR #6726, opened 2026-01-12
update httplib2 version to fix security vulnerability [FOEPD-2771]

jleven:

bump `httplib2` version to `0.19`

Running `pip` resulted in the following:

```
Successfully installed botocore-1.36.2 httplib2-0.19.0 pyparsing-2.4.7 s3transfer-0.11.3
```

Read through [CHANGELOG](https://github.com/httplib2/httplib2/blob/master/CHANGELOG) and I don't see any breaking changes.

I have not thoroughly testing running the application.

        notes for FiftyOne users.

Security fix for httplib2

## Summary by CodeRabbit

* **Chores**
  * Updated google-api-python-client minimum to >=1.7.12 to ensure compatibility with newer API fixes.
  * Relaxed httplib2 constraint to require >=0.19 by removing the previous upper bound, improving compatibility with recent transports.
  * Added shapely (>=1.7.1) to support geometry operations required by newer features.

### DIFF
--- a/requirements/extras.txt
+++ b/requirements/extras.txt
@@ -1,6 +1,6 @@
-google-api-python-client>=1.6.5
+google-api-python-client>=1.7.12
 google-cloud-storage>=1.36
-httplib2==0.19
+httplib2>=0.19
 ipywidgets>=7.5
 notebook>=5.3
 pydicom>=2.2.0


afoley587 (approved):
Looks good to me! Thanks!

This `requirements` file isn't used in any of our wheels, images, etc. Should we update this elsewhere as well?

PR #6729, opened 2026-01-12
Add rerunnable flag to specify if a given DO can be rerun

CamronStaley:

Add the ability to specify if a given DO or pipeline stage is rerunnable. If they are not rerunnable it will not show the rerun option in the UI and calling manually will throw a value error.

Unit tests + created a test operator and delegated it:

```
import asyncio

import fiftyone as fo
import fiftyone.operators.executor as foe

uri = "@voxel51/stu-testing/stu_pipeline"

request_params = {
    "dataset_name": "quickstart",
    "operator_uri": uri,
    "params": {"message": "Hello, World!"},
    "request_delegation": True,
}

result = asyncio.run(foe.execute_or_delegate_operator(uri, request_params))
print(result)
```

I changed the rerunnable param and tested execute_or_delegate to verify the docs are created as expected. Then I tested trying to rerun each of them with the new `fiftyone delegated rerun` command:

```
(OSS) ➜  fiftyone git:(feat/rerunnable-flag) ✗ fiftyone delegated rerun 69652a2cd02754776c863cf9
  File "/Users/camronstaley/Workplace/envs/OSS/bin/fiftyone", line 8, in <module>
    sys.exit(main())
  File "/Users/camronstaley/Workplace/fiftyone/fiftyone/core/cli.py", line 5269, in main
    args.execute(args)
  File "/Users/camronstaley/Workplace/fiftyone/fiftyone/core/cli.py", line 5252, in <lambda>
    parser.set_defaults(execute=lambda args: command.execute(parser, args))
  File "/Users/camronstaley/Workplace/fiftyone/fiftyone/core/cli.py", line 3185, in execute
    dos.rerun_operation(ObjectId(args.id))
  File "/Users/camronstaley/Workplace/fiftyone/fiftyone/operators/delegated.py", line 461, in rerun_operation
    raise ValueError(f"Delegated operation {doc_id} is not rerunnable")
ValueError: Delegated operation 69652a2cd02754776c863cf9 is not rerunnable
```

        notes for FiftyOne users.

Now if you set "rerunnable" to false on the operator config of your operation users will not be able to rerun that operation. Additionally if you provide rerunnable false or true as part of a pipeline stage we will rerun that operation automatically when the pipeline stages are rerun. The pipeline stage will overwrite whatever is set on the operation, None will default back to whatever is set on the operation.

Also, you can now use `fiftyone delegated rerun operationId` to rerun delegated operations.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added a rerunnable flag across operators and pipeline stages; persisted in operator configs and stage serialization (defaults to true).
  * CLI: added a "rerun" delegated-operation command that creates a rerun operation and prints the new operation ID.

* **Behavior Changes / Bug Fixes**
  * Prevents rerunning operations marked non-rerunnable and requires parent operations to be in a terminal state before allowing child reruns.

### DIFF
--- a/fiftyone/operators/delegated.py
+++ b/fiftyone/operators/delegated.py
@@ -461,11 +461,7 @@ class DelegatedOperationService(object):
             raise ValueError(f"Delegated operation {doc_id} is not rerunnable")
         if doc.parent_id:
             parent_doc = self._repo.get(_id=doc.parent_id)
-            if (
-                parent_doc
-                and parent_doc.run_state
-                not in ExecutionRunState.TERMINAL_STATES
-            ):
+            if parent_doc.run_state not in ExecutionRunState.TERMINAL_STATES:
                 raise ValueError(
                     f"Cannot rerun delegated operation {doc_id} because its "
                     f"parent operation {doc.parent_id} is not yet completed"


fiftyone/operators/delegated.py
```diff
@@ -454,8 +457,18 @@ def rerun_operation(self, doc_id):
         doc = self._repo.get(_id=doc_id)
         if doc is None:
             raise ValueError(f"No delegated operation with {doc_id=} found")
+        if not doc.rerunnable:
+            raise ValueError(f"Delegated operation {doc_id} is not rerunnable")
         if doc.parent_id:
-            raise ValueError("Cannot rerun a child delegated operation.")
+            parent_doc = self._repo.get(_id=doc.parent_id)
+            if parent_doc and parent_doc.run_state not in {
```
yrahal: what happens if/when parent_doc is `None`?

fiftyone/operators/delegated.py
```diff
@@ -454,8 +457,18 @@ def rerun_operation(self, doc_id):
...
+                ExecutionRunState.COMPLETED,
```
yrahal: probably outside the scope of this review, but it'd be great to have a helper to just check if a state is final instead of checking them one by one (I assume there are other places where we're doing similar checks)

fiftyone/operators/delegated.py
```diff
@@ -454,8 +457,18 @@ def rerun_operation(self, doc_id):
...
+                ExecutionRunState.FAILED,
+            }:
+                raise ValueError(
+                    f"Cannot rerun delegated operation {doc_id} because its "
+                    f"parent operation {doc.parent_id} is not yet completed"
+                )
         return self._repo.queue_operation(**doc.__dict__)
```
yrahal, line 469: Why isn't the state of the doc checked (not just the parent)? Are we allowed to rerun multiple instances of the same doc in parallel?

is the state of the operation checked in the requeue function? If so, it would be good to mention it via a comment here.

fiftyone/operators/delegated.py
CamronStaley: This should never happen but if it did, then the user would be able to rerun the child operation which I guess would be fine in the strange case that the parent DO was deleted. For context deleting a parent will delete all the children so its unlikely / should be impossible to end up in this state.

fiftyone/operators/delegated.py
CamronStaley: This follows how it's done elsewhere, but I can add a constant var to be reused if we ever did happen to add a new final run state in the future.

fiftyone/operators/delegated.py
CamronStaley: Rerunning just creates a clone of a document and puts it into "scheduled" state. If a DO isn't a child, then there's really no implications to creating a clone of it so we are okay with that. 

If a DO is a child, then rerunning has implications for the parent / pipeline which is why we are restricting that for now. Maybe eventually we will allow rerunning children within a pipeline and in that case we'd need to require them to be completed so that we don't have conflicts in the pipeline.

fiftyone/operators/delegated.py
yrahal: In that case I'd remove the `if parent_doc` check and let it fail in such a weird edge case instead or explicitly guard against it instead of the silent check.

fiftyone/operators/delegated.py
yrahal: no worries for this PR

fiftyone/operators/delegated.py
yrahal: Isn't the code above restricting the children to be rerun only if the parent is not in a final state and allowing otherwise?

CamronStaley: we are throwing an exception if the parent is not in a terminal state which is what we want. We should only allow rerunning if the parent is in a terminal state.

yrahal: Thanks. That's how I understood it from the code but wasn't sure given the previous comment :-p

PR #6731, opened 2026-01-12
FOEPD-2908: Fix scipy-torch regression

sashankaryal:

The `import_deps_test.py` test sets `sys.modules["torch"] = None` (and other optional deps) to verify that `import fiftyone` doesn't require these packages. This worked with scipy 1.14.x but broke with scipy 1.17.0 (released two days ago). scipy 1.17.0 added array API compatibility code that checks for torch types. When `sys.modules["torch"] = None`, the `getattr(None, "Tensor")` in scipy's code raises `AttributeError` which isn't caught, causing the test to fail.

The solution replaces `None` with placeholder classses that raise when actually `used` but do not raise when accessed for type inspection (like what scipy is doing...).

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Tests**
  * Improved handling of missing optional dependencies in the test infrastructure: unavailable packages are now simulated so tests raise clear module-not-found errors on access, making test behavior more consistent and reliable when optional libraries (e.g., TensorFlow, PyTorch, torchvision, etc.) are absent.

### DIFF
--- a/tests/isolated/import_deps_test.py
+++ b/tests/isolated/import_deps_test.py
@@ -12,14 +12,35 @@ from types import ModuleType
 
 
 class _UnavailableModule(ModuleType):
-    """A module placeholder that raises ModuleNotFoundError on attribute access."""
+    """A module placeholder that simulates an unavailable optional dependency.
+
+    Returns placeholder classes for attribute access that:
+    - Won't match real types in isinstance/issubclass checks (returns False)
+    - Raise ModuleNotFoundError when actually used (instantiated, called, etc.)
+    """
 
     def __init__(self, name):
         super().__init__(name)
         self._name = name
+        self._cache = {}
+
+    def _make_placeholder(self, attr):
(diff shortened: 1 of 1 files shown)


tests/isolated/import_deps_test.py
```diff
@@ -18,12 +50,12 @@
 #
 # https://docs.python.org/3/reference/import.html#the-module-cache
 #
-sys.modules["tensorflow"] = None
-sys.modules["tensorflow_datasets"] = None
-sys.modules["torch"] = None
-sys.modules["torchvision"] = None
-sys.modules["flash"] = None
-sys.modules["pycocotools"] = None
+sys.modules["tensorflow"] = _UnavailableModule("tensorflow")
```
kaixi-wang, line 53: should we just `del sys.modules[<pkg>]` instead?

sashankaryal: i don't think that will do it...

PR #6757, opened 2026-01-15
[HA] FOEPD-2633: Schema Manager configure component type in primitives and label attributes

lanzhenw:
- Updates the schema manager FE to handle list type attributes
- Add/Edit/Remove/Reorder a label attribute and configure component input types in Schema Manager
- Allow user to modify primitive fields and configure its component input types in GUI
- Refactor and cleanup: use domain hooks, no direct import of atoms in components

  * GUI: add/edit/delete attribute workflow with inline cards and reusable action controls
  * Primitive fields: dedicated configuration UI (types, component, values, range, default, read-only)

This PR completes the following features:

FOEPD-2633 GUI for Field Schema MVP
FOEPD-2795 HA M4: configure UI types for primitive field
FOEPD-2796 HA M4: configure UI types for label attribute 

https://github.com/user-attachments/assets/2beb655d-c7b6-4297-bf8d-cedf2b6cafb1

live at https://lanzhen.ephem.fiftyone.ai/datasets

        notes for FiftyOne users.

Allow user to configure input types and used component in the schema manager on primitive fields and label attributes. 

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Add/edit/delete attributes directly in-place with a dedicated “new attribute” card, richer attribute forms (type, component, values, range, defaults) and a reusable actions row.
  * Improved primitive field support with specialized editors for JSON, dates, component-driven forms, and configurable input types.

* **UI Improvements**
  * Simplified header displaying field info as read-only, conditional “Scan” action, streamlined layout, real-time validation, and clearer inline errors.

### DIFF
--- a/e2e-pw/src/oss/specs/smoke-tests/schema-manager.spec.ts
+++ b/e2e-pw/src/oss/specs/smoke-tests/schema-manager.spec.ts
@@ -117,9 +117,8 @@ test.describe.serial("schema manager", () => {
       classes: ["value"],
     });
 
-    // Save
+    // Save (automatically navigates back to field list)
     await jsonEditor.save();
-    await schemaManager.back();
 
     // Activate
     await row.assert.hasCheckbox();


plugins/operators/annotation.py
```diff
@@ -150,10 +150,110 @@ def config(self):
...
+            subfield = list_types[attr_type]()
+            print(f"[DEBUG] Adding list field: {attr_path}")
+            dataset.add_sample_field(attr_path, ftype, subfield=subfield)
+        elif attr_type in type_to_ftype:
+            ftype = type_to_ftype[attr_type]
+            print(f"[DEBUG] Adding field: {attr_path} with type {ftype}")
+            dataset.add_sample_field(attr_path, ftype)
+        else:
+            print(f"[DEBUG] Unknown attr_type: {attr_type}, not adding field")
+
```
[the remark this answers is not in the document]
lanzhenw: good catch

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -317,7 +319,10 @@ def _generate_field_label_schema(collection, field_name, scan_samples):
     if isinstance(field, fof.BooleanField):
         fn = _handle_bool
     elif isinstance(field, (fof.FloatField, fof.IntField)):
-        fn = _handle_float_or_int
+        is_int = isinstance(field, fof.IntField)
+        fn = lambda c, f, il, s, ss: _handle_float_or_int(
+            c, f, il, s, ss, is_int
+        )
```
lanzhenw: @benjaminpkane doing this because `_handle_float_or_int` wants a is_int flag to determine the default step value for int or float.

plugins/operators/annotation.py
```diff
@@ -150,10 +156,137 @@ def config(self):
...
+                        f"got {type(values).__name__}"
+                    )
+                # For list types, validate element types
+                if is_list_type:
+                    expected_elem = attr_type.split("<")[1].rstrip(">")
+                    type_map = {
+                        "str": str,
+                        "int": int,
+                        "float": (int, float),
+                        "bool": bool,
```
lanzhenw: to cut.

plugins/operators/annotation.py
```diff
@@ -150,10 +156,145 @@ def config(self):
...
+            if not isinstance(values, list):
+                raise ValueError(
+                    f"Values for attribute '{attr_name}' must be a list, "
+                    f"got {type(values).__name__}"
+                )
+
+            # For list types, validate element types
+            if is_list_type:
+                expected_elem = attr_type.split("<")[1].rstrip(">")
+                type_map = {
```
erik-nieh: maybe move to constant outside this

plugins/operators/annotation.py
```diff
@@ -150,10 +156,145 @@ def config(self):
...
+
+        # Create new attribute fields first if specified
+        if new_attributes:
+            _add_new_attributes(ctx.dataset, field, new_attributes)
+
         ctx.dataset.update_label_schema(field, label_schema)
         return {"label_schema": label_schema}
 
 
+def _add_new_attributes(dataset, field, new_attributes) -> None:
```
erik-nieh: I think the size and complexity of this function warrants refactoring it into smaller parts/functions

plugins/operators/annotation.py
```diff
@@ -150,10 +156,145 @@ def config(self):
...
+
+    # Map label schema types to field types
+    type_to_ftype = foac.TYPES_TO_FIELD_TYPE
+
+    # Components that require values
+    values_components = {"radio", "dropdown", "checkboxes"}
+
+    # Validate each attribute entry up-front
+    for attr_name, attr_schema in new_attributes.items():
+        if not isinstance(attr_name, str):
```
erik-nieh: since all of these `raise` safe to just scoop them all into a different function and call it I think

fiftyone/core/annotation/validate_label_schemas.py
```diff
@@ -590,6 +593,29 @@ def _validate_range_setting(field_name, value, _type):
     )
 
 
+def _validate_step_setting(field_name, value, _range=None, _type=None):
```
erik-nieh: within this function probably ok to just use `range`/`type` rather than private convention

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/constants.ts
```diff
@@ -26,3 +39,208 @@ const SYSTEM_READ_ONLY_FIELDS_SET = new Set<string>(
...
+    { id: "dropdown", label: "Dropdown", icon: IconName.Search },
+    { id: "text", label: "Text", icon: IconName.Text },
+  ],
+};
+
+// Types that are numeric (values must be numbers)
+export const NUMERIC_TYPES = ["int", "float", "list<int>", "list<float>"];
+
+// Types that don't need/support a default value
+export const NO_DEFAULT_TYPES = ["bool", "date", "datetime", "dict", "id"];
```
erik-nieh, line 127: In addition to `types`, it seems like schema is enforcing this on `components` as well, for example a slider with a default is being rejected.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/AddAttributeCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+  onSave: (name: string, config: AttributeConfig) => void;
+  onCancel: () => void;
+}
+
+const AddAttributeCard = ({
+  existingAttributes,
+  onSave,
+  onCancel,
+}: AddAttributeCardProps) => {
+  const [formState, setFormState] = useState<AttributeFormData>(
```
erik-nieh, line 29: There is a small bug related to this, I have not dug deeply. But if you switch between component "types" the `values` I think do not get refreshed.

For example starting on a dropdown with existing values:

Image

if I click "text", values will be reset. If I click back on dropdown, the previously existing values are missing.

erik-nieh: Image

lanzhenw: switching between text and dropdown needs to reset the values, because otherwise the default value for text on can't get validated if it's not within the values. However, values are not required if the user select text input. I think values shouldn't change when changing between dropdown/radio, and that's the only exception case.

plugins/operators/annotation.py
```diff
@@ -150,10 +156,182 @@ def config(self):
     def execute(self, ctx):
         field = ctx.params.get("field", None)
         label_schema = ctx.params.get("label_schema", None)
-        ctx.dataset.update_label_schema(field, label_schema)
```
lanzhenw, line 153: move add new attributes to inside update_label_schema

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -313,7 +317,12 @@ def _generate_field_label_schema(collection, field_name, scan_samples):
     if isinstance(field, fof.BooleanField):
         fn = _handle_bool
     elif isinstance(field, (fof.FloatField, fof.IntField)):
-        fn = _handle_float_or_int
+        is_int = isinstance(field, fof.IntField)
+
+        def _handle_numeric(c, f, il, s, ss, _is_int=is_int):
```
kaixi-wang: this is pretty unreadable

fiftyone/core/annotation/utils.py
```diff
@@ -206,3 +206,165 @@ def _is_supported_primitive(field):
...
+        )
+
+    if not isinstance(values, list):
+        raise ValueError(
+            f"Values for attribute '{attr_name}' must be a list, "
+            f"got {type(values).__name__}"
+        )
+
+
+def _get_attribute_base_path(dataset, field):
```
kaixi-wang: is there not a another function in the sdk for this? Seems like this isn't annotation specific

fiftyone/core/annotation/utils.py
```diff
@@ -206,3 +206,165 @@ def _is_supported_primitive(field):
...
+    # Get the base path for attributes
+    base_path = _get_attribute_base_path(dataset, field)
+    if base_path is None:
+        return
+
+    # Add each attribute field
+    for attr_name, attr_schema in new_attributes.items():
+        attr_type = attr_schema.get("type", "str")
+        attr_path = f"{base_path}.{attr_name}"
+        _add_attribute_field(dataset, attr_path, attr_type, type_to_ftype)
```
kaixi-wang: Dataset.update_label_schema() is calling into foa.add_new_attributes(), but those foa methods are really only used by plugins right? That creates a bit of a circular dependency — core calling annotation utilities that just turn around and call core primitives like add_sample_field.
Also, add_new_attributes seems like it's mostly just wrapping add_sample_field with some path construction. Do we need a separate method for that?
Since label schema is a subset of sample schema anyway, could the plugin just own this workflow? Something like: plugin adds the fields it needs via add_sample_field, then calls update_label_schema. Keeps core simpler and avoids the back-and-forth.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/AddAttributeCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+    createDefaultFormData()
+  );
+  const [isDirty, setIsDirty] = useState(false);
+
+  const nameError = getAttributeNameError(formState.name, existingAttributes);
+  const formErrors = getAttributeFormErrors(formState);
+  const showError = isDirty && nameError;
```
erik-nieh, line 36: There are a few cases I think one is the slider component where defaults are not valid ( because they need user input) But this error will not show because the form is not dirty. I don't think this is incorrect but it may be unexpected to the user because there's no prompt or instructions

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/AttributeCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Standalone attribute card component with RichList wrapper.
+ * Use this for the "Add attribute" card that renders independently.
+ */
+const AttributeCard = (props: AttributeCardProps) => {
+  const listItem = createAttributeCardItem(props);
+
+  return (
+    <div style={{ marginBottom: "1rem" }}>
```
erik-nieh: This is fine but consider using a stack with spacing I think it will do the same thing. Even if there's only one element

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/AttributesSection.tsx
```diff
@@ -13,60 +13,180 @@ import {
...
-        actions: <EditAction onEdit={() => onEditAttribute(name)} />,
-      });
+        data: {
+          canSelect: false,
+          canDrag: true,
+          primaryContent: name,
+          secondaryContent: (
+            <>
+              {secondaryParts.join(" · ")}
+              {isM4Enabled && config.read_only && (
```
erik-nieh, line 157: How do we typically handle removing feature flags? would we just set the default to True at some point or do we try to remove them from code?

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/ComponentTypeButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+  isSelected,
+  onClick,
+  largeText = false,
+}: ComponentTypeButtonProps) => {
+  const theme = useTheme();
+
+  return (
+    <div style={{ flex: 1 }}>
+      <Clickable onClick={onClick}>
+        <div
```
erik-nieh, line 35: I wonder how we can handle this in design system in the future. it seems like this is custom enough that maybe we shouldn't

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/ListDefaultInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+      },
+    }),
+    [choices, isNumeric]
+  );
+
+  const handleChange = (_path: string, newValues: unknown) => {
+    if (Array.isArray(newValues)) {
+      // Convert to numbers if numeric type
+      const processed = isNumeric
+        ? newValues.map((v) => {
```
erik-nieh: maybe move this function to it's own declaration

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/ListDefaultInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+            const num = parseFloat(String(v));
+            return isNaN(num) ? v : num;
+          })
+        : newValues;
+      onChange(processed);
```
erik-nieh: Maybe not necessary - do we want to de-dupe this? Preventing users from having duplicate values in their options?

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/PrimitiveFieldContent.tsx
```diff
@@ -0,0 +1,235 @@
...
+  // Derived state
+  const isNumericType = NUMERIC_TYPES.includes(schemaType);
+  const isIntegerType = schemaType === "int" || schemaType === "list<int>";
+
+  // Visibility flags
+  const showValues = componentNeedsValues(component);
+  const showRange = isNumericType && componentNeedsRange(component);
+
+  // Validation errors
+  const errors = useMemo(() => {
```
erik-nieh, line 82: It might be nice to move this validation logic to its own function maybe in a different file so we could use it elsewhere since it seems so useful

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/PrimitiveFieldContent.tsx
```diff
@@ -0,0 +1,235 @@
...
+      range: null as string | null,
+    };
+
+    // Values validation - required for radio/dropdown/checkboxes
+    if (showValues && values.length === 0) {
+      result.values = "At least one value is required";
+    }
+
+    // Range validation - required for slider
+    if (showRange) {
```
erik-nieh: I also worry a little bit that we are probably doing the same validation on the back end and as we talked about last week what happens when these things are out of sync or how often will they be out of sync? I don't think we should change anything here because it seems necessary though

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/PrimitiveFieldContent.tsx
```diff
@@ -0,0 +1,235 @@
...
+      onConfigChange(newConfig);
+    },
+    [config, onConfigChange]
+  );
+
+  const handleValuesChange = useCallback(
+    (newValues: string[]) => {
+      if (!onConfigChange) return;
+      const convertedValues = isNumericType
+        ? newValues.map((v) => parseFloat(v)).filter((n) => !isNaN(n))
```
erik-nieh, line 113: I think this is similar logic to the other component maybe we could share it

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/RangeInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const max = range?.max || "";
+
+  const labelVariant = largeLabels ? TextVariant.Lg : TextVariant.Md;
+  const labelColor = largeLabels ? TextColor.Primary : TextColor.Secondary;
+
+  return (
+    <div>
+      <div style={{ width: "100%", display: "flex", gap: "1rem" }}>
+        <div style={{ flex: 1 }}>
+          <Text
```
erik-nieh, line 42: I think FormField can do some of this: https://voodo.dev.fiftyone.ai/?path=/docs/components-formfield--docs&globals=backgrounds.value:dark;backgrounds.grid:!false

if not do you mind providing feedback why? So we could update it

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/ValuesList.tsx
```diff
@@ -0,0 +1,164 @@
...
+  onValuesChange,
+  isNumeric = false,
+  isInteger = false,
+  error = null,
+  largeLabels = false,
+}: ValuesListProps) => {
+  const [newValue, setNewValue] = useState("");
+  const [inputError, setInputError] = useState<string | null>(null);
+
+  const validateValue = (val: string): string | null => {
```
erik-nieh: maybe another validation function to move to a shared file

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/EditFieldLabelSchema.tsx
```diff
@@ -91,12 +44,19 @@ const EditFieldLabelSchema = ({ field }: { field: string }) => {
...
     <EditContainer>
-      <Header field={field} setField={setCurrentField} />
+      <Header field={field} />
 
       {isM4Enabled && (
-        <div className="my-4">
-          <div className="flex items-center justify-between mb-1">
-            <Text variant={TextVariant.Xl}>Read-only</Text>
+        <div style={{ marginTop: "1rem", marginBottom: "1rem" }}>
+          <div
```
erik-nieh, line 51: Maybe more `<Stack>` opportunities

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -3,12 +3,78 @@
...
+ * Safely extract attribute names from a schema value.
+ * Attributes are stored as an array of objects with 'name' field.
+ */
+const getAttributeNames = (value: unknown): Set<string> => {
+  if (value && typeof value === "object" && "attributes" in value) {
+    const attrs = (value as LabelSchema).attributes;
+    if (Array.isArray(attrs)) {
+      return new Set(
+        attrs
+          .filter((attr) => attr && typeof attr === "object" && "name" in attr)
```
erik-nieh: This predicate is shared several times, could probably be moved to its own function and passed in

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/constants.ts
```diff
@@ -26,3 +39,208 @@ const SYSTEM_READ_ONLY_FIELDS_SET = new Set<string>(
...
+
+export const ATTRIBUTE_TYPE_OPTIONS = [
+  { id: "str", data: { label: "String" } },
+  { id: "int", data: { label: "Integer" } },
+  { id: "float", data: { label: "Float" } },
+  { id: "bool", data: { label: "Boolean" } },
+  { id: "date", data: { label: "Date" } },
+  { id: "datetime", data: { label: "Datetime" } },
+  { id: "dict", data: { label: "Dictionary" } },
+  { id: "list<str>", data: { label: "String list" } },
```
erik-nieh: could this be a `Record<type, label>` instead of array? I think only piece of data is the label

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/constants.ts
```diff
@@ -26,3 +39,208 @@ const SYSTEM_READ_ONLY_FIELDS_SET = new Set<string>(
...
+  { id: "list<int>", data: { label: "Integer list" } },
+  { id: "list<float>", data: { label: "Float list" } },
+];
+// Component options by type (matches annotation/constants.py)
```
erik-nieh: Might be worth adding the full github link with commit hash

Also might be worth (at some point, not this PR) moving this into the schema itself. I guess some frontend mapping is needed for the icon though.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/utils.ts
```diff
@@ -155,3 +204,242 @@ export const buildFieldSecondaryContent = (
...
+    errors.values = "At least one value is required";
+  } else if (needsValues && isNumeric) {
+    const invalid = data.values.find((v) => isNaN(parseFloat(v)));
+    if (invalid !== undefined) {
+      errors.values = "Values must be valid numbers";
+    }
+  }
+
+  // Range validation (for slider)
+  if (needsRange) {
```
erik-nieh: I think breaking each of these into their own functions, even if not exported, would be helpful

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/AttributesSection.tsx
lanzhenw: We would actually remove it from code.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/ListDefaultInput.tsx
lanzhenw: we should.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/PrimitiveFieldContent.tsx
lanzhenw: Users may do bulk edit, so if something will fail, users need to be warned early to avoid frustrations.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/constants.ts
lanzhenw: on primitives, default is not allowed in the BE.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/AddAttributeCard.tsx
lanzhenw: right now, the validation is onBlur, and default is expected to be within range or values if provided for numbers.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/ComponentTypeButton.tsx
lanzhenw: adding an `equalWidth` prop to `Stack` can be helpful. 

equalWidth means children should have flex:1

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -313,7 +313,12 @@ def _generate_field_label_schema(collection, field_name, scan_samples):
...
```
kaixi-wang: Is this necessary? You could use a partial if you need to pass the kwarg rather than define this nested function. However, is_int doesn't seem to even be used in _handle_float_or_int?

fiftyone/core/annotation/utils.py
```diff
@@ -206,3 +206,165 @@ def _is_supported_primitive(field):
...
+
+    if attr_type.startswith("list<"):
+        subfield = type_to_ftype[attr_type]()
+        dataset.add_sample_field(attr_path, fof.ListField, subfield=subfield)
+    else:
+        ftype = type_to_ftype[attr_type]
+        dataset.add_sample_field(attr_path, ftype)
+
+
+def add_new_attributes(dataset, field, new_attributes) -> None:
```
kaixi-wang: What about moving all the validation back to update_label_schema and making this more pure? Otherwise, you end up constantly validating and revalidating. For example, currently the code goes `dataset.update_label_schema` -> `add_new_attributes` -> `foa.validate_label_schemas(self, label_schema, fields=field)`

fiftyone/core/annotation/utils.py
```diff
@@ -206,3 +206,165 @@ def _is_supported_primitive(field):
...
+    """Add a single attribute field to the dataset.
+
+    Args:
+        dataset: the dataset
+        attr_path: the full path for the attribute
+        attr_type: the attribute type string
+        type_to_ftype: mapping of type strings to field types
+    """
+    # Skip if field already exists
+    if dataset.get_field(attr_path) is not None:
```
kaixi-wang: what if it's a different type?

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/GUIContent/PrimitiveFieldContent.tsx
```diff
@@ -0,0 +1,214 @@
...
+        <div onBlur={() => handleBlur("range")}>
+          <RangeInput
+            range={range}
+            onRangeChange={handleRangeChange}
+            error={touched.range ? errors.range : null}
+            largeLabels={largeLabels}
+          />
+        </div>
+      )}
+    </Stack>
```
erik-nieh, line 210: @lanzhenw I am not sure if it is intentional because its not necessary for GA but preview from this PR is missing: https://github.com/voxel51/fiftyone/commit/39d120a8da3605e4bdb3fa9a9dee78d63eff01bc

erik-nieh: My mistake I forgot that preview was not in schema only in field creation. I see it in the other pr!

PR #6777, opened 2026-01-20
Add rerunnable as part of delegation

CamronStaley:

When a pipeline is delegated we should resolve the rerunnable so that we don't have to do that on the backend.

unit tests

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Improvements**
  * Pipelines now auto-populate stage rerunnable status from operator configurations during validation, improving execution consistency and adding warnings when an operator cannot be resolved.
* **Tests**
  * Added a test verifying per-stage rerunnable resolution; note: the same test scenario was duplicated in the test suite.

### DIFF
--- a/fiftyone/operators/executor.py
+++ b/fiftyone/operators/executor.py
@@ -311,7 +311,8 @@ async def execute_or_delegate_operator(
                             stage.rerunnable = child.config.rerunnable
                         else:
                             logger.warning(
-                                f"Pipeline stage operator '{stage.operator_uri}' does not exist"
+                                "Pipeline stage operator '%s' does not exist"
+                                % stage.operator_uri
                             )
             except Exception as e:
                 return ExecutionResult(


fiftyone/operators/executor.py
```diff
@@ -303,6 +303,17 @@ async def execute_or_delegate_operator(
                     raise TypeError(
                         "Pipeline stages must be of type PipelineStage"
                     )
+                registry = OperatorRegistry()
+                for stage in pipeline.stages:
+                    if stage.rerunnable is None:
+                        child = registry.get_operator(stage.operator_uri)
+                        if child:
+                            stage.rerunnable = child.config.rerunnable
```
kaixi-wang, line 311: If a pipeline stage just refers to an operator, how does resolving the operator from the registry give a different value. Does runnable get set/removed at runtime somewhere?

CamronStaley: The pipeline stage refers to the operator if the pipeline stage config rerunnable value is none. Basically if you set the rerunnable stage (defaults to None) it is used, if you didn't, then we pull the child's rerunnable value (which default to True).

CamronStaley: https://github.com/voxel51/fiftyone/blob/5688b006cde9919426192ad0db4329809ac96df0/fiftyone/operators/_types/pipeline.py#L35

Here is where it's defined on the pipeline stage

fiftyone/operators/executor.py
```diff
@@ -303,6 +303,17 @@ async def execute_or_delegate_operator(
                     raise TypeError(
                         "Pipeline stages must be of type PipelineStage"
                     )
+                registry = OperatorRegistry()
+                for stage in pipeline.stages:
+                    if stage.rerunnable is None:
+                        child = registry.get_operator(stage.operator_uri)
```
swheaton, line 309: should we wrap this call like we discussed? what if operator isn't available here

PR #6779, opened 2026-01-21
[HA] Make `attributes` a list for label schemas

benjaminpkane:

Convert `attributes` setting to a list for label schemas to support ordering

New label schema validation tests

-   [ ] Other

## Summary by CodeRabbit

* **Refactor**
  * Converted label schema attributes from a mapping to a list of named attribute objects, updating creation, validation, and update flows accordingly.

* **Documentation**
  * Updated examples and wording to reflect attributes-as-list format and normalized JSON-style quoting.

* **Tests**
  * Updated tests and added validation tests to match the list-based attributes structure and new validation scenarios.

* **Chores**
  * Added a new public NAME constant used by annotation schemas.


mcdoh (approved):
LGTM

fiftyone/core/annotation/constants.py
```diff
@@ -51,6 +51,7 @@
 CLASSES = "classes"
 COMPONENT = "component"
 DEFAULT = "default"
+NAME = "name"
```
lanzhenw, line 53: where is `name` used?

fiftyone/core/annotation/generate_label_schemas.py
```diff
@@ -155,84 +155,80 @@ def generate_label_schemas(sample_collection, fields=None, scan_samples=True):
...
-                'values': ['validation'],
-            },
-            'uniqueness': {
-                'type': 'float',
-                'component': 'slider',
-                'range': [0.15001302256126986, 1.0],
+                "component": "text",
+                "type": "detections",
             },
+            "tags": {"type": "list<str>", "component": "text"},
```
lanzhenw, line 228: tags no longer have values? This isn't actual change in schema right?

fiftyone/core/annotation/constants.py
benjaminpkane: It's used in a couple places, validate_label_schemas.py, mixins.py

fiftyone/core/annotation/generate_label_schemas.py
benjaminpkane: I changed the example to use `scan_samples=False` so the values are no longer present

fiftyone/core/annotation/validate_label_schemas.py
```diff
@@ -453,14 +453,20 @@ def _raise_unknown_setting_error(name, field_name):
...
 def _validate_attribute(
-    attribute,
     class_name,
     collection,
     field_name,
     label_schema,
     path,
     subfields,
 ):
+    label_schema = dict(**label_schema)
```
exupero: [nit] I presume the intention here is to copy the label_schema to avoid mutating the input? If so, `.copy()` communicates that more explicitly:
```suggestion
    label_schema = label_schema.copy()
```

fiftyone/core/annotation/validate_label_schemas.py
```diff
@@ -496,9 +502,9 @@ def _validate_attribute(
 
 
 def _validate_attributes(collection, field_name, class_name, attributes):
-    if not isinstance(attributes, dict):
+    if not isinstance(attributes, list):
         raise ValueError(
-            f"'attributes' setting for field '{field_name}' must be a 'dict'"
+            f"'attributes' setting for field '{field_name}' must be a 'list'"
         )
```
exupero: [nit] the type doesn't need to be quoted (it's not on line 581 below):
```suggestion
            f"'attributes' setting for field '{field_name}' must be a list"
```

fiftyone/core/odm/mixins.py
```diff
@@ -2059,14 +2059,18 @@ def _change_nested_label_schema(
...
-        return
+            break
 
-    if new_path is None:
-        del attributes[name]
-    else:
-        new_name = new_path.split(".")[-1]
-        attributes[new_name] = attributes.pop(name)
+    if delete_index is not None:
+        attributes.pop(delete_index)
```
exupero: Do we support a version of Python that chokes on deleting an item from a list while iterating over it? I tried on 3.10 and didn't have a problem.
```python
    for idx, attribute in enumerate(attributes):
        if name == attribute[foac.NAME]:
            if new_path is None:
                attributes.pop(idx)
            else:
                new_name = new_path.split(".")[-1]
                attribute[foac.NAME] = new_name

            return
```
If there is a reason to defer the deletion until after the loop, we should have a test of that behavior to make sure no one refactors the logic away.

tests/unittests/annotation/validate_label_schemas_tests.py
```diff
@@ -812,14 +818,81 @@ def test_validate_detections_label_field_schema(self):
...
+                        {
+                            "name": "confidence",
+                            "component": "text",
+                            "type": "float",
+                        },
+                    ],
+                    "type": "detection",
+                },
+                fields="detection",
+            )
```
exupero, line 857: This looks like 3 separate tests?

fiftyone/core/odm/mixins.py
benjaminpkane: Returning in the loop is cleaner perhaps. Will update

tests/unittests/annotation/validate_label_schemas_tests.py
benjaminpkane: Multiple assertions for one test is just what I'm used to 😅

PR #6789, opened 2026-01-21
BE Support annotating patches views via SampleFieldRoute

kaixi-wang:

FOEPD-2951
Update the SampleFieldRoute to handle updates coming from patches views such that both the src and generated datasets are updated. For deletes, the patch sample is deleted.

- Added tests

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Multi-operation JSON Patch for sample fields with optional sync to generated datasets; patch endpoints now return full sample payloads and ETags.
  * New explicit root-delete signaling (public exception) to let callers handle root-level deletions.

* **Bug Fixes**
  * Stronger field/path validation, clearer 404s, improved If-Match/ETag semantics, datetime patch support, and better patch error reporting/logging.
  * Dataset lookup accepts either dataset ID or name.

* **Tests**
  * Expanded unit tests for patching, deletion, sync, and root-delete scenarios.


fiftyone/server/routes/sample.py
```diff
@@ -476,78 +487,144 @@ class SampleField(HTTPEndpoint):
...
+                    path,
+                    field_id,
+                    data,
+                    delete=is_delete,
+                )
+                if generated_sample is not None:
+                    return utils.json.JSONResponse(
+                        utils.json.serialize(generated_sample),
+                        headers={"ETag": etag},
+                    )
```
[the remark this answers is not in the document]
kaixi-wang: Generated datasets would not be loaded by another user to edit as a new dataset is created upon load with a name not exposed to the user (unless loading a saved view). It is only important to have consistency and concurrency checks for the src.

fiftyone/server/utils/json/jsonpatch/methods.py
```diff
@@ -176,8 +182,8 @@ def copy(
     from_pointer = to_json_pointer(from_)
 
     value = get(src, from_pointer)
-    add(src, pointer, value)
-    return src
+
+    return add(src, pointer, value)
```
kaixi-wang, line 186: modified these because add does not always modify the src (in the case add to root) so we need to use the returned src

tests/unittests/server/utils/jsonpatch/test_jsonpatch_methods.py
```diff
@@ -368,62 +367,39 @@ def test_ok(person: Person):
...
-    src = mock.Mock()
-
-    with (
-        mock.patch.object(methods, "get") as mock_get,
-        mock.patch.object(methods, "remove") as mock_remove,
-        mock.patch.object(methods, "add") as mock_add,
-    ):
-        pointer = jsonpointer.JsonPointer("/a/b/c")
-        from_pointer = jsonpointer.JsonPointer("/d/e/f")
+    src = {"a": {"b": {}}, "d": {"e": {"f": "value"}}}
```
kaixi-wang, line 384: using real values to confirm results and not just calls. I can switch back, but I think this will catch more bugs if they arise

fiftyone/server/routes/sample.py
```diff
@@ -476,78 +488,144 @@ class SampleField(HTTPEndpoint):
...
+                sample_id,
+                dataset_id,
+            )
             raise HTTPException(
                 status_code=400, detail="Invalid If-Match header"
             )
 
-        sample = get_sample(dataset_id, sample_id, if_last_modified_at)
+        # Skip checking the last_modified_at on the src when editing through a generated dataset as the
+        # timestamp reflect the generated sample's last_modified_at, not the source sample's timestamp.
```
CamronStaley: so are we just not using last modified at for generated datasets? I thought we needed that for consistency or is a generated dataset unlikely to have two people modifying a sample? I am not super familiar with generated datasets just curious

fiftyone/server/routes/sample.py
```diff
@@ -476,78 +488,144 @@ class SampleField(HTTPEndpoint):
...
+            except Exception:
+                logger.exception(
+                    "Failed to sync to generated dataset `%s` associated with dataset `%s`",
+                    generated_dataset_name,
+                    dataset_id,
+                )
 
         return utils.json.JSONResponse(
-            utils.json.serialize(updated_field), headers={"ETag": etag}
+            utils.json.serialize(sample), headers={"ETag": etag}
```
CamronStaley, line 638: returning the sample rather than the updated fields seems like a change in behavior. Originally we did return the entire sample but were told its too much data transfer and they just want the updated fields. Have they communicated otherwise now?

fiftyone/server/utils/datasets.py
```diff
@@ -53,3 +72,85 @@ def get_sample_from_dataset(dataset, sample_id: str):
...
+        # Update single label fields directly
+        target = field_value
+
+    # Apply patches - RootDeleteError would have been caught by caller
+    try:
+        _, errors = apply_jsonpatch(
+            target, patch_operations, transform_fn=deserialize
+        )
+    except RootDeleteError:
+        # Should not reach here if delete flag was passed correctly
```
CamronStaley: so why doesn't the caller just catch and swallow that error and then not pass a delete flag and allow this one to hit the error and handle it? Just seems weird that the caller has to handle that logic if we could catch it here. Perhaps you're avoiding a side effect of running the code above this that I am unaware of

fiftyone/server/routes/sample.py
kaixi-wang: every time you click the to_patches() button in the app, it'll regenerate a new dataset if you haven't loaded one in the current session (I think there's some FE relay passing of the dataset name in the stage kwargs so that it doesn't just continuously regenerate,. There's no backend handling that would resolve the same non-saved_view patches view for multiple users or even multiple local client sessions.

fiftyone/server/routes/sample.py
kaixi-wang: I think I needed this for patches samples to update correctly... I would run to test, but save on the FE is completely broken right now

CamronStaley: Okay I don't see any FE changes in this so I am really curious if this is a breaking change assuming they're using it currently for regular samples.

fiftyone/server/utils/datasets.py
kaixi-wang: Oops I refactored the code this should be removed

fiftyone/server/routes/sample.py
kaixi-wang: As far as I can tell, they were never using it (only one place calls patchSample and label id is never included to construct the path).
https://github.com/voxel51/fiftyone/blob/develop/app/packages/annotation/src/util/labelPersistence.ts#L47

kaixi-wang: Also, I think I may have had issues with the etag/grid updating when not passing the sample

PR #6801, opened 2026-01-23
fix saving and deleting labels

mcdoh:

Fix saving/deleting labels.

https://github.com/user-attachments/assets/3bca8974-16e1-4242-b94f-31caa9dc2367

https://github.com/user-attachments/assets/2e92292b-0f5c-44fe-ad62-74d532bca866

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Support for primitive label types and a unified label-proxy used across annotation operations.
  * Added constants to represent primitive label categories.

* **Bug Fixes**
  * Improved upsert and delete flows with stronger conversion and error handling for various annotation types.
  * Added 2D detection bounding-box overlay support so detection annotations include relative bounds.

### DIFF
--- a/app/packages/annotation/src/hooks/useRegisterAnnotationCommandHandlers.ts
+++ b/app/packages/annotation/src/hooks/useRegisterAnnotationCommandHandlers.ts
@@ -136,18 +136,7 @@ export const useRegisterAnnotationCommandHandlers = () => {
       async (cmd) => {
         try {
           const labelId = cmd.label.data._id;
-          const labelProxy = convertToLabelProxy(cmd.label);
-
-          if (!labelProxy) {
-            eventBus.dispatch("annotation:deleteError", {
-              labelId,
-              type: "delete",
-              error: new Error("Failed to convert label to proxy"),
-            });
-            return false;
-          }
-
-          const success = await deleteLabel(labelProxy, cmd.schema);
+          const success = await deleteLabel(cmd.label, cmd.schema);
 
           if (success) {
             eventBus.dispatch("annotation:deleteSuccess", {


app/packages/annotation/src/hooks/useRegisterAnnotationCommandHandlers.ts
```diff
@@ -8,10 +8,70 @@ import {
...
+      };
+    }
+
+    // For 3D detections, no bounding box needed
+    return {
+      type: DETECTION,
+      data: annotationLabel.data,
+      path: annotationLabel.path,
+    };
+  } else if (annotationLabel.type === CLASSIFICATION) {
```
erik-nieh, line 55: I here think you can just 
```
return annotationlabel
``` 
type/data/path all come from it, I think they are valid LabelProxy as is

app/packages/utilities/src/index.ts
```diff
@@ -152,6 +152,8 @@ export const KEYPOINT = "Keypoint";
 export const KEYPOINTS = "Keypoints";
 export const POLYLINE = "Polyline";
 export const POLYLINES = "Polylines";
+export const PRIMITIVE = "Primitive";
+export const PRIMITIVES = "Primitives";
```
erik-nieh, line 156: I think unused in this PR, possibly it's imported/used somewhere else you were fixing that was broken

app/packages/annotation/src/hooks/useRegisterAnnotationCommandHandlers.ts
erik-nieh: also if this is true, I think you can just test if `overlay instanceof BoundingBoxOverlay`, make the change, and otherwise just return `label` as is for everything except Detection

app/packages/annotation/src/hooks/useRegisterAnnotationCommandHandlers.ts
```diff
@@ -60,9 +134,20 @@ export const useRegisterAnnotationCommandHandlers = () => {
     DeleteAnnotationCommand,
     useCallback(
       async (cmd) => {
-        const labelId = cmd.label.data._id;
         try {
-          const success = await deleteLabel(cmd.label, cmd.schema);
+          const labelId = cmd.label.data._id;
+          const labelProxy = convertToLabelProxy(cmd.label);
```
tom-vx51: this shouldn't be needed for delete; `boundingBox` is the only difference and it's ignored in the deletion case

app/packages/utilities/src/index.ts
mcdoh: Constant values in FiftyOne are currently a mess. While `./utilities/src/constants.ts` is a better location for these values doing so incurs further changes. I just wanted to establish values for `"Primitive"` and `"Primitives"` (currently unused) so that we can hopefully move towards using one collection of constants.

PR #6807, opened 2026-01-26
[HA] Add a new field (label/primitive) in schema manager

lanzhenw:

FOEPD-2546: HA M4: As a dataset manager, I want to create a new label field in the Schema Manager UI
FOEPD-2790: HA M4:  As a dataset manager, I want to create a new Primitive field in the Schema Manager UI
FOEPD-2793: HA M4: As a dataset manager, I want the option to "activate" (or not) the field I'm editing/creating in the Annotation Schema

- updated `dataset.update_label_schema()` to also takes `allow_new_field` and `allow_new_attrs` params

A user can create a new label field or primitive field with label schema configured in schema manager

https://github.com/user-attachments/assets/cdcc53e0-5395-42b3-af38-ea3d5d4de6c5

        notes for FiftyOne users.

When configuring their dataset’s annotation schema, dataset managers should have the option of creating a new label field for their dataset and annotation schema directly from the Schema Manager UI. 

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added ability to create new label and primitive fields through a dedicated creation workflow.
  * Enhanced field editor with improved primitive value rendering and validation.
  * Implemented new field activation to automatically add created fields to active schemas.
  * Extended schema validation to support optional new attributes and fields during annotation.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/NewFieldSchema/NewFieldSchema.tsx
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/NewFieldSchema/NewFieldSchema.tsx
@@ -20,7 +20,7 @@ import {
   TextVariant,
   ToggleSwitch,
 } from "@voxel51/voodo";
-import { useCallback, useMemo, useRef, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import {
   ATTRIBUTE_TYPE_OPTIONS,
   CATEGORY_LABEL,
@@ -77,14 +77,13 @@ const NewFieldSchema = () => {
   const currentMediaType = useMediaType();
   const is3dMedia = !!(currentMediaType && is3d(currentMediaType));
 
-  // Make sure 3D Detection attributes initialize correctly
-  const initializedRef = useRef(false);
-  if (!initializedRef.current && currentMediaType !== null) {
-    initializedRef.current = true;
+  // Initialize correct attributes for 3D media
+  useEffect(() => {
     if (is3dMedia) {
       setAttributes(getDefaultAttributesForType(labelType, true));
     }
-  }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [is3dMedia]);
 
   // Get label type options based on media type
   const labelTypeOptions = useMemo(


fiftyone/core/annotation/validate_label_schemas.py
```diff
@@ -434,11 +434,16 @@ def _validate_attribute(
     path,
     subfields,
 ):
+    # Check both field.fields (in-memory) and direct schema lookup.
```
kaixi-wang: Can you help me understand what the intention of `in-memory` here is?

plugins/operators/annotation.py
```diff
@@ -357,3 +357,146 @@ def execute(self, ctx):
                 errors.append(str(exception))
 
         return {"errors": errors}
+
+
+def _add_base_label_subfields(dataset, base_path):
+    """Add base subfields shared by all label types (id, label, tags, confidence)."""
+    dataset.add_sample_field(f"{base_path}.id", fof.ObjectIdField)
```
kaixi-wang: Is there something weird with the schema manager UI that requires this? Without full context, it feels like this should be handled automatically by the sdk by using the label class. Default fields do not seem like something a plugin should be manually needing to add?

plugins/operators/annotation.py
```diff
@@ -357,3 +357,146 @@ def execute(self, ctx):
...
+
+    This adds the standard Detection/Classification subfields to the dataset's
+    data schema so that the label schema validation passes.
+    """
+    if field_type == "detections":
+        # Detections has a nested 'detections' list of Detection objects
+        base_path = f"{field_name}.detections"
+        _add_base_label_subfields(dataset, base_path)
+        # Detection also has 'index'
+        dataset.add_sample_field(f"{base_path}.index", fof.IntField)
```
kaixi-wang: same comment as above. This feels like it would be really easy to break if new functionality/attributes are added to the sdk

plugins/operators/annotation.py
```diff
@@ -357,3 +357,146 @@ def execute(self, ctx):
...
+        ctx.dataset.update_label_schema(field_name, label_schema)
+
+        # Activate the field (prepend to make it appear at top)
+        active = ctx.dataset.active_label_schemas or []
+        ctx.dataset.active_label_schemas = [field_name] + [
+            f for f in active if f != field_name
+        ]
+
+        ctx.dataset.save()
+        ctx.trigger("reload_dataset")
```
kaixi-wang: this seems like it would make adding multiple fields pretty tedious/slow

fiftyone/core/annotation/validate_label_schemas.py
lanzhenw: good catch. should always use `collection.get_field()`

plugins/operators/annotation.py
lanzhenw: Product requirements require the newly added field to immediately show up on schema manager after creation. 
Is there other way to update the data schema?

plugins/operators/annotation.py
lanzhenw: The problem:  
When we call
```
dataset.add_sample_field("my_detections", fof.EmbeddedDocumentField, embedded_doc_type=fol.Detections)  
```
Only the top-level field is created in the dataset schema. The nested subfields (id, label, confidence, index, tags) are defined on the Detection Python class, but they don't exist in the dataset's schema yet.  

When do nested fields get added normally?  
When we store actual data:
```
  sample.my_detections = fo.Detections(detections=[fo.Detection(label="cat", confidence=0.9)])
  sample.save()  
```
At this point, MongoDB sees the structure and the SDK adds the nested fields to the schema.  

Why is this a problem for Schema Manager?  We want to create a field AND set up its label schema before any data exists. We don't users to mess up with the default attributes, but from UX perspective, it makes sense to let the users see these default attributes in the attribute list.  

That's why we manually add them:  
```
  dataset.add_sample_field("my_detections.detections.id", fof.ObjectIdField)
  dataset.add_sample_field("my_detections.detections.confidence", fof.FloatField)
```
They need to exist in the data schema because label schema gets validated against data schema.

plugins/operators/annotation.py
lanzhenw: removed it.

plugins/operators/annotation.py
lanzhenw: took it offline. Could accept adding the hardcoded defaults temporarily. The good solution should be have separate sdk methods to handle default label attributes.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/utils.ts
```diff
@@ -356,6 +368,123 @@ export interface AttributeFormErrors {
...
+      const invalidDefault = data.listDefault.find(
+        (d) => !data.values.includes(String(d))
+      );
+      if (invalidDefault !== undefined) {
+        errors.default = "All defaults must be from the provided values";
+      }
+    }
+  }
+
+  return errors;
```
erik-nieh: small syntax error here, needs closing bracket

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -7,6 +7,42 @@ import { useAtom, useAtomValue, useSetAtom } from "jotai";
...
+        attrs
+          .filter((attr) => attr && typeof attr === "object" && "name" in attr)
+          .map((attr) => attr.name)
+      );
+    }
+  }
+  return new Set();
+};
+
 import { getAttributeNames, hasAttributes, isNamedAttribute } from "../utils";
```
erik-nieh, line 10: @lanzhenw I think there might be some merge issues here

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/hooks.ts
```diff
@@ -508,3 +526,23 @@ export const useFullSchemaEditor = () => {
...
+export const useNewFieldMode = () => {
+  const [isNewField, setIsNewField] = useAtom(isNewFieldMode);
+  return { isNewField, setIsNewField };
+};
+
+/**
+ * Hook to exit new field mode (convenience hook)
+ */
+export const useExitNewFieldMode = () => {
+  const setNewFieldMode = useSetAtom(isNewFieldMode);
```
erik-nieh, line 546: can just use the hook above too

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/state.ts
```diff
@@ -20,6 +20,15 @@ import { isSystemReadOnlyField } from "./constants";
...
 export { currentField } from "../state";
 
+// =============================================================================
+// New Field Mode State
+// =============================================================================
+
+/**
+ * Whether we're in "add new field" mode
+ */
+export const isNewFieldMode = atom(false);
```
erik-nieh, line 30: it might just be better to put this atom in the hooks file that uses it and not export it. this would enforce to some degree people using the hooks rather than the atom directly

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/utils.ts
```diff
@@ -356,6 +368,123 @@ export interface AttributeFormErrors {
   default: string | null;
 }
 
+/**
+ * Validate attribute form data and return field-specific errors.
+ * Used for both UI display and canSave logic.
+ */
+export const getAttributeFormErrors = (
```
erik-nieh: I think this can be removed? or the other declaration modified with new behavior?

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
lanzhenw: 😭 Yeah, I realized that. I will go through and do a clean up.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/utils.ts
lanzhenw: yes, this is the merge conflict issue.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/state.ts
lanzhenw: I'd prefer to keep all atoms in state.ts for consistency - it gives us a single source of truth for state definitions and makes it easier to find/audit all atoms in the schema manager.

plugins/operators/annotation.py
lanzhenw: Update: removed these code, since validation will support `allow_new_attrs` and `allow_new_field` in https://github.com/voxel51/fiftyone/pull/6835

plugins/operators/annotation.py
```diff
@@ -179,3 +219,123 @@ def execute(self, ctx):
...
+            # Validate and set the label schema (with flags for new fields/attrs)
+            validate_label_schemas(
+                ctx.dataset,
+                {field_name: label_schema},
+                allow_new_attrs=True,
+                allow_new_fields=True,
+            )
+            label_schemas = ctx.dataset.label_schemas
+            label_schemas[field_name] = copy.deepcopy(label_schema)
+            ctx.dataset._doc.label_schemas = label_schemas
```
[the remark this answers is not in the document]
kaixi-wang: agree. try to stick to using public functions in operators when they exist

plugins/operators/annotation.py
```diff
@@ -6,15 +6,46 @@
...
+PRIMITIVE_TYPE_TO_FIELD = {
+    "int": fof.IntField,
+    "float": fof.FloatField,
+    "str": fof.StringField,
+    "bool": fof.BooleanField,
+    "date": fof.DateField,
+    "datetime": fof.DateTimeField,
+    "list<int>": fof.IntField,
+    "list<float>": fof.FloatField,
+    "list<str>": fof.StringField,
```
kaixi-wang: Can you re-use FIELD_TYPE_TO_TYPES  from fiftyone.core.annotation.constants.py? You can use dict comprehension to derive PRIMITIVE_TYPE_TO_FIELD for convenience to have key be the string name, but this way updating it in constants won't just break the operator if someone forgets.

plugins/operators/annotation.py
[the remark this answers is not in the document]
lanzhenw: implemented in the latest commit

fiftyone/core/annotation/constants.py
```diff
@@ -138,6 +138,18 @@
 }
 
 
+### Label schema type to field/label class (reverse mappings)
+
+
+# Reverse of FIELD_TYPE_TO_TYPES (type string -> field class)
+TYPE_TO_FIELD = {
```
kaixi-wang, line 146: non-blocking nit: this is kinda hard to understand what the intent here is (why altering). might be cleaner to use a helper method and flatten out the mapping above.

plugins/operators/annotation.py
```diff
@@ -182,3 +189,119 @@ def execute(self, ctx):
...
+            ctx.dataset.update_label_schema(
+                field_name,
+                label_schema,
+                allow_new_attrs=True,
+                allow_new_fields=True,
+            )
+
+            # Activate the field (prepend to make it appear at top)
+            active = ctx.dataset.active_label_schemas or []
+            ctx.dataset.active_label_schemas = [field_name] + [
```
kaixi-wang: is there a reason you can't/wouldn't use ctx.dataset.activate_label_schemas here?

plugins/operators/annotation.py
```diff
@@ -182,3 +189,119 @@ def execute(self, ctx):
                 errors.append(str(exception))
 
         return {"errors": errors}
+
+
+class CreateAndActivateField(foo.Operator):
```
kaixi-wang, line 194: Could we consolidate CreateAndActivateField and ActivateField with just ActivateField with a create=True parameter?

lanzhenw: - CreateAndActivateField: Creates a new field in the dataset schema, builds its label schema, then activates it to the top of the list
  - ActivateLabelSchemas: Just moves existing fields to the visible list

They correspond to two CTA and different UI flows. They need different parameters, since `ActivateLabelSchemas` only needs a list of field names. Combining them would mean most parameters are only relevant when `create=True`. The overlap is just a single line.

plugins/operators/annotation.py
lanzhenw: activate_label_schemas() would add the new field to the bottom, but we need it to be on the top. 

I can add a `prepend=True` and update `activate_label_schemas()` so it's re-usable here.

kaixi-wang (approved):

PR #6821, opened 2026-01-27
[HA] Add undo/redo support for primitive edits; integrate with auto-save

tom-vx51:

* Adds undo/redo support for primitive edits
* Integrates primitive edits with auto-save

local

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added staged mutation manager for transient sample edits.
  * Integrated undo/redo flow for primitive sample edits.

* **Refactor**
  * Switched value sourcing to the new mutation manager and removed the older value hook.
  * Built delta supplier to emit staged-change deltas from staged mutations.
  * Improved command execution flow with error handling.

* **Tests**
  * Updated tests to reflect changed initial execution behavior.

### DIFF
--- a/app/packages/commands/src/context/context.manager.test.ts
+++ b/app/packages/commands/src/context/context.manager.test.ts
@@ -160,7 +160,7 @@ describe("CommandContextManager", () => {
       new KeyboardEvent("keydown", { ctrlKey: true, key: "x" })
     );
     expect(cmdFn).toBeCalledTimes(1);
-    expect(execFn).toBeCalledTimes(0);
+    expect(execFn).toBeCalledTimes(1);
     expect(undoFn).toBeCalledTimes(0);
     expect(CommandContextManager.instance().getActiveContext().canUndo()).toBe(
       true
@@ -170,7 +170,7 @@ describe("CommandContextManager", () => {
     );
     CommandContextManager.instance().getActiveContext().undo();
     expect(undoFn).toBeCalledTimes(1);
-    expect(execFn).toBeCalledTimes(0);
+    expect(execFn).toBeCalledTimes(1);
     expect(CommandContextManager.instance().getActiveContext().canUndo()).toBe(
       false
     );
@@ -178,7 +178,7 @@ describe("CommandContextManager", () => {
       true
     );
     CommandContextManager.instance().getActiveContext().redo();
-    expect(execFn).toBeCalledTimes(1);
+    expect(execFn).toBeCalledTimes(2);
     expect(undoFn).toBeCalledTimes(1);
   });
 });
--- a/app/packages/commands/vitest.config.ts
+++ b/app/packages/commands/vitest.config.ts
@@ -1,4 +1,4 @@
-import { defineConfig } from "vitest/config.js";
+import { defineConfig } from "vitest/config";
 
 export default defineConfig({
   test: {


app/packages/commands/src/context/CommandContext.ts
```diff
@@ -324,9 +324,15 @@ export class CommandContext {
     }
     if (resolved) {
       const result = resolved.execute();
-      if (result && isUndoable(result)) {
-        this.actions.push(result as Undoable);
```
marcus-kestler, line 328: The original intent was that if the command returned an action, that was for undo only.  We didn't execute the action, because the command has it (since it returns it), and could execute it.
I'm fine with this approach too, just want to call out the behavior change.

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/PrimitiveEdit.tsx
```diff
@@ -51,9 +61,50 @@ export default function PrimitiveEdit({
...
+    [type, value]
+  );
+
+  // need to use a ref to access field value in command callback;
+  // command will run before the next render loop when `fieldValue` is updated.
+  const transientFieldValue = useRef<Primitive>(fieldValue as Primitive);
+
+  // undoable command which handles primitive edits
+  const editCommand = useCreateCommand(
+    KnownContexts.Modal,
```
marcus-kestler: I've got in flight changes that add a new context for annotations.  We'll need to move this to that context when that lands.  Can you add a //TODO: Move to annotation context?

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/PrimitiveEdit.tsx
```diff
@@ -51,9 +61,50 @@ export default function PrimitiveEdit({
...
+      setFieldValue(data as Primitive);
+
+      if (editCommand.enabled) {
+        editCommand.callback();
+      }
+    },
+    [editCommand]
+  );
 
   const persistData = useCallback(async () => {
```
erik-nieh, line 114: are we able to remove this & the manual save/discard?

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/PrimitiveEdit.tsx
```diff
@@ -51,9 +61,50 @@ export default function PrimitiveEdit({
...
+          );
+        },
+        // restore current value on undo
+        () => {
+          sampleMutationManager.stageMutation(path, value);
+        }
+      );
+    }, [path, sampleMutationManager, type, value]),
+    () => true
+  );
```
sashankaryal, line 100: the `DelegatingUndoable` callbacks don't properly capture their values
                                                                                          
Redo seems to read from `transientFieldValue.current` at execution time. After undoing, this ref still holds the last user-typed value, so redo skips intermediate states.
                                    
**expectation:**
action: original -> a -> ab

undo: ab -> a -> original
redo: original -> a -> ab

**reality:**
action: original -> a -> ab

undo: ab -> a -> original
_redo: original -> ab (reads stale ref?)_

We might want to capture values before / after edit at _creation time_ in `DelegatingUndoable` closure, so that each undoable has its own immutable snapshot of before/after values

app/packages/annotation/src/persistence/useSampleMutationManager.ts
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Mapping of sample path to transient values.
+ */
+export type MutationMap = Record<string, Primitive>;
+
+/**
+ * Centralized {@link MutationMap} for stability across hook instances.
+ */
+const stagedMutationsAtom = atom<MutationMap>({});
+
```
sashankaryal, line 16: nit: should this be sample scoped, too? hope we don't leak state across sample lifecycles

app/packages/commands/src/utils.ts
```diff
@@ -1,3 +1,12 @@
...
 }
+
+/**
+ * Return whether the provided object adheres to the {@link Action} interface.
+ *
+ * @param data Object to check
+ */
+export const isAction = (data: object): boolean => {
+  return "execute" in data && typeof data.execute === "function";
+};
```
sashankaryal: Should probably accept `unknown` and not `object` and return type guard: `data is Action`. That way caller isn't forced to pre-narrow

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/PrimitiveEdit.tsx
tom-vx51: yes, I was going to remove the manual save in the same PR where I unflag auto-save (coming soon)

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/PrimitiveEdit.tsx
sashankaryal: BTW I tested this with `filepath`. I tried new values using clipboard since I didn't want to use text edit (but you can repro it with that, too)

tom-vx51: > so that each undoable has its own immutable snapshot of before/after values

this was the intent. I thought I tested this but maybe changed something after, will take a look

app/packages/annotation/src/persistence/useSampleMutationManager.ts
tom-vx51: considered it, still can. Currently there's a `useEffect` which resets the state on sample id change

app/packages/commands/src/utils.ts
tom-vx51: i did this but linter complained. I'll figure out a fix

app/packages/commands/src/context/CommandContext.ts
tom-vx51: ah okay, I had inferred the intent from the `CommandFunction` type but understand now. This change feels more intuitive to me; I think the equivalent would otherwise be
```tsx
useCreateCommand(
  ...,
  () => {
    const undoable = new DelegatingUndoable(...);
    undoable.execute();
    return undoable;
  },
  ...
);
```

not terrible, but sort of blurs the line on what's responsible for execution

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/PrimitiveEdit.tsx
tom-vx51: i am dumb and broke it with a refactor

PR #6826, opened 2026-01-27
Update Enterprise Migration Docs

Anddraca:

Updates Enterprise migration documentation to reference the new Enterprise Migration tool.

n/a; no functional changes

        notes for FiftyOne users.

Updates Enterprise migration documentation to reference the new Enterprise Migration tool

-   [ ] Other

## Summary by CodeRabbit

* **Documentation**
  * Introduced Enterprise Migration Tool workflow (v2.14.0+), including installation and configuration
  * Added usage guidance with examples for per-service and full deployments, include/exclude controls, and version-specific migrations
  * Added bidirectional (migrate/revert) workflows and separate guidance for pre-tool downgrades
  * Updated commands to use the new tool, clarified environment prerequisites, warnings, admin workflow impacts, and post-upgrade SDK requirements

### DIFF
--- a/docs/source/enterprise/migrations.rst
+++ b/docs/source/enterprise/migrations.rst
@@ -164,12 +164,12 @@ provide the version which you want to restore.
 
 .. code-block:: shell
 
-    # Migrate from v2.12.0 to v2.13.0
-    fiftyone-migrator migrate 2.13.0
+    # Migrate from v2.14.0 to v2.15.0
+    fiftyone-migrator migrate 2.15.0
     
     # Oops, need to revert this migration!
-    # Migrate from v2.13.0 to v2.12.0
-    fiftyone-migrator migrate 2.12.0
+    # Migrate from v2.15.0 to v2.14.0
+    fiftyone-migrator migrate 2.14.0
 
 Downgrading your deployment without the Enterprise Migration Tool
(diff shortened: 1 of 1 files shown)


docs/source/enterprise/migrations.rst
```diff
@@ -85,15 +85,67 @@ must manually upgrade your database by installing the newest version of the
 FiftyOne Enterprise SDK locally, assuming admin privileges, and running the
 following command:
 
+Beginning with FiftyOne Enterprise `v2.15.0`, there is a new migration tool
```
tom-vx51: [nit] this is true for 2.14 (2.13? i've lost track of numbers) - only worth noting because 2.14 introduced a new data model for VAL which requires a migration provided by this tool

Anddraca: > [nit] this is true for 2.14 (2.13? i've lost track of numbers) - only worth noting because 2.14 introduced a new data model for VAL which requires a migration provided by this tool

Good catch. I can change the referenced version to 2.14.0 -- I had put it as 2.15.0 since that was the first time we introduced docs around it, but if it is relevant for users back to 2.14, it would be confusing as written

Anddraca: updated

PR #6830, opened 2026-01-28
[HA] Migrate annotation notifications to use voodo's activity toast

tom-vx51:

* Add support for using VOODO's `ActivityToast` component
* Integrate annotation notifications with `ActivityToast`

local

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Introduced activity toast notifications for improved user feedback. Success messages confirm when changes are saved, while error toasts alert users to issues. Toasts automatically dismiss after a brief duration for a streamlined experience.

### DIFF
--- a/app/packages/core/package.json
+++ b/app/packages/core/package.json
@@ -23,7 +23,7 @@
         "@mui/material": "^5.9.0",
         "@react-spring/web": "^9.4.3",
         "@use-gesture/react": "^10.3.0",
-        "@voxel51/voodo": "0.0.8",
+        "@voxel51/voodo": "0.0.10",
         "@xstate/react": "1.3.3",
         "classnames": "^2.2.6",
         "color": "^4.2.3",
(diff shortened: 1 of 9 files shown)


app/packages/core/package.json
```diff
@@ -23,7 +23,7 @@
         "@mui/material": "^5.9.0",
         "@react-spring/web": "^9.4.3",
         "@use-gesture/react": "^10.3.0",
-        "@voxel51/voodo": "0.0.8",
+        "@voxel51/voodo": "0.0.10",
```
lanzhenw, line 26: let's just use *. I modified it to ^number earlier and regretted.

erik-nieh: @lanzhenw  Is this just because of the sort of rapid changes? or do you think it should be done more permanently

tom-vx51: The problem with `*` is that you lose control over the versioning. It works okay for our other packages because they're all local and so the latest version is always used. With packages that come from npm, `*` just indicates *any* version, so e.g. if you have an old version in your `node_modules`, it will always just use that rather than checking for updates. 

We don't specify a version in the voodo docs, but it works in that case because we explicitly do `npm update @voxel51/voodo@latest` in our build. 

I agree it's a bit annoying when we're quickly iterating on versions, but I don't think we can drop versioning for this package unfortunately.

benjaminpkane (approved):
✨

PR #6845, opened 2026-01-29
Read Only Primitives

erik-nieh:

1. Changes Read Only primitives to render Label/Text:

image

2. Changes read only for primitives so it respects if ANY of the following are true:
- field is read only
- the primitive is a child of a label and that label is read only (so ALL primitives on that label will be read only)
- the primitive itself is set as read only

3. If there is ever a case where we cannot determine based on `type`/`component` what we are supposed to render, we render read-only (instead of undefined) - this fixes a bug/edge case.

Tested locally

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added a read-only label display for showing immutable field values in forms and annotations.
  * New compact widget for presenting a field label and its value as plain text.

* **Improvements**
  * Support for per-field read-only flags in schemas to prevent editing of specific fields.
  * Enhanced read-only handling across schema processing and annotation forms.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/AnnotationSchema.tsx
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/AnnotationSchema.tsx
@@ -28,7 +28,7 @@ function updateSchemaReadOnly(schema: SchemaType, readOnly: boolean) {
     return { ...schema, readOnly };
   }
   // if parent/field readOnly is false don't override the
-  // field's readOnly flag
+  // primitive's readOnly flag
   return { ...schema, readOnly: schema.read_only };
 }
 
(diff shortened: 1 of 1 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/AnnotationSchema.tsx
```diff
@@ -14,23 +15,48 @@ import {
...
+  // if parent/field readOnly is false don't override the
+  // field's readOnly flag
+  return { ...schema, readOnly: schema.read_only };
+}
+
 const useSchema = (readOnly: boolean) => {
   const config = useAtomValue(currentSchema);
+  const isLabelReadOnly = config.read_only;
+  // respect either the field OR the parent schema's readOnly flag
+  const effectiveReadOnly = readOnly || isLabelReadOnly;
```
mcdoh, line 39: I'm thinking one atom to cover this would be ideal but probably best to mark for a refactor later.

PR #6850, opened 2026-01-30
Add YOLOE-26 open-vocabulary segmentation models to zoo

CharlesCNorton:
## Summary

Adds YOLOE-26 open-vocabulary segmentation: 5 visual-prompt variants (`FiftyOneYOLOEVPModel`) and 5 prompt-free variants (`FiftyOneYOLOModel`, 4,585 built-in classes).

Visual-prompt path: `YOLOEVPGetItem` builds the `visual_prompts` payload during preprocessing; `FiftyOneYOLOEVPModel._predict_all_visual_prompts` runs per-image inference and resets the predictor via `self._set_predictor`; `YOLOEVPSegmentationOutputProcessor` handles the per-image class-name remap and `to_instances` conversion.

### Visual-prompt
| Model | Size |
|---|---|
| `yoloe26n-seg-torch` | 11 MB |
| `yoloe26s-seg-torch` | 29 MB |
| `yoloe26m-seg-torch` | 67 MB |
| `yoloe26l-seg-torch` | 75 MB |
| `yoloe26x-seg-torch` | 163 MB |

### Prompt-free (4,585 classes)
| Model | Size |
|---|---|
| `yoloe26n-seg-pf-torch` | 15 MB |
| `yoloe26s-seg-pf-torch` | 34 MB |
| `yoloe26m-seg-pf-torch` | 72 MB |
| `yoloe26l-seg-pf-torch` | 80 MB |
| `yoloe26x-seg-pf-torch` | 167 MB |

## Usage

```python
import fiftyone as fo
import fiftyone.zoo as foz

model = foz.load_zoo_model("yoloe26m-seg-torch")
dataset.apply_model(model, label_field="vp_preds", prompt_field="ground_truth")

pf_model = foz.load_zoo_model("yoloe26m-seg-pf-torch")
dataset.apply_model(pf_model, label_field="pf_preds")

```

### Unit tests (`tests/unittests/utils/test_ultralytics.py`, 28 tests)

- `_detections_to_visual_prompts`: normalized→absolute conversion, multi-class dedup ordering, source non-mutation.
- `YOLOEVPGetItem`: required keys, transform contract, force-RGB load, `visual_prompts`/`vp_classes` emission, None/empty prompt fall-through.
- `FiftyOneYOLOEVPModel.collate_fn`: batch assembly; missing-key default.
- Dispatch: VP path on non-empty prompts; fall-through on all-None, empty-list, and non-dict inputs.
- `_predict_all_visual_prompts`: per-image predict kwargs, confidence threshold (including the falsy-zero case), predictor restore on success and on exception, `strict=True` zip length-mismatch behavior.
- `YOLOEVPSegmentationOutputProcessor`: per-image names_map remap, fall-through when `vp_classes` is None.
- `_get_yoloe_vp_predictor`: real-import success, `AttributeError` with chained `__cause__` when missing.

### Inference parity vs direct ultralytics

Verified on `yoloe26n-seg-torch` (50 quickstart samples), `yoloe26s-seg-torch` (10), `yoloe26m-seg-torch` (5). With predictor state reset between consecutive predict calls (which the wrapper does via `_set_predictor` after every batch), the wrapper produces `fol.Detections` bit-identical to `ultralytics.YOLO(...).predict(visual_prompts=..., predictor=YOLOEVPSegPredictor, ...)`: labels, normalized bboxes, confidences, mask shapes, mask sums, and mask SHA-1 hashes all match. Without that reset, the upstream `YOLOEVPSegPredictor` accumulates state across calls and shifts a few near-threshold detections in larger runs.

Wrapper determinism verified: two consecutive `apply_model` runs on the same dataset produce byte-identical detections.

- [x] Visual-prompt model loads and segments per the supplied prompt field
- [x] Prompt-free model loads and segments from built-in vocabulary
- [x] Mixed batches (some samples with prompts, some without) dispatch correctly
- [x] All 10 manifest entries resolve to the correct model class and output processor

## Summary by CodeRabbit

* **New Features**
  * Adds per-image visual-prompt (box) inference so items can include prompts to guide detection; prompt-aware results are mapped back into app-friendly prediction format and original image/shape context is preserved.

* **Tests**
  * Adds comprehensive unit tests validating prompt conversion, dataset-item behavior, batching/collation, prediction routing, predictor restoration, and error handling.

* **Chores**
  * Adds graceful fallback when visual-prompt support is unavailable.

### DIFF
--- a/fiftyone/utils/ultralytics.py
+++ b/fiftyone/utils/ultralytics.py
@@ -23,6 +23,7 @@ import fiftyone.zoo.models as fozm
 ultralytics = fou.lazy_import("ultralytics")
 torch = fou.lazy_import("torch")
 torchvision = fou.lazy_import("torchvision")
+_yoloe = fou.lazy_import("ultralytics.models.yolo.yoloe")
 
 
 def convert_ultralytics_model(model):
@@ -707,6 +708,171 @@ class FiftyOneYOLOModel(fout.TorchImageModel):
             classes=self.config.filter_classes,
         )
 
+class YOLOEVPGetItem(fout.GetItem):
+    """A :class:`GetItem` that loads images and box prompt detections for
+    :class:`FiftyOneYOLOEVPModel` instances.
+
+    Args:
+        transform (None): a callable that maps a ``PIL.Image.Image`` to a
+            dict containing ``"img"`` (a ``torch.Tensor``) and ``"orig_img"``
+            (a ``numpy.ndarray``) keys
+        field_mapping (None): the user-supplied dict mapping keys in
+            :attr:`required_keys` to field names of their dataset
+    """
+
+    def __init__(self, transform=None, field_mapping=None, **kwargs):
+        super().__init__(field_mapping=field_mapping, **kwargs)
+        self._transform = transform
+
+    def __call__(self, d):
+        img = fout._load_image(d["filepath"], use_numpy=False, force_rgb=True)
+        if self._transform is None:
+            raise TypeError(
+                "YOLOEVPGetItem requires a transform to produce tensors"
+            )
+        result = self._transform(img)
+
+        prompt = d["prompt_field"]
+        if prompt is not None and prompt.detections:
+            bboxes, cls_indices, classes = _detections_to_visual_prompts(
+                prompt, img.width, img.height
+            )
+            result["visual_prompts"] = {
+                "bboxes": np.array(bboxes),
+                "cls": np.array(cls_indices),
+            }
+            result["vp_classes"] = classes
+        else:
+            result["visual_prompts"] = None
+            result["vp_classes"] = None
+
+        return result
+
+    @property
+    def required_keys(self):
+        return ["filepath", "prompt_field"]
+
+
+class FiftyOneYOLOEVPModelConfig(FiftyOneYOLOModelConfig):
+    """Configuration for a :class:`FiftyOneYOLOEVPModel`."""
+
+    pass
+
+
+class FiftyOneYOLOEVPModel(FiftyOneYOLOModel):
+    """YOLOE model with visual prompt (box prompt) support.
+
+    Args:
+        config: a :class:`FiftyOneYOLOEVPModelConfig`
+    """
+
+    @staticmethod
+    def collate_fn(batch):
+        visual_prompts = [item.get("visual_prompts") for item in batch]
+        vp_classes = [item.get("vp_classes") for item in batch]
+        orig_images = [img.get("orig_img") for img in batch]
+        orig_shapes = [_get_image_dims(img)[::-1] for img in orig_images]
+        images = torch.stack([img.get("img") for img in batch])
+        return {
+            "orig_imgs": orig_images,
+            "images": images,
+            "orig_shapes": orig_shapes,
+            "visual_prompts": visual_prompts,
+            "vp_classes": vp_classes,
+        }
+
+    def _predict_all(self, imgs):
+        # Inputs from build_get_item + fout.get_model_inputs_from_get_item
+        # arrive as a list of already-transformed per-sample dicts. Collate
+        # them here rather than letting the parent's transform-then-collate
+        # path try to re-transform the dicts.
+        if (
+            isinstance(imgs, list)
+            and imgs
+            and isinstance(imgs[0], dict)
+            and "img" in imgs[0]
+        ):
+            imgs = self.collate_fn(imgs)
+
+        visual_prompts = (
+            imgs.get("visual_prompts") if isinstance(imgs, dict) else None
+        )
+        if visual_prompts and any(vp is not None for vp in visual_prompts):
+            return self._predict_all_visual_prompts(
+                imgs["orig_imgs"],
+                imgs["visual_prompts"],
+                imgs["vp_classes"],
+            )
+
+        return super()._predict_all(imgs)
+
+    def _predict_all_visual_prompts(
+        self, orig_images, visual_prompts_list, vp_classes_list
+    ):
+        vp_predictor_cls = _get_yoloe_vp_predictor()
+        default_args = self._model.predictor.args
+        default_rect = default_args.rect
+        default_retina_masks = default_args.retina_masks
+        default_conf = default_args.conf
+
+        all_labels = []
+        try:
+            if not (
+                len(orig_images)
+                == len(visual_prompts_list)
+                == len(vp_classes_list)
+            ):
+                raise ValueError(
+                    "orig_images, visual_prompts_list, and vp_classes_list "
+                    "must have equal length"
+                )
+            # YOLOEVPSegPredictor requires per-image visual_prompts dicts,
+            # so batched prediction is not supported here.
+            for orig_img, visual_prompts, classes in zip(
+                orig_images, visual_prompts_list, vp_classes_list
+            ):
+                if visual_prompts is None:
+                    all_labels.append(fol.Detections())
+                    continue
+
+                results = self._model.predict(
+                    orig_img,
+                    visual_prompts=visual_prompts,
+                    predictor=vp_predictor_cls,
+                    rect=default_rect,
+                    retina_masks=default_retina_masks,
+                    save=False,
+                    mode="predict",
+                    verbose=False,
+                    device=self._device,
+                    conf=self.config.confidence_thresh
+                    if self.config.confidence_thresh is not None
+                    else default_conf,
+                )
+
+                labels = self._output_processor(
+                    results,
+                    None,
+                    vp_classes=classes,
+                    confidence_thresh=self.config.confidence_thresh,
+                    classes=self.config.filter_classes,
+                )
+
+                if isinstance(labels, list):
+                    labels = labels[0] if labels else fol.Detections()
+
+                all_labels.append(labels)
+        finally:
+            self._set_predictor(self.config, self._model)
+
+        return all_labels
+
+    def build_get_item(self, field_mapping=None):
+        return YOLOEVPGetItem(
+            transform=self._transforms,
+            field_mapping=field_mapping,
+        )
+
 
 def _get_image_dims(img):
     if isinstance(img, torch.Tensor):
@@ -1015,6 +1181,45 @@ class UltralyticsSegmentationOutputProcessor(
         )
 
 
+class YOLOEVPSegmentationOutputProcessor(
+    UltralyticsSegmentationOutputProcessor
+):
+    """Output processor for YOLOE visual-prompt segmentation.
+
+    The VP code path produces ultralytics ``Result`` objects directly (no
+    raw-tensor postprocess step). When ``vp_classes`` is provided, this
+    processor remaps each ``Result.names`` to the prompt's first-seen unique
+    class names before delegating to ``to_instances``. When ``vp_classes`` is
+    not provided (i.e., the prompt-free path through the same model), this
+    processor falls through to the standard segmentation behavior.
+    """
+
+    def __call__(
+        self,
+        output,
+        frame_size,
+        confidence_thresh=None,
+        classes=None,
+        vp_classes=None,
+        **kwargs,
+    ):
+        if vp_classes is not None:
+            names_map = dict(enumerate(vp_classes))
+            for r in output:
+                r.names = names_map
+            return self._parse_output(
+                output, frame_size, confidence_thresh, classes
+            )
+
+        return super().__call__(
+            output,
+            frame_size,
+            confidence_thresh=confidence_thresh,
+            classes=classes,
+            **kwargs,
+        )
+
+
 class UltralyticsPoseOutputProcessor(
     fout.OutputProcessor, UltralyticsPostProcessor
 ):
@@ -1061,3 +1266,30 @@ class UltralyticsOBBOutputProcessor(
         return obb_to_polylines(
             preds, confidence_thresh=confidence_thresh, classes=classes
         )
+
+
+def _detections_to_visual_prompts(detections, img_width, img_height):
+    dets = detections.detections
+    boxes = np.array([d.bounding_box for d in dets])
+    boxes[:, 2] += boxes[:, 0]
+    boxes[:, 3] += boxes[:, 1]
+    boxes[:, 0] *= img_width
+    boxes[:, 2] *= img_width
+    boxes[:, 1] *= img_height
+    boxes[:, 3] *= img_height
+
+    labels = [d.label for d in dets]
+    classes = list(dict.fromkeys(labels))
+    label_to_idx = {c: i for i, c in enumerate(classes)}
+    cls_indices = [label_to_idx[label] for label in labels]
+
+    return boxes.tolist(), cls_indices, classes
+
+
+def _get_yoloe_vp_predictor():
+    try:
+        return _yoloe.YOLOEVPSegPredictor
+    except (AttributeError, ImportError) as e:
+        raise AttributeError(
+            "Visual prompts require ultralytics>=8.4.0 with YOLOE support"
+        ) from e
--- a/fiftyone/utils/ultralytics.py
+++ b/fiftyone/utils/ultralytics.py
@@ -809,12 +809,6 @@ class FiftyOneYOLOEVPModel(FiftyOneYOLOModel):
     def _predict_all_visual_prompts(
         self, orig_images, visual_prompts_list, vp_classes_list
     ):
-        vp_predictor_cls = _get_yoloe_vp_predictor()
-        default_args = self._model.predictor.args
-        default_rect = default_args.rect
-        default_retina_masks = default_args.retina_masks
-        default_conf = default_args.conf
-
         all_labels = []
         try:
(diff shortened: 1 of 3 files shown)


fiftyone/utils/ultralytics.py
```diff
@@ -610,6 +612,23 @@ def _forward_pass(self, imgs):
...
+        else:
+            self._curr_visual_prompts = None
+
+        return self._predict_all(imgs)
+
+    def _get_field(self):
+        if "prompt_field" in self.needs_fields:
+            return self.needs_fields["prompt_field"]
+
+        return next(iter(self.needs_fields.values()), None)
```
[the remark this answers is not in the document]
lorenzoriano: Agree with @coderabbit, why return an arbitrary field if what we want is "pompt_field" ?

fiftyone/utils/ultralytics.py
```diff
@@ -701,6 +723,67 @@ def _predict_all(self, imgs):
...
         )
 
+    def _predict_all_visual_prompts(self, imgs):
+        if self._preprocess and self._transforms is not None:
+            imgs = [self._transforms(img) for img in imgs]
+            if self.has_collate_fn:
+                imgs = self.collate_fn(imgs)
+
+        orig_images = imgs["orig_imgs"]
+        width_height = imgs["orig_shapes"]
```
lorenzoriano: this assumption already exists in the codebase, but is there a chance those keys don't exist?

lorenzoriano (changes requested):
This is not binding, but would you consider a separate class here? Something like:

```
class FiftyOneYOLOEVPModel(fout.TorchSamplesMixin, FiftyOneYOLOModel):
    """YOLOE model with visual prompt support."""
    
    def __init__(self, config):
        fout.TorchSamplesMixin.__init__(self)
        FiftyOneYOLOModel.__init__(self, config)
        self._curr_visual_prompts = None
    
    def predict_all(self, imgs, samples=None):
        ...
    def _predict_all_visual_prompts(self, imgs):
        ...
```

I think it'll be cleaner instead of adding the dual modality to FiftyOneYOLOModel

lorenzoriano (approved):

fiftyone/utils/ultralytics.py
```diff
@@ -1055,3 +1159,49 @@ def __call__(
...
+                    % type(value)
+                )
+            prompts.append(value)
+        else:
+            prompts.append(None)
+
+    return prompts
+
+
+def _detections_to_visual_prompts(detections, img_width, img_height):
```
manushreegangwar, line 1305: You can vectorize this https://github.com/voxel51/fiftyone/blob/c0c5f49b9326fc9505a1078afd43e3e078b5898d/fiftyone/utils/sam.py#L398

fiftyone/utils/ultralytics.py
```diff
@@ -1055,3 +1159,49 @@ def __call__(
         return obb_to_polylines(
             preds, confidence_thresh=confidence_thresh, classes=classes
         )
+
+
+def _parse_visual_prompts(samples, field_name):
+    prompts = []
+    for sample in samples:
```
manushreegangwar: This can be very slow. Use https://docs.voxel51.com/api/fiftyone.core.collections.html#fiftyone.core.collections.SampleCollection.values

fiftyone/utils/ultralytics.py
```diff
@@ -701,6 +701,110 @@ def _predict_all(self, imgs):
...
+    """YOLOE model with visual prompt support.
+
+    This subclass adds visual-prompt-based inference to YOLOE models,
+    allowing users to supply bounding box detections as prompts.
+
+    Args:
+        config: a ``FiftyOneYOLOModelConfig``
+    """
+
+    def __init__(self, config):
```
manushreegangwar: Leaning towards implementing this as a `GetItem` for box prompts similar to https://github.com/voxel51/fiftyone/pull/6371

manushreegangwar (changes requested):

CharlesCNorton:
@manushreegangwar @lorenzoriano All addressed!

fiftyone/utils/ultralytics.py
```diff
@@ -701,6 +701,131 @@ def _predict_all(self, imgs):
...
+    """A :class:`GetItem` that loads images and box prompt detections for
+    :class:`FiftyOneYOLOEVPModel` instances.
+
+    Args:
+        transform (None): a preprocessing transform to apply
+        field_mapping (None): the user-supplied dict mapping keys in
+            :attr:`required_keys` to field names of their dataset
+    """
+
+    def __init__(self, transform=None, field_mapping=None, **kwargs):
```
lorenzoriano, line 723: non blockintg nit: There should be more info on this variable (is it a callable?) and a type info, better if it specifies input and output type

fiftyone/utils/ultralytics.py
```diff
@@ -701,6 +701,131 @@ def _predict_all(self, imgs):
...
+        return {
+            "orig_imgs": orig_images,
+            "images": images,
+            "orig_shapes": orig_shapes,
+            "prompts": prompts,
+        }
+
+    def _predict_all(self, imgs):
+        if self._preprocess and self._transforms is not None:
+            transformed = [self._transforms(img) for img in imgs]
```
lorenzoriano: shouldn't this be handled by YOLOEVPGetItem, or at least reusing the same transform?

fiftyone/utils/ultralytics.py
```diff
@@ -1055,3 +1180,32 @@ def __call__(
...
+    classes = list(dict.fromkeys(labels))
+    label_to_idx = {c: i for i, c in enumerate(classes)}
+    cls_indices = [label_to_idx[label] for label in labels]
+
+    return boxes.tolist(), cls_indices, classes
+
+
+def _get_yoloe_vp_predictor():
+    try:
+        from ultralytics.models.yolo.yoloe import YOLOEVPSegPredictor
```
lorenzoriano: do you need a mid function import? Can you use a lazy import?

fiftyone/utils/ultralytics.py
```diff
@@ -701,6 +701,131 @@ def _predict_all(self, imgs):
...
+        orig_shapes = [_get_image_dims(img)[::-1] for img in orig_images]
+        images = torch.stack([img.get("img") for img in batch])
+        return {
+            "orig_imgs": orig_images,
+            "images": images,
+            "orig_shapes": orig_shapes,
+            "prompts": prompts,
+        }
+
+    def _predict_all(self, imgs):
```
lorenzoriano, line 796: There's a problem with this issue:  when preprocess=False (which apply_model always sets at models.py), \_predict_all skips the entire visual processing block and falls straight through to super()._predict_all. The prompts in
the collated dict are never read. _predict_all_visual_prompts is never called.  
You can test this with this unit test (AI generated):

```
class TestYOLOEVPPredictAllDataloaderPath(unittest.TestCase):

      def _make_collated_batch_with_prompts(self):
          import torch
          import numpy as np
          from PIL import Image as PILImage

          orig_img = PILImage.fromarray(np.zeros((100, 100, 3), dtype=np.uint8))
          img_tensor = torch.zeros(3, 100, 100)

          fake_prompt = MagicMock()
          fake_prompt.detections = [MagicMock()]

          return {
              "orig_imgs": [orig_img],
              "images": img_tensor.unsqueeze(0),
              "orig_shapes": [(100, 100)],
              "prompts": [fake_prompt],
          }

      def test_visual_prompts_dispatched_when_preprocess_false(self):
          from fiftyone.utils.ultralytics import FiftyOneYOLOEVPModel

          model = FiftyOneYOLOEVPModel.__new__(FiftyOneYOLOEVPModel)
          model._preprocess = False
          model._transforms = MagicMock()
          model._model = MagicMock()

          collated = self._make_collated_batch_with_prompts()

          vp_call_args = []

          def fake_vp(orig_images, width_height, prompts):
              vp_call_args.append((orig_images, width_height, prompts))
              return ["fake_label"]

          def fake_super_predict_all(self_or_imgs, imgs=None):
              return ["plain_label"]

          model._predict_all_visual_prompts = fake_vp

          import fiftyone.utils.ultralytics as fou_ul
          with patch.object(fou_ul.FiftyOneYOLOModel, "_predict_all", fake_super_predict_all):
              result = model._predict_all(collated)

          self.assertEqual(
              len(vp_call_args),
              1,
              "Expected _predict_all_visual_prompts to be called once, "
              "but it was never called. Visual prompts are silently ignored "
              "when preprocess=False (the apply_model path).",
          )
          self.assertEqual(result, ["fake_label"])
          model._transforms.assert_not_called()
```

To fix you can move the prompt check outside the _preprocess guard so it runs on the already-collated dict on the dataloader path, but please double check the logic

lorenzoriano (changes requested):

fiftyone/zoo/models/manifest-torch.json
lorenzoriano, line 1: The 5 prompt free models declare  "type": "fiftyone.utils.ultralytics.FiftyOneYOLOEVPModel". That class requires prompt_field in its required_keys. apply_model without a prompt_field= kwarg will raise ValueError.   
Here is a minimal unit test:

```python

  def test_pf_model_fails_without_prompt_field():
      """
      Prompt-free zoo models use FiftyOneYOLOEVPModel, which requires
      'prompt_field' in its field_mapping. Calling apply_model without
      prompt_field= raises ValueError before inference starts.
      """
      get_item = YOLOEVPGetItem(transform=None, field_mapping={})
      fields = list(get_item.field_mapping.values())
      assert "prompt_field" in fields

      ds = fo.Dataset()
      ds.add_sample(fo.Sample(filepath="/tmp/test.jpg"))
      with pytest.raises(ValueError, match="prompt_field"):
          ds.select_fields(fields)
```

To fix change those 5 manifest entries to `fiftyone.utils.ultralytics.FiftyOneYOLOModel` (the same as for all yoloe11 models), which has no prompt_field dependency.

fiftyone/utils/ultralytics.py
lorenzoriano: I'm referring to `transform`

lorenzoriano (changes requested):

CharlesCNorton:
@lorenzoriano done!

lorenzoriano (approved):

fiftyone/utils/ultralytics.py
```diff
@@ -701,6 +702,137 @@ def _predict_all(self, imgs):
...
+    """YOLOE model with visual prompt (box prompt) support.
+
+    Args:
+        config: a :class:`FiftyOneYOLOEVPModelConfig`
+    """
+
+    @staticmethod
+    def collate_fn(batch):
+        prompts = [item.pop("prompt", None) for item in batch]
+        orig_images = [img.get("orig_img") for img in batch]
```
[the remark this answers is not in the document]
manushreegangwar: Is there a reason for using `pop` for prompts?

fiftyone/utils/ultralytics.py
```diff
@@ -1055,3 +1187,30 @@ def __call__(
...
+    label_to_idx = {c: i for i, c in enumerate(classes)}
+    cls_indices = [label_to_idx[label] for label in labels]
+
+    return boxes.tolist(), cls_indices, classes
+
+
+def _get_yoloe_vp_predictor():
+    try:
+        return _yoloe.YOLOEVPSegPredictor
+    except (ImportError, AttributeError) as e:
```
manushreegangwar: Does this raise an ImportError? It should raise AttributeError.

fiftyone/utils/ultralytics.py
```diff
@@ -701,6 +702,137 @@ def _predict_all(self, imgs):
...
+        prompts = imgs.get("prompts") if isinstance(imgs, dict) else None
+        if prompts and any(p is not None for p in prompts):
+            return self._predict_all_visual_prompts(
+                imgs["orig_imgs"], imgs["orig_shapes"], prompts
+            )
+
+        return super()._predict_all(imgs)
+
+    def _predict_all_visual_prompts(self, orig_images, width_height, prompts):
+        vp_predictor_cls = _get_yoloe_vp_predictor()
```
manushreegangwar: this should be done via `self._set_predictor`

https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/ultralytics.py

fiftyone/utils/ultralytics.py
```diff
@@ -701,6 +702,137 @@ def _predict_all(self, imgs):
...
+            # so batched prediction is not supported here.
+            for orig_img, wh, prompt in zip(
+                orig_images, width_height, prompts, strict=True
+            ):
+                if not prompt or not prompt.detections:
+                    all_labels.append(fol.Detections())
+                    continue
+
+                w, h = wh
+                bboxes, cls_indices, classes = _detections_to_visual_prompts(
```
manushreegangwar: this is a preprocessing step and should be handled during preprocessing. Similar to https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/ultralytics.py#L619-L639

fiftyone/utils/ultralytics.py
```diff
@@ -701,6 +702,137 @@ def _predict_all(self, imgs):
...
+                labels = to_instances(
+                    results,
+                    confidence_thresh=self.config.confidence_thresh,
+                    classes=self.config.filter_classes,
+                )
+
+                if isinstance(labels, list):
+                    labels = labels[0] if labels else fol.Detections()
+
+                all_labels.append(labels)
```
manushreegangwar: this should be handled via output processor https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/ultralytics.py#L703

fiftyone/utils/ultralytics.py
```diff
@@ -701,6 +702,137 @@ def _predict_all(self, imgs):
...
+    def _predict_all(self, imgs):
+        prompts = imgs.get("prompts") if isinstance(imgs, dict) else None
+        if prompts and any(p is not None for p in prompts):
+            return self._predict_all_visual_prompts(
+                imgs["orig_imgs"], imgs["orig_shapes"], prompts
+            )
+
+        return super()._predict_all(imgs)
+
+    def _predict_all_visual_prompts(self, orig_images, width_height, prompts):
```
manushreegangwar: this can be done via a separate `_forward_pass_prompts`

See https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/sam.py#L1138-L1159 for context

fiftyone/utils/ultralytics.py
manushreegangwar: Can you test this with `model.predict_all(args)`, where `args` is constructed using https://github.com/voxel51/fiftyone/blob/develop/fiftyone/utils/torch.py#L2934

See this https://github.com/voxel51/fiftyone/blob/develop/tests/intensive/sam_image_model_tests.py#L397 for context

manushreegangwar (changes requested):
Few comments and questions. Based on the existing patterns, all the preprocessing should happen before before the forward pass and output processing should be done via output processor.

Also, the tests in the PR description don't cover visual prompting. Can you update it show how it has been tested? Also, have you done parity testing against direct ultralytics inference?

fiftyone/utils/ultralytics.py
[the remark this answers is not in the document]
CharlesCNorton: got it

CharlesCNorton:
> Few comments and questions. Based on the existing patterns, all the preprocessing should happen before before the forward pass and output processing should be done via output processor.
> 
> Also, the tests in the PR description don't cover visual prompting. Can you update it show how it has been tested? Also, have you done parity testing against direct ultralytics inference?

I haven't done that yet. Will address all comments and update ASAP.

CharlesCNorton:
forgot a thing or two. adding now.

fiftyone/utils/ultralytics.py
```diff
@@ -707,6 +708,193 @@ def _predict_all(self, imgs):
...
+                "YOLOEVPGetItem requires a transform to produce tensors"
+            )
+        result = self._transform(img)
+
+        prompt = d["prompt_field"]
+        if prompt is not None and prompt.detections:
+            bboxes, cls_indices, classes = _detections_to_visual_prompts(
+                prompt, img.width, img.height
+            )
+            result["visual_prompts"] = {
```
manushreegangwar, line 745: nit: assumes `result` is a dict but this is never checked

fiftyone/utils/ultralytics.py
```diff
@@ -707,6 +708,193 @@ def _predict_all(self, imgs):
...
+
+    Args:
+        config: a :class:`FiftyOneYOLOEVPModelConfig`
+    """
+
+    @staticmethod
+    def collate_fn(batch):
+        visual_prompts = [item.get("visual_prompts") for item in batch]
+        vp_classes = [item.get("vp_classes") for item in batch]
+        orig_images = [img.get("orig_img") for img in batch]
```
manushreegangwar: would be better to loop over the batch once and populate all the lists

manushreegangwar (approved):

PR #6854, opened 2026-01-30
Adding one additional default operator that should be delegated

prernadh:

Adding one additional operator to [this](https://github.com/voxel51/fiftyone/pull/6843) pull request which also should have the option to be delegated.

## Summary by CodeRabbit

* **New Features**
  * Enhanced operator execution with flexible modes, now supporting both delegated and immediate execution options for improved control over operation processing.


brimoor (commented):
@prernadh `add_group_slice` just declares a new group slice on the dataset, it doesn't actually touch the samples and thus should be instant for any size dataset and should not ever need to be delegated

PR #6858, opened 2026-02-02
Fallback to drop+replace when trying to convert an existing index to unique

brimoor:
## Release Notes

- Introduces a fallback strategy to ensure that converting an existing index to unique via `create_index()` always succeeds

## Tested by

On `develop`, I'm getting a unit test failure:

```
$ pytest tests/unittests/dataset_tests.py -k test_indexes
FAILED tests/unittests/dataset_tests.py::DatasetTests::test_indexes - pymongo.errors.OperationFailure: collMod does not support converting an index to 'unique' or to 'prepareUnique' mode
```

Now on this branch, the test passes.

## Context/discussion

I recently upgraded my machine and am now running:

```
macOS Sequoia 15.5
Python 3.11.14
MongoDB 6.0.27
pymongo 4.9.2
```

and the unit tests started failing. I am perplexed by the error, because [this page](https://www.mongodb.com/docs/manual/core/index-unique/convert-to-unique) does claim that the `collMod` pattern that we're using to upgrade an existing index to unique *should work*...

Also, tests are passing on CI/CD and no one else has complained about this failure, so I am stumped...

If someone can find an explanation, that would be tremendous. In the meantime, this at least patches things...

## Summary by CodeRabbit

* **New Features**
  * 3D annotation overlays, projected cuboid/polyline views, image-slice panel and crosshair/projection tooltips.
  * Inline code editor component for JSON/schema editing.
* **UI / UX**
  * Quick-edit affordance on hover, updated schema manager UI (scan/cancel, setup flow), clearer annotation buttons.
  * Longer persistence toast for slow saves.
* **Bug Fixes**
  * Improved index error recovery and safer version-mismatch handling during saves.
* **Performance**
  * Client-side fetch caching and reduced test matrix for CI.

### DIFF
--- a/fiftyone/core/collections.py
+++ b/fiftyone/core/collections.py
@@ -10959,7 +10959,15 @@ class SampleCollection(object):
             ):
                 raise ValueError(f"Cannot modify default index '{index_name}'")
 
+        #
         # Convert existing index to unique, if necessary
+        # https://www.mongodb.com/docs/manual/core/index-unique/convert-to-unique
+        #
+        # Although MongoDB documents this strategy for more efficently
+        # upgrading an existing index to unique on MongoDB 6+, it does not seem
+        # to work in all cases, so we gracefully fallback to the "delete and
+        # replace" strategy if necessary
+        #
         if convert_to_unique:
             logger.info(
                 f"Converting existing index '{index_name}' to unique "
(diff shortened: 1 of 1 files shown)


brimoor:
> That error is perplexing if you're running mongo 6+ 🤔

indeed it is, but that's what was happening on my machine on `develop` with the setup I reported

PR #6860, opened 2026-02-02
FOEPD-3047 - Clean up the undo stack when leaving the annotation panel or current sample

marcus-kestler:

This PR clears the undo stack when the user leaves the annotation panel or navigates to another sample.

Manually locally.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Undo/redo history is now cleared during modal navigation and when exiting annotation mode to prevent unintended behavior.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/Annotate.tsx
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/Annotate.tsx
@@ -20,6 +20,7 @@ import useLabels from "./useLabels";
 import { usePrimitivesCount } from "./usePrimitivesCount";
 import { useAnnotationContextManager } from "./useAnnotationContextManager";
 import useDelete from "./Edit/useDelete";
+import { KnownContexts, useUndoRedo } from "@fiftyone/commands";
 
 const showImportPage = atom((get) => !get(activeLabelSchemas)?.length);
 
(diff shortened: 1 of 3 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/Annotate.tsx
```diff
@@ -120,13 +121,15 @@ const Annotate = ({ disabledReason }: AnnotateProps) => {
...
   const contextManager = useAnnotationContextManager();
+  const { clear: clearUndo } = useUndoRedo(KnownContexts.ModalAnnotate);
   useDelete();
 
   useEffect(() => {
     contextManager.enter();
 
     return () => {
       contextManager.exit();
+      clearUndo();
```
tom-vx51, line 132: can we add this to the context manager's exit logic instead?

edit: we already call `deactivateCommandContext` in the `exit` logic; clearing the undo stack feels appropriate to include in the deactivation logic no? https://github.com/voxel51/fiftyone/blob/develop/app/packages/core/src/components/Modal/Sidebar/Annotate/useAnnotationContextManager.ts#L195

marcus-kestler: We don't clear the undo stack on deactivate.  @sashankaryal had the use case of side by side editors which would make this bad.  You want the key bindings to change when selecting a different editor, but you want them to have individual undo stack in that case, hence the current behavior.
I did initially put the clear in the context manager's exit, but it did not work.

PR #6887, opened 2026-02-05
Allow cached fetch in the app

sashankaryal:

This PR adds a feature that allows fetch function to use an LRU cache for responses if opted in. Default is off to not disrupt status-quo.

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Release Notes

* **Performance & Optimization**
  * Implemented an intelligent caching layer for API requests with automatic deduplication to reduce network calls and improve response times.
  * Added automatic cache clearing when switching datasets to ensure fresh data is always displayed.
  * Enhanced request handling to prevent duplicate concurrent requests for the same data.

### DIFF
--- a/app/packages/utilities/package.json
+++ b/app/packages/utilities/package.json
@@ -28,6 +28,7 @@
         "@microsoft/fetch-event-source": "^2.0.1",
         "fetch-retry": "^5.0.3",
         "lodash": "^4.17.23",
+        "lru-cache": "^11.0.1",
         "mime": "^2.5.2",
         "path-browserify": "^1.0.1"
     }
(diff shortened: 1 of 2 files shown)


app/packages/utilities/src/fetch.ts
```diff
@@ -3,15 +3,45 @@ import {
...
+  sizeCalculation: (value) => {
+    try {
+      return JSON.stringify(value).length;
+    } catch {
+      return 1;
+    }
+  },
+});
+
+// Tracks in-flight requests to dedupe parallel fetches with the same cache key
```
imanjra, line 34: Nice!

PR #6895, opened 2026-02-05
Fix undo for annotation labels.

marcus-kestler:
This fixes the problem where undoing a field edit did not refresh the UI until autosave runs.
I introduced an event that is fired from UpdateLabelCommand to notify the UI to refresh.  There is an origin field to differentiate it as coming from the undo path.  There had to be some special handling of undo that reverts to the empty state.

Locally, manually

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * External annotation update events with origin tracking so updates from different sources correctly refresh the UI.
  * Sidebar edits can signal replace-or-merge behavior for label data, enabling more flexible edits and undo behavior.
  * Improved color mapping and overlay sync so visual updates propagate reliably.

* **Bug Fixes**
  * Ensured record invalidation so newly created annotations appear immediately in patches view.

### DIFF
--- a/app/packages/core/src/components/Modal/Lighter/useBridge.ts
+++ b/app/packages/core/src/components/Modal/Lighter/useBridge.ts
@@ -6,6 +6,7 @@ import {
   useAnnotationEventBus,
   useAnnotationEventHandler,
 } from "@fiftyone/annotation";
+import { type AnnotationLabel } from "@fiftyone/state";
 import {
   type LighterEventGroup,
   type Scene2D,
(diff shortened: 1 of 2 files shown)


sashankaryal (changes requested):
too complicated... I think we can come up with a simpler approach. see 3d's `useSyncWorkingToSidebar.ts` (and if not - we should still not have special purpose flags or couple commands like `UpdateLabelCommand()` with side-effects, my 2c.)

PR #6897, opened 2026-02-05
FiftyOne Browser File Upload Service

ritch:
## Summary

Adds file upload and delete routes to the FiftyOne app server. `POST /files/upload?path=<dest>` streams a binary via the `fiftyone.core.storage` interface/module, auto-incrementing the filename on collision (`photo.png` → `photo_1.png`), and returns `{"path": "<resolved>"}`. `DELETE /files?path=<target>` removes a file idempotently (204). Both are gated behind `FIFTYONE_ALLOW_BROWSER_FILE_OPERATIONS` (default `false`), and restricted to paths within `FIFTYONE_BROWSER_FILE_OPERATIONS_DIR` (default `~`). All filesystem work runs in worker threads via `anyio.to_thread.run_sync` so the event loop is never blocked. Partial files are cleaned up on failure.

### Routes

| Method | Path | Query Params | Body | Success | Description |
|--------|------|-------------|------|---------|-------------|
| `POST` | `/files/upload` | `path` (required) | Binary stream | `201 {"path": "..."}` | Upload a file, auto-increments on collision |
| `DELETE` | `/files` | `path` (required) | — | `204` (no body) | Delete a file, idempotent |

### Config

| Option | Env Var | Default | Description |
|--------|---------|---------|-------------|
| `allow_browser_file_operations` | `FIFTYONE_ALLOW_BROWSER_FILE_OPERATIONS` | `false` | Enable/disable the file operations endpoints |
| `browser_file_operations_dir` | `FIFTYONE_BROWSER_FILE_OPERATIONS_DIR` | `~` | Security boundary — all paths must resolve within this directory |

## Try it out

Start the app with `FIFTYONE_ALLOW_BROWSER_FILE_OPERATIONS=true` then:

```sh
# Upload text
curl -X POST "http://localhost:5151/files/upload?path=%2Ftmp%2Ftest.txt" \
  --data-binary "hello" -w "\n%{http_code}\n"
# → {"path":"/tmp/test.txt"}  201

# Upload again (auto-increments)
curl -X POST "http://localhost:5151/files/upload?path=%2Ftmp%2Ftest.txt" \
  --data-binary "hello again" -w "\n%{http_code}\n"
# → {"path":"/tmp/test_1.txt"}  201

# Upload an image
curl -X POST "http://localhost:5151/files/upload?path=%2Ftmp%2Fphoto.png" \
  --data-binary @photo.png -H "Content-Type: image/png" -w "\n%{http_code}\n"
# → {"path":"/tmp/photo.png"}  201

# Delete
curl -X DELETE "http://localhost:5151/files?path=%2Ftmp%2Ftest.txt" -w "\n%{http_code}\n"
# → 204 (idempotent — returns 204 even if already deleted)
```

## Summary by CodeRabbit

* **New Features**
  * Configurable browser file operations toggle and base directory (defaults to user home).
  * Streamed file upload endpoint with automatic filename collision resolution and created-path response.
  * Idempotent file deletion endpoint.

* **Security / Validation**
  * Strict path validation and confinement to the configured directory with structured error responses and HTTP status mapping.

* **Tests**
  * Extensive unit tests for uploads, deletions, error cases, path edge-cases, and large/binary streaming.


fiftyone/core/config.py
```diff
@@ -312,6 +312,18 @@ def __init__(self, d=None):
...
+            d,
+            "allow_browser_file_operations",
+            env_var="FIFTYONE_ALLOW_BROWSER_FILE_OPERATIONS",
+            default=False,
+        )
+        self.browser_file_operations_dir = self.parse_path(
+            d,
+            "browser_file_operations_dir",
+            env_var="FIFTYONE_BROWSER_FILE_OPERATIONS_DIR",
+            default=None,
```
ritch: Still need to move the actual default here, which should be the home directory. If users want to be completely open they can set this to `/` or `C:\`. We can consider another setting.

fiftyone/server/routes/files.py
```diff
@@ -0,0 +1,63 @@
...
+| Copyright 2017-2026, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
+from starlette.endpoints import HTTPEndpoint
+from starlette.requests import Request
+from starlette.responses import JSONResponse, Response
+
+from fiftyone.server.files import (
```
ritch: I realize this is non standard. I'll rework this to be `import fifstyone.server.files as fsf`.

fiftyone/server/files.py
```diff
@@ -0,0 +1,348 @@
...
+    # Normalise backslashes to forward slashes so that traversal attacks
+    # using "\..\..\" are caught on Unix (where \ is a valid filename
+    # character rather than a path separator).
+    safe_path = path.replace("\\", "/")
+
+    try:
+        normalized = fos.realpath(safe_path)
+        base = fos.realpath(allowed_base)
+    except Exception as e:
+        raise FileOperationError(
```
ritch: Just realizing that this could be an IO error so we should look into catching specific errors if possible.

fiftyone/server/files.py
```diff
@@ -0,0 +1,348 @@
...
+            code=FileOperationErrorCode.PATH_INVALID,
+            message=f"Invalid path: {e}",
+            details={"path": path},
+        )
+
+    if not (normalized.startswith(base + os.sep) or normalized == base):
+        raise FileOperationError(
```
ritch: This will need to be re-worked in teams where there are multiple "drives" / buckets. I guess the multiple drives scenario exists in OSS. Not sure how robust this needs to be.

ritch: I want the default configuration to be totally safe and avoid being able to delete `/` or anything silly like that without jumping through some hoops.

fiftyone/server/files.py
```diff
@@ -0,0 +1,348 @@
...
+# =========================================================================
+
+
+def _sync_write_chunks(path: str, chunks: list) -> None:
+    """Writes *chunks* to *path* via :mod:`fiftyone.core.storage`.
+
+    Creates parent directories as needed.  This function is designed to be
+    called from a worker thread via :func:`anyio.to_thread.run_sync`.
+    """
+    fos.ensure_basedir(path)
```
ritch: I'll move this around. This is going to slow down every write...will move to a one time call.

fiftyone/server/files.py
```diff
@@ -0,0 +1,348 @@
...
+            f.write(chunk)
+
+
+def _sync_try_remove(path: str) -> None:
+    """Best-effort removal of *path*.  Ignores errors."""
+    try:
+        if fos.exists(path):
+            fos.delete_file(path)
+    except Exception:
+        pass
```
ritch: We shouldn't ignore all errors here...rethinking this.

imanjra (approved):
LGTM

fiftyone/server/files/operations.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def _sync_try_remove(path: str) -> None:
+    """Best-effort removal of *path*.  Logs a warning on failure."""
+    try:
+        if fos.exists(path):
+            fos.delete_file(path)
+    except Exception:
+        logger.warning(
+            "Failed to clean up partial file: %s", path, exc_info=True
```
ritch, line 55: I'll reword this, its not quite accurate. It should really be a 404.

PR #6908, opened 2026-02-06
chore(tests): deprecations and unawaited

afoley587:

Currently, our `async def` tests aren't being awaited. You might see this in our test logs ([example](https://github.com/voxel51/fiftyone/actions/runs/21739719765/job/62712545861)):

```
tests/unittests/operators/delegated_tests.py::DelegatedOperationServiceTests::test_set_completed_in_async_context
  /opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/unittest/case.py:549: RuntimeWarning:
  
  coroutine 'DelegatedOperationServiceTests.test_set_completed_in_async_context' was never awaited
  
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.

tests/unittests/operators/notification_service_tests.py::TestMongoChangeStreamNotificationServiceAsync::test_handle_change_insert
tests/unittests/operators/notification_service_tests.py::TestMongoChangeStreamNotificationServiceAsync::test_handle_change_delete
tests/unittests/operators/notification_service_tests.py::TestMongoChangeStreamNotificationServiceAsync::test_run_with_change_stream
tests/unittests/operators/notification_service_tests.py::TestMongoChangeStreamNotificationServiceAsync::test_run_with_fallback_to_polling
tests/unittests/operators/notification_service_tests.py::TestMongoChangeStreamNotificationServiceAsync::test_poll
  /opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/pytest_asyncio/plugin.py:761: DeprecationWarning:
  
  The event_loop fixture provided by pytest-asyncio has been redefined in
  /home/runner/work/fiftyone/fiftyone/tests/unittests/operators/notification_service_tests.py:143
  Replacing the event_loop fixture with a custom implementation is deprecated
  and will lead to errors in the future.
  If you want to request an asyncio event loop with a scope other than function
  scope, use the "scope" argument to the asyncio mark when marking the tests.
  If you want to return different types of event loops, use the event_loop_policy
  fixture.
```

It's because we're running `async def` tests in a syncronous test class:

```
class DelegatedOperationServiceTests(unittest.TestCase):
    @pytest.mark.asyncio
    async def test_something...
```

We were adding `pytest.mark.asyncio` on the tests, but we're mixing `unittest` with `pytest` and it doesn't seem to work like we expect. We can use the `unittest.IsolatedAsyncioTestCase` base class:

```
class DelegatedOperationAsyncServiceTests(unittest.IsolatedAsyncioTestCase):
    async def test_something...
```

When I first ran this, I started seeing a lot of errors related to `pytest.fixtures`. So I moved those fixtures into an `asyncSetUp` where needed.

We also don't need to patch `asyncio.create_task` - we're not trying to avoid the task creation.

No more async warnings in this section of code

```
(.venv) alexanderfoley@Alexanders-MacBook-Pro fiftyone % pytest ./tests/unittests/operators/notification_service_tests.py                                                                     
============================================================================================================= test session starts =============================================================================================================
platform darwin -- Python 3.10.14, pytest-8.1.1, pluggy-1.5.0
rootdir: /Users/alexanderfoley/Projects/AS-1126/fiftyone
configfile: pytest.ini
plugins: asyncio-0.23.8, dash-2.18.1, timeout-2.4.0, mock-3.10.0, aspectlib-2.0.0, anyio-4.8.0, cov-4.0.0
asyncio: mode=strict
collected 10 items                                                                                                                                                                                                                            

tests/unittests/operators/notification_service_tests.py ..........                                                                                                                                                                      [100%]

============================================================================================================= 10 passed in 10.61s =============================================================================================================
```

## Release Notes

<!--
-   [x] Selected option
-->

        notes for FiftyOne users.

-   [ ] Other


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Tests**
  * Added async test suite for delegated operations with validation for completed operation states.
  * Refactored notification service tests to use streamlined async testing approach.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/tests/unittests/operators/notification_service_tests.py
+++ b/tests/unittests/operators/notification_service_tests.py
@@ -11,6 +11,7 @@ import datetime
 import unittest
 from unittest.mock import AsyncMock, MagicMock, patch
 
+import pytest
 from pymongo.errors import OperationFailure
 
 from fiftyone.operators.message import MessageData, MessageMetadata
@@ -137,34 +138,31 @@ class TestMongoChangeStreamNotificationService(unittest.TestCase):
         )
 
 
-class TestMongoChangeStreamNotificationServiceAsync(
-    unittest.IsolatedAsyncioTestCase
-):
-    async def asyncSetUp(self):
-        self.remote_notifier = MagicMock()
-        self.remote_notifier.broadcast_to_store = AsyncMock()
+class TestMongoChangeStreamNotificationServiceAsync:
+    @pytest.fixture
+    def notification_service(self):
+        remote_notifier = MagicMock()
+        remote_notifier.broadcast_to_store = AsyncMock()
 
-        self.collection = MagicMock()
-        self.db = MagicMock()
-        self.db.__getitem__.return_value = self.collection
+        collection = MagicMock()
+        db = MagicMock()
+        db.__getitem__.return_value = collection
 
-        # Patch async DB connection getter used by the service
-        # Initialize collection manually for testing
-        self.get_db_patcher = patch("fiftyone.core.odm.get_async_db_conn")
-        self.mock_get_db = self.get_db_patcher.start()
-        self.addCleanup(self.get_db_patcher.stop)
-        self.mock_get_db.return_value = self.db
+        with patch("fiftyone.core.odm.get_async_db_conn") as mock_get_db:
+            mock_get_db.return_value = db
+            notification_service = MongoChangeStreamNotificationService(
+                collection_name="test_collection",
+                remote_notifier=remote_notifier,
+            )
+            # Initialize collection manually for testing
+            notification_service._collection_async = collection
+            yield notification_service, collection, remote_notifier
 
-        self.service = MongoChangeStreamNotificationService(
-            collection_name="test_collection",
-            remote_notifier=self.remote_notifier,
-        )
-
-        self.service._collection_async = self.collection
-
-    async def test_handle_change_insert(self):
+    @pytest.mark.asyncio
+    async def test_handle_change_insert(self, notification_service):
+        service, collection, remote_notifier = notification_service
         callback = MagicMock()
-        self.service.subscribe("test_store", callback)
+        service.subscribe("test_store", callback)
 
         change = {
             "operationType": "insert",
@@ -177,8 +175,8 @@ class TestMongoChangeStreamNotificationServiceAsync(
             "wallTime": datetime.datetime.now(),
         }
 
-        await self.service._handle_change(change)
-        await asyncio.gather(*self.service._background_tasks)
+        await service._handle_change(change)
+        await asyncio.gather(*service._background_tasks)
 
         callback.assert_called_once()
         called_message = callback.call_args[0][0]
@@ -187,11 +185,12 @@ class TestMongoChangeStreamNotificationServiceAsync(
         assert called_message.value == "test_value"
         assert called_message.metadata.dataset_id == "test_dataset"
         assert called_message.metadata.operation_type == "insert"
-        self.remote_notifier.broadcast_to_store.assert_awaited_once()
 
-    async def test_handle_change_delete(self):
+    @pytest.mark.asyncio
+    async def test_handle_change_delete(self, notification_service):
+        service, collection, remote_notifier = notification_service
         callback = MagicMock()
-        self.service.subscribe("test_store", callback)
+        service.subscribe("test_store", callback)
 
         # Mock a delete operation where fullDocument is not available
         change = {
@@ -206,8 +205,8 @@ class TestMongoChangeStreamNotificationServiceAsync(
             "wallTime": datetime.datetime.now(),
         }
 
-        await self.service._handle_change(change)
-        await asyncio.gather(*self.service._background_tasks)
+        await service._handle_change(change)
+        await asyncio.gather(*service._background_tasks)
 
         callback.assert_called_once()
         called_message = callback.call_args[0][0]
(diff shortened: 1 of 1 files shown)


fiftyone/service/util.py
```diff
@@ -111,7 +111,7 @@ def get_listening_tcp_ports(process):
     Returns:
         generator of integers
     """
-    for conn in process.connections(kind="tcp"):
+    for conn in process.net_connections(kind="tcp"):
```
afoley587: Addresses the following deprecation warning

```
  /opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/fiftyone/service/util.py:114: DeprecationWarning: connections() is deprecated and will be removed; use net_connections() instead
    for conn in process.connections(kind="tcp"):
```

kaixi-wang: Does the app run with this? I think I tried to address this awhile back and had issues with net_connections not being on Mac or something

afoley587: Let me try - I can also back out this change if we don't want this change. I think the async stuff is more pressing!

Here's a REPL test:

```
>>> process = psutil.Process()
>>> process
psutil.Process(pid=80802, name='python3.10', status='running')
>>> process.net_connections(kind="tcp")
[]
>>> 
```

afoley587: Maybe this CR comment is what you ran into before and we should back this change out

https://github.com/voxel51/fiftyone/pull/6908#discussion_r2774368876

afoley587: I rolled this back so this is a "tests only" pr

tests/unittests/operators/notification_service_tests.py
exupero, line 1: I'm finding that on develop the `async def` tests in this module do fail when I add `assert False`. So are these changes necessary?

swheaton: interesting that DelegatedOperationServiceTests in fiftyone-teams is already an IsolatedAsyncioTestCase

afoley587: @exupero - It looks like those tests you pointed out don't inherit from `unittest.TestCase`, so are exempt from this PR. 

With respect to the `assert False`, Maybe we're just mixing two patterns, asyncio does actually patch it, but the warnings are being emitted. Maybe not needed, but it does make our test outputs a bit cleaner. What're your thoughts?

afoley587: Ah, I see `TestMongoChangeStreamNotificationServiceAsync` wasn't a `unittest.TestCase`.

The warnings seem to be coming from the `patch("asyncio.create_task")` (which we don't need). I can also remove those patches.

They leave this deprecation warning though

```
  /Users/alexanderfoley/.asdf/installs/python/3.10.14/lib/python3.10/site-packages/pytest_asyncio/plugin.py:761: DeprecationWarning: The event_loop fixture provided by pytest-asyncio has been redefined in
  /Users/alexanderfoley/Projects/AS-1126/fiftyone/tests/unittests/operators/notification_service_tests.py:141
  Replacing the event_loop fixture with a custom implementation is deprecated
  and will lead to errors in the future.
  If you want to request an asyncio event loop with a scope other than function
  scope, use the "scope" argument to the asyncio mark when marking the tests.
  If you want to return different types of event loops, use the event_loop_policy
  fixture.
  
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

swheaton: ultimately would be nice to convert all to pytest because the UnitTest way is 👎🏼 lol

swheaton: totally in for removing warnings though so will dismiss my RFC and you can make the decision

swheaton: or maybe not ... github UI continues to baffle me

afoley587: We can keep pytest and remove the event loop patching if that's what the developers prefer

afoley587: @swheaton - removed the event loop patching and create_task patching:

```
(.venv) alexanderfoley@Alexanders-MacBook-Pro fiftyone % pytest ./tests/unittests/operators/notification_service_tests.py
============================================================================================================= test session starts =============================================================================================================
platform darwin -- Python 3.10.14, pytest-8.1.1, pluggy-1.5.0
rootdir: /Users/alexanderfoley/Projects/AS-1126/fiftyone
configfile: pytest.ini
plugins: asyncio-0.23.8, dash-2.18.1, timeout-2.4.0, mock-3.10.0, aspectlib-2.0.0, anyio-4.8.0, cov-4.0.0
asyncio: mode=strict
collected 10 items                                                                                                                                                                                                                            

tests/unittests/operators/notification_service_tests.py ..........                                                                                                                                                                      [100%]

============================================================================================================= 10 passed in 10.07s =============================================================================================================
```

exupero: > With respect to the `assert False`, Maybe we're just mixing two patterns, asyncio does actually patch it

@afoley587 I don't understand this comment. To clarify, I added `assert False` to each async test in this module in order to verify that the tests fail. With the first test in the PR, develop's version of the test passes even with `assert False`, confirming that it's not running properly, but these tests did fail with `assert False`, even on develop, so they're apparently not affected (because the class doesn't inherit from unittest.TestCase, correct?). It sounds like your changes to this file are instead to clean up some deprecation warnings in the output, which is fine with me as long as the tests still run and fail when expected.

afoley587: In this file, you are totally correct. The `(unittest.IsolatedAsyncioTestCase)` wasn't needed, I misread the test. The tests were running as expected (unlike the delegated operator tests).

I have changed this file to:

1. keep using `pytest`
2. remove the `event_loop` fixture (which caused warnings)
3. removed the `patch("asyncio.create_task")` (which caused warnings)

PR #6912, opened 2026-02-06
Add MapAnything to model zoo

CharlesCNorton:
## Summary

- Add Meta MapAnything universal 3D reconstruction model to the model zoo
- New `fiftyone/utils/mapanything.py` wrapper with dual output modes
- One model variant: MapAnything Apache (1B params, Apache 2.0)

## Model Added

| Model | Size | Description |
|-------|------|-------------|
| `map-anything-apache-torch` | 4.23 GB | Universal feed-forward metric 3D reconstruction from images |

## Usage

```python
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset(quickstart, max_samples=5)

# Depth heatmaps
model = foz.load_zoo_model(map-anything-apache-torch)
dataset.apply_model(model, label_field=ma_depth)

# 3D point clouds
model = foz.load_zoo_model(map-anything-apache-torch, output_type=pointcloud)
dataset.apply_model(model, label_field=ma_points)
```

## Testing

106 tests passing across 9 categories:

**Config (16 tests)**
- Default values, custom values, type inheritance
- All parameters: hf_repo, output_type, use_amp, amp_dtype
- Combined options, unknown keys, edge cases

**Input Conversion (14 tests)**
- PIL RGB/RGBA/grayscale/palette/CMYK/1-bit
- Numpy uint8 with pixel preservation
- File I/O: PNG, JPEG
- Dimensions: tiny (1px), large (4096px), odd, extreme aspect ratios

**Depth Output (12 tests)**
- Heatmap shape preservation for 518x518 and 392x518
- Normalization to [0, 1], zero-depth safety, uniform depth
- Float dtype, large depth ranges, small values, NaN/Inf checks

**Pointcloud Output (12 tests)**
- Classification label with points3d attribute
- Full mask, partial mask, empty mask
- Combined model mask and depth validity filtering
- Coordinate finiteness, different spatial sizes

**Inference Control Flow (14 tests)**
- Empty input, single, multiple images
- Infer called per image with correct kwargs
- AMP on/off, memory-efficient flag, minibatch size, mask edges
- Temp file cleanup

**Input Variations (10 tests)**
- PIL RGB/RGBA/grayscale, numpy, filepath
- 32px, 4K, portrait, extreme landscape/portrait

**Determinism (6 tests)**
- Same input same output for depth and pointcloud
- Different predictions produce different outputs
- Output count matches input count

**Batch Processing (12 tests)**
- Batch sizes 1, 2, 4, 16
- Batch vs individual consistency for depth and pointcloud
- Mixed input types, sizes, and color modes in single batch
- Per-item independence, different masks/depths across batch

**Model Properties & Manifest (10 tests)**
- media_type, stored attributes
- Manifest entry: license, source URL, type path, tags, size_bytes

## Summary by CodeRabbit

* **New Features**
  * MapAnything integration: depth heatmaps and 3D pointcloud outputs, selectable output modes, mixed-precision inference, and broad input-format support with on-demand runtime installation.
* **Tests**
  * Comprehensive unit test suite covering config defaults/validation, input conversions, depth/pointcloud processing, batching, determinism, edge cases, and cleanup.

---
**Enterprise sync PR**: https://github.com/voxel51/fiftyone-teams/pull/3008


fiftyone/zoo/models/manifest-torch.json
```diff
@@ -1,5 +1,38 @@
...
+                ],
+                "cpu": {
+                    "support": true
+                },
+                "gpu": {
+                    "support": true
+                }
+            },
+            "tags": ["depth", "3d", "pointcloud", "reconstruction", "torch"],
+            "training_data": [],
```
PeterOttoMoeller, line 33: why empty?

fiftyone/zoo/models/manifest-torch.json
```diff
@@ -1,5 +1,38 @@
...
+            "base_name": "map-anything-apache-torch",
+            "base_filename": null,
+            "version": null,
+            "description": "Universal feed-forward metric 3D reconstruction from images producing depth maps, camera poses, intrinsics, and point clouds",
+            "source": "https://huggingface.co/facebook/map-anything-apache",
+            "author": "Nikhil Keetha, et al.",
+            "license": "Apache 2.0",
+            "size_bytes": 4230000000,
+            "default_deployment_config_dict": {
+                "type": "fiftyone.utils.mapanything.MapAnythingModel",
```
PeterOttoMoeller, line 13: potentially dumb question: why does this live under utils?

fiftyone/utils/mapanything.py
```diff
@@ -0,0 +1,268 @@
...
+        super().__init__(d)
+
+        self.hf_repo = self.parse_string(
+            d, "hf_repo", default=DEFAULT_MAPANYTHING_MODEL
+        )
+        self.output_type = self.parse_string(
+            d, "output_type", default="depth"
+        )
+
+        _valid_output_types = ("depth", "pointcloud")
```
PeterOttoMoeller: nitpicking: could we make this a class variable instead of a local variable inside __init__()?

fiftyone/utils/mapanything.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def _load_model(self, config):
+        """Load the MapAnything model from HuggingFace."""
+        fou.ensure_torch()
+        import torch
+
+        if config.device is not None:
+            device = torch.device(config.device)
+        elif torch.cuda.is_available():
+            device = torch.device("cuda")
```
PeterOttoMoeller, line 156: what happens here when there are multiple cuda devices?

fiftyone/utils/mapanything.py
```diff
@@ -0,0 +1,268 @@
...
+        else:
+            device = torch.device("cpu")
+
+        model = mapanything_models.MapAnything.from_pretrained(
+            config.hf_repo
+        ).to(device)
+        model.eval()
```
PeterOttoMoeller, line 163: just curious: does this disable gradient tracking? Does it have any other (maybe unwanted) side effects?

fiftyone/utils/mapanything.py
```diff
@@ -0,0 +1,268 @@
...
+        import torch
+
+        if not isinstance(imgs, list):
+            imgs = [imgs]
+
+        if len(imgs) == 0:
+            return []
+
+        outputs = []
+        with torch.no_grad():
```
PeterOttoMoeller, line 194: I suppose we also ensure no gradients here....

fiftyone/utils/mapanything.py
```diff
@@ -0,0 +1,268 @@
...
+        if isinstance(img, np.ndarray):
+            return Image.fromarray(img).convert("RGB")
+
+        pil_img = fout.ToPILImage()(img)
+        return pil_img.convert("RGB")
+
+    def _predict_all(self, imgs):
+        """Run MapAnything inference on a list of images."""
+        fou.ensure_torch()
+        import torch
```
PeterOttoMoeller: we want some kind of conditional import here, right? 

Seems to me though, that we always need torch, even if the model is only instantiated, not used for inference, because we call 

`super().__init__(config)` in `__init__()` which calls (the overwritten) `_load_model()` which needs torch

So I guess I am wondering if there is an execution path there that does NOT need torch

fiftyone/utils/mapanything.py
```diff
@@ -0,0 +1,268 @@
...
+                )
+                tmp.close()
+                pil_img.save(tmp.name)
+
+                try:
+                    views = mapanything_image.load_images([tmp.name])
+                finally:
+                    os.unlink(tmp.name)
+
+                preds = self._model.infer(
```
PeterOttoMoeller, line 209: wait, does this support inference on multiple views simultaneously?

PeterOttoMoeller (dismissed):
I am not sure if changes are really necessary :-) 

But  think you should probably at least have a look at and consciously close all comments, to be on the safe side

fiftyone/zoo/models/manifest-torch.json
CharlesCNorton: No real reason beyond matching the other wrappers. Happy to move it if you'd prefer it elsewhere.

fiftyone/zoo/models/manifest-torch.json
CharlesCNorton: Left it blank rather than list a subset; I can populate it from the model card if useful.

fiftyone/utils/mapanything.py
CharlesCNorton: Moved them to class attributes.

fiftyone/utils/mapanything.py
CharlesCNorton: It uses the default device; `config.device` targets a specific GPU. No multi-GPU sharding.

fiftyone/utils/mapanything.py
CharlesCNorton: `eval()` doesn't affect gradients; those are handled by the `no_grad()` in `_predict_all`. Nothing else is affected.

fiftyone/utils/mapanything.py
CharlesCNorton: Yes, the `no_grad()` wraps the loop.

fiftyone/utils/mapanything.py
CharlesCNorton: Agreed, there's no torch-free path, so it's a plain `ensure_torch()` + import now.

fiftyone/utils/mapanything.py
CharlesCNorton: Not currently: single-view per sample, and the joint multi-view path isn't exposed through `apply_model`. Could add it later.

kevin-dimichel:
removing aloha. no aloha owned files.

benjaminpkane (approved):

PR #6922, opened 2026-02-09
Fix/field creation

mcdoh:

When a new Field is created via Schema Manager and that field is then used to create new Labels the overlay is not visible in the scene. Furthermore, these newly-created labels aren't persisted to the backend.

https://github.com/user-attachments/assets/8c3b7019-f873-400a-94b9-a7e6c3b916b1

https://github.com/user-attachments/assets/ced3e912-088d-4872-96c1-6bc8b4dc153d

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Preserve server-enriched properties when updating detections, classifications, and polylines by merging updates rather than replacing.
  * Tolerate missing or stale schemas by inferring minimal schema info for reliable delta generation.
  * Newly created fields are activated after creation; errors surface as user notifications.
  * Improved loading/recognition of additional label types from samples.
  * Renderer now respects active label schemas more consistently for display.


app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/NewFieldSchema/NewFieldSchema.tsx
```diff
@@ -222,7 +229,6 @@ const NewFieldSchema = () => {
       callback: (createResult) => {
         if (createResult.error) {
           setIsCreating(false);
-          console.error("Failed to create field:", createResult.error);
```
kaixi-wang, line 225: Can you leave console logs? Otherwise it feels impossible to debug

app/packages/core/src/components/Modal/Lighter/LighterSampleRenderer.tsx
```diff
@@ -115,9 +117,20 @@ const LighterSetupImpl = (props: {
     fos.lookerOptions({ modal: true, withFilter: false })
   );
 
+  // Read activePaths directly from Jotai to bypass Recoil's filterPaths,
```
kaixi-wang, line 126: If a field isn't reflected in the recoil graphql, then shouldn't that be fixed? Otherwise won't you risk the backend rejecting the updates, corrupting the sample by adding a field not declared in the schema, or inconsistencies with the field not being visible in other parts of the app?

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/NewFieldSchema/NewFieldSchema.tsx
```diff
@@ -234,10 +240,17 @@ const NewFieldSchema = () => {
...
-                );
+                const {
+                  active_label_schemas,
+                  label_schemas,
+                } = schemasResult.result;
+
+                addToActiveSchema(active_label_schemas);
+                activateFields.execute({
+                  fields: active_label_schemas,
+                });
```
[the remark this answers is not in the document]
kaixi-wang: This seems like a valid issue. One of the issues I've seen in general testing locally is the lack of feedback for errors and the continuous retries leading to the server getting stuck in an error loop and loss of data for things never actually saved to the db but seemingly saved because of the FE cache

app/packages/core/src/components/Modal/Lighter/LighterSampleRenderer.tsx
lanzhenw: this is the side effect of removing the ctx.trigger(‘reload_dataset’) in the operator and only updating the activeLabel atom in the onComplete callback. In this case, looks like a reload in the operator  would be a cleaner fix.

lanzhenw: https://github.com/voxel51/fiftyone/pull/6925
This would be the fix from operator side.

mcdoh: I added a call to `useRefresh` to get Relay to re-fetch from the backend. That's an asynchronous call so I kept the client-side merge in place to keep things aligned.

lanzhenw: `reload_dataset` basically does the same thing under the hood. I like how you handled error notifications and FE controlled refresh timing. 

I will close my PR.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/NewFieldSchema/NewFieldSchema.tsx
```diff
@@ -234,34 +251,63 @@ const NewFieldSchema = () => {
...
+                      addToActiveSchema(new Set(active_label_schemas));
+                      setLabelSchemasData(label_schemas);
+                      setActiveLabelSchemas(active_label_schemas);
+                      refreshSchema();
+                      exitNewFieldMode();
+                    },
+                  }
                 );
+              } else {
+                exitNewFieldMode();
```
lanzhenw: should only return to annotation sidebar on success.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/NewFieldSchema/NewFieldSchema.tsx
```diff
@@ -234,34 +251,63 @@ const NewFieldSchema = () => {
...
               if (schemasResult.result) {
-                setLabelSchemasData(schemasResult.result.label_schemas);
-                setActiveLabelSchemas(
-                  schemasResult.result.active_label_schemas
+                const {
+                  active_label_schemas,
+                  label_schemas,
+                } = schemasResult.result;
+
+                activateFields.execute(
```
lanzhenw: In `create_and_activate_field` operator, we already called `activate_label_schemas` operator. It would be redundant to call it again.

lanzhenw: In `create_and_activate_field` operator, we already called `activate_label_schemas` operator. It will be a redundant call. 

https://github.com/voxel51/fiftyone/blob/release/v1.13.0/plugins/operators/annotation.py#L235

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/NewFieldSchema/NewFieldSchema.tsx
lanzhenw: ✂️

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/NewFieldSchema/NewFieldSchema.tsx
```diff
@@ -234,34 +250,41 @@ const NewFieldSchema = () => {
...
-                );
-              }
+                const {
+                  active_label_schemas,
+                  label_schemas,
+                } = schemasResult.result;
+
+                addToActiveSchema(new Set(active_label_schemas));
+                setLabelSchemasData(label_schemas);
+                setActiveLabelSchemas(active_label_schemas);
```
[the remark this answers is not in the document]
lanzhenw: addressed

PR #6931, opened 2026-02-09
FOEPD-3057 Add a hook to sync the sidebar

marcus-kestler:

Uses a hook based approach to sync the sidebar.

Manually

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Improvements**
  * Overlay edits now reliably sync into the sidebar annotation editor with undo/redo awareness and prevention of redundant updates.
  * Pending annotation edits are handled more robustly to preserve correct "before" state during rapid typing.
  * Position updates use clearer bounds handling for more consistent transforms.

* **Refactor**
  * Simplified synchronization flow by removing redundant two‑way command syncing.

* **Chore**
  * Added an additional exported label-related type.

### DIFF
--- a/app/packages/commands/src/actions/ActionManager.ts
+++ b/app/packages/commands/src/actions/ActionManager.ts
@@ -11,7 +11,11 @@ export type UndoStateListener = (
   redoEnabled: boolean
 ) => void;
 
-export type ActionListener = (actionId: string, isUndo: boolean) => void;
+export type ActionListener = (
+  actionId: string,
+  isUndo: boolean,
+  isRedo: boolean
+) => void;
 
 /**
  * Manages the execution of actions and supports
@@ -75,7 +79,7 @@ export class ActionManager {
       try {
         await undoable.undo();
         this.redoStack.push(undoable);
-        this.fireActionListeners(undoable.id, true);
+        this.fireActionListeners(undoable.id, true, false);
         this.fireUndoListeners();
         return true;
       } catch (error) {
@@ -100,7 +104,7 @@ export class ActionManager {
       try {
         await undoable.execute();
         this.undoStack.push(undoable);
-        this.fireActionListeners(undoable.id, false);
+        this.fireActionListeners(undoable.id, false, true);
         this.fireUndoListeners();
         return true;
       } catch (error) {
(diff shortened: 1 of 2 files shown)


app/packages/core/src/components/Modal/Lighter/useSyncOverlayToSidebar.ts
```diff
@@ -0,0 +1,92 @@
...
+      overlayLabel as Record<string, unknown>
+    ) as AnnotationLabel["data"];
+
+    if (isEqual(coercedLabel, lastSyncedDataRef.current)) {
+      return;
+    }
+
+    // Update the atom with the latest authoritative data from the overlay
+    // We perform a full replacement to ensure fields are cleared correctly on undo
+    store.set(editingValue, {
```
tom-vx51: this will have undesired behavior if the stack has edits from multiple labels - changes will be applied to the sidebar irrespective of which label the change belongs to. You can use this helper to update an atom by label ID https://github.com/voxel51/fiftyone/blob/develop/app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts#L156-L162

app/packages/core/src/components/Modal/Lighter/useSyncOverlayToSidebar.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Hook that synchronizes changes from Lighter overlays to the sidebar editing atoms.
+ * It ensures that the sidebar remains the authoritative UI for editing, but
+ * pulls data from the renderer when it changes (e.g. via commands, undo/redo).
+ */
+export function useSyncOverlayToSidebar(scene: Scene2D | null) {
+  const useEventHandler = useLighterEventHandler(
+    scene?.getEventChannel() ?? UNDEFINED_LIGHTER_SCENE_ID
+  );
+  const store = getDefaultStore();
```
tom-vx51: I think we can remove the store if you migrate to the helper in my other comment, but in general we should avoid using the jotai store directly as it breaks hook reactivity. `useAtomCallback` is typically what you want if you need synchronous values

app/packages/core/src/components/Modal/Lighter/useSyncOverlayToSidebar.ts
marcus-kestler: Addressed.

app/packages/core/src/components/Modal/Lighter/useSyncOverlayToSidebar.ts
marcus-kestler: Converted to useAtomCallback.

PR #6934, opened 2026-02-09
Improve sample pagination performance

kaixi-wang:

Remove redundant collection/dataset-level calls for getting schema and media paths from inside loops and compute values once to significantly improve performance on datasets with many label fields

Running locally in teams, average time for samples page request dropped from 22sec to 13 sec.

Without changes:
image

With these changes:
image

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Refactor**
  * Improved metadata and media-field handling to reduce redundant work, speeding up sample pagination and page load times when browsing datasets.
* **Chores**
  * Minor code cleanup to remove unused imports and streamline server-side processing.

### DIFF
--- a/fiftyone/server/metadata.py
+++ b/fiftyone/server/metadata.py
@@ -77,18 +77,11 @@ async def get_metadata(
     filepath = sample["filepath"]
     metadata = sample.get("metadata", None)
 
-    if additional_media_fields is not None:
-        (
-            opm_field,
-            detections_fields,
-            additional_fields,
-        ) = additional_media_fields
-    else:
-        (
-            opm_field,
-            detections_fields,
-            additional_fields,
-        ) = _get_additional_media_fields(collection)
+    (
+        opm_field,
+        detections_fields,
+        additional_fields,
+    ) = additional_media_fields if additional_media_fields is not None else _get_additional_media_fields(collection)
 
     filepath_result, filepath_source, urls = _create_media_urls(
         collection,


fiftyone/server/metadata.py
```diff
@@ -59,25 +58,37 @@ async def get_metadata(
...
+            opm_field,
+            detections_fields,
+            additional_fields,
+        ) = additional_media_fields
+    else:
+        (
+            opm_field,
+            detections_fields,
+            additional_fields,
+        ) = _get_additional_media_fields(collection)
```
jleven: When there are no additional media fields, this will still be called every time. Is that okay?

Better would be to have a way to distinguish between “there are no additional media fields” and “I don’t know if there are any”

jleven: In Typescript, it would be null vs undefined

kaixi-wang: Not when called from paginate samples which calls `_get_additional_media_fields(collection)` and will return `(None, None, [])`. However if additional_media_fields is not passed, then I think we have no choice but to check since this is called per sample.

fiftyone/server/samples.py
```diff
@@ -137,10 +137,18 @@ async def paginate_samples(
 
     metadata_cache = {}
     url_cache = {}
+    additional_media_fields = (
+        fosm._get_additional_media_fields(view) if samples else None
```
jleven, line 141: Should this else case be None? Or rather (None, None, [])
?

fiftyone/server/metadata.py
exupero: A "value or default" pattern seems more idiomatic, and prevents the two branches from diverging in what vars it assigns:
```suggestion
    (
        opm_field,
        detections_fields,
        additional_fields,
    ) = additional_media_fields if additional_media_fields is not None else _get_additional_media_fields(collection)
```

fiftyone/server/samples.py
kaixi-wang: Not necessary because if there are no samples, _create_sample_item won't be called anyways

jleven (approved):

PR #6938, opened 2026-02-10
[HA] Ensure _cls attribute is set for newly-created labels FOEPD-3119

tom-vx51:

Adds explicit logic to set `_cls` to ensure proper typing when creating labels.

local

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Refactor**
  * Introduced explicit discriminators for label types to make annotation overlays more consistent and type-safe.
* **Chores**
  * Broadened exported overlay symbols so more overlay types are available from the public package surface.
* **Notes**
  * No user-facing behavior changes; improvements are internal and aim to reduce integration friction.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useCreate.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useCreate.ts
@@ -17,8 +17,7 @@ import { useCallback } from "react";
 import type { LabelType } from "./state";
 import { defaultField, editing, savedLabel } from "./state";
 import { useQuickDraw } from "./useQuickDraw";
-import { DetectionLabel } from "@fiftyone/looker";
-import { ClassificationLabel } from "@fiftyone/looker/src/overlays/classifications";
+import { ClassificationLabel, DetectionLabel } from "@fiftyone/looker";
 
 const useCreateAnnotationLabel = () => {
   const { scene, addOverlay, overlayFactory } = useLighter();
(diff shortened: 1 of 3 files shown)


app/packages/looker/src/overlays/detection.ts
```diff
@@ -66,6 +66,7 @@ const getIndexIdFromInstanceIdForLabel = (
 };
 
 export interface DetectionLabel extends RegularLabel {
+  _cls: "Detection";
```
kaixi-wang, line 69: is there a reason for why this is defined as an interface but PolylineLabel is a type?

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useCreate.ts
```diff
@@ -17,6 +17,8 @@ import { useCallback } from "react";
 import type { LabelType } from "./state";
 import { defaultField, editing, savedLabel } from "./state";
 import { useQuickDraw } from "./useQuickDraw";
+import { DetectionLabel } from "@fiftyone/looker";
+import { ClassificationLabel } from "@fiftyone/looker/src/overlays/classifications";
```
mcdoh: Should we not provide an export of `ClassificationLabel` from _looker_ so that we can just `import { ClassificationLabel, DetectionLabel } from "@fiftyone/looker";` and imply rough equivalence?

mcdoh (dismissed):
Minor request, take it or leave it.

app/packages/looker/src/overlays/detection.ts
tom-vx51: not that i'm aware of, the typing is a bit of a mess :(

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useCreate.ts
tom-vx51: yeah, we should. I can update this case specifically. Unrelated to the imports, these types are just suggestions currently :(; we should migrate to some strongly-typed factory method similar to the overlay creation

PR #6941, opened 2026-02-10
Fix point label support in _to_sam_points

CharlesCNorton:

Fix field name mismatch in `_to_sam_points` that prevents object-level point labels from working. Line 390 checks for `"sam_labels"` but line 389 reads `keypoint.sam2_labels`, so `sam2_labels` is silently ignored (all points default to positive) and `sam_labels` crashes with `AttributeError`.

Fix checks `sam2_labels` first, falls back to `sam_labels`, then defaults to all ones.

Added 6 unit tests to `test_sam2.py`:
  - `sam2_labels` respected as point labels
  - `sam_labels` respected as fallback
  - No labels defaults to all positive
  - `sam2_labels` takes priority when both set
  - NaN points filtered with labels staying aligned
  - All-negative labels passed through

Manually verified with SAM2 image model on COCO images (two cats, two dogs, two zebras, horse+rider, two bears) confirming negative points exclude regions as expected.

## What areas of FiftyOne does this PR affect?

- [X] Core: Core `fiftyone` Python library changes

## Summary by CodeRabbit

* **New Features**
  * Improved label handling for point annotations: the system now prefers a newer label format when present and falls back to legacy formats or defaults when absent, ensuring consistent per-point labels.

* **Tests**
  * Added comprehensive unit tests covering label selection, precedence, NaN point filtering, and defaulting behavior across scenarios.

### DIFF
--- a/fiftyone/utils/sam.py
+++ b/fiftyone/utils/sam.py
@@ -387,9 +387,9 @@ def _to_sam_points(points, w, h, keypoint):
     scaled_points = np.array(points[valid_rows]) * np.array([w, h])
     labels = (
         np.array(keypoint.sam2_labels)[valid_rows]
-        if "sam2_labels" in keypoint
+        if "sam2_labels" in keypoint and keypoint.sam2_labels is not None
         else np.array(keypoint.sam_labels)[valid_rows]
-        if "sam_labels" in keypoint
+        if "sam_labels" in keypoint and keypoint.sam_labels is not None
         else np.ones(len(scaled_points))
     )
     return scaled_points.astype(np.float32), labels.astype(np.uint32)


fiftyone/utils/sam.py
```diff
@@ -387,7 +387,9 @@ def _to_sam_points(points, w, h, keypoint):
     scaled_points = np.array(points[valid_rows]) * np.array([w, h])
     labels = (
```
manushreegangwar, line 388: Have you tested this with `dataset.apply_model`? I would expect `Keypoint.labels` to be honored here since `sam_labels` and `sam2_labels` are not standard attributes of https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/labels.py#L1074

CharlesCNorton: @manushreegangwar ok I tested with `dataset.apply_model`, `sam2_labels` survives the round-trip and negative points visibly change the mask. `Keypoint` has `.label` (string) but no `.labels` — the existing code already uses these attrs so this just fixes the mismatch.

PR #6951, opened 2026-02-11
Fix/stay classy

mcdoh:

- fix erroneous "unknown label" error
- ensure new Classifications are added to the _Labels list_
- remove un-labeled Classification tags from the scene if abandoned

Locally

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Classification overlays now refresh correctly when overlays are added or removed, ensuring classifications recompute.
  * Uncommitted temporary labels no longer remain selected or leave the UI in an inconsistent interactive state.

* **Chores**
  * Split render vs storage paths so new labels are both rendered and persisted reliably.
  * Improved stability with unmount/stale guards and safer label-update flows.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
@@ -173,7 +173,16 @@ const pathMap = selector<{ [key: string]: string }>({
 });
 
 /**
- * Hook which provides a method for updating data in a label atom.
+ * Returns a callback that updates the {@link AnnotationLabelData} for a label
+ * identified by its overlay ID.
+ *
+ * The callback looks up the label's individual atom in the {@link labelMap},
+ * replaces its `data` field, and returns whether the update succeeded.
+ *
+ * @returns A callback with signature
+ *   `(id: string, data: AnnotationLabelData) => boolean` that returns `true`
+ *   if the label was found and updated, or `false` if no label with the given
+ *   ID exists.
  */
 const useUpdateLabelAtom = () => {
   return useAtomCallback(
@@ -295,6 +304,8 @@ export default function useLabels() {
   }, [active, removeOverlay, setLabels, setLoading]); // omit: [currentLabels]
 
   useEffect(() => {
+    // Flipped to `true` by the cleanup function so in-flight async work
+    // from a superseded effect invocation can bail out before mutating state.
     let stale = false;
 
     if (modalSample?.sample && active) {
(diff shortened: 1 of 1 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
```diff
@@ -177,16 +177,17 @@ const pathMap = selector<{ [key: string]: string }>({
...
 
       if (targetAtom) {
         const currentValue = get(targetAtom);
         set(targetAtom, { ...currentValue, data });
-      } else {
-        console.warn(`Unknown label id ${id}`);
+        return true;
       }
+
+      return false;
```
marcus-kestler, line 199: Are we returning false on success true on failure?  Is it ok to silently let this fail and return true?

mcdoh: If the label for the provided ID was found and that label is updated we return true. For IDs for which no label is found we return false. With a negative response we can assume we're dealing with a new label and add it to our collection.

app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
```diff
@@ -306,27 +310,62 @@ export default function useLabels() {
...
-          result.forEach((annotationLabel) => addLabel(annotationLabel));
+          result.forEach((annotationLabel) =>
+            addLabelToRenderer(annotationLabel)
+          );
           setLoading(LabelsState.COMPLETE);
         });
       } else if (loadingState === LabelsState.COMPLETE) {
         // refresh label data
         getLabelsFromSample().then((result) => {
+          if (stale) return;
```
erik-nieh, line 338: I don't fully understand this workflow, but what is happening in these cases? Are we abandoning some process early? It might just be worth adding a comment around `stale` to indicate what is happening/why we need to handle it.

app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
```diff
@@ -306,27 +310,62 @@ export default function useLabels() {
...
         });
       }
     }
-  }, [active, getFieldType, loadingState, modalSample?.sample, paths]);
+
+    return () => {
+      stale = true;
+    };
+  }, [
+    active,
```
erik-nieh, line 367: this large set of dependencies makes me think this should be more separate hooks but probably a refactor for later

app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
erik-nieh: Might be worth putting that as a comment on this line @mcdoh

erik-nieh (commented):
Just some comments, I'm not sure I know this well enough to approve

PR #6955, opened 2026-02-11
Fix/read only

mcdoh:

Make read-only read-only again.

https://github.com/user-attachments/assets/1135580a-0253-4308-914b-dcd3af37c284

https://github.com/user-attachments/assets/cf440e63-b3c2-4e02-aaa3-7c810cd79f8b

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed read-only field handling to properly prevent editing of detection overlays and annotations when associated fields are marked as read-only.
  * Detection boxes and quick-draw selections now respect read-only status, preventing unwanted modifications to protected field data.

### DIFF
--- a/app/packages/lighter/src/overlay/BoundingBoxOverlay.ts
+++ b/app/packages/lighter/src/overlay/BoundingBoxOverlay.ts
@@ -653,7 +653,10 @@ export class BoundingBoxOverlay
    * @param draggable - Whether the overlay should be draggable.
    */
   setDraggable(draggable: boolean): void {
-    this.isDraggable = draggable;
+    if (this.isDraggable !== draggable) {
+      this.isDraggable = draggable;
+      this.markDirty();
+    }
   }
 
   /**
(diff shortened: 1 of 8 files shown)


app/packages/lighter/src/overlay/BoundingBoxOverlay.ts
```diff
@@ -297,16 +297,23 @@ export class BoundingBoxOverlay
...
         Math.pow((point.y - this.moveStartPoint.y) / scale, 2)
     );
 
     if (distance > this.CLICK_THRESHOLD) {
       const resizeRegion = this.getResizeRegion(worldPoint, scale);
-      this.moveState = !this.hasValidBounds()
-        ? "SETTING"
-        : resizeRegion || "DRAGGING";
+      if (!this.hasValidBounds()) {
+        this.moveState = "SETTING";
```
erik-nieh, line 311: would be nice if these were enums but I see they existed this way before

PR #6963, opened 2026-02-11
Fix/absolute bounds

mcdoh:

The display of absolute bounds in the Annotation sidebar broke due to a fix I implemented earlier. I added a variable, `stale`, to avoid logging _"unknown label id"_ errors to the dev console. If we don't want to make this proposed change I can roll back the `stale` work.

locally

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal state management to improve performance and reliability in the annotation workflow.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
@@ -9,7 +9,7 @@ import {
   useModalSample,
 } from "@fiftyone/state";
 import { DETECTION } from "@fiftyone/utilities";
-import { atom, useAtom, useAtomValue, useSetAtom } from "jotai";
+import { atom, useAtomValue, useSetAtom } from "jotai";
 import { splitAtom, useAtomCallback } from "jotai/utils";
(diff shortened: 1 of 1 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
```diff
@@ -292,6 +292,13 @@ export default function useLabels() {
   const prevSliceRef = useRef(currentSlice);
   const updateLabelAtom = useUpdateLabelAtom();
 
+  // Use a ref for the loading state machine to avoid having it as an effect
+  // dependency. When loadingState was both a dep and mutated inside the effect,
+  // the UNSET→LOADING state change triggered a re-render whose cleanup set
+  // stale=true, causing the async result to be discarded and the state reset
+  // to UNSET — an infinite loop that prevented labels from ever loading.
+  const loadingRef = useRef(LabelsState.UNSET);
```
imanjra, line 300: I added a hook `useUnboundStateRef` which could be handy for things like this... It has an added benefit of upadting ref value as they change

PR #6973, opened 2026-02-12
keep Explore and Annotate in sync

mcdoh:

Keep Explore and Annotate Fields in sync.

locally

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added automatic synchronization of field activation between Explore and Annotate modes for fields with annotation schemas.

* **Bug Fixes**
  * Prevented a race condition that could leave label loading stuck when switching modes, ensuring toggling fields and mode changes reliably update activation state.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/useFieldActivationSync.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/useFieldActivationSync.ts
@@ -44,6 +44,10 @@ const useFieldActivationSync = () => {
 
   const activateOperator = useOperatorExecutor("activate_label_schemas");
   const deactivateOperator = useOperatorExecutor("deactivate_label_schemas");
+  const activateOperatorRef = useRef(activateOperator);
+  const deactivateOperatorRef = useRef(deactivateOperator);
+  activateOperatorRef.current = activateOperator;
+  deactivateOperatorRef.current = deactivateOperator;
 
   const prevRecoilRef = useRef<string[] | null>(null);
   const prevModeRef = useRef(mode);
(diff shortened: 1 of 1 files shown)


app/packages/core/src/components/Modal/Sidebar/useFieldActivationSync.ts
```diff
@@ -0,0 +1,137 @@
...
+        set(activeFields({ modal: true }), fields);
+      },
+    []
+  );
+
+  const addSchemas = useSetAtom(addToActiveSchemas);
+  const removeSchemas = useSetAtom(removeFromActiveSchemas);
+
+  const activateOperator = useOperatorExecutor("activate_label_schemas");
+  const deactivateOperator = useOperatorExecutor("deactivate_label_schemas");
```
lanzhenw, line 46: Annotation sidebar cannot mutate the `dataset.active_label_schemas`. The jotai atom is exclusively controlled by the schema manager

lanzhenw:
Since the requirement is "both explore/schema manager must agree", we can use a derived selector or atom that intersects the two sources. 

```
  // A read-only derived atom that computes the intersection
  export const visibleAnnotationFields = atom((get) => {
    const jotaiActive = get(activeLabelSchemas) ?? [];
    const recoilActive = /* read from recoil bridge */;
    return jotaiActive.filter(f => recoilActive.includes(f));
  });
```
We then need to make sure all the consumers in annotation sidebar read `visibleAnnotationFields` instead of `activeLabelSchemas`.

PR #6986, opened 2026-02-12
Add backspace as a delete shortcut. (Old API).

marcus-kestler:

Bind the backspace key to the delete command.
Had to use the raw old API b/c multiple keybindings of the same command are supported in the hooks for the old API.

Pressed the keys.

        notes for FiftyOne users.

-   [ ] Other

## Release Notes

**Bug Fixes**
- Improved undo/redo functionality for delete operations within the annotation editor, ensuring complete and proper restoration of annotation labels, overlays, and associated metadata
- Enhanced keyboard shortcut responsiveness and reliability for Delete and Backspace key commands when editing annotations
- Resolved potential state inconsistencies that could occur during annotation deletion workflows and undo/redo operations

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useDelete.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useDelete.ts
@@ -4,16 +4,16 @@ import { useLighter } from "@fiftyone/lighter";
 import * as fos from "@fiftyone/state";
 
 import { useAtomValue } from "jotai";
-import { useMemo } from "react";
+import { useCallback, useEffect, useMemo, useRef } from "react";
 import { useRecoilValue } from "recoil";
 import { current } from "./state";
 import useExit from "./useExit";
 import { useLabelsContext } from "../useLabels";
 import {
+  CommandContextManager,
   DelegatingUndoable,
   KnownCommands,
   KnownContexts,
-  useKeyBindings,
 } from "@fiftyone/commands";
 
 export default function useDelete() {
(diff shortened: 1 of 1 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useDelete.ts
```diff
@@ -28,115 +28,152 @@ export default function useDelete() {
   const exit = useExit();
   const setNotification = fos.useNotification();
 
-  const undoable = useMemo(() => {
+  // Use refs to capture latest values without causing re-renders
```
jleven, line 31: I appreciate you jumping in to try to get "backspace" to work, but I don't think this is behaving the way the comment implies.

jleven: As I understand it, this will still rerender whenever any of these values changes

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useDelete.ts
```diff
@@ -28,115 +28,152 @@ export default function useDelete() {
...
+    commandBusRef.current = commandBus;
+    removeLabelFromSidebarRef.current = removeLabelFromSidebar;
+    removeOverlayRef.current = removeOverlay;
+    setNotificationRef.current = setNotification;
+    exitRef.current = exit;
+    addLabelToSidebarRef.current = addLabelToSidebar;
+  });
+
+  // Create stable command handler that uses refs
+  const commandHandler = useCallback(() => {
```
jleven, line 56: Ah, you want the commandHandler to be stable, even as the values it uses change!

PR #6993, opened 2026-02-12
FOEPD-3133/FOEPD-3140: Annotate sidebar should also reflect explore sidebar visibility setting

lanzhenw:

This PR creates a jotai selector/derived atom for annotate sidebar to consume that is derived from both explore sidebar's active fields (recoil atom) and the schema manager's active label schemas (jotai atom). 

It does not sync bi-directionally to either active fields or active label schema. 

https://github.com/user-attachments/assets/dbcba84f-67fb-4e96-8229-df995a3fbb34

This also fixes the edge case (when there are active label schema, but some detection fields are not active in explore sidebar side)
Screenshot 2026-02-12 at 12 39 03 PM

disable the create buttons if the embeddedDoctype is not enabled in the explore sidebar

https://github.com/user-attachments/assets/530dd224-ac93-43a0-b6bd-18ceb6f3d826

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Field visibility is synchronized between the Explore sidebar and Annotate view; hiding or showing fields in one view updates the other and persists across modal mounts/unmounts.
  * Improved filtering for embedded document types so only relevant embedded fields appear in visible field lists.
  * Annotation action buttons (Classification, Detection, 3D Polyline/Cuboid) now disable when no applicable fields exist, preventing invalid actions.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/state.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/state.ts
@@ -1,25 +1,9 @@
-import {
-  CLASSIFICATION,
-  CLASSIFICATIONS,
-  DETECTION,
-  DETECTIONS,
-  POLYLINE,
-  POLYLINES,
-} from "@fiftyone/utilities";
 import { atom, useAtom } from "jotai";
 import { atomFamily } from "jotai/utils";
 import { capitalize } from "lodash";
 import { LabelSchemaMeta } from "./useSchemaManager";
 import { useMemo } from "react";
-
-const EMBEDDED_DOC_TYPES = new Set([
-  CLASSIFICATION,
-  CLASSIFICATIONS,
-  DETECTION,
-  DETECTIONS,
-  POLYLINE,
-  POLYLINES,
-]);
+import { PRIMITIVE_FIELD_TYPES } from "./SchemaManager/constants";
 
 // Tab state for GUI/JSON toggle
 export const activeSchemaTab = atom<"gui" | "json">("gui");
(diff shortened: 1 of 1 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/state.ts
```diff
@@ -24,6 +25,37 @@ export const labelSchemaData = atomFamily((field: string) => {
...
+ * pass through).
+ */
+export const exploreActiveFields = atom<string[] | null>(null);
+
+/**
+ * Intersection of activeLabelSchemas and exploreActiveFields.
+ * Display consumers should read this instead of activeLabelSchemas so that
+ * hiding a field in the Explore sidebar also hides it in Annotate.
+ */
+export const visibleLabelSchemas = atom((get) => {
```
tom-vx51, line 42: I might be missing something, but would it be easier to make this a hook (and remove `exploreActiveFields`)?

Something like
```typescript
export const useVisibleLabelSchemas = () => {
  const activeExploreFields = useRecoilValue(activeFields);
  const activeAnnotateFields = useAtomValue(activeLabelSchemas);

  return activeAnnotateFields.filter(...)
};
```

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/state.ts
```diff
@@ -198,11 +198,11 @@ export const isNew = atom((get) => {
...
 });
 
-const fieldsOfType = atomFamily((type: LabelType) =>
+export const fieldsOfType = atomFamily((type: LabelType) =>
   atom((get) => {
     const fields = new Array<string>();
 
-    for (const field of get(activeLabelSchemas) ?? []) {
+    for (const field of get(visibleLabelSchemas) ?? []) {
       if (type && IS[type].has(get(fieldType(field)))) {
```
lanzhenw, line 206: @tom-vx51 the tricky part is `fieldsOfType` reads visibleLabelSchemas inside the jotai getter function. 

Since we can't call a hook inside a Jotai atom getter, we'd either need to refactor that whole atom chain into hooks too, or still sync via useEffect to a writable atom — which ends up being the same bridge pattern.

Open to suggestions if you see a cleaner way around this

erik-nieh: If we can I'd like to avoid these atom chains in the future, it made several things more difficult in HA.

tom-vx51: oh, I see. Agreed that we need to clean these up; they aren't helping us at this point. That'll be a bigger task and we can do it in its own set of PRs.

erik-nieh (approved):

PR #6998, opened 2026-02-12
Fix compute metadata for url without Content-Length

swheaton:

- Metadata and ImageMetadata get their size from "Content-Length" header but if it's not there, they will consume the full buffer to get the size as a backup.

added unit tests for this plus other missing behavior

manual simulated testing

```python
# chunked_server.py
import os
from http.server import HTTPServer, BaseHTTPRequestHandler

DATA_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "data")
CHUNK_SIZE = 8192

class ChunkedHandler(BaseHTTPRequestHandler):
    def do_GET(self):
        path = os.path.normpath(os.path.join(DATA_DIR, self.path.lstrip("/")))
        if not path.startswith(DATA_DIR):
            self.send_response(403)
            self.end_headers()
            return

        try:
            with open(path, "rb") as f:
                data = f.read()
        except FileNotFoundError:
            self.send_response(404)
            self.end_headers()
            return

        self.send_response(200)
        self.send_header("Transfer-Encoding", "chunked")
        self.end_headers()

        for i in range(0, len(data), CHUNK_SIZE):
            chunk = data[i : i + CHUNK_SIZE]
            self.wfile.write(f"{len(chunk):X}\r\n".encode())
            self.wfile.write(chunk)
            self.wfile.write(b"\r\n")
        self.wfile.write(b"0\r\n\r\n")


if __name__ == "__main__":
    host, port = "0.0.0.0", 9000
    os.makedirs(DATA_DIR, exist_ok=True)
    print(f"Serving files from {DATA_DIR} on http://{host}:{port}")
    HTTPServer((host, port), ChunkedHandler).serve_forever()
```
run `python chunked_server.py`

```python
import os
import shutil
import fiftyone as fo

data_path = "/path/to/server/data_dir"
image_file_path = "/path/to/image.jpg"
image_name = os.path.basename(image_file_path)

m1 = fo.ImageMetadata.build_for("http://localhost:9000/" + image_name)
m2 = fo.ImageMetadata.build_for(image_file_path)
assert m1 == m2

m1 = fo.Metadata.build_for("http://localhost:9000/" + image_name)
m2 = fo.Metadata.build_for(image_file_path)
assert m1 == m2
```

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Improvements**
  * More robust metadata extraction for remote assets: reliably computes remote content size when Content-Length is missing and improves image dimension handling for URL sources, reducing incorrect size/dimension reports.

* **Tests**
  * Added extensive unit tests covering URL and local metadata extraction, streaming responses with/without Content-Length, seek/read behaviors, and multiple image formats.

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -3241,7 +3241,12 @@ class ResponseStream(object):
         """
         size = self._response.headers.get("Content-Length")
         if size is not None:
-            return int(size)
+            try:
+                size = int(size)
+                if size >= 0:
+                    return size
+            except Exception:
+                pass
 
         self._bytes.seek(0, io.SEEK_END)
         size = self.tell() + sum(map(len, self._iterator))


fiftyone/core/utils.py
```diff
@@ -3220,12 +3220,34 @@ def read(self, size=None):
...
-            self._bytes.seek(position, whence)
+        self._bytes.seek(position, whence)
+        return self.tell()
 
     def tell(self):
         return self._bytes.tell()
 
+    def consume_size(self):
+        """Returns the total size of the response content in bytes but
+        invalidates the stream for subsequent reads.
```
exupero: The indeterminism feels risky and worth calling out:
```suggestion
        may invalidate the stream for subsequent reads.
```

tests/unittests/metadata_tests.py
```diff
@@ -5,15 +5,140 @@
...
+        img_data = _make_test_image(width=50, height=50, mode="L")
+        mock_get.return_value = _make_mock_response(
+            img_data, content_length=len(img_data)
+        )
+
+        metadata = fom.ImageMetadata.build_for("http://example.com/gray.png")
+
+        self.assertEqual(metadata.width, 50)
+        self.assertEqual(metadata.height, 50)
+        self.assertEqual(metadata.num_channels, 1)
```
exupero, line 89: Should this test have a check of `metadata.size_bytes`?

fiftyone/core/utils.py
swheaton: Was trying to convey via docstring and name - "consume". Got any better naming ideas?
Alternatively I could just scrap the optimization

swheaton:
> Do we know why `Content-Length` is missing? Is it always missing, or only sometimes? Does it depend on the user's images?

I'm not sure exactly but a user was seeing it. Some potentials:
- server side encryption
- multipart uploads
- some type of proxy in between
- some load balancers
- chunked transfer-encoding

> Does this behavior introduce a backwards incompatibility?
It should not because we will always check the header first which is what it did before. Previously it would just fail in this case.

fiftyone/core/utils.py
exupero: I'm fine with the method name, I was just explaining why I suggested changing "invalidates" to "may invalidate", since the method isn't guaranteed to consume the stream and that may be important in some scenarios, so the docstring should identify that it doesn't always invalidate the stream.

swheaton:
> > > Does this behavior introduce a backwards incompatibility?
> > 
> > 
> > It should not because we will always check the header first which is what it did before. Previously it would just fail in this case.
> 
> Currently, when the `Content-Length` header is missing, does it raise an error? Or does it return `0` or something else that `int` can handle but which isn't correct? If it raises an error, I agree this isn't a backwards incompatibility. If it was not erroring and the body content was not consumed but now it is, that seems like a backwards incompatibility.

It would just throw a KeyError. Also the request isn't returned or anything, it's just consumed in this local context which is why I don't care if we throw away all the bytes

PR #7001, opened 2026-02-12
FOEPD-3179: [HA] Reset labels when sample ID changes

sashankaryal:
When we navigate between samples, we need `labels` to reset so that sidebar reflects new labels for the given sample as per the schema. In 2D, this is handled by a cleanup function in a `useEffect()` that has `scene` has a dependency and resets labels when `scene` changes. However, we need this to run when sample ID changes, too.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved annotation label state reset when switching between samples.

* **Refactoring**
  * Optimized internal state management for better component lifecycle handling and stability during annotation operations.

### DIFF
--- a/app/packages/state/src/accessors/modal.ts
+++ b/app/packages/state/src/accessors/modal.ts
@@ -93,4 +93,8 @@ export const usePreferredGroupAnnotationSlice = () =>
 /**
  * Gets the current sample ID.
  */
-export const useCurrentSampleId = () => useRecoilValue(currentSampleId);
+export const useCurrentSampleId = () => {
+  const loadable = useRecoilValueLoadable(currentSampleId);
+
+  return loadable.state === "hasValue" ? loadable.contents : null;
+};


app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
```diff
@@ -300,32 +295,23 @@ export default function useLabels() {
...
-        throw new Error("no type");
-      }
-
-      return type as LabelType;
-    },
+    ({ snapshot }) =>
+      async (path: string) => {
+        const loadable = await snapshot.getLoadable(field(path));
+        const type = loadable
+          .getValue()
```
sashankaryal: all formatting change by precommit hook - i changed nothing

app/packages/core/src/components/Modal/Sidebar/Annotate/useLabels.ts
```diff
@@ -300,32 +295,23 @@ export default function useLabels() {
...
 
-  // This effect resets labels when the annotation slice changes for grouped datasets
-  useEffect(() => {
-    if (prevSliceRef.current !== currentSlice && currentSlice) {
-      prevSliceRef.current = currentSlice;
-      setLabels([]);
-      loadingRef.current = LabelsState.UNSET;
-      setLoading(LabelsState.UNSET);
-    }
-  }, [currentSlice, setLabels, setLoading]);
```
sashankaryal, line 327: this effect subsumed by effect below which adds dependency on `modalSample`

app/packages/state/src/accessors/modal.ts
```diff
@@ -88,3 +89,8 @@ export const useModalSampleSchema = (): Schema =>
  */
 export const usePreferredGroupAnnotationSlice = () =>
   useAtom(preferredGroupAnnotationSliceAtom);
+
+/**
+ * Gets the current sample ID.
+ */
+export const useCurrentSampleId = () => useRecoilValue(currentSampleId);
```
tom-vx51: this claims to be a loadable here https://github.com/voxel51/fiftyone/blob/develop/app/packages/operators/src/state.ts#L875 - am I misinterpreting or should this also?

sashankaryal: good point ... will use `useRecoilValueLoadable`

tom-vx51 (approved):

PR #7003, opened 2026-02-13
Fix management SDK docs failing to render

AdonaiVera:

The Enterprise Management SDK documentation pages are rendering [empty](https://docs.voxel51.com/enterprise/management_sdk.html), all automodule directives silently fail because Sphinx cannot import the management modules.

Commit `f6a4e7e7bc` in `fiftyone-teams` (`[FOEPD-2828] Allow FOM to be called using CTX`) added new imports to `fiftyone/management/connection.py`:

```python
import fiftyone.internal.api_requests as fiar
import fiftyone.internal.util as fiu
```

These modules exist only in **fiftyone-teams**, not in the open-source `fiftyone` package.

During the docs build, `connection.py` is symlinked from `fiftyone-teams`, but the `internal/` directory still comes from the OSS `fiftyone` package. Because of this mismatch, the imports fail and Sphinx silently skips all Management SDK pages, leaving them empty.

I fixed this during the docs build by symlinking the `internal/` directory from `fiftyone-teams` and also symlinking `utils/decorators.py`. This follows the same pattern already used for `management/` and `api/`. The `internal/` directory is excluded from `sphinx-apidoc`, so no enterprise-only code is exposed in the generated docs.

## How is this patch tested?

* Built the docs locally
* Deployed to [dev](https://docs.dev.voxel51.com/enterprise/management_sdk.html)
* Verified that the Management SDK docs render correctly

* [x] No. You can skip the rest of this section.

* [x] Build: Build and test infrastructure changes
* [x] Documentation: FiftyOne documentation changes
* [ ] Other

[FOEPD-2828]: https://voxel51.atlassian.net/browse/FOEPD-2828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * Switched docs build to an environment-driven approach for optional team resources; no symlink creation or post-generation cleanup remains.
  * Removed the separate docs-specific dependency file and eliminated installing the local team package in the docs CI.

* **Documentation**
  * Enabled conditional AutoAPI integration when an environment path is provided and updated doc directives to use AutoAPI for service components.
  * Simplified README instructions accordingly.

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -7,7 +7,7 @@ typing_extensions==4.15.0
 
 # precommit requirements
 pre-commit==4.5.1
-pylint==4.0.4
+pylint==3.3.9
 jupyter-client==8.8.0
 
 # docs requirements
@@ -41,4 +41,4 @@ sphinxcontrib-jquery==4.1
 sphinx-pushfeedback==0.1.8
 sphinx-docsearch==0.2.0
 sphinx-remove-toctrees==1.0.0.post1
-sphinx-autoapi>=3.6.1
+sphinx-autoapi==3.6.1


afoley587 (approved):
This LGTM!

Just a thought - should other repos that have docs publish their own? Or do we like centralized builds? Breaking them apart would let us independently publish things and adding new modules wouldn't break things like this.

Will defer to you - just a thought!

docs/generate_docs.bash
```diff
@@ -88,6 +88,28 @@ if [[ -n "${PATH_TO_TEAMS}" ]]; then
 
     ln -sfn "${PATH_TO_TEAMS}/fiftyone/management" "${PATH_TO_FIFTYONE_DIR}/management"
     ln -sfn "${PATH_TO_TEAMS}/fiftyone/api" "${PATH_TO_FIFTYONE_DIR}/api"
+    rm -rf "${PATH_TO_FIFTYONE_DIR}/internal.bak"
+    mv "${PATH_TO_FIFTYONE_DIR}/internal" "${PATH_TO_FIFTYONE_DIR}/internal.bak"
```
brimoor: Hmm, the more symlinking of files/directories from FOE into OSS we do, the messier and messier this is getting.

On one hand, if this leaves the OSS files in a dirty state after a docs build, the changes are to files/directories under source control, so `git diff` *should* show it, right?

On the other hand, it would definitely be desirable to solve this in a way that required no linking of files at all, including the existing `management/` and `api/` folders, which seemed okay at the time but now that the necessary surgery is getting more complex, feels like a less tenable solution.

I wonder if there's a way to build the FOE-specific docs that uses a fully separate FOE install?

Curious what @swheaton thinks here as the person who I believe is most familiar with the Management SDK. Both about @AdonaiVera's other ideas [here](https://github.com/voxel51/fiftyone/pull/7003#issuecomment-3895135743) and my thoughts above.

brimoor: To be clear, I'm okay merging this PR as a temporary solution as long as the cleanup is complete and doesn't result in any `git diff` in OSS. Definitely need the docs builds working again ASAP!

AdonaiVera: Hi @brimoor
I modified the approach in this PR. No symlinking at all anymore, instead of the symlink approach, we're now using sphinx-autoapi, which parses Python source files via static analysis, no imports, no mocks, no symlinks, no pip install of fiftyone-teams needed. I tested in local and generate the enterprise docs correctly, now testing in CI here.

docs/source/enterprise/management_sdk.rst
```diff
@@ -27,15 +27,15 @@ API reference
 Connections
 -----------
 
-.. automodule:: fiftyone.management.connection
+.. autoapimodule:: fiftyone.management.connection
    :members: test_api_connection, reload_connection
```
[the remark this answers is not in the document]
AdonaiVera: Thank you, Rabbit, but I think this is not a real issue because Sphinx just ignores unknown directives with a warning. This means the Enterprise Management SDK will be empty locally if the user doesn’t want to render it, which is fine.

requirements/docs.txt
```diff
@@ -41,3 +41,4 @@ sphinxcontrib-jquery==4.1
 sphinx-pushfeedback==0.1.8
 sphinx-docsearch==0.2.0
 sphinx-remove-toctrees==1.0.0.post1
+sphinx-autoapi>=3.6.1
```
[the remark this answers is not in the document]
AdonaiVera: Good catch!

PR #7008, opened 2026-02-14
FOEPD-3181: toggle field annotation visibility also active on save or disgard

lanzhenw:

In the “Edit field schema” part of the Schema Manager, the “Visible field” toggle takes immediate effect: When the user changes the state of the toggle, the field is immediately moved to/from the annotation schema and either “Active fields” or “Hidden fields” section of the Schema Manager UI.

Desired Behavior
Changing the state of the toggle doesn’t take immediate effect. The user must click the “Save” button in “Edit field schema” for it to take effect.

changed visibility of multiple fields, then save: the field visibility would not move until user clicks "save"

(Details)

        notes for FiftyOne users.

Changing the state of the toggle doesn’t take immediate effect. The user must click the “Save” button in “Edit field schema” for it to take effect.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Introduced per-field visibility management with automatic persistence when saving schema changes.

* **Improvements**
  * Streamlined field visibility logic in schema editing workflow for enhanced schema manager integration.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
@@ -17,7 +17,10 @@ import {
   labelSchemaData,
   removeFromActiveSchemas,
 } from "../../state";
-import { useSchemaManager } from "../../useSchemaManager";
+import {
+  useSchemaManager,
+  type UpdateSchemaRequest,
+} from "../../useSchemaManager";
 import { currentLabelSchema } from "../state";
 import { reconcileComponent } from "../utils";
 
@@ -154,7 +157,10 @@ const useSave = (field: string, visibilityChanged: boolean) => {
       const labelSchema = current ? reconcileComponent(current) : current;
 
       try {
-        await updateSchema({ field, label_schema: labelSchema });
+        await updateSchema({
+          field,
+          label_schema: labelSchema,
+        } as UpdateSchemaRequest);
       } catch (error) {
         console.error("Failed to save label schema:", error);
         setIsSaving(false);
@@ -313,12 +319,11 @@ export default function useLabelSchema(field: string) {
   // this view). Visibility is local state so it resets automatically on unmount.
   // After a successful save, savedLabelSchema is already updated before unmount,
   // so re-opening the field will show the saved state correctly.
-  const setCurrentSchema = useSetAtom(currentLabelSchema(field));
   useEffect(() => {
     return () => {
-      setCurrentSchema(undefined);
+      currentLabelSchema.remove(field);
     };
-  }, [setCurrentSchema]);
+  }, [field]);
 
   // Wrap discard to also revert visibility
   const originalDiscard = validate.discard;
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/hooks.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/hooks.ts
@@ -37,7 +37,6 @@ import {
   hiddenFieldTypes,
   isNewFieldMode,
   jsonValidationErrors,
-  pendingActiveSchemas,
   selectedActiveFields,
   selectedHiddenFields,
   sortedInactivePaths,
@@ -650,14 +649,11 @@ export const useIsLargeDataset = () => {
  */
 export const useSchemaManagerCleanup = () => {
   const setCurrentFieldAtom = useSetAtom(currentField);
-  const setPendingActiveSchemas = useSetAtom(pendingActiveSchemas);
 
   useEffect(() => {
     return () => {
       // Reset field editing state
       setCurrentFieldAtom(null);
-      // Reset pending visibility state
-      setPendingActiveSchemas(null);
     };
   }, []);
 };
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/state.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/state.ts
@@ -186,16 +186,6 @@ export const hasJsonChanges = atom((get) => {
 });
 
 // =============================================================================
-// Pending Visibility State
-// =============================================================================
-
-/**
- * Pending active schemas list. Mirrors activeLabelSchemas but only persisted
- * on save via set_active_label_schemas. null means not yet initialized.
- */
-export const pendingActiveSchemas = atom<string[] | null>(null);
-
-// =============================================================================
 // Edit Field Label Schema State
 // =============================================================================
 


app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -64,6 +65,77 @@ const useHasChanges = (one: unknown, two: unknown) => {
   }, [one, two]);
 };
 
+const sameSet = (a: string[], b: string[]): boolean => {
```
tom-vx51: don't think it matters here, but this assumes that the elements in the arrays are distinct

For example:
```typescript
sameSet(["foo", "foo"], ["foo", "bar"]) === true
```

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -64,6 +65,77 @@ const useHasChanges = (one: unknown, two: unknown) => {
...
+
+  const isFieldVisible = effective.includes(field);
+
+  const visibilityChanged = useMemo(
+    () => !sameSet(effective, activeSchemas ?? []),
+    [effective, activeSchemas]
+  );
+
+  const toggleVisibility = useCallback(() => {
+    const current = pending ?? activeSchemas ?? [];
```
tom-vx51: [nit] can use `effective` here

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -164,6 +240,37 @@ const useSave = (field: string) => {
...
 
+          // Persist pending visibility changes
+          if (visibilityChanged && pending) {
+            // Build ordered list: old items first, new appended at end
+            const oldSet = new Set(activeSchemas ?? []);
+            const kept = (activeSchemas ?? []).filter((f) =>
+              pending.includes(f)
+            );
+            const added = pending.filter((f) => !oldSet.has(f));
+            const ordered = [...kept, ...added];
```
tom-vx51: it seems like we already maintain the pending order in `useVisibility`, do we need this additional logic?

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -164,6 +240,37 @@ const useSave = (field: string) => {
...
+
+            setActiveSchemas.execute(
```
tom-vx51: noting that `useSchemaManager()` provides a typed interface for this https://github.com/voxel51/fiftyone/blob/develop/app/packages/core/src/components/Modal/Sidebar/Annotate/useSchemaManager.ts#L209-L216

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
lanzhenw: The additional logic is for an edge case: restoring original position for fields toggled off then on. 

e.g.
active fields: ["a", "b", "c"], user toggles "a" off, and then toggles "a" on again. 
pending would be = ["b", "c", "a"]
with save handler it would still be ["a", "b", "c"].

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
lanzhenw: i will add a comment

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -64,6 +66,80 @@ const useHasChanges = (one: unknown, two: unknown) => {
   }, [one, two]);
 };
 
+// Assumes elements in each array are distinct (true for field name lists).
+const sameSet = (a: string[], b: string[]): boolean => {
```
erik-nieh: I think this is fine but lodash has an equivalent that I believe is _more_ restrictive (no distinct assumption): https://lodash.com/docs/#isEqual

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -64,6 +66,80 @@ const useHasChanges = (one: unknown, two: unknown) => {
...
+    if (effective.includes(field)) {
+      setPending(effective.filter((f) => f !== field));
+    } else {
+      // Append new field at the end
+      setPending([...effective, field]);
+    }
+  }, [field, effective, setPending]);
+
+  const discardVisibility = useCallback(() => {
+    const current = pending ?? activeSchemas ?? [];
```
erik-nieh: I think `current` and `effective` are the same? might be able to use just `effective`

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -64,6 +66,80 @@ const useHasChanges = (one: unknown, two: unknown) => {
...
+    const wasActive = original.includes(field);
+    const isPending = current.includes(field);
+
+    if (wasActive === isPending) return; // no change for this field
+
+    if (wasActive && !isPending) {
+      // Was active, user hid it — revert by re-inserting at original position
+      const newPending = [...current];
+      const originalIndex = original.indexOf(field);
+      let insertAt = newPending.length;
```
erik-nieh: might be worth moving `insertAt` to it's own function for readability

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -64,6 +66,80 @@ const useHasChanges = (one: unknown, two: unknown) => {
...
+  // Lazily initialize pending from current active schemas
+  const effective = useMemo(
+    () => pending ?? activeSchemas ?? [],
+    [pending, activeSchemas]
+  );
+
+  const isFieldVisible = effective.includes(field);
+
+  const visibilityChanged = useMemo(
+    () => !sameSet(effective, activeSchemas ?? []),
```
erik-nieh: will this return true even if a change was made to a totally different `field`? That might be fine or even intentional I'm not completely sure.

erik-nieh (dismissed):
Some comments but generally looks good!

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
lanzhenw: Yes, currently not only visibility, we allow user to modify a field, go to a different field continue to make changes, and save all of them at once.

lanzhenw: @erik-nieh Thanks for raising the concern. I checked with Mike and verified we actually want to save per-field. 
In the new commits, I updated the logic so that field changes and visibility changes will be saved/discarded per field.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
lanzhenw: replaced with lodash. good call!

PR #7012, opened 2026-02-14
[FOEPD-2257] Enable annotating PatchesViews

kaixi-wang:

Enable editing detection patches while in a patches view

Implemented:
- [x]  Annotate mode enabled
- [x] Can edit and save label field attribute values
- [x] Can not change field
- [x] Can not delete patches
- [x] Can not create new patches
- [x] In Annotate mode, only the edit view is available (no list view in Annotation mode) and the back arrow goes to Explore mode

TODO/Next Up:
- [ ] Sync zoom between annotate mode and explore (copy explore behavior) [FOEPD-3003]
- [ ]  Handle patches view empty Annotate side panel if patches field is not an active field [FOEPD-3261]

(Details)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added generated-view-aware annotation and patching (field-level patch metadata, optional routing to source dataset).
  * Auto-select single labels when editing in patches view.

* **Bug Fixes**
  * Improved patch error handling and delete behavior for generated views.
  * Made patches/patches-view UI respect read-only and unsupported actions.

* **Chores**
  * Added diagnostic logging and Extended JSON transformation for patch payloads.
  * Tests updated to cover generated-view flows.

[FOEPD-3003]: https://voxel51.atlassian.net/browse/FOEPD-3003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FOEPD-3261]: https://voxel51.atlassian.net/browse/FOEPD-3261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/packages/annotation/src/util/labelPersistence.test.ts
+++ b/app/packages/annotation/src/util/labelPersistence.test.ts
@@ -45,7 +45,8 @@ describe("handleLabelPersistence", () => {
 
     vi.mocked(buildLabelDeltas).mockReturnValue([]);
     vi.mocked(buildJsonPath).mockImplementation(
-      (basePath, deltaPath) => `${basePath ?? ""}.${deltaPath}`
+      (basePath, deltaPath) =>
+        `/${[...(basePath?.split(".") || []), deltaPath].join("/")}`
     );
     vi.mocked(buildAnnotationPath).mockReturnValue("predictions.detections");
   });


app/packages/annotation/src/deltas.ts
```diff
@@ -54,6 +54,26 @@ const isFieldType = (field: Field, fieldType: FieldType): boolean => {
...
+  label: LabelProxy,
+  isGenerated: boolean
+): string => {
+  let basePath = label.path;
+  if (isGenerated) {
+    // Patches views flatten the structure so we need to adjust the path to reflect the src sample.
+    if (label.type === "Detection") basePath = `${basePath}.detections`;
+  }
+  return basePath;
+};
```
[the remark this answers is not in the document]
kaixi-wang: Classifications and polylines are not in patches views

app/packages/annotation/src/persistence/useAnnotationDeltaSupplier.ts
```diff
@@ -14,10 +14,17 @@ export const useAnnotationDeltaSupplier = (): DeltaSupplier => {
...
+    const resultLighter = supplyLighterDeltas();
+    const resultSidebar = supplySidebarDeltas();
+
+    const deltas = [
+      ...result3d.deltas,
+      ...resultLighter.deltas,
+      ...resultSidebar.deltas,
     ];
+
+    // Metadata for generated views only comes from Lighter
```
tom-vx51, line 27: will this always be true? I don't know enough about generated views to say whether it precludes 3d etc.

app/packages/annotation/src/persistence/useLighterDeltaSupplier.ts
```diff
@@ -53,14 +57,36 @@ const buildAnnotationLabel = (overlay: BaseOverlay): LabelProxy | undefined => {
...
+    }
+
+    let metadata;
+    if (isGenerated && firstChangedOverlay) {
+      // Must include label metadata for generated views
+      // so that the backend can identify the correct label to update
+      const labelProxy = buildAnnotationLabel(firstChangedOverlay);
+      if (labelProxy) {
+        metadata = {
+          labelId: (labelProxy.data as { _id?: string })._id,
```
tom-vx51: can we just reference `overlay.id` here? Or `overlay.label._id`? The cast makes us lose type safety

app/packages/annotation/src/persistence/useLighterDeltaSupplier.ts
```diff
@@ -53,14 +57,36 @@ const buildAnnotationLabel = (overlay: BaseOverlay): LabelProxy | undefined => {
...
+          firstChangedOverlay = overlay;
+        }
+      }
+    }
+
+    let metadata;
+    if (isGenerated && firstChangedOverlay) {
+      // Must include label metadata for generated views
+      // so that the backend can identify the correct label to update
+      const labelProxy = buildAnnotationLabel(firstChangedOverlay);
```
tom-vx51, line 84: this sounds like we can only support a single label update, but the loop above doesn't enforce that limit. Should this be a separate block entirely? e.g. `if (isGeneratedView) { // stricter enforcement of patch rules } else { // existing behavior }`?

app/packages/annotation/src/util/labelPersistence.test.ts
```diff
@@ -259,13 +289,54 @@ describe("handleLabelPersistence", () => {
...
+    expect(buildLabelDeltas).toHaveBeenCalledWith(
+      mockSample,
+      mockAnnotationLabel,
+      mockSchema,
+      mockOpType,
+      true
+    );
+    expect(buildJsonPath).toHaveBeenCalledWith(null, "label");
+    expect(mockPatchSample).toHaveBeenCalledWith(
+      [{ path: "/label", value: "cat", op: "replace" }],
```
tom-vx51, line 335: noticing this path is prefixed with `/` which is omitted from the other deltas. Looks to be an artifact of passing `null` and wanted to make sure this is intentional

app/packages/annotation/src/util/labelPersistence.ts
```diff
@@ -48,13 +69,23 @@ export const doPatchSample = async ({
   if (sampleDeltas.length > 0) {
     try {
       let updatedSample: Sample;
+      // For generated views, use _sample_id so that the sampleId and datasetId
+      // always refer to the persistent source.
+      const sampleId = isGenerated
+        ? (sample as Sample & { _sample_id?: string })._sample_id
```
tom-vx51, line 75: not your fault but the sample types suck :(

app/packages/annotation/src/util/labelPersistence.ts
```diff
@@ -48,13 +69,23 @@ export const doPatchSample = async ({
...
 
       try {
         const response = await patchSample({
           datasetId,
-          sampleId: sample._id,
+          sampleId: sampleId,
           deltas: sampleDeltas,
           versionToken,
+          // Pass generated view parameters for field-level updates
+          labelId: isGenerated ? labelId : undefined,
```
tom-vx51: [nit] these come in as `undefined` if it's not a generated view, right? If so we can omit the ternary and just pass them directly

app/packages/annotation/src/util/labelPersistence.ts
```diff
@@ -70,6 +101,12 @@ export const doPatchSample = async ({
         }
       }
 
+      if (isGenerated && opType === "delete") {
+        // Response contains the source sample and the patch sample was deleted
```
tom-vx51: Would the existing source sample still have this deleted label though?

app/packages/annotation/src/util/labelPersistence.ts
```diff
@@ -85,6 +122,18 @@ export const doPatchSample = async ({
         console.warn("received empty sample data; deltas may be stale");
       }
     } catch (error) {
+      // For delete operations on generated views (patches/clips/frames), a 404
```
tom-vx51, line 119: can you elaborate here? what's returning the 404?

app/packages/annotation/src/util/labelPersistence.ts
```diff
@@ -138,12 +197,23 @@ export const handleLabelPersistence = async ({
...
+    // Convert label delta to sample delta
+    // Operation path is the label path for patches views
+    path: buildJsonPath(isGenerated ? null : annotationLabel.path, delta.path),
   }));
 
-  return await applyPatch(sampleDeltas);
+  // For SampleField patch updates on generated views
+  const patchOptions = isGenerated
+    ? {
+        labelId: (annotationLabel as { data?: { _id?: string } }).data?._id,
```
tom-vx51, line 206: Noting for myself we should clean this up. This path is only used for delete and is a subset of `useGetLabelDelta`

app/packages/annotation/src/deltas.ts
```diff
@@ -324,42 +370,40 @@ export const buildDetectionMutationDelta = (
...
-  upsertArrayElement(
-    newArray,
-    mergedDetection,
-    (det) => det._id === label.data._id
-  );
-
-  const newLabel = {
+  const { detections } = existingLabel;
+  const exists = detections.some((det) => det._id === label.data._id);
+  const newDetections = exists
```
tom-vx51: can we leave this logic as it was for consistency with other methods here? We can still set `existingDetection` before the `if (isGenerated)` and use the value in that block

app/packages/core/src/client/annotationClient.ts
```diff
@@ -140,15 +147,47 @@ const doFetch = <A, R>(
...
+  // This enables syncing changes from the source to the currently loaded generated dataset.
+  // We do this instead of making a separate request with the generated dataset _id
+  // because we want to ensure permissions are checked against the src dataset.
+  const queryParams = new URLSearchParams();
+  if (request.generatedDatasetName) {
+    if (
+      !request.generatedSampleId ||
+      request.generatedSampleId === request.sampleId
+    ) {
+      throw new MalformedRequestError(
```
tom-vx51: Can we change this to a plain `Error`? `MalformedRequestError` implies that the server rejected it

app/packages/core/src/client/annotationClient.ts
```diff
@@ -140,15 +147,47 @@ const doFetch = <A, R>(
...
+  }
+
+  const queryString = queryParams.toString();
+  const pathWithQuery = queryString
+    ? `${encodeURIPath(pathParts)}?${queryString}`
+    : encodeURIPath(pathParts);
+
+  // Convert ObjectId strings to Extended JSON ({ $oid: "..." }) so the server
+  // can deserialize them as bson.ObjectId rather than storing plain strings.
+  const deltas = request.deltas.map((delta) =>
```
tom-vx51, line 183: not a big deal, but this method is intended to be "dumb" and only know how to interact with the server endpoint. If we can move this transformation out of this method I think it would make more sense with other business logic

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/Header.tsx
```diff
@@ -92,11 +103,20 @@ const Header = () => {
   const isAnnotatingPolyline = current3dAnnotationMode === "polyline";
   const isAnnotatingCuboid = current3dAnnotationMode === "cuboid";
 
-  const handleExit = () => {
+  // In patches view with single label, clicking back should go to explore mode
```
tom-vx51, line 108: is this a product requirement or a technical requirement?

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/Header.tsx
```diff
@@ -92,11 +103,20 @@ const Header = () => {
...
+  const isPatches = useRecoilValue(fos.isPatchesView);
+  const labelCount = useAtomValue(labels).length;
+  const setModalMode = useSetAtom(fos.modalMode);
+  const shouldExitToExplore = isPatches && labelCount === 1;
+
+  const handleExit = useCallback(() => {
+    if (shouldExitToExplore) {
+      void setModalMode(fos.ModalMode.EXPLORE);
```
tom-vx51: this should use `exitAnnotationMode()` from `useAnnotationController` which handles some additional logic https://github.com/voxel51/fiftyone/blob/develop/app/packages/annotation/src/hooks/useAnnotationController.ts#L43

app/packages/core/src/components/Modal/Sidebar/Annotate/Annotate.tsx
```diff
@@ -65,22 +67,65 @@ const Loading = () => {
...
+  const isPatches = useRecoilValue(isPatchesView);
+  const isGenerated = useRecoilValue(isGeneratedView);
+  const [entries] = useEntries();
+  const { scene } = useLighter();
+  const loadingState = useAtomValue(labelsState);
+
+  // Track the previous loading state to detect fresh loads
+  const prevLoadingState = useRef<LabelsState>(LabelsState.UNSET);
+
+  // Auto-edit single label in patches view
```
tom-vx51: iiuc this is meant to replicate the quick-edit flow; can we reuse that logic instead? I suspect there are race conditions here w.r.t the scene/overlays not being initialized when this runs.

app/packages/core/src/components/Modal/Sidebar/Annotate/useFocus.ts
```diff
@@ -48,13 +51,25 @@ export default function useFocus() {
           return;
         }
 
+        // In generated views (patches/clips/frames), don't exit edit mode on deselect
```
tom-vx51, line 54: I'm curious why we want this - currently we treat no selection as a valid state irrespective of the number of labels

app/packages/annotation/src/persistence/useAnnotationDeltaSupplier.ts
kaixi-wang: I double checked with @sashankaryal that it is currently true (that we don't support 3d patches). I don't know about the roadmap.

In the patches creation code it has: 
```
 fova.validate_collection(sample_collection, media_type=fom.IMAGE)
```

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/Header.tsx
kaixi-wang: product request to not have the list view, but without list view, there isn't a way to get to the explore view, so this felt like the most natural/simple fix

app/packages/annotation/src/deltas.ts
kaixi-wang: sure. so the new array was intentional?

app/packages/core/src/client/annotationClient.ts
kaixi-wang: where do you suggest? My thinking is that the transform to objectId is just for the request and part of the serialization/transport layer. I don't really see it as business logic as it doesn't have any functional meaning in the application layer.

app/packages/annotation/src/util/labelPersistence.ts
kaixi-wang: I was seeing requests to the backend for the deleted patches sample that were returning 404s.

app/packages/annotation/src/util/labelPersistence.ts
kaixi-wang: oh are you suggesting setting updateSample with the response? Would that make the modalSample change to the source sample? Or does it just update the cache?

app/packages/core/src/components/Modal/Sidebar/Annotate/useFocus.ts
kaixi-wang: Would exiting edit mode go to list view? I think that's why I did this. 
Also quickdraw should not be active (cannot create new)

app/packages/core/src/components/Modal/Sidebar/Annotate/Annotate.tsx
kaixi-wang: I haven't had any issues testing locally, but I can take another pass. Iirc, I think I had this code around before the annotationController was fully implemented in teams

app/packages/annotation/src/util/labelPersistence.ts
tom-vx51: yeah we normally update with the response, want to make sure we're not missing data here. With the existing refresh, it would update the modal sample (and also the sample in the grid)

app/packages/annotation/src/util/labelPersistence.ts
tom-vx51: hm we should investigate. We should expect `200` for all of these operations

app/packages/core/src/client/annotationClient.ts
tom-vx51: Can leave it for now. We have some existing transformation logic for data coming from the server, but not to the sever. I'll see if it makes sense to consolidate into a more holistic interceptor chain

fiftyone/constants.py
```diff
@@ -14,7 +14,7 @@
 from importlib.metadata import metadata
 
 
-CLIENT_TYPE = "fiftyone"
+CLIENT_TYPE = "fiftyone-teams"
```
tom-vx51: Guessing this should remain `fiftyone`

app/packages/annotation/src/util/labelPersistence.ts
kaixi-wang: Delete is currently disabled from the UI so I didn't bother further debugging why the deleted generated sample would throw an error for now.

tom-vx51: fair enough, do you mind creating a ticket so we don't lose track?

kaixi-wang: created FOEPD-3260

app/packages/annotation/src/util/labelPersistence.ts
kaixi-wang: removed special handling and just letting refreshSample handle it. The extra unnecessary grid reloading is a problem that regular updates also trigger and can be handled separately.

app/packages/annotation/src/util/labelPersistence.test.ts
kaixi-wang: yeah intentional because buildJsonPath prefixes the result with `/`...

the others/existing are wrong in terms of mocking realistic values, although I guess you can say they are testing the logic correctly

tom-vx51 (approved):
>> Want to make sure we're correctly handling updates in useLighterDeltaSupplier - perhaps some additional branching/validation for these single-label deltas.

> Not sure what you're thinking, but I didn't want to overengineer a solution without another case

Just want to make sure we're not dropping edits. If you're happy with the behavior, I'll leave it at your discretion

PR #7041, opened 2026-02-19
[HA] label default to dropdown

erik-nieh:

defaults label with no classes to dropdown
fixes type issues in schema helper

Screenshot from 2026-02-19 09-29-28

Screenshot from 2026-02-19 09-29-13

Locally

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added support for numeric values alongside string values in schema configuration for radio buttons, tags, and select dropdowns.

* **Improvements**
  * Enhanced validation rules to permit empty class and value lists in label schema definitions.
  * Updated default UI component selection logic for label fields with empty or missing class configurations.


app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/schemaHelpers.ts
```diff
@@ -301,3 +305,16 @@ export function generatePrimitiveSchema(
...
+ * Currently labelSchema will default to text for a label that has no
+ * classes. This function remaps this to dropdown (which will have no values).
+ * @param component - The label component from the schema
+ * @returns The parsed component type
+ */
+export function parseLabelComponent(component?: ComponentType): ComponentType {
+  if (!component) return "dropdown";
+  if (component === "text") return "dropdown";
+  return component;
+}
```
lanzhenw: it should be handled before this step at the root.

lanzhenw (changes requested):
Thanks for looking into this! The fix works at the UI level, but I think we should address this at the root cause instead. User should not be allowed to define text input for labels (even in quick edit)

  The reason the label defaults to text when there are no classes is because of a change I introduced in #6927, where _create_label_field in plugins/operators/annotation.py was updated to fall back to text when classes is empty. 6927 was intended to be consistent with quick edit ux and the existing backend validation logic. 

  I think the right approach is:
  1. Revert the logic in annotation.py (from #6927) so that when there are no classes, the component defaults to dropdown instead of text:
```
  if classes and len(classes) <= foac.CHECKBOXES_OR_RADIO_THRESHOLD:
      component = foac.RADIO
  else:
      component = foac.DROPDOWN
```
  2. Update the backend validation — currently `_validate_values_setting` in `validate_label_schemas.py ` raises an error if classes is an empty list, which would block a dropdown component with no classes from passing validation. That check needs to be relaxed to allow empty classes for dropdown and radio components.
  3. Once the backend correctly produces dropdown for no-classes labels, the frontend parseLabelComponent workaround here wouldn't be needed. But the type improvements in this PR (exporting FieldType/ComponentType and widening values to string[] | number[]) are nice cleanups

lanzhenw (approved):
clean fix! thank you. 
[nit] You may have to update the backend test with the validation logic change. Or add a backend test case for it.

app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/schemaHelpers.ts
```diff
@@ -101,7 +102,7 @@ export const createSlider = (
 
 export const createRadio = (
   name: string,
-  choices: string[],
+  choices: string[] | number[],
```
kaixi-wang, line 105: Can you use Union instead for python 3.9 support

erik-nieh: @kaixi-wang is this being compiled by docs in python?

swheaton: this is ts not python kacey?

kaixi-wang: 🤦‍♀️ too early

PR #7047, opened 2026-02-19
[HA]: Address memory leaks in annotate mode

sashankaryal:
Fixes memory leaks caused by retained event handler that closes on Scene _and_ quickdraw handler closing on Scene

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved stale-state issues in annotation tools and quick-draw transitions for more reliable editing.
  * Improved media initialization so images and overlays appear consistently after load.

* **Refactor**
  * Simplified modal rendering behavior and public props, consolidating native vs. custom/annotate flows.
  * Centralized bounds-change handling and tightened cleanup for overlays and scene events.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useQuickDraw.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/Edit/useQuickDraw.ts
@@ -283,7 +283,9 @@ export const useQuickDraw = () => {
           currentScene.renderLoopActive
         ) {
           currentScene.exitInteractiveMode();
-          addOverlay(currentLabel.overlay as BaseOverlay);
+          if (currentLabel.overlay) {
+            addOverlay(currentLabel.overlay as BaseOverlay);
+          }
         }
 
         // Track last-used detection field and label for auto-assignment


app/packages/core/src/components/Modal/ModalLooker.tsx
```diff
@@ -48,8 +44,8 @@ const ModalLookerNoTimeline = React.memo((props: LookerProps) => {
       id={id}
       data-cy="modal-looker-container"
       style={{
-        width: props.ghost ? 0 : "100%",
-        height: props.ghost ? 0 : "100%",
+        width: "100%",
+        height: "100%",
```
sashankaryal, line 48: matches looker's rendering style if we place it at parent's level (visually, you'll see a small margin top-down which feels clean)

app/packages/lighter/src/react/useLighter.ts
```diff
@@ -19,6 +19,18 @@ export const useLighter = () => {
   const scene = useAtomValue(lighterSceneAtom);
   const isReady = !!scene;
   const registeredCallbacks = useRef<Set<() => void>>(new Set());
+
+  // useLighter() consumers may cache callback that closes over `scene` and that can result in memory leak.
```
tom-vx51, line 23: not opposed to this change, but shouldn't those callbacks be recreated on scene change and drop the ref they're holding?

PR #7053, opened 2026-02-20
fix: re-fetch frame labels on cache miss for long videos (#6347)

### LINKED ISSUE #6347: [BUG] Frame Labels (detections) not visisble on GUI on what seems to be a buffer limitation
### Describe the problem

Hi, so I recently upgraded from an old version of fiftyone 0.25 to the latest 1.8.0. 
I manage long videos in my dataset, of 20 minutes, 15FPS and several detections per frame.

The problem I'm experiencing now is the following:
- I open a video on fityone with several detections. (17.446 frames)
- After a period of time of buffer loading, that I have identified as 5:40 min for a video of 15FPS and 3:20min for 30FPS (half), the detections are not seen anymore. If Im on the mark 1:00 and the buffer loads more than 1min + 5:40min (15FPS) the detections before the 1:00min mark wont be visible anymore. 
- This can be easily tested when the video buffer is full loaded. I will only be able to see detections on the image that are between (end_video_time - 5:40 min, end_video_time)
- This also happens on videos with double the frames but half of the time, independently of the amount of "frame labels or detections" in that video.
- It definately seems like a buffer somewhere of around (5100 frames, 5:40min x 15FPS) in the code.
- In previous versions I didnt experience this problem. 

Probably highly related to this line of code that was not around in version 0.25
https://github.com/voxel51/fiftyone/blob/48ecd2a99622f9819e0a3a6f8f5215a59c3cba8a/app/packages/looker/src/constants.ts#L28

### Code to reproduce issue

Unfortunately I cannot record the screen due to the data being used but to make this reproductible a random video at 30FPS with more than 3:20min should be enough to appreciate the issue. 
Add a random detection on every frame and allow for the entire buffer to be loaded. See how the first part of the video is missing the detections overlay but the last 3:20 minutes will have the detections overlayed.

```
# commands and/or screenshots here
```

### System information

-   **OS Platform and Distribution** : Linux Ubuntu 22.04
-   **Python version** (`python --version`): 3.10.12
-   **FiftyOne version** (`fiftyone --version`): 1.8.0

### Other info/logs

As a workaround I was wondering if I could reinstall the fiftyone binaries application with a bigger buffer but I dont know how to build the web application. If you can provide some steps on how to do this it would be appreciated.

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community

HemantSudarshan:
**Summary:** 
Fixes an issue where frame labels (detections) permanently disappear on videos longer than ~5100 frames by detecting LRU cache misses and correctly restarting the frame stream.

**Motivation:** 
Resolves #6347. The `@fiftyone/looker` frame reader holds frames in an LRU cache limited to 5100 entries to protect browser memory. On long videos, when a user scrubs backward to a frame that was evicted from the cache, the frame stream mistakenly interpreted the active stream state as sufficient. It would trigger the forward-only `requestFrameChunk` path instead of restarting the stream, causing the labels to vanish permanently since the requested frame was never fetched.

**Root Cause:** 
The closure returned by `acquireReader()` in `frame-reader.ts` only checked `!nextRange` (stream completed) to decide whether to restart the stream. When `nextRange` was set (stream in progress) but the requested frame had been evicted, it continued requesting chunks forward, abandoning the requested missing frame.

**Fix:** 
Added a `frameCache.has(frameNumber)` check. When the requested frame is not in the LRU cache, the stream restarts precisely from that frame using the existing `setStream()` worker mechanism. 

**Changes:**
- `app/packages/looker/src/lookers/frame-reader.ts`: Added cache-miss detection (2-line fix).
- `app/packages/looker/src/lookers/frame-reader.test.ts`: Created comprehensive test file with 10 unit tests.

**Testing:**
- Added 10 unit tests covering initialization, streaming, cache-miss refetch, buffering state, and cache eviction.
- Follows the existing `vi.hoisted()` and `DummyWorker` pattern established in Looker testing.
- Existing E2E tests (`group-video-label.spec.ts`) continue to serve as regression guards.
- Manual verification recommended with a 6000+ frame video: `ffmpeg -f lavfi -i color=c=blue:s=320x240:d=400 -r 15 long_test.mp4`.

**Backward Compatibility:**
-   No public API changes
-  No function signature changes
-  `MAX_FRAME_STREAM_SIZE` unchanged (keeps browser memory safe)
-  No migration required

**Checklist:**
- [x] Code follows project style guide (2-space indent, K&R braces)
- [x] Self-review completed
- [x] Tests added and passing
- [x] No breaking changes
- [x] CI/CD compatible

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive test suite for the frame reader covering worker lifecycle, stream management, chunk handling, cache and eviction scenarios, buffering updates, restart behavior, and cleanup.

* **Performance**
  * Improved frame-streaming logic to detect cached frames and in-flight fetches, reducing unnecessary stream restarts and redundant frame requests.

### DIFF
--- a/app/packages/looker/src/lookers/frame-reader.test.ts
+++ b/app/packages/looker/src/lookers/frame-reader.test.ts
@@ -37,8 +37,10 @@ vi.mock("./shared", () => {
     };
 });
 
-vi.mock("@fiftyone/utilities", () => {
+vi.mock("@fiftyone/utilities", async (importOriginal) => {
+    const actual = await importOriginal<typeof import("@fiftyone/utilities")>();
     return {
+        ...actual,
         sizeBytesEstimate: vi.fn(() => 1),
     };
 });


app/packages/looker/src/lookers/frame-reader.test.ts
```diff
@@ -0,0 +1,354 @@
...
+        const options = createMockOptions({ frameCount: 200, frameNumber: 100 });
+        const requestFrames = acquireReader(options);
+
+        // Load frames 100-129. After this, nextRange advances to [130, ...].
+        simulateFrameChunk(
+            Array.from({ length: 30 }, (_, i) => ({ frame_number: i + 100 })),
+            [100, 129]
+        );
+
+        // Request frame 50 â it was never loaded and is behind the stream
```
benjaminpkane: "â"  text is unexpected. It is used a few times

benjaminpkane:
```suggestion
        // Request frame 50 as it was never loaded and is behind the stream
```

app/packages/looker/src/lookers/frame-reader.test.ts
```diff
@@ -0,0 +1,354 @@
...
+        const requestFrames = acquireReader(options);
+
+        // Simulate loading all frames (end >= frameCount sets nextRange to null)
+        simulateFrameChunk(
+            [{ frame_number: 1 }, { frame_number: 2 }],
+            [1, 2]
+        );
+
+        // Stream is now complete (nextRange === null)
+        // Request a frame â should restart the stream via setStream
```
benjaminpkane:
```suggestion
        // Requesting a frame should restart the stream via setStream
```

benjaminpkane (approved):
Looks good. Thanks 🤖

PR #7056, opened 2026-02-22
Normalize CVAT mask shapes based on expected label type

vittorio-prodomo:
## Summary

- When downloading CVAT annotations, mask (brush) shapes are now converted to polylines/polygons/segmentation when the target label field expects those types, instead of always routing to the detections bucket
- This mirrors the branching logic already used for `"polygon"` shapes in `_parse_annotation()` (lines 5967-5989), where `expected_label_type` is checked to decide whether a polygon should be routed to `"segmentation"`, `"detections"` (as an instance mask), or `"polylines"` (as a regular polygon/polyline). The new code applies the same pattern to `"mask"` shapes: it calls `cvat_shape.to_instance()` to decode the RLE mask into a `Detection`, then converts it via `Detection.to_polyline(tolerance=2)` when the expected label type calls for it
- Currently, mask shapes always land in the `"detections"` bucket regardless of what the user's label field expects. When the field is e.g. `"polygons"`, these labels become "unexpected" during `load_annotations()`, triggering an interactive prompt and splitting annotations across multiple fields. This splitting is not necessarily a problem on its own, but it becomes one during export — as far as I'm aware, FiftyOne doesn't support exporting multiple label fields at once (e.g., by merging them), so having annotations split across fields means you can't cleanly export everything in one go. This change routes them to the correct bucket based on the expected label type instead
- I understand that mask-to-polygon is a conversion more prone to data loss compared to the reverse operation, so I'm open to suggestions from the maintainers — for example, rather than converting silently, we could still prompt the user and give them the choice (either split annotations across multiple fields or attempt to merge them into the expected field)

## Motivation

I really appreciate the effort that has gone into integrating CVAT into FiftyOne workflows — it's been great to use. The reason I'm proposing this change is that I typically tell my CVAT annotators they can use polygons and the brush/mask tool interchangeably, and it would be nice if FiftyOne handled both the same way when loading annotations.

One concrete reason I encourage this is that I'm not a fan of how CVAT currently handles COCO export of mask shapes — it dumps them directly as RLE masks, which can cause problems downstream. Many well-known codebases used for format conversion or dataset ingestion completely ignore RLE masks because they still follow an old, unwritten convention in COCO annotations where RLE masks were essentially always `iscrowd=1` and vice versa. So in practice, allowing annotators to freely use the brush tool while still getting clean polygon output on the FiftyOne side would be very helpful.

## Test plan
- Existing CVAT tests should pass with no regressions (default detections/instances path is unchanged)
- Manual: annotate with the brush/mask tool in CVAT on a polygons-type field, confirm labels load into the correct single field without prompts

## Release Notes

* **Bug Fixes**
  * Improved handling of mask data during CVAT annotation parsing to correctly convert masks to polylines, polygons, and segmentation formats based on expected label types.

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -5923,6 +5923,7 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
         frame_id = frame_data.get("frame_id", None)
 
         label = None
+        label_type = None
 
         if anno_type in ("shapes", "track"):
             shape_type = anno["type"]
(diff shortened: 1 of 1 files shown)


brimoor:
@ehofesmann could you give this a review as CVAT codeowner? 😄

ehofesmann (approved):
Thanks for the contribution @vittorio-prodomo ! This has been a long needed extension of our polyline/mask loading logic and is working great!

I confirmed that the CVAT test suite still passes. Also the default polyline conversion tolerance is not a major concern for me. Other annotation platforms follow similar patterns when using mask tools for polyline annotation. If it is an issue, then users can always load the labels into instance masks in FiftyOne and convert them to polylines after the fact  with more flexibility.

PR #7058, opened 2026-02-23
CVAT integration: support >10 projects

### LINKED ISSUE #7024: [BUG] Fetching CVAT project name/ID occasionally fails if there are more than 10 CVAT projects
### Describe the problem

We leverage FiftyOne's CVAT integration for annotation work and we specify project IDs in our `.annotate()` calls. We now have more than 10 CVAT projects and have noticed that the annotation calls (annotate/load_annotations) sometimes fail. 

The issue appears to be that `CVATAnnotationAPI._parse_project_details()` tries to get the project name via `get_project_name()`, which in turns calls `_get_value_from_search()`. The latter tries to look up the project via CVAT's `GET api/projects`, but fails to find it if the looked-for project ID isn't among the first page results (first 10 results). 

There is a query parameter (`?id=X`) but it doesn't seem to actually filter the API results. Would it make sense to get the results via `_get_paginated_results()` instead?

### Code to reproduce issue

To reproduce the issue, there needs to be more than 10 projects in CVAT and the project targeted by `.annotate()` must not be among the first page results for CVAT's `GET api/projects`.

### System information

...

### Other info/logs

```
  File "XXX/fiftyone/core/collections.py", line 9259, in load_annotations
    return foua.load_annotations(
  File "XXX/fiftyone/utils/annotations.py", line 1038, in load_annotations
    annotations = results.backend.download_annotations(results)
  File "XXX/fiftyone/utils/cvat.py", line 3324, in download_annotations
    annotations = api.download_annotations(results)
  File "XXX/fiftyone/utils/cvat.py", line 4602, in download_annotations
    _, project_id = self._parse_project_details(
  File "XXX/fiftyone/utils/cvat.py", line 4952, in _parse_project_details
    raise ValueError("Project '%d' not found" % project_id)
ValueError: Project '15' not found
```

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community

sqeph:
Resolves #7024

## Release Notes

- Fixes a bug in the [CVAT integration](https://docs.voxel51.com/integrations/cvat.html) where annotations could fail to load when calling `load_annotations()` if there are >10 CVAT projects

## Description

In reference to #7024 , this PR extends `_get_value_from_search()` to support paginated results (by delegating to `_get_paginated_results()`).

Not familiar with how your automated CVAT tests are set up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CVAT integration to properly handle paginated API responses, ensuring all results are retrieved across multiple pages instead of only processing the first page.

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -4932,8 +4932,8 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
         self, search_url_fcn, target, target_key, value_key
     ):
         search_url = search_url_fcn(target)
-        resp = self.get(search_url).json()
-        for info in resp["results"]:
+        results = self._get_paginated_results(search_url)
+        for info in results:
             if info[target_key] == target:
                 return info[value_key]
 


fiftyone/utils/cvat.py
```diff
@@ -4932,8 +4932,8 @@ def _get_value_from_search(
         self, search_url_fcn, target, target_key, value_key
     ):
         search_url = search_url_fcn(target)
-        resp = self.get(search_url).json()
-        for info in resp["results"]:
+        results = self._get_paginated_results(search_url)
```
brimoor, line 4935: @ehofesmann I see the `self.get(...).json()` pattern elsewhere in the integration. Do any/all of those also suffer from the same page limit issue?

ehofesmann: In all of the other instances of this pattern, it is pulling a single resource like one task/project/job by id, so it wouldn't result in the same pagination issue that this PR addresses.

PR #7062, opened 2026-02-23
Refactor computing embeddings with dataloader for patches

kaixi-wang:

Refactor the patch embeddings data loader (TorchImagePatchesDataset) from a map-style Dataset to a streaming IterableDataset that reads detection metadata and resolves corresponding media in the dataloader worker, bypassing the need to materialize all bounding boxes and file paths into memory upfront. 

Retains a list-based path (_iter_list_detections) for callers that pass explicit image_paths/patches rather than a samples view

Benefits:
- memory efficient multi-worker Dataloaders
- enable batch_size > 1
-  Decode images from bytes (used by the prefetcher) instead of always reading from a file path to avoid additional file checks
- remove need to iterate over both samples and dataloader

Manual testing with compute_patch_embeddings with 'clip-vit-base32-torch` on detections field.
- [x] improved performance
- [x] output embeddings are the same

        notes for FiftyOne users.

-   [ ] Other

### DIFF
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -13,8 +13,6 @@ import multiprocessing
 import os
 import pickle
 import sys
-from typing import Any, Optional, List
-import warnings
 
 import cv2
 import numpy as np
@@ -27,11 +25,8 @@ import eta.core.utils as etau
 import fiftyone.core.config as foc
 import fiftyone.core.labels as fol
 import fiftyone.core.models as fom
-import fiftyone.core.media as fomd
 import fiftyone.core.odm as foo
 import fiftyone.core.utils as fou
-import fiftyone.utils.image as foui
-import fiftyone.core.collections as focol
 import fiftyone.core.view as fov
 
 fou.ensure_torch()
@@ -2379,6 +2374,11 @@ class TorchImagePatchesDataset(torch.utils.data.IterableDataset):
         num_workers (0): the number of DataLoader workers. When > 0 and using
             ``samples``/``patches_field``, the dataset is partitioned across
             workers via ``$bucketAuto``
+        read_fn (None): an optional callable that takes a file path and returns
+            bytes. Used for concurrent I/O prefetching. By default uses
+            :func:`fiftyone.core.storage.read_file`
+        prefetch_workers (8): the number of threads to use for concurrent
+            image I/O prefetching. Set to 0 to disable prefetching
     """
 
     def __init__(
@@ -2398,7 +2398,8 @@ class TorchImagePatchesDataset(torch.utils.data.IterableDataset):
         alpha=None,
         skip_failures=False,
         num_workers=0,
-        _prefetch_fn=None,
+        read_fn=None,
+        prefetch_workers=8,
     ):
         self.handle_missing = handle_missing
         self.transform = transform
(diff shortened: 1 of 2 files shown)


fiftyone/utils/torch.py
```diff
@@ -2392,189 +2397,298 @@ def __init__(
...
-                bboxes = [[b] if b is not None else None for b in bboxes]
-        else:
-            raise ValueError(
-                "Either `patches` or `patches_field` must be provided"
-            )
 
-        num_patches = 0
-        patch_edges = [0]
-        patches = []
+def _prepare_db_detection_config(samples, patches_field, num_workers):
```
brimoor, line 2544: This is largely an alternative implementation of the logic that exists here today: https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/map/batcher/id_batch.py

We need to build on top of the existing patterns we have rather than reimplementing them for bespoke use cases.

To be clear, `$bucketAuto` is potentially interesting as an alternative to `values("id", _enforce_natural_order=False)`. But if it is better, it should be considered for use everywhere, not just here.

fiftyone/utils/torch.py
```diff
@@ -2307,16 +2302,18 @@ def _parse_inputs(
         return image_paths, sample_ids, targets, str_targets
 
 
-class TorchImagePatchesDataset(Dataset):
-    """A :class:`torch:torch.utils.data.Dataset` of image patch tensors
+class TorchImagePatchesDataset(torch.utils.data.IterableDataset):
```
brimoor, line 2305: Switching to `torch.utils.data.IterableDataset` seems reasonable based purely on the stated design of  that Torch class. But I would definitely want the ML team to review this part to see what they think is the best way to split datasets up for processing by multiple data loaders, which is what this PR is optimizing in one specific case.

cc @jacobsela as codeowner of `to_torch()`.

fiftyone/core/models.py
```diff
@@ -1803,64 +1808,94 @@ def _embed_patches_data_loader(
...
 
-                if patches is not None:
-                    embeddings = []
-                    for patches_batch in fou.iter_slices(patches, batch_size):
-                        embeddings_batch = model.embed_all(patches_batch)
-                        embeddings.append(embeddings_batch)
+    with contextlib.ExitStack() as exit_stack:
+        bulk_ctx = None
+        if save:
+            bulk_ctx = exit_stack.enter_context(foc.BulkWriteContext(coll))
```
brimoor, line 1838: The design goal of `SaveContext` was to provide a simple syntax for expressing a series of edits to samples in a `SampleCollection` in such a way that the underlying mongo operations can be issued as bulk writes.

If there are ways to make that more efficient, I really want to see us make those changes inside `SaveContext` rather than building a bespoke alternative to that pattern for this one use case!

brimoor: cc @exupero

fiftyone/utils/torch.py
kaixi-wang: Yes I initially wanted to incorporate it into the existing batcher, but it was turning out to be a larger refactor than anticipated because the batcher requires Samples currently, so I ended up isolating it to here. 

However, this method of batching is currently used in hub for splitting distributed DOs because of performance, and I do want to make it an option for map_sample.

PR #7066, opened 2026-02-23
Add tsc to CI and yarn scripts

marcus-kestler:
Uses a top level tsconfig.json and a tsconfig.spec.json for source and test respectively.  Source uses the recommended configs for vite-react and node-lts as it's basis.  The test config is more flexible.
I removed all of the nested tsconfigs.  This should be kept as a best practice to standardize the rules over all packages.  This can obviously be overridden for special cases if needed, and use the root as the base.

Unfortunately the "export" section of the tsconfig does not merge and must specify the package exclude list in both source and test builds.  (I looked all over for a solution to this, but if someone knows a way, please let me know). 

This also fixes all tsc errors in the command package (I needed something to compile).

Also note, I removed all the different version of typescript from the packages and added it to the root to keep this the same as well.

Verified the compiler is running on the correct packages, tests work, etc.
Pushed a non-compiling commit to verify CI blocks correctly and then corrected it verifying it works after.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Chores**
  * Enhanced build pipeline with explicit compilation steps for application source code and test suites to improve build transparency.
  * Refactored build configuration to separate Vite and Vitest concerns for better maintainability.
  * Improved internal error handling for context management and key sequence matching.

### DIFF
--- a/app/vite.config.ts
+++ b/app/vite.config.ts
@@ -1,10 +1,10 @@
-import { defineConfig } from "vite";
+import { defineConfig as defineViteConfig, mergeConfig } from "vite";
+import { defineConfig as defineVitestConfig } from "vitest/config";
 import relay from "vite-plugin-relay";
-import "vitest/config";
 
 const { DISABLE_COVERAGE } = process.env;
 
-export default defineConfig({
+const vitestConfig = defineVitestConfig({
   test: {
     environment: "jsdom",
     setupFiles: ["./vitest.setup.ts"],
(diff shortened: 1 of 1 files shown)


mnk-nasir:
Great to see this refinement in the v1.13.x pipeline. I've been following the recent HA (Human Annotation) updates, and this looks like a solid step toward stabilizing the new Quick Edit/Schema Manager workflows. Have we verified if this impacts any of the existing plugin hooks, or is it strictly an internal App fix?

afoley587 (approved):
Aloha changes seem good to me!

tom-vx51 (approved):

PR #7069, opened 2026-02-23
Fix data_map path-matching bugs in import_annotations()

### LINKED ISSUE #7057: [BUG] CVAT import_annotations() fails to match CVAT filenames containing subdirectory paths
### Motivation
  It would be hugely beneficial to have to employ the least number of work-arounds possible when trying to simply and quickly get a CVAT-labeled dataset directly into FiftyOne, configuring this as an easy way to:
  
  - Facilitate a "first import" of a CVAT dataset into FiftyOne without intermediate steps (e.g., without doing CVAT -> export to disk -> import into Fiftyone)
  - Having the possibility to periodically keep the FiftyOne dataset "mapped" with the latest additions/modifications to the labels done in CVAT (this is different from sporadically sending quick annotations job from FiftyOne to CVAT, which creates new tasks and has different use cases).

  ### Describe the problem

  [`fiftyone.utils.cvat.import_annotations()`](https://github.com/voxel51/fiftyone/blob/fff48b8f63e4799253ec5261a43b361fb43d349d/fiftyone/utils/cvat.py#L49) silently fails to import any annotations when CVAT stores filenames with subdirectory structure (e.g. `coco8-seg/images/train/000000000009.jpg` rather than just
  `000000000009.jpg`). This is common when data is uploaded to CVAT as a directory tree or zip archive.

  The root cause is that the `data_map` dictionary — which maps CVAT filenames to local filepaths — is built using `os.path.basename()`, stripping all directory components. But CVAT's task metadata API returns
  the full relative path as the frame name. The lookup `data_map.get(filename, None)` then fails for every frame, and all annotations are silently ignored.

  There are **three related bugs** in the `data_map` construction, plus one downstream consequence:

  ---

  #### Bug 1 — `data_path=None`: basenames of existing samples don't match CVAT filenames

  https://github.com/voxel51/fiftyone/blob/f488054dabf83489ad898241c0990e93ef1ab67c/fiftyone/utils/cvat.py#L171

  `os.path.basename(f)` produces `"000000000009.jpg"`, but CVAT returns `"coco8-seg/images/train/000000000009.jpg"`. No match.

  #### Bug 2 — `data_path` is a directory: same basename issue

  https://github.com/voxel51/fiftyone/blob/f488054dabf83489ad898241c0990e93ef1ab67c/fiftyone/utils/cvat.py#L175-L181

  Same problem — files are listed recursively but keyed by basename only.

  #### Bug 3 — `download_media=True` with relative `data_path`: path mismatch after round-trip

  https://github.com/voxel51/fiftyone/blob/f488054dabf83489ad898241c0990e93ef1ab67c/fiftyone/utils/cvat.py#L311-L312

  When `download_media=True`, the code bypasses `data_map` and constructs the filepath directly via `os.path.join(data_dir, filename)`. Media downloads correctly and samples are created. However, if `data_path`
   is a relative path (e.g. `"my_downloads/"`), the samples are stored with the **relative** filepath. FiftyOne then converts it to an **absolute** path internally when persisting to the database.

  Later, `_build_sparse_frame_id_map` reads back the absolute paths from the database and tries to look them up in `cvat_id_map`, which still has relative-path keys. Every lookup returns `None`, so
  `frame_id_map` ends up empty and all annotations are silently skipped.

  ---

  **In all three cases** the result is the same: a warning about ignored files followed by no annotations being imported, with no error raised.

  ### Code to reproduce issue

```python
  python
  import fiftyone as fo

  # Assumes a CVAT task where files have subdirectory structure
  # e.g. frame names like "subdir/image.jpg" rather than just "image.jpg"

  # Bug 1: data_path=None
  ds = fo.Dataset()
  ds.add_sample(fo.Sample(filepath="/path/to/subdir/image.jpg"))
  fouc.import_annotations(ds, task_ids=[TASK_ID])
  # Warning: Ignoring annotations for N files in CVAT (eg subdir/image.jpg)
  # that do not appear in the provided data map

  # Bug 2: data_path is a directory
  ds2 = fo.Dataset()
  fouc.import_annotations(ds2, task_ids=[TASK_ID],
                          data_path="/path/to/data/root/",
                          insert_new=True)
  # Same warning — no annotations imported

  # Bug 3: download_media=True with relative data_path
  ds3 = fo.Dataset()
  fouc.import_annotations(ds3, task_ids=[TASK_ID],
                          data_path="relative_dir",
                          download_media=True, insert_new=True)
  # Media downloads, samples are created, but labels are empty

  # Workaround for Bugs 1 & 2: build a dict manually with relative-path keys
  import os
  data_dir = "/path/to/data/root"
  data_map = {
      os.path.relpath(os.path.join(root, f), data_dir): os.path.join(root, f)
      for root, _, files in os.walk(data_dir)
      for f in files
  }
  fouc.import_annotations(ds, task_ids=[TASK_ID], data_path=data_map)
  # Works correctly

  # Workaround for Bug 3: use an absolute data_path
  fouc.import_annotations(ds3, task_ids=[TASK_ID],
                          data_path=os.path.abspath("relative_dir"),
                          download_media=True, insert_new=True)
```

  System information

  - OS Platform and Distribution: Linux Ubuntu (kernel 6.8.0-94-generic)
  - Python version: 3.10.12
  - FiftyOne version: 1.14.0
  - FiftyOne installed from: source (develop branch at f488054d)

  Other info/logs

  Suggested fixes:

  - Bugs 1 & 2: Use os.path.relpath(f, data_path) instead of os.path.basename(f) when building data_map from a directory. For the data_path=None case, a fallback chain could try the full basename first, then
  progressively longer relative suffixes — though the right approach may warrant discussion.
  - Bug 3: Convert data_dir to an absolute path before constructing filepaths (e.g. data_dir = os.path.abspath(data_path)), so that the paths stored in cvat_id_map and in the database are consistent.

  Willingness to contribute

  - [x] Yes. I can contribute a fix for this bug independently
  - [ ] Yes. I would be willing to contribute a fix for this bug with guidance
  from the FiftyOne community
  - [ ] No. I cannot contribute a bug fix at this time

  ---

vittorio-prodomo:
The bug is in the `data_map` construction block of `import_annotations()`:

https://github.com/voxel51/fiftyone/blob/306772a7832af440517c0a147627117b11f01bce/fiftyone/utils/cvat.py#L170-L185

## Summary

Fixes three related bugs where CVAT filenames with subdirectory structure (e.g. `images/train/img.jpg`) silently fail to match local filepaths, causing all annotations to be dropped.

- **Bug 1** (`data_path=None`): Replace the `{basename: filepath}` dict with a `_BasenameLookup` class that indexes by basename and disambiguates collisions via suffix matching against the CVAT relative path. O(N) to build, O(1) per lookup for unique basenames, O(k) tiebreak for k collisions.
- **Bug 2** (`data_path` is a directory): Use `os.path.relpath()` instead of `os.path.basename()` to preserve subdirectory keys in the data map.
- **Bug 3** (relative `data_path` + `download_media=True`): Convert `data_dir` to an absolute path so filepaths persisted to the database are consistent with keys in `cvat_id_map`.

Closes #7057

## Test plan

- [ ] Verify `import_annotations()` with a CVAT task where frame names include subdirectory paths (e.g. `subdir/image.jpg`)
- [ ] Verify that duplicate basenames across different subdirectories resolve correctly when CVAT provides the relative path
- [ ] Verify `download_media=True` with a relative `data_path` correctly imports annotations after download
- [ ] Run `python -m pytest tests/unittests/import_export_tests.py -k cvat -x -q`

## Summary by CodeRabbit

* **New Features**
  * Faster CVAT annotation import lookups for large datasets when basenames are unique.

* **Bug Fixes**
  * Improved disambiguation for files with identical basenames across subdirectories to reduce incorrect matches.
  * More consistent handling of data-path directories and absolute path resolution during imports.

* **Tests**
  * Added comprehensive unit tests covering basename resolution, ambiguity handling, path normalization, and edge cases.

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -300,15 +300,24 @@ class _BasenameLookup(dict):
 
     Subclasses :class:`dict` so that downstream code expecting a dict-like
     ``data_map`` (e.g. ``data_map.get(filename)``) continues to work.
+    All standard dict methods (``in``, ``len``, ``items``, iteration, etc.)
+    route through the same resolution logic to avoid behavioral
+    inconsistencies.
     """
 
     def __init__(self, filepaths):
+        super().__init__()
         self._index = defaultdict(list)
-        for f in filepaths:
+        self._filepaths = list(filepaths)
+        for f in self._filepaths:
             self._index[os.path.basename(f)].append(f)
-        super().__init__(self._index)
 
-    def get(self, filename, default=None):
+    def _resolve(self, filename):
+        """Resolve a CVAT filename to a local filepath.
+
+        Returns the filepath if unambiguously resolved, or :data:`_MISSING`
+        otherwise.
+        """
         basename = filename.replace(os.sep, "/").rsplit("/", 1)[-1]
         if candidates := self._index.get(basename):
             if len(candidates) == 1:
@@ -337,15 +346,38 @@ class _BasenameLookup(dict):
                         len(matches),
                     )
 
-            return default
+        return _MISSING
 
-        return default
+    def get(self, filename, default=None):
+        result = self._resolve(filename)
+        return default if result is _MISSING else result
 
     def __getitem__(self, filename):
-        result = self.get(filename, _MISSING)
+        result = self._resolve(filename)
         if result is _MISSING:
             raise KeyError(filename)
         return result
+
+    def __contains__(self, filename):
+        return self._resolve(filename) is not _MISSING
+
+    def __len__(self):
+        return len(self._filepaths)
+
+    def __iter__(self):
+        return iter(self._filepaths)
+
+    def __bool__(self):
+        return len(self._filepaths) > 0
+
+    def keys(self):
+        return {f: None for f in self._filepaths}.keys()
+
+    def values(self):
+        return {f: f for f in self._filepaths}.values()
+
+    def items(self):
+        return {f: f for f in self._filepaths}.items()
 
 
 def _parse_task_metadata(
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -308,7 +308,8 @@ class _BasenameLookup(dict):
     def __init__(self, filepaths):
         super().__init__()
         self._index = defaultdict(list)
-        self._filepaths = list(filepaths)
+        self._filepaths = tuple(dict.fromkeys(filepaths))
+        self._filepath_set = set(self._filepaths)
         for f in self._filepaths:
             self._index[os.path.basename(f)].append(f)
 
@@ -318,6 +319,10 @@ class _BasenameLookup(dict):
         Returns the filepath if unambiguously resolved, or :data:`_MISSING`
         otherwise.
         """
+        # Fast path: exact filepath match (e.g. from iterating this object)
+        if filename in self._filepath_set:
+            return filename
+
         basename = filename.replace(os.sep, "/").rsplit("/", 1)[-1]
         if candidates := self._index.get(basename):
             if len(candidates) == 1:
@@ -371,7 +376,7 @@ class _BasenameLookup(dict):
         return len(self._filepaths) > 0
 
     def keys(self):
-        return {f: None for f in self._filepaths}.keys()
+        return dict.fromkeys(self._filepaths).keys()
 
     def values(self):
         return {f: f for f in self._filepaths}.values()
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -174,7 +174,7 @@ def import_annotations(
     elif etau.is_str(data_path):
         if os.path.isdir(data_path):
             data_map = {
-                os.path.relpath(f, data_path): f
+                os.path.relpath(f, data_path).replace(os.sep, "/"): f
                 for f in etau.list_files(
                     data_path, abs_paths=True, recursive=True
                 )
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -168,13 +168,13 @@ def import_annotations(
     data_dir = None
     existing_filepaths = sample_collection.values("filepath")
     if data_path is None:
-        data_map = {os.path.basename(f): f for f in existing_filepaths}
+        data_map = _BasenameLookup(existing_filepaths)
     elif etau.is_str(data_path) and data_path.endswith(".json"):
         data_map = etas.read_json(data_path)
     elif etau.is_str(data_path):
         if os.path.isdir(data_path):
             data_map = {
-                os.path.basename(f): f
+                os.path.relpath(f, data_path): f
                 for f in etau.list_files(
                     data_path, abs_paths=True, recursive=True
                 )
@@ -182,7 +182,7 @@ def import_annotations(
         else:
             data_map = {}
 
-        data_dir = data_path
+        data_dir = os.path.abspath(data_path)
     else:
         data_map = data_path
 
@@ -286,6 +286,41 @@ def import_annotations(
     finally:
         anno_backend.delete_run(dataset, anno_key)
         api.close()
+
+
+class _BasenameLookup:
+    """Maps CVAT filenames to local filepaths via a basename index.
+
+    Builds a ``{basename: [filepaths]}`` index once, then resolves CVAT
+    filenames in O(1) when basenames are unique or O(k) when *k* paths
+    share the same basename (disambiguated by suffix matching).
+    """
+
+    def __init__(self, filepaths):
+        self._index = defaultdict(list)
+        for f in filepaths:
+            self._index[os.path.basename(f)].append(f)
+
+    def get(self, filename, default=None):
+        basename = filename.replace(os.sep, "/").split("/")[-1]
+        candidates = self._index.get(basename)
+        if not candidates:
+            return default
+
+        if len(candidates) == 1:
+            return candidates[0]
+
+        # Multiple files share the same basename — use the CVAT relative
+        # path to disambiguate, but only if it contains directory components.
+        # A bare basename (no "/") matches all candidates and is ambiguous.
+        normalized = filename.replace(os.sep, "/")
+        if "/" in normalized:
+            suffix = "/" + normalized
+            for f in candidates:
+                if f.replace(os.sep, "/").endswith(suffix):
+                    return f
+
+        return default
 
 
 def _parse_task_metadata(
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -288,39 +288,64 @@ def import_annotations(
         api.close()
 
 
-class _BasenameLookup:
+_MISSING = object()  # sentinel for _BasenameLookup.__getitem__
+
+
+class _BasenameLookup(dict):
     """Maps CVAT filenames to local filepaths via a basename index.
 
     Builds a ``{basename: [filepaths]}`` index once, then resolves CVAT
     filenames in O(1) when basenames are unique or O(k) when *k* paths
     share the same basename (disambiguated by suffix matching).
+
+    Subclasses :class:`dict` so that downstream code expecting a dict-like
+    ``data_map`` (e.g. ``data_map.get(filename)``) continues to work.
     """
 
     def __init__(self, filepaths):
         self._index = defaultdict(list)
         for f in filepaths:
             self._index[os.path.basename(f)].append(f)
+        super().__init__(self._index)
 
     def get(self, filename, default=None):
         basename = filename.replace(os.sep, "/").rsplit("/", 1)[-1]
-        candidates = self._index.get(basename)
-        if not candidates:
+        if candidates := self._index.get(basename):
+            if len(candidates) == 1:
+                return candidates[0]
+
+            # Multiple files share the same basename — use the CVAT
+            # relative path to disambiguate, but only if it contains
+            # directory components.  A bare basename (no "/") matches
+            # all candidates and is ambiguous.
+            normalized = filename.replace(os.sep, "/")
+            if "/" in normalized:
+                suffix = "/" + normalized
+                matches = [
+                    f
+                    for f in candidates
+                    if f.replace(os.sep, "/").endswith(suffix)
+                ]
+                if len(matches) == 1:
+                    return matches[0]
+
+                if len(matches) > 1:
+                    logger.warning(
+                        "Ambiguous CVAT filename '%s' matched %d local "
+                        "paths; skipping",
+                        filename,
+                        len(matches),
+                    )
+
             return default
 
-        if len(candidates) == 1:
-            return candidates[0]
-
-        # Multiple files share the same basename — use the CVAT relative
-        # path to disambiguate, but only if it contains directory components.
-        # A bare basename (no "/") matches all candidates and is ambiguous.
-        normalized = filename.replace(os.sep, "/")
-        if "/" in normalized:
-            suffix = "/" + normalized
-            for f in candidates:
-                if f.replace(os.sep, "/").endswith(suffix):
-                    return f
-
         return default
+
+    def __getitem__(self, filename):
+        result = self.get(filename, _MISSING)
+        if result is _MISSING:
+            raise KeyError(filename)
+        return result
 
 
 def _parse_task_metadata(
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -302,7 +302,7 @@ class _BasenameLookup:
             self._index[os.path.basename(f)].append(f)
 
     def get(self, filename, default=None):
-        basename = filename.replace(os.sep, "/").split("/")[-1]
+        basename = filename.replace(os.sep, "/").rsplit("/", 1)[-1]
         candidates = self._index.get(basename)
         if not candidates:
             return default
--- a/tests/unittests/cvat_utils_tests.py
+++ b/tests/unittests/cvat_utils_tests.py
@@ -9,7 +9,7 @@ Unit tests for ``fiftyone.utils.cvat`` helpers.
 import os
 import unittest
 
-import fiftyone as fo
+import fiftyone as fo  # noqa: F401  bootstrap modules to avoid circular imports
 from fiftyone.utils.cvat import _BasenameLookup
 
 
@@ -181,6 +181,35 @@ class TestBasenameLookup(unittest.TestCase):
         lookup = _BasenameLookup([path])
         self.assertEqual(lookup.get("d/e/f/img.jpg"), path)
         self.assertEqual(lookup.get("img.jpg"), path)
+
+    def test_iterated_keys_round_trip(self):
+        """Iterated keys must resolve back through get/contains/getitem."""
+        paths = [
+            "/data/images/train/cat.jpg",
+            "/data/images/val/dog.jpg",
+        ]
+        lookup = _BasenameLookup(paths)
+        for key in lookup:
+            self.assertIn(key, lookup)
+            self.assertEqual(lookup[key], key)
+            self.assertEqual(lookup.get(key), key)
+
+    def test_iterated_keys_round_trip_ambiguous_basenames(self):
+        """Round-trip must work even when basenames collide."""
+        paths = [
+            "/data/images/train/cat.jpg",
+            "/data/images/val/cat.jpg",
+        ]
+        lookup = _BasenameLookup(paths)
+        for key in lookup:
+            self.assertIn(key, lookup)
+            self.assertEqual(lookup[key], key)
+
+    def test_duplicate_filepaths_deduplicated(self):
+        paths = ["/data/a.jpg", "/data/b.jpg", "/data/a.jpg"]
+        lookup = _BasenameLookup(paths)
+        self.assertEqual(len(lookup), 2)
+        self.assertEqual(list(lookup), ["/data/a.jpg", "/data/b.jpg"])
 
 
 if __name__ == "__main__":
--- a/tests/unittests/cvat_utils_tests.py
+++ b/tests/unittests/cvat_utils_tests.py
@@ -1,0 +1,187 @@
+"""
+Unit tests for ``fiftyone.utils.cvat`` helpers.
+
+| Copyright 2017-2026, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
+import os
+import unittest
+
+import fiftyone as fo
+from fiftyone.utils.cvat import _BasenameLookup
+
+
+class TestBasenameLookup(unittest.TestCase):
+    """Tests for :class:`_BasenameLookup`."""
+
+    # ------------------------------------------------------------------
+    # Construction
+    # ------------------------------------------------------------------
+
+    def test_empty(self):
+        lookup = _BasenameLookup([])
+        self.assertEqual(len(lookup), 0)
+        self.assertFalse(lookup)
+
+    def test_single_file(self):
+        lookup = _BasenameLookup(["/data/images/cat.jpg"])
+        self.assertEqual(len(lookup), 1)
+        self.assertTrue(lookup)
+
+    # ------------------------------------------------------------------
+    # .get() — unique basenames
+    # ------------------------------------------------------------------
+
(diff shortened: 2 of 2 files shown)


fiftyone/utils/cvat.py
```diff
@@ -288,6 +288,41 @@ def import_annotations(
...
+    share the same basename (disambiguated by suffix matching).
+    """
+
+    def __init__(self, filepaths):
+        self._index = defaultdict(list)
+        for f in filepaths:
+            self._index[os.path.basename(f)].append(f)
+
+    def get(self, filename, default=None):
+        basename = filename.replace(os.sep, "/").split("/")[-1]
```
Burhan-Q: If `filename` was "path/to/file/" wouldn't this not return an empty string? Maybe strip the `filename` of "/" first?

Additionally I think for very long paths, if you use, `filename.replace(os.sep, "/").rsplit("/", 1)[-1]`. Check with:

```python
import timeit

p = "path/" * 60
timeit.repeat(lambda: p.split("/")[-1])

timeit.repeate(lambda: p.rsplit("/", 1)[-1])
```

With 60 splits, using `rsplit("/", 1)` was 10 times faster when I tested. Even with a smaller 10 splits, the speed up using `rsplit` was 2.1 times faster.

fiftyone/utils/cvat.py
```diff
@@ -288,6 +288,41 @@ def import_annotations(
...
+        candidates = self._index.get(basename)
+        if not candidates:
+            return default
+
+        if len(candidates) == 1:
+            return candidates[0]
```
Burhan-Q:
```suggestion
        return (
            candidates[0]
            if len(candidates := self._index.get(basename) or [])
            else default
        )
```

Burhan-Q: I'm a fan of using the walrus operator to help condense code, although here it might not be _as_ nice to use.

Burhan-Q (commented):
Just throwing on a few ideas.

fiftyone/utils/cvat.py
vittorio-prodomo: Good point on `rsplit("/", 1)[-1]` — it expresses the intent better ("give me the last segment") and avoids the unnecessary split of the full path. Will adopt that.

Regarding the trailing slash concern though, I don't think we need to guard against that. The `frame["name"]` value comes directly from CVAT's `Image.path` database field, which is set server-side via `os.path.relpath()`:

https://github.com/cvat-ai/cvat/blob/9470130de19156280ad614191a427a6a535e6c6c/cvat/apps/engine/task.py#L1650-L1658

...and then served as-is through the data meta API:

https://github.com/cvat-ai/cvat/blob/9470130de19156280ad614191a427a6a535e6c6c/cvat/apps/engine/views.py#L1485-L1490

Since it's always a relative file path produced by `os.path.relpath` on the server, it won't have trailing slashes.

fiftyone/utils/cvat.py
vittorio-prodomo: The suggested rewrite would only handle the zero-or-one-candidate case, but the method continues below with multi-candidate disambiguation — when multiple files share the same basename, it uses the CVAT relative path to pick the right one via suffix matching. The walrus version would short-circuit past that logic, silently returning the first candidate even when there are collisions.

Burhan-Q: Yup, I read that section _after_ posting my comments and just didn't connect the dots 🤦

Burhan-Q:
```python
        if candidates := self._index.get(basename):
            if len(candidates) == 1:
                return candidates[0]
            # Multiple files share the same basename
            normalized = filename.replace(os.sep, "/")
            if "/" in normalized:
                suffix = "/" + normalized
                matches = [
                    f for f in candidates if f.replace(os.sep, "/").endswith(suffix)
                ]
                if len(matches) == 1:
                    return matches[0]
            return default  # ambiguous case, basename matches multiple values
        return default  # for all false-y values of candidates
```

Perhaps something like this? Also, I noticed that this block:
```python
                    for f in candidates:
                        if f.replace(os.sep, "/").endswith(suffix):
                            return f
```
will exit on the first match. This _could be_ an issue for a (likely rare) case like:
```
/data/project/images/train/cat.jpg
/data/archive/images/train/cat.jpg
```
since there are two files with `images/train/cat.jpg` as a basename. As originally designed, it would return the first match, but it's no obvious to me if that's the correct decision or not. I added a `matches` to handle a case where there could be more than one of the same basename file, and return the default value instead.

My instinct would be to raise an exception in ambiguous cases instead, but it's not clear to me if that's the correct pattern to use here. I think at the very least it would make sense to log a warning, otherwise there could be a silent issue with multiple file matches.

Finally, I'll also note that this design mutates the `data_map` on this branch to a different type, as it will be a `dict` on all other conditional branches. This breaks the downstream contract of what the expected object will be. Presently this isn't an issue, but it could be addressed using:
```python
_MISSING = object()  # sentinel value

class _BasenameLookup(dict):
    """Maps CVAT filenames to local filepaths via a basename index.

    Builds a ``{basename: [filepaths]}`` index once, then resolves CVAT
    filenames in O(1) when basenames are unique or O(k) when *k* paths
    share the same basename (disambiguated by suffix matching).
    """

    def __init__(self, filepaths):
        self._index = defaultdict(list)
        for f in filepaths:
            self._index[os.path.basename(f)].append(f)
        super().__init__(self._index)

    def get(self, filename, default=None):
        basename = filename.replace(os.sep, "/").rsplit("/", 1)[-1]
        if candidates := self._index.get(basename):
            if len(candidates) == 1:
                return candidates[0]
            # Multiple files share the same basename
            normalized = filename.replace(os.sep, "/")
            if "/" in normalized:
                suffix = "/" + normalized
                matches = [
                    f for f in candidates if f.replace(os.sep, "/").endswith(suffix)
                ]
                if len(matches) == 1:
                    return matches[0]
            return default  # ambiguous case, basename matches multiple values
        return default  # for all false-y values of candidates

    def __getitem__(self, filename):
        result = self.get(filename, _MISSING)  # for when None is legitimately returned
        if result is _MISSING:
            raise KeyError(filename)
        return result
```
It only adds a bit more overhead, but keeps the downstream contract of a dictionary-like object, and keeps the indexing and `.get()` method output consistent.

vittorio-prodomo: Great catches — applied all three in 936ed45:

1. **Walrus guard** — `get()` now uses `if candidates := ...` at the top.
2. **Ambiguous multi-match** — suffix matches are collected into a list; if more than one match we return `default` and log a `logger.warning()` so it's not silent.
3. **Subclass `dict`** — `_BasenameLookup(dict)` with `super().__init__(self._index)` and a sentinel-based `__getitem__`, so the downstream dict contract is preserved.

fiftyone/utils/cvat.py
```diff
@@ -288,6 +288,66 @@ def import_annotations(
...
+
+    Subclasses :class:`dict` so that downstream code expecting a dict-like
+    ``data_map`` (e.g. ``data_map.get(filename)``) continues to work.
+    """
+
+    def __init__(self, filepaths):
+        self._index = defaultdict(list)
+        for f in filepaths:
+            self._index[os.path.basename(f)].append(f)
+        super().__init__(self._index)
```
Burhan-Q:
```suggestion
        super().__init__()
```

Burhan-Q: Looking back I realize I added this, but it's not needed. It will mean that other `dict` methods won't access the data at all (before the data was copied and accessible, but would return data differently). Removing it means that `"x" in data_map` won't  work, nor would `data_map.items()`, but this removes data duplication.

fiftyone/utils/cvat.py
```diff
@@ -288,6 +288,66 @@ def import_annotations(
...
+
+    def get(self, filename, default=None):
```
Burhan-Q:
```suggestion
    # NOTE: no other dict methods accessed, avoid method overrides until needed
    def get(self, filename, default=None):
```

Burhan-Q: Should help document the decision here about why other methods don't return data

Burhan-Q (commented):
A couple additional comments. In addition, even though there isn't an existing suite of tests for anything in the `fou.cvat` module, I think that including minimal unit tests for the new `_BasenameLookup` class would be a good place to start.

fiftyone/utils/cvat.py
vittorio-prodomo: But is this really worth it, for the goal of avoiding data duplication? I admit I'm new to all this, so this is just my opinion, but I thought the whole point of sublcassing dict was to preserve the downstream contract. Correct me if I'm wrong, but an empty underlying dict means that `in`, `len()`, `items()`, iteration, and truthiness checks would all likely misbehave silently if any downstream code ever touches them.

I don't know the entire codebase enough to back up this claim, but it seems to me the duplication here would be shallow, since it's just references to the same list objects, not copies of the filepath strings. So maybe the overhead is negligible in practice.

Burhan-Q: Yes and no. The data duplication might be a minor concern for smaller datasets, but will become much more of a concern for larger datasets. It's not necessarily inconsequential for everyone, and we have to be mindful of all use cases.

Not only that, but the behavior of the other `dict` methods aren't the same as the logic used for the `.get()` or `.__getitem__()` methods. This means there will be unexpected behavior:

```python
for k,v in data_map.items():
    ...  # here v would be a list, not a str
```

I was trying to follow the YAGNI principle that until there's a need, we won't include overrides for the other `dict` methods. It would keep the PR slimmer, but it also means the class doesn't _fully_ respect the `dict` object contract.

Given the above, even though there's no use of the other `dict` methods, maybe it just makes more sense to ensure that the behavior acts the same 100% by overriding the other dunder methods. I believe the additional code would be minimal and it would remove the need for data duplication.

vittorio-prodomo: agreed, overriding the dunder methods is the way to go. No data duplication, no behavioral surprises, no matter how downstream code touches the object.

pushed `aeacdf06` which:
- uses `super().__init__()` with no args (empty underlying dict)
- extracts a private `_resolve()` so `get()`, `__getitem__`, and `__contains__` all go through the same resolution logic
- adds `__contains__`, `__len__`, `__iter__`, `__bool__`, `keys()`, `values()`, `items()`
- also adds `tests/unittests/cvat_utils_tests.py` with 28 tests covering resolution, disambiguation, ambiguous cases, dict protocol compliance, and edge cases (as you suggested)

fiftyone/utils/cvat.py
vittorio-prodomo: both addressed in `aeacdf06`:
- `super().__init__()` as you suggested, but with full dunder overrides now so the "no other dict methods accessed" comment isn't needed anymore
- added `tests/unittests/cvat_utils_tests.py` with 28 tests for `_BasenameLookup` covering unique lookups, disambiguation, ambiguous cases (bare basename + multi-suffix), `KeyError` on missing/ambiguous `__getitem__`, `__contains__`, iteration, and edge cases

Burhan-Q:
Okay, @vittorio-prodomo apologies for the delay. The issue you're looking to resolve is that for any dataset with non-unique filenames, but unique parent directory/directories, fetched from CVAT into FiftyOne will either get dropped or have filename collisions:

```tree
.
|__ 0001
|   |__ A
|     |__ image000.jpg  # collision
|     |__ image001.jpg  # collision
|     ...
|__ 0002
   |__ A
   |  |__ image000.jpg  # collision
   |  |__ image010.jpg  # survives
   |  ...
   |__ B
   |  |__ image000.jpg  # survives
   |  |__ image001.jpg  # survives 
   | ...
```

Based this, there needs to be a way to include the data-path relative directories as part of the filename lookups to avoid collisions, but also to maintain the original path structure. Then there's the additional issue where the data-path is a relative path, and since FiftyOne normalizes to an absolute path, all relative keys from the database never match the absolute data-path, so the annotations are silently dropped. I think the simplest solutions to the three problems are:

## 1. Fix map construction to stop clobbering

For [L176-180](https://github.com/voxel51/fiftyone/blob/1235a8dbbcb511de28fbf8670d415554ea40682a/fiftyone/utils/cvat.py#L176-L180) (directory case): use relative paths from `data_path` as keys instead of `basenames`. Mirroring how `_load_data_map` already works in [`fiftyone/utils/data/importers.py:858-861`](https://github.com/voxel51/fiftyone/blob/1235a8dbbcb511de28fbf8670d415554ea40682a/fiftyone/utils/data/importers.py#L858-L861):

```python
data_map = {
    os.path.relpath(f, data_path): f
    for f in etau.list_files(data_path, abs_paths=True, recursive=True)
  }
```

Keys are relative paths from `data_path`, preserving directory structure and matching what CVAT stores for path-structured data"

For[ L171](https://github.com/voxel51/fiftyone/blob/1235a8dbbcb511de28fbf8670d415554ea40682a/fiftyone/utils/cvat.py#L171) (`data_path=None`) use the full `filepath` as the key:

```python
data_map = {f: f for f in existing_filepaths}
```

This guarantees zero collisions since `filepaths` in a sample collection are unique by definition.

## 2. Update the lookup to handle key format differences

On [L308-309](https://github.com/voxel51/fiftyone/blob/1235a8dbbcb511de28fbf8670d415554ea40682a/fiftyone/utils/cvat.py#L308-L309) CVAT's `frame["name"]` might be a bare `basename`, a relative path, or something else. A suffix-match fallback handles all cases:

```python
filename = frame["name"]
filepath = data_map.get(filename, None)
if filepath is None:
    for key, fp in data_map.items():
        if key.endswith(os.sep + filename) or key.endswith("/" + filename):
            filepath = fp
            break
```

When `data_map` keys are full absolute paths ([L171](https://github.com/voxel51/fiftyone/blob/1235a8dbbcb511de28fbf8670d415554ea40682a/fiftyone/utils/cvat.py#L171) case), `key.endswith("/0001/A/image000.jpg")` uniquely resolves the correct file. When keys are relative paths ([L176](https://github.com/voxel51/fiftyone/blob/1235a8dbbcb511de28fbf8670d415554ea40682a/fiftyone/utils/cvat.py#L176) case), it still works. The original exact-match path remains the fast path for the non-collision case.

## 3. Upfront normalization to absolute paths

For [L185](https://github.com/voxel51/fiftyone/blob/1235a8dbbcb511de28fbf8670d415554ea40682a/fiftyone/utils/cvat.py#L185) ensure that `data_dir` passed to `cvat_id_map` keys and database filepaths are _both_ absolute, meaning that lookups in `_build_sparse_frame_id_map` will succeed.

```python
data_dir = os.path.abspath(data_path)
```

I think these changes would resolve the issues you've raised, but also minimize the scope of the changes. Let me know what you think, and if you can, please give it a try to see if it works for both the standard (all unique filenames) and the path-prefix cases.

vittorio-prodomo:
@Burhan-Q Thanks for taking the time to sketch out a complete alternative. Walking through each fix:

  ## Fix #1a (directory case) — already in

  Your proposal:

  > For L176-180 (directory case): use relative paths from `data_path` as keys instead of `basenames`. [...]
  > ```python
  > data_map = {
  >     os.path.relpath(f, data_path): f
  >     for f in etau.list_files(data_path, abs_paths=True, recursive=True)
  >   }
  > ```

  is the current state, with one small addition — POSIX normalization of the relpath keys for Windows compatibility (flagged by CodeRabbit in an earlier round, commit 0bc9ded05):

  https://github.com/voxel51/fiftyone/blob/70feb3c3c9c5baf5e5f0021eb2fcd559701d2892/fiftyone/utils/cvat.py#L176-L181

  ## Fix #3 (absolute `data_dir`) — already in

  Same for:

  > ```python
  > data_dir = os.path.abspath(data_path)
  > ```

  already in place at:

  https://github.com/voxel51/fiftyone/blob/70feb3c3c9c5baf5e5f0021eb2fcd559701d2892/fiftyone/utils/cvat.py#L185

  ---

  So what remains under discussion is your suggested fix 1b (the `data_path=None` branch) and the number 2 (the suffix-match fallback). I wanted to flag two concerns with those.

  ## Time complexity on the `data_path=None` branch

  With `data_map = {f: f for f in existing_filepaths}`, the keys are absolute local filepaths. CVAT's `frame["name"]`, on the other hand, reflects whatever filename was sent at upload time. Looking at the upload paths I can see:

  - **fiftyone's own upload path**: `_parse_local_files` sends `filename_maker.get_output_path(path)` to CVAT as `client_files[N]`:

    https://github.com/voxel51/fiftyone/blob/70feb3c3c9c5baf5e5f0021eb2fcd559701d2892/fiftyone/utils/cvat.py#L4452-L4483

    and `UniqueFilenameMaker.get_output_path` returns either a bare basename or a `rel_dir`-relative path:

    https://github.com/voxel51/fiftyone/blob/70feb3c3c9c5baf5e5f0021eb2fcd559701d2892/fiftyone/core/utils.py#L2125-L2131

  - **`server_files` uploads**: CVAT stores paths relative to its mounted share root (e.g. `subdir/image.jpg`), not absolute client paths.

  None of these produce absolute paths on the client's local filesystem, so the exact-match fast path on `{f: f for f in existing_filepaths}` would miss in all of the code paths I can see, and every lookup
  falls through to the suffix scan:

  ```python
  for key, fp in data_map.items():
      if key.endswith(os.sep + filename) or key.endswith("/" + filename):
          ...
  ```

  For a sample collection with `N` filepaths and a CVAT task with `M` frames, that's `O(N·M)` `endswith` calls per task.

  You acknowledge this implicitly:

  > When `data_map` keys are full absolute paths (L171 case), `key.endswith("/0001/A/image000.jpg")` uniquely resolves the correct file.

  i.e. the suffix scan is doing all the work on that branch. `_BasenameLookup` indexes by basename so the common case (unique basenames) stays `O(M)` per task, and `O(M·k)` only when `k` filepaths share a basename (if there's an upload path I'm missing that does produce absolute-path frame names, let me know — it would change this analysis).

  ## Silent first-match on ambiguous cases

  The fallback returns on the first `endswith` hit:

  ```python
  for key, fp in data_map.items():
      if ...:
          filepath = fp
          break
  ```

  This silently picks one file in ambiguous cases. You also already flagged this pattern with your past comment on an earlier revision of `_BasenameLookup._resolve`:

  > will exit on the first match. This _could be_ an issue for a (likely rare) case like:
  > ```
  > /data/project/images/train/cat.jpg
  > /data/archive/images/train/cat.jpg
  > ```
  > since there are two files with `images/train/cat.jpg` as a basename. [...] I think at the very least it would make sense to log a warning, otherwise there could be a silent issue with multiple file
  matches.

  The current `_BasenameLookup._resolve` implements exactly that, collecting all suffix-match candidates and logging a warning when ambiguity remains, instead of silently picking one:

  https://github.com/voxel51/fiftyone/blob/70feb3c3c9c5baf5e5f0021eb2fcd559701d2892/fiftyone/utils/cvat.py#L336-L352

  ---

  Happy to keep iterating on the shape of the class itself, I perfectly understand the instinct to keep scope tight. Let me know what you think about my analysis.

Burhan-Q:
@vittorio-prodomo 

> CVAT stores paths relative to its mounted share root

I think this point is really the core of the problem. FiftyOne can't control how CVAT handles file paths and can only be aware of what happens on the FiftyOne end. I don't want to point my finger and say "go ask CVAT to change what they do" because they'd likely just point you back here, and it's just not helpful in the first place. Integrating other services is _always_ a challenge, since everyone has differing opinions on the mechanics of how to handle data and operations.

I think that the _best_ solution here is not a code change. Instead I think it's a documentation enhancement to explicitly inform users looking to import data from CVAT into FiftyOne, about the pitfalls due to the relative file paths used in CVAT. Such an update should also advise users to use unique file names and a relatively flat file structure when planning to work with both CVAT and FiftyOne, since doing so will reduce the likelihood of name collisions.

I'm not going to block a code change or even advocate against fixing the CVAT import issue with a code change. It's just that from my perspective, the current design is adding a lot more maintenance burden. Despite our efforts to maintain the `data_map` contract as being `dict`-like, it's behavior is still different _enough_ to introduce unnecessary complexity. The simplified version isn't perfect either for the reasons we've both called out, which is why I really think this has to be something documented and not necessary fixed.

In summary, I propose to modify the pull request to instead address the issue by updating the documentation and resetting the code changes. Documentation updates will inform users of the possible issue of name collisions and provide guidance on how to to structure their data when working with CVAT & FiftyOne to avoid missing data on import.

PR #7083, opened 2026-02-25
FOEPD-3026: improve ux on empty labels

lanzhenw:

When the “LABELS” section of the object/field list is empty and expanded, there’s a helpful empty state explaining how the user should check to be sure their field is enabled in the Explore tab

https://github.com/user-attachments/assets/1440596c-2438-469b-bbc6-da02064b2c91

        notes for FiftyOne users.

Improve empty annotation label ui

-   [ ] Other

## Release Notes

* **New Features**
  * Added an empty state message in the annotation sidebar that displays when no labels are available for annotation, with helpful guidance directing users to enable their fields in Explore for labeling.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/Annotate.tsx
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/Annotate.tsx
@@ -1,6 +1,6 @@
 import { LoadingSpinner } from "@fiftyone/components";
 import { EntryKind, isGeneratedView } from "@fiftyone/state";
-import { Typography } from "@mui/material";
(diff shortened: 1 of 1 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/Annotate.tsx
```diff
@@ -52,13 +52,25 @@ const Container = styled.div`
   overflow: auto;
 `;
 
+const EmptyLabelsContainer = styled.div`
```
erik-nieh, line 55: I think this is fine, it would be nice to use a `<Stack>` from voodo - but I'm curious if it actually solves the padding/gap this does correctly.

PR #7089, opened 2026-02-25
Handle input schema prompt for patches view

kaixi-wang:

- Fix blank sidebar when a user with manage permissions tries to annotate a patches view without first adding the patches field to the active label schema
- Differs from the normal schema manager input prompt because users should not be able to add another field or modify the dataset schema from the generated dataset, else this could lead to inconsistent states and lost data
- This also allows a user with manage access to immediately add the field to the active label schema without leaving the modal or resetting the view, which can be expensive.

- added tests
- run locally

https://github.com/user-attachments/assets/89720c78-37b6-4a11-ac03-f32757e6e148

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Detects missing source fields in generated views and surfaces a "Field not in label schema" prompt with an action to add the field to the label schema.
  * Import/setup UI now conditionally shows either the required-field prompt or a setup "Add schema" action.
  * Permissions-aware behavior: prompts and actions reflect manage permissions.

* **Tests**
  * Expanded unit and end-to-end tests covering missing-field detection, prompt flows, activation, permissions, and schema-manager scenarios.

### DIFF
--- a/e2e-pw/src/oss/specs/smoke-tests/schema-manager.spec.ts
+++ b/e2e-pw/src/oss/specs/smoke-tests/schema-manager.spec.ts
@@ -246,7 +246,6 @@ test.describe.serial("schema manager", () => {
     page,
     grid,
     modal,
-    schemaManager,
   }) => {
     // Navigate to patches view
     await fiftyoneLoader.waitUntilGridVisible(page, detectionDatasetName, {
@@ -256,6 +255,7 @@ test.describe.serial("schema manager", () => {
     // Open first patch in modal
     await grid.openFirstSample();
     await modal.assert.isOpen();
+    await modal.waitForSampleLoadDomAttribute();
     await modal.sidebar.switchMode("annotate");
 
     // The required field prompt should appear since "predictions" has no active schema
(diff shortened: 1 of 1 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/useSourceFieldToActivate.ts
```diff
@@ -0,0 +1,72 @@
...
+    kwarg: "field",
+    guard: ({ isPatches }) => isPatches,
+  },
+  // TODO: add entries for ToClips, ToFrames, etc.
+];
+
+/**
+ * Extracts the source field name from the view stages for stages with internal state
+ */
+export function getSourceFieldFromStages(
```
tom-vx51, line 27: gross :( we manually inspect the view stages in a few different places in the app, but it creates tight coupling and is buried in random places. Not your fault/don't need to block this PR, but we should find an actual solution for this. Is this all client-side or could we query a server endpoint for relevant metadata?

tom-vx51 (approved):
nothing blocking, but would like to find a better solution for view inspection

app/packages/core/src/components/Modal/Sidebar/Annotate/useSourceFieldToActivate.ts
kaixi-wang: This is all client-side. I don't see the benefits of a server endpoint here (you'd be passing the same view to the server and then doing the same inspecting in python). 

My ideal would be that generated views are never materialized and we'd never have to deal with this syncing. A patchies view is just a flattened array field stored in a separate one-off dataset and sample collection. It can't even be reused unless accessing through a saved view so it really has no relevance outside of the current session.

tom-vx51: The benefit would be that we can actually parse the view on the server. Right now the FE needs to have knowledge of how views are constructed, how view stages are serialized, and rely on raw string comparisons.

PR #7090, opened 2026-02-25
Fix stale sample model after patches view edits

kaixi-wang:

Fix cache issue when returning to the full dataset view after making annotation edits in the patches view and opening up the sample modal. 

- test locally
- added unit tests

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where modal views for generated samples could display stale cached data. Modal records are now properly cleaned up during updates to ensure fresh data is fetched on reopening.

* **Tests**
  * Added comprehensive test suite for sample update functionality, covering store record updates, cleanup operations, and integration scenarios.

### DIFF
--- a/app/packages/state/src/hooks/useUpdateSamples.ts
+++ b/app/packages/state/src/hooks/useUpdateSamples.ts
@@ -3,77 +3,6 @@ import { commitLocalUpdate, useRelayEnvironment } from "react-relay";
 import type { ModalSample } from "../recoil";
 import { stores } from "./useLookerStore";
 
-/**
- * For generated view samples (patches/clips/frames), update the source
- * sample's cached label data in the Relay store so the modal shows fresh
- * data when the user returns to the source dataset view.
- *
- * The patch sample's `_id` is the label's `_id` within the source sample.
- * We find that label in the source sample's fields and replace it with the
- * updated data from the patch sample.
- */
-function updateSourceSampleLabel(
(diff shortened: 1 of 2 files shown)


app/packages/state/src/hooks/useUpdateSamples.ts
```diff
@@ -36,6 +36,13 @@ export const useUpdateSamples = () => {
               }
             }
           }
+
+          // For generated views (patches/clips/frames), delete the source
+          // sample's cached modal record so the next modal open fetches fresh data
+          const sourceSampleId = (sample as Record<string, unknown>)._sample_id;
+          if (typeof sourceSampleId === "string") {
+            store.delete(`${sourceSampleId}-modal`);
```
tom-vx51, line 44: can we set the value here instead of just removing the entry? i.e. by updating line 25 to
```
const ids = [id, `${id}-modal`, `${sample._sample_id}-modal`];
```

kaixi-wang: The sample is the generated/patches sample (the flattened field) and not equivalent to the src sample, so we don't want to overwrite the src sample with the patch data (unless setValue is doing something smart). If we want to set the value, we would have to lookup the src sample and manually set the field based off the generated sample data, which looked really hacky.

Also I tried just record.invalidateRecord(), but that wasn't effective and the cache was still stale.

FWIW I really dislike that we need to deal with this and why patches can't just be a flattened field on the same dataset :(

PR #7093, opened 2026-02-25
Fix plugin docs crash on empty headings

AdonaiVera:

The doc build was failing with `AssertionError: Empty title` during both the HTML and Markdown build steps when processing plugin pages generated by `generate_plugin_docs.py`.

I fixed applying two conditions
- After `_remove_emojis`, strip heading lines whose text is empty (`re.sub(r"^#+\s*$", "", ...)`) so ghost headings don't reach either builder.
- Check whether any heading with actual text content remains. If not, prepend `# {display_name}` as a guaranteed fallback title

Built in local

        notes for FiftyOne users.

-   [X] Documentation: FiftyOne documentation changes
-   [ ] Other

## Summary by CodeRabbit

* **Documentation**
  * Improved plugin documentation generation: removes stray/standalone heading lines and normalizes headings in README files.
  * If a plugin README lacks any headings, a top-level title is automatically added using the plugin display name.
  * Preserves existing behavior for community plugin notes so original content remains intact.

### DIFF
--- a/docs/scripts/generate_plugin_docs.py
+++ b/docs/scripts/generate_plugin_docs.py
@@ -129,10 +129,10 @@ class PluginDocGenerator:
         )
         self.load_dataset_pattern = re.compile(r"\bdef\s+load_dataset\s*\(")
         self.empty_heading_pattern = re.compile(
-            r"^[ \t]{0,3}#{1,6}[ \t]*$",
+            r"^\s*#{1,6}\s*$",
             flags=re.MULTILINE,
         )
-        self.heading_pattern = re.compile(r"^[ \t]{0,3}#{1,6}[ \t]+\S")
+        self.heading_pattern = re.compile(r"^\s*#{1,6}\s+\S")
 
     def _remove_emojis(self, text: str) -> str:
         """Remove emoji and miscellaneous symbols from a string.


docs/scripts/generate_plugin_docs.py
```diff
@@ -825,6 +825,18 @@ def generate_plugins_ecosystem_rst(self, all_plugins: List[Plugin]) -> str:
                 frontmatter = self._generate_frontmatter(seo_metadata)
                 github_badge = self._generate_github_badge(plugin)
 
+                processed_readme = re.sub(
+                    r"^#+\s*$", "", processed_readme, flags=re.MULTILINE
+                )
+
+                has_heading = any(
+                    line.lstrip().startswith("#")
+                    and len(line.lstrip("#").strip()) > 0
```
Burhan-Q:
```suggestion
                    and len(line.replace("#", "").strip()) > 0
```

Burhan-Q: Given the Coderabbit review, but suggesting an easier change.

docs/scripts/generate_plugin_docs.py
```diff
@@ -825,6 +825,18 @@ def generate_plugins_ecosystem_rst(self, all_plugins: List[Plugin]) -> str:
                 frontmatter = self._generate_frontmatter(seo_metadata)
                 github_badge = self._generate_github_badge(plugin)
 
+                processed_readme = re.sub(
+                    r"^#+\s*$", "", processed_readme, flags=re.MULTILINE
```
Burhan-Q: Might want to compile the pattern before the loop starts. Should help processing run a bit quicker.

AdonaiVera: Yes, sir, thank you for the suggestion, we have a `_compile_regex_patterns` function where we register all the regex, I just moved it to there

docs/scripts/generate_plugin_docs.py
```diff
@@ -825,6 +825,18 @@ def generate_plugins_ecosystem_rst(self, all_plugins: List[Plugin]) -> str:
...
+                )
+
+                has_heading = any(
+                    line.lstrip().startswith("#")
+                    and len(line.lstrip("#").strip()) > 0
+                    for line in processed_readme.splitlines()
+                )
+                if not has_heading:
+                    processed_readme = f"# {display_name}\n\n" + processed_readme
+
```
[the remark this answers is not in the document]
AdonaiVera: Nice! I just applied your suggested patterns but moved both into _compile_regex_patterns() so they compile once at instantiation     rather than per plugin processed, consistent with the class's existing convention. Thx

docs/scripts/generate_plugin_docs.py
AdonaiVera: Thanks for the suggestion Burhan, this solution does fix the reported case, but it strips # from the entire line including content, so a heading like "## #" would be misidentified as empty. I'm going to apply rabbit suggestion here, thank you

docs/scripts/generate_plugin_docs.py
```diff
@@ -128,6 +128,11 @@ def _compile_regex_patterns(self):
             r"\bdef\s+download_model\s*\("
         )
         self.load_dataset_pattern = re.compile(r"\bdef\s+load_dataset\s*\(")
+        self.empty_heading_pattern = re.compile(
+            r"^[ \t]{0,3}#{1,6}[ \t]*$",
```
Burhan-Q:
```suggestion
            r"^\s*#{1,6}\s*$",
```

Burhan-Q: Previous pattern `^[ \t]{0,3}` would not match 4 spaces, 3 space + tab, new-line + spaces + tabs, etc. at start of string. New pattern `^\s*` will match 0 or more white space characters at start of line, and 0 or more white space characters at the end of the line. If looking to prevent infinite matching of white space characters, could use `r"^(\s){,64}#{1,6}(\s){,64}$"` instead.

docs/scripts/generate_plugin_docs.py
```diff
@@ -128,6 +128,11 @@ def _compile_regex_patterns(self):
...
+            flags=re.MULTILINE,
+        )
+        self.heading_pattern = re.compile(r"^[ \t]{0,3}#{1,6}[ \t]+\S")
```
Burhan-Q:
```suggestion
        self.heading_pattern = re.compile(r"^\s*#{1,6}\s+\S")
```

Burhan-Q: Same as other pattern. Additionally, kept `\S` ending since I presume that the aim is to ensure that there's _at least_ $1$ non-white space character on the heading line, as opposed to using `/S+` or `/S{3,}` to ensure that there's at least more than a single non-white space character for the header.

docs/scripts/generate_plugin_docs.py
Burhan-Q: Simple test case for original pattern:

```python
import re

s = " " * 4
s += "#"
p = re.compile(r"^[ \t]{0,3}#{1,6}[ \t]*$")

p.search(s) is None
>>> True
```

Burhan-Q (approved):

PR #7094, opened 2026-02-26
Allow plugins to override grid and modal default file rendering

sashankaryal:

This PR extends FiftyOne plugin panels API to allow plugins to declare file-renderer claims and automatically take over modal or grid rendering for matching file extensions, mime types, or media types.

Some unit tests and manual testing with dummy plugins.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added support for custom sample renderers in plugins, enabling personalized rendering of media files in grid and modal views with automatic fallback on errors.

* **Documentation**
  * Added comprehensive guide for developing custom sample renderers with configuration options and examples.


app/packages/core/src/components/Grid/GridCustomRendererItem.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+} as const;
+
+const TestBridge = ({ children }: React.PropsWithChildren) => <>{children}</>;
+
+describe("GridRenderClaimsLooker", () => {
+  let consoleErrorSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    // Suppress React error boundary noise in test output
+    consoleErrorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
```
ritch, line 47: [nit]

We don't want to see any console.error calls at all?

Just shouting out this pattern:

```
      try {
        doSomething()
      } catch (error) {
        console.error(
           `Some useful info about ${error}`
        );
      }
```

docs/source/plugins/developing_plugins.rst
```diff
@@ -4710,6 +4710,84 @@ Using the custom component as the view for a Python operator field:
...
+          style={{ width: "100%", height: "100%", border: 0 }}
+        />
+      );
+    }
+
+    registerComponent({
+      name: "PDFRenderer",
+      label: "PDF renderer",
+      component: PdfRenderer,
+      type: PluginComponentType.Panel,
```
ritch: Why overload `Panel` instead of adding a new type eg. `SampleRenderer`?

docs/source/plugins/developing_plugins.rst
```diff
@@ -4710,6 +4710,84 @@ Using the custom component as the view for a Python operator field:
...
+          mediaTypes: ["unknown"],
+          modeExtensions: [
+            RENDER_CLAIM_MODE_EXTENSION_GRID,
+            RENDER_CLAIM_MODE_EXTENSION_MODAL_ANNOTATE,
+          ],
+        },
+      },
+    });
+
+If multiple panels match the same sample, the one with the highest
```
ritch: Glad to see the behavior is defined, but I'm guessing the user would want to define which renderer wins. Not requesting a change but just mentioning something I think might resurface if this becomes widely adopted.

docs/source/plugins/developing_plugins.rst
```diff
@@ -4710,6 +4710,84 @@ Using the custom component as the view for a Python operator field:
...
+      component: PdfRenderer,
+      type: PluginComponentType.Panel,
+      activator: () => true,
+      panelOptions: {
+        surfaces: "grid modal",
+        priority: 100,
+        renderClaims: {
+          extensions: [".pdf"],
+          mimeTypes: ["application/pdf"],
+          mediaTypes: ["unknown"],
```
ritch, line 4889: Do you explain this anywhere? Kind of confusing IMO.

docs/source/plugins/developing_plugins.rst
```diff
@@ -4710,6 +4710,84 @@ Using the custom component as the view for a Python operator field:
...
+``renderClaims.modeExtensions``:
+
+-   Include ``RENDER_CLAIM_MODE_EXTENSION_GRID`` to enable rendering in the
+    grid
+-   Include ``RENDER_CLAIM_MODE_EXTENSION_MODAL_ANNOTATE`` to enable
+    rendering in the modal's annotate mode
+
+The claim-based renderer component receives two props:
+
+-   ``ctx``: a rendering context object containing ``sample``,
```
ritch: Should flesh this out more and describe each is if this were the reference OR link to an API doc.

Also should probably mention that you can use `fos` and other core modules, espeically the atoms around labels or other important features.

docs/source/plugins/developing_plugins.rst
```diff
@@ -4710,6 +4710,84 @@ Using the custom component as the view for a Python operator field:
...
+    import {
+      PluginComponentType,
+      RENDER_CLAIM_MODE_EXTENSION_GRID,
+      RENDER_CLAIM_MODE_EXTENSION_MODAL_ANNOTATE,
+      registerComponent,
+    } from "@fiftyone/plugins";
+
+    function PdfRenderer({ ctx, url }) {
+      return (
+        <iframe
```
ritch, line 4870: Might be more useful to include a more complete example...either here or in fiftyone/plugins.

app/packages/core/src/components/Grid/useRenderer.ts
```diff
@@ -55,15 +62,33 @@ export default function ({
...
+            id,
+            getFontSize()
+          );
+        } catch (e) {
+          console.error("Failed to create plugin renderer, using default:", e);
+          looker = createLooker.current?.(
+            { ...result, symbol: id },
+            { fontSize: getFontSize() }
+          ) as fos.Lookers;
         }
```
benjaminpkane: If `renderClaims` already has a reference to `createLooker`, shouldn't it always handle looker creation?

app/packages/core/src/components/Grid/useGridRenderClaimsLooker.ts
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Hook that wraps default looker creation with render claims support.
+ */
+export function useGridRenderClaimsLooker(
+  createDefaultLooker: ReturnType<typeof fos.useCreateLooker>
+) {
+  const dataset = useRecoilValue(fos.dataset);
+  const schema = useRecoilValue(
+    fos.fieldSchema({ space: fos.State.SPACE.SAMPLE })
+  );
```
benjaminpkane: Looks like this could be a common hook to hide recoil? `usePanelContext` or similar

app/packages/core/src/components/Grid/useGridRenderClaimsLooker.ts
```diff
@@ -0,0 +1,143 @@
...
+      }
+
+      const matchedPanel = getMatchingRenderClaimsPanel(panelPlugins, ctx, {
+        surface: "grid",
+      });
+
+      return (
+        !!matchedPanel &&
+        !!ctx.selectedMediaUrl &&
+        !!getRawComponent(matchedPanel.name)
```
benjaminpkane: Why would a matched panel not have a selected media URL or raw component?

app/packages/core/src/components/Modal/modal-spaces-utils.ts
```diff
@@ -13,8 +13,9 @@ export const useModalSpaces = () => {
 
   const panelsPredicate = useCallback(
     (panel: PluginComponentRegistration) =>
-      panel.panelOptions?.surfaces === "modal" ||
-      panel.panelOptions?.surfaces === "grid modal",
+      (panel.panelOptions?.surfaces === "modal" ||
+        panel.panelOptions?.surfaces === "grid modal") &&
+      !panel.panelOptions?.renderClaims,
```
benjaminpkane: Why does renderClaims invalidate this?

app/packages/plugins/src/index.ts
```diff
@@ -469,14 +486,17 @@ class PluginComponentRegistry {
   hasScript(name: string) {
     return this.scripts.has(name);
   }
+  getRawComponent(name: string) {
+    return this.rawComponents.get(name);
```
benjaminpkane: Why is `rawComponent` required? The plugin system now holds two different references to a plugin component?

app/packages/core/src/components/Grid/useRenderer.ts
```diff
@@ -21,6 +22,12 @@ export default function ({
   const createLooker = fos.useCreateLooker(false, true, lookerOptions);
   const getFontSize = useFontSize(id);
   const selectSample = useSelectSample(records);
+  const renderClaims = useGridRenderClaimsLooker(createLooker);
+
+  // We need returned `renderer` to be referentially stable, which requires referential stability for `showItem`
+  // since `showItem` depends on `renderClaims`, we use a ref
+  const renderClaimsRef = useRef(renderClaims);
+  renderClaimsRef.current = renderClaims;
```
benjaminpkane: Can `useGridRenderClaimsLooker` return a ref? It is only used by grid, so it seems doable

app/packages/core/src/components/Modal/ModalFileRendererOverride.tsx
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Renders a plugin-claimed renderer for non-native media in the modal,
+ * or falls back to MetadataLooker when no matching plugin is found.
+ */
+export const ModalFileRendererOverride = React.memo(
+  ({ sample, modalMediaField, isAnnotate }: ModalFileRendererOverrideProps) => {
+    const dataset = useRecoilValue(fos.dataset);
+    const schema = useRecoilValue(
+      fos.fieldSchema({ space: fos.State.SPACE.SAMPLE })
+    );
```
benjaminpkane: `usePanelContext` applies here too

benjaminpkane (changes requested):
I'm also realizing selection isn't supported for these custom renderers? That is a requested change. Also noting coordination for https://github.com/voxel51/fiftyone/pull/7109

And to be specific, I think this feature needs to deploy to an ephem before approval.

app/packages/core/src/components/Grid/GridTagBubbles.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+const getTagAttribute = (path: string | undefined, value: string) => {
+  const tagValue = value.replace(/[\s.,/]/g, "-").toLowerCase();
+  const pathValue = path ?? "undefined";
+  return ["tags", "_label_tags"].includes(pathValue)
+    ? `tag-${pathValue}-${tagValue}`
+    : `tag-${pathValue}`;
+};
+
+export default function GridTagBubbles({ sample }: GridTagBubblesProps) {
```
sashankaryal, line 24: Used a different path for this because existing constructs that render tag/primitive "bubbles" is very coupled to looker and I didn't want to touch that code

app/packages/core/src/components/Modal/ModalSampleRenderer.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const schema = fos.useModalSampleSchema();
+
+    const sampleRenderers = useActivePlugins(
+      PluginComponentType.SampleRenderer
+    );
+
+    const fallback = <MetadataLooker sample={sample} />;
+
+    if (!dataset) {
+      return fallback;
```
ritch: nit: create fallback only when used.

```tsx
return <MetadataLooker sample={sample} />;
```

app/packages/core/src/components/Modal/ModalSampleRenderer.tsx
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Renders a matched sample renderer for non-native modal media, or falls back
+ * to the built-in metadata renderer when no renderer is available.
+ */
+export const ModalSampleRenderer = React.memo(
+  ({ sample, modalMediaField }: ModalSampleRendererProps) => {
+    const dataset = fos.useCurrentDataset();
+    const schema = fos.useModalSampleSchema();
+
+    const sampleRenderers = useActivePlugins(
```
ritch, line 63: I thought`useActivePlugins` depended on `ctx` being provied as the second arg...?

ritch: I'll assume this just means `active = true`.

ritch (approved):
Plugin API and implementation LGTM. I did not look too closely at the looker/grid aspects.

app/packages/core/src/components/Modal/ModalSampleRenderer.tsx
sashankaryal: ctx gets constructed to `{dataset, schema}` if not passed

app/packages/plugins/src/sample-renderer.ts
```diff
@@ -0,0 +1,400 @@
...
+}
+
+function isMatchMediaObject(
+  supports:
+    | MatchMedia
+    | ((ctx: SampleRendererMatchContext<unknown>) => boolean)
+    | undefined
+): supports is MatchMedia {
+  return Boolean(supports) && typeof supports !== "function";
+}
```
benjaminpkane: Unused so I removed it

docs/source/plugins/developing_plugins.rst
```diff
@@ -4710,6 +4710,74 @@ Using the custom component as the view for a Python operator field:
...
+
+The sample renderer component receives a single ``ctx`` prop containing the
+sample renderer render context:
+
+-   ``sample`` and ``media``
+-   ``surface`` (``"modal"`` or ``"grid"``)
+-   ``dataset`` and ``schema``
+
+Here's an example of a hypothetical plugin config that renders PDF files in
+both the grid and modal:
```
benjaminpkane, line 4857: This looks like a real plugin config, not a hypothetical one 🚀 
```suggestion
Here's an example plugin config that renders PDF files in both the grid and modal:
```

app/packages/core/src/components/Grid/useGridRenderClaimsLooker.ts
sashankaryal: Done

app/packages/spaces/src/hooks.ts
```diff
@@ -41,6 +42,8 @@ import {
 } from "./types";
 import { getNodes } from "./utils";
 
+type SpacePanelRegistration = PanelRegistration | PlotRegistration;
```
benjaminpkane, line 45: This isn't my domain, so I was surprised to see this. A `SpacePanel` is either a `Panel` or `Plot`? A comment would help

app/packages/plugins/src/sample-renderer.ts
```diff
@@ -0,0 +1,400 @@
...
+};
+
+/**
+ * Declarative media matchers used to determine renderer compatibility.
+ */
+export type MatchMedia = {
+  extensions?: string[];
+  mimeTypes?: string[];
+  mediaTypes?: string[];
+};
```
benjaminpkane: `NormalizedMatchMedia` duplicated `MatchMedia` so I removed it

app/packages/plugins/src/sample-renderer-registration.typecheck.ts
```diff
@@ -0,0 +1,74 @@
+import {
```
benjaminpkane, line 1: I think I understand the purpose of this file, but I've never seen it before. Is there any more context you can share? Is `.typecheck.ts` a convention?

app/packages/plugins/src/index.ts
```diff
@@ -245,37 +265,50 @@ export function safePluginActivator(
   return false;
 }
 
+export function useGetDefaultPluginCtx() {
```
benjaminpkane: I don't believe this function needed to be exported

app/packages/plugins/src/index.ts
```diff
@@ -556,5 +668,29 @@ export function usePluginSettings<T>(
...
+  getFileExtension,
+  getMatchingSampleRenderer,
+  getSampleRendererComponent,
+  getSelectedMediaPath,
+  hasMatchMediaMatchers,
+  isSampleRendererGridEnabled,
+  matchesMatchMedia,
+  normalizeMatchMedia,
+  sortSampleRenderersByPriority,
+  supportsSampleRenderer,
```
benjaminpkane: Unclear why all these exported. Some of them seem like internal utils

app/packages/spaces/src/hooks.ts
benjaminpkane: Looks like `Plot` is deprecated

sashankaryal: It is...

app/packages/plugins/src/sample-renderer-registration.typecheck.ts
sashankaryal: There was an internal thread about TS ergonomics and I thought it'd be nice to capture that in code since you can do TS assertions... not really an established convention, though 🤷  it can be, now.

app/packages/relay/src/selectorWithEffect.tsx
```diff
@@ -24,6 +24,20 @@ export function SelectorEffectContext({
...
+ * This is useful when a piece of state is primarily derived from another atom
+ * or selector, but writes must still flow through an external sync layer such
+ * as the Relay writer/session bridge. The returned selector:
+ *
+ * - reads exactly like a normal selector via `options.get`
+ * - looks up a matching setter from the effect store using `itemKey` or
+ *   `options.key`
+ * - optionally transforms the write payload via `options.set`
+ * - optionally mirrors the final value into `state` for local Recoil updates
+ */
```
sashankaryal, line 40: I was touching this in the middle - but removed again, but decided to leave the claude generated docstring

benjaminpkane:
Looks great. I have three final requests.

1. I see a banner was added for notifying users that customer rendering failed. Since it is implemented in `@fiftyone/app`, the Enterprise version will need an implementation too. Has that been done? I would be remiss if I did not mention that using a notification Toast would be sufficient from POV (easier to maintain). Either way, requesting confirmation of equivalent behavior in Enterprise.

2. I also see that grid error handling is well-defined, but I cannot open the modal and see a fallback rendering when custom modal rendering raises an error. This is inconsistent, and should be addressed.

https://github.com/user-attachments/assets/2a818fce-5ef2-40b6-8980-46329765b1e8

3. Can we add event tracking to gauge the popularity of this feature after its release? If this is already the case, is that documented somewhere?

sashankaryal:
> Looks great. I have three final requests.
> 
> 1. I see a banner was added for notifying users that customer rendering failed. Since it is implemented in `@fiftyone/app`, the Enterprise version will need an implementation too. Has that been done? I would be remiss if I did not mention that using a notification Toast would be sufficient from POV (easier to maintain). Either way, requesting confirmation of equivalent behavior in Enterprise.
> 2. I also see that grid error handling is well-defined, but I cannot open the modal and see a fallback rendering when custom modal rendering raises an error. This is inconsistent, and should be addressed.
> 
>  Screen.Recording.2026-03-26.at.10.06.58.AM.mov 
> 3. Can we add event tracking to gauge the popularity of this feature after its release? If this is already the case, is that documented somewhere?

re 1: fine switching to a toast

re 2: regression on my part, will fix, thanks for flagging. I believe it should be possible to write basic e2e specs for this feature, will attempt in subsequent PR

re 3: sure

app/packages/core/src/components/Modal/ModalSampleRenderer.tsx
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Renders a matched sample renderer for non-native modal media, or falls back
+ * to the built-in metadata renderer when no renderer is available.
+ */
+export const ModalSampleRenderer = React.memo(
+  ({ sample, modalMediaField }: ModalSampleRendererProps) => {
+    const dataset = fos.useCurrentDataset();
+    const schema = fos.useModalSampleSchema();
+    const { isDisabled: isDatasetRendererDisabled } =
+      fos.useGridCustomRendererFailover(dataset?.name);
```
benjaminpkane, line 61: I was surprised to see a `useGrid` hook here. A comment would help

PR #7102, opened 2026-02-26
FOEPD-3147 Resolve Stacking Issue With HA Label Tooltip and Schema Manager

nverjinski:
- Updated the `z-index` of `ModalBackground` to be higher than the `Tooltip`'s `z-index`

1. Open a sample with object detections in Annotate
2. Hover over a label instance, and click `ctrl` to persist the tooltip
3. Open the schema manager and verify the tooltip is **behind** the resulting modal

Before:

https://github.com/user-attachments/assets/94c5f7ad-5de5-48a3-a1f0-7208f0b7fc33

After:

https://github.com/user-attachments/assets/e97c84c5-efad-405d-b996-414ede4824c9

## Release Notes
- Fixed bug where tooltips would improperly render above Schema Manager

        notes for FiftyOne users.

- Fixed bug where tooltips would improperly render above Schema Manager

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed modal dialog stacking so modals consistently appear above other page elements.

* **Style**
  * Minor formatting and whitespace improvements in tooltip-related UI code with no user-visible behavior changes.

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/styled.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/styled.ts
@@ -186,7 +186,7 @@ export const ModalBackground = styled.div`
   background: rgba(0, 0, 0, 0.8);
   top: 0;
   left: 0;
-  z-index: 20001;
+  z-index: 2000;
   display: flex;
   justify-content: center;
   align-items: center;
--- a/app/packages/core/src/components/Modal/TooltipInfo.tsx
+++ b/app/packages/core/src/components/Modal/TooltipInfo.tsx
@@ -29,12 +29,12 @@ import { QuickEditEntry } from "./Sidebar/Annotate";
 
 const TOOLTIP_HEADER_ID = "fo-tooltip-header";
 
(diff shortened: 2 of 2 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/styled.ts
```diff
@@ -186,7 +186,7 @@ export const ModalBackground = styled.div`
   background: rgba(0, 0, 0, 0.8);
   top: 0;
   left: 0;
-  z-index: 1001;
+  z-index: 20001;
```
lanzhenw: This may cause dropdown selection and tooltip on the schema manager to be behind the modal. Let's check. 

https://github.com/voxel51/design-system/blob/develop/src/components/Select/Select.tsx
Line 228 is the dropdown z-index for above modal
https://github.com/voxel51/design-system/blob/develop/src/components/Tooltip/Tooltip.tsx
Line 99 is the tooltip z-index for above modal

so, the tooltip will behind the schema manager modal. 20001 is too big. We usually have modals around 1000-2000. What's the other tooltip's z-index? If there's a conflict, we may have to bump the design-system z-index

PR #7109, opened 2026-02-27
Support alternative selecting samples

lanzhenw:

This PR adds a second way of selecting samples on the samples grid and provided a list of UI icons for style customization. Why? Python panels sometimes need more than one way to select samples on the grid. Yet, currently it is not possible in Fiftyone. This PR adds a configurable alternative way of selection on samples grid, so that users can have better UX with grid selection in few shots, similarity search panels and the future VAL panel. 

## Release Notes

* **New Features**
  * Added selection style customization with multiple icon options (checkmark, thumbs up/down, pin, star, bookmark, X).
  * Added alternative selection mode accessible via alt-click.
  * Enhanced selection metadata tracking to support per-sample selection state and type information.
  * Improved selection UI with dynamic icons that change based on selection mode and style.
  
  There will be a follow-up PR that add alternative selection to labels on samples grid and sample modal in the same pattern. 

```
  export type SelectionIconStyle =
  | "checkmark"
  | "green-checkmark"
  | "red-checkmark"
  | "thumbsup"
  | "thumbsdown"
  | "pin"
  | "star"
  | "x"
  | "bookmark"
```

Screenshot 2026-03-03 at 6 30 32 PM

How to test it?
  
 ## in python SDK
  
  ```
  import fiftyone as fo
  import fiftyone.zoo as foz

  ids = dataset.limit(10).values("id")

  # Set selection style to thumbsup/thumbsdown
  session.set_sample_selection_style(default="thumbsup", alt="thumbsdown")

  # Select with mixed types via selected_samples
  session.selected_samples = [
      {"id": ids[0], "type": "default"},  # thumbsup
      {"id": ids[1], "type": "alt"},       # thumbsdown
      {"id": ids[2], "type": "default"},   # thumbsup
      {"id": ids[3], "type": "alt"},       # thumbsdown
  ]

  # Verify: selected returns flat IDs, selected_samples has types
  print("selected:", session.selected)
  print("selected_samples:", session.selected_samples)
  print("style:", session.sample_selection_style)

  # --- 2. Backward compat: session.selected = flat list (all default) ---
  session.selected = [ids[4], ids[5]]
  print("selected_samples:", session.selected_samples)
  # should be type "default"

  # --- 3. selected setter preserves existing types ---
  session.selected_samples = [
      {"id": ids[0], "type": "alt"},
      {"id": ids[1], "type": "default"},
  ]

  # Now add id2 via the flat setter — existing types preserved
  session.selected = [ids[0], ids[1], ids[2]]
  print("\nAfter preserving types + adding new:")
  print("selected_samples:", session.selected_samples)
  # ids[0] should still be "alt", ids[1] "default", ids[2] "default"

  # --- 4. Clear selection ---
  session.clear_selected()
  print("\nAfter clear:", session.selected, session.selected_samples)

  # --- 5. Style resets on dataset switch ---
  session.set_sample_selection_style(default="star", alt="pin")
  print("\nStyle before switch:", session.sample_selection_style)
  # In the UI, go to a different dataset
  print("Style after switch:", session.sample_selection_style)
  ```
  
## To test alt-click in App 
load a dataset with samples, then alt click, you should see normal checkbox
Turn on alt-click by setting the sample selections tyle:
```
  session.dataset = dataset
  session.set_sample_selection_style(default="thumbsup", alt="thumbsdown")
```
click a sample -> thumbsup icon
alt+click a sample -> thumbsdown icon
shift+click -> range select 
shift+alt+click -> range alt select
go to a group dataset, set sample selection style, check carousel thumbnail show the correct icons

## To test the changes in execution context 

```
  # In a plugin, test:
  # ctx.ops.set_selected_samples(["id1", "id2"])           # backward compat
  # ctx.ops.set_selected_samples([{"id": "id1", "type": "alt"}])  # new
  # ctx.ops.set_sample_selection_style(default="star", alt="x")
  # ctx.ops.clear_sample_selection_style()
  # ctx.selected           -> flat list
  # ctx.selected_samples   -> list of dicts
  ```

        notes for FiftyOne users.

adds a configurable alternative way of selection on samples grid, so that users can have better UX with grid selection in few shots, similarity search panels and the future VAL panel. 

-   [ ] Other

## Release Notes

* **New Features**
  * Added customizable selection styling with multiple icon options (checkmark, thumbs up/down, star, pin, bookmark, and more).
  * Introduced Alt+click for alternative selection mode alongside standard selection.
  * Enhanced sample selection with per-sample type tracking and metadata.
  * Added selection style configuration methods for default and alternative selection modes.

* **Bug Fixes**
  * Improved robustness of selection data handling with better support for missing or undefined values.

* **Tests**
  * Added comprehensive unit tests for new selection utilities and event handling.

### CODE REVIEW

File: app/packages/core/src/components/Grid/useSelectSample.ts [hunk 1]
```diff
@@ -106,36 +109,67 @@ export default (records: Records) => {
...
-            ? selected.delete(sampleId)
-            : selected.add(sampleId);
-        }
 
-        if (selectedObjects.has(sampleId)) {
-          selectedObjects.delete(sampleId);
+        if (shiftKey && !current.has(sampleId)) {
+          // Shift-click (or shift+alt-click) range add
+          const newSelected = addRange(index, current, records);
+          for (const id of newSelected) {
```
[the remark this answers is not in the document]
lanzhenw: good catch

fiftyone/core/session/session.py
```diff
@@ -814,7 +819,62 @@ def select_samples(
...
+    def clear_selection_style(self) -> None:
+        """Clears the selection style, reverting to default checkmark."""
+        style = {"default": "checkmark"}
+        self._state.selection_style = style
+        # Convert all alt meta to default
+        if self._state.selected_meta:
+            self._state.selected_meta = {
+                k: {"type": "default"} for k in self._state.selected_meta
+            }
+        self._client.send_event(SetSelectionStyle(style=style))
```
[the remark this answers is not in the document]
lanzhenw: This finding is based on the old code where clear_selection_style was converting all alt meta to default.
It's removed — now it just resets the style. No meta normalization happens, so there's no stale meta to propagate. issue resolved.

app/packages/core/src/components/Grid/useEscape.ts
```diff
@@ -27,6 +27,7 @@ const useEscape = () => {
               confirm("Are you sure you want to clear your current selection?")
             ) {
               reset(fos.selectedSamples);
+              reset(fos.selectedMeta);
```
ritch: This is highlighting a code smell we have. It leaks the coupling and implementation details of selection. Instead we should use the domain store pattern that is used elsewhere.

The rest of the app should only know about this domain store and internally it just happens to use recoil.

```tsx
export type SelectionStore = {
  // state
  selectedSamples: string[];
  selectedMeta: SelectionMeta;
  hasSelection: boolean;

  // actions
  _setSelectedSamples: (ids: string[]) => void;
  _setSelectedMeta: (meta: SelectionMeta) => void;
  clear: () => void; // clears both

  // convenience actions
  addSample: (id: string, meta: SelectionMetaItem) => void;
  removeSample: (id: string) => void;
};
```

The concept is covered pretty well in [this article](https://dev.to/blindpupil/domain-driven-architecture-in-the-frontend-i-1f41).

lanzhenw: In the new proposed syntax (see below), we would remove the concept of samples and labels meta and thus condense the source of truth to one session atom. 

```
# Selected sample IDs, regardless of selection type
ctx.selected = [list, of, all, ids]

# NEW PROPERTY shows selected samples with 'type' key
ctx.selected_samples = [
    {
        "sample_id": "XXX",
        "type": "default" or "alt",
    },
    ...
]


ctx.selected_samples = [
    {
        "sample_id": "XXX",
        "type": "default" or "alt",
    },
    ...
]

# NEW ATTRIBUTE: 'type' key shows selection type for selected labels
ctx.selected_labels = [
    {
        "sample_id": "XXX",
        "label_id": "YYY",
        "instance_id": "ZZZ" or None,
        "field": "ground_truth",
        "frame_number": 1 or None,
        "type": "default" or "alt",  # NEW KEY
    },
    ...
]

# NEW SYNTAX: select samples with custom selection types
ctx.ops.set_selected_samples(
    [
        {
            "sample_id": "XXX",
            "type": "default" or "alt",
        },
        ...
    ]
)
# NEW SYNTAX: select labels with custom selection types
ctx.ops.set_selected_labels(
    [
        {
            "sample_id": "XXX",
            "label_id": "YYY",
            "instance_id": "ZZZ" or None,  # optional
            "field": "ground_truth",
            "frame_number": 1,  # frame labels only
            "type": "default" or "alt",  # NEW OPTIONAL KEY
        },
        ...
    ],
)
```

lanzhenw:
@coderabbitai review

app/packages/app/src/useEvents/utils.ts
```diff
@@ -49,12 +49,29 @@ export const processState = (
...
+    return map;
   }
 
-  return new Set<string>(state.selected || []);
+  // Fallback: flat selected list — all default type
+  const map = new Map<string, SelectionType>();
+  for (const id of state.selected || []) {
+    map.set(id, "default");
+  }
+  return map;
```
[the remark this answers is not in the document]
lanzhenw: good catch

app/packages/operators/src/built-in-operators.ts
```diff
@@ -37,7 +37,6 @@ import {
 } from "./operators";
 import { useShowOperatorIO } from "./state";
 import usePanelEvent from "./usePanelEvent";
-import { Clear } from "@mui/icons-material";
```
lanzhenw, line 40: It was never used. stray import

fiftyone/core/session/constants.py
```diff
@@ -0,0 +1,26 @@
...
+    "x",
+    "bookmark",
+}
+
+VALID_SELECTION_TYPES = {"default", "alt"}
+
+DEFAULT_SELECTION_STYLE = {
+    "default": "checkmark",
+    "alt": "checkmark",
+}
```
swheaton, line 26: Maybe this should be in the top level constants.py? I think it does not currently create issues that you import this into `fiftyone.operators`, but it seems a little odd because operators are not always run in the app context.
On the other hand, something like "checkmark" is a frontend concept so one could argue doesn't belong in the top level constants either...
Dealer's choice.

fiftyone/operators/executor.py
```diff
@@ -830,7 +831,34 @@ def spaces(self):
...
         """The list of selected sample IDs (if any)."""
-        return self.request_params.get("selected", [])
+        selected = self.request_params.get("selected", None)
+        if selected is not None:
+            return selected
+
+        return [
+            s["sample_id"] if isinstance(s, dict) else s
+            for s in self.selected_samples
+        ]
```
swheaton: I know the tests in the operators modules are lacking. But it would be nice to have a simple test that using "selected" gets what you want - either selected from request params or the alternate mode from "selected_samples"
The ones below are obvious so it's fine.

fiftyone/operators/executor.py
```diff
@@ -830,7 +831,34 @@ def spaces(self):
...
+
+    @property
+    def selected_samples(self):
+        """A list of selected sample dicts (if any).
+
+        Each dict has ``sample_id`` and ``type`` (``"default"`` or ``"alt"``).
+        """
+        return self.request_params.get("selected_samples", [])
```
swheaton: This will ensure we return an empty list even if this is in the request params for some reason
```
{
  "selected_samples": None
}
```
```suggestion
        return self.request_params.get("selected_samples") or []
```

fiftyone/operators/executor.py
```diff
@@ -830,7 +831,34 @@ def spaces(self):
...
+        return [
+            s["sample_id"] if isinstance(s, dict) else s
+            for s in self.selected_samples
+        ]
+
+    @property
+    def selected_samples(self):
+        """A list of selected sample dicts (if any).
+
+        Each dict has ``sample_id`` and ``type`` (``"default"`` or ``"alt"``).
```
swheaton: This doesn't have to be the only 2 types right? As long as all the types in the list are represented in `sample_selection_style`. It's just that the default sample selection style only has default and alt.

fiftyone/operators/operations.py
```diff
@@ -9,6 +9,9 @@
 import json
 
 from bson import json_util
+
+from fiftyone.core.session.constants import VALID_ICON_STYLES
+from fiftyone.core.session.session import _normalize_selected_samples
```
swheaton: Hmm again this is a gray area where I want to say this normalize selected samples helper should be somewhere more common, but this file is specifically for helping the plugin interact with the app plugin system.

swheaton (changes requested):
Only reviewed `fiftyone/operators/`. Few comments, nothing major. I only request changes for comment resolution

fiftyone/operators/executor.py
lanzhenw: That's right. I updated the docstring text to point to the actual source `sample_selection_style`, but also listed "default" and "alt" (these two would be fairly stable for some time).

fiftyone/operators/operations.py
lanzhenw: I call this helper in `session.py`, `operators/operations.py`, and `server/mutation.py`. I put it under `/session`, as it seems the central piece that connects all of them. Do you have a suggestion on a better place?

fiftyone/operators/executor.py
lanzhenw: Added the test

fiftyone/core/session/session.py
```diff
@@ -788,33 +796,97 @@ def selected(self) -> t.List[str]:
...
+            for s in (self._state.selected_samples or [])
+            if isinstance(s, dict)
+        }
+        samples = [
+            {"sample_id": sid, "type": existing.get(sid, "default")}
+            for sid in (sample_ids or [])
+        ]
+        samples = _normalize_selected_samples(samples)
+        self._state.selected_samples = samples
+        self._client.send_event(SelectSamples(samples=samples))
```
benjaminpkane:
```suggestion
        normalized = _normalize_selected_samples(samples)
        self._state.selected_samples = normalized
        self._client.send_event(SelectSamples(samples=normalized))
```

fiftyone/core/state.py
```diff
@@ -45,6 +46,9 @@ class StateDescription(etas.Serializable):
         sample_id (None): a :attr:`fiftyone.core.sample.Sample.id`
         selected (None): the list of currently selected samples
         selected_labels (None): the list of currently selected labels
+        selected_samples (None): a list of dicts with sample selection info
+        sample_selection_style (None): a dict mapping selection types to
+            icon style names (e.g. ``{"default": "checkmark", "alt": "thumbsdown"}``)
```
benjaminpkane: Should we remove `selected`? `selected_samples` is the new format, correct?

fiftyone/core/state.py
```diff
@@ -202,6 +234,8 @@ def from_dict(cls, d):
             sample_id=d.get("sample_id", None),
             selected=d.get("selected", []),
```
benjaminpkane, line 223: Another place where `selected` is still referenced

fiftyone/operators/executor.py
```diff
@@ -830,7 +831,35 @@ def spaces(self):
...
+            return selected
+
+        return [
+            s["sample_id"] if isinstance(s, dict) else s
+            for s in self.selected_samples
+        ]
+
+    @property
+    def selected_samples(self):
+        """A list of selected sample dicts (if any).
```
benjaminpkane: nit: for consistency with other docs
```suggestion
        """A list of selected sample dicts, if any.
```

fiftyone/operators/executor.py
```diff
@@ -830,7 +831,35 @@ def spaces(self):
...
+
+        Each dict has ``sample_id`` and ``type`` (``"default"`` or ``"alt"``),
+        where type corresponds to a key in :attr:`sample_selection_style`.
+        """
+        return self.request_params.get("selected_samples") or []
+
+    @property
+    def sample_selection_style(self):
+        """The current sample selection style config (if any).
```
benjaminpkane:
```suggestion
        """The current sample selection style config, if any.
```

fiftyone/operators/operations.py
benjaminpkane: Moving to `fiftyone.core.session.utils` and making it a public is an option, i.e. `normalize_selected_samples`

tests/unittests/operators/executor_tests.py
```diff
@@ -90,6 +90,52 @@ def test_execution_context(self):
...
+            operator_uri="test_op",
+            request_params={
+                "dataset_name": "test",
+                "selected": ["ccc"],
+                "selected_samples": [
+                    {"sample_id": "ccc", "type": "alt"},
+                ],
+            },
+        )
+        self.assertEqual(ctx.selected, ["ccc"])
```
benjaminpkane: This behavior is a bit surprising to me. If `selected` and `selected_samples` are provided can we raise an error?

tests/unittests/session_events_tests.py
```diff
@@ -46,3 +59,412 @@ def test_select_labels(self):
...
+
+    @drop_datasets
+    def test_selected_property_derives_from_selected_samples(self):
+        """selected is computed from selected_samples."""
+        state = StateDescription()
+        state.selected_samples = [
+            {"sample_id": "a" * 24, "type": "default"},
+            {"sample_id": "b" * 24, "type": "alt"},
+        ]
+        self.assertEqual(state.selected, ["a" * 24, "b" * 24])
```
benjaminpkane: `StateDescription` is not exposed to users for the most part. Is `state.selected` necessary to maintain? I'd be in favor of removing it

tests/unittests/session_events_tests.py
```diff
@@ -46,3 +59,412 @@ def test_select_labels(self):
...
+
+    @drop_datasets
+    def test_selected_property_empty(self):
+        """selected is [] when selected_samples is []."""
+        state = StateDescription()
+        self.assertEqual(state.selected, [])
```
benjaminpkane: If `state.selected` is removed this is unneeded

tests/unittests/session_events_tests.py
```diff
@@ -46,3 +59,412 @@ def test_select_labels(self):
...
+        state = StateDescription.from_dict(d)
+        self.assertEqual(state.selected_samples, [])
+        self.assertEqual(state.selected, [])
+
+    # -------------------------------------------------------------------
+    # Operations.set_selected_samples — backward & forward compat
+    # -------------------------------------------------------------------
+
+    @drop_datasets
+    def test_ops_set_selected_samples_with_id_list(self):
```
benjaminpkane: Is there an existing set of operators tests? It feels a little awkward adding these tests to `session_events_tests.py`

app/packages/app/src/useEvents/utils.ts
```diff
@@ -49,12 +50,30 @@ export const processState = (
     workspace,
   };
 };
-const resolveSelected = (state: { selected?: string[] }) => {
+const resolveSelected = (state: {
+  selected?: string[];
+  selected_samples?: Array<{ sample_id: string; type: SelectionType }>;
+}) => {
```
benjaminpkane, line 55: I think `selected_samples` should be kept, and `selected` can be removed. Maintaining both feels like extra overhead. Is there a reason we need to support both?

fiftyone/core/session/session.py
```diff
@@ -788,33 +796,97 @@ def selected(self) -> t.List[str]:
...
+        self._state.selected_samples = []
+        self._client.send_event(SelectSamples(samples=[]))
 
     def select_samples(
         self,
-        ids: t.Optional[t.Union[str, t.Iterable[str]]] = None,
+        ids: t.Optional[t.Union[str, t.Iterable]] = None,
         tags: t.Optional[t.Union[str, t.Iterable[str]]] = None,
     ) -> None:
-        """Selects the specified samples in the current view in the App,
```
benjaminpkane, line 804: Good eye

fiftyone/core/session/session.py
```diff
@@ -788,33 +796,97 @@ def selected(self) -> t.List[str]:
...
+            ids = [ids]
 
-        self._state.selected = list(ids)
-        self._client.send_event(SelectSamples(self._state.selected))
+        normalized = _normalize_selected_samples(list(ids))
+        self._state.selected_samples = normalized
+        self._client.send_event(SelectSamples(samples=normalized))
+
+    @property
+    def sample_selection_style(self) -> t.Dict:
```
benjaminpkane, line 870: I was surprised that this didn't have a setter. Could we add one? And also keep `set_sample_selection_style`

app/packages/relay/src/mutations/setSelected.ts
```diff
@@ -2,6 +2,12 @@ import { graphql } from "react-relay";
 
 import r from "../resolve";
 
+/**
+ * Backward-compatible mutation that accepts a flat list of sample IDs.
+ * All samples are set to "default" selection type.
+ *
+ * Prefer setSelectedSamples for new code that needs selection type support.
+ */
```
benjaminpkane, line 10: Good note

app/packages/looker/src/state.ts
```diff
@@ -192,6 +192,18 @@ export interface BaseOptions {
...
+    | "checkmark"
+    | "green-checkmark"
+    | "red-checkmark"
+    | "thumbsup"
+    | "thumbsdown"
+    | "pin"
+    | "star"
+    | "x"
+    | "bookmark"
+    | null;
```
benjaminpkane, line 213: 💯

app/packages/state/src/recoil/types.ts
```diff
@@ -1,6 +1,27 @@
...
+export type SelectionIconStyle =
+  | "checkmark"
+  | "green-checkmark"
+  | "red-checkmark"
+  | "thumbsup"
+  | "thumbsdown"
+  | "pin"
+  | "star"
+  | "x"
+  | "bookmark";
```
benjaminpkane, line 14: `@fiftyone/looker` exports `ImageOptions`, so the below also works
```suggestion
export type SelectionIconStyle = ImageOptions["selectionIcon"];
```

fiftyone/core/session/constants.py
```diff
@@ -0,0 +1,26 @@
...
+    "star",
+    "x",
+    "bookmark",
+}
+
+VALID_SELECTION_TYPES = {"default", "alt"}
+
+DEFAULT_SELECTION_STYLE = {
+    "default": "checkmark",
+    "alt": "checkmark",
```
benjaminpkane, line 25: I was surprised to see `checkmark` is `alt`'s default. This is ok for us?

benjaminpkane (commented):
Very comprehensive work. The comments are minor. Just two questions. Otherwise LGTM

Should we add a version of the summary when two different selections are in use? The below was printed by my session's `repr` when 1 `alt` and 1 `default` selection were chosen
```
Dataset:          cifar10
Media type:       image
Num samples:      60000
Selected samples: 2
Selected labels:  0
Session type:     remote
```

Should we remove the use of `selected` from `StateDescription` or are we worried about backward compatibility?

tests/unittests/operators/executor_tests.py
lanzhenw: the FE will send both for backward compatibility, raising an error would break the normal flow. but I updated how it is handled: we always derive `selected` from `selected_samples`, and fall back to `param.selected` when `param.selected_samples` is not available.

fiftyone/operators/operations.py
swheaton: no I guess I don't have a better suggestion but I like Ben's ^^

swheaton (approved):
Thanks for adding the tests!
I think Ben may have some open comments but it's good on the backend (minimal changes)

fiftyone/core/session/constants.py
lanzhenw: Yes, it's intentional, so that people who are not aware of alternative selection aren't surprised. When a panel requires alternative selection, it will set the selection accordingly.

benjaminpkane (approved):
Very thorough tests! Feature looks great! Adding groups carousel support too 🚀 

My notes for potential future work:

* The checkbox icon the modal remains fixed. I had thumbs up/down configured (see screenshot)
* `./e2e-pw` POM updates to support new selection modes. Maybe a smoke test too?

Image

lanzhenw:
Screenshot 2026-04-03 at 4 11 26 PM

fixed the sample selection. 
Will add a smoke test.

PR #7117, opened 2026-03-02
chore(github): constraints and dependabot support

afoley587:

We are proposing we begin to constrain packages at the major level to make sure we test new versions before they get released. This begins to move us to a world where we can be confident in our dependencies and our code and begin testing various versions across various ecosystems. We have been bit by unexpected changes, especially among the `0.x` versions, in the past.

We are also proposing that we use dependabot to automate the widening of these constraints. These will submit a PR to our codebase which will run tests and, if they pass, we can merge.

0.x libraries are treated differently as we have seen issues with those before.

> [!NOTE]
> This does not introduce lock files at this point. My homework on `pip lock` seems to show that it'll generate a lockfile for the current OS only. If we wanted to migrate over to `uv` or `poetry`, I am definitely in favor of those for reproducible builds (lock files aren't packaged with wheels so won't impact people downloading fo with `pip install fiftyone`).

Dependabot was tested in private repo with various `setup.py` configurations

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Chores**
  * Added a Dependabot configuration for weekly pip updates with PR limits, a grouped stable package set, and ignore rules for selected packages.
  * Switched install dependency handling from dynamic resolution to an explicit, statically declared dependency list for more predictable installs.
  * Added support to override the packaged version via an environment variable during packaging.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,7 +16,6 @@ updates:
   # ---------------------------------------------------------------------------
   - package-ecosystem: "pip"
     directory: "/"
-    registries: '*'
     schedule:
       interval: "weekly"
       day: "wednesday"


.github/dependabot.yml
```diff
@@ -5,3 +5,63 @@ updates:
...
+    directory: "/"
+    registries: '*'
+    schedule:
+      interval: "weekly"
+      day: "wednesday"
+      time: "14:00"
+      timezone: "UTC"
+    open-pull-requests-limit: 5
+    # Dependabot needs to evaluate setup.py to read install_requires
+    insecure-external-code-execution: "allow"
```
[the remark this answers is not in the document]
afoley587: Removed.

setup.py
```diff
@@ -136,9 +46,60 @@ def get_install_requirements(install_requires, choose_install_requires):
...
         exclude=["app", "eta", "package", "requirements", "tests", "tools"]
     ),
-    install_requires=get_install_requirements(
-        INSTALL_REQUIRES, CHOOSE_INSTALL_REQUIRES
-    ),
+    install_requires=[
+        # python-packaged or meta libraries (unconstrained)
+        "Deprecated", # Don't constrain python-packaged `Deprecated`
+        "packaging",  # Don't constrain python-packaged `packaging`
+        "setuptools", # Don't constrain python-packaged `packaging`
```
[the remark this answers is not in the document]
afoley587: Fixed.

.github/dependabot.yml
```diff
@@ -5,3 +5,61 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  # ---------------------------------------------------------------------------
+  # Fiftyone Python Dependabot configuration
```
benjaminpkane:
```suggestion
  # FiftyOne Python Dependabot configuration
```

setup.py
```diff
@@ -136,9 +46,60 @@ def get_install_requirements(install_requires, choose_install_requires):
...
+        "universal-analytics-python3>=1.0.1,<2",
+        # ML Libraries (unconstrained)
+        "matplotlib",
+        "numpy",
+        "opencv-python-headless",
+        "pandas",
+        "rtree",
+        "scikit-learn",
+        "scikit-image",
+        "scipy",
```
jleven: @afoley587 -- after further discussion, we are on board with putting major version upper bounds on ML libraries, so long as our dependabot automations enable us to be responsive to any new major versions quickly -- which I believe they will.

jleven:
```suggestion
        # ML Libraries
        "matplotlib<4",
        "numpy<3",
        "opencv-python-headless<5",
        "pandas<4",
        "rtree<2",
        "scikit-learn<2",
        "scikit-image<1",
        "scipy<2",
```

.github/dependabot.yml
```diff
@@ -5,3 +5,61 @@ updates:
...
+    versioning-strategy: "auto"
+    ignore:
+      # Ignore patches for all bumps.
+      # All of the stable packages in setup.py
+      # are constrained to a major, so DB won't bump them
+      # unless a new major comes out.
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+      # Only do minor versions for these 0.x packages to avoid making 
+      # the 0.x -> 1.0 bump.
```
jleven: How will we know when a major version release does happen for one of these `0.x` libraries? For example, `starlette 1.0` is coming soon!

afoley587: We could remove these `ignores` and dependabot will continue to bump libraries 0.x version by 0.x version. When the 1.x is hit, it would submit a PR and we could test a bit more thoroughly. If we like that, I can remove the majority of this `ignore` section.

jleven: Yeah, I think that makes sense to me

afoley587: I removed the `ignores` for `0.x` packages!

PR #7122, opened 2026-03-03
Refactor SAM and SAM2 image model integration

manushreegangwar:

This PR refactors SAM and SAM2 integration.

- [x] Update SAM and SAM2 to use `GetItem`
- [x] Add combination prompts (box plus points)

1. Compared model outputs on this branch against outputs on `develop`
- Auto segmentation against auto segmentation :heavy_check_mark: 
- Box prompted segmentation against box prompted segmentation :heavy_check_mark: 
- Point prompted segmentation against point prompted segmentation -- **_minor error due to floating point differences between using predictor.predict vs model.forward in some cases_**

2. Compare model outputs on this branch against outputs using `segment-anything` and `sam2` inference and add parity tests -- WIP

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added SAM/SAM2 integration with explicit prompt modes (auto/box/point/combined), configurable prompt field mappings, unified prompt validation, and improved mask→detection output processing (confidence thresholds, class filtering, auto-generation).

* **Refactor**
  * Consolidated predictor/auto-generator initialization and inference flows; unified input pipeline and centralized output post-processing for more consistent predictions.

* **Tests**
  * Removed the SAM2 unit test suite.

### DIFF
--- a/fiftyone/utils/sam.py
+++ b/fiftyone/utils/sam.py
@@ -182,6 +182,7 @@ class SegmentAnythingImageGetItem(fout.GetItem):
         point_transform=None,
         **kwargs,
     ):
+        field_mapping = {} if field_mapping is None else dict(field_mapping)
         # Used for managing backward compatibility with prompt_field in field mapping.
         self._has_prompt_field = (
             field_mapping.pop("prompt_field", None) is not None
@@ -709,7 +710,7 @@ class SegmentAnythingModel(fout.TorchImageModel):
         """
         if samples is not None:
             raise RuntimeError(
-                "Use of SamplesMixin has been deprecated."
+                "Use of SamplesMixin has been deprecated. "
                 "Use SegmentAnythingGetItem to get inputs for the predict_all method."
             )
 
@@ -830,6 +831,11 @@ class SegmentAnythingModel(fout.TorchImageModel):
                     results[key].append(val)
 
         # Collapse prompt type
+        prompt_types = results["prompt_type"]
+        if not all(pt == prompt_types[0] for pt in prompt_types):
+            raise ValueError(
+                "All samples in a batch must have the same prompt_type"
+            )
         results["prompt_type"] = results["prompt_type"][0]
 
         return results


fiftyone/utils/sam.py
```diff
@@ -40,14 +44,347 @@ def __init__(self, d):
...
+        # For each point prompt, three masks are generated. Mask index is used to select
+        # which one of the three masks to choose. When not provided, mask with the
+        # highest score is chosen.
         self.points_mask_index = self.parse_int(
             d, "points_mask_index", default=None
         )
         if self.points_mask_index and not 0 <= self.points_mask_index <= 2:
             raise ValueError("mask_index must be 0, 1, or 2")
 
+        self.get_item_cls = self.parse_int(
```
jacobsela: should this be configurable?

fiftyone/utils/sam.py
```diff
@@ -40,14 +44,347 @@ def __init__(self, d):
...
+        use_numpy (False): whether to use numpy arrays rather than PIL images
+            and Torch tensors when loading data
+        box_transform (None): SAM specific box transform function to apply
+        validate_prompts (False): Whether to validate combination prompts
+    """
+
+    def __init__(
+        self,
+        field_mapping=None,
+        transform=None,
```
jacobsela, line 179: From [here](https://github.com/voxel51/fiftyone/pull/7122/changes#diff-92544b2f1a47fa2f592a393088274ed7b2312304eea461545d176e086967d3f9R534) it seems that `transform`, `use_numpy`, and `box_transform` shouldn't be configurable. Is this the case?

fiftyone/utils/sam.py
manushreegangwar: In theory, if you want to write your own GetItem that doesn't extract prompts from Detections and Keypoints, you should be able to use that without overriding `SegmentAnythingModel.build_get_item`. Also, if other SAM (2 and 3) models need a new GetItem class, I can make a config only change.

fiftyone/utils/sam.py
manushreegangwar: True. You should use the transforms from the SAM predictor. I can enable setting them via `get_item_args` to make get_item more configurable

fiftyone/utils/sam.py
```diff
@@ -122,205 +459,146 @@ class SegmentAnythingModel(fout.TorchSamplesMixin, fout.TorchImageModel):
...
-        for sample in samples:
-            value = sample.get_field(field_name)
-            if value is not None:
-                prompts.append(value)
-            else:
+        get_item = etau.get_class(get_item_cls)
+        get_item_args = self.config.get_item_args or {}
+        field_mapping = {} if field_mapping is None else field_mapping
+
+        if "prompt_field" in field_mapping:
```
manushreegangwar, line 740: @brimoor What are your thoughts on mainintaing this backward compatibility?

brimoor: I personally like the simplicity of this syntax, which is [currently documented in the model zoo](https://docs.voxel51.com/model_zoo/models/segment_anything_2_hiera_base_plus_image_torch.html):

```py
# Segment inside boxes
dataset.apply_model(
    model,
    label_field="segmentations",
    prompt_field="ground_truth",  # can contain Detections or Keypoints
)
```

brimoor: So my two cents is that I'd prefer to not only retain that syntax (which you've already done), but I would advocate for not raising warnings telling the user that they should change their behavior.

Isn't providing both a box and points more of a niche use case?

manushreegangwar: Okay, I can maintain this syntax and remove the warning.

One thing to note is that being explicit with the type of prompts will reduce some checks in the implementation and these checks will grow as we support more types of prompts. I am also planning to add mask prompts and with SAM3, we will also have exemplar prompts on the Sample.

manushreegangwar: `Isn't providing both a box and points more of a niche use case?` 

When using `apply_model`, I agree that it is a niche use case. For one sample inference, I can see it being used more often.

fiftyone/utils/sam.py
lorenzoriano: Shouldn't this be parse_string?

fiftyone/utils/sam.py
```diff
@@ -40,14 +44,404 @@ def __init__(self, d):
...
 
+        self.get_item_cls = self.parse_int(
+            d,
+            "get_item_cls",
+            default="fiftyone.utils.sam.SegmentAnythingImageGetItem",
+        )
+        self.get_item_args = self.parse_dict(d, "get_item_args", default=None)
+
+
+class _SAMPredictor:
```
lorenzoriano, line 69: docstrings, including the methods below

fiftyone/utils/sam.py
```diff
@@ -40,14 +44,404 @@ def __init__(self, d):
...
+        self,
+        field_mapping=None,
+        transform=None,
+        use_numpy=None,
+        box_transform=None,
+        point_transform=None,
+        validate_prompts=False,
+        **kwargs,
+    ):
+        self.mode = self._set_mode(field_mapping)
```
lorenzoriano, line 191: I'm not sure I like the idea of setting the mode using field_mapping. It feels like abusing a dict that has a different purpose. What about a specific config or runtime parameter here, and GetItem reads it but does't have to infer

fiftyone/utils/sam.py
```diff
@@ -40,14 +44,404 @@ def __init__(self, d):
...
+    def _set_mode(self, field_mapping):
+        prompt_fields = ["box_prompt_field", "point_prompt_field"]
+        is_present = {f: False for f in prompt_fields}
+
+        for key in field_mapping:
+            if key in prompt_fields:
+                is_present[key] = True
+
+        if not any(is_present.values()):
+            return SAMPromptMode(1)  # auto
```
lorenzoriano: why not SAMPromptModel.auto (and the same for box and point below)?

fiftyone/utils/sam.py
```diff
@@ -40,14 +44,404 @@ def __init__(self, d):
...
+        self.mode = self._set_mode(field_mapping)
+
+        super().__init__(field_mapping=field_mapping, **kwargs)
+        self.transform = transform
+        self.use_numpy = use_numpy
+        self.box_transform = box_transform
+        self.point_transform = point_transform
+        self.validate_prompts = validate_prompts
+
+    def _set_mode(self, field_mapping):
```
lorenzoriano, line 199: Honestly, if you set the explicit parametwer (see my other comment), then in GetItem you can do something like:

```
      @property
      def required_keys(self):
          common_keys = ["id", "filepath"]

          if self.prompt_mode == SAMPromptMode.auto:
              return common_keys

          if self.prompt_mode == SAMPromptMode.box_only:
              return common_keys + ["box_prompt_field"]

          if self.prompt_mode == SAMPromptMode.point_only:
              return common_keys + ["point_prompt_field"]

          if self.prompt_mode == SAMPromptMode.box_point_combo:
              return common_keys + ["box_prompt_field", "point_prompt_field"]
```

which is a lot cleaner

fiftyone/utils/sam2.py
```diff
@@ -440,10 +305,6 @@ def predict(self, video_reader, sample):
         ) = video_reader.frame_size
```
lorenzoriano, line 360: <p class='graphite__hidden'><i>[Re: line +301]</i></p>
Is negative_field_name being used?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/voxel51/fiftyone/7122?utm_source=unchanged-line-comment">Graphite</a>.</p>

fiftyone/utils/sam.py
```diff
@@ -122,274 +516,235 @@ class SegmentAnythingModel(fout.TorchSamplesMixin, fout.TorchImageModel):
...
+            outputs.append(out)
         return outputs
 
-    def _forward_pass_points(self, imgs):
-        sam_predictor = self._load_predictor()
-        self._output_processor = fout.InstanceSegmenterOutputProcessor(
-            self._curr_classes
+    def _forward_pass(self, arg):
+        self._sam_predictor.set_image(
+            arg["image"],
```
lorenzoriano: I think this doesn't reflect the TorchImageModel contract if you read arg["image"]: TorchImageModel.predict() still forwards raw images into _predict_all()

lorenzoriano (changes requested):

fiftyone/utils/sam.py
jacobsela: I think that since the GI is already configurable, we don't have to expose them at all. If the user wants to override them for whatever reason they can plug in a different GI.

fiftyone/utils/sam.py
jacobsela: It can be used to propagate a single human labeled detection on each sample to all of the data in a large scale batch setting, arguably it is quite important.

fiftyone/utils/sam.py
```diff
@@ -40,14 +44,411 @@ def __init__(self, d):
...
+    def _parse_output(self, output, frame_wh, confidence_thresh, classes):
+        width, height = frame_wh
+
+        boxes = output["boxes"]
+        labels = output["labels"]
+        masks = output["masks"]
+        scores = output["scores"]
+
+        if isinstance(masks, torch.Tensor):
+            masks = masks.detach().cpu().numpy()
```
jacobsela: make the gpu --> cpu transfers non-blocking

fiftyone/utils/sam.py
```diff
@@ -40,14 +44,411 @@ def __init__(self, d):
...
+        if isinstance(scores, torch.Tensor):
+            scores = scores.detach().cpu().numpy()
+
+        boxes = [None] * len(masks) if len(boxes) == 0 else boxes
+        detections = []
+
+        for box, label, mask, score in zip(boxes, labels, masks, scores):
+            if isinstance(mask, torch.Tensor):
+                score = score.detach().cpu().numpy()
```
jacobsela: here as well (this is especially important for loops like this)

fiftyone/utils/sam.py
jacobsela, line 39: please add documentation for any new arguments.

fiftyone/utils/sam.py
```diff
@@ -122,274 +523,235 @@ class SegmentAnythingModel(fout.TorchSamplesMixin, fout.TorchImageModel):
...
-
-            boxes = np.array([d.bounding_box for d in detections.detections])
-            boxes_xyxy = _to_abs_boxes(boxes, w, h)
-            sam_boxes = np.round(boxes_xyxy).astype(int)
-            input_boxes = torch.tensor(sam_boxes, device=sam_predictor.device)
-            transformed_boxes = sam_predictor.transform.apply_boxes_torch(
-                input_boxes, (h, w)
-            )
+    def _predict_all(self, args):
+        if self.config.raw_inputs is False:
```
jacobsela: what is this?

fiftyone/utils/sam.py
```diff
@@ -122,274 +523,235 @@ class SegmentAnythingModel(fout.TorchSamplesMixin, fout.TorchImageModel):
...
-            outputs.append(
-                {
-                    "boxes": torch.tensor(boxes_xyxy),
-                    "labels": labels,
-                    "masks": masks,
-                    "scores": scores,
-                }
-            )
+        outputs = []
+        for arg in args:
```
jacobsela: this isn't batched inference.

fiftyone/utils/sam.py
```diff
@@ -122,274 +523,235 @@ class SegmentAnythingModel(fout.TorchSamplesMixin, fout.TorchImageModel):
...
+                    "boxes": boxes,
+                    "multimask_output": False,
+                },
+            )
+            out_boxes = arg["boxes_xyxy"]
+
+        elif prompt_type == "point":
+            # Each point prompt has varying number of points.
+            out_boxes, out_scores, out_masks = [], [], []
+            for points, labels in zip(point_coords, point_labels):
```
jacobsela: same here, no for loops in batched inference.

fiftyone/utils/sam.py
manushreegangwar: I completely agree with the sentiment about using `field_mapping` for setting the mode. However, I also don't like the alternative of expecting a user to set the mode and also provide the prompts. 

Here's my take:

- With the current implementation, we are using `field_mapping` to set the mode, which in turn decides what the `required_keys` should be. This is circular since `required_keys` decide whether a `field_mapping` has valid keys.

- If we set an explicit `mode` parameter, it solves the problem of relying on `field_mapping` for setting the keys. Great but now we are introducing a new parameter that is going to break the exisiting `dataset.apply_model` for sam models unless we make `mode` optional. If we make `mode` optional, we still need a way to figure out what mode to operate in. The option is to go back to using `field_mapping` unless we add some logic in `apply_model` to handle this. 

I still think [optional keys](https://github.com/voxel51/fiftyone/blob/34a1076ba76ea9e41eb81741ffebe59c33bf4a9c/fiftyone/utils/sam.py#L267) are worth considering but this requires changes to `GetItem`

fiftyone/utils/sam.py
manushreegangwar: See https://github.com/voxel51/fiftyone/pull/7122/changes#r2897145337

fiftyone/utils/sam.py
manushreegangwar: We can't stack torch image tensors for inference. SAM processes one image and the respective prompts at a time.

fiftyone/utils/sam.py
manushreegangwar: Yes. SAMPredictor can only process one image at a time. How do you envision batching to work in SAM/SAM2? 

https://github.com/facebookresearch/sam2/blob/main/sam2/sam2_image_predictor.py#L175 is not really batched inference

fiftyone/utils/sam.py
manushreegangwar: We can't batch point prompts because each prompt can have varying number of points.

For the first instance prompt, we have 8 points and for the second prompt, we can have 2, etc.

fiftyone/utils/sam.py
manushreegangwar: I can enable bypassing `predict` in `SAMPredictor` for `apply_model` and use SAM's forward pass. For interactivity, we can preserve `SAMImagePredictor.predict`.  How does that sound?

fiftyone/utils/sam.py
```diff
@@ -122,274 +658,323 @@ class SegmentAnythingModel(fout.TorchSamplesMixin, fout.TorchImageModel):
...
+
+                    results[key].append(torch.stack(padded_points))
+                    results["point_labels"].append(torch.stack(padded_labels))
+                else:
+                    results[key].append(val)
+
+        # Collapse prompt type
+        results["prompt_type"] = results["prompt_type"][0]
+
+        return results
```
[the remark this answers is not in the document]
manushreegangwar: Mixed batches are not supported. Added a check

manushreegangwar:
Closing this in favor of graphite stack https://app.graphite.com/submit/voxel51/fiftyone/7179

PR #7124, opened 2026-03-04
feat: allow for using the SAM2 video predictor on flattened image datasets

neerajaabhyankar:

1. [minor] Ability to handle mask prompts. Currently, when the prompt field has detections, only bounding boxes are read. SAM2 uses its own segmentation to find masks within it. With this commit, it also takes the mask from the detection into consideration via the `add_new_mask` API call.
2. [related to the above; changes existing behavior] In the video propagator, if only bounding box prompts are provided, only bounding boxes are returned. Masks are only returned if prompts contain them.
3. [major] Ability to pass a flattened image dataset into the `SegmentAnything2VideoModel`. This is [supported by the original SAM2 code already](https://github.com/facebookresearch/sam2/blob/2b90b9f5ceec907a1c18123530e92e794ad901a4/sam2/utils/misc.py#L172).

## JIRA

https://voxel51.atlassian.net/browse/ML-356

- https://github.com/voxel51/fiftyone/pull/7143 for videos
- Added a new test `test_sam2_video_flattened` for images

        notes for FiftyOne users.

SAM2 models now support flattened image datasets for propagating labels across video frames. Users can specify `media_mode="image"` while loading the video model from the zoo. An adequate batch size -- larger than the set of images to propagate to -- must be passed when applying the model.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * SegmentAnything2 gains configurable media mode (video or image), frame-based batching, batch-size support, and improved handling for per-frame prompts and masks.
  * New helpers to load video frames from video files or image-sequence folders and automatic dispatch based on media type.
  * Support for converting box-relative masks into full-image masks.

* **Tests**
  * Added tests for flattened video workflows and frame-based processing to validate image/video inference paths.

### DIFF
--- a/fiftyone/utils/sam2.py
+++ b/fiftyone/utils/sam2.py
@@ -522,7 +522,25 @@ class SegmentAnything2VideoModel(fom.SamplesMixin, fom.Model):
 
         # The existing video path will handle
         # prompt extraction, registration, propagation, and detection construction
-        sample_detections = self.predict(mock_reader, mock_video_sample)
+        try:
+            sample_detections = self.predict(mock_reader, mock_video_sample)
+        except Exception as e:
+            if "mat1 and mat2 must have the same dtype" in str(e):
+                """
+                ------------------------------------------------------------
+                This error typically occurs due to a mixed-precision dtype mismatch in SAM2's
+                internal transformer layers. This is a known issue in SAM2's implementation
+                that is more likely to occur when propagating across
+                multiple sequences with different label sets, as this triggers
+                more complex memory attention paths in SAM2's tracking stack.
+                ------------------------------------------------------------
+                """
+                raise RuntimeError(
+                    f"Error propagating to all frames; \
+                please try with a shorter sequence with continuous frames and consistent labels."
+                )
+            else:
+                raise e
 
         return [
             sample_detections.get(i + 1, fol.Detections())
--- a/fiftyone/utils/sam2.py
+++ b/fiftyone/utils/sam2.py
@@ -851,7 +851,7 @@ def load_fiftyone_video_frames(
     img_mean=(0.485, 0.456, 0.406),
     img_std=(0.229, 0.224, 0.225),
     async_loading_frames=False,
-    compute_device=torch.device("cuda"),
+    compute_device=None,
 ):
     """
     The signature of this function matches
@@ -875,7 +875,7 @@ def load_fiftyone_video_frames(
             image_size=image_size,
             img_mean=img_mean,
             img_std=img_std,
-            async_loading_frames=async_loading_frames,
+            async_loading_frames=False,
         )
     else:
         raise NotImplementedError("Unsupported media type")
@@ -888,8 +888,11 @@ def load_fiftyone_video_frames_from_video_file(
     offload_video_to_cpu,
     img_mean=(0.485, 0.456, 0.406),
     img_std=(0.229, 0.224, 0.225),
-    compute_device=torch.device("cuda"),
+    compute_device=None,
 ):
+    if compute_device is None:
+        compute_device = "cuda:0" if torch.cuda.is_available() else "cpu"
+
     img_mean = torch.tensor(img_mean, dtype=torch.float32)[:, None, None]
     img_std = torch.tensor(img_std, dtype=torch.float32)[:, None, None]
 
--- a/fiftyone/utils/sam2.py
+++ b/fiftyone/utils/sam2.py
@@ -8,9 +8,7 @@ wrapper for the FiftyOne Model Zoo.
 """
 
 import logging
-import contextlib
 import os
-import shutil
 import tempfile
 from typing import List, Optional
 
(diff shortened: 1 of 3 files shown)


fiftyone/utils/sam2.py
```diff
@@ -417,7 +431,12 @@ def __init__(self, config):
 
     @property
     def media_type(self):
-        return "video"
+        return self.media_mode
+
+    @property
+    def ragged_batches(self):
+        # Frames are resized to a fixed size, so batching is safe
+        return False
```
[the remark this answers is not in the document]
neerajaabhyankar: This path will be hit when we load the model with `media_mode="image"` but try to apply it to a video dataset. Given that the default mode of the model is video, I guess it is okay to have an error raised in this case?

tests/intensive/model_zoo_tests.py
```diff
@@ -116,6 +116,17 @@ def test_sam2_video():
     )
 
 
+def test_sam2_video_flattened():
+    models = ["segment-anything-2-hiera-tiny-video-torch"]
+    _apply_video_models_on_flattened_dataset(
+        models,
+        max_samples=1,
+        max_frames=3,
+        batch_size=8,  # must be >= the number of frames
```
neerajaabhyankar, line 125: Not sure what's the best way to warn in the case where `batch_size` < frames.
`batch_size` is consumed by the `apply_model` API and may not be visible from inside `sam2.py`, where we only see batches yielded by the dataloader.

fiftyone/utils/sam2.py
```diff
@@ -116,6 +120,9 @@ def __init__(self, d):
         d = self.init(d)
         super().__init__(d)
 
+        self.batch_size = self.parse_int(d, "batch_size", default=None)
+        self.media_mode = self.parse_string(d, "media_mode", default="video")
```
manushreegangwar: Can you add what these additional args are for in the docstring?

fiftyone/utils/sam2.py
```diff
@@ -399,14 +410,17 @@ def __init__(self, config):
...
-                "Failed to monkey patch sam2.utils.misc.load_vide_frames: %s",
+            logger.error(
+                "Failed to monkey patch sam2.utils.misc.load_video_frames: %s",
                 e,
             )
 
         self.model = self._load_model(config)
+        self.media_mode = (
+            getattr(config, "media_mode", None) or media_mode or "video"
+        )
```
manushreegangwar: Why the need for the input argument `media_mode` when it has been added to the config?

fiftyone/utils/sam2.py
```diff
@@ -432,7 +451,7 @@ def _load_model(self, config):
             )
         return model
 
-    def predict(self, video_reader, sample):
+    def predict(self, video_reader, sample=None):
```
manushreegangwar: Handle `sample=None`. Add a check / raise an error.

fiftyone/utils/sam2.py
```diff
@@ -592,7 +727,9 @@ def _forward_pass_boxes(self, video_reader, sample):
                     fol.Detection(
                         label=label,
                         bounding_box=bounding_box,
-                        mask=mask,
+                        mask=mask
+                        if self._curr_prompt_type == "masks"
+                        else None,
```
manushreegangwar, line 725: Why is this not populating the `mask` for box prompts?

fiftyone/utils/sam2.py
```diff
@@ -441,39 +460,126 @@ def predict(self, video_reader, sample):
...
+                has_prompt = True
+                break
+
+        if not has_prompt:
+            return [fol.Detections() for _ in samples]
+
+        # We construct a lightweight container that mimics a video sample for flattened frames
+        # The sample has a `frames` dict and delegates to
+        # the existing video path via `_forward_pass_boxes`.
+        class _MockVideoSample(object):
```
manushreegangwar: This seems rather hacky to me

manushreegangwar (changes requested):
In `apply_model`, there is currently no support for applying video model to image frames which this PR is doing but implicitly. Let's think about how we can re-design the SAM video model integration to do this in a cleaner manner.

fiftyone/utils/sam2.py
neerajaabhyankar: You're right; was initially trying out an override mechanism but forgot to clean it up

fiftyone/utils/sam2.py
neerajaabhyankar: For the label propagation case, it seemed a bit non-intuitive to me if the model returns masks when I have only prompted boxes.
- I'm not sure how easy it is in the HA workflow to delete just the mask but keep the box, if the user wants that.
- The other way round seems definitely doable: first run click2seg on just one image, and have all masks populated in propagations. Although correct me if you feel otherwise!

I'm open to returning masks by default, but having a config param where I can disable masks in such cases.

fiftyone/utils/sam2.py
neerajaabhyankar: Digging into this a bit, I didn't see any pathway via the `apply_model` calls that expose the case where sample=None; I've removed the default for now. Any errors could be propagated as-is, since I'm not certain what interpretable exception we can raise here.
Let me know if there's a test that would fail.

fiftyone/utils/sam2.py
neerajaabhyankar: I agree. Tried simplifying it a bit and renaming to sound less hacky 😄 but open to thoughts on doing it differently.

PR #7129, opened 2026-03-05
Enable uv for install

gboeer:
The install script `install.sh` previously hardcoded `python -m pip` as the package installer, which can be incompatible with [uv](https://github.com/astral-sh/uv)-managed environments where the correct invocation is uv pip. This change introduces an optional -u flag that switches the pip backend to uv pip for the duration of the install. When -u is passed, the script validates that uv is available in PATH and exits early with a clear error if not. All existing install paths and flags are unaffected, the default behavior remains python -m pip, so no changes are required for users not using uv.

Edit:
By accident this PR includes the (small) changes from #7127 as well

## Summary by CodeRabbit

* **New Features**
  * Installer adds a new flag for extended installation configuration options.
  * Installer now detects and exposes an alternate Python package backend when available and uses it for package installation.

* **Improvements**
  * Improved installer compatibility with robust fallback behavior across different environments.
  * Minor user-facing message tweaks for clearer output.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -38,12 +38,25 @@ SHIFT
 GOTO parse
 :endparse
 
+:: Resolve pip backend
+where uv >nul 2>&1
+IF NOT ERRORLEVEL 1 (
+  set PIP=uv pip
+) else (
+  where pip >nul 2>&1
+  IF ERRORLEVEL 1 (
+    echo ERROR: Neither 'uv' nor 'pip' found in PATH. Please install one before running this script.
+    exit /b 1
+  )
+  set PIP=pip
+)
+
 :: Do this first so pip installs with a built app
 if %BUILD_APP%==true (
   echo ***** INSTALLING FIFTYONE-APP *****
   :: TODO - Add nvm and yarn installs
   cd app
-  echo "Building the App. This will take a minute or two..."
+  echo Building the App. This will take a minute or two...
   call yarn install > nul 2>&1
   call yarn build:win32
   cd ..
@@ -51,7 +64,7 @@ if %BUILD_APP%==true (
 
 IF %USE_FIFTY_ONE_DB%==true (
   echo ***** INSTALLING FIFTYONE-DB *****
-  pip install fiftyone-db
+  %PIP% install fiftyone-db
 ) else (
   echo ***** USING LOCAL MONGODB *****
 )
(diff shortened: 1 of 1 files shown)


install.sh
Burhan-Q, line 1: I too am a big fan of `uv`. I actually had intended to open a PR today to introduce the use of `uv` for the install script as well. I'll share my change with you to consider using since I feel it's much simpler.

install.sh
```diff
@@ -79,7 +82,15 @@ fi
...
-PIP="$PYTHON -m pip"
+if [ "$USE_UV" = true ]; then
+    if ! command -v uv >/dev/null 2>&1; then
+        echo "ERROR: -u flag given but 'uv' not found in PATH."
+        exit 1
+    fi
+    PIP="uv pip"
+else
+    PIP="$PYTHON -m pip"
+fi
```
Burhan-Q:
```suggestion
if command -v uv >/dev/null 2>&1; then
    PIP="uv pip"
else
    PIP="$PYTHON -m pip"
fi
```

Burhan-Q: No flag needed, and if user has `uv` installed, it's automatically used, otherwise falls back to `pip`. I have tested this on my Linux PC. If using this route, then all uses of `$USE_UV` would need to be removed.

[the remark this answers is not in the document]
Burhan-Q: I agree that this change should be reflected in both the `install.bat` and the `install.sh` for parity. I think for the modification I proposed, the equivalent in the `.bat` would be:

```bat
where uv >nul 2>&1
if %ERRORLEVEL% == 0 (
    set PIP=uv pip
) else (
    set PIP=pip
)
```

Then updating all locations to use the `%PIP%` variable.

> [!CAUTION]
> Not tested

gboeer: Yes I see. I only kept it explicit with  the -u switch, in case anybody wants to use standard pip even if uv is installed systemwide.

gboeer: I thinks its an edge case though and maybe a bit over cautious. Indeed, who ever wants to switch back after working with uv ;)

gboeer: > No flag needed, and if user has `uv` installed, it's automatically used, otherwise falls back to `pip`. I have tested this on my Linux PC. If using this route, then all uses of `$USE_UV` would need to be removed.

I removed the explicit switch again for the implicit version now, assuming anyone who has uv installed actually wants to use it for the install.

I tested this with the normal install and developer install and it completed successfull in both cases.

install.bat
```diff
@@ -38,20 +38,33 @@ SHIFT
...
+IF NOT ERRORLEVEL 1 (
+  set PIP=uv pip
+) else (
+  where pip >nul 2>&1
+  IF ERRORLEVEL 1 (
+    echo ERROR: Neither 'uv' nor 'pip' found in PATH. Please install one before running this script.
+    exit /b 1
+  )
+  set PIP=pip
+)
```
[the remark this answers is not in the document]
gboeer: No, we have agreed to implicitly favor UV.

PR #7135, opened 2026-03-06
Add RF-DETR detection and segmentation models to zoo

CharlesCNorton:
## Summary

- Add RF-DETR (Roboflow, ICLR 2026) detection and segmentation models to the model zoo
- New `fiftyone/utils/rfdetr.py` wrapper around the `rfdetr` package
- 11 models total: 5 detection + 6 instance segmentation, all COCO-pretrained

### Detection
| Model | Params | COCO AP₅₀:₉₅ | Resolution |
|-------|--------|---------------|------------|
| `rfdetr-nano-coco-torch` | 30.5M | 48.4 | 384 |
| `rfdetr-small-coco-torch` | 32.1M | 53.0 | 512 |
| `rfdetr-medium-coco-torch` | 33.7M | 54.7 | 576 |
| `rfdetr-base-coco-torch` | 29.0M | 48.8 | 560 |
| `rfdetr-large-coco-torch` | 33.9M | 56.5 | 704 |

### Instance Segmentation
| Model | Params | COCO AP₅₀:₉₅ | Resolution |
|-------|--------|---------------|------------|
| `rfdetr-seg-nano-coco-torch` | 33.6M | 40.3 | 312 |
| `rfdetr-seg-small-coco-torch` | 33.7M | 43.1 | 384 |
| `rfdetr-seg-medium-coco-torch` | 35.7M | 45.3 | 432 |
| `rfdetr-seg-large-coco-torch` | 36.2M | 47.1 | 504 |
| `rfdetr-seg-xlarge-coco-torch` | 38.1M | 48.8 | 624 |
| `rfdetr-seg-2xlarge-coco-torch` | 38.6M | 49.9 | 768 |

## Usage
```python
import fiftyone as fo
import fiftyone.zoo as foz

# Detection
model = foz.load_zoo_model("rfdetr-medium-coco-torch")
dataset.apply_model(model, label_field="rfdetr")

# Instance segmentation
seg_model = foz.load_zoo_model("rfdetr-seg-medium-coco-torch")
dataset.apply_model(seg_model, label_field="rfdetr_seg")

```

## Notes

RF-DETR is not yet in HuggingFace `transformers` (PR pending), so this integration
wraps the `rfdetr` pip package directly rather than using the existing transformers
integration path. The wrapper converts `supervision.Detections` output to FiftyOne
`Detections` with normalized bounding boxes and cropped instance masks.

## Test plan

- [x] `foz.load_zoo_model()` loads all detection variants
- [x] `foz.load_zoo_model()` loads all segmentation variants
- [x] `dataset.apply_model()` produces `Detections` with labels, confidence, bounding boxes
- [x] Segmentation models produce instance masks cropped to bounding boxes
- [x] All 11 manifest entries resolve correctly
- [ ] Extended testing across all model sizes in progress

## Summary by CodeRabbit

* **New Features**
  * RF-DETR integration for detection and segmentation with public model configs and wrappers.
  * Accepts many image input formats (converted to RGB), lazy weight download, device validation, class filtering, and configurable confidence threshold (default 0.5).
  * Produces dataset-ready normalized bounding boxes and optional per-instance cropped masks; class names exposed.

* **Refactor**
  * Centralized image-to-RGB conversion and detection-to-frame helpers for consistent output handling.

* **Tests**
  * Unit tests covering device handling, input conversion, threshold/metadata propagation, filtering, and output formatting.


fiftyone/utils/rfdetr.py
```diff
@@ -0,0 +1,451 @@
...
+    For pretrained COCO models this returns the standard 80-class COCO
+    mapping with non-contiguous integer keys (1-90 with gaps).
+
+    Args:
+        model: an :class:`rfdetr.detr.RFDETR` instance
+
+    Returns:
+        a dict mapping ``{int: str}``
+    """
+    return model.class_names
```
lorenzoriano: I really don't think we need this method, it's 1 line, in the property below you can just return that

fiftyone/utils/rfdetr.py
```diff
@@ -0,0 +1,451 @@
...
+        # RF-DETR handles its own preprocessing (resize + normalize).
+        # Return an identity transform that keeps the original image as-is.
+        return _IdentityTransform(), True
+
+    def _forward_pass(self, imgs: list[Any]) -> dict[str, Any]:
+        pil_images, sizes = _imgs_to_pil(imgs)
+
+        threshold = self.config.confidence_thresh
+        if threshold is None:
+            threshold = 0.5
```
lorenzoriano: This ia bit arbitrary. Since TorchImageModelConfig already defaults confidence_thresh to None, the default belongs in the model config's **init** or should at least be a named module-level constant.

fiftyone/zoo/models/manifest-torch.json
lorenzoriano, line 1: Not a blocking issue, but having this big refactor in this file is out of scope and should be better moved in a different PR

fiftyone/utils/rfdetr.py
```diff
@@ -0,0 +1,451 @@
...
+def _normalize_rfdetr_device(device):
+    """Normalize FiftyOne device strings to the subset RF-DETR accepts."""
+    if device is None:
+        return None
+
+    if isinstance(device, torch.device):
+        device = str(device)
+
+    if device.startswith("cuda:"):
+        return "cuda"
```
lorenzoriano: Is this necessary for rfdetr? It might silently break systems with multiple GPUs. I'd rather print a big a warning or, even better, throw an error, instead of silently moving to a diffeent GPU

fiftyone/utils/rfdetr.py
```diff
@@ -0,0 +1,451 @@
...
+                    mask_crop = mask_crop > 0.5
+
+                kwargs["mask"] = mask_crop
+
+        detections.append(fol.Detection(**kwargs))
+
+    return fol.Detections(detections=detections)
+
+
+def _tensor_to_pil(img: torch.Tensor) -> Image.Image:
```
lorenzoriano: here is already ToPILImage and generic image loading in [torch.py](http://torch.py:2623). I would prefer a shared helper, something like to_rgb_pil(img) or imgs_to_rgb_pil(imgs) in fiftyone.utils.torch or fiftyone.utils.image, not RF-DETR-local

fiftyone/utils/rfdetr.py
```diff
@@ -0,0 +1,451 @@
...
+    Args:
+        model: an :class:`rfdetr.detr.RFDETR` instance
+
+    Returns:
+        a dict mapping ``{int: str}``
+    """
+    return model.class_names
+
+
+def _sv_detections_to_fo(
```
lorenzoriano, line 213: isn't this a duplicate of DetectionOutputProcessor and InstanceSegmenterOutputProcessor in torch.py?

fiftyone/utils/rfdetr.py
lorenzoriano: same for _numpy_to_pil(), _imgs_to_pil()

fiftyone/utils/rfdetr.py
```diff
@@ -0,0 +1,451 @@
...
+    def _load_model(self, config: Any) -> Any:
+        return _load_rfdetr_model(
+            config.model_type,
+            device=_normalize_rfdetr_device(config.device),
+        )
+
+    def _build_transforms(self, config: Any) -> tuple[Any, bool]:
+        # RF-DETR handles its own preprocessing (resize + normalize).
+        # Return an identity transform that keeps the original image as-is.
+        return _IdentityTransform(), True
```
lorenzoriano: I think this is not correct.  ragged_batches=True causes _parse_batch_size() in models.py to downgrade
batch_size > 1 to 1, with a  "Model does not support batching"
warning.  Since predict_all is fully overridden and never stacks images into a tensor, you can change to: return None, False. But please write a unit test for it!

Bonus points: you don't need  _IdentityTransform  as None is already handled by the
parent (if self._transforms is not None).

fiftyone/utils/rfdetr.py
lorenzoriano, line 1: General comment on this file. Some functions are redundant.

- Keep RF-DETR model-loading code local
- Replace the local PIL conversion stack with shared utilities if a second wrapper needs the
same behavior
- Refactor the bbox/mask conversion into fiftyone.utils.torch, because that logic is already
effectively shared and currently duplicated

fiftyone/utils/rfdetr.py
```diff
@@ -0,0 +1,451 @@
...
+            bounding_box=bounding_box,
+            confidence=score,
+        )
+
+        if has_masks and masks is not None:
+            # Crop the full-image boolean mask to the bounding box region
+            # x1/y1/x2/y2 are already clamped to image bounds above
+            mask_full = masks[i]
+            r1, r2 = round(y1), round(y2)
+            c1, c2 = round(x1), round(x2)
```
lorenzoriano: round() on both bounds can shrink the crop enough to drop boundary pixels or even produce an empty mask for small instances. This is an edge case, but one that could be problematic. using floor/ceil
semantics would preserve the full instance region. The same issue is in torch.py so, if you end up using that one, you change it there (but write unit tests that target this very specific behavior)

lorenzoriano (changes requested):
Airight a lot of stuff here, especially around refactoring. sorry!

fiftyone/zoo/models/manifest-torch.json
CharlesCNorton: very true took it out

CharlesCNorton:
@lorenzoriano All addressed! Please let me know how it looks!

fiftyone/utils/rfdetr.py
```diff
@@ -0,0 +1,320 @@
...
+            confidence=score,
+            mask=mask,
+        )
+        if detection is not None:
+            detections.append(detection)
+
+    return fol.Detections(detections=detections)
+
+
+class _RFDETRBaseModel(fout.TorchSamplesMixin, fout.TorchImageModel):
```
lorenzoriano: Do you actually need to inherit TorchSampleMixin? predict_all discards the samples

fiftyone/utils/torch.py
lorenzoriano, line 1: There are behavior changes in both DetectorOutputProcessor.\_parse_output and InstanceSegmenterOutputProcessor._parse_output that go beyond DETR, they will affect every other model. The changes are not bad, but I'm worried that we're changing the behavior of other classes. If anytning, they should belong to a different PR. And they should have their own tests. You either make this code DETR spefic, or make a second PR with these changes (and several unit tests to preserve and guarantee behavior). Right now I can't find tests for this code path

lorenzoriano (changes requested):

lorenzoriano (approved):

PR #7140, opened 2026-03-06
FOEPD-3174: Add a new field in schema manager should activate it 

lanzhenw:

Bug: Newly created fields in the Schema Manager aren't available for annotation until a page refresh because the field isn't "checked" in the Explore sidebar, and visibleLabelSchemas requires fields to be active in both the Annotate and Explore systems.

Root cause: The Explore sidebar's active fields come from Recoil (backed by the Relay store), while the Annotate sidebar uses Jotai. When a new field is created: (1) the Relay store doesn't know about it until a re-fetch, and (2) the Recoil cached field selection doesn't include it. A Recoil→Jotai sync effect overwrites any direct Jotai updates once the re-fetch completes.

And also did 3387 to update the text copies in the schema manager and the tech debt that schema manager was not using the same interface through useSchemaManager() hook

https://github.com/user-attachments/assets/3d05dd1d-bc9a-4cb8-83fb-6271243ffa91

        notes for FiftyOne users.

Bug: Newly created fields in the Schema Manager aren't available for annotation until a page refresh because the field isn't "checked" in the Explore sidebar

-   [ ] Other

## Release Notes

* **UI Updates**
  * Updated tooltip text in the Annotate tab for improved clarity on field availability
  * Refined field schema management labels and messaging terminology
  * Corrected capitalization in field preview section

* **Refactor**
  * Improved internal schema management architecture with enhanced asynchronous handling and event-driven updates

### DIFF
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
@@ -20,6 +20,10 @@ import {
   useSchemaManager,
   type UpdateSchemaRequest,
 } from "../../useSchemaManager";
+import {
+  dispatchSchemaManagerEvent,
+  useSchemaManagerEventBus,
+} from "../events";
 import { currentLabelSchema } from "../state";
 import { reconcileComponent } from "../utils";
 
@@ -146,6 +150,7 @@ const useSave = (field: string, visibilityChanged: boolean) => {
   const notify = useNotification();
   const [current] = useCurrentLabelSchema(field);
   const setCurrentField = useSetAtom(currentField);
+  const { dispatch } = useSchemaManagerEventBus();
 
   return {
     isSaving,
@@ -163,13 +168,13 @@ const useSave = (field: string, visibilityChanged: boolean) => {
       } catch (error) {
         console.error("Failed to save label schema:", error);
         setIsSaving(false);
-        document.dispatchEvent(new CustomEvent("schema-manager-save-complete"));
+        dispatchSchemaManagerEvent(dispatch, "schema-manager:save-complete");
         return;
       }
 
       setSaved(current);
       setIsSaving(false);
-      document.dispatchEvent(new CustomEvent("schema-manager-save-complete"));
+      dispatchSchemaManagerEvent(dispatch, "schema-manager:save-complete");
 
       // Determine activation change: first save auto-activates,
       // otherwise apply the visibility toggle for this field
@@ -209,6 +214,7 @@ const useScan = (field: string) => {
   const [, setCurrent] = useCurrentLabelSchema(field);
   const { createSchemas } = useSchemaManager();
   const limit = useQueryPerformanceSampleLimit();
+  const { dispatch } = useSchemaManagerEventBus();
   return {
     isScanning,
     scan: async () => {
@@ -220,7 +226,7 @@ const useScan = (field: string) => {
         }
       } finally {
         setIsScanning(false);
-        document.dispatchEvent(new CustomEvent("schema-manager-scan-complete"));
+        dispatchSchemaManagerEvent(dispatch, "schema-manager:scan-complete");
       }
     },
     cancelScan: () => {
@@ -237,6 +243,7 @@ const useValidate = (field: string) => {
   const [, setCurrent] = useCurrentLabelSchema(field);
   const discard = useDiscard(field, () => setErrors([]));
   const { validateSchemas } = useSchemaManager();
+  const { dispatch } = useSchemaManagerEventBus();
 
   const resetErrors = useCallback(() => {
     setErrors([]);
@@ -261,15 +268,13 @@ const useValidate = (field: string) => {
           setErrors(result.errors);
         }
 
-        if (!result.errors.length) {
+        if (!result.errors?.length) {
           setCurrent(parsed);
           setIsValid(true);
-          document.dispatchEvent(new CustomEvent("schema-manager-valid-json"));
+          dispatchSchemaManagerEvent(dispatch, "schema-manager:valid-json");
         } else {
           setIsValid(false);
-          document.dispatchEvent(
-            new CustomEvent("schema-manager-invalid-json")
-          );
+          dispatchSchemaManagerEvent(dispatch, "schema-manager:invalid-json");
         }
       } catch (e) {
         if (e instanceof SyntaxError) {
--- a/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/hooks.ts
+++ b/app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/hooks.ts
@@ -27,7 +27,6 @@ import {
   activeSchemaTab,
   addToActiveSchemas,
   currentField,
-  exploreActiveFields,
   fieldAttributeCount,
   fieldType,
   fieldTypes,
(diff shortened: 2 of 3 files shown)


app/packages/core/src/components/Modal/Sidebar/Annotate/useSchemaManager.ts
```diff
@@ -133,6 +133,21 @@ export type UpdateSchemaResponse = {
...
+  field_type: string;
+  read_only?: boolean;
+  label_schema_config?: Record<string, unknown>;
+  schema_config?: Record<string, unknown>;
+};
+
+export type CreateAndActivateFieldResponse = {
+  field_name: string;
+  label_schema: FieldSchema;
+  error?: string;
```
lanzhenw, line 155: validateSchemas will skipErrors and return the error

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -255,42 +249,35 @@ const useValidate = (field: string) => {
...
       } catch (e) {
         if (e instanceof SyntaxError) {
           setErrors([e.message]);
         }
-
-        setIsValidating(false);
         setIsValid(false);
+      } finally {
+        setIsValidating(false);
       }
```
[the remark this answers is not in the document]
lanzhenw: This is minor for the json editor. SyntaxError still sets the error message, that's the key.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/hooks.ts
```diff
@@ -391,20 +412,20 @@ export const useToggleFieldVisibility = (field: string) => {
...
+    activateSchemas({ fields: Array.from(selected) });
     setSelected(new Set());
     setMessage({
       msg: `${selected.size} schema${
         selected.size > 1 ? "s" : ""
       } moved to active fields`,
       variant: "success",
     });
-  }, [activateFields, addToActiveSchema, selected, setSelected, setMessage]);
+  }, [activateSchemas, addToActiveSchema, selected, setSelected, setMessage]);
```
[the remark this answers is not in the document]
lanzhenw: this is the same fire-and-forget pattern before we change, not a regression.

plugins/operators/annotation.py
```diff
@@ -239,6 +239,8 @@ def execute(self, ctx):
             # Activate the field (prepend to make it appear at top)
             ctx.dataset.activate_label_schemas(field_name, prepend=True)
 
+            ctx.trigger("reload_dataset")
```
kaixi-wang: I'm weary of calling reload_dataset. It could really make things unusable for large datasets.

lanzhenw: Good call. I dropped the reload

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
```diff
@@ -208,26 +207,21 @@ const useSave = (field: string, visibilityChanged: boolean) => {
...
-          },
+      try {
+        const result = await createSchemas({ field, limit });
+        if (result.label_schema) {
+          setCurrent(result.label_schema);
         }
-      );
+      } finally {
+        setIsScanning(false);
+        document.dispatchEvent(new CustomEvent("schema-manager-scan-complete"));
```
tom-vx51: I know this is refactoring existing behavior, but any reason we need to use custom events? Can we use our event library instead?

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/NewFieldSchema/NewFieldSchema.tsx
```diff
@@ -194,88 +195,76 @@ const NewFieldSchema = () => {
...
+
+      const { active_label_schemas, label_schemas } = await listSchemas({});
+
+      setLabelSchemasData(label_schemas);
+      setActiveFields(active_label_schemas);
+
+      // Add the new field to exploreActiveFields so it's
+      // immediately visible (visibleLabelSchemas intersects
+      // activeLabelSchemas with exploreActiveFields).
+      addToExploreActiveFields(trimmedName);
```
tom-vx51, line 241: this works, but feels like we are ignoring the original source of truth. How does this get populated initially? Is it possible to re-trigger that behavior so that everything gets refreshed while keeping the server the source of truth?

My concern is that we're creating coupling here--if any other part of the app cares about the available fields, we would need to ensure this logic updates those behaviors as well. If we can maintain a single update path, then we can be confident that all consumers (current and future) are updated appropriately.

app/packages/core/src/components/Modal/Sidebar/Annotate/useSchemaManager.ts
```diff
@@ -133,6 +133,21 @@ export type UpdateSchemaResponse = {
   label_schema: FieldSchema;
 };
 
+export type CreateAndActivateFieldRequest = {
+  field_name: string;
+  field_category: "label" | "primitive";
+  field_type: string;
+  read_only?: boolean;
+  label_schema_config?: Record<string, unknown>;
```
tom-vx51: can we make these config fields stricter types?

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/EditFieldLabelSchema/useLabelSchema.ts
lanzhenw: yeah this is refactoring existing code. It is used in the json-editor POM. I thought this has been the pattern to determine certain process is ready (to avoid set time out) on e2e.

app/packages/core/src/components/Modal/Sidebar/Annotate/SchemaManager/NewFieldSchema/NewFieldSchema.tsx
lanzhenw: This is the new requirement: adding a new field label schema, should also turn on the field in the explore sidebar side. It actually requires a mutation. And we only do this to "newly created field".

lanzhenw: Agreed this isn't ideal, but right now I don't see a clean way around it given the current architecture. The root issue is that there's no lightweight mechanism to re-fetch just the active paths (entries used by explore sidebar) from the server:

  - refreshSchema() only triggers aggregation refreshes, not _activeFields
  - The GraphQL-backed _activeFields atom would need a full dataset query re-fetch to pick up the new
  field. And that would be more expensive for large datasets.

kaixi-wang: Why do you need to refetch? Technically it's initiated on the client so you could just update it clientside simultaneously

lanzhenw: No, we should avoid refetch imo. The current code is updating the client side `activeField` atom to brdige the gap. But it would be nice to be able to sync the explore sidebar entries w/t a full reload, that would fixing it from the source of truth.

PR #7143, opened 2026-03-07
fix: sam2 video unit test functional

neerajaabhyankar:

Previously, the sam2 intensive test was failing due to a syntax error.
```
dataset.match_frames(F("frame_number" > 1))
```
instead of
```
dataset.match_frames(F("frame_number") > 1).set_field(
```
It was also not checking anything with asserts, unlike most other tests in the file.

This PR fixes the bug and adds some asserts.
This is also a precursor to https://github.com/voxel51/fiftyone/pull/7124, a work-in-progress PR where I also test SAM2 on flattened image datasets.

This patch is itself a test.

        notes for FiftyOne users.

-   [x] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected frame-number predicate handling and the field reference used during video processing.

* **Tests**
  * Added configurable frame-limit support and pre-trimming when loading video samples.
  * Added post-processing validation to ensure processed frames contain expected labels and detections.
  * Updated tests to exercise the new frame-limit behavior.

### DIFF
--- a/tests/intensive/model_zoo_tests.py
+++ b/tests/intensive/model_zoo_tests.py
@@ -111,6 +111,7 @@ def test_sam2_video():
     _apply_video_models(
         models,
         max_samples=2,
+        max_frames=3,
         apply_kwargs={_SAM_PROMPT_FIELD: "frames.detections"},
     )
 
@@ -244,6 +245,7 @@ def _apply_video_models(
     confidence_thresh=None,
     pass_confidence_thresh=False,
     max_samples=10,
+    max_frames=10,
     model_kwargs=None,
     apply_kwargs=None,
 ):
@@ -260,10 +262,11 @@ def _apply_video_models(
         dataset_name=fo.get_default_dataset_name(),
         max_samples=max_samples,
     )
+    dataset.match_frames(F("frame_number") <= max_frames).keep_frames()
 
     if _SAM_PROMPT_FIELD in kwargs:
-        dataset.match_frames(F("frame_number" > 1)).set_field(
-            "frames.detections", None
+        dataset.match_frames(F("frame_number") > 1).set_field(
+            kwargs[_SAM_PROMPT_FIELD], None
         ).save()
 
     for idx, model_name in enumerate(model_names, 1):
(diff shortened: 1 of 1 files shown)


tests/intensive/model_zoo_tests.py
manushreegangwar, line 281: Does this use `_SAM_PROMPT_FIELD` as the `prompt_field`? Where is `prompt_field` added to `kwargs`?

tests/intensive/model_zoo_tests.py
```diff
@@ -279,6 +282,11 @@ def _apply_video_models(
             model, label_field=label_field, batch_size=batch_size, **kwargs
         )
 
+        for sample in dataset:
+            last_frame = sample.frames[max_frames]
```
manushreegangwar: Can't this have fewer than `max_frames`?

neerajaabhyankar: not for the specific unit test's case, but added a safeguard just in case.

tests/intensive/model_zoo_tests.py
neerajaabhyankar: The test passes in `apply_kwargs={_SAM_PROMPT_FIELD: "frames.detections"}`
And then `kwargs.update(apply_kwargs)` is called earlier in this function

manushreegangwar: Ah I misread `_SAM_PROMPT_FIELD` as string. It is set to `prompt_field`

tests/intensive/model_zoo_tests.py
```diff
@@ -279,6 +282,12 @@ def _apply_video_models(
             model, label_field=label_field, batch_size=batch_size, **kwargs
         )
 
+        for sample in dataset:
+            if len(sample.frames) > 0:
+                last_frame = sample.frames[len(sample.frames)]
```
manushreegangwar, line 287: nit: -1 should also work

neerajaabhyankar: I get this `fiftyone.core.frame_utils.FrameError: Frame numbers must be 1-based integers; found -1`

PR #7144, opened 2026-03-07
Expose additional Depth Anything V3 upstream parameters

CharlesCNorton:
## Summary

- Add `ref_view_strategy` config param for multi-view reference view selection (`first`, `middle`, `saddle_balanced`, `saddle_sim_range`)
- Add `align_to_input_ext_scale` config param for pose scale alignment to input extrinsics
- Surface `Prediction.scale_factor` (metric scale) on output heatmaps when returned by upstream
- Add `conf_thresh_percentile`, `num_max_points`, `show_cameras` params to `compute_3d_exports` for GLB export control
- Document `process_res_method` supported values (`upper_bound_resize`, `lower_bound_resize`)

All new parameters use upstream defaults and are fully backward-compatible.

## Test plan

- [x] 24 unit tests pass (config parsing, output processor, scale_factor surfacing)
- [x] GPU inference with `depth-anything-v3-base-torch` using explicit `ref_view_strategy` and `align_to_input_ext_scale`
- [x] GPU inference with `depth-anything-v3-metric-large-torch` confirming `scale_factor` guard handles absence

## Summary by CodeRabbit

* **New Features**
  * Depth Anything V3 now propagates per-prediction scale factors to depth heatmaps.
  * Added reference-view selection and input-extrinsic scale alignment options for inference.
  * 3D exports now support confidence thresholds, point limits, camera display, and Gaussian-splat videos.
  * Video exports select the latest generated MP4 artifact.

* **Bug Fixes**
  * Improved handling of unavailable export components and incomplete prediction outputs.

* **Tests**
  * Expanded coverage for configuration, inference options, scale factors, and 3D exports.

## numpy 2 compatibility

Also makes DA3 usable in numpy 2 environments. `depth_anything_3` pins `numpy<2`, but only its 3D export stack needs that; a full dependency install silently downgrades the host numpy. The wrapper now installs the package with `--no-deps` plus its actual inference dependencies (`evo`, `e3nn`, both numpy-2-clean), and registers a stub for `depth_anything_3.utils.export` when the real module cannot import, so depth prediction works everywhere and `compute_3d_exports` raises with guidance in environments without the export stack. Verified on numpy 2.4.4: da3-base and da3-giant load and predict, host numpy untouched, all 28 unit tests pass.

---
**Enterprise sync PR**: https://github.com/voxel51/fiftyone-teams/pull/3354

### DIFF
--- a/fiftyone/utils/depth_anything.py
+++ b/fiftyone/utils/depth_anything.py
@@ -7,6 +7,7 @@ wrapper for the FiftyOne Model Zoo.
 |
 """
 
+import glob
 import logging
 import os
 from typing import Any, Dict, List, Optional, Tuple
@@ -412,6 +413,7 @@ class DepthAnythingV3Model(fout.TorchImageModel):
         overwrite: bool = False,
         skip_failures: bool = False,
         progress: Optional[bool] = None,
+        *,
         conf_thresh_percentile: float = 40.0,
         num_max_points: int = 1_000_000,
         show_cameras: bool = True,
@@ -483,3 +485,9 @@ class DepthAnythingV3Model(fout.TorchImageModel):
                 sample["da3_export_path"] = os.path.join(
                     sample_export_dir, "gs_ply", "0000.ply"
                 )
+            elif export_format == "gs_video":
+                video_paths = sorted(
+                    glob.glob(os.path.join(sample_export_dir, "gs_video", "*.mp4"))
+                )
+                if video_paths:
+                    sample["da3_export_path"] = video_paths[-1]
--- a/fiftyone/utils/depth_anything.py
+++ b/fiftyone/utils/depth_anything.py
@@ -144,6 +144,19 @@ class DepthAnythingV3OutputProcessor(fout.OutputProcessor):
                 if len(sky_masks.shape) == 2:
                     sky_masks = sky_masks[np.newaxis, ...]
 
+        scale_factors = output.get("scale_factor")
+        if isinstance(scale_factors, torch.Tensor):
+            scale_factors = scale_factors.detach().cpu().numpy()
+        if scale_factors is not None and np.ndim(scale_factors) > 0:
+            if len(scale_factors) < len(depth_maps):
+                logger.warning(
+                    "Scale factors (%d) fewer than depth maps (%d), "
+                    "skipping scale_factor",
+                    len(scale_factors),
+                    len(depth_maps),
+                )
+                scale_factors = None
+
         width, height = frame_size
         results = []
 
@@ -196,8 +209,18 @@ class DepthAnythingV3OutputProcessor(fout.OutputProcessor):
             if heatmap.is_metric:
                 heatmap.max_depth = float(max_depth)
 
-            if output.get("scale_factor") is not None:
-                heatmap.scale_factor = output["scale_factor"]
+            if scale_factors is not None:
+                value = (
+                    scale_factors
+                    if np.ndim(scale_factors) == 0
+                    else scale_factors[i]
+                )
+                if isinstance(value, np.ndarray):
+                    value = value.item()
+                elif hasattr(value, "item"):
+                    value = value.item()
+
+                heatmap.scale_factor = float(value)
 
             results.append(heatmap)
 
--- a/fiftyone/utils/depth_anything.py
+++ b/fiftyone/utils/depth_anything.py
@@ -41,6 +41,12 @@ da3_api = fou.lazy_import(
 )
 
 DEFAULT_DA3_MODEL = "depth-anything/da3-base"
+_VALID_REF_VIEW_STRATEGIES = (
+    "first",
+    "middle",
+    "saddle_balanced",
+    "saddle_sim_range",
+)
 
 
 class DepthAnythingV3OutputProcessor(fout.OutputProcessor):
@@ -190,6 +196,9 @@ class DepthAnythingV3OutputProcessor(fout.OutputProcessor):
             if heatmap.is_metric:
                 heatmap.max_depth = float(max_depth)
 
+            if output.get("scale_factor") is not None:
+                heatmap.scale_factor = output["scale_factor"]
+
             results.append(heatmap)
 
         if results:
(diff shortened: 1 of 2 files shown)


fiftyone/utils/depth_anything.py
```diff
@@ -429,6 +504,9 @@ def compute_3d_exports(
                     infer_gs=infer_gs,
```
lorenzoriano, line 590: <p class='graphite__hidden'><i>[Re: line +502]</i></p>
Here you're not passing ref_view_strategy and align_to_input_ext_scale, is that intentional?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/voxel51/fiftyone/7144?utm_source=unchanged-line-comment">Graphite</a>.</p>

fiftyone/utils/depth_anything.py
```diff
@@ -40,6 +41,12 @@ def _ensure_depth_anything_3() -> None:
...
 DEFAULT_DA3_MODEL = "depth-anything/da3-base"
+_VALID_REF_VIEW_STRATEGIES = (
+    "first",
+    "middle",
+    "saddle_balanced",
+    "saddle_sim_range",
+)
 
 
 class DepthAnythingV3OutputProcessor(fout.OutputProcessor):
```
lorenzoriano, line 120: <p class='graphite__hidden'><i>[Re: lines +69 to +73]</i></p>
missing confidence_thresh and classes
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/voxel51/fiftyone/7144?utm_source=unchanged-line-comment">Graphite</a>.</p>

lorenzoriano (dismissed):
just a couple of small things

fiftyone/utils/depth_anything.py
```diff
@@ -26,12 +31,66 @@
...
+
+
+def _install_depth_anything_3() -> None:
+    # depth-anything-3 pins numpy<2, but its inference stack (including evo,
+    # which pose alignment uses) runs on numpy 2; only the 3D export stack
+    # needs the pin. Installing the full dependency set would downgrade the
+    # host numpy, so the package is installed without dependencies, evo is
+    # installed explicitly, and `_stub_da3_export()` absorbs the export
+    # stack when it cannot import
+    for pkg in ("evo", "e3nn"):
```
AdonaiVera: I found a couple more packages we need while testing locally, adding them here.

```
_DA3_INFERENCE_DEPS = (
    "evo",
    "e3nn",
    "omegaconf",
    "einops",
    "addict",
    "safetensors",
)
```

fiftyone/utils/depth_anything.py
```diff
@@ -429,6 +565,9 @@ def compute_3d_exports(
                     infer_gs=infer_gs,
                     export_dir=sample_export_dir,
                     export_format=export_format,
+                    conf_thresh_percentile=conf_thresh_percentile,
+                    num_max_points=num_max_points,
+                    show_cameras=show_cameras,
```
AdonaiVera, line 599: Here, we had to add these six parameters because compute_3d_exports, _forward_pass, and compute_multiview_depth all invoke the same underlying model, but only the export path wasn't forwarding them. As a result, user-configured settings (such as ref_view_strategy) were applied during depth prediction but silently ignored during 3D exports, causing inconsistent outputs from the same model configuration. This is a correctness issue, not just a style concern, the exported GLB/PLY/video could differ from the depth heatmap generated with the same settings. By forwarding all six configuration values, all three inference entry points now behave consistently and produce matching results regardless of which method the user uses.

```
self._model.inference(
    [sample.filepath],
    process_res=self.config.process_res,
    process_res_method=self.config.process_res_method,
    infer_gs=infer_gs,
    use_ray_pose=self.config.use_ray_pose,
    ref_view_strategy=self.config.ref_view_strategy,
    align_to_input_ext_scale=self.config.align_to_input_ext_scale,
    export_feat_layers=self.config.export_feat_layers or [],
    export_dir=sample_export_dir,
    export_format=export_format,
    conf_thresh_percentile=conf_thresh_percentile,
    num_max_points=num_max_points,
    show_cameras=show_cameras,
)
```

fiftyone/utils/depth_anything.py
```diff
@@ -444,3 +583,9 @@ def compute_3d_exports(
                 sample["da3_export_path"] = os.path.join(
                     sample_export_dir, "gs_ply", "0000.ply"
                 )
+            elif export_format == "gs_video":
+                video_paths = sorted(
+                    glob.glob(os.path.join(sample_export_dir, "gs_video", "*.mp4"))
```
AdonaiVera: Small suggestion here, `glob.glob()` treats characters like `[`, `]`, and `*` as wildcard patterns rather than literal characters. If the export directory name contains any of those characters, the match can fail and the exported file won't be found. Wrapping `sample_export_dir` with `glob.escape()` ensures the directory name is treated literally, while still allowing `"gs_video"` and `"*.mp4"` to work as expected.


```suggestion
                    glob.glob(
                        os.path.join(
                            glob.escape(sample_export_dir),
                            "gs_video",
                            "*.mp4",
                        )
                    )
```

AdonaiVera (changes requested):
Overall, this looks good. The tests cover the main scenarios well, including default values, custom configs, invalid inputs, and different `scale_factor` cases. I also like that they use stubs and `monkeypatch` instead of depending on the real model, which keeps them fast and focused. The API signature check is a nice addition, and the new tests follow the same structure as the rest of the file.

A few suggestions:

* `test_compute_3d_exports_uses_keyword_only_args_and_sets_gs_video_path` uses hardcoded Windows paths (`C:\exports` and `C:\data\image.png`). It would be better to use `tmp_path` so the test works cleanly across platforms and doesn't leave directories behind.
* One thing I think is still missing is a test that verifies the config is actually forwarded to `self._model.inference`. Right now the tests only check that the values are parsed correctly, but not that they reach the model. A regression like the recent `compute_3d_exports` forwarding bug would still pass the current test suite.
* It could also be worth adding a test for the `glob.escape()` case by using an export directory with characters like `[`, `]`, or `*`. That would make sure this edge case stays covered. (I already added the suggestion code)
* Just a small formatting note, a few of the new lines are longer than the project's Black line length, so they'll probably get wrapped automatically. Can you run formatting here, please.

AdonaiVera (approved):
Thanks for applying all the suggestions, looks good to go! Before we merge, can you rebase with the latest changes from dev so we make sure CI is good? I'll run CI after to confirm everything passes.

kevin-dimichel:
Removing Aloha. No aloha owned files.

PR #7177, opened 2026-03-12
FOEPD 3003 Create Viewport State Bridging Between Explore (Looker) and Annotate (Lighter) Modes 

nverjinski:
### Description
Add functionality to persist a `ViewportState` (`panX, panY, zoom`) when toggling between `EXPLORE` and `ANNOTATE` modes in the modal. This is a useful QoL improvement for a user performing data annotation. When the user zooms and pans in `EXPLORE` mode and determines they need to make an adjustment in `ANNOTATE` mode, clicking the `ANNOTATE` tab will initialize the image at the current zoom and pan level instead of defaulting to the original zoom and pan state.

This PR can be thought of in two parts:
- A ViewportState transfer mechanism to transfer state between Looker and Lighter
- An extension of Lighter to be able to understand when it needs to fit to content, calculate the desired bounds, and move the viewport to that content. 

### Changes 
- Defined a bridge in `fiftyone/state` to track `ViewportState` when switching between `Looker` and `Lighter`
- Extended `Looker` and `Lighter` to be capable of capturing or setting a `ViewportState`
- Extended `Looker` to be capable of initializing with a specific `ViewportState`
- Updated relevant modal renderers to capture and restore `ViewportState` when they mount or unmount

- Extended `LighterSampleRenderer` to calculate a zoom target from a `Sample` and pass that along as an option to get the renderer to initialize at that viewport.

Before:

https://github.com/user-attachments/assets/ba2a25f2-8afd-4c0a-9d62-b23175b38ea1

After:

https://github.com/user-attachments/assets/ad5e3da0-d750-4610-9999-3ac50ca9dfb2

1. Select an image from a dataset
2. In `EXPLORE` mode, pan and zoom in or out on a feature within the image
3. Click the `ANNOTATE` tab to switch modes
4. Verify the `ANNOTATE` tab (Lighter) initializes to the same pan and zoom level you were at when you were in `EXPLORE` mode (Looker)
5. Click the right arrow button to navigate to the next image
6. Verify the image is zoomed and panned to its default display values

## Release Notes
- Toggling between Explore and Annotate modes within the view modal maintains viewport state.

        notes for FiftyOne users.

- Toggling between Explore and Annotate modes within the view modal maintains viewport state.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Modal viewer now saves/restores per-sample zoom/pan, can auto-zoom to detected content, and defers showing media until viewport restore/auto-zoom completes for a smoother revisit.
  * Viewport can be programmatically fitted to content and restored across renderer switches for consistent visual state.

* **Chores**
  * Viewer configuration and public hooks/APIs expanded to capture, persist, and synchronously read viewport snapshots and new auto-zoom options.


app/packages/core/src/components/Modal/use-looker.ts
```diff
@@ -130,6 +133,29 @@ function useLooker<L extends fos.Lookers>({
...
+    };
+  }, [looker, sample, setViewportState]);
+
+  // Seed the viewport to restore when the Looker first fully loads.
+  useEffect(() => {
+    const saved = modalBridge.getModalViewport();
+    if (saved?.sampleId === sample.sample._id) {
+      looker.setInitialViewport(saved);
+    }
+  }, [looker, sample]);
```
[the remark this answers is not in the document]
nverjinski: False positive.

kaixi-wang:
I pulled this branch and tested patches views and it doesn't look like the zoom level is preserved. 
Repro: Click ToPatches button and select field with detections. In a patches view, the modal should preserve the zoom level from the grid.

https://github.com/user-attachments/assets/e0c021e3-c2f7-44b5-b67e-201ae9fe0187

benjaminpkane (changes requested):
@erik-nieh  added the lighter toolbar pom which has zoom in/out methods. Here is an [example spec](https://gist.github.com/benjaminpkane/0e3d2fd3ce855ccbc93952cbc92353d7)

app/packages/looker/src/lookers/image.ts
```diff
@@ -78,7 +78,14 @@ export class ImageLooker extends AbstractLooker<ImageState> {
...
       LookerUtils.toggleZoom(this.state, this.currentOverlays);
     } else if (this.state.setZoom && this.state.overlaysPrepared) {
-      if (this.state.options.zoom) {
+      if (this.pinnedViewport) {
+        this.state.scale = this.pinnedViewport.scale;
+        this.state.pan = [
+          this.pinnedViewport.panX,
+          this.pinnedViewport.panY,
+        ] as Coordinates;
+        this.pinnedViewport = null;
```
benjaminpkane: Could we add `pinnedViewport` as an option to `BaseOptions`? Looker isn't really designed to have state directly attached to it.

nverjinski: No problem. I renamed it to `initialViewport` and put it in `BaseOptions`.

app/packages/core/src/components/Modal/Lighter/LighterSampleRenderer.tsx
```diff
@@ -45,6 +48,20 @@ export const LighterSampleRenderer = ({
   const sampleRef = useRef(sample);
   sampleRef.current = sample;
 
+  const options = useRecoilValue(
+    fos.lookerOptions({ modal: true, withFilter: false })
+  );
```
benjaminpkane: I think we should add this as a hook to `@fiftyone/state`. In `./src/accessors/modal.ts`.

```ts
/**
 * Gets the looker options for the modal.
 *
 * @param withFilter - Whether to apply frontend label filtering. Defaults to `false`.
 */
const useModalLookerOptions = (withFilter = false) => {
  return useRecoilValue(lookerOptions({ modal: true, withFilter }));
};
```

app/packages/lighter/src/constants.ts
```diff
@@ -59,3 +59,9 @@ export const HANDLE_OFFSET_Y = 3;
  */
 export const SELECTED_ALPHA = 0.1;
 export const SELECTED_COLOR = 0x000000;
+
+/**
+ * Default padding applied on each side when auto-zooming to content.
+ * 0.1 per side = 0.2 total, matching Looker's zoomPad: 0.2 default.
+ */
```
benjaminpkane, line 75: No need to reference `Looker`. Might be confusing to future readers, noting we hope to remove it eventually.

```suggestion
/**
 * Default padding applied on each side when auto-zooming to content.
 * 0.1 per side = 0.2 total
 */
```

app/packages/looker/src/lookers/image.ts
```diff
@@ -78,7 +78,14 @@ export class ImageLooker extends AbstractLooker<ImageState> {
...
     } else if (this.state.setZoom && this.state.overlaysPrepared) {
-      if (this.state.options.zoom) {
+      if (this.state.options.initialViewport) {
+        this.state.scale = this.state.options.initialViewport.scale;
+        this.state.pan = [
+          this.state.options.initialViewport.panX,
+          this.state.options.initialViewport.panY,
+        ] as Coordinates;
+        this.state.options = { ...this.state.options, initialViewport: null };
+      } else if (this.state.options.zoom) {
```
benjaminpkane, line 88: Hacky, but looks good! Exactly what I would do

app/packages/looker/src/state.ts
```diff
@@ -528,3 +530,13 @@ export interface FrameChunkResponse extends FrameChunk {
...
+
+/**
+ * The zoom and pan state of a modal viewer, used to transfer camera position
+ * when switching between Explore (Looker) and Annotate (Lighter) modes.
+ */
+export interface ViewportState {
+  readonly scale: number;
+  readonly panX: number;
+  readonly panY: number;
+}
```
benjaminpkane: It is just a suggestion, but I don't think the Explore/Annotate Lighter/Looker details are needed
```suggestion
/**
 * The zoom and pan state of a modal viewer
 */
export interface ViewportState {
  readonly scale: number;
  readonly panX: number;
  readonly panY: number;
}
```

app/packages/state/src/accessors/modal.ts
```diff
@@ -98,3 +100,15 @@ export const useCurrentSampleId = () => {
...
+ * Gets the saved modal viewport (zoom/pan) state.
+ */
+export const useModalViewport = (): ModalViewportState | null =>
+  useAtomValue(__unsafeModalViewportAtom);
+
+/**
+ * Setter for persisting the modal viewport (zoom/pan) state.
+ */
+export const useSaveModalViewport = () =>
+  useSetAtom(__unsafeModalViewportAtom);
```
benjaminpkane: Nice!

app/packages/lighter/src/core/Scene2D.ts
```diff
@@ -900,6 +982,51 @@ export class Scene2D {
     this.config.renderer.addTickHandler(async () => {
       await this.renderFrame();
     });
+
+    // The initial zoom requires two conditions before it can run:
+    // lighter:resize — Dimension correctness guarantee and Pixi readiness
+    // lighter:canonical-media-bounds-changed — image bounds are known
+    //
+    // These events can arrive in either order. Apply the zoom when the second
+    // condition arrives, regardless of ordering.
```
nverjinski: I think a meaningful improvement may be to create a single event that tells us when everything is ready rather than wait upon two separate events. 

Alternatively, `lighter:resize` is being used as a proxy for pixi being ready. Perhaps we could just add an event `lighter:renderer-ready` at the end of `initializePixiJS` and listen to that to avoid the ambiguity of relying on `lighter:resize` ?

app/packages/core/src/components/Modal/Lighter/LighterSampleRenderer.tsx
```diff
@@ -8,17 +8,20 @@ import {
...
   useLighterSetupWithPixi,
 } from "@fiftyone/lighter";
-import type { Sample } from "@fiftyone/state";
+import type { ModalViewportState, Sample } from "@fiftyone/state";
 import * as fos from "@fiftyone/state";
 import { getSampleSrc } from "@fiftyone/state";
+import { modalBridge, useSaveModalViewport } from "@fiftyone/state";
 import { useAtomValue } from "jotai";
-import React, { useEffect, useMemo, useRef, useState } from "react";
+import React, { useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
```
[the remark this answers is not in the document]
nverjinski: This is expected. We do not intend to subscribe reactively to viewport state updates.

app/packages/lighter/src/overlay/ImageOverlay.ts
```diff
@@ -118,6 +154,18 @@ export class ImageOverlay extends BaseOverlay implements CanonicalMedia {
     }
   }
 
+  /**
+   * Returns the container dimensions by reading directly from the DOM.
+   * This is the authoritative source of truth for container dimensions.
+   */
```
nverjinski, line 124: This helper function is necessary to solve the following problem:

Before, `ImageOverlay` would read container dimensions from Pixi's internal state rather than the DOM (`this.renderer.getContainerDimensions`). Sometimes, when switching to a new sample, there was a timing window where the image element's `onLoad` could run and obtain stale dimensions from Pixi before Pixi's resize observer fired and called `app.renderer.resize()`. Any code in `ImageOverlay` that read dimensions before that resize observer callback delivered, most critically the `onload` handler, would get the previous sample's dimensions. This caused a chain of events that lead to an incorrect rect being stored in the `coordinateSystem` transform. `applyInitialZoom()` in `Scene2D.ts` would use this poisoned `coordinateSystem` to calculate a bad `worldRect` which led to obvious viewport problems. 

The render loop self-heals within one frame once correct bounds arrive, but `applyInitialZoom` is a one-shot operation, it fires once when the barrier resolves, commits the wrong pan to the viewport, and never runs again.

The solution is for `ImageOverlay` to read container dimensions directly from the DOM, which is the source of truth, instead of Pixi's cached state.

app/packages/lighter/src/core/Scene2D.ts
```diff
@@ -148,11 +158,20 @@ export class Scene2D {
...
 
   constructor(private readonly config: Scene2DConfig) {
     this.sceneOptions = config.options;
     this.sceneId = config.sceneId;
 
+    if (config.options?.zoom) {
+      this.shouldInitialZoom = true;
+      this.initialZoomPad = config.options.zoomPad ?? 0;
+      this.initialZoomTarget = config.options.zoomTarget ?? null;
+    }
```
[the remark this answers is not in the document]
nverjinski: This is not an issue. `zoomTarget` is a synchronously precomputed target rect to zoom to. It will only be populated if there is content to zoom to. If `shouldInitialZoom` is true and `initialZoomTarget` is false, that means there is nothing to zoom to and the default behavior of rendering the entire image is actually desired.

app/packages/lighter/src/react/useLighterSetup.ts
```diff
@@ -32,25 +46,32 @@ export type LighterOptions = Partial<ReturnType<typeof useLookerOptions>>;
...
   options: LighterOptions,
-  sceneId: string
+  sceneId: string,
 ) => {
   const [scene, setScene] = useAtom(lighterSceneAtom);
+
+  // Frozen at mount time — modalViewportState only changes on Looker/Lighter unmount 
+  // so this value is stable for the entire lifetime of the component.
+  const initialViewportRef = useRef(options.initialViewport);
+  const rendererRef = useRef<PixiRenderer2D | null>(null);
```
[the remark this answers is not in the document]
nverjinski: This is actually safe. The comment describes why

app/packages/core/src/components/Modal/Lighter/LighterSampleRenderer.tsx
```diff
@@ -117,20 +147,49 @@ const LighterSetupImpl = (props: {
...
+  useLayoutEffect(() => {
+    return () => {
+      if (scene && !scene.isDestroyed && sampleId) {
+        setViewportState({
+          sampleId,
+          ...scene.getViewportState(),
+        });
+      }
+    };
+  }, [scene, sampleId]);
```
benjaminpkane: This could go in its own hook, e.g. `useRetrieveViewport`

e2e-pw/src/oss/poms/modal/sample-canvas/index.ts
```diff
@@ -165,6 +137,92 @@ export class SampleCanvasPom {
...
+    await this.page.mouse.wheel(0, deltaY);
+  }
+
+  /**
+   * Capture a clean screenshot of the sample canvas.
+   * Moves the cursor to the canvas center and optionally holds Shift so
+   * renderers hide annotation overlays before snapping the screenshot.
+   * @param hideOverlays Whether to hold Shift to suppress annotation overlays
+   */
+  async screenshot(hideOverlays = false): Promise<Buffer> {
```
benjaminpkane: We should stick to one screenshotting style using `assert.hasScreenshot(...)`. Let's remove this

e2e-pw/src/oss/poms/modal/sample-canvas/index.ts
```diff
@@ -165,6 +137,92 @@ export class SampleCanvasPom {
...
+   *
+   * A rightward drag adds +N to panX; a downward drag adds +N to panY.
+   *
+   * @param direction The direction to pan
+   * @param offsetPixels How many pixels to drag
+   */
+  async pan(
+    direction: "left" | "right" | "up" | "down",
+    offsetPixels: number
+  ) {
```
benjaminpkane: I'd rather we not add this just yet. It is certainly useful. I worry about this POM accumulating too many methods, though

Requiring specs to be imperative until we get further along I think may lead to less friction. Product specifications can then be captured by e2e screenshot specs.

And by product specifications, I mean imperative descriptions of how the canvas behaves due keyboard and mouse events. Mouse move, mouse down, mouse up, etc.

e2e-pw/src/oss/poms/modal/sample-canvas/index.ts
```diff
@@ -165,6 +137,92 @@ export class SampleCanvasPom {
...
+    await this.page.mouse.down();
+    await this.page.mouse.move(endX, endY, { steps: 10 });
+    await this.page.mouse.up();
+  }
+
+  /**
+   * Zoom the canvas by scrolling the mouse wheel at the canvas center.
+   * @param deltaY The wheel scroll delta (negative = zoom in, positive = zoom out)
+   */
+  async zoom(deltaY: number) {
```
benjaminpkane: Same as my comment above

e2e-pw/src/oss/specs/smoke-tests/viewport-bridge.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  await fiftyoneLoader.executePythonCode(`
+    import fiftyone as fo
+    import fiftyone.zoo as foz
+
+    dataset_name = "${datasetName}"
+    dataset = foz.load_zoo_dataset(
+      "quickstart", max_samples=5, dataset_name=dataset_name
+    )
+    dataset.persistent = True
+  `);
```
benjaminpkane: Let's use the `datasetFactory.createBlankDataset` util

e2e-pw/src/oss/specs/smoke-tests/viewport-bridge.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    await modal.waitForSampleLoadDomAttribute();
+
+    const after = await modal.sampleCanvas.screenshot(hideOverlays);
+
+    // Use a Jimp-based comparison with a tiny per-pixel tolerance instead of
+    // a strict byte-level comparison.
+    const beforeImg = await Jimp.read(before);
+    const afterImg = await Jimp.read(after);
+    const diff = jimpDiff(beforeImg, afterImg, 0.02);
+    expect(diff.percent).toBeLessThan(0.001);
```
benjaminpkane: Please use `assert.hasScreenshot`

e2e-pw/src/oss/specs/smoke-tests/viewport-bridge.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    const lookerImg = await Jimp.read(lookerBuf);
+    const lighterImg = await Jimp.read(lighterBuf);
+
+    // Allow per-pixel color distance up to 15% of the maximum (255) to account
+    // for the difference in rendering pipelines (Canvas 2D vs WebGL/Pixi.js).
+    const diff = jimpDiff(lookerImg, lighterImg, 0.15);
+
+    // Less than 10% of pixels may differ; a completely wrong viewport would
+    // push this near 1.0.
+    expect(diff.percent).toBeLessThan(0.1);
```
benjaminpkane: Also `assert.hasScreenshot`

e2e-pw/src/oss/specs/smoke-tests/viewport-bridge.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    const hideOverlays = true;
+    const PAN_X = 150;
+    const ZOOM_IN = -600;
+
+    await grid.openFirstSample();
+    await modal.waitForSampleLoadDomAttribute();
+    await modal.sampleCanvas.assert.is(SampleCanvasType.LOOKER);
+
+    await modal.sampleCanvas.zoom(ZOOM_IN);
+    await modal.sampleCanvas.pan("right", PAN_X);
```
benjaminpkane: Let's write the imperative version of this for the spec (the wheel method will need to be added to the `sampleCanvas` POM

```ts
await modal.sampleCanvas.move(...)
await modal.sampleCanvas.wheel(...)
await modal.sampleCanvas.down(...)
```

e2e-pw/src/oss/poms/modal/index.ts
```diff
@@ -305,6 +305,17 @@ export class ModalPom {
...
+  async waitForLighterReady() {
+    return this.page.waitForFunction(
+      () =>
+        document
+          .querySelector(`[data-cy=lighter-sample-renderer] canvas`)
+          ?.getAttribute("lighter-ready") === "true",
+      undefined,
+      { timeout: Duration.Seconds(20) }
+    );
+  }
```
benjaminpkane, line 325: Ok for now 👍

e2e-pw/src/oss/poms/modal/sample-canvas/index.ts
```diff
@@ -185,7 +243,7 @@ export class SampleCanvasPom {
  * Sample canvas asserter
  */
 class SampleCanvasAsserter {
-  constructor(private readonly sampleCanvasPom: SampleCanvasPom) {}
+  constructor(private readonly sampleCanvasPom: SampleCanvasPom) { }
```
benjaminpkane: Is pre-commit installed in your in local environment? The formatting should be deterministic

app/packages/lighter/src/react/useLighterSetup.ts
```diff
@@ -2,11 +2,14 @@
  * Copyright 2017-2026, Voxel51, Inc.
  */
 
+import type { ViewportState } from "@fiftyone/looker";
```
benjaminpkane: As discussed let's avoid referring to `@fiftyone/looker` in Lighter related code. This type can be added to `@fiftyone/lighter`

app/packages/core/src/components/Modal/use-looker.ts
```diff
@@ -130,6 +131,30 @@ function useLooker<L extends fos.Lookers>({
...
+  }, [looker]);
+
+  // Seed the saved viewport into the looker after mount. This runs after
+  // unmounting component's useLayoutEffect cleanup has already written to the atom.
+  useLayoutEffect(() => {
+    const savedViewport = modalBridge.getModalViewport();
+    if (savedViewport?.sampleId === sample.sample._id) {
+      looker.updateOptions({ initialViewport: savedViewport }, true);
+    }
+  }, [looker, sample]);
```
benjaminpkane: This could go into its own hook (the hook can be in this file)

```suggestion
  useSyncViewport();
```

app/packages/core/src/components/Modal/Lighter/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (w <= 0 || h <= 0) return;
+
+    minX = Math.min(minX, x);
+    minY = Math.min(minY, y);
+    maxX = Math.max(maxX, x + w);
+    maxY = Math.max(maxY, y + h);
+    found = true;
+  };
+
+  for (const value of Object.values(sample)) {
```
benjaminpkane: This function could use some unit tests. Minimally, `embedded.fields` should be supported. In Python, a dataset with an `embedded.detections` field can be created using the below

```py
 dataset.add_sample_field("embedded", fo.EmbeddedDocumentField, embedded_doc_type=fo.DynamicEmbeddedDocument)
 dataset.add_sample_field("embedded.detections", fo.EmbeddedDocumentField, embedded_doc_type=fo.Detections)
```

app/packages/core/src/components/Modal/Lighter/LighterSampleRenderer.tsx
```diff
@@ -45,6 +48,20 @@ export const LighterSampleRenderer = ({
...
+
+  const sampleId = sample?.sample?._id;
+  const savedViewportState = modalBridge.getModalViewport();
+  const initialViewport =
+    savedViewportState?.sampleId === sampleId ? savedViewportState : null;
+
+  // Zoom to content only when options.zoom is set and there is no saved
+  // viewport to restore. Restoring a viewport takes precedence over auto-zoom.
+  const optionsZoom = ("zoom" in options && options.zoom) as boolean | undefined;
+  const effectiveZoom = !!optionsZoom && !initialViewport;
```
benjaminpkane: Can we create a separate hook?

```ts
const useDeferShow = (id: string) => {
  const options = useModalLookerOptions();

  const savedViewportState = modalBridge.getModalViewport();
  const initialViewport =
    savedViewportState?.sampleId === id ? savedViewportState : null;

  // Zoom to content only when options.zoom is set and there is no saved
  // viewport to restore. Restoring a viewport takes precedence over auto-zoom.
  const optionsZoom = ("zoom" in options && options.zoom) as
    | boolean
    | undefined;
  return !!optionsZoom && !initialViewport;
};
```

benjaminpkane (changes requested):
Great stuff! The Lighter stuff definitely proved to be tricky. I think the solution is good. But, as discussed, if we added new lighter events based around initialization, I think we could decouple the feature from lighter by running a callback.

I am not requesting we change the implementation. We can discuss with @sashankaryal and others offline.

I am requesting the comments in the e2e spec are resolved. And that we remove references to Looker in Lighter-only code.

app/packages/lighter/src/core/Scene2D.ts
benjaminpkane: Yes, agree!

benjaminpkane:
Note, for screenshot testing, the linux versions can be added using this [guide](https://github.com/voxel51/fiftyone/blob/develop/e2e-pw/README.md#screenshot-testing). Assuming you are running Mac 🍎

e2e-pw/src/oss/poms/modal/sample-canvas/index.ts
nverjinski: pre-commit was not installed.

lanzhenw:
Nice work!

Not a request for change for this PR, just thinking out loud:

This PR's `extractZoomTarget` walks raw sample JSON, checks for `_cls === 'detection'` and extracts `bounding_box` arrays. Each new label type would require a new branch in that function. 

The `Scene2D` already has overlays: `Map<string, BaseOverlay>` and each overlay has bounds. A cleaner setup could be:

- Add an abstract bounds getter to `BaseOverlay` (or a `getZoomBounds(): Rect | null` method) — `BoundingBoxOverlay` and `KeypointOverlay` already have it, ClassificationOverlay would return null.
- Add a `getContentBounds()` method to `Scene2D` that unions all overlay bounds (excluding the image overlay). This is a one-liner once all overlays expose bounds uniformly.
- Defer the initial zoom until after`Scene2D` has loaded its overlays (listen for a `lighter:overlays-loaded` event or similar), then call `scene.fitToRect(scene.getContentBounds())`.

This avoids the raw JSON parsing entirely, makes the zoom target calculation automatic for any future overlay type, and keeps the knowledge of "what are my bounds" inside each overlay where it belongs. The `deferShow` mechanism in the PR already gates rendering until the viewport is positioned, so waiting slightly longer for overlays to load is architecturally compatible.

if we are okay with defer the zooming until after overlays are loaded, this method would be easier to extend to other label types.

benjaminpkane:
> Nice work!
> 
> Not a request for change for this PR, just thinking out loud:
> 
> This PR's `extractZoomTarget` walks raw sample JSON, checks for `_cls === 'detection'` and extracts `bounding_box` arrays. Each new label type would require a new branch in that function.
> 
> The `Scene2D` already has overlays: `Map<string, BaseOverlay>` and each overlay has bounds. A cleaner setup could be:
> 
> * Add an abstract bounds getter to `BaseOverlay` (or a `getZoomBounds(): Rect | null` method) — `BoundingBoxOverlay` and `KeypointOverlay` already have it, ClassificationOverlay would return null.
> * Add a `getContentBounds()` method to `Scene2D` that unions all overlay bounds (excluding the image overlay). This is a one-liner once all overlays expose bounds uniformly.
> * Defer the initial zoom until after`Scene2D` has loaded its overlays (listen for a `lighter:overlays-loaded` event or similar), then call `scene.fitToRect(scene.getContentBounds())`.
> 
> This avoids the raw JSON parsing entirely, makes the zoom target calculation automatic for any future overlay type, and keeps the knowledge of "what are my bounds" inside each overlay where it belongs. The `deferShow` mechanism in the PR already gates rendering until the viewport is positioned, so waiting slightly longer for overlays to load is architecturally compatible.
> 
> if we are okay with defer the zooming until after overlays are loaded, this method would be easier to extend to other label types.

Great points! We should use overlay bounds, I agree

app/packages/state/src/accessors/modal.ts
```diff
@@ -108,6 +113,35 @@ export const useCurrentSampleId = () => {
...
+ * Gets the looker options for the modal.
+ *
+ * @param withFilter - Whether to apply frontend label filtering. Defaults to `false`.
+ */
+export const useModalLookerOptions = (withFilter = false) => {
+  return useRecoilValue(lookerOptions({ modal: true, withFilter }));
+};
+
+/**
+ * Gets the current "Crop to content" setting in the modal when in a view
```
nverjinski: This ended up not being used. Could remove.

app/packages/core/src/components/Modal/Lighter/useViewport.ts
```diff
@@ -0,0 +1,243 @@
...
+      !rendererReady ||
+      !labelsLoaded ||
+      appliedRef.current
+    )
+      return;
+
+    const complete = () => {
+      appliedRef.current = true;
+
+      /*
```
nverjinski, line 160: I want to draw special attention to the comment for the code below.

Changing `Scene2d.renderFrame()` to be synchronous is an option, but it would have implications. This function is responsible for firing off callbacks `before` and `after` a frame render. Currently, the codebase registers NO async callbacks so this async capability is not being used at the moment, although I'm sure it was designed to be async for future proofing reasons. 

Anyways, changing `Scene2d.renderFrame()` to be synchronous would net us some minor performance gains (although speeding up scene/pixi initialization and overlay registering/rendering would be more valuable) as well as avoiding this hack, but it would come at the expense of not being able to register async render callbacks.

app/packages/events/src/hooks/useEventBus.ts
```diff
@@ -19,6 +19,7 @@ export const useEventBus = <T extends EventGroup>(
     return {
       on: dispatcher.on.bind(dispatcher),
       off: dispatcher.off.bind(dispatcher),
+      once: dispatcher.once.bind(dispatcher),
```
benjaminpkane, line 22: Awesome awesome awesome

app/packages/core/src/components/Modal/Lighter/LighterSampleRenderer.tsx
```diff
@@ -198,12 +115,52 @@ const LighterSetupImpl = (props: {
...
+    // Set the image overlay as canonical media for coordinate transformations
+    scene.setCanonicalMedia(mediaOverlay);
+  }, [scene, sceneId]);
+
+  const useEventHandler = useLighterEventHandler(
+    scene?.getEventChannel() ?? UNDEFINED_LIGHTER_SCENE_ID
+  );
+  useEventHandler(
+    "lighter:viewport-init-complete",
+    useCallback(() => onReveal(), [onReveal]),
```
benjaminpkane: `onReveal` is already memoized, yes?
```suggestion
    onReveal,
```

app/packages/core/src/components/Modal/Lighter/useViewport.ts
```diff
@@ -0,0 +1,260 @@
...
+      setBounds(bounds);
+    }, []),
+    { once: true }
+  );
+
+  return bounds;
+};
+
+/**
+ * Waits for at least one spatial overlay to be present in the scene.
```
benjaminpkane, line 61: To the extent it concerns initialization, I think all overlays are spatial
```suggestion
 * Waits for at least one overlay to be present in the scene.
```

e2e-pw/src/oss/specs/smoke-tests/viewport-bridge.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    grid,
+    modal,
+  }) => {
+    await grid.openFirstSample();
+    await modal.waitForSampleLoadDomAttribute();
+    await modal.sampleCanvas.assert.is(SampleCanvasType.LOOKER);
+
+    // Zoom in by scrolling the wheel at the canvas centre.
+    await modal.sampleCanvas.move(0.5, 0.5);
+    await modal.sampleCanvas.wheel(0, -600);
```
benjaminpkane: Yes yes yes!

e2e-pw/src/oss/specs/smoke-tests/quick-edit.spec.ts
```diff
@@ -234,7 +234,7 @@ test.describe.serial("quick edit", () => {
     // Transition to quick edit via the tooltip
     await modal.sampleCanvas.tooltip.quickEdit();
     await modal.sampleCanvas.assert.is(SampleCanvasType.LIGHTER);
-    await modal.sampleCanvas.move(0.9, 0.9, "default");
+    await modal.sampleCanvas.move(0.9, 0.9, "crosshair");
```
benjaminpkane, line 237: This is expected? I am surprised that this feature has changed behavior here

e2e-pw/src/oss/poms/modal/sample-canvas/index.ts
```diff
@@ -108,15 +108,24 @@ export class SampleCanvasPom {
    *
    * @param x The x coordinate between [0, 1]
    * @param y The y coordinate between [0, 1]
-   * @param cursor An optional cursor value to expect after moving
+   * @param cursor An optional cursor value to expect after moving. When
+   *   provided, the move is retried until the cursor matches. This is
+   *   necessary because the cursor is event-driven — it only updates when a
+   *   mouse event fires — so the underlying state (e.g. detection mode) may
+   *   not have settled yet on the first move attempt.
```
benjaminpkane, line 115: I am learning Playwright as we go. We cannot reliably expect one mouse event always have the same outcome? Or `hasCursor` may assert itself before the browser event resolves?

I think this is ok because it is isolated in the implementation. Just want to be sure we aren't baking in flakiness into canvas e2e if it can be avoided

app/packages/lighter/src/events/index.ts
```diff
@@ -192,6 +192,10 @@ export type LighterEventGroup = {
   "lighter:zoomed": { scale: number };
   /** Emitted when the viewport is panned/moved */
   "lighter:viewport-moved": { x: number; y: number; scale: number };
+  /** Emitted by useViewport once the initial viewport has been applied (or no action was needed) */
+  "lighter:viewport-init-complete": Record<string, never>;
```
benjaminpkane, line 196: I would argue this isn't needed as a lighter event. `useViewport` could encapsulate this (as a consumer of lighter events). I see the event is also dispatched outside of lighter and in React. Not blocking, but I think it meets YAGNI

app/packages/core/src/components/Modal/Lighter/useViewport.ts
```diff
@@ -0,0 +1,260 @@
...
+       * On the first tick, Pixi renders the canvas from the current state of its scene graph.
+       * Because renderFrame is async, the renderOverlay() calls that mutate the scene graph (adding
+       * bounding box Graphics objects to the Pixi stage) haven't run yet by the time Pixi renders,
+       * they're suspended in a pending microtask. So the first canvas paint shows nothing (or whatever
+       * was there before). On the second tick, Pixi renders the now-mutated scene graph and the overlays appear.
+       *
+       * HACK: We use a double-tick pattern to ensure the reveal is after the image and all overlays
+       * have been painted to the canvas. A cleaner solution may be to make Scene2D.renderFrame
+       * synchronous which would cause scene graph mutations to happen in the same tick as
+       * Pixi's per-tick render.
```
benjaminpkane: Yes, something to consider

app/packages/core/src/components/Modal/Lighter/useViewport.ts
benjaminpkane: Async always makes things tricky! Good stuff

This UX is correct from my testing, and things are isolated so a great solution I think. Out of curiosity, can we guarantee the 2-tick pattern? In theory async could run over many ticks, no?

benjaminpkane (commented):
Awesome awesome awesome! An intentional implementation ✨ 

I pushed an update to Looker scaling so we have parity. I think this is ready to go once e2e screenshots assert expectations (as discussed offline).

What could be nice in addition to the existing assertions is a max/min zoom spec that ensures wheeling matches between looker/lighter at the boundaries (0.1x scale and 10x scale)

app/packages/core/src/components/Modal/Lighter/useViewport.ts
benjaminpkane: Actually, disregard. I am wrong. Jet lag

e2e-pw/src/oss/specs/smoke-tests/quick-edit.spec.ts
nverjinski: In `QuickEdit` mode, the expected cursor **should** be `crosshair` if we're not hovering a detection. I believe this was working before because it was asserting `default` cursor before the underlying state for switching to `QuickEdit` mode had fully settled. The `page.mouse.move()` resolves when the event is dispatched, not when the app finishes reacting to it, so the assertion was racing against the renderer initialization. The test was passing, but the assertion was simply wrong.

e2e-pw/src/oss/poms/modal/sample-canvas/index.ts
nverjinski: `page.mouse.move()` resolves when the event is dispatched, not when the app finishes reacting to it. So yes, `hasCursor` can fire before the renderer has processed the event and updated the cursor style. The `toPass()` retry is not baking in flakiness, it's the same behavior for polling until the DOM reflects expected state, just as `toBeVisible()` does. It'll poll for 5 seconds (default) before failing the test, same as `toBeVisible`.

app/packages/lighter/src/events/index.ts
nverjinski: I wanted to lean in to using the event bus which is why it ended up as an event instead of a prop, but yeah, its probably not necessary.

I agree with your point about the event being dispatched outside of Lighter.

nverjinski:
Responding to your comment:
>Async always makes things tricky! Good stuff
>This UX is correct from my testing, and things are isolated so a great solution I think. Out of curiosity, can we guarantee the 2-tick pattern? In theory async could run over many ticks, no?

The two-tick pattern is reliable because all overlay `renderImpl` methods do synchronous Pixi scene graph mutations. A future overlay that awaits something before drawing would break this. The overlay promise is fire and forget, so if `executeRenderCallbacks("after")` fires before the async mutations land, it would cause the scene to reveal before that overlay is painted. Worth noting in the comment, but at the moment this is not an issue.

benjaminpkane (approved):
Looks good. Great stuff #momentum

e2e-pw/src/oss/specs/smoke-tests/viewport-bridge.spec.ts
```diff
@@ -0,0 +1,103 @@
...
```
benjaminpkane: I'm still not seeing this wheel result in a zoom unless my eyes deceive me. Better to remove this wheel call if it's a noop in these tests still imo

nverjinski: The `wheel` was working, it was just small and difficult to see. The `canvas.ts` component was capturing the wheel event no matter what magnitude was passed and scaling by about `9%`. I reworked the `wheel` function in the sample-canvas pom to be more intuitive and updated the tests that use this capability to show this zoom more obviously.

Issue #7128, opened 2026-03-05
[FR] Custom Metrics in Scenario Analysis


Lucasmogsan:
### Proposal Summary

Custom `EvaluationMetric` plugin aggregates should be recomputed when a Scenario Analysis is created in the App.

### Motivation

Custom metrics (e.g. pixel-level keypoint distance) correctly populate the model evaluation, but is not adjusting to new Scenario Analysis. This makes Scenario Analysis (in app) unusable for custom metrics.

### What areas of FiftyOne does this feature affect?

-   [x] App: FiftyOne application
-   [x] Core: Core `fiftyone` Python library
-   [ ] Server: FiftyOne server

### Willingness to contribute

The FiftyOne Community welcomes contributions! Would you or another member of your organization be willing to contribute an implementation of this feature?

-   [ ] Yes. I can contribute this feature independently
-   [ ] Yes. I would be willing to contribute this feature with guidance from the FiftyOne community
-   [x] No. I cannot contribute this feature at this time

PR #7207, opened 2026-03-19
Recompute custom metrics per scenario subset in model evaluation panel

Lucasmogsan:

When the App builds Scenario Analysis data, `get_subset_def_data` computes standard metrics (precision, recall, F1, IoU, etc.) by calling `results.metrics()` inside a `use_subset()` context manager. However, custom `EvaluationMetric `plugin aggregates were never recomputed for each scenario subset — the panel returned `results.custom_metrics` from the original full-dataset evaluation run, so the values never changed across scenarios.

This fix invokes `compute_custom_metrics()` against the subset view (`results.samples` within the context) inside `get_subset_def_data`, so each scenario subset gets its own custom metric aggregate. The original custom metrics are saved and restored after computation to avoid side effects.

It is tested with a custom plugin (with pixel-distance measure) and works with:

```python
dataset = fo.load_dataset("fiftyone_fix")
results = dataset.load_evaluation_results("test_eval")

# Define your scenario subsets (same format the panel uses)
scenario_def = {"type": "field", "field": "cam_type", "value": "iOS"}

with results.use_subset(scenario_def):
    # Standard metrics work out of the box
    print(results.metrics())

    # Recompute custom metrics for this subset
    results.custom_metrics = None
    results.backend.compute_custom_metrics(
        results.samples, results.key, results
    )
    print(results.custom_metrics)
```

NB: This is still not showing in the GUI app - a feature that I believe should be added?

        notes for FiftyOne users.

Custom `EvaluationMetric` plugin values are now recomputed per scenario subset in the Model Evaluation panel, instead of showing the same full-dataset aggregate for every subset.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Custom metrics are now computed and included in subset evaluation data within the model evaluation panel, providing metrics tailored to each specific subset.
  * Improved error resilience ensures metric computation failures gracefully degrade without disrupting evaluations.

### DIFF
--- a/plugins/panels/model_evaluation/__init__.py
+++ b/plugins/panels/model_evaluation/__init__.py
@@ -880,8 +880,22 @@ class EvaluationPanel(Panel):
                     metrics["incorrect"],
                 ) = self.get_correct_incorrect(results)
 
+            # Recompute custom metrics for this subset
+            original_custom_metrics = results.custom_metrics
+            try:
+                results.custom_metrics = None
+                results.backend.compute_custom_metrics(
+                    results.samples, results.key, results
+                )
+                custom_metrics = self.get_custom_metrics(results)
+            except Exception:
+                custom_metrics = None
+            finally:
+                results.custom_metrics = original_custom_metrics
+
             return {
                 "metrics": metrics,
+                "custom_metrics": custom_metrics,
                 "distribution": len(results.ytrue_ids),
                 "confusion_matrix": self.get_confusion_matrix(results),
                 "confidences": confidences,


Lucasmogsan:
@AdonaiVera is this something you intend to merge?

brimoor:
@Lucasmogsan thanks for identifying the issue in https://github.com/voxel51/fiftyone/issues/7128 and contributing a fix here! I'll give this a review and get it merged 😄

Lucasmogsan:
@brimoor will it be possible to merge this? Thanks

Lucasmogsan:
@brimoor not to be pushy but this would be very nice so I can get rid of the source code just for this small change :-)

PR #7224, opened 2026-03-21
[HA] ensure server-provided fields are preserved in delta calculation

tom-vx51:

This fixes a bug where server-provided data could be omitted on client-side labels, resulting in erroneous patch requests. Labels are now properly merged during delta calculations.

local, unit tests

Specific steps to repro:
* Create a new `Classification` field on a dataset
* Use the `edit_field_values` operator to init `<new_field>.label` to some value for all samples
* Edit the classification label in the annotate tab

Previous behavior:
* Persistence gets into a bad state due to missing local data

New behavior:
* Persistence works as expected

        notes for FiftyOne users.

Fixed a bug which could cause erroneous patch requests while annotating samples

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Delta/patch generation for annotations updated to preserve existing server-enriched fields during partial updates.

* **Tests**
  * Expanded delta calculation test suite with broader edge-case coverage for nested, empty and mixed updates.

* **Chores**
  * Reorganized imports and utilities; added a shared object-check helper and cleaned up client-side imports.

### DIFF
--- a/app/packages/annotation/src/deltas.ts
+++ b/app/packages/annotation/src/deltas.ts
@@ -13,7 +13,7 @@ import type {
   PrimitiveValue,
   Sample,
 } from "@fiftyone/state";
-import type { Field, Primitive } from "@fiftyone/utilities";
+import { Field, isObject, Primitive } from "@fiftyone/utilities";
 import { get } from "lodash";
 import type { OpType } from "./types";
 import { arePrimitivesEqual, isPrimitiveFieldType } from "./util";
@@ -318,11 +318,7 @@ export const buildSingleMutationDelta = <
   //  (`_cls`, `_id`, `tags`) are preserved when the overlay only has a
   //  subset of fields
   const merged =
-    typeof data === "object" &&
-    data !== null &&
-    !Array.isArray(data) &&
-    typeof existingLabel === "object" &&
-    !Array.isArray(existingLabel)
+    isObject(data) && isObject(existingLabel)
       ? { ...existingLabel, ...data }
       : data;
 
--- a/app/packages/annotation/src/deltas.ts
+++ b/app/packages/annotation/src/deltas.ts
@@ -1,19 +1,19 @@
-import { isDetection3d } from "@fiftyone/core";
-import { JSONDeltas } from "@fiftyone/core/src/client";
+import { isDetection3d } from "@fiftyone/core/src/utils/labels";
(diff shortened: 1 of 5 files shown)


app/packages/annotation/src/deltas.ts
```diff
@@ -305,15 +305,28 @@ export const buildDeletionDeltas = (
...
+export const buildSingleMutationDelta = <
   T extends AnnotationLabel["data"] | Primitive
 >(
   sample: Sample,
   path: string,
   data: T
 ): JSONDeltas => {
   const existingLabel = <T>extractNestedField(sample, path) ?? {};
-  return generateJsonPatch(existingLabel, data);
+
```
erik-nieh, line 316: There is an existing isObject that might be useful, but I think it does not explicitly check the null case: 

https://github.com/voxel51/fiftyone/blob/develop/app/packages/core/src/client/util.ts#L89

erik-nieh (dismissed):

app/packages/annotation/src/deltas.ts
```diff
@@ -305,15 +305,28 @@ export const buildDeletionDeltas = (
...
+  //  (`_cls`, `_id`, `tags`) are preserved when the overlay only has a
+  //  subset of fields
+  const merged =
+    typeof data === "object" &&
+    data !== null &&
+    !Array.isArray(data) &&
+    typeof existingLabel === "object" &&
+    !Array.isArray(existingLabel)
+      ? { ...existingLabel, ...data }
+      : data;
```
mozabes, line 323: Is there no way for us to type this better than a generic T?  Can we be more explicit with the typing to make it so we can not have to be so verbose with checking typeofs and isArrays?

lanzhenw (dismissed):
lgtm

app/packages/annotation/src/deltas.ts
tom-vx51: yeah it's currently scoped to 
```typescript
T extends AnnotationLabel["data"] | Primitive
```
and the wrench here is `Primitive`. The guards are probably more than necessary, but it's important that we don't corrupt data here, so I'd rather have some extra caution. 

Per Erik's point, we can at least hide it in a utility function

app/packages/annotation/src/deltas.ts
tom-vx51: updated, thanks!

lanzhenw (approved):

PR #7227, opened 2026-03-22
Fix numpy output serialization in KeypointOutputProcessor and embed_all

CharlesCNorton:
## Summary

- Wrap numpy scalar division in `float()` in `KeypointOutputProcessor._parse_output`
  (same fix as #6407 for the parent class)
- Handle `ModelOutput` return type from `get_image_features()` in
  `ZeroShotTransformerEmbeddingsMixin.embed_all`

`KeypointOutputProcessor._parse_output` divides numpy arrays by Python ints, producing
`np.float32` values that pymongo rejects. #6407 fixed the same issue in the child class
`KeypointDetectorOutputProcessor` but the parent class was missed. This crashes all vitpose
and pose-estimation-transformer models on `dataset.apply_model()`.

`ZeroShotTransformerEmbeddingsMixin.embed_all` calls
`self._model.get_image_features(**args).detach()`, assuming the return value is always a
tensor. For siglip, monet-zero, pubmed-clip, and zero-shot-classification-transformer,
`get_image_features()` returns a `BaseModelOutputWithPooling` which doesn't have `.detach()`.
The fix extracts `.pooler_output` when the return is not a tensor.

Reproduced all 11 affected failures with numpy 2.3.5. Applied both patches and verified
all 11 models pass. Ran a full 216-model zoo sweep with patches applied to confirm zero
regressions (203 pass vs 192 without patches — the 11 difference is exactly these fixes).

## Release Notes

- [x] Yes.

Fixed `InvalidDocument` error when running vitpose models and `AttributeError` when
calling `embed()` on siglip/clip-based models with numpy >= 2.0.

- [x] Core: Core fiftyone Python library changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure keypoint coordinates are produced as native Python floats for consistent numeric formatting.
  * Improve transformer embedding extraction to handle varied model output shapes and raise a clear error when expected features are missing.

* **Tests**
  * Added unit tests validating keypoint coordinate serialization as native floats.

### DIFF
--- a/fiftyone/utils/torch.py
+++ b/fiftyone/utils/torch.py
@@ -1617,7 +1617,7 @@ class KeypointOutputProcessor(OutputProcessor):
                     # Low confidence
                     points.append((float("nan"), float("nan")))
                 else:
-                    points.append((p[0] / width, p[1] / height))
+                    points.append((float(p[0] / width), float(p[1] / height)))
             _keypoints.append(
                 fol.Keypoint(
                     points=points,
--- a/fiftyone/utils/transformers.py
+++ b/fiftyone/utils/transformers.py
@@ -532,7 +532,12 @@ class ZeroShotTransformerEmbeddingsMixin(EmbeddingsMixin):
                 args[k] = v.to(self.device)
             features = self._model.get_image_features(**args)
             if not isinstance(features, torch.Tensor):
-                features = features.pooler_output
+                features = getattr(features, "pooler_output", None)
+                if features is None:
+                    raise ValueError(
+                        "get_image_features() returned a non-tensor "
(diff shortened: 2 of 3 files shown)


Burhan-Q (approved):
I'm a bit ignorant WRT the `.pooler_output` attribute, so I kind of side with the Code Rabbit comment as a solid idea to guard on, but I trust others know better than me here.

Burhan-Q:
Follow up I guess. Is there any unit test that could be included that would help prevent this from being missed again?

CharlesCNorton:
@Burhan-Q Done and done!

Burhan-Q (approved):
Image

PR #7230, opened 2026-03-23
improve background compute onboarding

imanjra:

improve background compute onboarding

Using an operator which required background compute

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved orchestrator setup requirement handling by removing generic warning alerts and replacing them with a dedicated setup flow.

* **Style**
  * Updated footer UI to display a "Set up now" link and documentation guidance when orchestrator setup is required, providing clearer guidance to users.

### DIFF
--- a/app/packages/operators/src/components/OperatorPromptFooter.tsx
+++ b/app/packages/operators/src/components/OperatorPromptFooter.tsx
@@ -1,10 +1,16 @@
 import { Button } from "@fiftyone/components";
-import { CircularProgress, Button as MUIButton } from "@mui/material";
+import { CircularProgress, Link, Stack } from "@mui/material";
+import {
+  IconName,
+  RichCard,
+  Variant,
+  Button as VoodoButton,
+} from "@voxel51/voodo";
 import { useCallback } from "react";
+import { SubmitButtonOption } from "../OperatorPalette";
 import SplitButton from "../SplitButton";
 import { BaseStylesProvider } from "../styled-components";
 import { onEnter } from "../utils";
-import { SubmitButtonOption } from "../OperatorPalette";
 
 export default function OperatorPromptFooter(props: OperatorFooterProps) {
   const {
@@ -18,7 +24,7 @@ export default function OperatorPromptFooter(props: OperatorFooterProps) {
     submitButtonOptions,
     submitOptionsLoading,
     hasSubmitButtonOptions,
-    showWarning,
+    requiresOrchestratorSetup,
   } = props;
 
   const handleSubmit = useCallback(() => {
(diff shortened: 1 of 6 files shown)


app/packages/operators/src/components/OperatorPromptFooter.tsx
```diff
@@ -18,70 +24,80 @@ export default function OperatorPromptFooter(props: OperatorFooterProps) {
...
-              disabled={disableSubmit}
-              disabledReason={disabledReason}
-              options={submitButtonOptions}
-              submitOnEnter
-              onSubmit={onSubmit}
-            />
-          </BaseStylesProvider>
-        )}
-      </>
+              <VoodoButton variant={Variant.Secondary}>Set up now</VoodoButton>
```
timmydoza, line 48: I agree with coderabbit here. Don't think we should nest a `<button>` in an `<a>` if we can help it. 

For now it's probably fine but maybe we make a ticket and come back to this soon. Something like
```jsx
<VoodoButton as="a" variant={Variant.Secondary}>Set up now</VoodoButton>
```
or
```jsx
<VoodoButton href="foo" variant={Variant.Secondary}>Set up now</VoodoButton>
```
could work. MUI uses the latter approach, where it automatically uses an anchor element when the `href` prop is present.

app/packages/operators/src/components/OperatorPromptFooter.tsx
```diff
@@ -18,70 +24,80 @@ export default function OperatorPromptFooter(props: OperatorFooterProps) {
...
+      {onCancel && (
+        <BaseStylesProvider>
+          <Button onClick={onCancel} onKeyDown={onEnter(onCancel)}>
+            {cancelButtonText}
+          </Button>
+        </BaseStylesProvider>
+      )}
+      {onSubmit && !hasSubmitButtonOptions && !submitOptionsLoading && (
+        <BaseStylesProvider>
+          <Button
```
timmydoza, line 78: Is this button inside of a `<form>` element? If so, `type="submit"` would be nice. It would handle `onKeyDown={onEnter(handleSubmit)}` automatically.

timmydoza (commented):
Just had some minor thoughts, but overall it looks good!

I'd like to run the code locally before I ✅ . What's the flow in OSS in order for me to see this UI? Additionally, would it possible to provide screenshots of new UI changes in the PR description?

app/packages/operators/src/components/OperatorPromptFooter.tsx
timmydoza: Nevermind - looks like the form is a sibling. Still, this could work if the form had an ID, but that's outside the scope of this PR.

timmydoza (approved):
> What's the flow in OSS in order for me to see this UI? Additionally, would it possible to provide screenshots of new UI changes in the PR description?

Never mind, I got it working! ✅

app/packages/operators/src/components/OperatorPromptFooter.tsx
imanjra: Yah, I agree. We can address this as a follow-up with voodo

PR #7240, opened 2026-03-25
Fix #2176: Docs update file_hash precision for large integers

### LINKED ISSUE #2176: [BUG] file_hash lost precision in the fiftyone app
### System information

-   **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:  Centos7
-   **FiftyOne version (run `fiftyone --version`)**: 0.16.6
-   **Python version**: 3.7.12

### Describe the problem

`file_hash` lost precision in the fiftyone app which show `1990983151808212700` but actually `1990983151808212765`. 

image

### Code to reproduce issue

```python
import fiftyone.zoo as foz

for sample in dataset.select_fields().iter_samples(progress=True, autosave=True, batch_size=60.0):
    sample["file_hash"] = fou.compute_filehash(sample.filepath)

view = dataset[:1]
sample = view.first()
print(sample.file_hash) # result in 1990983151808212765

with session.no_show():
    session.view = view #  but show out 1990983151808212700 in the App
```

### Other info / logs

### Willingness to contribute

FiftyOne codebase?

        from the FiftyOne community.

diogossilva5:
Fixes file_hash precision loss when the value exceeds JavaScript's safe integer limit (2**53 - 1). Previously, the backend serialized these integers as numbers and, once they reached the frontend, parsing could round the value, causing a mismatch between the backend-computed file_hash and the value shown in the app.
Before converting data to JSON, the fix checks numeric values and, when it finds an integer larger than JavaScript can represent precisely, it serializes that value as a string. This preserves the exact value while it travels between the backend and the app, without changing file_hash computation.
A deterministic regression test was also added with an out-of-range value to ensure the app-received value remains exactly equivalent to the backend value.

- Adds _coerce_unsafe_ints() which recursively traverses objects and converts integers outside the JS safe range into strings
- Ensures compatibility with nested structures (lists, tuples, dicts) and NumPy integer types
- Preserves safe integers, booleans, floats, None, and special values (NaN, ±inf) unchanged
- Adds a recursion depth guard via MAX_COERCE_RECURSION_DEPTH to prevent pathological inputs

- End-to-end regression test
- Nested structure handling (dicts, lists, tuples, NumPy integers, and primitive types)
- Boundary conditions (behavior at ±JS_MAX_SAFE_INT)
- Special values (None, booleans, floats, NaN, and infinities remain unchanged)
- Large-scale input
- Recursion safety (deeply nested inputs raise a ValueError)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Documentation**
  * Clarified default file-hash behavior in documentation: the default returns a Python integer derived from raw file bytes, which can exceed JavaScript’s safe integer range and be unsafe to compare or display; recommends using a stable hexadecimal digest (e.g., SHA-1) when a consistent displayable identifier is required.
  * Cleared notebook execution metadata and added the same precision warning as an inline comment.

### DIFF
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -2530,7 +2530,10 @@ def compute_filehash(filepath, method=None, chunk_size=None):
     Args:
         filepath: the path to the file
         method (None): an optional ``hashlib`` method to use. If not specified,
-            the builtin ``str.__hash__`` will be used
+            the builtin ``str.__hash__`` will be used, which returns an
+            integer. The default hashing method will display non-unique values in the app or
+            if accessing filehash values via Javascript. If this is problematic, you
+            should include the argument `method="sha1"` for unique string values.
         chunk_size (None): an optional chunk size to use to read the file, in
             bytes. Only applicable when a ``method`` is provided. The default
             is 64kB. If negative, the entire file is read at once
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -2527,19 +2527,19 @@ def get_module_name(path, start=None):
 def compute_filehash(filepath, method=None, chunk_size=None):
     """Computes the hash of the given file.
 
-    Args:
-        filepath: the path to the file
-        method (None): an optional ``hashlib`` method to use. If not specified,
-            the builtin ``str.__hash__`` will be used, which returns an
-            integer. The default hashing method will display non-unique values in the app or
-            if accessing filehash values via Javascript. If this is problematic, you
-            should include the argument `method="sha1"` for unique string values.
-        chunk_size (None): an optional chunk size to use to read the file, in
-            bytes. Only applicable when a ``method`` is provided. The default
-            is 64kB. If negative, the entire file is read at once
+        Args:
+            filepath: the path to the file
+            method (None): an optional ``hashlib`` method to use. If not specified,
+                Python's builtin ``hash()`` on the file bytes will be used, which
+    +           returns an integer. This value may be unsafe to compare/display via
+    +           JavaScript due to numeric precision limits. If this is problematic,
+    +           pass ``method="sha1"`` to get a stable hexadecimal string digest.
+            chunk_size (None): an optional chunk size to use to read the file, in
+                bytes. Only applicable when a ``method`` is provided. The default
+                is 64kB. If negative, the entire file is read at once
 
-    Returns:
-        the hash
+        Returns:
+            the hash
     """
     if method is None:
         with open(filepath, "rb") as f:
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -2538,8 +2538,8 @@ def compute_filehash(filepath, method=None, chunk_size=None):
             bytes. Only applicable when a ``method`` is provided. The default
             is 64kB. If negative, the entire file is read at once
 
-        Returns:
-            the hash
+    Returns:
+        the hash
     """
     if method is None:
         with open(filepath, "rb") as f:
--- a/fiftyone/core/utils.py
+++ b/fiftyone/core/utils.py
@@ -2527,16 +2527,16 @@ def get_module_name(path, start=None):
 def compute_filehash(filepath, method=None, chunk_size=None):
     """Computes the hash of the given file.
 
-        Args:
-            filepath: the path to the file
-            method (None): an optional ``hashlib`` method to use. If not specified,
-                Python's builtin ``hash()`` on the file bytes will be used, which
-    +           returns an integer. This value may be unsafe to compare/display via
-    +           JavaScript due to numeric precision limits. If this is problematic,
-    +           pass ``method="sha1"`` to get a stable hexadecimal string digest.
-            chunk_size (None): an optional chunk size to use to read the file, in
-                bytes. Only applicable when a ``method`` is provided. The default
-                is 64kB. If negative, the entire file is read at once
+    Args:
+        filepath: the path to the file
+        method (None): an optional ``hashlib`` method to use. If not specified,
+            Python's builtin ``hash()`` on the file bytes will be used, which
+            returns an integer. This value may be unsafe to compare/display via
+            JavaScript due to numeric precision limits. If this is problematic,
+            pass ``method="sha1"`` to get a stable hexadecimal string digest.
+        chunk_size (None): an optional chunk size to use to read the file, in
+            bytes. Only applicable when a ``method`` is provided. The default
+            is 64kB. If negative, the entire file is read at once
 
         Returns:
             the hash
--- a/docs/source/recipes/image_deduplication.ipynb
+++ b/docs/source/recipes/image_deduplication.ipynb
@@ -673,7 +673,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 15,
+   "execution_count": null,
    "metadata": {},
    "outputs": [
     {
@@ -697,6 +697,9 @@
    ],
    "source": [
     "import fiftyone.core.utils as fou\n",
+    "\n",
+    "# NOTE: The default method may return integers exceeding JavaScript's safe limit (2^53 - 1), causing precision loss in the app or JS.\n",
+    "# If this is problematic, use ``method=\"sha1\"`` for string-based hashes.\n",
     "\n",
     "for sample in dataset:\n",
     "    sample[\"file_hash\"] = fou.compute_filehash(sample.filepath)\n",


Burhan-Q (commented):
Thank you for the contribution! 🙌 I think that the proposed change has much larger implications, see the [Code Rabbit comment](https://github.com/voxel51/fiftyone/pull/7240#discussion_r2988005081). The `fiftyone.core.utils.compute_filehash` function accepts hash methods for the `hashlib` module, but defaults to the builtin Python `hash()` method. The builtin `hash()` is fast, but may introduce values that are beyond the JS unsafe integer limit, so using an alternative hash method would be the correct solution.

https://github.com/voxel51/fiftyone/blob/541f1b2e3062776ce443cabd28cd95ed5c4867d5/fiftyone/core/utils.py#L2527-L2540


The SHA1 hashing algorithm returns strings like "f7359ad3c5f03d12e3d8b89d397ea3d9260282a1" for each file. Which won't have issues with the JS safe integer limit. I believe that given the option for alternative hashing methods, that will generate a string hash, makes the need to enforce this integer limit unnecessary.

Since I expect the first argument to be about hashing speed, I'll address this here. The builtin `hash()` call is faster than using SHA1 with my quick test for the 200 images. It adds up on a much larger dataset, but it's a viable alternative if you're seeing filehash collisions due to the integer limit. FWIW, I tested using a 8600 image dataset and found that the builtin `hash()` method could process all images in ~0.473 seconds where the SHA1 hash took ~0.734 seconds. I think that the coercion of integers with the proposed change, given the various conditionals and recursion, would likely result in a higher computational time cost than swapping to SHA1 over the builtin `hash()` method. 

<details><summary>Ran speed test using the "quickstart" dataset images.</summary>
<p>

```python
from pathlib import Path
from timeit import repeat

import fiftyone as fo

img_path = Path(fo.config.dataset_zoo_dir) / "quickstart/data"  # already downloaded, 200 images
img_files = list(img_path.glob("*.jpg")

for method in ("sha1", None, "sha224", "sha256"):  # these were fastest in my testing
    result = repeat(
        lambda: [fou.compute_filehash(f, method) for f in img_files],
        repeat=5,
        number=100
    )
    avg = sum(result) / 5
    each = avg / 100
    print(
        f"{method or 'builtin hash()'}:",
        f"  Average 100 repeats = {avg:.4f} seconds.",
        f"  Per run average = {each:.4f} seconds for {len(img_files)} images.",
        sep="\n"
    )

>>> sha1:
>>>   Average 100 repeats = 0.9820 seconds.
>>>   Per run average = 0.0098 seconds for 200 images.
>>> builtin hash():
>>>   Average 100 repeats = 0.6357 seconds.
>>>   Per run average = 0.0064 seconds for 200 images.
>>> sha224:
>>>   Average 100 repeats = 0.9724 seconds.
>>>   Per run average = 0.0097 seconds for 200 images.
>>> sha256:
>>>   Average 100 repeats = 0.9745 seconds.
>>>   Per run average = 0.0097 seconds for 200 images.
```

</p>
</details>

diogossilva5:
Thanks for the detailed feedback! I investigated further and found that compute_filehash has no production call site in the core codebase — the file_hash field appears only in recipes/examples (e.g., docs/source/recipes/image_deduplication.ipynb), where users create it manually.

I agree that the global encoder change is too broad and could have unintended side effects on other numeric fields.

I'd like to adjust this PR to:

Revert the encoder changes
Update the deduplication recipe to use method="sha1" instead of the default hash()
Add a note to the compute_filehash docstring warning that method=None returns an integer that may exceed JavaScript's safe integer limit (2^53 - 1)
Does this approach make sense? Or is there an internal flow creating file_hash that I might have missed?

Burhan-Q:
@diogossilva5 I mostly agree with the direction you're thinking, but I'd also like to hear from others on the @voxel51/developers team as well.
> Revert the encoder changes
> Update the deduplication recipe to use method="sha1" instead of the default hash()
> Add a note to the compute_filehash docstring warning that method=None returns an integer that may exceed JavaScript's safe integer limit (2^53 - 1)
> Does this approach make sense? Or is there an internal flow creating file_hash that I might have missed?

* I definitely agree to revert the encoder changes.
* I don't think that the recipe needs to be changed to use `method="sha1"`, but I do think that including a note like: 
  ```
  The default hashing method will display non-unique values in the app or 
  if accessing filehash values via Javascript. If this is problematic, you 
  should include the argument `method="sha1"` for unique string values.
  ```
  would be useful. I suspect most users will not have any issues with the `hash()` method since all filehash value analysis is executed via Python and the "true" values are stored in MongoDB.
* I think the docstring addition makes sense, again I would say pointing out that it's only important for direct Javascript access or when visually comparing filehash values in the app, otherwise the default `hash()` method is usable in all other cases

Burhan-Q:
Here's example code from Claude to demonstrate the issue.

```python
import fiftyone as fo
from fiftyone.core.json import stringify

JS_MAX_SAFE_INTEGER = (1 << 53) - 1

def find_collision_pair():
      """Find two distinct Python integers beyond JS_MAX_SAFE_INTEGER that
      collide when truncated to float64 precision.
  
      Returns two distinct ints that map to the same JS number.
      """
      # Any two integers that differ only in the low bits beyond 53-bit
      # mantissa precision will collide. We construct such a pair directly.
      #
      # For a 64-bit int, the float64 representation loses the lowest bits.
      # Two integers that round to the same float64 are a collision pair.
      base = 2**60  # well beyond JS_MAX_SAFE_INTEGER
      # These two values differ by 1, but both map to the same float64
      a = base + 1
      b = base + 2
      # Walk until we find a pair that truly collides in float64
      while float(a) != float(b):
          a += 1
          b += 1
      assert a != b, "Values must be distinct Python ints"
      assert float(a) == float(b), "Values must collide in float64"
      return a, b

if __name__ == "__main__":
    hash_a, hash_b = find_collision_pair()
    
    print(f"File hash values are distinct: {hash_a != hash_b}")
    # output: File hash values are distinct: True
    
    # Create a minimal dataset
    dataset = fo.Dataset()

    # Create two samples with synthetic filehash values
    dataset.add_samples(
        [
            fo.Sample(filepath="/tmp/nonexistent_a.jpg"),
            fo.Sample(filepath="/tmp/nonexistent_b.jpg")
        ]
    )
    
    # Store filehash values as a custom field (as a user might do)
    dataset.set_values("filehash", [hash_a, hash_b])
    
    session.wait()
```

Filehash values rendered in the app are identical:

image

Both filehash values are now $1152921504606847000$

fiftyone/core/utils.py
```diff
@@ -2527,16 +2527,19 @@ def get_module_name(path, start=None):
...
+        Args:
+            filepath: the path to the file
+            method (None): an optional ``hashlib`` method to use. If not specified,
+                Python's builtin ``hash()`` on the file bytes will be used, which
+    +           returns an integer. This value may be unsafe to compare/display via
+    +           JavaScript due to numeric precision limits. If this is problematic,
+    +           pass ``method="sha1"`` to get a stable hexadecimal string digest.
+            chunk_size (None): an optional chunk size to use to read the file, in
+                bytes. Only applicable when a ``method`` is provided. The default
+                is 64kB. If negative, the entire file is read at once
```
Burhan-Q:
```suggestion
    Args:
        filepath: the path to the file
        method (None): an optional ``hashlib`` method to use. If not specified,
            Python's builtin ``hash()`` on the file bytes will be used, which
            returns an integer. This value may be unsafe to compare/display via
            JavaScript due to numeric precision limits. If this is problematic,
            pass ``method="sha1"`` to get a stable hexadecimal string digest.
        chunk_size (None): an optional chunk size to use to read the file, in
            bytes. Only applicable when a ``method`` is provided. The default
            is 64kB. If negative, the entire file is read at once
```

fiftyone/core/utils.py
```diff
@@ -2527,16 +2527,19 @@ def get_module_name(path, start=None):
...
 
-    Returns:
-        the hash
+        Returns:
+            the hash
```
Burhan-Q:
```suggestion
    Returns:
        the hash
```

Burhan-Q (changes requested):
Requesting some minor changes. Removed `+`'s and corrected indentation of docstring edits. Aside from that, the only other suggestion might be to update the "Compute file hashes" section in `docs/source/recipes/image_deduplication.ipynb`, but I would consider it optional.

Burhan-Q (approved):
Looks good! I'll approve, but I think it might require approval from one of the engineering teams

diogossilva5:
I've added a short note in the image deduplication notebook to mention the JavaScript precision limitation as well.

diogossilva5:
Hi all! This PR should be ready for final review.
It looks like it's currently waiting on a code owner review from the backend team, and the CI workflows may still need approval to run. Could someone from the backend squad take a look and, if needed, trigger the workflows?
Thanks a lot for your time!

diogossilva5:
The failing check is Build docs / build and it fails at checkout of voxel51/fiftyone-teams with Input required and not supplied: token. This appears to be a CI secret/permissions issue unrelated to this bugfix PR changes. Could a maintainer rerun or advise override?

Burhan-Q:
@diogossilva5 AFAIK this is expected to fail for external contributors. Still needs approval from code owners, who are often tied up with engineering priorities. I'll be certain to let the teams know, and appreciate your patience in the meantime.

brimoor (approved):

PR #7242, opened 2026-03-25
fix #1982: prevent duplicate frozen image output in Jupyter

### LINKED ISSUE #1982: [BUG] Freezing a plotly plot stores two copies of the image in the notebook output
When running the two cells below in a Jupyter notebook:

```py
import fiftyone as fo
import fiftyone.zoo as foz

locations = dataset.values("location.point.coordinates")
num_objects = dataset.values(F("ground_truth.detections").length())

plot = fo.location_scatterplot(locations=locations, sizes=num_objects)
plot.show()
```

```py
plot.freeze()
```

the intention is that `plot.freeze()` replaces the interactive plot with a screenshot, so that when the notebook is closed and reopened, the static image version of the plot will be there.

Currently, `plot.freeze()` appears to work in Jupyter notebooks (`jupyter notebook`), but, upon reopening the notebook, one sees that two copies of the image were actually stored:

Screen Shot 2022-08-04 at 9 49 06 PM

When running the same code in JupyterLab (`jupyter lab`), the duplicate images *immediately* appear in the cell output upon calling `plot.freeze()`.

The relevant code that will need to be tweaked in some way is here:
https://github.com/voxel51/fiftyone/blob/9ee5a680519f274c52a52816a7bd23c775a992f2/fiftyone/core/plots/plotly.py#L1842-L1885

As a side note, it would be preferable if `FigureWidget(...)` also *did not* appear in the cell output when the notebook is reopened. That is likely there because `self._widget.close()` makes the plot invisible in the active notebook, but doesn't actually delete it from the cell output.

jaimear2005:
Issue #1982 reports that `plot.freeze()` does not fully replace FiftyOne's interactive Plotly output in Jupyter. After `plot.show()` and `plot.freeze()`, the notebook ends up persisting two identical static images instead of a single screenshot. The fix ensures that the interactive widget and the frozen image are rendered through a single `display_id` handle, so `freeze()` updates the existing output rather than creating an additional one.

- Updated `_show()` so it now displays the Plotly widget directly with display_id in the same handle that freeze later updates.
- Removed the separate empty placeholder output and separate widget output path in `_show()`.
- Added a safety guard in `_screenshot()` so it returns cleanly if no handle exists.

On Jupyter Notebook or Jupyter Lab.
Before:

https://github.com/user-attachments/assets/9e5585d2-c237-4c51-8a1a-f72d8e73043a

After:

https://github.com/user-attachments/assets/7b6027fc-3cbe-477e-91d7-1c54bf85fc56

        notes for FiftyOne users.

Fixed an issue in Jupyter Notebook/Lab where freezing FiftyOne Plotly plots (`plot.freeze()`) could persist duplicate static images in the cell output.
Freeze now cleanly replaces the interactive widget with a single snapshot.

-   [ ] Other


Burhan-Q:
@jaimear2005 this looks like a clean fix. Thanks for opening a PR. Just a couple questions for clarification and understanding:
1. The videos you showed definitely helped, but I'd also like to understand if changes made to the dataset view and/or the plot (zoom, pan, etc.) are reflected in the image output by the "freeze" call. Have you tested this? If not, could you please give it a try with both before and after your changes? A video demo would be very helpful if you wouldn't mind.
2. I noticed on the linked issue:
> the intention is that plot.freeze() replaces the interactive plot with a screenshot, so that when the notebook is closed and reopened, the static image version of the plot will be there.

and I'm curious if you've tested the close/reopen persistence as well? A video here would also be insightful.

I'm not a big user of `IPython` or notebooks in general, so I don't have a lot of intuition on how these changes would impact things. Look forward to hearing back from you regarding these questions 🚀

jaimear2005:
@Burhan-Q thanks for your feedback and sorry for the delay to respond, I've tested what you suggested. First, I confirmed that changes made to the plot, such as pan and zoom, are reflected in the image output by the `freeze()` call, both after and before my changes.
Here is before:

https://github.com/user-attachments/assets/fd563776-0b03-4392-8164-f3aa9b4c4bc6

And after: 

https://github.com/user-attachments/assets/48fd64e5-332d-4b42-882b-4322f9dc4448

Then, a video confirming the close/reopen persistence:

https://github.com/user-attachments/assets/f1259015-58a1-4c52-af72-e193294d9cbf

Burhan-Q:
@jaimear2005 thank you for the updated videos! That looks like it checks all the boxes. I'm going to let the CI tests run. Would be great if we could hear from @brimoor but I'd give my 👍 on this change.

Burhan-Q (approved):

Burhan-Q:
Docs builds always fail for contributors. I'm going to run a build locally on this branch, just b/c I'm curious.

Also, if you weren't aware, we have a Discord server https://discord.gg/fiftyone-community Definitely consider joining 🚀

benjaminpkane (approved):

benjaminpkane:
Needs Backend approval. LGTM

brimoor (approved):
LGTM

PR #7251, opened 2026-03-27
[FOEPD-3394] Add Cross-Origin Isolation and CORS headers to server routes

yrahal:

Add Cross-Origin Isolation headers (COOP: same-origin, COEP: credentialless) to both the Python server and Vite dev server config, enabling SharedArrayBuffer for multi-threaded WASM inference.

Use credentialless instead of require-corp for COEP to avoid breaking cross-origin media loading.

Configure Vite to properly resolve runtime files with optimizeDeps.exclude for dev and a build plugin for prod.

Drtive-by: set weights URL to Voxel51-hosted.

- yarn build, check the fiftyone/server/static/assets/ dir for wasm files. Run fiftyone app launch and check segmentation works at localhost:5151 (with mutlithreading)
- python fiftyone/server/main.py with yarn dev and check segmentation works at localhost:5153 (with mutlithreading)
- Ephem deployment (multi threading needs a nextjs config change in the teams repo though)

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **Performance**
  * Worker thread allocation now scales to available CPU cores when cross-origin isolation is enabled, improving parallelism.

* **Configuration**
  * Model asset base URL switched to a CDN, updating model download sources.
  * WASM and model assets can be served from a configurable assets path in production; build emits fixed asset filenames and includes .onnx models as assets.
  * In non-development builds, a custom WASM path override is applied when specified.

* **Infrastructure**
  * Cross-origin isolation headers added to dev and production servers to enable secure features that rely on isolation.

### DIFF
--- a/app/packages/app/vite.config.ts
+++ b/app/packages/app/vite.config.ts
@@ -36,6 +36,8 @@ async function loadConfig() {
           },
           closeBundle() {
             const distAssets = path.resolve(__dirname, "dist/assets");
+            if (!fs.existsSync(distAssets))
+              return;
             const keep = new Set(ortWasmFiles);
             for (const f of fs.readdirSync(distAssets)) {
               if (f.includes("ort-wasm") && !keep.has(f)) {
--- a/app/packages/app/vite.config.ts
+++ b/app/packages/app/vite.config.ts
@@ -1,5 +1,5 @@
-import fs from "fs";
-import path from "path";
+import fs from "node:fs";
+import path from "node:path";
 import reactRefresh from "@vitejs/plugin-react-refresh";
 import nodePolyfills from "rollup-plugin-polyfill-node";
 import { defineConfig } from "vite";
--- a/app/packages/app/vite.config.ts
+++ b/app/packages/app/vite.config.ts
@@ -1,3 +1,5 @@
+import fs from "fs";
+import path from "path";
 import reactRefresh from "@vitejs/plugin-react-refresh";
 import nodePolyfills from "rollup-plugin-polyfill-node";
 import { defineConfig } from "vite";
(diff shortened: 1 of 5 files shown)


app/packages/app/vite.config.ts
```diff
@@ -20,7 +22,26 @@ async function loadConfig() {
...
+      {
+        name: "copy-ort-wasm",
+        apply: "build",
+        buildStart() {
+          const ortDist = path.dirname(require.resolve("onnxruntime-web"));
+          for (const f of ["ort-wasm-simd-threaded.jsep.wasm", "ort-wasm-simd-threaded.jsep.mjs"]) {
+            this.emitFile({ type: "asset", fileName: `assets/${f}`, source: fs.readFileSync(path.join(ortDist, f)) });
+          }
+        },
+      },
```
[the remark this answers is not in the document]
yrahal: Intentionally hardcoded. Dynamic discovery would emit all WASM variants when we only need the JSEP pair. If ort renames these files in a future version, fs.readFileSync throws and the build fails immediately, which is the correct behaviour since we'd need to review what changed.

app/packages/app/vite.config.ts
```diff
@@ -20,7 +22,39 @@ async function loadConfig() {
...
+              }
+            }
+          },
+        };
+      })(),
     ],
+    assetsInclude: ["**/*.onnx"],
+    define: {
+      "import.meta.env.ORT_WASM_PATH": JSON.stringify("/assets/"),
+    },
```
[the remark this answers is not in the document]
yrahal: The app is always served at root

fiftyone/server/app.py
```diff
@@ -97,6 +97,8 @@ async def dispatch(
     ) -> Response:
         response = await call_next(request)
         response.headers["x-colab-notebook-cache-control"] = "no-cache"
+        response.headers["Cross-Origin-Opener-Policy"] = "same-origin"
+        response.headers["Cross-Origin-Embedder-Policy"] = "credentialless"
```
lanzhenw: This would apply to all routes. It would be better if we can limit the range of where we apply the COOP and COEP headers. 

I am not an expert on CORP. Claude suggests that COOP requirement will break Jupyter notebooks and google colab iframes, as 
```
Jupyter server (localhost:8888)
  └── <iframe src="http://localhost:5151">   ← FiftyOne App
```

yrahal: Great catch, thank you @lanzhenw . I don't think we can scope the headers to specific routes or workers.  since they are a page-level property set by the browser on the top-level document response and the workers inherit it from their parent page.  I added a condition to detect if we are in a notebook context so the headers are skipped and the app loads normally in the iframe (in which case SAM2 falls back to single-threaded but still works).

tom-vx51 (approved):
looks good to me locally. Probably worth deploying to an ephem to validate the build changes against the prod config

lanzhenw (approved):

yrahal:
> looks good to me locally. Probably worth deploying to an ephem to validate the build changes against the prod config

Working in an ephem. Multi threading is disabled because I'd need to change the nextjs config on the teams repo, so it works but slower until we add that too.

I'll merge this and open a PR for the above.

fiftyone/server/app.py
lanzhenw: The fix is neat

PR #7252, opened 2026-03-28
Fix CVAT text attributes coerced to int on annotation import

joaquinhuigomez:
Fixes #7221

`_parse_value()` aggressively coerces string values to numeric types. When a CVAT text attribute contains all-digit strings like `"0001"`, it gets cast to `int(1)`, losing leading zeros and semantic meaning.

Added an `attr_type` parameter to `_parse_value()` so callers can indicate the CVAT attribute type. When the source type is `"text"`, numeric coercion is skipped and the original string is preserved.

The attribute type information is now threaded from the CVAT task labels (which carry `input_type` on each attribute spec) through the annotation parsing pipeline via a new `attr_type_map` dict. All existing callers that don't provide type info continue to work as before — the parameter defaults to `None`, preserving backward compatibility.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CVAT annotation import: text attributes remain text (empty strings → null), numeric and boolean values keep correct types, and tag/classification attributes use their declared types for more accurate imports.

* **New Features**
  * Added a configuration option to preserve legacy text-coercion behavior when needed.

* **Tests**
  * Updated tests to align with the revised annotation parsing behavior.

### DIFF
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -3366,9 +3366,7 @@ class CVATBackend(foua.AnnotationBackend):
     def download_annotations(self, results):
         api = self.connect_to_api()
 
-        coerce_text_attrs = getattr(
-            results.config, "coerce_text_attrs", True
-        )
+        coerce_text_attrs = getattr(results.config, "coerce_text_attrs", True)
 
         logger.info("Downloading labels from CVAT...")
         annotations = api.download_annotations(
@@ -4733,9 +4731,11 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
                     frame_stop -= offset
 
                 # Download task data
-                attr_id_map, attr_type_map, _class_map_rev = (
-                    self._get_attr_class_maps(task_id)
-                )
+                (
+                    attr_id_map,
+                    attr_type_map,
+                    _class_map_rev,
+                ) = self._get_attr_class_maps(task_id)
 
                 if coerce_text_attrs:
                     attr_type_map = None
@@ -4918,8 +4918,7 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
                 i["name"]: i["id"] for i in label["attributes"]
             }
             attr_type_map[label["id"]] = {
-                i["id"]: i.get("input_type", None)
-                for i in label["attributes"]
+                i["id"]: i.get("input_type", None) for i in label["attributes"]
             }
 
         # AL: not sure why we didn't just reverse keys/vals initially
@@ -6069,9 +6068,7 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
                     attr_ind = 0
                     while label is None and attr_ind < num_attrs:
                         attr = anno["attributes"][attr_ind]
-                        attr_type = _attr_types.get(
-                            attr["spec_id"], None
-                        )
+                        attr_type = _attr_types.get(attr["spec_id"], None)
                         label = _parse_value(
                             attr["value"], attr_type=attr_type
                         )
@@ -6094,7 +6091,10 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
             else:
                 label_type = "classifications"
                 cvat_tag = CVATTag(
-                    anno, class_map, attr_id_map, server_id_map,
+                    anno,
+                    class_map,
+                    attr_id_map,
+                    server_id_map,
                     attr_type_map=attr_type_map,
                 )
                 label = cvat_tag.to_classification()
@@ -7136,9 +7136,7 @@ class CVATLabel(object):
 
         # Parse attributes
         attr_id_map_rev = {v: k for k, v in attr_id_map[cvat_id].items()}
-        _attr_types = (
-            attr_type_map.get(cvat_id, {}) if attr_type_map else {}
-        )
+        _attr_types = attr_type_map.get(cvat_id, {}) if attr_type_map else {}
         for attr in attrs:
             name = attr_id_map_rev[attr["spec_id"]]
             attr_type = _attr_types.get(attr["spec_id"], None)
--- a/fiftyone/utils/cvat.py
+++ b/fiftyone/utils/cvat.py
@@ -3358,11 +3358,13 @@ class CVATBackend(foua.AnnotationBackend):
 
         return results
 
-    def download_annotations(self, results):
+    def download_annotations(self, results, coerce_text_attrs=False):
         api = self.connect_to_api()
 
         logger.info("Downloading labels from CVAT...")
-        annotations = api.download_annotations(results)
+        annotations = api.download_annotations(
+            results, coerce_text_attrs=coerce_text_attrs
+        )
         logger.info("Download complete")
 
         return annotations
@@ -4644,12 +4646,15 @@ class CVATAnnotationAPI(foua.AnnotationAPI):
 
         return results
 
-    def download_annotations(self, results):
+    def download_annotations(self, results, coerce_text_attrs=False):
         """Download the annotations from the CVAT server for the given results
         instance and parses them into the appropriate FiftyOne types.
 
         Args:
             results: a :class:`CVATAnnotationResults`
+            coerce_text_attrs (False): whether to coerce text attributes to
+                numeric types. By default, text attribute values are preserved
+                as strings
 
         Returns:
             the annotations dict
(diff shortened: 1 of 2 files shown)


joaquinhuigomez:
Addressed both CodeRabbit findings:

1. **Early return for text type in `_parse_value`** — changed from `if attr_type != "text"` guard to `if attr_type == "text": return str(value)` so text attributes skip boolean/None normalization entirely.

2. **Thread `attr_type` through scalar-tag branch** — the tag annotation path at line ~6049 now looks up the attribute's input type from `attr_type_map` before calling `_parse_value`, matching the pattern already used in `CVATShape`.

joaquinhuigomez:
Fixed the empty-string text scalar issue — `_parse_value` now returns `None` for `""` when `attr_type == "text"`, consistent with the existing behavior for non-text types (line 7714).

Regarding the XML importer path (`load_cvat_image_annotations` / `load_cvat_video_annotations`): that's a separate code path from the REST API download flow this PR targets. The XML format doesn't carry `input_type` metadata the same way the REST API does, so threading it there would require a different approach and broader changes. I'd prefer to keep this PR scoped to the REST download path that #7221 reported against, and handle the XML path in a follow-up if needed.

Burhan-Q (commented):
Great work identifying and fixing this issue. The root cause analysis is spot on, and the fix in `_parse_value()` itself is reasonable.

I had one design concern I wanted to raise: the `attr_type_map` is currently threaded as a parameter through

`download_annotations` → `_parse_shapes_tags` → `_parse_annotation` → `CVATShape`/`CVATTag` → `CVATLabel.__init__` → `_parse_value`
 
which adds up to about 20 places it's needed in the code, spanning 8 functions. Most of these are pure pass-thru `_parse_shapes_tags` and `_parse_annotation` don't use the map directly, they just relay it.

Since `_parse_shapes_tags`, `_parse_annotation`, and `download_annotations` are all methods on `CVATAnnotationAPI`, I think it might be simpler to store the map as instance state (`self._attr_type_map`) on the API object when it's built per-task in `download_annotations`. That way, `_parse_annotation` can read `self._attr_type_map` directly and only pass it to the `CVATShape`, `CVATTag`, and `CVATLabel` constructors that actually need it. This would eliminate the parameter from the signatures of `_parse_shapes_tags` and `_parse_annotation` entirely, cutting down the number of areas impacted by the proposed changes.

I also had a concern about backward compatibility. There may be users who have pipelines where text attributes containing text values like "42" or "true" are intentionally coerced to `int` or `bool` values by the current behavior (see [Hyrum's Law](https://www.hyrumslaw.com/)). Making text preservation automatic would silently change their results. 
One way I can think to handle this would be to gate the behavior behind an opt-in config parameter on `CVATBackendConfig`, something like ~~`preserve_text_values=False`~~ (see note below). When enabled, `download_annotations` would populate  `self._attr_type_map` from the CVAT task labels; when disabled (the default), `self._attr_type_map` stays `None` and existing coercion behavior is preserved. The `Config` base class handles serialization automatically for public attributes, so it would round-trip through `annotate()` and `load_annotations()` without any additional plumbing. This would need to be updated in the documentation as well, but would permit users to opt in with:

```py
dataset.annotate(anno_key, ..., preserve_text_values=True)  # see preferred argument name below
```

One final note regarding the changes on this PR. Including unit tests to cover the new, and old, behaviors will be immensely helpful. Adding or updating tests will ensure that changes, if merged, will have safeguards to ensure the new behavior going forward.

To summarize, the two suggestions are:
1. Store `attr_type_map` on `CVATAnnotationAPI` to cut the intermediate pass-through plumbing
2. Default to existing behavior, let users enable text preservation explicitly, likely needs documentation updates for CVAT
3. Include unit tests to cover all proposed changes, (for either the current proposed changes or based on my recommendations)

Happy to discuss if I'm missing something about why the current approach is preferable. I appreciate the thorough work tracing the type info through the pipeline. Ultimately merging the changes is not up to me, but I'm providing feedback with what I think could give the best odds of the changes being accepted.

---

### [edit]

> [!IMPORTANT]
> Using opt-in arg name `coerce_text_attrs` is preferred over the suggested `preserve_text_values`. The design eeps the same behavior, `coerce_text_attrs=False` as the default which uses the same behavior as today (no coercing), `coerce_text_attrs=True` to enforce coercing CVAT `#text` attributes as string values.

brimoor:
@Burhan-Q FYI I added some thoughts to the motivating issue [here](https://github.com/voxel51/fiftyone/issues/7221#issuecomment-4170857365).

TLDR, I think this PR makes sense. On nomenclature, I have a slight preference for `coerce_text_attrs=False`, but `preserve_text_attrs=True` is okay too if you prefer that.

On these points:

> 1. Store attr_type_map on CVATAnnotationAPI to cut the intermediate pass-through plumbing
> 2. Default to existing behavior, let users enable text preservation explicitly, likely needs documentation updates for CVAT
> 3. Include unit tests to cover all proposed changes, (for either the current proposed changes or based on my recommendations)

my personal opinion is:

1. :x: Not necessary to merge
2. ✅ Definitely necessary! 
3. :x: Not necessary to merge, but if tests are added, the place to do that is [here](https://github.com/voxel51/fiftyone/blob/develop/tests/intensive/cvat_tests.py)

Burhan-Q:
I agree with argument name `coerce_text_attrs=False` for the default behavior, with `coerce_text_attrs=True` as the new opt-in behavior. I've appended my original comment to reflect my agreement.

On the point of adding `attr_type_map` to `CVATAnnotationAPI`, it was a recommendation to simplify the changes. Sounds like that's not necessary, but I do still think it would be a nice addition personally, but I won't block on it.

@joaquinhuigomez please make the appropriate updates when you're able to. If you need some time, that's understandable. Please leave a comment or react with 👍 on this comment to acknowledge you're able to make updates soon, otherwise I'll assume you're busy if there's no response after a few days. You could also ping me on the `#github-contribution` channel on the Discord if that's easier https://discord.gg/fiftyone-community

joaquinhuigomez:
On it — will update with coerce_text_attrs=False as default.

joaquinhuigomez:
Added `coerce_text_attrs=False` to `download_annotations`. When False (default), text attributes are preserved as strings. Set to True for the old coercion behavior.

Burhan-Q:
@joaquinhuigomez thanks for making the update. I'd like to approve the tests to run, but could you run `pre-commit` and push any changes? I suspect that there might be some formatting updates

Burhan-Q:
I noticed that my understanding of the `coerce_text_attrs` argument name was **incorrect** (and I missed the detail in [this comment](https://github.com/voxel51/fiftyone/issues/7221#issuecomment-4170857365) that explicitly mentioned what the default setting/behavior needs to be). The default value _actually_ needs to be `coerce_text_attrs=True` to maintain the current behavior and then using `coerce_text_attrs=False` would be the _new_ design, which would keep all `#text` attributes as strings.

Additionally, in the [issue comment](https://github.com/voxel51/fiftyone/issues/7221#issuecomment-4170857365), a requirement I didn't transfer was:

> ...`coerce_text_attrs` being added to `CVATBackendConfig`

@joaquinhuigomez hopefully this task list will help:

- [ ] Add `coerce_text_attrs=True` to `CVATBackendConfig`: making the flag accessible via `dataset.annotate(...)` and `fouc.import_annotations(...)`
- [ ] Read the flag from config in `CVATBackend.download_annotations`: connects config to the download path; removes the orphaned function parameter
- [ ] Correct the default to `True` on `CVATAnnotationAPI.download_annotations`: currently `False` (this was my fault)
- [ ] Fix test callers of `_get_attr_class_maps`: 2 call sites in `tests/intensive/cvat_tests.py` that will crash from the 2→3 tuple change; [L60](https://github.com/voxel51/fiftyone/blob/01d632500208305a8f8dcbb6b120a440964a4ad6/tests/intensive/cvat_tests.py#L60) & [L189](https://github.com/voxel51/fiftyone/blob/01d632500208305a8f8dcbb6b120a440964a4ad6/tests/intensive/cvat_tests.py#L189) (on the `develop` branch)
- [ ] Document in `import_annotations` docstring: so users discover the kwarg
- [ ] Run `pre-commit` on all changes to apply formatting
- [ ] (OPTIONAL) Add new unit tests: for `_parse_value` with and without `attr_type="text"`

joaquinhuigomez:
Updated per the task list:
- Default flipped to `coerce_text_attrs=True`
- Added to `CVATBackendConfig` so it flows through `dataset.annotate()` and `import_annotations()`
- `CVATBackend.download_annotations` reads from config instead of taking a parameter
- Fixed the 2 test call sites for the tuple change
- Still need to run pre-commit (don't have the hooks installed locally — will sort that out)

Burhan-Q (commented):
Thanks for the updates @joaquinhuigomez let me know if you'd like me to run pre-commit and push the changes. For now I'll run the tests (some are expected to fail). Otherwise the changes are looking good!

Burhan-Q (approved):
I think this has met the necessary requirements, nice work @joaquinhuigomez

PR #7274, opened 2026-03-31
Ontology odm

erik-nieh:
## 📋 What changes are proposed in this pull request?

This pull request adds a new top-level document to support the ontology proposal. Please see 3543 for full details on this proposal and the implementations it will contain.

- new ODM
- auto managed created at and last modified at
- unit tests that also contain some of the taxonomy and conditional attribute document structure.

## 🧪 How is this patch tested? If it is not, please explain why.

- new unit tests around this document
- working in a Python shell to import and create documents in local database.

        notes for FiftyOne users.

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * Added ontology persistence for storing and managing taxonomies and annotation ontologies with name/version uniqueness, automatic created/last-modified timestamps, slug generation/sanitization, and version-aware retrieval.
  * Validation enforces allowed ontology types and rejects invalid legacy or malformed values.

* **Tests**
  * Added comprehensive tests covering creation, versioning, uniqueness (case-insensitive), slug behavior, validation, serialization round-trips, timestamps, querying by type, and deletion.

### DIFF
--- a/fiftyone/core/odm/ontology.py
+++ b/fiftyone/core/odm/ontology.py
@@ -7,6 +7,7 @@ Ontology documents.
 """
 
 from enum import Enum
+from datetime import datetime
 
 from mongoengine.fields import DynamicField
 
@@ -56,3 +57,9 @@ class OntologyDocument(Document):
     root = DynamicField()
     created_at = DateTimeField()
     last_modified_at = DateTimeField()
+
+    def save(self, *args, **kwargs):
+        if not self.in_db and self.created_at is None:
+            self.created_at = datetime.utcnow()
+
+        return super().save(*args, **kwargs)
--- a/fiftyone/core/odm/ontology.py
+++ b/fiftyone/core/odm/ontology.py
@@ -66,7 +66,11 @@ class OntologyDocument(Document):
     last_modified_at = DateTimeField()
 
     def save(self, *args, **kwargs):
+        now = datetime.now(timezone.utc)
+
         if not self.in_db and self.created_at is None:
-            self.created_at = datetime.now(timezone.utc)
+            self.created_at = now
+
+        self.last_modified_at = now
 
         return super().save(*args, **kwargs)
--- a/fiftyone/core/odm/ontology.py
+++ b/fiftyone/core/odm/ontology.py
@@ -7,7 +7,7 @@ Ontology documents.
 """
 
 from enum import Enum
-from datetime import datetime
+from datetime import datetime, timezone
 
 from mongoengine.fields import DynamicField
 
@@ -60,6 +60,6 @@ class OntologyDocument(Document):
 
     def save(self, *args, **kwargs):
         if not self.in_db and self.created_at is None:
-            self.created_at = datetime.utcnow()
+            self.created_at = datetime.now(timezone.utc)
 
         return super().save(*args, **kwargs)
--- a/fiftyone/core/odm/ontology.py
+++ b/fiftyone/core/odm/ontology.py
@@ -20,10 +20,18 @@ from .document import Document
 
 
 class OntologyType(str, Enum):  # TODO - update to StrEnum
-    """Allowed values for :attr:`OntologyDocument.type` in storage and APIs."""
+    """Allowed values for :attr:`OntologyDocument.type` in storage and APIs.
+
+    ``TAXONOMY`` — a reusable hierarchical class tree. Multiple annotation
+    ontologies can reference the same taxonomy by name.
+
+    ``ANNOTATION_ONTOLOGY`` — a container that defines classes, attributes
+    (with optional conditional display logic), and references to taxonomies.
+    This is the document that gets connected to a label schema on a field.
+    """
 
     TAXONOMY = "taxonomy"
-    CONDITIONAL_ATTRIBUTES = "conditional_attributes"
+    ANNOTATION_ONTOLOGY = "annotation_ontology"
 
 
 _ONTOLOGY_TYPE_VALUES = tuple(t.value for t in OntologyType)
--- a/fiftyone/core/odm/ontology.py
+++ b/fiftyone/core/odm/ontology.py
@@ -1,0 +1,58 @@
+"""
+Ontology documents.
+
+| Copyright 2017-2026, Voxel51, Inc.
+| `voxel51.com <https://voxel51.com/>`_
+|
+"""
+
+from enum import Enum
+
+from mongoengine.fields import DynamicField
+
+from fiftyone.core.fields import (
+    DateTimeField,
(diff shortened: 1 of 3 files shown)


fiftyone/core/odm/ontology.py
```diff
@@ -0,0 +1,65 @@
...
+from fiftyone.core.fields import (
+    DateTimeField,
+    IntField,
+    StringField,
+)
+
+from .document import Document
+
+
+class OntologyType(str, Enum):  # TODO - update to StrEnum
```
Burhan-Q, line 22: Any reason not to update now? It looks like it would be a simple swap to include in this PR, but I'm guessing I'm missing something

fiftyone/core/odm/ontology.py
```diff
@@ -0,0 +1,65 @@
...
+    version = IntField(required=True, default=1)
+    type = StringField(required=True, choices=_ONTOLOGY_TYPE_VALUES)
+    description = StringField(default=None)
+    root = DynamicField()
+    created_at = DateTimeField()
+    last_modified_at = DateTimeField()
+
+    def save(self, *args, **kwargs):
+        if not self.in_db and self.created_at is None:
+            self.created_at = datetime.utcnow()
```
Burhan-Q: For ease of future updates, you might consider using the [Python docs recommended pattern](https://docs.python.org/3.9/library/datetime.html#datetime.datetime.utcnow):

```python
from datetime import datetime, timezone
# from datetime import datetime, UTC

# in Python >=3.11, UTC is a constant that can be imported
UTC = timezone.utc

datetime.now(UTC)
```

In Python 3.12, `datetime.utcnow()` is [deprecated](https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcnow) (it still works, so it won't break anything, but shows a deprecation warning).

Burhan-Q (commented):
Just a few early comments to share for consideration

fiftyone/core/odm/ontology.py
erik-nieh: not supported in Python 3.9. I think it got added in 3.11

fiftyone/core/odm/ontology.py
```diff
@@ -0,0 +1,65 @@
...
+            },
+            "type",
+        ],
+    }
+
+    name = StringField(required=True)
+    version = IntField(required=True, default=1)
+    type = StringField(required=True, choices=_ONTOLOGY_TYPE_VALUES)
+    description = StringField(default=None)
+    root = DynamicField()
```
kaixi-wang: I don't think we use DynamicField() or direct mongoengine types anywhere in the sdk. We typically stick to fiftyone.core.fields types. Does DictField() work or is there something about DynamicField in particular?

erik-nieh: Yep, that will work fine

fiftyone/core/odm/ontology.py
```diff
@@ -0,0 +1,64 @@
...
+    description = StringField(default=None)
+    root = DictField()
+    created_at = DateTimeField()
+    last_modified_at = DateTimeField()
+
+    def save(self, *args, **kwargs):
+        if not self.in_db and self.created_at is None:
+            self.created_at = datetime.now(timezone.utc)
+
+        return super().save(*args, **kwargs)
```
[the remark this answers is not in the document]
Burhan-Q: `datetime.utcnow` is deprecated in Python 3.12, but it's also timezone aware (UTC). Using `datetime.utcnow` will output the same result as `datetime.now(timezone.utc)` which is the recommended pattern.

fiftyone/core/odm/ontology.py
```diff
@@ -0,0 +1,76 @@
...
+    Multiple datasets can reference the same ontology by name. Each
+    save creates a new versioned document; ``(name, version)`` is unique.
+    """
+
+    meta = {
+        "collection": "ontologies",
+        "strict": False,
+        "indexes": [
+            {
+                "fields": ["name", "version"],
```
kaixi-wang, line 53: should we sanitize the name and/or make it case insensitive?

fiftyone/core/odm/ontology.py
```diff
@@ -0,0 +1,76 @@
...
+                "unique": True,
+            },
+            "type",
+        ],
+    }
+
+    name = StringField(required=True)
+    version = IntField(required=True, default=1)
+    type = StringField(required=True, choices=_ONTOLOGY_TYPE_VALUES)
```
kaixi-wang, line 62: is `choices` actually used to enforce the type anywhere? Do you need to add an explicit check in save that type in _ONTOLOGY_TYPE_VALUES?

fiftyone/core/odm/ontology.py
```diff
@@ -0,0 +1,76 @@
...
+    description = StringField(default=None)
+    root = DictField()
+    created_at = DateTimeField()
+    last_modified_at = DateTimeField()
+
+    def save(self, *args, **kwargs):
+        now = datetime.now(timezone.utc)
```
kaixi-wang, line 69: I think this might already be handled since this is a subclass of Document (https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/odm/document.py#L788)

erik-nieh: The issue I found with this is it only acts on updates, not inserts. I'd prefer if this field was always populated even though the created_at documents this It's just nicer to be able to treat this as non null throughout the entire codebase

fiftyone/core/odm/ontology.py
erik-nieh: Yes Mongoengine validates this.[ I added a test that covers this case. ](https://github.com/voxel51/fiftyone/blob/ontology-odm/tests/unittests/ontology_tests.py#L159-L167)

fiftyone/core/odm/ontology.py
erik-nieh: i found a pattern in a different document that uses slugs so i added it to this and updated the index and using this constraints. i will update the other prs to query this on this instead of just name

fiftyone/core/odm/ontology.py
kaixi-wang: oh cool TIL

fiftyone/core/odm/ontology.py
kaixi-wang: i wonder if the sdk doesn't do it on inserts for some optimization to check if something was every updated? 🤷‍♀️ 
At least here since you are setting both created_at and last_updated_at with now, we can still derive that.

kaixi-wang (approved):
lgtm. thanks for addressing comments

PR #7275, opened 2026-03-31
Add 'once' convenience method to EventDispatcher

benjaminpkane:
## 📋 What changes are proposed in this pull request?

Add a convenience method for listening to an event once and only once

## 🧪 How is this patch tested? If it is not, please explain why.

Unit tests

-   [ ] Other

## Summary by CodeRabbit

* **New Features**
  * One-time event handlers: register via an options parameter to fire once and self-unregister; can be re-registered later.

* **Bug Fixes**
  * Unregistering before or after firing is safe; thrown errors in handlers are logged without breaking dispatch; async handlers still fire only once.

* **Documentation**
  * Hook examples updated to show one-time handler usage.

* **Tests**
  * Added comprehensive tests covering once behavior, payloads, unregistering, re-registration, async and error cases.

### DIFF
--- a/app/packages/events/src/dispatch/dispatcher.ts
+++ b/app/packages/events/src/dispatch/dispatcher.ts
@@ -1,4 +1,4 @@
-import { EventGroup, EventHandler } from "../types";
+import type { EventGroup, EventHandler } from "../types";
 
 type DispatchData<T> = T extends undefined | null ? [data?: T] : [data: T];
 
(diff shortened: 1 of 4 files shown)


app/packages/events/src/dispatch/dispatcher.ts
```diff
@@ -166,10 +201,11 @@ export class EventDispatcher<T extends EventGroup> {
     ...args: DispatchData<T[E]>
   ): void {
     const data = args[0] as T[E];
-    const typeHandlers = this.handlers[event];
-    if (!typeHandlers || typeHandlers.length === 0) {
+    if (!this.handlers[event]?.length) {
       return;
     }
+    // Make a copy of the handlers at the time of dispatch
+    const typeHandlers = [...this.handlers[event]];
```
benjaminpkane, line 224: Claude helped write most of this PR. The tests it wrote revealed a race condition with dispatching events to all handlers at the time of the request. Specifically, in `"should not affect other handlers registered with on"`, the regular `on` handler was only called once. Copying the array here fixed the issue

mcdoh (approved):

mozabes (approved):

app/packages/events/src/dispatch/dispatcher.ts
tom-vx51: yeah this works, I think our handler counts will be small enough that the copy isn't a concern, but we should re-evaluate if we start emitting a lot of high-frequency events.

The cause of the issue is that `once` calls `this.off` explicitly, which modifies the handler array in-place _while it's being iterated over_ in `dispatch`. 

This generally results in undefined behavior, and languages like Java don't allow it at all, instead raising a [ConcurrentModificationException](https://docs.oracle.com/javase/8/docs/api/java/util/ConcurrentModificationException.html). 

Making a copy of the array allows `dispatch` to iterate over a _different_ array and avoids the issue. If we need zero-copy behavior in the future, we can instead keep a record of items that should be removed and modify the handler array once `dispatch` is complete

tom-vx51 (approved):
nice

